Issue #515, opened 2023-08-25
[BETA]  mojo matmul.mojo 


guidorice:
### Bug description

The matmul.mojo example program does not return the expected output, and prints 

> An error occurred in Python.

### Steps to reproduce

```
@guidorice ➜ /workspaces/mojo/examples (main) $ mojo matmul.mojo 
Throughput of a 128x128 matrix multiplication in Python: 
An error occurred in Python.
Throughput of a 512x512 matrix multiplication in Mojo using a naive algorithm: 
5.486891495394973 GFLOP/s <> -9223372036854775808 x speedup over Python
Throughput of a 512x512 matrix multiplication in Mojo using vectorization: 
29.406540250582058 GFLOP/s <> -9223372036854775808 x speedup over Python
Throughput of a 512x512 matrix multiplication in Mojo using the stdlib `vectorize`: 
24.141552979977298 GFLOP/s <> -9223372036854775808 x speedup over Python
Throughput of a 512x512 {vectorized + parallelized} matrix multiplication in Mojo: 
2.7173885207260922 GFLOP/s <> -9223372036854775808 x speedup over Python
Throughput of a 512x512 {tiled + vectorized + parallelized} matrix multiplication in Mojo: 
2.9034890049746509 GFLOP/s <> -9223372036854775808 x speedup over Python
Throughput of a 512x512 {tiled + unrolled + vectorized + parallelized} matrix multiplication in Mojo: 
2.8672076606454677 GFLOP/s <> -9223372036854775808 x speedup over Python
```

PS first impression is it's looking really awesome. Very responsive, lightweight binaries, great CLI interface, repl, help msg. 👍🏽

### System information

- What OS did you do install modular CLI/mojo on ?

On a github workspace, 32GB instance.

```
@guidorice ➜ /workspaces/mojo/examples (main) $ mojo deviceinfo.mojo 
System information: 
    OS          :  linux
    CPU         :  znver3
    Arch        :  x86_64-unknown-linux-gnu
    Num Cores   :  8
    CPU Features:  avx2
```

- Provide version information for Mojo by pasting the output of `mojo -v`

mojo 0.2.0 (4c0ef274)

- Provide Modular CLI version by pasting the output of `modular -v`

modular 0.1.1

shashankprasanna:
Hi @guidorice thank you for reporting!
Unfortunately we've hardcoded alias test_dir = "/root/mojo-examples"
[...]

Issue #514, opened 2023-08-25
[BETA]: mojo hello_interop.mojo 


guidorice:
### Bug description

The hello_interop example program does not return the expected output, and the exit code is not an error:

## reproduce steps

```
@guidorice ➜ /workspaces/mojo/examples (main) $ mojo hello_interop.mojo 
Hello Mojo 🔥!
9
6
3
An error occurred in Python.
@guidorice ➜ /workspaces/mojo/examples (main) $ echo $?
0
```

## expected output (src)

```
Hello Mojo 🔥!
9
6
3
Hello from Python!
...
```

### Steps to reproduce

- Include relevant code snippet that did not work as expected.
- If applicable, add screenshots to help explain the problem.
- Include anything else that might help us debug the issue.

PS first impression is it's looking really awesome. Very responsive, lightweight binaries, great CLI interface, repl, help msg. 👍🏽 

### System information

```shell
- What OS did you do install modular CLI/mojo on ?

On a github workspace, 32GB instance.


@guidorice ➜ /workspaces/mojo/examples (main) $ mojo deviceinfo.mojo 
System information: 
    OS          :  linux
    CPU         :  znver3
    Arch        :  x86_64-unknown-linux-gnu
    Num Cores   :  8
    CPU Features:  avx2
```

- Provide version information for Mojo by pasting the output of `mojo -v`

mojo 0.2.0 (4c0ef274)

- Provide Modular CLI version by pasting the output of `modular -v`

modular 0.1.1
```

**Comment by jackos** [comment]
Thanks for raising, for Github Codespaces in `hello_interop.mojo` you will need to change line 29 to:

```
alias test_dir = "/workspaces/mojo/examples"
```

And also install numpy into the Python instance it's using:

```bash
python3.8 -m pip install numpy
```

We'll work on resolving both of these, thanks for the positive words!

shashankprasanna:
Hi @guidorice thank you for reporting!
Unfortunately we've hardcoded `alias test_dir = "/root/mojo-examples"` 
Please change it to `alias test_dir = "."` or the directory from which you're running `mojo hello_interop.mojo`

PR #517, opened 2023-08-25
Fix interop in examples

jackos:
Fixes issue #515 and #514 

Will look for the modules in the current directory or `./examples`, in case the user is running from the root of the repo, and print an error if it can't find the required modules.

It will also install `numpy` if it's missing, as failing when missing a pip module currently doesn't give an informative error.

### DIFF
--- a/examples/matmul.mojo
+++ b/examples/matmul.mojo
@@ -66,16 +66,16 @@ struct Matrix:
 fn run_matmul_python(M: Int, N: Int, K: Int) -> Float64:
     var gflops: Float64 = 0.0
     let python = Python()
-    alias test_dir = "/root/mojo-examples"
     try:
-        Python.add_to_path(test_dir)
+        Python.add_to_path(".")
+        Python.add_to_path("./examples")
         let pymatmul_module: PythonObject = Python.import_module("pymatmul")
         if pymatmul_module:
             gflops = pymatmul_module.benchmark_matmul_python(
                 M, N, K
             ).to_float64()
         else:
-            print("Python matmul module not found")
+            print("pymatmul module not found")
     except e:
         print(e.value)
         pass
--- a/examples/hello_interop.mojo
+++ b/examples/hello_interop.mojo
@@ -26,14 +26,11 @@ def main():
     print("Hello Mojo 🔥!")
     for x in range(9, 0, -3):
         print(x)
-    alias test_dir = "/root/mojo-examples"
     try:
-        Python.add_to_path(test_dir)
-        let test_module: PythonObject = Python.import_module("simple_interop")
-        if test_module:
-            _ = test_module.test_interop_func()
-        else:
-            print("Could not locate module: ", test_module)
+        Python.add_to_path(".")
+        Python.add_to_path("./examples")
+        let test_module = Python.import_module("simple_interop")
+        test_module.test_interop_func()
     except e:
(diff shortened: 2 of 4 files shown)


guidorice:
Github Workspace has python3.8 and python3.10 installed. How did know to select 3.8 env to install numpy into? (or how does mojo select virtual envs?)

examples/matmul.mojo
```diff
@@ -162,8 +160,7 @@ fn matmul_tiled_parallelized(C: Matrix, A: Matrix, B: Matrix):
                     C.store[nelts](
                         m,
                         n + x,
-                        C.load[nelts](m, n + x)
-                        + A[m, k] * B.load[nelts](k, n + x),
+                        C.load[nelts](m, n + x) + A[m, k] * B.load[nelts](k, n + x),
```
dualvtable, line 163: Jack - this line was formatted by `mojo format`; otherwise the code would fail the lint step.

examples/matmul.mojo
```diff
@@ -191,8 +188,7 @@ fn matmul_tiled_unrolled_parallelized(C: Matrix, A: Matrix, B: Matrix):
...
```
dualvtable, line 191: Same as above.

examples/matmul.mojo
```diff
@@ -250,10 +246,7 @@ fn main():
         M,
         M,
         python_gflops,
-        (
-            "Throughput of a 512x512 matrix multiplication in Mojo using"
-            " vectorization: "
-        ),
+        "Throughput of a 512x512 matrix multiplication in Mojo using vectorization: ",
```
dualvtable, line 249: Same as above.

dualvtable:
lgtm - thanks for doing the PR!

dualvtable (approved):
+1 from me.

examples/matmul.mojo
jackos: @dualvtable I did a `mojo format` on all the files and that's what it tuned into, so decided to leave it in the PR. This repo doesn't have those lint checks on it.

jackos:
> Github Workspace has python3.8 and python3.10 installed. How did you know to select 3.8 env to install numpy into? (or how does mojo select virtual envs?)

It's a peculiarity of Github Codespaces because they install the main python on path (3.10) to a folder in `~/.python` while the system uses 3.8 in `/usr/bin/python3.8`. It selects the libpython to link to during install, we have a `TODO` to allow the user to select which libpython to link to. But you can change it manually now in the file `.modular/modular.cfg` under `python_lib`

guidorice:
@jackos  that makes sense- thanks. On a related note, with `mojo build` it would be good to provide some hints to user about how to resolve/set MOJO_PYTHON_LIBRARY (I tried a few paths on my system, but no luck)

```shell
@guidorice ➜ /workspaces/mojo/examples (main) $ mojo build matmul.mojo 
@guidorice ➜ /workspaces/mojo/examples (main) $ ./matmul 
Throughput of a 128x128 matrix multiplication in Python: 
Mojo/Python interoperability error: Unable to locate a suitable libpython, please set `MOJO_PYTHON_LIBRARY`
Segmentation fault (core dumped)
```
I assume it build the binary against a specific version of libpython, and so that same one must be set in the env var MOJO_PYTHON_LIBRARY?

PR #544, opened 2023-09-01
[Docs] Improves matmul and mandelbrot parallel notebooks and examples

jackos:
- Uses the new ComplexSIMD and Tensor type from the std lib for mandelbrot instead of implementing them from scratch
- Speeds mandelbrot up by 168x using simplified version from Abdul's blog post, and parallel now 20x faster than sequential
- Matmul notebook now 73,000x faster than python
- Adds a mandelbrot.mojo like the matmul.mojo


examples/matmul.mojo
```diff
@@ -202,31 +205,33 @@ fn matmul_tiled_unrolled_parallelized(C: Matrix, A: Matrix, B: Matrix):
...
+        fn test_fn():
+            _ = func(C, A, B, rt)
+
+        let secs = Float64(Benchmark().run[test_fn]()) / 1_000_000_000
+        # Prevent the matrices from being freed before the benchmark run
+        _ = (A, B, C)
+        let gflops = ((2 * M * N * K) / secs) / 1e9
+        let speedup: Float64 = gflops / base_gflops
+        # print(gflops, "GFLOP/s", speedup, " speedup")
+        print(str)
```
dualvtable, line 233: Non-blocking suggestion, but at some point, we probably want the pretty print fn to reduce the precision. Perhaps in the next update:

```python
def float_printer(f: Float32):
    let builtin = Python.import_module('builtins')
    let pyf = builtin.float(f)
    builtin.print(builtin.str("{:4.2f}").format(pyf))
```

dualvtable (approved):
LGTM - thanks for the changes!

examples/matmul.mojo
jackos: Yes good idea, maybe we can even pop that in the std lib :)

lattner:
Wowwow, nice Jack!

PR #545, opened 2023-09-01
Fix tensor being destroyed before benchmark finished

jackos:
Example will also continue to work with release 0.2.0


dualvtable (approved):
Ah - this is the older API for `simd_store`. Looks good!

PR #557, opened 2023-09-02
Mandelbrot and matmul improvements

jackos:
Sync changes from main


Mogball (approved):
Seems like this should be done automatically

PR #635, opened 2023-09-08
Create a_taste_of_dynamism.mojo

rd4com:
what do you think?

### DIFF
--- a/examples/a_taste_of_dynamism.mojo
+++ b/examples/a_taste_of_dynamism.mojo
@@ -1,0 +1,30 @@
+#a taste of dynamism in pure mojo
+fn main():
+    try:
+
+        var db=object([])
+        var user = object([])
+        
+        var field = object(["array"])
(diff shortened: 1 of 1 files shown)


drunkwcodes:
I think this is good but still too early to write them.

Nonetheless, it is necessary in regard to https://github.com/modularml/mojo/issues/476 .
This makes people have less trial and error times.

rd4com:
```
struct singleton:
    @staticmethod
    def increase(inout scope):
        scope+=1
    @staticmethod
    def myinit(inout scope):
        scope = 10
    @staticmethod
    def change_type(inout scope):
        scope = "hello world"

def main():
    print("ok")
    var singletonscope:object = object(0)
    singleton.myinit(singletonscope)
    singleton.increase(singletonscope)
    print(singletonscope)
    singleton.change_type(singletonscope)
    print(singletonscope)
```

PR #743, opened 2023-09-12
fix docker build command

MIDORIBIN:
The build command '\\' was added because it was missing.


Zarzamora8 (approved):

jackos:
Thanks for raising this PR, this Dockerfile has changed since so closing

PR #770, opened 2023-09-13
Add an "output tiled" version of matmul, performs quite a bit better when single threaded

dsharlet:
In my experience, the strategy used in `matmul_tiled_output` in this PR is the best "simple" matrix multiply approach. I have C++ versions of this and the mojo strategies, the C++ version of `matmul_tiled_output` is ~2x faster than any of the other strategies.

When disabling parallelism, the implementation in this PR is 25-50% faster than the fastest existing implementation. I'm on WSL, where parallelism performs very strangely/not well in all examples.

I think that it wouldn't take much work on the mojo compiler to make the generated inner loop code match the C++ version. I think most of what is going wrong is things LLVM should be giving you for "free" anyways. Some examples:
- I've seen a lot of examples where it seems like common subexpressions aren't being eliminated, including redundant loads and redundant address arithmetic computations.
- In the C++ version, LLVM seems to be able to lift store-load sequences out of loops, which enables the accumulators to stay in registers between iterations of `k`.
- (Not exactly LLVM) `unroll` doesn't seem to actually unroll, is this just a hint?

https://github.com/dsharlet/mojo_comments has more notes and observations (disassembly, etc.) from playing around with this example in mojo.

Anyways, it's been fun playing with mojo, it's a nice language and I'm happy to see a language that gives programmers low level control over the generated code.


Mogball:
`unroll` definitely unrolls the loop. It's not a hint

jackos (approved):
Hi @dsharlet this is great thanks, I've integrated it with the latest version of Mojo, and some other changes I made upstream. This was a big help, much appreciated.

PR #851, opened 2023-09-19
Fix deviceinfo.mojo: only the last found CPU feature was being shown

kipras:
The "CPU Features" row of the output from _mojo/examples/deviceinfo.mojo_ showed only one CPU feature (the last one detected). This fixes it.


jackos (approved):
Fantastic, thanks @kipras

PR #883, opened 2023-09-22
fix: hello.mojo

kirklin:


lattner:
It's correct as is, thanks for the suggestion ;-)

PR #901, opened 2023-09-23
Docker compose

N-E-W-T-O-N:
I have created a docker-compose.yml for a better utilization of Dockerfile.

@dualvtable please let me know for any improvement.


jackos (changes requested):
Hi this is great thanks @N-E-W-T-O-N builds and starts the container in a single command. The files have moved around since you raised this though. Do you mind placing the docker-compose.yaml into `examples/docker` please?

N-E-W-T-O-N:
Ok I did the little rearangement as you asked

examples/docker/build-image.sh
```diff
@@ -1,65 +1,65 @@
-#!/bin/sh
```
jackos, line 1: All the line endings in this file have changed to windows style `CR` `LF`, which makes it fail on Linux. You can revert this from the base of the repo by doing `git checkout main -- examples/docker/build-image.sh`

jackos (changes requested):
Everything else looks good, thanks again.

examples/docker/build-image.sh
N-E-W-T-O-N: Change the end of line sequence as asked.

PR #954, opened 2023-09-29
11

mehranbahramm:


jackos:
Hi, thanks for the PR, we already have hello world in the examples though

PR #1122, opened 2023-10-20
example/benchmarks: compare against numpy dot

zachgrayio:
Generally no one is doing such a naive matmul I don't think, I'm thinking something like this is a bit more realistic. 

IMO, something like this is great to show that mojo has quite a bit to offer over numpy's matrix ops as well (a question that I had until I ran this)

```
$ mojo matmul.mojo
Python:        	   0.006 GFLOPS
Python (Numpy):	   1.603 GFLOPS
Naive:         	   5.920 GFLOPS         1052.24x faster than Python
Vectorized:    	  20.414 GFLOPS         3628.55x faster than Python
Parallelized:  	  43.906 GFLOPS         7804.28x faster than Python
Tiled:         	  47.673 GFLOPS         8473.98x faster than Python
Unrolled:      	  55.627 GFLOPS         9887.68x faster than Python
Accumulated:   	 436.572 GFLOPS        77601.01x faster than Python
```

### DIFF
--- a/examples/matmul.mojo
+++ b/examples/matmul.mojo
@@ -84,7 +84,7 @@ def run_matmul_numpy() -> Float64:
     let pymatmul: PythonObject = Python.import_module("pymatmul")
     let py = Python.import_module("builtins")
 
-    let gflops = pymatmul.benchmark_matmul_python_numpy(M, N, K).to_float64()
+    let gflops = pymatmul.benchmark_matmul_numpy(M, N, K).to_float64()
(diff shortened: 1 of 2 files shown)


jackos (approved):
Hi @zachgrayio we all really like this idea, I added a column for comparing each benchmark to numpy, and made the shape for the numpy benchmark match mojo for fair comparison.

Thanks for raising this great idea!

zachgrayio:
awesome this is great TY!

PR #1142, opened 2023-10-23
This example demonstrates a simple Mojo class with a property, an ini…

PentesterPriyanshu:
i have added a simple example.mojo file 
This example demonstrates a simple Mojo class with a property, an initializer, and a method. It prints the value of the property to the console.


PentesterPriyanshu:
@jackos   please review and merge it

jackos:
Hi @PentesterPriyanshu this is Swift code, which won't run with Mojo 😄

Issue #1163, opened 2023-10-26
Readme.md


CrypticRevenger:
### What can we do better?

Table of Contents: Create a table of contents at the beginning of the README to help users quickly navigate to the sections they're interested in.
More About the Playground: Provide a brief overview of what users can do in the Mojo Playground. Mention if there are any limitations, offer some tips, or include a link to a more detailed guide.

CrypticRevenger:
@modularbot please review the PR #1164.

scottamain:
The Mojo repo is very limited at this point, so I don't think it needs a table of contents yet. The best place to learn more about Mojo is in the Mojo Manual: https://docs.modular.com/mojo/manual/

PR #1164, opened 2023-10-26
Update Readme.md #1163

CrypticRevenger:


lattner:
@scottamain fyi

scottamain:
Hi @CrypticRevenger, I really appreciate your excitement to contribute to the Mojo repo! 🔥 However, this pull request is a wholesale rewrite of the README without any explanation or rationalization. I see a few ideas for content in here that I like, but the overall approach in this PR is not aligned with the style and content that we want right now.

This raises a very important problem though: **We don't have a contributing guide and we really need one**. That would help give contributors like you a better starting point and ensure that we can reach an agreement on an idea before putting in the effort to make changes. So I prefer that we not accept this change, because it's not aligned with our style, rewriting the README is not a priority right now, and it will require too much work to make this PR right.

What I prefer is that you [create a docs issue](https://github.com/modularml/mojo/issues/new/choose) and tell us specifically what information you see missing that you think is important to include in the README, and we can discuss making some smaller, more incremental changes.

Thank you! 🙏

PR #1199, opened 2023-10-30
fix typos

hairyputtar:
parameteric --> parametric
seperate --> separate
perfromance --> performance
performace --> performance
evaluting --> evaluating


jackos (approved):
Great thanks!

PR #1288, opened 2023-11-14
improve readability in matmul examples

mikowals:
- use `_` to iterate a variable that won't be used rather than defining `i` on two lines in a row
- omit the 0 starting value in `range` when its unnecessary
- use the `@unroll`decorator where possible
- group `alias` definitions near the start of the matmul example that uses them.  There are still some hidden in the nested tile functions but at least most are easier to find.

There is also the `M * 2` num_workers that  probably needs a comment or some other way to make it more obvious.  On my M1 Pro using just `M` was faster though the `swizzling` example generally did not show much performance improvement even after reducing the workers.  But I assume `M * 2` is a specific tweak that works well on some systems, might make large performance differences between examples, and might need to be found and tweaked by others.


jackos (approved):
Great thanks!

PR #1357, opened 2023-12-02
[examples] Don't hard-code matrix size for Python in examples/matmul

dev0x13:
Apparently the `let gflops = pymatmul.benchmark_matmul_python(128, 128, 128).to_float64()` in `run_matmul_python` should be  `let gflops = pymatmul.benchmark_matmul_python(M, N, K).to_float64()` for a fair comparison.


dev0x13:
Okay, I now see [this](https://github.com/modularml/mojo/commit/7cd671e10acd1ee5386cfeb739c96dbdc3f29b55) commit explaining such a change.

PR #1389, opened 2023-12-05
Update hello-world.md

scw:
Four spaces for example indentation, as shown in other examples and mentioned as the preferred format (as is with Python).


scottamain (approved):
Thanks!

PR #1403, opened 2023-12-05
remove TODO as traits is landed

YichengDWu:


scottamain (approved):
Great catch, thanks! :D

PR #1406, opened 2023-12-05
Fix a few typos in docs

rikhuijzer:
Found by running [`typos`](https://github.com/crate-ci/typos) with the following `_typos.toml` config:

```toml
[default.extend-words]
inout = "inout"
ND = "ND" # NDBuffer
nane = "nane" # Wrong on purpose in an example.
mis = "mis" # mis-features
```


scottamain (approved):
That's great thank you! We were actually just having a chat about an action like that one. :D

PR #1445, opened 2023-12-08
Proposal For An Actor System Based On Mojo

reid-spencer:
This is currently a work in progress.  There are no code changes, just a proposal written in the proposals section. This was pre-approved by Chris Lattner in a conversation in June 2023. 

I will keep working on this as I have time, but it is far enough along that I'm looking for feedback and assistance from interested parties. 

I will take it out of draft mode when its a little further along. 

Signed-off-by: Reid Spencer <reid@ossuminc.com>

### DIFF
--- a/proposals/actors/READM-ME-FIRST.md
+++ b/proposals/actors/READM-ME-FIRST.md
@@ -3,7 +3,7 @@
 * _Author_: [Reid Spencer](https://github.com/reid-spencer)
 * _Prior Work_: [Swift Concurrency Manifesto](https://gist.github.com/lattner/31ed37682ef1576b16bca1432ea9f782)
 * _Date_: September-December 2023
-* _Project Working Name_ : Macsy ( Mojo ACtor SYstem, pronounced "moxy")
+* _Project Working Name_ : Moxy
 
 ## Overview
 Mojo gives us the opportunity to re-imagine how an Actor System may 
--- a/proposals/actors/READM-ME-FIRST.md
+++ b/proposals/actors/READM-ME-FIRST.md
@@ -1,0 +1,46 @@
+# Mojo Actor System Proposal
+
+* _Author_: [Reid Spencer](https://github.com/reid-spencer)
+* _Prior Work_: [Swift Concurrency Manifesto](https://gist.github.com/lattner/31ed37682ef1576b16bca1432ea9f782)
+* _Date_: September-December 2023
+* _Project Working Name_ : Macsy ( Mojo ACtor SYstem, pronounced "moxy")
+
+## Overview
+Mojo gives us the opportunity to re-imagine how an Actor System may 
+work in the world of heterogeneous computing. Actor Systems have been 
+actively researched since the 1970s and have been in use in a variety of
+manifestations since then. This proposal wants to carry that model forward into
+the 2020s and beyond.
+ 
+We outline here a long-term view of how to tackle a very large problem.
+It explores *one possible* approach to adding an actor model to Mojo. 
+Importantly, we also outline how this can be done in careful iterations
+over time.
+
+## Participation
+This proposal hopes to catalyze positive discussion that leads Mojicians to a
+best-possible design for actors in Mojo. It is not intended to be a
+specification for how actors will be implemented in Mojo but, certainly, that
+specification will make reference to this Proposal. This document should
+encourage participation from the community of Mojicians interested in an
+actor model for Mojo and not, despite its origin, be the exclusive product of
+the primary author's mind.
+
+## Contents
+* [About Actors](about-actors.md)
+  * A refresher on actors and actor systems 
+* [Pain Points Addressed](pain-points-addressed.md) 
+  * A recitation of the software development pain points that actors alleviate
+* [Actor System Features](actor-system-features.md)
+  * A list of what we expect Mojo Actor System be
+* [Not Actor System Features](not-actor-system-features.md)
+  * A list of things we don't want the Mojo Actor System to be or have
+* [Mojo Features Needed](mojo-features-needed.md)
+  * The features of the Mojo compiler we think are needed to implement the
+  Mojo Actor System
+* [Expected Optimizations](expected-optimizations.md)
+  * An exploration of some of the optimizations we expect from basing an 
+  actor model on Mojo 
+* [Milestones](milestones.md)
+  * A rough estimate of the order in which the goals of this proposal should
+    be delivered
--- a/proposals/actors/about-actors.md
+++ b/proposals/actors/about-actors.md
@@ -187,3 +187,19 @@ these benefits:
   maintaining a facade of a single scalabe computing entity (a cluster of
   nodes)
 
+### Actor processing is side-effect free
+Since actor state is private and we intend to make it impossible to break
+the encapsulation of it (your program won't compile if you try), message
+processing should be considered free of side-effects from the functional
+programming point of view. While the actor's behavior may change the 
+state of the actor, that is a local effect that has no bearing on anything
+else. In other words, the only way to observe the effect is by sending 
+another message, or implying the effect in a response message. It can't 
+be measured or observable directly. 
+
+A consequence of this is that functional programming enthusiasts can 
+and should consider a message send to be equivalent to an asynchronous,
+non-blocking, function call.  The sender need not wait for a response
+and there may not be a response. If there is a response, it too is sent
+by another asynchronous non-blocking message to the sender of the original
+message.  
--- a/proposals/actors/about-actors.md
+++ b/proposals/actors/about-actors.md
@@ -1,0 +1,189 @@
+## About Actors
+While some familiarity with actors and actor systems is expected in this 
+proposal, this section provides a light coverage of actors for the unfamiliar. 
+If you are thoroughly versed, you can probably skip this section. 
+
+### Background
+Actors have a deep theoretical basis and have been explored by academia since
+they were invented in 1973 by Carl Hewitt at MIT.
+
+A full explanation of [the Actor Model is available on Wikipedia](https://en.wikipedia.org/wiki/Actor_model), and there
+are other academic references on the [c2 wiki page](http://wiki.c2.com/?ActorsModel). These are good place to 
+start reading if you'd like to dive into some of the theoretical
+fundamentals that back the actor model.
+
+The actor model also has a proven commercial track record with  [Erlang](https://www.erlang.org) and
+[Akka](http://akka.io), which successfully power a large number of highly scalable and reliable
+systems. Erlang is used by Ericcson in most of its telecommunications equipment. 
+Akka is used by Tesla to [operate a digital twin
+system across America for managing its fleet of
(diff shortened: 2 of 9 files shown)


lattner:
This is cool reid, thank you for putting this together.  We're quite a bit too early to invest in this area IMO (we need to get traits much further along and complete lifetimes) but I think this is a very likely long term direction.  If you're interested, Actors got built into swift with a more complex model than was in the manifesto, you can read about it here, or on several swift-evolution proposals:
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency#Actors

I do hope we can eschew the complexity, a lot of it is due to legacy interop with apple frameworks.  OTOH we may need such things to work with legacy python and other libs though.

reid-spencer:
@lattner - I understand the language earliness, but I think there's value in starting the Actor's project early. So, I have started already: https://github.com/ossuminc/moxy (extremely nascent).  In the proposal, I've tried to minimize the requirements on Mojo. The recent introduction of Traits allowed me to get started. All else that is needed is default implementations in the traits.  Later on, when Mojo has matured, it would be interesting to integrate an ASIC or GPU to help with extremely fast message dispatch.  All in good time. I'm happy to start this work without the involvement of Modular's time/resources; at least for now. 

Thanks for the reference to the Actor implementation in Swift.  I am examining several actor systems to try and glean the winning strategies from their patterns. Akka is my strongest entry knowledge, but I'm open to merging the best ideas from other ecosystems. 

I plan to leave interoperability to the end of actor development and not sacrifice simplicity or performance. In other words, interoperability will have its own complexity and costs, as an add-on.

Brian-M-J:
I have a suggestion for the mojo-features-needed.md [concurrency section](https://github.com/ossuminc/mojo/blob/actors/proposals/actors/mojo-features-needed.md#concurrency): senders and receivers, a.k.a. [`std::execution`](https://github.com/NVIDIA/stdexec). Hylo (formerly Val) also has this (mentioned in this [talk](https://youtu.be/ws-Z8xKbP4w?t=2752)).

Basically, it's a universal abstraction to express all concurrency and parallelism. Of particular note are the theoretical results in this [paper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2504r0.html) (The paper refers to senders as 'computations'):
>1) Result 1 (general solution for scheduling tasks): There is a general method of scheduling tasks that works with all concurrent problems, that is safe (without the need of additional blocking synchronization), and is as efficient as it can be (under the greedy assumption).

Note: The greedy assumption is as follows -
>By greedy assumption, we mean that adding a task to be executed as soon as possible is a good choice. We cannot know what tasks would soon be available for execution, so, trying to hold back tasks is not necessarily a good strategy. Using the greedy assumption typically leads to good efficiency, but it is not guaranteed to have the maximum efficiency.

>2) Result 2 (general applicability of computations): All concurrent programs can be expressed with computations in a safe manner, without requiring blocking synchronization and with high efficiency for computation execution (under the greedy assumption).
>3) Result 3 (general solution with computations): There is a general method of scheduling computations that works with all concurrent problems, that is safe (without the need of additional blocking synchronization), and is as efficient as it can be (under the greedy assumption).
>4) Result 4 (general top-down decomposition): Any concurrent program can be modeled as a computation and can be decomposed in terms of computations; if those computations are large enough, they can be further decomposed until reaching unit computations.

This [article](https://accu.org/journals/overload/29/165/teodorescu/) states many benefits of senders and receivers:
>- ability to represent all computations (from simple ones to complex ones)
>- ability to represent computations that have inner parts that can be executed on different execution contexts (e.g., different threads, different computation units, different machines)
>- can use computations to solve all concurrency problems (without using synchronisation primitives in the user code)
>- composability
>- proper error handling and cancellation support
>- no memory allocations needed to compose basic computations
>- no blocking waits needed to implement most of the algorithms
>- allows flexibility in specialising algorithms, thus allowing implementors to create highly efficient implementations
>- ability to interoperate with coroutines

Senders and receivers enable [structured concurrency](https://www.youtube.com/watch?v=Xq2IMOPjPs0). Using them, you can make higher level abstractions for concurrency and parallelism:
>It’s also worth mentioning that the term computation applies to multiple paradigms. It can be easily used to describe imperative work, it can be well assimilated by functional programmers, it can apply to reactive programming and to all stream-based paradigms; although we haven’t talked about it, we can think of computations also in the context of the actor model.

So we can use senders and receivers to implement the actor system, while allowing users to build any concurrency abstraction they need without the need for locks.

## It's not all sunshine and rainbows
As explained in this [video](https://www.youtube.com/watch?v=uSG240pJGPM), senders and receivers have a lot of advantages, but the C++ implementation has some flaws too:
### Safety
- The C++ implementation doesn't guarantee the absence of race conditions. Maybe by using lifetimes or reference capabilities from Pony, Mojo can make up for this.
### Performance
- There is a little dynamic memory allocation needed.
### Ease of Use
- It feels like a whole new programming paradigm, so there can be some work done to make it more beginner friendly.
- It requires function colouring, which spreads throughout the program.
- The rules that programmers need to keep in their heads to avoid edge cases and weird bugs will also need to be taught.

If Mojo can implement senders and receivers in a better way by avoiding these pitfalls, I think it will have the best concurrency model out of all programming languages.

Brian-M-J:
Also, for the [Message Transmission Acceleration](https://github.com/ossuminc/mojo/blob/actors/proposals/actors/expected-optimizations.md#message-transmission-acceleration) section, I think a [DPU](https://en.wikipedia.org/wiki/Data_processing_unit) would be a better fit, as it is already being used for encryption and decryption. We can also use it for serialization and deserialization.

JoeLoser:
Going to close this PR for now since there hasn't been much activity on this in several months.  Feel free to reopen when we're ready to take on this kind of work.

nmsmith:
@Brian-M-J I've looked into the senders and receivers proposal. I see people promising big things ("a global solution to concurrency"), but notably, they don't seem to be able to back up their claims with evidence. I would expect to see an example of a non-trivial concurrent program that is dramatically easier to write with senders as opposed to being written with async/await and tasks etc. All I see are toy examples.

I am really skeptical that something big has been discovered. If it had, more people would have noticed by now. Senders and receivers have been around for 4+ years.

Brian-M-J:
> I would expect to see an example of a non-trivial concurrent program that is dramatically easier to write with senders as opposed to being written with async/await and tasks etc.

I'd be able to show it to you if it was open source 🙂. In terms of _applications_, I'd say the most prolific users of S&R are at Meta. In terms of libraries, there's plenty of open source stuff out there (see [users of HPX](https://hpx.stellar-group.org/hpx-users/)).

I guess talks like [this](https://youtu.be/XaNajUp-sGY) would be good demonstrations at least.

I guess another thing to note is that S&R is meant for high performance, so the C++ implementation might not be the most beautiful library. Starting from a blank slate (Hylo, Mojo (somewhat)) means that some syntactical niceties can be added to make it easier to use.

> I am really skeptical that something big has been discovered. If it had, more people would have noticed by now.

As you noted, S&R is relatively new, so it isn't widely used yet. Though I wouldn't say that only a few people have noticed it. Clearly Nvidia has noticed, because they host the reference implementation. Meta has noticed it, because they've been using it in production for years. Bloomberg has noticed, because a few of their employees are authors of the proposals. Adobe has noticed, because there are plans to remodel the stlab library in terms of S&R, and Hylo is supported by the company. The users of libraries like HPX, folly etc. have noticed. The C++ committee has noticed, because they've voted favorably on it multiple times.

nmsmith:
> I guess talks like [this](https://youtu.be/XaNajUp-sGY) would be good demonstrations at least.

That's another toy example. All he's done is create an event loop that spawns asynchronous tasks one-at-a-time. This is trivial to do in any language with async/await and a `Task` type.

I would _love_ if S&R has solved some major problems with modelling concurrent systems, but I don't see it.

nmsmith:
At the risk of stating the obvious, there are a lot of projects out there developed by well-meaning, passionate people, who promise that they have created something important. But most of the time, that doesn't turn out to be the case. I've been burned a lot in the past by believing that a project is as important as the contributors say it is, and then I've begun to experiment with it, only to eventually discover that I've wasted my time.

Brian-M-J:
> I would _love_ if S&R has solved some major problems with modelling concurrent systems, but I don't see it.

Well, the only thing I can show you when I don't have access to the code is this: Hylo has gotten rid of function colouring (see [here](https://docs.hylo-lang.org/language-tour/concurrency)).

BTW when they say "no `actor` abstractions" it just means that actors would be a library thing. You can build an actor system using S&R.

I myself am working on a Mojo implementation of S&R based on this [talk](https://youtu.be/xiaqNvqRB2E) in a private repo. That way when Mojo's coroutines drop with some benchmarks, I'll have something concrete to compare to. The biggest problem is that I don't know what I'm doing 😅. I'm not a low level / library expert or anything.

nmsmith:
Getting rid of function coloring is a worthy goal, no doubt about that. But that is orthogonal to S&R. The reason most PLs have colored functions is because the thread-based concurrency model that they already had is compatible with implicit suspension and implicit migration of tasks between threads. In the case of Python, the main reason you can't implicitly switch tasks is because a Python program is a big soup of shared mutable state with no synchronization/critical sections, so two tasks can easily race each other.

Brian-M-J:
> On top of that, you'd want a way to perform transactions on shared state. This is a big design space worth exploring. S&R doesn't really have a solution here.

One of the authors of p2300 wrote an [article](https://accu.org/journals/overload/28/158/teodorescu/) on how to use tasks to replace locks. The same ideas are repeated in [p2504](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2504r0.html#conc-tasks-specialized-ser).

nmsmith:
I've seen that. The idea is that if you can **statically** identify all of the places in your codebase where a variable is being accessed by multiple tasks—and if at least one of those tasks mutates the variable—then you (or your compiler) can conceivably restructure the program (cut it up into subtasks) such that any time a task needs to access the shared variable, you defer the access to a scheduler/executor.

This is a good observation, and I strongly agree that a task-based concurrency model should aim to do this. (Concurrency without explicit locking would be amazing!) But this is—again—completely orthogonal to S&R. S&R doesn't give me a simple way to _write_ that restructured program, from what I can tell.

nmsmith:
Actually, I'd only read the second article you linked, but the [first article](https://accu.org/journals/overload/28/158/teodorescu/) is more interesting IMO, because it actually discusses the "program restructuring" problem I'm referring to:
> Let’s assume that one has an application that can be broken down in tasks relatively easily. But, at some point, deep down in the execution of a task one would need a lock. Ideally one would break the task in 3 parts: everything that goes before the lock, the protected zone, and everything that goes after. Still, easier said than done; this can be hard if one is 20 layers of function-calls – it’s not easy to break all these 20 layers of functions into 3 tasks and keep the data dependencies correct. If breaking the task into multiple parts is not easily doable, then one can also use the fork-join model to easily get out of the mess.

This proposed solution—forking a new task to mutate the shared variable—makes a lot of sense. It's worth exploring further.

npuichigo:
@nmsmith maybe take a look at how NVIDIA leverages P2300 with CUDA to do async computation on GPUs? https://www.youtube.com/watch?v=nwrgLH5yAlM

PR #1451, opened 2023-12-09
docs and example update

rahuldshetty:
In the HeapArray example, set the "cap" size as the total allocation for the pointer rather than the "size" of the current allocated elements. 

image

Signed-off-by: Rahul D Shetty [35rahuldshetty@gmail.com](mailto:35rahuldshetty@gmail.com)


scottamain (approved):
Nice, thank you!

PR #1455, opened 2023-12-09
Correct a roadmap example to match reality

reid-spencer:
A comment is a cut & paste error and should
be deleted. Additionally, the 0.6.0 compiler
generates a different error message not
related to the point being discussed.

Signed-off-by: Reid Spencer <reid.spencer@ossuminc.com>

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -509,10 +509,18 @@ def pick_func(cond):
         def bar0(): return 42
         result = bar0
     else:
-        def bar1(): return 3 # error: redeclaration of 'bar'
+        def bar1(): return 3 
         result = bar1
     return result
 ```
+In mojo 0.6.0, this yields:
+```
+pick_func.mojo:9:12: error: cannot implicitly convert 'fn() raises capturing -> object' value to 'object' in return value
+    return result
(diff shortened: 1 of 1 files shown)


docs/roadmap.md
```diff
@@ -509,10 +509,18 @@ def pick_func(cond):
...
         result = bar0
     else:
-        def bar1(): return 3 # error: redeclaration of 'bar'
+        def bar1(): return 3 
         result = bar1
     return result
 ```
+In mojo 0.6.0, this yields:
+```
+pick_func.mojo:9:12: error: cannot implicitly convert 'fn() raises capturing -> object' value to 'object' in return value
```
arthurevans, line 518: I think this is just an error in the code example. It appears that `pick_func()` needs to indicate that it returns a function:

```
def pick_func(cond) -> def() capturing:
```
Then the rest of the example should work fine.

arthurevans (changes requested):
The comment is indeed a typo. But we should fix the code example rather than describing the current error.

arthurevans:
Thanks for the PR. Code example has been fixed, so I'm going to close this.

PR #1491, opened 2023-12-15
[mojo-examples] Fix raytracing notebook

jackos:
Fix after change of name from utils.vector to collections.vector


examples/notebooks/RayTracing.ipynb
```diff
@@ -599,7 +599,7 @@
         "from algorithm import parallelize\n",
         "from math.limit import inf\n",
         "# DynamicVector doesn't work in jupyter environment yet\n",
-        "from utils.vector import _OldDynamicVector as DynamicVector\n",
+        "from collections.vector import _OldDynamicVector as DynamicVector\n",
```
arthurevans, line 602: This will break with the currently-released SDK.

PR #1534, opened 2023-12-21
nbody.mojo - remove unnecessary memory writes and @unroll inner loop

mikowals:
Edit: I found a second optimisation unrolling the inner loop so the total improvement is 6.53 -> 1.13 seconds on M1 Pro.  That change is discussed in the next comment.

------------------
On M1 Pro the result of `benchmark()` falls from 6.53 to 4.40 seconds.  I realise this code is reproducing a specific benchmark but some of the other competitors do this optimisation so I think it is fine.

One of the [Julia code examples](https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/nbody-julia-8.html) explains it well. 

```
    @inbounds for i=1:4
        bi = bodies[i]
        # For body i, since velocity is the only part of the object
        # changing on each iteration, it's more efficient to update
        # outside the vector.
        iv = bi.v
        for j=i+1:5
            iv = iv .- Δx[k] .* (mag[k] * bodies[j].m)
            bodies[j] = Body(bodies[j].x,
                             bodies[j].v .+ Δx[k] .* (mag[k] * bi.m),
                             bodies[j].m)
            k += 1
        end
        bodies[i] = Body(bi.x, iv, bi.m)
    end
```

### DIFF
--- a/examples/nbody.mojo
+++ b/examples/nbody.mojo
@@ -63,9 +63,7 @@ fn advance(inout bodies: StaticTuple[NUM_BODIES, Planet], dt: Float64):
     for i in range(NUM_BODIES):
         var body_i = bodies[i]
 
-        # @unroll(NUM_BODIES - 1) errors.
-        # Bug reported at https://github.com/modularml/mojo/pull/1534
-        @unroll(4)
+        @unroll(NUM_BODIES - 1)
         for j in range(NUM_BODIES - i - 1):
             var body_j = bodies[j + i + 1]
             let diff = body_i.pos - body_j.pos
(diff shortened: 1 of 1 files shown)


mikowals:
Adding @unroll decorator to the inner loop of `advance` reduces the time to 1.13 seconds.  

Ideally the amount of unrolling could be linked to the alias `NUM_BODIES` but that led to #1542 .  So I have hardcoded `@unroll(4)`.  

I can also rewrite using `unroll` from algorithm for both the outer and inner loops which works fine with the alias but does make the code a bit less readable.

jackos (approved):
Sorry for the delay @mikowals, just double-checked that @unroll with a number more than the actual loops wouldn't cause UB. It's a lot faster, thanks!

PR #1578, opened 2023-12-30
Fix broken vector links

isuckatcs:
Some of the links in the docs still point to the old location of `DynamicVector`.

This patch intends to make the links point to the correct location.


scottamain (approved):
Great thanks!

PR #1585, opened 2023-12-30
Fix typo under docs/manual/values/value-semantics.ipynb

sraza-onshape:
Just a small spelling error, in the section titled "Value semantics in `def` vs `fn`".


scottamain (approved):
Thank you!

PR #1661, opened 2024-01-25
Fix various typos in documentation

nmsmith:
Plus rewording of a few sentences to improve their clarity.

### DIFF
--- a/docs/manual/parameters/index.ipynb
+++ b/docs/manual/parameters/index.ipynb
@@ -20,7 +20,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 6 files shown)


multikitty (approved):
Looks good.

docs/manual/parameters/index.ipynb
```diff
@@ -601,8 +601,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Recall that Mojo can infer parameter values in a parametric function, based on\n",
-    "the parametric values attached to an argument value (see the `rsqrt[]()`\n",
+    "Recall that when a parametric function is called, Mojo can infer the parameter values.\n",
+    "In particular, it can use the parameter values attached to an argument value (see the `sqrt[]()`\n",
```
arthurevans, line 605: "In particular" suggests that there are other ways to infer parameters. Maybe:
```suggestion
    "That is, it can use the parameter values attached to an argument value (see the `sqrt[]()`\n",
```

We need to add a section that defines parameter inference explicitly, but it's going to take a little more rewriting. Meanwhile, this is an improvement.

arthurevans (changes requested):
Thanks. Nice fixes. One change requested.

arthurevans (approved):

PR #1687, opened 2024-01-27
Typo fix

LiamSwayne:
A simple typo fix that adds "is".

Before: "(compatibility a massive and important goal of Mojo)"

After: "(compatibility is a massive and important goal of Mojo)"


arthurevans (approved):
Thanks for the fix!

PR #1711, opened 2024-01-31
Submit proposal for new `inferred` parameters

Mogball:
This PR submits a new proposal document reviewed by the "language committee" to introduce a new `inferred` keyword when declaring parameters. These parameters would only be inferrable from other parameters or arguments in a function, but are skipped when binding functions at callsites. This will enable autoparameterization of parameters and omitting redundant parameters when calling certain functions.

For example:

```python
fn scalar_func[inferred dt: DType, x: Scalar[dt]]() -> Scalar[dt]:
    return x * 2

print(scalar_func[Int64(5)]())
```

### DIFF
--- a/proposals/inferred-parameters.md
+++ b/proposals/inferred-parameters.md
@@ -1,0 +1,144 @@
+# Inferring Parameters from Other Parameters
+
+A common feature in programming language with generics is the ability to infer the value of generics/templates/parameters from the argument types. Consider C++:
+
+```cpp
+template <typename T>
+void inferMe(T x) {}
+
+int x = 1;
+inferMe(x);
+// Equivalent to
+inferMe<int>(x);
+```
+
+Mojo is a parametric language and also supports this feature in a variety of use cases that make code significantly less verbose:
+
+```python
+fn infer_me[dt: DType, size: Int](x: SIMD[dt, size]): pass
+
+infer_me(Int32())
+# Equivalent to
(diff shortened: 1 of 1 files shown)


proposals/inferred-parameters.md
```diff
@@ -0,0 +1,144 @@
...
+```python
+scalar_param[infer_stuff_from_me=Int32()]()
+
+# One would like to write:
+scalar_param[Int32()]()
+```
+
+Laszlo Kindrat and I proposed several options to remedy this and members of the “Mojo Language Committee” (key stakeholders across Modular) met to discuss these ideas, summarized below.
+
+We decided to move forward with the following option. Mojo will introduce a new keyword, `inferred`, as a specifier for parameters only. `inferred` parameters must precede all non-inferred parameters in the parameter list, and they **cannot** be specified by a caller — they can **only** be inferred from other parameters. This allows us to express:
```
arthurevans, line 62: To match the description in the PR, maybe add "arguments" in here somehow.

```suggestion
We decided to move forward with the following option. Mojo will introduce a new keyword, `inferred`, as a specifier for parameters only. `inferred` parameters must precede all non-inferred parameters in the parameter list, and they **cannot** be specified by a caller — they can **only** be inferred from other parameters or from function arguments. This allows us to express:
```

arthurevans (approved):
LGTM, one optional clarification.

PR #1748, opened 2024-02-13
Various matmul example improvements, remove accumulate implementation

jackos:
Puts the public repo in sync with matmul improvements.

Fixes https://github.com/modularml/mojo/issues/1426


soraros:
There are places where `Scalar` can be used instead of `SIMD[_, 1]`.

ematejska (approved):

PR #1756, opened 2024-02-15
Changing main from fn to def

jackos:
Simplifies main function by changing to `def` to remove compiler warnings on unused Python variables.


shashankprasanna (approved):
looks good

PR #1789, opened 2024-02-21
First iteration Mojo Public CI Workflow

dlumma:


.github/ISSUE_TEMPLATE/workflows/ci.yml
JoeLoser, line 1: I think you meant for this workflow file to be in
.github/workflows/ci.yml not within the issue templates directory.

JoeLoser (approved):
LGTM now, thanks!  Next step would be to pull down the latest `mojo` compiler using the installer with an auth token and make sure something simple like `mojo --version` works?

PR #1794, opened 2024-02-22
CI update to get the modular installer, install mojo compiler

dlumma:
This PR includes logic to download the modular installer and use it to install the mojo compiler.


modocache (approved):
Great, thank you!

PR #1856, opened 2024-03-02
Update BoolMLIR.ipynb for Mojo 2024.1.0

gabrieldemarmiesse:
Two changes occured in Mojo 2024.1.0 which made some code in this notebook deprecated:
* Using `let` now raises a warning. We should use `var` instead.
* Initializers for `@register_passable` values can (and should!) now be specified with inout self arguments just like memory-only types.

See the changelog here: https://docs.modular.com/mojo/changelog#v241-2024-02-29

We currently show deprecated syntax in the notebook in the documentation. See https://docs.modular.com/mojo/notebooks/BoolMLIR#compile-time-constants

### DIFF
--- a/examples/notebooks/BoolMLIR.ipynb
+++ b/examples/notebooks/BoolMLIR.ipynb
@@ -159,7 +159,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "let b = OurBool()"
+    "var b = OurBool()"
    ]
   },
   {
@@ -182,7 +182,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 6,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


examples/notebooks/BoolMLIR.ipynb
```diff
@@ -284,12 +282,12 @@
    "source": [
     "`OurFalse` is declared to be of type `OurBool`, and then assigned an `i1` type -- in this case, the `OurBool` constructor we added is called implicitly.\n",
     "\n",
-    "With true and false constants, we can also simplify our original `__init__` constructor for `OurBool`. Instead of constructing an MLIR value, we can simply return our `OurFalse` constant:"
+    "We can also make use of Mojo's default arguments to create automatically a `False` value when the constructor of `OurBool` is called without arguments. "
```
arthurevans, line 285:
```suggestion
    "We can also make use of Mojo's default arguments to automatically create a `False` value when the constructor of `OurBool` is called with no arguments. "
```

arthurevans (approved):
LGTM with a minor wording tweak. Thanks!

PR #1859, opened 2024-03-02
Update RayTracing.ipynp to use latest syntax of `@register_passable`

gabrieldemarmiesse:
Similar to https://github.com/modularml/mojo/pull/1856, Mojo 2024.1.0 uses a simplified syntax to make the constructor of `@register_passable` objects. The release notes are saying that we should prefer this syntax and that the old one will be removed in a future release.

See the changelog here: https://docs.modular.com/mojo/changelog#v241-2024-02-29


arthurevans (approved):
LGTM, thanks.

PR #1959, opened 2024-03-15
Fix md: use var instead of let

junweizeng:
Use `var` instead of `let`in `packages.md` file.

The reason is as follows:

> Mojo 24.1 removed support for let from the compiler IR (but it still parses let for migration), subsequent releases will remove the let keyword entirely.

Quote from https://github.com/modularml/mojo/discussions/1456#discussioncomment-8672522.


arthurevans (approved):
LGTM, thanks.

PR #1962, opened 2024-03-15
[Examples] Add ability to run and test examples

JoeLoser:
Add a `lit.cfg.py` in the `examples` directory which sets up the lit config for
example Mojo files (notebooks not yet supported).  Add a wrapper script,
`run-examples.sh` which creates a build directory and invokes `lit` using this
configuration.

Go one step further and create a GitHub workflow that puts all of the pieces
together to download the latest stable Mojo compiler and test the examples using
the provided script.

The `scripts/run-examples.sh` is something that can be run both locally and in
CI.

Add common artifacts to the `.gitignore` such as the new `build` directory, and
other Python-related things.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,18 @@
+# Build directory
+/build
+
+# Reference: https://github.com/github/gitignore/blob/main/Python.gitignore
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
--- a/scripts/run-examples.sh
+++ b/scripts/run-examples.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+set -euo pipefail
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT="${SCRIPT_DIR}"/../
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -0,0 +1,2 @@
+build
```
modocache, line 1: I believe you'll want to constrain this a little more -- I believe this will ignore any directories that include a `build` path component. Based on the lit config below, I'm guessing this should be:

```suggestion
/build
```

.gitignore
```diff
@@ -0,0 +1,2 @@
...
+*.pyc
```
modocache, line 2: This is a good idea, but perhaps we should copy GitHub's gitignore for Python? https://github.com/github/gitignore/blob/main/Python.gitignore

I don't think we need every single line from that file, but I do think we could steal a few things. For example, [the first few lines](https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Python.gitignore#L1-L4) cover several kinds of Python caches, which I think may get created when running `lit` here.

scripts/run-examples.sh
```diff
@@ -0,0 +1,27 @@
...
+REPO_ROOT="${SCRIPT_DIR}"/../
+EXAMPLES_DIR="${REPO_ROOT}"/examples
+
+echo "Creating build directory"
+BUILD_DIR="${REPO_ROOT}"/build
+mkdir -p "${BUILD_DIR}"
+
+# Change into the examples dir where lit.cfg.py is
+# and then invoke lit to run the examples.
+(cd "${EXAMPLES_DIR}" && lit -sv .)
```
modocache, line 27: I'm surprised one can't just invoke `lit -sv "${EXAMPLES_DIR}"` here. I sort of thought that was the point of setting the `exec_root` in the lit config -- to have lit execute from a specific location, regardless of the location from which the user invoked `lit`.

modocache (approved):

.gitignore
JoeLoser: Valid point; just changed! Thanks.

ConnorGray (approved):

.gitignore
JoeLoser: Thanks for pointing me at that — I just added everything from https://github.com/github/gitignore/blob/main/Python.gitignore into our `.gitignore` and mentioned the origin reference.

scripts/run-examples.sh
JoeLoser: Thanks for keeping me accountable here.  I had to tweak the `test_source_root` as it was only using `Path.cwd()`, but it really needs to account for where it may be run as seen with:

```
config.test_source_root = Path(__file__).parent.resolve()
```

This makes it so test discovery works as expected and one doesn't need to invoke `lit` only from the examples directory (and thus `cd` there first).  It just works as you'd expect now 🎉

PR #1964, opened 2024-03-16
[CI] Explicitly specify `types:` under PR section

JoeLoser:
To ensure the workflow triggers on all types of status changes under the `pull_request` trigger, explicitly specify all the state changes.

While here, go ahead and add our open-source license to the top of each workflow.


dlumma (commented):
👍

PR #2001, opened 2024-03-26
[CI] Add workflow permissions

JoeLoser:
Add a `permissions` section to all workflow files to allow third-party GitHub actions to have read permissions.  This will fix the issue where the `dorny/paths-filter` step doesn't have proper read permissions to query the file list modified in a pull request, resulting in a hard error from the GitHub APIs it uses.


dlumma (approved):
Nice!

PR #2035, opened 2024-03-28
Refactor: Replace __get_lvalue_as_address with Reference

ubaidsk:
Fixes https://github.com/modularml/mojo/issues/2027

### DIFF
--- a/stdlib/src/collections/vector.mojo
+++ b/stdlib/src/collections/vector.mojo
@@ -243,5 +243,5 @@ struct InlinedFixedVector[
             An iterator to the start of the vector.
         """
         return Self._iterator(
-            0, self.current_size, __get_lvalue_as_address(self)
+            0, self.current_size, Reference(self).get_unsafe_pointer()
         )
--- a/stdlib/src/sys/ffi.mojo
+++ b/stdlib/src/sys/ffi.mojo
@@ -126,7 +126,8 @@ struct DLHandle(CollectionElement):
                 "dlsym", DTypePointer[DType.int8]
             ](self.handle.address, name)
             return (
-                Pointer(__get_lvalue_as_address(opaque_function_ptr))
+                Reference(opaque_function_ptr)
+                .get_unsafe_pointer()
                 .bitcast[result_type]()
                 .load()
             )
@@ -197,17 +198,14 @@ fn _get_dylib_function[
     alias func_cache_name = name + "/" + func_name
     var func_ptr = _get_global_or_null[func_cache_name]()
     if func_ptr:
-        return (
-            Pointer(__get_lvalue_as_address(func_ptr))
-            .bitcast[result_type]()
-            .load()
-        )
+        return func_ptr.bitcast[result_type]().load()
 
     var dylib = _get_dylib[name, init_fn, destroy_fn](payload)
     var new_func = dylib._get_function[func_name, result_type]()
     external_call["KGEN_CompilerRT_InsertGlobal", NoneType](
         StringRef(func_cache_name),
-        Pointer(__get_lvalue_as_address(new_func))
+        Reference(new_func)
+        .get_unsafe_pointer()
         .bitcast[Pointer[NoneType]]()
         .load(),
     )
--- a/stdlib/src/builtin/object.mojo
+++ b/stdlib/src/builtin/object.mojo
@@ -231,7 +231,7 @@ struct _Function:
     fn __init__[FnT: AnyRegType](value: FnT) -> Self:
         # FIXME: No "pointer bitcast" for signature function pointers.
         var f = Pointer[Int16]()
-        Pointer(__get_lvalue_as_address(f)).bitcast[FnT]().store(value)
+        Reference(f).get_unsafe_pointer().bitcast[FnT]().store(value)
         return Self {value: f}
 
     alias fn0 = fn () raises -> object
@@ -246,7 +246,8 @@ struct _Function:
     @always_inline
     fn invoke(owned self) raises -> object:
         return (
-            Pointer(__get_lvalue_as_address(self.value))
+            Reference(self.value)
+            .get_unsafe_pointer()
             .bitcast[Self.fn0]()
             .load()()
         )
@@ -254,7 +255,8 @@ struct _Function:
     @always_inline
     fn invoke(owned self, arg0: object) raises -> object:
         return (
-            Pointer(__get_lvalue_as_address(self.value))
+            Reference(self.value)
+            .get_unsafe_pointer()
             .bitcast[Self.fn1]()
             .load()(arg0)
         )
@@ -262,7 +264,8 @@ struct _Function:
     @always_inline
     fn invoke(owned self, arg0: object, arg1: object) raises -> object:
         return (
-            Pointer(__get_lvalue_as_address(self.value))
+            Reference(self.value)
+            .get_unsafe_pointer()
             .bitcast[Self.fn2]()
             .load()(arg0, arg1)
         )
@@ -272,7 +275,8 @@ struct _Function:
         owned self, arg0: object, arg1: object, arg2: object
     ) raises -> object:
         return (
-            Pointer(__get_lvalue_as_address(self.value))
+            Reference(self.value)
+            .get_unsafe_pointer()
             .bitcast[Self.fn3]()
             .load()(arg0, arg1, arg2)
         )


lattner:
Please don't do this as a literally mechanical change.  Please refactor the individual uses to use the proper apis on Reference where possible

Shaikh-Ubaid:
Thanks @lattner. I updated it. Can you take a look again, please?

docs/roadmap.md
```diff
@@ -605,7 +605,7 @@ fn printIntRef(x: UnsafeIntRef):
 
 var c: Int = 10
 # "reference" operator
-var ref = UnsafeIntRef(__get_lvalue_as_address(c)) # &Int -> Pointer[Int]
```
Shaikh-Ubaid, line 608: Should this example have initially been 
`var ref = UnsafeIntRef(Pointer(__get_lvalue_as_address(c)))`?

JoeLoser: Probably, yes.  In general, this example is out-of-date since the use above even uses `__get_address_as_lvalue` which got removed recently.  For the purposes of this PR, perhaps just revert the change to this file, and the @modularml/mojo-standard-library will work to update this roadmap section in a follow-up PR if that works for you.

stdlib/src/collections/vector.mojo
```diff
@@ -243,5 +243,5 @@ struct InlinedFixedVector[
             An iterator to the start of the vector.
         """
         return Self._iterator(
-            0, self.current_size, __get_lvalue_as_address(self)
+            0, self.current_size, Reference(self)
```
JoeLoser, line 246: **Suggestion** I think this should be `Reference(self).get_unsafe_pointer()`.  The old code worked because `__get_lvalue_as_address(self)` returns a `!kgen.pointer` which would be used to construct the `Pointer` type that is the expected argument to `_deref_iter_impl`.  Here, you're trying to construct (and then later invoke `_deref_iter_impl`) with a `Reference` type instead of a `Pointer` type.  Unlike `!kgen.pointer,` `Pointer` is not able to implicitly be constructed from a `Reference`.

stdlib/src/memory/anypointer.mojo
```diff
@@ -91,7 +91,7 @@ struct AnyPointer[T: Movable](
             An AnyPointer which contains the address of the argument.
         """
         return __mlir_op.`pop.pointer.bitcast`[_type = Self.pointer_type](
-            __get_lvalue_as_address(arg)
+            Reference(arg)
```
JoeLoser, line 94: **Suggestion** I think this should be `Reference(self).get_unsafe_pointer()` to be explicit.

stdlib/src/memory/unsafe.mojo
```diff
@@ -717,7 +717,7 @@ struct Pointer[
             A Pointer struct which contains the address of the argument.
         """
         return __mlir_op.`pop.pointer.bitcast`[_type = Self.pointer_type](
-            __get_lvalue_as_address(arg)
+            Reference(arg)
```
JoeLoser, line 720: **Suggestion** I think this should be `Reference(self).get_unsafe_pointer()` to be explicit.

JoeLoser (commented):
Generally looks good, but I think we're missing a few places that need to specify `get_unsafe_pointer()` still as I commented in the PR.

Additionally, can you please rebase with the latest `nightly` branch?  CI is using the `nightly/mojo` compiler that was released today which is incompatible with the version of the code you have checked out on your current branch.

stdlib/src/collections/vector.mojo
Shaikh-Ubaid: Thanks for the detailed clarification!

Shaikh-Ubaid:
Thanks @JoeLoser. I updated the PR as per the suggestions.

stdlib/src/sys/ffi.mojo
```diff
@@ -198,7 +198,7 @@ fn _get_dylib_function[
     var func_ptr = _get_global_or_null[func_cache_name]()
     if func_ptr:
         return (
-            Pointer(__get_lvalue_as_address(func_ptr))
+            Reference(func_ptr).get_unsafe_pointer()
```
soraros, line 201: Can't we simply use `func_ptr` here?

Shaikh-Ubaid: I think we can. It seems `func_ptr` already is a `Pointer`, so we need not construct a reference from it and convert back to pointer.

I updated it. Thanks!

Shaikh-Ubaid:
From the CI, it looks like some formatting changes are expected. I will update them and get back.

JoeLoser:
> From the CI, it looks like some formatting changes are expected. I will update them and get back.

Yep, we require code to be formatted with `mojo format`.  As seen [in the CI logs](https://github.com/modularml/mojo/actions/runs/8513844780/job/23319812130?pr=2035#step:8:34), you can fix this locally by just running `mojo format stdlib/src/builtin/object.mojo stdlib/src/sys/ffi.mojo -l 80` which is what [CI uses](https://github.com/modularml/mojo/blob/nightly/.github/workflows/standard_library_tests_and_examples.yml#L111).

Shaikh-Ubaid:
I did the formatting changes. I also attempted running the tests. The tests pass for me in the `nightly` branch. But they fail in the PR branch:
```console
(mojo) ubaid@ubaids-MacBook-Pro mojo % ./stdlib/scripts/run-tests.sh
Included from /Users/ubaid/Desktop/OpenSource/mojo/stdlib/src/memory/__init__.mojo:1:
/Users/ubaid/Desktop/OpenSource/mojo/stdlib/src/memory/anypointer.mojo:94:23: error: cannot form a reference to an argument that might instantiate to @register_passable type
            Reference(arg).get_unsafe_pointer()
                      ^
Included from /Users/ubaid/Desktop/OpenSource/mojo/stdlib/src/memory/__init__.mojo:1:
/Users/ubaid/Desktop/OpenSource/mojo/stdlib/src/memory/unsafe.mojo:716:23: error: cannot form a reference to an argument that might instantiate to @register_passable type
            Reference(arg).get_unsafe_pointer()
                      ^
mojo: error: failed to parse the provided Mojo
```

Shaikh-Ubaid:
`error: cannot form a reference to an argument that might instantiate to @register_passable type`

Since `@register_passable` types are to be always passed by copy, it seems we cannot have/create a Reference for it.

JoeLoser:
> `error: cannot form a reference to an argument that might instantiate to @register_passable type`
> 
> Since `@register_passable` types are to be always passed by copy, it seems we cannot have/create a Reference for it.

Interesting!  This makes sense.  The compiler is currently correctly rejecting these things right now.  The short story is that register values don't have a lifetime value associated with them, so you can't form a `Reference` to it.  So, in the short-term, I'd recommend we keep these two uses as before with the `__get_lvalue_as_address`.  `@register_passable` is going through some wider changes right now, and this may fall out from that work when it's done.  Until then, I think we'll have to keep these two uses around before we remove the builtin from the compiler.  FYI @Mogball.

Shaikh-Ubaid:
Thanks Joe! Tests pass for me locally now:
```console
(mojo) ubaid@ubaids-MacBook-Pro mojo % ./stdlib/scripts/run-tests.sh              
Successfully created /Users/ubaid/Desktop/OpenSource/mojo/stdlib/scripts/../../build/stdlib.mojopkg
Packaging up the test_utils.

Testing Time: 6.60s

Total Discovered Tests: 78
  Unsupported:  7 (8.97%)
  Passed     : 71 (91.03%)
(mojo) ubaid@ubaids-MacBook-Pro mojo % ./examples/run-examples.sh                 
                                      -- Testing: 5 tests, 5 workers --                                      

Testing Time: 12.01s

Total Discovered Tests: 5
  Passed: 5 (100.00%)
```

Shaikh-Ubaid:
>So, in the short-term, I'd recommend we keep these two uses as before with the __get_lvalue_as_address. @register_passable is going through some wider changes right now, and this may fall out from that work when it's done

Let me know if you would like me to create an issue for it.

JoeLoser (approved):

JoeLoser:
> > So, in the short-term, I'd recommend we keep these two uses as before with the __get_lvalue_as_address. @register_passable is going through some wider changes right now, and this may fall out from that work when it's done
> 
> Let me know if you would like me to create an issue for it informing that those two use cases of `__get_lvalue_as_address` need to be removed in future.

I'll take a look internally to see if we already have an existing issue for this reg-passable problem before creating a new GitHub issue.  Now that we open-sourced the standard library, the @modularml/mojo-standard-library will likely be moving to publicly created GitHub issues as much as we can (we'll announce more info soon).  For this one, it's mostly in the compiler camp now in order for the library to remove these last two uses, so I'll work with them internally.  Appreciate your hard work in this contribution! 🎉

Shaikh-Ubaid:
Thanks for your reviews and suggestions on this Joe!

PR #2038, opened 2024-03-28
[stdlib] Add `String.removeprefix` and `String.removesuffix`

gabrieldemarmiesse:
Apologies since adding methods is not listed as "Changes we accept" in the readme. I had already implemented those methods in my own repo and it was very low effort to submit a first PR. Furthermore I believe adding those two methods is not controversial
and has a very low bikeshedding potential.

The two methods `removeprefix` and `removesuffix` are present in python. See the docs: 
* https://docs.python.org/3/library/stdtypes.html#str.removeprefix
* https://docs.python.org/3/library/stdtypes.html#str.removesuffix

Both were introduced in python 3.9.

The documentation was added, the argments were made positional-only like in CPython.

Unit tests were added in the corresponding module.

The docstrings were taken from the python documentation.

### DIFF
--- a/stdlib/test/builtin/test_string.mojo
+++ b/stdlib/test/builtin/test_string.mojo
@@ -545,6 +545,20 @@ fn test_endswith() raises:
     assert_false(str.endswith("llo", 2, 3))
 
 
+def test_removeprefix():
+    assert_equal(String("hello world").removeprefix("hello"), " world")
+    assert_equal(String("hello world").removeprefix("world"), "hello world")
+    assert_equal(String("hello world").removeprefix("hello world"), "")
+    assert_equal(String("hello world").removeprefix("llo wor"), "hello world")
+
+
+def test_removesuffix():
+    assert_equal(String("hello world").removesuffix("world"), "hello ")
+    assert_equal(String("hello world").removesuffix("hello"), "hello world")
+    assert_equal(String("hello world").removesuffix("hello world"), "")
+    assert_equal(String("hello world").removesuffix("llo wor"), "hello world")
+
+
 def test_intable():
     assert_equal(int(String("123")), 123)
 
@@ -590,5 +604,7 @@ def main():
     test_hash()
     test_startswith()
     test_endswith()
+    test_removeprefix()
+    test_removesuffix()
     test_intable()
     test_string_mul()
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -999,6 +999,47 @@ struct String(Sized, Stringable, IntableRaising, KeyElement, Boolable):
             return self._endswith_impl(suffix, start)
         return self[start:end]._endswith_impl(suffix)
 
+    fn removeprefix(self, prefix: String, /) -> String:
+        """If the string starts with the prefix string, return `string[len(prefix):]`.
+        Otherwise, return a copy of the original string.
+
+        ```mojo
+        print(String('TestHook').removeprefix('Test'))
+        # 'Hook'
+        print(String('BaseTestCase').removeprefix('Test'))
+        # 'BaseTestCase'
+        ```
+        Args:
+          prefix: The prefix to remove from the string.
+
+        Returns:
+          A new string with the prefix removed if it was present.
+        """
+        if self.startswith(prefix):
+            return self[len(prefix) :]
+        return self
+
+    fn removesuffix(self, suffix: String, /) -> String:
+        """If the string ends with the suffix string, return `string[:-len(suffix)]`.
+        Otherwise, return a copy of the original string.
+
+        ```mojo
+        print(String('TestHook').removesuffix('Hook'))
+        # 'Test'
+        print(String('BaseTestCase').removesuffix('Test'))
+        # 'BaseTestCase'
+        ```
+
+        Args:
+          suffix: The suffix to remove from the string.
+
+        Returns:
+          A new string with the suffix removed if it was present.
+        """
+        if self.endswith(suffix):
+            return self[: -len(suffix)]
+        return self
+
     @always_inline
     fn _endswith_impl(self, suffix: String, start: Int = 0) -> Bool:
         return self.rfind(suffix, start) + len(suffix) == len(self)


ConnorGray:
Hey, @gabrieldemarmiesse thanks for your contribution! 

The ["Changes we accept"](https://github.com/modularml/mojo/blob/main/CONTRIBUTING.md#changes-we-accept) section documents change types we know we are okay with accepting, but is not an exhaustive list of all change types we're willing to accept. This change you've submitted is an example of the kind of change we're open to accepting as well :)

This change looks good, and thank you for the attention to detail in referencing the Python equivalent of these new methods. 

Though, a minor point on the naming of these methods. It's true that Mojo is seeking to provide a great onboarding experience to developers familiar with Python, and that means fidelity to Python syntax and libraries in many ways. However, at our current stage we're also looking to improve the consistency of the Mojo standard library, and part of that is a stronger convention for the naming of types and functions in the Mojo standard library.

The official [Style Guide section on naming](https://github.com/modularml/mojo/blob/main/stdlib/docs/style-guide.md#identifier-naming-conventions) documents that `snake_case()` is the preferred naming convention for new functions and methods. Admittedly the Mojo standard library is not perfectly consistent about this at the moment. But it is the convention we're aiming towards.

Could you please rename these new methods, respectively, to `String.remove_prefix` and `String.remove_suffix`?

Let me know if you have any questions, and thank you again for your help growing the Mojo🔥 standard library! :)

gabrieldemarmiesse:
I'm not especially in favor of deviating from the naming of Python's functions and methods, but this is not necessarily the place to discuss this. I made the change that you asked, we should be good to merge :)

gryznar:
I do agree with consistency and do not copy-paste Python naming. Thank you @ConnorGray for following this!

stdlib/test/builtin/test_string.mojo
gryznar, line 1: How about using "Mojo" here instead of "world"? It will be nice if example would also include 🔥to check if truncating non-ASCII characters works

gabrieldemarmiesse: It's not clear that bikeshedding on the exact words used in the unit tests is productive, so we shouldn't change "world" into "Mojo".

Non-ascii characters are not supported in mojo strings yet, it only works sometimes, in some situations. I believe we should add such a test only when we have proper support for utf-8 & friends, otherwise the unit test might break from one release to the next without indicating a legitimate bug in the method.

loftusa:
Isn't the goal of Mojo to be a drop-in replacement for Python? How could that be achieved if Mojo is deviating from Python function names? I'd expect that most users would want to be able to just run their python scripts in the mojo environment and have it work, regardless of whether some functions in python's standard library follow naming conventions.

gryznar:
Mojo is not aiming to copy paste CPython. This approach has a lot of flaws

gabrieldemarmiesse:
If you feel strongly about this, please open an issue so that users can voice their opinions. It will be much easier to discuss than in a pull request

gryznar:
No, pr right now looks good and I would like to prevent IMHO wrong changes

Senhaji-Rhazi-Hamza:
> Mojo's `String` is not Python `str`
> 
> Mojo is not aiming to copy paste CPython. This approach has a lot of flaws, e.g.:
> 
> * fat stdlib - not acceptable for system programming language
> * performance hits - relying on CPython implementation may restrict Mojo's performance
> * violation on style guide contributions. Mojo has defined rules to follow to be consistent
> * maintainability issues - Python stdlib is huge due to all batteries includes. Maintaining such is Mojo is waste of developer's time
> 
> Impression that code from CPython can be directly copied to Mojo is harmful illusion. Migration from Python to Mojo is a task for separate tool. If someone would to use almost drop-in replacement, he probably should look at PyPy or Taichi Lang

I don't think it's about copying, it's about having a similar API that integrate seemlessly with existing code, also the manual https://docs.modular.com/mojo/manual/ here, states ->  "We also designed Mojo as a superset of Python because we love Python and its community" 

Small decisions like that, have unexpected consequences -> would have been more coherent to keep the same api (function signatures in this case)

gryznar:
And that's why, there should be "_" to keep consistency. I am against messing up like in Python.

Senhaji-Rhazi-Hamza:
> And that's why, there should be "_" to keep consistency. I am against messing up like in Python.

People overtime will probably correct it in the next python versions, so will be good to correct it afterward, i think there is a distinction to be made, for new code (that doesn't exist in python) and existing code, i do agree that python is messing it up there, but there is a practical & pragmatic trade off there to be considered and discussed.

Thak you for the work you're doing ! we have huge expectation from mojo on the next comming years

ConnorGray:
Hey folks,

First off, I want to say we really appreciate the enthusiasm from the community on this topic. We all want Mojo🔥 to be a success, and it's fantastic to see folks engaging, weighing the options, discussing tradeoffs, and working with us towards that success.

Secondly, and I apologize to @gabrieldemarmiesse for the churn here, but after some further discussion amongst the standard library team, wider Mojo team, and consideration of the different approaches community members here have outlined, we think it is probably best for the time being if we follow the Python naming conventions for these particular functions, and revert to the intuitive naming this PR originally chose: `String.removeprefix` and `String.removesuffix`. We should also do similarly for subsequent improvements to this particular module, and similar modules that have an existing adherence to Pythonic naming and structure.

I want to apologize to the community as well for the mixed messaging. This is one of many nuanced design choices we and community will need to navigate as we work to build out the standard library as Mojo continues to mature. We discovered here that we have some work to do regarding naming. While we have long had strong Python interoperability/compatibility as a guiding "north star", we're learning alongside the community about how we should apply and weigh that principle to individual small design decisions. 

To that end, we'll be continuing to discuss and improve our current [Style Guide](https://github.com/modularml/mojo/blob/main/stdlib/docs/style-guide.md), especially in relation to issues like this that arise as we and the community grow.

Thank you again to everyone who contributed to this discussion. Mojo🔥 is still young, and there's a lot of exciting work to be done, and we're looking forward to learning and growing with all of you :). Special thanks to @gabrieldemarmiesse for the PR, and bearing with us as we iron out these growing pains.

gryznar:
@ConnorGray it's a pitty that Mojo will sacrifice its consistency to copy Python functionalities..

JoeLoser (approved):

Issue #2017, opened 2024-03-28
[mojo-stdlib] Add `list.pop(index)` API


JoeLoser:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Currently, our `List` type only supports `pop_back` to remove and return the last element in the list.  However, we support `__setitem__` for inserting a value at an arbitrary index.  Naturally, we should be allowed to remove elements at an arbitrary position that is in range.

Note the Python `list` datatype supports `pop(index)` too as seen [here](https://docs.python.org/3/tutorial/datastructures.html).  If no index is specified, it removes the last element and returns it.  I think preserving that behavior is desirable.

### What is your motivation for this change?

Parity with Python `List` API and to allow for specific-index removal.

whym1here:
Could you point to the current `pop_back` implementation?

JoeLoser:
> Could you point to the current `pop_back` implementation?

Of course.  [This](https://github.com/modularml/mojo/blob/76d05839487905d6c7459c7135fdb5bb9f23f336/stdlib/src/collections/list.mojo#L229) is the current implementation.  I recommend we add a new function: `pop(index: Int)` and preserve the `pop_back` for now.  We can deprecate it and remove it in a normal migration manner after we have the `pop` function.

sazid:
Does this look like a valid implementation of the `pop(index)` behavior?

```mojo
@always_inline
fn pop(inout self, index: Int) -> T:
    """Pops the value at the given index and returns it.

    Returns:
        The popped value.
    """
    debug_assert(-self.size <= index < self.size, "index must be within bounds")

    current_len = len(self)

    var normalized_idx = index
    if index < 0:
        normalized_idx += len(self)

    var ret_val = (self.data + normalized_idx).take_value()
    self.size -= 1

    # Move all elements to the left by one index.
    for i in range(normalized_idx, current_len-1):
        self[i] = self[i+1]

    if self.size * 4 < self.capacity:
        if self.capacity > 1:
            self._realloc(self.capacity // 2)

    return ret_val^
```

Chalked it up on my windows machine, if this looks good I'll open and submit a PR after building and testing it on linux/mac.

whym1here:
I could be wrong here but I see an issue in this implementation.
- currently `fn __getitem__(self, i: Int) -> T:` returns a copy of element at `i` 
[...]

PR #2041, opened 2024-03-28
[mojo-stdlib]Add list.pop(index) API to stdlib

jiex-liu:
This pull request serves to add `list.pop(index)` API to the stdlib as requested by issue #2017.

This API should handle the same functionality as the python `list.pop`

Unit tests includes:

- popping both positive and negative index
- making sure returned value is consistent
- making sure list len is consistent

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -61,7 +61,9 @@ struct _ListIter[
         inout self,
     ) -> Reference[T, list_mutability, list_lifetime]:
         self.index += 1
-        return self.src[].__get_ref[list_mutability, list_lifetime](self.index - 1)
+        return self.src[].__get_ref[list_mutability, list_lifetime](
+            self.index - 1
+        )
 
     fn __len__(self) -> Int:
         return len(self.src[]) - self.index
@@ -436,9 +438,7 @@ struct List[T: CollectionElement](CollectionElement, Sized):
     ](
         self: Reference[Self, mutability, self_life].mlir_ref_type,
         i: Int,
-    ) -> Reference[
-        T, mutability, self_life
-    ]:
+    ) -> Reference[T, mutability, self_life]:
         """Gets a reference to the list element at the given index.
 
         Args:
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -61,9 +61,7 @@ struct _ListIter[
         inout self,
     ) -> Reference[T, list_mutability, list_lifetime]:
         self.index += 1
-        return self.src[].__get_ref[list_mutability, list_lifetime](
-            self.index - 1
-        )
+        return self.src[].__get_ref[list_mutability, list_lifetime](self.index - 1)
 
     fn __len__(self) -> Int:
         return len(self.src[]) - self.index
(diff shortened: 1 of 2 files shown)


lsh:
I think this might want a return type of `Optional[T]` rather than just `T`, because of the case where the `List` is empty.

See also Swift's [`Array.popLast`](https://developer.apple.com/documentation/swift/array/poplast\(\)) and Rust's [`Vec::pop`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.pop)

LJ-9801:
That is a good point. Also `pop_back()` needs to be modified to deal with this situation as well. Either `debug_assert()` the list when it's empty or simply just return a `Optional[T](None)` and do not decrement the size when the list is empty.

lsh:
Another point is to maybe return `None` if the `index` is out of bounds.

whym1here:
If we want `pop` to behave similar to [`pop` in python](https://docs.python.org/3/tutorial/datastructures.html) then you can 
change:
```mojo
fn pop(inout self, i: Int) -> T:
```
to:
```mojo
fn pop(inout self, i: Int = -1) -> T:
```

So that `pop_back` can be deprecated in the future.

stdlib/test/collections/test_list.mojo
```diff
@@ -84,6 +84,34 @@ def test_list():
     assert_equal(0, len(list))
     assert_equal(2, list.capacity)
 
+    # Test pop with index
+    for i in range(6):
+        list.append(i)
+    
+    # try poping from index 3 for 3 times
+    for i in range(3, 6):
+        assert_equal(i, list.pop(3))
```
Laotree, line 93: Pop 3 times, it hides the order of pop out. You can consider popping once to check instead of popping 3 times to check the remaining members.

stdlib/test/collections/test_list.mojo
```diff
@@ -84,6 +84,34 @@ def test_list():
     assert_equal(0, len(list))
     assert_equal(2, list.capacity)
 
+    # Test pop with index
```
JoeLoser, line 87: **Suggestion** Do you mind moving this into a separate function for testing this `pop` API functionality?

stdlib/src/collections/list.mojo
```diff
@@ -239,6 +237,31 @@ struct List[T: CollectionElement](CollectionElement, Sized):
...
+    fn pop(inout self, i: Int = -1) -> T:
+        """Pops a value from the list at the given index.
+
+        Args:
+            i: The index of the value to pop.
+
+        Returns:
+            The popped value.
+        """
+        debug_assert(-self.size <= i < self.size, "pop index out of range")
```
JoeLoser, line 250: **Thought** This is fine for now.  However, I think we should consider making this API return a `Optional[T]` and embrace algebraic datatypes even though it notably has us differ from Python.  In this particular case, we're _already_ differing from Python since in Python, this function would raise for an index error, whereas this would abort.  This would get changed once we have parametric raises though.  Right now, it's good to _not_ have this function raise until we have parametric raises. 

This is related to [this comment](https://github.com/modularml/mojo/pull/2041#issuecomment-2026512721) from @lsh.

WDYT @modularml/mojo-standard-library?

JoeLoser (approved):

JoeLoser:
Can you please rebase with the latest nightly branch? CI is using the `nightly/mojo` compiler that was released today which is incompatible with the version of the code you have checked out on your current branch.

stdlib/test/collections/test_list.mojo
LJ-9801: will do

JoeLoser:
Just merged — thanks so much for the contribution!

PR #2082, opened 2024-03-30
[mojo-stdlib] Allow `my_optional is None` and `my_optional is not None`

gabrieldemarmiesse:
Just as the title says, it's a very common idiom in python, is even recommended by PEP8: https://peps.python.org/pep-0008/#programming-recommendations

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.
Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

While the warning does not apply to Mojo, Python users are quite used to it and it's a small quality of life improvement which is inexpensive to support.

We could also add the same methods to the `Bool` struct if the maintainers agree.

### DIFF
--- a/stdlib/src/collections/optional.mojo
+++ b/stdlib/src/collections/optional.mojo
@@ -147,6 +147,32 @@ struct Optional[T: CollectionElement](CollectionElement, Boolable):
             return self._value.get[T]()[]
         return default
 
+    fn __is__(self, other: NoneType) -> Bool:
+        """Return `True` if the Optional has no value.
+
+        It allows you to use the following syntax: `if my_optional is None:`
+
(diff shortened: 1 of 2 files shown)


soraros:
Wow, pretty cool syntactic/typing trick!

gabrieldemarmiesse:
> You might also want to add this to OptionalReg.
Yep, that can be the topic of another PR

JoeLoser (approved):
LGTM, thank you!  Let me think a bit on whether we want to do the same for `Bool` and circle back with you.

PR #2085, opened 2024-03-30
[mojo-stdlib] Add method `Dict.update()`

gabrieldemarmiesse:
Here is the python documentation about this method: https://docs.python.org/3/library/stdtypes.html#dict.update

There should be another overload: `fn update(self, **kwargs: V):` which we can't do yet since it would only work with keys that are strings and https://github.com/modularml/mojo/issues/1876 is not fixed yet.

This is why we force the argument to be positional-only. If we allowed it to be passed by keyword, it would break when we will intruduce the other overload of `Dict.update`.

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -638,6 +638,17 @@ struct Dict[K: KeyElement, V: CollectionElement](Sized, CollectionElement):
             0, 0, Reference(self)
         )
 
+    fn update(inout self, other: Self, /):
+        """Update the dictionary with the key/value pairs from other, overwriting existing keys.
+
+        The argument must be positional only.
+
+        Args:
+            other: The dictionary to update from.
+        """
+        for entry in other.items():
+            self[entry[].key] = entry[].value
+
     @staticmethod
     @always_inline
     fn _new_entries(reserved: Int) -> List[Optional[DictEntry[K, V]]]:
--- a/stdlib/test/collections/test_dict.mojo
+++ b/stdlib/test/collections/test_dict.mojo
@@ -192,6 +192,48 @@ def test_dict_copy_calls_copy_constructor():
     assert_equal(6, copy["a"].copy_count)
 
 
+def test_dict_update_nominal():
+    var orig = Dict[String, Int]()
+    orig["a"] = 1
+    orig["b"] = 2
+
+    var new = Dict[String, Int]()
+    new["b"] = 3
+    new["c"] = 4
+
+    orig.update(new)
+
+    assert_equal(orig["a"], 1)
+    assert_equal(orig["b"], 3)
+    assert_equal(orig["c"], 4)
+
+
+def test_dict_update_empty_origin():
+    var orig = Dict[String, Int]()
+    var new = Dict[String, Int]()
+    new["b"] = 3
+    new["c"] = 4
+
+    orig.update(new)
+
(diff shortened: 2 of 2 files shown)


stdlib/test/collections/test_dict.mojo
```diff
@@ -192,6 +192,48 @@ def test_dict_copy_calls_copy_constructor():
...
+    new["c"] = 4
+
+    orig.update(new)
+
+    assert_equal(orig["b"], 3)
+    assert_equal(orig["c"], 4)
+
+
+def test_dict_update_empty_new():
+    orig = Dict[String, Int]()
```
JoeLoser, line 224: **Question** Is there a reason why to omit the `var` here (e.g. `var orig = ...` instead of `orig = ...`)?  Ditto for the `new` variable a few lines below.

JoeLoser (commented):
Looks good, thank you!  Just had one quick question around missing a `var` in a specific test.

stdlib/src/collections/dict.mojo
```diff
@@ -638,6 +638,17 @@ struct Dict[K: KeyElement, V: CollectionElement](Sized, CollectionElement):
             0, 0, Reference(self)
         )
 
+    fn update(inout self, other: Self, /):
```
StandinKP, line 641: Just a question. What is the use of / here?

gabrieldemarmiesse: It means the argument is positional-only: https://docs.modular.com/mojo/manual/functions#positional-only-and-keyword-only-arguments

stdlib/test/collections/test_dict.mojo
gabrieldemarmiesse: Sometimes I do it and sometimes I don't, I don't think too much about it. Since the compiler allows both in `def` and that there is no guidelines in the style guide, I didn't pay too much attention to it. 

If the maintainers want `var` to be present even in `def` functions, I would advise to add it to the style guide, and even better, automate the check for it. Otherwise, they might end up spending quite some time telling contributors to add `var` when they forgot it. I would say the same if we have conventions about where to use `fn` vs `def`. I guess that even reviewers might miss a `def` making it's way into the stdlib. It's very easy to miss.

My personal recommendation about `var` vs `nothing` would be to no care about it, but I don't feel strongly about this, as long as we (contributors and maintainers) don't spend too much time on the matter, it feels a bit like bikeshedding. 

I added `var` in this test, as I believe this is what you wanted, this should be good to merge :)

JoeLoser (approved):

stdlib/test/collections/test_dict.mojo
JoeLoser: Thanks for the reply.  I'm still getting used to the fact that we can even omit the `var` or `let` keyword now when within a `def` context, so bare with me.  My intent wasn't to bikeshed — sorry if it came off that way.  We don't have an official policy on it right now as you pointed out (and I think that's OK, as long as we don't bikeshed/mandate a particular style without it being written down/expected by everyone in the Style Guide or the like).  I suspect we'll figure this one out as we go and get more mileage on `mojo` code now that `var` or `let` isn't required in a `def`.  But either way, it's not a biggie.  I won't bother you (or other contributors) about this sort of thing now that I'm reminded that it's fine 😄

gabrieldemarmiesse: No worries :)

Issue #2024, opened 2024-03-28
[mojo-stdlib] Replace tests using `FileCheck` with `testing` module where possible


JoeLoser:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Many tests use `FileCheck` still which isn't great for many reasons for a standard library. We should move these over to use functions like `assert_equal`, `assert_true`, and friends from the `testing` module where possible.

### What is your motivation for this change?

Moving things over to use our `testing` module helps "eat our own cooking" as we build out the test framework.  Additionally, often, the code is much shorter — while still achieving the same intent as the original code.

### Any other details?

There are several test files still using `FileCheck` as can be identified with:

```
git grep -l '| FileCheck'
```

I wouldn't worry too much about the examples to start with; in fact, I recommend we fix those _last_ only after we've changed as many of the standard library tests as we can.

Note: if multiple people plan on working on this, please coordinate to avoid two people working in porting the tests over for the same file.

fowlerlee:
@JoeLoser: is this issue still open for working on (if so I wish to contribute and fix it) ? @shenyihOng has committed to a different ticket (https://github.com/modularml/mojo/pull/2069) but there are still files that use the FileCheck error handling. I can help by fixing the remaining files work.

gabrieldemarmiesse:
#2451 is the last PR we can do in this direction. The only FileCheck left are:
* The tests for `debug_assert`
* The tests for `print`
* The tests for `trap`
* The tests for the docs and examples
Which have all effects that are impossible to grab and inspect from inside the Mojo process. So maybe let's close this issue then? No other improvement can be added from the community's side I believe

JoeLoser:
Sounds good, thank you for all of your hard work in this area, @gabrieldemarmiesse!  Getting the remaining tests off of `FileCheck` is a bit trickier/requires some library work.  I have some initial ideas for inspecting stdout and such using `dup`, but we may want to do that sort of "stdout" inspection at the unit testing frameworking level when we build that out in the coming months.

PR #2088, opened 2024-03-30
[mojo-stdlib] Added move constructor to Atomic

### LINKED ISSUE #2026: [mojo-stdlib] Add move constructor to `Atomic`
### Review Mojo's priorities

### What is your request?

Add a `moveinit` function to the `Atomic` type.  Currently, people do workarounds such as calling `atomic.fetch_add(0)` in a user-defined type's `moveinit` when trying to move a data member that is an `Atomic` type.

This will help simplify/clarify intent for code trying to move from an atomic, such as a class trying to define a `moveinit` function that has a data member that is an `Atomic` type.

### Any other details?

Please explicitly conform to the `Movable` trait once `moveinit` is implemented for `Atomic`.

StandinKP:
Fixes #2026

### DIFF
--- a/stdlib/test/os/test_atomic.mojo
+++ b/stdlib/test/os/test_atomic.mojo
@@ -88,7 +88,7 @@ fn test_atomic_floating_point():
     print(atom.value)
 
 
-fn test_atomic_move_constructor():
+def test_atomic_move_constructor():
     var atom: Atomic[DType.index] = 3
     var atom2 = atom^
     assert_equal(atom2.value, 3)
@@ -106,7 +106,7 @@ fn test_atomic_move_constructor():
     assert_equal(atom2.value, 0)
 
 
-fn main():
+def main():
     test_atomic()
     test_atomic_floating_point()
     test_atomic_move_constructor()
--- a/stdlib/test/os/test_atomic.mojo
+++ b/stdlib/test/os/test_atomic.mojo
@@ -13,6 +13,7 @@
 # RUN: %mojo -debug-level full %s | FileCheck %s
 
 from os.atomic import Atomic
+from testing import assert_equal
 
 
 # CHECK-LABEL: test_atomic
@@ -52,7 +53,7 @@ fn test_atomic():
 
 
 # CHECK-LABEL: test_atomic_floating_point
-fn test_atomic_floating_point__():
+fn test_atomic_floating_point():
     print("== test_atomic_floating_point")
 
     var atom: Atomic[DType.float32] = Float32(3.0)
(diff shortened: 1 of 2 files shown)


LJ-9801:
Can you please add test cases?

StandinKP:
> Can you please add test cases?

Added test for testing the move. I have used Filecheck for now as this will be tackled in another issue (#2024)
@LJ-9801 lmk if any other comments

stdlib/test/os/test_atomic.mojo
```diff
@@ -52,7 +52,7 @@ fn test_atomic():
 
 
 # CHECK-LABEL: test_atomic_floating_point
-fn test_atomic_floating_poInt__():
+fn test_atomic_floating_point__():
```
JoeLoser, line 55: **Suggestion** Pre-existing, but do you mind dropping (removing) the two trailing underscores in this function name while you're here?

stdlib/test/os/test_atomic.mojo
```diff
@@ -87,6 +87,44 @@ fn test_atomic_floating_poInt__():
     print(atom.value)
 
 
+# CHECK-LABEL: test_atomic_move_constructor
```
JoeLoser, line 90: **Suggestion** I saw your comment in the commit message, but it's fine to use `assert_equal` and friends even in a test file that pipes its output to `FileCheck`.  So, I'd recommend not using `FileCheck` here and just using `assert_equal` and such from the `testing` module if you don't mind.  

This isn't blocking this PR, but would be a good habit to not create more `FileCheck` usages that we want to remove soon.

JoeLoser:
Looks good — just left a minor comment about switching the test to use `assert_equal` and such instead of `FileCheck` if you're up for it.  If not, no biggie, we can merge and refactor in a follow-on.

stdlib/test/os/test_atomic.mojo
StandinKP: Sure. Done

stdlib/test/os/test_atomic.mojo
StandinKP: Sure I have updated to use assert_equal

stdlib/test/os/test_atomic.mojo
```diff
@@ -87,6 +88,25 @@ fn test_atomic_floating_poInt__():
     print(atom.value)
 
 
+fn test_atomic_move_constructor():
```
JoeLoser, line 91: This needs to be a `def` since `assert_equal` may raise.  Ditto with changing `fn main():` to `def main():`.  After that, LGTM!

StandinKP: Done

JoeLoser (approved):

PR #2089, opened 2024-03-30
[mojo-stdlib] Replace `register_passable` `__init__ -> Self` usages in `stdlib/src/collections`

MertCelik:
Replaces `register_passable __init__ -> Self` usages in `stdlib/src/collections`. 2 of 6 PRs for https://github.com/modularml/mojo/issues/2037

- [x] Replace `/builtin` #2075
- [x] Replace `/collections` #2089
- [ ] Replace `/memory`
- [ ] Replace `/os`
- [ ] Replace `/utils`
- [ ] Replace misc

PR #2075, opened 2024-03-29
[mojo-stdlib] Replace `register_passable __init__ -> Self` usages in `stdlib/src/builtin`

MertCelik:
Replaces `register_passable __init__ -> Self` usages in `stdlib/src/builtin`. First of 6 PRs for https://github.com/modularml/mojo/issues/2037

- [x] Replace `/builtin` #2075
- [ ] Replace `/collections`
- [ ] Replace `/memory`
- [ ] Replace `/os`
- [ ] Replace `/utils`
- [ ] Replace misc

PR #2090, opened 2024-03-30
[mojo-stdlib] Replace `register_passable` `__init__ -> Self` usages in `stdlib/src/memory`

MertCelik:
Replaces `register_passable __init__ -> Self` usages in `stdlib/src/memory`. 3 of 6 PRs for https://github.com/modularml/mojo/issues/2037

- [x] Replace `/builtin` #2075
- [x] Replace `/collections` #2089
- [x] Replace `/memory` #2090
- [ ] Replace `/os`
- [ ] Replace `/utils`
- [ ] Replace misc


stdlib/src/memory/anypointer.mojo
```diff
@@ -42,15 +42,13 @@ struct AnyPointer[T: Movable](
...
+    fn __init__(inout self):
         """Create a null pointer.
 
         Returns:
             A null pointer.
         """
-        return Self {
-            value: __mlir_attr[`#interp.pointer<0> : `, Self.pointer_type]
-        }
+        self.value = __mlir_attr[`#interp.pointer<0> : `, Self.pointer_type]
```
StandinKP, line 51: This should be `self.pointer_type` instead of `Self.pointer_type`?

JoeLoser (commented):
Looks good, but I think you'll just need to fix up the formatting.  This can be done via `mojo format -l 80 stdlib/src/memory/anypointer.mojo` in your case.

stdlib/src/memory/anypointer.mojo
soraros: Why is that so? `Self.pointer_type` already carries `T`.

JoeLoser (approved):

PR #2096, opened 2024-03-30
[mojo-stdlib] Add `Bytes` struct, not public yet

gabrieldemarmiesse:
We add here a Bytes stuct, which is similar to python's `bytes`: https://docs.python.org/3/library/stdtypes.html#bytes-objects. One big difference is that it's mutable. 

We could also call it `BytesArray` since Python's `bytesarray` is like `bytes` but mutable: https://docs.python.org/3/library/stdtypes.html#bytearray-objects. 

I'm not sure we need the immutability for bytes in Mojo. So if we have only one of `bytes` and `bytesarray`, let's call it `Bytes`.

We can merge this PR without any additional features, they'll come later on. This struct should not be public and documented yet.

When more features are available for `Bytes`, we can start using it in `Path.read_bytes()` for example and make it public.

### DIFF
--- a/stdlib/src/builtin/bytes.mojo
+++ b/stdlib/src/builtin/bytes.mojo
@@ -1,0 +1,110 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Defines the Bytes type.
+
+The Bytes type is the equalivalent of the bytes type in Python. It is an
+array of values between 0 and 255. It is used to represent binary data.
+
+This type cannot be used without imports yet because it's not ready for a wider audience.
+"""
+from collections import Optional
+
+
+struct Bytes(Sized, CollectionElement):
+    """A mutable sequence of bytes. Behaves like the python version.
+
+    Note that `some_bytes[i]` returns an UInt8.
+    some_bytes *= 2 modifies the sequence in-place. Same with +=.
+
+    Also `__setitem__` is available, meaning you can do `some_bytes[7] = 105` or
+    even `some_bytes[7] = some_other_bytes` (the latter must be only one byte long).
+
+    You can create bytes from a list of UInt8 values:
+    ```mojo
+    var some_bytes = Bytes(List[UInt8](1, 2, 3, 4))
+    print(some_bytes)
+    # b'\x01\x02\x03\x04'
+    ```
+
+    or you can create bytes, set to 0, by specifying the size:
+    ```mojo
+    var some_bytes = Bytes(4)
+    print(some_bytes)
+    # b'\x00\x00\x00\x00'
+    ```
+
+    An empty constructor means a sequence of 0 bytes.
+    ```mojo
+    var some_bytes = Bytes()
+    print(some_bytes)
+    # b''
+    ```
+    """
+
+    alias _storage_type = List[UInt8]
+
+    var _data: Self._storage_type
+
+    fn __init__(inout self, owned data: Self._storage_type, /):
+        """Creates a Bytes object from a list of UInt8 values.
+
+        Args:
+            data: The list of UInt8 values to create the Bytes object from. Positional-only.
+        """
+        self._data = data^
+
+    fn __init__(inout self, size: Int = 0, /):
+        """Creates a Bytes object of a given size, filled with 0s.
+
+        Args:
+            size: The size of the Bytes object. Defaults to 0.
+        """
+        self.__init__(size, capacity=size)
+
+    fn __init__(inout self, size: Int, /, *, capacity: Int):
+        """Creates a Bytes object of a given size, filled with 0s.
+
+        Args:
+            size: The size of the Bytes object. Defaults to 0.
+            capacity: The capacity of the underlying `List[UInt8]`.
+        """
+        self._data = Self._storage_type(capacity=capacity)
+        for i in range(size):
+            self._data.append(0)
+
+    fn __len__(self) -> Int:
+        """Returns the number of bytes present in the `Bytes` object."""
+        return len(self._data)
+
+    fn __getitem__(self, index: Int) -> UInt8:
+        """Returns the byte at the given index.
+
+        The returned value is an UInt8 (so in the range 0-255).
+        """
+        return self._data[index]
+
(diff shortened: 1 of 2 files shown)


soraros:
-1 before a proposal.

What's the benefit of this over simply `List[UInt8]`? Python has to provide wrapper type because high level types can't be composed in performance preserving ways (`list[byte]` won't work). We don't have that problem in Mojo.

gabrieldemarmiesse:
I respectfully disagree, I very much like the #2099 proposal and I have no doubt it will be accepted. This PR goes in this direction as the interal data is a `List[Uint8]`. This pull request provides a struct that is not used yet throughout the codebase, so should not bring git conflicts.

There are two other major benefits of using a specific struct to represent bytes instead of using an array of UInt8.
* Strong typing: It allows a library author to say that a function does not expects just list of number but actual bytes. This can prevent type errors and should be extremely common for serialization libraries.
* Specialized methods: Python provides many methods to work with `bytes`. All methods that work on `str` work on `bytes`,  making bytes easy to work with when serializing and deserializing bytes (`.removeprefix()`, `.startswith()`, `.join()`, ...). Python also provide [methods to convert between `str` and `bytes`](https://docs.python.org/3/howto/unicode.html#converting-to-bytes): 

Overall, having a specific struct to represent plain bytes makes a lot of sense, both in Mojo and in python. Beyond the performance, python users very much like `bytes`, same as Rust users like the [Bytes](https://docs.rs/bytes/latest/bytes/) crate.

lsh:
I think if you want to use this type with methods like `removeprefix`, then the representation might be better suited as a copy on write data structure like:
```mojo
struct Bytes[is_mutable: __mlir_type.i1, lifetime: AnyLifetime[is_mutable].type]:
    # Reference of Buffer since Buffer doesn't have a lifetime
    alias type = Variant[List[UInt8], Reference[Buffer[DType.uint8], is_mutable, lifetime]]
```
I can't think of any language that requires allocation for the same method ([Rust](https://doc.rust-lang.org/std/primitive.slice.html#method.strip_prefix), [Zig](https://ziglang.org/documentation/master/std/#std.mem.trimLeft)). If this data structure can only be used in an owned capacity, that will massively limit use cases.

soraros:
> -1 before *a more detailed design proposal*.

Seems you conveniently ignored this point. The benefit you mentioned is true only when `Bytes` is more thought out and provides more features. I like the Strong typing idea a lot (even if we keep `List[UInt8]`, I will most probably use a type alias). However, for the Mojo as of now, I can't even imagine using `Bytes` inside stdlib: people will have to access the internal buffer and use the list.

> - Specialized methods: Python provides many methods to work with bytes. All methods that work on str work on bytes, making bytes easy to work with when serializing and deserializing bytes (`.removeprefix()`, `.startswith()`, `.join()`, ...). Python also provide [methods to convert between str and bytes](https://docs.python.org/3/howto/unicode.html#converting-to-bytes):

This is true (in a bad way) because of Python's own limitations. Language like Rust or Swift typically provide these features for the generic `List[T]` when `T: Equitable` for instance. Converting to string is trivially `String(some_bytes)` if we keep `List[UInt8]` when #2099 happens.

JoeLoser:
Thanks for the PR.  We're going to discuss https://github.com/modularml/mojo/pull/2099 this week and then we'll circle back here on what we want to do with this `Bytes` struct.  If you have more fleshed out ideas on what you're thinking for `Bytes`, I would love to see a proposal — especially if it can address some of @soraros's concerns he brought up.

gabrieldemarmiesse:
Thanks @JoeLoser, I'm a bit afraid of making a proposal that would end up in a lot of bikeshedding and just repeating what was said in https://github.com/modularml/mojo/issues/2113 and https://github.com/modularml/mojo/issues/2164 . I do not think that would be very productive. Let's first sort out those issues and decide if we want Mojo to (long term) be able to run any Python code (at least, that's how I read [this part of the "Why Mojo"](https://docs.modular.com/mojo/why-mojo#a-member-of-the-python-family)), then we can decide what to do about this PR. Would that work for you?

JoeLoser:
> Thanks @JoeLoser, I'm a bit afraid of making a proposal that would end up in an endless debate and just repeating what was said in #2113 and #2164 . I do not think that would be very productive. Let's first sort out those issues and decide if we want Mojo to (long term) be able to run any Python code (at least, that's how I read [this part of the "Why Mojo"](https://docs.modular.com/mojo/why-mojo#a-member-of-the-python-family)), then we can decide what to do about this PR. Would that work for you?

Sounds good to me!  Thanks for pushing on this.  I'm spinning up a working group to discuss design things like this internally with the standard library team, and I'll make sure this is included when we hopefully meet for the first time either late this week or early next.  In general, my early reactions are that we definitely need a `Bytes` type of some sort FYI.

When we sit down to discuss this, I'll take an action item to circle back here and let you know.  My intent of the working group is to be similar to the existing Mojo language feature request working group where we (mostly) discuss compiler/language things; this would be the standard library counterpart.

JoeLoser:
Hey @gabrieldemarmiesse, thank you so much for seeding this discussion with the `Bytes` type provided.  As a team, we met earlier this week to discuss this and went back and forth on this on a couple of different things.  In general, the overall takeaway was a weak +1.  At this time, the Stdlib team would recommend holding off on this `Bytes` type **for now** as it's currently proposed. 

Our ask to you:
- Can you please provide example use cases for the `Bytes` type?  What additional methods do you envision adding to make this more useful over `List[UInt8]`? 

Here are some takeaways from our discussion:
- We've accepted https://github.com/modularml/mojo/pull/2099 (which we discussed right before this PR). `List[UInt8]` can get a lot of mileage right now before we need a `Bytes` type.
- As @lsh points out, we may want to go down a path where we define a “Bytes” type with mutability being part of the type.  Then, in our Mojo -> Python interop layer (details still TBD), we could infer mutability to encode this property correctly to maintain parity with the existing types of `bytes` and `bytesarray`.  In Mojo, we don't need separate types to conceptually represent mutability.
- There is an open discussion that we don't have an answer to (haven't asked the language subteam) around how this may work with binary literals like in Python.

The general summary is that because `bytes` in Python is more tightly tied to language semantics (as the type of `b”...”` string literals) than the average standard library type, we want to proceed a bit more cautiously here. Mojo doesn’t have binary string literals yet, so there wouldn’t be a clear use for this type yet, and when Mojo does support that feature we may realize we want to go with a slightly different representation. (For example, perhaps something built on `Reference`).

gabrieldemarmiesse:
Sure, as with timedelta, there is no rush to take a decision. I believe that as the stdlib matures, we'll see more and more places where the methods provided by python's bytes would be useful :)

JoeLoser:
> Sure, as with timedelta, there is no rush to take a decision. I believe that as the stdlib matures, we'll see more and more places where the methods provided by python's bytes would be useful :)
> 
> I'll put this PR in draft to avoid futher discussion until the stdlib and language has matured a bit more

Thanks for understanding and pushing us to think about these things.  Like you said, it's just a bit early would be the key takeaway right now.

gabrieldemarmiesse:
I'll close this PR and we'll reopen it when it's the right time for the stdlib to expand :)

PR #2099, opened 2024-03-31
[Proposal] Standardise the representation of byte sequence as a sequence of unsigned 8 bit integers

mzaks:
See the byte-as-uint8.md file

### DIFF
--- a/proposals/byte-as-uint8.md
+++ b/proposals/byte-as-uint8.md
@@ -6,7 +6,7 @@ Most noticeable example is the underlying data of string types `String`, `String
 # Motivation
 
 Logically a byte is an integer value between `0` and `255`. Lots of algorithms make use of arithmetics ground by this assumption.
-A signed 8 bit integer on the contrary represents values between `-128` and `127`. This introduces very settle bugs, when an algorithm written for unsigned 8 bit integer is used on a signed 8 bit integer.
+A signed 8 bit integer on the contrary represents values between `-128` and `127`. This introduces very subtle bugs, when an algorithm written for unsigned 8 bit integer is used on a signed 8 bit integer.
 
 Another motivation for this change is that Mojo aims to be familiar to Python users. Those Python users are familiar with the `bytes` class, which itself is working with values between `0` and `255`, not values between `-128` and `127`.
 
--- a/proposals/byte-as-uint8.md
+++ b/proposals/byte-as-uint8.md
@@ -1,0 +1,25 @@
+# Standardise the representation of byte sequence as a sequence of unsigned 8 bit integers
+
+At this point in time, a sequence of bytes is often represented as a sequence of signed 8 bit integers in Mojo standard library. 
+Most noticeable example is the underlying data of string types `String`, `StringLiteral`, `StringRef` and `InlinedString`, but also APIs like for example the hash function `fn hash(bytes: DTypePointer[DType.int8], n: Int) -> Int:`.
+
+# Motivation
+
+Logically a byte is an integer value between `0` and `255`. Lots of algorithms make use of arithmetics ground by this assumption.
+A signed 8 bit integer on the contrary represents values between `-128` and `127`. This introduces very settle bugs, when an algorithm written for unsigned 8 bit integer is used on a signed 8 bit integer.
+
+Another motivation for this change is that Mojo aims to be familiar to Python users. Those Python users are familiar with the `bytes` class, which itself is working with values between `0` and `255`, not values between `-128` and `127`.
+
+## Examples:
+
+### Division:
+A value `-4` represented as `Int8` has the same bit pattern as value `252` represented as `UInt8`.
+`-4 // 4` equals to `-1` (`bx11111111`), where `252 // 4` equals to `63` (`bx00111111`) as we can see the bit patterns are different.
+
+### Bit shift:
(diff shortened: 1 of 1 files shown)


proposals/byte-as-uint8.md
```diff
@@ -0,0 +1,23 @@
+# Standardise the representation of byte sequence as a sequence of unsigned 8 bit integers
+
+At this point in time, a sequence of bytes is often represented as a sequence of signed 8 bit integers in Mojo standard library. 
+Most noticeable example is the underlying data of string types `String`, `StringLiteral`, `StringRef` and `InlinedString`, but also APIs like for example the hash function `fn hash(bytes: DTypePointer[DType.int8], n: Int) -> Int:`.
+
+# Motivation
+
+Logically a byte is an integer value between `0` and `255`. Lots of algorithms make use of arithmetics ground by this assumption.
+A signed 8 bit integer on the contrary represents values between `-128` and `127`. Which introduces very settle bugs, when an algorithm written for unsigned 8 bit integer is used on a signed 8 bit integer.
```
VascoSch92, line 9:
```suggestion
A signed 8 bit integer on the contrary represents values between `-128` and `127`. This introduces very settle bugs, when an algorithm written for unsigned 8 bit integer is used on a signed 8 bit integer.
```

proposals/byte-as-uint8.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Examples:
```
gabrieldemarmiesse, line 11:
```suggestion

Another motivation for this change is that Mojo aims to be familiar to Python users. Those Python users are familiar with the `bytes` class, which itself is working with values between `0` and `255`, not values between `-128` and `127`.

## Examples:
```

JoeLoser (commented):
Thank you very much for going through the proposal process for this!  We're going to discuss this internally within the team this coming week and we'll circle back here.  Just wanted to give you a quick update here to let you know we haven't forgotten or missed this.

In general, I'm +1 for this — we're just trying to understand some of the existing history for the rationale why `Int8` was chosen for `String` to begin with to help guide our decision on how we want to move forward.

proposals/byte-as-uint8.md
```diff
@@ -0,0 +1,25 @@
+# Standardise the representation of byte sequence as a sequence of unsigned 8 bit integers
+
+At this point in time, a sequence of bytes is often represented as a sequence of signed 8 bit integers in Mojo standard library. 
+Most noticeable example is the underlying data of string types `String`, `StringLiteral`, `StringRef` and `InlinedString`, but also APIs like for example the hash function `fn hash(bytes: DTypePointer[DType.int8], n: Int) -> Int:`.
+
+# Motivation
+
+Logically a byte is an integer value between `0` and `255`. Lots of algorithms make use of arithmetics ground by this assumption.
+A signed 8 bit integer on the contrary represents values between `-128` and `127`. This introduces very settle bugs, when an algorithm written for unsigned 8 bit integer is used on a signed 8 bit integer.
```
JoeLoser, line 9: **Suggestion** `s/settle/subtle`

JoeLoser:
Thanks so much for going through the proposal process!  We as a team really appreciate the thought and motivation presented, and are happy to accept this proposal!  🎉 🎉 🎉 

Internally, we went back and forth whether to add an `alias Byte = UInt8` and don't feel strongly either way when the technical implementation for this proposal happens.  Also, to be explicit about our existing commitments and priorities — we'd love if the community is interested in changing the `Int8` to `UInt8` in the open-source parts of the standard library, and we're happy to do the rest for the closed source modules when we import said PR.  Otherwise, this isn't on our immediate radar given pre-existing internal commitments.  

Regarding the alias part, since an alias is a weak type, it's only for readability sake/to denote we're conceptually working with "bytes" since you can't overload functions/APIs on weak types, and we don't yet have a strong `Bytes` type (and there's no need to couple this proposal with https://github.com/modularml/mojo/pull/2096 for example).

Congrats again on the first standard library proposal to be accepted!  Now that we have a process internally for discussing these, we expect to be able to triage and discuss these in a much more timely fashion going forward.  Thanks for your patience with this one.

JoeLoser (approved):

proposals/byte-as-uint8.md
JoeLoser: @mzaks after this small typo is fixed, happy to land this 👍

gabrieldemarmiesse:
I made #2317 to track the progress of the implementation

mzaks:
Hm, why is the first commit not verified? @JoeLoser do you have an idea?

JoeLoser:
> Hm, why is the first commit not verified? @JoeLoser do you have an idea?

Looks like one was signed with your GPG key and one was not.  The GPG signing isn't required, so I went ahead and merged this.  We only require a sign-off using `git commit -s` to sign your commits (which is different than verifying your commits).  In any case, the final squashed and merged commit as seen [here](https://github.com/modularml/mojo/commit/46889965656de7d4949f054e483335883a1f83e2) shows it as GPG-verified, so that's good if that's what you were wanting. 😄

JoeLoser:
FYI I realized this was targeting the main branch and not nightly, but snuck in since it wasn't rebased where we have CI checking it targets the nightly branch. I'm happy to open a PR for this to target the nightly branch as well later today unless you beat me to it.

The main branch will automatically get synced each time we make a new MAX release and it'll pick up all the things in the nightly branch effectively that made it into the stable MAX release.

PR #2110, opened 2024-04-01
[stdlib] Check compiler versions before building

JoeLoser:
There's often breaking changes in the compiler that mandate an update to the latest `nightly/mojo` compiler in order to work with the latest standard library.  Currently, if a user installs a newer `nightly/mojo` compiler but hasn't updated the library source code, they'll get pages of bad errors depending on the breaking change, often "cryptic" MLIR errors.  This is not a great user experience, and does not tell contributors what they can/should do.

To improve this user experience, check the current installed compiler version in the script when building the standard library.  If there is a mismatch between the expected compiler version (stored in a file in the repo) and the installed compiler, give a good error message telling the user to update their `nightly/mojo` compiler.

Note:
- An alternative approach is to query to find the "latest compiler version released" and check against that, but that has the caveat that just because internally we publish a new `nightly/mojo` package, if someone hasn't updated their local copy of the standard library source code, they shouldn't be prevented from building just because something out-of-band changed/got updated.  This approach keeps the local flow working until they update their copy of the library source code.
- We'll probably want to rewrite these bash scripts into Python for a bit better maintenance soon.


ConnorGray (approved):

River707 (approved):
LG, we can come up with something better over time.

stdlib/scripts/build-stdlib.sh
```diff
@@ -19,6 +19,17 @@ REPO_ROOT="${SCRIPT_DIR}"/../..
...
 
+ACTUAL_COMPILER_VERSION=$(mojo --version | tr " " "\n" | sed -n 2p)
+EXPECTED_COMPILER_VERSION=$(<"${REPO_ROOT}"/stdlib/COMPATIBLE_COMPILER_VERSION)
+
+if [ "${EXPECTED_COMPILER_VERSION}" != "${ACTUAL_COMPILER_VERSION}" ]; then
+  echo "Mismatch in compiler versions! Cannot build the standard library."
+  echo "Expected compiler version: ${EXPECTED_COMPILER_VERSION}"
+  echo "Current installed compiler version: ${ACTUAL_COMPILER_VERSION}"
+  echo "Please run modular update nightly/mojo to get the latest compiler."
+  exit 1
```
ConnorGray, line 30: **Suggestion** I wonder if this should only be a warning, or if we should include a CLI flag to disable this check. Doing an exact match means this check may have a high false positive rate for incompatibility, which could be a burden as much as it helps.

JoeLoser: I'd like to do a "min compiler version compatibility" check, but that pushes complexity internally for us to _know_ when there was a breaking change.  We don't have that ability easily (automated) right now other than knowing our internal CI when we test with latest changes internally for the unreleased Mojo SDK combined with the public `mojo` nightly changes fail the CI when you install `nightly/mojo` (the previous released `nightly/mojo` pedantically, by definition).  We can make that better over time.  This whole PR is just a short-term "stop-gap" solution to improve the developer experience in my mind.

gabrieldemarmiesse: In the future, the issue will go away anywway no? If the mojo compiler and stdlib are both open source and in the same repo, everything will be in sync and it's going to be rainbows and butterflies :) That's ok if it's a bit rough until then

JoeLoser: Right, the issue would go away if the compiler and library were both open-source and built in tandem, if you will.  Right now, as I was publishing a new `nightly/mojo` package, there were some breaking changes in the compiler.  The user experience is pretty poor: pages full of cryptic MLIR errors.  😄

gabrieldemarmiesse: Since we just want to improve the error message, why do we need the "exit 1"? Would it be possible to just add a big fat warning before or after running the tests? With this, we get a good error message, but it doesn't prevent the tests from actually running if, by luck, everything is compatible, even with a version mismatch (meaning no breaking changes in the compiler)

JoeLoser: Given that this is a short-term solution, I'm going to stick with the `exit 1`.  You won't even be able to _build_ the new standard library code with the old compiler — much less run the tests.  I think hard-erroring for now is probably going to cause other developers less confusion than trying to understand why they're getting long error messages when trying to build the standard library itself.  These long error messages would make the proposed warning message from this `build-stdlib.sh` script itself, sort of defeating the purpose.

Issue #2133, opened 2024-04-02
[Enhancement] List.resize takes a T even when you are shrinking the list


bserletfun:
### Bug description

I think it would be cleaner to have List.grow that takes a T and List.shrink that does not.

### Steps to reproduce

- Include relevant code snippet or link to code that did not work as expected.
- If applicable, add screenshots to help explain the problem.
- If using the Playground, name the pre-existing notebook that failed and the steps that led to failure.
- Include anything else that might help us debug the issue.

### System information

```shell
Latest mojo
```

soraros:
Not a bug. Please write a feature request, and fill in all the fields.

bserletfun:
I think there's a bug: I can't provide a T when I'm shrinking the list...

mikowals:
This is added in nightly branch. See #2140.

ConnorGray:
Closing as fixed by #2140, thanks!

PR #2140, opened 2024-04-02
[mojo-stdlib] feature: Resize List smaller with single arg

mikowals:
Implements #2133

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -288,8 +288,8 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         debug_assert(
             new_size <= self.size,
             (
-                "New size must be smaller than or equal to current size when no new"
-                " value is provided."
+                "New size must be smaller than or equal to current size when no"
+                " new value is provided."
             ),
         )
         for i in range(new_size, self.size):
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/list.mojo
```diff
@@ -265,12 +265,37 @@ struct List[T: CollectionElement](CollectionElement, Sized):
...
+                _ = (self.data + i).take_value()
+            for i in range(self.size, new_size):
+                (self.data + i).emplace_value(value)
+            self.size = new_size
+
+    @always_inline
+    fn resize(inout self, new_size: Int):
+        """Resizes the list to the given new size.
+
+        With no new value provided, the new size must be smaller than the current one. Elements at the end
```
JoeLoser, line 282: **Quibble** Here, we say "smaller than" but the `debug_assert` is "smaller than or equal".  I think we just want to fixup this doc string to say "smaller than or equal to".

JoeLoser (approved):
Great, thank you for your contribution!  Just one minor wording on the doc string comment.  Once that is fixed, I'm happy to merge this. 🎉

stdlib/src/collections/list.mojo
mikowals: done.

JoeLoser: Thanks for the quick fix — merged!  Appreciate the contribution 🎉

PR #2146, opened 2024-04-02
[mojo-stdlib] Replace manual refcounted `List` with `Arc`

jayzhan211:
Parts of https://github.com/modularml/mojo/issues/2030

I try replacing List as the first step.

### DIFF
--- a/stdlib/src/builtin/object.mojo
+++ b/stdlib/src/builtin/object.mojo
@@ -21,6 +21,7 @@ from sys.intrinsics import _mlirtype_is_eq
 
 from memory import memcmp, memcpy
 from memory.unsafe import DTypePointer, Pointer
+from memory._arc import Arc
 
 from utils import StringRef
 from utils.loop import unroll
@@ -92,6 +93,26 @@ struct _RefCountedList:
     var impl: List[_ObjectImpl]
     """The list value."""
 
+struct _List(Movable):
+    var impl: List[_ObjectImpl]
+    """The list value."""
+
+    fn __init__(inout self):
+        self.impl = List[_ObjectImpl]()
+
+    fn __moveinit__(inout self, owned existing: Self):
+        self.impl = existing.impl
+        existing.impl = List[_ObjectImpl]()
+
+    fn __del__(owned self):
+        for i in range(len(self.impl)):
+            self.impl[i].destroy()
+
+struct _RCList:
+    var impl: Arc[_List]
+
+    fn __init__(inout self):
+        self.impl = Arc[_List](_List())
 
 @register_passable("trivial")
 struct _RefCountedListRef:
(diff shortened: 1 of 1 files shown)


stdlib/src/builtin/object.mojo
```diff
@@ -107,20 +120,11 @@ struct _RefCountedListRef:
 
     @always_inline
     fn copy(self) -> Self:
-        _ = self.lst.bitcast[_RefCountedList]()[].refcount.fetch_add(1)
+        _ = self.lst.bitcast[_RefCountedList]()[].impl
```
jayzhan211, line 123: I think this trigger `__init__` which do `increment`

stdlib/src/builtin/object.mojo
```diff
@@ -107,20 +120,11 @@ struct _RefCountedListRef:
...
-        var prev = ptr[].refcount.fetch_sub(1)
-        if prev != 1:
-            return
-
-        # Run the destructor on the list elements and then destroy the list.
-        var list = __get_address_as_owned_value(ptr.address).impl
-        for i in range(len(list)):
-            list[i].destroy()
-        ptr.free()
+        var ptr = self.lst.bitcast[_RefCountedList]()[].impl
```
jayzhan211, line 127: `__del__` is called after the end of this function, so do `decrement`

stdlib/src/builtin/object.mojo
```diff
@@ -75,6 +76,22 @@ struct _ImmutableString:
             return 0
         return -1 if self.length < rhs.length else 1
 
+struct _List(Movable):
```
jayzhan211, line 79: `List[_ObjectImpl]()` itself is not moveable, so I create another movable struct

jayzhan211:
Recommend to review [6476438](https://github.com/modularml/mojo/pull/2146/commits/647643875dc945bb1824c31bce56ed4aedfa4bb1), the formatter changes many unrelated code

stdlib/src/builtin/object.mojo
JoeLoser: **Question** Is this needed?  `_ObjectImpl` should implicitly conform to `Movable` trait since it is `register_passable`.  That's why the pre-existing code works fine with `List[_ObjectImpl]` today, right?

I think for this PR, it should work if we can just augment `_RefCountedList` struct to use an `Arc` internally and remove the internal `refcount` data member.  Does that not work for some reason?

jayzhan211: I retry it and it works. I previously saw the error saying that `List[_ObjectImpl]` is not movable, which is why I wrapped it into another movable struct. Maybe it is fixed in the new version 🤔

stdlib/src/builtin/object.mojo
JoeLoser: If I understand correctly, you're saying this will invoke `_RefCountedList.__init__` which then calls the `Arc` constructor which then eventually bumps the inner data refcount by `1`, preserving the same behavior as before.

stdlib/src/builtin/object.mojo
JoeLoser: **Question** Will this also run the destructors for the list elements themselves in the `Arc` destructor, or does it just delete the inner pointer to the list of elements held by `Arc`?

JoeLoser (commented):
Looks good! Just had a clarifying question or two before we merge.  Thanks for pushing on this!

stdlib/src/builtin/object.mojo
jayzhan211: I think it calls `__del__`  in Arc, and inner.free() is called. I'm not sure what exactly does free of List do. But I guess it calls `__del__` which `free` the list of elements.

jayzhan211: I try print out and ensure that inner.free() is called, and so does `__del__`. But there is one weird thing, I see tons of printout in List `__del__`, not sure why does it been called so many times.

jayzhan211: btw I run the test with

```mojo
from memory._arc import Arc

fn test_123():
    print("== test_123")
    var a = Arc[List[Int32]](List[Int32](1, 2, 3))
    var b = a
    b.set(List[Int32](4, 5, 6))

    b[].append(7)

    # CHECK: 4
    print(a[][0])
    # CHECK: 4
    print(b[][0])
    # CHECK: 4
    print(len(b[]))

fn main():
    test_123()
```

jayzhan211: The crazy priintouts is also reproduced without arc, so it seems to be another issue, nothing about Arc. If it is unexpected we can file an issue to track it.

```mojo
fn test_456():
    print("== test_456")
    var a = List[Int32](1, 2, 3)
    # CHECK: 1
    print(a[0])
```

JoeLoser (approved):
Thanks for the contribution and patience with this one!  First one is always the toughest as we rework these types to use `Arc`.  😄

stdlib/src/builtin/object.mojo
JoeLoser: Interesting that the destructor (even on top-of-tree here, without Arc) seems to be infinitely recursing.  That seems like a compiler bug of some sorts, but I don't see why yet.

jayzhan211:
@JoeLoser Thanks for your review!

PR #2148, opened 2024-04-02
[mojo-stdlib] Add list.insert(index, value) API to stdlib

whym1here:
This pr serves to add `list.insert(index, value)` API to the stdlib as requested by issue https://github.com/modularml/mojo/issues/2134.

This API should handle the same functionality as the [python `list.insert`](https://docs.python.org/3/tutorial/datastructures.html) and adds tests for the same.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -185,6 +185,36 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         self.size += 1
 
     @always_inline
+    fn insert(inout self, i: Int, owned value :T):
+        """Insert a value to the list at the given index.
+        `a.insert(len(a), value)` is equivalent to `a.append(value)`.
+
+        Args:
+            i: The index for the value.
+            value: The value to insert.
+        """
+        debug_assert(-self.size <= i <= self.size, "insert index out of range")
+        
+        var normalized_idx = i
+        if i < 0:
+            normalized_idx += len(self)
+
+        var earlier_idx = len(self)
+        var later_idx = len(self) - 1
+        self.append(value^)
+        
+        for _ in range(i, len(self) - 1): 
+            var earlier_ptr = self.data + earlier_idx
+            var later_ptr = self.data + later_idx
+
+            var tmp = earlier_ptr.take_value()
+            later_ptr.move_into(earlier_ptr)
+            later_ptr.emplace_value(tmp^)
+
+            earlier_idx -= 1
+            later_idx -= 1
+
+    @always_inline
     fn extend(inout self, owned other: List[T]):
         """Extends this list by consuming the elements of `other`.
 
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -185,7 +185,7 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         self.size += 1
 
     @always_inline
-    fn insert(inout self, i: Int, owned value :T):
+    fn insert(inout self, i: Int, owned value: T):
         """Inserts a value to the list at the given index.
         `a.insert(len(a), value)` is equivalent to `a.append(value)`.
 
@@ -194,7 +194,7 @@ struct List[T: CollectionElement](CollectionElement, Sized):
             value: The value to insert.
         """
         debug_assert(i <= self.size, "insert index out of range")
-        
+
         var normalized_idx = i
         if i < 0:
             normalized_idx = _max(0, len(self) + i)
@@ -202,8 +202,8 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         var earlier_idx = len(self)
         var later_idx = len(self) - 1
         self.append(value^)
-        
-        for _ in range(normalized_idx, len(self) - 1): 
+
+        for _ in range(normalized_idx, len(self) - 1):
             var earlier_ptr = self.data + earlier_idx
             var later_ptr = self.data + later_idx
 
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -186,24 +186,24 @@ struct List[T: CollectionElement](CollectionElement, Sized):
 
     @always_inline
     fn insert(inout self, i: Int, owned value :T):
-        """Insert a value to the list at the given index.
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/list.mojo
```diff
@@ -184,6 +184,36 @@ struct List[T: CollectionElement](CollectionElement, Sized):
...
+
+        Args:
+            i: The index for the value.
+            value: The value to insert.
+        """
+        debug_assert(-self.size <= i <= self.size, "insert index out of range")
+        
+        var normalized_idx = i
+        if i < 0:
+            normalized_idx += len(self)
```
Neroro64, line 200: Hi, I'm curious. What is the intention of this `normalized_idx`? 

Another, perhaps stupid question (to mojo team) is why we are allowing indices, sizes and capacity to have the `Int` where negative values are not valid?

Neroro64: Ah, for indices, is it for allowing syntax like `vec.insert(-2, foo)` where it inserts the value at position `len(vec)-2` ?

whym1here: Oh my bad it should use `normalized_index` instead of `i` in the `for` loop. Thanks for pointing it out.
Yeah It should behave similar to [`a.insert` in python](https://docs.python.org/3/tutorial/datastructures.html)
image

whym1here: Their is some ambiguity in python3 like this:
```python
a = []
a.insert(-2, 11)
print(a) # [11]
a = [1, 2, 3, 4]
a.insert(-10, 4)
print(a) # [4, 1, 2, 3, 4]
```

@Neroro64 , @JoeLoser  how should we handle -ve index in `insert`? python3 allow for something similar to above. 
how to handle the case when `len(a) == 0` and we `a.insert(-1, some_value)` ?

stdlib/src/collections/list.mojo
```diff
@@ -184,6 +184,36 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         (self.data + self.size).emplace_value(value^)
         self.size += 1
 
+    @always_inline
+    fn insert(inout self, i: Int, owned value :T):
+        """Insert a value to the list at the given index.
```
arthurevans, line 189:
```suggestion
        """Inserts a value to the list at the given index.
```

Minor adjustment for style consistency.

stdlib/src/collections/list.mojo
Neroro64: Sorry for late response. 
Personally, I don't like negative index because that doesn't really provide any value other than the syntactic sugar. Value types with proper "unit" would be safer and easier to work with, imo. 

But if we want the same behavior as python. I guess something like the following would do:
```python
if (index < 0):
    normalized_index = max(0, len(self) + index)
```

For python `list` any `insert(index, value)` where `index <= -len(self)` is equivalent to `insert(0, value)`

See implementation here: https://github.com/python/cpython/blob/c43f6a4dfaa1c1974141e2f7014a7f98ca3a3f93/Objects/listobject.c#L453

stdlib/src/collections/list.mojo
whym1here: Sure

stdlib/src/collections/list.mojo
whym1here: This should work. Thanks for the input.

JoeLoser (commented):
Looks good!  We'll need to rebase this PR/resolve conflicts and then add a test for the negative index case.

Currently, the PR is showing many commits from April 4th that already exist in the `nightly` branch.  If you use `git rebase -i upstream/nightly` where `upstream` is the corresponding git remote:

```
git remote -v
upstream        git@github.com:modularml/mojo.git (fetch)
upstream        git@github.com:modularml/mojo.git (push)
```

then you'll be able to see those commits not show up as duplicates in your PR.  Please let me know if you run into any issues and I'll be happy to help.

stdlib/src/collections/list.mojo
```diff
@@ -184,6 +184,36 @@ struct List[T: CollectionElement](CollectionElement, Sized):
...
+        """Inserts a value to the list at the given index.
+        `a.insert(len(a), value)` is equivalent to `a.append(value)`.
+
+        Args:
+            i: The index for the value.
+            value: The value to insert.
+        """
+        debug_assert(i <= self.size, "insert index out of range")
+        
+        var normalized_idx = i
```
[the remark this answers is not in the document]
whym1here:
```mojo
    @always_inline
    fn _normalize_index(inout self, index: Int) -> Int:
        return _max(0, index + len(self))
```

Something like this?

JoeLoser:
Looks like CI is failing due to formatting.  Do you mind running `mojo format -l 80 stdlib/src/collections/list.mojo stdlib/test/collections/test_list.mojo` to fix the formatting issues?  Thanks!

stdlib/src/collections/list.mojo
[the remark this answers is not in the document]
JoeLoser: Sure, feel free to open an issue for the refactor/consistency thing after we land this PR.  Seems like a nice little cleanup but no need to couple it with this PR's functionality addition.

stdlib/test/collections/test_list.mojo
```diff
@@ -296,6 +296,71 @@ def test_list_reverse_move_count():
...
+    vec.insert(len(vec), 1)
+    vec.insert(len(vec), 3)
+    vec.insert(1, 2)
+
+    assert_equal(len(vec), 3)
+    assert_equal(vec[0], 1)
+    assert_equal(vec[1], 2)
+    assert_equal(vec[2], 3)
+
+    vec.clear()
```
JoeLoser, line 314: **Suggestion** Instead of manually running `clear()` between the different parts of this function, it would read better to just use different variables, such as `v1 = List[Int]()`, `v2 = List[Int]()`, and so on.  What do you think?

JoeLoser (approved):
LGTM - just need to fix the formatting before we merge this.  Thank you for your contribution! 🚀

stdlib/test/collections/test_list.mojo
whym1here: Yeah seems like a good idea

PR #2158, opened 2024-04-03
[docs] Fix wrong `mojo format` command, add `-l 80`.

gabrieldemarmiesse:
The `mojo format` [command in the CI](https://github.com/modularml/mojo/blob/nightly/.github/workflows/standard_library_tests_and_examples.yml#L111) applies a limit of 80 chars with `-l 80`. This is not documented anywhere, resulting in broken formatting in pull requests. This PR fixes this issue.

### DIFF
--- a/stdlib/docs/style-guide.md
+++ b/stdlib/docs/style-guide.md
@@ -40,13 +40,14 @@ It adjusts indentation, spacing, and line breaks, making code more readable and
 consistent.
 
 ```bash
-> mojo format example.mojo
+> mojo format -l 80 example.mojo
 All done! ✨ 🍰 ✨
 1 file left unchanged.
 ```
 
 Unless otherwise noted, Mojo standard library code should follow the formatting
-produced by `mojo format`.
+produced by `mojo format -l 80`. Note that the default `mojo format` command is not enough.
+You need to specify that there is a limit of 80 characters per line with `-l 80`.
 
 #### Whitespace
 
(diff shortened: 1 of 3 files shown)


stdlib/docs/style-guide.md
```diff
@@ -40,13 +40,14 @@ It adjusts indentation, spacing, and line breaks, making code more readable and
...
 ```bash
-> mojo format example.mojo
+> mojo format -l 80 example.mojo
 All done! ✨ 🍰 ✨
 1 file left unchanged.
 ```
 
 Unless otherwise noted, Mojo standard library code should follow the formatting
-produced by `mojo format`.
+produced by `mojo format -l 80`. Note that the default `mojo format`command is not enough.
```
JoeLoser, line 49: **Quibble** Space after the backtick in `mojo format` part, please.

JoeLoser (commented):
Thank you for following up on this!  When I came across the incorrect behavior the other day, I had filed an internal bug for `mblack` (what `mojo format` calls), but I'll move that out to the public mojo repo here for tracking purposes.  It's confusing to everyone since the default help text for `mojo format --help` shows:

```
  -l, --line-length INTEGER       How many characters per line to allow.
                                  [default: 80]
```

but reading the code, the _actual default_ is `88` characters, which is why I changed CI to explicitly specify `-l 80` recently.  Thanks for fixing the docs to make this match reality.  In the near future, I expect us to change the default line length for `mojo format` _to actually be 80_.

CC: @modocache @River707

stdlib/docs/style-guide.md
gabrieldemarmiesse: Done

JoeLoser (approved):
Thank you!

PR #2163, opened 2024-04-03
[stdlib] Allow `is None` and `is not None` with `OptionalReg`

gabrieldemarmiesse:
Follow-up on https://github.com/modularml/mojo/pull/2082
 
Do you think that we should recommend `if a is None` instead of `if a`? It will be closer to what Python's PEP8 recommends.
If so, I can change in another PR the examples in the documentation to show this syntax first.

### DIFF
--- a/stdlib/src/collections/optional.mojo
+++ b/stdlib/src/collections/optional.mojo
@@ -312,6 +312,32 @@ struct OptionalReg[T: AnyRegType](Boolable):
         """
         return __mlir_op.`kgen.variant.take`[index = Int(0).value](self._value)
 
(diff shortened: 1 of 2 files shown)


gryznar:
In Python `if a is None` is not the same as `if a`. In second case it means: `if bool(a)`

soraros:
@gryznar Yea, but a) `a is not None` coincide with `bool(a)` and b) I think it's a better way to represent the "has a value" idea.

gryznar:
@soraros I agree, you are right

JoeLoser (approved):

PR #2181, opened 2024-04-04
[stdlib] Make Bool struct Intable

### LINKED ISSUE #2168: [Feature Request] Make `Bool` `Intable`
### Review Mojo's priorities

### What is your request?

As title.

I don't see reason to hold it off any longer.

### Any other details?

N/A

gabrieldemarmiesse:
Fix #2168  As the title says. I don't know if this implementation will give us the maximum amount of performance, (I don't know if I should have used MLIR directly) but it can always be improved later if needed.

Technically, in Python, a bool is an alias for 0 and 1, so a bool is an int. But I guess that doing `int(my_bool)` is close enough.

### DIFF
--- a/stdlib/src/builtin/bool.mojo
+++ b/stdlib/src/builtin/bool.mojo
@@ -46,7 +46,9 @@ trait Boolable:
 @lldb_formatter_wrapping_type
 @value
 @register_passable("trivial")
-struct Bool(Stringable, CollectionElement, Boolable, EqualityComparable):
+struct Bool(
+    Stringable, CollectionElement, Boolable, EqualityComparable, Intable
+):
     """The primitive Bool scalar value used in Mojo."""
 
     var value: __mlir_type.`!pop.scalar<bool>`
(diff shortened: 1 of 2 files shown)


bgreni:
I believe this would be a good opportunity to use `__mlir_op.'pop.cast'` to cast it to `__mlir_type.index` which is the underlying data type of `Int`?

gabrieldemarmiesse:
Very good idea, let me do this :)

stdlib/src/builtin/bool.mojo
```diff
@@ -245,3 +247,16 @@ struct Bool(Stringable, CollectionElement, Boolable, EqualityComparable):
...
+        """Convert this Bool to an integer.
+
+        Returns:
+            1 if the Bool is True, 0 otherwise.
+        """
+        return __mlir_op.`pop.cast_to_builtin`[_type = __mlir_type.index](
+            __mlir_op.`pop.cast`[_type = __mlir_type.`!pop.scalar<index>`](
+                self.value
+            )
+        )
```
soraros, line 262: There is a `Int` constructor from `!pop.scalar<index>`, so I think you could skip the `pop.cast_to_builtin`.

gabrieldemarmiesse: You are right, I fixed it, thanks!

JoeLoser (approved):
Makes sense, thank you for the contribution! 🎉

Issue #2051, opened 2024-03-29
[stdlib] Do not copy elements when using `FileHandle.read_bytes()`


VMois:
I was doing a one-billion row challenge with Mojo and tried reading 1 billion rows (around 13GB file) using `read_bytes()` and quickly ran out of memory. It does not happen with `read()`.

```mojo
alias input_file = "../1brc/measurements.txt"
alias chunk_size = 1024

fn main() raises:
    with open(input_file, "rb") as f:
        while len(f.read_bytes(chunk_size)):
            pass
```

After checking recently open-sourced `stdlib` (hooray), I found that while `read` passes buffer to String, `read_bytes` copies memory.

https://github.com/modularml/mojo/blob/6d2a7b552769358ec68a94b8fd1f6c2126d59ad9/stdlib/src/builtin/file.mojo#L187-L195

I am not a novice in Mojo, but I am far away from being good. I am not sure why it is a problem. I modified the read_bytes method and replaced the built stdlib in `.modular/pkg/packages.modular.com_mojo/lib/mojo`:

```mojo
var list = List[Int8](capacity=int(size_copy))
list.data = rebind[AnyPointer[Int8]](buf)
list.size = int(size_copy)
```
I am still unsure how `buf` becomes owned by a list here (similarly in the `read` method), but my modification does not run out of memory.

I would appreciate an explanation of why, and if my approach makes sense, I can open PR. Thank you

MoSafi2:
As #2182  is now merged, we can fix this issue.
@JoeLoser can you assign this to me?

JoeLoser:
> As #2182 is now merged, we can fix this issue. @JoeLoser can you assign this to me?

All yours, just assigned!  I think @jackos is working in this area for something else though.  @jackos do you mind chiming in here?

jackos:
Hey good catch @VMois yep it's an unnecessary memcpy and memory leak. The external function returns a pointer with the memory already allocated, so doing the memcpy to a different address means that the original pointer is never freed.

This would be a great PR to raise yeah! What you have there looks correct.

JoeLoser:
Fixed with https://github.com/modularml/mojo/commit/0b344fbd14fa80a33a3c07bfccba3dce661262de, thank you @MoSafi2 for the fix!

PR #2182, opened 2024-04-04
[stdlib] Add `List` constructor from (ptr, size)

### LINKED ISSUE #2170: [stdlib] [Feature Request] Allow List initalization from a raw pointer
### Review Mojo's priorities

### What is your request?

As per title, add an initalizer to list to allow initlization from a Pointer gived the length of the data similair to Rust's [from_raw_parts](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts).
Suggested signature: `(inout self: Self, owned ptr: AnyPointer,  length: Int, capacity: Int)`
This will allow fixing #2051 by eliminating unnecessary copying from the IO pointer and an empty list pointer

Currently, `FileHandle.read_bytes()` is **2x** slower than 24.1 implementation due to unnecessary copying of data which causes regression in many IO related tasks.

### Any other details?

_No response_

StandinKP:
Fixes #2170

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -61,7 +61,9 @@ struct _ListIter[
         inout self,
     ) -> Reference[T, list_mutability, list_lifetime]:
         self.index += 1
-        return self.src[].__get_ref[list_mutability, list_lifetime](self.index - 1)
+        return self.src[].__get_ref[list_mutability, list_lifetime](
+            self.index - 1
+        )
 
     fn __len__(self) -> Int:
         return len(self.src[]) - self.index
@@ -122,7 +124,9 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         for value in values:
             self.append(value[])
 
-    fn __init__(inout self: Self, owned ptr: AnyPointer[T], length: Int, capacity: Int):
+    fn __init__(
+        inout self: Self, owned ptr: AnyPointer[T], length: Int, capacity: Int
+    ):
         """Constructs a list from a pointer and its length.
 
         Args:
@@ -448,9 +452,7 @@ struct List[T: CollectionElement](CollectionElement, Sized):
     ](
         self: Reference[Self, mutability, self_life].mlir_ref_type,
         i: Int,
-    ) -> Reference[
-        T, mutability, self_life
-    ]:
+    ) -> Reference[T, mutability, self_life]:
         """Gets a reference to the list element at the given index.
 
         Args:
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -124,15 +124,17 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         for value in values:
             self.append(value[])
 
-    fn __init__(inout self: Self, ptr: AnyPointer[T], size: Int, capacity: Int):
+    fn __init__(
+        inout self: Self, data: AnyPointer[T], size: Int, capacity: Int
+    ):
         """Constructs a list from a pointer and its size.
 
         Args:
-            ptr: The pointer to the data.
+            data: The pointer to the data.
             size: The number of elements in the list.
             capacity: The capacity of the list.
         """
-        self.data = ptr
+        self.data = data
         self.size = size
         self.capacity = capacity
 
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -61,7 +61,9 @@ struct _ListIter[
         inout self,
     ) -> Reference[T, list_mutability, list_lifetime]:
         self.index += 1
-        return self.src[].__get_ref[list_mutability, list_lifetime](self.index - 1)
+        return self.src[].__get_ref[list_mutability, list_lifetime](
+            self.index - 1
+        )
 
     fn __len__(self) -> Int:
         return len(self.src[]) - self.index
@@ -448,9 +450,7 @@ struct List[T: CollectionElement](CollectionElement, Sized):
     ](
         self: Reference[Self, mutability, self_life].mlir_ref_type,
         i: Int,
-    ) -> Reference[
-        T, mutability, self_life
-    ]:
+    ) -> Reference[T, mutability, self_life]:
         """Gets a reference to the list element at the given index.
 
         Args:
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/list.mojo
```diff
@@ -124,6 +122,18 @@ struct List[T: CollectionElement](CollectionElement, Sized):
...
+        """Constructs a list from a pointer and its length.
+
+        Args:
+            ptr: The pointer to the data.
+            length: The number of elements in the list.
+            capacity: The capacity of the list.
+        """
+        self.data = ptr
+        self.size = length
+        self.capacity = capacity
```
LJ-9801, line 139: capacity has to be larger then length and we probably should not expose capacity to the user in this situation. Also the capacity should be the actual size of the pointer/array and it should probably need to be `_realloc()` if necessary.

StandinKP: Got it. So we should not set capacity on self right? Also then will we just derive the capacity from the pointer and then remove the capacity parameter?

LJ-9801: yes i believe so

StandinKP: I am trying to do the same but the AnyPointer doesn't have any param as size or capacity. How do we calculate the capacity? Am a bit new to doing low level operations. Thanks

arthurevans: The pointer doesn't have a capacity parameter. The user needs to know how much space they allocated—that's one reason pointers are "unsafe". 

If I understand the intention of this feature request, the method is intended to avoid a copy (for performance reasons) at the cost of some safety. So the user needs to know exactly what they're doing.

LJ-9801: In this case you can just take the pointer with a length. Then assign a capacity that is bigger then the length. The rest of method should be able to take care of the adjustment of sizing.

StandinKP: Yes even in [from_raw_parts](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts) it is mentioned the same that the user needs to take care of all the invariant checking as it is highly unsafe.

So I think keeping the capacity taken from the user makes sense. You can check the Rust reference above. Let me know if I need to change this

LJ-9801: > If I understand the intention of this feature request, the method is intended to avoid a copy (for performance reasons) at the cost of some safety. So the user needs to know exactly what they're doing.

Well wouldn't it be better to just have 1 unsafe argument then 2?

arthurevans:
Question for the standard library team: how should an unsafe API like this work, and how should we document it? And is this something that should be exposed as a constructor? I think we might want this to be a more explicitly-named method, like `take()`?

Both `extend()` and `steal_data()` have similar concerns. `extend()` takes the elements from another list and is supposed to leave the other list in an invalid state (it doesn't, but that's [another issue](https://github.com/modularml/mojo/issues/2185)).

`steal_data()`, which is kind of the reverse of this method, returns a pointer to the list's data, and nulls out its own pointer.

What would be the equivalent behavior here? If we leave the original anypointer intact, we have two references to the data, which is bad. If we took the anypointer as inout, we could copy it and then null out the original pointer so it can no longer access the data.

On the doc front, neither `extend()` or `steal_data()` are very explicit about the state they leave things in, although `steal_data()`'s name is pretty clear by itself.

StandinKP:
@JoeLoser any direction on how should I proceed with this?

JoeLoser:
> @JoeLoser any direction on how should I proceed with this?

I'll take a look at the content itself when I get a chance soon and then I'll start a discussion with the team around how we want to handle unsafe APIs like constructors such as this in general that @arthurevans brought up. I don't think we need to block the PR on that part though.

JoeLoser (commented):
Do you mind adding a test, please?  After we have this in place, then we can tackle https://github.com/modularml/mojo/issues/2051 as mentioned in https://github.com/modularml/mojo/issues/2170.

StandinKP:
> Do you mind adding a test, please? After we have this in place, then we can tackle #2051 as mentioned in #2170.

Yes. Sorry for the delay. I was in some work. Will add the tests

StandinKP:
@LJ-9801 @JoeLoser I was adding the test the following way:
```
def test_list_init_from_other_pointer():
    existing = List[Int](1, 2, 3)
    new = List[Int](existing.data, existing.size, existing.capacity)
```
The data is an AnyPointer, size and capacity are T (Int in this case) 
But the compiler crashes if I run the test. Even vscode says no matching function found in initialization? Am I missing something here?

LJ-9801:
You probably want to just pass in a raw pointer instead of the data of the list. This will cause a double free(maybe, not sure how the lifetime of the pointer of list is taken care of but it seems like a bad idea overall). VScode says this because this is a new method that has not been incorporated into the nightly compiler.

gabrieldemarmiesse:
@StandinKP I tried adding some tests and it worked correctly. You can copy my code, here is the commit: https://github.com/modularml/mojo/pull/2258/commits/e0198599abf51ec994bc66de4734e1aa46baa7a0

StandinKP:
Thanks @LJ-9801 @gabrieldemarmiesse for your inputs. I was essentially missing the `steal_data`
@JoeLoser I have pushed the tests please review when you get time. Thanks

stdlib/src/collections/list.mojo
```diff
@@ -124,6 +124,18 @@ struct List[T: CollectionElement](CollectionElement, Sized):
         for value in values:
             self.append(value[])
 
+    fn __init__(inout self: Self, ptr: AnyPointer[T], size: Int, capacity: Int):
```
gabrieldemarmiesse, line 127:
```suggestion
    fn __init__(inout self: Self, data: AnyPointer[T], size: Int, capacity: Int):
```
To get the pointer in a list, we do `my_list.data`, I don't think we need to introduce a new name that users will have to remember

StandinKP: Makes sense. Pushed changes

JoeLoser (approved):
Looks great.  Thank you for @gabrieldemarmiesse helping out here with the tests ❤️

PR #2183, opened 2024-04-04
[mojo-stdlib] Added offset fn to Anypointer

### LINKED ISSUE #2171: [Feature Request] Add an `offset` method to `AnyPointer`.
### Review Mojo's priorities

### What is your request?

As title.

- `Pointer` has it. In order to unify `Pointer` and `AnyPointer`, we should also add it to `AnyPointer`.
- `__add__` looks too cute, and prevents method chaining.

### Any other details?

Maybe make `__add__` less generic than `Pointer.__add__`, so we can prevent adding two pointers together (both pointer types are `Intable`). More concretely, now, we have:

```mojo
struct AnyPointer:
    fn __add__(self, rhs: Int) -> Self: ...
    fn __sub__(self, rhs: Int) -> Self: ...

struct Pointer:
    fn offset[T: Intable](self, offset: T) -> Self: ...
    fn __add__[T: Intable](self, rhs: T) -> Self: ...
    fn __sub__[T: Intable](self, rhs: T) -> Self: ...
```

I propose we move to:

```mojo
struct AnyPointer:
    fn offset[T: Intable](self, offset: T) -> Self: ... # retain a little bit of convenience 
    fn __add__(self, rhs: Int) -> Self: ...
    fn __sub__(self, rhs: Int) -> Self: ...
```

Or even better:

```mojo
struct AnyPointer:
    fn offset(self, offset: Int) -> Self: ...
    fn __add__(self, rhs: Int) -> Self: ...
    fn __sub__(self, rhs: Int) -> Self: ...
```

StandinKP:
Fixes #2171

### DIFF
--- a/stdlib/src/memory/anypointer.mojo
+++ b/stdlib/src/memory/anypointer.mojo
@@ -178,6 +178,19 @@ struct AnyPointer[T: Movable](
             dest.value
         ) = __get_address_as_owned_value(self.value)
 
+    @always_inline("nodebug")
+    fn offset(self, idx: Int) -> Self:
+        """Returns a new pointer shifted by the specified offset.
+
+        Args:
+            idx: The offset.
+
+        Returns:
+            The new Pointer shifted by the offset.
+        """
+        # Returns a new pointer shifted by the specified offset.
+        return Self.__from_index(int(self) + int(idx) * sizeof[T]())
+
     @always_inline
     fn __int__(self) -> Int:
         """Returns the pointer address as an integer.


JoeLoser (commented):
Looks right, but can you add a unit test or two please?

soraros:
@JoeLoser what's your opinion regarding other suggestions in #2171?

JoeLoser:
I'm actually going to suggest we no longer do this.  Internally, today, @lattner pushed on removing `lit.ref.offset` which is what `Pointer` used internally in its `offset` function.  So, in the next nightly release, `Pointer.offset` is removed.  So, we should hop on that ship and not introduce a new `offset` function to `AnyPointer`.  The original motivation in https://github.com/modularml/mojo/issues/2171 was for parity with `Pointer` to help us with moving uses of `Pointer` over to `AnyPointer`.  However, that argument is no longer sound since we soon remove `offset` from `Pointer` 😄 

I appreciate you looking into this though!

PR #2184, opened 2024-04-04
[mojo-stdlib] remove stringable trait in String

### LINKED ISSUE #2169: [Feature Request] Remove the `Stringable` constructor overload from `String`
### Review Mojo's priorities

### What is your request?

As title.

Now, we have traits and `str`, I see no reason to keep ad-hoc constructors on the `String` type itself. Removing this constructor also helps us control implicit conversion. Notably, both `Bool` and `Int` don't have the `*able` overloads.

### Any other details?

N/A

jiex-liu:
Fix #2169


soraros:
I meant remove the constructor from `Stringable`, and not the trait bound itself. Also, there is no consensus if we want the change at all. Let's wait a bit first.

PR #2186, opened 2024-04-04
[stdlib] Add `__rmod__` for `SIMD`

### LINKED ISSUE #1482: [Request] Implement `__mod__` in `Int` when rhs is SIMD type
### Bug description

This code snippet:

```
fn main():
    let array = DTypePointer[DType.int32].alloc(1000000)
    let bug = 1000 % (array[0] | 1)
```

produces the following error

```
/__w/modular/modular/Kernels/mojo/builtin/simd.mojo:890:5: error: function has recursive call to 'always_inline' function
/__w/modular/modular/Kernels/mojo/builtin/simd.mojo:899:22: note: call here recurses
/__w/modular/modular/Kernels/mojo/builtin/simd.mojo:890:5: note: back to function here
pooh@RAPTOR:~/aoc2023$ 
```

### Steps to reproduce

- minimal repro case included in the description

### System information

```shell
Win11 / WSL2 / Ubuntu 23.04
mojo 0.6.0 (d55c0025)
modular 0.2.2 (95d42445)
```

bgreni:
Add __rmod__ for SIMD types to allow expressions with an `Int` on the lhs

Fixes #1482

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -692,6 +692,19 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
             return mod + mask.select(rhs, Self(0))
 
     @always_inline("nodebug")
+    fn __rmod__(self, value: Self) -> Self:
+        """Returns `value mod self`.
+
+        Args:
+            value: The other value.
+
+        Returns:
+            `value mod self`.
+        """
+        constrained[type.is_numeric(), "the type must be numeric"]()
+        return value % self
+
+    @always_inline("nodebug")
     fn __pow__(self, rhs: Int) -> Self:
         """Computes the vector raised to the power of the input integer value.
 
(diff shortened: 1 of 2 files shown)


bgreni:
> I think this is the wrong fix. We should add a `Int` overload of `__rmod__` to `SIMD`, as all other simd operations have type `fn(SIMD, Int) -> SIMD` and not `fn(SIMD, Int) -> Int`.

Thanks for the feedback! I didn't realize that pattern had already been established. Turns out just adding a trivial implementation of `__rmod__` on SIMD does the trick.

JoeLoser (approved):
Looks good, thank you!  Do you mind fixing up the formatting that CI is failing on?  You can just run `mojo format -l 80 stdlib/src/builtin/simd.mojo` to make this pass.

bgreni:
@JoeLoser Thanks! It looks the line that was upsetting the formatter was resolved in a recent commit so I've rebased and I see no other changes when running it locally.

Issue #1876, opened 2024-03-04
[Feature Request] Allow to declare methods only if a type conforms to a specific trait


gabrieldemarmiesse:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

DynamicVector doesn't have an `__eq__` method currently because of this. We'd need something like this to do it:

```python
trait Equalable:
    fn __ne__(self, other: Self) -> Bool: ...

@value
struct CustomList[T: CollectionElement]:
    var internal_vector: DynamicVector[T]

    if hastrait(T, Equalable):
        def __eq__(self, other: Self) -> Bool:
            if len(self.internal_vector) != len(other.internal_vector):
                return False
            for i in range(len(self.internal_vector)):
                if self.internal_vector[i] != other.internal_vector[i]:
                    return False
            return True

@value
struct NotEqualable:
    var i: Int

def main():
    x = CustomList(DynamicVector[Int]()) # supports ==
    y = CustomList(DynamicVector[NotEqualable]()) # doesn't support ==
```

The method would be available conditionally, if the type `T` has the right trait. `hastrait` would work like `isinstance`. 

### What is your motivation for this change?

Otherwise we need to define the function for each type know, that's very painful and leads to ugly apis. The fact that DynamicVector and Dict don't have `__eq__` or `__str__` shows this.

soraros:
+1 for conditional conformance, -20 for `hastrait` as the means to support it.

To put it simply: we probably don't want C++ concepts but something closer to Rust, Swift or Haskell.

gryznar:
Moreover, why `hastrait` and not `has_trait`?

JoeLoser:
Thanks for filing the issue.  This was actually one of the first things I asked for when starting to experiment with traits in the library when they initially got implemented in the compiler.  Conditional conformance is on the compiler roadmap, but I don't have a clear timeline when it will be started right now.

The library is intentionally not providing `__eq__` and friends right now in `DynamicVector` because of limitations in using traits at this time.  

Stay tuned as conditional conformance is important from my perspective as we start pushing on traits more in the library.

soraros:
@JoeLoser Could you comment more about the approach Mojo may take? Like, leaning more towards the "Turing complete" `if trait` or Rust style "very static" conditional conformance?

gabrieldemarmiesse:
[...]

PR #2190, opened 2024-04-04
[mojo-stdlib] Add new `ComparableCollectionElement` trait and `index(list, value, start, end)` function APIs

arvindavoudi:
This pull request introduces the new `index` method to the `collections.List` struct within mojo-stdlib, mirroring the exact functionality found in Python's list, as detailed [here](https://docs.python.org/3/tutorial/datastructures.html).

This also allows for the specification of start and end parameters when searching for an item's index within a list, with both parameters accepting positive and negative numbers.

Also comprehensive suite of unit tests included meticulously designed to cover a wide range of scenarios for the `index` method's application.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -452,16 +452,17 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         # TODO: Once the min() and max() functions are available in Mojo,
         # TODO: we can simplify the entire if-else block into a single line using the ternary operator:
         # var normalized_end = self.size if end is None else min(max(end, 0), self.size)
+        var normalized_end: Int
         if end is None:
-            var normalized_end = self.size
+            normalized_end = self.size
         else:
-            if end < 0:
-                var normalized_end = self.size + end
+            if end.value()[] < 0:
+                normalized_end = self.size + end.value()[]
             else:
-                if end > self.size:
-                    var normalized_end = self.size
+                if end.value()[] > self.size:
+                    normalized_end = self.size
                 else:
-                    var normalized_end = end
+                    normalized_end = end.value()[]
 
         if not self.size:
             raise "Cannot find index of a value in an empty list."
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -445,7 +445,19 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             ```
         """
         var normalized_start = (self.size + start) if start < 0 else start
-        var normalized_end = self.size if end is None else (self.size + end if end < 0 else (self.size if end > self.size else end))
+        # TODO: Once the min() and max() functions are available in Mojo,
+        # TODO: we can simplify the entire if-else block into a single line using the ternary operator:
+        # var normalized_end = self.size if end is None else min(max(end, 0), self.size)
+        if end is None:
+            var normalized_end = self.size
+        else:
+            if end < 0:
+                var normalized_end = self.size + end
+            else:
+                if end > self.size:
+                    var normalized_end = self.size
+                else:
+                    var normalized_end = end
 
         if not self.size: raise "Cannot find index of a value in an empty list."
         if normalized_start >= self.size: raise "Given 'start' parameter (" + String(normalized_start) + ") is out of range. List only has " + String(self.size) + " elements."
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -473,9 +473,6 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             ) + " elements."
 
         for i in range(normalized_start, normalized_end):
-            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
-            # could degrade the performance of the List.index method if it were to replace the current implementation,
-            # as it would essentially perform the same loop twice.
             if self[i] == value:
                 return i
         raise "ValueError: Given element is not in list"
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -427,11 +427,19 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         Returns the index of the first occurrence of a value in a list, starting from the specified
         index (default 0). Raises an Error if the value is not found.
 
+        ```mojo
+        var my_list = List[Int](1, 2, 3)
+        print(__type_of(my_list).index(my_list, 2)) # Output: 1
+        ```
+
         Args:
             self: The list to search in.
             value: The value to search for.
             start: The starting index of the search (default 0).
             end: The ending index of the search (default None, which means the end of the list).
+
+        Parameters:
+            C: The type of the elements in the list. Must implement the `ComparableCollectionElement` trait.
 
         Returns:
             The index of the first occurrence of the value in the list.
@@ -439,14 +447,6 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         Raises:
             ValueError If the value is not found in the list.
 
-        Type Parameters:
-            C: The type of the elements in the list. Must implement the `ComparableCollectionElement` trait.
-
-        Example:
-            ```mojo
-            var my_list = List[Int](1, 2, 3)
-            print(__type_of(my_list).index(my_list, 2))  # Output: 1
-            ```
         """
         var normalized_start = (self.size + start) if start < 0 else start
         # TODO: Once the min() and max() functions are available in Mojo,
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -416,6 +416,59 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             earlier_idx += 1
             later_idx -= 1
 
+    @always_inline
+    fn index(self, owned value: T, start: Int = 0) raises -> Int:
+        var normalized_start = self.size + start if start < 0 else start
+        var normalized_end = self.size
+    
+        if normalized_start >= self.size: raise "Given 'start' parameter is out of range"
+    
+        var ret_val: Int = -1
+        for i in range(normalized_start, normalized_end):
+            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
+            # could degrade the performance of the List.index method if it were to replace the current implementation,
+            # as it would essentially perform the same loop twice.
+            if ((self.data + i).bitcast[T]()[]) == value:
+                ret_val = i
+                break
+        else:
+            pass
+            # TODO: Currently, attempting to raise a message, leads to a compiler error (#issue2188)
+            # TODO: Once this issue is resolved, the commented-out code below should be uncommented.
+            # TODO: And then the if statement below should get removed completely.
+            # Code:
+            # raise "Value does not exist in list"
+        if ret_val == -1:
+            raise "Value does not exist in list"
+        return ret_val
+
+    @always_inline
+    fn index(self, owned value: T, start: Int, end: Int) raises -> Int:
+        var normalized_start = (self.size + start) if start < 0 else start
+        var normalized_end = (self.size + end) if end < 0 else end
+    
+        if normalized_end > self.size: raise "Given 'end' parameter is out of range"
+        if normalized_start >= self.size: raise "Given 'start' parameter is out of range"
+    
+        var ret_val: Int = -1
+        for i in range(normalized_start, normalized_end):
+            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
+            # could degrade the performance of the List.index method if it were to replace the current implementation,
+            # as it would essentially perform the same loop twice.
+            if ((self.data + i).bitcast[T]()[]) == value:
+                ret_val = i
+                break
+        else:
+            pass
+            # TODO: Currently, attempting to raise a message, leads to a compiler error (#issue2188)
+            # TODO: Once this issue is resolved, the commented-out code below should be uncommented.
+            # TODO: And then the if statement below should get removed completely.
+            # Code:
+            # raise "Value does not exist in list"
+        if ret_val == -1:
+            raise "Value does not exist in list"
+        return ret_val
+
     fn clear(inout self):
         """Clears the elements in the list."""
         for i in range(self.size):
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -416,8 +416,9 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             earlier_idx += 1
             later_idx -= 1
 
-    @always_inline
-    fn index(self, owned value: T, start: Int = 0) raises -> Int:
+    # TODO: Modify these to regular methods when issue 1876 is resolved
+    @staticmethod
+    fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0) raises -> Int:
         """Returns the index of the first occurrence of a value in a list; raises Error if not found."""
         var normalized_start = self.size + start if start < 0 else start
         var normalized_end = self.size
@@ -430,15 +431,15 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
             # could degrade the performance of the List.index method if it were to replace the current implementation,
             # as it would essentially perform the same loop twice.
-            if ((self.data + i).bitcast[T]()[]) == value:
+            if ((self.data + i).bitcast[C]()[]) == value:
                 ret_val = i
                 break
         else:
             raise "Value does not exist in the list."
         return ret_val
 
-    @always_inline
-    fn index(self, owned value: T, start: Int, end: Int) raises -> Int:
+    @staticmethod
+    fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int, end: Int) raises -> Int:
         """Returns the index of the first occurrence of a value in a list; raises Error if not found."""
         var normalized_start = (self.size + start) if start < 0 else start
         var normalized_end = (self.size + end) if end < 0 else (self.size if end > self.size else end)
@@ -451,7 +452,7 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
             # could degrade the performance of the List.index method if it were to replace the current implementation,
             # as it would essentially perform the same loop twice.
-            if ((self.data + i).bitcast[T]()[]) == value:
+            if ((self.data + i).bitcast[C]()[]) == value:
                 ret_val = i
                 break
         else:
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -418,9 +418,13 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
 
     # TODO: Modify this to be regular method when issue 1876 is resolved
     @staticmethod
-    fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0, end: Optional[Int] = None) raises -> Int:
+    fn index[
+        C: ComparableCollectionElement
+    ](
+        self: List[C], owned value: C, start: Int = 0, end: Optional[Int] = None
+    ) raises -> Int:
         """
-        Returns the index of the first occurrence of a value in a list, starting from the specified 
+        Returns the index of the first occurrence of a value in a list, starting from the specified
         index (default 0). Raises an Error if the value is not found.
 
         Args:
@@ -459,8 +463,14 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
                 else:
                     var normalized_end = end
 
-        if not self.size: raise "Cannot find index of a value in an empty list."
-        if normalized_start >= self.size: raise "Given 'start' parameter (" + String(normalized_start) + ") is out of range. List only has " + String(self.size) + " elements."
+        if not self.size:
+            raise "Cannot find index of a value in an empty list."
+        if normalized_start >= self.size:
+            raise "Given 'start' parameter (" + String(
+                normalized_start
+            ) + ") is out of range. List only has " + String(
+                self.size
+            ) + " elements."
 
         for i in range(normalized_start, normalized_end):
             # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -431,12 +431,9 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
             # could degrade the performance of the List.index method if it were to replace the current implementation,
             # as it would essentially perform the same loop twice.
-            if ((self.data + i).bitcast[C]()[]) == value:
-                ret_val = i
-                break
-        else:
-            raise "Value does not exist in the list."
-        return ret_val
+            if self[i] == value:
+                return i
+        raise "ValueError: Given element is not in list"
 
     fn clear(inout self):
         """Clears the elements in the list."""
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -419,7 +419,31 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
     # TODO: Modify this to be regular method when issue 1876 is resolved
     @staticmethod
     fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0, end: Optional[Int] = None) raises -> Int:
-        """Returns the index of the first occurrence of a value in a list; raises Error if not found."""
+        """
+        Returns the index of the first occurrence of a value in a list, starting from the specified 
+        index (default 0). Raises an Error if the value is not found.
+
+        Args:
+            self: The list to search in.
+            value: The value to search for.
+            start: The starting index of the search (default 0).
+            end: The ending index of the search (default None, which means the end of the list).
+
+        Returns:
+            The index of the first occurrence of the value in the list.
+
+        Raises:
+            ValueError If the value is not found in the list.
+
+        Type Parameters:
+            C: The type of the elements in the list. Must implement the `ComparableCollectionElement` trait.
+
+        Example:
+            ```mojo
+            var my_list = List[Int](1, 2, 3)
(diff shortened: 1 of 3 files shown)


stdlib/src/collections/list.mojo
```diff
@@ -335,6 +335,61 @@ struct List[T: CollectionElement](CollectionElement, Sized):
...
+        for i in range(normalized_start, normalized_end):
+            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
+            # could degrade the performance of the List.index method if it were to replace the current implementation,
+            # as it would essentially perform the same loop twice.
+            if ((self.data + i).bitcast[T]()[]) == value:
+                ret_val = i
+                break
+        else:
+            pass
+            # TODO: Currently, attempting to raise a message, leads to a compiler error (#issue2188)
```
JoeLoser, line 356: FYI issue 2188 seems to be resolved.  Can you uncomment this part and see if the `raise` works?

stdlib/src/collections/list.mojo
```diff
@@ -335,6 +335,61 @@ struct List[T: CollectionElement](CollectionElement, Sized):
             earlier_idx += 1
             later_idx -= 1
 
+    @always_inline
+    fn index(self, owned value: T, start: Int = 0) raises -> Int:
+        """Returns the index of the first occurrence of a value in a list; raises ValueError if not found."""
+        var normalized_start = self.size + start if start < 0 else start
+        var normalized_end = self.size
+    
+        if normalized_start >= self.size: raise "Given 'start' parameter is out of range"
```
JoeLoser, line 344: **Suggestion** It may be good to include `normalized_start` as well as `len(self)`, i.e. the list size in the error message raised.  Ditto in the overload below.

stdlib/test/collections/test_list.mojo
```diff
@@ -286,6 +286,60 @@ def test_list_reverse_move_count():
...
     _ = vec^
 
+def test_list_index():
+    var test_list_a = List[Int](10,20,30,40,50)
+    
+    # Basic Functionality Tests
+    assert_equal(test_list_a.index(10), 0)  # Verify finding the first element's index.
+    assert_equal(test_list_a.index(30), 2)  # Verify finding the index of a middle element.
+    assert_equal(test_list_a.index(50), 4)  # Verify finding the last element's index.
+    with assert_raises():  # Verify error for a non-existent element.
```
JoeLoser, line 296: **Suggestion** You could specify `contains=<some_msg>` as a keyword argument to the `assert_raises` to verify the error message if you'd like.  This would be nice if you adopt my earlier comment about including the `normalized_start` and length of the list in the error message.

stdlib/test/collections/test_list.mojo
```diff
@@ -286,6 +286,60 @@ def test_list_reverse_move_count():
...
+        _ = test_list_a.index(60)
+
+    # Tests With Start Parameter
+    assert_equal(test_list_a.index(30, start=1), 2)  # Verify search beginning from a specific start index.
+    assert_equal(test_list_a.index(30, start=-4), 2)  # Verify element search with a negative start index.
```
JoeLoser, line 301: **Question** Is this one correct?  I think in python,

```python
xs = [10, 20, 30, 40, 50]
xs.index(30, -4, 2)
```

will raise.

stdlib/test/collections/test_list.mojo
```diff
@@ -286,6 +286,60 @@ def test_list_reverse_move_count():
...
+
+
+    var test_list_b = List[Int](10, 20, 30, 20, 10)
+
+    # Test finding the first occurrence of an item
+    assert_equal(test_list_b.index(10), 0)  # Verify first occurrence of '10'.
+    assert_equal(test_list_b.index(20), 1)  # Verify first occurrence of '20'.
+
+    # Test skipping the first occurrence with a start parameter
+    assert_equal(test_list_b.index(20, start=2), 3)  # Skip first '20', find second.
```
JoeLoser, line 330: **Question** Is this one correct as well?  Similar to my question about the test case above.

JoeLoser:
> Additionally, it's worth noting that the `list.index` method in Python accepts the parameters `value`, `start`, and `end`. However, these are not treated as keyword arguments, meaning that while passing the `start` and `end` parameters is optional, if you wish to specify `end`, you must also pass `start`. This limitation informed my decision to overload the index method as you can observe.
> 
> Personally, I would prefer the flexibility to omit the `start` parameter, especially since the CPython API is gradually embracing keyword arguments for functions that previously did not support them. This is evidenced by recent developments, such as this [PR#106488](https://github.com/python/cpython/pull/106488) for Python 3.13, which aligns with this preference.
> 
> I'm eager to hear your thoughts on this matter. Thanks in advance.

Thanks for bringing this up.  This is interesting from an API design evolution standpoint.  Given that we already have keyword arguments, and modern Python is moving in this direction, I'd personally like to intersect with where they're going given where the Mojo language is at right now, rather than trying to be exactly source-compatible with older Python versions exclusively.  However, from a pragmatic "now" perspective, I'd recommend we keep the API as you have it, even though that's not using the keyword argument form.  I think we should raise this up into a broader API evolution guidelines talking point.  I'll think about the right forum to raise this in (whether internally or as a discussion externally) and go from there.  Let's not let the (lack of) a decision on that block this PR though.

FYI: @modularml/mojo-standard-library @jopamer

JoeLoser (commented):
Generally looks good — please take a look at my comments though.  In addition, you'll need to rebase with the latest `nightly` branch and resolve the conflicts in one file.  Thanks for working on this!

stdlib/src/collections/list.mojo
arvindavoudi: Sure. I'm going to update it.

stdlib/src/collections/list.mojo
arvindavoudi: Great idea for improving error messages in Mojo, similar to the enhancements we've observed in CPython over recent years. I will implement this as well. Thanks

stdlib/test/collections/test_list.mojo
arvindavoudi: Um, It seems you misread the test case. We are not calling `index(30, -4, 2)`. Instead, the method is `index(30, -4)`, and then we use `assert_equal` to verify whether the result equals `2`. So the final code for that test is correct as below:

```mojo
assert_equal(test_list_a.index(30, start=-4), 2)
```

Also if you run your given code in Mojo, it will raise an error just like Python. I believe that the implemented `index` method does not have any issues here.

stdlib/test/collections/test_list.mojo
arvindavoudi: Yes, I think that's correct as I described above.

stdlib/test/collections/test_list.mojo
JoeLoser: Oh you're definitely right here. Sorry I misread the test case.

stdlib/src/collections/list.mojo
arvindavoudi: One quick question. In Python, if you give out-of-range start/end values, it does not raise any error concerning that bad given range. It just raises a value error in any case that it could not find the value. So, in Python, the code below works but in my implementation, it does not work:

```python
x = [10, 20, 30]
x.index(10, 0, 500) # Search for value; If does not exists, raise a Value Error. Don't care if start/end is out of range
```

So do you think we should keep this raising as it is? Or because of this incompatibility, it's better to drop it off?

I suddenly noticed about this so I personally prefer to have the same behavior in Mojo as well.

stdlib/test/collections/test_list.mojo
arvindavoudi: Done

stdlib/src/collections/list.mojo
arvindavoudi: I fixed it to use an appropriate approach. Neither losing clarity of the error messages you desired nor making incompatibility with Python's API.

arvindavoudi:
Now I think we must wait for [this issue](https://github.com/modularml/mojo/issues/2238) to get resolved.

JoeLoser:
> Now I think we must wait for [this issue](https://github.com/modularml/mojo/issues/2238) to get resolved.

Instead of blocking on a resolution to that issue, what do you think about making these free functions in the `list.mojo` file and the first argument is `List[T]` instead of the `self` argument now?  I'd recommend we move forward in this direction and can move them to be a member function once our traits are further along.

In general, anywhere where we need these trait refinements (like `__eq__` or `__str__` requirement), we could make them be free functions for now, like we do with `UnsafePointer` extension methods for example [here](https://github.com/modularml/mojo/blob/0a884b7941303e41d77e889101c34783c9e3a975/stdlib/src/memory/unsafe_pointer.mojo#L346).  FYI @gabrieldemarmiesse as I think you mentioned something on Discord recently about this sort of thing.

gabrieldemarmiesse:
@arvindavoudi You can use this implementation as reference: https://github.com/modularml/mojo/pull/2323  I cause here a small parser bug, but this should not happen if you use this strategy to implement `List.index()`

Users will be able to call it with `__type_of(my_list).index(my_list, value, start, end)`

arvindavoudi:
@JoeLoser, @gabrieldemarmiesse
Thanks for your concern. But one part is still ambiguous for me. In [issue 2263](https://github.com/modularml/mojo/issues/2263), you mentioned `count` implementation should be like below (which is very similar to `index`):

```mojo
def count[T: CollectionElement](arg: List[T], value: T):
   ...
```

But I didn't quite understand what approach we should take to be able to compare two values when they are both identified as CollectionElement, so then not comparable. In my PR this comparing part was specified as:

```mojo
if ((self.data + i).bitcast[T]()[]) == value:
```

I highly appreciate if you could clarify it for me as I can't really see any striking difference in here.

gabrieldemarmiesse:
I think it was a small oversight on @JoeLoser 's part, you are right. `CollectionElement`'s types are not `EqualityComparable`, so you can't compare easily elements.

You can follow #2323  to see how to do it in practice, but I'll give you a summary. There are 3 main steps for you:

### Make a new trait that combines `CollectionElement` and `EqualityComparable`:
in mojo it would look like this:

```mojo
trait ComparableCollectionElement(EqualityComparable, CollectionElement):
    pass
```
### Create a static method on the List struct
Name the first argument `self` (so we won't have to rename it later on) and specify the type as corresponding to the `ComparableCollectionElement` trait. The signature will look like this:

```mojo
@staticmethod
fn index[U:ComparableCollectionElement](self: List[U], owned value: U, start: Int = 0) raises -> Int:
```
(I looked at the code in your PR as a reference).

### Use the static method
To use the method, since it's a staticmethod, you need to access the type of the list object you're working with. It will look like this:

```mojo
var my_list = List[Int](10, 11, 12)
var index_of_11 = __type_of(my_list).index(my_list, 11)
# note that in the future, we can remove @staticmethod and users will replace the call by "var index_of_11 = my_list.index(11)"
# it will be easy to replace by searching "__type_of" in the project
```

Is this enough to get you started? If you have any other questions feel free to ask :)

JoeLoser:
As @gabrieldemarmiesse pointed out, that was just a small oversight on my part — sorry for the confusion (I've updated the GitHub issue to avoid future confusion).  We do need a composite trait to encompass that equality comparable requirement.  Mojo doesn't have in-place "anonymous composition" of traits, like (bikeshed syntax) 

```
def my_function[T: EqualityComparable + CollectionElement](value: T):
  pass
```

Instead, you have to create a named trait for each "anonymous" composition, e.g.

```
trait ComparableCollectionElement(EqualityComparable, CollectionElement):
    pass
```

This isn't ideal, but workable in the short-term until traits are further along in the compiler.  In this case, to do that "refinement" to ensure the given type can be guaranteed to have the `__eq__` function you want to use in your algorithm, we need said trait.

I don't have strong preferences on whether we make it a static method or free function outside of the struct until we have conditional conformances.  😄 

The key point is:
- We won't have conditional conformance anytime soon necessarily, so don't block this PR on that being available
- The reworking of `CollectionElement` trait to include `EqualityComparable` is complicated and won't be a short-term thing, so again, don't block this PR on that

So, we'll get creative in order for the library to move forward until we can get those nicer things in the future.  I have plans to push on (2) a bit sometime in the coming month or two.

arvindavoudi:
Thanks for your detailed clarification @gabrieldemarmiesse.
I rebased and implemented those two as free functions as @JoeLoser suggested.
I also decided to put the new `ComparableCollectionElement` trait beside `CollectionElement` in `builtin/value.mojo` so it must be globally accessible from now on..?

stdlib/src/collections/list.mojo
```diff
@@ -568,3 +568,46 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         """
         var ref = Reference(self)
         return _ListIter[T, mutability, self_life, False](len(ref[]), ref)
+
+# TODO: Convert functions below to list methods when issue 1876 is resolved.
+@always_inline
+fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0) raises -> Int:
```
gabrieldemarmiesse, line 574: If I may suggest, making this a `@staticmethod` instead of a free function. A few reasons I can think of:

* it allows a user to call this method without importing anything.
* It allows `__type_of(my_list).index(my_list, "hello")` which is easy to search and replace when #1876 is fixed
* When #1876 is fixed, users and us stdlib authors won't have to change the imports
* When #1876 is fixed, we won't have to move the code into the struct, avoiding a big diff due to indentation (for free functions, the indentation is 0 spaces, for methods it's 4 spaces)
* For built-in structs, users don't import anything to use them. So it's not obvious where they should import the free function from: `from ??? import index`. For us, standard library devs, it's obvious what the import is, but for Mojo users, they'll have to look at the docs to know the correct import.
* In this PR, the `index` method is imported in `stdlib/src/collections/__init__.mojo`, what do we do when multiple collections have a method with the same name, like `__eq__` ?
* For methods common to multiple types, it's going to get confusing. Let's say the user needs `__eq__` on both a List and a Dict in the same file, then users will have to do 
```mojo
from collections.list import __eq__
from collections.dict import __eq__
```
and it's going to get confusing.

For all those reasons, I suggest going for the syntax `__type_of(my_list).index(my_list, "hello")` which will lead to less issues in the future.

You can take a look at the way to make this syntax happen in this PR: https://github.com/modularml/mojo/pull/2323

stdlib/src/collections/list.mojo
```diff
@@ -414,6 +414,49 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        var ret_val: Int
+        for i in range(normalized_start, normalized_end):
+            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
+            # could degrade the performance of the List.index method if it were to replace the current implementation,
+            # as it would essentially perform the same loop twice.
+            if ((self.data + i).bitcast[C]()[]) == value:
+                ret_val = i
+                break
+        else:
+            raise "Value does not exist in the list."
```
gabrieldemarmiesse, line 436:
```suggestion
            raise "ValueError: Element is not in list."
```

To mimic the error message of python.

stdlib/src/collections/list.mojo
```diff
@@ -414,6 +414,49 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             earlier_idx += 1
             later_idx -= 1
 
+    # TODO: Modify these to regular methods when issue 1876 is resolved
+    @staticmethod
+    fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0) raises -> Int:
```
gabrieldemarmiesse, line 419: I don't think you need to implement the whole logic twice. You have two options:

```mojo
@staticmethod
fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0) raises -> Int:
    return Self.index(self, value^, start, len(self))
```
this will call the second overload where the full implementation is. Or you have a second option, implementing only one signature:

```mojo
@staticmethod
fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0, end: Optional[Int] = None) raises -> Int:
    if end is None:
        end = len(self)
    ...
```

stdlib/src/collections/list.mojo
```diff
@@ -414,6 +414,49 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        if normalized_start >= self.size: raise "Given 'start' parameter (" + String(normalized_start) + ") is out of range. List only has " + String(self.size) + " elements."
+
+        var ret_val: Int
+        for i in range(normalized_start, normalized_end):
+            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
+            # could degrade the performance of the List.index method if it were to replace the current implementation,
+            # as it would essentially perform the same loop twice.
+            if ((self.data + i).bitcast[C]()[]) == value:
+                ret_val = i
+                break
```
gabrieldemarmiesse, line 434: You can just do `return i` instead of the `break` here. That should still work, and you won't need `ret_val`

stdlib/src/collections/list.mojo
```diff
@@ -414,6 +414,49 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+
+        if not self.size: raise "Cannot find index of a value in an empty list."
+        if normalized_start >= self.size: raise "Given 'start' parameter (" + String(normalized_start) + ") is out of range. List only has " + String(self.size) + " elements."
+
+        var ret_val: Int
+        for i in range(normalized_start, normalized_end):
+            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
+            # could degrade the performance of the List.index method if it were to replace the current implementation,
+            # as it would essentially perform the same loop twice.
+            if ((self.data + i).bitcast[C]()[]) == value:
```
gabrieldemarmiesse, line 432:
```suggestion
            if self[i] == value:
```
woudln't that work too?

stdlib/test/collections/test_list.mojo
```diff
@@ -348,6 +348,69 @@ def test_list_insert():
     for i in range(len(v4)):
         assert_equal(v4[i], i + 1)
 
+def test_list_index():
+    var test_list_a = List[Int](10,20,30,40,50)
+    
+    # Basic Functionality Tests
+    assert_equal(__type_of(test_list_a).index(test_list_a, 10), 0)  # Verify finding the first element's index.
```
gabrieldemarmiesse, line 355: I think the comments are a bit too much here. Comments should be there to explain something being done implicitely in the code, or to provide context that the reader can't guess, or even explain complexe algorithms. Here the comments are explaining in english what the code does. I don't think they're useful. The comments separating each section are relevant though.

gabrieldemarmiesse (commented):
Thank you for the PR, could you run "mojo format ./" before pushing the new changes? Thanks! Also note that you can contest all my comments, I'm not modular staff nor do I represent them in any way 😁

stdlib/src/collections/list.mojo
arvindavoudi: Done

stdlib/src/collections/list.mojo
arvindavoudi: You are right. This method was used before the resolution of issue 2188. Done

stdlib/src/collections/list.mojo
arvindavoudi: Done

stdlib/test/collections/test_list.mojo
arvindavoudi: Thanks. Done. It's now definitely cleaner.

stdlib/src/collections/list.mojo
arvindavoudi: Thanks for bringing this up. I was also looking forward to implementing this, but I first wanted to have an acceptable version of `index`. Now that it's approved, I also going to merge those methods two by the help of `optional` in Mojo.
This also means we can pass `end` parameter without the need for specifying `start`. So I also added some new tests to cover this scenario.

arvindavoudi:
> could you run "mojo format ./" before pushing the new changes?

Sure. I run it but it seems there are more formatting issues available in `list.mojo` and `test_list.mojo` that are not related to this PR. I doubt whether to commit those in here or not.

stdlib/src/collections/list.mojo
```diff
@@ -418,6 +419,25 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             earlier_idx += 1
             later_idx -= 1
 
+    # TODO: Modify this to be regular method when issue 1876 is resolved
+    @staticmethod
+    fn index[C: ComparableCollectionElement](self: List[C], owned value: C, start: Int = 0, end: Optional[Int] = None) raises -> Int:
+        """Returns the index of the first occurrence of a value in a list; raises Error if not found."""
+        var normalized_start = (self.size + start) if start < 0 else start
+        var normalized_end = self.size if end is None else (self.size + end if end < 0 else (self.size if end > self.size else end))
```
gabrieldemarmiesse, line 427: That's a lot of logic in the same line, and can make it quite hard to read. Can you split it into multiple lines?

gabrieldemarmiesse:
It seems that you have git conflicts here. You'll need to either merge or rebase the "nightly" branch into your branch and fix the conflicts. Afterwards, make sure you have the latest version of mojo nightly with `modular clean && modular install nightly/mojo`, then you'll be able to run `mojo format ./`. 

If you have a hard time resolving the conflicts, or that you're afrait of loosing your changes, you can add me with commit rights to https://github.com/arvindavoudi/mojo and I'll fix the conflicts for you :)

stdlib/src/collections/list.mojo
```diff
@@ -418,6 +419,25 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+
+        if not self.size: raise "Cannot find index of a value in an empty list."
+        if normalized_start >= self.size: raise "Given 'start' parameter (" + String(normalized_start) + ") is out of range. List only has " + String(self.size) + " elements."
+
+        var ret_val: Int
```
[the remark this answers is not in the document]
arvindavoudi: Ah, my bad. Done.

stdlib/src/collections/list.mojo
arvindavoudi: Done.

stdlib/test/collections/test_list.mojo
```diff
@@ -348,6 +348,77 @@ def test_list_insert():
...
 
+def test_list_index():
+    var test_list_a = List[Int](10,20,30,40,50)
+    
+    # Basic Functionality Tests
+    assert_equal(__type_of(test_list_a).index(test_list_a, 10), 0)
+    assert_equal(__type_of(test_list_a).index(test_list_a, 30), 2)
+    assert_equal(__type_of(test_list_a).index(test_list_a, 50), 4)
+    with assert_raises(contains="ValueError: Given element is not in list"):
+        _ = __type_of(test_list_a).index(test_list_a, 60)
```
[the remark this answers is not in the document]
arvindavoudi: In regular Mojo scripts, when a function returns a value that is not assigned to a variable, a warning is displayed. To address this, I assigned all those to "_" but I removed them in the newly committed changes.

stdlib/src/collections/list.mojo
```diff
@@ -418,6 +419,25 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        for i in range(normalized_start, normalized_end):
+            # Note: Implementing __contains__ with O(n) time complexity in future, indicating it relies on linear search,
+            # could degrade the performance of the List.index method if it were to replace the current implementation,
+            # as it would essentially perform the same loop twice.
```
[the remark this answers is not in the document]
arvindavoudi: Um yeah. This suggests that if want to first check `value` availability in the list using `in`, this makes the algorithm iterate over the list two times if the value is available. One time for checking value availability and the second time for finding its index. In the first version which was not clean and simple at all, I thought these comments may lead to preventing future "refactors" that can corrupt the performance. Since it's much more understandable now, I also assume that's unnecessary, so I will remove them.

arvindavoudi:
Dear @gabrieldemarmiesse. I fixed everything you mentioned + the formatting. I hope there is no crucial problem anymore, ready to get merged :)

gabrieldemarmiesse:
Mmhhh the CI is still not passing. That's weird. It seems the formatting is still not okay, if you want I'll take a look and see if I can push the commits directly to your branch.

arvindavoudi:
Sure thanks. I have added you as a collaborator.

stdlib/test/collections/test_list.mojo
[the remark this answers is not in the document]
arvindavoudi: We can see warnings in the GitHub actions job result.

arvindavoudi:
@gabrieldemarmiesse. No need for any further changes, I fixed the remaining issues. Thanks a lot for your valuable cooperation during the submission of this PR.

gabrieldemarmiesse:
Great! Thank you for your hard work there!

JoeLoser (approved):

Hey @arvindavoudi,

Thanks so much for the contribution — love the tests!  🎉 🔥 

We're moving to a new infrastructure for merging contributions to Mojo (we're using a tool called [Copybara](https://github.com/google/copybara)), and your contribution has now been merged into our internal copy of the Mojo Standard Library. I've added the "merged-internally" label on this PR.

The changes in this PR will appear here in the mojo repo nightly branch when we do our next outbound synchronization at the time that the next Mojo nightly is released. That should happen later today.

ConnorGray:
Closing as this has been merged internally, per Joe's comment above.

stdlib/src/collections/list.mojo
```diff
@@ -416,6 +416,68 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             earlier_idx += 1
             later_idx -= 1
 
+    # TODO: Modify this to be regular method when issue 1876 is resolved
+    @staticmethod
+    fn index[
+        C: ComparableCollectionElement
+    ](
+        self: List[C], owned value: C, start: Int = 0, end: Optional[Int] = None
```
ConnorGray, line 424: This PR is amazing, thank you for the contribution @arvindavoudi and all the help @gabrieldemarmiesse.

One minor question we could possible address in a follow-up PR: Does this `value` argument used for comparision need to be `owned`, or could use the default `borrowed` argument convention? I.e.:

```mojo
self: List[C], value: C, ....
```

Logically I think `borrowed` would make sense, since the element isn't mutated or consumed by this method.

Is there some other reason this methods needs its own unique copy of the `value` argument?

JoeLoser: I had the same thought this morning — something we should follow-up on.

ConnorGray: No worries :) And yep, we can take of this internally as well.

Once we sync our internal changes out again, it will be easier for a community member to make this change as well if they wanted to. Its a small fix though, so I might just do it real quick internally if no one else gets to it first :)

ConnorGray: This has been fixed by our new intern @Dan13llljws in [this commit here](https://github.com/modularml/mojo/commit/f2f9e77ecf9780af4ecdf28ca170ca9276f586a9) — welcome to the team @Dan13llljws, and congratulations on your first Mojo🔥 contribution! 🎉  😄

PR #2192, opened 2024-04-04
[stdlib][draft] Add `datetime.timedelta` only the constructor and attributes

gabrieldemarmiesse:
This PR is in draft, it's missing some tests. It's mirroring the Python class "timedelta", the documentation can be found here: https://docs.python.org/3/library/datetime.html

Every functionality added here is compatible with CPython, but the constructor doesn't support floats yet.

A second constructor was made (it's not in python) for performance purposes, allowing one to skip all the math to convert between the different time units.

It's in draft but if the maintainers want to leave some comments, they are most welcome since only tests need some improvements.

This PR assumes that https://github.com/modularml/mojo/pull/2097 will be approved and merged in the future.


gryznar:
Please consider firstly this issue: https://github.com/modularml/mojo/issues/2164 and this one: https://github.com/modularml/mojo/issues/2113 before mirroring Python 1:1.

JoeLoser:
Hey @gabrieldemarmiesse, thank you for seeding an implementation of this `datetime.timedelta`.  As a team, we met yesterday to discuss this.

**Decision**
- Hold off for now

**Rationale**
- We'd like a unified understanding of how we're going to fill in the rest of this new `datetime` module before we commit to it.  For example, we can imagine a minimal `datetime` support in the core Mojo standard library, but perhaps the rest may be better suited as a 3rd party package which would be seamless to slurp in once we have that Mojo Package Manager in place down the road.
- We'd like to see some "reference call sites" to understand some motivated use cases in real world Mojo code before proceeding.
- We worry a bit about opening the flood gates here if we accept this PR and new PRs are opened just for "Python compatibility sake", if you will.  We're still figuring out some of the core details of how we're going to approach Mojo -> Python compatibility story.  So, we would rather walk a little slower here right now until we're further along and can share more details in the coming months.

I understand  this may not be the decision you were hoping for, but I don't want you to be discouraged.  Consider this more of a "not now" or "not quite yet" rather than a "never", please.

gabrieldemarmiesse:
Closing this PR. We'll reopen it when it is the right time for the stdlib to expand to new modules.

PR #2193, opened 2024-04-04
[tools] Add a `pre-commit` for `mojo format` and license check

gabrieldemarmiesse:
We currently rely on contributors applying themselves the `mojo format -l 80` and thinking themselves about the licence check. This PR adds a [pre-commit](https://pre-commit.com/), which, once installed, will do the mojo format and license check automatically before each commit. 

This is optional for contributors but useful, since, if later we add more steps in the pre-commit, they will automatically run (both on the contributors' laptops and in the CI). 

This should save some time for the maintainers as less mistakes will be made and the CI will check more stuff. Here, the CI will fail if the license is not present on every file.

I also simplified the contributing guide, notably removed the section about whitespaces, and line lenght which is already enforced by `mojo format`.

This also means that I had to fix all files missing the license (3) and all Mojo files which had an incorrect format (2)


.github/workflows/standard_library_tests_and_examples.yml
```diff
@@ -76,48 +76,14 @@ jobs:
...
-            git diff --name-only
-            echo -e "\nFull diff:\n"
-            git diff
-            exit 1
-          fi
+          pip install pre-commit
+          pre-commit install
+      
+      - name: Run pre-commit
+        run: pre-commit run --all-files
```
hamirmahal, line 86:
```suggestion
      - run: pre-commit run --all-files
```

hamirmahal (changes requested):

.github/workflows/standard_library_tests_and_examples.yml
gabrieldemarmiesse: Can you expand on why you would want this change?

hamirmahal: Sure! I think it's just simpler and thus easier to maintain.

hamirmahal: Usually `name` can provide some sort of documentation for a series of complex steps, but here, there's only 1 step, and that step says almost the same thing as `name`, so this documentation seems unnecessary.

gabrieldemarmiesse: Thank you for the opinion. I believe having steps in plain english helps the readability for people who are no familiar with the CI process. So I'd like to keep it. If the mainainers ask me to remove it, I'll comply though.

hamirmahal: You're welcome! That's fair.

hamirmahal: You're welcome! That's fair.

JoeLoser, line 86: **Note** This is a slight change in behavior if I understand correctly as previously we only ran `mojo format` on the select set of files changed in a PR.  This probably isn't a big deal since we don't have that many Mojo files in practice, but just calling it out.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
+repos:
+  - repo: local
+    hooks:
+      - id: mojo-format
+        name: mojo-format
+        entry: mojo format -l 80
```
JoeLoser, line 6: **Suggestion** Remove the `-l 80` as `mojo format` properly defaults to 80 characters line length now.  See https://github.com/modularml/mojo/pull/2244.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+        language: system
+        files: '\.(mojo|🔥)$'
+        stages: [commit]
+      - id: check-license
+        name: check-license
+        entry: mojo stdlib/scripts/check_license.mojo
```
JoeLoser, line 12: **Quibble** Perhaps rename to `check-licenses.mojo` (dash instead of underscore) for consistency in the other script names?  Don't feel strongly.

stdlib/docs/development.md
```diff
@@ -112,15 +112,23 @@ Otherwise, CI will fail in its lint and formatting checks.  The `mojo` compiler
 provides a `format` command.  So, you can format your changes like so:
 
 ```bash
-mojo format -l 80 <file1> <file2> ...
+mojo format -l 80 ./
```
JoeLoser, line 115: **Suggestion** Remove the parts about `-l 80` - see https://github.com/modularml/mojo/pull/2244. Here and elsewhere in this file when you rebase.

stdlib/scripts/check_license.mojo
```diff
@@ -0,0 +1,48 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+import sys
+from pathlib import Path
+
+# We can't check much more than this at the moment, because the license year
```
JoeLoser, line 17: 👍 if we cared enough, we could just write the script in Python instead of Mojo, but what you have is totally fine for our current needs IMO.

stdlib/scripts/check_license.mojo
```diff
@@ -0,0 +1,48 @@
...
+).strip()
+
+
+def main():
+    target_paths = sys.argv()
+    if len(target_paths) < 2:
+        raise Error("A file path must be given as a command line argument.")
+
+    one_file_failed = False
+    for i in range(len(target_paths)):
```
JoeLoser, line 33: **Thought** Wonder if it makes sense to add a `range` overload for `Sized` things like we do for `Intable` to avoid `range(len(x))` patterns.  Nothing to do here for this PR.  In this case, `sys.argv()` returns a `VariadicList` which implements `Sized`.

stdlib/scripts/check_license.mojo
```diff
@@ -0,0 +1,48 @@
...
+        if not file_path.read_text().startswith(LICENSE):
+            print(
+                "The license has been forgotten at the top of the file `"
+                + str(file_path)
+                + "`, please add it before commiting. "
+            )
+            one_file_failed = True
+
+    if one_file_failed:
+        print("At least one file is missing the license 💥 💔 💥")
```
JoeLoser, line 47: **Suggestion** We could collect the list of filenames missing licenses and print them all out here.  WDYT?

JoeLoser (commented):
Awesome, thank you for doing this!  Just a few comments.  Please rebase and then it should be good to go.

.github/workflows/standard_library_tests_and_examples.yml
gabrieldemarmiesse: It is extremely fast, 6 seconds. You can take a look in the CI.

.pre-commit-config.yaml
gabrieldemarmiesse: Ok, sure. Just noting that it seems we're deviating from black's default (88 chars per line) I hope it's not going to annoy people who migrate their python code to mojo in the future.

stdlib/scripts/check_license.mojo
gabrieldemarmiesse: Sure, I don't mind. I'll do this.

stdlib/scripts/check_license.mojo
gabrieldemarmiesse: in the future, we'll have iterators and replace all this by `for x in target_paths` so I don't think it's worth the effort

gabrieldemarmiesse:
Here is what it looks like when the pre-commit fails:
```
$ pre-commit run --all-files
mojo-format..............................................................Passed
check-license............................................................Failed
- hook id: check-license
- exit code: 1

The following files have missing licences 💥 💔 💥
--> stdlib/src/pathlib/__init__.mojo
Please add the license to each file before commiting.
The following files have missing licences 💥 💔 💥
--> stdlib/src/pathlib/path.mojo
Please add the license to each file before commiting.
```

stdlib/scripts/check_license.mojo
gabrieldemarmiesse: Done

.pre-commit-config.yaml
gabrieldemarmiesse: Done

.github/workflows/standard_library_tests_and_examples.yml
gabrieldemarmiesse: I added a comment to give a workaround if this ever gets a problem in the future

JoeLoser: Awesome — glad it's fast 😄

JoeLoser (approved):

.pre-commit-config.yaml
JoeLoser: I agree. We can lazily evaluate that in the future once the language and tools mature.  Right now we're quite a ways from the direct migration of Python to Mojo.

JoeLoser:
Thanks so much for doing this, @gabrieldemarmiesse — very much appreciated! ❤️

PR #2195, opened 2024-04-04
[Scripts] Add override for compiler version checking

JoeLoser:
Add an override ability for doing the compiler version checking.  This is opt-in based on the environment variable `MOJO_OVERRIDE_COMPILER_VERSION_CHECK`.  By default, we still do the compiler version checking.  Internally, this allows us to test to make sure the workflow upstream works for unreleased compilers that, by definition, would have a different compiler version than the one in the `COMPATIBLE_COMPILER_VERSION` file.

### DIFF
--- a/stdlib/scripts/build-stdlib.sh
+++ b/stdlib/scripts/build-stdlib.sh
@@ -22,12 +22,14 @@ mkdir -p "${BUILD_DIR}"
 ACTUAL_COMPILER_VERSION=$(mojo --version | tr " " "\n" | sed -n 2p)
 EXPECTED_COMPILER_VERSION=$(<"${REPO_ROOT}"/stdlib/COMPATIBLE_COMPILER_VERSION)
 
-if [ "${EXPECTED_COMPILER_VERSION}" != "${ACTUAL_COMPILER_VERSION}" ]; then
-  echo "Mismatch in compiler versions! Cannot build the standard library."
-  echo "Expected compiler version: ${EXPECTED_COMPILER_VERSION}"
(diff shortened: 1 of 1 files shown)


stdlib/scripts/build-stdlib.sh
```diff
@@ -22,12 +22,14 @@ mkdir -p "${BUILD_DIR}"
 ACTUAL_COMPILER_VERSION=$(mojo --version | tr " " "\n" | sed -n 2p)
 EXPECTED_COMPILER_VERSION=$(<"${REPO_ROOT}"/stdlib/COMPATIBLE_COMPILER_VERSION)
 
-if [ "${EXPECTED_COMPILER_VERSION}" != "${ACTUAL_COMPILER_VERSION}" ]; then
-  echo "Mismatch in compiler versions! Cannot build the standard library."
-  echo "Expected compiler version: ${EXPECTED_COMPILER_VERSION}"
-  echo "Current installed compiler version: ${ACTUAL_COMPILER_VERSION}"
-  echo "Please run modular update nightly/mojo to get the latest compiler."
-  exit 1
+if [ -z "${MODULAR_OVERRIDE_COMPILER_VERSION_CHECK:-}" ]; then
```
River707, line 25: Maybe `MOJO_STDLIB_OVERRIDE_COMPILER_VERSION_CHECK`?

River707 (approved):

stdlib/scripts/build-stdlib.sh
JoeLoser: Renamed 👍

PR #2214, opened 2024-04-05
[mojo-stdlib] Implement Corresponding Regular Methods for 4 Overloaded `Set` Operators

arvindavoudi:
## Implement Corresponding Regular Methods for 4 Overloaded `Set` Operators

This pull request ensures that, alongside the existing functionality of overloaded operators for `Set`, their corresponding regular methods are also fully implemented as they are in Python.

The following methods have been introduced to correspond with their respective operators:

- `__sub__` (-): Now corresponds to `Set.difference()`
- `__isub__` (-=): Now corresponds to `Set.difference_update()`
- `__iand__` (&=): Now corresponds to `Set.intersection_update()`
- `__ior__` (|=): Now corresponds to `Set.update()`

Previously, these magic methods lacked their equivalent regular methods. This update rectifies that by ensuring that calls to the magic methods are properly linked to their corresponding regular methods.

Note: No changes are needed for the `test_set.mojo` file, as this update merely relocates the existing code for the overloaded operators to their respective regular functions without adding any new functionality that needs to be tested.

### DIFF
--- a/stdlib/src/collections/set.mojo
+++ b/stdlib/src/collections/set.mojo
@@ -310,7 +310,7 @@ struct Set[T: KeyElement](Sized, EqualityComparable, Hashable, Boolable):
                 result.add(v[])
 
         return result^
-    
+
     fn difference(self, other: Self) -> Self:
         """Set difference.
 
@@ -326,7 +326,7 @@ struct Set[T: KeyElement](Sized, EqualityComparable, Hashable, Boolable):
             if e[] not in other:
                 result.add(e[])
         return result^
-    
+
     fn update(inout self, other: Self):
         """In-place set update.
 
@@ -338,7 +338,7 @@ struct Set[T: KeyElement](Sized, EqualityComparable, Hashable, Boolable):
         """
         for e in other:
             self.add(e[])
-    
+
     fn difference_update(inout self, other: Self):
         """In-place set difference update.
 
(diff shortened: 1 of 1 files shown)


arvindavoudi:
Since only the `mojo-format` test was not successful, I ran the `mojo format -l 80 set.mojo` in the terminal on my local machine. It seems the formatting error is caused by the lines where a variable name is next to the `^` operator without any whitespaces between. Like what we initially had in method `__moveinit__`.

As far as the initial formatting for these lines is kept the same and not changed at all, It leads me to speculate whether there's a conflict within the `mojo format` command. Specifically, the conflict might arise between the use of `var1 ^ var2` (where ^ signifies the bitwise XOR operation, handled by the `__xor__` magic method, which needs whitespace indeed) and the use of the `^` operator for indicating ownership transfer in Mojo's syntax (which does not need whitespace).

Should I open a new issue for it @JoeLoser? 

Thanks in advance.

arvindavoudi:
I also have removed the whitespaces that were incorrectly added at the beginning of some lines but I think the problem with requiring whitespace before `^` persists.

According to the stdlib codes which do not follow this whitespace rule before `^`, I haven't made any changes to it yet, as I'm waiting for your review.

arvindavoudi:
@JoeLoser It's alright now.

JoeLoser (approved):
Thanks for the contribution! 🎉 🔥

PR #2215, opened 2024-04-06
[stdlib] Add `reversed`

helehex:
Adds an initial implementation of `reversed()`, for getting a reversed iterator of a range or list.

This isn't meant to be final, since iterators are not fully worked out yet, but it works for now.

Changes:
- Added submodule `/builtin/reversed.mojo` 
- Added function `reversed()`
- Added trait `ReversibleRange`
- Implemented `ReversibleRange` for ranges
- Added method `__reversed__()` in `List`
- Changed `_ListIter` to allow for backwards iteration
- Added tests for ranges and reversed list iterator
- Fixed original ranges with negative size not passing the new tests

### DIFF
--- a/stdlib/src/builtin/range.mojo
+++ b/stdlib/src/builtin/range.mojo
@@ -49,6 +49,15 @@ fn _abs(x: Int) -> Int:
 
 
 @always_inline
+fn _sign(x: Int) -> Int:
+    if x > 0:
+        return 1
+    if x < 0:
+        return -1
+    return 0
+
+
+@always_inline
 fn _max(a: Int, b: Int) -> Int:
     return a if a > b else b
 
@@ -59,14 +68,14 @@ fn _max(a: Int, b: Int) -> Int:
 
 
 @register_passable("trivial")
-struct _ZeroStartingRange(Sized):
+struct _ZeroStartingRange(Sized, ReversibleRange):
     var curr: Int
     var end: Int
 
     @always_inline("nodebug")
     fn __init__(inout self, end: Int):
         self.curr = _max(0, end)
-        self.end = self.curr
+        self.end = end
 
     @always_inline("nodebug")
     fn __iter__(self) -> Self:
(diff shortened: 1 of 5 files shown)


helehex:
oups i had a premature submit

helehex:
The fix for ranges having negative length is mostly unnecessary since it never really matters in practice, so that could be changed back to avoid the extra operations.
Just have to remove the edge case section of `test_range_len()` in `test_range.mojo` aswell.

helehex:
`_StridedRangeIterator` now has a `__iter__()` method which returns itself.
So `reversed(range(...))` behaves as expected with `for` loops.

Somehow i missed that when testing, because i started off by having reversed() return a non iterator strided range.

helehex:
that looping test was a little excessive, it makes more sense to check the sums of both ranges.

helehex:
It occurred to me that i never checked this with the `@unroll` decorator, so i checked and everything seems to work good

JoeLoser (approved):
This is a huge step forward, even though we haven't fully worked out the long-term iterator support in the standard library.  Thank you so much for working on this! 🎉 

To land this, you'll need to rebase to resolve a few conflicts, but after that — happy to see it go in.

As a follow-up, we can go implement reverse iterator support for `Set` and `Dict`, right?  I'm happy to create some GitHub issues for that if anyone is interested in working on that.

jayzhan211:
> As a follow-up, we can go implement reverse iterator support for `Set` and `Dict`, right? I'm happy to create some GitHub issues for that if anyone is interested in working on that.

I'm interested to help on `Set` and `Dict`.

helehex:
ok i added the suggestions, and moved all the reversed range tests into one function (it was kind of split before)

JoeLoser:
> ok i added the suggestions, and moved all the reversed range tests into one function (it was kind of split before)

Merged! Thanks so much for this great contribution — we really appreciate it! 🎉

lattner:
This is great, thank you!

PR #2225, opened 2024-04-06
[stdlib] Handle whitespaces in `atol`

artemiogr97:
"atol" function now handles whitespaces before and after the number to parse such that int(String(" 10 ")) now gives 10 instead of raising an error.
Signed-off-by: Artemio Gazra Reyna <artemiogr97@gmail.com>

### DIFF
--- a/stdlib/test/builtin/test_string.mojo
+++ b/stdlib/test/builtin/test_string.mojo
@@ -200,11 +200,21 @@ fn test_string_indexing() raises:
 fn test_atol() raises:
     assert_equal(375, atol(String("375")))
     assert_equal(1, atol(String("001")))
+    assert_equal(5, atol(String(" 005")))
+    assert_equal(13, atol(String(" 013  ")))
     assert_equal(-89, atol(String("-89")))
+    assert_equal(-52, atol(String(" -52")))
+    assert_equal(-69, atol(String(" -69  ")))
 
     # Negative cases
     try:
         _ = atol(String("9.03"))
+        raise Error("Failed to raise when converting string to integer.")
+    except e:
+        assert_equal(str(e), "String is not convertible to integer.")
+
+    try:
+        _ = atol(String(" 10 1"))
         raise Error("Failed to raise when converting string to integer.")
     except e:
         assert_equal(str(e), "String is not convertible to integer.")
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Would it make sense to implement `.strip()` on `StringRef` and use this method here? See  https://docs.modular.com/mojo/stdlib/builtin/string#strip It could be more readable but maybe there something i'm missing.

artemiogr97:
@gabrieldemarmiesse I tough about just calling the `strip` method in the `String` class, but it completely depends on whether `atol` should handle whitespaces or not, python 3 does not have an `atol` function, but I think python 2 did, maybe we could consider that as the reference for mojo's implementation?
Also if you use `strip` a copy would be needed which would not be the best performance-wise (but I didn't benchmark it to really know for sure)

stdlib/test/builtin/test_string.mojo
```diff
@@ -209,6 +213,12 @@ fn test_atol() raises:
     except e:
         assert_equal(str(e), "String is not convertible to integer.")
 
+    try:
```
JoeLoser, line 216: **Suggestion** You can use [assert_raises](https://github.com/modularml/mojo/blob/e98f496c3a0d3565f1d6453412d5530894f4bc1e/stdlib/src/testing/testing.mojo#L272) from the `testing` module.  If you'd rather defer this and potentially refactor the other uses in this file to use `assert_raises`, that's totally fine.

JoeLoser:
> @gabrieldemarmiesse I tough about just calling the `strip` method in the `String` class, but it completely depends on whether `atol` should handle whitespaces or not, python 3 does not have an `atol` function, but I think python 2 did, maybe we could consider that as the reference for mojo's implementation? Also if you use `strip` a copy would be needed which would not be the best performance-wise (but I didn't benchmark it to really know for sure)

In my opinion, it's reasonable for `atol` to work as you have it where it parses the number out even if there is whitespace present.  This would then match the behavior of `atol` in C for example.  As for whether to use `strip()` to clean up the code, I'm OK personally with as you have it, but feel free to drop a TODO to consider using `strip()` instead, but as you pointed out, you'd incur a copy that can be avoided.

JoeLoser (approved):
Great.  Thank you for the contribution! 🚀

stdlib/test/builtin/test_string.mojo
artemiogr97: Oh, I didn't know assert_raises was already implemented, I'll probably change that I'm these days

Issue #2179, opened 2024-04-04
[stdlib] Optional.value should return a Reference


lsh:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

The Optional docs for `value()` state that it should return a `Reference` when that is possible. Since the underlying `Variant` already returns a reference which is then dereferenced, this should be possible now.

### What is your motivation for this change?

This is a planned feature per the comment on the method.

StandinKP:
I will take this task up. Thanks

StandinKP:
In this what should be passed into the lifetime component and the mutability component of the Reference?

StandinKP:
@lsh I was working on this

lsh:
@StandinKP didn't mean to step on any toes. This is a feature I wanted for a project, and I knew roughly how to wire up all the lifetimes and mutability in `Reference`, but am happy to close out in favor of whatever you have.

StandinKP:
> @StandinKP didn't mean to step on any toes. This is a feature I wanted for a project, and I knew roughly how to wire up all the lifetimes and mutability in `Reference`, but am happy to close out in favor of whatever you have.

Nah man don't worry. Don't close it. I'll review the PR and let you know if any comments. Next time it would be good to ask so we won't be repeating efforts and we can grow this faster🤟🏻🤩

PR #2226, opened 2024-04-06
[mojo-stdlib] `Optional.value()` returns a `Reference`

lsh:
Addresses #2179

### DIFF
--- a/stdlib/src/collections/optional.mojo
+++ b/stdlib/src/collections/optional.mojo
@@ -105,10 +105,7 @@ struct Optional[T: CollectionElement](CollectionElement, Boolable):
     ](self: Reference[Self, mutability, self_life].mlir_ref_type) -> Reference[
         T, mutability, self_life
     ]:
-        """Unsafely retrieve the value out of the Optional.
-
-        This function currently creates a copy. Once we have lifetimes
-        we'll be able to have it return a reference.
+        """Unsafely retrieve a reference to the value of the Optional.
 
         This doesn't check to see if the optional contains a value.
         If you call this without first verifying the optional with __bool__()
@@ -116,12 +113,14 @@ struct Optional[T: CollectionElement](CollectionElement, Boolable):
         value (for instance with `or_else`), you'll get garbage unsafe data out.
 
         Returns:
-            The contained data of the option as a T value.
+            A reference to the contained data of the option as a Reference[T].
         """
         debug_assert(Reference(self)[].__bool__(), ".value() on empty Optional")
-        return __mlir_op.`lit.ref.from_pointer`[
-            _type = Reference[T, mutability, self_life].mlir_ref_type
-        ](Reference(self)[]._value._get_ptr[T]().value)
+        alias RefType = Reference[T, mutability, self_life]
+        var ptr = Reference(self)[]._value._get_ptr[T]().value
+        return __mlir_op.`lit.ref.from_pointer`[_type = RefType.mlir_ref_type](
+            ptr
+        )
 
     fn take(owned self) -> T:
         """Unsafely move the value out of the Optional.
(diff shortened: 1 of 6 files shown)


stdlib/test/collections/test_dict.mojo
```diff
@@ -182,14 +182,14 @@ def test_dict_copy_add_new_item():
...
-    orig["a"] = CopyCounter()^
+    orig["a"] = CopyCounter() ^
 
     # test values copied to new Dict
     var copy = Dict(orig)
     # I _may_ have thoughts about where our performance issues
     # are coming from :)
-    assert_equal(5, orig["a"].copy_count)
-    assert_equal(6, copy["a"].copy_count)
+    assert_equal(4, orig["a"].copy_count)
```
lsh, line 191: The copy count went down by 1 from these changes. Is that expected?

stdlib/src/collections/optional.mojo
```diff
@@ -100,7 +100,11 @@ struct Optional[T: CollectionElement](CollectionElement, Boolable):
...
     @always_inline
-    fn value(self) -> T:
+    fn value[
+        mutability: __mlir_type.`i1`, self_life: AnyLifetime[mutability].type
+    ](self: Reference[Self, mutability, self_life].mlir_ref_type) -> Reference[
+        T, mutability, self_life
+    ]:
         """Unsafely retrieve the value out of the Optional.
 
         This function currently creates a copy. Once we have lifetimes
```
JoeLoser, line 110: **Suggestion** Please update the doc strings here as we no longer create a copy 😄

stdlib/src/collections/optional.mojo
```diff
@@ -114,8 +118,10 @@ struct Optional[T: CollectionElement](CollectionElement, Boolable):
         Returns:
             The contained data of the option as a T value.
         """
-        debug_assert(self.__bool__(), ".value() on empty Optional")
-        return self._value.get[T]()[]
+        debug_assert(Reference(self)[].__bool__(), ".value() on empty Optional")
+        return __mlir_op.`lit.ref.from_pointer`[
+            _type = Reference[T, mutability, self_life].mlir_ref_type
+        ](Reference(self)[]._value._get_ptr[T]().value)
```
JoeLoser, line 124: **Suggestion** For readability sake, it may be nicer to pull this out into a local variable:

```
var ptr = Reference(self)[]._velue.get_ptr[T]().value
```

And then just pass `ptr` to the argument when we call `__mlir_op.`lit.ref.from_pointer`.  WDYT?

stdlib/test/collections/test_dict.mojo
JoeLoser: Yep, this is expected since we no longer return a copy of the `T` from `Optional`.  Slowly getting our copies down to a more sane amount in `Dict` which is one of the perf problems in `Dict` right now. 😄

stdlib/test/collections/test_optional.mojo
```diff
@@ -52,6 +52,10 @@ def test_basic():
 
     assert_equal(1, (a^).take())
 
+    var a2 = Optional(1)
```
JoeLoser, line 55: **Thought** This is great for testing the non-const reference case.  We don't have a way to test that we return a "const reference to T" in `Optional.value()` right now since we got rid of `let` recently.  Do you mind dropping a `TODO` potentially to come back to this to add a "const reference" test when we have the language facilities to do so?

JoeLoser (approved):
Generally looks good - just need to update the API doc strings.  Thanks for doing this!

JoeLoser:
@lsh do you mind signing your commits, please?

Moosems:
Auto dereferencing 🎉🎉

PR #2244, opened 2024-04-08
Revert "[docs] Fix wrong `mojo format` command, add `-l 80`. (#2158)"

JoeLoser:
This reverts commit e305d91c2688432bacac4e72918d66541f57fec1.  As of the latest `nightly/mojo`, `mojo format` properly defaults its line length to `80` now instead of `88`.  This change didn't make it into the changelog yet, so I'll go write that up, but it did actually make it into today's `nightly/mojo` on April 8th.  So, go ahead and update the docs to no longer explicitly specify `-l 80` to `mojo format` everywhere.


arthurevans (approved):
LGTM.

Issue #2224, opened 2024-04-06
[BUG] `object` does not implement `__truediv__` or `__floordiv__`


Brian-M-J:
### Bug description

When running the code snippets linked below in the [Mojo Playground](https://docs.modular.com/mojo/playground), I get the following errors:

`__floordiv__` code:
```
/source/prog.mojo:2:20: error: 'object' does not implement the '__floordiv__' method
    return (x + y) // 2
           ~~~~~~~ ^
mojo: error: failed to parse the provided Mojo
```
It is supposed to print `1`

`__truediv__` code:
```
/source/prog.mojo:2:20: error: 'object' does not implement the '__truediv__' method
    return (x + y) / 2
           ~~~~~~~ ^
mojo: error: failed to parse the provided Mojo
```
It is supposed to print `1.5`

I consider this to be a bug as the exact same code works perfectly fine in Python.

### Steps to reproduce

- Include relevant code snippet or link to code that did not work as expected.

[`__truediv__` code](https://gist.github.com/modularbot/4aef64f85bccd1ad0fb1294a7a487e9f)

[`__floordiv__` code](https://gist.github.com/modularbot/64ca9ee9ce38d14552d79aa18ebeade2)

Just run the above snippets in the Playground.

### System information

```shell
I ran this on the Mojo Playground as of 06-04-2024.
```

LJ-9801:
this seems to be in the TODO list of the object struct implementation 
https://github.com/modularml/mojo/blob/738901dec1058612d8f01fd13e13a3e09103944f/stdlib/src/builtin/object.mojo#L1326
https://github.com/modularml/mojo/blob/738901dec1058612d8f01fd13e13a3e09103944f/stdlib/src/builtin/object.mojo#L1327

There are other dunder methods such as mod, lshiift and rshift that has not been implemented yet. I can open a PR for now for the truediv and div?

JoeLoser:
@LJ-9801 - these look like they all need to be implemented still for the builtin object type.  I'd definitely welcome a contributor for any/all that you're up for with this type.  Thank you for looking into this!

JoeLoser:
Fixed in https://github.com/modularml/mojo/commit/dbe3deed799b032b5243de291f4a2f316029bb31

PR #2247, opened 2024-04-09
[mojo-stdlib] implement shifts in object

jiex-liu:
This is for implementing shift operation for object #2224

### DIFF
--- a/stdlib/src/builtin/object.mojo
+++ b/stdlib/src/builtin/object.mojo
@@ -1356,6 +1356,34 @@ struct object(IntableRaising, Boolable, Stringable):
     # TODO __rshift__
 
     @always_inline
+    fn __lshift__(self, rhs: object) raises -> object:
+        """Left shift operator. Valid only for arithmetic types.
+
+        Args:
+            rhs: Right hand value.
+
+        Returns:
+            The left hand value left shifted by the right hand value.
+        """
+        return Self._arithmetic_binary_op[Float64.__lshift__, Int64.__lshift__](
+            self, rhs
+        )
+
+    @always_inline
+    fn __rshift__(self, rhs: object) raises -> object:
+        """Right shift operator. Valid only for arithmetic types.
+
+        Args:
+            rhs: Right hand value.
+
+        Returns:
+            The left hand value right shifted by the right hand value.
+        """
+        return Self._arithmetic_binary_op[Float64.__rshift__, Int64.__rshift__](
+            self, rhs
+        )
+
+    @always_inline
     fn __and__(self, rhs: object) raises -> object:
         """Bool AND operator. If the left hand value is False, return the
         left-hand value.
--- a/stdlib/src/builtin/object.mojo
+++ b/stdlib/src/builtin/object.mojo
@@ -582,6 +582,18 @@ struct _ObjectImpl(CollectionElement, Stringable):
         else:
             lhs = lhs.convert_int_to_float()
 
+    @staticmethod
+    fn coerce_integral_type(inout lhs: _ObjectImpl, inout rhs: _ObjectImpl):
+        """Coerces two values of integral type to the appropriate
+        lowest-common denominator type for performing bitwise operations.
+        """
+        if lhs.is_int() == rhs.is_int():
+            return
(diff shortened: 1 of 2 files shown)


LJ-9801:
@JoeLoser Hi, I have a questions in regards to all the bitwise operation for object. Python's behavior for shift and bitwise and/or/xor is different in that python will always cast the type of shift to an int where as the other bitwise operation will depend on the input types. 

In this pr I kind of just do a type check and then return a new object of type int for all the shift operator. Mean while I also found a lot of other bitwise operator such as and/or/xor that only support bool type
https://github.com/modularml/mojo/blob/abf2d97e88a2cc97e3029a578d2824895a88d0c1/stdlib/src/builtin/object.mojo#L1315
https://github.com/modularml/mojo/blob/abf2d97e88a2cc97e3029a578d2824895a88d0c1/stdlib/src/builtin/object.mojo#L1330

So I create another function name `_arithmetic_bitwise_op()` that handles these kind of operator for both int and bool. I did not use this function for the shift operation as it is not necessary, but it would be very helpful for completing the rest of the bitwise operator. 

Please let me know if it is OK to include that into this pr. The rest of the bitwise operator can be completed once this pr is merged. Thanks!

stdlib/src/builtin/object.mojo
```diff
@@ -577,6 +577,18 @@ struct _ObjectImpl(CollectionElement, Stringable):
         else:
             lhs = lhs.convert_int_to_float()
 
+    @staticmethod
+    fn coerce_integral_type(inout lhs: _ObjectImpl, inout rhs: _ObjectImpl):
```
JoeLoser, line 581: **Suggestion** Consider making this private (prefix with `_`).  Then, doc-gen won't generate public API docs for this as users shouldn't call this directly.

stdlib/src/builtin/object.mojo
```diff
@@ -1198,6 +1216,30 @@ struct object(IntableRaising, Boolable, Stringable):
...
+            bool_func: Boolean operator.
+
+        Returns:
+            The bitwise operation result.
+        """
+        lhs._arithmetic_integral_type_check()
+        rhs._arithmetic_integral_type_check()
+        var lhsValue = lhs._value
+        var rhsValue = rhs._value
+        _ObjectImpl.coerce_integral_type(lhsValue, rhsValue)
```
JoeLoser, line 1238: **Suggestion** Instead of `_ObjectImpl` you can use `Self` which is the same thing.

JoeLoser (approved):

JoeLoser:
Looks good, thank you!  You'll need to rebase to resolve the test file conflict before we can merge.

ConnorGray:
Hi @LJ-9801, thanks for this contribution! Joe and I talked about your question above re. bitwise operations. What you said about the other bitwise operators makes sense, but since it's separate from the problem of supporting shift operators, it would make review and discussion easier if we make those changes in a separate PR. This PR looks good as is though :)

Since `_arithmetic_bitwise_op` is not used in this PR, I suggest we leave it out of thise PR, and include it in a follow-up PR making the changes to the other bitwise operators you mentioned.

Thanks again for this contribution, very usefull! :fire:

LJ-9801:
@ConnorGray Yeah sure! i will do that.

stdlib/src/builtin/object.mojo
LJ-9801: @JoeLoser  I implement this function inside the `_ObjectImpl` struct which I believe is private in nature right?

JoeLoser: Oops, you're right here - sorry I missed that! 👍

LJ-9801:
@JoeLoser @ConnorGray the function has been removed from this PR. I will open a issue for this as well as all the bitwise operation!

JoeLoser (approved):

JoeLoser:
> @JoeLoser @ConnorGray the function has been removed from this PR. I will open a issue for this as well as all the bitwise operation!

Sounds good, feel free to do that.  I've just merged this PR.  Thanks again for the contribution! 🎉

PR #2250, opened 2024-04-09
[Proposal] Improve the hash module

mzaks:
This proposal is based on discussion started in https://github.com/modularml/mojo/issues/1744

### DIFF
--- a/proposals/imporved-hash-module.md
+++ b/proposals/imporved-hash-module.md
@@ -1,0 +1,134 @@
+# Improve hash module
+
+Current implementation of the hash module in standard library reflex the implementation of the Python hash module, which by itself is a good idea, but it has some flaws, which we should correct in Mojo.
+
+# Flaws of the `Hashable` trait
+
+The `Hashable` trait is designed as following:
+
+```mojo
+trait Hashable:
+    fn __hash__(self) -> Int:
+        ...
+```
+
+Which implies that a developer, who writes a new hashable struct needs to return an `Int`. 
+- This API does not provide guidance to how this `Int` value needs to be computed
+- It is impossible to exchange hashing algorithms on demand
+- `Int` as a type is variable length (based on the CPU arch), which might cause issues
+- Such trait design follows the call return principle which, when applied on complex types, will lead to unnecessary computations and memory allocations
+
+## Example
+
+```mojo
+@value
+struct Person(Hashable):
+    var name: String
+    var age: UInt8
+    var friends_names: List[String]
+
+    fn __hash__(self) -> Int:
+        var hashes = List[Int]()
+        hashes.append(hash(self.name))
+        hashes.append(hash(self.age))
+        for friend in self.friends_names:
+            hashes.append(hash(friend[]))
+        
+        # How to combine a hash of hashes ???
+```
+As you can see above we, computed hashes for all of the struct fields, but we are uncertain how to combine those values in a way which produces a good (non compromised) hash value.
+
+# Proposal
+
+In order to improve the hash module and address the flaws of the `Hashable` trait, we need to apply following steps.
+
+## Introduce a `Hasher` trait
+
+By introducing a `Hasher` trait we define an abstraction for the hashing algorithm itself, which allows streaming creation of the hash value. Bellow is a possible API of the `Hashable` trait.
+
+```mojo
+trait Hasher:
+    fn __init__(inout self):
+        ...
+    fn update[T: DType](inout self, value: SIMD[T, 1]):
+        ...
+    fn update(inout self, pointer: DTypePointer[DType.uint8], length: Int):
+        ...
+    fn finish(owned self) -> UInt64:
+        ...
+```
+
+## Implement a default `Hasher` in standard library
+
+The standard library should provide a default `Hasher` implementation, but it would be possible for the developers to implement, or choose other hash algorithms, if they better fit their use case.
+
+Bellow you can see a dummy implementation of a `DefaultHasher`
+
+```mojo
+struct DefaultHasher(Hasher):
+    var hash: UInt64
+
+    fn __init__(inout self):
+        self.hash = 42
+    fn update[T: DType](inout self, value: SIMD[T, 1]):
+        ...
+    fn update(inout self, pointer: DTypePointer[DType.uint8], length: Int):
+        ...
+    fn finish(owned self) -> UInt64:
+        return self.hash
+```
+
+## Redesign the `Hashable` trait to follow the data flow principles
+
+Given the `Hasher` trait, we can define the `Hashable` trait to adopt the data flow paradigm instead of call return.
+
+```mojo
+trait Hashable:
+    fn hash_with[H: Hasher](self, inout hasher: H):
+        ...
+```
+
+## Example for `Hashable` struct implementation
+
+The implementation of `Hashable`, where all the fields are `Hashable` is trivial and could be easily synthesized by the compiler. But even if not, the data flow API guides the developer toward a very simple solution
+
+```mojo
+@value
+struct Person(Hashable):
+    var name: String
+    var age: UInt8
+    var friends_names: List[String]
+
+    fn hash_with[H: Hasher](self, inout hasher: H):
+        # self.name.hash_with(hasher), when String is Hashable, otherwise:
+        hasher.update(self.name._as_ptr().bitcast[DType.uint8](), len(self.name))
+        # self.age.hash_with(hasher), when SIMD is hashable, otherwise
+        hasher.update(self.age)
+        # self.friends_names.hash_with(hasher), when List of Hashable types is Hashable, otherwise:
+        for friend in self.friends_names:
+            hasher.update(friend[]._as_ptr().bitcast[DType.uint8](), len(friend[]))
+```
+
+## Parameterized the `hash` function with `Hasher` type
+
+The `hash` function can be parameterized with the `Hasher` type, which gives the users control over the hashing algorithm.
+
+```mojo
+fn hash[T: Hashable, H: Hasher = DefaultHasher](value: T) -> UInt64:
+    var hasher = H()
+    value.hash_with(hasher)
+    return hasher^.finish()
+```
+
+Alternatively we could go with following API
+
+```mojo
+fn hash[T: Hashable, H: Hasher, hasher_factory: fn () -> H](value: T) -> UInt64:
+    var hasher = hasher_factory()
+    value.hash_with(hasher)
+    return hasher^.finish()
+```
+
+Where we allow complex and context dependent `Hasher` initialization.
+
+Similar parameterization should be applied for all hash based data structures like `Dict` and `Set`.
--- a/proposals/improved-hash-module.md
+++ b/proposals/improved-hash-module.md
@@ -48,13 +48,23 @@ By introducing a `Hasher` trait we define an abstraction for the hashing algorit
 
 ```mojo
 trait Hasher:
+    """Trait which every hash function implementer needs to implement."""
     fn __init__(inout self):
+        """Expects a no argument instantiation."""
         ...
-    fn update[T: DType](inout self, value: SIMD[T, 1]):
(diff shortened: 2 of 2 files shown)


proposals/imporved-hash-module.md
```diff
@@ -0,0 +1,134 @@
...
+```mojo
+struct DefaultHasher(Hasher):
+    var hash: UInt64
+
+    fn __init__(inout self):
+        self.hash = 42
+    fn update[T: DType](inout self, value: SIMD[T, 1]):
+        ...
+    fn update(inout self, pointer: DTypePointer[DType.uint8], length: Int):
+        ...
```
gabrieldemarmiesse, line 76: Can we add a comment saying that when Mojo supports it, we could have a default implementation for this?
Something like:
```suggestion
    fn update(inout self, pointer: DTypePointer[DType.uint8], length: Int):
        for i in range(length):
            self.update(pointer[i])
```
Then a user can override the method to implement something more efficient if they wish to do so.

theopomies:
Nice proposal 👍🏻
Please note there appears to be a typo in the filename: `imporved-hash-module.md` when it should actually be `improved-hash-module.md`.

proposals/imporved-hash-module.md
```diff
@@ -0,0 +1,134 @@
...
+        var hashes = List[Int]()
+        hashes.append(hash(self.name))
+        hashes.append(hash(self.age))
+        for friend in self.friends_names:
+            hashes.append(hash(friend[]))
+        
+        # How to combine a hash of hashes ???
+```
+As you can see above we, computed hashes for all of the struct fields, but we are uncertain how to combine those values in a way which produces a good (non compromised) hash value.
+
```
gabrieldemarmiesse, line 40: I believe to be fair, we should add an example of what the computing of the hash would look like if mojo evolves without this proposal. By following https://docs.python.org/3/reference/datamodel.html#object.__hash__ it would look like this

```mojo
@value
struct Person(Hashable):
    var name: String
    var age: UInt8
    var friends_names: List[String]
    fn __hash__(self) -> Int:
        var final_tuple = (self.name, self.age) + tuple(self.friends)
        return hash(final_tuple)
```
And this has a lot of flaw(lots of copying, possible memory allocation, unpredictable types possible for the tuple, use of the `object` struct, a struct must know how to hash it's members, etc...) but at least the proposal gives a fair chance to the "python's way" of doing the hashing.

proposals/imporved-hash-module.md
mzaks: See above, I think it is a detail not really important for this proposal.

proposals/imporved-hash-module.md
mzaks: I mentioned the "python's way" in the proposal.

proposals/imporved-hash-module.md
gabrieldemarmiesse: It was just a suggestion, fair enough

proposals/imporved-hash-module.md
```diff
@@ -0,0 +1,134 @@
...
+Bellow you can see a dummy implementation of a `DefaultHasher`
+
+```mojo
+struct DefaultHasher(Hasher):
+    var hash: UInt64
+
+    fn __init__(inout self):
+        self.hash = 42
+    fn update[T: DType](inout self, value: SIMD[T, 1]):
```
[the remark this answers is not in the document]
gabrieldemarmiesse: It's not the focus I agree, it's just helpful for readers like me who are not very familiar with hashing algorithms. Thanks for the proposal though. I think it's really good :)

proposals/improved-hash-module.md
```diff
@@ -0,0 +1,134 @@
...
+    var friends_names: List[String]
+
+    fn hash_with[H: Hasher](self, inout hasher: H):
+        # self.name.hash_with(hasher), when String is Hashable, otherwise:
+        hasher.update(self.name._as_ptr().bitcast[DType.uint8](), len(self.name))
+        # self.age.hash_with(hasher), when SIMD is hashable, otherwise
+        hasher.update(self.age)
+        # self.friends_names.hash_with(hasher), when List of Hashable types is Hashable, otherwise:
+        for friend in self.friends_names:
+            hasher.update(friend[]._as_ptr().bitcast[DType.uint8](), len(friend[]))
```
gabrieldemarmiesse, line 109: I propose an alternative here, as it allows the implementer of "hash_with" to use only one type of call instread of two, thus simplifying the API.
If we modify the `Hasher` trait to include a default implementation:
```mojo
trait Hasher:
    fn update(inout self, value: Hashable):
        # this is a default implementation and should not be changed
        value.hash_with[Self](self)
```
then, if the struct attribute has a `hash_with` method or not, the same method is called and users don't have to remember two different ways of calling the hasher:
```mojo
    fn hash_with[H: Hasher](self, inout hasher: H):
        # hasher.update(hasher), when String is Hashable, otherwise:
        hasher.update(self.name._as_ptr().bitcast[DType.uint8](), len(self.name))
        hasher.update(self.age)
        # hasher.update(hasher), when List of Hashable types is Hashable, otherwise:
        for friend in self.friends_names:
            hasher.update(friend[]._as_ptr().bitcast[DType.uint8](), len(friend[]))
```

proposals/improved-hash-module.md
```diff
@@ -0,0 +1,134 @@
...
+The implementation of `Hashable`, where all the fields are `Hashable` is trivial and could be easily synthesized by the compiler. But even if not, the data flow API guides the developer toward a very simple solution
+
+```mojo
+@value
+struct Person(Hashable):
+    var name: String
+    var age: UInt8
+    var friends_names: List[String]
+
+    fn hash_with[H: Hasher](self, inout hasher: H):
```
gabrieldemarmiesse, line 102: Python and Mojo both use dunder `__something__` to signify that methods have a special meaning in the stdlib and the language. It avoids naming conflicts with user methods in general. I don't have a strong opinion about this, I'm just pointing this out.
This method in this proposal shouldn't be a second class citizen compared to `__hash__`.

proposals/improved-hash-module.md
```diff
@@ -0,0 +1,134 @@
...
+## Example
+
+```mojo
+@value
+struct Person(Hashable):
+    var name: String
+    var age: UInt8
+    var friends_names: List[String]
+
+    fn __hash__(self) -> Int:
```
gabrieldemarmiesse, line 30: While the `__hash__` method has many flaws, what should we do with it once the proposal is implemented? Do we remove it and give up on backward compatibility? Can we make something work with the default hasher? Can someone implement both? How would that work? If you're open to the idea, I can maybe work out something that would allow a graceful migration.

mzaks: @gabrieldemarmiesse sorry for the late reply, I wanted to create a POC for the hasher first.

I think best case, we remove the `__hash__(self) -> Int` method. I think backwards compatibility is not a big deal at this point in time. Otherwise we can follow Swift convention, where the compiler handles implementation of the legacy hash method with a warning. Generally the `__hash__(self) -> Int` can be synthesized as following:
```mojo
fn __hash__(self) -> Int:
  return int(hash(self))
```
Given that `fn __hash__[H: Hasher](self, inout hasher: H):` is implemented see (https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa#file-main-mojo-L2)

And as the standard library will implement the hash function for all the basic types a hash function for a type where its fields are all hashable is also very simple to synthesize as you can see here: https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa#file-main-mojo-L35

By synthesize, I mean that the compiler could generate the function when the struct has is properly decorated, similar to what happens with `@value`.

proposals/improved-hash-module.md
mzaks: I agree, I created a small POC, where I changed the method names. https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa

If it is ok with everyone I will change the snippets in the proposal and reference the gist at the end.

proposals/improved-hash-module.md
mzaks: I simplified the API and provide a working example in https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa
Let me know if it resolves your concerns.

mzaks:
I created a small POC to address all the outstanding questions:
https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa

> Should we create an alias for the return type of finish(...) (e.g. alias HASH_RETURN_T = UInt64) rather than hard-coding to UInt64? This gives us flexibility to change the return type in the future or have different hashers return different sized things if they care about space savings in certain cases.

I suggest to parametrize the return type of `finish` with a default, (see)[https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa#file-main-mojo-L73]. It might be sensible to put the hash value type parameter directly on the Hasher trait, which would allow Hasher to have different implementations for different DTypes, not just a bitcast at the end.

```mojo
trait Hasher[hash_value_type: DType]:
    fn __init__(inout self):
        ...
    fn update(inout self, bytes: DTypePointer[DType.uint8], n: Int):
        ...
    fn finish(owned self) -> Scalar[hash_value_type]:
        ...
```
I did not do it at this point in time mainly because the compiler does not allow default parameters on traits. (`trait Hasher[hash_value_type: DType = DType.uint64]` does not work)

> We prefer the hash_with approach defaulting to the DefaultHasher rather than the hasher_factory approach. If you share more motivation/context on why the function approach may be useful, we may be better equipped to chime in.

I prefer it too now. I though of factory approach, because some Hashers might have complex initialization needs, e.g. often they need a random seed and a secret. But as you can see (here)[https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa#file-main-mojo-L61] I was able to come up with a strategy where users can influence the hasher behavior with compile time parameters and environment variables.

> When implementors (and users) are writing user-defined-types, it would be nice for them to not have to always worry about init, update, and finish to show up in their code: i.e. they only worry about hash or hash_with for example rather than the full API set for the trait requirements.

Yes, as you can see (here)[https://gist.github.com/mzaks/aa66c831dc5e177c2322d5088aac76aa#file-main-mojo-L35] it is trivial to implement the hash method when standard library implements hash methods for the basic types. As I mentioned (here)[https://github.com/modularml/mojo/pull/2250#discussion_r1580488380] it would be very easy to synthesize the method as well. The only concern I have right now is, what happens with references and do we want to provide a solution for circular references. We could say that structs with references do a shallow hash, where they only hash the reference address. Or we say we allow deep hash, where we would need to be able to check if the referenced value was already visited by the Hasher. This would be possible if we add `fn visited(inout self, id: ???) -> Bool` to the Hasher. As you can see I am not certain what the type of the reference `id` should be, maybe `Scalar[DType.address]`.

> In the future (keep things simple for now as you have it!), we may want to consider fancy things like write_length_prefix from [Rust](https://doc.rust-lang.org/stable/std/hash/trait.Hasher.html#method.write_length_prefix). Have you given any thought to this sort of thing?

To be honest I did not consider it till now. After examining the method docs I think it's a bit redundant. The collection implementer can add the length of the collection as an Int to the hasher, in order to solve the issue of `[1, 2, 3]` and `[[1], [2, 3]]` generating same hash value. I think a special method does not help much.

mzaks:
I had a call with @gabrieldemarmiesse yesterday where we discussed the proposal in depth. As a result I commit some improvements to the proposal today.

gabrieldemarmiesse:
Looks good! We can improve on the API and Python compatibility when the compiler has more flexibility :)

gabrieldemarmiesse (approved):

JoeLoser:
Thanks for updating this — I'll take a look at the updates early this week.  Very exciting to see this moving along!! 🚀

JoeLoser:
Landed in today's nightly: https://github.com/modularml/mojo/pull/2615. Thanks for the well thought out proposal! Happy to see this moving along.

PR #2262, opened 2024-04-11
[stdlib] Add trait `Boolable` to `Dict` and `List`

gabrieldemarmiesse:
This allows a very common pattern that's available in python:

```mojo
error_messages = List[String]()
for ... in ...:
    if ...:
        error_messages.append("some new error")

if error_messages:
    print("There were errors!!!")
    ....
    sys.exit(1)

```


JoeLoser (approved):
Looks good, thank you!

PR #2264, opened 2024-04-11
[stdlib] Fix `Int.MIN` error

helehex:
Fixes the compilation error you get when using `Int.MIN`

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -105,6 +105,11 @@ fn _unchecked_zero[type: DType, size: Int]() -> SIMD[type, size]:
     }
 
 
+# ===------------------------------------------------------------------------===#
+# SIMD
+# ===------------------------------------------------------------------------===#
+
+
 @lldb_formatter_wrapping_type
 @register_passable("trivial")
 struct SIMD[type: DType, size: Int = simdwidthof[type]()](
@@ -2465,11 +2470,6 @@ fn _neginf[type: DType]() -> Scalar[type]:
 # ===----------------------------------------------------------------------===#
 
 
-@always_inline("nodebug")
-fn _is_32_bit_system() -> Bool:
-    return sizeof[DType.index]() == sizeof[DType.int32]()
-
-
 @always_inline
 fn _max_finite[type: DType]() -> Scalar[type]:
     """Returns the maximum finite value of type.
@@ -2491,13 +2491,13 @@ fn _max_finite[type: DType]() -> Scalar[type]:
         return 32767
     elif type == DType.uint16:
         return 65535
-    elif type == DType.int32 or (type == DType.index and _is_32_bit_system()):
+    elif type == DType.int32 or type.is_index32():
         return 2147483647
     elif type == DType.uint32:
         return 4294967295
-    elif type == DType.int64 or (
-        type == DType.index and not _is_32_bit_system()
-    ):
+    elif type == DType.float32:
+        return 3.40282346638528859812e38
+    elif type == DType.int64 or type.is_index64():
         return 9223372036854775807
     elif type == DType.uint64:
         return 18446744073709551615
@@ -2538,11 +2538,9 @@ fn _min_finite[type: DType]() -> Scalar[type]:
         return -128
     elif type == DType.int16:
         return -32768
-    elif type == DType.int32 or (type == DType.index and _is_32_bit_system()):
+    elif type == DType.int32 or type.is_index32():
         return -2147483648
-    elif type == DType.int64 or (
-        type == DType.index and not _is_32_bit_system()
-    ):
+    elif type == DType.int64 or type.is_index64():
         return -9223372036854775808
     elif type.is_floating_point():
         return -_max_finite[type]()
--- a/stdlib/src/sys/info.mojo
+++ b/stdlib/src/sys/info.mojo
@@ -447,6 +447,32 @@ fn is_big_endian[
 
 
 @always_inline("nodebug")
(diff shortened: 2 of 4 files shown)


JoeLoser (commented):
Can you share more about your system where this fails to compile? Is it a 32-bit system?

helehex:
Host Information
  ================

  Target Triple: x86_64-unknown-linux
  CPU: skylake
  CPU Features: adx, aes, avx, avx2, bmi, bmi2, clflushopt, cmov, crc32, cx16, cx8, f16c, fma, fsgsbase, fxsr, invpcid, lzcnt, mmx, movbe, pclmul, popcnt, prfchw, rdrnd, rdseed, sahf, sgx, sse, sse2, sse3, sse4.1, sse4.2, ssse3, x87, xsave, xsavec, xsaveopt, xsaves

mojo 2024.4.818 (34399306)

modular 0.7.0 (487c2f88)

helehex:
i was assuming this would happen regardless of platform because of the logic, maybe i was mistaken?

abduld (commented):
can you add a test that Scalar[DType.index].MIN has a value. Also, would be great to consolidate these into an _is_index_32bit helper function to avoid duplication

helehex:
sure no problem

helehex:
how's that? it probably would've been fine to just `return sizeof[DType.index]() == sizeof[DType.int32]()`,
but figured i'd just be safe and explicit

helehex:
oh there's one whitespace it doesn't like hang on

JoeLoser (approved):
Looks good, thank you for the contribution! 🎉 

Do you mind rebasing to resolve the conflict?  Happy to merge after that is taken care of.

stdlib/src/builtin/simd.mojo
```diff
@@ -105,6 +105,29 @@ fn _unchecked_zero[type: DType, size: Int]() -> SIMD[type, size]:
     }
 
 
+@always_inline
+fn _is_index_32bit() -> Bool:
+    @parameter
+    if sizeof[DType.index]() == sizeof[DType.int32]():
+        return True
+    else:
+        return False
```
soraros, line 114: I don't think the param if is necessary, as the function itself is only called in a param if so it's guaranteed to be compile time.

helehex: it should probably go somewhere else aswell, like `dtype.mojo` or `info.mojo`

helehex:
seems like someone already solved it and thats what the conflict is

helehex:
maybe i'll move that helper function to `info.mojo`, then add a corresponding one in `dtype.mojo` that does both checks. does that sound good?

helehex:
oh i just noticed the import for sys sizeof in `dtype.mojo` isn't being used

stdlib/src/sys/info.mojo
helehex, line 1: how about this?

stdlib/src/sys/info.mojo
```diff
@@ -446,6 +446,32 @@ fn is_big_endian[
     ]
 
 
+@always_inline("nodebug")
```
helehex, line 449: how about this?

stdlib/src/sys/info.mojo
```diff
@@ -446,6 +446,32 @@ fn is_big_endian[
...
+
+    Parameters:
+        target: The target architecture.
+
+    Returns:
+        True if the maximum integral value is 64 bit, False otherwise.
+    """
+    return sizeof[DType.index, target]() == sizeof[DType.int64, target]()
+
+
```
helehex, line 474: how about this?

stdlib/src/sys/info.mojo
```diff
@@ -446,6 +446,32 @@ fn is_big_endian[
...
+fn is_32bit[target: __mlir_type.`!kgen.target` = _current_target()]() -> Bool:
+    """Returns True if the maximum integral value is 32 bit.
+
+    Parameters:
+        target: The target architecture.
+
+    Returns:
+        True if the maximum integral value is 32 bit, False otherwise.
+    """
+    return sizeof[DType.index, target]() == sizeof[DType.int32, target]()
```
helehex, line 459: what do you think about this?

JoeLoser: LGTM.  Seems like a much more natural place to put this function. 👍 

Thanks!

stdlib/src/builtin/simd.mojo
```diff
@@ -2538,11 +2538,11 @@ fn _min_finite[type: DType]() -> Scalar[type]:
         return -128
     elif type == DType.int16:
         return -32768
-    elif type == DType.int32 or (type == DType.index and _is_32_bit_system()):
+    elif type == DType.int32 or type.is_index32():
         return -2147483648
-    elif type == DType.int64 or (
-        type == DType.index and not _is_32_bit_system()
-    ):
+    elif type == DType.float32:
```
JoeLoser, line 2543: **Suggestion** I think this rebase part is not correct.  We want to just fall through to the bottom part of `elif type.is_floating_point()` which we changed internally recently.

JoeLoser (approved):
Looks good still, but please fixup the one part about the rebase went slightly awry (correct, but not quite what we want).

stdlib/src/builtin/simd.mojo
helehex: oh i see

helehex:
oh i didn't do a `rebase upstream/nightly` that time, should that matter? there's no conflicts

JoeLoser:
> oh i didn't do a `rebase upstream/nightly` that time, should that matter? there's no conflicts

Nah, force pushing is totally fine.  No biggie.  I'll merge this now! 😄

Issue #1626, opened 2024-01-16
[Feature Request] Enable SIMD vector of indices in `DTypePointer.simd_load`


leandrolcampos:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Add an overload to the `DTypePointer.simd_load` method that takes a SIMD vector containing `width` indices as an argument.

This overload should have a semantically equivalent implementation to the following sequence of scalar loads:

```python
for i in range(width):
    result[i] = self[int(indices[i])]
```

Whenever possible, this overload should use SIMD intrinsics, such as the AVX2 intrinsic function `_mm256_i32gather_ps`.

### What is your motivation for this change?

This change aims to enable the efficient implementation of mathematical functions that utilize table lookup algorithms (table-based numerical methods).

leandrolcampos:
Hi, @abduld.

Here are the [`gather` and `scatter` methods](https://github.com/leandrolcampos/mojo/blob/17a6eadf0c31fb4c90f5c29abae84ccb9704b9e7/stdlib/src/memory/unsafe.mojo#L1490-L1673) that I would add to `DTypePointer` along with their [respective unit tests](https://github.com/leandrolcampos/mojo/blob/17a6eadf0c31fb4c90f5c29abae84ccb9704b9e7/stdlib/test/memory/test_memory.mojo#L302-L417).

I would also update the [docstring of the `scatter` function](https://github.com/leandrolcampos/mojo/blob/17a6eadf0c31fb4c90f5c29abae84ccb9704b9e7/stdlib/src/sys/intrinsics.mojo#L677-L680) in `sys.intrinsics` to align it with the [documentation of the `llvm.masked.scatter.*` intrinsics](https://llvm.org/docs/LangRef.html#llvm-masked-scatter-intrinsics) regarding overlapping addresses.

All the best,

JoeLoser:
Fixed with https://github.com/modularml/mojo/commit/1a14a1e6f8ad9273db2c2bf67fcaef3011011195

PR #2268, opened 2024-04-12
[stdlib] Add the `gather` and `scatter` methods to `DTypePointer`

leandrolcampos:
This PR introduces the `gather` and `scatter` methods to the `DTypePointer` struct, as outlined in feature request #1626.

**Key Changes:**
- Implemented the `gather` method for `DTypePointer`.
- Implemented the `scatter` method for `DTypePointer`.
- Added corresponding unit tests to validate the functionality of these methods.
- Updated the docstring in `sys.intrinsics.scatter` to align with the LLVM masked scatter intrinsic documentation [[link](https://llvm.org/docs/LangRef.html#llvm-masked-scatter-intrinsics)]. These changes better reflect the behavior of the scatter intrinsic with overlapping addresses.

**Naming Consideration:**
- While staying consistent with the corresponding LLVM intrinsics is important, it’s also crucial to focus on the primary audience of Mojo. If Mojo developers are primarily expected to have a comfort with Python, using `default` instead of `passthru` might reduce the learning curve and avoid confusion. For developers familiar with Python, `default` conveys the idea that an argument serves as a fallback or standard value when some condition fails. This naming choice aims to enhance clarity and accessibility, making the API more intuitive for our primary user base.

Aware that this feature had been assigned to a maintainer since January, I reached out through comments on the original feature request to ask about any progress and to express my willingness to contribute. In the absence of updates or responses, I proceeded with this implementation. _I am fully prepared to adjust or withdraw my contribution should there be existing work that overlaps with what is presented here_.

Thank you for considering this contribution.

Best regards,

### DIFF
--- a/stdlib/src/memory/unsafe.mojo
+++ b/stdlib/src/memory/unsafe.mojo
@@ -1641,8 +1641,8 @@ struct DTypePointer[
             "alignment must be 0 or a power of two integer value",
         ]()
 
-        var pointer = offset.cast[DType.index]().fma(sizeof[type](), int(self))
-        return gather(pointer.cast[DType.address](), mask, default, alignment)
+        var base = offset.cast[DType.index]().fma(sizeof[type](), int(self))
+        return gather(base.cast[DType.address](), mask, default, alignment)
 
     @always_inline("nodebug")
     fn scatter[
@@ -1730,8 +1730,8 @@ struct DTypePointer[
             "alignment must be 0 or a power of two integer value",
         ]()
 
-        var pointer = offset.cast[DType.index]().fma(sizeof[type](), int(self))
-        scatter(val, pointer.cast[DType.address](), mask, alignment)
+        var base = offset.cast[DType.index]().fma(sizeof[type](), int(self))
+        scatter(val, base.cast[DType.address](), mask, alignment)
 
     @always_inline("nodebug")
     fn __int__(self) -> Int:
--- a/stdlib/src/memory/unsafe.mojo
+++ b/stdlib/src/memory/unsafe.mojo
@@ -1655,7 +1655,7 @@ struct DTypePointer[
         """Scatters values to memory based on offsets from the current pointer.
 
         This method performs a store operation, scattering each element from
-        the `val` SIMD vector at the memory address calculated by adding the
+        the `val` SIMD vector to the memory address calculated by adding the
         corresponding element from `offset` SIMD vector to the current pointer.
 
         If the same offset is targeted multiple times, the values are stored in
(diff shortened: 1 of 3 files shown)


stdlib/src/memory/unsafe.mojo
```diff
@@ -1550,6 +1550,188 @@ struct DTypePointer[
...
+                element whether to load from memory or to take from the
+                `default` SIMD vector.
+            default: The SIMD vector providing default values to be taken
+                where the `mask` SIMD vector is `False`.
+
+        Returns:
+            The SIMD vector containing the gathered values.
+        """
+        constrained[
+            offset_type.is_integral(),
```
abduld, line 1636:
```suggestion
            offset.type.is_integral(),
```

stdlib/src/memory/unsafe.mojo
```diff
@@ -1550,6 +1550,188 @@ struct DTypePointer[
...
+            "offset type must be an integral type",
+        ]()
+        constrained[
+            (alignment == 0) | _is_power_of_2(alignment),
```
abduld, line 1640: why do you allow 0 alignment?

stdlib/src/memory/unsafe.mojo
```diff
@@ -1550,6 +1550,188 @@ struct DTypePointer[
...
+    @always_inline("nodebug")
+    fn gather[
+        offset_type: DType,
+        /,
+        *,
+        width: Int = 1,
+        alignment: Int = Self._default_alignment,
+    ](
+        self,
+        offset: SIMD[offset_type, width],
```
abduld, line 1600:
```suggestion
        offset: SIMD[_, width],
```

abduld (approved):
nice

stdlib/src/memory/unsafe.mojo
leandrolcampos: Initially, I aligned the implementation of the `gather` and `scatter` methods with the constraints applied to the `alignment` argument in the corresponding LLVM intrinsic functions, which permit a zero value. However, considering the specific context of `DTypePointer` objects, which are always aligned to a non-zero value, I recognize that the flexibility provided by accepting an alignment of 0 may indeed be unnecessary. I have revised the PR accordingly to reflect this more restrictive approach. Do you agree with this change?

stdlib/src/memory/unsafe.mojo
leandrolcampos: Done (here and elsewhere).

stdlib/src/memory/unsafe.mojo
leandrolcampos: Done (here and elsewhere).

soraros:
Can we remove the overload and give `default`/`mask` default values?

leandrolcampos:
> Can we remove the overload and give `default`/`mask` default values?

In principle, I'd prefer not to remove the overloads, because they give me the opportunity to hide complexity when it is not needed.

Given that `gather` and `scatter` operations are naturally complex, I think overloads with simpler docstrings and APIs are welcome.

leandrolcampos:
@JoeLoser: please let me know if anything else is needed from my side. Thank you,

JoeLoser:
> @JoeLoser: please let me know if anything else is needed from my side. Thank you,

Just merged it!  Thank you so much for the contribution! 🎉

PR #2272, opened 2024-04-12
[stdlib] Do not copy elements when using `FileHandle.read_bytes()`

### LINKED ISSUE #2051: [stdlib] Do not copy elements when using `FileHandle.read_bytes()`

MoSafi2:
Fixes #2051, allows `FileHandle.read_bytes()` to return `List[Int8]` without additional copies, depends on #2182.

### DIFF
--- a/stdlib/src/builtin/file.mojo
+++ b/stdlib/src/builtin/file.mojo
@@ -33,8 +33,7 @@ with open("my_file.txt", "r") as f:
 
 from os import PathLike
 from sys import external_call
-
-from memory import AddressSpace, DTypePointer, Pointer
+from memory.unsafe import AddressSpace, DTypePointer, Pointer
 
 
 @register_passable
@@ -176,7 +175,7 @@ struct FileHandle:
         var err_msg = _OwnedStringRef()
 
         var buf = external_call[
-            "KGEN_CompilerRT_IO_FileReadBytes", Pointer[Int8]
+            "KGEN_CompilerRT_IO_FileReadBytes", AnyPointer[Int8]
         ](
             self.handle,
             Pointer.address_of(size_copy),
@@ -186,13 +185,8 @@ struct FileHandle:
         if err_msg:
             raise (err_msg^).consume_as_error()
 
-        var list = List[Int8](capacity=int(size_copy))
-
-        var list_ptr = Pointer[Int8].__from_index(int(list.data))
-
-        # Initialize the List elements and set the initialized size
-        memcpy(list_ptr, buf, int(size_copy))
-        list.size = int(size_copy)
+        # No-copy list initialization
+        var list = List[Int8](buf, int(size_copy), int(size_copy))
 
         return list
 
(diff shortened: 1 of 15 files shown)


stdlib/test/builtin/test_file.mojo
```diff
@@ -68,16 +68,19 @@ def test_file_read_bytes_multi():
 
     # CHECK: Lorem ipsum
     var bytes1 = f.read_bytes(12)
+    assert_equal(bytes1, "Lorem ipsum")
     print(String(bytes1))
 
     # CHECK: dolor
```
JoeLoser, line 74: **Suggestion** We can drop this `CHECK` and the `print` below your new assert.  Ditto below for the `print(len(s))` and its `CHECK` line.

JoeLoser (approved):
Looks good, thank you! Please see my comment about cleaning up the test now that we're using `assert_equal` 👍

jackos:
Nice one @MoSafi2! Ran through a leak profiler and this indeed gets rid of the memory leak too.

stdlib/test/builtin/test_file.mojo
MoSafi2: The unnecessary prints and comments are now dropped.

gabrieldemarmiesse:
I seems you need to run `mojo format` on your files to make the CI happy @MoSafi2

MoSafi2:
@gabrieldemarmiesse let's see if this commit is enough to get through the checks.

stdlib/src/builtin/file.mojo
```diff
@@ -33,8 +33,7 @@ with open("my_file.txt", "r") as f:
 
 from os import PathLike
 from sys import external_call
-
-from memory import AddressSpace, DTypePointer, Pointer
+from memory.unsafe import AddressSpace, DTypePointer, Pointer
```
JoeLoser, line 36: **Suggestion** Do you mind not making this change?  We'd prefer `from memory <blah>` rather than `from memory.unsafe <blah>`.  See https://github.com/modularml/mojo/commit/63d5cb0e5fd1b90fb4da5c8e4f5974fbc04c8633

PR #2273, opened 2024-04-12
[stdlib] String to Int with specified base

artemiogr97:


artemiogr97:
implementation for #2274

JoeLoser (commented):
Thanks for the contribution! I'll take a look at this in the morning. Sorry for the slower than usual response this week.

stdlib/src/builtin/int.mojo
```diff
@@ -18,7 +18,7 @@ These are Mojo built-ins, so you don't need to import them.
 from collections import KeyElement
 
 from builtin.hash import _hash_simd
-from builtin.string import _calc_initial_buffer_size
+from builtin.string import _calc_initial_buffer_size, atol
```
JoeLoser, line 21: **Suggestion** From the `builtin` module, you only need to explicitly export _private_ entities (those prefixed with an `_`).  In this case, `atol` is public, so there's no need to explicitly import it.  It (and any public entity in the `builtin` module) will always be imported into every Mojo program.

stdlib/src/builtin/string.mojo
```diff
@@ -149,23 +149,28 @@ fn chr(c: Int) -> String:
...
+        str_ref: A string to be parsed as an integer in the given base.
+        base: Base used for conversion, value must be between 2 and 36.
 
     Returns:
         An integer value that represents the string, or otherwise raises.
     """
     if not str_ref:
         raise Error("Empty String cannot be converted to integer.")
+    if base < 2 or base > 36:
+        raise Error("Base must be between 2 and 36 included.")
```
JoeLoser, line 170: **Quibble** Perhaps `s/included/inclusive`?

stdlib/test/builtin/test_string.mojo
```diff
@@ -222,35 +223,40 @@ fn test_atol() raises:
...
-    try:
+    with assert_raises(contains="String is not convertible to integer."):
         _ = atol(String(" 10 1"))
-        raise Error("Failed to raise when converting string to integer.")
-    except e:
-        assert_equal(str(e), "String is not convertible to integer.")
 
-    try:
+    with assert_raises(
+        contains="String is not convertible to integer with base 5."
```
JoeLoser, line 243: **Suggestion** I think it would be good to include the original user's requested string in this error message — what do you think?  For example, the error from Python is:

``` python
int("5", 5)
ValueError: invalid literal for int() with base 5: '5'
```

If we do this, I think we shouldn't special case the empty string error message either.  It's consistent in Python, for example:

``` python
int("")
ValueError: invalid literal for int() with base 10: ''
```

stdlib/src/builtin/string.mojo
```diff
@@ -149,23 +149,28 @@ fn chr(c: Int) -> String:
...
-        str_ref: A string to be parsed as a base-10 integer.
+        str_ref: A string to be parsed as an integer in the given base.
+        base: Base used for conversion, value must be between 2 and 36.
 
     Returns:
         An integer value that represents the string, or otherwise raises.
     """
     if not str_ref:
         raise Error("Empty String cannot be converted to integer.")
+    if base < 2 or base > 36:
```
JoeLoser, line 169: **Question** Any specific reason to exclude the special case of base `0`?  In Python:

```python
int("0", 1)
ValueError: int() base must be >= 2 and <= 36, or 0
```

which suggests we should handle base 0 too pedantically.

JoeLoser (commented):
Thanks for your contribution! 🎉 

Looks pretty good — just had a few comments on improving the error message even more and handling the special case of base 0.

stdlib/src/builtin/string.mojo
artemiogr97: No reason, I didn't realize that base 0 was an option, I guess I'll handle this case too

stdlib/test/builtin/test_string.mojo
artemiogr97: Sure seems like a good idea

artemiogr97:
I added the case for base=0 starting with "+" and using "_", but I had some problems running the test, it was disabled for some reason, also after activating it I had to comment the following lines:
```
    alias multi_byte = ord("α")
    assert_equal(multi_byte, 945)
    alias multi_byte2 = ord("➿")
    assert_equal(multi_byte2, 10175)
    alias multi_byte3 = ord("🔥")
    assert_equal(multi_byte3, 128293)```
I reenabled the test on commit 074962d, seems like the CI got the same error as me, so I commited a second time commenting the lines.
I leave you decide what to do, should we leave the test disabled? or enabled with the lines in comments?

**Comment by ConnorGray** [comment]
Hi @artemiogr97, thanks again for this contribution. Awesome work handling the base=0 case, its a very useful feature to have and great to be consistent with Python.

> ```
>    alias multi_byte = ord("α")
>    assert_equal(multi_byte, 945)
>    alias multi_byte2 = ord("➿")
>    assert_equal(multi_byte2, 10175)
>    alias multi_byte3 = ord("🔥")
>    assert_equal(multi_byte3, 128293)
> ```

I'm uncertain which error you originally saw, but there is a known error with evaluating `ord("...")` at compile time (like in an `alias`.

Does the error you mentioned go away if you switch these to `var`, e.g.:

```
    var multi_byte = ord("α")
    assert_equal(multi_byte, 945)
    var multi_byte2 = ord("➿")
    assert_equal(multi_byte2, 10175)
    var multi_byte3 = ord("🔥")
    assert_equal(multi_byte3, 128293)
```

Re. the `REQUIRES: Disabled` line, its okay to remove that in this PR. 

Thanks a bunch for this! :fire:

artemiogr97:
Hi @ConnorGray, using var instead of alias `fixed` the problem, so I replaced it in my last commit,  I also found 1 missing edge case in atol and fixed it

ConnorGray:
> Hi @ConnorGray, using var instead of alias "fixed" the problem

Thanks, glad the workaround unblocked this issue. That this doesn't currently work in `alias` is bug in the compiler that we'll aim to address in the future.

> I also found 1 missing edge case in atol and fixed it

Nice catch, thanks for fixing that!

This PR is looking good, however I'm going to hold off on merging it temporarily as we're about to perform some internal infrastructure changes that would be affected if we made changes to the Mojo repo. We should have those finished tomorrow sometime and I'll be able to merge this PR then.

Thanks again so much @artemiogr97, this is a :fire: contribution :)

artemiogr97:
Hi @ConnorGray seems like some stuff was broken after updating to the new nightly complier, I just fixed it in my last commit

ConnorGray:
**✅🟣 This contribution has been merged 🟣✅**

Hi @artemiogr97 thank you so much for fixing those issues!

Using the new release infrastructure I mentioned in my previous comment (we're using a tool called [Copybara](https://github.com/google/copybara)), your contribution change has now been merged into our internal copy of the Mojo Standard Library.

The changes in this PR will appear here in the mojo repo `nightly` branch when we do our next outbound synchronization at the time that the next Mojo nightly release is released.

Thank you so much again for making this :fire: contribution, and let me say thank you personally for bearing with me through the transition to this new infrastructure! Please let me know if you have any questions or concerns.

PR #2275, opened 2024-04-12
[stdlib] Remove unnecessary declaration

LiamSwayne:
`i` was an unnecessary and unclear declaration that could be substituted for `end`. This is not a behavioral change.

Reasoning:
- `i` was set to be equivalent to `end`, but was never used in a way in which it could not simply be substituted for `end`.
- `i` is not descriptive, and `end` does a better job describing it's purpose. `i` didn't have a comment either, also making it less descriptive than `end`.
- `var i` was used almost immediately after a different `i` was used as a counter in a for loop, and it could be confusing to use two variables with different purposes and the same name immediately after one another.
- Removing this declaration could result in a small performance improvement because an assignment is removed, but at the very least the code is simpler without this declaration.


JoeLoser (approved):
Looks good, thanks for the contribution! 🎉

PR #2279, opened 2024-04-13
[stdlib] Fix assert_equal[SIMD]() not raising if any elements are equal

gabrieldemarmiesse:
To reproduce the bug:
```mojo
with assert_raises():
    assert_equal(SIMD[DType.uint8, 2](1, 1), SIMD[DType.uint8, 2](1, 2))
```

This is because using `!=` returns `[False, True]` then, when using `.__bool__()` on this, since `__bool__()` is equivalent to a `reduce_and()`, it returns False, and the error is not triggered.

Long story short, doing `if a != b` doesn't do what one would expect. This bug is not present with `assert_not_equal()`

### DIFF
--- a/stdlib/src/testing/testing.mojo
+++ b/stdlib/src/testing/testing.mojo
@@ -151,7 +151,9 @@ fn assert_equal[
     Raises:
         An Error with the provided message if assert fails and `None` otherwise.
     """
-    if lhs != rhs:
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
Sigh, the `SIMD.eq` and `SIMD.ne` methods strike again by not returning just a `Bool`.  Thanks for noticing and fixing this before we start writing more checks using the `testing` module and `SIMD` types!

PR #2282, opened 2024-04-13
[stdlib] Remove FileCheck for `test_simd.mojo`

gabrieldemarmiesse:
Related to https://github.com/modularml/mojo/issues/2024

Note that https://github.com/modularml/mojo/pull/2279 should be merged first, then we can rebase this PR, it's going to be cleaner.


stdlib/test/builtin/test_simd.mojo
```diff
@@ -42,18 +39,11 @@ def test_cast():
...
 
 
-# CHECK-LABEL: test_simd_variadic
-fn test_simd_variadic():
-    print("== test_simd_variadic")
-
-    # CHECK: [52, 12, 43, 5]
-    print(SIMD[DType.index, 4](52, 12, 43, 5))
+def test_simd_variadic():
+    assert_equal(str(SIMD[DType.index, 4](52, 12, 43, 5)), "[52, 12, 43, 5]")
```
JoeLoser, line 43: **Suggestion** How about we actually use `SIMD.eq` rather than `str()` on both sides essentially?  `str` was only there for `FileCheck` since it's just a fancy regex matcher in short.  We can actually call real methods and not just do string matching shenanigans now that we're using the `testing` module.

JoeLoser (approved):
Looks good, thank you! Just one minor comment in a test.  Do you want to rebase now that https://github.com/modularml/mojo/commit/ee5dfe791effa9def469001e9cd5933996b96c6f landed, or want me to do "rebase and merge" after the test fix?

stdlib/test/builtin/test_simd.mojo
gabrieldemarmiesse: For this case in particular, I don't think switching would make sense. We would end up with 
```
assert_equal(SIMD[DType.index, 4](52, 12, 43, 5), SIMD[DType.index, 4](52, 12, 43, 5))
```
which seems kind of pointless as a test. To verify that the numbers are correctly stored in the SIMD, I think converting to string is a valid thing to do

gabrieldemarmiesse:
I merged nightly into this branch. I believe you can "squash and merge" @JoeLoser

stdlib/test/builtin/test_simd.mojo
JoeLoser: Oh, I see now - thanks for pointing that out.  The whole test seems pretty low value, honestly, but we can keep it for now; not a biggie.  We're in the process of syncing repos right now to push out a new `nightly/mojo`, so I'm holding off from merging this, but I'll merge it later today.

gabrieldemarmiesse: Sure thing!

JoeLoser:
@gabrieldemarmiesse I think you'll need to rebase again from more conflicts when we synced the repos/made a new `nightly/mojo` just a bit ago.  Once that's done, I'll merge this PR.  Thanks!

gabrieldemarmiesse:
Haha, we'll merge this PR eventually XD

JoeLoser:
> Haha, we'll merge this PR eventually XD

Merged!

PR #2283, opened 2024-04-13
[stdlib] Remove FileCheck from `test_file.mojo`

gabrieldemarmiesse:
Related to https://github.com/modularml/mojo/issues/2024

I had some issues concerning the conversion `List[Int8]` -> `String`, the null terminator was missing, so I added a check concerning this.


JoeLoser (approved):
So exciting to see our use of `FileCheck` go away 🎉 !  Thanks again for the contribution 🔥

PR #2289, opened 2024-04-13
[stdlib] Remove FileCheck for `test_bfloat16.mojo`

gabrieldemarmiesse:
Related to #2024


stdlib/test/builtin/test_bfloat16.mojo
```diff
@@ -35,46 +34,58 @@ def test_methods():
     assert_almost_equal(BFloat16(2.0), 2.0)
 
 
-fn test_bf_primitives():
+def test_bf_primitives():
```
Moosems, line 37: Why the change to using `def` instead of `fn`?

stdlib/test/builtin/test_bfloat16.mojo
```diff
@@ -35,46 +34,58 @@ def test_methods():
...
+    assert_equal(a == b, a_hp == b_hp)
+    assert_equal(a != b, a_hp != b_hp)
+    assert_equal(a <= b, a_hp <= b_hp)
+    assert_equal(a >= b, a_hp >= b_hp)
 
-def main():
-    # CHECK: 33.0
-    print(
+
+def check_float64_values():
```
Moosems, line 56: Done here as well

Moosems (commented):
Quick question

stdlib/test/builtin/test_bfloat16.mojo
gabrieldemarmiesse: Using `assert_*` function means that the function can now raise. So we have two choices, either we use `raising` to the function, either we use `def` instead of `fn`. I chose def, but I don't have a strong opinion about this

stdlib/test/builtin/test_bfloat16.mojo
```diff
@@ -10,11 +10,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
-# RUN: %mojo %s | FileCheck %s
+# RUN: %mojo -debug-level full %s
```
JoeLoser, line 13: Similar comment about dropping the `-debug-level full` as in other PRs.

stdlib/test/builtin/test_bfloat16.mojo
```diff
@@ -35,46 +34,58 @@ def test_methods():
...
-    print(a * b)
-    print(a == b)
-    print(a != b)
-    print(a <= b)
-    print(a >= b)
+    # higher precision
+    var a_hp = a.cast[DType.float64]()
+    var b_hp = b.cast[DType.float64]()
 
+    assert_almost_equal(a + b, (a_hp + b_hp).cast[DType.bfloat16]())
```
JoeLoser, line 46: Thanks for adding these!  Surprised the original code just had naked `print`s without any `CHECK` directives 🤷

stdlib/test/builtin/test_bfloat16.mojo
```diff
@@ -35,46 +34,58 @@ def test_methods():
...
+                        _type = __mlir_type[`!pop.scalar<bf16>`],
+                        value = __mlir_attr[
+                            `#pop.simd<"nan"> : !pop.scalar<bf16>`
+                        ],
+                    ]()
+                )
             )
-        )
+        ),
+        "nan",
```
JoeLoser, line 82: 👍 good to compare the string `nan` here since can't use the `math` module in the open source part of the standard library right now.

JoeLoser (approved):
Looks good, thank you so much for pushing on all of this `FileCheck` removal!  Happy to merge after we remove the `-debug-level full` part in the `RUN` directive.

stdlib/test/builtin/test_bfloat16.mojo
gabrieldemarmiesse, line 13:
```suggestion
# RUN: %mojo-no-debug %s
```

PR #2294, opened 2024-04-13
[mojo-stdlib] Create `InlineArray` type

lsh:
This PR creates an `InlineArray` type that takes any `CollectionElement` rather than just `AnyRegType`. See also [this thread](https://discord.com/channels/1087530497313357884/1228515158234501231) on Discord.

### DIFF
--- a/stdlib/src/utils/static_tuple.mojo
+++ b/stdlib/src/utils/static_tuple.mojo
@@ -328,7 +328,11 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
 
     @always_inline
     fn __init__(inout self, *, fill: Self.ElementType):
-        """Constructs an empty (undefined) array."""
+        """Constructs an empty array where each element is the supplied `fill`.
+
+        Args:
+            fill: The element to fill each index.
+        """
         _static_tuple_construction_checks[size]()
         self._array = __mlir_op.`kgen.undef`[_type = Self.type]()
 
--- a/stdlib/src/utils/static_tuple.mojo
+++ b/stdlib/src/utils/static_tuple.mojo
@@ -97,6 +97,70 @@ fn _create_array[
         return array
 
 
+@always_inline
+fn _set_array_elem_mem[
+    index: Int,
+    size: Int,
+    type: CollectionElement,
+](
+    val: type,
+    array: Reference[
+        __mlir_type[`!pop.array<`, size.value, `, `, type, `>`],
+        __mlir_attr.`1 : i1`,
+        _,
+    ],
+):
+    """Sets the array element at position `index` with the value `val`.
+
+    Parameters:
+        index: the position to replace the value at.
+        size: the size of the array.
+        type: the element type of the array
+
+    Args:
+        val: the value to set.
+        array: the array which is captured by reference.
+    """
+    var ptr = __mlir_op.`pop.array.gep`(
+        array.get_legacy_pointer().address, index.value
+    )
+    var p = UnsafePointer(ptr)
+    initialize_pointee(p, val)
+
+
+@always_inline
+fn _create_array_mem[
+    size: Int, type: CollectionElement
+](lst: VariadicList[type]) -> __mlir_type[
+    `!pop.array<`, size.value, `, `, type, `>`
+]:
+    """Sets the array element at position `index` with the value `val`.
+
+    Parameters:
+        size: the size of the array.
+        type: the element type of the array
+
+    Args:
+        lst: the list of values to set.
+
+    Returns:
+        The array with values filled from the input list.
+    """
+    debug_assert(size == len(lst), "mismatch in the number of elements")
+
+    var array = __mlir_op.`kgen.undef`[
+        _type = __mlir_type[`!pop.array<`, size.value, `, `, type, `>`]
+    ]()
+
+    @always_inline
+    @parameter
+    fn fill[idx: Int]():
+        _set_array_elem_mem[idx, size, type](lst[idx], array)
+
+    unroll[fill, size]()
+    return array
+
+
 # ===----------------------------------------------------------------------===#
 # StaticTuple
 # ===----------------------------------------------------------------------===#
@@ -240,3 +304,89 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
         )
         Pointer(ptr).store(val)
         self = tmp
+
+
+@value
+struct Array[ElementType: CollectionElement, size: Int](Sized):
+    """A statically sized array type which contains elements of homogeneous types.
+
+    Parameters:
+        ElementType: The type of the elements in the array.
+        size: The size of the array.
+    """
+
+    alias type = __mlir_type[
+        `!pop.array<`, size.value, `, `, Self.ElementType, `>`
+    ]
+    var array: Self.type
+    """The underlying storage for the array."""
+
+    @always_inline
+    fn __init__(inout self):
+        """Constructs an empty (undefined) array."""
+        _static_tuple_construction_checks[size]()
+        self.array = __mlir_op.`kgen.undef`[_type = Self.type]()
+
+    @always_inline
+    fn __init__(inout self, *elems: Self.ElementType):
+        """Constructs an array given a set of arguments.
+
+        Args:
+            elems: The element types.
+        """
+        self = Self()
+        var i = 0
+        for elem in elems:
+            var ptr = __mlir_op.`pop.array.gep`(
+                Reference(Reference(self)[].array).get_legacy_pointer().address,
+                i.value,
+            )
+            initialize_pointee(UnsafePointer[Self.ElementType](ptr), elem[])
+            i += 1
+
+    @always_inline
+    fn __init__(inout self, values: VariadicList[Self.ElementType]):
+        """Creates an array constant using the specified values.
+
+        Args:
+            values: The list of values.
+        """
+        _static_tuple_construction_checks[size]()
+        self.array = _create_array_mem[size, Self.ElementType](values)
+
+    @always_inline("nodebug")
+    fn __len__(self) -> Int:
+        """Returns the length of the array. This is a known constant value.
+
+        Returns:
+            The size of the list.
+        """
+        return size
+
+    fn __refitem__[
+        mutability: __mlir_type.i1, self_life: AnyLifetime[mutability].type
+    ](
+        self: Reference[Self, mutability, self_life]._mlir_type, index: Int
+    ) -> Reference[Self.ElementType, mutability, self_life]:
+        var ptr = __mlir_op.`pop.array.gep`(
+            Reference(Reference(self)[].array).get_legacy_pointer().address,
+            index.value,
+        )
+        return Reference[Self.ElementType, mutability, self_life](
+            UnsafePointer(ptr)[]
+        )
+
+    fn __refitem__[
+        mutability: __mlir_type.i1,
+        self_life: AnyLifetime[mutability].type,
+        index: Int,
+    ](self: Reference[Self, mutability, self_life]._mlir_type) -> Reference[
+        Self.ElementType, mutability, self_life
+    ]:
+        var ptr = __mlir_op.`pop.array.gep`(
+            Reference(Reference(self)[].array).get_legacy_pointer().address,
+            index.value,
+        )
+        return Reference[Self.ElementType, mutability, self_life](
+            UnsafePointer(ptr)[]
+        )
--- a/stdlib/src/utils/static_tuple.mojo
+++ b/stdlib/src/utils/static_tuple.mojo
@@ -387,15 +387,19 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
 
     @always_inline("nodebug")
     fn __refitem__[
-        mutability: __mlir_type.i1, self_life: AnyLifetime[mutability].type
+        mutability: __mlir_type.i1,
+        self_life: AnyLifetime[mutability].type,
+        IntableType: Intable,
     ](
-        self: Reference[Self, mutability, self_life]._mlir_type, index: Int
+        self: Reference[Self, mutability, self_life]._mlir_type,
+        index: IntableType,
     ) -> Reference[Self.ElementType, mutability, self_life]:
         """Get a `Reference` to the element at the given index.
 
         Parameters:
             mutability: The inferred mutability of the reference.
             self_life: The inferred lifetime of the reference.
+            IntableType: The inferred type of an intable argument.
 
         Args:
             index: The index of the item.
@@ -403,9 +407,9 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
         Returns:
             A reference to the item at the given index.
         """
-        debug_assert(-size <= index < size, "Index must be within bounds.")
-        var normalized_idx = index
-        if index < 0:
+        debug_assert(-size <= int(index) < size, "Index must be within bounds.")
+        var normalized_idx = int(index)
+        if normalized_idx < 0:
             normalized_idx += size
 
         return Reference(self)[]._get_reference_unsafe[mutability, self_life](
@@ -416,7 +420,8 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
     fn __refitem__[
         mutability: __mlir_type.i1,
         self_life: AnyLifetime[mutability].type,
-        index: Int,
+        IntableType: Intable,
+        index: IntableType,
     ](self: Reference[Self, mutability, self_life]._mlir_type) -> Reference[
         Self.ElementType, mutability, self_life
     ]:
@@ -425,17 +430,19 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
         Parameters:
             mutability: The inferred mutability of the reference.
             self_life: The inferred lifetime of the reference.
+            IntableType: The inferred type of an intable argument.
             index: The index of the item.
 
         Returns:
             A reference to the item at the given index.
         """
-        constrained[-size <= index < size, "Index must be within bounds."]()
+        alias i = int(index)
+        constrained[-size <= i < size, "Index must be within bounds."]()
 
-        var normalized_idx = index
+        var normalized_idx = i
 
         @parameter
-        if index < 0:
+        if i < 0:
             normalized_idx += size
 
         return Reference(self)[]._get_reference_unsafe[mutability, self_life](
--- a/stdlib/src/utils/static_tuple.mojo
+++ b/stdlib/src/utils/static_tuple.mojo
@@ -368,6 +368,7 @@ struct Array[ElementType: CollectionElement, size: Int](Sized):
         """
         return size
 
+    @always_inline("nodebug")
     fn __refitem__[
         mutability: __mlir_type.i1, self_life: AnyLifetime[mutability].type
     ](
@@ -381,6 +382,7 @@ struct Array[ElementType: CollectionElement, size: Int](Sized):
             UnsafePointer(ptr)[]
         )
 
+    @always_inline("nodebug")
     fn __refitem__[
         mutability: __mlir_type.i1,
         self_life: AnyLifetime[mutability].type,
--- a/stdlib/src/utils/static_tuple.mojo
+++ b/stdlib/src/utils/static_tuple.mojo
@@ -306,6 +306,11 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
         self = tmp
 
 
+# ===----------------------------------------------------------------------===#
+# Array
+# ===----------------------------------------------------------------------===#
+
+
 @value
 struct Array[ElementType: CollectionElement, size: Int](Sized):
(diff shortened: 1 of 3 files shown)


stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,122 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+        return AnyPointer(ptr).__refitem__()
+
+    # TODO: make generic over mutability
+    @always_inline
+    fn __refitem__[
+        self_life: MutLifetime, index: Int
+    ](
+        self: Reference[Self, __mlir_attr.`1: i1`, self_life].mlir_ref_type,
+    ) -> Reference[Self.ElementType, __mlir_attr.`1: i1`, self_life]:
+        constrained[-index < index < size, "index must be within bounds"]()
```
mikowals, line 414: Should this be `-size <= index < size`?

lsh: Good catch, fixed

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,142 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+    @always_inline
+    fn __init__(inout self, values: VariadicList[Self.ElementType]):
+        """Creates a tuple constant using the specified values.
+
+        Args:
+            values: The list of values.
+        """
+        _static_tuple_construction_checks[size]()
+        debug_assert(
+            0 < len(values) < size, "size of `values` must be within range"
```
mikowals, line 372: There is also `debug_assert(size == len(lst), ...)` inside `_create_array_collection_elem`. Maybe this should just be a comment to say `len(values)` is checked there?

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
         return Pointer(ptr)
+
+
+# ===----------------------------------------------------------------------===#
+# Array
+# ===----------------------------------------------------------------------===#
+
+
+@value
+struct Array[ElementType: CollectionElement, size: Int](Sized):
```
JoeLoser, line 325: **Suggestion** I'm good with having `Array` be defined here rather than in `collections/array.mojo` for now to make the migration easier for us internally (and we can do the move once we remove the `StaticTuple`).

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+    """A statically sized tuple type which contains elements of homogeneous types.
```
JoeLoser, line 326: **Suggestion** Since we don't have tuple in the name, WDYT about phrasing this something along the lines of 


> A fixed-size sequence of `size` homogenous elements where `size` is a constant expression.

or something like that.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+    Parameters:
+        ElementType: The type of the elements in the tuple.
+        size: The size of the tuple.
+    """
+
+    alias type = __mlir_type[
+        `!pop.array<`, size.value, `, `, Self.ElementType, `>`
+    ]
+    var array: Self.type
+    """The underlying storage for the static tuple."""
```
JoeLoser, line 337: **Suggestion** `s/static tuple/array`.  Ditto elsewhere in this struct definition - essentially let's use the word "array" instead of "tuple" - WDYT?

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+
+    @always_inline
+    fn __init__(inout self):
+        """Constructs an empty (undefined) tuple."""
+        _static_tuple_construction_checks[size]()
+        self.array = __mlir_op.`kgen.undef`[_type = Self.type]()
+
+    @always_inline
+    fn __init__(inout self, *elems: Self.ElementType):
+        """Constructs a static tuple given a set of arguments.
```
JoeLoser, line 347: **Suggestion** Ditto here in the API doc string - "Constructs an array ..." instead of static tuple.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+        """
+        _static_tuple_construction_checks[size]()
+        self = Self()
+        var i = 0
+        for e in elems:
+            self[i] = e[]
+            i += 1
+
+    @always_inline
+    fn __init__(inout self, values: VariadicList[Self.ElementType]):
```
JoeLoser, line 360: **Suggestion** Can we say `*values: Self.ElementType` please?  In general, APIs shouldn't spell out the direct type of `VariadicList` or `VariadicListMem` (this is a sharp edge); instead, they should use the language spelling of `*` to construct the right library type backing the variadic.  In this case, this is actually wrong since you want `VariadicListMem`.  Otherwise, the current code will actually never run destructors for the argument variadic pack.  See how `VariadicListMem` defines `__del__` for example but `VariadicList` can't (since it only works on `AnyRegType`).  This is a QoI thing in the compiler for even *accepting* this code, since you're passing `AnyType` to a thing accepting `AnyRegType`.  This should be a compiler error in a nice world.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+            self_life: The inferred lifetime of the variant type.
+            index: The index of the element.
+
+        Returns:
+            A mutable reference to the element at the given index.
+        """
+        constrained[-size <= index < size, "index must be within bounds"]()
+        var normalized_idx = index
+
+        if index < 0:
```
JoeLoser, line 432: **Suggestion** This could be run at compile time via `@parameter if` since `index` is a parameter (and not a runtime argument) in this context.  So, `normalized` index could be an `alias` too (we may need to define a smaller helper function and assign the result of calling to the `alias normalized_idx = compile_time_helper_function_to_compute_normalized_index()` for example).

After all that is said and done, what do you think about factoring out the "core logic" of the body of this function with the other `__refitem__` overload?  Worth it or nah?

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+        var array_copy = Reference[Self.type, mutability, self_life](
+            Reference(self)[].array
+        )
+        var ptr = __mlir_op.`pop.array.gep`(
+            array_copy.get_unsafe_pointer().address, normalized_idx.value
+        )
+        return AnyPointer(ptr).__refitem__()
+
+    fn as_ptr(inout self) -> AnyPointer[Self.ElementType]:
+        """Get a mutable pointer to the elements contained by this tuple.
```
JoeLoser, line 444: **Suggestion** `s/tuple/array` here again (and in the `Returns:` part).

stdlib/src/utils/static_tuple.mojo
```diff
@@ -254,3 +314,139 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+
+        var array_copy = Reference[Self.type, mutability, self_life](
+            Reference(self)[].array
+        )
+        var ptr = __mlir_op.`pop.array.gep`(
+            array_copy.get_unsafe_pointer().address, normalized_idx.value
+        )
+        return AnyPointer(ptr).__refitem__()
+
+    fn as_ptr(inout self) -> AnyPointer[Self.ElementType]:
```
JoeLoser, line 443: **Question** Do we want to expose this public API of `as_ptr()` yet or have it come in a follow-on?  Of note, `String._as_ptr()` is private (and not `as_ptr()` despite it getting used in very-public ways by users and in the library code probably in some places.

This is non-blocking, but curious to get your thoughts.  It would also be good to think in the future about API naming here (do we like `as_ptr()`?).  It's also inherently unsafe since it returns an `AnyPointer` as you know.

CC: @ConnorGray

stdlib/test/utils/test_tuple.mojo
```diff
@@ -98,8 +99,72 @@ def test_tuple_pointer():
...
+
+
+def test_array_str():
+    var arr = Array[String, 3]("hi", "hello", "hey")
+
+    assert_equal(arr[0], "hi")
+    assert_equal(arr[1], "hello")
+    assert_equal(arr[2], "hey")
+
+    # Test mutating a tuple through its __refitem__
```
JoeLoser, line 137: `s/tuple/array` 😄

JoeLoser (commented):
In general, looks pretty good! Thanks for pushing on this.  This will be a critical core type and will simplify several things. 🎉

stdlib/test/utils/test_tuple.mojo
```diff
@@ -98,8 +99,72 @@ def test_tuple_pointer():
     assert_equal(tup[2], 3)
 
 
+def test_array_int():
+    var arr = Array[Int, 3](0, 0, 0)
+
+    assert_equal(arr[0], 0)
```
JoeLoser, line 105: Can we also add tests for negative indexing support, please? Internally, we have a PR for supporting negative indexing, so for parity it'd be good to go ahead and add support here.

stdlib/src/utils/static_tuple.mojo
lattner: Given we have a List now (following Python convention) I think a bunch of folk will be confused about Array (the name in a lot of other languages) vs List.  This is a C-style inline array, so WDYT about a name like "InlineArray" to make this more explicit?

lattner:
This is super cool @lsh great work!

stdlib/src/utils/static_tuple.mojo
lsh: Agreed, removed this method.

stdlib/src/utils/static_tuple.mojo
lsh: I factored out part of this to `_get_unsafe_reference` since the `gep` is really the part worth isolating IMO, and the main difference is that one can be a compile time `if`.

mikowals:
There is a problem with how `__refitem__` provides `__getitem__` when `InlineArray` is used as an `alias`.  Maybe since in this case `self` is immutable in `__refitem__`?  A failing test to demonstrate this is:

```mojo
alias arr_const = InlineArray[Int, 3](1, 2, 3)
assert_equal(arr_const[0], 1)
```
A workaround is adding a `__getitem__` with immutable `self`.  
```mojo
fn __getitem__(self, idx: Int) -> Self.ElementType:
    return self.__refitem__(idx)
```

stdlib/src/utils/static_tuple.mojo
mikowals: I think keeping the feature but marking it private and subject to change with an underscore prefix would be better.  I believe the ability to access the underlying storage of `static_tuple` was requested specifically and only added recently so it would be a shame to lose it.

Also,  UnsafePointers  or References to `_array` are created several times in the code.  `UnsafePointer` now converts to an immortal `Reference` with suffix `[]`.  So `_as_ptr()` and `_as_ptr()[]` could replace all these and  `_get_reference_unsafe` could be removed.  

Since creating an immortal Reference from an UnsafePointer is new maybe others more knowledgable than me need to confirm its use is a good idea.

stdlib/src/utils/static_tuple.mojo
soraros: @lattner Shouldn't the emphasis more on `Static` (having a size known at compile time) than `Inline`?

There is one advantage in keeping `Array`: it feels nature when/if we have `NDArray`.

stdlib/src/utils/static_tuple.mojo
soraros: @mikowals Maybe you meant to put your comment in the `as_ptr` thread?

stdlib/src/utils/static_tuple.mojo
mikowals: I think keeping the feature but marking it private and subject to change with an underscore prefix would be better. I believe the ability to access the underlying storage of static_tuple was requested specifically and only added recently so it would be a shame to lose it.

Also, UnsafePointers or References to _array are created several times in the code.  UnsafePointer now converts to an immortal Reference with suffix []. So _as_ptr() and _as_ptr()[] could replace all these and _get_reference_unsafe could be removed.

Since creating an immortal Reference from an UnsafePointer is new maybe others more knowledgable than me need to confirm its use is a good idea.

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: -1 for "Array" as array is already defined in the python stdlib: https://docs.python.org/3/library/array.html (note that this module doesn't make much sense in mojo as python's array's use case is covered by Mojo's List). I can already see the confused faces of Python devs when I'll try to explain them that Mojo's array is different from python's array 😁 

I'm fine with whatever other name, StaticArray, InlineArray, CArray are all fine by me...

stdlib/src/utils/static_tuple.mojo
```diff
@@ -97,6 +97,70 @@ fn _create_array[
...
+        array.get_legacy_pointer().address, index.value
+    )
+    var p = UnsafePointer(ptr)
+    initialize_pointee(p, val)
+
+
+@always_inline
+fn _create_array_mem[
+    size: Int, type: CollectionElement
+](lst: VariadicList[type]) -> __mlir_type[
```
JoeLoser, line 134: **Question** I think I had an earlier comment around using `VariadicList`, but I think we want this to be `*lst: type` to mean "a variadic pack of types" rather than a `VariadicList`.  In general, you don't want to spell `VariadicList` or `VariadicListMem` directly in APIs.

JoeLoser:
> There is a problem with how `__refitem__` provides `__getitem__` when `InlineArray` is used as an `alias`. Maybe since in this case `self` is immutable in `__refitem__`? A failing test to demonstrate this is:
> 
> ```mojo
> alias arr_const = InlineArray[Int, 3](1, 2, 3)
> assert_equal(arr_const[0], 1)
> ```
> 
> A workaround is adding a `__getitem__` with immutable `self`.
> 
> ```mojo
> fn __getitem__(self, idx: Int) -> Self.ElementType:
>     return self.__refitem__(idx)
> ```

Thanks for calling this out.  @lattner would this sort of thing work with the latest top-of-tree `mojo` based on your changes yesterday with handling the case when both `__getitem__`  and `__refitem__` are present for a type?

stdlib/src/utils/static_tuple.mojo
JoeLoser: See https://github.com/modularml/mojo/pull/2294#discussion_r1569499182

stdlib/test/utils/test_tuple.mojo
```diff
@@ -15,6 +15,7 @@
 from testing import assert_equal, assert_false, assert_true
 
 from utils import StaticTuple, StaticIntTuple
+from utils.static_tuple import InlineArray
```
JoeLoser, line 18: **Suggestion** Instead of this, modify `utils/__init__.mojo` to public export `InlineArray`.  Then, you can just modify the line above here to import `InlineArray`.

JoeLoser (commented):
Almost there! Just one minor comment about the use of `VariadicList`, and then I'd be happy to merge this.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -240,3 +240,162 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+    alias type = __mlir_type[
+        `!pop.array<`, size.value, `, `, Self.ElementType, `>`
+    ]
+    var _array: Self.type
+    """The underlying storage for the array."""
+
+    # This constructor will always cause a compile time error if used.
+    # It is used to steer users away from uninitialized memory.
+    @always_inline
+    fn __init__(inout self):
```
lsh, line 268: I want to flag this constructor, which I'm using as a way to enforce safety guarantees. What do you think about this API? @JoeLoser @ConnorGray

stdlib/src/utils/static_tuple.mojo
```diff
@@ -240,3 +240,162 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+            False,
+            (
+                "Initialize with either a variadic list of arguments or a"
+                " default fill element."
+            ),
+        ]()
+        self._array = __mlir_op.`kgen.undef`[_type = Self.type]()
+
+    @always_inline
+    fn __init__(inout self, fill: Self.ElementType):
```
lsh, line 279: bikeshed: is `fill` the right word here?

arthurevans: `SIMD` just uses `value` for this kind of constructor. Not sure if that's a great precedent or not, although it's probably clear enough. From the name I wouldn't be sure if `fill` was a value or a boolean. Maybe `fill_value` or `fill_with`?

Also, have you tried to create an `InlineArray` of size 1? It looks like the two constructors would be ambiguous in that case. Making  this a keyword-only argument would avoid that potential issue.

`InlineArray[Int, 10](fill_with=0xFF)`

lsh: At least on my machine, the constructor of size 1 seems to work fine.

JoeLoser: They're not ambiguous as @lsh points out.  Consider the following smaller program that shows this is the case:

```
struct S:
    fn __init__(inout self, value: Int):
        print("single argument constructor")

    fn __init__(inout self, *values: Int):
        print("variadic argument constructor")

fn main():
    var s1 = S(42) # prints single argument constructor
    var s2 = S(42, 100) # prints variadic argument constructor
```

Re the name, I'm fine with `fill`, `default_value`, `value`, or any of them.  I don't feel strongly and we can easily change it at any time since this isn't a keyword-argument-only constructor (therefore it's just an implementation detail at this point).

stdlib/src/utils/static_tuple.mojo
JoeLoser: This makes sense, since otherwise a zero-argument call like so:

```
var arr = InlinedArray[Int, 10]()
```

would call the variadic constructor.  Doing it as an explicit constructor rather than a `@parameter if` in the variadic constructor makes sense for now.  I don't have a good gauge as to the compile time cost difference between the additional overload (constructor) vs. a `@parameter if` in the variadic constructor case.  Nothing to worry about for now as we're not at the level of compile time perf problems right now 😄

stdlib/src/utils/static_tuple.mojo
```diff
@@ -240,3 +240,162 @@ struct StaticTuple[element_type: AnyRegType, size: Int](Sized):
...
+            IntableType: The inferred type of an intable argument.
+            index: The index of the item.
+
+        Returns:
+            A reference to the item at the given index.
+        """
+        alias i = int(index)
+        constrained[-size <= i < size, "Index must be within bounds."]()
+
+        var normalized_idx = i
```
JoeLoser, line 393: **Quibble** I think we could do this arithmetic at compile time since `size` is a compile time parameter.  But totally fine for now. 😄

stdlib/src/utils/static_tuple.mojo
JoeLoser: FYI this constructor will need public doc strings since it's a public entity still.  We have a PR in-flight internally that adds a decorator (such as `@doc_private`) that we could apply to things like this in the future.  The API doc string checker wouldn't apply in such cases, and the constructor decl would be hidden in our generated docs too.

I moved your comment from above the constructor to be the doc string internally when I just imported this PR.  We also have a [PR in-flight](https://github.com/modularml/mojo/pull/2344) in the public `mojo` repo so the external CI here can do API doc string validation.

JoeLoser (approved):
This is great, @lsh!  Thanks so much for your patience and working on this, even amidst us making you rebase and rework the code a bit as we changed several things across nightly versions.

I've imported this PR internally and I'll keep you updated on the progress of it.

JoeLoser:

Hey @lsh,

As mentioned, we're moving to a new infrastructure for merging contributions to Mojo (we're using a tool called [Copybara](https://github.com/google/copybara)), and your contribution has now been merged into our internal copy of the Mojo Standard Library.  I've added the "merged-internally" label on this PR (FYI @ConnorGray).

The changes in this PR will appear here in the mojo repo nightly branch when we do our next outbound synchronization at the time that the next Mojo nightly is released. That should happen on this coming Monday.

JoeLoser:
Closing as this got merged into the latest `nightly` as https://github.com/modularml/mojo/commit/904ac4ec05dfed0b8f69fb622f8de1d0fcf9bd30

Issue #2306, opened 2024-04-15
[stdlib] Implement os.remove and os.unlink


artemiogr97:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Implement functions to delete files `os.remove` and `os.unlink`

### What is your motivation for this change?

These features are still missing in standard library

artemiogr97:
I have a branch with a prototype, I'm adding some tests to create the PR

Moosems:
Is there a secondary issue for file creation (`touch filename.ext` / `cop nul > filename.ext`)?

artemiogr97:
I don't think so, but there is not a "touch" function in python, so I don't know if it is needed in mojo, in python to create a file you do
```
file = open("my_file")
file.close()
```
or you use a context manager, you can already do the same in mojo

JoeLoser:
Fixed with https://github.com/modularml/mojo/commit/20737dcb02ebf091c77b931e5e8b8abb2d4bb164, thank you for the contribution @artemiogr97! 🎉

PR #2310, opened 2024-04-15
[stdlib] create os.remove and os.unlink

artemiogr97:
Implementation of #2306

### DIFF
--- a/stdlib/src/os/os.mojo
+++ b/stdlib/src/os/os.mojo
@@ -247,3 +247,53 @@ fn abort[
         print(message, flush=True)
 
     return abort[result]()
+
+
+# ===----------------------------------------------------------------------=== #
+# remove/unlink
+# ===----------------------------------------------------------------------=== #
+fn remove(path: String) raises:
+    """Removes the speficied file.
+
+    Args:
+      path: The path to the file.
+
+    """
+    var error = external_call["unlink", Int8](path._as_ptr())
+
+    if error != 0:
+        # TODO get error message, the following code prints it
+        # var error_str = String("Something went wrong")
+        # _ = external_call["perror", Pointer[NoneType]](error_str._as_ptr())
+        # _ = error_str
+        raise Error("Can not remove file: " + path)
+
+
+fn remove[pathlike: os.PathLike](path: pathlike) raises:
+    """Removes the speficied file.
+
+    Args:
+      path: The path to the file.
+
+    """
+    remove(path.__fspath__())
+
+
+fn unlink(path: String) raises:
+    """Removes the speficied file.
+
+    Args:
+      path: The path to the file.
+
+    """
+    remove(path)
+
+
+fn unlink[pathlike: os.PathLike](path: pathlike) raises:
+    """Removes the speficied file.
+
+    Args:
+      path: The path to the file.
+
+    """
+    remove(path.__fspath__())
--- a/stdlib/src/os/os.mojo
+++ b/stdlib/src/os/os.mojo
@@ -253,13 +253,15 @@ fn abort[
 # remove/unlink
 # ===----------------------------------------------------------------------=== #
 fn remove(path: String) raises:
-    """Removes the speficied file.
+    """Removes the specified file.
+    If the path is a directory or it can not be deleted, an error is raised.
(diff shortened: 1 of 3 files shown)


stdlib/src/os/os.mojo
```diff
@@ -244,3 +244,53 @@ fn abort[
         print(message, flush=True)
 
     return abort[result]()
+
+
+# ===----------------------------------------------------------------------=== #
+# remove/unlink
+# ===----------------------------------------------------------------------=== #
+fn remove(path: String) raises:
+    """Removes the speficied file.
```
JoeLoser, line 253: **Quibble** `s/speficied/specified`

JoeLoser, line 253: **Suggestion** Be clear about the semantics about what happens if the argument passed is a path to a directory.  In Python, according to their [docs](https://docs.python.org/3/library/os.html#os.remove), if a directory is passed, an error is raised.

Pedantically, should we also spell out whether the path is relative or absolute?

stdlib/src/os/os.mojo
```diff
@@ -244,3 +244,53 @@ fn abort[
...
+
+    Args:
+      path: The path to the file.
+
+    """
+    var error = external_call["unlink", Int8](path._as_ptr())
```
JoeLoser, line 259: **Suggestion** It doesn't matter too much in practice, but I'd use `Int` instead of `Int8` here as it more closely matches the UNIX function return type as seen [here](https://man7.org/linux/man-pages/man2/unlink.2.html).  This would follow the element of least surprise to future readers even though the return values all technically fit in the type domain of an `Int8` as you have it now.

stdlib/src/os/os.mojo
```diff
@@ -244,3 +244,53 @@ fn abort[
...
+
+    Args:
+      path: The path to the file.
+
+    """
+    remove(path.__fspath__())
+
+
+fn unlink(path: String) raises:
+    """Removes the speficied file.
```
JoeLoser, line 280: Ditto here about the small typo as above.

stdlib/test/os/test_remove.mojo
```diff
@@ -0,0 +1,75 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo -debug-level full %s
```
JoeLoser, line 13: **Suggestion** Change this to just `RUN: %mojo-no-debug %s` as we recently dropped the `-debug-level full` stuff in our stdlib tests in https://github.com/modularml/mojo/commit/b6a8f871174bde8211b43bfec50a5a7104e116df

stdlib/src/os/os.mojo
```diff
@@ -244,3 +244,53 @@ fn abort[
...
+
+    Args:
+      path: The path to the file.
+
+    """
+    var error = external_call["unlink", Int8](path._as_ptr())
+
+    if error != 0:
+        # TODO get error message, the following code prints it
```
JoeLoser, line 262: **Suggestion** Do you mind filing a GitHub issue to support capturing the error message on failure and link to it here in the code?  This will help others who are interested in improving this to be able to cross-reference things.

JoeLoser (commented):
Generally looks good, thank you for the contribution!  Just a few minor comments about being more specific about what's currently supported, etc.  Once that is addressed, I'm happy to merge this!

artemiogr97:
Sure, I'll check that later today

stdlib/src/os/os.mojo
artemiogr97: I'll wait for the merge to create the issue, to be able to reference the code in the nightly branch

JoeLoser (approved):

stdlib/src/os/os.mojo
abduld: This should be Int32 since it matches the C int type

PR #2313, opened 2024-04-16
[stdlib] Update stdlib corresponding to `2024-04-16` nightly/mojo

patrickdoc:
This updates the stdlib with the internal commits corresponding to today's nightly release: mojo 2024.4.1618 .

In the future, we may push these updates directly to nightly branch.


dlumma (approved):
👍 🥳

Issue #2293, opened 2024-04-13
[Feature Request] Allow StaticIntTuple mask for SIMD.shuffle


mikowals:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

As title.  I want to build the list of indices in the mask with a for-loop at compile time.  StaticIntTuple seems like the obvious candidate for such a mask but I other types could work.

I already implemented this and can finish the PR after inferred parameters are in nightly.

### What is your motivation for this change?

Reduce boilerplate writing overload functions for different mask sizes.  See Discord [discussion](https://discord.com/channels/1087530497313357884/1205809638751346688/1205817726086418463).

### Any other details?

The current `SIMD.shuffle` relies on `_shuffle_list` with the signature:

```mojo
fn _shuffle_list[ *mask: Int, output_size: Int = size ](self, other: Self) -> SIMD[type, output_size]:
```

For this feature the signature of `_shuffle_list` using `StaticIntTuple` would be like this:

```mojo
fn _shuffle_list[inferred output_size: Int, mask: StaticIntTuple[output_size]](self, other: Self) -> SIMD[type, output_size]:
```

Getting these to have the order of `output_size` and `mask` the same would be nicer.  For most cases it would be hidden behind `SIMD.shuffle`.

LJ-9801:
Any news on when the inferred parameters will be supported? Seems like a great feature to have.

mikowals:
According to [this](https://github.com/modularml/mojo/issues/1245) it is done in main branch so it should be in nightly branch soon and released this month or next I suppose.

mikowals:
And if you are keen for this shuffle functionality the implementation is trivial.

```
@always_inline("nodebug")
fn _shuffle_list[
    dtype: Dtype, size: Int, output_size: Int, mask: StaticIntTuple[output_size]
](a: SIMD[dtype: size], other: SIMD[dtype, size]) -> SIMD[dtype, output_size]:
    
    #could check that each entry in mask is greater than or equal to 0 and less than 2 * size
    return __mlir_op.`pop.simd.shuffle`[
        mask = mask.data.array,
        _type = __mlir_type[
            `!pop.simd<`, output_size.value, `, `, type.value, `>`
        ],
    ](a.value, b.value)
```

mikowals:
fixed by #2315

PR #2315, opened 2024-04-17
[stdlib] SIMD.shuffle with StaticIntTuple mask

mikowals:
Implement #2293.  New overloads for `SIMD.shuffle()` like this: 
```
fn shuffle[mask: StaticIntTuple[size]](self) -> Self:
```
additionally:
- added tests for `SIMD.join()` and `SIMD.shuffle()`
- simplified `join` implementation and always use shuffle rather than 2 inserts when mask size > 32
- cleaned up doc strings about shuffle output length
- minor refactor of original `_shuffle_list` to use `output_size` parameter rather than calculate `len(mask)` many times

As I mentioned in the issue, this can be nicer when inferred parameters are implemented but I guess that is true of most things.

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -1468,13 +1468,10 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
         @parameter
         fn _convert_variadic_to_pop_array[
             *mask: Int
-        ]() -> __mlir_type[
-            `!pop.array<`, variadic_len[mask]().value, `, `, Int, `>`
-        ]:
-            alias size = variadic_len[mask]()
+        ]() -> __mlir_type[`!pop.array<`, output_size.value, `, `, Int, `>`]:
             var array = __mlir_op.`kgen.undef`[
                 _type = __mlir_type[
-                    `!pop.array<`, variadic_len[mask]().value, `, `, Int, `>`
+                    `!pop.array<`, output_size.value, `, `, Int, `>`
                 ]
             ]()
 
@@ -1483,7 +1480,7 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
             fn fill[idx: Int]():
                 alias val = mask[idx]
                 constrained[
-                    0 <= val < 2 * size,
+                    0 <= val < 2 * output_size,
                     "invalid index in the shuffle operation",
                 ]()
                 var ptr = __mlir_op.`pop.array.gep`(
@@ -1491,7 +1488,7 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
                 )
                 __mlir_op.`pop.store`(val, ptr)
 
-            unroll[fill, size]()
+            unroll[fill, output_size]()
             return array
 
         alias length = variadic_len[mask]()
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/simd.mojo
```diff
@@ -1393,6 +1390,49 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+                    "invalid index in the shuffle operation",
+                ]()
+
+            unroll[_check, output_size]()
+
+            return True
+
+        constrained[
+            _check_indices[mask](), "invalid index in the shuffle operation"
+        ]()
```
soraros, line 1428: Why not simply
```mojo
fn _check():
  ...
unroll[_check, ...]()
```

mikowals: Good point.  I fixed it.

JoeLoser:
Thanks for the contribution! 🎉 

I'm out of office for the first part of this week (Monday - Wednesday), so I've added a few additional reviewers to this PR in hopes they can take a look before I'm back.

stdlib/src/builtin/simd.mojo
```diff
@@ -1669,92 +1735,17 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
-                29,
-                30,
-                31,
-                output_size = 2 * size,
-            ](other)
-
-        var res = SIMD[type, 2 * size]()
-        res = res.insert(self)
-        return res.insert[offset=size](other)
+        fn build_indices() -> StaticIntTuple[2 * size]:
```
JoeLoser, line 1738: **Suggestion** (Internal PR feedback): do you mind `always_inline`ing this function please?

stdlib/test/builtin/test_simd.mojo
```diff
@@ -325,6 +325,43 @@ def test_shift():
     )
 
 
+def test_shuffle():
+    print("== test_shuffle")
```
JoeLoser, line 329: **Suggestion** You can drop the `print` as it's only needed for `FileCheck`-based tests. Ditto for `test_join`, etc.

JoeLoser (approved):
Nice! Just a few quibbles.

Thanks so much for the contribution! 🎉

stdlib/src/builtin/simd.mojo
mikowals: Added it.

stdlib/test/builtin/test_simd.mojo
mikowals: I removed the prints.  The new location information for asserts is great!

JoeLoser:

Hey @mikowals,

Thanks so much for the contribution! 🎉

We're moving to a new infrastructure for merging contributions to Mojo (we're using a tool called [Copybara](https://github.com/google/copybara)), and your contribution has now been merged into our internal copy of the Mojo Standard Library. I've added the "merged-internally" label on this PR.

The changes in this PR will appear here in the mojo repo nightly branch when we do our next outbound synchronization at the time that the next Mojo nightly is released. That should happen within the next 24 hours.

JoeLoser:
🎉 🔥 This landed in tonight's nightly release in https://github.com/modularml/mojo/pull/2480!

PR #2316, opened 2024-04-17
[mojo-stdlib] Implement `__float__`,`__add__`,`__abs__` for Bool struct

ghost:
Implement some APIs for mojo Bool struct

### DIFF
--- a/stdlib/src/builtin/bool.mojo
+++ b/stdlib/src/builtin/bool.mojo
@@ -286,7 +286,6 @@ struct Bool(
             1 if the Bool is True, 0 otherwise.
         """
         return Int(
-
             __mlir_op.`pop.cast`[_type = __mlir_type.`!pop.scalar<index>`](
                 self.value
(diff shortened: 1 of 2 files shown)


soraros:
I'm against add any of the three APIs, especially `__add__`.

- If the idea is to mimic Python, even with the current implementation of `__add__`, we still can't add bool to e.g. float.
- When we have traits for numeric types, both `__float__` and `__abs__` can be implemented in more generic ways, like deriving from `Boolable` or `Intable`.

zhoujingya:
> I'm against add any of the three APIs, especially `__add__`.
> 
> * If the idea is to mimic Python, even with the current implementation of `__add__`, we still can't add bool to e.g. float.
> * When we have traits for numeric types, both `__float__` and `__abs__` can be implemented in more generic ways, like deriving from `Boolable` or `Intable`.

alright, this pull request canbe closed

PR #2318, opened 2024-04-17
[stdlib] Use `UInt8` instead of `Int8` to store `Error`'s message

gabrieldemarmiesse:
Related to https://github.com/modularml/mojo/issues/2317

I think we can do this safely by making PRs for one struct at a time. Here is the change for the `Error` struct. I believe that as we progress on the migration, the `bitcast` will go away


JoeLoser (approved):
Looks good, thank you for starting this work! 🎉 

@ConnorGray FYI I ran some internal tests with this PR and it looked fine.  Going to merge this.

PR #2323, opened 2024-04-17
[stdlib] Implement `List.__str__()`

gabrieldemarmiesse:
PR that can serve as a reference for https://github.com/modularml/mojo/pull/2190#issuecomment-2062101102

Note that it causes a bug that seems on the parser side. We get this:

```
RUN: at line 13: mojo /projects/open_source/mojo/stdlib/test/builtin/test_str.mojo
+ mojo /projects/open_source/mojo/stdlib/test/builtin/test_str.mojo
/projects/open_source/mojo/stdlib/test/builtin/test_str.mojo:18:5: error: cannot bind MLIR type 'None' to trait 'StringableCollectionElement'
def test_str_none():
    ^
/projects/open_source/mojo/stdlib/src/builtin/str.mojo:62:8: note: MLIR type cannot satisfy required trait function here
    fn __str__(self) -> String:
       ^
```

So basically, making an overload to allow `str(List[...])` makes `str(List[Int](1, 2, 3))` return `"[1, 2, 3]"`, but breaks `str(None)`. I think the parser doesn't pick the right overload.

I'll try to make a small reproducer and open an issue to track the resolution

### DIFF
--- a/stdlib/src/builtin/str.mojo
+++ b/stdlib/src/builtin/str.mojo
@@ -18,6 +18,7 @@ These are Mojo built-ins, so you don't need to import them.
 # ===----------------------------------------------------------------------=== #
 # Stringable
 # ===----------------------------------------------------------------------=== #
+from ..collections.list import StringableCollectionElement
 
 
 trait Stringable:
@@ -169,3 +170,19 @@ fn str[T: StringableRaising](value: T) raises -> String:
         If there is an error when computing the string representation of the type.
     """
     return value.__str__()
+
+
+@always_inline
+fn str[T: StringableCollectionElement](value: List[T]) -> String:
+    """Get the string representation of a list of strings.
+
+    Parameters:
+        T: The type conforming to Stringable and a CollectionElement.
+
+    Args:
+        value: The object to get the string representation of.
+
+    Returns:
+        The string representation of the object.
+    """
+    return __type_of(value).__str__(value)
--- a/stdlib/src/builtin/str.mojo
+++ b/stdlib/src/builtin/str.mojo
@@ -18,7 +18,6 @@ These are Mojo built-ins, so you don't need to import them.
 # ===----------------------------------------------------------------------=== #
 # Stringable
 # ===----------------------------------------------------------------------=== #
-from ..collections.list import StringableCollectionElement
 
 
 trait Stringable:
@@ -170,19 +169,3 @@ fn str[T: StringableRaising](value: T) raises -> String:
         If there is an error when computing the string representation of the type.
     """
     return value.__str__()
-
-
-@always_inline
-fn str[T: StringableCollectionElement](value: List[T]) -> String:
-    """Get the string representation of a list of strings.
-
-    Parameters:
-        T: The type conforming to Stringable and a CollectionElement.
-
-    Args:
-        value: The object to get the string representation of.
-
-    Returns:
-        The string representation of the object.
-    """
-    return __type_of(value).__str__(value)
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -69,6 +69,10 @@ struct _ListIter[
         return len(self.src[]) - self.index
 
 
+trait StringableCollectionElement(Stringable, CollectionElement):
+    pass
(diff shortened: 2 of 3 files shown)


stdlib/src/builtin/str.mojo
```diff
@@ -169,3 +170,19 @@ fn str[T: StringableRaising](value: T) raises -> String:
         If there is an error when computing the string representation of the type.
     """
     return value.__str__()
+
+
+@always_inline
+fn str[T: StringableCollectionElement](value: List[T]) -> String:
+    """Get the string representation of a list of strings.
```
JoeLoser, line 177: **Quibble** "list of strings" -> "list of stringable values" or something?  The `T` need not be a `String`, but just a `Stringable` like thing.

JoeLoser (commented):
This is great!  We've talked internally about this in the past (FYI @abduld and @ConnorGray). 

Would love to see a small GitHub issue and maybe @Mogball or @laszlokindrat knows where things are going wrong in the parser/overload selection.

stdlib/src/builtin/str.mojo
gabrieldemarmiesse: Oupsy, I shouldn't trust copilot

gabrieldemarmiesse:
@JoeLoser @Mogball @laszlokindrat  if you don't want to read this PR, I've made a small reproducible example in an issue. See #2336

gabrieldemarmiesse:
I have removed the `str()` overload. Now it's working, we can add it later when  #2336 is fixed. For the moment, we can say that to call conditional methods, users should do `__type_of(x).the_method(x, ...)`, with this syntax, it will be easy to change in the future, a "search everywhere" of `__type_of` will help, we can even make a regex to replace them automatically. I also allows us to use "self" as argument name in the implementation, which is nice to avoid changing it in the future, the diff will be smaller, only the signature will change.

stdlib/src/collections/list.mojo
```diff
@@ -568,3 +572,36 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+            self: The list to represent as a string.
+
+        Parameters:
+            U: The type of the elements in the list. Must implement the
+              traits `Stringable` and `CollectionElement`.
+
+        Returns:
+            A string representation of the list.
+        """
+        # TODO: Use `String.join()` when it's working with List.
```
JoeLoser, line 601: **Question** Can you elaborate/file a GitHub issue?  Once this lands, we could teach `String.join` to work with this new `StringableCollectionElement`, right?  Or we could start generic algorithms for joining things.

stdlib/src/collections/list.mojo
```diff
@@ -568,3 +572,36 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        var result = String("[")
```
JoeLoser, line 602: **Suggestion** Would it make sense to have the result be a `List[UInt8]` and initialize its capacity ahead of time with the pre-computed size?  Then we could fill in the buffer rather than doing some potential reallocations/copies in the `+=` calls.

JoeLoser (approved):

stdlib/test/collections/test_list.mojo
```diff
@@ -571,6 +571,19 @@ def test_constructor_from_other_list_through_pointer():
     assert_equal(some_list.capacity, capacity)
 
 
+def test_converting_list_to_string():
```
JoeLoser, line 574: **Suggestion** Perhaps add a test case for nested lists?  E.g. `List[List[Int]]`.

gabrieldemarmiesse: I can't do this, because `List[Int]` is not Stringable, even after this PR :(

stdlib/src/collections/list.mojo
gabrieldemarmiesse: Done. If str(x) is one char, we shouldn't get any allocations.

gabrieldemarmiesse:
@JoeLoser feel free to re-review when you have the time

JoeLoser (approved):

Issue #2325, opened 2024-04-18
[stdlib] Reverse iterator for Set and Dict


jayzhan211:
As a follow-up, we can go implement reverse iterator support for `Set` and `Dict`, right?  I'm happy to create some GitHub issues for that if anyone is interested in working on that.

_Originally posted by @JoeLoser in https://github.com/modularml/mojo/pull/2215#pullrequestreview-2001824680_
            
```[tasklist]
### Tasks
- [ ] Set
- [ ] Dict
```

I would like to work on this

JoeLoser:
Thanks for filing this.  Should we make a `set` reverse-iterable?  Of note, Python's `set` isn't reversible, which surprised me.  However, there is a `dict_reversekeyiterator` in Python for example so that

```
ys = {"hello": 42, "bar" : 100}
for y in reversed(ys):
   print(y)
```

is well-formed for example.

PR #2327, opened 2024-04-18
[stdlib] Support reverse for Dict keys

jayzhan211:
Part of #2325 

The behavior is align with Python, only keys is iterated

> reversed(d)
Return a reverse iterator over the keys of the dictionary. This is a shortcut for reversed(d.keys()).

I plan to support `d.values()` and `d.items()` in follow up PR too!

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -80,6 +80,7 @@ struct _DictEntryIter[
     @always_inline
     fn __next__(inout self) -> Self.ref_type:
         while True:
+
             @parameter
             if forward:
                 debug_assert(
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -785,10 +785,10 @@ struct Dict[K: KeyElement, V: CollectionElement](
     ) -> _DictKeyIter[
         K, V, mutability, self_life, False
     ]:
-        """Iterate backwards over the list, returning immutable references.
+        """Iterate backwards over the dict keys, returning immutable references.
 
         Returns:
-            A reversed iterator of immutable references to the list elements.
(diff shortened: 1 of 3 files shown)


stdlib/src/collections/dict.mojo
```diff
@@ -78,18 +80,33 @@ struct _DictEntryIter[
     @always_inline
     fn __next__(inout self) -> Self.ref_type:
         while True:
-            debug_assert(self.index < self.src[]._reserved, "dict iter bounds")
+            if forward:
```
JoeLoser, line 83: **Suggestion** Since `forward` is a parameter and not a runtime value (as it should be), we can evaluate this branch at compile-time using `@parameter if` instead of `if`.  Ditto below.

JoeLoser (approved):
Looks good, but see my comment about using `@parameter if` instead of runtime `if`.  Thanks for working on this!

jayzhan211:
Thanks for your review, fix it! @JoeLoser

stdlib/test/builtin/test_reversed.mojo
```diff
@@ -24,5 +24,62 @@ def test_reversed_list():
...
+
+    assert_equal(keys, "bd")
+
+    keys = String("")
+    for key in reversed(dict):
+        keys += key[]
+
+    assert_equal(keys, "db")
+
+    # Empty dict is iterable
```
JoeLoser, line 58: **Suggestion** This is good as you have it, but it seems cleaner to not use the same local variable for testing this part.  Instead, just use a new local variable that is an empty dict rather than popping the elements.

JoeLoser (approved):
Awesome, thank you for the contribution! ❤️ ❤️ ❤️

Issue #2331, opened 2024-04-18
[tools] [stdlib] API doc string validation


JoeLoser:
Now that other `mojo` tools understand the package name in the open-source standard library directory structure as of the recently `nightly/mojo` release, we can revive a fixed-up variant of the API doc string scripts removed in https://github.com/modularml/mojo/pull/2118.  I haven't checked, but `mojo doc` and friends should work.

I recommend we write it in Python probably for maintainability sake.  The high-level is we need to run API doc string validation script that runs on each module (directory) in `stdlib/src` and also `stdlib/test` (such as `test_utils`).  So we'll need to walk over those directories, find all the subdirectories (modules), and run `mojo doc -warn-missing-doc-strings` on each module.  The original PR had some bugs in that script with the arguments, but the ideas presented in this GitHub issue point in the right direction.

It's desirable to have this check externally in our CI since we have this check internally in a different shape/form tied in at our build system and CI level.  When we import PRs, we don't want things to pass external CI and then fail internal CI for things like this; it's not an effective use of maintainer's time.

FYI @ConnorGray (but external contributors also may be interested in this).  I can do this too if nobody else gets to it soon.

Note:
- I think the current top of tree in the `nightly` branch may fail some API doc string validation that we may need to fix up.

sarfarazsiddiquii:
hi @JoeLoser i can take this issue.
can you assign me?

JoeLoser:
> hi @JoeLoser i can take this issue. can you assign me?

Assigned! Feel free to ping if you want any help or run into issues, and I'm happy to help out here.

sarfarazsiddiquii:
Thanks, @JoeLoser, I'm considering developing a Python script to automate the validation of documentation strings. The script will be placed in the root directory and utilize the `subprocess` module to execute the `mojo doc -warn-missing-doc-strings` command on every module within the `stdlib/src and stdlib/test` directories.

It will iterate through these directories, run the command for each module, capture the output, and identify any warnings about missing documentation strings.

Please let me know if this proposed solution sounds appropriate, or if you have any other suggestions before I proceed with implementing the script

JoeLoser:
> Thanks, @JoeLoser, I'm considering developing a Python script to automate the validation of documentation strings. The script will be placed in the root directory and utilize the `subprocess` module to execute the `mojo doc -warn-missing-doc-strings` command on every module within the `stdlib/src and stdlib/test` directories.
> 
[...]

PR #2344, opened 2024-04-19
[tools] Fix API doc string validation

sarfarazsiddiquii:
This PR solves the issue (ref #2331).

The warnings from `stdlib/src/` have been resolved and integrated into CI.


JoeLoser:
Hey @sarfarazsiddiquii, thanks for the contribution!  Do you mind running `mojo format stdlib/src/collections/optional.mojo` to fix the CI error [here](https://github.com/modularml/mojo/actions/runs/8760398314/job/24045279833?pr=2344)?  I think it would also be great to mention the API doc string validation in the contributing or style guide, maybe somewhere after [this section](https://github.com/modularml/mojo/blob/nightly/stdlib/docs/style-guide.md#mojo-format).

JoeLoser:
I also think it would be nice to run the API doc string validation in https://github.com/modularml/mojo/blob/20737dcb02ebf091c77b931e5e8b8abb2d4bb164/stdlib/scripts/build-stdlib.sh#L40 for example after we build the `stdlib.mojopkg` (and before we run the tests).  What do you think?

I think we should incorporate this `mojo doc validate` command in our scripts in some shape or form so developers are notified of any issues before CI fails.  WDYT @ConnorGray?

stdlib/src/builtin/reversed.mojo
```diff
@@ -90,7 +90,12 @@ fn reversed[
...
         value: The list to get the reversed iterator of.
 
+    Parameters:
+        mutability: Indicates if the list is mutable or immutable.
+        self_life: The lifetime policy for the list.
+        T: The type of elements contained in the list.
+
     Returns:
-        The reversed iterator of the list.
+        _ListIter: The reversed iterator of the list.
```
ConnorGray, line 99: **Question** Why does this `Returns` need to be "named" with the type of the returned value?

sarfarazsiddiquii:
@JoeLoser Thanks for the feedback, according to your request i have made changes in the codebase and the documentation as well.

stdlib/src/builtin/reversed.mojo
sarfarazsiddiquii: Hi @ConnorGray, you are correct, there is no need for defining return type. I've removed the redundant return type name from the documentation string. It now reads:

`Returns: The reversed iterator of the list.`

sarfarazsiddiquii:
@JoeLoser i am still getting errors after new changes, can you help me?

stdlib/scripts/build-stdlib.sh
```diff
@@ -39,4 +39,11 @@ STDLIB_PACKAGE_NAME="stdlib.mojopkg"
 FULL_STDLIB_PACKAGE_PATH="${BUILD_DIR}"/"${STDLIB_PACKAGE_NAME}"
 mojo package "${STDLIB_PATH}" -o "${FULL_STDLIB_PACKAGE_PATH}"
 
+echo "Validating API documentation strings in the Standard Library."
```
gabrieldemarmiesse, line 42: I don't think you need to modify `stdlib/scripts/build-stdlib.sh`, verifying docstrings is more of something to do once a pull request is done, and a contributor will push just after. It's a separate workflow from building the stdlib or testing it.

stdlib/scripts/run-tests.sh
```diff
@@ -23,6 +23,12 @@ mkdir -p "${BUILD_DIR}"
 
 source "${SCRIPT_DIR}"/build-stdlib.sh
 
+echo "Validating API documentation strings in the Standard Library."
```
gabrieldemarmiesse, line 26: Same here, testing the docstrings is more something to do when a PR is nearly done. Something like a pre-commit. 
Contributors might want to run tests without checking the docstrings first.

stdlib/docs/style-guide.md
```diff
@@ -59,6 +59,15 @@ pre-commit install
 ```
 and that's it!
 
+ #### API Doc String Validation
+
+Mojo provides a command line utility, `mojo doc validate`, to validate the API doc strings in your code. This ensures that your doc strings are correctly formatted and consistent with the Mojo style guidelines.
```
gabrieldemarmiesse, line 64:
```suggestion
Mojo provides a command line utility, `mojo doc`, to validate the API doc strings in your code. This ensures that your doc strings are correctly formatted and consistent with the Mojo style guidelines.
```

stdlib/docs/style-guide.md
```diff
@@ -59,6 +59,15 @@ pre-commit install
...
+
+```bash
+> mojo doc validate example.mojo
```
gabrieldemarmiesse, line 67:
```suggestion
> mojo doc -warn-missing-doc-strings -o /dev/null example.mojo
```
`mojo doc validate` doesn't exist. It's `mojo doc`

stdlib/scripts/build-stdlib.sh
```diff
@@ -39,4 +39,11 @@ STDLIB_PACKAGE_NAME="stdlib.mojopkg"
...
+
+mojo doc validate "${STDLIB_PATH}" || {
```
gabrieldemarmiesse, line 44: `mojo doc validate` doesn't exist, it's `mojo doc`. This is what causes the CI to fail.

stdlib/scripts/run-tests.sh
```diff
@@ -23,6 +23,12 @@ mkdir -p "${BUILD_DIR}"
...
+mojo doc validate "${REPO_ROOT}/stdlib/src" || {
```
gabrieldemarmiesse, line 27: Same here, `mojo doc validate` doesn't exist. It's `mojo doc`, this is why the CI Doesn't understand what is going on (but I think the error message is not correct, `mojo doc` should output that the directory `validate` doesn't exist.)

sarfarazsiddiquii:
Thank you, @gabrieldemarmiesse, for pointing out the issues in my previous commits. I have resolved all the errors as suggested, and everything seems to be working as expected. Please let me know if there are any other changes required in this PR.

gabrieldemarmiesse:
Looks good! The only thing missing is the DCO. Take a look at the instructions here: https://github.com/modularml/mojo/pull/2344/checks?check_run_id=24070604633

sarfarazsiddiquii:
> Looks good! The only thing missing is the DCO. Take a look at the instructions here: https://github.com/modularml/mojo/pull/2344/checks?check_run_id=24070604633

Done.

.github/workflows/standard_library_tests_and_examples.yml
```diff
@@ -96,6 +96,14 @@ jobs:
 
           # Ensure `FileCheck` from the pre-installed LLVM 15 package is visible
           echo $(brew --prefix llvm@15)/bin/ >> $GITHUB_PATH
+      
+      - name: Validate documentation strings
+        run: |
+          mojo doc -warn-missing-doc-strings -o /dev/null stdlib/src/
+          if [ $? -ne 0 ]; then
```
JoeLoser, line 103: **Suggestion** Unfortunately, this won't work as you hope since the above command still exits with exit code `0` since it produces _warnings_ (not errors).  You can verify this locally:

```bash
 mojo doc -warn-missing-doc-strings -o /dev/null stdlib/src/
Included from /Users/joe/dev/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/joe/dev/mojo/stdlib/src/collections/dict.mojo:49:7: warning: public symbol 'StringableKeyElement' is missing a doc string
trait StringableKeyElement(KeyElement, Stringable):
      ^
Included from /Users/joe/dev/mojo/stdlib/src/os/__init__.mojo:1:
/Users/joe/dev/mojo/stdlib/src/os/os.mojo:274:4: warning: function takes parameters, but no 'Parameters' in doc string
fn remove[pathlike: os.PathLike](path: pathlike) raises:
   ^
Included from /Users/joe/dev/mojo/stdlib/src/os/__init__.mojo:1:
/Users/joe/dev/mojo/stdlib/src/os/os.mojo:298:4: warning: function takes parameters, but no 'Parameters' in doc string
fn unlink[pathlike: os.PathLike](path: pathlike) raises:
   ^
Included from /Users/joe/dev/mojo/stdlib/src/testing/__init__.mojo:1:
/Users/joe/dev/mojo/stdlib/src/testing/testing.mojo:115:4: warning: function takes parameters, but no 'Parameters' in doc string
fn assert_equal[T: Testable](lhs: T, rhs: T, msg: String = "") raises:
   ^
Included from /Users/joe/dev/mojo/stdlib/src/testing/__init__.mojo:1:
/Users/joe/dev/mojo/stdlib/src/testing/testing.mojo:175:4: warning: function takes parameters, but no 'Parameters' in doc string
fn assert_not_equal[T: Testable](lhs: T, rhs: T, msg: String = "") raises:
```

combined with `echo $?` which reports the exit code of `0`.

To "fix" this problem, you can use the `check-file-is-empty.py` script from https://github.com/modularml/mojo/pull/2118 that got removed (just bring it back for this PR).

Longer term, I think we should file a GitHub issue and have a discussion with @River707 and @modocache on their take on this command.  The current structure of producing warnings for invalid API doc strings **and** exiting with an exit code of `0`, in my opinion, pushes complexity to every user (such as with this `check-file-is-empty.py` script).  I'd rather have _some option_ (even if it's `-error-missing-doc-strings` or name to be bikeshedded) which would give me a non-zero exit code if there are any invalid doc strings.  I think that will be the common use case for users of this command.

stdlib/scripts/run-tests.sh
JoeLoser: I agree this is more akin to `mojo format` sort of thing - something needed to be fixed up prior to landing a PR.  After this lands, do you have any interest in exploring hooking this up with the `pre-commit` things you've set up recently?  I think that would be great.

.github/workflows/standard_library_tests_and_examples.yml
sarfarazsiddiquii: That is correct, i will look into the file and make changes accordingly by looking at the mentioned file.

sarfarazsiddiquii:
@JoeLoser In the commit, I restored the code in the file `check-file-is-empty.py`, and integrated it into `Standard Library tests and examples` which will check if the file `warnings.txt` contains any warnings. If the file is not empty, meaning it has warnings, the test will fail.

This commit caused license error can you help me with it?
Thank you.

JoeLoser:
> @JoeLoser In the commit, I restored the code in the file `check-file-is-empty.py`, and integrated it into `Standard Library tests and examples` which will check if the file `warnings.txt` contains any warnings. If the file is not empty, meaning it has warnings, the test will fail.
> 
> This commit caused license issue can you help me with it, and also review if the changes i made are correct? Thank you.

Hey! Sorry for the delay in response — I was on PTO earlier this week.  The CI failure [here](https://github.com/modularml/mojo/actions/runs/8775143436/job/24077140590?pr=2344#step:7:16) is trying to tell you that we need to add the license part at the top of the file of `stdlib/scripts/check-file-is-empty.py`.  However, you already have the license, so I suspect this is an issue with the license file checker itself.  Specifically, [this line](https://github.com/modularml/mojo/blob/d344d178aec2b80c5772261b8a6e441202a752de/stdlib/scripts/check-licenses.mojo#L38) assumes the first line will be the beginning of the license itself.  However, in your case, it's a shebang, meaning the check won't pass, pedantically.  To fix this, one approach would be to do a little bit of parsing (see if it starts with `!` and if so, skip that line (when you hit a newline character)) and assume the rest of the file is license + rest.  Then you can keep the same `startswith` logic for the license, etc.  Thoughts, @gabrieldemarmiesse?

gabrieldemarmiesse:
Sure I can do that. In the meantime, you can remove the shebang and just call the script with `python3 path_to_script.py` that should do the trick

sarfarazsiddiquii:
@JoeLoser I did what you suggested however the problem is not yet resolved.
Can you help me with it.

gabrieldemarmiesse:
Hi, I took at look at the pull request and saw that if we incorporated it in the pre-commit hook, it should make things much simpler.  I pulled your commits and modified the code a bit more and arrived at this result: https://github.com/modularml/mojo/compare/nightly...gabrieldemarmiesse:mojo:include_to_pre_commit  

@sarfarazsiddiquii  Do you think I should open a new PR with my branch? Or do you want to pull my commits to your branch maybe? It's as you wish. If I make a new PR with my branch, since your commits are there, you'll still be credited as one of the authors.

sarfarazsiddiquii:
> Hi, I took at look at the pull request and saw that if we incorporated it in the pre-commit hook, it should make things much simpler. I pulled your commits and modified the code a bit more and arrived at this result: [nightly...gabrieldemarmiesse:mojo:include_to_pre_commit](https://github.com/modularml/mojo/compare/nightly...gabrieldemarmiesse:mojo:include_to_pre_commit)
> 
> @sarfarazsiddiquii Do you think I should open a new PR with my branch? Or do you want to pull my commits to your branch maybe? It's as you wish. If I make a new PR with my branch, since your commits are there, you'll still be credited as one of the authors.

Thanks @gabrieldemarmiesse for helping me on this issue. I think you should open a new PR since it has both our contributions.

gabrieldemarmiesse:
Great :) #2444 was opened I think you can close this PR

PR #2349, opened 2024-04-20
[stdlib] Fix wrong instructions to use `List.__str__`

gabrieldemarmiesse:
It's unfortunate that this typo slipped through. My bad. I hope that in the future we'll have automatic tests of docstrings :)


JoeLoser (approved):
Doc testing is coming soon, don't you worry 😄

PR #2350, opened 2024-04-20
[stdlib] Test `assert_raises()` without FileCheck

gabrieldemarmiesse:
Related to #2024


JoeLoser (approved):
Nice!  Much cleaner — thank you! ❤️ 

This also minimizes our dependency on the `has_not` lit feature (i.e. requiring the `not` binary from LLVM).  There's only a handful of tests that still require that.

Issue #2018, opened 2024-03-28
[mojo-stdlib] Add tempdir / tempfile functionality to os/Pathlib


JoeLoser:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Add `tempdir` and `tempfile` functions to the `pathlib` module.

### What is your motivation for this change?

These are generally useful functions that exist in the Python ecosystem that we should have in our `pathlib` module.

kiran-4444:
Can I work on this?

JoeLoser:
> Can I work on this?

Of course, go ahead! Feel free to assign the issue to yourself so other contributors know.

artemiogr97:
@kiran-4444 are you still working on this?, I would like to work on it if you did not continue

kiran-4444:
> @kiran-4444 are you still working on this?, I would like to work on it if you did not continue

Hey, yeah please go ahead.

FarhanAliRaza:
Hello, @kiran-4444  are you working on this?

FarhanAliRaza:
I saw the code of CPython to check how they implement this feature in Python they do create a fileWrappper which has its own close function that deletes the file. 
I tried implementing that like this 

```
struct FileWrapper:
 fn __init__(
        inout self,
        file: FileHandle,


    ):

        self.file = file
```
I am getting an error Filehandle can not be copied because it does not implement the copy function.

FarhanAliRaza:
Another easier method could be to override the __del__ function of Filehandle struct or some other way extending that struct 

```
    @always_inline
    fn __del__(owned self):
        """Closes the file handle."""
        try:
            self.close()
        except:
            pass
```

would become 
```
    @always_inline
    fn __del__(owned self):
        """Closes the file handle."""
        try:
            self.close()
           # delete the file from the file system here
        except:
            pass
```

artemiogr97:
> Hello, @kiran-4444  are you working on this? 

Hello, I'm working on this, but the functions to delete files were not implemented yet, so I opened the issue #2306 and created a pull request for it

artemiogr97:
@JoeLoser maybe you can assign the issue to me? I don't think I can do it myself

JoeLoser:
This is done now, closing this out!  Thanks everyone!

PR #2352, opened 2024-04-20
[stdlib] tempfile and tempdir

artemiogr97:
Implementation for #2018


artemiogr97:
I opened the pull to have some feedback, some points to discuss:

- Are the read methods useful ?
  - personally I don't know in which case I would create a temporary file and read from it at the same time
- should I implement what I need in the os module directly in this pull ?
  - `TemporaryDirectory` is not implemented yet because `os.mkdir` and `os.rmdir` don't exist, I guess `shutil.rmtree` will be necessary too
  -  For `NamedTemporaryFile` there is a slight difference with respect to the python implementation, the name attribute could end up being relative but the functions `os.abspath`, and `os.normpath` still don't exist
     - also how different platforms are going to be handled in mojo, in python there are dedicated modules `posix` and `ntpath`, I guess the same should be done in mojo?

artemiogr97:
#2439 must be merged first

JoeLoser:
> #2439 must be merged first

Sorry for the delay here.  Is it possible to decouple this PR from #2439?  I think temporary file things would be useful even outside of the `rmtree` functionality, so I'd rather not couple them.

FYI @laszlokindrat @Dan13llljws @ConnorGray as we were talking about temp files today.

laszlokindrat:
> Sorry for the delay here. Is it possible to decouple this PR from #2439? I think temporary file things would be useful even outside of the `rmtree` functionality, so I'd rather not couple them.

I completely agree. Can you please rebase this on the latest nightly, and maybe start splitting it up? Also one concrete suggestion: could you please start with the low level `tempfile` APIs, i.e. `gettempdir` and `mkdtemp`, and maybe `mkstemp`, and then we can build the rest on it.

This super cool, we're super happy that you are pushing this forward!

artemiogr97:
> could you please start with the low level `tempfile` APIs, i.e. `gettempdir` and `mkdtemp`, and maybe `mkstemp`, and then we can build the rest on it.
@laszlokindrat I started working on this, I'm having some issues with `mkstemp`, in python it returns a tuple with the file object and the name, I'm not sure how to do this in mojo, the problem is that when I do the following I have an error:
```
tmp_file_and_name = msktemp()
tmp_file = tmp_file_and_name[0]  # error here: 'FileHandle' is not copyable because it has no '__copyinit__'
file_name = tmp_file_and_name[1]
```
I tried doing `tmp_file = tmp_file_and_name.get[0, FileHandle]()` but then I have a double free at runtime.
I guess that the best option is to return only the file handle for now? but we wont have access to the file name

artemiogr97:
I opened a new PR (#2742) that implements `gettempdir` and `mkdtemp`, and a second PR ( #2743) for `TemporaryDirectory` using `mkdtemp`

artemiogr97:
@JoeLoser, @laszlokindrat  I opened another PR #2762 for `NamedTemporaryFile` seems so be easier to implement than `mkstemp` for now, I opened it as a draft since I'm having some issues, I'll leave you all the details in that PR.

Also I guess I can close this PR and also #2439 since everything is now splitted in #2742, #2743 and #2762

PR #2359, opened 2024-04-20
[stdlib] Implement `Dict.__str__()`

gabrieldemarmiesse:
Same logic as `List.__str__()`

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -45,6 +45,14 @@ trait KeyElement(CollectionElement, Hashable, EqualityComparable):
     pass
 
 
+trait StringableKeyElement(KeyElement, Stringable):
+    pass
+
+
+trait StringableCollectionElement(CollectionElement, Stringable):
+    pass
+
+
 @value
 struct _DictEntryIter[
     K: KeyElement,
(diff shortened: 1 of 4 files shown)


stdlib/src/collections/dict.mojo
```diff
@@ -45,6 +45,14 @@ trait KeyElement(CollectionElement, Hashable, EqualityComparable):
     pass
 
 
+trait StringableKeyElement(KeyElement, Stringable):
+    pass
+
+
+trait StringableCollectionElement(CollectionElement, Stringable):
```
JoeLoser, line 52: **Suggestion** Let's move this right by `CollectionElement` and remove it from here and `list.mojo` - WDYT?  I'd prefer having a common spot for it rather than the trait defined twice (once in `list.mojo` and once here).

JoeLoser (approved):
Looks good, but WDYT about moving `StringableCollectionElement` to a common spot right next to `CollectionElement`?

stdlib/src/collections/dict.mojo
gabrieldemarmiesse: Sure thing

gabrieldemarmiesse:
We can merge this PR. I would bring to the attention of the maintainers that currently, the implementations of `__str__` in the stdlib have a small difference (bug?) compared to what is done in python and it can be noticeble. 

See https://github.com/modularml/mojo/issues/2207#issuecomment-2067721811 in which I detail the difference.

stdlib/src/collections/dict.mojo
gabrieldemarmiesse: Done

JoeLoser (approved):

stdlib/src/collections/dict.mojo
```diff
@@ -32,6 +32,7 @@ value types must always be Movable so we can resize the dictionary as it grows.
 See the `Dict` docs for more details.
 """
 from memory import UnsafePointer
+from builtin.value import StringableCollectionElement
```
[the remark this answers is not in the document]
gabrieldemarmiesse: That's how it works! I was wondering how did the compiler know what is auto-imported and what is not

PR #2370, opened 2024-04-21
[stdlib] Use `UnsafePointer` in the `String` struct

gabrieldemarmiesse:
Three things going on here:
* Added a String constructor from an `UnsafePointer`, this will help us phase out the `Pointer` constructor. Added corresponding test.
* Renamed Pointer -> LegacyPointer, for better information and it conveys the state of the code, better searchability too.
* Use DtypePointer instead of `LegacyPointer` when doing `memcpy`.

I'd appreciate a merge of the internal branch in nightly soon-ish otherwise it's a bit hard to guess what function has been changed or not internally. Especially the ones in _startup.mojo

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -18,7 +18,7 @@ These are Mojo built-ins, so you don't need to import them.
 from collections import List, KeyElement
 from sys import llvm_intrinsic, bitwidthof
 
-from memory import DTypePointer, Pointer, UnsafePointer, memcmp, memcpy
+from memory import DTypePointer, LegacyPointer, UnsafePointer, memcmp, memcpy
 
 
 from utils import StringRef, StaticIntTuple, StaticTuple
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):

stdlib/src/builtin/string.mojo
```diff
@@ -400,7 +400,22 @@ struct String(Sized, Stringable, IntableRaising, KeyElement, Boolable):
         self = str(value)
 
     @always_inline
-    fn __init__(inout self, ptr: Pointer[Int8], len: Int):
+    fn __init__(inout self, ptr: UnsafePointer[Int8], len: Int):
```
JoeLoser, line 403: **Suggestion** In the future, given how unsafe both this constructor and the pre-existing one below it are, it would be good to make these keyword-arguments.  I filed an issue internally about this sort of constructor for `StringRef` last month and Chris said:

> Driveby comment: if you do this, please make sure to require a keyword argument since it will be very unsafe.

PR #2377, opened 2024-04-22
[stdlib] Use `UnsafePointer` in `vector.mojo`

gabrieldemarmiesse:
With this change I get a weird error:

```
 ./stdlib/scripts/run-tests.sh && lit -sv stdlib/test
Included from /projects/open_source/mojo/stdlib/src/collections/__init__.mojo:1:
/projects/open_source/mojo/stdlib/src/collections/vector.mojo:250:48: error: '_VecIter' parameter #2 has 'fn(UnsafePointer[InlinedFixedVector[type, size], 0], Int, /) -> type' type, but value has type 'fn(selfptr: UnsafePointer[InlinedFixedVector[type, size], 0], i: Int, /) -> type'
    alias _iterator = _VecIter[type, Self, Self._deref_iter_impl]
                                           ~~~~^~~~~~~~~~~~~~~~~
/projects/open_source/mojo/stdlib/src/__init__.mojo:1:1: note: '_VecIter' declared here
^
```
That does seem like a compiler bug no? Those looks the same to me:
```
fn(UnsafePointer[InlinedFixedVector[type, size], 0], Int, /) -> type
fn(selfptr: UnsafePointer[InlinedFixedVector[type, size], 0], i: Int, /) -> type
```
Maybe I'm missing something?


Moosems:
Not the same, one has a named argument ((sarcasm) yeah, I think it's a compiler bug).

JoeLoser:
Thanks for pushing on this!  I'll import the PR internally and we'll take a look at the compiler bug and see if there's a workaround, etc.

lattner:
Thanks for poking me about this Joe.  @gabrieldemarmiesse the problem here is a disagreement about AnyRegType, I think it works if you change VecIter to:

```
@value
struct _VecIter[
    type: AnyRegType,
    vec_type: AnyType,
    deref: fn (UnsafePointer[vec_type], Int) -> type,
```

AnyRegType is a bad thing :), we need to move off it completely to use AnyType.

gabrieldemarmiesse:
Thanks @lattner it works well :) This PR is ready for review now

stdlib/src/collections/vector.mojo
```diff
@@ -243,5 +243,5 @@ struct InlinedFixedVector[
             An iterator to the start of the vector.
         """
         return Self._iterator(
-            0, self.current_size, Reference(self).get_legacy_pointer()
+            0, self.current_size, UnsafePointer(Reference(self))
```
soraros, line 246: Maybe:
```suggestion
            0, self.current_size, UnsafePointer(self)
```

JoeLoser:
Landed in https://github.com/modularml/mojo/pull/2559!

Issue #2346, opened 2024-04-20
[Feature Request] Add larger than and larger than or equal operator support to String


VMois:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Add larger than (__lt__) and larger than or equal (__le__) operator support to String.

### What is your motivation for this change?

Currently, it is not possible to compare Strings easily. It is very useful when sorting Strings.

zhoujingya:
maybe I can take this one?

VMois:
@zhoujingya, I have already made a snippet (needed it for a project to compare strings). Do you want me to share it with you, or would you prefer to do it yourself? My snippet is very simple. You might want to see if it is possible to use SIMDs when comparing.

VMois:
> maybe I can wok on this one if upstream think this work is worth?

I think @JoeLoser already marked it as a good first issue which means you are free to take it.

zhoujingya:
wow, very appreciate for your snippets.thanks

VMois:
Here is what I use as a quick fix. I would double check that :) And def look at SIMDs.

```mojo
@always_inline
fn __lt__(self, other: String) -> Bool:
    var p1 = self._buffer
    var p2 = other._buffer
    var min_len = len(self)
    if len(other) < min_len:
        min_len = len(other)
    for i in range(min_len):
        var ai = p1[i].cast[DType.uint8]()
        var bi = p2[i].cast[DType.uint8]()
        if ai > bi:
            return False
        if ai < bi:
            return True
    return len(self) <= len(other)

@always_inline
fn __le__(self, other: String) -> Bool:
    return self < other or self == other
```

zhoujingya:
@VMois ,so maybe SIMDs is another issue?

VMois:
> @VMois ,so maybe SIMDs is another issue?

Up to you. You can do a simple version or look at SIMD. I am just a random guy :)

soraros:
Fixed by #2409.

conorbronsdon:
`String` now conforms to `Comparable`, which provides `__gt__`, `__ge__`, and `__le__` via default implementations derived from `__lt__`. See the [String docs](https://docs.modular.com/mojo/std/collections/string/String/).

Thanks!

PR #2378, opened 2024-04-22
[mojo-stdlib] Implement compare operation for String

ghost:
This PR refers to issue #2346  , though the string test is disables now, I still write test cases for those operators

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -52,6 +52,23 @@ fn _ctlz(val: SIMD) -> __type_of(val):
     )
 
 
+@always_inline
+fn _str_compare(str1: String, str2: String) -> Int:
+    var min_len = len(str1) if len(str1) < len(str2) else len(str2)
+    var cmp = memcmp(str1._as_ptr(), str2._as_ptr(), min_len)
+    if cmp < 0:
+        return -1
+    elif cmp > 0:
+        return 1
+    # now check length for str1 and str2
+    if len(str1) == len(str2):
+        return 0
+    elif len(str1) > len(str2):
+        return 1
+    else:
+        return -1
+
+
 # ===----------------------------------------------------------------------===#
 # ord
 # ===----------------------------------------------------------------------===#
@@ -714,13 +731,8 @@ struct String(
         Returns:
             True if the Strings are equal and False otherwise.
         """
-        if len(self) != len(other):
-            return False
 
-        if int(self._as_ptr()) == int(other._as_ptr()):
-            return True
-
-        return memcmp(self._as_ptr(), other._as_ptr(), len(self)) == 0
+        return _str_compare(self, other) == 0
 
     @always_inline
     fn __ne__(self, other: String) -> Bool:
@@ -733,6 +745,54 @@ struct String(
             True if the Strings are not equal and False otherwise.
         """
         return not (self == other)
+
+    @always_inline
+    fn __gt__(self, other: String) -> Bool:
+        """Compares two Strings if self greater than the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are greater than other string.
+        """
+        return _str_compare(self, other) == 1
+
+    @always_inline
+    fn __ge__(self, other: String) -> Bool:
+        """Compares two Strings if self greater equal the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are greater equal the other string.
+        """
+        return _str_compare(self, other) == 0 or _str_compare(self, other) == 1
+
+    @always_inline
+    fn __lt__(self, other: String) -> Bool:
+        """Compares two Strings if self less than the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are less than other string.
+        """
+        return _str_compare(self, other) == -1
+
+    @always_inline
+    fn __le__(self, other: String) -> Bool:
+        """Compares two Strings if self less equal the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are less equal other string.
+        """
+        return _str_compare(self, other) == 0 or _str_compare(self, other) == -1
 
     @always_inline
     fn __add__(self, other: String) -> String:
--- a/stdlib/test/builtin/test_string.mojo
+++ b/stdlib/test/builtin/test_string.mojo
@@ -98,6 +98,20 @@ fn test_equality_operators() raises:
     assert_not_equal(s0, "notabc")
 
 
+fn test_compare_operators() raises:
+    assert_true(str("feefef") > str("feefe"))
+    assert_false(str("hello") > str("hello"))
+    assert_true(str("hello") >= str("hello"))
+    assert_false(str("hello") < str("hello"))
+    assert_true(str("hello") <= str("hello"))
+    assert_false(str("apple") > str("banana"))
+    assert_false(str("apple") > str("banana"))
+    assert_true(str("apple") > str("Banana"))
+    assert_false(str("apple123") > str("apple456"))
+    assert_false(str("appl2$") > str("banana"))
+    assert_false(str("") > str("a"))
+
+
 fn test_add() raises:
     var s1 = String("123")
     var s2 = String("abc")
@@ -697,6 +711,7 @@ def main():
     test_constructors()
     test_copy()
     test_equality_operators()
+    test_compare_operators()
     test_add()
     test_stringable()
     test_string_join()


stdlib/test/builtin/test_string.mojo
```diff
@@ -101,6 +101,20 @@ fn test_equality_operators() raises:
     assert_not_equal(s0, "notabc")
 
 
+fn test_compare_operators() raises:
+    assert_true("feefef".__str__().__gt__("feefe".__str__()))
```
gabrieldemarmiesse, line 105: You can use str() instead of `__str__()`, same for `>` instead of `__gt__`, etc...

stdlib/src/builtin/string.mojo
```diff
@@ -609,6 +609,64 @@ struct String(Sized, Stringable, IntableRaising, KeyElement, Boolable):
...
+    fn __gt__(self, other: String) -> Bool:
+        """Compares two Strings if self greater than the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are greater than other string.
+        """
+        if self == other:
```
gabrieldemarmiesse, line 622: At the worst case, this line is already one pass on every bytes in the string. It should be possible to get the result of `__gt__` with just a size comparison and a `memcmp`, which does, at the worst, one read on every byte

stdlib/src/builtin/string.mojo
```diff
@@ -609,6 +609,64 @@ struct String(Sized, Stringable, IntableRaising, KeyElement, Boolable):
...
+    fn __ge__(self, other: String) -> Bool:
+        """Compares two Strings if self greater equal the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are greater equal the other string.
+        """
+        return self == other or self > other
```
gabrieldemarmiesse, line 644: Here you read every byte twice, is there a way to read each byte once ?

gabrieldemarmiesse (commented):
What is usually done in this case is to make a private function that returns -1 (strictly lower), 0 (equal) or +1 (strictly greater) and call this function in each of the `__gt__`, `__eq__`, `__lt__`... methods. CPython implements a few comparaisons with this system and it's quite readable. This is only an idea, it can be implemented in other ways.

stdlib/src/builtin/string.mojo
```diff
@@ -609,6 +609,64 @@ struct String(Sized, Stringable, IntableRaising, KeyElement, Boolable):
...
+        if self == other:
+            return False
+        var _self = self._as_ptr()
+        var _other = other._as_ptr()
+        var _self_len = self.__len__()
+        var _other_len = other.__len__()
+        var len = _other_len if _self_len > _other_len else _self_len
+        var cmp = memcmp(_self, _other, len)
+        if cmp == 0:
+            return True if _self_len > _other_len else False
```
soraros, line 631: Why would you write ternary at all?

Moosems: As in what's the point of writing ternary statements?

stdlib/test/builtin/test_string.mojo
zhoujingya: 😊ok, thanks for your advice

stdlib/src/builtin/string.mojo
zhoujingya: ok, let me have a try

stdlib/src/builtin/string.mojo
zhoujingya: `if cmp == 0`, the longer string is greater? isn't it?

zhoujingya:
I tried to implement compare operation for `StringLiteral` too, but `StringLiteral` use self defined  `_memcmp` inside

stdlib/src/builtin/string.mojo
soraros: My point was `True if cond else False` is the same as `cond`.

stdlib/src/builtin/string.mojo
```diff
@@ -51,6 +51,25 @@ fn _ctlz(val: SIMD) -> __type_of(val):
     return llvm_intrinsic["llvm.ctlz", __type_of(val)](val, False)
 
 
+@always_inline
+fn unicode_compare[T: Stringable](str1: T, str2: T) -> Int:
```
mzaks, line 55: The name of this function should not mention unicode as it is a pure utf8 byte comparison. Unicode defines a very complex set of rules for comparison, which is called collation and is described in following [document](http://www.unicode.org/reports/tr10/tr10-41.html).

soraros, line 55: This is an internal function, maybe prefix the name with `_`. Also, I think we should restrict the input type to `String`.

zhoujingya: Thanks for your rigorous comments, I will modify this function name😊

stdlib/src/builtin/string.mojo
```diff
@@ -51,6 +51,23 @@ fn _ctlz(val: SIMD) -> __type_of(val):
     return llvm_intrinsic["llvm.ctlz", __type_of(val)](val, False)
 
 
+@always_inline
+fn _str_compare(str1: String, str2: String) -> Int:
+    var min_len = len(str1) if len(str1) < len(str2) else len(str2)
```
ConnorGray, line 56: Joe recently added `min` to a new `utils._math`, so we could import that in this file:

```mojo
from utils._math import min
```

and then use it here:

```suggestion
    var min_len = min(len(str1), len(str2))
```

ConnorGray (approved):
Thanks for your contribution @zhoujingya! I had only one minor piece of feedback (using `min`), but I'll address that when I import this PR internally.

Moosems (approved):
Connor Gray liked it and LGTM

stdlib/src/builtin/string.mojo
```diff
@@ -734,6 +746,54 @@ struct String(
...
+    fn __le__(self, other: String) -> Bool:
+        """Compares two Strings if self less equal the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are less equal other string.
+        """
+        return _str_compare(self, other) == 0 or _str_compare(self, other) == -1
```
laszlokindrat, line 795:
```suggestion
        return not self.__gt__(other)
```

stdlib/src/builtin/string.mojo
```diff
@@ -734,6 +746,54 @@ struct String(
...
+    fn __lt__(self, other: String) -> Bool:
+        """Compares two Strings if self less than the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are less than other string.
+        """
+        return _str_compare(self, other) == -1
```
laszlokindrat, line 783:
```suggestion
        return not self.__ge__(other)
```

stdlib/src/builtin/string.mojo
```diff
@@ -52,6 +52,23 @@ fn _ctlz(val: SIMD) -> __type_of(val):
     )
 
 
+@always_inline
+fn _str_compare(str1: String, str2: String) -> Int:
```
laszlokindrat, line 56: Can you please provide a summary of what the logic is in the docstring? Also, please document that this can return {-1, 0, 1}.

stdlib/src/builtin/string.mojo
```diff
@@ -734,6 +746,54 @@ struct String(
...
+    fn __ge__(self, other: String) -> Bool:
+        """Compares two Strings if self greater equal the other.
+
+        Args:
+            other: The rhs of the operation.
+
+        Returns:
+            True if self are greater equal the other string.
+        """
+        return _str_compare(self, other) == 0 or _str_compare(self, other) == 1
```
laszlokindrat, line 771: Please ensure `_str_compare` is called only once.
```suggestion
        return _str_compare(self, other) != -1
```

stdlib/test/builtin/test_string.mojo
```diff
@@ -98,6 +98,20 @@ fn test_equality_operators() raises:
...
+    assert_false(str("hello") > str("hello"))
+    assert_true(str("hello") >= str("hello"))
+    assert_false(str("hello") < str("hello"))
+    assert_true(str("hello") <= str("hello"))
+    assert_false(str("apple") > str("banana"))
+    assert_false(str("apple") > str("banana"))
+    assert_true(str("apple") > str("Banana"))
+    assert_false(str("apple123") > str("apple456"))
+    assert_false(str("appl2$") > str("banana"))
+    assert_false(str("") > str("a"))
```
laszlokindrat, line 112: Can you please modify these tests so that they use the dunder methods directly (as opposed to the operators).

laszlokindrat (commented):
Thanks for the patch! Any chance you could implement this in `StringRef`, and delegate from here (and from `StringLiteral`) to there as well?

laszlokindrat:
@zhoujingya Thanks for the patch! Are you still interested in pushing this forward?

laszlokindrat:
Closing this in favor of https://github.com/modularml/mojo/pull/2409.

PR #2384, opened 2024-04-23
[stdlib] Parameterize __getitem__ and __setitem__ in stdlib types

### LINKED ISSUE #2337: [stdlib] Use `Intable` in `getitem` and `setitem` consistently
Many `__getitem__` and `__setitem__` functions still take an `Int` as an argument.  This is OK, but forces call sites to explicitly construct an `Int` via `int(my_value)` for example to call `__getitem__` or `__setitem__`.  Instead, we can just make `__getitem__` and `__setitem__` take an `Intable` thing that always does the `int(value)` internally rather than pushing it onto all the callers.

One thing to note is for floating point indices, like `x[0.1]`, this would erroneously call `x[int(0.1)]` i.e. `x[0]`.  This is surprising and instead should be a hard error.  But we can put that on hold and stage this in phases:

1. Change all the `__getitem__` and `__setitem__` functions to work on `Intable` and ignore the floating point indexing issue mentioned above.
2. Introduce an `Integral` trait or something to ensure that callers can't pass a floating point value to these functions.  Use that instead of `Intable` everywhere.
3. Audit other functions using `Intable` in case they also have floating point indexing issues like mentioned above.

FYI @itramble

bgreni:
Fixes #2337

As mentioned in the referenced issue, the `__index__()` method allows types other than the base `Int` to be used when indexing containers without allowing inappropriate types that happen to implement `Intable` from being used (such as float point scalars).

Introduce the `Indexer` trait that requires the type to implement the `__index__()` method, and change all instances of `__getitem__` and `__setitem__` to accept any type that has the `Indexer` trait.

### DIFF
--- a/stdlib/src/builtin/bool.mojo
+++ b/stdlib/src/builtin/bool.mojo
@@ -59,7 +59,12 @@ trait Boolable:
 @value
 @register_passable("trivial")
 struct Bool(
-    Stringable, CollectionElement, Boolable, EqualityComparable, Intable
+    Stringable,
+    CollectionElement,
+    Boolable,
+    EqualityComparable,
+    Intable,
+    Indexer,
 ):
     """The primitive Bool scalar value used in Mojo."""
 
@@ -323,6 +328,15 @@ struct Bool(
             0 for -False and -1 for -True.
         """
         return __mlir_op.`index.casts`[_type = __mlir_type.index](self.value)
+
+    @always_inline("nodebug")
+    fn __index__(self) -> Int:
+        """Convert this Bool to an integer for indexing purposes
+
+        Returns:
+            Bool as Int
+        """
+        return self.__int__()
 
 
 # ===----------------------------------------------------------------------=== #
--- a/stdlib/src/builtin/builtin_list.mojo
+++ b/stdlib/src/builtin/builtin_list.mojo
@@ -146,16 +146,19 @@ struct VariadicList[type: AnyRegType](Sized):
         return __mlir_op.`pop.variadic.size`(self.value)
 
     @always_inline
-    fn __getitem__(self, index: Int) -> type:
+    fn __getitem__[indexer: Indexer](self, idx: indexer) -> type:
         """Gets a single element on the variadic list.
 
+        Parameters:
+            indexer: The type of the indexing value.
+
         Args:
-            index: The index of the element to access on the list.
+            idx: The index of the element to access on the list.
 
         Returns:
             The element on the list corresponding to the given index.
         """
-        return __mlir_op.`pop.variadic.get`(self.value, index.value)
+        return __mlir_op.`pop.variadic.get`(self.value, index(idx).value)
 
     @always_inline
     fn __iter__(self) -> Self.IterType:
@@ -358,18 +361,21 @@ struct VariadicListMem[
     # TODO: Fix for loops + _VariadicListIter to support a __nextref__ protocol
     # allowing us to get rid of this and make foreach iteration clean.
     @always_inline
-    fn __getitem__(self, index: Int) -> Self.reference_type:
+    fn __getitem__[indexer: Indexer](self, idx: indexer) -> Self.reference_type:
         """Gets a single element on the variadic list.
 
+        Parameters:
+            indexer: The type of the indexing value.
+
         Args:
-            index: The index of the element to access on the list.
+            idx: The index of the element to access on the list.
 
         Returns:
             A low-level pointer to the element on the list corresponding to the
             given index.
         """
         return Self.reference_type(
-            __mlir_op.`pop.variadic.get`(self.value, index.value)
+            __mlir_op.`pop.variadic.get`(self.value, index(idx).value)
         )
 
     @always_inline
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -139,6 +139,7 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
     Sized,
     Stringable,
     Truncable,
+    Indexer,
 ):
     """Represents a small vector that is backed by a hardware vector element.
 
@@ -482,6 +483,22 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
         return __mlir_op.`pop.cast`[_type = __mlir_type.`!pop.scalar<index>`](
             rebind[Scalar[type]](self).value
         )
+
+    @always_inline("nodebug")
+    fn __index__(self) -> Int:
+        """Returns the value as an int if it is an integral value
+
+        Contraints:
+            Must be an integral value
+
+        Returns:
+            The value as an integer
+        """
+        constrained[
+            type.is_integral() or type.is_bool(),
+            "expected integral or bool type",
+        ]()
+        return self.__int__()
 
     @always_inline
     fn __str__(self) -> String:
@@ -1731,8 +1748,11 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
     # ===-------------------------------------------------------------------===#
 
     @always_inline("nodebug")
-    fn __getitem__(self, idx: Int) -> Scalar[type]:
+    fn __getitem__[indexer: Indexer](self, idx: indexer) -> Scalar[type]:
         """Gets an element from the vector.
+
+        Parameters:
+            indexer: The type of the indexing value.
 
         Args:
             idx: The element index.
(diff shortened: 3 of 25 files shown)


stdlib/src/builtin/bool.mojo
```diff
@@ -261,6 +266,15 @@ struct Bool(
             )
         )
 
+    @always_inline("nodebug")
+    fn __index__(self) -> Int:
```
bgreni, line 270: I'm not sure if we actually want this, but I could see it being useful to avoid branching in some cases?

stdlib/src/builtin/simd.mojo
```diff
@@ -513,6 +508,19 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+    fn __index__(self) -> Int:
+        """Returns the value as an int if it is an integral value
+
+        Contraints:
+            Must be an integral value
+
+        Returns:
+            The value as an integer
+        """
+        constrained[type.is_integral(), "expected integral type"]()
```
gabrieldemarmiesse, line 521: NOICE

soraros: This may not be a good idea, for `SIMD` at least. This way, `SIMD[bool, 1]` will be cast to a `Int` but we might want to index with the original mask (which is size 1).

bgreni: > This may not be a good idea, for `SIMD` at least. This way, `SIMD[bool, 1]` will be cast to a `Int` but we might want to index with the original mask (which is size 1).

Now that you mention it `Scalar[bool]` actually won't be accepted here, which is inconsistent as I've currently given `Bool` the `Indexer` trait, but I was hoping for feedback from the team on whether that was acceptable or not.

We need some support for this in SIMD, however, since it would be expected that one could use any of the integral scalars to index the stdlib containers.

soraros: I just think we need a more complete design before adding the support to stdlib.

stdlib/src/builtin/value.mojo
```diff
@@ -145,3 +145,21 @@ trait RepresentableCollectionElement(CollectionElement, Representable):
     """
 
     pass
+
+
+trait Indexer:
```
gabrieldemarmiesse, line 150: In Python dunder methods are not supposed to be called directly. Instead, we use a method that will call the dunder method for us (and possibly do extra logic).
For example `int()` calls `__int__`, `*` calls `__mul__`, `.` calls `__getattr__`, etc...

In the case of `__index__`, it's `operator.index()`. Since we don't have an `operator` module yet, maybe let's add the `index` function in the builtins for now? I don't think we will want to convert all the `x.__index__()` to `index(x)` later on, let's do it now in this PR.

bgreni: Oh silly me, I originally didn't do that since I saw we had a bunch of free functions already in `utils/index.mojo` named index, but those start with a capital letter so there's no risk of it clashing. I'll fix this!

stdlib/src/builtin/bool.mojo
```diff
@@ -322,6 +327,15 @@ struct Bool(
         """
         return __mlir_op.`index.casts`[_type = __mlir_type.index](self.value)
 
+    @always_inline("nodebug")
```
laszlokindrat, line 332: Please add a test for this (we test these implementations directly using explicit calls to `__index__` instead of through the free functions).

stdlib/src/builtin/builtin_list.mojo
```diff
@@ -138,16 +138,19 @@ struct VariadicList[type: AnyRegType](Sized):
         return __mlir_op.`pop.variadic.size`(self.value)
 
     @always_inline
-    fn __getitem__(self, index: Int) -> type:
+    fn __getitem__[indexer: Indexer](self, idx: indexer) -> type:
```
laszlokindrat, line 149: style: here and elsewhere: we try to make type parameters PascalCase, could you please change these?
```suggestion
    fn __getitem__[IndexerType: Indexer](self, idx: IndexerType) -> type:
```
If you want, I'm also okay with calling these just `T` for brevity.

stdlib/src/builtin/bool.mojo
```diff
@@ -322,6 +327,15 @@ struct Bool(
...
+    fn __index__(self) -> Int:
+        """Convert this Bool to an integer for indexing purposes
```
laszlokindrat, line 334: nit
```suggestion
        """Convert this Bool to an integer for indexing purposes.
```

stdlib/src/builtin/bool.mojo
```diff
@@ -322,6 +327,15 @@ struct Bool(
...
+
+        Returns:
+            Bool as Int
```
laszlokindrat, line 337: Please provide a more details description here, e.g. "`1` if `True` and `0` otherwise."

stdlib/src/builtin/simd.mojo
```diff
@@ -472,6 +473,22 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
             rebind[Scalar[type]](self).value
         )
 
+    @always_inline("nodebug")
+    fn __index__(self) -> Int:
+        """Returns the value as an int if it is an integral value
```
laszlokindrat, line 489: nit here and below: missing periods
```suggestion
        """Returns the value as an int if it is an integral value.
```

stdlib/src/builtin/simd.mojo
```diff
@@ -472,6 +473,22 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+
+        Contraints:
+            Must be an integral value
```
laszlokindrat, line 492: It also needs to be a scalar, could you please add that here?

stdlib/src/builtin/value.mojo
```diff
@@ -179,3 +179,37 @@ trait RepresentableCollectionElement(CollectionElement, Representable):
     """
 
     pass
+
```
laszlokindrat: Could you please move these to `int.mojo`?

stdlib/src/builtin/value.mojo
```diff
@@ -179,3 +179,37 @@ trait RepresentableCollectionElement(CollectionElement, Representable):
...
+
+trait Indexer:
+    """This trait denotes a type that can be used to index a container that
+    handles integral index values.
+
+    This solves the issue of being able to index data structures such as `List` with the various
+    integral types without being too broad and allowing types that should not be used such as float point
+    values.
```
laszlokindrat, line 214: style nit: please wrap lines to 80 char

stdlib/src/builtin/value.mojo
```diff
@@ -179,3 +179,37 @@ trait RepresentableCollectionElement(CollectionElement, Representable):
...
+    """
+
+    fn __index__(self) -> Int:
+        """Return the index value
+
+        Returns:
+            The index value of the object
+        """
```
laszlokindrat, line 222: @JoeLoser Do we have styleguide recommendations on documentating method declarations in traits? And do they even generate docs correctly?

laszlokindrat (commented):
This is a great improvement, thank you! Internally, we have been discussing related issues, so your timing to do this is perfect. I have a bunch of mostly nit comments, and one big ask: could you please split this into two patches: one that introduces the `Indexer` trait and makes the types conform to it (along with tests for their `__index__` methods if they don't have them yet), and a second one that changes `__getitem__` et al to take `Indexer` instead of `Int`?

laszlokindrat:
Also, please add a changelog entry!

bgreni:
> This is a great improvement, thank you! Internally, we have been discussing related issues, so your timing to do this is perfect. I have a bunch of mostly nit comments, and one big ask: could you please split this into two patches: one that introduces the `Indexer` trait and makes the types conform to it (along with tests for their `__index__` methods if they don't have them yet), and a second one that changes `__getitem__` et al to take `Indexer` instead of `Int`?

@laszlokindrat Sounds good thanks for the feedback. Do you mean split this into two PRs, or simply two commits within this PR?

gabrieldemarmiesse:
@bgreni  I fused part of this PR with the one from @StandinKP ( https://github.com/modularml/mojo/pull/2386 ) as they are very much related. Maybe let's merge https://github.com/modularml/mojo/pull/2677 first and then both of you can rebase your PRs on nightly afterwards?

laszlokindrat:
> @laszlokindrat Sounds good thanks for the feedback. Do you mean split this into two PRs, or simply two commits within this PR?

Sorry, I should have been clearer: two separate PRs, please (PRs get squashed before merging internally)! Thanks!

laszlokindrat:
> @bgreni I fused part of this PR with the one from @StandinKP ( #2386 ) as they are very much related. Maybe let's merge #2677 first and then both of you can rebase your PRs on nightly afterwards?

I would like to wait for the split of this to happen. Then you can rebase `#2677` once `Indexer` is delivered separately.

gabrieldemarmiesse:
> I would like to wait for the split of this to happen first. Then you can rebase #2677 once `Indexer` is delivered separately.

Sounds good!

bgreni:
> > @laszlokindrat Sounds good thanks for the feedback. Do you mean split this into two PRs, or simply two commits within this PR?
> 
> Sorry, I should have been clearer: two separate PRs, please (PRs get squashed before merging internally)! Thanks!

Ok thanks, I'll do that

bgreni:
@laszlokindrat The PR for the introduction of the trait has been created #2685

bgreni:
@laszlokindrat #2722 has been created for the second part of applying `Indexer` to `__getitem__` `__setitem__` and `__refitem__`

laszlokindrat:
> @laszlokindrat #2722 has been created for the second part of applying `Indexer` to `__getitem__` `__setitem__` and `__refitem__`

@bgreni cool, thanks! Does that mean that we can close this PR?

bgreni:
> > @laszlokindrat #2722 has been created for the second part of applying `Indexer` to `__getitem__` `__setitem__` and `__refitem__`
> 
> @bgreni cool, thanks! Does that mean that we can close this PR?

Yes definitely, I will close it now

PR #2386, opened 2024-04-23
[mojo-stdlib] Added fn to normalize index calculation

### LINKED ISSUE #2251: [stdlib] Consistency in `List` struct for "normalized index calculation"
### Review Mojo's priorities

### What is your request?

Currently, [`List`](https://github.com/modularml/mojo/blob/nightly/stdlib/src/collections/list.mojo) type contains multiple implementation for "normalized index calculation".  Refactor this into a private function and re-use it everywhere in this struct. Idea for this issue: https://github.com/modularml/mojo/pull/2148#discussion_r1554624275.

This can be observed by searching `normalized_idx` in [`stdlib/src/collections/list.mojo`](https://github.com/modularml/mojo/blob/nightly/stdlib/src/collections/list.mojo)

 Consistency in `List` struct for "normalized index calculation".

### Any other details?

_No response_

StandinKP:
Fixes #2251


stdlib/src/collections/list.mojo
```diff
@@ -610,3 +600,9 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
                 result += ", "
         result += "]"
         return result
+
+    @always_inline
+    fn _normalize_index(self, index: Int) -> Int:
+        if index < 0:
+            return _max(0, len(self) + index)
```
soraros, line 607: Why do we need the `_max` here?

StandinKP: It would be here just to prevent the negative indexes. How would you suggest to do it?

soraros: Given that index outside `[-len(self), len(self))` is out of bound, we can simply skip the `_max` call.
```mojo
return index if index >= 0 else index + len(self)
```

StandinKP: @JoeLoser @gabrieldemarmiesse can I get your views on this ?

stdlib/src/collections/list.mojo
```diff
@@ -666,3 +656,9 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
                 result += ", "
         result += "]"
         return result
+
+    @always_inline
+    fn _normalize_index(self, index: Int) -> Int:
+        if index < 0:
```
abduld, line 662: Suggestion: invert the condition to reduce indentation. So 

```suggestion
        if index >= 0:
           return index 
        return complicated expr 
```

abduld (approved):
Thanks

stdlib/src/collections/list.mojo
StandinKP: The indentation would remain the same @abduld after your suggestion

stdlib/src/collections/list.mojo
JoeLoser: Hey @StandinKP — sorry for the slow reply here.  We discussed this the other day during our team meeting and here's my notes:

- The current code seems wrong because it will silently turn an out of bounds negative value into an index of the 0th element by using `_max`.  But  I don’t think the solution is to remove the `_max` and allow negative indices to segfault (or worse) at runtime.  Instead, let's `abort` on out-of-bound indices. 
- In the future, it would be nice to have one "normalized index" function for use in the library for any contiguous data structure we have in the library such as `List`, `String`, etc.  Nothing to do for this PR, but just pointing out how we'd like to clean this up over time.
- In the short (and long-term), I think aborting on out-of-bounds access is a reasonable approach, both here in `List` and elsewhere.

StandinKP: Sure. Makes sense. Do you want me to keep an issue filed to add the common "normalized function" and update all references?

StandinKP:
@JoeLoser @soraros I am facing this issue right now after rebasing with nightly.
```
❯ ./stdlib/scripts/run-tests.sh
Included from /Users/kaushalphulgirkar/Documents/Projects/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/kaushalphulgirkar/Documents/Projects/mojo/stdlib/src/collections/list.mojo:573:39: error: invalid initialization: could not deduce parameter 'type' of parent struct 'Reference'
        var normalized_idx = Reference(self)._normalize_index(i)
                             ~~~~~~~~~^~~~~~
/Users/kaushalphulgirkar/Documents/Projects/mojo/stdlib/src/__init__.mojo:1:1: note:  struct declared here
^
Included from /Users/kaushalphulgirkar/Documents/Projects/mojo/stdlib/src/memory/__init__.mojo:1:
/Users/kaushalphulgirkar/Documents/Projects/mojo/stdlib/src/memory/reference.mojo:222:8: note: function declared here
    fn __init__(inout self, value: Self._mlir_type):
       ^
```

Not able to figure how to exactly fix this. Any suggestions? Did we change the Reference initializations?

gabrieldemarmiesse:
Hi @StandinKP, indeed you are right :) there was a change with how Reference works right now, see https://github.com/modularml/mojo/commit/0fe4cb7c9610d0c9623f5b0becd2d9af94835b7c

I think that in your case, you don't need to do `Reference(self)` since `self` is already a reference. I hope it helps!

StandinKP:
Makes sense. Thanks @gabrieldemarmiesse

StandinKP:
@gabrieldemarmiesse if I use self._normalize_index it gives out `'Reference[List[T], is_mutable, lifetime, 0]' value has no attribute '_normalize_index'`
trying to figure out this but if you have any idea let me know Thanks

StandinKP:
@soraros can you provide some guidance here? Thanks

gabrieldemarmiesse:
@StandinKP I think we should merge https://github.com/modularml/mojo/pull/2677 first and then you can open one pull request per method to change. This will ensure the easy changes will be merged to nightly quickly and the harder changes (I know there are some out of bounds bugs in the tests that you will hit) can be merged afterwards. What do you think?

JoeLoser:
Closing as this is now implemented at https://github.com/modularml/mojo/blob/3cfdd29453d1bddd6d6dbde5fd13c3cf64eff825/stdlib/src/collections/_index_normalization.mojo#L48.  Thanks for pushing on this @StandinKP and @gabrieldemarmiesse!

PR #2395, opened 2024-04-24
[stdlib] Move `unsafe_pointer.mojo` off `LegacyPointer`.

soraros:


stdlib/src/memory/unsafe.mojo
```diff
@@ -221,6 +221,9 @@ struct LegacyPointer[
         )
         return Self {address: address}
 
+    fn __init__(p: UnsafePointer[type, address_space]) -> Self:
+        return Self(p.value)
+
```
ConnorGray, line 226: Unfortunately it is not safe in general to construct a `LegacyPointer` from an `UnsafePointer` because `LegacyPointer` is parameterized on a `type: AnyRegType`, which will effectively type-erase important information about the pointee type.

We also don't generally want to make it easy to go back from an `UnsafePointer` to a `LegacyPointer`.

I may have missed something, but I think this new initializer is not being used in this PR? If so, I suggest we remove it.

Otherwise I'm happy to get a better understanding of what the intended use-cases are 🙂

ConnorGray:
Thanks for tackling this @soraros! 🙂

It looks like may need to be rebased to fix a merge conflict, and also update the compiler version as we just released a new Mojo nightly.

I left a couple of initial comments, but then I noticed this is still in "Draft" mode so I'll hold off until you mark this ready for review😌

stdlib/src/memory/memory.mojo
```diff
@@ -413,9 +413,14 @@ fn stack_allocation[
...
-        count, Scalar[type], alignment=alignment, address_space=address_space
-    ]()
+    return DTypePointer(
+        stack_allocation[
+            count,
+            Scalar[type],
+            alignment=alignment,
+            address_space=address_space,
+        ]()
+    )
```
soraros, line 423: `DTypePointer` does have a constructor from `UnsafePointer`, but LSP complains.

stdlib/src/memory/unsafe.mojo
soraros: I added this in case the changed return type of `_malloc` becomes a problem. Perhaps there is nothing to worry about.

stdlib/src/memory/unsafe.mojo
```diff
@@ -800,9 +804,10 @@ struct DTypePointer[
         Returns:
             A new `DTypePointer` object which has been allocated on the heap.
         """
-        return _malloc[Self.element_type, address_space=address_space](
+        var p = _malloc[Self.element_type, address_space=address_space](
             count * sizeof[type](), alignment=alignment
         )
+        return Self(p)
```
soraros, line 810: This also doesn't work for some reason.

stdlib/src/memory/unsafe.mojo
ConnorGray: I think so, lets hold off. 

If the changed return type of `_malloc` does cause any problems I think those would be better handled on a case-by-case basis (and it would be good to get a compiler error so we know where that happens).

Thanks @soraros!

stdlib/src/memory/unsafe.mojo
soraros: @ConnorGray This should be a compiler bug, right?

soraros:
The error message doesn't make much sense.

laszlokindrat:
I cannot see any uses of `LegacyPointer` in `unsafe_pointer.mojo` anymore. Can we close this PR?

PR #2397, opened 2024-04-24
[SDLC] Mojo OSS Copybara baseline (#38479)

patrickdoc:
This change fixes a stdlib test to use `mojo` instead of the `mojo-no-debug` alias.

It also serves as the first Copybara commit to be copied to the `mojo` repo. In the `mojo` repo, it will also contain any deltas between the current state of the monorepo and `mojo`.

MODULAR_ORIG_COMMIT_REV_ID: 7656ee47e77d77f18fece65567e54fe8ab1c5d0e


ConnorGray (approved):
Amazing to have this working! :tada:

PR #2406, opened 2024-04-25
[mojo-stdlib] Add `List.count(value)` method

### LINKED ISSUE #2263: [stdlib] Add `count(List, value)` method
Similar to Python's `list.count(x)` method, we should have the same for our `List`.  It can't be a member method yet due to lack of conditional conformance in requiring the `__eq__` trait (it's not present in `CollectionElement` today).

As I suggest in https://github.com/modularml/mojo/pull/2190#issuecomment-2062101102, we could add this as a free function since we don't have conditional conformance e.g.

```
trait ComparableCollectionElement(EqualityComparable, CollectionElement):
    pass

def count[T: ComparableCollectionElement](arg: List[T], value: T):
   ...
```

in `list.mojo`.

StandinKP:
Add `List.count(value)` method to count the number of occurrences of a value in the `List`.

Fixes #2263

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -669,6 +669,16 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
     @staticmethod
     fn count[T: ComparableCollectionElement](self: List[T], value: T) -> Int:
         """Counts the number of occurrences of a value in the list.
+        Note that since we can't condition methods on a trait yet,
+        the way to call this method is a bit special. Here is an example below.
+
+        ```mojo
+        var my_list = List[Int](1, 2, 3)
+        print(__type_of(my_list).count(my_list, 1))
+        ```
+
+        When the compiler supports conditional methods, then a simple `my_list.count(1)` will
+        be enough.
 
         Args:
             self: The list to search.
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -20,10 +20,6 @@ from collections import List
 """
 
 
-from builtin.value import (
-    StringableCollectionElement,
-    ComparableCollectionElement,
-)
 from memory import UnsafePointer, Reference
 from memory.unsafe_pointer import move_pointee, move_from_pointee
 from .optional import Optional
@@ -670,19 +666,19 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         result += "]"
         return result
 
+    @staticmethod
+    fn count[T: ComparableCollectionElement](self: List[T], value: T) -> Int:
+        """Counts the number of occurrences of a value in the list.
 
-fn count[T: ComparableCollectionElement](list: List[T], value: T) -> Int:
-    """Counts the number of occurrences of a value in the list.
+        Args:
+            self: The list to search.
+            value: The value to count.
 
-    Args:
-        list: The list to search.
-        value: The value to count.
-
-    Returns:
-        The number of occurrences of the value in the list.
-    """
-    var count = 0
-    for i in range(len(list)):
-        if list[i] == value:
-            count += 1
-    return count
+        Returns:
+            The number of occurrences of the value in the list.
+        """
+        var count = 0
+        for i in range(len(self)):
+            if self[i] == value:
+                count += 1
+        return count
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -20,7 +20,10 @@ from collections import List
 """
 
 
-from builtin.value import StringableCollectionElement
+from builtin.value import (
+    StringableCollectionElement,
+    ComparableCollectionElement,
+)
 from memory import UnsafePointer, Reference
 from memory.unsafe_pointer import move_pointee, move_from_pointee
 from .optional import Optional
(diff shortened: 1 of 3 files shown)


stdlib/test/collections/test_list.mojo
```diff
@@ -589,6 +589,12 @@ def test_converting_list_to_string():
     assert_equal(__type_of(my_list3).__str__(my_list3), "[1.0, 2.0, 3.0]")
 
 
+def test_list_count():
```
JoeLoser, line 592: **Suggestion** I recommend adding a few more test cases:
- `count()` on an empty list
- `count()` when there are no matches to ensure it returns `0`

stdlib/src/collections/list.mojo
```diff
@@ -20,11 +20,13 @@ from collections import List
 """
 
 
-from builtin.value import StringableCollectionElement
+from builtin.value import (
```
JoeLoser, line 23: **Suggestion** You can remove this entirely.  For public entities in the `builtin` module, you don't need to import anything.  They will automatically be imported into every Mojo program.

stdlib/src/collections/list.mojo
```diff
@@ -604,3 +606,20 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
                 result += ", "
         result += "]"
         return result
+
+
+fn count[T: ComparableCollectionElement](list: List[T], value: T) -> Int:
```
JoeLoser, line 611: **Suggestion** Make this a static method like we do for `__str__` [here](https://github.com/modularml/mojo/blob/d344d178aec2b80c5772261b8a6e441202a752de/stdlib/src/collections/list.mojo#L567).  This will make it easier to swap (no syntax change required on call sites of your new function) when we push `EqualityComparable` into `CollectionElement` and then can make the method non-static.  What do you think?  @gabrieldemarmiesse may able to chime in more here as he explored the different options for things like this that require trait refinement (such as your `ComparableCollectionElement`).

gabrieldemarmiesse: The trait is fine, I'm in favor of the static method indeed as it will be easier to switch later when we have the compiler features we need. In your case the signature would look like this:

```mojo
struct List[...](...):
    ...
    @staticmethod
    fn count[U: ComparableCollectionElement](self: List[U], value: U) -> Int:
        ....


# then call it with
var my_list = List[Int](1, 2, 3, 2)
print(__type_of(my_list).count(my_list, 2))
```

stdlib/test/collections/test_list.mojo
StandinKP: Updated

stdlib/src/collections/list.mojo
StandinKP: I have pushed changes for the same

stdlib/src/collections/list.mojo
```diff
@@ -604,3 +602,20 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
                 result += ", "
         result += "]"
         return result
+
+    @staticmethod
+    fn count[T: ComparableCollectionElement](self: List[T], value: T) -> Int:
+        """Counts the number of occurrences of a value in the list.
```
gabrieldemarmiesse, line 608: You should add a small example explaining how to use this method in the docstring. It might not be intuitive for users to use `__type_of(...)`.

stdlib/test/collections/test_list.mojo
StandinKP: this test case is covered below on line 596 @gabrieldemarmiesse

gabrieldemarmiesse: oupsy, my bad. Ignore my comment then

stdlib/src/collections/list.mojo
StandinKP: @gabrieldemarmiesse should the note be similar to the __str__ method?

```Note that since we can't condition methods on a trait yet,
the way to call this method is a bit special. Here is an example below:

`mojo
var my_list = List[Int](1, 2, 3)
print(__type_of(my_list).count(my_list, 1))
`

When the compiler supports conditional methods, then a simple `my_list.count(1)` will
be enough.```

Comment (gabrieldemarmiesse) [reply stdlib/src/collections/list.mojo:608]: yes that looks good :)

Comment (StandinKP) [reply stdlib/src/collections/list.mojo:608]: Pushed the changes

**Comment by StandinKP** [comment]
@JoeLoser are we having any ticket to track adding the conditional conformance? Is that something we can work on or is it going to be internal development?

**Comment by gabrieldemarmiesse** [comment]
@StandinKP it's something that must be done internally. There is an issue here: https://github.com/modularml/mojo/issues/1876 to follow the progress if you want. It seems that it's not going to be available in the short term

**Review by JoeLoser — APPROVED** [review]

File: stdlib/src/builtin/value.mojo [hunk 5]
```diff
@@ -121,3 +121,15 @@ trait StringableCollectionElement(CollectionElement, Stringable):
     """
 
     pass
+
+
+trait ComparableCollectionElement(CollectionElement, EqualityComparable):
```
JoeLoser, line 126: **Suggestion** Do you mind rebasing with the latest `nightly` from today's release since this trait exists now at https://github.com/modularml/mojo/blob/ee5218693b57959d44e9b82dbc0774fb4d28db18/stdlib/src/builtin/value.mojo#L126?

StandinKP: I rebased but now it has added too many files🥲

StandinKP: Done

JoeLoser:
This landed internally and got pushed out already - see https://github.com/modularml/mojo/commit/c37a3ec3d26f1cb2ddafe9149e5504a291ab02c6.  Closing this PR. 

Thanks again for the contribution!! 🎉

PR #2409, opened 2024-04-25
[stdlib] String comparisons implemented

siitron:
For issue #2346 (as an alternative to #2378). All four comparisons (`__lt__`, `__le__`, `__gt__`, & `__ge__`) uses a single `__lt__` comparison (instead of checking less/greater than + potentially another "equals to"-check, for `__le__` & `__ge__`). Sorry if this is considered a duplicate PR, I only meant to give an alternative suggestion. This is my first ever PR on GitHub.

StringLiterals also get comparisons.


VMois:
Are there any reasons to close a PR? This looks like quite a nice implementation.

siitron:
@VMois Thank you! I closed this PR before because I couldn't pass the tests (because I didn't know that the math package wasn't supported to use because of cyclic dependencies). But I replaced `min` with if/else-statements and that nicely resulted in fewer comparisons.

I can't complete `./stdlib/scripts/run-tests.sh` locally due to some issues I have with WSL, but the current version should be fine now.

stdlib/test/builtin/test_string.mojo
```diff
@@ -98,6 +98,59 @@ fn test_equality_operators() raises:
...
+    var s2 = String("abc")
+    var s3 = String("ABC")
+    var s4 = String("ab")
+    var s5 = String("abcd")
+    var empty = String("")  # Explicit empty string for comparison
+    var default_string = String()  # Default constructed, assumed empty
+
+    # Test less than and greater than
+    assert_true(s0 < s1)  # "abc" < "def"
+    assert_false(s1 < s0)  # "def" is not less than "abc"
```
gabrieldemarmiesse, line 113: This PR is really cool! It might be possible that in the future, as devs add or modify tests, the comments get out of sync with the actual values of the strings. If you're worried about making long lines, what about this syntax?

```suggestion
    assert_false(str("def") < str("abc"))
```
Looking forward to your opinion on this :)

siitron: I really like this suggestion, and I fully agree with what you're saying! My tests were pretty sloppily written so I'm gonna go fix this now :)

VMois: I think tests always look a bit sloppy. But I do like a suggestion. Keep the tests as simple as possible.

stdlib/src/builtin/string_literal.mojo
```diff
@@ -129,6 +129,60 @@ struct StringLiteral(
...
+        Args:
+            rhs: The other StringLiteral to compare against.
+
+        Returns:
+            True if this StringLiteral is strictly less than the RHS StringLiteral and False otherwise.
+        """
+        var len1 = len(self)
+        var len2 = len(rhs)
+
+        if len1 < len2:
```
abduld, line 145: Suggestion: you can consolidate the duplication between this and the implementation in String by implementing the comparison on StringRef

siitron: Do you mean something like `return StringRef(self) < StringRef(rhs)` as `__lt__` for StringLiteral (and `return self._strref_dangerous() < rhs._strref_dangerous()` for String), etc.?

I didn't even consider StringRef (forgot it existed), and I tried to match `__lt__` with their respective `__eq__`-methods. Right now String, StringLiteral, and StringRef does `__eq__` in three separate ways, so I thought there was a reason for keeping them apart(??). Would it still be safe to use `memory.memcmp` in the StringRef `__lt__` implementation (considering it currently looks like StringLiteral has to use it's own local `_memcmp`, and StringRef's `__eq__` also avoids it), or should StringRef's `__lt__` be done like its `__eq__`?

What do you think of:
```mojo
    @always_inline("nodebug")
    fn __lt__(self, rhs: StringRef) -> Bool:
        """Compare this StringRef to the RHS using LT comparison.

        Args:
            rhs: The other StringRef to compare against.

        Returns:
            True if this StringRef is strictly less than the RHS StringRef and False otherwise.
        """
        var len1 = len(self)
        var len2 = len(rhs)

        if len1 < len2:
            return memcmp(self.data, rhs.data, len1) <= 0
        else:
            return memcmp(self.data, rhs.data, len2) < 0
```

laszlokindrat (approved):
This is great, thank you! Do you think you could do a followup where you move the core logic to `StringRef` and delegate to that from both `String` and `StringLiteral`?

siitron:
> This is great, thank you! Do you think you could do a followup where you move the core logic to `StringRef` and delegate to that from both `String` and `StringLiteral`?

Sure np, I'll create a new PR for the StringRef stuff!


Some thoughts (that I'll probably repeat in the PR):

@laszlokindrat Does this mean that StringLiteral no longer needs its private `_memcmp`-function (assuming `__eq__` and `__ne__` are also delegated to StringRef), i.e. StringLiteral (and StringRef) are ok to use `memory.memcmp` now? If yes, then I can make the 18 comparison methods all just use `__eq__` and `__lt__` in StringRef. I'm asking since StringLiteral's private `_memcmp` function has this comment: 
> \# Use a local memcmp rather than memory.memcpy to avoid #31139 and #25100. 

(I can't access "#31139 and #25100", I'm guessing its some internal stuff).

Also, for String to be delegated to StringRef I'd have to use String's `_strref_dangerous`-method which I was a bit hesitant to do

JoeLoser:

laszlokindrat:
> @laszlokindrat Does this mean that StringLiteral no longer needs its private `_memcmp`-function (assuming `__eq__` and `__ne__` are also delegated to StringRef), i.e. StringLiteral (and StringRef) are ok to use `memory.memcmp` now? If yes, then I can make the 18 comparison methods all just use `__eq__` and `__lt__` in StringRef. I'm asking since StringLiteral's private `_memcmp` function has this comment:
> 
> > # Use a local memcmp rather than memory.memcpy to avoid #31139 and #25100.
> 
> (I can't access "#31139 and #25100", I'm guessing its some internal stuff).

Good question. If you don't mind, would you just give it a go, and let's see what breaks? If things don't work, I will take a look at those tickets.
 
> Also, for String to be delegated to StringRef I'd have to use String's `_strref_dangerous`-method which I was a bit hesitant to do

I think this should be okay, since the lifetimes of self should be guaranteed to outlive method calls, so you can take a dangerous reference in these method implementations.

siitron:
> If you don't mind, would you just give it a go, and let's see what breaks? If things don't work, I will take a look at those tickets.

@laszlokindrat You can take a look at #2620 if you'd like.

PR #2375, opened 2024-04-22
[stdlib] Enhance Handling of Infinity and NaN in `assert_almost_equal`

leandrolcampos:
This PR enhances the `assert_almost_equal` function to correctly handle cases involving infinity and NaN.

According to `test_assert_almost_equal` added to `/test/testing/test_assertion.mojo`, the current implementation of `assert_almost_equal` results in errors in the following cases:

```mojo
    alias float_type = DType.float32
    alias _inf = inf[float_type]()
    alias _nan = nan[float_type]()
    ...
    _should_succeed(
        SIMD[float_type, 2](-_inf, _inf), SIMD[float_type, 2](-_inf, _inf)
    )
    ...
    _should_fail(
        SIMD[float_type, 2](-_inf, 0.0),
        SIMD[float_type, 2](_inf, 0.0),
        rtol=0.1,
    )
    _should_fail(
        SIMD[float_type, 2](_inf, 0.0),
        SIMD[float_type, 2](0.0, 0.0),
        rtol=0.1,
    )
    ...
    _should_fail(
        SIMD[float_type, 2](_nan, 0.0),
        SIMD[float_type, 2](0.0, 0.0),
        equal_nan=True,
    )
```

This PR also:
- Eliminates the use of `and` and `or` in the `_isclose` function due to the issue outlined in #2374.
- Explicitly reduces boolean vectors to boolean scalar values instead of counting on implicit conversions for clarity.
- Avoids arithmetic operations in `_isclose` and `assert_almost_equal` when the type is boolean, as these operations are not supported in this case.
- Clarifies the behavior of `assert_almost_equal` in the docstring, highlighting differences from similar functions such as `numpy.testing.assert_allclose`.
- Adds the `inf` function to `utils/_numerics` along with corresponding tests in `test/utils/test_numerics.mojo`.

PR #2412, opened 2024-04-26
[stdlib] SIMD conformance to EqualityComparable

helehex:
This allows `SIMD` to conform to `EqualityComparable`, without losing any of the original behavior.

It uses the 4th overload resolution rule to give the new methods lower precedence, while still conforming to `EqualityComparable`

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -146,6 +146,7 @@ struct SIMD[type: DType, size: Int](
     CeilDivable,
     CollectionElement,
     CollectionElementNew,
+    EqualityComparable,
     Floorable,
     Formattable,
     Hashable,
@@ -717,6 +718,26 @@ struct SIMD[type: DType, size: Int](
                 self.value, rhs.value
             )
 
+    # TODO there may be a better way to do this:
+    # the [overload_1: None = None] parameter is to give this overload lower precedence,
+    # while still conforming to EqualityComparable
+    @always_inline("nodebug")
+    fn __eq__[overload_1: None = None](self, rhs: Self) -> Bool:
+        """Compares two SIMD vectors using all-equal-to comparison.
+
+        This overload allows EqualityComparable conformance.
+
+        Parameters:
+            overload_1: Ignore this parameter, used to set the overload precedence.
+
+        Args:
+            rhs: The rhs of the operation.
+
+        Returns:
+            True if all lanes are equal, False otherwise.
+        """
+        return all(self == rhs)
+
     @always_inline("nodebug")
     fn __ne__(self, rhs: Self) -> Self._Mask:
         """Compares two SIMD vectors using not-equal comparison.
@@ -741,6 +762,26 @@ struct SIMD[type: DType, size: Int](
             return __mlir_op.`pop.cmp`[pred = __mlir_attr.`#pop<cmp_pred ne>`](
                 self.value, rhs.value
             )
+
+    # TODO there may be a better way to do this:
+    # the [overload_1: None = None] parameter is to give this overload lower precedence,
+    # while still conforming to EqualityComparable
+    @always_inline("nodebug")
+    fn __ne__[overload_1: None = None](self, rhs: Self) -> Bool:
+        """Compares two SIMD vectors using any-not-equal comparison.
+
+        This overload allows EqualityComparable conformance.
+
+        Parameters:
+            overload_1: Ignore this parameter, used to set the overload precedence.
+
+        Args:
+            rhs: The rhs of the operation.
+
+        Returns:
+            True if any lanes are not equal, False otherwise.
+        """
+        return any(self != rhs)
 
     @always_inline("nodebug")
     fn __gt__(self, rhs: Self) -> Self._Mask:
--- a/stdlib/test/builtin/test_math.mojo
+++ b/stdlib/test/builtin/test_math.mojo
@@ -81,16 +81,16 @@ def test_max():
 
 
 def test_round():
-    assert_equal(0.0, round(0.0))
-    assert_equal(1.0, round(1.0))
-    assert_equal(1.0, round(1.1))
-    assert_equal(1.0, round(1.4))
-    assert_equal(2.0, round(1.5))
-    assert_equal(2.0, round(2.0))
-    assert_equal(1.0, round(1.4, 0))
-    assert_equal(2.0, round(2.5))
-    assert_equal(1.5, round(1.5, 1))
-    assert_equal(1.61, round(1.613, 2))
+    assert_equal(round(0.0), 0)
+    assert_equal(round(1.0), 1)
+    assert_equal(round(1.1), 1)
+    assert_equal(round(1.4), 1)
+    assert_equal(round(1.5), 2)
+    assert_equal(round(2.0), 2)
+    assert_equal(round(1.4, 0), 1)
+    assert_equal(round(2.5), 2)
+    assert_equal(round(1.5, 1), 1.5)
+    assert_equal(round(1.613, 2), 1.61)
 
     var lhs = SIMD[DType.float32, 4](1.1, 1.5, 1.9, 2.0)
     var expected = SIMD[DType.float32, 4](1.0, 2.0, 2.0, 2.0)
--- a/stdlib/test/builtin/test_math.mojo
+++ b/stdlib/test/builtin/test_math.mojo
@@ -81,14 +81,14 @@ def test_max():
 
 
 def test_round():
-    assert_equal(0, round(0.0))
-    assert_equal(1, round(1.0))
-    assert_equal(1, round(1.1))
-    assert_equal(1, round(1.4))
-    assert_equal(2, round(1.5))
-    assert_equal(2, round(2.0))
-    assert_equal(1, round(1.4, 0))
-    assert_equal(2, round(2.5))
+    assert_equal(0.0, round(0.0))
+    assert_equal(1.0, round(1.0))
+    assert_equal(1.0, round(1.1))
+    assert_equal(1.0, round(1.4))
+    assert_equal(2.0, round(1.5))
+    assert_equal(2.0, round(2.0))
+    assert_equal(1.0, round(1.4, 0))
+    assert_equal(2.0, round(2.5))
     assert_equal(1.5, round(1.5, 1))
     assert_equal(1.61, round(1.613, 2))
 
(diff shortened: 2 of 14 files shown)


gabrieldemarmiesse (commented):
I wouldn't be against some more details (in the docs or something) since this seems a bit like black magic and using return types for overload resolution :)
Maybe the docstring would be a good place to explain how to actually use this overload?

helehex:
Yeah that might be good. I'm not even sure this is something that would get pulled because of that, but figured i'd give it a shot since it's just a few lines and gives expected behavior

JoeLoser (commented):
This is a pretty neat way to make `SIMD` conform to `EqualityComparable`, props to your cleverness!

This seems like something we should do in the short-term until we have conditional conformance and other things.  Accepting this PR would unlock a bunch of other future nice things.  To start with, we could then push this `EqualityComparable` requirement into `CollectionElement`, which would then mean we can now implement `__eq__ ` and ` __ne__` for collections.  Further, it would allow us to simplify things like `List.count` and `List.index` which currently use an ad-hoc composite trait to ensure the given type implements `__eq__` since that isn’t a requirement of `CollectionElement`.

I've added this PR to bring up with the team on Monday during our internal design discussions, and I'll circle back after we meet.  But this is super cool, thanks for pushing on this!

stdlib/test/builtin/test_simd.mojo
```diff
@@ -715,6 +715,41 @@ def test_mul_with_overflow():
...
+        return a != b
+
+    var s1 = SIMD[DType.int32, 2](1, 2)
+    var s2 = SIMD[DType.int32, 2](1, 6)
+    var s3 = SIMD[DType.int32, 2](6, 2)
+    var s4 = SIMD[DType.int32, 2](6, 6)
+
+    assert_equal(s1 == s1, SIMD[DType.bool, 2](True, True), "s1 == s1")
+    assert_equal(s1 != s1, SIMD[DType.bool, 2](False, False), "s1 != s1")
+    assert_equal(eq(s1, s1), True, "eq(s1, s1)")
```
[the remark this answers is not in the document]
helehex: The first should work, but i think the second one might cause issues.
however i should be able to do `assert_equal(s1, s1)` and `assert_not_equal(s1, s2)`

The downside of using this trick:
the behavior of `==` changes slightly, depending if you're in the context of `EqualityComparable`. so i was using the `eq` function to ensure that.

I'll try it out in a lil bit though

helehex:
sora pointed out that this would not be confluent under generalization, and we agreed that for now, `SIMD.__bool__()` should be constrained to size=1, so that if the logic of `bool(a != b)` changed out from under you, it at least gives a compiler error

helehex:
ok this feel like the right way to go
- Added some simd helper functions: `_all_equal`, `_any_equal`, `_all_not_equal`, `_any_not_equal`.
- Changed every case i could find of implicit reduce functions to be explicit. (tests pass but may have missed some)
- `SIMD.__bool__()` constrained to size=1
- Removed some things that are unnecessary now, like overloads for `assert_equal` and `assert_not_equal`
- Tweaked some tests to make more sense
- Fixed a pre-existing bug with `_isclose()` while changing to explicit reduce

soraros:
You might need to coordinate with @leandrolcampos a bit because of #2375.

helehex:
i'm still not sure this should or will get pulled. it feels good writing simd code like this, but the downside is you don't get an error in the editor when doing converting a simd size>1 to bool, such as with if statements. you get a compiler error but i can see that being confusing

helehex:
another trade-off that should be mentioned:
Ideally, everything should be boolable, and this breaks that
but like sora mentioned, when working with simd size > 1, the boolability causes trouble

JoeLoser:
Thanks for pushing on this!  I think this is a general direction we want to go with, but there's a couple things going on in this PR I'd recommend we decompose/split up into separate PRs.  Specifically, I'd recommend

1. Constrain `__bool__` to ensure size is `1` to avoid this footgun.  I'd recommend beefing up the `constrained` message to be something like 

```
constrained[width == 1, "The truth value of a SIMD vector with more than one element is ambiguous. Use `v.reduce_and()` or `v.reduce_or()`"]()
```

2. I'd like to see a proposal of just changing `SIMD.eq` to return a `Bool` instead of being cute with the overload rules.  I think this is a better long-term approach for generic programming and the footgun we have now with SIMD equality.  I'd recommend we reference things like Swift and Rust when considering this API change.  Take a look at https://github.com/apple/swift-evolution/blob/main/proposals/0229-simd.md#detailed-design and how Rust models its `simd_eq` and friends [here](https://doc.rust-lang.org/std/simd/cmp/trait.SimdPartialEq.html).  In addition, I'd recommend we also discuss how we want to model the other comparison operators.  This may change in the library once we have extensions in the language too.

3. Split up the constructor removal into standalone PR?  I don't think it _needs_ to be coupled with these changes, right?

@abduld I'm also curious to get your general thoughts here as I know I've brought up my concerns with the existing `SIMD.eq` in the past.

helehex:
everyone i've talked to likes how `simd.eq` returns a simd though

helehex:
I agree we need a proper proposal though.
For now i'll just open a separate pr for constraining `simd.__bool__`. That will also align better with `simd.__int__`

helehex:
removing the bool constructor wasn't necessary, but it helped weed out all the cases of implicit reduce and is also very much related

helehex:
it finally occurred to me that `reduce_and` has bitwise ambiguity

helehex:
@JoeLoser I made it cute again ;]

helehex:
There's a couple of tests that had to be modified. They were finding the path of least resistance through the `String` overload of `assert_equal()`. If `String` was `EqualityComparable`, those overloads could be removed and this could be avoided.

stdlib/test/builtin/test_math.mojo
```diff
@@ -63,12 +63,12 @@ def test_max():
...
-    assert_equal(1, round(1.1))
-    assert_equal(2, round(1.5))
-    assert_equal(2, round(1.9))
-    assert_equal(2, round(2.0))
+    assert_equal(0.0, round(0.0))
+    assert_equal(1.0, round(1.0))
+    assert_equal(1.0, round(1.1))
+    assert_equal(2.0, round(1.5))
+    assert_equal(2.0, round(1.9))
+    assert_equal(2.0, round(2.0))
```
helehex, line 72: There's a couple of tests that had to be modified. They were finding the path of least resistance through the `String` overload of `assert_equal()`. If `String` was `EqualityComparable`, those overloads could be removed and this could be avoided.

helehex: (when you add the `.0`, it uses the correct overload)

helehex:
I was gonna close this, but then I tried fixing it, went a little crazy, and now I'm moving on with my life.

helehex:
oh wait

helehex:
Much better :]

stdlib/src/builtin/simd.mojo
```diff
@@ -717,6 +718,26 @@ struct SIMD[type: DType, size: Int](
...
+    # the [__:None=None] parameter is to give this overload lower precedence,
+    # while still conforming to EqualityComparable
+    @always_inline("nodebug")
+    fn __eq__[uwu: None = None](self, rhs: Self) -> Bool:
+        """Compares two SIMD vectors using all-equal-to comparison.
+
+        This overload allows EqualityComparable conformance.
+
+        Parameters:
+            uwu: OwO.
```
JoeLoser, line 731: Can we give a better identifier than uwu? :)

JoeLoser (commented):
!sync

stdlib/src/builtin/simd.mojo
helehex: *sigh* I suppose.

stdlib/test/builtin/test_math.mojo
JoeLoser: Is this still the case on latest nightly?

JoeLoser:
!sync

helehex:
I'd be surprised if this doesn't break everything.. in that case don't worry about it Joe, I'm sure we'll figure out something better.

martinvuyk:
Hi, a bit late to the party but won't doing something like adding a conditional conformance overload to eq work ? i.e.  `fn __eq__(self: Scalar[type], rhs: Scalar[type]) -> Bool:`

helehex:
Yeah that might work now, when i tried it before it would complain because of it not conforming in the general case

martinvuyk:
> i was confusing it with having condition `__bool__()` conformance like that, hah..

Well I guess it could be used for that too ? But you wouldn't have the nice and clear error message that the constraint has

But yeah my idea was this:
```mojo
struct A[size: UInt]:
    var field: Int

    fn __init__(inout self, field: Int):
        self.field = field

    fn __eq__(self: A[1], rhs: A[1]) -> Bool:
        return self.field == rhs.field


fn main():
    print(A[1](1) == A[1](1))
    print(
        A[2](2) == A[2](2)
    )  # invalid call to '__eq__': could not deduce parameter 'size' of parent struct 'A'
```

This works fine in a separate file, but I don't know if it'd pass all the tests for SIMD.

The only bad thing I see about doing it like this is that the error message is not clear, the user would have to go look at the function signature to understand what is going on.

helehex:
make sure it works when passing it as an `EqualityComparable`

JoeLoser:
!sync

JoeLoser:
!sync

stdlib/test/builtin/test_math.mojo
helehex: Yes, it's because there is an overload of assert equal that takes strings.

helehex: Removing them right now causes problems because some tests use `assert_equal(StringLiteral, StringSlice[lifetime])` for example.

helehex: Maybe those could call `str()` explicity.

helehex: Also, there's lots of strange symbols in the `test_string_iter()` which are really confusing the language server.

helehex: I kind of like what it does to the text colors :]
but maybe there's a better way.

helehex:
I removed the string overloads for `assert_equal()` as well, since those were causing problems

helehex:
oh good a docstring problem, i wanted to change something anyways :]

helehex:
!sync

JoeLoser:
I'm closing this PR due to recent SIMD work from @laszlokindrat on changing comparison operators to return a `Bool` - see https://github.com/modular/modular/blob/1dbbbac2b427b32c496dc9da5f91cd9e5b151f81/mojo/stdlib/stdlib/builtin/simd.mojo#L1221.

PR #2414, opened 2024-04-26
[stdlib] Add unit tests for the SIMD __rfloordiv__() method and constraint check on the element types

peymanbr:
The __rfloordiv__() method is already supported by Int and Object. This commit adds support for this method to the SIMD type.

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -671,6 +671,7 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
         Returns:
             `floor(rhs / self)` value.
         """
+        constrained[type.is_numeric(), "the type must be numeric"]()
         return rhs // self
 
     @always_inline("nodebug")
--- a/stdlib/test/builtin/test_simd.mojo
+++ b/stdlib/test/builtin/test_simd.mojo
@@ -223,6 +223,18 @@ def test_floordiv():
     assert_equal(Float32(99) // Float32(-2), -50)
 
 
+def test_rfloordiv():
+    alias I = SIMD[DType.int32, 4]
+    var i = I(2, 4, -2, -4)
+    assert_equal(i.__rfloordiv__(2), I(1, 0, -1, -1))
+    assert_equal(i.__rfloordiv__(Int32(2)), I(1, 0, -1, -1))
+
+    alias F = SIMD[DType.float32, 4]
+    var f = F(3, -4, 1, 5)
+    assert_equal(f.__rfloordiv__(3), F(1, -1, 3, 0))
+    assert_equal(f.__rfloordiv__(Float32(3)), F(1, -1, 3, 0))
+
+
 def test_mod():
     assert_equal(Int32(99) % Int32(1), 0)
     assert_equal(Int32(99) % Int32(3), 0)
@@ -873,6 +885,7 @@ def main():
     test_round()
     test_roundeven()
     test_floordiv()
+    test_rfloordiv()
     test_mod()
     test_rotate()
     test_shift()


stdlib/test/builtin/test_simd.mojo
```diff
@@ -102,6 +102,24 @@ def test_floordiv():
     assert_equal(Float32(2) // Float32(-2), -1)
     assert_equal(Float32(99) // Float32(-2), -50)
 
+    assert_equal(
+        SIMD[DType.int32, 4](2, 4, 1, 5) // Int32(2),
```
gabrieldemarmiesse, line 106: isn't this actually calling `__floordiv__` instead of `__rfloordiv__`? Since the compiler will try `__floordiv__` first if the two elements are both SIMD. This raises the question, in what situation is this implementation of `__rfloordiv__` useful? In python, it's only used to handle other types. 

Maybe there is some compiler logic I'm missing here?

peymanbr: Correct that test and the one right below it call __floordiv__(). I added those tests to have more test coverage for __floordiv__(). I can exclude them in this PR and add them in another PR if its misleading.

stdlib/src/builtin/simd.mojo
```diff
@@ -653,6 +653,23 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+    @always_inline("nodebug")
+    fn __rfloordiv__(self, value: Self) -> Self:
+        """Returns the division of value and self rounded down to the nearest
+        integer.
+
+        # Constraints:
+            The element type of the SIMD vector must be numeric.
+
+        Args:
+            value: The value to divide on.
```
peymanbr, line 665: Change to "The value that is being divided."

stdlib/test/builtin/test_simd.mojo
gabrieldemarmiesse: Then is which situation is it possible to call the function`__rfloordiv__` that was just added?

abduld (approved):
Thanks

gabrieldemarmiesse:
If I may add, unless I'm missing something, the method `__rfloordiv__` added in this PR can't be called. A call to `//` between two SIMD elements will always all `__floordiv__`. Thus while the unit tests added in this PR are good and can be merged,  the added method might not be as useful as we previously thought.

peymanbr:
"The added `__rfloordiv__()` method is definitely called; otherwise, the unit tests fail without it.

gabrieldemarmiesse:
It seems that `__rfloordiv__` was already added by an internal commit. 

See https://github.com/modularml/mojo/blob/nightly/stdlib/src/builtin/simd.mojo#L661

It came from this commit: https://github.com/modularml/mojo/commit/f5aee408379ae336d5f877694ddc7128d9bd9bb1

@peymanbr I'm sorry that this was already done by someone else. Coordination is pretty hard on a project evolving as fast as Mojo. Maybe it should be possible to change this PR to only add unit tests? That should still be a great addition :)

peymanbr:
@gabrieldemarmiesse I understand that there are a lot of activities in this repo. Thank you for notifying me about the added functionality.

I think the constraint check on the element types is missing in the implementation. Therefore, I've added this check to the `__floordiv__()` method implementation and limited the PR to include this check and the unit tests. I've also updated the title of this issue to accurately reflect the changes made in the PR.

laszlokindrat:
Thanks for the patch! I had to do some refactoring of the `math` module, and ended up implementing `SIMD.__rfloordiv__`, sorry about that! I do however would appreciate a patch with tests for this, and you almost have them. Can you update them, so they directly use the `__rfloordiv__` function call instead of the `//` operator? This should be the preferred way to unittest these magic methods for the basic numeric types (because it is less likely that a test would accidentally pass due to some implicit conversion rules).

soraros:
@laszlokindrat I thought the reason we need the `r` variant because we always implicit conversion to work. I think it's valuable to also have tests excising the operator path.

peymanbr:
@laszlokindrat @gabrieldemarmiesse Thank you for your feedback.

I changed the added tests to call ` __rfloordiv__` function instead of the `//` operator and moved them to their own test function named `test_rfloordiv()` for better visibility.

laszlokindrat:
> @laszlokindrat I thought the reason we need the `r` variant because we always implicit conversion to work. I think it's valuable to also have tests excising the operator path.

Yes, but we need to figure a more robust way to ensure that these tests actually hit that conversion path. Although these are less important because the compiler has internal tests to ensure it resolves the magic methods in the correct precedence order. Either way, we need the method itself to have unit tests that aren't affected by conversion rules. I will try to write up some guidelines on this for the future, but for now, just ensure you test dunder methods directly for the core numeric types (and maybe others as well).

stdlib/test/builtin/test_simd.mojo
laszlokindrat: It can happen if there is some implicit conversion. It can also be required by some traits (e.g. `CeilDivable`). See https://github.com/modularml/mojo/pull/2414#issuecomment-2098934194

laszlokindrat (approved):
Thanks!

PR #2430, opened 2024-04-28
[stdlib] Implement mkdir and rmdir

artemiogr97:
Signed-off-by: Artemio Garza Reyna <artemiogr97@gmail.com>


stdlib/src/os/os.mojo
```diff
@@ -328,3 +328,60 @@ fn unlink[pathlike: os.PathLike](path: pathlike) raises:
...
+    """Creates a directory at the specified path.
+    If the directory can not be created an error is raised.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
+
+    Args:
+      path: The path to the directory.
+      mode: The mode to create the directory with.
+    """
+
+    var error = external_call["mkdir", Int](path._as_ptr(), mode)
```
abduld, line 348: Should have commented to this pr instead (btw thanks for splitting it up) but the result should be Int32 here

artemiogr97: Sure, could you just confirm that I should use Int32?, in an older PR I was told to use Int: https://github.com/modularml/mojo/pull/2310#discussion_r1569653405

Then I could do the change in `os.remove`  in this pulll if it is ok to merge both changes a the same time

abduld: Yes. the unlink should be Int32

abduld (approved):

JoeLoser:
Quick update here: this is still working on landing internally.  On CI, I'm seeing some tests fail related to your change.  On Graviton, some `assert_raises` in the tests aren't firing, but I can't repro locally on a M1.  I'll need to investigate this some more.

JoeLoser:

PR #2439, opened 2024-04-28
[stdlib] Implementation of  rmtree

artemiogr97:
#2430 must be merged first

### DIFF
--- a/stdlib/src/shutil/shutil.mojo
+++ b/stdlib/src/shutil/shutil.mojo
@@ -1,0 +1,74 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Implements a number of high-level operations on files and
+collections of files.
+
+You can import these APIs from the `shutil` package. For example:
+
+```mojo
+from shutil import rmtree
+```
+"""
+
+from os import listdir
+from os.path import islink, isdir, isfile
+
+
+fn rmtree(path: String, ignore_errors: Bool = False) raises:
+    """Removes the specified directory and all its contents.
+    If the path is a symbolic link, an error is raised.
+    If ignore_errors is True, errors resulting from failed removals will be ignored.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
+
+    Args:
+      path: The path to the directory.
+      ignore_errors: Whether to ignore errors.
+    """
+    if os.path.islink(path):
+        raise Error("`path`can not be a symbolic link")
+
+    for file_or_dir in listdir(path):
+        if isfile(path + "/" + file_or_dir[]):
+            try:
+                os.remove(path + "/" + file_or_dir[])
+            except e:
+                if not ignore_errors:
+                    raise Error(e)
+            continue
+        if isdir(path + "/" + file_or_dir[]):
+            try:
+                rmtree(path + "/" + file_or_dir[], ignore_errors)
+            except e:
+                if ignore_errors:
+                    continue
+                raise Error(e)
+
+    os.rmdir(path)
+
+
+fn rmtree[
+    pathlike: os.PathLike
+](path: pathlike, ignore_errors: Bool = False) raises:
+    """Removes the specified directory and all its contents.
+    If the path is a symbolic link, an error is raised.
+    If ignore_errors is True, errors resulting from failed removals will be ignored.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
+
+    Parameters:
+      pathlike: The a type conforming to the os.PathLike trait.
+
+    Args:
+      path: The path to the directory.
+      ignore_errors: Whether to ignore errors.
+    """
+    rmtree(path.__fspath__(), ignore_errors)
--- a/stdlib/test/shutil/test_shutil.mojo
+++ b/stdlib/test/shutil/test_shutil.mojo
@@ -1,0 +1,54 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo %s
+
+from shutil import rmtree
+from os import mkdir
+from os.path import exists
+from pathlib import Path
+
+from testing import assert_false, assert_raises
+
+
+fn _create_some_files_and_dirs(path: Path) raises:
+    mkdir(path / "nested")
+    mkdir(path / "nested" / "more_nested")
+    mkdir(path / "nested" / "more_nested" / "even_more_nested")
+
+    _ = open(path / "nested.txt", "w")
+    _ = open(path / "nested" / "more_nested.txt", "w")
+    _ = open(path / "nested" / "more_nested" / "even_more_nested.txt", "w")
+
+
+fn test_rmtree() raises:
+    var cwd_path = Path()
+    var my_dir_path = cwd_path / "my_dir"
+
+    assert_false(
+        my_dir_path.exists(),
+        "Unexpected dir " + my_dir_path.__fspath__() + " it should not exist",
+    )
+
+    with assert_raises():
+        rmtree(my_dir_path)
+
+    mkdir(my_dir_path)
+    _create_some_files_and_dirs(my_dir_path)
+
+    rmtree(my_dir_path)
+
+    assert_false(my_dir_path.exists())
+
+
+fn main() raises:
+    test_rmtree()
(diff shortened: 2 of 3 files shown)


stdlib/src/os/os.mojo
```diff
@@ -328,3 +328,60 @@ fn unlink[pathlike: os.PathLike](path: pathlike) raises:
...
+    """Creates a directory at the specified path.
+    If the directory can not be created an error is raised.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
+
+    Args:
+      path: The path to the directory.
+      mode: The mode to create the directory with.
+    """
+
+    var error = external_call["mkdir", Int](path._as_ptr(), mode)
```
abduld, line 348: The result should be an int32 instead of an Int. An Int in mojo corresponds to a ssize_t

stdlib/src/os/os.mojo
```diff
@@ -328,3 +328,60 @@ fn unlink[pathlike: os.PathLike](path: pathlike) raises:
...
+
+fn rmdir(path: String) raises:
+    """Removes the specified directory.
+    If the path is not a directory or it can not be deleted, an error is raised.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
+
+    Args:
+      path: The path to the directory.
+    """
+    var error = external_call["rmdir", Int](path._as_ptr())
```
abduld, line 374: Same

abduld (approved):
Pretty cool. Please make sure about the Int types though

stdlib/src/shutil/shutil.mojo
```diff
@@ -0,0 +1,74 @@
...
+```mojo
+from shutil import rmtree
+```
+"""
+
+from os import listdir
+from os.path import islink, isdir, isfile
+
+
+fn rmtree(path: String, ignore_errors: Bool = False) raises:
```
laszlokindrat, line 27: @JoeLoser @ConnorGray Do you have any opinions if we should make `ignore_errors` keyword-only? Python doesn't do this, but it's not clear if that's just a result of backward compatibility. Otherwise, I always feel a bit strange about flags that can be specified positionally.

stdlib/src/shutil/shutil.mojo
```diff
@@ -0,0 +1,74 @@
...
+    """Removes the specified directory and all its contents.
+    If the path is a symbolic link, an error is raised.
+    If ignore_errors is True, errors resulting from failed removals will be ignored.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
+
+    Args:
+      path: The path to the directory.
+      ignore_errors: Whether to ignore errors.
```
laszlokindrat, line 35: Idea: Once we have parametric raising, we could introduce overloads that have this as a parameter. In the meanwhile, would it be beneficial to have a separate function that ignores errors and never raises?

stdlib/test/shutil/test_shutil.mojo
```diff
@@ -0,0 +1,54 @@
...
+    mkdir(path / "nested" / "more_nested" / "even_more_nested")
+
+    _ = open(path / "nested.txt", "w")
+    _ = open(path / "nested" / "more_nested.txt", "w")
+    _ = open(path / "nested" / "more_nested" / "even_more_nested.txt", "w")
+
+
+fn test_rmtree() raises:
+    var cwd_path = Path()
+    var my_dir_path = cwd_path / "my_dir"
```
laszlokindrat, line 35: I'm not so sure about this. It would be nice if we had some `tempfile`-like utilities to create these ephemeral paths. At the same time, `tempfile` should probably depend on `shutil`, not the other way around.
Idea 1: Maybe we could bring up some basic `tempfile` utilities so that we can at least have a platform-independent temporary path.
Idea 2: we could just use python interop in these tests while we get there.

stdlib/test/shutil/test_shutil.mojo
```diff
@@ -0,0 +1,54 @@
...
+        "Unexpected dir " + my_dir_path.__fspath__() + " it should not exist",
+    )
+
+    with assert_raises():
+        rmtree(my_dir_path)
+
+    mkdir(my_dir_path)
+    _create_some_files_and_dirs(my_dir_path)
+
+    rmtree(my_dir_path)
```
laszlokindrat, line 48: Testing these kinds of functions can actually be tricky: what if the test fails for some reason and the directory isn't deleted? The next time the test is run, `mkdir` can fail. One way to fix this is by doing this in a context manager, like [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). In lieu of that, some manual cleanup logic could also work.

laszlokindrat (commented):
Thanks for the patch! This could be very nice to have, but there are a bunch of library design questions, as well as implementation details that we should iron out. My main question at this point: do we actually want this _right now_, given that Python interop already allows one to do this?

While we work on these, I have a couple of suggestions that could help moving this forward: 1) we could use an `os.makedirs` with an `exist_ok` argument; 2) we could explore (maybe in a design doc) if we could start a `tempfile` module to build some testing utilities. @artemiogr97 Would you be interested in working on either of these?

CC: @ConnorGray @rparolin @JoeLoser

artemiogr97:
> My main question at this point: do we actually want this _right now_, given that Python interop already allows one to do this?

I agree that maybe it is a bit soon to start implementing a new module such as `shutil` (or an equivalent).
At the beginning I wanted to implement #2018 but then I soon realized that a lot of the basic functionalities such as `os.remove/os.unlink`, `os.mkdir`, `os.rmdir` where missing so I started implementing them to finish #2352, and still there is a lot of stuff missing `os.makedirs`, `os.join`, `os.abspath`, ...

> While we work on these, I have a couple of suggestions that could help moving this forward: 1) we could use an `os.makedirs` with an `exist_ok` argument; 2) we could explore (maybe in a design doc) if we could start a `tempfile` module to build some testing utilities. @artemiogr97 Would you be interested in working on either of these?

for your point 1, yes, I would like to work on that, everything that could be done to "complete" the `os` module makes a lot of sense, but I have some doubts:
- should I care about windows compatibility at some point?
- should we start creating the equivalents of `posix/nt` modules to start handling os specific stuff?
- if windows is to be taken into account it is going to be hard to make sure that everything is ok since mojo is not available in windows yet

for 2 I'm not sure what there is to explore/discuss about a `tempfile` module, the equivalent python module just contains a couple of functions/classes but nothing really out of the ordinary, most of the functionality is already in the PR that I mentioned above but it depends on `rmtree` for now, so if needed I could modify it to remove `rmtree` and use it as a testing module for now

stdlib/src/shutil/shutil.mojo
artemiogr97: I that could make sense for now

stdlib/test/shutil/test_shutil.mojo
artemiogr97: I guess python interop is the best option for now, one idea to prevent testing `shutil` with a potential `tempfile` module (which logically would depend on `shutil`) would be to use something like `subprocess.run` and call something like `rm -rf my_dir`, but it does not seem like a short-term solution

stdlib/test/shutil/test_shutil.mojo
artemiogr97: yes, it's tricky, but doing some manual cleanup is not that simple in this case, what if what is failing is `rmdir/unlink` ? obviously `rmtree` is relying on those functions so what else could be done?

we have the same "issue" in `test_remove` and `test_mkdir_and_rmdir`

laszlokindrat:
> > While we work on these, I have a couple of suggestions that could help moving this forward: 1) we could use an `os.makedirs` with an `exist_ok` argument; 2) we could explore (maybe in a design doc) if we could start a `tempfile` module to build some testing utilities. @artemiogr97 Would you be interested in working on either of these?
> 
> for your point 1, yes, I would like to work on that, everything that could be done to "complete" the `os` module makes a lot of sense, ...

That's awesome, thank you! Please ping me if you have PRs for these, I'm happy to review. I'm assuming these `os` utilities shouldn't depend on `shutil`, right (i.e. the dependency is the other way around)?

> ... but I have some doubts:
> * should I care about windows compatibility at some point?

For now, don't worry about Windows. The existing code in `os` would already break on native Windows. But during the implementation, feel free to leave `TODO`s in places where you think some special handling for Windows might be needed.

> * should we start creating the equivalents of `posix/nt` modules to start handling os specific stuff?

I would be lazy about this stuff. If the complexity of handling multiple platforms within `os`/`sys` (as well as user code) would grow to the point where it makes sense, we can introduce these, but I don't think we're there yet.

> * if windows is to be taken into account it is going to be hard to make sure that everything is ok since mojo is not available in windows yet

Yep, don't worry about this for now.

> for 2 I'm not sure what there is to explore/discuss about a `tempfile` module, the equivalent python module just contains a couple of functions/classes but nothing really out of the ordinary, most of the functionality is already in the PR that I mentioned above but it depends on `rmtree` for now, so if needed I could modify it to remove `rmtree` and use it as a testing module for now

I see now that you already did quite a bit of work on `tempfile`, which is awesome. Let's discuss that on the other PR.

stdlib/src/shutil/shutil.mojo
JoeLoser: Making `ignore_errors` keyword-only SGTM.

JoeLoser:
What's the status of this @laszlokindrat @artemiogr97?  Can we move forward with this?  I saw https://github.com/modularml/mojo/pull/2430 already merged.

artemiogr97:
@JoeLoser I moved this implementation inside of `tempfile` to be able to finish #2743 and left a TODO to replace it when the equivalent function exists (either in `shutil` or an equivalent module) so I don't really need this right now.

I would say that it is up to the maintainers to decide whether the `shutil` module should be created now or we we should move this function to a temporary place such as `os.path`.

JoeLoser:
> @JoeLoser I moved this implementation inside of `tempfile` to be able to finish #2743 and left a TODO to replace it when the equivalent function exists (either in `shutil` or an equivalent module) so I don't really need this right now.
> 
> I would say that it is up to the maintainers to decide whether the `shutil` module should be created now or we we should move this function to a temporary place such as `os.path`.

Let's hold off _right this moment_ on eagerly creating `shutil`, but we can circle back on that.  Since you made `_rmtree` private when working on the temp file functions, I think it's safe to close this PR for right now.  Thanks again for all of your hard work on this!

PR #2454, opened 2024-04-30
[stdlib] Allow `StringRef` to work with both Int8 and UInt8

gabrieldemarmiesse:
This is related to https://github.com/modularml/mojo/issues/2317 

This is similar to https://github.com/modularml/mojo/pull/2448 but for StringRef.

With this, users and maintainers can work with both Int8 and UInt8 and will allow a graceful transition. This should also allow users to switch during the next release.


JoeLoser:
🎉 🔥 This landed in tonight's nightly release in https://github.com/modularml/mojo/pull/2480!

PR #2456, opened 2024-04-30
[stdlib] Complete most of the remaining methods of `Set`

arvindavoudi:
I've been working on this for a while and I believe this PR is a big step toward reaching the point that our collection datatypes are almost identical to Python. Long story short, First, let me introduce you to the new methods and operators that are being implemented with this PR:

## Methods
1- `issubset`
2- `isdisjoint`
3- `issuperset`
4- `symmetric_difference`
5- `symmetric_difference_update`
6- `discard`
7- `clear`

## Operators
1- `__le__`
2- `__lt__`
3- `__ge__`
4- `__gt__`
5- `__xor__`
6- `__ixor__`

## Renamed methods
- `remove_all` renamed to `difference_update`

## The EDGE CASES:
The methods are designed to replicate the same functionality we get from them in Python. So, This PR assures that we have no methods/operators left that are not coded yet. Cool, right? But there are still some points that must be considered. I list them below:

1- The only method not implemented yet is `copy`. I haven't implemented this because of the differences between Python and Mojo and the related concepts of shallow/deep copy. Any suggestion is welcomed.

2- I have observed `pop` method has slightly different functionality than the Pythonic version. In Python, `pop` method works by randomly deleting an item, which is not what we have in Mojo's sets. We can keep the current functionality and also add the Pythonic one.  They can be declared all in one single method but a default argument can determine if the user wants the random poping or the other.

3- In Python, some of the methods that accept another set (e.g. `union`), work with variadic arguments which means you can pass more than one set/collection to them. The newly added and the previously implemented methods do not still support such a thing. I'm deciding whether to add them here.

### DIFF
--- a/stdlib/src/collections/set.mojo
+++ b/stdlib/src/collections/set.mojo
@@ -463,3 +463,24 @@ struct Set[T: KeyElement](Sized, EqualityComparable, Hashable, Boolable):
                 return False
 
         return True
+
+    fn symmetric_difference(self, other: Self) -> Self:
+        """Returns the symmetric difference of two sets.
+        
+        Args:
+            other: The set to find the symmetric difference with.
+
+        Returns:
+            A new set containing the symmetric difference of the two sets.
+        """
+        var result = Set[T]()
+        
+        for element in self:
+            if element[] not in other:
+                result.add(element[])
+        
+        for element in other:
+            if element[] not in self:
+                result.add(element[])
+        
+        return result ^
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
This is awesome, thank you for the contribution!!! 🎉 🎉 🎉 

It just got merged internally and should be available in today's nightly release.

JoeLoser:
Landed in https://github.com/modularml/mojo/commit/328bbf537dbe39246e06fb7ac606426a503371d7 - thanks again for the contribution! 🎉 

I forgot to write a changelog entry for this, but I'll add it to my list unless you'd like to open a PR to update the changelog for this.  Sorry about that!

PR #2457, opened 2024-05-01
[stdlib] Support print to stderr

### LINKED ISSUE #2453: [Feature Request] Add a way for `print` to print to stderr
### Review Mojo's priorities

### What is your request?

In Python, we can print to stderr like so:

```python
import sys
print("hi", file=sys.stderr)
```

I'd like to accomplish the same thing in Mojo, but I believe the Mojo print function (as documented [here](https://docs.modular.com/mojo/stdlib/builtin/io#print)) can only print to `stdout` right now.

I'm developing a command-line tool written in Mojo, and I wish to print error messages. These are typically printed to `stderr`.

### Any other details?

This is a good first issue for new contributors, since it just requires some light plumbing to specify `_fdopen.STDOUT` or `_fdopen.STDERR` (defined [here](https://github.com/modularml/mojo/blob/432b2c1c966b8f2d9866d0b5257ab603aacd62d1/stdlib/src/builtin/io.mojo#L52-L54)) when calling `print`. Mojo supports keyword arguments, so one option would be to add a `file=` keyword argument similar to the one supported by Python's `print` function.

Tests also ought to be added to ensure that what's bring printed is actually being written to stderr.

GeauxEric:
Add keyword argument to `print` function to support stream to stderr.

Fix #2453 

Signed-off-by: Yun Ding <yunding.eric@gmail.com>


stdlib/test/builtin/test_print.mojo
```diff
@@ -72,8 +72,13 @@ fn test_print_sep():
     # CHECK: a/1/2xx
     print("a", 1, 2, sep="/", end="xx\n")
 
+fn test_print_stderr():
+    # TODO: check print to stderr
+    print("stderr", file=2)
```
gabrieldemarmiesse, line 77: To avoid magic numbers, I wonder if we can add `alias stderr = 2` in `sys`, same for `stdout`.
With this, we can do `print("something", file=sys.stderr)`. Later on, when we have mutable global variables, we can replace those numbers by actual file descriptors.

ConnorGray: I agree, we should have named constants for these, which would have better readability and consistency with Python.

I suggest we add a new file called `sys/_io.mojo` and add these constants there:

```mojo
alias stdout: Int = 1
alias stderr: Int = 2
```

These can be integer values for now, and we'll plan to adjust them to be "file objects" over time as the IO parts of the standard library shape up.

stdlib/src/builtin/io.mojo
```diff
@@ -89,8 +89,8 @@ struct _fdopen:
 
 
 @no_inline
-fn _flush():
-    with _fdopen(_fdopen.STDOUT) as fd:
+fn _flush(file: Int = _fdopen.STDOUT):
```
ConnorGray, line 92: Once we added [the named comments mentioned earlier](https://github.com/modularml/mojo/pull/2457/files#r1586789342), we should update this to:

```suggestion
fn _flush(file: Int = sys.stdout):
```

and do similarly in the other new `file` arguments.

ConnorGray:
Thanks for working on this @GeauxEric!🙂

I left a couple of small suggestions, but I noticed this PR is still in 'Draft' mode, so I'll hold off on leaving any more feedback until you indicate you're ready :)

gabrieldemarmiesse:
@GeauxEric could I suggest that you rebase this PR on top of #2463 ? This should make this pull request quite a bit smaller in the end, as there will be less code to change.

GeauxEric:
rebased https://github.com/modularml/mojo/pull/2463

stdlib/test/builtin/test_print.mojo
GeauxEric: done.

stdlib/src/builtin/io.mojo
GeauxEric: done.

stdlib/test/builtin/test_print.mojo
```diff
@@ -73,7 +74,17 @@ fn test_print_sep():
     print("a", 1, 2, sep="/", end="xx\n")
 
 
+fn test_print_stderr():
+    # TODO: check print to stderr
```
JoeLoser, line 78: Do you have ideas on how you'd like to test that we wrote to stderr here?

modocache: If I could offer a suggestion: since this test uses FileCheck anyway, you could add a check specifically for this. Something like this:

```
# RUN: %mojo %s 2>&1 1>/dev/null | FileCheck %s --check-prefix=CHECK-STDERR
# CHECK-STDERR: stderr
```

The `2>&1 1>/dev/null` should redirect all the `print` statements from other tests to `/dev/null`, silencing them, while the `print(..., file=sys.stderr)` statements should be printed to `stdout`, which `FileCheck` makes expectations against. As a result, the `CHECK-STDERR: stderr` expectation would fail if `stderr` was only being printed to `stdout`, not `stderr`.

JoeLoser:
Landed in today's nightly: https://github.com/modularml/mojo/pull/2615. Thanks for the contribution! @modocache is eager to start using this.

modocache:
Absolutely! Thanks a ton, @GeauxEric!

PR #2461, opened 2024-05-01
[docs] Add constraint documentation for various `reduce`-like functions in simd.mojo

### LINKED ISSUE #2237: [Docs] Constraints documentations need to be repeated for overloaded methods
### Where is the problem?

https://github.com/modularml/mojo/blob/a8f7393a0e1e3e0a307391b844e1f00e9c4a05d9/stdlib/src/memory/unsafe.mojo#L60-L255

### What can we do better?

Right now, overloaded methods that delegate to one another do not repeat the constraints in the docstring. This provides confusion to users who are looking at the one(s) that doesn't have the constraints documented. One such instance is [bitcast](https://github.com/modularml/mojo/blob/a8f7393a0e1e3e0a307391b844e1f00e9c4a05d9/stdlib/src/memory/unsafe.mojo#L60-L255) where there are missing constraints in the some of the docstring.

### Anything else?

_No response_

Brian-M-J:
Fixes #2237 for `reduce`-like functions in simd.mojo:
- Adds documentation for constraint on value of `size_out` not exceeding width of input vector in `reduce_add`, `reduce_mul`, `reduce_max` and `reduce_min`.

Signed-off-by: Brian-M-J <brianmj02@gmail.com>

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2006,6 +2006,7 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
             size_out: The width of the reduction.
 
         Constraints:
+            `size_out` must not exceed width of the vector.
             The element type of the vector must be integer or FP.
 
         Returns:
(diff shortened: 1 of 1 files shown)


Brian-M-J:
Is the constraint on `size_out` the only one for `reduce_add` or does "The element type of the vector must be integer or FP." also apply?

JoeLoser (approved):

PR #2463, opened 2024-05-02
[stdlib] Remove unnecessary arguments and overloads in print()

gabrieldemarmiesse:
The print function was using overloads and `rest.each` to be able to generate all the `_put` call necessary at compile time. It's possible to do the same thing without overloads and also with less arguments. This will increase maintainability and make the signature easily understandable in the docs.

I would suggest that we merge this simplification PR before merging https://github.com/modularml/mojo/pull/2457 which adds support for the `file` argument in print(). Or we could rebase https://github.com/modularml/mojo/pull/2457 on top of this PR too.

### DIFF
--- a/stdlib/src/builtin/io.mojo
+++ b/stdlib/src/builtin/io.mojo
@@ -374,7 +374,7 @@ fn print[
         Ts: The elements types.
 
     Args:
-        values: The remaining elements.
+        values: The elements to print.
         sep: The separator used between elements.
         end: The String to write after printing the elements.
         flush: If set to true, then the stream is forcibly flushed.
--- a/stdlib/src/builtin/io.mojo
+++ b/stdlib/src/builtin/io.mojo
@@ -379,17 +379,16 @@ fn print[
         end: The String to write after printing the elements.
         flush: If set to true, then the stream is forcibly flushed.
     """
-    alias number_of_values = values.__len__()
 
     @parameter
-    fn print_with_separator[i: Int]():
-        _put(values.get_element[i]()[])
+    fn print_with_separator[i: Int, T: Stringable](value: T):
+        _put(value)
 
         @parameter
-        if i < number_of_values - 1:
+        if i < values.__len__() - 1:
             _put(sep)
 
-    unroll[print_with_separator, number_of_values]()
+    values.each_idx[print_with_separator]()
 
     _put(end)
     if flush:
--- a/stdlib/src/builtin/io.mojo
+++ b/stdlib/src/builtin/io.mojo
@@ -359,27 +359,10 @@ fn _put(x: DType):
 
 
 @no_inline
-fn print(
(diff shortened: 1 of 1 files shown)


soraros:
You could use `rest.each_idx`.

stdlib/src/builtin/io.mojo
```diff
@@ -358,28 +358,11 @@ fn _put(x: DType):
...
-    """
-    _put(end)
-    if flush:
-        _flush()
-
-
 @no_inline
 fn print[
-    T: Stringable, *Ts: Stringable
+    *Ts: Stringable
```
JoeLoser, line 363: @Mogball Is this still needed (split first from rest of pack) for compiler optimization/speed, or is this just an artifact of where packs were at the time?

JoeLoser (commented):
In general, looks good.  It was originally written this way for compiler perf reasons (see my comment inline).  I don't know if that's still required or not, so I tagged a compiler engineer to comment.

gabrieldemarmiesse:
@soraros  thanks for the suggestion, I used each_idx

stdlib/src/builtin/io.mojo
Mogball: This is small valid as a small optimization. Invoking the function with a pack still has overhead if there is a single element. Although I don't really imagine `print` is performance sensitive

JoeLoser: Let's do the simple, intuitive thing than rather than trying to micro-optimize compile-time perf IMO.

Mogball: It's not compile-time perf, it's generated code perf

Mogball: In the future. Packs should zero-overhead, but it's a TODO atm

gabrieldemarmiesse: Knowing this, I believe it's safe to merge this PR as the overhead will disapear at some point without us having to change anything

JoeLoser: Yep, I just imported it and I'll merge it soon.

JoeLoser (approved):
Great cleanup!

JoeLoser:
> @JoeLoser shall we do the same for `_print_fmt`?

Makes sense to me.  Any known reason we shouldn't do this for the formatting, @ConnorGray?  I suspect you just copied after `print` with how it handled variadics.

JoeLoser:
This just landed internally.  The code-gen changed some debug-info stuff internally, so it took a bit of time for me to chase that down.  But it'll be available in today's nightly later.  Thank you!

ConnorGray:
> shall we do the same for _print_fmt?

Yep, I don't see any reason not to, that would be a nice simplification :)

JoeLoser:
Landed in https://github.com/modularml/mojo/commit/6ae4384abbf59f5e3e41159f6d88602a19528279 - thanks again for the contribution! 🎉

PR #2475, opened 2024-05-02
[stdlib] Fix accidental use of `assert_true` in tests

soraros:
Found when trying to remove `String` constructor from `Stringable`. Implicit conversion is just, evil.

@JoeLoser Shall we make the `msg` argument keyword only in the mean time to prevent this from happening?


JoeLoser (approved):
Oof.  Thanks for noticing this! +1 for making `msg` a keyword-only argument in a separate PR.

Just imported it and it should merge internally shortly and be available in tomorrow's (Friday) nightly.

JoeLoser:
🎉 🔥 This landed in tonight's nightly release in https://github.com/modularml/mojo/commit/11a6c4cf1bc29f0d1dd035a173a2e277fe30ac91

PR #2481, opened 2024-05-03
[stdlib] Use UInt8 as the storage type for `StringRef`

gabrieldemarmiesse:
Note to the maintainers:

Something we can see here is that it's a pain to have other structs/methods/functions access the attributes or our containers directly. Here since I changed the `.data` attribute, a lot of code broke, but it should not have been the case since methods `_as_ptr()` and `_as_uint8_ptr()` were provided.

I would advise to make the attributes of our struct private and to provide public getters and setters to allow other functions to use the pointers and such in a future-proof way. 

Long story short, for `StringRef`, that would mean:
* `var data: DTypePointer[DType.uint8]` -> `var _data: DTypePointer[DType.uint8]` 
* `fn _as_uint8_ptr(self)` -> `fn get_unsafe_ptr(self)`


helehex:
I agree with adding underscores to attributes that should not be accessed directly

PR #2483, opened 2024-05-03
[tools] Add a docstring check to the CI pre-commit

### LINKED ISSUE #2331: [tools] [stdlib] API doc string validation

JoeLoser:
Split up from https://github.com/modularml/mojo/pull/2483 since Copybara
internally isn't managing the `.pre-commit-config.yaml` file.  So this PR just
updates the config file itself.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,12 @@ repos:
         language: system
         files: '\.(mojo|🔥)$'
         stages: [commit]
+      - id: check-docstrings
+        name: check-docstrings
+        entry: python3 ./stdlib/scripts/check-docstrings.py
+        language: system
+        pass_filenames: false
+        stages: [commit]
       - id: check-license
         name: check-license
         entry: mojo stdlib/scripts/check-licenses.mojo


JoeLoser:
Example of doc string validation running in CI can be found [here](https://github.com/modularml/mojo/actions/runs/8946212305/job/24576472947?pr=2483#step:7:10).  Thanks for all your hard work, @gabrieldemarmiesse!  Now we finally have doc string validation going! 👏

PR #2491, opened 2024-05-03
[stdlib] Remove `__and__`/`__or__`/... from `Optional` and `OptionalReg`

soraros:
We now have guaranteed implicit conversion from `Boolable` to `Bool` which renders them useless.


stdlib/test/collections/test_optional.mojo
```diff
@@ -39,10 +39,6 @@ def test_basic():
 
     assert_equal(1, a.value()[])
 
-    # Test invert operator
-    assert_false(~a)
-    assert_true(~b)
-
```
laszlokindrat, line 45: I think Bool is getting an __invert__ method soon in another patch, can you rebase once that lands and keep these?

laszlokindrat: nevermind, I will do this internally when I land this

laszlokindrat (approved):
Thanks!

stdlib/test/collections/test_optional.mojo
soraros: Ah, sorry I forgot that.

laszlokindrat: I tried this but it actually doesn't work with implicit conversion to Bool. I'm not sure why, but I will just leave in the invert method for now.

JoeLoser:

PR #2493, opened 2024-05-03
[stdlib] handle escape characters with `repr()`

helehex:
guarantees a printable representation of the string when using `repr()`

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -107,10 +107,7 @@ fn chr(c: Int) -> String:
     # 4: 00000000 000aaabb bbbbcccc ccdddddd -> 11110aaa 10bbbbbb 10cccccc 10dddddd     a >> 18 | 0b11110000, b >> 12 | 0b10000000, c >> 6 | 0b10000000, d | 0b10000000
 
     if (c >> 7) == 0:  # This is 1 byte ASCII char
-        var p = DTypePointer[DType.int8].alloc(2)
-        p.store(c)
-        p.store(1, 0)
-        return String(p, 2)
+        return _chr_ascii(c)
 
     @always_inline
     fn _utf8_len(val: Int) -> Int:
@@ -133,6 +130,71 @@ fn chr(c: Int) -> String:
         p.store(i, ((c >> shift) & 0b00111111) | 0b10000000)
     p.store(num_bytes, 0)
     return String(p.bitcast[DType.int8](), num_bytes + 1)
+
+
+# ===----------------------------------------------------------------------===#
+# ascii
+# ===----------------------------------------------------------------------===#
+
+
+@always_inline("nodebug")
+fn _chr_ascii(c: Int8) -> String:
+    """Returns a string based on the given ASCII code point.
+
+    Args:
+        c: An integer that represents a code point.
+
+    Returns:
+        A string containing a single character based on the given code point.
+    """
+    return String(String._buffer_type(c, 0))
+
+
+@always_inline("nodebug")
+fn _repr_ascii(c: Int8) -> String:
+    """Returns a printable representation of the given ASCII code point.
+
+    Args:
+        c: An integer that represents a code point.
+
+    Returns:
+        A string containing a representation of the given code point.
+    """
+    alias ord_tab = ord("\t")
+    alias ord_new_line = ord("\n")
+    alias ord_carriage_return = ord("\r")
+    alias ord_back_slash = ord("\\")
+
+    if c == ord_back_slash:
+        return r"\\"
+    elif isprintable(c):
+        return _chr_ascii(c)
+    elif c == ord_tab:
+        return r"\t"
+    elif c == ord_new_line:
+        return r"\n"
+    elif c == ord_carriage_return:
+        return r"\r"
+    else:
+        var uc = c.cast[DType.uint8]()
+        if uc < 16:
+            return hex(uc, r"\x0")
+        else:
+            return hex(uc, r"\x")
+
+
+# TODO: This is currently the same as repr, should change with unicode strings
+@always_inline("nodebug")
+fn ascii(value: String) -> String:
+    """Get the ASCII representation of the object.
+
+    Args:
+        value: The object to get the ASCII representation of.
+
+    Returns:
+        A string containing the ASCII representation of the object.
+    """
+    return value.__repr__()
 
 
 # ===----------------------------------------------------------------------===#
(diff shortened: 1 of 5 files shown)


stdlib/src/builtin/string.mojo
```diff
@@ -135,6 +133,74 @@ fn chr(c: Int) -> String:
     return String(p.bitcast[DType.int8](), num_bytes + 1)
 
 
+# ===----------------------------------------------------------------------===#
+# ascii
+# ===----------------------------------------------------------------------===#
+
+
+@always_inline("nodebug")
+fn chr_ascii(c: Int8) -> String:
```
laszlokindrat, line 142: About this and the other ascii helpers: Could you please prepend them with `_`? At this point, I'm not sure where these will live, and which of them will survive, so I would rather not make them public.

stdlib/src/builtin/string_literal.mojo
```diff
@@ -144,6 +145,14 @@ struct StringLiteral(
         """
         return self
 
+    fn __repr__(self) -> String:
```
laszlokindrat, line 148: Please add tests for this, directly using this method (as opposed to the builting `repr` function).

stdlib/test/builtin/test_repr.mojo
```diff
@@ -47,7 +47,24 @@ def test_none_representation():
...
+    assert_equal(repr("\n"), r"'\n'")
+    assert_equal(repr("\x0d"), r"'\r'")
+    assert_equal(repr("\x0e"), r"'\x0e'")
+    assert_equal(repr("\x1f"), r"'\x1f'")
+    assert_equal(repr(" "), "' '")
+    assert_equal(repr("'"), '"\'"')
+    assert_equal(repr("A"), "'A'")
+    assert_equal(repr("\\"), r"'\\'")
+    assert_equal(repr("~"), "'~'")
+    assert_equal(repr("\x7f"), r"'\x7f'")
```
laszlokindrat, line 63: Can you please ensure that these use `String.__repr__`? Due to how implicit conversion work, I think this actually uses `StringLiteral.__repr__` as it stands right now.

laszlokindrat (commented):
Looks pretty good, thank you for the patch! There are a couple questions and requests for change, nothing major, though.

stdlib/src/builtin/string.mojo
```diff
@@ -135,6 +133,74 @@ fn chr(c: Int) -> String:
...
+    Args:
+        c: An integer that represents a code point.
+
+    Returns:
+        A string containing a single character based on the given code point.
+    """
+    var p = DTypePointer[DType.int8].alloc(2)
+    p.store(c)
+    p.store(1, 0)
+    return String(p, 2)
```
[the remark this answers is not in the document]
helehex: Is it ok if I use `String._buffer_type(...)` instead?

stdlib/test/builtin/test_repr.mojo
helehex: `StringLiteral.__repr__`, converts it to a String, then calls `String.__repr__` so there's not much reason to change it, unless we want to do something else for `StringLiteral.__repr__()`.
Should I use `String(..)` anyways?

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
helehex: I'll just do that since the underlying type may change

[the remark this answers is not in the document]
laszlokindrat: Sure, that's fine, thanks!

stdlib/test/builtin/test_repr.mojo
laszlokindrat: Yeah, you're right, which kind of goes to the point that we should ensure that this does not exercise `StringLiteral.__repr__` (because tests for those should be in another file, similarly enforcing that it does not use `String.__repr__`).

stdlib/test/builtin/test_repr.mojo
```diff
@@ -47,7 +47,24 @@ def test_none_representation():
     assert_equal(repr(None), "None")
 
 
+def test_escape_representation():
```
laszlokindrat, line 50: Could you please move this to `test_string.mojo`, once you ensure it only uses `String.__repr__`?

stdlib/src/builtin/string_literal.mojo
laszlokindrat: And please place them in `test_string_litera.mojo`.

stdlib/src/builtin/string.mojo
```diff
@@ -448,10 +534,18 @@ struct String(
 
         You don't need to call this method directly, use `repr(my_string)` instead.
         """
-        if "'" in self:
-            return '"' + self + "'"
+        alias ord_squote = ord("'")
+        var result = String()
+        var use_dquote = False
+        for idx in range(len(self._buffer) - 1):
+            var char = self._buffer[idx]
```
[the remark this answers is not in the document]
helehex: We could, but then we would have to account for the null terminator, and I'm not sure that would look any better.

stdlib/src/builtin/string.mojo
helehex: sure

stdlib/src/builtin/string_literal.mojo
helehex: ok that sounds good. I'm also going to move the `repr` escape string tests into `test_string.mojo` (also applying the suggestion below), and add corresponding ones into `test_string_literal.mojo`.

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
helehex: if we had ref slicing of list, that might look nice

stdlib/test/builtin/test_repr.mojo
helehex: (I'll do what I mentioned in the above suggestion.)

stdlib/test/builtin/test_repr.mojo
helehex: yup

stdlib/src/builtin/string_literal.mojo
helehex: it looks like things are moving towards not calling them directly in the tests, so I'm still going to use the `repr` function

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
laszlokindrat: right, okay, this is fine, thanks!

stdlib/src/builtin/string_literal.mojo
laszlokindrat: Where do you see things not moving in that direction?

laszlokindrat (approved):
Awesome, thank you for fixing this!

JoeLoser:

PR #2495, opened 2024-05-03
[stdlib] Refactor slice to use Optionals and fix negative step slices

### LINKED ISSUE #1944: [BUG]: `Slicing with [::-1]` becomes `[0: 9223372036854775807 : -1]` and has a negative length
### Bug description

I expect slicing with `[::-1]` to produce something that can be bounds checked and used to reverse an entire sequence. The correct length of `Slice(0, 9223372036854775807, -1)` should be 0 and not -9223372036854775807.

Or the compiler could force start and end to be provided when step is negative.

The new `Slice.__init__[T0: AnyRegType, T1: AnyRegType, T2: AnyRegType](start: T0, end: T1, step: T2) -> Self` allows working around this with custom slice. But this is a sharp edge when using the default Slice provided by sugared syntax.

### Steps to reproduce

This code:

```mojo
@value
struct Array:
    var size: Int
 
    fn __getitem__(self, s: Slice) -> Int:
        print("start:", s.start, "end:", s.end, "step:", s.step)
        print("slice length: ", len(s))
        print("s[0]:", s[0], "s[1]:", s[1])
        return 0
 
    fn main():
        var y = Array(1)
        print("y[::-1]")
        _ = y[::-1]
```

produces:

```console
y[::-1]
start: 0 end: 9223372036854775807 step: -1
slice length:  -9223372036854775807
s[0]: 0 s[1]: -1
```

### System information

```shell
MacOS using Mojo 24.1.0
```

### LINKED ISSUE #2046: [BUG] List slicing (`__getitem__`) with negative step is broken
### Bug description

The gist is probably #1944 prevents accurately setting the start index and end index in many cases when step is negative.

### Steps to reproduce

```mojo
from collections.list import List
 
fn main():
    var l = List[Int](1, 2, 3)
    var slice1 = l[1:0:-1]
    print(slice1[0]) # prints 3, expected 2
    slice1 = l[2:1:-1]
    print(slice1[0]) # prints 1, expected 3
    slice1 = l[:0:-1]
    print(len(slice1)) # prints 0, expected 2
    slice1 = l[2::-1]
    print(len(slice1)) # prints 1, expected 3
```

### System information

```shell
MacOS Sonoma 14.2.1 (23C71)
mojo 24.2.0 (c2427bc5)
modular 0.6.0 (04c05243)
```

### LINKED ISSUE #2142: [BUG] [Refactor] `Slice.end` should not be default to `_int_max_value()`
### Proposal
defaulting `Slice.end` to -1 has several benefit. First of all, it avoids unsafe memory access in case it is used in the wrong way. Secondly, I saw most of the struct implements a `_adjust_span(self, span: Slice) -> Slice` method to round the slice. This refactor could simplify a lot of these method throughout the stdlib

```mojo
fn _adjust_span(self, span: Slice) -> Slice:
        """Adjusts the span based on the list length."""
        var adjusted_span = span

        if adjusted_span.start < 0:
            adjusted_span.start = len(self) + adjusted_span.start
        

        if not adjusted_span._has_end(): # <--- this if statement can be removed
            adjusted_span.end = len(self)
        elif adjusted_span.end < 0:
            adjusted_span.end = len(self) + adjusted_span.end

        if span.step < 0:
            var tmp = adjusted_span.end
            adjusted_span.end = adjusted_span.start - 1
            adjusted_span.start = tmp - 1

        return adjusted_span
```

### Steps to reproduce

see #1944 and #2046

### System information

_No response_

bgreni:
- Use `Optional` for `Slice` `start` and `end`
- Fix negative step slice indexing
- Fix incorrect String unit tests
- Fix `_StridedRange` length with mismatched start and end indices

Fixes #2046
Fixes #1944
Fixes #2142

### DIFF
--- a/stdlib/src/builtin/builtin_slice.mojo
+++ b/stdlib/src/builtin/builtin_slice.mojo
@@ -16,27 +16,14 @@ These are Mojo built-ins, so you don't need to import them.
 """
 
 from sys.intrinsics import _mlirtype_is_eq
+from collections import OptionalReg
 
 
 @always_inline("nodebug")
-fn _int_max_value() -> Int:
-    # FIXME: The `slice` type should have the concept of `None` indices, but the
-    # effect of a `None` end index is the same as a very large end index.
-    return int(Int32.MAX)
-
-
-@always_inline("nodebug")
-fn _default_or[T: AnyTrivialRegType](value: T, default: Int) -> Int:
-    # TODO: Handle `__index__` for other types when we have traits!
-    @parameter
-    if _mlirtype_is_eq[T, Int]():
-        return __mlir_op.`kgen.rebind`[_type=Int](value)
-    else:
-        __mlir_op.`kgen.param.assert`[
-            cond = (_mlirtype_is_eq[T, NoneType]()).__mlir_i1__(),
-            message = "expected Int or NoneType".value,
-        ]()
-        return default
+fn _compare_optional(x: OptionalReg[Int], y: OptionalReg[Int]) -> Bool:
+    if x and y:
+        return x.value() == y.value()
+    return not x and not y
 
 
 @register_passable("trivial")
@@ -55,9 +42,9 @@ struct Slice(Stringable, EqualityComparable):
     ```
     """
 
-    var start: Int
+    var start: OptionalReg[Int]
     """The starting index of the slice."""
-    var end: Int
+    var end: OptionalReg[Int]
     """The end index of the slice."""
     var step: Int
     """The step increment value of the slice."""
@@ -75,24 +62,26 @@ struct Slice(Stringable, EqualityComparable):
         self.step = 1
 
     @always_inline("nodebug")
-    fn __init__[
-        T0: AnyTrivialRegType, T1: AnyTrivialRegType, T2: AnyTrivialRegType
-    ](inout self, start: T0, end: T1, step: T2):
+    fn __init__(
+        inout self,
+        start: OptionalReg[Int],
+        end: OptionalReg[Int],
+        step: OptionalReg[Int],
+    ):
         """Construct slice given the start, end and step values.
-
-        Parameters:
-            T0: Type of the start value.
-            T1: Type of the end value.
-            T2: Type of the step value.
 
         Args:
             start: The start value.
             end: The end value.
             step: The step value.
         """
-        self.start = _default_or(start, 0)
-        self.end = _default_or(end, _int_max_value())
-        self.step = _default_or(step, 1)
+        self.start = start
+        self.end = end
+
+        if step and step.value() == 0:
+            abort("Cannot build slice with a step of 0")
+
+        self.step = step.value() if step else 1
 
     fn __str__(self) -> String:
         """Gets the string representation of the span.
@@ -100,10 +89,10 @@ struct Slice(Stringable, EqualityComparable):
         Returns:
             The string representation of the span.
         """
-        var res = str(self.start)
+        var res = str(self.start.value()) if self.start else ""
         res += ":"
-        if self._has_end():
-            res += str(self.end)
+        if self.end:
+            res += str(self.end.value()) if self.end else ""
         res += ":"
         res += str(self.step)
         return res
@@ -120,8 +109,8 @@ struct Slice(Stringable, EqualityComparable):
             corresponding values of the other slice and False otherwise.
         """
         return (
-            self.start == other.start
-            and self.end == other.end
+            _compare_optional(self.start, other.start)
+            and _compare_optional(self.end, other.end)
             and self.step == other.step
         )
 
@@ -148,7 +137,7 @@ struct Slice(Stringable, EqualityComparable):
             The length of the slice.
         """
 
-        return len(range(self.start, self.end, self.step))
+        return len(range(self.start.value(), self.end.value(), self.step))
 
     @always_inline
     fn __getitem__(self, idx: Int) -> Int:
@@ -160,11 +149,62 @@ struct Slice(Stringable, EqualityComparable):
         Returns:
             The slice index.
         """
-        return self.start + index(idx) * self.step
+        return self.start.value() + idx * self.step
 
-    @always_inline("nodebug")
-    fn _has_end(self) -> Bool:
-        return self.end != _int_max_value()
+    fn indices(self, src_len: Int) -> (Int, Int, Int):
+        """Returns a tuple of 3 intergers representing the start, end, and step
+           of the slice if applied to a container of the given length.
+
+        Uses the target container length to normalize negative, out of bounds, or None indices.
+
+        Negative indices are wrapped using the length of the container.
+        ```mojo
+        s = slice(0, -1, 1)
+        s.indices(5) # returns (0, 4, 1)
+        ```
+
+        None indices are defaulted to the start or the end of the container based
+        on whether `step` is positive or negative.
+        ```mojo
+        s = slice(None, None, 1)
+        s.indices(5) # returns (0, 5, 1)
+        ```
+
+        Out of bounds indices are clamped using the size of the container.
+        ```mojo
+        s = slice(20)
+        s.indices(5) # returns (0, 5, 1)
+        ```
+
+        Args:
+            src_len: The length of the target container.
+
+        Returns:
+            A tuple containing three integers for start, end, and step.
+        """
+        var start = self.start
+        var end = self.end
+        var positive_step = self.step > 0
+
+        if not start:
+            start = 0 if positive_step else src_len - 1
+        elif start.value() < 0:
+            start = start.value() + src_len
+            if start.value() < 0:
+                start = 0 if positive_step else -1
+        elif start.value() >= src_len:
+            start = src_len if positive_step else src_len - 1
+
+        if not end:
+            end = src_len if positive_step else -1
+        elif end.value() < 0:
+            end = end.value() + src_len
+            if end.value() < 0:
+                end = 0 if positive_step else -1
+        elif end.value() >= src_len:
+            end = src_len if positive_step else src_len - 1
+
+        return (start.value(), end.value(), self.step)
 
 
 @always_inline("nodebug")
@@ -194,17 +234,11 @@ fn slice(start: Int, end: Int) -> Slice:
     return Slice(start, end)
 
 
-# TODO(30496): Modernize the slice type
 @always_inline("nodebug")
-fn slice[
-    T0: AnyTrivialRegType, T1: AnyTrivialRegType, T2: AnyTrivialRegType
-](start: T0, end: T1, step: T2) -> Slice:
+fn slice(
+    start: OptionalReg[Int], end: OptionalReg[Int], step: OptionalReg[Int]
+) -> Slice:
     """Construct a Slice given the start, end and step values.
-
-    Parameters:
-        T0: Type of the start value.
-        T1: Type of the end value.
-        T2: Type of the step value.
 
     Args:
         start: The start value.
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -898,7 +898,6 @@ struct String(
         buf.append(0)
         return String(buf^)
 
-    @always_inline
     fn __getitem__(self, span: Slice) -> String:
         """Gets the sequence of characters at the specified positions.
 
@@ -908,18 +907,24 @@ struct String(
         Returns:
             A new string containing the string at the specified positions.
         """
-
-        var adjusted_span = self._adjust_span(span)
-        var adjusted_span_len = adjusted_span.unsafe_indices()
-        if adjusted_span.step == 1:
-            return StringRef(self._buffer.data + span.start, adjusted_span_len)
+        var start: Int
+        var end: Int
+        var step: Int
+        start, end, step = span.indices(len(self))
+        var r = range(start, end, step)
+        if step == 1:
+            return StringRef(
+                self._buffer.data + start,
+                len(r),
+            )
 
         var buffer = Self._buffer_type()
-        buffer.resize(adjusted_span_len + 1, 0)
+        var result_len = len(r)
+        buffer.resize(result_len + 1, 0)
         var ptr = self.unsafe_uint8_ptr()
-        for i in range(adjusted_span_len):
-            buffer[i] = ptr[adjusted_span[i]]
-        buffer[adjusted_span_len] = 0
+        for i in range(result_len):
+            buffer[i] = ptr[r[i]]
+        buffer[result_len] = 0
         return Self(buffer^)
 
     @always_inline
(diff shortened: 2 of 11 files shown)


mikowals:
Could these tests using out of bounds values be added in `test_string.mojo`?  There should probably be similar tests in `test_list.mojo`.  The new `Slice.adjust()` itself has no tests yet.

```
assert_equal("", str[-50::-1])
assert_equal("Hello Mojo!!", str[-50::]) # fails
assert_equal("!!ojoM olleH", str[:-50:-1]) # fails
assert_equal("Hello Mojo!!", str[:50:])  # fails
assert_equal("H", str[::50])
assert_equal("!", str[::-50])
assert_equal("!", str[50::-50]) # fails
assert_equal("H", str[-50::50]) # fails
```

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -158,11 +146,31 @@ struct Slice(Sized, Stringable, EqualityComparable):
...
+
+    fn adjust(inout self, src_len: Int):
+        """Adjusts span.
+
+        Uses target container length to adjust negative or None indices.
+
+        Args:
+            src_len: The length of the target container.
+        """
+        var negative_step = self.step < 0
```
mikowals, line 163: Could just be `positive_step` and replace `not negative_step` uses below.

stdlib/src/builtin/string.mojo
```diff
@@ -735,7 +722,7 @@ struct String(
         var adjusted_span = self._adjust_span(span)
         if adjusted_span.step == 1:
             return StringRef(
-                self._buffer.data + span.start,
+                self._buffer.data + span.start.value(),
```
mikowals, line 725: Should use the `adjusted_span` here?

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -158,11 +146,31 @@ struct Slice(Sized, Stringable, EqualityComparable):
...
+
+        if not self.start:
+            self.start = 0 if not negative_step else src_len - 1
+        elif self.start.value() < 0:
+            self.start = self.start.value() + src_len
```
mikowals, line 168: Need bounds checks.  

- Given a negative start and positive step, Python bounds `start >= 0` 
- Given a negative start and negative step,  Python bounds `start >= -1`
- Given a positive start and negative step, Python bounds `start <= src_len - 1`

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -158,11 +146,31 @@ struct Slice(Sized, Stringable, EqualityComparable):
...
+
+        if not self.end:
+            self.end = src_len if not negative_step else -1
+        elif self.end.value() < 0:
+            self.end = self.end.value() + src_len
```
mikowals, line 173: Need bounds checks.

- Given a negative end and positive step, Python bounds `end >= 0`
- Given a negative end and negative step, Python bounds `end >= -1`
- Given a positive end and positive step, Python bounds `end <= src_len`

bgreni:
> Could these tests using out of bounds values be added in `test_string.mojo`? There should probably be similar tests in `test_list.mojo`. The new `Slice.adjust()` itself has no tests yet.
> 
> ```
> assert_equal("", str[-50::-1])
> assert_equal("Hello Mojo!!", str[-50::]) # fails
> assert_equal("!!ojoM olleH", str[:-50:-1]) # fails
> assert_equal("Hello Mojo!!", str[:50:])  # fails
> assert_equal("H", str[::50])
> assert_equal("!", str[::-50])
> assert_equal("!", str[50::-50]) # fails
> assert_equal("H", str[-50::50]) # fails
> ```

@mikowals I actually wasn't aware Python had this behaviour, thank you kindly for the review! I've made the necessary changes so the previously failing cases now pass.

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -16,27 +16,17 @@ These are Mojo built-ins, so you don't need to import them.
...
-@always_inline("nodebug")
-fn _default_or[T: AnyRegType](value: T, default: Int) -> Int:
-    # TODO: Handle `__index__` for other types when we have traits!
-    @parameter
-    if _mlirtype_is_eq[T, Int]():
-        return __mlir_op.`kgen.rebind`[_type=Int](value)
+fn _compare_optional(x: OptionalReg[Int], y: OptionalReg[Int]) -> Bool:
+    if x and y:
+        return x.value() == y.value()
+    elif not x and not y:
```
laszlokindrat, line 26:
```suggestion
    return not x and not y
```

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -75,35 +65,33 @@ struct Slice(Sized, Stringable, EqualityComparable):
...
+        self.step = step.value() if step else 1
 
     fn __str__(self) -> String:
         """Gets the string representation of the span.
 
         Returns:
             The string representation of the span.
         """
-        var res = str(self.start)
+        var res = str(self.start.value()) if self.start else ""
```
laszlokindrat, line 91: Are there any tests for `slice.__str__`? If not, could you please add some?

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -158,11 +146,41 @@ struct Slice(Sized, Stringable, EqualityComparable):
...
         """
-        return self.start + idx * self.step
+        return self.start.value() + idx * self.step
 
     @always_inline("nodebug")
     fn _has_end(self) -> Bool:
-        return self.end != _int_max_value()
+        return self.end.__bool__()
+
+    fn adjust(inout self, src_len: Int):
```
laszlokindrat, line 155: I'm not sure if this should be public. @ConnorGray WDYT, could we just rename it to `_adjust`?

laszlokindrat (commented):
Thanks! I have a couple small comments and questions, but generally looks good!

stdlib/src/builtin/builtin_slice.mojo
bgreni: Yes there are https://github.com/modularml/mojo/blob/main/stdlib/test/builtin/test_slice.mojo#L78-L82

stdlib/src/builtin/builtin_slice.mojo
ConnorGray: I agree, this should probably be private.

It would also be great to see some more docs for this. The current description somewhat assumes I know what it means to "adjust" this slice index. Is this really more like a "normalization" or "canonicalization" kind of operation?

ConnorGray:
Thanks for taking this on @bgreni! It's awesome to see us using types like `Optional` to make APIs like this more precise and ergonomic 😄

stdlib/src/builtin/builtin_slice.mojo
bgreni: I've changed the function name and added more details to the description with some examples.

laszlokindrat (approved):
LGTM, thank you!

laszlokindrat:
@bgreni A lot has changed over the past two weeks, could you please rebase this and ensure it passes CI?

bgreni:
> @bgreni A lot has changed over the past two weeks, could you please rebase this and ensure it passes CI?

Sorry I figured the `imported-internally` tag meant it was out of my hands. I just had to fix `fn __getitem__(self, slice: Slice)` for `Span` since it was recently introduced but otherwise all good!

laszlokindrat:
!sync

laszlokindrat:
!sync

laszlokindrat:
I have been wrestling with CI about this, and discovered that internally we depend on incorrect behavior in multiple places. This is unfortunate, but on the way I also learned that Python's `slice` has an [`indices`](https://docs.python.org/3/reference/datamodel.html#slice.indices) method, which is similar to your `_adjust` method; at the same time, Python does _not_ have a `__len__` for slices, and I think this makes sense given the ambiguity. I will remove the `__len__` method in favor of a new `unsafe_indices`. Could you please open a separate PR just for a new `Slice.indices` method? Once these both land, we can circle back to the original patch.

laszlokindrat:
@bgreni I removed `Slice.__len__` so you can rebase and get those changes after the next nightly. Let me know if you can implement `Slice.indices`.

bgreni:
> @bgreni I removed `Slice.__len__` so you can rebase and get those changes after the next nightly. Let me know if you can implement `Slice.indices`.

Yeah I'll get on that today thanks!

bgreni:
@laszlokindrat Just to clarify is the intent to consolidate the current broken behaviour into a `Slice.indices` method that returns a `Range` or something with the adjusted start, end, and step values? Then afterwards we return here and integrate those two changes with the corrected behaviour present in this patch?

laszlokindrat:
> @laszlokindrat Just to clarify is the intent to consolidate the current broken behaviour into a `Slice.indices` method that returns a `Range` or something with the adjusted start, end, and step values? Then afterwards we return here and integrate those two changes with the corrected behaviour present in this patch?

You raise a good point. Since the current slice implementation cannot provide a correct `Slice.indices` method, you can just do it as part of this PR. Please make sure you mention it in the changelog, though! BTW the python API returns a tuple of three ints IIRC. Thanks again for working on this, this will be a huge improvement!

bgreni:
@laszlokindrat In working on that I realized slicing a `Span` is sort of ill-defined at the moment. Since it can't copy the data, negative and > 1 step values don't work as expected. I think it makes sense for `Span` to be sliceable though and I have an idea to give `Span` an internal `Range` member that could be used in the indexing behaviour to fix this.

Are you ok with leaving the current behaviour for now and I can fix that after this lands in a following patch?

bgreni:
@laszlokindrat I've updated everything to use the `indices()` semantics and I'll rebase on your changes when nightly time hits.

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -16,27 +16,14 @@ These are Mojo built-ins, so you don't need to import them.
...
-    @parameter
-    if _mlirtype_is_eq[T, Int]():
-        return __mlir_op.`kgen.rebind`[_type=Int](value)
-    else:
-        __mlir_op.`kgen.param.assert`[
-            cond = (_mlirtype_is_eq[T, NoneType]()).__mlir_i1__(),
-            message = "expected Int or NoneType".value,
-        ]()
-        return default
+fn _compare_optional(x: OptionalReg[Int], y: OptionalReg[Int]) -> Bool:
```
laszlokindrat, line 23: Non-blocking: If you are interested, maybe we could add `__eq__` and `__ne__` to `OptionalReg` using the conditional conformance trick.

laszlokindrat (commented):
This is really great, thank you so much! I have mostly quibbles and a couple questions before I bring this in, and I will wait for you to rebase. Please also mention in the changelog that Slice now uses Optional, since this is a user visible change that also brings us closer to Python.

> Are you ok with leaving the current behaviour for now and I can fix that after this lands in a following patch?

Absolutely, thanks again!

stdlib/src/utils/span.mojo
```diff
@@ -179,15 +179,17 @@ struct Span[
...
-        var adjusted_span = self._adjust_span(slice)
+        var start: Int
+        var end: Int
+        var step: Int
+        start, end, step = slice.indices(len(self))
         debug_assert(
-            0 <= adjusted_span.start <= self._len
-            and 0 <= adjusted_span.end <= self._len,
-            "Slice must be within bounds.",
+            0 <= start <= self._len and 0 <= end <= self._len and step != 1,
```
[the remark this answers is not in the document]
bgreni: Well its possible to get a negative value if you use a negative step values like in the following test case (which is consistent with Pythons behaviour)

```mojo
s = slice(1, None, -1)
start, end, step = s.indices(5)
assert_equal(slice(start, end, step), slice(1, -1, -1))
```
But granted since we already aren't allowed a negative step for now you are correct that the other checks are redundant.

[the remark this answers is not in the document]
laszlokindrat: Ok, thanks for clarifying. Please remove the redundant checks and rebase to resolve the conflicts. I'll bring it in after that.

stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -75,35 +62,33 @@ struct Slice(Sized, Stringable, EqualityComparable):
...
             start: The start value.
             end: The end value.
             step: The step value.
         """
-        self.start = _default_or(start, 0)
-        self.end = _default_or(end, _int_max_value())
-        self.step = _default_or(step, 1)
+        self.start = start
+        self.end = end
+        self.step = step.value() if step else 1
```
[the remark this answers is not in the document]
laszlokindrat: BTW, no need to fix this in this patch if we don't comply with this; followup is fine.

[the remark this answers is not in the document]
bgreni: Thanks for pointing that out, currently that results in unexpected behaviour (apparently we also don't throw an error when giving `range` a 0 step either).

That might be better as a follow up as well since having to mark all slice getters with `raises` feels like a pretty significant commitment even if it is in the name of being python compliant.

laszlokindrat:
!sync

stdlib/src/builtin/builtin_slice.mojo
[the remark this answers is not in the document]
laszlokindrat: Yeah, let's not raise from these, but try to do either `debug_assert` or check and `abort`.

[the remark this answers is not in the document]
bgreni: Ok in that case I've just added a check and abort in the constructor

bgreni:
@laszlokindrat  Do we have some way to re-invoke the tests when they fail due to current flakey ones?

laszlokindrat:
> @laszlokindrat Do we have some way to re-invoke the tests when they fail due to current flakey ones?

Not sure, maybe you can add an empty commit? We squash and merge internally, so the commit history on your branch is not important.

laszlokindrat:
I attempted this marshal this through again, but it seems our internal uses of slice deeply depend on the current (admittedly incorrect) behavior. My suggestion for a plan forward is the following:
1) Introduce a new, corrected slice type with a new name, e.g. `SliceNew` and functions like `slice_new`.
2) I will internally migrate uses of `Slice` that I can to this new type.
3) Rename the existing `Slice` to `SliceOld` and the new `SliceNew` to `Slice` (and similarly for `slice_old`/`slice_new`).
4) Mark `SliceOld` and `slice_old` deprecated.

WDYT?

bgreni:
@laszlokindrat Could be a little confusing for nightly users, but if resolving the internal issues is highly burdensome then that would probably be the best way forward?

JoeLoser:
> @laszlokindrat Could be a little confusing for nightly users, but if resolving the internal issues is highly burdensome then that would probably be the best way forward?

+1 to @laszlokindrat's approach.  It's the best approach forward for us internally, as we have a good amount of misuse of it across a variety of teams in the codebase.  It'd be good for us to do the incremental approach.  In general, it becomes increasingly more complex for our small standard library team to fix the whole company's misuse of certain types and features.  Instead, we need to build safe alternatives and incrementally migrate them at times — such as the case here.

bgreni:
Right thanks Joe I understand now. Do want to add an overload `__getitem__` for the types in this pr that accepts `SliceNew` so nightly users can get the correct behaviour even if they can't use the shorthand syntax for it yet?

laszlokindrat:
> Right thanks Joe I understand now. Do want to add an overload `__getitem__` for the types in this pr that accepts `SliceNew` so nightly users can get the correct behaviour even if they can't use the shorthand syntax for it yet?

Maybe, my main ask is that we do this incrementally, since it's clearly gonna be a non-trivial process. So let's just start with a new types + unit tests, then maybe some helpers and phase it in to builtins, and so on. It might also make sense to add implicit conversion between the new and old to help ease the transition.

bgreni:
@laszlokindrat I've opened a new PR with a subset of these changes using the name `SliceNew` #2894. I'll close this one but we can always refer back to it when the time comes for additional changes to be introduced.

laszlokindrat:
@bgreni Some heroics were pulled internally, and we were able to land this patch more or less as is. I will clean up `SliceNew` and we won't have to phase this in after all. Thanks for being patient with this!

bgreni:
@laszlokindrat Oh wow how exciting! Thank you for all your work on this!

PR #2502, opened 2024-05-04
[stdlib] Constrain `simd.bool()` to `size == 1`

helehex:
Use explicit `reduce_or()`/`reduce_and()` instead
see https://github.com/modularml/mojo/pull/2412

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -348,17 +348,22 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
 
     @always_inline("nodebug")
     fn __bool__(self) -> Bool:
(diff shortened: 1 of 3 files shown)


soraros:
Actually, I wish they were called `any` and `all`.

helehex:
i was thinking the same thing, but they can be separate to avoid bitwise ambiguity

JoeLoser:
Just a quick update here: this has been approved internally, but I need to go fix up more uses internally this breaks before we can land it.

JoeLoser (approved):

helehex:
I was about to try adding `any`/`all` functions, but my Internet connection just died, so if anyone wants something todo

helehex:
I thought there was something else that should get done before `any`/`all`, but I can't remember now

PR #2504, opened 2024-05-04
[tools] Check-licenses-failing

artemiogr97:
the pre-commit hook was failing, poiting to the script in relative fixed the issue for me


artemiogr97:
false alarm, I forgot to compile my checkout 🤦

PR #2506, opened 2024-05-04
[stdlib]: Allow ListLiteral as an initializer to List

lsh:
The goal of this PR is to make `ListLiteral` useful in a layout independent way. With this design, it should be possible to e.g. change ListLiteral to be:

```mojo
struct ListLiteral[T: CollectionElement, inferred size: Int]:
    var data: InlineArray[T, size]
```

### DIFF
--- a/stdlib/test/collections/test_list.mojo
+++ b/stdlib/test/collections/test_list.mojo
@@ -75,6 +75,24 @@ def test_list_to_bool_conversion():
     assert_true(List[String](""))
 
 
+def test_initialize_from_list_literal():
+    var a = List[Int]([1, 2, 3, 4])
+    assert_equal(a[0], 1)
+    assert_equal(a[1], 2)
+    assert_equal(a[2], 3)
+    assert_equal(a[3], 4)
+
+    a.extend([7, 8, 9])
+    assert_equal(a[4], 7)
+    assert_equal(a[5], 8)
+    assert_equal(a[6], 9)
+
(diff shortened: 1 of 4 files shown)


stdlib/test/collections/test_list.mojo
```diff
@@ -693,7 +711,7 @@ def main():
     test_mojo_issue_698()
     test_list()
     test_list_clear()
-    test_list_to_bool_conversion()
+    # test_list_to_bool_conversion()
```
lsh, line 714: This test fails with this initializer but I'm not sure why

helehex:
i remember doing this for some of my internal collections, and something about the AnyRegType stuff going on meant that i ended up removing it for now, but i dont remember exactly what

JoeLoser:
This is super exciting, but I'd recommend we hold off right this moment on it.  I talked with @Mogball about this the other day, and, as discussed offline.

Rationale: `ListLiteral` is currently homogenous in the elements it accepts, but it's backed by a `Tuple` which supports heterogeneous elements.  Soon. this representation will change — likely to a `pop.array` for constraining the element types in `ListLiteral` to be homogenous.  Given this layout change, I'd rather not have people start depending on this sort of thing as it will lead to surprises and adding users that we're not ready for quite yet.

helehex:
Dang I kinda liked that

helehex:
could we do a `@nonmaterializable` struct around `pop.array` right now?

PR #2507, opened 2024-05-04
[stdlib] Add small string optimization to the `String` struct

gabrieldemarmiesse:
Fix part of https://github.com/modularml/mojo/issues/2467

This PR will stay in draft as it's too big to be merged at once, I'll split it further into multiple PRs. I also have some cleanup to do and some benchmarking. But it can give the general idea of how it works.

In a nutshell:
* I created a type `_InlineBytesList` which is like `InlineArray` but specialized in `Int8` and has a size and capacity (list-like)
* I created a type `_BytesListWithSmallSizeOptimization` which has the interface of a List, but can fall back to stack allocated memory  with `_InlineBytesList` for small sizes.
* Replaced `List[Int8]` in `String` by `_BytesListWithSmallSizeOptimization`. Adapt a few lines here and there.

The tests are passing locally. I'm open to feedback as I work on the cleanup+split.

### Benchmark
See  https://github.com/modularml/mojo/issues/2467#issuecomment-2106263163

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -20,7 +20,7 @@ from sys import llvm_intrinsic, bitwidthof
 
 from memory import DTypePointer, LegacyPointer, UnsafePointer, memcmp, memcpy
 
-from utils import StringRef, StaticIntTuple
+from utils import StringRef, StaticIntTuple, StaticTuple, InlineArray, Variant
 from utils._format import Formattable, Formatter, ToFormatter
 
 from .io import _snprintf
@@ -147,7 +147,10 @@ fn _chr_ascii(c: Int8) -> String:
     Returns:
         A string containing a single character based on the given code point.
     """
-    return String(String._buffer_type(c, 0))
+    var buffer = String._buffer_type()
+    buffer.append(c)
+    buffer.append(0)
+    return String(_buffer=buffer^)
 
 
 @always_inline("nodebug")
@@ -516,7 +519,7 @@ struct String(
 ):
     """Represents a mutable string."""
 
-    alias _buffer_type = List[Int8]
+    alias _buffer_type = _BytesListWithSmallSizeOptimization
     var _buffer: Self._buffer_type
     """The underlying storage for the string."""
 
@@ -553,7 +556,7 @@ struct String(
 
     # TODO: Remove this method when #2317 is done
     @always_inline
-    fn __init__(inout self, owned impl: Self._buffer_type):
+    fn __init__(inout self, owned impl: List[Int8]):
         """Construct a string from a buffer of bytes.
 
         The buffer must be terminated with a null byte:
@@ -578,7 +581,7 @@ struct String(
             impl[-1] == 0,
             "expected last element of String buffer to be null terminator",
         )
-        self._buffer = impl^
+        self._buffer = _BytesListWithSmallSizeOptimization(impl^)
 
     @always_inline
     fn __init__(inout self, owned impl: List[UInt8]):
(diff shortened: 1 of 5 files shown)


stdlib/src/builtin/string.mojo
```diff
@@ -1549,3 +1537,173 @@ fn _calc_initial_buffer_size[type: DType](n0: Scalar[type]) -> Int:
             )
 
     return 128 + 1  # Add 1 for the terminator
+
+
+@value
+struct _InlineBytesList[capacity: Int](Sized, CollectionElement):
```
lsh, line 1543: **Thought** I'm not convinced this is a better solution than making `InlineFixedVector` use an `InlineVector` instead of `StaticTuple`, but that's also less trivial an update.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -400,3 +400,11 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
         return Reference(self)[]._get_reference_unsafe[mutability, self_life](
             normalized_idx
         )
+
+    fn get_storage_unsafe_pointer(self) -> UnsafePointer[ElementType]:
```
lsh, line 404: We explicitly chose not to reveal this in the safe API in the original PR. If you _really_ think we should be adding it, we should at least prefix it with an `_`, but the general stance is that getting an unsafe pointer is not recommended.

gabrieldemarmiesse: There are wayyyy too many places in the stdlib that uses a string pointer, so I don't mind removing it but this PR is not the place to do it.

Concerning the naming, I'm just following the instructions given in the contributing guide. There is nothing stating that unsafe functions should be private, but we should be explicit about the unsafe part (like in swift):
https://github.com/modularml/mojo/blob/main/stdlib/docs/vision.md#objectives-and-goals

stdlib/src/builtin/string.mojo
gabrieldemarmiesse: That's a great idea! I didn't know InlineFixedVector  existed actually. There is not much code to change and can be a good way to split this PR into multiple ones. Definitly in my TODO list.

stdlib/src/utils/static_tuple.mojo
JoeLoser: Drive-by comment: we have an internal PR which is harmonizing _all_ the "get an unsafe pointer" APIs to `unsafe_ptr()`.

stdlib/src/builtin/string.mojo
JoeLoser: FYI `InlineFixedVector` is going to switch over to `AnyType` or `CollectionElement` after we fix the compiler bug(s) blocking https://github.com/modularml/mojo/pull/2377

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: It's pretty rare in python to abbreviate words, but I guess that folks coming from low level languages too used to write "ptr" everywhere XD I don't want to bikeshed too much, but just saying that there is no reason to use "ptr" instead of "pointer"

JoeLoser: Right, I don't feel strongly on the eventual unified/final name.  Mostly care about unifying the various different namings we have now across the types.

JoeLoser:
https://github.com/modularml/mojo/commit/ea1f2a548ce3da593a6b87672c112a56a5b3f768 landed in today's nightly, so feel free to rebase with that if you'd like.

gabrieldemarmiesse:
FYI, we got huge slowdowns (x2 in some cases) in nightly for the benchmarks presented above. After the rebase, the numbers for my branch didn't change si it makes my SSO implementation look better XD

gabrieldemarmiesse:
Closing in favor of https://github.com/modularml/mojo/pull/2827

PR #2510, opened 2024-05-04
[stdlib] add `SIMD.Mask` alias

helehex:
adds an alias for a boolean simd vector, and applies it throughout the `simd.mojo` file.

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -144,6 +144,8 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
         size: The size of the SIMD vector.
     """
 
+    alias Mask = SIMD[DType.bool, size]
+
     alias element_type = type
     var value: __mlir_type[`!pop.simd<`, size.value, `, `, type.value, `>`]
     """The underlying storage for the vector."""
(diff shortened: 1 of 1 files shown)


helehex:
i'd prefer a top level alias like `alias Mask = SIMD[DType.bool, _]`, but that doesn't work right now (alias Mask = SIMD[DType.bool, _])
getting rid of the default value for the simd `size` parameter fixes it though.

stdlib/src/builtin/simd.mojo
```diff
@@ -144,6 +144,8 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
         size: The size of the SIMD vector.
     """
 
+    alias Mask = SIMD[DType.bool, size]
```
laszlokindrat, line 147: Can you please rename this to `BoolMask`?

soraros: I still wish it's a top level alias, and since it's not possible right now, shall we hold off for now?

laszlokindrat: The change as is does improve the readability of the implementation, so I think we should have it. Maybe we can just name it `_Mask` or something?

laszlokindrat: I'm gonna amend this change locally so that this becomes a hidden alias.

PR #2519, opened 2024-05-05
[stdlib] Add `Dict.get(key) -> Optional[T]` and  `Dict.get(key, default) -> T` methods

martinvuyk:
added both functions to keep compatibility with python
```
fn get(self, key: K) -> Optional[V]:
    ...

fn get(self, key: K, default: V) -> V:
    ...
```

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -612,6 +612,18 @@ struct Dict[K: KeyElement, V: CollectionElement](
             return ev.value()[].value
         return None
 
+    fn get(self, key: K) -> Optional[V]:
+        """Get a value from the dictionary by key.
+
+        Args:
+            key: The key to search for in the dictionary.
+
+        Returns:
+            An optional value containing a copy of the value if it was present,
+            otherwise an empty Optional.
+        """
+        return self.find(key)
+
     fn pop(inout self, key: K, owned default: Optional[V] = None) raises -> V:
         """Remove a value from the dictionary by key.
 
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -632,8 +632,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
             default: Default value to return.
 
         Returns:
-            An optional value containing a copy of the value if it was present,
-            otherwise an empty Optional.
+            A copy of the value if it was present, otherwise default.
         """
         return self.find(key).or_else(default)
 
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -624,6 +624,19 @@ struct Dict[K: KeyElement, V: CollectionElement](
         """
         return self.find(key)
 
+    fn get(self, key: K, default: V) -> V:
+        """Get a value from the dictionary by key.
+
+        Args:
+            key: The key to search for in the dictionary.
+            default: Default value to return.
+
+        Returns:
+            An optional value containing a copy of the value if it was present,
+            otherwise an empty Optional.
+        """
+        return self.find(key).or_else(default)
+
     fn pop(inout self, key: K, owned default: Optional[V] = None) raises -> V:
         """Remove a value from the dictionary by key.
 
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/dict.mojo
```diff
@@ -612,6 +612,30 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
+        """Get a value from the dictionary by key.
+
+        Args:
+            key: The key to search for in the dictionary.
+
+        Returns:
+            An optional value containing a copy of the value if it was present,
+            otherwise an empty Optional.
+        """
+        return self.find(key)
```
gabrieldemarmiesse, line 625: **suggestion** `find` is not a Python dict method. See https://docs.python.org/3/library/stdtypes.html#mapping-types-dict . What is does it playing the role of the `get()` method you just added. So maybe, to avoid having two methods doing the same thing, you could rename `find()` into `get()`?

stdlib/src/collections/dict.mojo
```diff
@@ -612,6 +612,30 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
+        """Get a value from the dictionary by key.
+
+        Args:
+            key: The key to search for in the dictionary.
+            default: Default value to return.
+
+        Returns:
+            A copy of the value if it was present, otherwise default.
+        """
+        return self.find(key).or_else(default)
```
gabrieldemarmiesse, line 637: **suggestion**
If you apply the first suggestion, then this line becomes:
```suggestion
        return self.get(key).or_else(default)
```

stdlib/src/collections/dict.mojo
```diff
@@ -612,6 +612,30 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
+        Args:
+            key: The key to search for in the dictionary.
+
+        Returns:
+            An optional value containing a copy of the value if it was present,
+            otherwise an empty Optional.
+        """
+        return self.find(key)
+
+    fn get(self, key: K, default: V) -> V:
```
gabrieldemarmiesse, line 627: `self.find(key)` has already unit tests. But `get(self, key: K, default: V) -> V` doesn't have unit tests yet. Maybe you can add a few unit tests for this function in https://github.com/modularml/mojo/blob/nightly/stdlib/test/collections/test_dict.mojo ?

gabrieldemarmiesse (commented):
Many thanks for this pull request! It should help Python users be immediately familiar with the `Dict` of Mojo :)

gabrieldemarmiesse:
Also, if there is an issue related to `Dict.get`, don't hesitate to mention it here :) that will create a link between this pull request and the issue, it will help the maintainers doing the bookkeeping

martinvuyk:
> self.find(key) has already unit tests. But get(self, key: K, default: V) -> V doesn't have unit tests yet. Maybe you can add a few unit tests for this function in https://github.com/modularml/mojo/blob/nightly/stdlib/test/collections/test_dict.mojo ?
I didn't think of adding unit tests since it would be essentially testing `Optional.or_else()`
> suggestion find is not a Python dict method. See https://docs.python.org/3/library/stdtypes.html#mapping-types-dict . What is does it playing the role of the get() method you just added. So maybe, to avoid having two methods doing the same thing, you could rename find() into get()?
I didn't read or make any proposals to change the API so I left it alone (I also think it could either be renamed or made a "private" method as _find)

gabrieldemarmiesse:
>I didn't think of adding unit tests since it would be essentially testing Optional.or_else()

You are right! That's also testing everything in the body of the function `fn get(self, key: K, default: V) -> V:`, which right now contains only `return self.find(key).or_else(default)`, but might contain other pieces of code in the future. Unit tests are also here to ensure that the next contributor doesn't break your code in the future. 

It's very common for project to have tests covering all of their public APIs. Which means that if we don't add tests now, the next contributor can add some logic, assume that unit tests are present and merge this because the test suite is passing while they added broken code. So we'll likely want unit tests for all the public api, even for very simple functions. You can take a look at other simple public functions in the Mojo stdlib, it's extremely likely they'll have unit tests.

martinvuyk:
Added a short and simple unit test, `Dict.find()` didn't have one. Bunched both (find & get) in the same test since they are aliases for each other

JoeLoser (approved):

PR #2522, opened 2024-05-05
[stdlib] Remove unused `_from_bytes` static method in `String`

gabrieldemarmiesse:
This PR is a small piece of https://github.com/modularml/mojo/pull/2507 . I'm trying to split it into multiple PRs.

Those methods are unused anyway, we can drop them.


JoeLoser (changes requested):
These functions are used by some internal MAX packages.  We can rework these call sites, but can you share some motivation if this is in the way of SSO?

JoeLoser:
> Well, I don't have to refactor code that doesn't exist :p I didn't know it was used internally. I'll close this PR then

Yep, no problem at all. When we import PRs, it's on us to fix up any internal things since they're invisible to you and others from this repo.

Issue #2515, opened 2024-05-04
[stdlib] Audit uses of `StaticIntTuple`


JoeLoser:
Similar in spirit to https://github.com/modularml/mojo/issues/2514, it would be nice to audit our use of `StaticIntTuple` in the standard library which is a mix of `InlineArray[Int, size]` with some odd APIs that make it represents dimensions or tuple-like things.

Where it makes sense, it would be good to simplify APIs to not be coupled to this unique/non-generic type.  For example, [this Int._divmod](https://github.com/modularml/mojo/blob/fd304d2f3f60a51816b8672cafdea41da35bd456/stdlib/src/builtin/int.mojo#L684) function should just return a `Tuple[Int, Int]` instead of using this unique type.  There's no reason to prefer `StaticIntTuple[2]` over `Tuple[Int, Int]` here.

Feel free to ask questions about specific APIs if you're not sure whether to change it or not.

YichengDWu:
I propose to use 'NTuple[2, Int]' as the type alias for 'Tuple[Int, Int]', and 'NTuple[N, T]' in general, so that it can be easily dispatch on 'N'.

JoeLoser:
> I propose to use 'NTuple[2, Int]' as the type alias for 'Tuple[Int, Int]', and 'NTuple[N, T]' in general, so that it can be easily dispatch on 'N'.

Can you elaborate on the value of the type alias?  I'm not sure I understand the full value yet.  Are there places you have in mind in the library that would benefit from this?

YichengDWu:
```mojo
NTuple[2, Int] = Tuple[Int, Int]
NTuple[3, Int] = Tuple[Int, Int, Int]
```
Or In general I should be able to write
```mojo
alias NTuple[N, T] = Tuple[VariadicPack[T, N]]
```

Then I can dispatch on `NTuple[5,Int]` instead of `Tuple[Int, Int, Int, Int ,Int]`, it's just shorter.

artemiogr97:
@JoeLoser  while working on this I found a potential issue, StaticIntTuple is imported in builtin/int.mojo which exposes it everywhere as all builtins, is exposing stuff in this way intentional?

gabrieldemarmiesse:
>@JoeLoser while working on this I found a potential issue, StaticIntTuple is imported in builtin/int.mojo which exposes it everywhere as all builtins, is exposing stuff in this way intentional?

[...]

PR #2525, opened 2024-05-05
[stdlib] Uses of static int tuple

artemiogr97:
Implementation for #2515


stdlib/src/builtin/int.mojo
```diff
@@ -692,14 +691,14 @@ struct Int(
...
+            return Tuple(0, 0)
         var div: Int = self._positive_div(rhs)
         if rhs > 0 and self > 0:
-            return StaticIntTuple[2](div, self._positive_rem(rhs))
+            return Tuple(div, self._positive_rem(rhs))
         var mod = self - div * rhs
         if ((rhs < 0) ^ (self < 0)) and mod:
-            return StaticIntTuple[2](div - 1, mod + rhs)
-        return StaticIntTuple[2](div, mod)
+            return Tuple[Int, Int](div - 1, mod + rhs)
```
gabrieldemarmiesse, line 700:
```suggestion
            return div - 1, mod + rhs
```
This should be enough. By default, values separated by commas means that it's a tuple. Same for the other returns

stdlib/src/builtin/simd.mojo
```diff
@@ -33,7 +33,7 @@ from utils._numerics import FPUtils
 from utils._numerics import isnan as _isnan
 from utils._numerics import nan as _nan
 from utils._visualizers import lldb_formatter_wrapping_type
-from utils import StaticTuple
+from utils import StaticTuple, StaticIntTuple
```
gabrieldemarmiesse, line 36: **thoughts** I supposed this is here otherwise the compiler will complain that this does not exist. This is not the first time that I see this bug about the compiler not importing the builtins automatically. But it's really hard to reproduce.

gabrieldemarmiesse (commented):
Thanks a lot for the pull request! That's a nice cleanup! Just a small simplification I found and it should be ok to merge this

stdlib/src/builtin/string.mojo
```diff
@@ -925,13 +925,10 @@ struct String(
             UnsafePointer.address_of(self).bitcast[NoneType](),
         )
 
-    fn join[rank: Int](self, elems: StaticIntTuple[rank]) -> String:
+    fn join(self, *elems: Int) -> String:
```
soraros, line 928: I think we should remove this after switching its call sites to use `Tuple`. Maybe just leave a TODO here.

stdlib/src/builtin/simd.mojo
artemiogr97: Is `StaticIntTuple`  supposed to be a builtin? it is defined in utils 
As I mentionned here it was exposed as a builtin from the imports in the int module: https://github.com/modularml/mojo/issues/2515#issuecomment-2094760396

gabrieldemarmiesse: Then it seems you are importing it but not using it in simd.mojo. But maybe the PR is incomplete

artemiogr97: it was already used in simd.mojo, so I had to add the import here (it was missing but working "thanks" to the import in int.mojo)

gabrieldemarmiesse: Yeah that's really not a good thing how builtin imports work right now. I'll open an issue.

stdlib/src/builtin/string.mojo
artemiogr97: I just tried removing it, and it works, should it remove it now?

Also I just realized that `String.join` does not match the python implementation of `str.join`, in python only one argument is expected, not a sequence of arguments

stdlib/src/builtin/simd.mojo
```diff
@@ -1620,17 +1620,8 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
             `(self+other)[permutation[i]]`.
         """
 
-        @parameter
```
artemiogr97, line 1623: I didn't realize that I deleted this check, but I cannot get it to work, keeping ` <= mask[i] < 2 * size` gives the following error: 
```
no matching method in call to '__refitem__'mojo
static_tuple.mojo(341, 8): candidate not viable: callee expects at least 2 positional parameters, but 0 were specified
static_tuple.mojo(372, 8): candidate not viable: expected at most 1 positional argument, got 2
```

artemiogr97: I saw that the reference changed a lot in the last nightly release so after updating the error changed to `cannot use a dynamic value in call parameter` which makes a lot more sense, but I'm still not sure how to add the compile time check again

gabrieldemarmiesse:
For your information, https://github.com/modularml/mojo/pull/2097 was just merged internally, which means that you'll have conflicts in the `_divmod` method. You might want to wait for the next nightly and then rebase your changes. Between this and the `Reference` changes, I expect you'll have a fair amount of conflicts. If you want, then,  I can lend a hand :)

artemiogr97:
@gabrieldemarmiesse I solved the conflicts for the `Reference` changes but I don't think I will have time to do much more these days, if you want I can give you write access to my fork and feel free to do a merge, otherwise I'll do it in the weekend.

Personally I'm a little bit more worried about the issue I mentioned above, I think that the check was important and maybe I'll have to revert the changes in simd.mojo to keep it.

artemiogr97:
yep, and it's because of those holidays that I won't be home these days, i just sent you the invite 😉

gabrieldemarmiesse:
Happy holiday :)

stdlib/src/builtin/simd.mojo
gabrieldemarmiesse: Ok so I dig this one a bit and it's a compiler bug. The good news is that I found a workaround, as the struct `Tuple` had the same problem (I was wondering how Tuple handled this). So I found this: https://github.com/modularml/mojo/blob/nightly/stdlib/src/builtin/tuple.mojo#L182 I used the same trick here and it worked. It seems the issue is tracked internally, so this problem should go away at some point.

gabrieldemarmiesse:
Hey @artemiogr97 , I had some time to kill:
* I fixed the conflicts, it's now all find for `_divmod`, which doesn't require changes in nightly
* Put back the compile-time check where you had some issues. See https://github.com/modularml/mojo/pull/2525#discussion_r1595590187
This PR should be ready for review :)

stdlib/src/builtin/string.mojo
JoeLoser: Let's save the refactoring of `join` for a different PR, please.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -392,3 +392,17 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
             normalized_idx += size
 
         return self[]._get_reference_unsafe(normalized_idx)
+
+    # TODO(#38268): Remove this method when references and parameter expressions
```
JoeLoser, line 396: **Note to self** file this bug externally so we can track it easier

JoeLoser (approved):
Great cleanup! Thank you for this!

JoeLoser:
> > @gabrieldemarmiesse I think @JoeLoser was talking about removing the method later, the branch already had the tag 'imported internally' see: [#2525 (comment)](https://github.com/modularml/mojo/pull/2525#discussion_r1590313416)
> 
> No worries, it's easy to update the internal PR. I'll re-import this. Thanks for the ping. We're working on internal tooling to notify when external changes have been made from imported internal PRs.
> 
> This is breaking some internal things relying on the specific IR of these stdlib types, but we're digging into it. Nothing needed from your part at all. 😄

I marked this PR as "blocked", but there's nothing for you to do.  There's some compiler issues that need to get sorted out internally before we can land this change, unfortunately.

artemiogr97:
@JoeLoser I see that there are some conflicts now with this PR in several files, maybe it would be a good idea to split it?, also that way the "safe"/working bits could be merged.

JoeLoser:
> @JoeLoser I see that there are some conflicts now with this PR in several files, maybe it would be a good idea to split it?, also that way the "safe"/working bits could be merged.

I'll take care of this, no worries — thank you for flagging this!  I want to split up/land as many as I can internally that aren't crashing the compiler.  It'll be easier for me to do this in your internal PR.

I'll take a look at this in about an hour from now.

JoeLoser:
@artemiogr97 I tracked down the crash a bit, and it's due to the changes in `simd.mojo`.  I haven't root-caused it yet, but I will when I get some more time this week.  I backed off those changes (and the introduction of `get` function in `static_tuple.mojo` as it was only used in `simd.mojo` in this PR.  This should unblock this PR from landing in some shape at least here very shortly assuming internal CI passes after I reworked this PR.

PR #2526, opened 2024-05-05
[stdlib] add `Defaultable` trait

helehex:
adds a `Defaultable` trait for describing a type which has a constructor with no arguments

### DIFF
--- a/stdlib/src/builtin/value.mojo
+++ b/stdlib/src/builtin/value.mojo
@@ -99,6 +99,40 @@ trait Copyable:
         ...
 
 
+trait Defaultable:
+    """The `Defaultable` trait describes a type with a default constructor.
+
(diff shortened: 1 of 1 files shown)


stdlib/src/builtin/value.mojo
```diff
@@ -99,6 +99,40 @@ trait Copyable:
...
+
+        fn __init__(inout self):
+            self.bleep = T1()
+            self.bloop = T2()
+
+    var pair = Pair[Foo["Aigis"], Foo["Max"]]()
+    ```
+    """
+
+    fn __init__(inout self):
```
jayzhan211, line 131: Can we have the name `default`

JoeLoser (approved):
LGTM!  We can bikeshed the name another time.  I'll bring it up in Monday's team meeting.

Issue #2528, opened 2024-05-05
[BUG] check-license failing sometimes


artemiogr97:
### Bug description

The pre-commit hook for license checking is failing sometimes.
I had not been able to understand when it happens, from the log it seems like my stdlib.mojopkg is not found but running an script in the same terminal by doing `mojo path/to/my/script` works (and the stdlib in my build directory is used).

### Steps to reproduce

just do `git commit`, I get the following log:

```
[INFO] Stashing unstaged files to /home/temo/.cache/pre-commit/patch1714911965-57965.
mojo-format..............................................................Passed
check-docstrings.........................................................Passed
check-license............................................................Failed
- hook id: check-license
- exit code: 1

/home/temo/repos/mojo/build/stdlib.mojopkg:0:0: error: attempting to skip 4 bytes when only 0 remain
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:14:1: error: unable to load package '/home/temo/repos/mojo/build/stdlib.mojopkg'
import sys
^
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:14:8: error: unable to locate module 'sys'
import sys
       ^
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:15:6: error: unable to locate module 'pathlib'
from pathlib import Path
     ^
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:19:17: error: use of unknown declaration 'String'
alias LICENSE = String(
                ^~~~~~
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:27:11: error: could not find an 'object' type
def main():
          ^
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:27:11: error: could not find an 'Error' type
def main():
          ^
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:27:5: error: could not find an 'object' type
def main():
    ^
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:27:5: error: unable to locate module 'builtin'
def main():
    ^
/home/temo/repos/mojo/stdlib/scripts/check-licenses.mojo:28:1: error: 
    target_paths = sys.argv()
^
/home/temo/.modular/pkg/packages.modular.com_nightly_mojo/bin/mojo: error: failed to parse the provided Mojo source module
```

### System information

```shell
- What OS did you do install Mojo on ?
WSL ubuntu 22.04.4 in windows 11
[...]

PR #2531, opened 2024-05-05
[docs] `entr` and pre commit issue

artemiogr97:


artemiogr97:
"solves" #2528

JoeLoser (approved):
Looks good. Do people like the entr section at all? If we're just documenting sharp edges with it, does it make sense to put it in the contributing guide at all?

artemiogr97:
personally I really like it, I've been using it since I started contributing some weeks ago, I don't know how I managed to avoid the issue for so long 🤷‍♂️

JoeLoser:

PR #2533, opened 2024-05-05
[stdlib] Add `FileHandle.__iter__()` with newline as separator

martinvuyk:
To add pythonesque functionality

For now it returns a `List[String]` which doesn't implment the `__next__` method so its `__iter__ `needs to be manually called
```mojo
var f = open("/tmp/example.txt", "r")
for line in f.__iter__().__iter__():
    print(line[])
```
needs generators to be able to be called directly once.

The future implementation could look something like
```mojo
alias eol = String("\n").as_bytes()[0]
var content = self.read_bytes()
var i = 0
while i < content.size:
    var new_i = content[i:].index(eol).or_else(content.size)
    yield StringRef(content[i:new_i])
    i = new_i
```

### DIFF
--- a/stdlib/src/builtin/file.mojo
+++ b/stdlib/src/builtin/file.mojo
@@ -450,6 +450,44 @@ struct FileHandle:
         """The function to call when entering the context."""
         return self^
 
+    fn __iter__(self) raises -> List[String]:
+        """Iterate over lines in the file separated by the
+        newline character: \\n.
+
+        Returns:
+            A List of the file lines.
+
+        Examples:
+
+        ```mojo
+        var f = open("/tmp/example.txt", "r")
+        # TODO: this is ugly stil, since List doesn't implement __next__
+        # but once we have generators the problem should go away
+        for line in f.__iter__().__iter__():
+            print(line[])
+        ```
+        """
+        # TODO: when generators are implemented this should be something like
+        # alias eol = String("\n").as_bytes()[0]
+        # var content = self.read_bytes()
+        # var i = 0
+        # while i < content.size:
+        #     var new_i = content[i:].index(eol).or_else(content.size)
+        #     yield StringRef(content[i:new_i])
+        #     i = new_i
+        return self.read().split("\n")
+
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/file.mojo
```diff
@@ -450,6 +450,44 @@ struct FileHandle:
...
+        # alias eol = String("\n").as_bytes()[0]
+        # var content = self.read_bytes()
+        # var i = 0
+        # while i < content.size:
+        #     var new_i = content[i:].index(eol).or_else(content.size)
+        #     yield StringRef(content[i:new_i])
+        #     i = new_i
+        return self.read().split("\n")
+
+    fn __reversed__(self) raises -> List[String]:
```
soraros, line 480: I don't think file should really be a bidirectional iterator.

soraros:
I think we should implement a iterator object like [`_Listiter`](https://github.com/modularml/mojo/blob/00528f4c44f799966b3da0a866a00e24370c4b59/stdlib/src/collections/list.mojo#L44) instead of using `List`.

martinvuyk:
>I don't think file should really have a bidirectional iterator.

deleted it

>I think we should implement a iterator object like [_Listiter](https://github.com/modularml/mojo/blob/00528f4c44f799966b3da0a866a00e24370c4b59/stdlib/src/collections/list.mojo#L44) instead of using List.

I'm trying to get it to work but I don't know how to "force" the association of the read content to the lifetime of the `FileHandle` itself
```mojo
fn __iter__[
    mutability: __mlir_type.`i1`, self_life: AnyLifetime[mutability].type
](self) raises -> _FileIter[String, mutability, self_life]:
    """Iterate over lines in the file separated by the
    newline character: \\n.

    Returns:
        An iterator of immutable references to the file lines.

    Examples:

    ```mojo
    var f = open("/tmp/example.txt", "r")
    for line in f:
        print(line[])
    ```
    """
    var items = self.read().split("\n")
    var ref = Reference[List[String], mutability, self_life](items)
    return _FileIter[String, mutability, self_life](0, ref)
```

PR #2537, opened 2024-05-05
FloatLiteral missing operators to conform Python float [wrong]

msaelices:
There are some missing operators in the `FloatLiteral` struct that we would need to implement to conform the Python `float` type:

* `__round__(ndigits)`
* `__mod__(rhs)`
* `__divmod__(rhs)`

This PR implement them and add some test to check the methods.


msaelices:
Closed as it's included commits from another PR I want to isolate here. This is the right PR to review: https://github.com/modularml/mojo/pull/2538

Issue #2543, opened 2024-05-06
[BUG] alias using `__refitem__` fails to compile with "value doesn't have a memory type"


mikowals:
### Bug description

Trying to use a struct that implements `__refitem__` in an alias gives a compiler error: `error: value doesn't have a memory type`. Related to [excluded InlineArray tests](https://github.com/modularml/mojo/pull/2294#issuecomment-2065916684) and #2372.

### Steps to reproduce

```mojo
from memory import UnsafePointer
 
 
@value
struct Foo:
    var data: UnsafePointer[Int]
 
    fn __init__(inout self):
        pass
 
    @always_inline
    fn __refitem__[
        mutability: __mlir_type.`i1`, self_life: AnyLifetime[mutability].type
    ](
        self: Reference[Self, mutability, self_life]._mlir_type,
        i: Int,
    ) -> Reference[Int, mutability, self_life]:
        var ptr = Reference(self)[].data
        return ptr[]
 
 
fn main():
    alias foo = Foo()
    _ = foo.__refitem__[__mlir_attr.`0: i1`, __lifetime_of(foo)](0)
```

Compiler output:

```console
error: value doesn't have a memory type
    _ = foo.__refitem__[__mlir_attr.`0: i1`, __lifetime_of(foo)](0)
                                             ^~~~~~~~~~~~~~~~~~
mojo: error: failed to parse the provided Mojo source module
```

Using the `foo[0]` shorthand is a different error. I assume this is correct that it can not infer parameters but the root cause is the error given above when all parameters are explicit.

```console
error: invalid call to '__refitem__': callee expects 2 parameters, but 0 were specified
    _ = foo[0]
        ~~~^~~
/Users/mikowals/projects/mojo/stdlib/scratch.mojo:12:8: note: function declared here
    fn __refitem__[
       ^
mojo: error: failed to parse the provided Mojo source module
```

### System information

```shell
- macOS Sonoma Version 14.4.1 (23E224)
- mojo 2024.5.323 (1d9276ea)
- modular 0.7.2 (d0adc668)
```

mikowals:
I don't think this issue is relevant any more with the removal of `__refitem__` and new `ref [lifetime]` approach to returning references.

Issue #2540, opened 2024-05-06
[BUG] `__getitem__(self, Slice)` with `__refitem__` gives callee expects 2 parameters, but 0 were specified


mikowals:
### Bug description

A problem in the `__getitem__` or `__refitem__` sugar syntax using brackets. Encountered implementing #2432.

### Steps to reproduce

```mojo
from memory import UnsafePointer
 
@value
struct Foo:
    var data: UnsafePointer[Int]
 
    fn __init__(inout self):
        pass
 
    @always_inline
    fn __refitem__[
        mutability: __mlir_type.`i1`, self_life: AnyLifetime[mutability].type
    ](
        self: Reference[Self, mutability, self_life]._mlir_type,
        i: Int,
    ) -> Reference[Int, mutability, self_life]:
        var ptr = Reference(self)[].data
        return ptr[]
 
    @always_inline
    fn __getitem__(self, span: Slice) -> Self:
        return Self()
 
    @always_inline
    fn __str__(self) -> String:
        return "foo"
 
 
fn main():
    var foo = Foo()
    print(foo[0:1:1]) # error: invalid call to '__refitem__': callee expects 2 parameters ...
    # print(foo.__getitem__(slice(0, 1, 1))) # this one works
```

### System information

```shell
- MacOS Sonoma 14.4.1 (23E224)
- mojo 2024.5.323 (1d9276ea)
- modular 0.7.2 (d0adc668)
```

Issue #2432, opened 2024-04-28
[stdlib] Add `__refitem__` to `List`


JoeLoser:
As the [TODO](https://github.com/modularml/mojo/blob/d344d178aec2b80c5772261b8a6e441202a752de/stdlib/src/collections/list.mojo#L516) suggests, rename `__get_ref` to be `__refitem__`.  Then, we can remove `List.__getitem__`.  This will require fixing up a bunch of callers that previously were using `List.__getitem__` to now dereferenced the returned `Reference` from `List.__refitem__`, e.g. `my_list[some_index][]`.

StandinKP:
I can take this up @JoeLoser

JoeLoser:
> I can take this up @JoeLoser

I appreciate the enthusiasm, but unfortunately this one may be blocked on https://github.com/modularml/mojo/issues/2441 as @mikowals started taking a look at this I think.

JoeLoser:
Blocked by https://github.com/modularml/mojo/issues/2441

StandinKP:
No problem. LMK if I should take up anything on priority or I can browse through some of the older issues and pick sth up @JoeLoser

JoeLoser:
> No problem. LMK if I should take up anything on priority or I can browse through some of the older issues and pick sth up @JoeLoser

Will certainly do! I'm planning on adding several GitHub issues to the public repo this week, so stay tuned 😄

StandinKP:
Perfect🔥

JoeLoser:
Removed blocked label since it appears to have been a user error. See my comment at https://github.com/modularml/mojo/issues/2441#issuecomment-2082732920

PR #2544, opened 2024-05-06
[stdlib] use `__refitem__` in List

mikowals:
attempts #2432.  Encountered a couple of issues: 

- #2543 when using List in an alias. Worked around by using a StaticTuple in place of List.  Was required in only one place.
- #2540 when using `vec[::1]` or any other Slice.  Can be worked around by `__getitem__(slice(...))` in many cases but I don't think we want to do that.  And that `slice` syntax requires the endpoints so it isn't a workaround in all cases.

For special attention during review, I changed explicit use of `__refitem__` in other modules like `Dict` to just use vanilla `foo[index]`. I think those calls existed to avoid copies which should be default behaviour with `__refitem__` wiring.  So `var ev = self._entries.__get_ref(index)[]` became `self.entries[index]`.  Where specific mutability and lifetimes were being passed I kept the longform use of `__refitem__`.

### DIFF
--- a/stdlib/src/builtin/dtype.mojo
+++ b/stdlib/src/builtin/dtype.mojo
@@ -727,7 +727,7 @@ fn _get_runtime_dtype_size(type: DType) -> Int:
     statically known dtypes. Instead, we have to perform a dispatch to
     determine the size of the dtype.
     """
-    alias type_list = List[DType](
+    alias type_list = StaticTuple[DType, 16](
         DType.bool,
         DType.int8,
         DType.uint8,
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -96,7 +96,7 @@ struct _DictEntryIter[
             else:
                 debug_assert(self.index >= 0, "dict iter bounds")
 
-            var opt_entry_ref = self.src[]._entries.__get_ref(self.index)
+            var opt_entry_ref = self.src[]._entries.__refitem__(self.index)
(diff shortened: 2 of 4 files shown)


stdlib/src/builtin/dtype.mojo
```diff
@@ -727,7 +727,7 @@ fn _get_runtime_dtype_size(type: DType) -> Int:
     statically known dtypes. Instead, we have to perform a dispatch to
     determine the size of the dtype.
     """
-    alias type_list = List[DType](
+    alias type_list = StaticTuple[DType, 16](
```
JoeLoser, line 730: **Suggestion** Use `InlineArray` instead.  We're about to mark `StaticTuple` as deprecated and soon remove it.

JoeLoser (approved):
Great, thank you! I just asked @Mogball to take a look at https://github.com/modularml/mojo/issues/2540 when he gets a chance.

stdlib/src/builtin/dtype.mojo
mikowals: InlineArray gets the error: "cannot use a dynamic value in alias initializer".  From the line `alias concrete_type = type_list[idx]`.  This is the same error `List` also gets with the nightly 2024.5.1002.  I assume because `InlineArray` and `List` make very similar use of `__refitem__`.

mikowals:
No longer planned as the functionality of `__refitem__` is now handled by `__getitem__ -> ref [lifetime] T`

Issue #2547, opened 2024-05-06
[Feature Request] Constrained[False], or alternative, should not cause a return expected error


Benny-Nottonson:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

# Example Code

```mojo
fn expand_type[Type: DType]() -> DType:
    @parameter
    if Type.is_floating_point():
        return DType.float64
    elif Type.is_signed():
        return DType.int64
    elif Type.is_integral():
        return DType.uint64
        
    constrained[False, "Type must be numeric"]()
```
The following code should NOT cause a "return expected at end of function with results" compilation error.

Along with this, there should really just be `unreachable`.

### What is your motivation for this change?

If Mojo is allowing compile time verification of inputs to a function, it makes sense that extra code to return an invalid value should not be required, as the code added can by nature never run.

Benny-Nottonson:
Messed up the PR a few times but it's done now

lattner:
This is behaving correctly.  You need to put the unreachable thing in the parameter if like this:
```
fn expand_type[Type: DType]() -> DType:
    @parameter
    if Type.is_floating_point():
        return DType.float64
    elif Type.is_signed():
        return DType.int64
    elif Type.is_integral():
        return DType.uint64
    else:    
        constrained[False, "Type must be numeric"]()
```

PR #2550, opened 2024-05-06
[stdlib] Add `unreachable` check instead of `constrained[False]`

Benny-Nottonson:
As mentioned in #2547, there should realistically be an `unreachable` function, to indicate when a segment of code can never be reached. 

For example
```mojo
fn one_or_two[value: Int]() -> Int:
    if value == 1:
        return 1
    elif value == 2:
        return 2
    unreachable()
```

The above code demonstrates a great use case, where the alternative now would be:

```mojo
fn one_or_two[value: Int]() -> Int:
    constrained[value == 1 or value == 2]()
    if value == 1:
        return 1
    return 2
```

This is problematic in my opinion as it creates two issues:
1. You are now checking `value == 1` two separate times, the only way to avoid this is creating an unnecessary variable.
2. You are reducing readability and harming code structure by leading with the entire body of a conditional.

More information and further suggestions found at https://github.com/modularml/mojo/issues/2547 

Signed-off-by: benny-nottonson <bennynottonson@gmail.com>


stdlib/src/builtin/constrained.mojo
```diff
@@ -33,3 +33,16 @@ fn constrained[cond: Bool, msg: StringLiteral = "param assertion failed"]():
         cond = cond.__mlir_i1__(), message = msg.value
     ]()
     return
+
+
+@always_inline("nodebug")
+fn unreachable[msg: StringLiteral = "unreachable code"]():
```
Mogball, line 39: This is an extremely misleading name for the function. It doesn't actually indicate unreachable code in the control-flow sense, which will lead to lots of misuse of the function. For example,


```mojo
fn make_true() -> Boo: return True

fn infinite_loop():
    while make_true():
        pass
    unreachable()
```

Will fail to compile

Benny-Nottonson: I think that's more of a compiler edge case than a misnamed function.

Mogball: The compiler can't interprocedurally reason about the function call and prune dead code in order to make the program "correct". That would make the optimization level affect whether code compiles

Mogball: Plus, this function doesn't actually model a divergent function call. Type checking will still complain that a function like

```
fn foo() -> Int:
    unreachable()
```

Doesn't return an `Int` as expected.

Benny-Nottonson: But when is an example when the above code would be functioning as intended? Why is there an unreachable in the first place?

Benny-Nottonson: > Plus, this function doesn't actually model a divergent function call. Type checking will still complain that a function like
> 
> 
> 
> ```
> 
> fn foo() -> Int:
> 
>     unreachable()
> 
> ```
> 
> 
> 
> Doesn't return an `Int` as expected.

I'd argue that is a valid error for that function

Mogball: The point is that the name "unreachable" suggests divergent control flow, which is not what the function implements. The function you implemented is a `static_assert(false)`, which is something else.

Benny-Nottonson: Do you have an alternative name that is more correct, or different implementation?

Brian-M-J:
> ```mojo
> contrained[value == 1 or value == 2]()
> ```

I assume you mean `constrained`?

Benny-Nottonson:
> > ```mojo
> > contrained[value == 1 or value == 2]()
> > ```
> 
> I assume you mean `constrained`?

my mistake, fixed

PR #2555, opened 2024-05-06
[stdlib] now `strip` can take `chars` argument and `String` now have useful aliases

ormeilu:
Two major changes:
1. `strip`, `lstrip` and `rstrip` now can remove custom chars
2. `String` now have useful aliases like Python's https://github.com/python/cpython/blob/0085c3ae8f067abd4f6540d0f6dd2fb13107618e/Lib/string.py#L24C1-L32C62

- Naming is identical to python's
- Defaults to whitespace ` \n\t\r\f\v` without casting to Int8
- Tests are updated


stdlib/src/builtin/string.mojo
```diff
@@ -1208,43 +1208,46 @@ struct String(
         res.append(0)
         return String(res^)
 
-    fn strip(self) -> String:
-        """Return a copy of the string with leading and trailing whitespace characters removed.
+    fn strip(self, chars: String = " \n\t\r\f\v") -> String:
```
gabrieldemarmiesse, line 1211: **suggestion** since the list of characters for whitespaces is fixed, we could define somewhere `alias WHITESPACES = " \n\t\r\f\v"` and use it in the default values to ensure the behavior is the same.

stdlib/src/builtin/string.mojo
```diff
@@ -1208,43 +1208,46 @@ struct String(
...
+            chars: A set of characters to be removed. Defaults to whitespace.
 
         Returns:
-          A copy of the string with no leading whitespace characters.
+          A copy of the string with no leading characters.
         """
 
         var l_idx = 0
-        while l_idx < len(self) and isspace(ord(self[l_idx])):
+        while l_idx < len(self) and self[l_idx] in chars:
```
gabrieldemarmiesse, line 1250: **thoughts** In the default use case, we might end up with a slower function. Before, it was checking that the byte was in a given interval (something like `return a < x < b`, see the implementation of `isspace()`) and now it's checking the value against one of the 5 elements in the list. With simd, the performance might be similar. 

I don't believe it needs to be changed right now with overloads for more performance, I'm just mentionning for the maintainers that we can improve in two areas:
* Ensure the same code path is taken for isspace and the default use of `rstrip`, `lstrip`, to ensure we don't have different behaviors
* Have a benchmarking tool to allow us to verify that we don't have regressions. Here it's not clear if we'll get a performance regression or not. (it might not be the focus this early so it might not be too important).

gabrieldemarmiesse (commented):
Many thanks for the PR, this is a nice addition! The tests look good :) Let me know if you have trouble with the DCO

stdlib/src/builtin/string.mojo
toiletsandpaper: Yeah, good idea. I thought about it, but forgot somehow. I'll probably add everything from https://github.com/python/cpython/blob/0085c3ae8f067abd4f6540d0f6dd2fb13107618e/Lib/string.py#L24C1-L32C62 now and will add additional tests

toiletsandpaper: Done.

gabrieldemarmiesse: Maybe you can add other the constants in a separate pull request? I believe it will be more controversial than just `whitespaces`. Especially since there are functions already in the codebase that check for lowercase, uppercase, etc and those functions do not currently use those constants, they use a more optimized algorithm.

We wouldn't want to block the merge of strip, rstrip and lstrip because of a discussion concerning those new constants.

stdlib/src/builtin/string.mojo
toiletsandpaper: Yep, that will definitely slow down checks a bit, but not so much. But for now it's the only readable and flexible solution at the same time I see.

gabrieldemarmiesse:
You might want to sign-off your commits or the DCO CI job will not pass. See https://github.com/modularml/mojo/pull/2555/checks?check_run_id=24632558478 for the instructions.

stdlib/src/builtin/string.mojo
gabrieldemarmiesse: Indeed I'm not asking for changes, it's just a note for the maintainers

stdlib/src/builtin/string.mojo
toiletsandpaper: I don't really know how to open different PR from same branch on GitHub.

About functions - python already have functions and constants too

toiletsandpaper: Btw, current implementation is using `String.WHITESPACE`

toiletsandpaper:
> You might want to sign-off your commits or the DCO CI job will not pass. See https://github.com/modularml/mojo/pull/2555/checks?check_run_id=24632558478 for the instructions.

Thx. I thought GPG auto-sign would be enough

stdlib/src/builtin/string.mojo
```diff
@@ -438,6 +438,17 @@ struct String(
...
+    """ Usefull string aliases. """
+    alias WHITESPACE = String(" \n\t\r\f\v")
+    alias ASCII_LOWERCASE = String("abcdefghijklmnopqrstuvwxyz")
+    alias ASCII_UPPERCASE = String("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+    alias ASCII_LETTERS = String.ASCII_LOWERCASE + String.ASCII_UPPERCASE
+    alias DIGITS = String("0123456789")
+    alias HEXDIGITS = String.DIGITS + String("abcdef") + String("ABCDEF")
+    alias OCTDIGITS = String("01234567")
+    alias PUNCTUATION = String("""!"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~""")
+    alias PRINTABLE = String.DIGITS + String.ASCII_LETTERS + String.PUNCTUATION + String.WHITESPACE
```
gabrieldemarmiesse, line 450: A note for the maintainers: I don't have an opinion here, but I'm just pointing out that in Python, those constants are lowercase: https://docs.python.org/3/library/string.html#string-constants . We just need to be aware of this fact before merging.

stdlib/src/builtin/string.mojo
```diff
@@ -361,10 +361,10 @@ fn isupper(c: Int8) -> Bool:
     Returns:
         True if the character is uppercase.
     """
-    return _is_ascii_uppercase(c)
+    return _is_ASCII_UPPERCASE(c)
```
gryznar, line 364: What is the purpose to break consistency and not to keep snake_case? This is internal functionality. There is no need to keep this aligned with Python. Same to below

stdlib/src/builtin/string.mojo
```diff
@@ -1453,7 +1467,7 @@ fn _toggle_ascii_case(char: Int8) -> Int8:
 
 fn _calc_initial_buffer_size_int32(n0: Int) -> Int:
     # See https://commaok.xyz/post/lookup_tables/ and
-    # https://lemire.me/blog/2021/06/03/computing-the-number-of-digits-of-an-integer-even-faster/
+    # https://lemire.me/blog/2021/06/03/computing-the-number-of-DIGITS-of-an-integer-even-faster/
```
gabrieldemarmiesse, line 1470: I believe that would break the URL here, the "search and replace" might have been a bit too quick 😄

stdlib/src/builtin/string.mojo
gryznar: I would prefer to keep them uppercase. But this is question to Joe Loser I think

stdlib/src/builtin/string.mojo
```diff
@@ -438,6 +438,17 @@ struct String(
     var _buffer: Self._buffer_type
     """The underlying storage for the string."""
 
+    """ Usefull string aliases. """
+    alias WHITESPACE = String(" \n\t\r\f\v")
+    alias ASCII_LOWERCASE = String("abcdefghijklmnopqrstuvwxyz")
+    alias ASCII_UPPERCASE = String("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+    alias ASCII_LETTERS = String.ASCII_LOWERCASE + String.ASCII_UPPERCASE
+    alias DIGITS = String("0123456789")
+    alias HEXDIGITS = String.DIGITS + String("abcdef") + String("ABCDEF")
```
gryznar, line 447: If we would not follow Python, this can be renamed to: HEX_DIGITS and OCTDIGITS -> OCT_DIGITS

stdlib/src/builtin/string.mojo
gabrieldemarmiesse: You could make two branches if you want. https://github.com/modularml/mojo/pull/2555#discussion_r1591073450 , https://github.com/modularml/mojo/pull/2555#discussion_r1591067735 show that those new constants are more controversial than the strip function improvements. It could have value to separate those two changes (one PR for String.whitespace + strip,rstrip,lstrip and one PR for all the new constants) otherwise everything will be blocked until all discussions have been resolved. (it's just a tip from an old OSS dude to get your PRs merged faster, you're not obligated to follow it at all)

stdlib/src/builtin/string.mojo
```diff
@@ -361,10 +361,10 @@ fn isupper(c: Int8) -> Bool:
     Returns:
         True if the character is uppercase.
     """
-    return _is_ascii_uppercase(c)
+    return _is_ASCII_UPPERCASE(c)
 
 
-fn _is_ascii_uppercase(c: Int8) -> Bool:
+fn _is_ASCII_UPPERCASE(c: Int8) -> Bool:
```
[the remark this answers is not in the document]
toiletsandpaper: That was accidental by using `change all occurences` instead of normal renaming in my IDE.
Good catch, thx!

stdlib/src/builtin/string.mojo
toiletsandpaper: That was accidental by using `change all occurences` instead of normal renaming in my IDE.
Good catch, thx!

stdlib/src/builtin/string.mojo
toiletsandpaper: That was accidental by using `change all occurences` instead of normal renaming in my IDE.
Good catch, thx!

stdlib/src/builtin/string.mojo
toiletsandpaper: Yep, looks good to me, because we're already breaking python naming here, we can change this as well.

toiletsandpaper:
Am I missing something that needs to be done or it's just "sit and wait" part?

stdlib/test/builtin/test_string.mojo
```diff
@@ -664,8 +721,19 @@ fn test_lstrip() raises:
...
+    var str3 = String("mississippi")
+    assert_true(str3.lstrip("mis") == "ppi")
+
+    var str4 = String(" \n mississippimississippi")
+    assert_true(str4.lstrip("mis ") == "\n mississippimississippi")
+    assert_true(str4.lstrip("mis \n") == "ppimississippi")
+
 
 fn test_strip() raises:
+    print("start strip")
```
JoeLoser, line 734: **Suggestion** Remove this `print`.  It's not needed for anything.

stdlib/test/builtin/test_string.mojo
```diff
@@ -631,7 +631,55 @@ fn test_isspace() raises:
...
+    assert_true(String(",") in String.PUNCTUATION)
+    assert_true(String(".") in String.PUNCTUATION)
+    assert_true(String("\\") in String.PUNCTUATION)
+    assert_true(String("@") in String.PUNCTUATION)
+    assert_true(String('"') in String.PUNCTUATION)
+    assert_true(String("'") in String.PUNCTUATION)
+
+    var text = String("I love my Mom and Dad so much!!!\n")
+    for i in range(len(text)):
+        assert_true(text[i] in String.PRINTABLE)
```
JoeLoser, line 678: **Suggestion** This is good, but do you mind also adding a negative test (i.e. testing for a character that is not printable)?

JoeLoser (approved):
This is a great addition, thank you!

stdlib/test/builtin/test_string.mojo
toiletsandpaper: Done :)

stdlib/test/builtin/test_string.mojo
toiletsandpaper: Done :)

JoeLoser: Thanks! I just re-imported this internally and it should land here in a little bit.

JoeLoser:
Merged today in https://github.com/modularml/mojo/commit/682f225b164c661ea79ccc056346d662ef3e5d54 - thanks for the contribution!

PR #2563, opened 2024-05-07
[mojo-stdlib] Used explicit str in `float_literal.mojo`

StandinKP:
Changed for #2169

### DIFF
--- a/stdlib/src/builtin/float_literal.mojo
+++ b/stdlib/src/builtin/float_literal.mojo
@@ -135,7 +135,7 @@ struct FloatLiteral(
         Returns:
             A string representation.
         """
-        return self
+        return str(self)
 
     @always_inline("nodebug")
     fn __int_literal__(self) -> IntLiteral:


soraros:
@JoeLoser this is where I encountered the super confusing implicit conversion behaviour. Shall we convert to runtime type first?

StandinKP:
@soraros what do you exactly mean by converting to runtime types?

soraros:
I meant that the `__str__` method seems to rely on implicit conversion to concrete type `Float64` to work. Your current implementation will diverge if I understand correctly. We might need to make it more explicit (and correct) by using the following:
```mojo
  return str(Float64(self))
```
I'm not entirely sure it's the right fix, so let's wait for the team's input.

StandinKP:
Got it. Let's wait for the teams input and I can make the changes if required

stdlib/src/builtin/float_literal.mojo
```diff
@@ -135,7 +135,7 @@ struct FloatLiteral(
         Returns:
             A string representation.
         """
-        return self
```
JoeLoser, line 138: @laszlokindrat can you help me understand how this was working before?  Some weird materialization or implicit conversion rules?  I like being explicit in general, but it's still a bit mysterious how this works both before and after this change.

laszlokindrat: I suspect it would implicitly convert to SIMD which is Stringable and would be implicitly converted through that to `String`. But then I don't really understand how this current change fixes it. Either way @StandinKP could you please make this implementation to just return `Float64(self)` instead?

soraros: What about [this](https://github.com/modularml/mojo/pull/2563#issuecomment-2099184126)

laszlokindrat:
Will be done differently in the next nightly. Regardless, thanks for the patch, we're looking forward to more of your contributions!

PR #2564, opened 2024-05-07
[mojo-stdlib] Used explicit str in `int_literal.mojo`

StandinKP:
Changed for #2169


soraros:
@laszlokindrat FYI, this also needs the `str(Int(self))` treatment.

laszlokindrat:
Thanks! I missed today's nightly with this, but I will fix this slightly differently for the next release.

Issue #2514, opened 2024-05-04
[stdlib] Refactor uses of `StaticTuple` to use `InlineArray`


JoeLoser:
We recently got a generic array type (`InlineArray`) (see [here](https://github.com/modularml/mojo/blob/fd304d2f3f60a51816b8672cafdea41da35bd456/stdlib/src/utils/static_tuple.mojo#L251)) that works on memory-only types rather than `AnyRegType` like `StaticTuple` works on.  So, conceptually, `InlineArray` can replace all of these uses.

This issue tracks doing the work of replacing uses of `StaticTuple` with `InlineArray` so we can eventually remove `StaticTuple` and then re-home the `InlineArray` into the `collections` module where it belong.

vidur2:
Do you have any idea how to run the testing suite? Is the 'Test Examples' workflow comprehensive?

ConnorGray:
Hi @artemiogr97, re. your second question:

> in collections.vector, utils.inlined_string StaticTuple is instantiated with uninitialized values which is forbidden for InlinedArray

Yes, it would be good for us to preserve a way to create uninitialized `InlinedArray` values.

We may want to go a different direction down the line, but I think your suggestion for a `__init__(*, uninitialized: Bool)` seems like a reasonable solution to try for the time being🙂

JoeLoser:
Fixed with https://github.com/modularml/mojo/commit/6af1fa7f919fce883fa88ba3455807cd275ca7a0.  Thank you!

PR #2583, opened 2024-05-08
[stdlib] remove uses of StaticTuple where possible

artemiogr97:
Fix for #2514


artemiogr97:
some details here: https://github.com/modularml/mojo/issues/2514#issuecomment-2096561621 and here: https://github.com/modularml/mojo/issues/2514#issuecomment-2099531666

JoeLoser (approved):
Nice. Like the approach of the keyword argument for still allowing the uninitialized case. Thank you!

Issue #474, opened 2023-07-23
[Feature Request] Flesh out Base64 module


lsh:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Right now the Base64 module only has `b64encode`.

Ideally we'd at least have

```
alias Bytes = DynamicVector[UInt8] # in lieu of a more specific type
fn b64decode(bytes: borrowed Bytes) -> Bytes
fn b64encode(bytes: borrowed Bytes) -> String
fn b16decode(bytes: borrowed Bytes) -> Bytes
fn b16encode(bytes: borrowed Bytes) -> String
```

### What is your motivation for this change?

These methods are foundational for working with Cryptography.

### Any other details?

I think Base64 is not a great name for this module.

Consider:
```
import Base64

# redundant
Base64.b64encode(bytes)

# strange
Base64.b16encode(bytes)
```
I think simply `Encodings` would be a good name.

drunkwcodes:
But `encoding` in python is usually used as string to bytes encoding, such as utf-8, [GB 2312](https://en.wikipedia.org/wiki/GB_2312), [Big5](https://en.wikipedia.org/wiki/Big5) etc.

`b16encode` is also in `base64` module in [python](https://docs.python.org/3/library/base64.html), so I think it's fine to add those functions in `Base64` module in mojo.

lsh:
> But encoding in python is usually used as string to bytes encoding, such as utf-8, [GB 2312](https://en.wikipedia.org/wiki/GB_2312), [Big5](https://en.wikipedia.org/wiki/Big5) etc.

I'm not sure this is a particularly compelling argument. Python also uses `errors` as a common key word argument name, but I still think the `Error` module is descriptively named.

But I also don't want to distract from the main issue, and it was more of an offhand comment, so it might be best to move this somewhere else.

drunkwcodes:
@lsh I can not recall `errors` kwarg in python standard libraries. But errors are errors, there are no ambiguity.
Character encoding and base64 encoding are almost two totally different categories.

This kind of incompatibility with python is a non-starter for me. C'mon, SUPERSET!
So `Base64` is the best name for the module.

The module names with PascalCase are confusing me that they seem to be callable, like python classes. But this is trivial.

If this feature is pending to post-release, I'm willing to contribute to this module.

gryznar:
[...]

PR #2584, opened 2024-05-08
[mojo-stdlib] Add b16encode and b16decode to base64 module

kernhanda:
This change adds `b16encode` and `b16decode` to the `base64` module, as well as relevant tests.

The initial implementation is simple and can be optimized in future iterations (if needed).

#474


stdlib/src/base64/base64.mojo
```diff
@@ -151,3 +151,85 @@ fn b64decode(str: String) -> String:
...
+    """
+    alias lookup = "0123456789ABCDEF"
+    var b16chars = lookup.data()
+
+    var length = len(str)
+    var out = List[Int8](capacity=length * 2 + 1)
+
+    @parameter
+    @always_inline
+    fn s(idx: Int) -> Int:
```
gryznar, line 178: Question to maintainers. I am aware, that this is consistent with the other implementations here, but this name is not meaningful. How about adjusting that?

kernhanda: Are you asking about the `b16encode`/`b16decode` names or the local functions (in your snippet, `s`)?

kernhanda: done

PR #2587, opened 2024-05-08
[stdlib] Add `InlineList` struct (stack-allocated List)

gabrieldemarmiesse:
This struc is very useful to implement SSO, it's related to 
* https://github.com/modularml/mojo/issues/2467
* https://github.com/modularml/mojo/pull/2507

If this is merged, I can take advantage of this in my PR that has the SSO POC

About `InlineFixedVector`: `InlineList` is different. Notably, `InlineList` have its capacity decided at compile-time, and there is no heap allocation (unless the elements have heap-allocated data of course).

`InlineFixedVector` stores the first N element on the stack, and the next elements on the heap. Since not all elements are not in the same spot, it makes it hard to work with pointers there as the data is not contiguous.


stdlib/src/collections/inline_list.mojo
```diff
@@ -0,0 +1,100 @@
...
+from utils import InlineArray
+
+# ===----------------------------------------------------------------------===#
+# InlineList
+# ===----------------------------------------------------------------------===#
+
+
+# TODO: Provide a smarter default for the capacity.
+struct InlineList[ElementType: CollectionElement, capacity: Int = 16](Sized):
+    """A list allocated on the stack with a maximum size knwon at compile time.
```
JoeLoser, line 31:
```suggestion
    """A list allocated on the stack with a maximum size known at compile time.
```

stdlib/src/utils/static_tuple.mojo
```diff
@@ -276,6 +276,28 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
         ]()
         self._array = __mlir_op.`kgen.undef`[_type = Self.type]()
 
+    @always_inline
+    fn __init__(inout self, *, unsafe_uninitialized: Bool):
```
JoeLoser, line 280: FYI we already have this on top-of-tree nightly [here](https://github.com/modularml/mojo/blob/668489eee3a48564a453077d42421f290dcb4d14/stdlib/src/utils/static_tuple.mojo#L280).  Feel free to improve the docs though as yours are nice! 😄

stdlib/test/collections/test_inline_list.mojo
```diff
@@ -0,0 +1,94 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo %s
+
+from collections import InlineList, Set
+from testing import assert_equal, assert_false, assert_true, assert_raises
+
+
+def test_mojo_issue_698():
```
JoeLoser, line 19: **Question** Is this GH issue relevant for your new type?  It's not obviously a "yes" for me, since we append in the following test, so I'd just omit this if we can't justify what value it's adding.

stdlib/test/collections/test_inline_list.mojo
```diff
@@ -0,0 +1,94 @@
...
+    var destructor_counter = List[Int]()
+    var inline_list = InlineList[ValueToCountDestructor, capacity=32]()
+
+    var nb_elements_to_add = 8
+    for index in range(nb_elements_to_add):
+        inline_list.append(
+            ValueToCountDestructor(index, UnsafePointer(destructor_counter))
+        )
+
+    # Using .append() should trigger a move and not a copy+delete.
```
JoeLoser, line 78: 👍 would it be easier if we wrote this in terms of `MoveCounter` for counting number of moves/copies for each append call, or is that at the container/`InlineList` level?  See https://github.com/modularml/mojo/blob/668489eee3a48564a453077d42421f290dcb4d14/stdlib/test/test_utils/types.mojo

JoeLoser (commented):
Looks great!  Just a few minor comments and a rebase is needed to fixup the constructor added to `InlineArray`.

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: I updated the docstring without changing the function.

I think we can rename it "unsafe_uninitialized" in another PR. It's very unsafe so it might be worth it to be explicit about it.

stdlib/test/collections/test_inline_list.mojo
gabrieldemarmiesse: Thanks for the advice, I split this test into two and made the destructor test more robust. Let me know if it's an invalid way to test the destructor

stdlib/test/collections/test_inline_list.mojo
gabrieldemarmiesse: Indeed, not relevant, I was a bit too quick on the copy-pasting. I removed it.

JoeLoser:
Landed in today's nightly: https://github.com/modularml/mojo/pull/2615. Thanks for the great contribution! Excited to see this applied to your SSO work.

PR #2590, opened 2024-05-08
[stdlib] Support `__add__` and  `__mul__` operators for `List`

### LINKED ISSUE #2589: [Feature Request] Support `__add__` and `__mul__` operators for `List`
### Review Mojo's priorities

### What is your request?

We should support the `__add__` and `__mul__` operators for the standard `List` container. Where `__add__` is used to concatenate two lists, and `__mul__` is used to concat a list with itself 1 or more times.

This behaviour is present in Python, and therefore, should also be present in Mojo.

### Any other details?

_No response_

bgreni:
Fixes #2589

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -244,6 +244,75 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             later_idx -= 1
 
     @always_inline
+    fn __mul(inout self, x: Int):
+        """Appends the original elements of this list x-1 times.
+
+        ```mojo
+        var a = List[Int](1, 2)
+        a.__mul(2) # a = [1, 2, 1, 2]
+        ```
+
+        Args:
+            x: The multiplier number.
+        """
+        if x == 0:
+            self.clear()
+            return
+        var orig = List(self)
+        self.reserve(len(self) * x)
+        for i in range(x - 1):
+            self.extend(orig)
+
+    @always_inline("nodebug")
+    fn __mul__(self, x: Int) -> Self:
+        """Multiplies the list by x and returns a new list.
+
+        Args:
+            x: The multiplier number.
+
+        Returns:
+            The new list.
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/list.mojo
```diff
@@ -239,6 +239,73 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        var a = List[Int](1, 2)
+        a.__mul(2) # a = [1, 2, 1, 2]
+        ```
+
+        Args:
+            x: The multiplier number.
+        """
+        if x == 0:
+            self.clear()
+        var orig = List(self)
```
soraros, line 256: Should probably create a `List` with precomputed capacity to prevent allocation.

stdlib/src/collections/list.mojo
```diff
@@ -239,6 +239,73 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+    fn __imul__(inout self, x: Int):
+        """Multiplies the list by x in place.
+
+        Args:
+            x: The multiplier number.
+        """
+        self.__mul(x)
+
+    @always_inline("nodebug")
+    fn __add__(self, owned other: Self) -> Self:
```
soraros, line 287: I think this should return a new list. Without conditional conformance, I guess we can't achieve the desired semantics as it requires `Copy`.

stdlib/src/collections/list.mojo
```diff
@@ -239,6 +239,75 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        """Concatenats self with other and returns the result as a new list.
```
cgyurgyik, line 290:
```suggestion
        """Concatenates self with other and returns the result as a new list.
```

JoeLoser (approved):
Looks good, thank you!

JoeLoser:

PR #2595, opened 2024-05-09
[stdlib] Introduce non owning collection type

lsh:
This PR introduces a non owning collection type for contiguous arrays of `CollectionElement`. With this type, we will be able to tie together the array types in the standard library, and reduce necessary copies in some places.

### Motivation

The pointer & length data structure is common across most mainstream programming languages at this point ([Zig](https://ziglang.org/documentation/master/#Slices), [Rust](https://doc.rust-lang.org/std/primitive.slice.html), [C++](https://en.cppreference.com/w/cpp/container/span)). In order to mirror APIs developers are used to with the performance characteristic they expect, this type will be required.

Notably, it now allows functions that can work on `List` _or_ `Array` without an overload:
```mojo
fn copy[
    T: CollectionElement, lifetime: MutLifetime
](owned dst: Span[T, __mlir_attr.`1: i1`, lifetime], src: Span[T, _, _],):
    for i in range(len(src)):
        dst[i] = src[i]


fn main():
    var l = List[Int](1, 2, 3)
    var a = InlineArray[Int, 3](4, 5, 6)
    copy(Span(l), Span(a))
    for i in l:
        print(i[])
```
This code keeps the `List` and `InlineArray` alive long enough for the copy to work, which is not true for `Buffer` or `DTypePointer`.

This type will be especially necessary for traits such as `Read` and `Write`, which will want to take a `Span` as the relevant buffer, since taking a `List` or `Array` is use case dependent for such functions.

Another motivation for this change is to reduce unnecessary copies on methods such as slicing a `List`, which currently allocates a new list. Ideally, we would just return a `Span` without making a copy. The same could be said for `String.as_bytes()`, which does not need to allocate a `List`.

**Note about naming**

I'm using the name `Span` to mirror C++'s `std::span<T>`. Rust, Go, and Zig call this type a `slice`, but that term has an overloaded meaning in Python. It's also nice that it's short as a name. Similarly, Mojo already has the `Buffer` type. I'm open to other options if `Span` is not the right fit.

### DIFF
--- a/stdlib/src/utils/span.mojo
+++ b/stdlib/src/utils/span.mojo
@@ -111,6 +111,9 @@ struct Span[
     ](inout self, array: Reference[InlineArray[T, size], is_mutable, lifetime]):
         """Construct a Span from an InlineArray.
 
+        Parameters:
+            size: The size of the InlineArray.
+
         Args:
             array: The array to which the span refers.
         """
@@ -195,13 +198,16 @@ struct Span[
         ref[] = value
 
     @always_inline
-    fn __getitem__(self, span: Slice) -> Self:
+    fn __getitem__(self, slice: Slice) -> Self:
         """Get a new span from a slice of the current span.
+
+        Args:
+            slice: The slice specifying the range of the new subslice.
 
         Returns:
             A new span that points to the same data as the current span.
         """
-        var adjusted_span = self._adjust_span(span)
+        var adjusted_span = self._adjust_span(slice)
         debug_assert(
             0 <= adjusted_span.start < self._len
             and 0 <= adjusted_span.end < self._len,
--- a/stdlib/src/utils/span.mojo
+++ b/stdlib/src/utils/span.mojo
@@ -81,8 +81,8 @@ struct Span[
         lifetime: The lifetime of the Span.
     """
 
-    var data: UnsafePointer[T]
-    var len: Int
+    var _data: UnsafePointer[T]
+    var _len: Int
 
     @always_inline
     fn __init__(inout self, *, data: UnsafePointer[T], len: Int):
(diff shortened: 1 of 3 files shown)


stdlib/src/utils/span.mojo
```diff
@@ -0,0 +1,222 @@
...
+        """Unsafe construction from a pointer and length.
+
+        Args:
+            data: The underlying pointer of the span.
+            len: The length of the view.
+        """
+        self.data = data
+        self.len = len
+
+    @always_inline
```
lsh, line 98: Note that we probably want to get rid of this constructor in favor of returning it from `__getitem__(self: Reference[Self, _, _], slice: Slice) -> Span[T, _ , _]`

bgreni:
I'm a big fan of this addition! One thing to note is hopefully #2495 will be going in soon, which fixes the currently broken slice indexing logic that you've copied over into this new type.

helehex:
I like this, it would be very nice. Hopefully we can resolve [this](https://github.com/modularml/mojo/pull/2544) soon aswell. You can kind of get `collection.__refitem__(slice)` working by adding `__mlir_ref__` to the span/iter type, but it's not ideal.

stdlib/src/utils/span.mojo
JoeLoser: That's totally fine to do in a follow-up though IMO.  Let's introduce the new type and then apply its use throughout the library.

JoeLoser:
Landed in today's nightly: https://github.com/modularml/mojo/pull/2615

Thanks for the awesome contribution, Lukas! Happy to see it get applied in APIs soon in the library.

PR #2600, opened 2024-05-09
[stdlib] Add builtin `any()`/`all()` functions

helehex:
Adds builtin functions for testing if there are truthy elements in `List`, `Set`, `SIMD`.
Should change to check iterators eventually.

Also mentioned [here](https://github.com/modularml/mojo/pull/2502)

### DIFF
--- a/stdlib/src/builtin/bool.mojo
+++ b/stdlib/src/builtin/bool.mojo
@@ -17,6 +17,8 @@ These are Mojo built-ins, so you don't need to import them.
 
 from utils._visualizers import lldb_formatter_wrapping_type
 
+from collections import Set
+
 
 # ===----------------------------------------------------------------------=== #
 #  Boolable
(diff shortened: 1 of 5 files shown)


JoeLoser:
Nice!  This will be useful and I can't wait for us to get further along with iterators and algorithms.  Do you mind adding a changelog entry [here](https://github.com/modularml/mojo/blob/836626b7c7478126f21c6f57f200d88fd55a253b/docs/changelog.md#%EF%B8%8F-new) please for this work?

stdlib/src/builtin/bool.mojo
```diff
@@ -17,6 +17,8 @@ These are Mojo built-ins, so you don't need to import them.
 
 from utils._visualizers import lldb_formatter_wrapping_type
 
+from collections.set import Set as _Set
```
JoeLoser, line 20: **Question** Do we need the `as _Set` part here?  If so, why?  Why can't we just use `Set` as is below in the functions if we do `from collections import Set`?

stdlib/src/builtin/simd.mojo
```diff
@@ -2192,6 +2192,24 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
             "llvm.vector.reduce.or", Scalar[DType.bool], has_side_effect=False
         ](self)
 
+    @always_inline
+    fn reduce_all(self) -> Bool:
```
soraros, line 2196: Do we really want to make them public? Also, maybe `@paramter if` on `DType.bool` to prevent generating redundant IR?

JoeLoser: +1 on both points. I'd lazily make them public if desired in the future. Also agree re parameter if.

soraros: Actually, I guess we should probably add the parameter if to `cast`, i.e. making it a no-op when `target_type == type`.

stdlib/src/builtin/bool.mojo
helehex: Wasn't there an issue with importing things into builtin files getting exposed publicly?

stdlib/src/builtin/simd.mojo
helehex: I did it this way to match the other reduce functions, but since it's better to use the builtin `any`/`all` functions in this case, making them private might make sense.

stdlib/src/builtin/bool.mojo
helehex: ok seems fine to just import Set

stdlib/src/builtin/simd.mojo
helehex: thanks for the suggestions

helehex:
I have another one ready for bitwise `reduce_and()`/`reduce_or()`, but waiting on this and simd bool constrain to go through .

helehex:
now i feel like these all should've been done in one go to avoid the intermediate problems, but i guess there's a chance bitwise reduce might not be desired

helehex:
oh i forgot the changelog entry, hang on

helehex:
there's a typo in the changelog under the `repr()` section, I think `by [@mikowals]` should be `by [@gabrieldemarmiesse]`

stdlib/test/builtin/test_any_all.mojo
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo %s
+
+from testing import assert_true, assert_false
+from collections.set import Set
+
+
+def test_list_any():
+    var all_true = List(1, 2, 3, 4)
```
laszlokindrat, line 20: nit: Can you please add some tests with different types here and below (e.g. bool and some float type)

stdlib/src/builtin/simd.mojo
```diff
@@ -2192,6 +2192,32 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
             "llvm.vector.reduce.or", Scalar[DType.bool], has_side_effect=False
         ](self)
 
+    @always_inline
+    fn _reduce_all(self) -> Bool:
+        """Returns `True` if **all** elements in this vector are non-zero.
+
```
laszlokindrat, line 2198: Can you please add a note here and in `_reduce_any` how these are different from their publilc counterparts (e.g. "unlike the public `reduce_all` this function is not constrained to boolean element type")

stdlib/test/builtin/test_any_all.mojo
helehex: good idea

helehex: i also need to test empty iterables

stdlib/src/builtin/simd.mojo
helehex: i dont think there is a need to do that right now, were still not sure what will happen to it, and if `reduce_and`/`reduce_or` become bitwise operations, it will be mentioned in the docstring.

helehex:
i guess we might want `all()` of an empty iterator to return `True`?

artemiogr97:
> i guess we might want `all()` of an empty iterator to return `True`?

I'd say to match the python behavior (I have not checked)

Brian-M-J:
Here are the Python docs for reference:

[any](https://docs.python.org/3/library/functions.html#any):
> Return `True` if any element of the iterable is true. If the iterable is empty, return `False`.

[all](https://docs.python.org/3/library/functions.html#all):
> Return `True` if all elements of the iterable are true (or if the iterable is empty).

laszlokindrat (approved):
This is great, thank you for working on this!

Issue #2194, opened 2024-04-04
Crash on call to `DType.sizeof()` when also creating a `TensorShape` [BUG]


JakeEBrooks:
### Bug description

Please forgive my ignorance... it's my first time submitting an issue on GitHub :)

Earlier today I found a weird crash that happens when `TensorShape` is called within a function that returns an object containing a `DType`. If I call `sizeof` on the `DType` field I get a compiler crash (see below).

### Steps to reproduce

For me, running the following works fine:

```mojo
struct SomeStruct:
    var datatype: DType

    fn __init__(inout self, datatype: DType):
        self.datatype = datatype

fn create_struct() -> SomeStruct:
    #var ax = TensorShape(0)
    var datatype = DType.uint32
    return SomeStruct(datatype)

fn main():
    var astruct = create_struct()
    print(astruct.datatype.sizeof())
```

But if I uncomment the `TensorShape` bit in `create_struct`, I get:

```
/__w/modular/modular/Kernels/mojo/stdlib/builtin/dtype.mojo:430:44: error: failed to legalize operation 'pop.dtype.sizeof' that was explicitly marked illegal
/__w/modular/modular/Kernels/mojo/stdlib/builtin/dtype.mojo:430:44: note: see current operation: %61 = "pop.dtype.sizeof"(%60) : (!kgen.dtype) -> index
mojo: error: Failed to materialize symbols: { (exec, { KGEN_EE_JIT_GlobalConstructor, KGEN_EE_JIT_GlobalDestructor, main }) } (from the layer: failed to lower module to LLVM IR for archive compilation)
```

Interestingly, I noticed if you just provide an empty `TensorShape` it works. For example, the following works fine for me:

```mojo
struct SomeStruct:
    var datatype: DType

    fn __init__(inout self, datatype: DType):
        self.datatype = datatype

fn create_struct() -> SomeStruct:
    var ax = TensorShape()
    var datatype = DType.uint32
    return SomeStruct(datatype)

fn main():
    var astruct = create_struct()
    print(astruct.datatype.sizeof())
```

Hope I'm not missing anything, and that this is helpful!

### System information

```shell
- What OS did you do install Mojo on ?
Ubuntu 23.10
- Provide version information for Mojo by pasting the output of `mojo -v`
mojo 24.2.0 (c2427bc5)
- Provide Modular CLI version by pasting the output of `modular -v`
modular 0.6.0 (04c05243)
```

JakeEBrooks:
[...]

PR #2603, opened 2024-05-10
[stdlib] Add `bin()` builtin function

### LINKED ISSUE #2604: [Feature Request] Add `bin()` builtin function
### Review Mojo's priorities

### What is your request?

Implement the `bin()` builtin function to convert integral values to a binary string representation.

```mojo
fn main():
    print(bin(123)) # prints "0b1111011"
```

This is a builtin function included in Python.
https://docs.python.org/3/library/functions.html#bin

### Any other details?

_No response_

bgreni:
Fixes #2604

Implements the `bin()` builtin function from Python for converting integral values to a binary string representation.

Leaving this as a draft for now as Python dictates that any type that implements `__index__` can be passed into this function, and #2384 introduces an `Indexer` trait for that method. Which is also blocking the `Bool` specialization since implementing `fn bin(b: Bool)` allows `SIMD` values of `size > 1` to be implicitly converted.

### DIFF
--- a/stdlib/src/builtin/bin.mojo
+++ b/stdlib/src/builtin/bin.mojo
@@ -1,0 +1,86 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Implements the `bin()` function
+
+These are Mojo built-ins, so you don't need to import them.
+"""
+
+
+# Need this until we have constraints to stop
+# the compiler from matching this directly
+# to bin[type: DType](num: Scalar[type])
+@always_inline("nodebug")
+fn bin(b: Scalar[DType.bool]) -> String:
+    """Returns the binary representation of a scalar bool.
+
+    Args:
+        b: A scalar bool value.
+
+    Returns:
+        The binary string representation of b.
+    """
+    return bin(index(b))
+
+
+fn bin[type: DType](num: Scalar[type]) -> String:
+    """Return the binary string representation an integral value.
+
+    ```mojo
+    print(bin(123))
+    print(bin(-123))
+    ```
+    ```plaintext
+    '0b1111011'
+    '-0b1111011'
+    ```
+
+    Parameters:
+        type: The data type of the integral scalar.
+
+    Args:
+        num: An integral scalar value.
+
+    Returns:
+        The binary string representation of num.
+    """
+    constrained[type.is_integral(), "Expected integral value"]()
+    alias BIN_PREFIX = "0b"
+
+    if num == 0:
+        return BIN_PREFIX + "0"
+
+    # TODD: pre-allocate string size when #2194 is resolved
+    var result = String()
+    var cpy = abs(num)
+    while cpy > 0:
+        result += str(cpy & 1)
+        cpy = cpy >> 1
+
+    result = BIN_PREFIX + result[::-1]
+    return "-" + result if num < 0 else result
+
+
+@always_inline("nodebug")
+fn bin[T: Indexer](num: T) -> String:
+    """Returns the binary representation of an indexer type.
+
+    Parameters:
+        T: The Indexer type.
+
+    Args:
+        num: An indexer value.
+
+    Returns:
+        The binary string representation of num.
+    """
+    return bin(Scalar[DType.index](index(num)))
--- a/stdlib/test/builtin/test_bin.mojo
+++ b/stdlib/test/builtin/test_bin.mojo
@@ -1,0 +1,55 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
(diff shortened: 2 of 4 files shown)


stdlib/test/builtin/test_bin.mojo
```diff
@@ -0,0 +1,44 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo -debug-level full %s
```
JoeLoser, line 13: **Suggestion** Drop the `-debug-level full`

stdlib/test/builtin/test_bin.mojo
```diff
@@ -0,0 +1,44 @@
...
+    assert_equal(bin(0), "0b0")
+    assert_equal(bin(1), "0b1")
+    assert_equal(bin(-1), "-0b1")
+    assert_equal(bin(4), "0b100")
+    assert_equal(bin(Int(-4)), "-0b100")
+    assert_equal(bin(389703), "0b1011111001001000111")
+    assert_equal(bin(-10), "-0b1010")
+
+
+# def test_bin_bool():
```
JoeLoser, line 36: **Question** Did you mean to have this commented out?

stdlib/src/builtin/bin.mojo
```diff
@@ -0,0 +1,80 @@
...
+    Returns:
+        The binary string representation of num.
+    """
+    constrained[type.is_integral(), "Expected integral value"]()
+    alias BIN_PREFIX = "0b"
+
+    if num == 0:
+        return BIN_PREFIX + "0"
+
+    var result = String()
```
JoeLoser, line 59: **Suggestion** You could pre-allocate the string size here if you'd like to avoid some allocations by pre-computing the size.

JoeLoser (approved):
Generally looks good, thank you for the contribution! I'd recommend un-drafting this and not coupling it to https://github.com/modularml/mojo/pull/2384.  We can come back through once https://github.com/modularml/mojo/pull/2384 makes it through.  Thoughts, @laszlokindrat?

stdlib/src/builtin/bin.mojo
bgreni: Unless you had something else in mind I'm not sure we can at the moment due to #2194 since you can't currently call `bitwidth()`  for `index` size integers.

stdlib/test/builtin/test_bin.mojo
bgreni: I did since I ran into an issue with larger `SIMD` types being implicitly converted to `Bool`, but now that you point it out `Bool` will be implicitly converted to `Int` so it turns out the tests work anyways.

laszlokindrat:
> Generally looks good, thank you for the contribution! I'd recommend un-drafting this and not coupling it to #2384. We can come back through once #2384 makes it through. Thoughts, @laszlokindrat?

In https://github.com/modularml/mojo/pull/2384 I suggested to split up the patch into two: one that delivers `Indexer` and a second one that updates `__getitem__` and friends. Given that using that trait this patch would simplify dramatically, I would recommend waiting for `Indexer` instead of having to come back and delete/move a bunch of code.

stdlib/src/builtin/bin.mojo
```diff
@@ -0,0 +1,81 @@
+# ===----------------------------------------------------------------------=== #
```
laszlokindrat, line 1: Once `Indexer` is delivered, this module will be much simpler, potentially just a single function. It might make sense to move it under `builtin/int`. Let's see when that happens.

laszlokindrat:
`Indexer` will be available in the next nightly, please rebase when that comes out.

stdlib/src/builtin/bin.mojo
bgreni: My thought process was that the main implementation needs to be `bin[type: DType](num: Scalar[type])` in case something like `DType.int128` is added later on which could be larger than `Int`. In which case we need exactly two functions.

laszlokindrat: On that note, should we just generalize `bin` to accept SIMD sizes larger than one?

bgreni: I would be happy to do that, should that produce a list representation such as `"[0b1, 0b0]"`?

laszlokindrat: Why not just one long string?

laszlokindrat: Oh I see, leading zeros and signs would be a problem. Yeah, let's table this, I didn't think through. Maybe we can circle back to it later.

bgreni: Could be a bit awkward to read, but one long string would be easier to parse back into a `SIMD` if that was something we wanted to do.

stdlib/src/builtin/bin.mojo
```diff
@@ -0,0 +1,81 @@
...
+    Args:
+        b: A scalar bool value.
+
+    Returns:
+        The binary string representation of b.
+    """
+    return bin(int(b))
+
+
+fn bin[type: DType](num: Scalar[type]) -> String:
```
laszlokindrat, line 32: Python declares this with a positional-only argument, and I think that's a good idea. Could you please change these (and you can do the same to the `index` function).
```suggestion
fn bin[type: DType](num: Scalar[type], /) -> String:
```

bgreni: Oh TIL, yeah I definitely do that

laszlokindrat (approved):
Very nice, thank you!

Issue #2390, opened 2024-04-23
[Feature Request] Adding a builtin `sort`/`sorted` function


Moosems:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

I think it would be a great improvement to the stdlib to bring a `sort`/`sorted` function that uses a `Sortable` trait for things like large length string sorting.

### What is your motivation for this change?

I use the `sort` function all the time in python (especially with a key) and I'd love to see this make its way into Mojo

jayzhan211:
I'm interested to support feature.

jayzhan211:
I think we need to implement comparison for CollectionElement

```
'T' does not implement the '__lt__' method
```

JoeLoser:
This would be great! Right now we only have a basic sorting algorithm in the algorithm package that only works on AnyRegType if I recall correctly. Perhaps it makes sense to open source that as a reference and then we can make it work on AnyType, improve the performance of it, and use the ordering trait in https://github.com/modularml/mojo/issues/2462.

Do you or @mzaks already have a generic sorting algorithm on AnyType/some refined trait such as the ordering one? If so, would very much welcome a PR into the library and we can build out this area.

jayzhan211:
> Do you or @mzaks already have a generic sorting algorithm on AnyType/some refined trait such as the ordering one? If so, would very much welcome a PR into the library and we can build out this area.

I don't have :(  but willing to help on this! Perhaps building on what you have mentioned

mzaks:
> Do you or @mzaks already have a generic sorting algorithm on AnyType/some refined trait such as the ordering one? If so, would very much welcome a PR into the library and we can build out this area.

I updated [mojo-sort](https://github.com/mzaks/mojo-sort) so insertion sort and quick sort are working on nightly, I am currently using following signature: `fn quick_sort[D: CollectionElement, lt: fn (D, D) -> Bool](inout vector: List[D]):` to make it generic, but with `OrderingComparable` trait it will be even simpler. Fo rstd sort I would implement a mix of quick and insertions sort, use insertion sort for small count and quick sort for larger count.

[...]

PR #2517, opened 2024-05-05
[stdlib] add `Comparable` trait

helehex:
adds a `Comparable` trait for comparison testing conformance.
explicitly conform `FloatLiteral`, `IntLiteral`, `Int`, and `Set`.

PR #2609, opened 2024-05-10
[stdlib] Introduce list sorting for comparable elements

mzaks:
This is a big PR which addresses #2390 and makes following PRs:
https://github.com/modularml/mojo/pull/2517
https://github.com/modularml/mojo/pull/2378
https://github.com/modularml/mojo/pull/2409

obsolete.

It replaces the recently introduced `builtin/sort` module. I suppose that the replacement should not be problematic as the legacy sort module was introduced due to discussion in #2390 and is not used anywhere in the code base as far as I can tell.


gabrieldemarmiesse:
That's a pretty cool PR! To make the review easier (1000 lines of code to read in one go is a lot), do you think we could merge 
* https://github.com/modularml/mojo/pull/2517
* https://github.com/modularml/mojo/pull/2378
* https://github.com/modularml/mojo/pull/2409 

first or that would be incompatible? If we merge those three and then rebase this branch on top of nightly afterwards, that might make it more managable. Are there any conflict in the design philosophy between those and yours?

JoeLoser:
> That's a pretty cool PR! To make the review easier (1000 lines of code to read in one go is a lot), do you think we could merge
> 
> * [[stdlib] add `Comparable` trait #2517](https://github.com/modularml/mojo/pull/2517)
> * [[mojo-stdlib] Implement compare operation for String #2378](https://github.com/modularml/mojo/pull/2378)
> * [[stdlib] String comparisons implemented #2409](https://github.com/modularml/mojo/pull/2409)
> 
> first or that would be incompatible? If we merge those three and then rebase this branch on top of nightly afterwards, that might make it more managable. Are there any conflict in the design philosophy between those and yours?

All of the dependent PRs landed in today's nightly. Do you mind rebasing like Gabriel suggested?

mzaks:
The branch is rebased, still quite a large PR :)

What I can offer is to remove Radix sort from this PR and introduce it in the next one.

soraros:
@mzaks I'm in favour of removing Radix from this PR. I also think the Radix needs a lot more comments/design doc/reference.

soraros:
Also, can we make functions other than `sort` private (prefix by `_`)? I think they should be treated as experimental/unstable.

mzaks:
> Also, can we make functions other than `sort` private (prefix by `_`)? I think they should be treated as experimental/unstable. We could maybe also keep the old implementation if only for evaluation purpose, and only remove them when we build up the benchmark infrastructure.

My Idea was, that there is the `sort` function which users will use by default. This function decides which algorithm is best given the list size. And there are the actual sorting algorithms `quick_sort`, `insertion_sort` etc... which users can explicitly use, they have same API, so I don't see the point of hiding them.

Regarding the old std lib sort functions. I did a benchmark https://github.com/mzaks/mojo-sort/blob/main/benchmark.mojo and the old sort function performs worse through out the test. Radix sort performs incredible when the list size is higher. Hence why I added Radix sort from the get go, but now I removed it. I agree that it adds lots of complexity to this PR.

soraros:
> My Idea was, that there is the `sort` function which users will use by default. This function decides which algorithm is best given the list size. And there are the actual sorting algorithms `quick_sort`, `insertion_sort` etc... which users can explicitly use, they have same API, so I don't see the point of hiding them.

I agree that we can and probably should expose them. I only suggested to use the prefix as indicator for unstable APIs (signature may change, they should live in `algorithm`, etc.), we might want to separate them from stable ones like `sort`. Maybe we can move them to `utils.sort`, and have the free function `sort` be a thin wrapper.

> Regarding the old std lib sort functions. I did a benchmark https://github.com/mzaks/mojo-sort/blob/main/benchmark.mojo and the old sort function performs worse through out the test. Radix sort performs incredible when the list size is higher. Hence why I added Radix sort from the get go, but now I removed it. I agree that it adds lots of complexity to this PR.

I knew that you did the benchmark, I don't question your result at all, and the Redix sort is a great addition. I meant keeping the old implementation to test the benchmarking infrastructure in the Mojo repo (CI etc.), and use the future PR which removes them as test bed.

But, of course, these are merely my own views and suggests. As long as the team is ok with your approach, I'm fine with them too. Thank you for the great work.

stdlib/src/builtin/sort.mojo
```diff
@@ -15,400 +15,169 @@
...
+fn insertion_sort[D: ComparableCollectionElement](inout list: List[D]):
+    """Sort list of the order comparable elements in place with insertion sort algorithm.
 
     Parameters:
-        type: DType of the underlying data.
+        D: The order comparable collection element type.
 
     Args:
-        v: Input vector to sort.
+        list: The list of the order comparable elements which will be sorted inpace.
```
JoeLoser, line 114: **Quibble** `s/inpace/in-place`

stdlib/src/builtin/sort.mojo
```diff
@@ -15,400 +15,169 @@
...
+    @always_inline
+    @parameter
+    fn _partition(low: Int, high: Int) -> Int:
+        var pivot = list[high]
+        var i = low - 1
+        for j in range(low, high):
+            if list[j] <= pivot:
+                i += 1
+                list[j], list[i] = list[i], list[j]
+        list[i + 1], list[high] = list[high], list[i + 1]
```
JoeLoser, line 148: **Suggestion** Split the assignment across two lines please for readability.

```mojo
list[i+1] = list[high]
list[high] = list[i+1]
```

stdlib/src/builtin/sort.mojo
```diff
@@ -15,400 +15,169 @@
...
     Parameters:
-        type: DType of the underlying data.
+        D: The dtype of the scalar.
 
     Args:
-        buff: Input buffer.
-        len: The length of the buffer.
+        list: The list of the scalars which will be sorted inpace.
     """
+    _quick_sort(list, 0, len(list) - 1)
```
JoeLoser, line 82: **Question** Do we have tests for `quick_sort` on zero-length lists?  Anytime I see `len(list) - 1`, I worry about the end bounds, but I suspect `_quick_sort` does the right thing still.

stdlib/src/builtin/sort.mojo
```diff
@@ -15,400 +15,169 @@
...
-        _sort_partial_3[type, cmp_fn](array, 2, 3, 4)
-        _sort2[type, cmp_fn](array, 1, 4)
-        _sort_partial_3[type, cmp_fn](array, 0, 2, 3)
-        _sort_partial_3[type, cmp_fn](array, 1, 2, 3)
-        return
+    Args:
+        list: The list of the scalars which will be sorted inpace.
+    """
+    var count = len(list)
+    if count <= 64:
```
JoeLoser, line 180: **Question** From your experience, is this list size for the hybrid sort platform dependent or `64` is about the sweet spot across the board?

stdlib/test/builtin/test_sort.mojo
```diff
@@ -10,589 +10,245 @@
...
-    # CHECK: -1
-    # CHECK: 0
-    # CHECK: 1
-    for i in range(length):
-        print(list[i])
 
-    list[0] = 2
-    list[1] = -2
-    list[2] = 0
+def test_sort_random_numbers_u8():
```
JoeLoser, line 62: **Suggestion** Since the next 100 lines are repeated with the whole parametric thing being the DType, can we use a type list and instantiate a common test function parameterized over said `DType` instead? See https://github.com/modularml/mojo/blob/f5f5109541c31615a68a3c4b58bd1e75b59625f6/stdlib/src/builtin/dtype.mojo#L739 which has an example of putting all the dtypes in a type list and running a function over it to iterate over each type in the type list.  It would simplify this test code a fair bit IMO.

stdlib/test/builtin/test_sort.mojo
```diff
@@ -10,589 +10,245 @@
...
-    # CHECK-LABEL: test_partition_top_k_11_2
-    # CHECK-NOT: incorrect top-k
-    test_partition_top_k(11, 2)
-    # CHECK-LABEL: test_partition_top_k_4_1
-    # CHECK-NOT: incorrect top-k
-    test_partition_top_k(4, 1)
+    assert_sorted(list)
+
+
+def main():
```
JoeLoser, line 239: **Suggestion** Please add some tests for sorting a zero length (empty) list.

JoeLoser (approved):
This is **great**!!! Thanks so much for pushing on this.

I left a few minor comments that I'd like to see addressed before we land this.

stdlib/src/builtin/sort.mojo
mzaks: We would need to introduce a temp var and have a third statement:
```Mojo
var tmp = list[i+1]
list[i+1] = list[high]
list[high] = tmp
```

It is simpler to read but not "pythonic" :)

mzaks:
> I left a few minor comments that I'd like to see addressed before we land this.

The comments are addressed and the branch is rebased.

JoeLoser:
> > I left a few minor comments that I'd like to see addressed before we land this.
> 
> The comments are addressed and the branch is rebased.

Nice, thank you!  Would you like to add a changelog entry for your awesome work [here](https://github.com/modularml/mojo/blob/nightly/docs/changelog.md#%EF%B8%8F-new) as well?  Sorry I forgot to mention this yesterday.

gryznar:
There are few places with typos / incosistency:
- inpace
- in-pace
- in place

Everywhere should be : "in-place"

gryznar:
I am also sure, that for `DType` there is some name in stdlib, more meaningful than "D"

stdlib/src/builtin/sort.mojo
```diff
@@ -15,383 +15,173 @@
...
-fn sort(inout v: List[Int]):
-    """Sort the vector inplace.
-    The function doesn't return anything, the vector is updated inplace.
+    Parameters:
+        D: The order comparable collection element type.
 
     Args:
-        v: Input integer vector to sort.
+        list: The list of the order comparable elements which will be sorted inpace.
+        low: Int value identifying the lowes index of the list section to be sorted.
```
gryznar, line 136: *lowest

gryznar (commented):
One more nit. Changelog entry is not very consistent with the rest, but overall LGTM

mzaks:
> One more nit. Changelog entry is not very consistent with the rest, but overall LGTM

I rephrased it a bit :)

gryznar (approved):

stdlib/src/builtin/sort.mojo
```diff
@@ -15,383 +15,175 @@
...
-        stack.append(start)
-        stack.append(pivot)
+fn quick_sort[dtype: DType](inout list: List[Scalar[dtype]]):
+    """Sort list of scalars in-place with quick sort algorithm.
 
+    Parameters:
+        dtype: The dtype of the scalar.
 
-fn _quicksort[
-    type: CollectionElement, cmp_fn: fn (type, type) capturing -> Bool
```
JoeLoser, line 181: **Question** Previously, this quicksort function allowed specifying a comparator; we use this functionality internally for sorting by `greater_than` comparator for example instead of the default less than comparator.  Do you mind adding this level of flexibility and plumbing it through the various sorting algorithms?

stdlib/src/builtin/sort.mojo
```diff
@@ -15,383 +15,175 @@
...
+        high: Int value identifying the highest index of the list section to be sorted.
+    """
 
-    if cmp_fn[type](array[right], pivot_value):
-        right += 1
-    swap(array[end - 1], array[right])
-    return right
+    @always_inline
+    @parameter
+    fn _partition(low: Int, high: Int) -> Int:
```
JoeLoser, line 58: **Suggestion** Do you mind lifting this out of `_quick_sort` as a free function?  We use `_partition` internally  and it would be nice to preserve that level of flexibility for custom sorting things.

stdlib/src/builtin/sort.mojo
mzaks: There a wrinkle with that, currently all sort algorithms are comparison based so it is not a problem. But Radix sort is not comparison based, hence it would mean we would have an issue to call it from the sort function, if sort function allows configurable comparison operator. I guess if the main motivation is for sorting in descending order, then we can introduce an ascending/descending parameter, instead of a sorting function. This should work for radix sort as well. If there are other motivations, then I guess we can go with operator and avoid calling radix sort from the sort function. We could still introduce it as a standalone function in the module though. 

Let me know what you prefer.

stdlib/src/builtin/sort.mojo
mzaks: @JoeLoser I just wanted to lift the `_partition` function to upper level, but then I looked through the previous code again and I noticed that the top level `partition` and `_partition` functions  expect `cmp_fn: _cmp_fn_type` as a parameter, which I think is quite specific for the `builtin` sort module. Or rather, we need to decide if we want a complete set of functions which expect the `cmp_fn: _cmp_fn_type` parameter. Currently I implemented a set of functions for `Scalar` list elements (mainly because allows some performance optimisations) and  `ComparableCollectionElement` list elements. 

If we want to keep 

`alias _cmp_fn_type = fn[type: AnyRegType] (type, type) capturing -> Bool`

It will make sense to introduce a third set of functions just with this signature. To be honest I would be agains it as it increases maintenance complexity for a very specific use case, but I am ok with going along with it, if it important internally.

JoeLoser:
!sync

stdlib/src/builtin/sort.mojo
JoeLoser: That makes sense.  We have a couple of use cases internally for calling `quick_sort` with various comparators such as `>=`, `<=`, or `==` on the elements when doing the comparisons.  I don't think we should go the ascending/descending route since it's not much more work to be generic over any comparison function.  Perhaps we should add an explicit `sort_by` function that allows passing this generic functor working on comparison of two `ComparableCollectionElement` values?  Something like https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by for example.  With this particularly, we can decide whether we want to push the function (and default it to a generic less than function for sorting purposes) in the `sort` API or have it be a different named function (such as the `sort_by`).

I think we should go in this general direction and then think about various things as follow-ups — do we want various sorting APIs or have a one-stop-shop like in C++ [here](https://en.cppreference.com/w/cpp/algorithm/ranges/sort)?  Of note, you mentioned non-comparison based sorting algorithms wouldn't be suitable for this sort of function signature.

What do you think is best?  Also CC @laszlokindrat @rparolin in case you have thoughts.

stdlib/src/builtin/sort.mojo
JoeLoser: Similar to my other comments, I think we should introduce a generic range partition that works on `ComparableCollectionElement` I suppose.  I'm imagining something similar to `std::ranges::partition` from C++ [doc](https://en.cppreference.com/w/cpp/algorithm/ranges/partition).  What do you think about that?  I think partition as a general algorithm (outside of just the sorting use case) is useful and it'll evolve over time.

stdlib/src/builtin/sort.mojo
mzaks: The question is, how generic do we want the `sort` to be. 

The current internal functions signatures are:
`[type: AnyRegType, cmp_fn: _cmp_fn_type](array: Pointer[type])` 

and 

`[type: CollectionElement, cmp_fn: fn (type, type) capturing -> Bool](array: UnsafePointer[type])`

which are called from public signatures like:

`sort(inout buff: Pointer[Int], len: Int):`

`fn sort[type: DType](inout buff: Pointer[Scalar[type]], len: Int):`

`fn sort(inout v: List[Int]):`

`fn sort[type: DType](inout v: List[Scalar[type]]):`

and 

`fn sort[type: CollectionElement, cmp_fn: fn (type, type) capturing -> Bool](inout v: List[type]):`

The state of the sort module actually changed a lot since my PR was opened so I think it is best to take a step back. I will make this PR less aggressive I will just introduce a new `sort` function with following signature:

`fn sort[type: ComparableCollectionElement](inout v: List[type]):`

So the PR will introduce the initial concept of generic comparison sorting and then we can build upon that.

mzaks:
@JoeLoser I reframed the PR to be just an additive change to add sorting function for the list of comparable elements. This way we can finally merge this PR and introduce new smaller PRs to add partition function, introduce new sorting algorithms, maybe consolidate the module at some point.

JoeLoser:
!sync

PR #2611, opened 2024-05-10
[stdlib] Add unit tests for SIMD add, radd and iadd operations

peymanbr:
Add unit tests for the `__add__()`, `__radd__()`, and `__iadd__()` methods.


laszlokindrat (approved):
Awesoem, thanks for adding these!

PR #2612, opened 2024-05-10
[stdlib] Use UInt8 to store characters in `String`

gabrieldemarmiesse:
The fact that there are so many changes to do means that many functions access `String`'s internal buffer without going through the method provided (for example `unsafe_ptr()`). It might be worth it to rework a bit the API so that when we change the `String`'s  internals later on, we don't have to touch too many things. I'm notably thinking here about SSO and Unicode.

This PR is part of https://github.com/modularml/mojo/issues/2317

Thanks god ASCII chars are between 0 and 127...


stdlib/src/builtin/string.mojo
```diff
@@ -553,7 +553,7 @@ struct String(
 
     # TODO: Remove this method when #2317 is done
     @always_inline
-    fn __init__(inout self, owned impl: Self._buffer_type):
+    fn __init__(inout self, owned impl: List[Int8]):
```
JoeLoser, line 556: **Suggestion** Should we throw @deprecated on this then given the note below?

>
Note that you should use the constructor from `List[UInt8]` instead

JoeLoser (approved):
Nice!

stdlib/src/builtin/string.mojo
gabrieldemarmiesse: I agree. Maybe let's do that in another PR, I'll also remove the unused methods at the same time.

gabrieldemarmiesse:
Happy to see that https://github.com/modularml/mojo/commit/4433f959dd3fd3ffcba1259a4a9b4bd81d288c58 was merged! I think we can close this PR now

JoeLoser:
> Happy to see that [4433f95](https://github.com/modularml/mojo/commit/4433f959dd3fd3ffcba1259a4a9b4bd81d288c58) was merged! I think we can close this PR now

Yep, thanks for closing this out.  As I mentioned at https://github.com/modularml/mojo/pull/2509#issuecomment-2121008274, a lot of this stuff was intertwined internally, so it was easier for @rparolin to just make a bigger commit to knock it all out.  Thanks for pushing on this!

PR #2620, opened 2024-05-11
[stdlib] Delegate string comparisons to `StringRef`

siitron:
This is a follow-up to #2409.

String comparisons for `StringRef` are implemented. `StringRef` make use of `memory.memcmp` for all of its 6 comparisons now, hopefully this change is ok. 

`String`'s and `StringLiteral`'s comparisons (`__eq__`, `__ne__`, `__lt__`, `__le__`, `__gt__`, & `__ge__`) are delegated to `StringRef`. As a result:
1. `String` comparisons uses its `_strref_dangerous`-method. And,
2. `StringLiteral`'s local `_memcmp`-function is removed (which was used "rather than memory.memcpy to avoid #31139 and #25100").

The base logic for all 18 comparisons are implemented in `StringRef`'s `__ne__` and `__lt__` methods. All other comparisons uses some variation/negation of either `__ne__` or `__lt__`.


siitron:
Ubuntu tests got stuck for 6 hours (but the macos ones passed). I'm closing + reopening this to rerun the tests.

JoeLoser:
> Ubuntu tests got stuck for 6 hours (but the macos ones passed). I'm closing + reopening this to rerun the tests.

Still looks like they're hanging for you. I just posted about this on Discord, but I don't think the hang is due to your change?

laszlokindrat (approved):
LGTM, thanks for fixing this!

laszlokindrat:
FYI I had to modify this internally to avoid the `memcmp` recursion. It's the same problem that necessitated the `_memcmp` workaround, and I did a similar workaround as well.

siitron:
> FYI I had to modify this internally to avoid the `memcmp` recursion. It's the same problem that necessitated the `_memcmp` workaround, and I did a similar workaround as well.

@laszlokindrat Great that you got it to work! But, does this mean that String now have to use a non-vectorized version of memcmp when doing comparisons?

laszlokindrat:
> @laszlokindrat Great that you got it to work! But, does this mean that `String` now have to use a non-vectorized version of memcmp when doing comparisons?

Yes, unfortunately, although my understanding is that the code before your patch had a similar workaround. If you are interested, you can give it a try: the recursion probably comes from the constraint or some assert or print from the memcmp implemention. In theory, you might be able to refactor to work around this.

siitron:
> Yes, unfortunately, although my understanding is that the code before your patch had a similar workaround. If you are interested, you can give it a try: the recursion probably comes from the constraint or some assert or print from the memcmp implemention. In theory, you might be able to refactor to work around this.

Yes, the workaround is similar to what `StringLiteral` used before, but `String` can use the regular `memcmp`. I'm not sure what causes the need for the workaround though, I could maybe try and see if anything could be done but idk. It's at least nice to have it all in one place :)

siitron:
What's even the issue at the moment? I can still pass the tests locally after replacing _memcmp with memcmp. 🤔

PR #2630, opened 2024-05-12
[stdlib] [proposal] Flexible-compact-datetime

martinvuyk:
linked to PR #2623


JoeLoser:
Hi @martinvuyk!  Thanks so much for the proposal here along with some reference implementation in https://github.com/modularml/mojo/pull/2623.  This is no small feat!

This is an interesting area and something whose core parts do have value in the Mojo standard library.  However, right now, our team is quite small and so we're not actively accepting massive contributions (quite yet!) that introduce new modules wholesale like this.  Right now, as discussed in [our roadmap](https://github.com/modularml/mojo/blob/nightly/stdlib/docs/roadmap.md), we're focusing on a lot of the "basics", if you will, to build solid foundations to create space for working on some of these bigger things like the `datetime` module you're proposing here.

I'd recommend you keep developing the `datetime` library outside of this repo/standard library for now, and we revisit in Q3 or Q4 to see where the standard library is at.  I do also want to remind you that we will eventually have a Mojo package manager — this can be a great way to get wide use of a bigger module like this and then we can incorporate the feedback you get from your users for when parts or all of it make their way into the standard library.

In short, don't give up on this module development!  Consider our team's response truly a "not now" — not a "never".  We're just a small team trying to stay laser focused on getting the basic building blocks right at this time. ❤️

PR #2650, opened 2024-05-14
[stdlib] Fix atol() when explicitly setting base 2, 8, and 16

northstreet12:
Should atol() be fixed when explicitly setting base 2, 8, and 16?

```
atol(" 0xff ", 16)
atol(" 0o12 ", 8)
```

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -234,6 +234,20 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
             start = pos
         break
 
+    if str_ref[start] == "0" and start + 1 < str_len:
+        if base == 2 and (
+            str_ref[start + 1] == "b" or str_ref[start + 1] == "B"
+        ):
+            start += 2
+        elif base == 8 and (
+            str_ref[start + 1] == "o" or str_ref[start + 1] == "O"
+        ):
+            start += 2
+        elif base == 16 and (
+            str_ref[start + 1] == "x" or str_ref[start + 1] == "X"
+        ):
+            start += 2
+
     alias ord_0 = ord("0")
     # FIXME:
     #   Change this to `alias` after fixing support for __refitem__ of alias.
--- a/stdlib/test/builtin/test_string.mojo
+++ b/stdlib/test/builtin/test_string.mojo
@@ -306,7 +306,14 @@ fn test_atol() raises:
     assert_equal(10, atol("A", 16))
     assert_equal(15, atol("f ", 16))
     assert_equal(255, atol(" FF", 16))
+    assert_equal(255, atol(" 0xff ", 16))
+    assert_equal(255, atol(" 0Xff ", 16))
     assert_equal(18, atol("10010", 2))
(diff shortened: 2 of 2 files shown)


stdlib/src/builtin/string.mojo
```diff
@@ -248,6 +248,20 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
             start = pos
         break
 
+    if str_ref[start] == "0" and start + 1 < str_len:
```
artemiogr97, line 251: maybe for the cases of base 2, 8 and 16 you can call `_identify_base` and if the returned base do not match the expected base than you raise an error?, just to prevent duplicating the logic for the base identification twice

artemiogr97: That's true, which means that in case of 2, 8 and 16, the identified base must be the original base or 10

northstreet12: I understand `_identify_base ` function is designed especially for the cases that base is set to 0. So it requires int literals to  identify themselves without ambiguity. If an int literal start with '0x', it is recognized as real base 16. Also the same for real base 2 and 8. If an int literal comes without prefix such as '0x' (e.g 'ff' of base 16), it would return real base -1 and raise error. So it might be a little bit difficult to use `_identify_base ` for int literals which base is explicitly set.

northstreet12: When get a real base -1 returned, we might need to determine whether the int literal is actualy valid if it is explicitly set a base (e.g 'ff' of base 16), or it is totally invalid such as "$%^&!!". So can we just fix this simply?

artemiogr97: For bases 2, 8, and 16 it is ok if the identified base is the original, 10 or -1, if something is wrong atol it's going to fail later.
The idea behind what I said was just to prevent duplicating the logic for the base identifiers (0b, 0B, 0x, ...), but feel free to adapt the code if what I proposed does not work.

northstreet12: Yes, I understand. I tried to use `_identify_base` to avoid duplicating the logic, but processing return value -1 is little bit complicated (as said above in case of 'ff' with explicite base 16). So maybe we could simply fix existing problem, then to consider how we can combine the logic into one function. Thanks.

northstreet12:
Hi, any comments?

JoeLoser:
> Hi, any comments?

Connor is out of office this week, so I'll take a look at this tomorrow (Tuesday) during code reviews. Thanks!

northstreet12:
@JoeLoser Hi, can you take a look at this? Just a simple fix :)

northstreet12:
@JoeLoser Hi, can you take a look at this? Just a simple fix :)

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
> @JoeLoser Hi, can you take a look at this? Just a simple fix :)

Yep, sorry I missed this one this week. Just approved and imported the PR.

PR #2651, opened 2024-05-14
[stdlib] Bump compiler version to 2024.5.1407

northstreet12:
It seems the latest nightly mojo version is updated to 2024.5.1407


patrickdoc:
Thanks for pointing this out! We had a bug in our release process last night that led to only a partial release. It's been resolved now, so if you update `mojo` you should see `2024.5.1413` which matches the latest commit now: https://github.com/modularml/mojo/commits/nightly/

Apologies for the inconvenience!

northstreet12:
OK. It works well now.

PR #2665, opened 2024-05-15
[External] [stdlib] List.__contains__

rd4com:
Hello, here is ```List.__contains__```,
The conformance is done by hand with ```constrained```, ```_type_is_eq``` and ```rebind``` so that users can do:
```mojo
var x = List[Int](1,2,3)
if 3 in x: print("ok")
```
If value would be T2, we would not be able to do it
(this is the same PR as https://github.com/modularml/mojo/pull/2662 , started over again)


rd4com:
i'll close it and do it again, sorry

PR #2667, opened 2024-05-15
[External] [stdlib] List.__contains__

rd4com:
Hello, here is ```List.__contains__```,
The conformance is done by hand with ```constrained```, ```_type_is_eq``` and ```rebind``` so that users can do:
```mojo
var x = List[Int](1,2,3)
if 3 in x: print("ok")
```
If value would be T2, we would not be able to do it
(this is the same PR as https://github.com/modularml/mojo/pull/2662 and https://github.com/modularml/mojo/pull/2665 , started over again)

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -22,6 +22,7 @@ from collections import List
 
 from memory import UnsafePointer, Reference
 from memory.unsafe_pointer import move_pointee, move_from_pointee
+from sys.intrinsics import _type_is_eq
 from .optional import Optional
 
 # ===----------------------------------------------------------------------===#
@@ -753,3 +754,30 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             The UnsafePointer to the underlying memory.
         """
         return self.data
+
+    @always_inline
+    fn __contains__[
+        T2: ComparableCollectionElement
+    ](self: List[T2], value: T) -> Bool:
+        """Verify if a given value is present in the list.
+
+        ```mojo
+        var x = List[Int](1,2,3)
+        if 3 in x: print("x contains 3")
+        ```
+        Parameters:
+            T2: The type of the elements in the list. Must implement the
+              traits `EqualityComparable` and `CollectionElement`.
+
+        Args:
+            value: The value to find.
+
+        Returns:
+            True or False.
+        """
+
+        constrained[_type_is_eq[T, T2](), "value type is not self.T"]()
+        for i in self:
+            if i[] == rebind[T2](value):
+                return True
+        return False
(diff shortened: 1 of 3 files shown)


helehex:
Hey that's really awesome! I didn't know you could do that

stdlib/test/collections/test_list.mojo
```diff
@@ -741,6 +741,19 @@ def test_list_mult():
     assert_equal(len(List[Int](1, 2, 3) * 0), 0)
 
 
+def test_list_contains():
+    var x = List[Int](1, 2, 3)
+    assert_equal(0 in x, False)
```
JoeLoser, line 746: **Suggestion** Can you rewrite these to use `assert_true` and `assert_false`, respectively, please?  E.g.

```mojo
assert_false(0 in x)
assert_true(1 in x)
assert_false(4 in x)
```

stdlib/src/collections/list.mojo
```diff
@@ -753,3 +754,30 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        ```
+        Parameters:
+            T2: The type of the elements in the list. Must implement the
+              traits `EqualityComparable` and `CollectionElement`.
+
+        Args:
+            value: The value to find.
+
+        Returns:
+            True or False.
```
JoeLoser, line 776: **Suggestion** Reword to "True if the value is contained in the list, False otherwise." or something of that nature.  "True or False" doesn't help the reader of API docs much for this function.  They can infer that from the nature of the return type (`Bool`).  This doc string part should tell them _when_ it returns one or the other (i.e. the semantics of the function).

JoeLoser (approved):
Nice!  Thank you for the contribution.  I left two minor comments.  Do you also mind adding a changelog entry for this work [here](https://github.com/modularml/mojo/blob/nightly/docs/changelog.md#%EF%B8%8F-new)?

rd4com:
Hello @JoeLoser , glad you like it !

Your videos and the community stream helped a lot :partying_face: 

I'm learning branch in order to do multiple PR at same time !

stdlib/test/collections/test_list.mojo
```diff
@@ -741,6 +741,19 @@ def test_list_mult():
     assert_equal(len(List[Int](1, 2, 3) * 0), 0)
 
 
+def test_list_contains():
+    var x = List[Int](1, 2, 3)
+    assert_false(0 in x)
+    assert_true(1 in x)
+    assert_false(4 in x)
```
laszlokindrat, line 748: Normally we use `__contains__` directly in these test to avoid potential pitfalls from implicit conversions. In this case, I would recommend to do that, and maybe keep on instance of `0 in x` to make sure the syntax sugar works.

rd4com: Yes, sorry, i use the edit button on github because:
```
./scripts/run-tests.sh 
Mismatch in compiler versions! Cannot build the standard library.
Expected compiler version: 2024.5.1515
Current installed compiler version: 2024.5.1607
Please run modular update nightly/mojo to get the latest compiler.
```

PR #2668, opened 2024-05-15
[docs] Mojo Manual typo fixes

Brian-M-J:
Fix various typos in the Mojo Manual:

`functions.ipynb`:
- Changed "matter personal taste" to "matter of personal taste"
- Changed "Current there are some differences" to "Currently there are some differences"

`why-mojo.md`
- Changed "here in 2023" to "here in the present" (to make it age better as time goes on)
- Changed "an embedded domain-specific languages" to "embedded domain-specific languages"
- Added information about Python 3.13 to the `Improving CPython and JIT compiling Python` section
- Changed "the Python package ecosystems has" to "the Python package ecosystem has"
- Changed "but don't see" to "but we don't see"

`basics.ipynb`:
- Changed "we try not assume" to "we try not to assume"

`get-started.md`:
- Changed "double-check your code looks" to "double-check that your code looks"

`types.ipynb`:
- Changed "to string" to "to a string"
- Changed "They dictionary's key type" to "The dictionary's key type"
- Changed "`Set` type represent" to "`Set` type represents"
- Changed "The `Optional`" to "An `Optional`"
- Changed ",where" to ", where"
- Changed "is For" to "For"

`structs.ipynb`:
- Changed "declares `self` is a" to "declares `self` as a"
- Changed "the section about Value lifecycle" to "the Value lifecycle section"

`packages.md`:
- Changed "Mojo module may" to "A Mojo module may"

`ownership.ipynb`:
- Changed "who's" to "whose"
- Changed "the `def` and `fn` functions are interchangeable" to "`def` and `fn` functions are interchangeable"

`death.ipynb`:
- Changed "OK in the this case" to "OK in this case"
- Changed "field technically no longer used" to "field is technically no longer used"

`debugging.ipynb`:
- Changed "If this a a" to "If this is a"


arthurevans:
Thanks for the fixes! Can you please sign the commits as described in the [Contributor guide](https://github.com/modularml/mojo/blob/main/CONTRIBUTING.md#signing-your-work)?
All these fixes look good, so if it's easier you could squash them into one commit and sign that.

gryznar:
I found yet another nit in [types](https://docs.modular.com/mojo/manual/types):
image

Brian-M-J:
Fixed the sign-off check and applied @gryznar's suggestion.

I have some more ideas on fixes but I don't know what exactly to do about these:
- https://docs.modular.com/mojo/manual/python/types#python-types-in-mojo
  > Mojo doesn't have a standard dictionary type yet
  
  Mojo does have a standard [`Dict`](https://docs.modular.com/mojo/stdlib/collections/dict/Dict) type now.

- https://docs.modular.com/mojo/roadmap#no-list-or-dict-comprehensions
  > [...] Mojo's standard library has not yet grown a standard list or dictionary type.
  
  Now it has: [`List`](https://docs.modular.com/mojo/stdlib/collections/list/List) (`Dict` already mentioned above)

- https://docs.modular.com/mojo/manual/types#simd-and-dtype
  types_nit
  The code example isn't formatted as code. Without being able to use Mojo locally (I'm on a Windows machine and can't 
  install WSL), I can't really fix this.

- https://docs.modular.com/mojo/manual/traits
  > Traits are similar to Java interfaces, C++ concepts, Swift protocols, and Rust traits. If you're familiar with any of those 
      features, Mojo traits solve the same basic problem.
  
  Should we also mention [Python's protocols](https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols) or do those not really apply here as Python is dynamically typed and the protocols are more like hints?

arthurevans:
Thanks! Those roadmap issues were on my radar, so I can fix those.

We did leave Python protocols out on purpose.

arthurevans:
Fixed the three other issues listed above (https://github.com/modularml/mojo/pull/2668#issuecomment-2115608634), these should be in the next nightly and will make it to the doc site in due course.

PR #2669, opened 2024-05-15
[stdlib] Guard `SIMD.cast()` against equal types

helehex:
Checks if types are already equal when calling `SIMD.cast()`.
This avoids unnecessary code when generically casting simd vectors.

But, this could be used in place of rebind when you know the types are already equal.
if that's not what we want, then maybe we should constrain the types to be different, instead of doing this.


laszlokindrat (approved):
LGTM, thanks!

PR #2671, opened 2024-05-15
[stdlib] Change `SIMD.reduce_and()`/`SIMD.reduce_or()` to bitwise operations

helehex:
changes the behavior of simd `reduce_and()`/`reduce_or()` to be bitwise operations.

Theres a few things that are related to this:
[builtin `any()`/`all()` functions](https://github.com/modularml/mojo/pull/2600)
[constrain simd bool to size=1](https://github.com/modularml/mojo/pull/2502)


stdlib/src/builtin/simd.mojo
```diff
@@ -2172,40 +2170,94 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+            @always_inline
+            @parameter
+            fn and_reduce_body[
+                type: DType, width: Int
+            ](v1: SIMD[type, width], v2: SIMD[type, width]) -> SIMD[
+                type, width
+            ]:
+                return v1 & v2
+
+            return self.reduce[and_reduce_body, size_out]()
```
helehex, line 2203: Are there any architecture-specific checks that should go here?

helehex: For example, the `reduce_min()`/`reduce_max()` methods also run this section if `is_x86()`

laszlokindrat:
The other two related patches will be available in the next nightly. Could you please rebase after that, and add a changelog entry?

helehex:
It might also be a good to group/mention them together in the changelog, and explain this is how you check if a simd is truthy now (no more implicit conversion to bool unless size=1), to avoid common bug, etc..

stdlib/src/builtin/simd.mojo
```diff
@@ -2275,39 +2276,87 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
 
         @parameter
         if size == 1:
-            return self.cast[DType.bool]()[0].value
+            return rebind[SIMD[type, size_out]](self)
+
         return llvm_intrinsic[
-            "llvm.vector.reduce.and", Scalar[DType.bool], has_side_effect=False
+            "llvm.vector.reduce.and",
```
soraros, line 2316: I don't think this [intrinsic](https://llvm.org/docs/LangRef.html#llvm-vector-reduce-and-intrinsic) work for `size_out > 1`.

soraros: In fact, how can we reach this line at all?

stdlib/src/builtin/simd.mojo
```diff
@@ -2275,39 +2276,87 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
-        """Reduces the boolean vector using the `and` operator.
+    fn reduce_and[size_out: Int = 1](self) -> SIMD[type, size_out]:
+        """Reduces the vector using the bitwise and `&` operator.
+
+        Parameters:
+            size_out: The width of the reduction.
 
         Constraints:
-            The element type of the vector must be boolean.
+            `size_out` must not exceed width of the vector.
```
soraros, line 2286: Maybe consider adding a `constrained` for this?

stdlib/src/builtin/simd.mojo
```diff
@@ -2275,39 +2276,87 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+            fn and_reduce_body[
+                type: DType, width: Int
+            ](v1: SIMD[type, width], v2: SIMD[type, width]) -> SIMD[
+                type, width
+            ]:
+                return v1 & v2
+
+            return self.reduce[and_reduce_body, size_out]()
 
         @parameter
```
soraros, line 2311: Could you move this part before `size_out > 1`?

soraros:
Could you add some test cases for `size_out > 1`?

stdlib/src/builtin/simd.mojo
helehex: Its constrained already by calling reduce

stdlib/src/builtin/simd.mojo
```diff
@@ -2275,39 +2276,87 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+            size_out: The width of the reduction.
 
         Constraints:
-            The element type of the vector must be boolean.
+            `size_out` must not exceed width of the vector.
+            The element type of the vector must be integer or boolean.
 
         Returns:
-            True if all element in the vector is True and False otherwise.
+            The bitwise and `&` of all vector elements.
```
laszlokindrat, line 2290:
```suggestion
            The bitwise AND of all vector elements.
```

stdlib/test/builtin/test_simd.mojo
```diff
@@ -1008,6 +1008,53 @@ def test_indexer():
...
+    test_dtype[DType.bool]()
+    test_dtype[DType.int8]()
+    test_dtype[DType.int16]()
+    test_dtype[DType.int32]()
+    test_dtype[DType.int64]()
+    test_dtype[DType.uint8]()
+    test_dtype[DType.uint16]()
+    test_dtype[DType.uint32]()
+    test_dtype[DType.uint64]()
+    test_dtype[DType.float16]()
```
laszlokindrat, line 1052: Could you please add a cast for `bfloat16` as well? You will have to guard it with
```mojo
@parameter
    if not has_neon():
```

laszlokindrat (approved):
LGTM, thank you! Please add tests for `size_out > 1` as requested by @soraros.

stdlib/test/builtin/test_simd.mojo
helehex: sure 🌻

stdlib/src/builtin/simd.mojo
helehex: I'm gonna make it even simpler, since that statement isn't always accurate anyways.

helehex: if it's ok: "returns the reduced vector" makes sense in all cases

stdlib/src/builtin/simd.mojo
helehex: im not sure what you mean here

stdlib/src/builtin/simd.mojo
helehex: we should ask about having constraints as close to the caller as possible

helehex:
@laszlokindrat do you like the new test :]

helehex:
oh, do i have to format the markdown or something?

laszlokindrat:
!sync

laszlokindrat:
I'm seeing a lot of internal breakages on this, not sure what's going on. Could you please rebase after the next nightly?

helehex:
Yeah that's ok, I know there was an issue originally where I had to rebind the llvm call result, but that went away for some reason. Or maybe because it returns a `SIMD` instead of `Bool` now?

PR #2673, opened 2024-05-16
[stdlib] Allow the simplified syntax `my_list.__str__()`

gabrieldemarmiesse:


JoeLoser (approved):
Nice, thank you! Can we add a `Stringable` trait overload for `RepresentableCollectionElement` so that `List` could conform to `Stringable`, or is that not possible yet?

gabrieldemarmiesse:
We cannot add traits to the container with this method because the type of `self` must be `Self` 😢  (at least it's the error message I get)

Yeah I don't think we'll be able to add those traits to containers until we have conditional conformance

PR #2677, opened 2024-05-16
[stdlib] Add the `normalize_index` function

gabrieldemarmiesse:
This PR is some kind of mix between https://github.com/modularml/mojo/pull/2386 and https://github.com/modularml/mojo/pull/2384 which have issues (aborting) or have too many conflicts because the PR is too big.

This PR solves part of https://github.com/modularml/mojo/issues/2251 and https://github.com/modularml/mojo/issues/2337

We try here to give the ground work for indexing correctly. This function added can then be used wherever we work with sequences.

Two things I noticed during development:
1) The `debug_assert` does not run in unit tests. Is there any way to enable it? We currently have out-of-bounds bugs in our test suite.
2) The null terminator is causing pain, again, again, and again. Do we have any plans to make it optional when working with String? I opened https://github.com/modularml/mojo/issues/2678 to discuss this.

To avoid to fix those issues in this PR, I used the `normalize_index` on the `__refitem__` of `InlineArray` which doesn't have widespread use yet and isn't impacted by out-of-bounds bugs.

My recommendation would be to merge this PR then to rebase https://github.com/modularml/mojo/pull/2386 and https://github.com/modularml/mojo/pull/2384 on it. We should also afterwards fix the out of bounds bugs that can be triggered in the test suite by enabling debug_assert.

The diff might seem big, but no worries, it's mostly the licenses and docstrings :)


JoeLoser:
Re your (1) comment about `MOJO_ENABLE_ASSERTS`, we need to fix up the library and its tests.  I see we've regressed in the past month or two.  We were building/working fine with assertions enabled, but we don't have internal CI with this mode.

I think we can do something like `MOJO_ENABLE_ASSERTIONS=1 ./stdlib/scripts/run-tests.sh` (may need to propagate `MOJO_ENABLE_ASSERTIONS` through in the `lit.cfg.py`.  When I ran this recently in the open source, I think we had 12 test failures or so.  We should fix these up and then add CI for it (preferably in a matrix for with and without assertions enabled).  Thanks for the reminder; feel free to file a GitHub issue, and if you're so inclined, work on it or someone else can! 😄

gabrieldemarmiesse:
Here you go: https://github.com/modularml/mojo/issues/2687 :)

laszlokindrat:
`Indexer` will be available in the next nightly, please rebase after that!

gabrieldemarmiesse:
@laszlokindrat Feel free to review when you have the time. The rebase is done :)

stdlib/src/collections/index_normalization.mojo
```diff
@@ -0,0 +1,81 @@
...
+            + str(container_length)
+            + " (exclusive) but the index value "
+            + str(i)
+            + " was used. "
+            + "Aborting now to avoid an out-of-bounds access."
+        )
+
+
+@always_inline
+fn normalize_index[
```
laszlokindrat, line 50: This is a nice utility, but I don't think it should be a public method. In fact, I'm not sure this should be a public module at all. WDYT about renaming to `_index_normalization`?

stdlib/src/utils/static_tuple.mojo
```diff
@@ -369,27 +369,26 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
...
             index: The index of the item.
 
         Returns:
             A reference to the item at the given index.
         """
-        debug_assert(-size <= int(index) < size, "Index must be within bounds.")
-        var normalized_idx = int(index)
-        if normalized_idx < 0:
-            normalized_idx += size
+        var normalized_index = normalize_index[container_name="InlineArray"](
```
laszlokindrat, line 387: This is subtly different behavior: `debug_assert` can be disabled (in which case there is no bound checking, no overhead, but bad things can happen); the implementation you introduced, however, always performs the bounds check, which guarantees that this becomes more expensive.

@JoeLoser Any thoughts on whether this should be a "safe" method or not?

stdlib/src/collections/index_normalization.mojo
```diff
@@ -0,0 +1,81 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""The utilities provided in this module help normalize the access 
+to data elements in arrays."""
+
+
+fn _get_out_of_bounds_error_message[
```
laszlokindrat, line 17: Unfortunately, this will allocate heap memory, which is not allowed on certain platforms. This might be one reason why the `__refitem__` below uses a `debug_assert`.

laszlokindrat (commented):
Thanks for the patch! Let's have a discussion on some of these design and implementation questions, before we move forward.

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: Thanks for the review. This is intentional. I followed the instructions from @JoeLoser  here: https://github.com/modularml/mojo/pull/2386#discussion_r1589103000 

Since having a performant access to the elements in any array is desirable, we could get some inspiration from rust which does have the same challenge. Rust solves it by having safe access with `[i]` and unsafe access with [`get_unchecked`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked) . 

What I would recommend here is to do something similar, as unsafe things must be explicitly unsafe using an appropriate name (from the stdlib vision document). When this PR gets merged, I can open an issue (which can be a good first issue!) to add a second method to the InlineArray `unsafe_get()` where `normalized_index` isn't called, and we don't wrap the index when there are negative values. A debug_assert will be added in `unsafe_get()` for better debuggability.

I'll answer here concerning the heap memory allocation. I think there are a few points here to add:
* The heap allocation will be used only when the "safe" get is used. Thus if the platform doesn't support it, the developper can use `unsafe_get()`
* We should strive for good error messages in the general case. It's fine to have bad/no error messages in the other situations.
* When we have SSO and can control the size of the internal String buffer, we can make it big enough in this precise code path to make sure that the String bytes is always allocated on the stack (buffer) and never on the heap. When this is done, the platforms where heap allocation isn't possible will be able to use the safe method to get an item by index.

stdlib/src/collections/index_normalization.mojo
gabrieldemarmiesse: Answered above :)

stdlib/src/collections/index_normalization.mojo
gabrieldemarmiesse: It's not very clear what is public/private in the stdlib. I added an underscore to the module name as asked.

stdlib/src/utils/static_tuple.mojo
laszlokindrat: Okay, thanks for clarifying, I'm glad you already had this discussion. I'm okay with the goal here, but it might be risky to merge this since it _will_ introduce overhead for existing (internal) users. I can push this forward, but it would be great if you could fast follow with the implementation of an unsafe accessors.

stdlib/src/collections/_index_normalization.mojo
```diff
@@ -0,0 +1,81 @@
...
+            + " (exclusive) but the index value "
+            + str(i)
+            + " was used. "
+            + "Aborting now to avoid an out-of-bounds access."
+        )
+
+
+@always_inline
+fn normalize_index[
+    IndexType: Indexer, ContainerType: Sized, container_name: String
```
laszlokindrat, line 51: nit: slight improvement for usability and comptime:
```suggestion
    inferred IndexType: Indexer, inferred ContainerType: Sized, container_name: StringLiteral
```

laszlokindrat (approved):
LGTM, thank you! Please rebase on the latest nightly, and I'll pull this in.

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: I'll add it in this PR instead. Let's not take unnecessary risks with internal breakages.

gabrieldemarmiesse: @laszlokindrat  I added unsafe_get to have fast access :)

laszlokindrat: amazing, thanks!

laszlokindrat:
!sync

laszlokindrat:
@gabrieldemarmiesse Sorry this has been taking so long, but we are running into several internal CI failures related to this and a lot has changed recently. Could you please rebase this on the next nightly? Note that there will be a `_get_reference_unsafe` methods that probably makes your `unsafe_get` obsolete.

gabrieldemarmiesse:
@laszlokindrat  done!

ConnorGray:
Now that this has been merged, we should start using it more widely. @gabrieldemarmiesse @laszlokindrat is there any reason I might have missed that means I should hold off on filing an issue for using this more widely?

gabrieldemarmiesse:
I think we can open an issue. At the time of merge there were a few PRs in flight, we didn't know if there was work left to do on the subject.

Now that the dust has settled, it's true that it would be a good issue to beginners. Let's not forget in the issue to mention that we would like on function change per PR to avoid a big PR changing all the indexing for the stdlib

ConnorGray:
Thanks Gabriel. Filed https://github.com/modularml/mojo/issues/2948 to track this. From grepping through the stdlib code it looks like there aren't that many uses of this pattern overall, so the size of resulting PRs may end up not to be too huge. I included a note though asking contributors not to do more than 4-5 call-sites in one PR.

PR #2683, opened 2024-05-16
[stdlib] Add method `strip()` to `StringRef`

fknfilewalker:
This PR adds a `strip` method to `StringRef`. This PR is helpful for #2649.

Where can I find the test cases for StringRef?


stdlib/src/utils/stringref.mojo
```diff
@@ -425,6 +425,23 @@ struct StringRef(
...
+    fn strip(self) -> StringRef:
+        """Gets a StringRef with leading and trailing whitespaces removed.
+
+        For example, `"  mojo  "` returns `"mojo"`.
+
+        Returns:
+            A StringRef with leading and trailing whitespaces removed.
+        """
+        var start: Int = 0
+        var end: Int = len(self) - 1  # without '\0'
```
laszlokindrat, line 437: `StringRef` doesn't have to point to a null terminated string (we have a length), so you don't need to decrement here.

stdlib/src/utils/stringref.mojo
```diff
@@ -425,6 +425,23 @@ struct StringRef(
...
+        var ptr = self.unsafe_ptr()
+        while start < end and isspace(ptr[start]):
+            start += 1
+        while end > start and isspace(ptr[end]):
+            end -= 1
+        return StringRef(ptr + start, end - start + 1)
```
laszlokindrat, line 443: see comment above
```suggestion
        return StringRef(ptr + start, end - start)
```

laszlokindrat (commented):
Nice! Please add some tests.

stdlib/src/utils/stringref.mojo
fknfilewalker: this is needed otherwise
```python
isspace(ptr[end])
```
in
```python
while end > start and isspace(ptr[end]):
```
points to `'\0'` and the while is never entered

fknfilewalker:
```python
isspace(ptr[end])
```
in
```python
while end > start and isspace(ptr[end]):
```
could point to a '\0' right? so I need to check that too, since StringRef can also contain that?

fknfilewalker: I guess it would be nice to have something like `isnullterminator()`, similar to `isspace()`

laszlokindrat: It could point to anything, since it's out of bounds (e.g. this could be a reference to a substring), so you could do
```mojo
while end > start and isspace(ptr[end-1]):
```
or something like that. I guess your current algorithm probably still works, since the first loop doesn't check the last character, but the second loop will. But please remove the comment about the `\0` and maybe clarify what the idea is here instead.

fknfilewalker: Yes you are right

laszlokindrat (approved):
Nice, thanks for adding this! Followup idea: `lstrip` and `rstrip` for `StringRef`, and delegate `String.rstrip` and friends to this StringRef implementation.

fknfilewalker:
> and delegate `String.rstrip` and friends to this StringRef implementation.

`String.rstrip` and `lstrip` allow the removal of entire char sets which cannot be done with `StringRefs` since this would mean changing the underlying memory?

PR #2690, opened 2024-05-16
[stdlib] Bump compatible compiler version to 2024.5.1619

patrickdoc:


JoeLoser (approved):

gryznar:
@patrickdoc, @JoeLoser maybe it will be a good idea to add compiler changelog? As far as it sources are closed, it is not easy to track changes, especially in nightly branch

PR #2691, opened 2024-05-16
[stdlib] Add more safeguards in the `String` constructors

gabrieldemarmiesse:

Fix https://github.com/modularml/mojo/issues/2392

Also related to https://github.com/modularml/mojo/issues/2678

### Description of the problem:

sddefault

EDIT: Please note that now, an empty string will allocate one byte on the heap for the null terminator. This will degrade the performance of some programs, the tradeoff is more safety. This performance penalty will go away with SSO, as the null terminator will be on the stack, which I hope can be implemented soon-ish when https://github.com/modularml/mojo/issues/2637 is fixed

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -687,6 +687,8 @@ struct String(
         Args:
             impl: The buffer.
         """
+        if len(impl) == 0:
+            impl.append(0)
         debug_assert(
             impl[-1] == 0,
             "expected last element of String buffer to be null terminator",
@@ -704,6 +706,7 @@ struct String(
     fn __init__(inout self):
         """Construct an uninitialized string."""
         self._buffer = Self._buffer_type()
+        self._buffer.append(0)
 
     @always_inline
     fn __init__(inout self, str: StringRef):
@@ -783,8 +786,10 @@ struct String(
         """
         # we don't know the capacity of ptr, but we'll assume it's the same or
         # larger than len
-        self._buffer = Self._buffer_type(
-            unsafe_pointer=ptr.bitcast[UInt8](), size=len, capacity=len
+        self = Self(
+            Self._buffer_type(
+                unsafe_pointer=ptr.bitcast[UInt8](), size=len, capacity=len
+            )
         )
 
     @always_inline
@@ -798,9 +803,13 @@ struct String(
             ptr: The pointer to the buffer.
             len: The length of the buffer, including the null terminator.
         """
-        self._buffer = Self._buffer_type()
-        self._buffer.data = UnsafePointer(ptr.address)
-        self._buffer.size = len
+        self = Self(
+            Self._buffer_type(
+                unsafe_pointer=UnsafePointer(ptr.address),
+                size=len,
+                capacity=len,
+            )
+        )
 
     @always_inline
     fn __init__(inout self, ptr: DTypePointer[DType.uint8], len: Int):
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/string.mojo
```diff
@@ -609,6 +613,7 @@ struct String(
     fn __init__(inout self):
         """Construct an uninitialized string."""
         self._buffer = Self._buffer_type()
+        self._buffer.append(0)
```
ConnorGray, line 616: Note that this change means that constructing an empty string will now allocate, where it did not beforec. That may be what we want, but its a notable change to the default representation of a `String`.

E.g. a `List[String]` default initialized to empty `String` and then later filled in would now do O(N) allocations.

stdlib/test/builtin/test_string.mojo
```diff
@@ -828,6 +828,17 @@ def test_string_mul():
...
+    assert_equal(str(String()), String())
+    assert_equal(str(String()), String(""))
+    assert_equal(str(String()), str(String("")))
+    assert_equal(str(String()), "")
+    assert_equal(str(String()), str(""))
+    assert_equal(len(String()), 0)
+
+    assert_equal(String(), String(List[Int8]()))
+
+
```
ConnorGray, line 841: One of the main benefits I can see from this change is that an empty `String` is now validly null terminator, so it can be passed to functions that are expecting a non NUL string pointer.

So if thats the desired behavior, we should add some tests for that🙂 

But, I'd argue most functions that could accept an empty string should arguably be written to accept a NUL pointer, because as mentioned then they'll optimally allow the caller to avoid allocating space just to store an empty string.1

stdlib/src/builtin/string.mojo
gabrieldemarmiesse: > E.g. a List[String] default initialized to empty String and then later filled in would now do O(N) allocations.

I'm not sure I understand this part. Can you give a small code example?

Pratically speaking, when we have SSO, empty strings will have null termintors allocated on the stack, so it will be very cheap. I'd advise that until then, we err on the side of caution about safety in String.

stdlib/test/builtin/test_string.mojo
gabrieldemarmiesse: >So if thats the desired behavior, we should add some tests for that🙂

Do you mean adding a test where if do `assert_equal(String()._buffer[0], 0)`?

stdlib/src/builtin/string.mojo
ConnorGray: That's a compelling point regarding the small-string optimization.

I'm convinced now that NUL terminating by default is not necessarily a bad behavior to aim for long term🙂 

> Can you give a small code example? I'm not sure I see the regression you are talking about.

Not a regression per se, but consider:

```mojo
var list = List[String](String(), String(), String(), String(), String())

for count in len(list):
    list[count] = str(count)
```

Before this change, this program only allocates a `String` for each element once.

After this change, first five empty strings will be allocated, and then overwritten with the final content of the list as initialized by the `for` loop.

I don't necessarily think this is a convincing reason on its own not to move forward with this change. I'm just offering this as an example of the change in behavior this PR brings.

gabrieldemarmiesse: Thank you for the insight. Indeed it's a good example where we may have degraded performance. With SSO, that shouldn't be an issue. I'd be interested to know if a null terminator needs to be a requirement for String, but I guess this can be discussed after SSO is done

stdlib/test/builtin/test_string.mojo
gabrieldemarmiesse: Done :)

gabrieldemarmiesse:
I'll close this PR in favor of the SSO work. We can always see after SSO if this is still relevant.

PR #2699, opened 2024-05-17
[stdlib] Add contains to InlineArray

lsh:


.helix/languages.toml
```diff
@@ -0,0 +1,19 @@
+[language-server.mojo-lsp]
```
JoeLoser, line 1: **Suggestion** Did you mean to check this in?  If so, can you split this out into a separate PR?  This file won't be managed by copybara, so when I import this PR internally, it would never make its way back out.  The fileset managed by copybara is explicit and not all of the files in the mojo repo right now.

JoeLoser (approved):

JoeLoser:
!sync

.helix/languages.toml
JoeLoser: I imported the PR right now, but it won't include this file when I land it FYI.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -432,3 +433,36 @@ struct InlineArray[ElementType: CollectionElement, size: Int](Sized):
...
+
+        Returns:
+            True if the value is contained in the array, False otherwise.
+        """
+        constrained[
+            _type_is_eq[T, Self.ElementType](),
+            "T must be equal to Self.ElementType",
+        ]()
+
+        @unroll
```
JoeLoser, line 464: FYI I removed this internally since `@unroll` is removed now.  We introduce a `@parameter for` that I think we can apply here, but I decoupled it for now so we can test the hot-off-the-presses language feature in a follow-up PR.  I dropped a TODO here so we don't forget.

PR #2702, opened 2024-05-17
[stdlib] List.__eq__ and List.__ne__ for Rhs_t:ComparableCollectionElement

rd4com:
Hello,

here is an implementation of  ```List.__eq__``` and ```List.__ne__```  that makes it possible to do:

```mojo
var x = List(1,2,3)
var y = List(1,2,3)
if x == y: print("same values")
if x != y: print("not same values")
```
&nbsp;

One of the downside is that part of the implementation is done with ```constrained```, 

this is not ideal because users won't be aware of some errors until ```m̀ojo build main.mojo``` shows the constrained failed.

But it is necessary to have the ```constrained``` because rhs could be a ```list[Bool]``` and self a ```list[Int]```, for example.


docs/changelog.md
```diff
@@ -275,6 +275,17 @@ what we publish.
   trailing whitespaces. ([PR #2683](https://github.com/modularml/mojo/pull/2683)
   by [@fknfilewalker](https://github.com/fknfilewalker))
 
+- `List` now conditionally supports ```__eq__``` and ```__ne__```. For example:
+  ```python
```
ChristopherLR, line 279:
```suggestion
  ```mojo
```

stdlib/test/collections/test_list.mojo
```diff
@@ -759,6 +759,32 @@ def test_list_contains():
     # assert_equal(List(0,1) in y,False)
 
 
+def test_list_eq():
+    var a = List[Int](1, 2, 3)
+    var b = List[Int](1, 2, 3)
+    var c = List[Int](0, 1, 2)
+    var d = List[Int](0, 1, 2, 3)
+    assert_true(a == b)
```
ChristopherLR, line 767: Wonder if it's worth using `assert_equal` and `assert_not_equal`? Guessing we get the lhs + rhs values propagated
```suggestion
    assert_equal(a, b)
```

rd4com: @ChristopherLR Let's use both ? if you could create a new suggestion with both ?

ChristopherLR:
```suggestion
    assert_true(a == b)
    assert_equal(a, b)
```

stdlib/test/collections/test_list.mojo
```diff
@@ -759,6 +759,32 @@ def test_list_contains():
...
+    assert_false(a == c)
+    assert_false(a == d)
```
ChristopherLR, line 769:
```suggestion
    assert_false(a == d)
    assert_not_equal(a, d)
```

rd4com: ```assert_false``` require the arguments to be ```Testable``` which includes ```EqualityComparable```,

this would not work because List is not ```EqualityComparable``` and ```Stringable```;

What the ```__eq__``` implementation does is to exist if T of a specific list is ```ComparableCollectionElement```.

&nbsp;

(see ```fn __str__[U: RepresentableCollectionElement](self: List[U]) -> String:``` of ```List```)





It would require the "conditional conformance"

rd4com:
Do not accept the PR please even if test passes, 

@soraros  pointed out that rebind should not work with memory-only types!

rd4com:
Ready for reviews ! :partying_face:

stdlib/test/collections/test_list.mojo
ChristopherLR: Thanks for the info! I'm assuming that at some stage `assert_equal` will work for collections? I wonder If it's worth (in a separate MR) attempting to enhance the testing library to work for `collections` that contain `ComparableCollectionElement`?

rd4com: Not sure, @ChristopherLR , if you think it's a good idea you can always develop it

rd4com:
When @soraros offered that review on another PR, `rebind` was used on `value:T`(`CollectionElement`), 
now, rebind is used on the `Reference[T]`, which is `register_passable` (not memory only).

This is also why there is a: `[stdlib] Fix Tuple.__contains__` PR  (#2782 )

*see `VariadicPack.get_element`*: https://github.com/modularml/mojo/blob/bf73717d79fbb79b4b2bf586b3a40072308b6184/stdlib/src/builtin/builtin_list.mojo#L579)

JoeLoser:
> When @soraros offered that review on another PR, `rebind` was used on `value:T`(`CollectionElement`), now, rebind is used on the `Reference[T]`, which is `register_passable` (not memory only).
> 
> This is also why there is a: `[stdlib] Fix Tuple.__contains__` PR (#2782 )
> 
> _see `VariadicPack.get_element`_:
> 
> https://github.com/modularml/mojo/blob/bf73717d79fbb79b4b2bf586b3a40072308b6184/stdlib/src/builtin/builtin_list.mojo#L579

Makes sense, thanks for clarifying! So, if I understand correctly, this PR should be ready to go then, yeah?  If so, do you mind resolving the merge conflicts?

Mogball:
> > Do not accept the PR please even if test passes,
> > @soraros pointed out that rebind should not work with memory-only types!
> 
> @soraros do you mind elaborating given the test case with a memory only type? @Mogball do you have any suggestions in not using rebind here and other places we're trying to implement these sort of things without conditional conformance?

My suggestion is to never use rebind. We should rename it to `__rebind`

rd4com:
I'm happy to follow @Mogball 's expert analysis,
let's just close the PR and revisit the idea once new features lands,
i would not be comfortable forming a reference out of an bitcasted `UnsafePointer`,
here is a suggestion for the renaming of rebind: `UnsafeRebind` :+1:  *(ping to @Mogball)*
@JoeLoser can i close this PR ?

stdlib/src/collections/list.mojo
```diff
@@ -790,3 +790,61 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             if i[] == rebind[T2](value):
                 return True
         return False
+
+    @always_inline
+    fn __eq__[
+        inferred Rhs_t: ComparableCollectionElement
```
JoeLoser, line 796: **Question** Do we need `ComparableCollectionElement` or would `EqualityCollectionElement` work instead (`CollectionElement` + `EqualityComparable`)?

stdlib/test/collections/test_list.mojo
```diff
@@ -759,6 +759,98 @@ def test_list_contains():
...
+    var c = List[Int](0, 1, 2)
+    var d = List[Int](0, 1, 2, 3)
+    assert_true(a.__eq__(b))
+    assert_false(a.__eq__(c))
+    assert_false(a.__eq__(d))
+    assert_true(a == b)
+    assert_false(a == c)
+    assert_false(a == d)
+
+    var e = List[StringLiteral]("Mojo", "Is", "Awesome")
```
JoeLoser, line 777: **Suggestion** Want to use `alias` instead here for the variables and we can make sure `eq` and `ne` work in comp-time?

rd4com:
@JoeLoser , did my best to implement both reviews but it does not work anymore

JoeLoser:
> @JoeLoser , did my best to implement both reviews, the tests do not passes

@laszlokindrat any idea if this is a parser issue in the test failures?

laszlokindrat:
> @laszlokindrat any idea if this is a parser issue in the test failures?

Possibly. @rd4com Can you try adding `EqualityCollectionElement` to `Int` (i.e. make it explicitly conform) and see if the tests still fail.

stdlib/src/collections/list.mojo
```diff
@@ -858,6 +858,64 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
+        Parameters:
+            Rhs_t: The inferred type of rhs.T.
+
+        Returns:
+            True if this List is equal to the RHS List and False otherwise.
+        """
+
+        @parameter
+        if not _type_is_eq[T, Rhs_t]():
+            return False
```
gabrieldemarmiesse, line 881: Isn't this situation always a user error? 

As a user, if I'm comparing two list of struct with different types, I would most of the time want the compiler to tell me at compile-time since the result is always the same. 

Maybe let's disallow this, it will remove the need for a `rebind`, that will make this PR easier to merge, and if later on, we see the need for this function to check for equality of the types, we can add it. I would advocate for the "wait-and-see" approach.

Looking at mypy, they provide a command line argument to disallow this:
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict-equality

if enabled, you get the following error:

```
trying_stuff.py:5: error: Non-overlapping equality check (left operand type: "list[int]", right operand type: "list[str]")  [comparison-overlap]
Found 1 error in 1 file (checked 1 source file)
```
@rd4com what do you think about this solution?

soraros: I think we should use the new parametric self to implement this.

rd4com: @soraros , @laszlokindrat , @JoeLoser , @gabrieldemarmiesse 
I messed up the git tree and tried everything:
feel free to make a PR by using this one, i give up this one.

rd4com:
@soraros , @laszlokindrat , @JoeLoser , @gabrieldemarmiesse
I messed up the git tree and tried everything:
feel free to make a PR by using this one, i give up this one.

PR #2704, opened 2024-05-17
[stdlib] Add three __delitem__ method overloads for List

rd4com:
This PR includes three `__delitem__` method overloads for `List`:
- `fn __delitem__(inout self, position: Int)`
- `fn __delitem__(inout self, *positions: Int)`
- `fn __delitem__(inout self, owned positions: List[Int])`

The implementation ensure that an index will not be deleted twice, 
even if provided multiple times.
The order of removal is the reverse of sorted indexes.

&nbsp;

Example:
```mojo
  x = List[Int](0,10,20,30,40)
  x.__delitem__(0,4)
  for i in x: print(i[]) #10,20,30
  ```
&nbsp;

@rparolin proposed using `__delitem__` directly instead of `del_batch_of_indexes`

#### Suggestion for the future:

```del my_list[0,1,4,5]```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -470,6 +470,19 @@ what we publish.
 - Added `SIMD.__repr__` to get the verbose string representation of `SIMD` types.
 ([PR #2728](https://github.com/modularml/mojo/pull/2728) by [@bgreni](https://github.com/bgreni))
 
+- `List` now has a `__delitem__` method.
+  ([PR #2704](https://github.com/modularml/mojo/pull/2704) by [@rd4com](https://github.com/rd4com))
+
+  It provides a way to delete one or multiple(in a batch) elements by index:
+  - Even if provided twice, it won't delete two times the same index.
+  - The order of removal is reverse of sorted provided indexes.
+
+  ```mojo
+  x = List[Int](0,10,20,30,40)
+  x.__delitem__(0,4)
+  for i in x: print(i[]) #10,20,30
+  ```
+
 ### 🦋 Changed
 
 - `Coroutine` now requires a lifetime parameter. This parameter is set
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -858,6 +858,73 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         """
         return self.data
 
+    @always_inline
+    fn __delitem__(inout self, owned positions: List[Int]):
+        """
+        Remove one or multiples elements at specific positions in a list.
+        Order of removal is reverse of sorted positions.
+        Will only remove once, any position specified multiple times.
+
+        *Example:*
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: docs/changelog.md [hunk 1]
```diff
@@ -275,6 +275,18 @@ what we publish.
...
   by [@fknfilewalker](https://github.com/fknfilewalker))
 
+- `List` now has a `delete_batch_of_indexes` method that provides a way to delete multiple elements by index in a batch.
+
+  - Even if provided twice, it won't delete two times the same index.
+  - The order of removal is reverse of sorted provided indexes.
+
+  ```mojo
+  x = List[Int](0,10,20,30,40)
+  x.delete_batch_of_indexes(0,4)
```
rparolin, line 285: Since we lack the language feature to remove subranges via the `del` keyword, I propose we implement the `__delitem__` member function and call it directly until we land the language feature enabling the use `del`.  @JoeLoser Thoughts?

rparolin (changes requested):

docs/changelog.md
rd4com: Hello, 

new commit changed the method name to `__delitem__` and includes an overload that takes only one position as an integer.

That way, it won't create a set and sort an array if there is only one index. ( ```_ = self.pop(position)```)

- Should there be any debug_asserts that asserts that `len(self) > index` and `index >= 0` ?

- Should the methods just don't do ```self.pop(position)``` if position is out of bound ?

JoeLoser:
Sorry I'm just coming back to this.  Can you share a bit more about your use case for this?

rd4com:
Yep, the idea is to be able to pop multiple items from the list.
If `x=List("a","b","c","d")` and the user pop `0, 1 , 2`,
`"d"` would be the only item in the list.
(it is done by popping from highest index to lowest index)

Why it can be useful to users:
If we would pop `0` first, then `1` would loose it's first meaning and thus remove `"c"`.
With a method like this, users can just pop multiple indexes without worries.

Not sure about if it is necessary or if python have a similar method,
if it is not, maybe just leave it aside or stash it ?

JoeLoser:
> Yep, the idea is to be able to pop multiple items from the list. If `x=List("a","b","c","d")` and the user pop `0, 1 , 2`, `"d"` would be the only item in the list. (it is done by popping from highest index to lowest index)
> 
> Why it can be useful to users: If we would pop `0` first, then `1` would loose it's first meaning and thus remove `"c"`. With a method like this, users can just pop multiple indexes without worries.
> 
> Not sure about if it is necessary or if python have a similar method, if it is not, maybe just leave it aside or stash it ?

It makes sense, but I'm not so sure it belongs in the core `List` API as is.  I'd recommend maintaining this functionality downstream in your codebase if you find it useful and/or stashing the feature as you mentioned.  Thanks for suggested API though!

PR #2705, opened 2024-05-17
[stdlib] Add `split()` to `StringRef`

fknfilewalker:
This PR adds a `split()` function to `StringRef` and delegates `String.split()` to it

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -172,6 +172,10 @@ what we publish.
       return a
   ```
 
+- `StringRef` now implements `split()` which can be used to split a
+  `StringRef` into a `List[StringRef]` by a delimiter.
+  ([PR #2705](https://github.com/modularml/mojo/pull/2705) by [@fknfilewalker](https://github.com/fknfilewalker))
+
 ### 🦋 Changed
 
 - More things have been removed from the auto-exported set of entities in the `prelude`
--- a/stdlib/src/utils/stringref.mojo
+++ b/stdlib/src/utils/stringref.mojo
@@ -574,6 +574,41 @@ struct StringRef(
             end -= 1
         return StringRef(ptr + start, end - start)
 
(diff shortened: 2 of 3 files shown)


fknfilewalker:
Would it be better to first count the occurrence of the delimiter to properly reserve space?

stdlib/test/builtin/test_string.mojo
```diff
@@ -223,6 +223,47 @@ fn test_stringref_from_dtypepointer() raises:
     assert_equal(a, b)
 
 
+fn test_stringref_split() raises:
```
JoeLoser, line 226: **Suggestion** Should we put this test in `stdlib/test/builtin/test_stringref.mojo` instead?

JoeLoser:
> Would it be better to first count the occurrence of the delimiter to properly reserve space?

Would this require two passes through the input string then?

jackos (approved):
Thanks @fknfilewalker squashed and rebased, and did @JoeLoser's recommendation of moving the testes into `stringref.mojo`. There's been changes to `string.mojo` so that `split` now has an overload where sep=None. If you want to simplify those a PR would be welcome thanks.

jackos:
!sync

martinvuyk:
AFAIK we are migrating away from `StringRef` to `StringSlice`, it looks like this implementation will be easily moved to `StringSlice`.

There is also another detail here that there are some edge cases covered by `String.split()` that this needs to be updated to ([see here](https://github.com/modularml/mojo/blob/02fc624b1fc05795577b096c5616f4cc02e0a316/stdlib/src/collections/string.mojo#L1612)) and it is still missing the `maxplit` argument.

@fknfilewalker I fixed some issues on the current nightly `String.split()` versions in PR #3528 that you can take inspiration from for a follow up PR if you want. Otherwise let me know and I'll tackle it once this is merged.

Issue #2658, opened 2024-05-15
[stdlib] Implement `__contains__` for `Tuple`, `List`, `ListLiteral` (almost)


laszlokindrat:
Now that we have `ComparableCollectionElement`, we can try to implement `__contains__` for some common collection types using a workaround similar to what was employed in https://github.com/modularml/mojo/pull/2190. It's possible that the compiler would be able to correctly pick up a `__contains__` overload with `@staticmethod`, e.g. something like this might just work:
```mojo
struct Bucket[T: CollectionElement]:
    @staticmethod
    fn __contains__[S: ComparableCollectionElement](self: Bucket[S], elem: s) -> Bool: ...

var b: Bool = MyStuff() in Bucket[MyStuff]()
```
Even if `in` doesn't work with this, the implementation shouldn't have to change much once we roll out conditional conformance.

ChristopherLR:
Tried something like this:

```mojo
struct Bucket[T: CollectionElement]:
    var elements: List[T]

    fn __init__(inout self, elements: List[T]):
        self.elements = elements

    @staticmethod
    fn __contains__[C: ComparableCollectionElement](self: Bucket[C], borrowed elem: C) -> Bool:
        for i in range(self.elements.size):
            if elem.__eq__(self.elements[i]):
                return True
        return False
```

However, i'm getting:

```mojo
error: special method may not be a static method
    fn __contains__[C: ComparableCollectionElement](self: Bucket[C], borrowed elem: C) -> Bool:
       ^
mojo: error: failed to parse the provided Mojo source module
```

If I remove `@staticmethod`:
```mojo
error: 'self' argument must have type 'Bucket[T]', but actually has type 'Bucket[C]'
    fn __contains__[C: ComparableCollectionElement](self: Bucket[C], borrowed elem: C) -> Bool:
                                                    ^     ~~~~~~~~~
mojo: error: failed to parse the provided Mojo source module
```

ChristopherLR:
Is there something similar to c++ `enable_if` that i'm missing? https://en.cppreference.com/w/cpp/types/enable_if

rd4com:
@laszlokindrat , @ChristopherLR , if self can be a reference? this parametrization could do the job:

```mojo
struct Bucket[T: CollectionElement]:
    var elements: List[T]

    fn __init__(inout self):
        self.elements = List[T]()

    fn __contains__[C: ComparableCollectionElement](
        self: Reference[Bucket[C]], elem: C
[...]

PR #2709, opened 2024-05-17
[stdlib] Tuple.__contains__

rd4com:
Hello,

here is an implementation of ```Tuple.__contains__```,

it makes it possible to do: 
```mojo
  var x = Tuple(1,2,True)
  if 1 in x: print("x contains 1")
```

@laszlokindrat suggested a ping for the PR in #2658

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -378,11 +378,6 @@ what we publish.
   module to the new `utils.numerics` module: `isfinite`, `isinf`, `isnan`,
   `nan`, `nextafter`, and `ulp`. The functions continue to be exposed in the
   `math` module.
-- `Tuple()` now supports `__contains__`. For example:
-  ```mojo
-  var x = Tuple(1,2,True)
-  if 1 in x: print("x contains 1")
-  ```
 
 ### ❌ Removed
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -311,6 +311,13 @@ what we publish.
 - Added `atof()` function which can convert a `String` to a `float64`.
   ([PR #2649](https://github.com/modularml/mojo/pull/2649) by [@fknfilewalker](https://github.com/fknfilewalker))
 
+- ```Tuple``` now supports ```__contains__```([PR #2709](https://github.com/modularml/mojo/pull/2709) by
+  [@rd4com](https://github.com/rd4com))
+  ```mojo
+  var x = Tuple(1,True,"Mojo is awesome")
+  if "Mojo is awesome" in x: print("Mojo is awesome 🔥")
+  ```
+
 ### 🦋 Changed
 
 - The `let` keyword has been completely removed from the language. We previously
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -378,6 +378,11 @@ what we publish.
   module to the new `utils.numerics` module: `isfinite`, `isinf`, `isnan`,
   `nan`, `nextafter`, and `ulp`. The functions continue to be exposed in the
   `math` module.
+- `Tuple()` now supports `__contains__`. For example:
+  ```mojo
+  var x = Tuple(1,2,True)
+  if 1 in x: print("x contains 1")
+  ```
 
 ### ❌ Removed
 
--- a/stdlib/src/builtin/tuple.mojo
+++ b/stdlib/src/builtin/tuple.mojo
@@ -203,20 +203,22 @@ struct Tuple[*element_types: Movable](Sized, Movable):
         """
 
         var result = False
+
         @parameter
         for i in range(len(VariadicList(element_types))):
+
             @parameter
-            if _type_is_eq[T,element_types[i]]():
+            if _type_is_eq[T, element_types[i]]():
                 var tmp_ref = self.__refitem__[i]()
                 var tmp = rebind[
                     Reference[
                         T,
                         tmp_ref.is_mutable,
                         tmp_ref.lifetime,
-                        tmp_ref.address_space
+                        tmp_ref.address_space,
                     ]
                 ](tmp_ref)
                 if tmp[].__eq__(value):
                     result = True
- 
+
         return result
--- a/stdlib/src/builtin/tuple.mojo
+++ b/stdlib/src/builtin/tuple.mojo
@@ -183,33 +183,40 @@ struct Tuple[*element_types: Movable](Sized, Movable):
 
     @always_inline("nodebug")
     fn __contains__[
-        RHS_T: ComparableCollectionElement
-    ](self: Reference[Self], owned rhs: RHS_T) -> Bool:
+        T: ComparableCollectionElement
+    ](self: Self, value: T) -> Bool:
         """Verify if a given value is present in the tuple.
 
         ```mojo
         var x = Tuple(1,2,True)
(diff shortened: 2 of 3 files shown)


stdlib/test/builtin/test_tuple.mojo
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo %s
+
+from testing import assert_true, assert_false
+
+
+def test_tuple_contains():
+    var a = (1, True)
```
laszlokindrat, line 19: Can you please change this to not have `1` (or) `0` in it? Since Bool would convert to that implicitly, it can get tricky if something goes wrong with the call emission.

stdlib/test/builtin/test_tuple.mojo
```diff
@@ -0,0 +1,27 @@
...
+    assert_true(1 in a)
+    assert_true(a.__contains__(1))
+    assert_true(True in a)
```
laszlokindrat, line 22: Can you please add equivalent tests that call into `__contains__` directly? I'm still paranoid about these features.

stdlib/test/builtin/test_tuple.mojo
```diff
@@ -0,0 +1,27 @@
...
+    assert_false(2 in a)
```
laszlokindrat, line 23: Can you please add an `assert_false` case with some different type, e.g. `assert_false("foo" in a)`?

laszlokindrat (approved):
Pretty cool, thank you!

rd4com:
@laszlokindrat Yes, sorry, the tests are more complete now, do you see more cases to add ?

There is a test failing in test_list.mojo for macosx, not sure what to do ?

stdlib/src/builtin/tuple.mojo
```diff
@@ -178,3 +180,36 @@ struct Tuple[*element_types: Movable](Sized, Movable):
...
+        Returns:
+            True if the value is contained in the tuple, False otherwise.
+        """
+        var result = False
+
+        @parameter
+        fn SingleIteration[Index: Int]():
+            if _type_is_eq[RHS_T, element_types[Index.value]]():
+                result |= rhs.__eq__(
+                    rebind[RHS_T](self.get[Index, element_types[Index.value]]())
```
soraros, line 210: @laszlokindrat I don't think `rebind` will work for mem type.

rd4com: Good observation! I changed the implementation to not use rebind and changed rhs to owned instead of borrowed.

Tests runs fine on ubuntu locally but somehow fails on macosx here, and stop the testing process of ubuntu?

Please see results on the (macos-14)  tab: 
```mojo
Please file a bug report.
[6505:28327:20240517,222150.691559:WARNING crash_report_exception_handler.cc:257] UniversalExceptionRaise: (os/kern) failure (5)
```

docs/changelog.md
```diff
@@ -275,6 +275,13 @@ what we publish.
   trailing whitespaces. ([PR #2683](https://github.com/modularml/mojo/pull/2683)
   by [@fknfilewalker](https://github.com/fknfilewalker))
 
+- `Tuple()` now supports `__contains__`. ([PR #2709](https://github.com/modularml/mojo/pull/2709)
+  by [@rd4com](https://github.com/rd4com)) For example:
+  ```mojo
+  var x = Tuple(1,2,True)
```
soraros, line 281: Nit: maybe reformat it? Same for the doc string.

stdlib/src/builtin/tuple.mojo
laszlokindrat: Why does the rhs need to be owned?

rd4com: In hope to not worry about borrowed references and manually extend lifetime with _=^,
Last commit passes tests, but there must be a better way to cast a reference than that ?

stdlib/src/builtin/tuple.mojo
```diff
@@ -178,3 +180,36 @@ struct Tuple[*element_types: Movable](Sized, Movable):
             The tuple element at the requested index.
         """
         return rebind[T](self[i])
+
+    @always_inline("nodebug")
+    fn __contains__[
+        RHS_T: ComparableCollectionElement
```
laszlokindrat, line 186: Please use PascalCase for type parameters.

stdlib/src/builtin/tuple.mojo
laszlokindrat: > In hope to not worry about borrowed references and manually extend lifetime with _=^,

Unfortunately, `owned` can cause unintended copies, so I would rather avoid it.

stdlib/src/builtin/tuple.mojo
rd4com: Yes, sorry about that, i do remember it all the time i see parameters now so thanks  :+1: 

Have got a better implementation, safer with references, please do not merge this one yet if it was the plan;

I'll do a force-push later-on, still don't know if it will work with the rebased!

stdlib/src/builtin/tuple.mojo
rd4com: Hello, i forced-push a commit with changes related to both suggestions (no owned rhs and no rebind of T),
 i suggest rebinding the reference instead of T, since ```Reference``` is ```register_passable``` but not always T.
Extended test with memory-only types passes for ```Tuple.__contains__``` on last commit!
If both of you likes it, we could resolve this specific review conversation

rd4com:
The force-push worked with the rebased branch :partying_face: :
- it includes a new ```@parameter``` optimization that returns false if ```T``` not in ```self.element_types```
  *(before ```var result:Bool = False```)*
- no rebind of memory-only types, we rebind the ```Reference```, which is ```@register_passable```

Todo:
- when ```@parameter for``` can do returns, we will be able to return True the first time that ```self[i] == value```,
instead of continuing doing result |= for every ```_type_is_eq[element_types[i],T]()```.
There would be no need for ```var result:Bool```

ℹ️ Note that the rebase is on ```2024.5.1805``` not  ```2024.5.1905```

laszlokindrat:
!sync

laszlokindrat (approved):
Thanks again! I think this is a useful functionality, and while we can't do it perfectly yet, it does deliver a useful feature, especially for people coming from Python. There are very likely some sharp edges around the current behavior. If you can think of any, please open a PR to document them.

rd4com:
@laszlokindrat , This is great! this is a cool feature, i'd be happy to use it too :smile: .
&nbsp;

Once auto-dereference lands:
- It will probably require some changes, as it might rebind the dereferenced value instead of the ```Reference```.

If there is any way you can add this to sort of a todo list to make sure ```Tuple.__contains__``` still works?

laszlokindrat:
> @laszlokindrat , This is great! this is a cool feature, i'd be happy to use it too 😄 .  
> 
> Once auto-dereference lands:
> 
> * It will probably require some changes, as it might rebind the dereferenced value instead of the `Reference`.
> 
> If there is any way you can add this to sort of a todo list to make sure `Tuple.__contains__` still works?

The tests should prevent it from being broken.

PR #2710, opened 2024-05-17
[stdlib] Add `startswith()` and `endswith()` to `StringRef`

fknfilewalker:
Also delegates String functions to StringRef


fknfilewalker:
Especially for this PR it would be nice to have a way of creating a `StringRef` from a `String` with a left and right offset, without resorting to ptrs

JoeLoser (approved):

JoeLoser:
!sync

PR #2711, opened 2024-05-17
[stdlib] Make `String.split()` default to whitespace & fix behavior to be pythonic

### LINKED ISSUE #2686: [stdlib] [Feature Request] Refactor `String` to not have so many raises and use Optional
### Review Mojo's priorities

### What is your request?

As title.

Could have a `fn () raising -> Int` and an `fn () -> Optional[Int]`
```mojo
fn atol(str: String, base: Int = 10) raises -> Int:
    return _atol(str._strref_dangerous(), base)

fn atol_safe(str: String, base: Int = 10) -> Optional[Int]:
    try:
      return _atol(str._strref_dangerous(), base)(str, base)
    except:
      return None
```

This could just return a list with one item, it being the whole String. `if not delimiter: return List[String](self)`
```mojo
    fn split(self, delimiter: String) raises -> List[String]:
        """Split the string by a delimiter.

        Args:
          delimiter: The string to split on.

        Returns:
          A List of Strings containing the input split by the delimiter.

        Raises:
          Error if an empty delimiter is specified.
        """
        if not delimiter:
            raise Error("empty delimiter not allowed to be passed to split.")
```

This will most certainly always be checked against anyway with `if index == -1: ...` so it could also be made an Optional[Int]
```mojo
fn find(self, substr: String, start: Int = 0) -> Int:
        """Finds the offset of the first occurrence of `substr` starting at
        `start`. If not found, returns -1.

        Args:
          substr: The substring to find.
          start: The offset from which to find.

        Returns:
          The offset of `substr` relative to the beginning of the string.
        """

        return self._strref_dangerous().find(
        )
```

### Any other details?

_No response_

martinvuyk:
Closes #2686
`String.split()` now defaults to whitespace and has pythonic behavior in that it removes all adjacent whitespaces by default.


martinvuyk:
Closes #2686, closes #2713

siitron:
Maybe this should be split into two separate issues/PRs, I don't see how they relate.

martinvuyk:
> Maybe this should be split into two separate PRs so that they can be reviewed independently? I feel like you address multiple things at once here.

ok I'll get the `atol_safe() -> Optional[Int]` into another PR

stdlib/src/builtin/string.mojo
```diff
@@ -1405,36 +1441,172 @@ struct String(
...
+                    output.append(self[lhs:])
+                    break
+                items += 1
+
+            output.append(self[lhs:rhs])
+            lhs = rhs + len(separator)
+
+        return output
+
+    fn split[separator: String = "", maxsplit: Int = -1](self) -> List[String]:
```
laszlokindrat, line 1534: Can you please remove this overload? It's not clear at this point if such optimization would pull it's weight.

stdlib/src/builtin/string.mojo
```diff
@@ -570,11 +570,44 @@ fn _is_ascii_lowercase(c: UInt8) -> Bool:
 # ===----------------------------------------------------------------------===#
 
 
+fn isspace_python(s: String) -> Bool:
```
laszlokindrat, line 573: Can you please move this to a hidden method on the struct, i.e. `String._isspace_python()`?

stdlib/src/builtin/string.mojo
```diff
@@ -570,11 +570,44 @@ fn _is_ascii_lowercase(c: UInt8) -> Bool:
...
+    """Determines whether the given String is a python
+    whitespace String. This corresponds to Python's
+    [universal separators](
+        https://docs.python.org/3/library/stdtypes.html#str.splitlines)
+    `" \\t\\n\\r\\f\\v\\x1c\\x1e\\x85\\u2028\\u2029"`.
```
laszlokindrat, line 578: please format to line width

stdlib/src/builtin/string.mojo
```diff
@@ -1405,36 +1441,172 @@ struct String(
...
+        [universal separators](
+            https://docs.python.org/3/library/stdtypes.html#str.splitlines)
+        `" \\t\\n\\r\\f\\v\\x1c\\x1e\\x85\\u2028\\u2029"`.
+
+        Returns:
+            True if the String is one of the whitespace characters
+                listed above, otherwise False."""
+        return isspace_python(self)
+
+    fn split(self, owned separator: String, maxsplit: Int = -1) -> List[String]:
```
laszlokindrat, line 1456: nit: use `sep` instead of `separator` for python compatilibitly. Also, why does this have to be `owned`?
```suggestion
    fn split(self, sep: String, maxsplit: Int = -1) -> List[String]:
```

stdlib/src/builtin/string.mojo
```diff
@@ -1405,36 +1441,172 @@ struct String(
             substr._strref_dangerous(), start=start
         )
 
-    fn split(self, delimiter: String) raises -> List[String]:
-        """Split the string by a delimiter.
+    fn isspace(self) -> Bool:
```
laszlokindrat, line 1444: This is a public API. Could you please add (direct) tests for this?

laszlokindrat (commented):
Thanks for working on this! I have a couple asks before I do a more thorough review, but I like the direction!

martinvuyk:
@laszlokindrat added you suggestions though I just put the isspace_python inside the String.isspace method since it's basically just checking for a String inside a List. Changed the tests to target that method and added another one for cases that shouldn't be a space

stdlib/test/builtin/test_string.mojo
```diff
@@ -594,11 +594,58 @@ fn test_rfind() raises:
...
+        String("\x1e"),
+        String("\x85"),
+        String(_line_sep_utf8),
+        String(_paragraph_sep_utf8),
+    )
+    for i in _universal_separators:
+        assert_true(i[].isspace())
+
+    for i in List[String]("not", "space", "", "s", "a", "c"):
+        assert_false(i[].isspace())
```
laszlokindrat, line 620: Can you please refactor into a separate test case?

stdlib/src/builtin/string.mojo
```diff
@@ -1405,36 +1408,109 @@ struct String(
...
+            String("\v"),
+            String("\f"),
+            String("\x1c"),
+            String("\x1e"),
+            String("\x85"),
+            String(List[UInt8](0x20, 0x5C, 0x75, 0x32, 0x30, 0x32, 0x38)),
+            String(List[UInt8](0x20, 0x5C, 0x75, 0x32, 0x30, 0x32, 0x39)),
+        )
+
+    fn split(self, sep: String = "", maxsplit: Int = -1) -> List[String]:
```
laszlokindrat, line 1439: The behavior of returning an empty string for `sep=""` is kind of strange and actually breaks an invariant: `sep.join(some_str.split(sep)) == some_str` should be true for any `sep` and any `some_str`. Python simply rejects empty separators (with `ValueError: empty separator`), and we could certainly do this. The (IMO better) alternative is to return a list of the characters. WDYT?

stdlib/test/builtin/test_string.mojo
```diff
@@ -594,11 +594,58 @@ fn test_rfind() raises:
...
+
+    # empty separators default to whitespace
+    var d = String("hello world").split("")
+    assert_true(d[0] == "hello", d[1] == "world")
+    d = String("hello \t\n\n\v\fworld").split("\n")
+    assert_true(d[0] == "hello \t" and d[1] == "" and d[2] == "\v\fworld")
```
laszlokindrat, line 626: While explicitly tests are good, a great way to improve coverage for this would be to check for the invariant `sep.join(some_str.split(sep)) == some_str`.

stdlib/src/builtin/string.mojo
laszlokindrat: You can check this by running `"   ".split(" ")` and `"   ".split()` in python.

martinvuyk: > to return a list of the characters. WDYT?
I think it makes a lot of sense, though it adds yet another branch to the code that could be parametrized... but yeah it's inevitable. I hadn't actually tested the `.split("")` with python so thanks for the save.

martinvuyk:
@laszlokindrat had to add an exception to the test. If the separator to test for `sep.join(some_str.split(sep)) == some_str` is a whitespace char the text has to be split and joined again otherwise if it was " " for example it would eat up all adjacent whitespace-like chars
```mojo
    for i in range(len(text)):
        var sep = text[i]
        var joined = String("")
        for it in text.split(sep):
            joined += sep + it[]
        joined = joined[len(sep) :]

        var chars = String("")
        if sep.isspace():
            assert_true(sep in String.WHITESPACE)
            for item in text.split(sep):
                chars += sep + item[]
            chars = chars[len(sep) :]
        else:
            chars = text
        assert_true(chars == joined)
```

stdlib/src/builtin/string.mojo
siitron: `split("")` shouldn't be possible IMO, since there's an infinite number of empty strings between every symbol. If you want to get a list of the characters of a string I would opt for `list(some_string)`

martinvuyk: I understand but how would we deal with it then? IMO we shouldn't raise in a func that is just splitting, the other option is to return an empty list, or a list with the whole string as the first and only item. And I do like the idea of returning chars since it would also provide an API for splitting a string into its chars, instead of iterating and building them or using `List[String](String.as_bytes())` (I'm not sure this implicit casting even works or if the user would have to build a String for each item in the byte List)

bgreni: I'm on board with returning a list of the characters from `.split("")`. I'm not typically a fan of deviating from Python behaviour but it's not really a useful safeguard to begin with and it's not worth "colouring" this method with `raises` to avoid an altered behaviour from exactly one possible input.

stdlib/test/builtin/test_string.mojo
```diff
@@ -593,12 +593,105 @@ fn test_rfind() raises:
...
+        String(_paragraph_sep_utf8),
+    )
+    for i in _universal_separators:
+        assert_true(i[].isspace())
+
+    for i in List[String]("not", "space", "", "s", "a", "c"):
+        assert_false(i[].isspace())
+
+
 fn test_split() raises:
```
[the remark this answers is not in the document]
martinvuyk: nice catch, fixed it now in the coming commit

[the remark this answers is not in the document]
bgreni: I think it might be better to use a `sep: Optional[String] = None` for the separator to denote no user inputed `sep` rather than a magic string. It matches the Python method signature better and allows `"a  b".split(None)` which while odd is a valid Python expression.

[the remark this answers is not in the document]
martinvuyk: I added it just to see if everything goes ok, the only problem with `sep: Optional[String] = None` is that it needs to be an `Optional[StringLiteral]` currently because it can't cast an embedded StringLiteral to a String implicitly yet.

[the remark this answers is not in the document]
bgreni: Adding a `sep: StringRef` overload solves that problem for now.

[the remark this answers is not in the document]
laszlokindrat: As @bgreni suggested, adding an overload is the best we can do here (we have run into this issue before, it's a known limitation ATM). So you will basically end up with
```mojo
fn split(self, sep: StringRef, maxsplit: Int = -1) -> List[String]: ...
fn split(self, maxsplit: Int = -1) -> List[String]: ...  # delegate to the one above
```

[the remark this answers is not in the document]
martinvuyk: I think I just solved it with the trait I made WDYT? I'm working on solving the issue of `sep.join(some_str.split(sep)) == some_str` when sep=" ", there is an extra space in the original and not in the joined result for the test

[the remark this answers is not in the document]
bgreni: I think adding a trait for this might be overkill unless it could be useful otherwise?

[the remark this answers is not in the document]
martinvuyk: It's just a wrapper around `StringableCollectionElement, Sized, EqualityComparable` until we can do `fn split[T: StringableCollectionElement & Sized &  EqualityComparable]` or solve the problem of casting when inside an `Optional`

martinvuyk:
@laszlokindrat  any idea why this keeps timing out with no error?

laszlokindrat:
@martinvuyk Can you confirm that the timeout issue is resolved?

martinvuyk:
> @martinvuyk Can you confirm that the timeout issue is resolved?

it's solved. Using self.endswith(sep) removed the issue. I think it might be  a problem with `fn __getitem__`, btw. why is this recursive? should I open a PR fixing this?
```mojo
fn __getitem__[IndexerType: Indexer](self, i: IndexerType) -> String:
    var idx = index(i)
    if idx < 0:
        return self.__getitem__(len(self) + idx)

    debug_assert(0 <= idx < len(self), "index must be in range")
    var buf = Self._buffer_type(capacity=1)
    buf.append(self._buffer[idx])
    buf.append(0)
    return String(buf^)
```

laszlokindrat:
> btw. why is this recursive? should I open a PR fixing this?
> 
> ```mojo
> fn __getitem__[IndexerType: Indexer](self, i: IndexerType) -> String:
>     var idx = index(i)
>     if idx < 0:
>         return self.__getitem__(len(self) + idx)
> 
>     debug_assert(0 <= idx < len(self), "index must be in range")
>     var buf = Self._buffer_type(capacity=1)
>     buf.append(self._buffer[idx])
>     buf.append(0)
>     return String(buf^)
> ```

Probably `debug_assert`, since it uses `print` under the hood. This is a known issue, for now we need to work around it. Also, please use `Int` directly now instead of `IndexerType` (we now have implicit conversions).

stdlib/src/builtin/string.mojo
siitron: Splitting on an empty string is ambiguous and conceptually wrong. If the goal is to be a superset of Python then I don't see why we should change Python's behaviour (unless Python changes it too).

As a fix we could add an overload like:
```mojo
fn split[sep: StringRef](maxsplit=-1):
    constrained[sep != "", "empty separator"]()
```

laszlokindrat: I gave this some thought, and I think I mostly agree with @siitron and @bgreni: splitting on an empty string is ambiguous, and we shouldn't risk unexpected behavior by allowing it (even when it might feel natural to some of us). Let's keep it simple, and make this function raising. If this turns out to be a perf problem, there are ways we can deal with it, but our default should be following Python and providing safe, unsurprising APIs.

stdlib/src/builtin/string.mojo
```diff
@@ -1406,36 +1408,135 @@ struct String(
...
+        _ = String("hello world").split(" ") # ["hello", "world"]
+        # Splitting adjacent separators
+        _ = String("hello,,world").split(",") # ["hello", "", "world"]
+        # Splitting with maxsplit
+        _ = String("1,2,3").split(",", 1) # ['1', '2,3']
+        # Splitting an empty string or filled with whitespaces
+        _ = String("      ").split() # []
+        _ = String("").split() # []
+        # Splitting a string with leading, trailing, and middle whitespaces
+        _ = String("      hello    world     ").split() # ["hello", "world"]
```
laszlokindrat, line 1438: Please remove these since they describe the other overload (where you already have these).

stdlib/test/builtin/test_string.mojo
```diff
@@ -594,11 +594,58 @@ fn test_rfind() raises:
 
 
 fn test_split() raises:
-    # Reject empty delimiters
-    with assert_raises(
-        contains="empty delimiter not allowed to be passed to split."
-    ):
-        _ = String("hello").split("")
+    # empty separators default to whitespace
+    var d = String("hello world").split()
```
laszlokindrat, line 598: Do you have any tests for multi-character values of `sep`? E.g.
```mojo
"aaaaabaaa".split("aa")
"abababa".split("aba")
```

laszlokindrat (commented):
Thanks for moving this forward! Do you think you could use the `isspace` patch in the implementation of this?

laszlokindrat:
https://github.com/modularml/mojo/pull/2793 has landed internally and will be included in the next nightly. Could you please rebase after that and use `isspace` where appropriate?

martinvuyk:
@laszlokindrat I just used _isspace temporarily since it's iterating byte per byte. I think we first need to have a _StringIter that returns a StringRef according to the full utf8 multi byte length. I have a few ideas for a PR on that but can we land this meanwhile?

stdlib/test/builtin/test_string.mojo
```diff
@@ -602,11 +602,63 @@ fn test_rfind() raises:
...
+    assert_true(len(String(" ").split()) == 0)
+    assert_true(len(String("").split(" ")) == 1)
+    assert_true(len(String(" ").split(" ")) == 2)
+    assert_true(len(String("  ").split(" ")) == 3)
+    assert_true(len(String("   ").split(" ")) == 4)
+
+    d = String("abababaaba").split("aba")
+    assert_true(d[0] == "" and d[1] == "b" and d[2] == "" and d[3] == "")
+
+    # separator = "" returns all char split
```
laszlokindrat, line 649: Does this currently fail? If it's a known limitation, please add a TODO, otherwise you can uncomment or remove it.

stdlib/test/builtin/test_string.mojo
```diff
@@ -602,11 +602,63 @@ fn test_rfind() raises:
...
+
+    assert_true(len(String("").split()) == 0)
+    assert_true(len(String(" ").split()) == 0)
+    assert_true(len(String("").split(" ")) == 1)
+    assert_true(len(String(" ").split(" ")) == 2)
+    assert_true(len(String("  ").split(" ")) == 3)
+    assert_true(len(String("   ").split(" ")) == 4)
+
+    d = String("abababaaba").split("aba")
+    assert_true(d[0] == "" and d[1] == "b" and d[2] == "" and d[3] == "")
```
laszlokindrat, line 647: Probably should also check the size of `d` as well, right? Otherwise, what if `d` has a fifth element?

stdlib/src/builtin/string.mojo
```diff
@@ -1462,37 +1463,122 @@ struct String(
...
+
+        Examples:
+
+        ```mojo
+        # Splitting an empty string or filled with whitespaces
+        _ = String("      ").split() # []
+        _ = String("").split() # []
+        # Splitting a string with leading, trailing, and middle whitespaces
+        _ = String("      hello    world     ").split() # ["hello", "world"]
+        # Splitting adjacent universal newlines: # TODO
```
laszlokindrat, line 1537: Please move this TODO out of the docstring and into the body.

stdlib/src/builtin/string.mojo
```diff
@@ -1462,37 +1463,122 @@ struct String(
...
+                items += 1
+
+            output.append(self[lhs:rhs])
+            lhs = rhs + sep_len
+
+        if self.endswith(sep):
+            output.append("")
+        return output
+
+    fn split(self, *, maxsplit: Int = -1) -> List[String]:
```
laszlokindrat, line 1518: In Python, this is also specifiable positionally:
```suggestion
    fn split(self, maxsplit: Int = -1) -> List[String]:
```

laszlokindrat:
> @laszlokindrat I just used _isspace temporarily since it's iterating byte per byte. I think we first need to have a _StringIter that returns a StringRef according to the full utf8 multi byte length. I have a few ideas for a PR on that but can we land this meanwhile?

Makes sense, thanks for pushing this forward!

laszlokindrat (approved):
Very nice, thanks! Just a couple more small things, but looks great otherwise! Could you please also update the PR description and add this to the changelog?

laszlokindrat:
Looks really good, can you address the remaining two comments, or should I do it when I bring it in?

laszlokindrat:
!sync

siitron:
Just a heads up: `.split()` still doesn't match Python's behaviour fully as it doesn't accept `None` as `sep`

PR #2713, opened 2024-05-17
[stdlib] Make String.split behaviour more like Python

bgreni:
Related to #2686 as I noticed our implementation of `String.split` doesn't behave like the Python version

- Allow delimiter arg to be optional, default to all whitespace if is None
- Implement the slightly distinct behaviour between passing an explicit delimiter and not doing so
- Add maxsplit arg

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -1333,11 +1333,15 @@ struct String(
             substr._strref_dangerous(), start=start
         )
 
-    fn split(self, delimiter: String) raises -> List[String]:
+    @always_inline
+    fn split(
+        self, sep: StringRef, maxsplit: Int = Int.MAX
+    ) raises -> List[String]:
         """Split the string by a delimiter.
 
         Args:
-          delimiter: The string to split on.
+          sep: The string to split on.
+          maxsplit: The maximum number of splits to allow.
 
         Returns:
           A List of Strings containing the input split by the delimiter.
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/string.mojo
```diff
@@ -1333,36 +1333,93 @@ struct String(
...
           A List of Strings containing the input split by the delimiter.
 
         Raises:
           Error if an empty delimiter is specified.
         """
-        if not delimiter:
+        return self.split(str(delimiter), maxsplit)
+
+    fn split(
+        self, delimiter: Optional[String] = None, maxsplit: Int = Int.MAX
```
soraros, line 1355: Instead of having `delimiter` taking an `Optional`, we should probably use an overload.

bgreni: What would the practical advantage be of doing it that way you think?

soraros: Move the implementation in the two branches to overloads.
```moojo
fn split(self, maxsplit: Int = ...)
fn split(self, delimiter: String, maxsplit: Int = ...)
```

bgreni: Hmmm I wasn't convinced at first but on second thought I think you're right

stdlib/src/builtin/string.mojo
```diff
@@ -1333,36 +1333,93 @@ struct String(
             substr._strref_dangerous(), start=start
         )
 
-    fn split(self, delimiter: String) raises -> List[String]:
+    @always_inline
+    fn split(
+        self, delimiter: StringRef, maxsplit: Int = Int.MAX
+    ) raises -> List[String]:
```
laszlokindrat, line 1339: Why do we need this overload? I could understand having `StringRef.split` first, and maybe delegating to that?

stdlib/src/builtin/string.mojo
laszlokindrat: I would prefer fewer overloads in this case. Firstly, it wouldn't be necessary to define the default value for `maxsplit` twice. Secondly, a call to `some_string.split(bar)` could become ambiguous if `bar` is implicitly convertible to both `String` and `Int`. Given that we cannot make these arguments keyword-only (because the Python API dictates that these are positional-or-keyword passable), a single overload is the safest way.

laszlokindrat, line 1355: Could you please change `delimiter` to `sep` to follow python's argument names?

soraros: Yea, I think it's actually the other way around: it doesn't make much sense to have this function taking an optional value.

stdlib/src/builtin/string.mojo
bgreni: From my testing implicit conversion between `StringLiteral` and `Optional[String]` doesn't work so any invocations like
`String.split(",")` must become `String.split(String(","))` without this.

martinvuyk:
please close this PR since I'm working on the same issue in #2711 and I opened the issue itself #2686

stdlib/src/builtin/string.mojo
siitron: I think it makes sense to have it take an Optional since `.split(None)` is completely valid in Python.

Also, `maxsplit` should default to `-1`. (since `.split(maxsplit=-1)` is the same as doing `.split()`)

stdlib/test/builtin/test_string.mojo
```diff
@@ -573,6 +573,23 @@ fn test_split() raises:
...
+    assert_equal(
+        String("abs  gr").split(" ", maxsplit=1).__str__(), "['abs', ' gr']"
+    )
+    assert_equal(String("a b c").split(maxsplit=1).__str__(), "['a', 'b c']")
+    assert_equal(String("1<>2<>3").split("<>").__str__(), "['1', '2', '3']")
+    assert_equal(String("abc").split().__str__(), "['abc']")
+    assert_equal(
+        String("a b c d e").split(maxsplit=2).__str__(), "['a', 'b', 'c d e']"
+    )
+
```
siitron, line 592: Here are two more edge cases, I don't think they'll pass at the moment:
```suggestion
    assert_equal(String("a ").split(" ", 1).__str__(), "['a', '']")
    assert_equal(String("a ").split(None, 1).__str__(), "['a']")
```

martinvuyk: please top reviewing this PR since I've solved these issues and other in #2711 and I opened the issue itself #2686

PR #2714, opened 2024-05-17
[stdlib] Fix #2692 assert_raises should report calling location

### LINKED ISSUE #2692: [Feature Request] assert_raises should report calling location
### Review Mojo's priorities

### What is your request?

It would be useful if the error message produced when `assert_raises` fails, and therefore it raises itself https://github.com/modularml/mojo/blob/f5f5109541c31615a68a3c4b58bd1e75b59625f6/stdlib/src/testing/testing.mojo#L386

included the location of the caller who didn't raise.

Currently if you have a single test file with multiple assert_raises conditions, like this file:

https://github.com/modularml/mojo/blob/f5f5109541c31615a68a3c4b58bd1e75b59625f6/stdlib/test/collections/test_list.mojo#L404-L414

If one of the conditions fails, the error message is just:

```
Unhandled exception caught during execution: AssertionError: Didn't raise
```

with no indication of where the assertion is coming from

### Any other details?

_No response_

rd4com:
Hello, 

here is an implementation for #2692

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -367,3 +367,4 @@ what we publish.
 - Made several improvements to dictionary performance. Dicts with integer keys
   are most heavily affected, but large dicts and dicts with large values
   will also see large improvements.
+- [#2692](https://github.com/modularml/mojo/issues/2692) Fix assert_raises to include calling location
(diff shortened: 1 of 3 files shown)


JoeLoser (approved):

keith:
thanks for this!

PR #2717, opened 2024-05-17
[tools] Add markdownlint to the pre-commit

gabrieldemarmiesse:
Fix https://github.com/modularml/mojo/issues/2716

This pre-commit hook will also fix easy mistakes :) It also doesn't require any manual installation. It just works™

The CI will trigger an error if a markdown file does not conform to the rules of markdownlint

Example of pre-commit failure in the CI: https://github.com/modularml/mojo/actions/runs/9132279705/job/25113289498?pr=2663

### DIFF
--- a/stdlib/docs/development.md
+++ b/stdlib/docs/development.md
@@ -142,7 +142,7 @@ standard library, test it, and raise a PR.
 
 First, follow everything in the [prerequisites](#prerequisites).
 
-__IMPORTANT__ We'll be in the `mojo/stdlib` folder for this tutorial, check and
+**IMPORTANT** We'll be in the `mojo/stdlib` folder for this tutorial, check and
 make sure you're in that location if anything goes wrong:
 `cd [path-to-repo]/stdlib`
 
@@ -263,7 +263,7 @@ ls **/*.mojo | entr sh -c "./scripts/build-stdlib.sh && mojo main.mojo"
 Now, every time you save a Mojo file, it packages the standard library and
 runs `main.mojo`.
 
-__Note__: you should stop `entr` while doing commits, otherwise you could have
+**Note**: you should stop `entr` while doing commits, otherwise you could have
 some issues, this is because some pre-commit hooks use mojo scripts
 and will try to load the standard library while it is being compiled by `entr`.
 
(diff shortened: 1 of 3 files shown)


stdlib/docs/development.md
```diff
@@ -263,7 +263,7 @@ ls **/*.mojo | entr sh -c "./scripts/build-stdlib.sh && mojo main.mojo"
 Now, every time you save a Mojo file, it packages the standard library and
 runs `main.mojo`.
 
-**Note**: you should stop `entr` while doing commits, otherwise you could have
+__Note__: you should stop `entr` while doing commits, otherwise you could have
```
JoeLoser, line 266: **Question** Is this change because of differences in the `markdownlint-cli` version we use internally, or something else?

gabrieldemarmiesse: That may be possible I can try it out. Give me a sec

gabrieldemarmiesse: I tried with 0.34.0 and I still got 

```
stdlib/docs/development.md:266:1 MD050/strong-style Strong style should be consistent [Expected: underscore; Actual: asterisk]
```
Looking at the documentation it makes sense: https://github.com/DavidAnson/markdownlint/blob/main/doc/md050.md

By default markdownlint just wants all emphasis to be the same. Being `__something__` or `**something**`. I don't know what the tool installed on your machine is doing but I would say that the pre-commit is right here.

If you want we can add some configuration to enforce one style or the other, that would avoid any blurry line concerning this rule. What do you think?

gabrieldemarmiesse:
@JoeLoser done. We now always force `**something**` instead of `__something__`. There is very little usage of the second form. I hope it's also the case in the internal codebase. If not, you can run the pre-commit with `pre-commit run --all-files` to fix everything in one go.

stdlib/docs/development.md
JoeLoser: I see, that makes sense.  I don't feel strongly on the version, but I'd just like our internal and external to be consistent with the style they pick to avoid drift.  @modularml/docs do you have opinions here?

JoeLoser:
!sync

JoeLoser:
I just landed this internally, but our copybara isn't managing .pre-commit-config.yaml, ditto with the workflow files etc. We should probably have it manage the pre commit config yaml file for easy (thoughts @patrickdoc?). The net effect is the config and md file change landed, and once those propagate in the nightly tonight, then we can send a different smaller PR for the pre commit config file itself. Or is it easier to just merge this wholesale into the nightly branch now, @patrickdoc? There shouldn't be any conflicts on theory during the auto rebase, but didn't want to cause any issues.

PR #2718, opened 2024-05-17
[stdlib] Enable assertions in unit tests

gabrieldemarmiesse:
This is a PR to keep track of what is left to do. Do not review. I will make one PR per bug. When all bugs are fixed, only the diff about the config change will remain here. And then we can merge this PR.

I excluded all test files which cannot currently run with assertions enabled.


gabrieldemarmiesse:
@JoeLoser I think we can merge this PR before we fix everything to avoid adding new bugs in the meantime.

stdlib/test/lit.cfg.py
```diff
@@ -47,7 +47,8 @@
 # Substitute %mojo for just `mojo` itself
 # since we're not supporting `--sanitize` initially
 # to allow running the tests with LLVM sanitizers.
-config.substitutions.insert(0, ("%mojo", "mojo"))
+config.substitutions.insert(0, ("%mojo", "mojo -D MOJO_ENABLE_ASSERTIONS"))
```
JoeLoser, line 50: **Thought** I'd like us to go a different direction, if possible.  I would like the stdlib to build and all tests pass in both modes (`mojo` and `mojo -D MOJO_ENABLE_ASSERTIONS`).  I recommend tackling this from the CI matrix level and plumbing the variable `(-D MOJO_ENABLE_ASSERTIONS`) from the matrix into the `lit.cfg.py` here which feeds into the tests.  This is more generic/composes better with other knobs in the future (think sanitizers and the like).  Thoughts?

gabrieldemarmiesse: I don't mind doing this. I believe we'd need to document the rationale as it's not obvious why we need to run the tests twice. Is there a situation where a test wouldn't fail with assertions enabled, but would fail with assertions disabled?

JoeLoser: That is a valid point regarding running the tests twice and what that does to CI time.  I'm slightly paranoid about a case where our standard library only works with assertions enabled and not without (since without assertions is the default and common case for ends users/non-contributors).  One case is an assert that has side-affects (which is bad practice and you shouldn't do, of course, but I've seen these sort of things way too many times in C++ code).

I could be convinced to only run with assertions, but curious to get others input (CC: @laszlokindrat @rparolin).

gabrieldemarmiesse: I'm not too worried about CI time as it can be two jobs running at the same time. I'm just wondering if it's worth the complexity.

> We can always err on the side of caution and run it twice in the CI. We can make running with assertions the default so that the contributors don't run all the tests twice locally, I don't want it to affect the DX. Let's wait for other inputs.

gabrieldemarmiesse:
It seems a bug was introduced in `test_bit.mojo`. I can exlude this file too from the list of tests to run with assertions. Let me know what is decided.

stdlib/test/lit.cfg.py
JoeLoser: This is on my list to enable internally ASAP and then we can adopt this PR.  It's a bit complicated since use different lit configs etc internally.  I'll sort it all out though.

gabrieldemarmiesse:
In `test_bit.mojo` an assertion is triggered [here](https://github.com/modularml/mojo/blob/nightly/stdlib/src/builtin/simd.mojo#L1624) because the operation [`bit_floor`](https://github.com/modularml/mojo/blob/nightly/stdlib/src/bit/bit.mojo#L455) make use of [poison values](https://llvm.org/docs/LangRef.html#poison-values) (the offending expression is `1 << (bit_width(val) - 1)` which can shift with negative values). 

I'm reading through the description and it seems using poison values is fine (but dangerous) as long as they are overwritten afterwards, which is the case in `bit_floor`. 

Is the `debug_assert` still valid then? Should we disallow poison values? Asking maintainers here since it's not my area of expertise.

JoeLoser:
!sync

JoeLoser (approved):
I just landed this internally.  When reworking the commit message for some other internal changes I had to make to make this atomically land, I messed up something needed for Copybara, so the normal automation message didn't post here.  This will be available in the next nightly.  Thanks for pushing on this, @gabrieldemarmiesse! 🎉

gabrieldemarmiesse:
Noice! This will definitly allow us to be more confident in our future PRs :)

JoeLoser:
> In `test_bit.mojo` an assertion is triggered [here](https://github.com/modularml/mojo/blob/nightly/stdlib/src/builtin/simd.mojo#L1624) because the operation [`bit_floor`](https://github.com/modularml/mojo/blob/nightly/stdlib/src/bit/bit.mojo#L455) make use of [poison values](https://llvm.org/docs/LangRef.html#poison-values) (the offending expression is `1 << (bit_width(val) - 1)` which can shift with negative values).
> 
> I'm reading through the description and it seems using poison values is fine (but dangerous) as long as they are overwritten afterwards, which is the case in `bit_floor`.
> 
> Is the `debug_assert` still valid then? Should we disallow poison values? Asking maintainers here since it's not my area of expertise.

Normally I'd think that bit shifting negative values should be UB.  I'd vote for removing the `debug_assert`.  What do you think, @ConnorGray @laszlokindrat @rparolin?

PR #2719, opened 2024-05-17
[stdlib] Optional `start` and `length` arg for `String._strref_dangerous`

fknfilewalker:
Normally when one wants to create a truncated StringRef, a ptr to the String together with offsets need to be used. This is annoying and dangerous and could be done more safely during the `_strref_dangerous` call. This also adds bound guards


stdlib/src/builtin/string.mojo
```diff
@@ -1136,14 +1136,31 @@ struct String(
         elems.each[add_elt]()
         return result
 
-    fn _strref_dangerous(self) -> StringRef:
-        """
-        Returns an inner pointer to the string as a StringRef.
+    fn _strref_dangerous(self, start: Int = 0, length: Int = -1) -> StringRef:
```
laszlokindrat, line 1139: Please make these keyword-only to force callers to be explicit. Also, please use `Optional[Int]` for `length`:
```suggestion
    fn _strref_dangerous(self, *, start: Int = 0, length: Optional[Int] = None) -> StringRef:
```

laszlokindrat (commented):
Thanks for the patch! @JoeLoser Do you have thoughts on this?

JoeLoser:
> Thanks for the patch! @JoeLoser Do you have thoughts on this?

In general, use of `_strref_dangerous` and friends will go away in favor of some of the `StringSlice` (safe string slice) work going on internally, right?  I need to look closer at it before I can give a good review here.

laszlokindrat:
> > Thanks for the patch! @JoeLoser Do you have thoughts on this?
> 
> In general, use of `_strref_dangerous` and friends will go away in favor of some of the `StringSlice` (safe string slice) work going on internally, right? I need to look closer at it before I can give a good review here.

Yeah, although I think this could be useful in the meanwhile, and low risk since it's a hidden API anyway.

@fknfilewalker Could you please rebase after the next nightly, and change this patch to update the potential uses in `https://github.com/modularml/mojo/pull/2710` and maybe other places? That would give us a definitive picture of how much this actually helps.

stdlib/src/builtin/string.mojo
```diff
@@ -1211,14 +1211,27 @@ struct String(
...
+            length: Length of StringRef. If None, the length is the remaining length of the string. If
+                the length is greater than the remaining length of the string, the length is truncated.
+
+        Returns:
+            A StringRef of the inner string.
         """
-        return StringRef(self.unsafe_ptr(), len(self))
+        if start >= len(self):
+            return StringRef(self.unsafe_ptr() + len(self), 0)
+        var _length = min(length.or_else(len(self)), len(self) - start)
```
laszlokindrat, line 1233: Does this not work?
```suggestion
        var _length = min(length or len(self), len(self) - start)
```

stdlib/src/builtin/string.mojo
```diff
@@ -1211,14 +1211,27 @@ struct String(
...
+            start: Offset from the start of the string. If the start is greater than the
+                length of the string, an empty string is returned.
+            length: Length of StringRef. If None, the length is the remaining length of the string. If
+                the length is greater than the remaining length of the string, the length is truncated.
+
+        Returns:
+            A StringRef of the inner string.
         """
-        return StringRef(self.unsafe_ptr(), len(self))
+        if start >= len(self):
```
laszlokindrat, line 1231: If we have these safety checks in the semantics, it would be good to ensure that negative `start` and `length` values are also clamped/handled.

docs/changelog.md
```diff
@@ -367,6 +367,9 @@ what we publish.
 - Added `clear` method  to `Dict`.
   ([PR 2627](https://github.com/modularml/mojo/pull/2627) by [@artemiogr97](https://github.com/artemiogr97))
 
+- Optional `start` and `length` argument added for `String._strref_dangerous()`.
```
laszlokindrat, line 370: Since this is a private API, could you please remove it from the changelog?

laszlokindrat (approved):
Thank you! I have a few more requests, but LGTM otherwise!

stdlib/src/builtin/string.mojo
fknfilewalker: no, `min(length, len(_ref) - start)` also does not work. Seems like the automatic unpacking does not work.
I am getting
`candidate not viable: argument #0 cannot be converted from 'Optional[Int]' to 'Int'`

laszlokindrat: To use `min`, you definitely need to unpack it manually, and I think that's the intended behavior. I don't know if we want this with `or`; the desired behavior is not obvious if `length` has a value of `0`. Anyway, thanks for checking!

laszlokindrat:
Also, can you please update `String.{end,start}swith` (and any other places you're aware of) to use these?

fknfilewalker:
> Also, can you please update `String.{end,start}swith` (and any other places you're aware of) to use these?

For now I only checked string.mojo and there are not many places where it is really helpful and since you want to replace it with `span` anyway maybe we should just drop this?

laszlokindrat:
Is it really only one place where it's useful? I would have expected more. If you find one or two more places, I think we should land it, otherwise you're probably right, it's not pulling its weight.

laszlokindrat:
@fknfilewalker Did you find any other uses for this? If not, I think I will just close it for now, and we can revive later if it becomes relevant again.

ConnorGray:
@fknfilewalker Thanks for your work on this so far.

Regarding whether we want to move forward with this, I think at the moment this change seems a bit under-motivated. From looking at where this new functionality is used, I think every call site is an example of a place that we will eventually want to migrate to using the relatively new `StringSlice` type.

We really appreciate you taking the time to make this contribution, but I think we're going to hold off on accepting it.

There are other valuable and worthwhile directions for improvements to the string handling code that could be made and we'd be eager to accept—if you're interested in helping out we'd love to accept those kinds of changes 🙂.

For example, as mentioned, we could transition away from using `_strref_dangerous()` in the places you highlighted and instead start using `StringSlice`-based indexing.

fknfilewalker:
I support this decision :)

PR #2720, opened 2024-05-17
[stdlib] Fix incorrect number of variadic arguments in SIMD

gabrieldemarmiesse:
Fix part of https://github.com/modularml/mojo/issues/2687

I wonder if it's considered undefined behavior


stdlib/src/builtin/simd.mojo
```diff
@@ -326,8 +326,14 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
                 ]
             ](elems[0].value)
             return
-
-        debug_assert(size == num_elements, "mismatch in the number of elements")
+        # TODO: Make this a compile-time check when possible.
+        debug_assert(
```
JoeLoser, line 330: I think we want this to be part of an `else` branch rather than implicit IIUC. I think it's false-firing for the case where `num_elements == 1` where the docs say:

>  If there is just one input value, then it is splatted to all elements
        of the SIMD vector. Otherwise, the input values are assigned to the
        corresponding elements of the SIMD vector.

Can you look into this?  I agree for the tests adjusted with not just one element, those just look like incorrect tests.

stdlib/test/builtin/test_simd.mojo
```diff
@@ -154,22 +153,17 @@ def test_len():
...
-    var i5 = UI64(10, 20, 30, 40)
-    assert_equal(16, i5.__len__())
+
     var i6 = UI64(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15)
     assert_equal(16, i6.__len__())
 
     @parameter
     if not has_neon():
         alias BF16 = SIMD[DType.bfloat16, 2]
-        var f1 = BF16(0.0)
```
JoeLoser, line 165: As mentioned in the other comment, here and below, I think the code intent was for specifying one element to mean "splat this value `N` times" where `N` is the simd width.

stdlib/src/builtin/simd.mojo
gabrieldemarmiesse: You are right. I removed too many tests. I went a bit too fast on this bug. I fixed the tests, it should be okay now

JoeLoser (approved):

JoeLoser:
!sync

stdlib/src/builtin/simd.mojo
JoeLoser: No problem, thanks for fixing them up!

PR #2722, opened 2024-05-17
[stdlib] Apply Indexer trait to stdlib containers

bgreni:
The second half of #2384

I have intentionally omitted `static_tuple.mojo` to avoid conflicting with #2677

### DIFF
--- a/stdlib/test/builtin/test_list.mojo
+++ b/stdlib/test/builtin/test_list.mojo
@@ -25,6 +25,8 @@ fn test_variadic_list() raises:
         assert_equal(nums[0], 5)
         assert_equal(nums[1], 8)
         assert_equal(nums[2], 6)
+        assert_equal(nums[True], 8)
+        assert_equal(nums[Int32(0)], 5)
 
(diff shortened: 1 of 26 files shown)


stdlib/test/builtin/test_list.mojo
```diff
@@ -38,7 +38,14 @@ fn test_repr_list() raises:
     assert_equal(__type_of(empty).__repr__(empty), "[]")
 
 
+def test_indexing():
+    var s = slice(1, 10)
+    assert_equal(s[True], 2)
+    assert_equal(s[UInt64(0)], 1)
```
laszlokindrat, line 44: Here and in all the other tests: Could you please add one for just using an `IntLiteral`:
```mojo
    assert_equal(s[2], 3)
```

laszlokindrat (approved):
Awesome, thank you! Before I bring this is, could you please add a few tests for IntLiteral indexers, and also update the changelog (you could just modify your existing entry for `Indexer`).

PR #2723, opened 2024-05-17
[stdlib] Fix list test warnings

bgreni:
If you break any of the `List` tests you get flooded with warnings and it makes it annoying to find the message telling you what failed.


laszlokindrat (approved):
nice fix, thanks!

PR #2726, opened 2024-05-17
[stdlib] Fix wrong slice check in span

gabrieldemarmiesse:
Fix part of https://github.com/modularml/mojo/issues/2687 

both the start and the end of span can be `len(original_string)`, that's not an issue. In the case where the start is equal to the lenght of th string, it just gives an empty span. This is similar to the slicing `my_string[:0]` which also gives an empty span.


JoeLoser:
!sync

PR #2728, opened 2024-05-18
[stdlib] Add `__repr__` in `SIMD`

bgreni:
Implements `__repr__` for `SIMD`

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -427,6 +427,9 @@ what we publish.
 
 - Added a new `tempfile` module. Similarly to Python, this will contain
   utilities for creating and working with temporary files and directories.
+  
+- Added `SIMD.__repr__` to get the verbose string representation of `SIMD` types.
+([PR #2728](https://github.com/modularml/mojo/pull/2728) by [@bgreni](https://github.com/bgreni))
 
 ### 🦋 Changed
 
--- a/stdlib/src/builtin/dtype.mojo
+++ b/stdlib/src/builtin/dtype.mojo
@@ -641,6 +641,25 @@ fn _integral_type_of[type: DType]() -> DType:
     return type.invalid
 
 
+fn _scientific_notation_digits[type: DType]() -> StringLiteral:
+    """Get the number of digits for the scientific notation representation of a float.
+    """
+    constrained[type.is_floating_point(), "Expected float point type"]()
+
+    @parameter
+    if type == DType.bfloat16 or type == DType.float16:
+        return "4"
+
+    @parameter
+    if type == DType.float32 or type == DType.tensor_float32:
+        return "8"
+
+    @parameter
+    if type == DType.float64:
+        return "16"
+    return "0"
+
+
 # ===-------------------------------------------------------------------===#
 # _uint_type_of_width
 # ===-------------------------------------------------------------------===#
--- a/stdlib/src/builtin/io.mojo
+++ b/stdlib/src/builtin/io.mojo
@@ -179,7 +179,8 @@ fn _snprintf[
 
 @no_inline
 fn _snprintf_scalar[
-    type: DType
+    type: DType,
+    float_format: StringLiteral = "%.17g",
 ](buffer: UnsafePointer[UInt8], size: Int, x: Scalar[type]) -> Int:
     alias format = _get_dtype_printf_format[type]()
 
@@ -195,9 +196,9 @@ fn _snprintf_scalar[
         type == DType.float16 or type == DType.bfloat16 or type == DType.float32
     ):
         # We need to cast the value to float64 to print it.
-        return _float_repr(buffer, size, x.cast[DType.float64]())
+        return _float_repr[float_format](buffer, size, x.cast[DType.float64]())
     elif type == DType.float64:
-        return _float_repr(buffer, size, rebind[Float64](x))
+        return _float_repr[float_format](buffer, size, rebind[Float64](x))
     return 0
 
 
@@ -207,11 +208,13 @@ fn _snprintf_scalar[
 
 
 @no_inline
-fn _float_repr(buffer: UnsafePointer[UInt8], size: Int, x: Float64) -> Int:
+fn _float_repr[
+    format: StringLiteral = "%.17g"
+](buffer: UnsafePointer[UInt8], size: Int, x: Float64) -> Int:
     # Using `%.17g` with decimal check is equivalent to CPython's fallback path
     # when its more complex dtoa library (forked from
     # https://github.com/dtolnay/dtoa) is not available.
-    var n = _snprintf(buffer, size, "%.17g", x.value)
+    var n = _snprintf(buffer, size, format, x.value)
     # If the buffer isn't big enough to add anything, then just return.
     if n + 2 >= size:
         return n
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -40,9 +40,12 @@ from utils.numerics import (
 )
 from utils._visualizers import lldb_formatter_wrapping_type
 from utils import InlineArray
-
-from .dtype import _integral_type_of, _get_dtype_printf_format
-from .io import _snprintf_scalar, _printf, _print_fmt
+from .dtype import (
+    _integral_type_of,
+    _get_dtype_printf_format,
+    _scientific_notation_digits,
+)
+from .io import _snprintf_scalar, _snprintf, _printf, _print_fmt
 from .string import _calc_initial_buffer_size, _calc_format_buffer_size
 
 # ===------------------------------------------------------------------------===#
@@ -139,6 +142,7 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
     Sized,
     Stringable,
     Truncable,
+    Representable,
 ):
     """Represents a small vector that is backed by a hardware vector element.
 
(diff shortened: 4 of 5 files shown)


stdlib/src/builtin/simd.mojo
```diff
@@ -510,6 +511,24 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
 
         return String.format_sequence(self)
 
+    @always_inline
+    fn __repr__(self) -> String:
+        """Get the representation of the SIMD value eg. "SIMD[DType.int8, 2](1, 2)".
+        Returns:
+            The representation of the SIMD value
```
laszlokindrat, line 518: nit:
```suggestion
            The representation of the SIMD value.
```

stdlib/test/builtin/test_simd.mojo
```diff
@@ -90,6 +90,18 @@ def test_convert_simd_to_string():
     )
 
 
+def test_simd_repr():
+    assert_equal(
```
laszlokindrat, line 94: Can you please add tests for floating point types? And when you do, please ensure you cover special floating point values (nan, inf, -inf, -0.0).

stdlib/test/builtin/test_simd.mojo
```diff
@@ -90,6 +90,18 @@ def test_convert_simd_to_string():
...
+        SIMD[DType.int32, 4](1, 2, 3, 4).__repr__(),
```
laszlokindrat, line 95: Please modify this test to use negative values as well.

laszlokindrat (commented):
Thanks, this could be a good improvement! I wonder what we should do with floating point values. We looked at this recently with @ConnorGray for numpy, but I can't remember what we concluded. Since we don't have good utilities for controlling floating point printing precision globally (the way numpy does), it might be safest to print all useful digits (the number of which depends on the float type), and use scientific notation.

Either way, please add a changelog entry.

laszlokindrat:
Could you please rebase this after tomorrow's nightly? Also, I really think the implementation should be verbose here, and print all significant digits and use scientific notation.

bgreni:
> Could you please rebase this after tomorrow's nightly? Also, I really think the implementation should be verbose here, and print all significant digits and use scientific notation.

Sorry could you maybe give an example of the output you're expecting?

laszlokindrat:
Something like this:
```mojo
assert_equal(
    SIMD[DType.float32, 1](1232.4).__repr__(),
    "SIMD[DType.float32, 1](1.2324000e+03)",  #  8 digits for float32, 16 for float64, etc
)
```
Looking at this now, I'm not actually sure about the trailing zeros, but in repr, I would prefer scientific notation.

bgreni:
Ok I'll do that thanks!

bgreni:
@laszlokindrat I've got the scientific notation working, hopefully you don't find the required changes too invasive. For now I haven't worried about trailing zeroes until you've made a concrete decision on that front.

docs/changelog.md
```diff
@@ -406,6 +406,9 @@ what we publish.
 
 - The Mojo Language Server now supports renaming local variables.
 
+- Added `SIMD.__repr__` to get the representation of `SIMD` types.
```
laszlokindrat, line 409:
```suggestion
- Added `SIMD.__repr__` to get the verbose string representation of `SIMD` types.
```

stdlib/src/builtin/dtype.mojo
```diff
@@ -641,6 +641,24 @@ fn _integral_type_of[type: DType]() -> DType:
...
+        return "4"
+
+    @parameter
+    if type == DType.float32 or type == DType.tensor_float32:
+        return "8"
+
+    @parameter
+    if type == DType.float64:
+        return "16"
+    return "0"
```
laszlokindrat, line 659: Is it possible to use `constrained` here?
```mojo
constrained[type.is_floating_point(), ...]
```

stdlib/src/builtin/io.mojo
```diff
@@ -180,7 +180,12 @@ fn _snprintf[
 @no_inline
 fn _snprintf_scalar[
     type: DType
-](buffer: UnsafePointer[UInt8], size: Int, x: Scalar[type]) -> Int:
+](
+    buffer: UnsafePointer[UInt8],
+    size: Int,
+    x: Scalar[type],
+    float_format: StringLiteral = "%.17g",
```
laszlokindrat, line 187: Here and in `_float_repr` and `_format_scalar` below: Can you please make `float_format` a parameter? Soon `StringLiteral` won't be allowed at runtime, and this would prevent some churn.

stdlib/src/builtin/simd.mojo
```diff
@@ -576,13 +580,47 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
         return String.format_sequence(self)
 
+    @always_inline
+    fn __repr__(self) -> String:
+        """Get the representation of the SIMD value eg. "SIMD[DType.int8, 2](1, 2)".
+        Returns:
+            The representation of the SIMD value.
+
+        """
+        var values: String
```
laszlokindrat, line 590: Could this `values` be replaced with a `StringRef` or `StringSlice`? Or does the concatenation not work then?

stdlib/src/builtin/simd.mojo
```diff
@@ -576,13 +580,47 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
     fn format_to(self, inout writer: Formatter):
         """
         Formats this SIMD value to the provided formatter.
 
         Args:
             writer: The formatter to write to.
         """
+        self.format_to(writer, False)
+
+    fn format_to(self, inout writer: Formatter, use_scientific_notation: Bool):
```
laszlokindrat, line 616: Why not use a default value instead of an overload?
```suggestion
    fn format_to(self, inout writer: Formatter, use_scientific_notation: Bool = False):
```

laszlokindrat (commented):
Thanks for pushing this forward! The behavior with respect to trailing zeros is fine for now; we can iterate later if needed. Could you please rebase and ensure CI is happy?

stdlib/src/builtin/simd.mojo
bgreni: Wouldn't that be unsafe/unnecessary since assigning `values` is the final use of `output` and it's being moved anyway?

stdlib/src/builtin/simd.mojo
bgreni: Doing so results in a compiler error in `SIMD.__str__()` when it tries to call `String.format_sequence(self)` since I thats not the function signature it expects of a `Formattable` type (I think?)

`invalid call to 'format_sequence': callee with non-empty variadic pack argument expects 0 positional operands, but 1 was specified`

laszlokindrat: Right, thanks! Could you please just mention this in a comment or the docstring?

stdlib/src/builtin/simd.mojo
laszlokindrat: The `size > 1` branch will probably need to make a copy. `StringSlice` should be safe btw.

stdlib/src/builtin/dtype.mojo
bgreni: There doesn't seem to be a way to prevent line 658 in `simd.mojo` from evaluating in so it'll produce compilation errors when calling `__repr__` with integral and other types?

stdlib/src/builtin/dtype.mojo
```diff
@@ -641,6 +641,24 @@ fn _integral_type_of[type: DType]() -> DType:
     return type.invalid
 
 
+fn _scientific_notation_digits[type: DType]() -> StringLiteral:
+    """Get the number of digits for the scientific notation representation of a float
```
laszlokindrat, line 645: nit:
```suggestion
    """Get the number of digits for the scientific notation representation of a float.
```

stdlib/src/builtin/simd.mojo
```diff
@@ -576,13 +580,48 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+
+    # This overload is required to keep SIMD compliant with the Formattable trait,
+    # and the call to `String.format_sequence(self)` in SIMD.__str__ will fail to compile
+    fn format_to(self, inout writer: Formatter, use_scientific_notation: Bool):
+        """
+        Formats this SIMD value to the provided formatter.
+
+        Args:
+            writer: The formatter to write to.
+            use_scientific_notation: Whether floats should use scientific notation.
```
laszlokindrat, line 623: Please add a note that this is ignored for integers.

stdlib/src/builtin/dtype.mojo
laszlokindrat: Yeah, you probably have to guard the callsite with a parameter if.

stdlib/src/builtin/simd.mojo
```diff
@@ -576,13 +580,48 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+
+    @always_inline
     fn format_to(self, inout writer: Formatter):
         """
         Formats this SIMD value to the provided formatter.
 
         Args:
             writer: The formatter to write to.
         """
+        self.format_to(writer, False)
```
laszlokindrat, line 613: nit: to avoid ambiguity in case we add other flags:
```suggestion
        self.format_to(writer, use_scientific_notation=False)
```

stdlib/src/builtin/simd.mojo
```diff
@@ -576,13 +580,48 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+    fn __repr__(self) -> String:
+        """Get the representation of the SIMD value eg. "SIMD[DType.int8, 2](1, 2)".
+        Returns:
+            The representation of the SIMD value.
+
+        """
+
+        var output = String()
+        var writer = output._unsafe_to_formatter()
+        self.format_to(writer, True)
```
laszlokindrat, line 593: Same here
```suggestion
        self.format_to(writer, use_scientific_notation=True)
```

stdlib/src/builtin/dtype.mojo
bgreni: Oh right silly me I had to add `type.is_floating_point()` to the condition but that works now!

laszlokindrat (approved):
Awesome, thanks!

laszlokindrat:
!sync

PR #2733, opened 2024-05-18
[stdlib] Allow natural list.__repr__() syntax for List

msaelices:
No need for the `@staticmethod` and the `__type_of()` trick in `List.__repr__()`


JoeLoser (approved):
Great, thank you!

JoeLoser:
!sync

PR #2735, opened 2024-05-18
[stdlib] Apply `any()`/`all()` for `SIMD` cases

helehex:
Switch to using the builtin `any()`/`all()` functions instead of `SIMD.reduce_and()`/`SIMD.reduce_or()` where applicable. This simplifies the code *and* reads very nice.
I tested the mandelbrot notebook example to ensure it still works.


helehex:
oh, should i have ran the notebook once after clearing the outputs? i noticed the large -diff

JoeLoser (approved):
Nice cleanup!  Don't worry about the metatdata of the notebook.  We're not always great about clearing the cells prior to it being checked in.  That's one of the problems of checking in ipynb into version control if you're not careful.

JoeLoser:
!sync

JoeLoser:
!sync

PR #2736, opened 2024-05-18
[stdlib] implement `bit_ceil` and `bit_floor`

### LINKED ISSUE #2682: [stdlib] Implement missing functions in the `bit` module
For example: `bit_ceil`, `bit_floor`, `has_single_bit`, and others. The immediate goal is to achieve (or get pretty close to) parity with [Cpp's `<bit>`](https://en.cppreference.com/w/cpp/header/bit). We shouldn't blindly follow the naming, and we can deviate from or extend some of these APIs slightly if we have a good reason to. Many of these functions might be simple wrappers around intrinsics.

See comment: https://github.com/modularml/mojo/commit/3f9fa057a9587f26e217344e663cc77513c0002a#diff-20568bce7de07632ece5bf3e933c7e2eec19706a949dcae48ddd4299309ddd10R249

jiex-liu:
fix #2682

### DIFF
--- a/stdlib/src/bit/bit.mojo
+++ b/stdlib/src/bit/bit.mojo
@@ -307,6 +307,7 @@ fn bit_width[
 # ===----------------------------------------------------------------------===#
 # bit_ceil
 # ===----------------------------------------------------------------------===#
+# reference: https://en.cppreference.com/w/cpp/numeric/bit_ceil
 
 
 @always_inline("nodebug")
@@ -354,22 +355,19 @@ fn bit_ceil[
     """
     constrained[type.is_integral(), "must be integral"]()
 
-    alias ones = SIMD[type, simd_width].splat(1)
-
-    var less_then_one = (val <= ones).select(ones, val)
-
-    return (val > ones).select(1 << bit_width(val - ones), less_then_one)
+    return (val <= 1).select(1, 1 << bit_width(val - 1))
 
 
 # ===----------------------------------------------------------------------===#
 # bit_floor
 # ===----------------------------------------------------------------------===#
+# reference: https://en.cppreference.com/w/cpp/numeric/bit_floor
 
 
 @always_inline("nodebug")
 fn bit_floor(val: Int) -> Int:
     """Computes the largest power of 2 that is less than or equal to the input
-    value.
+    value. Any integral value less than or equal to 0 will be floored to 0.
 
     Args:
         val: The input value.
@@ -388,7 +386,8 @@ fn bit_floor[
     type: DType, simd_width: Int
 ](val: SIMD[type, simd_width]) -> SIMD[type, simd_width]:
     """Computes the largest power of 2 that is less than or equal to the input
-    value for each element of a SIMD vector.
+    value for each element of a SIMD vector. Any integral value less than or
+    equal to 0 will be floored to 0.
 
     Parameters:
         type: `dtype` used for the computation.
@@ -407,11 +406,7 @@ fn bit_floor[
     """
     constrained[type.is_integral(), "must be integral and unsigned"]()
 
-    alias zeros = SIMD[type, simd_width].splat(0)
-
-    var less_then_zero = (val <= zeros).select(zeros, val)
-
-    return (val > zeros).select(1 << (bit_width(val) - 1), less_then_zero)
+    return (val <= 0).select(0, 1 << (bit_width(val) - 1))
 
 
 # ===----------------------------------------------------------------------===#
--- a/stdlib/src/bit/bit.mojo
+++ b/stdlib/src/bit/bit.mojo
@@ -355,7 +355,9 @@ fn bit_ceil[
     """
     constrained[type.is_integral(), "must be integral"]()
 
-    return (val <= 1).select(1, 1 << bit_width(val - 1))
+    alias ones = SIMD[type, simd_width].splat(1)
+
+    return (val > 1).select(1 << bit_width(val - ones), ones)
 
 
 # ===----------------------------------------------------------------------===#
@@ -406,7 +408,9 @@ fn bit_floor[
     """
     constrained[type.is_integral(), "must be integral and unsigned"]()
 
-    return (val <= 0).select(0, 1 << (bit_width(val) - 1))
+    alias zeros = SIMD[type, simd_width].splat(0)
+
+    return (val > 0).select(1 << (bit_width(val) - 1), zeros)
 
 
 # ===----------------------------------------------------------------------===#
--- a/stdlib/src/bit/bit.mojo
+++ b/stdlib/src/bit/bit.mojo
@@ -254,10 +254,25 @@ fn bit_not[
 
 
 @always_inline
+fn bit_width(val: Int) -> Int:
+    """Computes the minimum number of bits required to represent the integer.
+
+    Args:
+        val: The input value.
+
+    Returns:
+        The number of bits required to represent the integer.
+    """
+    alias bitwidth = bitwidthof[Int]()
+
+    return bitwidth - countl_zero(~val if val < 0 else val)
+
+
+@always_inline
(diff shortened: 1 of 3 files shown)


LJ-9801:
Implementation of `bit_ceil` is based on [c++ bit_ceil](https://en.cppreference.com/w/cpp/numeric/bit_ceil)

soraros:
Should we include the reference in doc string or comment in the future?

stdlib/src/bit/bit.mojo
```diff
@@ -289,6 +306,64 @@ fn bit_width[
...
+        that is greater than or equal to the integer at position `i` of the input
+        value.
+    """
+    constrained[type.is_integral(), "must be integral"]()
+
+    alias ones = SIMD[type, simd_width].splat(1)
+    alias zeros = SIMD[type, simd_width].splat(0)
+
+    var ret = (val <= ones).select(ones, val)
+    ret = (val & (val - ones) == zeros).select(val, ret)
```
soraros, line 362: Can we skip this line since there is no early return?

LJ-9801:
@soraros ops yeah you are right. I'll include it in the docstring for now. @laszlokindrat which one do you think is more suitable?

stdlib/src/bit/bit.mojo
LJ-9801: yep we could skip this. Thanks for pointing this out!

stdlib/src/bit/bit.mojo
```diff
@@ -253,6 +253,23 @@ fn bit_not[
...
+fn bit_width(val: Int) -> Int:
+    """Computes the minimum number of digits required to represent the integer.
+
+    Args:
+        val: The input value.
+
+    Returns:
+        The number of digits required to represent the integer.
+    """
+    var bitwidth = bitwidthof[Int]()
```
laszlokindrat, line 266: Prefer doing this at compile time.
```suggestion
    alias bitwidth = bitwidthof[Int]()
```

stdlib/src/bit/bit.mojo
```diff
@@ -253,6 +253,23 @@ fn bit_not[
...
+    if val < 0:
+        return bitwidth - countl_zero(~val)
+    else:
+        return bitwidth - countl_zero(val)
```
laszlokindrat, line 270: style nit: slightly more pythonic:
```suggestion
    return bitwidth - countl_zero(~val if val < 0 else val)
```

stdlib/test/bit/test_bit.mojo
```diff
@@ -12,11 +12,57 @@
...
+
+
+def test_bit_ceil_simd():
+    alias simd_width = 4
+    alias type = DType.int8
+
+    alias var1 = SIMD[type, simd_width](-2, -1, 3, 4)
+    assert_equal(
+        bit_ceil[type, simd_width](var1), SIMD[type, simd_width](1, 1, 4, 4)
+    )
```
laszlokindrat, line 35: here and below: please rely on parameter inference instead of passing explicitly
```suggestion
    assert_equal(
        bit_ceil(var1), SIMD[type, simd_width](1, 1, 4, 4)
    )
```

stdlib/src/bit/bit.mojo
```diff
@@ -253,6 +253,23 @@ fn bit_not[
 # ===----------------------------------------------------------------------===#
 
 
+@always_inline
+fn bit_width(val: Int) -> Int:
```
laszlokindrat, line 257: Can you please add tests for this? Also, while you're at it, could you also add tests for the SIMD overload?

stdlib/src/bit/bit.mojo
```diff
@@ -253,6 +253,23 @@ fn bit_not[
...
+    """Computes the minimum number of digits required to represent the integer.
```
laszlokindrat, line 258: You can also say "binary digits", but either way, we should avoid ambiguity.
```suggestion
    """Computes the minimum number of bits required to represent the integer.
```

laszlokindrat (commented):
Nice, thanks! I have a few requests for change, nothing major. This will be an excellent improvement for the `bit` module!

laszlokindrat:
> > Should we include the reference in doc string or comment in the future?
> 
> @soraros ops yeah you are right. I'll include it in the docstring for now. @laszlokindrat which one do you think is more suitable?

I would prefer to not reference external documentation. Can you please just ensure that the docstrings are (more or less) self explanatory?

soraros:
@laszlokindrat My idea was that performant numerical code can contain a large amount of black magic, thus not very self-explanatory. If we are indeed using an existing implementation (from a paper, libc++, etc) and include a reference to them, it would be less surprising to future maintainers when they see the said magic. If you think a link is not fitting for the doctoring, maybe we could simply leave a comment.

LJ-9801:
@laszlokindrat I have fixed all the requests. In terms of the reference, I do agree with @soraros in that numerical code does sometime contains "black magic". While these two functions seems pretty ok to understand, there might be functions that has wizardry implementation in the future and adding a reference can definitely help future maintainer. I have taken out the c++ reference. If there is a consent, I can put it back in.

stdlib/src/bit/bit.mojo
```diff
@@ -289,6 +304,116 @@ fn bit_width[
...
+        that is greater than or equal to the integer at position `i` of the input
+        value.
+    """
+    constrained[type.is_integral(), "must be integral"]()
+
+    alias ones = SIMD[type, simd_width].splat(1)
+
+    var less_then_one = (val <= ones).select(ones, val)
+
+    return (val > ones).select(1 << bit_width(val - ones), less_then_one)
```
soraros, line 361: Could we simply write this?
```suggestion
	return (val <= 1).select(1, 1 << bit_width(val - 1))
```

laszlokindrat:
> If you think a link is not fitting for the doctoring, maybe we could simply leave a comment.

Yeah, a link within a comment in the implementation is fine, since that's a message to developers working on the stdlib. I just didn't want the generated docs to reference external links (and also, docstrings should explain _what_ the function does, not _how_). @LJ-9801 could you please do this?

laszlokindrat (approved):
LGTM, thanks!

JoeLoser:
!sync

stdlib/src/bit/bit.mojo
LJ-9801: just fixed it, thank!

soraros: Why do you need two `select` at all?

stdlib/src/bit/bit.mojo
```diff
@@ -409,9 +411,9 @@ fn bit_floor[
 
     alias zeros = SIMD[type, simd_width].splat(0)
 
-    var less_then_zero = (val <= zeros).select(zeros, val)
+    var less_then_zero = (val <= 0).select(zeros, val)
 
-    return (val > zeros).select(1 << (bit_width(val) - 1), less_then_zero)
+    return (val > 1).select(1 << (bit_width(val) - 1), less_then_zero)
```
soraros, line 416: `val > 0`?

stdlib/src/bit/bit.mojo
LJ-9801: ohhh my bad i misread your suggestions.

LJ-9801:
@laszlokindrat @JoeLoser mojo crash on `test_reversed.mojo`? all the tests passed on my M1 mac tho.

stdlib/src/bit/bit.mojo
LJ-9801: the added line seems to cause a crash in the reverse test sometime, I still kept the `ones` and did 
```mojo
(val > 1).select(1 << bit_width(val - ones), ones)
```
to avoid the crashing.

laszlokindrat:
!sync

PR #2740, opened 2024-05-18
[stdlib] Replacing `StringRef`'s local `_memcmp` with `memory.memcmp` for string comparisons.

siitron:
Continuation of #2620.
A local `_memcmp`-method has previously been used when comparing strings in order to avoid indirect recursions. Unless this issue is still present then we may use `memory.memcmp` instead which would be preferred since the local `_memcmp` isn't vectorized.


laszlokindrat:
!sync

laszlokindrat:
I checked, this is breaking some internal stuff due to the recursion we discussed in https://github.com/modularml/mojo/pull/2620. Not sure how to fix it, but I think it has to do with the error printing of the constraint in `memcmp`. If you want you can try to refactor that to have a (potentially hidden) version that is unsafe and not constrained explicitly.

PR #2741, opened 2024-05-18
[Docs] correct output, fix indent in Heterogeneous variadic arguments

kengz:
The current [Heterogeneous variadic arguments](https://docs.modular.com/mojo/manual/functions#heterogeneous-variadic-arguments) (see screenshot) - last example has wrong output (`Bob0` should be `Bob`), and indentation of the example code is also off (had to edit after copying into Mojo playground)

Screenshot 2024-05-12 at 9 31 51 AM


KenJones-Modular (approved):

KenJones-Modular:
!sync

PR #2743, opened 2024-05-18
[stdlib] Temporary directory

artemiogr97:
Implementation of `TemporaryDirectory`, a wrapper around `mkdtemp` that can be used as a context nanager.

### DIFF
--- a/stdlib/src/tempfile/tempfile.mojo
+++ b/stdlib/src/tempfile/tempfile.mojo
@@ -144,11 +144,7 @@ fn mkdtemp(
     Raises:
         If the directory can not be created.
     """
-    var final_dir: Path
-    if not dir:
-        final_dir = Path(_get_default_tempdir())
-    else:
-        final_dir = Path(dir.value()[])
+    var final_dir = Path(dir.value()[]) if dir else Path(_get_default_tempdir())
 
     for _ in range(TMP_MAX):
         var dir_name = final_dir / (prefix + _get_random_name() + suffix)
@@ -172,33 +168,35 @@ fn _rmtree(path: String, ignore_errors: Bool = False) raises:
     If the path is a symbolic link, an error is raised.
     If ignore_errors is True, errors resulting from failed removals will be ignored.
     Absolute and relative paths are allowed, relative paths are resolved from cwd.
+
     Args:
       path: The path to the directory.
       ignore_errors: Whether to ignore errors.
     """
     if os.path.islink(path):
-        raise Error("`path`can not be a symbolic link")
+        raise Error("`path`can not be a symbolic link: " + path)
 
     for file_or_dir in os.listdir(path):
-        if os.path.isfile(path + "/" + file_or_dir[]):
+        var curr_path = os.path.join(path, file_or_dir[])
+        if os.path.isfile(curr_path):
             try:
-                os.remove(path + "/" + file_or_dir[])
+                os.remove(curr_path)
             except e:
                 if not ignore_errors:
-                    raise Error(e)
+                    raise e
             continue
-        if os.path.isdir(path + "/" + file_or_dir[]):
+        if os.path.isdir(curr_path):
             try:
-                _rmtree(path + "/" + file_or_dir[], ignore_errors)
+                _rmtree(curr_path, ignore_errors)
             except e:
                 if ignore_errors:
                     continue
-                raise Error(e)
+                raise e
     try:
         os.rmdir(path)
     except e:
         if not ignore_errors:
-            raise Error(e)
+            raise e
 
 
 struct TemporaryDirectory:
@@ -225,13 +223,8 @@ struct TemporaryDirectory:
             ignore_cleanup_errors: Whether to ignore cleanup errors.
         """
         self._ignore_cleanup_errors = ignore_cleanup_errors
-        var final_dir: Path
-        if not dir:
-            final_dir = Path(_get_default_tempdir())
-        else:
-            final_dir = Path(dir.value()[])
 
-        self.name = mkdtemp(suffix, prefix, final_dir.__fspath__())
+        self.name = mkdtemp(suffix, prefix, dir)
 
     fn __enter__(self) -> String:
         return self.name
--- a/stdlib/src/tempfile/tempfile.mojo
+++ b/stdlib/src/tempfile/tempfile.mojo
@@ -164,3 +164,84 @@ fn mkdtemp(
         except:
             continue
     raise Error("Failed to create temporary file")
+
+
+# TODO use shutil.rmtree (or equivalent) when it exists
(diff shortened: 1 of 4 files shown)


artemiogr97:
Part of the implementation for #2018

laszlokindrat:
https://github.com/modularml/mojo/pull/2742 has landed internally and will be included in the next nightly. Could you please rebase after that? Also, could you please update the PR description with a summary of the contents of this patch?

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -164,3 +164,84 @@ fn mkdtemp(
...
+    """Removes the specified directory and all its contents.
+    If the path is a symbolic link, an error is raised.
+    If ignore_errors is True, errors resulting from failed removals will be ignored.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
+    Args:
+      path: The path to the directory.
+      ignore_errors: Whether to ignore errors.
+    """
+    if os.path.islink(path):
+        raise Error("`path`can not be a symbolic link")
```
JoeLoser, line 180: **Suggestion** Include the path in this error message.

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -164,3 +164,84 @@ fn mkdtemp(
...
+
+    for file_or_dir in os.listdir(path):
+        if os.path.isfile(path + "/" + file_or_dir[]):
```
JoeLoser, line 183: **Suggestion** Can we use a portable separator rather than `/` everywhere?

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -164,3 +164,84 @@ fn mkdtemp(
...
             continue
     raise Error("Failed to create temporary file")
+
+
+# TODO use shutil.rmtree (or equivalent) when it exists
+fn _rmtree(path: String, ignore_errors: Bool = False) raises:
+    """Removes the specified directory and all its contents.
+    If the path is a symbolic link, an error is raised.
+    If ignore_errors is True, errors resulting from failed removals will be ignored.
+    Absolute and relative paths are allowed, relative paths are resolved from cwd.
```
JoeLoser, line 174: **Suggestion** I think a newline is needed here to please the doc string checking.

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -164,3 +164,84 @@ fn mkdtemp(
...
+
+        Args:
+            suffix: Suffix to use for the directory name.
+            prefix: Prefix to use for the directory name.
+            dir: Directory in which the directory will be created.
+            ignore_cleanup_errors: Whether to ignore cleanup errors.
+        """
+        self._ignore_cleanup_errors = ignore_cleanup_errors
+        var final_dir: Path
+        if not dir:
```
JoeLoser, line 229: **Suggestion** Flip this so `if dir` is the first part?  You'll save the `not`.

JoeLoser:
> #2742 has landed internally and will be included in the next nightly. Could you please rebase after that? Also, could you please update the PR description with a summary of the contents of this patch?

+1 to rebasing and adding a nice PR description!

stdlib/src/tempfile/tempfile.mojo
laszlokindrat: Or better yet, use a ternary if possible
```suggestion
        var final_dir = Path(dir.value()[] if dir _get_default_tempdir())
        # or
        var final_dir = Path(dir.value()[]) if dir Path(_get_default_tempdir())
```

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -164,3 +164,84 @@ fn mkdtemp(
...
+
+    var name: String
+    """The name of the temporary directory."""
+    var _ignore_cleanup_errors: Bool
+    """Whether to ignore cleanup errors."""
+
+    fn __init__(
+        inout self,
+        suffix: String = "",
+        prefix: String = "tmp",
```
laszlokindrat, line 215: In Python these can also be `None`. Can we also use `Optional[String] = None` for these?

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -164,3 +164,84 @@ fn mkdtemp(
         except:
             continue
     raise Error("Failed to create temporary file")
+
+
+# TODO use shutil.rmtree (or equivalent) when it exists
+fn _rmtree(path: String, ignore_errors: Bool = False) raises:
```
laszlokindrat, line 170: Can we use `PathLike` for the `path` argument?

ksandvik:
Yes, we could avoid Strings for path definitions, but rather PathLike traits, this for cross-platform support.

stdlib/src/tempfile/tempfile.mojo
artemiogr97: `listdir` returns a list of strings and `os.path.join` only takes `String` for now, so found it easier to use a `String` for now

stdlib/src/tempfile/tempfile.mojo
artemiogr97: `Optional[String]` does not play along nicely with `StringLiteral`, that's why I used default values for these, it should be possible to allow to use `None` but I think we will need 4 overloads to cover all the cases:
```
fn __init__(inout self, suffix: String = "", prefix: String = "tmp", ...)
fn __init__(inout self, suffix: Optional[String] = None, prefix: String = "tmp", ...)
fn __init__(inout self, suffix: String = "", prefix: Optional[String] = None, ...)
fn __init__(inout self, suffix: Optional[String] = None, prefix: Optional[String] = None, ...)
```

laszlokindrat: Yeah, this is a known issue, but workaround with the overloads can be simpler:
```mojo
fn __init__(inout self, suffix: String, prefix: String = "tmp", ...)
fn __init__(inout self, suffix: NoneType = None, prefix: String = "tmp", ...)
fn __init__(inout self, suffix: String = "", prefix: NoneType = None, ...)
fn __init__(inout self, suffix: NoneType = None, prefix: NoneType = None, ...)
```

laszlokindrat:
!sync

stdlib/src/tempfile/tempfile.mojo
laszlokindrat: Okay, thanks! Would it be possible to fix this in a followup?

stdlib/src/tempfile/tempfile.mojo
laszlokindrat: Could you please address this in a followup?

laszlokindrat (approved):
Awesome, thank you for working on this.

PR #2745, opened 2024-05-18
[stdlib] Fix out of bounds access in `List.index()`

gabrieldemarmiesse:

There were multiple bugs related to clipping there.

Long story short, the behavior of `list.index()` in python is this one: given a `start` and `end`, python will look for the element in `my_list[start:end]` and report the result, (`start` is added to the result to give the index with respect to the original list).

You can take a look at the description of the `index` method here: https://docs.python.org/3/tutorial/datastructures.html#more-on-lists

Since there is slicing semantics applied to `start` and `end`, we should do multiple things:
1) default to the start and the end of the list
2) normalize negative values by doing `+ len(my_list)`
3) clip both start and end between `0` and `len(my_list)`

The last step wasn't done correctly. Especially for the `stop` argument where the clipping was applied only when negative values were found (this caused the out of bounds bug in the tests).
Effectively 

```mojo
return end if end > 0 else min(end + size, size)
```
 is equivalent to 
```mojo
return end if end > 0 else end + size
```
since the min is applied only when `end <= 0`. So `end + size <= size`

This test can cause some flakyness in our CI: `test_list_a.index(10, start=5, stop=50)` for a list of size 6.
The stop was positive, so it was never clipped, thus too many values (out of bounds) were tried, causing some to match, sometimes.

Another bug was that because of this condition, `end = 0` means "check until the end of the list" while in python, with the slicing semantics, `end = 0` means "do nothing" (empty slice).

### TL;DR
Multiple clipping bugs. Slicing semantics applied to `start` and `stop` now like in Python. No more flakyness in our CI. No more out of bounds access. Corresponding tests added.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -508,18 +508,24 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         Raises:
             ValueError: If the value is not found in the list.
         """
-        var size = self[].size
-        var normalized_start = max(size + start, 0) if start < 0 else start
+        var start_normalized = start
 
-        @parameter
-        fn normalized_stop() -> Int:
-            if stop is None:
-                return size
-            else:
-                var end = stop.value()[]
-                return end if end > 0 else min(end + size, size)
+        var stop_normalized: Int
+        if stop is None:
+            # Default end
+            stop_normalized = len(self[])
+        else:
+            stop_normalized = stop.value()[]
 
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
Nice!  I really appreciate the detailed commit message and explanation of the problem(s) at hand.  It made this review easy.  Thank you for that and fixing the bugs!

JoeLoser:
!sync

PR #2747, opened 2024-05-18
[stdlib] Fix invalid code point check

gabrieldemarmiesse:

See https://stackoverflow.com/questions/27415935/does-unicode-have-a-defined-maximum-number-of-code-points for the correct range. You can also use Python's `chr` function to make sure the range is correct


JoeLoser (approved):

JoeLoser:
!sync

PR #2749, opened 2024-05-19
[stdlib] Add Result type

### LINKED ISSUE #2748: [Feature Request] [stdlib] [proposal] Add `ErrorReg`, `Result`, and `ResultReg` types
### Review Mojo's priorities

### What is your request?

Result is basically a copy paste of Optional just with an error attribute.

"""Defines Result, a type modeling a value which may or may not be present.
With a `Result.err` field which gives the error returned by the function.

Result values can be thought of as a type-safe nullable pattern.
Your value can take on a value or `None`, and you need to check
and explicitly extract the value to get it out.
```mojo
from collections import Result
var a = Result(1)
var b = Result[Int](Error())
if a:
    print(a.value()[])  # prints 1
if b:  # bool(b) is False, so no print
    print(b.value()[])
var c = a.or_else(2)
var d = b.or_else(2)
print(c)  # prints 1
print(d)  # prints 2
```

And if more information about the returned Error is wanted it is available.
```mojo
from collections import Result
var a = Result(1)
var b = Result[Int](Error("something went wrong"))
var c = Result[Int](Error("error 1"))
var d = Result[Int](Error("error 2"))
if a:
    print(a.err)  # prints ""
if not b:
    print(b.err) # prints "something went wrong"

if c.err:
    print("c had an error")

# TODO: pattern matching
if d.err == "error 1":
    print("d had error 1")
elif d.err == "error 2":
    print("d had error 2")
```
.
"""

```mojo
from utils import Variant


@value
struct _NoneType(CollectionElement):
    pass


# Result


@value
struct Result[T: CollectionElement](CollectionElement, Boolable):
    """A type modeling a value which may or may not be present.

    Result values can be thought of as a type-safe nullable pattern.
    Your value can take on a value or `None`, and you need to check
    and explicitly extract the value to get it out.

    Currently T is required to be a `CollectionElement` so we can implement
    copy/move for Result and allow it to be used in collections itself.

    ```mojo
    from collections import Result
    var a = Result(1)
    var b = Result[Int](None)
    if a:
        print(a.value()[])  # prints 1
    if b:  # bool(b) is False, so no print
        print(b.value()[])
    var c = a.or_else(2)
    var d = b.or_else(2)
    print(c)  # prints 1
    print(d)  # prints 2
    ```

    And if more information about the returned Error is wanted it is available.
    ```mojo
    from collections import Result
    var a = Result(1)
    var b = Result[Int](None, Error("something went wrong"))
    var c = Result[Int](None, Error("error 1"))
    var d = Result[Int](None, Error("error 2"))
    if a:
        print(a.err)  # prints ""
    if not b:
        print(b.err) # prints "something went wrong"

    if c.err:
        print("c had an error")

    # TODO: pattern matching
    if d.err == "error 1":
        print("d had error 1")
    elif d.err == "error 2":
        print("d had error 2")
    ```

    Parameters:
        T: The type of value stored in the `Result`.
    """

    # _NoneType comes first so its index is 0.
    # This means that Results that are 0-initialized will be None.
    var _value: Self._type
    var err: Error
    """The Error inside the `Result`."""

    fn __init__(inout self):
        """Construct an empty `Result`."""
        self._value = Self._type(_NoneType())
        self.err = Error()

    fn __init__(inout self, owned value: T):
        """Construct a `Result` containing a value.

        Args:
            value: The value to store in the `Result`.
        """
        self._value = Self._type(value^)
        self.err = Error()

    fn __init__(inout self, value: Error):
        """Construct an empty `Result`.

        Args:
            value: Must be an `Error`.
        """
        self._value = Self._type(_NoneType())
        self.err = value

    @always_inline
    fn value(
        self: Reference[Self, _, _]
    ) -> Reference[T, self.is_mutable, self.lifetime]:
        """Retrieve a reference to the value of the `Result`.

        This check to see if the `Result` contains a value.
        If you call this without first verifying the `Result` with __bool__()
        eg. by `if my_result:` or without otherwise knowing that it contains a
        value (for instance with `or_else`), the program will abort

        Returns:
            A reference to the contained data of the `Result` as a Reference[T].
        """
        if not self[].__bool__():
            abort(".value() on empty `Result`")

        return self[].unsafe_value()

    @always_inline
    fn unsafe_value(
        self: Reference[Self, _, _]
    ) -> Reference[T, self.is_mutable, self.lifetime]:
        """Unsafely retrieve a reference to the value of the `Result`.

        This doesn't check to see if the `Result` contains a value.
        If you call this without first verifying the `Result` with __bool__()
        eg. by `if my_result:` or without otherwise knowing that it contains a
        value (for instance with `or_else`), you'll get garbage unsafe data out.

        Returns:
            A reference to the contained data of the `Result` as a Reference[T].
        """
        debug_assert(self[].__bool__(), ".value() on empty Result")
        return self[]._value[T]

    @always_inline
    fn _value_copy(self) -> T:
        """Unsafely retrieve the value out of the `Result`.

        Note: only used for Results when used in a parameter context
        due to compiler bugs.  In general, prefer using the public `Result.value()`
        function that returns a `Reference[T]`.
        """

        debug_assert(self.__bool__(), ".value() on empty Result")
        return self._value[T]

    fn take(inout self) -> T:
        """Move the value out of the `Result`.

        The caller takes ownership over the new value, which is moved
        out of the `Result`, and the `Result` is left in an empty state.

        This check to see if the `Result` contains a value.
        If you call this without first verifying the `Result` with __bool__()
        eg. by `if my_result:` or without otherwise knowing that it contains a
        value (for instance with `or_else`), you'll get garbage unsafe data out.

        Returns:
            The contained data of the `Result` as an owned T value.
        """
        if not self.__bool__():
            abort(".take() on empty `Result`")
        return self.unsafe_take()

    fn unsafe_take(inout self) -> T:
        """Unsafely move the value out of the `Result`.

        The caller takes ownership over the new value, which is moved
        out of the `Result`, and the `Result` is left in an empty state.

        This check to see if the `Result` contains a value.
        If you call this without first verifying the `Result` with __bool__()
        eg. by `if my_option:` or without otherwise knowing that it contains a
        value (for instance with `or_else`), the program will abort!

        Returns:
            The contained data of the option as an owned T value.
        """
        debug_assert(self.__bool__(), ".unsafe_take() on empty Result")
        return self._value.unsafe_replace[_NoneType, T](_NoneType())

    fn or_else(self, default: T) -> T:
        """Return the underlying value contained in the `Result` or a default
        value if the `Result`'s underlying value is not present.

        Args:
            default: The new value to use if no value was present.

        Returns:
            The underlying value contained in the Result or a default value.
        """
        if self.__bool__():
            return self._value[T]
        return default

    fn __is__(self, other: NoneType) -> Bool:
        """Return `True` if the `Result` has no value.

        It allows you to use the following syntax: `if my_result is None:`

        Args:
            other: The value to compare to (None).

        Returns:
            True if the `Result` has no value and False otherwise.
        """
        return not self.__bool__()

    fn __isnot__(self, other: NoneType) -> Bool:
        """Return `True` if the `Result` has a value.

        It allows you to use the following syntax: `if my_result is not None:`.

        Args:
            other: The value to compare to (None).

        Returns:
            True if the `Result` has a value and False otherwise.
        """
        return self.__bool__()

    fn __bool__(self) -> Bool:
        """Return true if the `Result` has a value.

        Returns:
            True if the `Result` has a value and False otherwise.
        """
        return not self._value.isa[_NoneType]()

    fn __invert__(self) -> Bool:
        """Return False if the `Result` has a value.

        Returns:
            False if the `Result` has a value and True otherwise.
        """
        return not self

# ResultReg

@register_passable("trivial")
struct ResultReg[T: AnyRegType](Boolable):
    """A register-passable `ResultReg` type.

    This struct `ResultReg` contains a value. It only works with trivial register
    passable types at the moment.

    Parameters:
        T: The type of value stored in the `ResultReg`.
    """

    alias _mlir_type = __mlir_type[`!kgen.variant<`, T, `, i1>`]
    var _value: Self._mlir_type
    var err: ErrorReg
    """The Error inside the `ResultReg`."""

    fn __init__(inout self):
        """Create a `ResultReg` with a value of None."""
        self = Self(None)
        self.err = ErrorReg()

    fn __init__(inout self, value: T):
        """Create a `ResultReg` with a value.

        Args:
            value: The value.
        """
        self._value = __mlir_op.`kgen.variant.create`[
            _type = Self._mlir_type, index = Int(0).value
        ](value)
        self.err = ErrorReg()

    fn __init__(inout self, value: ErrorReg):
        """Create a `ResultReg` without a value from an `ErrorReg`.

        Args:
            value: The `ErrorReg`.
        """
        self._value = __mlir_op.`kgen.variant.create`[
            _type = Self._mlir_type, index = Int(1).value
        ](__mlir_attr.false)
        self.err = value

    @always_inline
    fn value(self) -> T:
        """Get the `Result` value.

        Returns:
            The contained value.
        """
        return __mlir_op.`kgen.variant.take`[index = Int(0).value](self._value)

    fn __is__(self, other: NoneType) -> Bool:
        """Return `True` if the `Result` has no value.

        It allows you to use the following syntax: `if my_result is None:`

        Args:
            other: The value to compare to (None).

        Returns:
            True if the `ResultReg` has no value and False otherwise.
        """
        return not self.__bool__()

    fn __isnot__(self, other: NoneType) -> Bool:
        """Return `True` if the `ResultReg` has a value.

        It allows you to use the following syntax: `if my_result is not None:`

        Args:
            other: The value to compare to (None).

        Returns:
            True if the Result has a value and False otherwise.
        """
        return self.__bool__()

    fn __bool__(self) -> Bool:
        """Return true if the `ResultReg` has a value.

        Returns:
            True if the `ResultReg` has a value and False otherwise.
        """
        return __mlir_op.`kgen.variant.is`[index = Int(0).value](self._value)
```

Better Error handling than raising and giving people tools to write nicer code

### Any other details?

I skimmed over the Error struct's code and it seems it can be done replacing its fields with just a StringLiteral (?)

martinvuyk:
Add Result, a type modeling a value which may or may not be present. With an Error in the case of failure.

Closes #2748


JoeLoser (commented):
`Result` is definitely a useful type.  I wouldn't eagerly create `ResultReg` or `ErrorReg` yet.  In general, memory-only flavors of types are the way things are trending.  If your counterexample is `OptonalReg`, I wouldn't base future decisions off of that.  That is used in a few load-bearing places internally that aren't easy to move to `Optional` quite yet.

Given that, do you mind splitting out `Result` and its tests into a separate PR?  Any thoughts here, @ConnorGray?

martinvuyk:
>  In general, memory-only flavors of types are the way things are trending

ok that makes sense.

> do you mind splitting out Result and its tests into a separate PR?

Yeah np, I can also just delete everything else from this branch since I already have it in another repo

remorses:
Mojo already has errors and `Optional`, you can combine both to get the same thing, why add a Result type? It adds another way to make the same thing, making the std code more confusing

martinvuyk:
@remorses Optional is for when the function can only return a value or nothing, the information about what went wrong inside the function is lost. 

This is a pattern often used in python:
```python
try:
    something()
except ValueError:
    pass
except IndexError:
    pass
...
```
That is the same as returning a Result with different errors.

If you still aren't convinced I suggest you have a look at Rust's [Result](https://doc.rust-lang.org/std/result/enum.Result.html) and [Option](https://doc.rust-lang.org/std/option/enum.Option.html) enums.

remorses:
The problem here is that in Python there is no way to handle an error in an expression, this is also the case for `match` and `if` statements.

This becomes a problem when you want to assign a variable directly with the result of an `if`, `try` or `match`. The best solution here would be to add a way to create an expression out of a group of statements. Zig has [something similar](https://ziglang.org/documentation/master/#toc-Blocks) and Rust [too](https://doc.rust-lang.org/reference/expressions/block-expr.html), but usually you don't need them because most things in Rust are expressions and not statements.



```mojo
# with without anything becomes a way to turn a block into an expression
var result = with:
  try:
    something()
    break "success"
  except ValueError:
    break "ValueError"
  except IndexError:
    break "IndexError"
```

Same thing in Rust:

```rust
let result = match something() {
    Ok(_) => "success".to_string(),
    Err(e) => match e {
        "ValueError" | "IndexError" => "".to_string(),
        _ => "other error".to_string(),
    },
};
```

martinvuyk:
> create an expression out of a group of statements

That is a completely different discussion. And even then, it still needs a Result type to map to.

With the current implementation one can do:
```mojo
fn func_that_can_err[A: CollectionElement]() -> Result[A]:
    return Error("failed")

fn return_early_if_err[T: CollectionElement, A: CollectionElement]() -> Result[T]:
    var result: Result[A] = func_that_can_err[A]()
    if not result:
        # the internal err gets transferred to a Result[T]
        return result
        # its also possible to do:
        # return None, Error("func_that_can_err failed")
    var val = result.value()
    var final_result: T
    ...
    return final_result
```
but you have to know the string message inside the error beforehand if you want to do any condition based on that.


I also have a proposal that I haven't formalized and I wanted to bring it up in the July 1 meeting that is having a parametrized pseudo-type for Error:
```mojo
struct Error2[T: StringLiteral](Stringable, Boolable):
    """This type represents a parametric Error."""

    alias type = T
    """The type of Error."""
    var message: String
    """The Error message."""
    ...
```
with this result type one could do this
```mojo
fn do_something(i: Int) -> Result2[Int, "IndexError"]:
    if i < 0:
        return None, Error2["IndexError"]("index out of bounds: " + str(i))
    return 1

fn do_some_other_thing() -> Result2[String, "OtherError"]:
    var a = do_something(-1)
    if a.err == "OtherError": # returns bool(err) and err.type == value
        return a # error gets transferred
    elif a.err == "IndexError":
        return a # error message gets transferred
    elif a.err: # some undefined error after an API change
        return a
    return "success"
```

ConnorGray:
Hi @martinvuyk, thanks for taking the time to add this type, and think through what a `Result` type in Mojo 🔥 could look like.

A few comments on this PR:

* The `Result` type in Rust is particularly ergonomic because of Rust language features like the `?` ([`Try`](https://doc.rust-lang.org/std/ops/trait.Try.html)) operator, and Rust's general support for rich sum types (`enum`) and pattern matching (`match`). 

  Mojo does not have those features today, and they aren't currently on our foreseeable roadmap.

* *However*, the `Result` approach to error handling—the reification of "can succeed or fail" as "just another type" that programmers can manipulate and build upon—is a powerful idea in standard library and language design, and I think it'd be reasonable for us to enable the community to use their creativity to explore this design space 🧠 

  - Can `Result` work ergonomically in Mojo today? 🤔 
  - Can `Result` and `raises`-based error handling play nicely together; can you provide "adaptors" between them that programmers find useful?

  And I'm sure many other design questions are worth investigating 🔎 .

Therefore, although I think its very reasonable for folks to point out that what's right for Rust may not be right for Mojo, I think this is the kind of moment where what's possible in this space isn't something anyone really knows for certain. And, consequently, the most positive path forward is to unleash some creativity and see what a `Result` type in Mojo _could_ look like 🙂

Perhaps it won't pan out, but the community has got the brain power and the enthusiasm that I don't want to bet against them and make that experimentation more difficult by siloing away a proven-useful pattern like this.

To that end, I suggest we:

* Move this `Result` type to a new `experimental/result.mojo` top-level module in the standard library.

  This is something the standard library team has discussed and thinks will be generally useful for us to help the community coalesce around core utility types like this, that would otherwise be fragment and make it difficult for projects in the Mojo ecosystem to interoperate. (For example, if projects instead defined their own versions of a `Result` / `std::expected`-like type, duplicated across many folks code bases.)

* Include a **clear** disclaimer that `Result` is not intended to be used in public parts of Mojo stdlib code _outside_ the experimental module.

  I strongly agree with @remorses that we do _not_ want to fragment the way Mojo standard library code represents errors.

  Although we're okay with including `Result` in a clearly-labeled `experimental` part of the stdlib, we're _not_ endorsing this as a replacement or alternative to `raises`-based error handling. We don't want to find ourselves in the situation of having two common but different ways of indicating that a function can result in an error.

  To be crystal clear: folks contributing to the standard library should continue to write code that uses a `raises` / `raise` interface, not `Result`. Any PR introducing a public interface based on `Result` error handling will not be accepted (unless its part of the `experimental` module.)

Please let me know if you think this path forward sounds reasonable to you 🙂

---

P.S. regarding the language feature-related issues that @remorses was highlighting: I generally agree that this; the ergonomics for `Result` based error handling will likely struggle without more dedicated language features. However, I don't think those language features are entirely necessary for `Result` to be usable or even useful. For example, methods on `Result` will enable a "functional" style of working with errors (see all the methods on the `Result` type from Rust for an example of what I mean) that I think has potential advantages over `raises` even _without_ dedicated language features.

That's my hypothesis at least, but I wouldn't say that's a certain outcome 🤷‍♂️. Hence why I think the right think for Mojo and the community is experimentation. We're all fortunate to be working on Mojo relatively early in its life, so experimentation and _making mistakes_ is relatively low risk but potentially very valuable work to see what Mojo can do 🏎️ 🔥 🙂

gabrieldemarmiesse:
@ConnorGray would it make sense for this experimental namespace to be located in a repository `modularml/stdlib-experimental` ? With this setup, it would be possible for external contributors to become maintainers of this repository. We can see that the Modular staff is really busy and having them review features that may not end up in the stdlib can impact Mojo's development speed overall. 

Furthermore, some contributors like @martinvuyk  clearly write code faster than the Mojo maintainers can review it. If we let the community take ownership of this new repository, we can share the review workload.

Finally, it would be a good way to give more responsabilities to the community, who currently has a quite low "glass ceiling" (cannot become maintainers, cannot publish packages, cannot participate in stdlib's meetings, can only contribute to a delayed view of an internal codebase), without impacting the development of internal products.

As the syntax is also changing, the work necessary to handle a breaking change in the Mojo syntax will be lower as we won't garantee that the nightly will always work with this `modularml/stdlib-experimental`  repository.

To garantee that  `modularml/stdlib-experimental` is actually used by the community, we can bundle it into each release, we'd just have a bit of work to sync them before each release, that seems managable.

I think the biggest pains of this repo, from a contributors' perspective, come from the fact that internal code depends on it, which would not be the case for  `modularml/stdlib-experimental` and thus can become a "proper open-source first" Mojo codebase.

PR #2752, opened 2024-05-19
[stdlib] Roundable trait with ndigits param in the round operation and Round to nearest even logic

msaelices:
* Add `Roundable.__round__(ndigits)` method
* Implement the method in all the types that adhere to `Roundable`
* Round to the next even number when the diff is exactly 0.5, recommended way in [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754), which also Python adheres to.
* Tests


stdlib/src/builtin/float_literal.mojo
```diff
@@ -275,6 +275,7 @@ struct FloatLiteral(
             return self
 
         alias one = __mlir_attr.`#kgen.int_literal<1> : !kgen.int_literal`
+        alias minus_one = __mlir_attr.`#kgen.int_literal<-1> : !kgen.int_literal`
```
laszlokindrat, line 278: nit: to be consistent with neg_zero
```suggestion
        alias neg_one = __mlir_attr.`#kgen.int_literal<-1> : !kgen.int_literal`
```

stdlib/src/builtin/math.mojo
```diff
@@ -235,6 +235,9 @@ trait Roundable:
     fn __round__(self) -> Self:
         ...
 
+    fn __round__(self, ndigits: IntLiteral) -> Self:
```
laszlokindrat, line 238: Requiring `ndigits` to be `IntLiteral` will make this impossible to use at runtime. Can you please change this to accept just an `Int`?

stdlib/src/builtin/math.mojo
```diff
@@ -250,3 +253,20 @@ fn round[T: Roundable](value: T) -> T:
         The rounded value of the object.
     """
     return value.__round__()
+
+
+@always_inline
+fn round[T: Roundable](value: T, ndigits: IntLiteral) -> T:
```
laszlokindrat, line 259: See my other comment.

stdlib/src/builtin/int_literal.mojo
```diff
@@ -269,6 +269,47 @@ struct IntLiteral(
...
+
+        Returns:
+            The quotient and remainder of the division.
+        """
+        var quotient: Self = self.__floordiv__(rhs)
+        var remainder: Self = self - (quotient * rhs)
+        return quotient, remainder
+
+    @always_inline("nodebug")
+    fn __round__(self, ndigits: IntLiteral) -> Self:
```
laszlokindrat, line 287: Both here and in the implementation for `FloatLiteral`: Given that `ndigits` is realistically not bigger than what we can store in `Int`, `IntLiteral` is not necessary. This might cause some calculations to happen in finite precision (i.e. return `Int` instead of `IntLiteral`); you can add some special methods for these mixed type computation.

stdlib/test/builtin/test_float_literal.mojo
```diff
@@ -95,6 +92,12 @@ def test_round():
     assert_equal(FloatLiteral.__round__(-1.198, 2), -1.2)
     assert_equal(FloatLiteral.__round__(-1.123, 2), -1.12)
 
+    # Test rounding to nearest even number
+    assert_equal(FloatLiteral.__round__(1.5, 0), 2.0)
+    assert_equal(FloatLiteral.__round__(2.5, 0), 2.0)
+    assert_equal(FloatLiteral.__round__(-2.5, 0), -2.0)
+    assert_equal(FloatLiteral.__round__(-1.5, 0), -2.0)
```
laszlokindrat, line 99: Please add tests with negative value of `ndigits`.

laszlokindrat (commented):
Thanks for working on this! I have a few concerns with the `Int`/`IntLiteral` stuff, but looking good otherwise!

laszlokindrat:
@msaelices Are you still working on this? This is a very nice improvement, and I will love to bring it in!

stdlib/src/builtin/float_literal.mojo
msaelices: Agreed. Done here: https://github.com/msaelices/mojo/commit/c6cf50f3962561f05d943a5b8016c068e39b6e6f

stdlib/src/builtin/math.mojo
msaelices: Done: https://github.com/msaelices/mojo/commit/9de5acca6db1dad73f160a084687cdfaba9f5ed4

stdlib/src/builtin/math.mojo
msaelices: Done in same commit: https://github.com/msaelices/mojo/commit/9de5acca6db1dad73f160a084687cdfaba9f5ed4

stdlib/src/builtin/int_literal.mojo
msaelices: Done: https://github.com/msaelices/mojo/commit/9de5acca6db1dad73f160a084687cdfaba9f5ed4

msaelices:
> @msaelices Are you still working on this? This is a very nice improvement, and I would love to bring it in!

Sure. I'm working on this now.

stdlib/test/builtin/test_float_literal.mojo
msaelices: Actually, negative `ndigits` was not supported. Implemented here: https://github.com/msaelices/mojo/commit/1657b19a075259fc2908a65e8d539fba68fb8db4

msaelices:
@laszlokindrat could you please take another look?

stdlib/src/builtin/float_literal.mojo
```diff
@@ -250,18 +250,26 @@ struct FloatLiteral(
         if not self._is_normal():
             return self
 
+        alias one = __mlir_attr.`#kgen.int_literal<1> : !kgen.int_literal`
+        alias neg_one = __mlir_attr.`#kgen.int_literal<-1> : !kgen.int_literal`
```
laszlokindrat, line 254:
```suggestion
        alias one = __mlir_attr.`#kgen.int_literal<1> : !kgen.int_literal`
        alias neg_one = __mlir_attr.`#kgen.int_literal<-1> : !kgen.int_literal`
```

stdlib/src/builtin/float_literal.mojo
```diff
@@ -250,18 +250,26 @@ struct FloatLiteral(
         if not self._is_normal():
             return self
 
+        alias one = __mlir_attr.`#kgen.int_literal<1> : !kgen.int_literal`
```
laszlokindrat, line 253: Why not just do this?
```suggestion
        alias one: IntLiteral = 1
```

stdlib/src/builtin/float_literal.mojo
```diff
@@ -250,18 +250,26 @@ struct FloatLiteral(
...
+        var abs_diff = abs(self - truncated)
+        var plus_one = one if self > 0 else neg_one
+        if abs_diff == 0.5:
+            # Round to the nearest even number.
+            if truncated % 2 == 0:
+                result = Self(truncated)
+            else:
+                result = Self(truncated + plus_one)
+        elif abs_diff > 0.5:
+            result = Self(truncated + plus_one)
```
laszlokindrat, line 266: here and below: It seems result is only set once, then returned. Can you just early return?
```suggestion
                return Self(truncated)
            else:
                return Self(truncated + plus_one)
        elif abs_diff > 0.5:
            return Self(truncated + plus_one)
```

stdlib/src/builtin/float_literal.mojo
```diff
@@ -275,24 +283,39 @@ struct FloatLiteral(
...
-        if abs(target) - abs(truncated) <= 0.5:
-            result = truncated
-        elif self > 0:
-            result = truncated + 1
+        var abs_diff = abs(target - truncated)
+        var plus_one = one if self > 0 else neg_one
+        if abs_diff == 0.5:
+            # Round to the nearest even number.
+            if truncated % 2 == 0:
+                result = Self(truncated)
```
laszlokindrat, line 308: Same as above: maybe early return?

stdlib/src/builtin/int_literal.mojo
```diff
@@ -269,6 +269,47 @@ struct IntLiteral(
         """
         return self
 
+    @always_inline("nodebug")
+    fn __divmod__(self, rhs: Self) -> Tuple[Self, Self]:
```
laszlokindrat, line 273: non-blocking: maybe in a followup: Maybe it's time for a `DivModable` trait or something. Then we could hook these up into the builtin `divmod` function.

stdlib/src/builtin/simd.mojo
```diff
@@ -1142,6 +1142,18 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
         """
         return llvm_intrinsic["llvm.round", Self, has_side_effect=False](self)
 
+    @always_inline("nodebug")
+    fn __round__(self, ndigits: Int) -> Self:
```
laszlokindrat, line 1146: non-blocking: SIMD might benefit from having a separate overload that does this elementwise between two vectors. Might get tricky for the different dtypes, though. Anyway, just a followup idea.

stdlib/src/builtin/int_literal.mojo
```diff
@@ -269,6 +269,47 @@ struct IntLiteral(
         """
         return self
 
+    @always_inline("nodebug")
```
laszlokindrat, line 272: Can you please add one or two simple tests for this? If you want, you can split it out into a separate patch, to get it in faster, since it's a simple change.

stdlib/src/builtin/int_literal.mojo
```diff
@@ -269,6 +269,47 @@ struct IntLiteral(
...
+
+        Returns:
+            The quotient and remainder of the division.
+        """
+        var quotient: Self = self.__floordiv__(rhs)
+        var remainder: Self = self - (quotient * rhs)
+        return quotient, remainder
+
+    @always_inline("nodebug")
+    fn __round__(self, ndigits: Int) -> Self:
```
laszlokindrat, line 287: I can't see tests for this or its counterpart in `Int`. Maybe forgot to push it?

laszlokindrat (commented):
Great progress, thanks for addressing my reviews! It seems you could split this up into several independent patches for the different types. That would make it easier to land these. Then the patch for the trait could be added after they all landed, and it should be simple. It's up to you, but I think it would speed things up :)

msaelices:
> Great progress, thanks for addressing my reviews! It seems you could split this up into several independent patches for the different types. That would make it easier to land these. Then, the patch for the trait could be added after they all landed, and it should be simple. It's up to you, but I think it would speed things up :)

I think I prefer not to split the PR. It would be harder for me to cherry-pick all the commits or redo my work and track all of them.

stdlib/src/builtin/float_literal.mojo
msaelices: It does not compile:
```
❯ ./stdlib/scripts/run-tests.sh                                                                                                                                                                    
Included from /home/msaelices/src/mojo/stdlib/src/builtin/__init__.mojo:1:                                                                                                                         
/home/msaelices/src/mojo/stdlib/src/builtin/float_literal.mojo:264:30: error: no matching function in initialization                                                                               
                result = Self(truncated + plus_one)                                                                                                                                                
                         ~~~~^~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                
Included from /home/msaelices/src/mojo/stdlib/src/builtin/__init__.mojo:1:                                                                                                                         
/home/msaelices/src/mojo/stdlib/src/builtin/float_literal.mojo:51:8: note: candidate not viable: argument #1 cannot be converted from 'Int' to '!kgen.float_literal'                               
    fn __init__(inout self, value: Self.fp_type):                                                                                                                                                  
       ^                                                                                                                                                                                           
Included from /home/msaelices/src/mojo/stdlib/src/builtin/__init__.mojo:1:                                                                                                                         
/home/msaelices/src/mojo/stdlib/src/builtin/float_literal.mojo:60:8: note: candidate not viable: argument #1 cannot be converted from 'Int' to 'IntLiteral'                                        
    fn __init__(inout self, value: IntLiteral):                                                                                                                                                    
       ^                                                                                                                                                                                           
Included from /home/msaelices/src/mojo/stdlib/src/builtin/__init__.mojo:1:                                                                                                                         
/home/msaelices/src/mojo/stdlib/src/builtin/simd.mojo:180:8: note: candidate not viable: expected at most 1 positional argument, got 2   
```

msaelices: Maybe there is a workaround. I'm still learning the MLIR stuff and Mojo.

stdlib/src/builtin/float_literal.mojo
msaelices: Good idea! Done: https://github.com/msaelices/mojo/commit/5b61c0cab6af500cc0b8f2d8f61ce1c05cee2009

stdlib/src/builtin/float_literal.mojo
msaelices: This time would be more difficult, as we compute a `result` value just for a final operation in the last lines:
```
... # computing result
if ndigits >= 0:                                                                                                                                                                                                         
    result /= Self(multiplier)                                                                                                                                                                                           
elif ndigits < 0:                                                                                                                                                                                                        
    result *= Self(multiplier)
return result
```

stdlib/src/builtin/int_literal.mojo
msaelices: Sure: https://github.com/msaelices/mojo/commit/78e4ee596860fa305c6548871f05cbdfcc484ca0

stdlib/src/builtin/int_literal.mojo
msaelices: Added here: https://github.com/msaelices/mojo/commit/6039f4e21db55f795608f5faf378c8e035f4679e

msaelices:
@laszlokindrat Could you please TAL? :ping_pong:

stdlib/src/builtin/float_literal.mojo
laszlokindrat: Okay, thanks for checking!

laszlokindrat (approved):
Very nice, thank you for working on this!

stdlib/src/builtin/int_literal.mojo
```diff
@@ -269,6 +269,47 @@ struct IntLiteral(
...
+        for _ in range(-ndigits):
+            multiplier = __mlir_op.`kgen.int_literal.binop`[
+                oper = __mlir_attr.`#kgen<int_literal.binop_kind mul>`
+            ](multiplier, ten)
+        alias Pair = Tuple[Self, Self]
+        var t: Pair = self.__divmod__(Self(multiplier))
+        if t[1] * 2 >= multiplier:
+            t[0] += 1
+            t[1] -= multiplier
+        return self - t[1]
```
laszlokindrat, line 311: I'm running into compiler crashes with this internally, and I'm not even sure how it could pass CI. It probably has to do with `Tuple` and `IntLiteral`.

Anyway, it seems you don't actually use t[0], right? Could you please remove `__divmod__` and change this to just use `%`?

msaelices: Sure, It's probably because a problem I had in the beginning and now something was implemented in upstream so now it could be simplified:

Done: https://github.com/msaelices/mojo/commit/e6f55f290b4c1baef25b17ef11fd3d6053d088e9

laszlokindrat:
!sync

laszlokindrat:
!sync

PR #2754, opened 2024-05-19
[stdlib] Add List constructor from Span

lsh:
Being able to convert a Span to an owned value is useful, and List makes the most sense for this type. We could also consider a `to_list()` method on Span instead if that is a better fit.


JoeLoser:
!sync

PR #2762, opened 2024-05-20
[stdlib] Implement `NamedTemporaryFile` in `tempfile` module

artemiogr97:
Implementation of `NamedTemporaryFile` a wrapper around a `FileHandle` that can be used as a context manager that is deleted after closing it if needed.

### DIFF
--- a/stdlib/src/tempfile/tempfile.mojo
+++ b/stdlib/src/tempfile/tempfile.mojo
@@ -270,30 +270,32 @@ struct NamedTemporaryFile:
             dir: Directory in which the file will be created.
             delete: Whether the file is deleted on close.
         """
+
         var final_dir = Path(dir.value()) if dir else Path(
             _get_default_tempdir()
         )
 
         self._delete = delete
+        self.name = ""
 
         for _ in range(TMP_MAX):
             var potential_name = final_dir / (
                 prefix + _get_random_name() + suffix
             )
-            try:
-                if os.path.exists(potential_name) == False:
-                    # TODO for now this name could be relative,
-                    # python implementation expands the path,
-                    # but several functions are not yet implemented in mojo
-                    # i.e. abspath, normpath
-                    self.name = potential_name.__fspath__()
-                    self._file_handle = FileHandle(
-                        potential_name.__fspath__(), mode=mode
-                    )
-                    return
-            except:
-                continue
-        raise Error("Failed to create temporary file")
+            if not os.path.exists(potential_name):
+                # TODO for now this name could be relative,
+                # python implementation expands the path,
+                # but several functions are not yet implemented in mojo
+                # i.e. abspath, normpath
+                self.name = potential_name.__fspath__()
+                break
+        else:
+            raise Error("Failed to create temporary file")
+
+        try:
+            self._file_handle = FileHandle(self.name, mode=mode)
+        except:
+            raise Error("Failed to create temporary file")
 
     @always_inline
     fn __del__(owned self):
--- a/stdlib/src/tempfile/tempfile.mojo
+++ b/stdlib/src/tempfile/tempfile.mojo
@@ -317,7 +317,7 @@ struct NamedTemporaryFile:
         """Moves constructor for the file handle.
 
         Args:
-          existing: The existing file handle.
+            existing: The existing file handle.
         """
         self._file_handle = existing._file_handle^
         self._delete = existing._delete
@@ -331,7 +331,7 @@ struct NamedTemporaryFile:
             size: Requested number of bytes to read.
 
         Returns:
-          The contents of the file.
+            The contents of the file.
         """
         return self._file_handle.read(size)
 
@@ -343,7 +343,7 @@ struct NamedTemporaryFile:
             size: Requested number of bytes to read.
 
         Returns:
-          The contents of the file.
+            The contents of the file.
         """
         return self._file_handle.read_bytes(size)
 
@@ -366,7 +366,7 @@ struct NamedTemporaryFile:
         """Write the data to the file.
 
         Args:
-          data: The data to write to the file.
+            data: The data to write to the file.
         """
         self._file_handle.write(data)
 
--- a/stdlib/src/tempfile/tempfile.mojo
+++ b/stdlib/src/tempfile/tempfile.mojo
@@ -239,3 +239,133 @@ struct TemporaryDirectory:
             return True
         except:
             return False
+
+
+struct NamedTemporaryFile:
+    """A handle to a temporary file."""
+
+    var _file_handle: FileHandle
+    """The underlying file handle."""
+    var _delete: Bool
+    """Whether the file is deleted on close."""
+    var name: String
+    """Name of the file."""
+
+    fn __init__(
+        inout self,
+        mode: String = "w",
+        suffix: String = "",
+        prefix: String = "tmp",
+        dir: Optional[String] = None,
+        delete: Bool = True,
+    ) raises:
+        """Create a named temporary file. Can be used as a context manager.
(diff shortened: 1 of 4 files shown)


artemiogr97:
@JoeLoser, @laszlokindrat, I'm having some issues with this:
The implementation that I had in #2352 used to work in an older release, in the current release I had the following compiler error:

```
for _ in range(TMP_MAX):  #error here: field 'self._file_handle' destroyed out of the middle of a value, preventing the overall value from being destroyed
    var potential_name = final_dir / (
        prefix + _get_random_name() + suffix
    )
    if os.path.exists(potential_name):
        continue
    try:
        self.name = potential_name
        self._file_handle = FileHandle(potential_name, mode=mode)
        return
    except:
        continue
```

I had to put the if condition inside the for to be able to compile:
````
for _ in range(TMP_MAX):
            var potential_name = final_dir / (
                prefix + _get_random_name() + suffix
            )
            try:
                if os.path.exists(potential_name) == False:
                    self.name = potential_name
                    self._file_handle = FileHandle(potential_name, mode=mode)
                    return
            except:
                continue
```
But now there is a runtime error which I don't really understand, I also had the same tests on my old PR and everything was ok.

laszlokindrat:
@artemiogr97 I think this is unblocked now. Are you still planning to work on this?

artemiogr97:
@laszlokindrat yes, I was planning to check again this week

laszlokindrat:
> @laszlokindrat yes, I was planning to check again this week

Awesome, thanks, please ping me when you have something ready for review.

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -239,3 +239,133 @@ struct TemporaryDirectory:
...
+        )
+
+        self._delete = delete
+
+        for _ in range(TMP_MAX):
+            var potential_name = final_dir / (
+                prefix + _get_random_name() + suffix
+            )
+            try:
+                if os.path.exists(potential_name) == False:
```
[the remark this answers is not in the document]
Dan13llljws: I am still looking into the crash, but 

```mojo
        var final_dir = Path(dir.value()) if dir else Path(
            _get_default_tempdir()
        )

        self._delete = delete
        self.name = ""

        for _ in range(TMP_MAX):
            var potential_name = final_dir / (
                prefix + _get_random_name() + suffix
            )
            if not os.path.exists(potential_name):
                self.name = potential_name.__fspath__()

        try:
            self._file_handle = FileHandle(self.name, mode=mode)
        except:
            raise Error("bad")
```
seems to work?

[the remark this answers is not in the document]
artemiogr97: thanks @Dan13llljws 

@laszlokindrat this should be ready for review now

artemiogr97:
> @JoeLoser, @laszlokindrat, I'm having some issues with this: The implementation that I had in #2352 used to work in an older release, in the current release I had the following compiler error:
> 
> ```
> for _ in range(TMP_MAX):  #error here: field 'self._file_handle' destroyed out of the middle of a value, preventing the overall value from being destroyed
>     var potential_name = final_dir / (
>         prefix + _get_random_name() + suffix
>     )
>     if os.path.exists(potential_name):
>         continue
>     try:
>         self.name = potential_name
>         self._file_handle = FileHandle(potential_name, mode=mode)
>         return
>     except:
>         continue
> ```
> 
> I had to put the if condition inside the for to be able to compile:
> 
> ```
> for _ in range(TMP_MAX):
>             var potential_name = final_dir / (
>                 prefix + _get_random_name() + suffix
>             )
>             try:
>                 if os.path.exists(potential_name) == False:
>                     self.name = potential_name
>                     self._file_handle = FileHandle(potential_name, mode=mode)
>                     return
>             except:
>                 continue
> ```
> 
> But now there is a runtime error which I don't really understand, I also had the same tests on my old PR and everything was ok.

Turns out that adding an else after the for loop makes the compiler happier, style-wise I prefer the first implementation that I had so I changed it again

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -239,3 +239,137 @@ struct TemporaryDirectory:
...
+            var potential_name = final_dir / (
+                prefix + _get_random_name() + suffix
+            )
+            if os.path.exists(potential_name):
+                continue
+            try:
+                # TODO for now this name could be relative,
+                # python implementation expands the path,
+                # but several functions are not yet implemented in mojo
+                # i.e. abspath, normpath
```
laszlokindrat, line 291: Followup idea: Maybe it would be better to do this earlier than later, given that it would be already useful in a few places.

stdlib/test/tempfile/test_tempfile.mojo
```diff
@@ -177,7 +177,59 @@ fn test_temporary_directory() raises -> None:
     assert_false(exists(tmp_dir), "Failed to delete temp dir " + tmp_dir)
 
 
+fn test_named_temporary_file_deletion() raises:
```
laszlokindrat, line 180: Here and elsewhere in the file: could you please use `def` instead of `fn` + `raises`? We're trying to close the gap between these and prefer `def` when possible.

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -239,3 +239,137 @@ struct TemporaryDirectory:
...
+
+        Args:
+          existing: The existing file handle.
+        """
+        self._file_handle = existing._file_handle^
+        self._delete = existing._delete
+        self.name = existing.name^
+
+    @always_inline
+    fn read(self, size: Int64 = -1) raises -> String:
```
laszlokindrat, line 327: Q: Just to confirm, providing these wrappers is necessary because we need to allow this to be used both within a context manager and as a normal object, right? (And also because in python this is actually a function).

This is fine, but I wonder if we could introduce a trait for things that implement these FileHandle-like interfaces.

stdlib/src/tempfile/tempfile.mojo
```diff
@@ -239,3 +239,137 @@ struct TemporaryDirectory:
...
+        """Reads the data from the file.
+
+        Args:
+            size: Requested number of bytes to read.
+
+        Returns:
+          The contents of the file.
```
laszlokindrat, line 334: style nit: here and elsewhere: please ensure correct indentation in the docstring.
```suggestion
            The contents of the file.
```

laszlokindrat (approved):
Great, thank you! I mostly have nits and questions. Could you please add a changelog entry, and update the PR title and description with a summary of the patch?

stdlib/src/tempfile/tempfile.mojo
artemiogr97: yes, I don't know if there is a better way to do this in mojo for now, just as a reference the python implementation does not wraps all the methods of the underlying object, it reimplements  `__getattr__` to directly return the methods of the real file object: https://github.com/python/cpython/blob/e83ce850f433fd8bbf8ff4e8d7649b942639db31/Lib/tempfile.py#L489-L506

I guess it would make sense to define some traits for IO stuff, the equivalent of the clases defined here: https://docs.python.org/3/library/io.html

stdlib/src/tempfile/tempfile.mojo
artemiogr97: yes, I don't think I will have some time soon, but I opened two new issues to implement those functions: #2961 and #2962

JoeLoser (approved):
This is great, thank you for working on this! @Dan13llljws is excited to use this new functionality soon 😄

JoeLoser:
!sync

PR #2768, opened 2024-05-21
[stdlib] Add count() method for InlineList

northstreet12:
The PR adds count() method for InlineList

*count*
```mojo
var my_list = InlineList[Int](1, 2, 3)
print(my_list.count(1))
```

### DIFF
--- a/stdlib/src/collections/inline_list.mojo
+++ b/stdlib/src/collections/inline_list.mojo
@@ -200,6 +200,34 @@ struct InlineList[ElementType: CollectionElement, capacity: Int = 16](Sized):
         return False
 
     @always_inline
+    fn count[C: ComparableCollectionElement](self: Self, value: C) -> Int:
+        """Counts the number of occurrences of a value in the list.
+
+        ```mojo
+        var my_list = InlineList[Int](1, 2, 3)
+        print(my_list.count(1))
+        ```
+        Parameters:
+            C: The type of the elements in the list. Must implement the
+              traits `EqualityComparable` and `CollectionElement`.
+
+        Args:
+            value: The value to count.
+
+        Returns:
+            The number of occurrences of the value in the list.
+        """
+        constrained[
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/inline_list.mojo
```diff
@@ -195,3 +195,31 @@ struct InlineList[ElementType: CollectionElement, capacity: Int = 16](Sized):
...
+        """
+        constrained[
+            _type_is_eq[ElementType, C](), "value type is not self.ElementType"
+        ]()
+
+        var count = 0
+        for elem in self:
+            if value == rebind[C](elem[]):
+                count += 1
+        return count
```
gabrieldemarmiesse, line 225: This code is exactly the same as the one found in `List`. Is it possible to group them together to avoid duplication?

JoeLoser: +1 to grouping them together.  This may even start seeding what will be our more generic algorithms across collections.

northstreet12: Is it to say, we might need one helper struct to hold methods like count for all the collections structs? Or we could wait for implementation of struct inheritance?

gabrieldemarmiesse: You can make a normal function with a trait instead of a specific list. Then you can call this function from both `List` and `InlineList`

northstreet12: If we do so, we might need an ‘ElementCountable’ trait which inherits an ‘Iterable’ trait. It seems to be a little difficult implementing an ‘Iterable’ trait. Any suggestions?

JoeLoser: Valid point re the lack of `Iterable` trait right now.  Let's keep this as-is and refactor in follow-ups once the library is further along.

JoeLoser (approved):

JoeLoser:
!sync

Issue #2761, opened 2024-05-20
[Feature Request] [mojo-lang] [proposal] Make fstring formatting a priority and define what features will be available


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Fstring formatting i.e. `f"my text with an embedded value: {some_var}, it's so simple"` is one of the best features of Python IMO, it helps really simplify dealing with strings. What I don't like, is how complicated the `"".format()` PEP specification is...

What I would propose is that we simplify things and let the programmer decide how to represent their types. Make it so that f strings are syntax sugar for string concatenation.

If someone wants to represent, say, an int's hex value, then they should just do f"some hex_value: {hex(10)}", instead of the `StringLiteral.format()` method needing to have a bunch of edge cases and different handcoded support for each stdlib type...

We can also simplify things even more by saying that  `f"my text with an embedded value: {some_var}, it's so simple"` generates code that looks like `("my text with an embedded value: " + (some_var) + ", it's so simple")` that way chained f strings would be simpler to implement.
```mojo
var a  = f"Just imagining that {f"I can chain {f"my text with an embedded value: {some_var}, it's so simple"} that I can almost taste"} the absolute joy of writing in this language"
```
Would codegen into
```mojo
var a = "Just imagining that" + ("I can chain  " + ("my text with an embedded value: " + (some_var) + ", it's so simple") + "that I can almost taste") + "the absolute joy of writing in this language"
```

### What is your motivation for this change?

Adding support for Python fstrings is important. But I think we shouldn't implement it exactly the same. We could do a `StringLiteral.format()` equal to Python's, but I don't think fstrings should have templating capabilities since that just overcomplicates things and we can leave those cases to `StringLiteral.format()` anyway

bgreni:
The ubiquitous use and utility of `format` is precisely why we cannot intentionally shortcut the implementation for Mojo. It needs to work the same as Python programmers expect it to eventually. However as you mentioned the standard is quite complex and given it doesn't quite align with the short term priorities for Mojo it's unlikely to be prioritized anytime soon I would bet.

I agree having trivial stand in support for now would be nice though!

laszlokindrat:
[...]

PR #2771, opened 2024-05-21
[stdlib] Add `String.format` method

rd4com:
Partial " fix for https://github.com/modularml/mojo/issues/2761 

Design is actively refined in #2761 as suggested by @laszlokindrat .

The current PR provides one `format` method overload,
it replaces occurrences of `{i}` by `args[i]`,

```mojo
print(
    String("{1} {0} {1}").format(
        "Mojo",  "🔥"
    )
)
# 🔥 Mojo 🔥
```

It also support automatic indexing (`{}`)
```mojo
print(
    String("Hello {} {}").format("Beautiful", "world")
)
```

```mojo
print(
    String("{} {}").format(True, 1.125)
)
```

We took great care to make sure it follows python behaviours.

The struct already support `**kwargs` for another patch.


stdlib/src/builtin/string_literal.mojo
```diff
@@ -346,3 +346,36 @@ struct StringLiteral(
...
+
+        Values need to implement `Stringable`.
+
+        If any of the replacement values contains curly braces,
+
+        an error is raised: `"A replacement value contains {}"`.
+
+        """
+
+        var res = String(self)
```
bgreni, line 376: Since you're creating a string anyways. It would be preferable to implement the logic in `String` itself and then here simply do something like this so we can have this functionality in both types.
```mojo
var res = String(self)
return res.format_simple(kwargs)
```

stdlib/test/builtin/test_string_literal.mojo
```diff
@@ -142,6 +142,43 @@ fn test_repr() raises:
...
+        "Number: {n}".format_simple(
+            n=123,
+        )
+        == "Number: 123"
+    )
+
+    with assert_raises(contains="A replacement value contains {}"):
+        _ = "Should raise simple error: {n}".format_simple(n="{123}")
+
+    assert_true(
```
bgreni, line 161: Use `assert_equals` so if a test fails it will print the two inputs and we can easily see which test failed and how.

laszlokindrat (commented):
Thanks for the patch! Before we move forward with this, let's discuss the bigger picture as well as some implementation and design details in the related issue: https://github.com/modularml/mojo/issues/2761

laszlokindrat:
Can you please rebase on the next nightly, and update the PR description to reflect the recant changes?

stdlib/src/builtin/string_literal.mojo
rd4com: Yes, thanks to both of you for the great suggestion :+1:

stdlib/test/builtin/test_string_literal.mojo
rd4com: Yes, it is an excellent idea, need help for this one, as it already raises in the `with` block

laszlokindrat: @rd4com can you clarify what you need help with?

laszlokindrat:
I'm very excited about this patch, thank you for pushing it forward! Please update the PR description and let me know when it's ready for review again!

stdlib/test/builtin/test_string_literal.mojo
rd4com: Thanks :+1: , how to both `assert_raise` and `assert_equal` (because there is no value if it raises)

rd4com:
@laszlokindrat If people are happy, i'm happy too ! I learned a lot about git with rebases :partying_face: 

Does the PR description mean the one on github or the commit message ?

laszlokindrat:
> Does the PR description mean the one on github or the commit message ?

On github; we squash and merge internally, and the github PR description becomes the commit message that lands (unless we manually edit, which makes us sad usually).

stdlib/test/builtin/test_string_literal.mojo
laszlokindrat: It's one or the other; if it raises, you can't compare it's output. And if it had an output, it did not raise, so you can use `assert_equal`. Maybe I'm misunderstanding, do you have a concrete example?

rd4com: I think i'm confused, it is expected that it raises in the next example:
```mojo
with assert_raises(contains="args[1] was provided but not used"):
     print("First {0}".format("First", True))
```
No value is returned, not sure how to ```assert_equal```;

Maybe @bgreni was refering to an ```assert_true``` below an ```assert_raise``` ?

laszlokindrat: Yes, he was referring to the assert_true, and I think you already changed that.

rd4com:
Thanks for explaining :+1: 
This one is good to go i think,

WDYT ? @bgreni @laszlokindrat

stdlib/test/builtin/test_string.mojo
```diff
@@ -1082,6 +1082,73 @@ def test_indexing():
...
+    assert_equal(String("Is {0} even? {1}").format(2, True), "Is 2 even? True")
+
+    assert_equal(String("{0} {1}").format("{0}", True), "{0} True")
+
+    with assert_raises(
+        contains="Recursion with a replacement value containing bracket"
+    ):
+        print(String("Is {0} even? {1}").format("{1}", True))
+
+    with assert_raises(contains="args[1] was provided but not used"):
```
bgreni, line 1108: Whats the behaviour if the number of args is less expected? That should raise as well.

stdlib/src/builtin/string.mojo
```diff
@@ -1957,6 +1957,58 @@ struct String(
...
+        var occurences = InlineList[Int, len_Ts]()
+
+        @parameter
+        for i in range(len_Ts):
+            occurences.append(self.count("{" + str(i) + "}"))
+
+        @parameter
+        for i in range(len_Ts):
+            if self.count("{" + str(i) + "}") != occurences[i]:
+                raise ("Recursion with a replacement value containing brackets")
```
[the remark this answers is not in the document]
rd4com: Yes, just played the example in `🐍 python` and it quite handles it gracefully !

rd4com:
Reviews by @bgreni :+1:  required to implement more and my head is still spinning.

Making progress toward :snake: with automatic indexing and kwargs! 

**current patch does not work in the parameter domain**

Not sure if it's possible, but let's not give up :partying_face: 

Please make more reviews and suggestions to solve the parameter domain

stdlib/src/utils/_format.mojo
```diff
@@ -147,3 +147,83 @@ fn write_to[*Ts: Formattable](inout writer: Formatter, *args: *Ts):
...
 
     args.each[write_arg]()
+
+
+# ===----------------------------------------------------------------------===#
+# Format method structures
+# ===----------------------------------------------------------------------===#
+
+
+@value
```
laszlokindrat, line 157: I think this can live in `string.mojo` as `_FormatCurlyEntry`. There is no need to add a new file in `utils` at this point.

stdlib/src/builtin/string.mojo
```diff
@@ -1971,6 +1972,115 @@ struct String(
             )
         return String(buf^)
 
+    @always_inline
+    fn _format[
+        *Ts: Stringable
+    ](
+        inout self,
+        args: VariadicPack[_, _, Stringable, Ts],
+        kwargs: OwnedKwargsDict[String],
```
laszlokindrat, line 1981: As I mentioned, let's keep the scope limited initially. Variadic keyword-only arguments are very much unstable, so I would prefer we start with just positional substitutions. Then you can extend in a followup patch.

stdlib/src/builtin/string.mojo
```diff
@@ -1971,6 +1972,115 @@ struct String(
             )
         return String(buf^)
 
+    @always_inline
+    fn _format[
+        *Ts: Stringable
+    ](
+        inout self,
+        args: VariadicPack[_, _, Stringable, Ts],
```
laszlokindrat, line 1980: I'm not sure variadic packs work in parameter contexts, this might be the issue.

stdlib/src/builtin/string_literal.mojo
```diff
@@ -340,3 +341,37 @@ struct StringLiteral(
...
+
+        Returns:
+            The formated template.
+
+        Variadic arguments are required to implement `Stringable`.
+
+        """
+        var res = String(self)
+        res._format(args, kwargs)
+        return res^
```
laszlokindrat, line 377: Why not just
```suggestion
        return String(self)._format(args, kwargs)
```

stdlib/src/builtin/string.mojo
```diff
@@ -1957,6 +1957,58 @@ struct String(
...
+            Ts: The types of the variadic arguments.
+
+        Returns:
+            A `String` built by replacing `{i}` with `args[i]`.
+
+        Values need to implement `Stringable`.
+
+        """
+        var res: String
+        String.__copyinit__(res, self)
```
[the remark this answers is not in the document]
laszlokindrat: Yeah, why not just assign and let the compiler copy?

stdlib/src/builtin/string.mojo
```diff
@@ -1971,6 +1972,115 @@ struct String(
             )
         return String(buf^)
 
+    @always_inline
+    fn _format[
+        *Ts: Stringable
+    ](
+        inout self,
```
laszlokindrat, line 1979: Why is it necessary to mutate self in this method? In the implementation you create `res`, you could just return it.

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
laszlokindrat: But see my other comment: I'm not sure why `_format` needs to make in-place modification in the first place.

stdlib/src/utils/_format.mojo
rd4com: Yep, this review can be implemented

stdlib/src/builtin/string.mojo
rd4com: Yes, this review can be implemented

stdlib/src/builtin/string.mojo
rd4com: :+1: , this review can be implemented,
it would require to refine a previous review, which asked to do `String(self).format(args)`:
- `*args` cannot be passed to another function that takes `*args`, 
only to one that takes `VariadicPack`or `VariadicListMem`.

*(`str` with `VariadicPack` don't work in the parameter domain.)*

If we design it to work in the parameter domain,
both `StringLiteral` and `String` need an duplicate implementation that takes `*args`.
*(`str` works with "simple" `*args` in the parameter domain)*

Need an answer for this review, 
can both `.format()` be a duplicate of exactly the same logic?

stdlib/src/builtin/string_literal.mojo
rd4com: Yes, this review can be implemented too

stdlib/src/builtin/string.mojo
rd4com: Yes, this review can be implemented (`borrowed` self and `return` a result)

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
rd4com: This review can be implemented, just thought explicit was better than implicit,
other comment can be implemented too :+1:

stdlib/src/builtin/string.mojo
laszlokindrat: I think if you change the implementation in `StringLiteral` to just use `String.format`, you would not have to forward `*args`, and you could just get rid of `_format` completely.

rd4com: Yes, lets not forward `*args` but how because `String.format` need it?

laszlokindrat: You can just implement `String.format`. Then `StringLiteral` can call it directly. If you're unsure, just address the other reviews, and we'll get back to this later.

rd4com: @laszlokindrat Sure, here is a commit !

stdlib/src/builtin/string.mojo
```diff
@@ -1979,6 +1980,101 @@ struct String(
...
+        Returns:
+            The formated template.
+
+        Variadic arguments are required to implement `Stringable`.
+
+        """
+        var Entries = _FormatCurlyEntry.create_entries(self)
+        var manual_indexing_count = 0
+        var automatic_indexing_count = 0
+        var kwargs_indexing_count = 0
```
laszlokindrat, line 2023: nits: This and its uses can be removed in this patch.
```suggestion
```

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2205,83 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
         return 128 + 1  # Add 1 for the terminator
+
+
+# ===----------------------------------------------------------------------===#
+# Format method structures
+# ===----------------------------------------------------------------------===#
+
+
+@value
+struct _FormatCurlyEntry:
```
laszlokindrat, line 2216: Q: Now that things are cleaning up, why do we need this struct at all? Looks like you never instantiate it, and it just has a single static method. Can we just have an ordinary function instead?

stdlib/src/builtin/string.mojo
```diff
@@ -1979,6 +1980,101 @@ struct String(
...
+        Parameters:
+            Ts: The types of the variadic arguments.
+
+        Returns:
+            The formated template.
+
+        Variadic arguments are required to implement `Stringable`.
+
+        """
+        var Entries = _FormatCurlyEntry.create_entries(self)
```
laszlokindrat, line 2020: style nit:
```suggestion
        var entries = _FormatCurlyEntry.create_entries(self)
```

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2205,83 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+
+    Purpose of the variant types:
+
+    - `String` for **kwargs
+
+        ```mojo
+        "{message}".format(message="hello")
+        ```
+
+        value field contains `message`
```
laszlokindrat, line 2228:
```suggestion
```

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2205,83 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+
+    This struct only supports theses:
+
+    `{number|name|nothing}`
+
+    """
+
+    var first_curly: Int
+    var last_curly: Int
+    var value: Variant[String, Int, NoneType]
```
laszlokindrat, line 2256: Can you please go through and update the code to ensure that complexity from the keyword handling is removed?

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2205,83 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+        Args:
+            arg: The "format" part provided by the user.
+
+        Returns:
+            A `List` of structured format entries.
+        """
+        var Entries = List[Self]()
+        var start = Optional[Int](None)
+        for i in range(len(arg)):
+            if arg[i] == "{":
```
laszlokindrat, line 2271: What happens with nested braces? e.g. `"{{0}}".format("foo")`? And what if it's inbalanced, e.g. `"{{0}".format("foo")`?

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2205,83 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+                continue
+            if arg[i] == "}":
+                if start:
+                    var start_value = start.value()[]
+                    var tmp = Self(start_value, i, None)
+                    if i - start_value != 1:
+                        var tmp2 = arg[start_value + 1 : i]
+                        try:
+                            var tmp3 = int(tmp2)
+                            tmp.value = tmp3
```
laszlokindrat, line 2282: What's raising here and what kind of exception are we trying to catch?

stdlib/src/builtin/string.mojo
```diff
@@ -1979,6 +1980,101 @@ struct String(
...
+            raise "Cannot both use manual and automatic indexing for *args"
+
+        for e in Entries:
+            if e[].value.isa[String]():
+                raise "Index " + e[].value[String] + " not in kwargs"
+            if manual_indexing_count:
+                if e[].value.isa[Int]() and e[].value[Int] >= len(args):
+                    raise ("Index " + str(e[].value[Int]) + " not in *args")
+            if automatic_indexing_count:
+                if automatic_indexing_count > len(args):
```
laszlokindrat, line 2042: Since `len` of anything is necessarily non-negative, this check is redundant.
```suggestion
            if automatic_indexing_count > len(args):
```

stdlib/src/builtin/string.mojo
rd4com: Would love to, but we cannot return a list of tuples yet,  what else could we return ?

stdlib/src/builtin/string.mojo
rd4com: If the `int` raises it means it is a keyword for **kwargs (`{name}`),
if it does not raises, it is a number that can be used for manual indexing (`{0} {1}`)
If the slice is len(0), it is `NoneType`, for automatic indexing (`{} {}`)

I think we need thoses 3, even if not using kwargs, in order to raise errors and ensure python behaviour;

stdlib/src/builtin/string.mojo
rd4com: :+1:  @laszlokindrat,
New commit contains more tests, including thoses two!
We can always add more if needed;
Let's keep this review conversation unresolved for now, so we remember later ?

stdlib/src/builtin/string.mojo
rd4com: Yes it should, but having the one for keyword let's us raise an error if the user do:
```mojo
"{} {foo}".format("bar")
#❌ Index foo not in kwargs
```
Even if we do not support keyword args, the error need to be raised no ?
Maybe we should parametrize `_FormatCurlyEntry.create_entries[raises_on_kwargs:Bool]` ?

stdlib/src/builtin/string.mojo
laszlokindrat: Okay, thanks for explaining these! Can you please put these in a comment here? Also, would it be possible to handle these cases in some more explicit way? We don't have typed errors yet, so exception-as-control-flow is not as expressive as in Python.

stdlib/src/builtin/string.mojo
laszlokindrat: Fair point, this is fine, thank you!

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2198,76 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+
+    This struct only supports theses:
+
+    `{number|name|nothing}`
+
+    """
+
+    var first_curly: Int
+    var last_curly: Int
+    var value: Variant[String, Int, NoneType]
```
laszlokindrat, line 2242: I think Python refers to these as `field`. Either way, something other than `value` would be better.
```suggestion
    var field: Variant[String, Int, NoneType]
```

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2198,76 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+
+    `*_curly` are the indexes of the enclosing curlies
+
+    This struct only supports theses:
+
+    `{number|name|nothing}`
+
+    """
+
+    var first_curly: Int
```
laszlokindrat, line 2240: Instead of the large docstring above, could you please add simple descriptions for these fields? E.g. "the index of an opening brace around a substitution field", or something.

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2198,76 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+    Purpose of the variant types:
+
+
+    - `Int` for manual indexing of *args
+
+        ```mojo
+        "{0}".format("hello")
+        ```
+
+        value field contains `0`
```
laszlokindrat, line 2222: It would be better to move this description to the `create_entries` function, since that's where this logic actually lives. Also, just to decrease cognitive load, let's use an example that doesn't refer to `format`, e.g. describing the result of
```mojo
_FormatCurlyEntry.create_entries("{0} foo {1}")
```

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2198,76 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+    fn create_entries(arg: String) -> List[Self]:
+        """Used internally by the `format()` method.
+
+        Args:
+            arg: The "format" part provided by the user.
+
+        Returns:
+            A `List` of structured format entries.
+        """
+        var Entries = List[Self]()
```
laszlokindrat, line 2254: style nit:
```suggestion
        var entries = List[Self]()
```

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2198,76 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+        var start = Optional[Int](None)
+        for i in range(len(arg)):
+            if arg[i] == "{":
+                start = i
+                continue
+            if arg[i] == "}":
+                if start:
+                    var start_value = start.value()[]
+                    var tmp = Self(start_value, i, None)
```
laszlokindrat, line 2263: Here and elsewhere: Can you please come up with more meaningful variable names? For example `tmp` here is something like `current_entry`, right?

stdlib/src/builtin/string.mojo
laszlokindrat: Why not just raise directly from `create_entries`? This `create_entries` function is what performs the actual parsing, so the rules of what the fields/entries can look like should be defined and checked here, for example `raise "Cannot both use manual and automatic indexing"`. Right now, the fields cannot be strings yet, which you can enforce here. Basically, I want to make we check as much as we can here before handing over the entry list to format later.

We don't have the ability to raise parametrically yet, but that's okay, just mark it raising.

stdlib/src/builtin/string.mojo
```diff
@@ -1979,6 +1980,94 @@ struct String(
...
+
+        """
+        var entries = _FormatCurlyEntry.create_entries(self)
+        var manual_indexing_count = 0
+        var automatic_indexing_count = 0
+        for e in entries:
+            if e[].value.isa[Int]():
+                manual_indexing_count += 1
+                if e[].value[Int] >= len(args):
+                    raise ("Index " + str(e[].value[Int]) + " not in *args")
```
laszlokindrat, line 2027: Here and in all other errors: Can you please check what Python throws and ensure that we emit a similar/identical error messages?

stdlib/src/builtin/string.mojo
```diff
@@ -1979,6 +1980,94 @@ struct String(
...
+            elif e[].value.isa[NoneType]():
```
laszlokindrat, line 2028: suggestion: Would be nice to have helpers that allow these checks to be done like this:
```suggestion
            elif e.is_none():
```

stdlib/src/builtin/string.mojo
```diff
@@ -1979,6 +1980,94 @@ struct String(
...
+        var start = 0
+
+        if manual_indexing_count:
+            for e in entries:
+                debug_assert(start < len(self), "start >= len(self)")
+                res += self[start : e[].first_curly]
+                if e[].value.isa[Int]():
+
+                    @parameter
+                    for i in range(len(VariadicList(Ts))):
```
laszlokindrat, line 2048: suggestion: to make the syntax less crowded, you could introduce
```mojo
alias num_pos_args = len(VariadicList(Ts))
```

stdlib/src/builtin/string.mojo
```diff
@@ -1979,6 +1980,94 @@ struct String(
...
+                automatic_indexing_count += 1
+            elif e[].value.isa[String]():
+                raise "Index " + e[].value[String] + " not in kwargs"
+            if manual_indexing_count and automatic_indexing_count:
+                raise "Cannot both use manual and automatic indexing"
+
+        if automatic_indexing_count > len(args):
+            raise ("Automatic indexing require more args in *args")
+
+        var res: String = ""
```
laszlokindrat, line 2038: followup idea: It would be nice if we could preallocate the correct result size here. Let's keep this patch simple, but in a followup you could probably modify `create_entries` to count the number of characters that are are not in the fields, right?

stdlib/src/builtin/string.mojo
```diff
@@ -2109,3 +2198,76 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+                    if i - start_value != 1:
+                        var tmp2 = arg[start_value + 1 : i]
+                        try:
+                            var tmp3 = int(tmp2)
+                            tmp.value = tmp3
+                        except:
+                            tmp.value = tmp2
+                    Entries.append(tmp)
+                start = None
+        return Entries
```
laszlokindrat, line 2273: Can we move these?
```suggestion
                    Entries.append(tmp^)
                start = None
        return Entries^
```

stdlib/test/builtin/test_string.mojo
```diff
@@ -1090,6 +1090,87 @@ def test_indexing():
...
+    assert_equal(
+        String(" {} , {} {} !").format(
+            "Hello",
+            "Beautiful",
+            "World",
+        ),
+        " Hello , Beautiful World !",
+    )
+
+    assert_equal(String("{{0}}").format("foo"), "{foo}")
```
laszlokindrat, line 1125: In Python, double braces are an escape sequence for `format`, so `"{{0}}".format("foo")` will actually yield `{0}`. Can we not support that for some reason?

stdlib/test/builtin/test_string.mojo
```diff
@@ -1090,6 +1090,87 @@ def test_indexing():
...
+
+    # Cannot call potentially raising function function in alias initializer
+    # alias A = "Mojo"
+    # alias B = !
+    # alias Result = String("{0} {1}").format(A,B)
+
+    # @parameter
+    # if Result != "Mojo !":
+    #    raise "Assertion failed (alias): "
+
```
laszlokindrat, line 1172:
```suggestion
```

laszlokindrat (commented):
This is looking really good, thank you! I did a more in-depth review, and I have a few questions, most importantly about error handling and compliance with Python's behavior, but I think we're getting pretty close!

laszlokindrat:
Also, please rebase on the latest nightly! Lots have been changing with references, allowing this code to be simplified.

rd4com:
Thanks so much for taking the time to review this well @laszlokindrat ,
i'll work on it today ! 

For speed, have a look at: [branch_add_format](https://github.com/rd4com/mojo_branch/tree/branch_add_format) *(iterator of reference)*

stdlib/test/builtin/test_string.mojo
rd4com: I think we can support that easily with small changes

stdlib/src/builtin/string.mojo
rd4com: Yes thought about it too, something like `capacity = len(self) - len(args) ?`
Let's keep this review open so we don't forget about it ?
That way we can do optimization at the end ?

stdlib/src/builtin/string.mojo
rd4com: Yep, that makes sense lol

stdlib/test/builtin/test_string.mojo
rd4com: :partying_face:  Done but we will need to decomplexify it,
i'm learning `🐍python.format` on the fly and it's relatively complex (for me),
let's continue the hill climb !

stdlib/src/builtin/string.mojo
rd4com: Yes ! Did some improvements, we might need to do another review of that part

stdlib/src/builtin/string.mojo
rd4com: ➡️ Updated the test to reflect what python does with theses (raising errors and escaping)

stdlib/src/builtin/string.mojo
rd4com: Yes, we are now raising from `create_entries` :+1:

rd4com:
:partying_face: It works, the hill climb is steep; is there more python behaviours for this patch?
I think we'll need to decomplexify first, then we can build on it again;

stdlib/src/builtin/string.mojo
```diff
@@ -1970,6 +1971,79 @@ struct String(
...
+        var current_automatic_arg_index = 0
+        for e in entries:
+            debug_assert(pos_in_self < len(self), "pos_in_self >= len(self)")
+            res += self[pos_in_self : e[].first_curly]
+
+            if e[]._is_escaped_brace():
+                if e[].field[Bool]:
+                    res += "}"
+                else:
+                    res += "{"
```
laszlokindrat, line 2023: style suggestion
```suggestion
                res += "}" if e[].field[Bool] "{"
```

stdlib/src/builtin/string.mojo
```diff
@@ -2100,3 +2174,169 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+
+    Does not contains any substitution values.
+
+    """
+
+    var first_curly: Int
+    """the index of an opening brace around a substitution field"""
+
+    var last_curly: Int
+    """the index of an closing brace around a substitution field"""
```
laszlokindrat, line 2199: nit:
```suggestion
    """The index of an opening brace around a substitution field."""

    var last_curly: Int
    """The index of an closing brace around a substitution field."""
```

stdlib/src/builtin/string.mojo
```diff
@@ -2100,3 +2174,169 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+
+
+@value
+struct _FormatCurlyEntry:
+    """
+    Internally used by the `format()` method.
+
+    Specifically to structure fields.
+
+    Does not contains any substitution values.
```
laszlokindrat, line 2191: nit:
```suggestion
    Does not contain any substitution values.
```

stdlib/src/builtin/string.mojo
```diff
@@ -2100,3 +2174,169 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+    var field: Variant[
+        String,  # kwargs indexing (`{field_name}`)
+        Int,  # args manual indexing (`{3}`)
+        NoneType,  # args automatic indexing (`{}`)
+        Bool,  # for escaped curlies ('{{')
+    ]
+    """Store the substitution field."""
+
+    fn _is_escaped_brace(ref [_]self) -> Bool:
+        return self.field.isa[Bool]()
```
laszlokindrat, line 2210: nit: here and below: since the struct itself is already hidden, there is no need to make its methods hidden too, especially since they are used outside the struct itself.
```suggestion
    fn is_escaped_brace(ref [_]self) -> Bool:
        return self.field.isa[Bool]()
```

stdlib/src/builtin/string.mojo
```diff
@@ -2100,3 +2174,169 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+
+        - `String` for **kwargs indexing
+            (value field contains `foo`)
+
+        - `Bool` for escaped curlies
+            (value field contains False for `{` or True for '}')
+        """
+        var manual_indexing_count = 0
+        var automatic_indexing_count = 0
+        var kwargs_indexing_count = 0
```
laszlokindrat, line 2250: What is the purpose of this? It's incremented, but doesn't seem to be used.

stdlib/src/builtin/string.mojo
```diff
@@ -2100,3 +2174,169 @@ fn _calc_format_buffer_size[type: DType]() -> Int:
...
+                            entries.append(curren_entry)
+                            skip_next = True
+                            continue
+                    # if it is not an escaped one, it is an error
+                    raise (
+                        "there is a single curly } left unclosed or unescaped"
+                    )
+
+        if raised_automatic_index:
+            raise "Automatic indexing require more args in *args"
```
laszlokindrat, line 2330: non-blocking: I'm not sure I understand why you moved this here. The architecture as I see it should be is the following: `_FormatCurlyEntry` is a helper class that takes _only_ the template string, parses it, and returns a data structure that `String.format` can easily use to perform its own checks and the actual substitution. So `_FormatCurlyEntry.create_entries` raises only if there is something fundamentally wrong with the template string (e.g. both manual and automatic indexing is used, there are inbalanced braces, etc). The other errors are the responsibility of `String.format`.

This is important, because it would be good to retain this separation of concern as we build out other features, for example `"{foo:0.5}".format(435.563244535)`. WDYT?

laszlokindrat (approved):
Awesome, thank you! I have a few more nits and a non-blocking question, but that's it. Please add a changelog entry!

laszlokindrat:
Also please update the PR description to make sure its up to date with what the patch became.

stdlib/src/builtin/string.mojo
```diff
@@ -24,6 +24,7 @@ from memory import DTypePointer, LegacyPointer, UnsafePointer, memcmp, memcpy
 
 from utils import StringRef, StaticIntTuple, Span, StringSlice
 from utils._format import Formattable, Formatter, ToFormatter
+from collections.dict import OwnedKwargsDict
```
laszlokindrat, line 27: not needed yet
```suggestion
```

stdlib/src/builtin/string.mojo
```diff
@@ -1970,6 +1971,79 @@ struct String(
...
+
+        Parameters:
+            Ts: The types of the substitution values.
+              Are required to implement `Stringable`.
+
+        Returns:
+            The template with the given values substituted.
+
+        """
+        alias NUM_POS_ARGS = len(VariadicList(Ts))
```
laszlokindrat, line 2008: style: this is not a global alias, so please use `snake_case`
```suggestion
        alias num_pos_args = len(VariadicList(Ts))
```

stdlib/src/builtin/string.mojo
rd4com: Yes, that makes sense, sorry moved too much there!
If we return only a `List[_FormatCurlyEntry]`, we would need to iterate it two times in `format`.
We'll probably need to return the counts with the list, in order to raise errors before starting.



`format` need to raise for `count_*>=len(args)`  and not `create_entries`?

:+1: All for re-use and separations

stdlib/src/builtin/string.mojo
rd4com: @laszlokindrat are we good with the tests ?

stdlib/src/builtin/string.mojo
rd4com: @laszlokindrat should move the `raises` in this patch ?

laszlokindrat: How about in a followup? I'm eager to land this, and a followup with this changes should be straightforward to review.

stdlib/src/builtin/string.mojo
laszlokindrat: Ok please remove `kwargs_indexing_count`.

rd4com:
@laszlokindrat 
We won't be able to use: `res += "}" if e[].field[Bool] "{"`
*(error: expecting an 'else' followed by an expression)*

Changes:
- used `^` for `entries.append(curren_entry^)`

laszlokindrat:
!sync

laszlokindrat:
!sync

laszlokindrat:
!sync

PR #2774, opened 2024-05-21
[docs] Add fix-ups to docs/changelog.md

Brian-M-J:
This PR adds a few typo/grammar fixes to `docs/changelog.md`:

- Changed "the sequence must be a parameter value and the induction variables are also parameter values in that sequence" to "the sequence and the induction values in the sequence must be parameter values"
- Changed "the `inferred` passing kind on parameters" to "the `inferred` parameter convention" (which is more consistent with the term "argument convention" already used by Mojo"
- Changed "specified by the user. This allows users" to "specified by the user. This allows programmers" ("user" was referring to both writers and users of functions)
- Changed "to `DType.int32`" to "to be `DType.int32`"
- Changed "treated as `owned` convention" to "treated according to the `owned` convention"
- Changed 'This is "worked"' to 'This "worked"'
- Changed "mutated on the body of the function" to "mutated in the body of the function"
- Changed "Added new `as_bytes_slice()` methods to `String` and `StringLiteral`" to "Added a new `as_bytes_slice()` method to `String` and `StringLiteral`"
- Changed "functionalty" to "functionality"
- Changed "method which allow" to "method which allows"
- Changed "`ListLiteral` and `Tuple` now only requires" to "`ListLiteral` and `Tuple` now only require"

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -19,9 +19,8 @@ what we publish.
 ### ⭐️ New
 
 - Mojo has introduced `@parameter for`, a new feature for compile-time
-  programming. `@parameter for` defines a for loop where the sequence must be a
-  parameter value and the induction variables are also parameter values in that
-  sequence. For example:
+  programming. `@parameter for` defines a for loop where the sequence and the
+  induction values in the sequence must be parameter values. For example:
 
   ```mojo
   fn parameter_for[max: Int]():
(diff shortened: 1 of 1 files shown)


JoeLoser (approved):
Thank you!

JoeLoser:
!sync

PR #2777, opened 2024-05-21
[stdlib] Implement gcd for two or more values

bgreni:
The `gcd` implementation in Python accepts a variadic list of integers. The existing mojo implementation lives in the closed source part of the stdlib and I see the comment mentioning that the open source module cannot depend on the closed source one so I've written a simple one here.

Since mojo does not support unpacking a `List` into a `VariadicList` I've also included a couple overloads to support the `List` and `Span` types as well.

The existing `gcd` implementation also does not have matching behaviour when negative integers are used.

### DIFF
--- a/stdlib/test/builtin/test_math.mojo
+++ b/stdlib/test/builtin/test_math.mojo
@@ -84,6 +84,26 @@ def test_pow():
     assert_equal(pow(I(0, 1, 2, 3), int(2)), I(0, 1, 4, 9))
 
 
+def test_gcd():
+    var l = List(2, 4, 6, 8, 16)
+    var il = InlineArray[Int, 5](4, 16, 2, 8, 6)
+    assert_equal(gcd(Span(il)), 2)
+    assert_equal(gcd(2, 4, 6, 8, 16), 2)
+    assert_equal(gcd(l), 2)
+    assert_equal(gcd(88, 24), 8)
+    assert_equal(gcd(0, 0), 0)
+    assert_equal(gcd(1, 0), 1)
+    assert_equal(gcd(-2, 4), 2)
+    assert_equal(gcd(-2, -4), 2)
+    assert_equal(gcd(24826148, 45296490), 526)
+    assert_equal(gcd(0, 9), 9)
+    assert_equal(gcd(4, 4), 4)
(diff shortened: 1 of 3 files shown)


stdlib/src/builtin/math.mojo
```diff
@@ -250,3 +250,72 @@ fn round[T: Roundable](value: T) -> T:
...
     """
     return value.__round__()
+
+
+# ===----------------------------------------------------------------------=== #
+# gcd
+# ===----------------------------------------------------------------------=== #
+
+
+fn gcd(owned m: Int, owned n: Int) -> Int:
```
soraros, line 260: No need to specify `owned` for reg type like `Int`.

stdlib/src/builtin/math.mojo
```diff
@@ -250,3 +250,72 @@ fn round[T: Roundable](value: T) -> T:
...
+        m: The first integer.
+        n: The second integrer.
+
+    Returns:
+        The greatest common divisor of the two integers.
+    """
+    while n != 0:
+        var t = m % n
+        m = n
+        n = t
```
soraros, line 273:
```suggestion
    while n:
        m, n = n, m % n
```

laszlokindrat (approved):
Very nice, thanks! Could you please move this to `builtin/_math.mojo`? I can make sure internally that we delegate to this.

Maybe `lcm` for followup? :)

bgreni:
> Maybe `lcm` for followup? :)

Certainly!

stdlib/test/builtin/test_math.mojo
```diff
@@ -84,10 +84,31 @@ def test_pow():
...
+
+
 def main():
     test_abs()
     test_divmod()
     test_max()
     test_min()
     test_round()
-    test_pow()
+    # test_pow()
```
laszlokindrat, line 113: Why disable this?

laszlokindrat:
!sync

stdlib/test/builtin/test_math.mojo
bgreni: Oh sorry a leftover from my rebasing, fixed now!

laszlokindrat:
!sync

Issue #2773, opened 2024-05-21
[Feature Request] [stdlib] [proposal] Rename `InlineList` to `Array`


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

InlinedFixedVector might get deprecated #2613

What is necessary:
  - [ ] rename `/stdlib/src/collections/inline_list.mojo` to  `/stdlib/src/collections/array.mojo`
  - [ ] rename `struct InlineList ...` to  `struct Array ...` ~~and  `struct InlinedFixedVector` to `struct Vector`~~
  - [ ] rename all mention in file docstrings
  - [ ] rename all mention in docs

### What is your motivation for this change?

Array is a name and structure that is known by many programmers from other languages, simpler to understand than saying inlinelist

This way we will have 2 easy to teach list-like types:
- `Array` An Array allocated on the stack with a maximum size known at compile time.
- `List` A dynamically-allocated homogeneous List.
~~- `Vector` A dynamically-allocated vector with small-vector optimization and a fixed
    maximum capacity.~~

gryznar:
+1 for such simplification and consistency :)

JoeLoser:
I don't think we should rename `InlineList` to `Array`.  Have you seen the original conversation that added the type at https://github.com/modularml/mojo/pull/2294#discussion_r1571549530?

martinvuyk:
I hadn't seen that specific comment. This is literally the first time I hear that Python has an array type (might have heard once and forgot) and I've been using Python since I'm a kid... people mostly use a list and forget about it (we all know Python isn't for performance), those who care about performance might have code that uses python's array (who? most will have migrated no numpy/use pandas), but most likely in a very easy way to migrate by just using Mojo's List constructor.

I think devs from the following languages might come into Mojo:
- Go: Array as fixed size
- Rust: Array as fixed size
- Zig: Array as fixed size
- Javascript: Array as heterogeneous List
- PHP: Array as heterogeneous List
- Ruby: Array as heterogeneous List
- Java: Array as fixed size
- Python: Array as homogeneous List
- C: Array as fixed size
- C++: Array as fixed size
- Swift: Array as homogeneous List
- C#: Array as fixed size
- Ocaml: Array as fixed size
- Perl: Array as fixed size
- SQL: (DBs that support it) Array as fixed size

[...]

PR #2780, opened 2024-05-22
[stdlib] Copy and rename impl and tests for `inline_list.mojo` to `array.mojo`

martinvuyk:
First stage to renaming `InlineList` to `Array` as per issue #2773


JoeLoser (changes requested):
As I mentioned at https://github.com/modularml/mojo/issues/2773#issuecomment-2125199600, I don't think we want to rename `InlineArray` to `Array` at this time.

PR #2781, opened 2024-05-22
[stdlib] Rename `InlinedFixedVector` to `Vector`

martinvuyk:
First stage to renaming `InlinedFixedVector` to `Vector` as per issue #2773

### DIFF
--- a/stdlib/src/collections/__init__.mojo
+++ b/stdlib/src/collections/__init__.mojo
@@ -17,7 +17,4 @@ from .inline_list import InlineList
 from .list import List
 from .optional import Optional, OptionalReg
 from .set import Set
-from .vector import (
-    CollectionElement,
-    InlinedFixedVector,
-)
+from .vector import CollectionElement, Vector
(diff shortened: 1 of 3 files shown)


helehex:
i would like if vector would eventually refer to a safe wrapper around dtype pointer

helehex:
I believe theres a chance DTypePointer will get deprecated eventualy but im not sure

martinvuyk:
> i would like if vector would eventually refer to a safe wrapper around dtype pointer, but this is probably good, at least for now

I'm not sure if that's even possible since in theory you could try to alloc something larger than your available heap

JoeLoser (changes requested):
I'm not a big fan of this change right now.  I'd rather us push `CollectionElement` through to this collection before we name it something more generic.  The bad name shys people away from this type intentionally right now still.

Also, do we even need this type once our `List` type supports small buffer optimization?  See https://github.com/modularml/mojo/pull/2613 from @gabrieldemarmiesse.

gabrieldemarmiesse:
Yep, I don't see any use cases of InlinedFixedVector in the future, if my proposal for small buffer optimisation get through, that is. I'm unsure of the general feelings of the maintainers on the subject.

martinvuyk:
Hadn't seen that PR, makes sense.

PR #2782, opened 2024-05-22
[stdlib] Fix Tuple.__contains__

rd4com:
Hello,

not sure why latest commit on https://github.com/modularml/mojo/pull/2709 were not used for ```nightly 2024.5.2119```,

they contained more tests and were passing on mac aswel.

*(probably because all the commits have the same name, i'll make sure to name them properly!)*

Maybe it is because they were pushed after the ```merged internally``` label?

Anyway, here is the latests commits pushed on https://github.com/modularml/mojo/pull/2709

It should be faster, safer, and taking into accounts all the reviews !

*(ping to @laszlokindrat)*


rd4com:
:partying_face:  In addition,

https://github.com/modularml/mojo/commit/53bfb7ca169c51302f7b6dfcf67156a355823eb6 makes it possible to return from a ```@parameter for```.

We can now return True the first time that `self[i] == value`,
instead of continuing doing `result |=` for every `_type_is_eq[element_types[i],T]()`.
It also remove `var result:Bool = False`.

stdlib/test/builtin/test_tuple.mojo
```diff
@@ -56,13 +56,12 @@ def test_tuple_contains():
...
     assert_true(String("Mojo is awesome") in d)
+    assert_false("Mojo is awesome" in d)
     assert_true(d.__contains__(String("Mojo is awesome")))
 
     assert_false(String("Hello world") in d)
     assert_false(d.__contains__(String("Hello world")))
 
 
 def main():
-    # FIXME(MSTDL-516)
```
JoeLoser, line 66: Don't fully enable this on Mac quite yet. There's a double free we need to chase down in one of these tests with yesterday's nightly.

JoeLoser (commented):
Can we please also add some tests to make sure this works in the compile time domain, eg contains works when used with a Tuple that is an alias (compile time variable)? An internal user, @abduld, just noticed it didn't work on top of tree in the compile time domain.

rd4com:
Yes, good idea,  here is an addition of the same tests with alias :+1:

stdlib/test/builtin/test_tuple.mojo
```diff
@@ -56,11 +56,57 @@ def test_tuple_contains():
...
+    alias c_alias = (1, 0)
+    assert_false(c_alias.__contains__(True))
+    assert_false(c_alias.__contains__(False))
+    assert_false(True in c_alias)
+    assert_false(False in c_alias)
+
+    alias d_alias = (123, True, String("Mojo is awesome"))
+
+    assert_true(String("Mojo is awesome") in d_alias)
+    assert_false("Mojo is awesome" in d_alias)
```
JoeLoser, line 104: This will surprise users until `StringLiteral` materializes to a `String`.  When this happens @laszlokindrat, would this sort of gotcha go away, or no still since this is comp-time domain?

laszlokindrat: This is not compile time domain, so this will go away when `StringLiteral` becomes NM. Also, eventually we will have the language features so that we can have types that are comparable to other types parametrically.

laszlokindrat:
@rd4com Yeah, sorry about the mixup. I probably brought it in before your later patches, and then never synced again. Our processes will improve, but right now this unfortunately is a manual step (and I also have to make internal changes as well sometimes, for our internal code, so mistakes can happen).

stdlib/test/builtin/test_tuple.mojo
JoeLoser: Oh fair point, `assert_false` is already runtime domain despite `d_alias` being comp-time.

rd4com: It should be possible to add more tests, asserting the aliases with:
```mojo
@parameter
if "ok" in a_alias: ...
else: raise("assertion failed")
```
Should i create them ?

JoeLoser: I wouldn't bother right now; I just landed this guy internally.  WDYT @laszlokindrat @abduld?

rd4com:
@laszlokindrat no worries, i pushed commits poorly named, my bad;
Happy about this ```tuple.__contains__``` effort, hope people enjoy it :tada:

stdlib/test/builtin/test_tuple.mojo
JoeLoser: No worries, your fix actually fixed the double free issue we were seeing. @Dan13llljws has a PR internally to re-enable this on mac.  Thanks!

PR #2789, opened 2024-05-22
[stdlib] Cleanup `String` constructors to reuse null terminator checks

gabrieldemarmiesse:
We would like to reuse as much as possible the main constructor (from List[UInt8]) which has more checks. And it will also simplify the SSO work as we centralize the entrypoints of the String creation. 

A few simple cleanups too:
* Do not add the null terminator if the list  storage was already initialized to 0
* Remove unnecessary bitcasts now that we always work with UInt8

All the diffs here are independant from one another but I didn't have the courage to do 5 PRs

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -693,14 +693,7 @@ struct String(
             impl[-1] == 0,
             "expected last element of String buffer to be null terminator",
         )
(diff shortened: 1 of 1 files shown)


JoeLoser (approved):

JoeLoser:
!sync

PR #2793, opened 2024-05-22
[stdlib] Add `String.isspace()` conformant with Python

martinvuyk:
For better compatibility with python


stdlib/src/builtin/string.mojo
```diff
@@ -1408,6 +1408,35 @@ struct String(
...
+            String("\n"),
+            String("\r"),
+            String("\v"),
+            String("\f"),
+            String("\x1c"),
+            String("\x1e"),
+            String("\x85"),
+            String(List[UInt8](0x20, 0x5C, 0x75, 0x32, 0x30, 0x32, 0x38, 0)),
+            String(List[UInt8](0x20, 0x5C, 0x75, 0x32, 0x30, 0x32, 0x39, 0)),
+        )
```
laszlokindrat, line 1438: Could this work? It would guarantee that the list is built at compile time. If it doesn't work, you could also try some collection of `StringLiteral`.
```suggestion
        alias spaces = List[String](
            String(" "),
            String("\t"),
            String("\n"),
            String("\r"),
            String("\v"),
            String("\f"),
            String("\x1c"),
            String("\x1e"),
            String("\x85"),
            String(List[UInt8](0x20, 0x5C, 0x75, 0x32, 0x30, 0x32, 0x38, 0)),
            String(List[UInt8](0x20, 0x5C, 0x75, 0x32, 0x30, 0x32, 0x39, 0)),
        )
        return self in spaces
```
Or better yet, if the alias could be pushed into a private alias on the struct, you could ensure in the tests that its contents don't accidentally change.

stdlib/test/builtin/test_string.mojo
```diff
@@ -702,6 +702,37 @@ fn test_isspace() raises:
     assert_false(isspace(ord(".")))
 
 
+fn test_isspace_unicode() raises:
```
laszlokindrat, line 705: Could you please unify these tests and the ones in `test_isspace`?

laszlokindrat (approved):
Looks good, thanks for improving this! I have a few requests, most importantly regarding compile time initalization of the list of space characters.

I also wanted to ask you for a followup: Could you do this for `StringRef`, and then just delegate to that? Then the top-level `isspace` function could be removed (or at least made private withing `string.mojo`) in favor of that.

stdlib/src/builtin/string.mojo
martinvuyk: > compile time initalization of the list of space characters

What I fear might happen is that making it an alias concatenates all the chars in the list (that's why I didn't do it at first)

image

I will try it and make a test for it to see if some of the bytes that compose a unicode separator get counted as being in the alias. I think we need to find of a better way to deal with different UTF standards...

> a private alias on the struct

That's also another topic that I wanted to talk about. Currently many other methods use String.WHITESPACE, in my `String.split()` PR I'm adding utf8 chars that are whitespace into that variable, but many others should also take unicode separators into account and they don't. If we change it to a `List[String]` and refactor those functions to iterate 4 bytes at a time, we will pay a performance penalty but we will be more conformant with Python. WDYT?

> Could you do this for StringRef, and then just delegate to that? 

I thought by looking at the code that we wanted to have most complex string operations to be on string.mojo (ord, atol, char, etc. are all there)

> the top-level isspace function could be removed

That function behaves like C's `isspace()` so maybe renaming it? It will be much faster than `String.isspace()`since it will be checking for less chars and only in a StringRef directly with `memcmp`

laszlokindrat: > > a private alias on the struct
> 
> That's also another topic that I wanted to talk about. Currently many other methods use String.WHITESPACE, in my `String.split()` PR I'm adding utf8 chars that are whitespace into that variable, but many others should also take unicode separators into account and they don't. If we change it to a `List[String]` and refactor those functions to iterate 4 bytes at a time, we will pay a performance penalty but we will be more conformant with Python. WDYT?

We can extend Python's API with either parameter or argument flags to control this behavior. For now, I wouldn't worry too much about performance and try to get to the functionality first.

> 
> > Could you do this for StringRef, and then just delegate to that?
> 
> I thought by looking at the code that we wanted to have most complex string operations to be on string.mojo (ord, atol, char, etc. are all there)

Fair point, and `StringRef` should also mostly go away in favor of `StringSlice. We would probably need a similar method on that, though.

> 
> > the top-level isspace function could be removed
> 
> That function behaves like C's `isspace()` so maybe renaming it? It will be much faster than `String.isspace()`since it will be checking for less chars and only in a StringRef directly with `memcmp`

I checked, `String.isspace` is actually just an implementation detail of `String` and `StringRef`, so we should just make it private.

stdlib/src/builtin/string.mojo
```diff
@@ -572,11 +572,11 @@ fn _is_ascii_lowercase(c: UInt8) -> Bool:
 # ===----------------------------------------------------------------------===#
 
 
-fn isspace(c: UInt8) -> Bool:
+fn _isspace(c: UInt8) -> Bool:
```
laszlokindrat, line 575: Can you please mention in the changelog that this is no longer public and `String.isspace` is preferred instead?

stdlib/src/builtin/string.mojo
```diff
@@ -585,11 +585,15 @@ fn isspace(c: UInt8) -> Bool:
...
-    return c == ord_space or ord_tab <= int(c) <= ord_carriage_return
+    alias c_spaces = (
+        ord(" "),
+        ord("\t"),
+        ord("\n"),
+        ord("\r"),
+        ord("\f"),
+        ord("\v"),
+    )
+    return int(c) in c_spaces
```
laszlokindrat, line 596: If possible, we should just do this with a 256 element jump table. A hidden top-level alias could do this.

stdlib/src/builtin/string.mojo
```diff
@@ -1406,6 +1410,35 @@ struct String(
...
+
+        Returns:
+            True if the String is one of the whitespace characters
+                listed above, otherwise False."""
+        # TODO add line and paragraph separator as stringliteral
+        # once unicode escape secuences are accepted
+        # Unicode Line Separator: \ u2028
+        # Unicode Paragraph Separator: \ u2029
+        # 0 is to build a String with null terminator
+        return self in List[String](
```
laszlokindrat, line 1428: The perf implications of this still bother me. The cpython implementation checks for the length first, and dispatches a lookup table based solution for single character strings. I think this is definitely something we should do (see my other comment).

Also, the cpython implementation actually returns true if _all_ elements in a string are spaces, i.e. not just for single element strings. This was a surprise to me, but we should follow it for now, and we can think about different APIs later.

laszlokindrat (commented):
Thanks for working on this and addressing my reviews, I really appreciate it! I did a bit of reading on the python implementation of this, please see my comments and let me know what you think!

stdlib/src/builtin/string.mojo
martinvuyk: > do this with a 256 element jump table. A hidden top-level alias could do this

Do you have any tips on how to approach that? all examples I've found are with switch in C, or talk about logic truth tables (or FPGAs which have nothing to do here)...

Won't it be the same performance by doing `Tuple[UInt8](...).__contains__` ?

And another question, standalone this works fine but in stringref when using the `UnsafePointer[SIMD[uint8, 1], 0]` it doesn't (tests fail). Is there some reason for that? will this approach even make a performance difference?
```mojo
fn main():
    alias spaces = SIMD[DType.uint8, 8](
        ord(" "),
        ord("\t"),
        ord("\n"),
        ord("\r"),
        ord("\f"),
        ord("\v"),
        ord("\v"),
        ord("\v"),
    )
    var item1: UInt8 = ord(" ")
    print((spaces & item1).cast[DType.bool]().reduce_or())
    var item2 = UnsafePointer[SIMD[DType.uint8, 1], 0](UInt8(ord(" ")))
    print((item2[0] & spaces).cast[DType.bool]().reduce_or())
```

laszlokindrat: Okay, I think what we need is a module level variable, not an alias. I think something like this might work:
```mojo
alias _SpacesTableType = InlineArray[UInt8, 256]

fn _get_spaces_table() -> _SpacesTableType:
    var table = _SpacesTableType(0)
    ...  # manually set locations associated with space types to 1, e.g. `table[ord(" ")] = 1` (no need to loop over these, just write them out)
    return table

var _SPACES_TABLE = _get_spaces_table()

fn isspace(char: UInt8) -> Bool:
   return _SPACES_TABLE[char]

```
This is an aggressive tradeoff: takes 256 bytes of memory but the lookup is very fast. I don't know if we can align this table in memory, but maybe we should.

There is a chance this won't work at all (because globals are not really baked out in mojo), but it's worth pursuing.

martinvuyk: I'm not sure if I like using so much memory without explicit command from the developer, is there some way we can make this opt in? For micro-controller development 256 bytes counts for a lot. And also other methods like `splitlines()` will need to define their own character combinations anyway, since you have stuff like windows `\r\n`.

laszlokindrat: We are very far from being able to run in any memory constrained environment and Python itself employs this jump table optimization for performance. When we start looking at MCUs and other embedded system, we will be able to specialize the function for different targets. For the systems we run on today, however, this optimization is needed for performance; looping through collections of characters/strings and looking for spaces is common, and it has to be fast.

martinvuyk: > probably you can just compare the underlying List

ok done, used memcmp

What do you think of this approach? it works, I was doing the logic wrong earlier.
```mojo
fn isspace(c: UInt8) -> Bool:
    alias spaces = SIMD[DType.uint8, 8](
        ord(" "),
        ord("\t"),
        ord("\n"),
        ord("\r"),
        ord("\f"),
        ord("\v"),
        ord("\v"),
        ord("\v"),
    )
    return (spaces ^ c).reduce_min() == 0
```

> We might have to employ similar optimizations there. Let's cross that bridge when we get there.

we might just use the same LUT and do some hashing to get bigger sequences in there (?). It would be kinda cheating but for example:

\x85 or 120 56 53 or 0x78 0x38 0x35 is NEXT LINE (NEL) character
\x1c or 92 120 49 99 or 0x5c 0x78 0x31 0x63 is INFORMATION SEPARATOR FOUR character
\x1e or 92 120 49 101 or 0x5c 0x78 0x31 0x65 is INFORMATION SEPARATOR TWO character

we could just grab the last byte of each and `|` it with a `0b10000000` (though other multi byte sequences might end up getting returned as true)

Also I've been thinking of how will we deal in other algos with the fact that some bytes might have `11xxxx...` as their first 2 bits and we should look for the nex byte while `10xxx...` is their first 2 bits. Or if we are already reading the first byte then we could just directly do
```mojo
var num_bytes = 1
if b >> 7 != 0:
    num_bytes = (bit_reverse(b) & 0b1111)
```
and we get the number of bytes to grab next. I'm worried for the performance either way.

stdlib/src/builtin/string.mojo
```diff
@@ -574,22 +574,28 @@ fn _is_ascii_lowercase(c: UInt8) -> Bool:
 
 fn isspace(c: UInt8) -> Bool:
```
laszlokindrat, line 575: I thought you already made this hidden, maybe something went wrong with a merge/rebase?

stdlib/src/builtin/string.mojo
laszlokindrat: Can you please just implement the LUT for length 1? It is a known fast implementation with little implementation complexity. Right now, we do not care about 256 bytes, and simd reduction may or may not be fast depending on the machine, the reduce implementation, and the compiler optimizations we can do. Once we have that, we can start a separate discussion on how to optimize these with different tradeoffs on different platforms, but only if needed.

stdlib/src/builtin/string.mojo
martinvuyk: Since it is only for a byte I thought of leaving it as is, tried putting it into String as a static method but the tests started failing.

stdlib/src/builtin/string.mojo
```diff
@@ -1406,6 +1412,73 @@ struct String(
             substr._strref_dangerous(), start=start
         )
 
+    fn isspace[str_len: Int = -1](self) -> Bool:
```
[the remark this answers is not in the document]
martinvuyk: The default is detemining it at runtime, I wanted to leave and overload for when we start checking for longer utf8 sequences for other methods like `splitlines()` and `split()` WDYT?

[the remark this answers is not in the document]
laszlokindrat: As I said, we can think about optimizing this later, but in this patch this is premature. It would be better to have small patches, that do fewer things; makes it much easier to review and land them internally.

martinvuyk:
@laszlokindrat  added everything, pipeline keeps failing because of other stuff unrelated to this PR

laszlokindrat:
!sync

laszlokindrat (approved):
This is great, thanks for bearing with me! As the first followup, I would request that we generalize this so it works with arbitrary length strings (e.g. python returns `True` for this: `"            ".isspace()`).

I will fix the linting failure internally, it's probably something about the changelog.

PR #2701, opened 2024-05-17
[stdlib] Support Dict popitem

jayzhan211:
close https://github.com/modularml/mojo/issues/2355
close #2542

PR #2794, opened 2024-05-22
[stdlib] Implement `Dict.popitem()` method

msaelices:
- [X] Implement `Dict.popitem()` method
- [X] Added changelog entry


JoeLoser:
Have you seen https://github.com/modularml/mojo/pull/2701 by @jayzhan211?

msaelices:
Closed in favor of: https://github.com/modularml/mojo/pull/2701

JoeLoser:
Since `Dict.pop` got reverted in https://github.com/modularml/mojo/commit/b9453e94bc6911c10bc55b429dda3c109f30682c, want to re-open this and we can see how it does with regard to flakiness?

msaelices:
> Since `Dict.popitem` got reverted in [b9453e9](https://github.com/modularml/mojo/commit/b9453e94bc6911c10bc55b429dda3c109f30682c), want to re-open this and we can see how it does with regard to flakiness?
> 
> I'd recommend adding a changelog entry and updating the raise to be more descriptive like `raise "KeyError: popitem(): dictionary is empty"` from https://github.com/modularml/mojo/pull/2701/files#diff-861767ee9c0157172e9901c202431c23b1609089ab926bdf524ee3e722b209bfR828.

Re-opened and synced from `nightly`. Also improved the raise sentence.

docs/changelog.md
```diff
@@ -18,6 +18,10 @@ what we publish.
 
 ### ⭐️ New
 
+- `Dict` now support `popitem`, which removes and returns the last item in the `Dict`.
```
JoeLoser, line 21:
```suggestion
- `Dict` now supports `popitem`, which removes and returns the last item in the `Dict`.  The item is returned as a (key, value) pair.
```

JoeLoser (approved):
Looks good — just one minor wording suggestion on the changelog entry, and then I'm happy to import and merge this.

stdlib/src/collections/dict.mojo
```diff
@@ -799,6 +799,23 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
+
+        Returns:
+            A tuple containing the key and value of the removed item.
+
+        Raises:
+            "KeyError" if the dictionary is empty.
+        """
+        if self.size == 0:
+            raise "KeyError: popitem(): dictionary is empty"
+        var entry = self._entries.pop(0)
```
jayzhan211, line 813: Shouldn't we follow the default behaviour like python in LIFO order?

jayzhan211:
Can we try another approach that is not using `reverse(dict.item)`.
We can iterate in the forward pass, the get the last item too. If this kind of approach has no panic issue, then we can know the issue is because of `reverse(dict)` but not `pop`.

What do you think @JoeLoser ?

JoeLoser:
> Can we try another approach that is not using `reverse(dict.item)`. We can iterate in the forward pass, the get the last item. If this kind of approach has no panic issue, then we can know the issue is because of `reverse(dict)` but not `pop`. And, we can also have the default behaviour that is consistent as python's FILO
> 
> What do you think @JoeLoser @msaelices ?

SGTM!

stdlib/src/collections/dict.mojo
msaelices: You are completely right. Fixed here: https://github.com/msaelices/mojo/commit/a034f7bf71fbf9fd99026d67d0093d34da55d2ae

msaelices:
> Can we try another approach that is not using `reverse(dict.item)`. We can iterate in the forward pass, the get the last item. If this kind of approach has no panic issue, then we can know the issue is because of `reverse(dict)` but not `pop`. And, we can also have the default behaviour that is consistent as python's FILO
> 
> What do you think @JoeLoser @msaelices ?

Why do we need to iterate? To get the FILO behaviour we can just do `self.entries.pop(self.size - 1)`

stdlib/src/collections/dict.mojo
```diff
@@ -799,6 +799,23 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
+
+        Returns:
+            A tuple containing the key and value of the removed item.
+
+        Raises:
+            "KeyError" if the dictionary is empty.
+        """
+        if self.size == 0:
+            raise "KeyError: popitem(): dictionary is empty"
+        var entry = self._entries.pop(self.size - 1)
```
jayzhan211, line 813:
```mojo
    fn _insert(inout self, owned entry: DictEntry[K, V]):
        self._maybe_resize()
        var found: Bool
        var slot: Int
        var index: Int
        found, slot, index = self._find_index(entry.hash, entry.key)

        self._entries[index] = entry^
        if not found:
            self._set_index(slot, index)
            self.size += 1
            self._n_entries += 1
```

I think `index` in the _entries is computed by hash function, which mean the element index is not ordered. You pop the item based on the index of `self.size - 1`, which I think it is not always the `last item` we expect.

You can try to pop and insert the dict to mix the dictionary, I think it is possible to get the unexpected result.

jayzhan211: iterator including reversed iterator both check if the index in entries exist item or not. 

> var opt_entry_ref = self.src[]._entries.__get_ref(self.index

```mojo
    @always_inline
        while True:

            @parameter
            if forward:
                debug_assert(
                    self.index < self.src[]._reserved, "dict iter bounds"
                )
            else:
                debug_assert(self.index >= 0, "dict iter bounds")

            var opt_entry_ref = self.src[]._entries.__get_ref(self.index)
            if opt_entry_ref[]:

                @parameter
                if forward:
                    self.index += 1
                else:
                    self.index -= 1

                self.seen += 1
                return opt_entry_ref[].value()[]

            @parameter
            if forward:
                self.index += 1
            else:
                self.index -= 1
```

msaelices: I'm not sure of it. Please review the `_DictEntryIter`, I'm deleted non-relevant fragments for this question:
```
@value
struct _DictEntryIter[
    ...  # omitted
    forward: Bool = True,
]:
    """Iterator over immutable DictEntry references.

    Parameters:
        ...  # omitted
    """
    ... # omitted
    var index: Int
    var src: Reference[Dict[K, V], dict_mutability, dict_lifetime]

    @always_inline
        while True:
            ...  # omitted

            var opt_entry_ref = self.src[]._entries.__get_ref(self.index)
            if opt_entry_ref[]:

                @parameter
                if forward:
                    self.index += 1
                else:
                    self.index -= 1

                return opt_entry_ref[].value()[]

            @parameter
            if forward:
                self.index += 1
            else:
                self.index -= 1

       ... # omitted
```

When we call `self.items()` in the `Dict`, a `_DictEntryIter` obj is returned. If you call `reversed(self.items())`, the same obj is created with the `forward` param set to `False`

So, as you can see in the code above, the index is an integer that is being incremented or decremented while looping over the `_entries`, calling to the `dict._entries.__get_ref(index)`.

This code calls the `self._entries.pop(self.size - 1)` being `self._entries` a `List` object, which works with integers indexes. 

So, I think that the code is right. I am not 100% sure as I'm just landing on the Mojo's world. I will complete the current tests to make sure it works as expected

msaelices: Forget about it. I think you are right. I will refactor the code.

msaelices: @jayzhan211 fixed here: https://github.com/msaelices/mojo/commit/c3112cd4939a46d6383177a148e5c1445ee47fd1

In the end, I finished implementing a very similar approach to yours, as the tests raised fatal errors if we added a new key after calling to `popitem()`. I think it's because what you said, the `_entries` list is bigger than the amount of entries with holes on it when we remove items from the dict. 

Now, with the current implementation, I'm not sure if we will find the same issue as in your PR code.

msaelices: @jayzhan211 FYI: It seems @gabrieldemarmiesse has found the cause of the flakiness in your code which it was NOT caused by your code but a boundary bug in the `reversed()` logic. 

Now I guess we can re-introduce your code or also this one. Let's @JoeLoser decide.

msaelices: @jayzhan211  Just in case, I've added you to the changelog entry: https://github.com/msaelices/mojo/commit/bd0b912cbfc47eae69f0ca86cd63d7c40b326513

jayzhan211 (approved):
👍

stdlib/src/collections/dict.mojo
jayzhan211: > @jayzhan211 FYI: It seems @gabrieldemarmiesse has found the cause of the flakiness in your code which it was NOT caused by your code but a boundary bug in the `reversed()` logic. See #2896
> 
> Now I guess we can re-introduce your code or also this one. Let's @JoeLoser decide.

I don't mind, at least the code looks good to me

gabrieldemarmiesse:
@msaelices The PR looks good. But since https://github.com/modularml/mojo/pull/2896 was merged into nightly and fixed the flakyness exposed by #2701 , the `popitem()` implementation of @jayzhan211 was added back to nightly. Maybe we can then close this pull request?

msaelices:
> @msaelices The PR looks good. But since #2896 was merged into nightly and fixed the flakyness exposed by #2701 , the `popitem()` implementation of @jayzhan211 was added back to nightly. Maybe we can then close this pull request?

Sure. Closed.

PR #2795, opened 2024-05-22
[tools] Simplify doc validation

JoeLoser:
Now that `mojo doc` command learned `--validate-doc-strings`, we don't need to have the wrapper script that inspects `stderr` that `mojo doc -diagnose-missing-doc-strings` previously used in the `check-docstrings.py`.

We'll update the docs in a separate PR since this `.pre-commit-config.yaml` isn't managed by Copybara.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
         stages: [commit]
       - id: check-docstrings
         name: check-docstrings
-        entry: python3 ./stdlib/scripts/check-docstrings.py
+        entry: mojo doc --validate-doc-strings stdlib/src
         language: system
         pass_filenames: false
         stages: [commit]


JoeLoser:
FYI @gabrieldemarmiesse

JoeLoser:
When I introduced a local doc string error and tried committing, here is the developer experience now:

```
it commit -am "bad docstring"
mojo-format..............................................................Passed
check-docstrings.........................................................Failed
- hook id: check-docstrings
- exit code: 1

Included from /Users/joe/dev/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/joe/dev/mojo/stdlib/src/collections/list.mojo:46:9: error: 'forward' is defined at index 3, but specified in doc string at index 2
        ^
Included from /Users/joe/dev/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/joe/dev/mojo/stdlib/src/collections/list.mojo:43:5: error: parameter 'list_lifetime' is not documented
    Parameters:
```

.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
         stages: [commit]
       - id: check-docstrings
         name: check-docstrings
-        entry: python3 ./stdlib/scripts/check-docstrings.py
+        entry: mojo doc --validate-doc-strings stdlib/src
```
gabrieldemarmiesse, line 12: Should we remove the python file in this PR or in another one? Is the `check-docstrings.py` managed by copybara?

gabrieldemarmiesse (approved):
Looks good! It's much simpler. I'm just wondering what we should do with the wrapper script.

.pre-commit-config.yaml
JoeLoser: It's managed by copybara so I plan on removing it in a separate PR and also update the contributor docs. Does that work for you?

gabrieldemarmiesse: Yep

River707 (approved):

Issue #2756, opened 2024-05-20
[BUG] Popping a String from a dictionary then using the value leads to a double free crash.


thatstoasty:
### Bug description

Popping a String from a dictionary, then using the value leads to a double free crash.

### Steps to reproduce

* Include relevant code snippet or link to code that did not work as expected.
* If applicable, add screenshots to help explain the problem.
* If using the Playground, name the pre-existing notebook that failed and the steps that led to failure.
* Include anything else that might help us debug the issue.

Minimal reproducible example:

```mojo
fn main() raises:
    var context = Dict[String, String]()
    context["a"] = "b"
 
    print(context.pop("a"))
```

```
mojo(76499,0x1eb727ac0) malloc: Double free of object 0x13a7283b0
mojo(76499,0x1eb727ac0) malloc: *** set a breakpoint in malloc_error_break to debug
Please submit a bug report to https://github.com/modularml/mojo/issues and include the crash backtrace along with all the relevant source codes.
Stack dump:
0. Program arguments: mojo run test.mojo
Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0 mojo 0x0000000104513db4 llvm_strlcpy + 51480
1 mojo 0x00000001045120a0 llvm_strlcpy + 44036
2 mojo 0x00000001045144a0 llvm_strlcpy + 53252
3 libsystem_platform.dylib 0x0000000183a4b584 _sigtramp + 56
4 libsystem_pthread.dylib 0x0000000183a1ac20 pthread_kill + 288
5 libsystem_c.dylib 0x0000000183927a20 abort + 180
6 libsystem_malloc.dylib 0x0000000183837aa8 malloc_vreport + 896
7 libsystem_malloc.dylib 0x000000018385bea8 malloc_zone_error + 104
8 libsystem_malloc.dylib 0x000000018385ad00 free_tiny_botch + 40
9 libKGENCompilerRTShared.dylib 0x000000010747ba7c KGEN_CompilerRT_AlignedFree + 156
10 libKGENCompilerRTShared.dylib 0x0000000300004a58 KGEN_CompilerRT_AlignedFree + 8467812472
11 mojo 0x00000001048a2394 __jit_debug_register_code + 1013256
12 mojo 0x000000010447436c
13 mojo 0x000000010445cc6c
14 dyld 0x00000001836920e0 start + 2360
mojo crashed!
Please file a bug report.
[...]

PR #2796, opened 2024-05-22
[stdlib] mark dict entry as destroyed in `Dict.pop()`

helehex:
see [#2756](https://github.com/modularml/mojo/issues/2756)
this is an attempt to fix errors that sometimes arise when using the pop method. I'm not sure this is the best way, so feedback is much appreciated

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -216,6 +216,15 @@ struct DictEntry[K: KeyElement, V: CollectionElement](
         self.key = other.key
         self.value = other.value
 
+    fn reap_value(owned self) -> V:
+        """Take the value from an owned entry.
+
+        Returns:
+            The value of the entry.
+        """
+        __mlir_op.`lit.ownership.mark_destroyed`(__get_mvalue_as_litref(self))
+        return self.value^
+
 
 alias _EMPTY = -1
 alias _REMOVED = -2
@@ -813,7 +822,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
             var entry_value = entry[].unsafe_take()
             entry[] = None
             self.size -= 1
-            return entry_value.value^
+            return entry_value^.reap_value()
         raise "KeyError"
 
     fn popitem(inout self) raises -> DictEntry[K, V]:
--- a/stdlib/test/collections/test_dict.mojo
+++ b/stdlib/test/collections/test_dict.mojo
@@ -538,6 +538,21 @@ def test_dict_popitem():
         _ = dict.popitem()
 
 
+def test_pop_string_values():
+    var dict = Dict[String, String]()
+    dict["mojo"] = "lang"
+    dict["max"] = "engine"
+    dict["a"] = ""
+    dict[""] = "a"
+
+    assert_equal(dict.pop("mojo"), "lang")
+    assert_equal(dict.pop("max"), "engine")
+    assert_equal(dict.pop("a"), "")
+    assert_equal(dict.pop(""), "a")
+    with assert_raises(contains="KeyError"):
+        _ = dict.pop("absent")
+
+
 fn test_clear() raises:
     var some_dict = Dict[String, Int]()
     some_dict["key"] = 1
@@ -594,6 +609,7 @@ def main():
     test_owned_kwargs_dict()
     test_bool_conversion()
     test_find_get()
+    test_pop_string_values()
     test_clear()
     test_init_initial_capacity()
     test_dict_setdefault()


helehex:
ok, tests passed locally, but test_reversed is still failing, so this fix may not work

stdlib/src/collections/dict.mojo
```diff
@@ -740,7 +740,12 @@ struct Dict[K: KeyElement, V: CollectionElement](
             var entry_value = entry[].unsafe_take()
             entry[] = None
             self.size -= 1
-            return entry_value.value^
+            _ = entry_value.key^
+            var value = entry_value.value^
+            __mlir_op.`lit.ownership.mark_destroyed`(
```
JoeLoser, line 745: @Mogball @lattner is this the "blessed" way to do this right now?

helehex: i imagine it's not since the test_reversed still failed... maybe it's not related, but i was really hoping it was

helehex: it did fix a crash locally for me though. (the exmaple given in #2756)

gabrieldemarmiesse:
> ok, tests passed locally, but test_reversed is still failing, so this fix may not work

@helehex I made a PR recently fixing the flakyness in `test_reverse.mojo`. Can you rebase this PR on nightly and see if that works now?

rd4com:
@helehex, based on https://github.com/modularml/mojo/issues/2822 last comment,
and that @gabrieldemarmiesse  fixed the nasty bug,
it should now work !

helehex:
Anyone know why this cant go through? last time I checked, pop still lets garbage data through @JoeLoser

JoeLoser:
> Anyone know why this cant go through? last time I checked, `pop()` still returns garbage data. @JoeLoser

Is there still an issue with pop on latest nightly? Originally, we were seeing failures around the time you opened this PR, but I thought I remember @gabrieldemarmiesse chasing down the root cause and fixing it.

helehex:
I just checked, and the test provided here still fails in the latest nightly.
They were unrelated. If I remember correctly, the issue here is that the `entry.value` gets destroyed twice.

helehex:
Sorry, after reading through this thread again, I realize I should've communicated better what was going on.

helehex:
Here's a few notes:
- I moved the functionality into a helper method on `DictEntry`, after seeing Chris's response in #2822. I called it `reap_value()`, It makes sense conceptually to me with how it passes itself as owned.
- The fix Gabriel provided allowed the tests to pass in this branch, but was unrelated to the particular issue.
- What the issues basically comes down to, as far as I can tell right now, is a segfault when doing something like this:
  ```mojo
  var entry = DictEntry[String, String]("a", "b")
  print(entry.value^)
  ```

rd4com:
Yes, this PR would make it possible the move a value out of the dictionary.
(`V` of `DictEntry[K,V]` is not an `Optional`)

I think that what we return now is a weird copy that gets double freed.
(No problem when `V` is a trivial type like `Int`)

If we turn `V` into an `Optional`, 

we'd have an massive overhead of `the_dict._reserved()` * 8 bytes.
(If `V` is `Int`, this is huge, even `None` entries would have the overhead)

So this PR move the value out by scavaging an `DictEntry[K,V]` with the blessed thing.

### The problem that could be a bug (without PR)

- only fully initialized instances can have their destructor used

Currently, we move the value out, and `DictEntry` is destroyed anyway.

So if the user received an instance of a type that has `UnsafePointer` as fields,
The pointers are probably freed.

helehex:
!sync

stdlib/src/collections/dict.mojo
```diff
@@ -216,6 +216,15 @@ struct DictEntry[K: KeyElement, V: CollectionElement](
         self.key = other.key
         self.value = other.value
 
+    fn reap_value(owned self) -> V:
```
JoeLoser, line 219: Should we make this private? I don't expect users to need to interact with this API directly, right?

helehex: hmm I'm not sure actually.
```mojo
def main():
    print(str(take()))

def take() -> String:
    var entry = DictEntry[String, String]("a", "b")
    return entry.value^
```
Doing this still causes a segfault and the language server to crash.. do you know if this is a bug, or maybe it should give me an error in the editor?

helehex: As things are right now, it seems useful to have the function available and documented.

helehex: Adding ```__mlir_op.`lit.ownership.mark_destroyed`(__get_mvalue_as_litref(entry))``` after `entry` is the solution to it, but that's definitely a private function.

helehex: No, I don't think this should be private right now.

JoeLoser:
!sync

PR #2798, opened 2024-05-23
[Examples] Fix plt figure dimensions in RayTracing example notebook

SynodicMonth:
The figsize in plt.figure should be in (width, height). 
It was inverted, making the fig look like this:
before

After swapping the dimension:
after


JoeLoser (approved):

JoeLoser:
!sync

PR #2800, opened 2024-05-23
[stdlib] Add method `unsafe_get` to `List`

gabrieldemarmiesse:
See https://github.com/modularml/mojo/pull/2677#discussion_r1605618673 for the background about this method.

We are currently missing methods to access elements in collections without any bounds checks or wraparound, for maximum performance. I suggest that we introduce `unsafe_get` and `unsafe_set` to our List-like collections. This is equivalent to 
* https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked
* https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked_mut
in Rust.

This should prove useful to makers of high performance libraries like Max :p 

We can then make `__getitem__` and `__setitem__` as safe as we want without impacting the power users.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -788,7 +788,43 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         if i < 0:
             normalized_idx += self[].size
 
-        return (self[].data + normalized_idx)[]
+        return self[].unsafe_get(normalized_idx)
+
+    @always_inline
+    fn unsafe_get[
+        IndexerType: Indexer,
+    ](self: Reference[Self, _, _], idx: IndexerType) -> Reference[
+        Self.T, self.is_mutable, self.lifetime
+    ]:
+        """Get a reference to an element of self without checking index bounds.
+
+        Users should opt for `__getitem__` instead of this method as it is unsafe.
+        If an index is out of bounds, this method will not abort, it will be considered
+        undefined behavior.
+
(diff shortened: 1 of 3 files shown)


stdlib/src/collections/list.mojo
```diff
@@ -788,7 +788,43 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         if i < 0:
             normalized_idx += self[].size
 
-        return (self[].data + normalized_idx)[]
+        return self[].unsafe_get(normalized_idx)
+
+    @always_inline
+    fn unsafe_get[
+        IndexerType: Indexer,
```
bgreni, line 795: Note that `Indexer` types can now be implicitly converted to `Int` so such methods were recently reverted to simply accepting `Int` args again https://github.com/modularml/mojo/commit/72c5a47872ffd8ca8e58a827be20da21c5c3d90f

gabrieldemarmiesse: Here we have a choice between implicit conversion and explicit conversion. I believe we'll choose explicit to clearly indicate the intent. That should avoid bugs in the future

bgreni: Frankly I agree with you 😄, I just wanted to bring it to your attention in case you hadn't seen it since you'll likely get some pushback from the team.

JoeLoser (approved):

JoeLoser:
!sync

Issue #2467, opened 2024-05-02
[Feature Request] Unify SSO between `InlinedString` and `String` type


gabrieldemarmiesse:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

We currently have https://docs.modular.com/mojo/stdlib/utils/inlined_string#inlinedstring and String. Does it make sense to have SSO enabled in String too? Is it something that's currently possible or would that be too much trouble for the compiler?

### What is your motivation for this change?

I guess that having SSO enabled on String would provide speed improvements. Furthermore, as we are writing more and more functions that work with Strings, I believe that we'll be avoiding a big refactor if we want SSO afterwards.

### Any other details?

Maybe that would require to make the internal buffer of String actually private? I wonder if some functions use the underlying pointer or List directly.

JoeLoser:
Thanks for filing this!  We actually discussed this exact topic earlier in the week in the team's design meeting.  Here are my notes:

- One first step is to remove `_ArrayMem` in `utils/inlined_strings` and use `InlineArray` directly.  We may run into trouble with our buffer formatting things in `builtin/int.mojo` and `builtin/hex.mojo` since they use the default constructor, which previously left undef values in the backing `pop.array` whereas we don’t allow undef values in default constructor in the new `InlineArray` for safety purposes, so you’ll have to pick a default fill value (such as 0) I suspect (FYI @lsh).

With SSO specifically, we talked about a few things like ABI concerns and the like, but decided on the following:
- Add SSO parameter to the existing `String` type.  Make it a keyword-only parameter such as `_Size` that defaults to the fixed SSO size (such as the existing value of `24`).  This will make it so it's clearly a "use at your own risk" if you explicitly set this parameter in caller code (by needing to specify the keyword parameter).  This is because different parameters for the SSO value would generate a distinct `String` type in our elaborator and have the same consequences as "template explosion" in C++ and make it difficult to use different types across APIs, e.g. `String` and `String[_Size=32]` for example.  One thing to note is nothing is ABI stable in the library, and so we have freedom to change the details of the SSO which can affect the layout of the type.  This is also one reason why the keyword was chosen to have an underscore prefix.
- Remove `InlinedString`.
- Keep `FixedString` as is.  It has value still, but we can safely remove `InlinedString` which was being that initial stepping stone as we were thinking about SSO.

[...]

PR #2805, opened 2024-05-23
[stdlib] Add `Array` type

### LINKED ISSUE #2804: [Feature Request] [stdlib] [proposal] Add `Array` type
### Review Mojo's priorities

### What is your request?

Since the proposal to rename `InlineList` to `Array` fell through (#2773), I thought of adding a dynamic Array that follows most of Python's behavior but lets the developer decide if and how  much small buffer optimization to use for register passable types.

Originally I wanted to do it for `Scalar[type]` and have the small buffer optimizations be automatic according  to the bitwidth of each type, but `List[T]` and `InlineArray[T]` expect `CollectionElement`. 

I'm also not sure if the resizing is worth it and what kind of performance hit it generates, the idea is that this be thought and applied according to each use case WDYT?.

Having stack allocated Arrays for numeric types.

### Any other details?

_No response_

martinvuyk:
Closes #2804

Since the proposal to rename `InlineList` to `Array` fell through (#2773), I thought of adding a dynamic Array that follows most of Python's behavior but lets the developer decide if and how  much small buffer optimization to use for DType subtypes.

The current implementation has basically become a user friendly wrapper around SIMD.

The current implementation lets the programmer decide what amount of "capacity" to allocate for the Array. Under the hood it rounds up to the next upper power of two for the underlying SIMD. Though every method has then to take that delta into account (everything is parametrized).

Every operation on the Array is done in a vectorized manner except extending, appending, concatenating, etc. So using `array.__contains__` can be a lot faster depending on the hardware it's running on.

You then have calculations like the dot product, cosine between arrays, applying a function to the array, and many other future ease of use features that can be added that vectorize the ops wherever possible.

Arrays of different lengths can interact with each other in many methods where it makes sense (concatenation, appending values from other to self, etc.).

Another important aspect for the future is the ease of use for going `List[T] <-> Array[T]` taking SBO into account.

It would also be awesome to add some benchmarks.

Examples:

```mojo
from collections import Array
alias Arr = Array[DType.uint8, 3]
var a = Arr(1, 2, 3)
var b = Arr(1, 2, 3)
print((a - b).sum()) # prints 0
print(a.avg()) # prints 2
print(a * b) # dot product: 14
print(a.cross(b)) # cross product: Array(0, 0, 0)
print(2 in a) # prints True
print(a.index(2).value() if a.index(2) else -1) # prints 1
print((Arr(2, 2, 2) % 2).sum()) # 0
print((Arr(2, 2, 2) // 2).sum()) # 3
print((Arr(2, 2, 2) ** 2).sum()) # 12
```


gabrieldemarmiesse:
@martinvuyk  If possible, I would recommend to avoid starting working on a new function/struct that's not present in Python and that has not been approved yet by the Mojo maintainers in the issues. 

While I'm sure that the new structs/functions contributors are writing have value, there is information internal to Modular that we don't have, it's possible that the Modular staff has already thought about some alternative, or has a plan for another API. Since we're in the dark here, let's proceed with caution.

It know it's annoying as it can delay your work, but it may also avoid some wasted work on PRs. (well, hopefully not really wasted, since I'm sure you learned a lot while writing those PRs!)

martinvuyk:
> @martinvuyk If possible, I would recommend to avoid starting working on a new function/struct that's not present in Python and that has not been approved yet by the Mojo maintainers in the issues.
> 
> While I'm sure that the new structs/functions contributors are writing have value, there is information internal to Modular that we don't have, it's possible that the Modular staff has already thought about some alternative, or has a plan for another API. Since we're in the dark here, let's proceed with caution.
> 
> It know it's annoying as it can delay your work, but it may also avoid some wasted work on PRs. (well, hopefully not really wasted, since I'm sure you learned a lot while writing those PRs!)
> 
> I'm currently waiting for the go of the maintainers to implement small buffer optimization in `List`, in which case this pull request would not be needed anymore. You can follow the progress here: [#2467 (comment)](https://github.com/modularml/mojo/issues/2467#issuecomment-2129157967)

Thank you for the comment, I'm waiting for confirmation but it's really fun to tinker in this language. The main difference between this type and SBO is that this is 100% on the stack. It tries to follow Python's Array as well, but only using SIMD. My hope is that this becomes the type that is used for high performance IO or be the user friendly interface to SIMD, since List lives in the heap and SBO will only get you so far when you want Arrays of different widths and do vectorized ops on them.

If, for example, someone wants to do high performance operations on strings that are parallel like uppercasing SBO will give you faster acces but not really faster ops, whereas SIMD does.

I've seen your PR and do think it is necessary and will have a big impact on List perf., but I don't think List should be the end  all be all. The main reason ppl. will come to Mojo is for out of the box support for the most innovative CPU and GPU ops, to get that, people will expect an Array to be there and be performant. If we stay only with SBO and iterating over arrays in a classic way... every high perf. C++ lib will be faster than the Mojo stdlib. Mojo  should be the place for easy to use SIMD ops IMO.

There  are some things that I'm not even sure if they are possible, since  I don't know the layout in memory  of SIMD and DTypePointer's pointee and many other Mojo  internals (I don't understand MLIR), so this  PR is really just an experiment. I have  no problem killing it if the Mojo team tells me to. Code is just code.

martinvuyk:
I will progressively bring these algorithms into `List` once we have better overload or generic `@parameter if` constraining mechanisms

PR #2806, opened 2024-05-23
Implement  `List.copy()`

msaelices:
- [X] Implement `List.copy()`
- [X] Tests for `List.copy()`


soraros:
Why do we want this method at all? `List` is a value type, and doesn't even have the notion of shallow copy.

msaelices:
> Why do we want this method at all? `List` is a value type, and doesn't even have the notion of shallow copy. 

Actually, it's not shallow copy, it depends on the type of the value, but it iterates over the list appending all the values, as you can see in the constructor I'm calling to:

image

I don't know if you want this method. It's just to match the Python `list.copy()` one

artemiogr97:
Isn't creating a shallow copy dangerous with how mojo works?, the `List` struct takes ownership of its elements, which means that it is responsible for destroying them when the list is destroyed, creating a shallow copy would mean that we will try to delete the elements twice.

Right now you don't have a double free issue because the destructor uses the length of the list to know what to destroy at the end, in your test try doing `len(l2)` it returns 0.

msaelices:
> Isn't creating a shallow copy dangerous with how mojo works?, the `List` struct takes ownership of its elements, which means that it is responsible for destroying them when the list is destroyed, creating a shallow copy would mean that we will try to delete the elements twice.
> 
> Right now you don't have a double free issue because the destructor uses the length of the list to know what to destroy at the end, in your test try doing `len(l2)` it returns 0.

Thanks for the `len()` catch! Fixed here for consistency reasons: https://github.com/msaelices/mojo/commit/0a088eda2ff639d2037de52452b5a7095c4398c8

But, I think you are right. We cannot perform shallow copies as `List.__del__` and other methods call `destroy_pointee` to delete the value that the pointers point to.

Should we implement a deep copy instead of not implement a `copy()` method at all?

ConnorGray:
Thanks for this contribution @msaelices. Apologies for the delay in a response to this issue, which is now a couple months since its last update. I think we should hold off on adding this for the time being, for the following reasons:

1. The current behavior implemented by this PR is I think largely subsumed by the existing `List.__copyinit__` functionality, which does a "deep" copy of the list.

2. A true "shallow" copy of the `List` would lead to double-free behavior. Any situations where a shallow copy is desired are probably better written to leverage the lifetimes system using the `Reference` type and/or `borrowed` argument convention arguments.

3. There is a bit of in-progress work trying to settle on how copies for stdlib types can be made explicit (see the `ExplicitlyCopyable` trait) that overlaps somewhat with the `.copy()` method convention. We should avoid adding a new method inconsistent with that. We can always revisit down the line though.

Thanks again for this contribution, and please let me know if you have any questions or concerns.

PR #2807, opened 2024-05-23
[stdlib] Add `assert_is()` and `assert_is_not()` functions

helehex:
Adds functions for asserting identity comparisons, along with some helper traits to make it possible.
Having the address information in the error message isn't always helpful, but sometimes it is.

### DIFF
--- a/stdlib/src/testing/testing.mojo
+++ b/stdlib/src/testing/testing.mojo
@@ -133,7 +133,9 @@ fn assert_equal[T: Testable](lhs: T, rhs: T, msg: String = "") raises:
         An Error with the provided message if assert fails and `None` otherwise.
     """
     if lhs != rhs:
-        raise _assert_equal_error(str(lhs), str(rhs), msg, __call_location())
+        raise _assert_cmp_error["`left == right` comparison"](
+            str(lhs), str(rhs), msg, __call_location()
+        )
 
 
 # TODO: Remove the String and SIMD overloads once we have more powerful traits.
@@ -151,7 +153,9 @@ fn assert_equal(lhs: String, rhs: String, msg: String = "") raises:
         An Error with the provided message if assert fails and `None` otherwise.
     """
     if lhs != rhs:
-        raise _assert_equal_error(lhs, rhs, msg, __call_location())
+        raise _assert_cmp_error["`left == right` comparison"](
+            lhs, rhs, msg, __call_location()
+        )
 
 
 @always_inline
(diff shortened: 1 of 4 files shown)


gabrieldemarmiesse (commented):
I would, if possible hold off on using the `is` keyword until the references are fully flushed out and we have decided exactly what `is` means with references. Are there structs that you would like to test with those new functions right now?

In the future, when `is` is used persuasively in the language, after the reference work has been done, I believe it would be very useful to have `assert_is` and `assert_is_not`. Would it be possible for you to re-submit the PR at this time?

helehex:
it works for PythonObjects and there are already places in the stdlib tests where this can be applied. The reason i'm passing them by reference is to have address information in the error message, which is not necessary and could be removed or improved later.

gabrieldemarmiesse:
In this case, since we are working with the current `__is__` which has nothing to do with references, we should remove the addressed, they're not useful, as two objects where `obj1 is obj2` doesn't necessarily means that the addresses of both objects are the same. The rest looks fine and can be merged :)

stdlib/src/testing/testing.mojo
```diff
@@ -300,29 +308,64 @@ fn assert_almost_equal[
...
-        + "\n  right: "
-        + rhs
-    )
-    if msg:
-        err += "\n  reason: " + msg
-    return _assert_error(err, loc)
+@always_inline
+fn assert_is[
+    T: StringableIdentifiable
+](lhs: Reference[T, _, _], rhs: Reference[T, _, _], msg: String = "") raises:
```
ConnorGray, line 314: Thanks for this, these are a natural extension of the assertion functions! 🙂 

I agree with the suggestion that we hold off on defining these in terms of `Reference`, and we avoid printing the `UnsafePointer.address_of(<lhs/rhs>)` below.

The issue is that it gives a false impression about what `is` actually does. `is` is like a "deep identity" check. As the test added later on demonstrates, two `PythonObject` instances on the stack will have different _stack_ addresses (if printed in the error case below), but still be `is` equivalent because their interior pointer points to the same reference-counted object.

For that reason, I suggest we change this signature to `lhs: T, rhs: T, ..`.

(Not for this PR, but if we really wanted to get the "interior" addresses for the purpose of printing them, that is something that types conforming to `Identifiable` would need to expose explicitly. Hypothetically, not suggesting this, there could be `Identifiable.__is_address__(self) -> Int` that e.g. for PythonObject would return the reference counted pointer address.)

ConnorGray:
!sync

PR #2810, opened 2024-05-24
[stdlib] Add `String.splitlines`

YichengDWu:
`String` now has a `splitlines()` method, which allows splitting strings at line
  boundaries. This method supports [universal newlines](https://docs.python.org/3/glossary.html#term-universal-newlines)


martinvuyk:
you will need to add support for python's [universal separators](https://docs.python.org/3/library/stdtypes.html#str.splitlines), I have a PR #2793 you can look at to see all utf8 equivalents you can support once `String.isspace()` is merged. You could also pull the examples from the Python docs and use them as examples in the docstrings and as testcases

YichengDWu:
Thanks for the comment. I believe `/n`, `/r` and `/r/n` covers the most use cases. I will take a look if I can add all separators.

stdlib/src/builtin/string.mojo
```diff
@@ -1446,6 +1446,48 @@ struct String(
 
         return output
 
+    fn splitlines(self, keepends: Bool = False) -> List[String]:
```
mzaks, line 1449: I think it is better to move `keepends: Bool = False` to parameter. As following:

`fn splitlines[keepends: Bool = False](self) -> List[String]:`

This way the branch in the while loop will be backed at compile time:
```Mojo
@parameter
if keepends:
   output.append(self[current_offset : loc + 1])
else:
   output.append(self[current_offset:eol])
```

YichengDWu: Good point.

stdlib/src/builtin/string.mojo
```diff
@@ -1446,6 +1446,49 @@ struct String(
...
+
+        Returns:
+          A List of Strings containing the input split by line boundaries.
+        """
+        var output = List[String]()
+        var length = len(self)
+        var current_offset = 0
+
+        while current_offset < length:
+            # Find the position of the next newline character
```
martinvuyk, line 1463:
```suggestion
            # Find the position of the next newline character
            # FIXME: this should take all universal newlines into account
            # see https://docs.python.org/3/library/stdtypes.html#str.splitlines
```

laszlokindrat:
Thanks for the patch, I'm very excited to see this land. Before we start to review this more seriously, let's wait for https://github.com/modularml/mojo/pull/2793 to land. There are several interesting discussions there that I would like to resolve before we implement more functionality like these.

stdlib/src/builtin/string.mojo
```diff
@@ -1439,6 +1439,51 @@ struct String(
 
         return output
 
+    fn splitlines[keepends: Bool = False](self) -> List[String]:
```
laszlokindrat, line 1442: Please follow the Python API for this. We can optimize later.
```suggestion
    fn splitlines(self, keepends: Bool = False) -> List[String]:
```

YichengDWu: Done!

laszlokindrat:
https://github.com/modularml/mojo/pull/2793 has landed internally and will be included in the next nightly. Could you please rebase after that? You can also preemptively read about some of the discussions we had there.

YichengDWu:
Will do, thanks for the reminder!

YichengDWu:
I'm confused with the new `isspace` implementation. @laszlokindrat @martinvuyk 

In mojo
```mojo
 23> var a = String("\\x1c")
(String) a = "\\x1c"
 24> print(a.isspace()) 
 25.  
True
```
v.s. in python

```python
In [31]: a = "\\x1c"

In [32]: a.isspace()
Out[32]: False
```

Is this inconsistency  intended?

martinvuyk:
> ```mojo
> \\x1c
> ```

escape secuences have only one \ . the double \\ is for the value to appear as only having 1 \ in the docstrings. Unicode escape sequences are still not supported, that's why they are inside a List as an alias inside the isspace func.

The current isspace implementation is not complete yet, I have another PR that will probably land in the next 2 days. I messed up the special separator characters (the values inside the lists are incorrect). And isspace still needs to evaluate all substrings in a string not just 1.

For the purposes of your splitlines implementation, I'm not sure how much isspace will help you in the end since you will have to check anyway for combinations of characters (\n, \r\n, etc.) , and the _isspace(c: UInt8) function returns true for "normal" whitespace (" ") as well...

YichengDWu:
This one is more concerning for me:
```mojo
 76> var a = String("\x85")
(String) a = "\x85"
 77> var c = a.isspace() 
 78.  
(Bool) c = False
```
Should `\x85` be in the `_SPACES_TABLE`?

YichengDWu:
Aha, there is a dirty bug in mojo
```moj
 14> print(ord("\x85")) 
 15.  
5
 15> print(ord("\x05"))
5
```
`"\x85"` itself if **not** encoded in utf-8.

.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
         stages: [commit]
       - id: check-docstrings
         name: check-docstrings
-        entry: python3 ./stdlib/scripts/check-docstrings.py
```
laszlokindrat, line 12: Why this change?

stdlib/src/builtin/string.mojo
```diff
@@ -1496,6 +1524,46 @@ struct String(
...
         return output
 
+    fn splitlines(self, keepends: Bool = False) -> List[String]:
+        """Split the string at line boundaries.
+
+        Arguments:
+        keepends: If True, line breaks are kept in the resulting strings.
+
+        Returns:
+        A List of Strings containing the input split by line boundaries.
```
laszlokindrat, line 1534: nit: indentation
```suggestion
            keepends: If True, line breaks are kept in the resulting strings.

        Returns:
            A List of Strings containing the input split by line boundaries.
```

stdlib/src/builtin/string.mojo
```diff
@@ -1496,6 +1524,46 @@ struct String(
...
+                var char = self[i]
+                var next_char = self[i + 1] if i + 1 < length else ""
+
+                if _isnewline(char):
+                    loc = i
+                    if char == "\r" and next_char == "\n":
+                        eol_length = 2
+                    break
+
+            if loc == -1:
```
laszlokindrat, line 1554: Can you try if this works?
```suggestion
            else:
```

stdlib/src/builtin/string.mojo
```diff
@@ -598,6 +598,34 @@ fn _isspace(c: UInt8) -> Bool:
...
+    table[ord("\x1e")] = 1
+    return table
+
+
+alias _NEWLINES_TABLE = _get_newlines_table()
+
+
+fn _isnewline(c: String) -> Bool:
+    # TODO: add \u2028 and \u2029 when they are properly parsed
+    # FIXME: \x85 is parsed but not encoded in utf-8
```
laszlokindrat, line 623: I haven't looked, but isn't this a problem with the implementation of `ord`?

laszlokindrat (approved):
Cool, thanks! Could you please add a PR description as well?

.pre-commit-config.yaml
YichengDWu: I did not make this change, this appeared somehow after I updated the branch after https://github.com/modularml/mojo/commit/d51dad80d863fac2a268f9215e22b0a989ca9176

stdlib/src/builtin/string.mojo
YichengDWu: No. `ord` is correctly implemented as far as I can tell. It recognize unicode characters.
```mojo
  3> var s = String("\xc2\x85"); 
(String) s = "\U00000085"
  4> print(s._buffer.__str__()) 
  5.  
[SIMD[DType.uint8, 1](194), SIMD[DType.uint8, 1](133), SIMD[DType.uint8, 1](0)]
  
5> print(ord(s))
133

```

The problem is that `\x85` *itself* is not encoded in utf-8 but in ASCII.

```
   6> var s = String("\x85")
(String) s = "\x85"
  7> print(s._buffer.__str__())
[SIMD[DType.uint8, 1](133), SIMD[DType.uint8, 1](0)]

```

For the some reason, `\u2028` is not recognized 
```
  8> var s = String("\u2028")
[User] error: Expression [7]:1:17: invalid escape sequence
var s = String("\u2028")
                ^

[User] error: Expression [7]:1:17: invalid escape sequence
var s = String("\u2028")
                ^

[User] error: Expression [7]:1:17: invalid escape sequence
var s = String("\u2028")
                ^

[User] error: Expression [7]:1:17: invalid escape sequence
var s = String("\u2028")
                ^

```

The is probably a bug in mojo's parser.

stdlib/src/builtin/string.mojo
YichengDWu: It does! nice trick.

stdlib/src/builtin/string.mojo
YichengDWu: `ord("\x85")` treats "\x85" as a unicode character "\xc2\x85" but in fact it is in ASCII hence `ord("\x85")==5`

YichengDWu:
PR description added. CI failure look unrelated.

laszlokindrat:
!sync

stdlib/src/builtin/string.mojo
martinvuyk: I opened an issue about it 3 days ago #2842

> ord("\x85") treats "\x85" as a unicode character

that is how python does it and \x85 is the way that it is annotated even though as you said it has a preceding \xc2 byte

YichengDWu: Yeah definitely a bug.

Issue #2813, opened 2024-05-24
[BUG] `DTypePointer[bool]` packs bits inconsistently


helehex:
### Bug description

When using `DTypePointer[bool]` `store()`/`load()` with different widths, you get inconsistent results.

### Steps to reproduce

```mojo
var ptr = DTypePointer[DType.bool].alloc(4)
ptr.store(0, True)
ptr.store(1, True)
ptr.store(2, True)
ptr.store(3, True)
print(ptr.load[width=4](0))
```

prints: `[True, False, False, False]`

### System information

```shell
Host Information
  ================

  Target Triple: x86_64-unknown-linux
  CPU: skylake
  CPU Features: adx, aes, avx, avx2, bmi, bmi2, clflushopt, cmov, crc32, cx16, cx8, f16c, fma, fsgsbase, fxsr, invpcid, lzcnt, mmx, movbe, pclmul, popcnt, prfchw, rdrnd, rdseed, sahf, sgx, sse, sse2, sse3, sse4.1, sse4.2, ssse3, x87, xsave, xsavec, xsaveopt, xsaves

mojo 2024.5.2405 (fc1ed0a3)

modular 0.8.0 (39a426b5)
```

MadAlex1997:
Probably Related to [2492](https://github.com/modularml/mojo/issues/2492), I am pretty sure there are issues at the DType pointer level with the i1 representation of `SIMD[DType.bool,1]` and the 8 Bit C style bool used in the pointers.

bgreni:
Seems as though this has been fixed with recent changes?

```mojo
var ptr = DTypePointer[DType.bool].alloc(4)
SIMD[size=4].store(ptr, SIMD[DType.bool, 4](True, False, True, True))
print(SIMD[size=4].load(ptr)) # prints [True, False, True, True]
```

rparolin:
Looks like the patch mentioned above by @bgreni resolved the issue.  I'll close it but we can definitely reopen it if knock-ons are discovered.  Thank you!

helehex:
looks like when `store`/`load` moved to `SIMD`, the fix wasn't

helehex:
```mojo
var ptr = DTypePointer[DType.bool].alloc(4)
SIMD[DType.bool, 1].store(ptr, 0, True)
SIMD[DType.bool, 1].store(ptr, 1, True)
SIMD[DType.bool, 1].store(ptr, 2, True)
SIMD[DType.bool, 1].store(ptr, 3, True)
[...]

PR #2815, opened 2024-05-24
[stdlib] Patch bool dtypepointer load

bgreni:
This PR offers a temporary fix to address #2813 in case the team cannot currently prioritize this, and I imagine a proper fix requires changes at a deeper level than OS contributors have access to?

### DIFF
--- a/stdlib/src/memory/unsafe.mojo
+++ b/stdlib/src/memory/unsafe.mojo
@@ -910,11 +910,23 @@ struct DTypePointer[
                 ]()
             return v
 
-        return (
-            self.address.offset(offset)
-            .bitcast[SIMD[type, width]]()
-            .load[alignment=alignment]()
-        )
+        # TODO: This is a temp fix for #2813
+        @parameter
+        if type.is_bool():
+            var v = SIMD[type, width]()
+
+            @parameter
+            for i in range(width):
+                v[i] = self.address.offset(int(offset) + i).load[
+                    alignment=alignment
+                ]()
+            return v
+        else:
+            return (
+                self.address.offset(offset)
+                .bitcast[SIMD[type, width]]()
+                .load[alignment=alignment]()
+            )
 
     @always_inline("nodebug")
     fn store[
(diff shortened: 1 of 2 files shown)


helehex:
Thanks for patching this

stdlib/test/memory/test_memory.mojo
```diff
@@ -518,6 +518,25 @@ def test_indexing():
...
+    ptr.store(1, True)
+    ptr.store(2, True)
+    ptr.store(3, True)
+    assert_equal(
+        ptr.load[width=4](0), SIMD[DType.bool, 4](True, True, True, True)
+    )
+
+    var ptr2 = DTypePointer[DType.int32].alloc(4)
+    ptr2.store(0, 1)
+    ptr2.store(1, 2)
```
helehex, line 533: Could you also add the inverse of this test, where you `store[width=4]()` and `load[width=1]()`?

helehex:
i suppose there a chance the original behavior of packing the bits is desired? but im not sure.. we need some more input

bgreni:
> I suppose there's a chance that the original behavior of packing the bits is desired?

I think that fact that my change works could point to `.bitcast[SIMD[DType.bool, width]]().load()` not working as intended? Regardless I think this code should work as expected and any special logic required to make it work should be implemented internally. Seems like a pretty major footgun that has no need to exist in a brand new language.

stdlib/test/memory/test_memory.mojo
bgreni: Good catch, `store[width=4]()` also doesn't work currently.

stdlib/src/memory/unsafe.mojo
```diff
@@ -904,11 +904,25 @@ struct DTypePointer[
...
+        if type.is_bool():
+            var v = SIMD[type, width]()
+
+            @parameter
+            fn do_load[i: Int]():
+                v[i] = self.address.offset(int(offset) + i).load[
+                    alignment=alignment
+                ]()
+
+            unroll[do_load, width]()
```
soraros, line 918:
```suggestion
            @parameter
            for i in range(width):
                v[i] = self.address.offset(int(offset) + i).load[
                    alignment=alignment
                ]()
```

bgreni: Thanks for pointing that out!

JoeLoser:
FYI @Dan13llljws in case you notice any issues as I know you're poking in this area

PR #2820, opened 2024-05-25
[tools] Add a template for pull requests

gabrieldemarmiesse:
Note that since nightly is not the default branch in github, this template won't be seen until the next release. This annoyance should be fixed if https://github.com/modularml/mojo/issues/2556 is addressed.

For more information about pull request templates: 
* https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates#pull-request-templates
* https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,10 @@
+<!--
+Thanks for submitting a pull request, 
+your contribution is really appreciated!
+
+If possible, add a link to the issue you are 
+trying to solve in the pull request description.
+
+If your pull request is big (> 100 lines), consider splitting it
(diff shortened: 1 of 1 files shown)


JoeLoser (commented):
Great! Should we mention something about the title one liner to include a [category] for convenience of new contributors since that is checked by CI?

gabrieldemarmiesse:
I thought about it and I believe that won't be good. The longer the template is, the less likely it is that the contributors are going to read it entirely. I think we should put only inside information that cannot be given by the CI, and limit ourselves as much as possible with the overall length

JoeLoser (approved):

PR #2825, opened 2024-05-25
[stdlib] Add optional small buffer optimization to `List`, take 2

gabrieldemarmiesse:
This PR solves part of #2467 

This PR is part of three PRs to read and merge in the following order
* #2825 
* #2826 
* #2827 

### The small buffer behavior

We use `InlineArray` to store up to `small_buffer_size` elements on the stack. 
When the user requests a capacity of `N`, if `N <= small_buffer_size`, then the small buffer is used. We make the pointer of the `List` point to the `InlineArray`. And we set the `self.capacity` to `small_buffer_size` since we can store up to `small_buffer_size` elements before doing a reallocation.

### Constructing from existing pointers
When constructing from an existing pointer, to avoid copying the values, the pointer is saved directly in `self.data` without moving the values, even if the size of the pointer is smaller than the buffer. This is to avoid moves for performance reasons. To reallocate the values on the small buffer (if it is big enough) one can simply do a copy of the `List`.

### Alternative considered: `Variant`
You might wonder, why not using a `Variant` since we don't need to store the `capacity` and the pointer `data` (and the `size` could be smaller, like an `UInt8`) when using the small buffer? This is because by setting the pointer to the stack allocated `InlineArray`, there is no overhead introduced compared to a normal list when doing to following operations: reading, checking bounds, getting the size, getting the capacity. The exact same steps are done, with just different values inside the variables. You might get some speedups when reading the values in the buffer due to cache locality.

### Checking where are the elements stored
Sometimes we need to know if the `InlineArray` is used or if we allocated data with `UnsafePointer.alloc`. When this is the case, we can do the following check `if self.data == self._small_buffer.unsafe_ptr()`. Note that this produces a compiler bug when materializing. See the next PR.

### Backward compatibility
By default, the small buffer size is `0`. Some `@parameter if self.sbo_enabled` here and there make sure that we don't add a single instruction when the small buffer size is null. So the default behavior of `List` doesn't change at all.
A notable difference is that `List[Int]` now means a list with no buffer optimization. Using `List[Int, _]` can be used in a signature to signify that the function works with any small buffer sizes.

### Unit tests
I take all the tests in `test_list.mojo` and run them with 10 different small buffer sizes. I also sometimes mix buffer sizes to make sure that methods like `extends` work with different SBO sizes.

I added a materialization test. But it only works with a SBO size of 0 due to the following compiler bug: https://github.com/modularml/mojo/issues/2637 

Since the bug won't be solved anytime soon, I made #2826 as a workaround


JoeLoser:
Any initial perf numbers with this compiler workaround approach?

gabrieldemarmiesse:
I'm not there yet, I still need some more work before being able to run the benchmarks :) I'll ping you with a summary when everything is done. If that was possible to review quickly afterwards, I would be grateful, as I may get a lot of conflicts otherwise.

JoeLoser:
> I'm not there yet, I still need some more work before being able to run the benchmarks :) I'll ping you with a summary when everything is done. If that is possible to review quickly afterwards, I would be grateful, as I may get a lot of conflicts otherwise.

Sounds good, thanks! I'll review ASAP once it's ready.

JoeLoser:
!sync

JoeLoser:
!sync

stdlib/src/utils/span.mojo
```diff
@@ -101,7 +101,12 @@ struct Span[
         self._len = len
 
     @always_inline
-    fn __init__(inout self, list: Reference[List[T], is_mutable, lifetime]):
+    fn __init__[
+        small_buffer_size: Int
```
JoeLoser, line 105: Do you mind adding doc strings for this? There's a bug with https://github.com/modularml/mojo/pull/2795 that I'll fix soon, but for now, use the wrapper Python script to detect and fix the errors. The fix forward is we need to also specify the `--diagnose-missing-doc-strings` option. I don't like that, but we're discussing that separately internally right now. When I'm by a laptop next, I'll submit a fix for that unless you beat me to it.

JoeLoser:
!sync

stdlib/src/collections/list.mojo
```diff
@@ -826,7 +929,9 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         )
         return (self[].data + idx_as_int)[]
 
-    fn count[T: ComparableCollectionElement](self: List[T], value: T) -> Int:
+    fn count[
+        U: ComparableCollectionElement
+    ](self: Reference[List[U, Self.small_buffer_size], _, _], value: U) -> Int:
```
gabrieldemarmiesse, line 934: I had to use references in those situations otherwise the compiler wasn't happy at all with the extra parameter. The behavior didn't change. Same for other methods like `__str__`

JoeLoser: Can you file compiler bugs please?  Your PRs are definitely hitting a few and I want to make sure they're all captured!  Perhaps a comment in the code too referencing such bugs?

gabrieldemarmiesse: I don't mind filing those bugs but can it be after the PR is merged? Making a minimal reproducible example can sometimes take a few minutes but it can also take a lot of time. I can make another PRs to add the TODOs later if that's okay with you.

JoeLoser: Of course, totally fine.

JoeLoser:
Unfortunately this is hitting a nasty compiler crash for internal things with 500+ stack frame trace 😞 .

So it won't land today.  I'll have to investigate this and file compiler repros, especially since it's not something repro'ing in the public stdlib alone.

nmsmith:
The parameter name `small_buffer_size` isn't very illuminating. When I read it, I think "okay, the buffer is small, but why should I care?" IMO what you want to convey is that the list has an _inline_ buffer. This is the property that matters for performance. If you tell somebody the buffer is inline—and if they know anything about how memory works—then they will conclude that having an inline buffer might help with performance.

TL;DR: Maybe the parameter should be called `inline_buffer_size`.

nmsmith:
Also if I'm reading the code right, the "buffer size" is actually the **capacity** of the list. So `inline_capacity` would be even more accurate.

gabrieldemarmiesse:
@nmsmith getting this merged is already tons of work. So I'd be happy if naming discussions happen after the merge. Mojo hasn't made any promises about backward compatibility so we can change parameter names whenever we feel comfortable with a new name :)

gabrieldemarmiesse:
I've hit an issue in the unit test. Not sure if it's a compiler bug or an issue with my code. I'll take a look at it later on.

JoeLoser:
!sync

nmsmith:
Storing list elements inline seems like it's going to make constructing views/slices/spans of a list more challenging. Is there a plan for how this would be done? It's not sufficient for the view to point to the list's buffer, because the buffer no longer contains all of the elements.

One crazy idea worth exploring is whether Mojo should support applying the small buffer optimization to views/slices/spans of lists. With the right model for references, maybe an immutable reference can store its target inline.

JoeLoser (commented):
It'd be interesting to explore this again now that we have SSO for String.  In the SSO for String, we steal bits from the `capacity` flag to represent whether the data is inline or out of line (see the [design doc](https://github.com/modular/max/blob/e606578390fb2ce84db772a0b3ae29e63d221f1d/mojo/proposals/string-design.md#representation-implementation).  I wonder if we want to try that approach here first and then we can graduate towards considering an approach of users explicitly controlling the SBO size.  I think starting with the simpler approach where users can't control the SBO size would be better for now.  WDYT @lattner?

gabrieldemarmiesse:
@JoeLoser Let me know when you have the bandwidth to tackle (I mean review) PRs on the subject. I'll update the PR then

JoeLoser:
> @JoeLoser Let me know when you have the bandwidth to tackle (I mean review) PRs on the subject. I'll update the PR then

Hey @gabrieldemarmiesse, I'm in a good spot to review these sort of PRs now.  Taking ideas from how SSO is implemented and applying it to `List` could work nicely.  As mentioned earlier, I'd hold off on a size parameter to `List` to control the inline buffer size as it may cause type explosion across APIs.  I'd rather start simple with a calculated/automatic "small buffer" optimization and then we can explore the size as a parameter in the future if that works.

soraros:
I don't think we can attempt this before migrating everything off `l.data` and using `l.unsafe_ptr()` instead.

JoeLoser:
> I don't think we can attempt this before migrating everything off `l.data` and using `l.unsafe_ptr()` instead.

Can you please file an issue for this? Should be a mostly mechanical change.

soraros:
> Can you please file an issue for this? Should be a mostly mechanical change.

Filed #4754.

gabrieldemarmiesse:
@JoeLoser do you have some ideas in mind on how that would work in practice for this proof of concept ? For example, should we do copy-on-write like String has? Or do we stick to something simpler, like inlining small types and disabling SBO for types bigger than 16 bytes? Is the complexity worth it then since it's only going to kick in for very few types? Can we afford adding an InlineArray to the List type to store some data there? If yes, how big should we allow the List struct to become? Do we aim to have the List[T] struct be the same size for all T possible? If it's something that you haven't though about, it's fine, I can use my jugement to come up with something decent :)

gabrieldemarmiesse:
Closing this as it requires a new approach anyway

PR #2832, opened 2024-05-26
[stdlib] Add some logging to `test_reverse.mojo` to flush out a flaky bug

gabrieldemarmiesse:
See https://github.com/modularml/mojo/issues/2369 this bug is appearing more and more. Some logging should help us understand what fails exactly.

### DIFF
--- a/stdlib/test/builtin/test_reversed.mojo
+++ b/stdlib/test/builtin/test_reversed.mojo
@@ -16,6 +16,7 @@ from testing import assert_equal
 
 
 def test_reversed_list():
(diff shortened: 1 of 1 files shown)


JoeLoser (commented):
Maybe let's switch this test to mojo build and then run that executable? That supposedly doesn't crash from when Daniel looked at this the other day. That suggests a non deterministic crash in the JIT code path, ie a compiler bug and not a stdlib bug.

gabrieldemarmiesse:
Closing as I found the cause. Making a PR for the fix now.

PR #2837, opened 2024-05-26
[stdlib] Fix `String.__getitem__`

martinvuyk:
As title. `String.__getitem__` has no reason to be recursive, and I think a `utils.index.normalize_idx()` utility is needed because I've seen wildly different normalization impls over the code and none take over and under indexing into account.


laszlokindrat:
I think this is a duplicate of https://github.com/modularml/mojo/pull/2677, right?

laszlokindrat:
> > I think this is a duplicate of #2677, right?
> 
> sorry I hadn't seen that PR, but that one just aborts with out of bounds (which is not "safe" behavior IMO).

That's intentional, because we don't want to use `debug_assert` (it would cause a recursive definition).

> And I thought Indexer should be avoided now.

I fixed this in #2677 before landing it internally, and it will be included in the next nightly. Could you please check after that if this patch is still relevant, and if not, please close it.

martinvuyk:
> Could you please check after that if this patch is still relevant, and if not, please close it.

ok I'll wait until that helper is released, `String.__getitem__` still needs to be fixed. I'll delete the helper part.

> that one just aborts with out of bounds

This is unrelated to whether debug_assert is used.
What I mean is, I did this:
`return min(idx, end - 1) if idx > -1 else max(start, end + idx)`
which defaults to index 0 and len(item) -1 instead of aborting, which I think is "safer" default behavior than exploding the program.

laszlokindrat:
> > that one just aborts with out of bounds
> 
> This is unrelated to whether debug_assert is used. What I mean is, I did this: `return min(idx, end - 1) if idx > -1 else max(start, end + idx)` which defaults to index 0 and len(item) -1 instead of aborting, which I think is "safer" default behavior than exploding the program.

Thanks for clarifying, I see what you meant now! Aborting is safer than `debug_assert`, because the latter can (and in release builds is expected to) be disabled. In these methods, we want to guarantee that in any build we prevent callers from accessing out of bounds. Since we cannot use `raises` (it would propagate through almost everything), this is the best we have right now. At some point, the language might have a better mechanism for handling these, but not sure when.

PR #2847, opened 2024-05-27
[stdlib] List __getitem__ returns auto-dereferenced ref

### LINKED ISSUE #2432: [stdlib] Add `__refitem__` to `List`

mikowals:
With this, `List.__getitem__()` no longer makes copies when returning a value.  I also added a test to show that setting an individual field using sugared `my_list[0].value = 1` no longer produces extra copies.  Passing that test is why `__getitem__` receives a `Reference[Self, ...]` argument rather than just `self`.

I removed uses of `__get_ref` in `list.mojo` and `test_list.mojo` other than leaving one test to show that `__get_ref` was still functioning correctly until it is removed.  I left usage of `List.__get_ref` in other modules untouched. From a few quick efforts to eliminate `__get_ref` it was leading to some lifetime casting decisions that seemed complex and numerous enough they should be handled separately.

This supersedes #2544 which I will close. It should close #2432 and #2540 as `__refitem__` and it related difficulties are no longer relevant.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -64,10 +64,10 @@ struct _ListIter[
         @parameter
         if forward:
             self.index += 1
-            return self.src[].__get_ref(self.index - 1)[]
+            return self.src[][self.index - 1]
         else:
             self.index -= 1
-            return self.src[].__get_ref(self.index)[]
+            return self.src[][self.index]
 
     fn __len__(self) -> Int:
         @parameter
@@ -722,10 +722,11 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         return res^
 
     @always_inline
-    fn __getitem__(self, idx: Int) -> T:
+    fn __getitem__(
+        self: Reference[Self, _, _], idx: Int
+    ) -> ref [self.lifetime] T:
         """Gets a copy of the list element at the given index.
 
-        FIXME(lifetimes): This should return a reference, not a copy!
 
         Args:
             idx: The index of the element.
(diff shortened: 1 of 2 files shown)


JoeLoser:
!sync

JoeLoser:
Unfortunately this is hitting some compiler crashes internally. I'll file a GitHub issue, but until it's solved, I don't see an easy workaround for this really.

stdlib/src/collections/list.mojo
```diff
@@ -751,10 +751,11 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         return res^
 
     @always_inline
-    fn __getitem__(self, idx: Int) -> T:
+    fn __getitem__(
+        self: Reference[Self, _, _], idx: Int
+    ) -> ref [self.lifetime] T:
         """Gets a copy of the list element at the given index.
```
GeauxEric, line 757: nit.
`Gets a copy of` -> `Gets a reference of`

mikowals: Nice catch!  I went with `Gets the list element at the given index` because that is what the caller actually ends up with after auto-dereferencing.  It could be more detailed but I am not sure how much should get included in the doc string that will likely be seen often and encountered early when someone starts using Mojo.

JoeLoser:
!sync

ConnorGray:
Hey @mikowals I just wanted to share an update on this PR. @lattner personally had a look at the bug holding this up🙂

He mentioned the error message we were seeing could use some improvement, but realized that the issue was an ambiguity between whether the compiler should call `List.__getitem__` or `List.__setitem__` at a particular callsite.

He was able to fix this by creating a patch that updated `List.__getitem__` to return a `ref [_] T` and additionally removed `List.__setitem__` entirely. That patch was merged yesterday evening, and should be visible during the next nightly release.

The other callsites updated and tests you added in this PR are still awesome to have, so I'm going rebase this PR and merge those changes in :)

Thank you again for pushing on this improvement and your patience while it was investigated and fixed 🙂

If you're interested in continuing work in this vein ⛏️ , there's still a lot of low-hanging fruit in terms of places in the stdlib that need to be updated to use reference—we'd love to work with you on any future contributions that interest you🙂

ConnorGray:
!sync

lattner:
Amazing, thank you @mikowals !!

Issue #2822, opened 2024-05-25
[Feature Request] returning values in `__del__` overload


rd4com:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Thought about the `Dict` for a while,
it would be nice to be able to move a value out of `__del__()` (because it is a place where struct can be not whole).

Maybe a new taking del could help with return DictEntry.value^ ?

An example of what could be accomplished:
```mojo
@destructor
fn __del2__(owned self)->V: 
    self.key^.__del__()
    return self.value^
var x = instance^.__del2__()
```

(Not sure about the decorator)

### What is your motivation for this change?

Having `DictEntry.value` as an `Optional` could be the thing,
but since performance is a priority, maybe it would become slower ? (no benchmarks done)

The challenge is to be able to take the value field of a `DictEntry` as an owned value,

that way, users would not get a copy when doing `Dict.pop("")` and it could fix https://github.com/modularml/mojo/issues/2756 .

It is possible that it returns the value as a copy, because `return entry_value.value^`  is different than `return entry_value^.value^`.

&nbsp;

#### Reference to https://github.com/modularml/mojo/issues/2756

Please have a look at the `Dict.pop`, it is weird that we are doing `return entry_value.value^`,

how can the value entry_value be destroyed if it is not whole ?  ([Documentation: field lifetimes](https://docs.modular.com/mojo/manual/lifecycle/death#field-lifetimes))
This might be the source of the double free ?

JoeLoser:
@stumpOS do you have ideas here with respect to CheckLifetimes?

lattner:
You cannot make this specifically a 'del' but you can already make consuming methods, and they can return values.  You have to explicitly disable the destructor of 'self' or else it will be called, which is the trick.  This is equivalent to Rust's "mem::forget".  We don't have syntax for this, but you can write it out manually with mlir magic:

```
  fn take_value(owned self) -> V:
      # don't run the destructor on 'self'
      __mlir_op.`lit.ownership.mark_destroyed`(__get_mvalue_as_litref(self))
     # Can now pillage the pieces of 'self' that I want.  Any unused subfields
     # will be individual destroyed.
     return self.some_part^
```

PR #2850, opened 2024-05-27
[stdlib] Fix draft for 2756 (not to merge), changes DictEntry.value:V to DictEntry.value: Optional[V]

rd4com:
Hello,

This PR is just to help designing a solution for #2756 ,

It also makes the `Dict.pop` return the `DictEntry.value` as an owned value by moving the value.

But making it an `Optional` could affect the performance, and add complexity to the Dict.

*(See feature request #2822 for another approach)*

&nbsp;

The good part is that is seem to fix the #2756, so at least we know what could be the problem.

This PR is quite a design change, but it can be used by staff to see if it fixes some bugs

I don't recommend this solution that adds an Optional, but could be wrong !

Hope it is useful :+1:

### DIFF
--- a/stdlib/src/builtin/object.mojo
+++ b/stdlib/src/builtin/object.mojo
@@ -575,7 +575,7 @@ struct _ObjectImpl(CollectionElement, Stringable):
                 "'"
                 + str(entry[].key)
                 + "' = "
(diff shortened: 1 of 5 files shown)


rd4com:
Wrong branch

PR #2859, opened 2024-05-27
[stdlib] implement `has_single_bit`

### LINKED ISSUE #2682: [stdlib] Implement missing functions in the `bit` module

jiex-liu:
fix #2682

### DIFF
--- a/stdlib/src/bit/__init__.mojo
+++ b/stdlib/src/bit/__init__.mojo
@@ -22,6 +22,7 @@ from .bit import (
     bit_width,
     rotate_bits_left,
     rotate_bits_right,
+    has_single_bit,
     bit_ceil,
     bit_floor,
 )
--- a/stdlib/src/bit/bit.mojo
+++ b/stdlib/src/bit/bit.mojo
@@ -246,8 +246,6 @@ fn bit_not[
     return __mlir_op.`pop.xor`(val.value, neg_one.value)
 
 
-# TODO: implement bit_ceil, bit_floor, has_single_bit, et al
-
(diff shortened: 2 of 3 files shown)


JoeLoser (approved):
Looks good, thank you!

JoeLoser:
!sync

soraros:
- Should this be called `is_power_of_2`? `math` has that already.
- We should replace `_is_power_of_2` with this function.

JoeLoser:
> * Should this be called `is_power_of_2`? `math` has that already.
> * We should replace `_is_power_of_2` with this function.
> * We could write `(val != 0) & (val & (val - 1) == 0))` in both implementations (`Int` also).
> * Why is this preferred over `popcount(val) == 1`?

For (2), I already have a PR up for review internally; I also removed `is_power_of_2` from `math` in that same PR. 

For (3), feel free to use `!=` instead of `>`, I'd welcome a PR on top of this.

PR #2865, opened 2024-05-28
[stdlib] Add method `_update_with_bytes()` to `_Hasher`

gabrieldemarmiesse:
This is a follow-up on this comment: https://github.com/modularml/mojo/pull/2823#discussion_r1615226363

With this, the actual Hasher trait is complete and we can start implementing some hashers here and there, and adding the corresponding `__hash__` methods.

See https://github.com/modularml/mojo/blob/nightly/proposals/improved-hash-module.md for more information on the subject

### DIFF
--- a/stdlib/test/builtin/test_hasher.mojo
+++ b/stdlib/test/builtin/test_hasher.mojo
@@ -86,7 +86,7 @@ def test_complexe_hash_with_hasher():
 
 
 @value
-struct ComplexeHashableStructWithList(_HashableWithHasher):
+struct ComplexHashableStructWithList(_HashableWithHasher):
     var _value1: SomeHashableStruct
     var _value2: SomeHashableStruct
     var _value3: List[UInt8]
@@ -105,7 +105,7 @@ struct ComplexeHashableStructWithList(_HashableWithHasher):
 
 def test_update_with_bytes():
     var hasher = DummyHasher()
-    var hashable = ComplexeHashableStructWithList(
+    var hashable = ComplexHashableStructWithList(
         SomeHashableStruct(42), SomeHashableStruct(10), List[UInt8](1, 2, 3)
     )
     hasher.update(hashable)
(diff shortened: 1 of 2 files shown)


stdlib/test/builtin/test_hasher.mojo
```diff
@@ -79,8 +85,36 @@ def test_complexe_hash_with_hasher():
     assert_equal(_hash_with_hasher[DummyHasher](hashable), 52)
 
 
+@value
+struct ComplexeHashableStructWithList(_HashableWithHasher):
```
gryznar, line 89: typo `Complexe` -> `Complex`

stdlib/test/builtin/test_hasher.mojo
```diff
@@ -79,8 +85,36 @@ def test_complexe_hash_with_hasher():
...
+    var _value1: SomeHashableStruct
+    var _value2: SomeHashableStruct
+    var _value3: List[UInt8]
```
gryznar, line 92: It is quite strange, that other type `List[UInt8]` has the same naming convention `_value*` as of type `SomeHashableStruct`

stdlib/test/builtin/test_hasher.mojo
gabrieldemarmiesse: Thanks, I fixed it!

stdlib/test/builtin/test_hasher.mojo
gabrieldemarmiesse: The time we would spend changing or debating this name would be greater than the time lost by readers of the code who will be slightly inconvenienced by it. Furthermore it's not user-facing code, so the impact of a badly chosen name is reduced since less people are going to read this code. I would prefer that we avoid bikeshedding names of dummy variables in unit tests.

mzaks (approved):

JoeLoser (approved):

JoeLoser:
!sync

PR #2868, opened 2024-05-28
[stdlib] Add `String.__iter__()` and fix `String.isspace()`

martinvuyk:
Add an iterator for String that takes multi byte utf8 encoding into account (other encodings remain as `TODO`). 

Used a comparison in the utf8 bit range to know if a byte is a 1 byte, continuation byte, a 2 byte start, a 3 byte start, or a 4 byte start.

The algorithm is made to be faster for languages with mostly ASCII characters, because the amount of continuation bytes in a string are counted in the Iterator constructor. That way when all of the characters with continuation bytes have been iterated over (the continuation_bytes variable becomes 0), the `_utf8_byte_type` function does not need to be called and a Reference of length 1 is directly returned on every subsequent iteration.

The iterator returns a StringSlice of the corresponding byte length (without null terminator length).


martinvuyk:
@laszlokindrat the lookup table is creating problems in the pipelines (works on my machine lol), any idea how to solve this?
the pipeline error is `non-constant inputs`
the code is this
```mojo
fn _get_utf8_first_byte_table() -> InlineArray[UInt8, 256]:
    var table = InlineArray[UInt8, 256](unsafe_uninitialized=True)

    @parameter
    for i in range(256):
        if i < 0b1000_0000:
            table[i] = 0
        elif i < 0b1100_0000:
            table[i] = 1  # is continuation byte
        elif i < 0b1110_0000:
            table[i] = 2  # is 2 byte long
        elif i < 0b1111_0000:
            table[i] = 3  # is 3 byte long
        else:
            table[i] = 4  # is 4 byte long
    return table


alias _UTF8_FIRST_BYTE_TABLE = _get_utf8_first_byte_table()
```

laszlokindrat:
Maybe it should be a `@parameter if` within the loop?

martinvuyk:
It seems that the problem is `_isspace()` but even returning it to the original impl (without editing anything) i get
`failed to interpret function @stdlib::builtin::string::_isspace(stdlib::builtin::simd::SIMD[{uint8}, {1}])`
`non-constant inputs`

martinvuyk:
@laszlokindrat  I'm guessing the check_licence step does not recompile the stdlib so maybe push the fix anyway? (setting the table as var not alias)

YichengDWu:
Can you add more tests?

"""mojo
    var hello = String("السلام عليكم")
    var hello1 = String("Dobrý den")
    var hello2 = String("Hello")
    var hello3 = String("שָׁלוֹם")
    var hello4 = String("नमस्ते")
    var hello5 = String("こんにちは")
    var hello6 = String("안녕하세요")
    var hello7 = String("你好")
    var hello8 = String("Olá")
    var hello9 = String("Здравствуйте")
"""

laszlokindrat:
The license checker problem has been resolved, and it turns out the root cause has to do with the global lookup table. Global have many gotchas (very much not a stable feature ATM), and there are two that are relevant here: 1) functions using global `var` declarations cannot be used at compile time (this is what caused the license checker to fail); 2) global `alias` declaration don't actually result in a global (i.e. static) variable, and each user will have local allocations. So the short story is, we cannot do this in a sane way yet.

At the same time, I learned that we actually emit code for large if-elif-else statements that llvm can cleverly optimize either using arithmetic transformations or turning it a jump table. This is actually better than a LUT, since the efficacy of these optimizations have likely been evaluated properly, so we should just rely on them (more-or-less blindly) until we have benchmarking infra for the stdlib. You can see how I changed the `_isspace` implementation [here](https://github.com/modularml/mojo/commit/7b00f400a62b6e1401f003a2b5599814b7e66311).

So, could you please rebase on the latest nightly, and modify your PR accordingly?

martinvuyk:
> The license checker problem has been resolved, and it turns out the root cause has to do with the global lookup table. Globals have many gotchas (very much not a stable feature ATM), and there are two that are relevant here: 1) functions using global `var` declarations cannot be used at compile time (this is what caused the license checker to fail); 2) global `alias` declarations don't actually result in a global (i.e. static) variable, and each use can result in local allocations. So the short story is, we cannot do this in a sane way yet.
> 
> At the same time, I learned that we actually emit code for large if-elif-else statements that llvm can cleverly optimize either using arithmetic transformations or turning it a jump table. This is actually better than a LUT, since the efficacy of these optimizations have likely been evaluated properly, so we should just rely on them (more-or-less blindly) until we have benchmarking infra for the stdlib. You can see how I changed the `_isspace` implementation [here](https://github.com/modularml/mojo/commit/7b00f400a62b6e1401f003a2b5599814b7e66311).
> 
> So, could you please rebase on the latest nightly, and modify your PR accordingly?

will do, that looks interesting to read about. In the LLVM devs we trust

martinvuyk:
@laszlokindrat any idea what is going on here?

I'm currently changing the _isnewline to this:
```mojo
fn _isnewline_start(
    ptr: UnsafePointer[UInt8], read_ahead: Int = 1
) -> (Bool, Int):
    """Returns if the first item in the pointer is the start of
    a newline sequence, and its length.
    """
    alias `\n` = UInt8(ord("\n"))
    alias `\r` = UInt8(ord("\r"))
    alias `\f` = UInt8(ord("\f"))
    alias `\v` = UInt8(ord("\v"))
    alias `\x1c` = UInt8(ord("\x1c"))
    alias `\x1d` = UInt8(ord("\x1d"))
    alias `\x1e` = UInt8(ord("\x1e"))

    # TODO add line and paragraph separator as stringliteral
    # once unicode escape secuences are accepted
    var rn = "\r\n"
    var next_line = List[UInt8](0xC2, 0x85)
    """TODO: \\x85"""
    var unicode_line_sep = List[UInt8](0xE2, 0x80, 0xA8)
    """TODO: \\u2028"""
    var unicode_paragraph_sep = List[UInt8](0xE2, 0x80, 0xA9)
    """TODO: \\u2029"""

    @always_inline
    fn compare(
        item1: UnsafePointer[UInt8], item2: UnsafePointer[UInt8], amnt: Int
    ) -> Bool:
        var ptr1 = DTypePointer(item1)
        var ptr2 = DTypePointer(item2)
        return memcmp(ptr1, ptr2, amnt) == 0

    if read_ahead > 1:
        if compare(ptr, rn.unsafe_uint8_ptr(), 2):
            return True, 2

    var val = _utf8_byte_type(ptr[0])
    if val == 0:
        return (
            Bool(
                ptr[0] == `\n`
                or ptr[0] == `\r`
                or ptr[0] == `\f`
                or ptr[0] == `\v`
                or ptr[0] == `\x1c`
                or ptr[0] == `\x1d`
                or ptr[0] == `\x1e`
            ),
            1,
        )
    elif val == 2:
        print(compare(ptr, next_line.unsafe_ptr(), 2), 2)
        return compare(ptr, next_line.unsafe_ptr(), 2), 2
    elif val == 3:
        return (
            compare(ptr, unicode_line_sep.unsafe_ptr(), 3)
            or compare(ptr, unicode_paragraph_sep.unsafe_ptr(), 3)
        ), 3
    return False, 1
```

In the case of the separator `\x85`
```mojo
elif val == 2:
    print(compare(ptr, next_line.unsafe_ptr(), 2), 2) # prints `True 2`
    return compare(ptr, next_line.unsafe_ptr(), 2), 2
```
and its caller recieves 
```mojo
var res = _isnewline_start(ptr.offset(i), read_ahead)
print(hex(ptr.offset(i)[0]), res[0], res[1]) # prints `0xc2 False 2`
```

laszlokindrat:
@martinvuyk not sure, but please wait with `_isnewline` because I already fixed it's footgun (but it didn't make it into the last nightly). And if you do end up changing it, please do that in a subsequent patch.

martinvuyk:
@laszlokindrat I think it's good to go, added changelog description and updated the PR's as well.

docs/changelog.md
```diff
@@ -548,6 +548,9 @@ what we publish.
 - `String.split()` now defaults to whitespace and has pythonic behavior in that
   it removes all adjacent whitespaces by default.
 
+- `String` now has a `String.__iter__()` method which returns a `StringRef`
```
laszlokindrat, line 551: Is it possible to use `StringSlice` instead of `StringRef` for improved safety? We're hoping to remove most uses of `StringRef` in favor of safer alternatives.

martinvuyk:
@laszlokindrat didn't know that type existed, that also solves the problem of extending the string's lifetime inside the for loop

stdlib/test/builtin/test_string.mojo
```diff
@@ -1082,6 +1073,78 @@ def test_indexing():
...
+        "ेत्समन",
+        "はちにんこ",
+        "요세하녕안",
+        "好你",
+        "álO",
+        "етйувтсвардЗ",
+    )
+    var utf8_lengths = List(5, 12, 9, 5, 7, 6, 5, 5, 2, 3, 12)
+    var item_idx = 0
+    for item in items:
```
laszlokindrat, line 1132: We don't have `enumerate` yet, so it's better to iterate with the index
```suggestion
    for item_idx in range(len(items)):
    var item = items[item_idx]
```

stdlib/test/builtin/test_string.mojo
```diff
@@ -1082,6 +1073,78 @@ def test_indexing():
...
+        return c
+
+    assert_equal(123, atol(conc(vs)))
+
+    concat = String("")
+    for v in vs.__reversed__():
+        concat += v
+    assert_equal(321, atol(concat))
+
+    # TODO
```
laszlokindrat, line 1093: Is something blocking this?

stdlib/test/builtin/test_string.mojo
```diff
@@ -1082,6 +1073,78 @@ def test_indexing():
...
+        "はちにんこ",
+        "요세하녕안",
+        "好你",
+        "álO",
+        "етйувтсвардЗ",
+    )
+    var utf8_lengths = List(5, 12, 9, 5, 7, 6, 5, 5, 2, 3, 12)
+    var item_idx = 0
+    for item in items:
+        var amnt = 0
```
laszlokindrat, line 1133: What does `amnt` stand for?

stdlib/test/builtin/test_string.mojo
```diff
@@ -1082,6 +1073,78 @@ def test_indexing():
...
+        var byte_idx = 0
+        for v in item[]:
+            var byte_len = len(v)
+            assert_equal(item[][byte_idx : byte_idx + byte_len], v)
+            byte_idx += byte_len
+            amnt += 1
```
laszlokindrat, line 1139: This is cool, but can we somehow test this directly at least on one example? E.g. comparing the lengths of and iterating through and  `"mojo🔥"` and `["m", "o", "j", "o", "🔥"]` or something like this.

stdlib/src/builtin/string.mojo
```diff
@@ -660,6 +666,87 @@ fn isprintable(c: UInt8) -> Bool:
...
+    elif b < 0b1110_0000:
+        return 2  # is 2 byte long
+    elif b < 0b1111_0000:
+        return 3  # is 3 byte long
+    else:
+        return 4  # is 4 byte long
+
+
+@value
+struct _StringIter[
```
laszlokindrat, line 683: @ConnorGray Could you please take a look at this?

stdlib/src/builtin/string.mojo
```diff
@@ -660,6 +666,87 @@ fn isprintable(c: UInt8) -> Bool:
...
+        lifetime: The lifetime of the underlying string data.
+        forward: The iteration direction. `False` is backwards.
+    """
+
+    var index: Int
+    var continuation_bytes: Int
+    var ptr: UnsafePointer[UInt8]
+    var len: Int
+
+    fn __init__(inout self, unsafe_pointer: UnsafePointer[UInt8], len: Int):
```
laszlokindrat, line 701: nit: to avoid some nasty errors when someone tries to use `len(...)`
```suggestion
    fn __init__(inout self, unsafe_pointer: UnsafePointer[UInt8], length: Int):
```

stdlib/src/builtin/string.mojo
```diff
@@ -660,6 +666,87 @@ fn isprintable(c: UInt8) -> Bool:
...
+
+    fn __next__(inout self) -> StringSlice[is_mutable, lifetime]:
+        @parameter
+        if forward:
+            var byte_len = 1
+            if self.continuation_bytes > 0:
+                var value = _utf8_byte_type(int(self.ptr[self.index]))
+                if value != 0:
+                    byte_len = int(value)
+                    self.continuation_bytes -= int(value) - 1
```
laszlokindrat, line 721:
```suggestion
                    self.continuation_bytes -= byte_len - 1
```

stdlib/src/builtin/string.mojo
laszlokindrat, line 701: @ConnorGray did we decide to make these constructors take keyword-only `unsafe_pointer` arguments?
```suggestion
    fn __init__(inout self, *, unsafe_pointer: UnsafePointer[UInt8], len: Int):
```

stdlib/src/builtin/string.mojo
```diff
@@ -660,6 +666,87 @@ fn isprintable(c: UInt8) -> Bool:
...
+
+    fn __iter__(self) -> Self:
+        return self
+
+    fn __next__(inout self) -> StringSlice[is_mutable, lifetime]:
+        @parameter
+        if forward:
+            var byte_len = 1
+            if self.continuation_bytes > 0:
+                var value = _utf8_byte_type(int(self.ptr[self.index]))
```
laszlokindrat, line 718: nit: `value` is not a very meaningful variable name
```suggestion
                var byte_type = _utf8_byte_type(int(self.ptr[self.index]))
```

laszlokindrat (commented):
Generally LGTM, but I would like some more 👀  on this before we bring it in. @JoeLoser @ConnorGray

laszlokindrat:
Also, please rebase on the latest nightly.

stdlib/src/builtin/string.mojo
```diff
@@ -660,6 +666,87 @@ fn isprintable(c: UInt8) -> Bool:
...
+    if b < 0b1000_0000:
+        return 0
+    elif b < 0b1100_0000:
+        return 1  # is continuation byte
+    elif b < 0b1110_0000:
+        return 2  # is 2 byte long
+    elif b < 0b1111_0000:
+        return 3  # is 3 byte long
+    else:
+        return 4  # is 4 byte long
```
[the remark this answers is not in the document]
martinvuyk: > Could this be done by just counting the leading ones?

That seems like an interesting idea. I'm not sure which would be more performant though. I've googled a bit and it seems leading zeroes speed varies some by ISA. 

```mojo
from bit import countl_zero
return max(4, countl_zero(~b))
```
what I'm seeing here is 1 branch, 1 negation op (no idea what cost), countl_zero (better perf in RISC than x86, but both are fast) for all first bytes

```mojo
    if b < 0b1000_0000:
        return 0
    elif b < 0b1100_0000:
        return 1  # is continuation byte
    elif b < 0b1110_0000:
        return 2  # is 2 byte long
    elif b < 0b1111_0000:
        return 3  # is 3 byte long
    else:
        return 4  # is 4 byte long
```
most ASCII alpabets will do just 1 operation, comparing in the first if and returning 0. The second most common will be a continuation byte since it is 1 for 2 byte sequences, 2 for 3 byte and 3 for 4 byte. But I have no idea what the cost is of this code having 5 branches.

Maybe `max(4, countl_zero(~b))` could be rewritten as `(countl_zero(~b) ^ 0b100 - 0b100) & 0b0111_1111` any idea what perf diff this would make or am I overcomplicating this?

stdlib/test/builtin/test_string.mojo
martinvuyk: `UnsafePointer` does not have a `store` or `__setitem__` method

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
laszlokindrat: I'm not too worried about the perf implications and seems difficult to predict the real performance on a modern out-of-order machine. I suggest doing this mostly to make the code shorter and more expressive. But since you suggested using masks, a branchless solution is probably nice and I'm confident it will be performant enough:
```mojo
# note that both compilers and HW are really efficient at these bitwise logic
countl_zero(~(b & 0b1111_0000))
```

stdlib/src/utils/string_slice.mojo
```diff
@@ -116,6 +116,9 @@ struct StringSlice[
     fn __str__(self) -> String:
         return String(str_slice=self)
 
+    fn __len__(self) -> Int:
```
laszlokindrat, line 119: Could you please add a docstring and a test for this?

stdlib/src/utils/stringref.mojo
```diff
@@ -136,6 +136,18 @@ struct StringRef(
 
         return DTypePointer[DType.uint8](ptr)
 
+    @always_inline
+    fn __init__(inout self, owned ptr: UnsafePointer[UInt8], length: Int):
```
laszlokindrat, line 140: Q: Why is this needed? The `@value` decorator should generate a similar constructor, without `owned`.

stdlib/test/builtin/test_string.mojo
laszlokindrat: Thanks, please describe this in the TODO.

stdlib/src/builtin/string.mojo
```diff
@@ -1450,44 +1560,43 @@ struct String(
...
-            0x20, 0x5C, 0x75, 0x32, 0x30, 0x32, 0x39, 0
-        )
+        var unicode_paragraph_sep = List[UInt8](0xE2, 0x80, 0xA9)
         """TODO: \\u2029"""
 
         @always_inline
-        fn compare(item1: List[UInt8], item2: List[UInt8], amnt: Int) -> Bool:
-            var ptr1 = DTypePointer(item1.unsafe_ptr())
-            var ptr2 = DTypePointer(item2.unsafe_ptr())
+        fn compare(
```
laszlokindrat, line 1575: Q: should this be part of the public API? If not, could you please rename it to `_compare`. Otherwise, could you please add a docstring?

laszlokindrat (approved):
LGTM, thank you! I have a couple small things, and I would love a review from @ConnorGray before we bring this in.

stdlib/src/builtin/string.mojo
martinvuyk: I thought it was irrelevant since it's a helper func declared within the "private" scope of the main body. But ok I'll just but an _ in front

laszlokindrat: Oh I see, I thought for some reason that this was a function on `String` 🤦. It's fine either way, thanks!

laszlokindrat:
!sync

stdlib/test/builtin/test_string.mojo
```diff
@@ -1107,6 +1098,93 @@ def test_indexing():
...
+    assert_equal(123, atol(conc(vs)))
+
+    concat = String("")
+    for v in vs.__reversed__():
+        concat += v
+    assert_equal(321, atol(concat))
+
+    # TODO: UnsafePointer does not have a store or __setitem__ method
+    # for v in vs:
+    #     v.unsafe_ptr().store(0, "1")
```
ConnorGray, line 1120: **Suggestion** I think this could be written as `v.unsafe_ptr()[0] = "1"`.

`UnsafePointer` does define a `__getitem__` method, [seen here](https://github.com/modularml/mojo/blob/388cbdb8de2e174175f1b0ff8e88aa85da2d68c7/stdlib/src/memory/unsafe_pointer.mojo#L197-L199).

(However, that method was previously called `__refitem__`, and was only changed to `__getitem__` using new language features in the last few days; so if you searched for `__getitem__` that's why you may not have seen it 🙂.)

stdlib/src/utils/string_slice.mojo
```diff
@@ -116,6 +116,15 @@ struct StringSlice[
...
+        """Returns the StringSlice byte length.
```
ConnorGray, line 120: Same as above, for consistency with Python (and for `StringSlice` to be consistent with `String`) I think the `__len__` for a string should be its length in Unicode code points.

Similarly, we can promote / add a `byte_length()` method for use in cases where the encoded length is needed.

stdlib/src/builtin/string.mojo
```diff
@@ -1116,10 +1226,10 @@ struct String(
 
     @always_inline
     fn __len__(self) -> Int:
-        """Returns the string length.
+        """Returns the string byte length.
```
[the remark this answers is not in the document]
laszlokindrat: Good catch, thanks Connor! Let's be lazy and see if `_byte_length()` is required by anything other the string implementations before making it public.

laszlokindrat:
@martinvuyk Could you please address Connor's comments in a followup after the next nightly? Feel free to ping me for a review. Thanks!

ConnorGray:
Apologies for leaving comments after this already merged, I started reviewing without seeing this had already merged in the meantime! This generally looks good though, thank you for contributing this! 🙂 

The review comments are fairly minor and should be easy to roll into future work.

stdlib/src/builtin/string.mojo
```diff
@@ -658,6 +664,90 @@ fn isprintable(c: UInt8) -> Bool:
...
+
+    fn __init__(
+        inout self, *, unsafe_pointer: UnsafePointer[UInt8], length: Int
+    ):
+        self.index = 0 if forward else length
+        self.ptr = unsafe_pointer
+        self.length = length
+        self.continuation_bytes = 0
+        for i in range(length):
+            if _utf8_byte_type(int(unsafe_pointer[i])) == 1:
```
ConnorGray, line 708: **Question** Is this `int(..)` needed? `unsafe_pointer[i]` should return a `UInt8`, and it looks like `_utf8_byte_type` takes a `UInt8` argument. I might have missed something, or perhaps this is leftover from a previous iteration of this implementation?

martinvuyk: > leftover from a previous iteration of this implementation?

yep, I forgot that one

PR #2872, opened 2024-05-29
[Proposals] Add proposal for ref convention

nmsmith:
This is the proposal for the new ref conventions that Chris and I have been working on.


JoeLoser:
!sync

lattner (approved):

PR #2883, opened 2024-05-29
[stdlib] Fix `_isspace` use at compile time in licence checker

laszlokindrat:


laszlokindrat:
!sync

JoeLoser (approved):
Looks good, thanks for unbreaking the CI!  As discussed offline, we'll work to prevent this sort of issue in the future for nightlies.  Appreciate you digging into and fixing this!

PR #2884, opened 2024-05-29
[stdlib] Implement pythonic lcm function

bgreni:
Similar changes as #2777 but for the `lcm` function.

Suggested followup https://github.com/modularml/mojo/pull/2777#pullrequestreview-2082274419

### DIFF
--- a/stdlib/src/builtin/_math.mojo
+++ b/stdlib/src/builtin/_math.mojo
@@ -267,3 +267,74 @@ fn gcd(*values: Int) -> Int:
         if result == 1:
             return result
     return result
+
+
+# ===----------------------------------------------------------------------=== #
+# lcm
+# ===----------------------------------------------------------------------=== #
+
+
+fn lcm(owned m: Int, owned n: Int, /) -> Int:
+    """Computes the least common multiple of two integers.
+
+    Args:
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/_math.mojo
```diff
@@ -267,3 +267,71 @@ fn gcd(*values: Int) -> Int:
...
+    """Computes the least common multiple of two integers.
+
+    Args:
+        m: The first integer.
+        n: The second integer.
+
+    Returns:
+        The least common multiple of the two integers.
+    """
+    return abs((m // gcd(m, n)) * n) if m > n else abs((n // gcd(m, n)) * m)
```
laszlokindrat, line 287: Probably need to check for zero divisor. Please also add a test.
```suggestion
    if d: = gcd(m, n):
        return abs((m // d) * n if m > n else (n // d) * m)
    return 0 
```

laszlokindrat (approved):
Cool, thank you!

stdlib/src/builtin/_math.mojo
bgreni: ooooo walrus operator

laszlokindrat:
!sync

PR #2888, opened 2024-05-30
[stdlib] Add struct `UnsafeMaybeUninitialized`

gabrieldemarmiesse:
This struct is private at the moment and for internal use only. Once we are sure that it's ready for public use, we'll call it `MaybeUninitialized`. Heavily borrowed from https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html

This can allow us to add a destructor to `InlineArray` without impacting `InlineList` and SBO. We can make `InlineList` use a `InlineArray[MaybeUninitialized[T]]`.

@ConnorGray  for review since you have a good idea of what to do with it. 
See https://github.com/modularml/mojo/issues/2869#issuecomment-2137962617 for background.

Three things to note:
1) I didn't measure the performance impact of this struct compared to working directly with uninitialized memory. Even if there is a performance penalty right now, do we really care at this point since we don't have benchmarks yet? Even if the implementation could be faster, this PR gives us unit tests and a struct with method signatures to iterate on.
2) This struct is private, I didn't want to put too much pressure on this PR to have something perfect on the first try. Furthermore, we don't have good docstrings yet, those can be added later on.
3) I use 
```
`__mlir_type[`!pop.array<`, _size.value, `, `, Self.ElementType, `>`]`
```
but this may not be the best MLIR type to do this. I'm not very experienced with MLIR, so if there is something more appropriate that we can use, let me know.


stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -0,0 +1,77 @@
...
+    CollectionElement
+):
+    alias type = __mlir_type[
+        `!pop.array<`, _size.value, `, `, Self.ElementType, `>`
+    ]
+    var _array: Self.type
+
+    @always_inline
+    fn __init__(inout self):
+        """The mpemory is now considered uninitialized."""
```
gabrieldemarmiesse, line 28:
```suggestion
        """The memory is now considered uninitialized."""
```

stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -0,0 +1,77 @@
...
+    @always_inline
+    fn __moveinit__(inout self, owned other: Self):
+        """Calling this method assumes that the memory is initialized."""
+        self = Self()
+        move_pointee(src=other.unsafe_ptr(), dst=self.unsafe_ptr())
+
+    @always_inline
+    fn write(inout self, owned value: Self.ElementType):
+        """Calling this method assumes that the memory is uninitialized."""
+        self.unsafe_ptr()[] = value^
```
ConnorGray, line 52:
```suggestion
        self.unsafe_ptr().init_pointee_move(value^)
```

**Request** The `=` operator assumes that the LHS is initialized and will run the destructor on it, causing this to try to destroy uninitialized memory. `init_pointee_move()` assumes the pointee is uninitialized and so will not run the destructor.

We should add a test for this as well:

```mojo
var storage = MaybeUninitialized[ImpossibleToDestroy]()

# Should not abort
storage.write(ImpossibleToDestroy(5))
```

As a sanity check, please confirm that that test above fails with the current code 🙂

stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -0,0 +1,77 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+
+# TODO: Make this public when we are certain of the design.
+# TODO: Move _size into an alias when the bug https://github.com/modularml/mojo/issues/2889
+# is fixed.
+struct _MaybeUninitialized[ElementType: CollectionElement, _size: Int = 1](
```
ConnorGray, line 18: **Suggestion** I suggest we remove this `_size` parameter and always use a fixed-size of `1`. To model arrays of possibly-uninitialized elements, authors could use `InlineArray[MaybeUninitialized[T], size]`. That composes and generalizes a bit better I think that `MaybeUninitialized[T, size]`.

E.g. `List[MaybeUninitialized[T]]` would work just as well.

stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -0,0 +1,77 @@
...
+        self = Self()
+        move_pointee(src=other.unsafe_ptr(), dst=self.unsafe_ptr())
+
+    @always_inline
+    fn write(inout self, owned value: Self.ElementType):
+        """Calling this method assumes that the memory is uninitialized."""
+        self.unsafe_ptr()[] = value^
+
+    @always_inline
+    fn assume_initialized(
```
ConnorGray, line 55:
```suggestion
    fn unsafe_assume_initialized(
```
**Suggestion** Add `unsafe` to the name because `MaybeUninitialized` does not itself have `Unsafe` in its name.

stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -0,0 +1,77 @@
...
+        """Calling this method assumes that the memory is initialized."""
+        return self[].unsafe_ptr()[]
+
+    @always_inline
+    fn unsafe_ptr(self) -> UnsafePointer[Self.ElementType]:
+        """Get a pointer to the underlying element."""
+        return UnsafePointer(self._array).bitcast[Self.ElementType]()
+
+    @always_inline
+    fn assume_initialized_destroy(inout self):
```
ConnorGray, line 67:
```suggestion
    fn unsafe_assume_initialized_destroy(inout self):
```

Same suggestion as above.

stdlib/src/memory/maybe_uninitialized.mojo
ConnorGray, line 18:
```suggestion
struct MaybeUninitialized[ElementType: CollectionElement](
```

I think we can call this `MaybeUninitialized`. I think we we will want something very similar to this, and we don't have backwards compatibility to worry about yet in the stdlib. So lets avoid the churn🙂

ConnorGray (approved):
Thank you for putting this together @gabrieldemarmiesse, this will be very useful!

I left a few minor comments for things we should improve. Once they're fixed I'll go ahead and merge this.

ConnorGray:
!sync

stdlib/src/memory/maybe_uninitialized.mojo
gabrieldemarmiesse: There is already this test in my unit test file, see line 60. And this works correctly:
```mojo
def test_write_does_not_trigger_destructor():
    var a = _MaybeUninitialized[ImpossibleToDestroy]()
    a.write(ImpossibleToDestroy(42))
```
So unless I'm missing something no change is needed here. I do agree it's strange but it works.

stdlib/src/memory/maybe_uninitialized.mojo
gabrieldemarmiesse: See the comment just above, line 17. The `_size` here is not meant to be modified and we indeed wish to make it fixed. But we can't because of a bug. The issue has already been opened about this and the link provided in the comment.

stdlib/src/memory/maybe_uninitialized.mojo
gabrieldemarmiesse: Since all methods of `MaybeUninitialized` are unsafe (like all the methods of `UnsafePointer`), maybe let's change the struct name to `UnsafeMaybeUninitialized`. What do you think?

stdlib/src/memory/maybe_uninitialized.mojo
gabrieldemarmiesse: Same as above. Can we rename the struct instead?

stdlib/src/memory/maybe_uninitialized.mojo
gabrieldemarmiesse: Great to know, I'll change it.

gabrieldemarmiesse:
@ConnorGray  here is a summary of the changes:
* Used the latest syntax for lifecycle manipulation on pointers
* Renamed `_MaybeUninitialized` -> `UnsafeMaybeUninitialized`
* Provided explicit move and copy operation as the caller is responsible for each step of the lifecycle. `__moveinit__` and `__copyinit__` are still present, but just as no-op to avoid footguns. Maybe we should add an "abort" in them to make sure they are never executed?
* Didn't change the assignation and the `_size` for the reasons mentionned above.

stdlib/src/memory/maybe_uninitialized.mojo
ConnorGray: Understood. I wonder if if we use a literal `1` in the `pop.array` type, does that run into the same bug?

stdlib/src/memory/maybe_uninitialized.mojo
ConnorGray: That makes sense, I agree with the rename of the type itself.

stdlib/src/memory/maybe_uninitialized.mojo
ConnorGray: Something strange must be happening. I tried this simplified example which demonstrates that `ptr[] = ImpossibleToDestroy(5)` does run a destructor:

```mojo
@value
struct ImpossibleToDestroy(Movable):
    var value: Int

    fn __del__(owned self):
        abort("ABORT")


fn main():
    var ptr = UnsafePointer[ImpossibleToDestroy].alloc(1)

    ptr[] = ImpossibleToDestroy(5)
```

So if that's not in this case, that indicates `ImpossibleToDestroy` used with `UnsafeMaybeUninitialized` is not behaving as we expect. It could be a compiler bug or something else.

Do you think you might be able to work to create a minimal reproduction of the bug?

ConnorGray:
!sync

stdlib/src/memory/maybe_uninitialized.mojo
gabrieldemarmiesse: Yep. Opened an issue here: https://github.com/modularml/mojo/issues/2951

ConnorGray:
!sync

stdlib/src/memory/maybe_uninitialized.mojo
ConnorGray: Thank you Gabriel! I'll resolve this thread.

ConnorGray: For what its worth, I'm hypothesizing that because the value is initialized to `kgen.undef`, its undefined behavior for the compiler to run the destructor, so in the spirit of "UB lets the compiler do whatever it wants", the compiler (arguably correctly) deduces it should not run a destructor on uninitialized memory and removes it.

PR #2894, opened 2024-05-30
[stdlib] Add SliceNew

bgreni:
Subset of changes from #2495

Introduce `SliceNew` with corrected behaviour when using negative step values.
Currently internal code relies on the old incorrect `Slice` behaviour, so we must repair
this feature incrementally.


tjdwill:
Hi, I'm here from #2902; thank you for informing of this thread. I see that this is an issue you've been working on, and I have a few questions about the proposed path forward (and some general programming questions if you don't mind):

0. From my understanding, the plan is to create a `SliceNew` type to incrementally replace the current `Slice`. The current `Slice` will eventually phased out, and, ultimately, `SliceNew` will be renamed to `Slice`. Is that correct?
1. Are the `T._adjust_span` methods found in `List` and `String` also being phased out completely?
2. General understanding question: when reading over `SliceNew.indices`, I noticed the following clause for `start` and `end`:

```mojo
 elif start.value() < 0:
            start = start.value() + src_len
            if start.value() < 0:
                start = 0 if positive_step else -1
``` 

Is the inner `if` clause checking for an overflow that would cause the start value to become negative after reassignment? Also, I'm still learning to reason with Options, so if I were to call `.value` on an Option that has no value, what would return?

I tend to ask a lot of questions in the process of learning, but please let me know if this is the wrong place.

bgreni:
@tjdwill 
> I tend to ask a lot of questions in the process of learning, but please let me know if this is the wrong place.

Of course, question are always good! I'll be happy to clarify.

> 0. From my understanding, the plan is to create a `SliceNew` type to incrementally replace the current `Slice`. The current `Slice` will eventually phased out, and, ultimately, `SliceNew` will be renamed to `Slice`. Is that correct?

Correct.

> 1. Are the `T._adjust_span` methods found in `List` and `String` also being phased out completely?

Eventually yes in favour of the new `Slice.indices` semantics that align with how python works.

> 2. General understanding question: when reading over `SliceNew.indices`, I noticed the following clause for `start` and `end`:
> 
> ```mojo
>  elif start.value() < 0:
>             start = start.value() + src_len
>             if start.value() < 0:
>                 start = 0 if positive_step else -1
> ```
> 
> Is the inner `if` clause checking for an overflow that would cause the start value to become negative after reassignment? Also, I'm still learning to reason with Options, so if I were to call `.value` on an Option that has no value, what would return?
> 

Calling `value` on an empty optional would result in a crash, similar to dereferencing a null pointer. The responsibility is on the program to always handle the cases where an `Optional` is `None`. Here we handle this by checking and given them a default value if they are empty.

```mojo
if not start:
            start = 0 if positive_step else src_len - 1
```

The inner if is to match the current Python behaviour where any very large negative values will be defaulted to either 0 or -1 to prevent out of bounds accesses.
```python
slice(1, -50).indices(4) # returns (1, 0, 1)
```

tjdwill:
Got it, so it's basically going "if the value is still negative after the initial adjustment, set it to 0 or -1 according to the sense of the step". Thank you for explaining that as well as clarifying Option.

laszlokindrat (approved):
Thank you for doing this, not a smooth ride, but highly appreciated!

laszlokindrat:
!sync

laszlokindrat:
In the next patch, could you please try phasing it in to the open source lib?

stdlib/src/builtin/range.mojo
```diff
@@ -179,9 +179,10 @@ struct _StridedRange(Sized, ReversibleRange, _StridedIterable):
 
     @always_inline("nodebug")
     fn __len__(self) -> Int:
-        # FIXME(#38392)
-        # if (self.step > 0) == (self.start > self.end):
-        #     return 0
+        if (self.step > 0 and self.start > self.end) or (
```
laszlokindrat, line 182: I think this was leftover accidentally. Could you please put it in a new  patch?

PR #2897, opened 2024-05-30
[stdlib] Fix when sep is suffix is `String.split`

### LINKED ISSUE #2879: [BUG] [stdlib] Unexpected behaviour by `String`'s `split` method when `sep` is a suffix
### Bug description

Splitting strings when the separator is a suffix of the string and `maxsplit` is less than the total number of occurrences of `sep` in the string results in incorrect output. These cases produce an extra empty string as the final element.
E.g. `String("abbaaaabbba").split("a", maxsplit=5)` incorrectly results in `['', 'bb', '', '', '', 'bbba', '']` instead of `['', 'bb', '', '', '', 'bbba']`.

### Steps to reproduce

MRE:

```mojo
String("a").split("a", 0)
```

results in `['a', '']` 
(expected `['a']`)

### System information

```shell
- OS: Debian (WSL)
- Mojo version: mojo 2024.5.2912 (6861a424)
- Modular CLI version: modular 0.8.0 (39a426b5)
```

bgreni:
Fixes #2879

Fix the case in `String.split` where the `sep` is the string suffix, but `maxsplit` has been reached

### DIFF
--- a/stdlib/test/builtin/test_string.mojo
+++ b/stdlib/test/builtin/test_string.mojo
@@ -699,6 +699,29 @@ fn test_split() raises:
     assert_equal(res4[0], "he")
     assert_equal(res4[1], "o")
 
+    # related to #2879
+    # TODO: replace string comparison when __eq__ is implemented for List
+    assert_equal(
+        String("abbaaaabbba").split("a").__str__(),
+        "['', 'bb', '', '', '', 'bbb', '']",
+    )
+    assert_equal(
+        String("abbaaaabbba").split("a", 8).__str__(),
+        "['', 'bb', '', '', '', 'bbb', '']",
+    )
+    assert_equal(
+        String("abbaaaabbba").split("a", 5).__str__(),
+        "['', 'bb', '', '', '', 'bbba']",
+    )
+    assert_equal(String("aaa").split("a", 0).__str__(), "['aaa']")
+    assert_equal(String("a").split("a").__str__(), "['', '']")
+    assert_equal(String("1,2,3").split("3", 0).__str__(), "['1,2,3']")
+    assert_equal(String("1,2,3").split("3", 1).__str__(), "['1,2,', '']")
+    assert_equal(String("1,2,3,3").split("3", 2).__str__(), "['1,2,', ',', '']")
+    assert_equal(
+        String("1,2,3,3,3").split("3", 2).__str__(), "['1,2,', ',', ',3']"
+    )
+
 
 fn test_splitlines() raises:
     # Test with no line breaks
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -1631,7 +1631,7 @@ struct String(
             output.append(self[lhs:rhs])
             lhs = rhs + sep_len
 
-        if self.endswith(sep):
+        if self.endswith(sep) and (len(output) <= maxsplit or maxsplit == -1):
             output.append("")
         return output
 


stdlib/test/builtin/test_string.mojo
```diff
@@ -683,6 +683,22 @@ fn test_split() raises:
     assert_equal(res4[0], "he")
     assert_equal(res4[1], "o")
 
+    # related to #2879
+    assert_equal(
+        String("abbaaaabbba").split("a").__str__(),
```
martinvuyk, line 688:
```suggestion
        str(String("abbaaaabbba").split("a")),
```

stdlib/test/builtin/test_string.mojo
```diff
@@ -683,6 +683,22 @@ fn test_split() raises:
...
+        "['', 'bb', '', '', '', 'bbb', '']",
+    )
+    assert_equal(
+        String("abbaaaabbba").split("a", 8).__str__(),
```
martinvuyk, line 692:
```suggestion
        str(String("abbaaaabbba").split("a", 8)),
```

stdlib/test/builtin/test_string.mojo
```diff
@@ -683,6 +683,22 @@ fn test_split() raises:
...
+        "['', 'bb', '', '', '', 'bbb', '']",
+    )
+    assert_equal(
+        String("abbaaaabbba").split("a", 5).__str__(),
```
martinvuyk, line 696:
```suggestion
        str(String("abbaaaabbba").split("a", 5)),
```

stdlib/test/builtin/test_string.mojo
```diff
@@ -683,6 +683,22 @@ fn test_split() raises:
...
+        "['', 'bb', '', '', '', 'bbba']",
+    )
+    assert_equal(String("aaa").split("a", 0).__str__(), "['aaa']")
```
martinvuyk, line 699:
```suggestion
    assert_equal(str(String("aaa").split("a", 0)), "['aaa']")
```

stdlib/test/builtin/test_string.mojo
```diff
@@ -683,6 +683,22 @@ fn test_split() raises:
...
+    assert_equal(String("a").split("a").__str__(), "['', '']")
```
martinvuyk, line 700:
```suggestion
    assert_equal(str(String("a").split("a")), "['', '']")
```

martinvuyk:
just some nits to be a little more pythonic. Great catch btw

martinvuyk (approved):

stdlib/test/builtin/test_string.mojo
bgreni: `str` doesn't work with `List` yet :(

martinvuyk: oh had no idea srry

JoeLoser (approved):
Nice fix! Just a minor suggestion in the tests.  Happy to merge after that.  Thank you!

stdlib/test/builtin/test_string.mojo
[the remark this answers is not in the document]
bgreni: `str()` still doesn't work with `List` yet

JoeLoser:
!sync

jdiggins:
Hey Team, I have a couple more test cases to consider here:

If maxsplit is 0, we should not append "" to end of output
```
d = String("1,2,3").split("3", 0)
assert_true(len(d) == 1)
assert_true(d[0] == "1,2,3")
```

If our split count == max count, and our sep occurs at the end, we still want to append "" to end of output
```
d = String("1,2,3").split("3", 1)
assert_true(len(d) == 2)
assert_true(d[0] == "1,2," and d[1] == "")
d = String("1,2,3,3").split("3", 2)
assert_true(len(d) == 3)
assert_true(d[0] == "1,2," and d[1] == "," and d[2] == "")
d = String("1,2,3,3,3").split("3", 2)
assert_true(len(d) == 3)
assert_true(d[0] == "1,2," and d[1] == "," and d[2] == ",3")
```

The tricky part with the current implementation is that `items` doesn't always equal the number of splits that occurred, but there are cases where it can. Something like this should get us closer to behaving as expected:
```
diff --git a/stdlib/src/builtin/string.mojo b/stdlib/src/builtin/string.mojo
index 5cb6fc53..8231dddb 100644
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -1646,8 +1646,9 @@ struct String(
             output.append(self[lhs:rhs])
             lhs = rhs + sep_len
 
-        if self.endswith(sep):
+        if self.endswith(sep) and (len(output) <= maxsplit or maxsplit == -1):
             output.append("")
+

```

JoeLoser:
!sync

Issue #2869, opened 2024-05-28
[stdlib] [BUG] `InlineArray` does not invoke the destructors of its elements


JoeLoser:
`InlineArray` contains values of `AnyType` and copies on construction but does not invoke the destructors of its elements.  We need to fix this.

GeauxEric:
Was attempting on this issue, but I found the `ElementType` is actually `CollectionElement`, not `AnyType` in the latest nightly branch: https://github.com/modularml/mojo/blob/nightly/stdlib/src/utils/static_tuple.mojo#L250

Am I missing something here?

rd4com:
@JoeLoser  it seem like an important one:
`InlineList` uses it and possibly more types,
it might also be blocking `SBO` by @gabrieldemarmiesse.
I'll put this issue on the chat to see if somebody can work on it!

rd4com:
#### Challenge: *values need to be initialized, before calling destructor on the elements*

This `__init__` makes it possible to create an array with uninitialized elements:
 ```mojo
   fn __init__(inout self, *, unsafe_uninitialized: Bool):
  ```

maybe `unsafe_uninitialized` need to be a parameter of `InlineArray`, with a default value of `False` ?

We could then do `@parameter if not unsafe_uninitialized: destroy_pointee all elements`

&nbsp;

How to be sure of which elements are initialized, if an user create a `unsafe_uninitialized` instance,

and initialize only some elements with `UnsafePointer` and `initialize_pointee_*`:

- should user free them all manually with a `DestroyElement` method ?

&nbsp;


Possible incomplete implementation:
```mojo
fn __del__(owned self):
    @parameter
    if not unsafe_uninitialized:
        @parameter
        for i in range(size):
            destroy_pointee(self.unsafe_ptr().offset(i))

```

gabrieldemarmiesse:
Some more information:
* `InlineList` is not affected by this bug because when I coded it I assumed that the InlineArray wasn't running destructors, thus I did It myself: https://github.com/modularml/mojo/blob/nightly/stdlib/src/collections/inline_list.mojo#L155 
* `InlineList` wants to manage itself the destructor of its elements because it does not want to run the destructor of all elements when the array is not completely filled. So just adding a `destroy_pointee` in the `__del__` is not a good solution, we should do better.
* `Inlinelist` sees `InlineArray` like a pointer. It's a piece of memory that is not aware of things like destructors.
[...]

PR #2899, opened 2024-05-31
[stdlib] Add __copyinit__ & __moveinit__ for InlineList

northstreet12:
Do we need `__copyinit__` & `__moveinit__` for InlineList?

### DIFF
--- a/stdlib/src/collections/inline_list.mojo
+++ b/stdlib/src/collections/inline_list.mojo
@@ -72,7 +72,7 @@ struct _InlineListIter[
 
 
 # TODO: Provide a smarter default for the capacity.
-struct InlineList[ElementType: CollectionElement, capacity: Int = 16](Sized):
+struct InlineList[ElementType: CollectionElement, capacity: Int = 16](Sized, CollectionElement):
     """A list allocated on the stack with a maximum size known at compile time.
 
     It is backed by an `InlineArray` and an `Int` to represent the size.
--- a/stdlib/src/collections/inline_list.mojo
+++ b/stdlib/src/collections/inline_list.mojo
@@ -111,6 +111,26 @@ struct InlineList[ElementType: CollectionElement, capacity: Int = 16](Sized):
         for value in values:
             self.append(value[])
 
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/inline_list.mojo
```diff
@@ -111,6 +111,26 @@ struct InlineList[ElementType: CollectionElement, capacity: Int = 16](Sized):
         for value in values:
             self.append(value[])
 
+    fn __copyinit__(inout self, existing: Self):
+        """Creates a deepcopy of the given list.
```
gabrieldemarmiesse, line 115:
```suggestion
        """Creates a copy of the given list.
```
The fact that it's deep or not is up to the underlying data type. If you store shared references in the InlineList, then it's not going to be a deep copy. I'm not sure there is a notion of deepcopy in Mojo, there is just copy.

gabrieldemarmiesse (commented):
This PR looks good, but to avoid suprises I would advise that we wait for the InlineArray semantics to be sorted out first. Let's not build too much on top of shaky fundations. You can follow the progress here: https://github.com/modularml/mojo/issues/2869

If you wish to, you can also add the trait `CollectionElement` to the `InlineList` in this PR

stdlib/src/collections/inline_list.mojo
northstreet12: Agreed. Updated the patch.

northstreet12:
> This PR looks good, but to avoid suprises I would advise that we wait for the InlineArray semantics to be sorted out first. Let's not build too much on top of shaky fundations. You can follow the progress here: #2869
> 
> If you wish to, you can also add the trait `CollectionElement` to the `InlineList` in this PR

OK, I'll follow https://github.com/modularml/mojo/issues/2869. Thanks.

JoeLoser:
@ConnorGray is this still blocked or can we move forward with it?

JoeLoser:
`InlineList` has been removed, so going this PR.  Thank you though!

PR #2902, opened 2024-05-31
[stdlib] Proposed Fix for String and List Slicing

tjdwill:
Identified and fixed discrepancy between Python and Mojo slicing behavior. Full details below:

# Mojo Pull Request Write-up

## Synoptic Changelog

In `stdlib`:

1. src/builtin/range.mojo
    - Adjusted `_StridedRange.__len__` definition to return 0 as needed.

2. src/builtin/string.mojo
    - Fix `String._adjust_span`: Adjust index calculations
    - Changed to call Slice._adjust_span
    - Removed `String._adjust_span` method.
    
3. src/builtin/builtin_slice.mojo
    - Changed `__init__` method to account for `NoneType` `Slice.start` value.
    - Added `Slice._adjust_span` method to replace those in `String` and `List`
    - Added a `Slice._has_start` method to match `Slice._has_end`
    - Adjusted `__str__` to account for slices of form `a[::step]`.
    
4. test/builtin/test_string.mojo
    - Changed `test_string_indexing` to add additional test cases; changed current test cases to reflect Python behavior.

5. test/builtin/test_range.mojo
    - Re-enabled edge case indexing tests.

6. test/builtin/test_slice.mojo
    - Added `test_has_start`
    - Added more cases in `test_slice_stringable`

7. test/collections/test_list.mojo
    - Added tests for list slicing that matches Python behavior

## Introduction

I am submitting this pull request in attempt to fix what I think is a bug in the slicing behavior of both strings and lists. I may be incorrect, but comparing the results of slicing operations between Python and Mojo reveals a divergence. My assumption in undertaking this pull request is that Mojo's Python emulation must be one-to-one.

## Background

While studying the source code, I came across the `String._adjust_span` method:

```mojo
@always_inline
fn _adjust_span(self, span: Slice) -> Slice:
    """Adjusts the span based on the string length."""
    var adjusted_span = span

    if adjusted_span.start < 0:
        adjusted_span.start = len(self) + adjusted_span.start

    if not adjusted_span._has_end():
        adjusted_span.end = len(self)
    elif adjusted_span.end < 0:
        adjusted_span.end = len(self) + adjusted_span.end

    if span.step < 0:
        var tmp = adjusted_span.end
        adjusted_span.end = adjusted_span.start - 1
        adjusted_span.start = tmp - 1

    return adjusted_span
```

From my understanding, the purpose of this method is to calculate the proper indices for the `String.__getitem__` method. The end and start adjustments in the 'step' conditional caught my eye and led to experimentation in both Python and Mojo to compare the behaviors of the two:

Python:

```python

>> s = "Hello Mojo!!"
>> s[:3:-1]
'!!ojoM o'
>> # Example 2
>> s[0:3:-1]
''
```

Mojo:
```mojo
    1> var s: String = "Hello Mojo!!"
    2. print(s[:3:-1])
leH
    3> # Example 2
    4. print(s[0:3:-1])
leH
```

There are more examples, but I think this is sufficient to show divergence in results.

## Investigating the Divergence

The divergence in behavior is the result of two reasons from what I could see. First, the calculation of indices does not match the behavior of Python for negative steps nor in some cases where `None` values for 'start' and/or 'end'. The second reason is that `_StridedRange` does not give the correct length value when compared to Python's behavior for `range` objects. 

### How it works currently

`String._adjust_span` does a decrement and swap of both 'start' and 'end' every
time 'step' is negative. So, for a slice such as `s[3:0:-1]`, for example, the
resulting slice object would be described `Slice(-1, 2, -1)`. In the `String.__getitem__` function, this slice has its length taken to determine the number of iterations in the `for` loop using `Slice.unsafe_indices` (nightly). This method is defined as:

```mojo
@always_inline
fn unsafe_indices(self) -> Int:
    """Return the length of the slice.

    Only use this function if start/end is guaranteed to be not None.

    Returns:
        The length of the slice.
    """

    return len(range(self.start, self.end, self.step))
```

It calls on the length function for the underlying object produced by `range`. Given the call signature, this object is of type `_StridedRange`, whose `__len__` function is defined as:

```mojo
# FIXME comment removed for readability
@always_inline("nodebug")
    fn __len__(self) -> Int:
        return _div_ceil_positive(abs(self.start - self.end), abs(self.step))
```

Calling this function on the example slice `Slice(-1, 2, -1)` generated from `String._adjust_span` results in a length of 3 when in Python, the length would be 0. This slice is then used for iterating over the String:

```mojo
...
var buffer = Self._buffer_type()
buffer.resize(adjusted_span_len + 1, 0)
var ptr = self.unsafe_uint8_ptr()
for i in range(adjusted_span_len):
    buffer[i] = ptr[adjusted_span[i]]
    # i=0: buffer[0] = ptr[(-1 + 0*-1)] = ptr[-1]
    # i=1: buffer[1] = ptr[(-1 + 1*-1)] = ptr[-2]
    # i=2: buffer[2] = ptr[(-1 + 2*-1)] = ptr[-3]
...
```

We can see that the created pointer is indexed using the retrieved index from the Slice object. So, given the length of the Slice object is 3 in the current implementation, the resulting buffer is filled with data. However, due to the indices, the pointer is referenced using negative indices. Since the pointer is an `DTypePointer[DType.int8]`, I think this would read the wrong data, but I'm not too sure on that part.

In any case, I don't think this is the desired behavior.

## Fixes

First, I fixed the length calculation of `_StridedRange` to account for the cases where zero should be returned. Given some slice object, span, these conditions are:

- span.step < 0 and span.start < span.end
- span.step > 0 and span.start > span.end

Otherwise, the length is calculated using the div ceil method as before.

Next was fixing the index calculations in `String._adjust_span`. To do so, I introduced additional logic in the relevant `Slice.__init__`. Essentially, similar logic that is used when `Slice(start: T0, end: T1=NoneType, step: T2)` is applied when `T0` is `NoneType`. `Slice._has_start` was added to match `Slice._has_end`, and `Slice.__str__` was adjusted accordingly.

Since `List` also has an `_adjust_span` method that identical to `String._adjust_span`, I added a new `Slice._adjust_span[T: Sized](object: T, span: Self) -> Self` static method. Both `List` and `String` had their respective `_adjust_span` method removed and now their `__getitem__` methods use `Slice._adjust_span`. This way, changes in that method are applied to both types. Also, it seemed right to have `Slice` produce additional `Slice` objects rather than `List` and `String`.

Tests were also modified to expect Python-like behavior. Currently all modified tests pass.

## Conclusion

In this pull request, I aim to rectify a perceived discrepancy between slicing behaviors of Mojo `String`. In the process, I adjusted the length calculation of `_StridedRange`, made multiple modifications to `Slice`, and adjusted `List`. Relevant tests were changed and added to, and all relevant tests pass. I hope the above report explains my thought process cogently.

### DIFF
--- a/stdlib/src/builtin/builtin_slice.mojo
+++ b/stdlib/src/builtin/builtin_slice.mojo
@@ -90,9 +90,12 @@ struct Slice(Stringable, EqualityComparable):
             end: The end value.
             step: The step value.
         """
-        self.start = _default_or(start, 0)
-        self.end = _default_or(end, _int_max_value())
         self.step = _default_or(step, 1)
+        if self.step < 0:
+            self.start = _default_or(start, _int_max_value())
+        else:
+            self.start = _default_or(start, 0)
+        self.end = _default_or(end, _int_max_value())
 
     fn __str__(self) -> String:
         """Gets the string representation of the span.
@@ -100,7 +103,10 @@ struct Slice(Stringable, EqualityComparable):
         Returns:
             The string representation of the span.
         """
-        var res = str(self.start)
+
+        var res = String()
+        if self._has_start():
+            res += str(self.start)
         res += ":"
         if self._has_end():
             res += str(self.end)
@@ -163,8 +169,37 @@ struct Slice(Stringable, EqualityComparable):
         return self.start + index(idx) * self.step
 
     @always_inline("nodebug")
+    fn _has_start(self) -> Bool:
+        return self.start != _int_max_value()
+
+    @always_inline("nodebug")
     fn _has_end(self) -> Bool:
         return self.end != _int_max_value()
+
+    @staticmethod
+    @always_inline("nodebug")
+    fn _adjust_span[T: Sized](object: T, span: Self) -> Self:
+        """Adjusts the provided slice based on the length of the object"""
+
+        var adjusted_span = span
+         
+        if adjusted_span.start < 0:
+            adjusted_span.start = len(object) + adjusted_span.start
+        if adjusted_span.end < 0:
+            adjusted_span.end = len(object) + adjusted_span.end
+
+        if adjusted_span.step < 0:
+            if not adjusted_span._has_start():
+                adjusted_span.start = len(object) - 1
+            if not adjusted_span._has_end():
+                adjusted_span.end = -1
+        else:
+            if not adjusted_span._has_start():
+                adjusted_span.start = 0
+            if not adjusted_span._has_end():
+                adjusted_span.end = len(object)
+        
+        return adjusted_span
 
 
 @always_inline("nodebug")
--- a/stdlib/src/builtin/range.mojo
+++ b/stdlib/src/builtin/range.mojo
@@ -179,10 +179,11 @@ struct _StridedRange(Sized, ReversibleRange, _StridedIterable):
 
     @always_inline("nodebug")
     fn __len__(self) -> Int:
-        # FIXME(#38392)
-        # if (self.step > 0) == (self.start > self.end):
-        #     return 0
-        return _div_ceil_positive(abs(self.start - self.end), abs(self.step))
+        if (self.step > 0 and self.start > self.end) or 
+            (self.step < 0 and self.start < self.end):
+            return 0
+        else:
+            return _div_ceil_positive(abs(self.start - self.end), abs(self.step))
 
     @always_inline("nodebug")
     fn __getitem__(self, idx: Int) -> Int:
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -927,7 +927,7 @@ struct String(
             A new string containing the string at the specified positions.
         """
 
-        var adjusted_span = self._adjust_span(span)
+        var adjusted_span = Slice._adjust_span(self, span)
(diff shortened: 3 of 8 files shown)


bgreni:
Please see this discussion https://github.com/modularml/mojo/pull/2495#issuecomment-2139256548

I've been working on a fix for this for a while, but unfortunately due to some internal code depending on the incorrect behaviour, it cannot be fixed all at once. Resulting in us needing to first introducing a new slice type (#2894) which the team can use to incrementally resolve the problematic internal code and we can go from there.

tjdwill:
> Please see this discussion [#2495 (comment)](https://github.com/modularml/mojo/pull/2495#issuecomment-2139256548)

Hello, 

Thank you for informing me of this. I'm not sure what to do in this scenario, but I do have a few questions based on what you've shared. I'll ask in the other pull request thread.

laszlokindrat:
Thanks for the contribution! As @bgreni pointed out, we already have a plan for fixing this and some patches in flight. It would be aweseome if you could review these, since you have clearly thought about this problem!

tjdwill:
> Thanks for the contribution! As @bgreni pointed out, we already have a plan for fixing this and some patches in flight. It would be aweseome if you could review these, since you have clearly thought about this problem!

Absolutely, I'll be sure to keep reading and learning. Thank you!

PR #2905, opened 2024-05-31
[stdlib] Remove attribute `self._reserved` from `Dict`

gabrieldemarmiesse:
### Rationale:
Our Dict implementation is too complicated for what it does. We have to keep `self._reserved` updated so that it's always equal to `len(self._entries)`. That's 1) taking memory (64 bits) for nothing 2) is a potential source of bugs as `self._reserved` could become out of sync with `len(self._entries)` in future code changes.

We could use `len(self._entries)` everywhere, but from an algorithmic point of view, I guess it's also easier to read `self._reserved`.

Since we don't have `@property` available in Mojo yet, I settled with 
```mojo
fn _reserved(self) -> Int:
     return len(self._entries)
```

### Using all the capacity available in the List
I also added another optimization here, to make the dict compatible in the future with small buffer optimization:
If we ask `self._entries` to allocate for N elements, but the List allocator tell us that now the capacity of the list is M, then we use M as the size of the list (and also the capacity). 

Concretely, let's say that we have small buffer optimization and that we ask for a list with a capacity of 8, it may be possible that the list has a small buffer of size 16, thus giving us a list of capacity 16 (the minimum) when we asked 8. Then we can use the 16 slots available. This is also good in case we later on allow Dict to grab another structure's pointer/list, as we'll use all the memory available.

If the second optimization is not a good idea, I can remove it. There might be some requirement (always be a power of 2 for example) that I'm not aware of and in this case we shouldn't let `List` decide of the capacity of our Dict.


stdlib/src/collections/dict.mojo
```diff
@@ -444,9 +445,14 @@ struct Dict[K: KeyElement, V: CollectionElement](
         """Initialize an empty dictiontary."""
         self.size = 0
         self._n_entries = 0
-        self._reserved = Self._initial_reservation
-        self._index = _DictIndex(self._reserved)
-        self._entries = Self._new_entries(self._reserved)
+        self._entries = Self._new_entries(Self._initial_reservation)
+        self._index = _DictIndex(len(self._entries))
```
soraros, line 449: Nit: I guess it makes more sense to use `Self._initial_reservation` here to make clear that it's compile time constant?

gabrieldemarmiesse: `len(self._entries)` can be greater than or equal to `Self._initial_reservation`, but they're not equal in all cases. See the section "Using all the capacity available in the List" of the pull request description

soraros: My point exactly: should we leave that to `List` at all? Your call anyways, shouldn't be blocking.

ConnorGray:
!sync

ConnorGray:
!sync

ConnorGray:
!sync

PR #2906, opened 2024-05-31
[stdlib] Add `Quaternion` and `DualQuaternion` to the complex package

martinvuyk:
Given that the complex package already exists I just put it in an identically named folder. 

Two structs were added to allow for some more complex coordinate systems to be included in the stdlib, which allow for some very efficient computations.

`Quaternion` is the most used in real life applications. `DualQuaternion` is used for advanced kinematics software because of the speedup they provide over matrices. 

Only basic tests were included because of how extensive hand calculations can be. 

`Quaternion`: 
  - A structure often used to represent rotations. Allocated on the stack with very efficient vectorized operations.

`DualQuaternion`: 
  - Structure nascently used to represent 3D transformations and rigid body kinematics. Allocated on the stack with very efficient vectorized operations.


martinvuyk:
@laszlokindrat this is the same error that is happening in another PR for me when I try to add tests for a module that didn't exist before, is there something hardcoded? Happens locally as well, but it fails saying 
`unable to locate module 'testing'`
`unable to locate module 'complex'`
`could not find an 'Error' type`
` unable to locate module 'builtin'` 

btw we can split this PR in two if you'd like, DualQuaternion would be an extension to Quaternion after it's merged.

PR #2910, opened 2024-05-31
[stdlib] Implement `collections.Counter`

msaelices:
Support for Python's [collections.Counter](https://docs.python.org/3/library/collections.html#collections.Counter)

Done:
- [X] Implement some of the `Dict` methods: `keys()`, `values()`, `get()`, `pop()`, `clear()`, `update()`
- [X] Implement some specific `Counter` methods: `total()`, `most_common()`, `elements()`, `substract()`
- [X] Implement `__add__`, `__iadd__`, `__sub__` and `__isub__`
- [x] Implement `__eq__`, `__lt__`, `__le__`, `__gt__`, `__ge__`.

Pending to match Python signature:
- [ ] Ideally, it should accept all the iterable types, but for now, it will only work with `List`. I think we need to wait for proper support from Mojo
- [ ] More `Dict` methods, like `popitem()`, `fromkeys()`, `__contains__()`, etc.
- [ ] Implement `__sub__` and `__isub__`
- [ ] Implement `__and__`, `__iand__`, `__or__`, `__ior__` and `__neg__`

Ticket: https://github.com/modularml/mojo/issues/2908


stdlib/src/collections/counter.mojo
```diff
@@ -0,0 +1,355 @@
...
+    def __getitem__(self, key: V) -> Int:
+        """Get the count of a key.
+
+        Args:
+            key: The key to get the count of.
+
+        Returns:
+            The count of the key.
+        """
+        return self._data.get(key, 0)
```
soraros, line 78:
```suggestion
        return self.get(key, 0)
```

stdlib/src/collections/counter.mojo
```diff
@@ -0,0 +1,355 @@
...
+        """Iterate over the keyword dict's keys as immutable references.
+
+        Returns:
+            An iterator of immutable references to the Counter values.
+        """
+        return self[]._data.__iter__()
+
+    fn __len__(self) -> Int:
+        """The number of elements currently stored in the Counter."""
+        return self._data.size
```
soraros, line 101:
```suggestion
        return len(self._data)
```

stdlib/src/collections/counter.mojo
```diff
@@ -0,0 +1,355 @@
...
+        """Get a value from the Counter.
+
+        Args:
+            value: The value to search for in the counter.
+            default: Default count to return.
+
+        Returns:
+            A copy of the value if it was present, otherwise default.
+        """
+        return self._data.find(value).or_else(default)
```
soraros, line 149:
```suggestion
        return self._data.get(value, default)
```

stdlib/src/collections/counter.mojo
```diff
@@ -0,0 +1,355 @@
...
+        """Return a list of the n most common elements and their counts from the most common to the least.
+
+        Args:
+            n: The number of most common elements to return.
+
+        Returns:
+            A list of the n most common elements and their counts.
+        """
+        var items: List[CountTuple[V]] = List[CountTuple[V]]()
+        for item_ref in self._data.items():
```
soraros, line 228:
```suggestion
        for item_ref in self.items():
```

stdlib/src/collections/counter.mojo
```diff
@@ -0,0 +1,355 @@
...
+    # Special methods for counter
+
+    fn total(self) -> Int:
+        """Return the total of all counts in the Counter.
+
+        Returns:
+            The total of all counts in the Counter.
+        """
+        var total = 0
+        for count_ref in self._data.values():
```
soraros, line 214:
```suggestion
        for count_ref in self.values():
```

stdlib/src/collections/counter.mojo
```diff
@@ -0,0 +1,355 @@
...
+
+    fn update(inout self, other: Self):
+        """Update the Counter, like dict.update() but add counts instead of replacing them.
+
+        Args:
+            other: The Counter to update this Counter with.
+        """
+        for item_ref in other.items():
+            var item = item_ref[]
+            self._data[item.key] = self._data.get(item.key, 0) + item.value
```
soraros, line 261:
```suggestion
            self[item.key] = self.get(item.key, 0) + item.value
```

soraros (commented):
Thank you for your contribution!

A few comments:
- Maybe use `K` or `KeyType` for the key type.
- `ref` argument should be preferred over `self: Reference[...]`, see [examples here](https://github.com/modularml/mojo/commit/c7c6d49df09d817d1d6ca11b9a4c91f1d56d486b).
- We could use the `@value` for `__copyinit__` and `__moveinit__`.
- I think we should use public APIs whenever possible (`len(_data)` over `_data.size`; `self[] = ...` over `self._data[] = ...`, etc). The inline suggestions are not exhaustive.
- Why do we need `CountTuple`? Wouldn't `Tuple` work?

stdlib/src/collections/counter.mojo
```diff
@@ -0,0 +1,355 @@
...
+
+    The value type must implement the `KeyElement` trait, as its values are
+    stored in the dictionary as keys. `KeyElement` includes
+    `Movable`, `Hashable`, and `EqualityComparable`.
+
+    Usage:
+
+    ```mojo
+    from collections import Counter
+    var c = Counter[String](["a", "a", "a", "b", "b", "c", "d", "c", "c"])
```
soraros, line 31: I think you need to write `List(...)` instead.

msaelices: Unfortunatelly, you are right. Fixed: https://github.com/msaelices/mojo/commit/bda132997618e46081ed22f186a9dbfd23f1a016

stdlib/src/collections/counter.mojo
msaelices: Done

stdlib/src/collections/counter.mojo
msaelices: Done

laszlokindrat:
@msaelices Thanks for the patch! This would be amazing to have, but I would like to request that before we move forward with reviewing this in detail, you split up this patch. When introducing a new data structure, we should generally start with something bare bones, and as a general note, we should try to refrain from these large patches. There is a line somewhere around 300-400 lines above which it becomes increasingly difficult to review anything thoroughly.

For this particular patch, the comparison operators, as well as `__add__`/`__iadd__` certainly are not part of the core functionality. Since the implementation uses a `Dict`, I would like the first patch be restricted to the basic `__getitem__`/`__setitem__` interfaces. Then you can extend to the full dictionary interface (e.g. `pop`/`get` etc), and continue a few (semantically related) methods at a time. This way, we can do a much better job discussing API and implementation details, ensure test coverage, and share the load between reviewers.

As I said, counters would be awesome to have, so I really appreciate you moving this forward!

msaelices:
> @msaelices Thanks for the patch! This would be amazing to have, but I would like to request that before we move forward with reviewing this in detail, you split up this patch. When introducing a new data structure, we should generally start with something bare bones, and as a general note, we should try to refrain from these large patches. There is a line somewhere around 300-400 lines above which it becomes increasingly difficult to review anything thoroughly.
> 
> For this particular patch, the comparison operators, as well as `__add__`/`__iadd__` certainly are not part of the core functionality. Since the implementation uses a `Dict`, I would like the first patch be restricted to the basic `__getitem__`/`__setitem__` interfaces. Then you can extend to the full dictionary interface (e.g. `pop`/`get` etc), and continue a few (semantically related) methods at a time. This way, we can do a much better job discussing API and implementation details, ensure test coverage, and share the load between reviewers.
> 
> As I said, counters would be awesome to have, so I really appreciate you moving this forward!

I can split the patch. Please note that the `__add__`, `__iadd__` operations and the comparison operators have a different logic than the Dict ones. For example: `{"a": 3, "b": 1}` is equal to `{"a": 3, "b": 1, "c": 0}`, so there is a reason for implementing them.

Could you please enumerate the list of PRs with the needed functionality on them? So I can fork and cherry-pick the commits for each one?

laszlokindrat:
> > @msaelices Thanks for the patch! This would be amazing to have, but I would like to request that before we move forward with reviewing this in detail, you split up this patch. When introducing a new data structure, we should generally start with something bare bones, and as a general note, we should try to refrain from these large patches. There is a line somewhere around 300-400 lines above which it becomes increasingly difficult to review anything thoroughly.
> > For this particular patch, the comparison operators, as well as `__add__`/`__iadd__` certainly are not part of the core functionality. Since the implementation uses a `Dict`, I would like the first patch be restricted to the basic `__getitem__`/`__setitem__` interfaces. Then you can extend to the full dictionary interface (e.g. `pop`/`get` etc), and continue a few (semantically related) methods at a time. This way, we can do a much better job discussing API and implementation details, ensure test coverage, and share the load between reviewers.
> > As I said, counters would be awesome to have, so I really appreciate you moving this forward!



> I can split the patch. Please note that the `__add__`, `__iadd__` operations and the comparison operators have a different logic than the Dict ones. For example: `{"a": 3, "b": 1}` is equal to `{"a": 3, "b": 1, "c": 0}`, so there is a reason for implementing them.

I'm not arguing against implementing them. I would just like to deliver these in later steps.

> Could you please enumerate the list of PRs with the needed functionality on them? So I can fork and cherry-pick the commits for each one?

I think this gives you an idea of how to schedule these:

> Since the implementation uses a `Dict`, I would like the first patch be restricted to the basic `__getitem__`/`__setitem__` interfaces. Then you can extend to the full dictionary interface (e.g. `pop`/`get` etc), and continue a few (semantically related) methods at a time. 

There is no hard-and-fast rule on how big a patch should be, but IMO patches under 200 lines are much easier to review. As I mentioned above, semantic grouping is also important when you are building a well defined API.

msaelices:
> > > @msaelices Thanks for the patch! This would be amazing to have, but I would like to request that before we move forward with reviewing this in detail, you split up this patch. When introducing a new data structure, we should generally start with something bare bones, and as a general note, we should try to refrain from these large patches. There is a line somewhere around 300-400 lines above which it becomes increasingly difficult to review anything thoroughly.
> > > For this particular patch, the comparison operators, as well as `__add__`/`__iadd__` certainly are not part of the core functionality. Since the implementation uses a `Dict`, I would like the first patch be restricted to the basic `__getitem__`/`__setitem__` interfaces. Then you can extend to the full dictionary interface (e.g. `pop`/`get` etc), and continue a few (semantically related) methods at a time. This way, we can do a much better job discussing API and implementation details, ensure test coverage, and share the load between reviewers.
> > > As I said, counters would be awesome to have, so I really appreciate you moving this forward!
> 
> > I can split the patch. Please note that the `__add__`, `__iadd__` operations and the comparison operators have a different logic than the Dict ones. For example: `{"a": 3, "b": 1}` is equal to `{"a": 3, "b": 1, "c": 0}`, so there is a reason for implementing them.
> 
> I'm not arguing against implementing them. I would just like to deliver these in later steps.
> 
> > Could you please enumerate the list of PRs with the needed functionality on them? So I can fork and cherry-pick the commits for each one?
> 
> I think this gives you an idea of how to schedule these:
> 
> > Since the implementation uses a `Dict`, I would like the first patch be restricted to the basic `__getitem__`/`__setitem__` interfaces. Then you can extend to the full dictionary interface (e.g. `pop`/`get` etc), and continue a few (semantically related) methods at a time.
> 
> There is no hard-and-fast rule on how big a patch should be, but IMO patches under 200 lines are much easier to review. As I mentioned above, semantic grouping is also important when you are building a well defined API.

Ok, I've created the first PR, forked from an early stage of this one and synced from `nightly`: https://github.com/modularml/mojo/pull/2937

msaelices:
> Thank you for your contribution!
> 
> A few comments:
> 
> * Maybe use `K` or `KeyType` for the key type.

I think `V` is better. The decision of storing the value as a key to me is an internal implementation details we may or may not expose.

> * `ref` argument should be preferred over `self: Reference[...]`, see [examples here](https://github.com/modularml/mojo/commit/c7c6d49df09d817d1d6ca11b9a4c91f1d56d486b).

Done.

> * We could use the `@value` for `__copyinit__` and `__moveinit__`.

Done.

> * I think we should use public APIs whenever possible (`len(_data)` over `_data.size`; `self[] = ...` over `self._data[] = ...`, etc). The inline suggestions are not exhaustive.

Done.

> * Why do we need `CountTuple`? Wouldn't `Tuple` work?

It's not working because tuple does not conform `CollectionElement`:

image

laszlokindrat:
@msaelices The first part of this patch has landed in the latest nightly. Can you please rebase and split out the next batch?

msaelices:
@laszlokindrat could you please take another look?

stdlib/test/collections/test_counter.mojo
```diff
@@ -32,6 +32,206 @@ def test_counter_getitem():
...
+    c2["b"] = 2
+    c2["c"] = 0
+
+    assert_false(c1 < c2)
+    assert_true(c1 <= c2)
+    assert_false(c1 > c2)
+
+    c2["b"] = 3
+    assert_true(c1 < c2)
+    assert_true(c1 <= c2)
```
laszlokindrat, line 157: Looks like `>=` is untested. Could you please add a cast for that?

laszlokindrat (commented):
Thanks! Could you please add unit tests for `get`, `pop`, `clear`, `__bool__`, and `__len__`?

stdlib/test/collections/test_counter.mojo
laszlokindrat: Also for this and the other comparison methods, could you please test the dunder methods directly, instead of using the operators?

msaelices: Done: https://github.com/msaelices/mojo/commit/dc2d2079c9a39764fc585103bdf70bd923e103a3

msaelices:
> test/collections/test_counter.mojo



> Thanks! Could you please add unit tests for `get`, `pop`, `clear`, `__bool__`, and `__len__`?

Done: 
https://github.com/msaelices/mojo/commit/e48f37cd0f91a92be676a475d789d8ac3d532054
https://github.com/msaelices/mojo/commit/1380750b9d28b097cd8f5caefbfbe6a523c32f1d

msaelices:
@laszlokindrat could you PTAL?

stdlib/src/collections/counter.mojo
```diff
@@ -76,3 +73,363 @@ struct Counter[V: KeyElement](
...
+        """The number of elements currently stored in the Counter."""
+        return len(self._data)
+
+    fn __bool__(self) -> Bool:
+        """Check if the Counter is empty or not.
+
+        Returns:
+            `False` if the Counter is empty, `True` if there is at least one element.
+        """
+        return len(self).__bool__()
```
soraros, line 95: I think it should be the other way around, as `__len__` is the one we defined here.
```suggestion
        return bool(self.__len__())
```
Actually, this should also be good. I don't know which style should we prefer. @laszlokindrat Do we always use explicit method call in implementations?

```suggestion
        return bool(len(self))
```

stdlib/src/collections/counter.mojo
```diff
@@ -76,3 +73,363 @@ struct Counter[V: KeyElement](
...
+    fn __lt__(self, other: Self) -> Bool:
+        """Check if all counts are less than in the other Counter.
+
+        Args:
+            other: The other Counter to compare to.
+
+        Returns:
+            True if all counts are less than in the other Counter, False otherwise.
+        """
+        return self <= other and self != other
```
soraros, line 159: Should we rely on loop fusion here?

msaelices: Not sure if it's worth it. I think currently is much cleaner to read. Not sure if @laszlokindrat has some opinion

stdlib/src/collections/counter.mojo
msaelices: Done: https://github.com/msaelices/mojo/commit/df2f4b1a905afb1f6ced8fdf6fccc8ec91dc2077

msaelices:
@laszlokindrat could you PTAL?

stdlib/src/collections/counter.mojo
```diff
@@ -76,3 +73,363 @@ struct Counter[V: KeyElement](
             count: The count to store in the Counter.
         """
         self._data[value] = count
+
+    fn __iter__(ref [_]self: Self) -> _DictKeyIter[V, Int, __lifetime_of(self)]:
+        """Iterate over the keyword dict's keys as immutable references.
+
+        Returns:
+            An iterator of immutable references to the Counter values.
```
ConnorGray, line 81:
```suggestion
            An iterator of references to the Counter values.
```

**Nit** It looks like these references use parameter mutability, so aren't always immutable :)

ConnorGray:
@msaelices Thank you for this contribution! 🔥 to have this type working like in Python.

I just wanted to let you know that Laszlo is out this week so may not get to reviewing this for a little bit. I'll have a look as well, but since this is a large feature and I don't have all the context I may wait for Laszlo to come back before proceeding with this.

stdlib/src/collections/counter.mojo
```diff
@@ -11,14 +11,11 @@
...
-from collections.dict import Dict
+from collections.dict import Dict, _DictKeyIter, _DictValueIter, _DictEntryIter
 
 
-struct Counter[V: KeyElement](
-    #    Sized,
-    #    CollectionElement,
-    #    Boolable
-):
+@value
```
ConnorGray, line 17: **Request** Since you're significantly expanding the number of methods on this type, I think it would be good for readability if we added some of the standard header comments to visually separate and group different kinds of methods: https://github.com/modularml/mojo/blob/main/stdlib/docs/style-guide.md#code-header-comments

The goal is to have a conventional grouping and ordering for methods so readers can "pattern match" when reading different type definitions and have a good sense at glance about the semantics and way a type is meant to be used.

You can see how those headers are used in most of the other prominent standard library types as well🙂

msaelices: Sure, done here: https://github.com/msaelices/mojo/commit/cfc97a6e09f5ae934d913d3118028f5eaf6e1c64

stdlib/src/collections/counter.mojo
```diff
@@ -76,3 +82,400 @@ struct Counter[V: KeyElement](
...
+        @parameter
+        @always_inline
+        fn is_eq(keys: _DictKeyIter[V, Int, _]) -> Bool:
+            for e_ref in keys:
+                var e = e_ref[]
+                if self.get(e, 0) != other.get(e, 0):
+                    return False
+            return True
+
+        return is_eq(self.keys()) and is_eq(other.keys())
```
soraros, line 133: `is_eq` captures `other`, so `is_eq(other.keys())` doesn't do anything, right?

stdlib/src/collections/counter.mojo
```diff
@@ -76,3 +82,400 @@ struct Counter[V: KeyElement](
...
+        Returns:
+            True if all counts are greater than or equal to the other Counter, False otherwise.
+        """
+        return other <= self
+
+    # ===-------------------------------------------------------------------===#
+    # Binary operators
+    # ===-------------------------------------------------------------------===#
+
+    fn __add__(self, other: Self) raises -> Self:
```
soraros, line 204: Why does this raise?

stdlib/src/collections/counter.mojo
```diff
@@ -76,3 +82,400 @@ struct Counter[V: KeyElement](
...
+
+    fn values(ref [_]self: Self) -> _DictValueIter[V, Int, __lifetime_of(self)]:
+        """Iterate over the Counter's values as references.
+
+        Returns:
+            An iterator of references to the Counter values.
+        """
+        return self._data.values()
+
+    fn items(ref [_]self: Self) -> _DictEntryIter[V, Int, __lifetime_of(self)]:
```
soraros, line 310:
```suggestion
    fn items(self: Self) -> _DictEntryIter[V, Int, __lifetime_of(self)]:
```

stdlib/src/collections/counter.mojo
msaelices: This is the Mojo version of the Python `Counter.__eq__()` that is based on two list comprehension joined, so a double loop:

image

It's needed to loop for both the `self` and `other` keys, as the keys listing could not be overlapped.

stdlib/src/collections/counter.mojo
msaelices: Thanks for the catch! Done here: https://github.com/msaelices/mojo/commit/d5f109f1f4dbb07f8de94366da6344473264a7fa

msaelices: Also implemented the useful `Counter.__contains__()` method here: https://github.com/msaelices/mojo/commit/0f22d2800ea813ec2742b1d9bc935b325219de61

stdlib/src/collections/counter.mojo
msaelices: I've copied the docstring from the `Dict.__iter__()` method. Are you sure it's wrong? should we change the `Dict` one too?

stdlib/src/collections/counter.mojo
msaelices: Fixed here: https://github.com/msaelices/mojo/commit/fd373bd1e0b46483994a5ea14a2df5fd9c8a046b

stdlib/src/collections/counter.mojo
msaelices: I think to be more precise, we can remove the `ref [_]` from the declaration. Fixed here: https://github.com/msaelices/mojo/commit/f46a98a95b0e1e8c3f695822020fd3b47ba59eb3

stdlib/src/collections/counter.mojo
soraros: Ah, I misread the code.

ConnorGray:
!sync

laszlokindrat:
The patch looks good to me, and we really appreciate the contribution! As I mentioned before, please try to keep the patch size smaller in the future. There is a lot that goes into a review, not just reading the code (e.g. ensuring there are appropriate unit tests, the changelog entry is correct, docstrings are useful, etc.). When we land, we also need to ensure it does not break internal code or cause a regression, which is harder for large patches. So again, we love that you contribute, and if you keep the patches smaller, they will land much faster, and are less likely to be reverted :)

PR #2913, opened 2024-06-01
[stdlib] file handle write borrowed slice of data

### LINKED ISSUE #2909: [stdlib] Add safe, non-allocating `FileHandle.write(..)` overload
We should add a `FileHandle.write(Span[UInt8, _, _])` overload for safely writing a borrowed sequence of bytes.

Currently, the `FileHandle` type used to read and write data to an opened file descriptor currently provides three `write(..)` method overloads:

* `FileHandle.write(String)` — requires allocating a `String`
* `FileHandle.write(StringRef)` — use of the unsafe and easy-to-misuse `StringRef` type
* `FileHandle.write(UnsafePointer[UInt8], Int)` — requires use of `UnsafePointer`

This means that all currently existing ways to write to a file are either non-optimal or requires dipping down into low-level unsafe code.

GeauxEric:
fix #2909 

Signed-off-by: Yun Ding <yunding.eric@example.com>


stdlib/src/builtin/file.mojo
```diff
@@ -414,6 +414,14 @@ struct FileHandle:
         """
         self._write(data.unsafe_ptr(), len(data))
 
+    fn write(self, data: Span[UInt8, _, _]) raises:
+        """Write a borrowed sequence of data to the file.
+
+        Args:
+          data: The data to write to the file.
+        """
+        self._write(data.unsafe_ptr().bitcast[Int8](), len(data))
```
gabrieldemarmiesse, line 423: Is this bitcast necessary? We are currently transitionning to making bytes UInt8 values.

Also since this is the same code as `fn write(self, data: String) raises:` without the bitcast, would it be possible to write generic code that would work for both using traits?

stdlib/test/builtin/test_file.mojo
```diff
@@ -160,6 +160,18 @@ def test_file_write():
...
 
 
+def test_file_write_span():
+    var content: String = "The quick brown fox jumps over the lazy dog"
+    var TEMP_FILE = Path(TEMP_FILE_DIR) / "test_file_write_span"
+    var f = open(TEMP_FILE, "w")
+    f.write(content.as_bytes_slice())
+    f.close()
+
+    var read_file = open(TEMP_FILE, "r")
```
JoeLoser, line 170: **Suggestion** Can we use a context manager here, e.g.

```
with open(TEMP_FILE, "r") as f:
  ...
```

JoeLoser (approved):

JoeLoser (approved):
Looks great, thanks for jumping on this so quickly!  I just left a small suggestion in the test, but happy to import and merge after that.  Thank you!

stdlib/test/builtin/test_file.mojo
GeauxEric: done.
I fixed a few other tests to use `with`.
note in `test_file_get_raw_fd`
```
    # since JIT and build give different file descriptors, we test by checking
    # if we printed to the right file.
    var f1 = open(Path(TEMP_FILE_DIR) / "test_file_dummy_1", "rw")
    var f2 = open(Path(TEMP_FILE_DIR) / "test_file_dummy_2", "rw")
    var f3 = open(Path(TEMP_FILE_DIR) / "test_file_dummy_2", "rw")
```
Currently Mojo does not support several statements in one `with`, e.g. `with xxx, yyy:`. This is supported in Python, the compiler team might want to support that.

JoeLoser:
!sync

JoeLoser (approved):

JoeLoser:
!sync

Issue #2730, opened 2024-05-18
[Feature Request] Consolidate `bin()` and `hex()` implementations


bgreni:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Immediately upon completing #2603 I realized `hex.mojo` already contains a general implementation for such things. These should be consolidated into a single implementation that accepts any size of `SIMD` or `Indexer` type. Having done this it would be trivial to add `oct` as well.

### What is your motivation for this change?

Reduce code duplication.

bgreni:
> Can I work on this?

Apologies for not opening a draft sooner but I've just put up a PR for it #2790

PR #2914, opened 2024-06-01
[stdlib] Add oct function

bgreni:
Add the `oct` builtin function to get the octal representation of integers.

Follow up from #2730


ConnorGray:
!sync

PR #2915, opened 2024-06-01
[stdlib] Fix possible bug in List.resize with default value T

rd4com:
There might be a bug in:
 `List.resize(inout self, new_size: Int, value: T)`

It calls `destroy_pointee` on uninitialized values (`self.reserve(new_size)`),
it is used with trivial type `Int` (no-op destructor) in stdlib so no bug yet?

Can someone have a look internally?

Not sure if it a bug or not, but it seem important to check!


stdlib/src/collections/list.mojo
```diff
@@ -568,8 +568,6 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
             self.resize(new_size)
         else:
             self.reserve(new_size)
-            for i in range(new_size, self.size):
```
gabrieldemarmiesse, line 571: Your change is correct, but it was not a bug. Those two lines were actually useless and no-op. If `new_size > self.size` then the range is null. Thus the `destroy_pointee()` can never be executed.
You can try this in a standalone script:
```mojo
def main():
    for i in range(10, 5):
        print(i)  # <- never executed
    print("done")
```

gabrieldemarmiesse (approved):

stdlib/src/collections/list.mojo
ConnorGray: Agreed. Still great to remove these useless lines those, thanks @rd4com!

ConnorGray:
!sync

ConnorGray (approved):

ConnorGray:
!sync

ConnorGray:
!sync

PR #2925, opened 2024-06-02
[proposal] Add `Deque` struct to the stdlib

gabrieldemarmiesse:
* https://github.com/modularml/mojo/issues/2659 

@JoeLoser I did the API design proposal as requested. We should be able to easily discuss the api with this format.

If you want to look at the rendered markdown, you can read it here: 
https://github.com/gabrieldemarmiesse/mojo/blob/add_deque_proposal/proposals/add-deque-struct.md

### DIFF
--- a/proposals/add-deque-struct.md
+++ b/proposals/add-deque-struct.md
@@ -1,0 +1,153 @@
+# Proposal: Add a `Deque` struct to the stdlib
+
+## Goals
+
+The goals of this document are
+
+1) Explain what problems we want to solve with `Deque`, and what problems we
+   don't want to solve.
+2) Explain how the internal storage of this double-ended
+   queue is managed. Give explored alternatives.
+3) Propose an API which can be used as a "todo list"
+   for contributors to implement.
+
+## 1 - Why we want to add a `Deque` struct to the stdlib
+
+All major languages have a double-ended queue in their stdlib.
+And they are useful for a number of reasons.
+The major one is that it allows users to work with
+FIFO (first-in first-out) and LIFO (last-in first-out) queues,
+which are very common in programming.
+
+For context, we are still at the start of Mojo's stdlib.
+Given the number of maintainers,
+we want to create a new data structure which
+doesn't add too much code and complexity to the codebase.
+
+With those constraints we give the following goals:
+
+1) Provide a data structure which allows users to
+   implement a fast FIFO and LIFO
+2) Has a familiar api with already existing Mojo
+   collections (`List`, `InlineList`, `String`, `VariadicList`, ...)
+3) Has a familiar api to Python users
+4) Can be efficiently converted from and into a `List`.
+5) Has fast access to its elements (cache-friendly).
+
+The non-goals are:
+
+1) Fast insert in the middle of
+   the `Deque` (motivation for this use case is not obvious)
+2) Thread-safety (require a certain amount of complexity)
+
+## 2 - Internal storage
+
+`Deque` will allocate a buffer of capacity `capacity` and
+use a similar algorithm as `List` to make it grow when necessary.
+
+To avoid reallocation when doing frequent append
+and pop on different sides of the
+`Deque`, which is common for FIFO queues,
+we'll use this buffer as a [circular buffer](https://en.wikipedia.org/wiki/Circular_buffer).
+
+This is how Rust's
+[`VecDeque`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html)
+is implemented. Note that it differs from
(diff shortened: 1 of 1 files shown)


proposals/add-deque-struct.md
```diff
@@ -0,0 +1,153 @@
...
+
+With those constraints we give the following goals:
+
+1) Provide a data structure which allows users to
+   implement a fast FIFO and LIFO
+2) Has a familiar api with already existing Mojo
+   collections (`List`, `InlineList`, `String`, `VariadicList`, ...)
+3) Has a familiar api to Python users
+4) Can be efficiently converted from and into a `List`.
+5) Has fast access to its elements (cache-friendly).
```
laszlokindrat, line 35: IMO 4) should not be a goal for the data structure we will call `deque` (or `Deque`). Neither CPP or Python makes this promise, and those are the languages/stdlibs closest to Mojo, and compatibility with Python is a stated goal (although, I admit, the level of this compatibility is not well defined). CPP does promise 5), and I think it's reasonable to aim for it (but maybe not right away).

proposals/add-deque-struct.md
```diff
@@ -0,0 +1,153 @@
...
+- Not thread safe
+- Needs to move all existing elements when growing the buffer, O(N) worst case.
+- Fast conversion to and from `List`
+- Intergrates nicely with small buffer optimization
+
+**Python**:
+
+- Slow indexing in the middle
+- Not cache friendly
+- Thread safe
```
laszlokindrat, line 79: This should be qualified: Thread safety is promised for pop/append on the ends, but not in the middle: https://docs.python.org/3/library/collections.html#collections.deque

proposals/add-deque-struct.md
```diff
@@ -0,0 +1,153 @@
...
+- Adding elements (left or right) is O(1) in the worst case scenario
+- No contiguous internal buffer for conversion to and from other structs
+- Doesn't intergrate well with small buffer optimization
+
+## 3 - API
+
+While we use Rust's implementation of the double-ended queue, that does not mean
+that we need to follow Rust's API or that we can't provide an API fammiliar to
+Python and Mojo's users. Here is the proposed API, which can be used as roadmap
```
laszlokindrat, line 88: As you mention above, Python's `collections.deque` _promises_ O(1) insertion on the ends. This is part of the API contract, so you cannot fully decouple implementation for API. But either way, I would strongly argue for just following Python's `deque` to begin with. This is what Python developers first reach for when they want a queue, and CPP makes similar tradeoffs while also mitigating the cache problems you mention above.

laszlokindrat (commented):
Thanks for writing this up, excellent doc! My main quibble is calling the data structure you propose a `Deque`. The two most important examples for us to follow are those of CPP and Python, and both of them make different decisions from what you propose here, which is why I think our double ended queue should follow those. That being said, I'm not against having a data structure like you describe in the stdlib, but I would prefer we just call it something explicit like `RingBuffer` or `RingQueue` or something like that.

proposals/add-deque-struct.md
```diff
@@ -0,0 +1,153 @@
...
+the number of maintainers and dynamism of the codebase.
+
+### Tradeoffs between the Rust and Python implementation
+
+**Rust**:
+
+- Efficient indexing
+- Cache friendly
+- Not thread safe
+- Needs to move all existing elements when growing the buffer, O(N) worst case.
```
laszlokindrat, line 71: This would mean that the contents need to be `Movable`. This could be a reasonable restriction in the short term, but eventually queues of non-movable types should be supported as well (when we enable in-place construction).

gabrieldemarmiesse:
@laszlokindrat Overall, I don't have a strong opinion on the internal implementation. Python and C++ use a double-linked list of arrays and Rust uses a ring buffer. I chose the ring buffer because it would be simpler as a first step, but I don't mind starting directly with the double-linked list of arrays. If possible I would prefer to avoid having to do both 😅  I don't think we need both in the stdlib. Someone else can always add a lib with the other.

Indeed if we drop the fast access to middle elements requirement, then we can easily switch to the other implementation.

If we decide that the arrays in the double-linked-list do not have a size known at compile-time, we can easily steal pointers from other data structures to append to the deque or initialize the deque with.

JoeLoser:
Given that Deque has been moving along in implementation from https://github.com/modularml/mojo/pull/3759 and previous, I'm going to go ahead and close this proposal.  Thank you for pushing on the original write up @gabrieldemarmiesse as it motivated the implementation which is almost finalized in landing (2/3 of the PRs have landed so far) 🎉

PR #2926, opened 2024-06-02
[stdlib] `memcmp` refactor for string comparisons

siitron:
This is a follow-up to #2740.
Edit:
Refactored the `memcmp` implementation to give `StringRef` a `constrained`-less version for vectorized string comparisons that avoid an internal recursion bug.

Old:
`StringRef`'s local `_memcmp` is now vectorized, but without using `constrained` in an attempt to avoid indirect recursions.

`StringRef`'s `_memcmp` is pretty much identical to `memory.memcmp` now, but I've also replaced `_align_down(count, simd_width)` with `count - simd_width` (as the upper limit for the simd loop) because in cases where `count` is divisible by `simd_width` (and `count > simd_width`, e.g. `count = 64, simd_width = 32`), then the last simd comparison would be done twice (at the end of the loop and once after).

### DIFF
--- a/stdlib/src/memory/memory.mojo
+++ b/stdlib/src/memory/memory.mojo
@@ -42,8 +42,9 @@ fn _align_down(value: Int, alignment: Int) -> Int:
 
 
 @always_inline
-fn _memcmp_impl(s1: DTypePointer, s2: __type_of(s1), count: Int) -> Int:
-    constrained[s1.type.is_integral(), "the input dtype must be integral"]()
+fn _memcmp_impl_unconstrained(
+    s1: DTypePointer, s2: __type_of(s1), count: Int
+) -> Int:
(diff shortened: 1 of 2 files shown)


siitron:
@laszlokindrat Do you mind checking if this is affected by the recursion thing

laszlokindrat:
@siitron Thanks for the patch! I will try it, but I would prefer to refactor `memcmp` instead of duplicating code.

laszlokindrat:
!sync

laszlokindrat:
@siitron Looks like it passed our internal CI. Can you please try to refactor the `memcmp` implementation so that we can resolve this problem without code duplication?

laszlokindrat:
(Sorry, I closed it accidentally)

laszlokindrat (approved):
Nice, thanks for fixing this!

PR #2932, opened 2024-06-03
[stdlib] Add a `UInt` struct with a few methods

gabrieldemarmiesse:
Fix part of 
* https://github.com/modularml/mojo/issues/2919


stdlib/src/builtin/uint.mojo
```diff
@@ -0,0 +1,86 @@
...
+"""Implements the UInt class.
+
+These are Mojo built-ins, so you don't need to import them.
+"""
+
+
+@lldb_formatter_wrapping_type
+@value
+@register_passable("trivial")
+struct UInt:
```
JoeLoser, line 22: **Question** Do you think we can get a lot of mileage/it would be sufficient to just have this be an alias to `UInt32` on 32-bit systems and `UInt64` otherwise?  @abduld just suggested this internally, which, sounds quite simple and would meet our needs if I understand correctly?

abduld: The implementation would be 

```
alias UInt = UInt64 if bitwidthof[DType.index]() else UInt32
```

gabrieldemarmiesse: Out of curiosity, is there any reason we're not doing this for Int then?

bgreni: I've also never been convinced that `Int` couldn't just be an alias so adding another seems unnecessary. If the user really needs an unsigned integer they can easily use one of the existing `Scalar` types?

abduld: Int is really the same as Scalar[DType.index], but is implemented using the mlir index dialect rather than the compiler dialects (like pop) that is used by SIMD. I encourage one to try to replace the implementation of Int with an alias

laszlokindrat: `Int` is special because it is the materialization target of `IntLiteral`. It also allows us to implement a bunch of other things that we couldn't do with `SIMD` because we lack proper conditional conformances. Turning `Int` into an alias is also not a simple task because the compiler depends on it. I'm +1 on trying to define `UInt` as an alias, but -1 on trying to change the `Int` implementation (at this point).

Mogball: I would strongly recommend not touching Into. It's special in a lot of ways

Mogball (commented):
Using `index` is the right approach here IMO. Do you plan to use the unsigned versions of the `index` dialect operations? E.g. https://mlir.llvm.org/docs/Dialects/IndexOps/#indexdivu-indexdivuop?

gabrieldemarmiesse:
> Using index is the right approach here IMO. Do you plan to use the unsigned versions of the index dialect operations? E.g. https://mlir.llvm.org/docs/Dialects/IndexOps/#indexdivu-indexdivuop?

If we go with `index` as internal representation this is indeed the plan. I already wrote more code using it, but I wanted to keep this pull request small so I left it for a future PR. 

Maybe a small meeting internally would help clear up the future plans for this PR and the use of SIMD in `Int` and `UInt`? It seems every maintainer has a different opinion here.

Mogball:
> Maybe a small meeting internally would help clear up the future plans for this PR and the use of SIMD in Int and UInt? It seems every maintainer has a different opinion here.

That's a good idea. We're definitely not all on the same page about this. The leaf types in the language have to be handled with care and can't get that far away from the compiler, because they're the types that will be optimized the most often!

JoeLoser:
> @JoeLoser do you want to schedule something?

Yeah, I'll set something up internally for us to discuss and get on the same page. Thanks for pushing on this @gabrieldemarmiesse as a forcing function. ❤️

stdlib/src/builtin/uint.mojo
Mogball: Concretely, `Int` is backed by an `__mlir_type.index` because the `index` opset is more mature than the `pop` opset. The POP opset is also semantically different: we have `index.add_nsw` but `pop.add_nsw` doesn't make as much sense because it would only work for a subset of SIMD dtypes. Using an `index` has other benefits because it leads to terse IR, faster compile times, and also has lots of other load-bearing weirdness in MAX that would be very difficult to change.

gabrieldemarmiesse: @JoeLoser Did you have the time to sync on the subject?

JoeLoser:
!sync

PR #2934, opened 2024-06-03
[stdlib] Simplify iterator logic in `dict.mojo`

gabrieldemarmiesse:
There is no need to increment or decrement the index in both branches of the if. We just need to do it once, before the if statement, since `self.index` is not used afterwards.

It also makes it easier to understand the purpose of the if statement.


laszlokindrat (approved):
Nice catch, thanks!

laszlokindrat:
!sync

PR #2945, opened 2024-06-04
[stdlib] Abort on list resize to bigger sizes without fill value

gabrieldemarmiesse:
Fix https://github.com/modularml/mojo/issues/2938

Note that I had to remove the `@always_inline` to avoid infinite recursion of `@always_inline` functions.


stdlib/src/collections/list.mojo
```diff
@@ -581,13 +580,16 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
...
             new_size: The new size.
         """
-        debug_assert(
-            new_size <= self.size,
-            (
-                "New size must be smaller than or equal to current size when no"
-                " new value is provided."
-            ),
-        )
+        if self.size < new_size:
```
laszlokindrat, line 583: Somehow this makes me feel like we should have an `always_assert` or something similar that is not dependent on the build type. That's the only issue with `debug_assert` here, right?

gabrieldemarmiesse:
Do not merge, I found a very strange flaky CI error, I don't want it to get to nightly.

stdlib/src/collections/list.mojo
soraros: If we do that, the string is going to be materialised unconditionally, which is not good, right?

.github/workflows/standard_library_tests_and_examples.yml
```diff
@@ -25,6 +25,7 @@ permissions:
 jobs:
   test-examples:
     strategy:
+      fail-fast: false
```
laszlokindrat, line 28: I'm assuming this was left here accidentally, right?

laszlokindrat:
!sync

.github/workflows/standard_library_tests_and_examples.yml
gabrieldemarmiesse: I had another PR merged into this one to make the debug easier. It got merged into nightly. So this diff should go away now.

stdlib/src/collections/list.mojo
soraros: Could you explain why variadic can help mitigate this problem?

gabrieldemarmiesse:
Okay there is definitly an issue here somewhere. Let's not merge this PR. I have no idea what is going on.

stdlib/src/collections/list.mojo
laszlokindrat: > I wouldn't worry too much about `debug_assert` as I think it will be dead code eliminated in the release, right?

Yes, but it still poses a problem if you want a debug build on a target that cannot heap allocate.

If you pass a variadic of Stringables, the implementation of the function can ensure that they only get actually stringified (and thefore for needing a heap allocation) if the passed condition fails. So instead of
```mojo
my_assert(cond, "some message with id" + str(id))
```
you would have
```mojo
my_assert(cond, "some message with id", id)
```
There might be control flow sinking happening in the first case, but it's not guaranteed, while in the second case an implementation like this would make sure that no heap allocation happens if the condition is true:
```mojo
fn my_assert[*Ts: Stringable](cond: Bool, *msgs: *Ts):
    if not cond:
        @parameter
        if sys.can_heap_allocate():  # doesn't exist, but you get the point
            msg: String = ""
            for x in msgs:  # might have to fiddle with the exact syntax here
                msg += str(x)  # many reallocs, but who cares if you are already aborting?
            abort(msg)
        else:
            alias static_msg = "something went wrong"
            abort(static_msg)
```

laszlokindrat:
I ran our internal CI on this. Strangely, it actually fails due to always_inline recursion. Can't really see why, since you removed the decorator 🤷

gabrieldemarmiesse:
@laszlokindrat does it fail only on macos internally? It's what happens with the external CI.

laszlokindrat:
> @laszlokindrat does it fail only on macos internally? It's what happens with the external CI.

No, it failed on other architectures. But it seems it's now also failing the external CI.

laszlokindrat:
@gabrieldemarmiesse I'm just circling back to this. Could you please rebase this on the latest nightly?

gabrieldemarmiesse:
@laszlokindrat I don't know if you have the permission to do so, but feel free to click on the button

image

when you want my PRs to be updated. I know it can feel weird to push a commit to someone's fork but I'll allow it :p 

Also good news! The CI is green! 🎉

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

laszlokindrat:
FYI I had to remove the dynamically built abort message to land this internally.

PR #2947, opened 2024-06-04
[CI] Run all unit tests with assertions enabled and disabled

gabrieldemarmiesse:
Follow-up on https://github.com/modularml/mojo/pull/2718#discussion_r1609049240


gabrieldemarmiesse:
It's pretty funny that it only fails on macos haha. I'll debug a bit more tonight.

gabrieldemarmiesse:
FYI I think that `has_not` is not present in linux in the CI. Thus the test `test_debug_assert_mojo_enable_assertions.mojo` only runs on macos. This is why it only fails in the macos build. Locally I can make it crash on linux as expected. It's the `# REQUIRES: has_not` which is the cause of this (very likely). But since we don't have too much info about this in the logs, it's all speculation.

JoeLoser:
> For your information, I think that `has_not` is not present in linux in the CI. Thus the test `test_debug_assert_mojo_enable_assertions.mojo` only runs on macos. This is why it only failed in the macos build. Locally I can make it crash on linux as expected. It's the `# REQUIRES: has_not` which is the cause of this (very likely). But since we don't have too much info about this in the logs, it's all speculation.
> 
> I don't think it's too much of an issue right now especially since we're moving away from lit. Let's not spend too much time on this. The PR can be reviewed.

This is a package manager problem, and not a big deal.  The problem is the `not` binary ships in the Homebrew package of LLVM, but not on Linux.  Internally, we don't have this issue since we build LLVM from source for our whole stack, so we always have the `not` binary available.

JoeLoser (approved):
LGTM! Copybara isn't managing the `.github` directory, so I'll import and merge this PR, but then you'll need to open a different PR tomorrow/in the future for the `.github` workflow file changes itself, and we can merge that PR directly into the `nightly` branch.

JoeLoser (approved):
Thanks for working on this!

JoeLoser:
!sync

PR #2949, opened 2024-06-05
[stdlib] Consolidate range structs and fix length calculation

bgreni:
Having three different range structs is unnecessary and makes it difficult to do something like write a function that accepts a range as an argument since `range` will currently return 3 different types depending on which override you happen to use.

- Remove `_ZeroStartingRange` and `_SequentialRange` as everything can be done with `_StridedRange`
- Rename `_StridedRange` to `_Range`
- Apply index normalization to `_Range.__getitem__`
- Fix `_Range` length when start index is after end index.

### DIFF
--- a/stdlib/src/builtin/range.mojo
+++ b/stdlib/src/builtin/range.mojo
@@ -20,6 +20,7 @@ from python import PythonObject
 
 # FIXME(MOCO-658): Explicit conformance to these traits shouldn't be needed.
 from builtin._stubs import _IntIterable, _StridedIterable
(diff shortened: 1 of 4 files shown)


ConnorGray:
!sync

laszlokindrat (approved):
Love seeing code being deleted, thank you!

laszlokindrat:
!sync

laszlokindrat:
@bgreni Unfortunately this is some internally performance critical code, so we currently cannot take this patch. Let's circle back to it later!

PR #2950, opened 2024-06-05
[stdlib] Fix Span slicing behaviour

bgreni:
- Allow more sophisticated slicing behaviour in `Span` by storing the currently active slice inside the span
- Alter `String` constructor from `StringSlice` to copy in a for loop instead of `memcpy` to correct behaviour when the 
  underlying `Span` has been sliced

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -847,15 +847,14 @@ struct String(
         """
 
         # Calculate length in bytes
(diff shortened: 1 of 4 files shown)


bgreni:
Unblocked by the introduction of 3cedf4d4fb5d96df1d61ef7907acf3239059c972

JoeLoser:
!sync

JoeLoser (approved):
Nice fix!  Thank you!

PR #2952, opened 2024-06-05
[stdlib] Fixed recursive calls to String.__getitem__ for negative idxs

tjdwill:
Currently, `String.__getitem__(self, idx: Int)` has a behavior in which really large indices result in many recursive calls, causing the REPL to hang. Given,

```mojo
1> var s: String = "there"
2. print(s[-1000000000000000])
3.     

```

The REPL hangs. At the time of writing, it is still hanging after an hour.

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -924,6 +924,9 @@ struct String(
 
     fn __getitem__[IndexerType: Indexer](self, idx: IndexerType) -> String:
         """Gets the character at the specified position.
+        
+        Parameters:
+            IndexerType: The inferred type of an indexer argument.
 
         Args:
             idx: The index value.
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -924,7 +924,7 @@ struct String(
 
     fn __getitem__[IndexerType: Indexer](self, idx: IndexerType) -> String:
         """Gets the character at the specified position.
-        
+
         Parameters:
             IndexerType: The inferred type of an indexer argument.
 
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -17,6 +17,7 @@ These are Mojo built-ins, so you don't need to import them.
 
 from bit import countl_zero
 from collections import List, KeyElement
+from collections._index_normalization import normalize_index
 from sys import llvm_intrinsic, bitwidthof
 from sys.ffi import C_char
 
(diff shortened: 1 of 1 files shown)


stdlib/src/builtin/string.mojo
```diff
@@ -930,12 +930,13 @@ struct String(
         Returns:
             A new string containing the character at the specified position.
         """
+        var index = idx
         if idx < 0:
-            return self.__getitem__(len(self) + idx)
+            index = len(self) + index
```
gabrieldemarmiesse, line 935: Thank you for the pull request, indeed this needed to be fixed! The is a function to do this operation. Can you take a look at `normalize_index`? See https://github.com/modularml/mojo/issues/2948 for more information on the subject :)

gabrieldemarmiesse (commented):
Nice catch!

stdlib/src/builtin/string.mojo
tjdwill: I'll have a look, thank you!

gabrieldemarmiesse:
If you are having trouble with the formatting, make sure to use the pre-commit. See the install steps here: https://github.com/modularml/mojo/blob/nightly/stdlib/docs/development.md#formatting-changes

tjdwill:
Thank you for that, so this should now automatically re-format any changes I make on commit?


Also, I do have a few questions if you don't mind answering:

1) How does `normalize_index` function know if `Int` has the `Indexer` trait?
2) What does the `//` do in the `normalize_index` parameter list?
3) Speaking of parameter lists, are they basically used for generic programming? As in, one is able to specify the characteristics of the types the function takes?
4) Given `normalize_index` takes three parameters in the param list, why are we allowed to only pass in one (the container_name)? How does the compiler know the argument passed applies to that specific parameter?

gabrieldemarmiesse:
Hi @tjdwill , here are the answers for your questions:
1) Traits are explained here in the docs, it explains pretty well what it can do: https://docs.modular.com/mojo/manual/basics#traits
2) It marks the inferred parameters, see https://github.com/modularml/mojo/blob/nightly/docs/changelog-released.md which explains this new feature
3) Yes, you can find out more in the page about parameters in the docs: https://docs.modular.com/mojo/manual/parameters/
4) It's because the other parameters are inferred, so you cannot specify them from the callee side. 

I hope that was helpful. It can be hard to follow as the mojo language is growing very fast.

tjdwill:
> I hope that was helpful. It can be hard to follow as the mojo language is growing very fast.


It was, thank you. I'll spend some time re-reading the docs.

Also, if I have general/further questions outside of a pull request context, would those go on the Discussions page?

gabrieldemarmiesse:
I would advise to use the discord for general questions. People are very active there.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

PR #2957, opened 2024-06-05
[stdlib] Cleanup sort module

mzaks:
Just a few renaming and removal of the Comparable elements list sorting section as it can be delegated to other sort function

### DIFF
--- a/stdlib/src/builtin/sort.mojo
+++ b/stdlib/src/builtin/sort.mojo
@@ -27,6 +27,7 @@ from sys import bitwidthof
 alias _cmp_fn_type = fn[type: AnyTrivialRegType] (type, type) capturing -> Bool
 
 
+@always_inline
 fn _insertion_sort[
     type: AnyTrivialRegType, cmp_fn: _cmp_fn_type
 ](array: Pointer[type], start: Int, end: Int):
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/sort.mojo
```diff
@@ -210,15 +215,16 @@ fn _quicksort[
 # ===----------------------------------------------------------------------===#
 # partition
 # ===----------------------------------------------------------------------===#
+@always_inline
```
laszlokindrat, line 218: +1 on always inlining the hidden methods, but I'm concerned about potential code bloat with decorating the public APIs. Did you benchmark the effects of these, and if so, did you find a perf reason for doing this?

mzaks: Totally agree with you. The implication of code bloat slipped my mind. I removed the always inline decorators from public API.

laszlokindrat (approved):
Makes sense, thanks for fixing this!

laszlokindrat:
!sync

soraros:
It's probably too late. I think we should make the element type always inferred:
```mojo
sort[T: Comparable, cmp: fn(T, T) -> Bool](x: S[T]): ...
```
to
```mojo
sort[T: Comparable, //, cmp: fn(T, T) -> Bool](x: S[T]): ...
```

soraros:
@laszlokindrat Related, is there any reason why a generic `_cmp_fn_type` is used instead of `fn(type, type) -> Bool`?

PR #2959, opened 2024-06-05
[stdlib] Delay string creation in debug_assert for better perf

gabrieldemarmiesse:
See https://github.com/modularml/mojo/pull/2945#discussion_r1625636497 for background.

By delaying the conversion to string and the concatenation after the `if` statement, we make sure that there is no additional instruction when not using `-D MOJO_ENABLE_ASSERTIONS`. Futhermore, in platforms where using `.alloc` is problematic, it can help.

@laszlokindrat since you were participating in the conversation.


stdlib/src/builtin/debug_assert.mojo
```diff
@@ -23,7 +23,9 @@ from builtin._location import __call_location, _SourceLocation
 
 
 @always_inline
-fn debug_assert[stringable: Stringable](cond: Bool, msg: stringable):
+fn debug_assert[
+    *stringable: Stringable
+](cond: Bool, *message_parts: *stringable, sep: StringLiteral = " "):
```
laszlokindrat, line 28: Please don't use `StringLiteral` for runtime arguments. We will have `StaticString` soon, so you can wait for that, or move this to a parameter.

laszlokindrat: Or better yet, just don't have this functionality. Users can manually add spaces if they really need to.

laszlokindrat (approved):
Very nice, thank you!

patrickdoc:
!sync

Issue #933, opened 2023-09-27
[mojo-compiler] CompTime interpreter should be able to fold `pop.call_llvm_intrinsic`


soraros:
### Bug description

`math.bit` functions doesn't run at compile time.

### Steps to reproduce

Consider the following code:

```mojo
import math.bit as bit

fn main():
  alias n = bit.ctlz(10)

```

It produces the following error message:

```
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:70:1: error: no viable expansions found
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:70:1: note:   call expansion failed - no concrete specializations
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:51:1: note:     no viable expansions found
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:65:57: note:       call expansion failed - no concrete specializations
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:29:1: note:         no viable expansions found
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:42:18: note:           call expansion failed - no concrete specializations
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:62:5: note:             no viable expansions found
/__w/modular/modular/Kernels/mojo/builtin/_startup.mojo:63:22: note:               call expansion failed - no concrete specializations
/workspaces/ubuntu/hello.mojo:3:1: note:                 no viable expansions found
def main():
^
/workspaces/ubuntu/hello.mojo:4:3: note:                   failed to evaluate 'apply'
  alias n = bit.ctlz(10)
  ^
/__w/modular/modular/Kernels/mojo/math/bit.mojo:81:1: note:                     failed to interpret function @$math::$bit::ctlz($builtin::$int::Int)_concrete
/__w/modular/modular/Kernels/mojo/math/bit.mojo:90:44: note:                       failed to fold operation pop.call_llvm_intrinsic(10 : index, #pop<simd false> : !pop.scalar<bool>)
mojo: error: failed to run the pass manager
```

### System information

```shell
- OS: Docker on Intel macOS
- mojo -v: mojo 0.3.0 (f64f9601)
- modular -v: modular 0.1.4 (6b54d308)
```

Mogball:
Thanks for filing! We haven't implement compile-time interpretation of LLVM intrinsics, and it's not clear that's a worthwhile endeavor. That said, there are other ways we can make this work.

[...]

PR #2960, opened 2024-06-05
[stdlib] Fix `String.split()` and start fixing `String.__len__()`

### LINKED ISSUE #2880: [Feature Request] [stdlib] Make `String`'s `split` method accept `None` as the separator
### Review Mojo's priorities

### What is your request?

Currently, `String`'s `split` method does not support `None` as the separator to split on. In Python it's fully possible to perform splits such as `some_string.split(None)` and `some_string.split(None, maxsplit=1)` and `some_string.split(None, 2)`. This is the same as not providing any `sep` at all.

Python interoperability

### Any other details?

_No response_

martinvuyk:
Fix string split to take in NoneType as input and its body to use `isspace(self)` method. Fixes #2880

Leave the fix for `lstrip` and `rstrip` to use `String.isspace()` as `TODO` once llvm intrinsics can be used at comp time

Added a method `String.byte_length()` but left the builtin `String.__len__()` alone for now since many other methods assume it returns byte length and it will require mayor refactoring of too many places for 1 PR. Added deprecation warning to `_byte_length()`


martinvuyk:
@laszlokindrat no idea what this error is. 

When calling `fn lstrip(self) -> String:`
```text
/Users/runner/work/mojo/mojo/stdlib/src/builtin/string.mojo:1874:8: note:                             failed to interpret function @stdlib::builtin::string::String::lstrip(stdlib::builtin::string::String)
    fn lstrip(self) -> String:
       ^
/Users/runner/work/mojo/mojo/stdlib/src/builtin/string.mojo:1881:18: note:                               failed to evaluate call
        for s in self:
                 ^
/Users/runner/work/mojo/mojo/stdlib/src/builtin/string.mojo:1194:8: note:                                 failed to interpret function @stdlib::builtin::string::String::__iter__[__mlir_type.i1,__mlir_type.!lit.lifetime<*(0,0)>](stdlib::builtin::string::String%)_REMOVED_ARG
    fn __iter__(ref [_]self) -> _StringIter[__lifetime_of(self)]:
       ^
/Users/runner/work/mojo/mojo/stdlib/src/builtin/string.mojo:1200:48: note:                                   failed to evaluate call
        return _StringIter[__lifetime_of(self)](
                                               ^
/Users/runner/work/mojo/mojo/stdlib/src/builtin/string.mojo:700:8: note:                                     failed to interpret function @stdlib::builtin::string::_StringIter::__init__(stdlib::builtin::string::_StringIter[$0, $1, $2]=&,stdlib::memory::unsafe_pointer::UnsafePointer[stdlib::builtin::simd::SIMD[{uint8}, {1}], {{0}}],stdlib::builtin::int::Int)_REMOVED_ARG,forward=1
    fn __init__(
       ^
/Users/runner/work/mojo/mojo/stdlib/src/builtin/string.mojo:708:31: note:                                       failed to evaluate call
            if _utf8_byte_type(unsafe_pointer[i]) == 1:
                              ^
/Users/runner/work/mojo/mojo/stdlib/src/builtin/string.mojo:[66](https://github.com/modularml/mojo/actions/runs/9390428176/job/25860270012?pr=2960#step:9:67)7:4: note:                                         failed to interpret function @stdlib::builtin::string::_utf8_byte_type(stdlib::builtin::simd::SIMD[{uint8}, {1}])
fn _utf8_byte_type(b: UInt8) -> UInt8:
   ^
/Users/runner/work/mojo/mojo/stdlib/src/sys/intrinsics.mojo:192:10: note:                                           failed to fold operation pop.call_llvm_intrinsic{fastmathFlags: #pop<fmf none>, hasSideEffects: false, intrin: "llvm.ctlz" : !kgen.string}(#pop<simd 223> : !pop.scalar<ui8>, false)
        ](arg0, arg1)
         ^
mojo: error: failed to run the pass manager```

**Comment by laszlokindrat** [comment]
@martinvuyk Something (not sure where) is calling `lstrip` at compile time, but it can only work at runtime because its implementation somewhere uses an llvm intrinsic, and we don't support them at compile time yet. This is a known limitation that we are planning to lift.

**Comment by martinvuyk** [comment]
@laszlokindrat ready for review :D

File: .gitignore [hunk 1]
```diff
@@ -19,3 +19,5 @@ venv.bak/
 
 # MacOS
 .DS_Store
+
+install_id
```
laszlokindrat, line 23: Is this accidental?

stdlib/src/builtin/string.mojo
```diff
@@ -1785,43 +1772,41 @@ struct String(
...
+        ).split()  # ["hello", "world"]
         ```
         .
         """
-        # TODO: implement and document splitting adjacent universal newlines:
-        # _ = String(
-        #     "hello \\t\\n\\r\\f\\v\\x1c\\x1e\\x85\\u2028\\u2029world"
-        # ).split()  # ["hello", "world"]
 
+        _ = sep
```
laszlokindrat, line 1783: What is this for?

stdlib/src/builtin/string.mojo
```diff
@@ -1974,8 +1959,12 @@ struct String(
         Returns:
             A copy of the string with no trailing whitespaces.
         """
-        # TODO: should use self.__iter__ and self.isspace()
         var r_idx = len(self)
+        # TODO: should use this once llvm intrinsics can be used at comp time
```
laszlokindrat, line 1963: Here and in the other TODOs: Could you please find the (or open a new) ticket for tracking this, and reference it here?

stdlib/src/builtin/string.mojo
```diff
@@ -1666,49 +1681,22 @@ struct String(
...
-                return False
-            elif no_null_len == 2 and not _compare(
-                ptr, next_line.unsafe_ptr(), 2
-            ):
-                return False
-            elif no_null_len == 3 and not (
-                _compare(ptr, unicode_line_sep.unsafe_ptr(), 3)
-                or _compare(ptr, unicode_paragraph_sep.unsafe_ptr(), 3)
-            ):
+            if not s.isspace():
```
laszlokindrat, line 1699: Given that you now have a `StringSlice.isspace()` implementation, why is a loop needed here? Could you just simply `return self.as_string_slice().isspace()` here?

stdlib/src/utils/string_slice.mojo
```diff
@@ -312,3 +313,45 @@ struct StringSlice[
...
+            return True
+        elif no_null_len == 2 and _compare(ptr, next_line.unsafe_ptr(), 2):
+            return True
+        elif no_null_len == 3 and (
+            _compare(ptr, unicode_line_sep.unsafe_ptr(), 3)
+            or _compare(ptr, unicode_paragraph_sep.unsafe_ptr(), 3)
+        ):
+            return True
+        _ = next_line, unicode_line_sep, unicode_paragraph_sep
+        return False
```
laszlokindrat, line 357: This implementation only works for a single unicode character, right? It would be better to not have that restriction, given that String doesn't have it either.

laszlokindrat (commented):
Thanks for the patch! I have a couple questions, but I like the direction you're going! Could you please rebase on the latest nightly?

.gitignore
martinvuyk: oh yeah forgot to take it out, it was a couple of weeks ago when it kept being created

stdlib/src/builtin/string.mojo
martinvuyk: I think not since `StringSlice.isspace()` uses the length of the slice as a discriminant
```mojo
        @always_inline
        fn _compare(
            item1: UnsafePointer[UInt8], item2: UnsafePointer[UInt8], amnt: Int
        ) -> Bool:
            var ptr1 = DTypePointer(item1)
            var ptr2 = DTypePointer(item2)
            return memcmp(ptr1, ptr2, amnt) == 0

        var no_null_len = len(self)
        var ptr = self.unsafe_ptr()
        if no_null_len == 1 and _isspace(ptr[0]):
            return True
        elif no_null_len == 2 and _compare(ptr, next_line.unsafe_ptr(), 2):
            return True
        elif no_null_len == 3 and (
            _compare(ptr, unicode_line_sep.unsafe_ptr(), 3)
            or _compare(ptr, unicode_paragraph_sep.unsafe_ptr(), 3)
        ):
            return True
```

stdlib/src/utils/string_slice.mojo
martinvuyk: oh so you mean moving the whole implementation into `StringSlice.isspace()`? That would mean having to iterate over the `StringSlice` checking for utf8 byte sequence type etc. , so basically moving/copying the `_StringIter` implementation over to `_StringSliceIter` and using that ? Some small overhead might be incurred when someone first calls String.__iter__ and then does .isspace() on the StringSlice though. Should I do that anyway?

stdlib/src/builtin/string.mojo
martinvuyk: Added the issue number #933

stdlib/src/utils/string_slice.mojo
laszlokindrat: It would be better to do that right away. It's a pattern we're trying to move to: implement for `StringSlice` and delegate from `String`. It will also reduce the amount of churn we have to do internally to review and land PRs. Thanks!

martinvuyk:
Hey @laszlokindrat I ended up touching a lot more places than I thought I'd have to. Changed `String._byte_length()` to `String.byte_length()` since it is a method that is pretty commonly used before using pointers.

I used a `@deprecated` decorator to help me find all places in the stdlib that use `String.__len__` assuming it's byte_length before doing pointer stuff and changed it to `String.byte_length()`.

I also added some code in places where we should use unicode code point indexing and currently assume ASCII, but we still need issue #933 solved to be able to switch to that. And also the code is the first thing that came to me, not necessarily efficient.

laszlokindrat:
Thanks! Can we split up this patch then? The length related stuff will touch a lot more things internally, and it's also semantically separate.

martinvuyk:
I honestly don't know where to even split this since `fn split()` needs byte_length and so does `_StringSliceIter` instantiation. I didn't change any behavior regarding length methods except for `StringSlice.__len__` that was already marked as returning unicode length
```mojo
    fn __len__(self) -> Int:
        """Nominally returns the _length in Unicode codepoints_ (not bytes!).

        Returns:
            The length in Unicode codepoints.
        """
        # FIXME(MSTDL-160):
        #   Actually perform UTF-8 decoding here to count the codepoints.
        return len(self._slice)
```
```mojo
    fn __len__(self) -> Int:
        """Nominally returns the _length in Unicode codepoints_ (not bytes!).

        Returns:
            The length in Unicode codepoints.
        """
        var unicode_length = self.byte_length()

        for i in range(unicode_length):
            if _utf8_byte_type(self._slice[i]) == 1:
                unicode_length -= 1

        return unicode_length
```

> The length related stuff will touch a lot more things internally

That's what I was afraid of. For now the only breaking change is `StringLiteral`, `StringSlice` and `String` got their `_byte_length()` method renamed to `byte_length()` to be used before doing pointer stuff (using the internal method directly was a bad idea anyway). I could also just add a `_byte_length()` method with a deprecation warning to all WDYT?

laszlokindrat:
Let's start with the `_byte_length` -> `byte_length` renaming, as well as all the miscellaneous updates you did to move from `len` to `byte_length`. Then we can iterate again. Thanks for you patience with this, I promise it will eventually land :)

stdlib/src/builtin/string.mojo
```diff
@@ -1008,11 +1008,10 @@ struct String(
         Construct a String from several `Formattable` arguments:
 
         ```mojo
-        from testing import assert_equal
-
         var string = String.format_sequence(1, ", ", 2.0, ", ", "three")
-
-        assert_equal(string, "1, 2.0, three")
+        print(string) # "1, 2.0, three"
```
laszlokindrat, line 1012: Could you please remove these and the commented section below?

laszlokindrat (commented):
Thanks! Can you please add a changelog entry describing the changes?

stdlib/src/builtin/string.mojo
martinvuyk: do you mean removing the whole example or undoing my changes ?

laszlokindrat: This particular `print` statement, as well as all the `#FIXME ` below that has large code examples. Since it's commented out, we cannot test it.

stdlib/src/builtin/string.mojo
```diff
@@ -1094,13 +1104,27 @@ struct String(
         var start: Int
         var end: Int
         var step: Int
-        start, end, step = span.indices(len(self))
+        # FIXME: this should work with unicode codepoints, but this doesn't yet
+        # work at compile time because of issue #933
```
laszlokindrat, line 1108: If you want to leave a reminder, to improve this, it's better to just do this:
```suggestion
        # TODO(#933): implement this for unicode when we support llvm intrinsic evaluation at compile time
```

stdlib/src/builtin/string.mojo
martinvuyk: > This particular print statement

if I do that then the user would only see
Examples:
```mojo
var string = String.format_sequence(1, ", ", 2.0, ", ", "three")
```
instead of seeing the result of the print, since the assert_equals gets hidden by the `%#`

> all the #FIXME  below that has large code examples

ok

laszlokindrat: oh sorry, this is an example, nevermind 🤦

laszlokindrat: So then you can either have an assert or the print. Is there a reason why this example needs to be updated at all?

docs/changelog.md
```diff
@@ -139,6 +139,11 @@ what we publish.
 
 - Added `StringSlice(..)` initializer from a `StringLiteral`.
 
+- Added a `byte_length()` method to `String`, `StringSlice`, and `StringLiteral`
+and deprecated their private `_byte_length()` methods. Added a warning to
+`String.__len__` method that it will return length in Unicode codepoints in the
+future and `StringSlice.__len__` now does return the Unicode codepoints length.
```
laszlokindrat, line 145: Please make sure you give credit to yourself :)

docs/changelog.md
```diff
@@ -139,6 +139,11 @@ what we publish.
 
 - Added `StringSlice(..)` initializer from a `StringLiteral`.
 
+- Added a `byte_length()` method to `String`, `StringSlice`, and `StringLiteral`
+and deprecated their private `_byte_length()` methods. Added a warning to
+`String.__len__` method that it will return length in Unicode codepoints in the
```
laszlokindrat, line 144: IIUC, this change doesn't actually do this, right? Looks like `String.__len__` just return the byte length right now.

docs/changelog.md
```diff
@@ -139,6 +139,11 @@ what we publish.
 
 - Added `StringSlice(..)` initializer from a `StringLiteral`.
 
+- Added a `byte_length()` method to `String`, `StringSlice`, and `StringLiteral`
```
laszlokindrat, line 142: In the PR title and description you mention some changes to `String.split` I don't see them in this patch. Can you please update it, or describe them here?

stdlib/src/builtin/string.mojo
martinvuyk: I found it ugly , IMO asserts shouldn't be used in examples and we should stick with print and showing with a comment what happens more similarly to what is done in Python. But I also like the idea of hiding an assert to ensure documentation is up to date with method behavior

laszlokindrat: Okay, I guess I agree that print is probably nicer in an example. If you change it, please remove the commented out asserts below.

docs/changelog.md
martinvuyk: It is actually just a small fix adding `sep: NoneType = None` as default for `String.split()`

stdlib/src/builtin/string.mojo
martinvuyk: They aren't commented, they will run as tests if the file [gets linked in `mojo test`](https://docs.modular.com/mojo/tools/testing#codeBlockContent_m3Ux), do you still want me to remove them?

laszlokindrat: whoah, TIL, thanks, it's fine

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

PR #2963, opened 2024-06-05
[stdlib] Remove SliceNew.__getitem__

### LINKED ISSUE #2953: [Feature Request] Remove `SliceNew.__getitem__`
### Review Mojo's priorities

### What is your request?

In a similar line of thinking as c8aa372, we should probably remove `SliceNew.__getitem__` as Python slices are not subscript-able, and the current implementation is broken anyways.

```mojo
var s = slice_new(1, 10)
print(s[-1]) # prints 0
```

This functionality already serves the same purpose as `range` and doesn't exist in Python.

### Any other details?

_No response_

bgreni:
Fixes #2953

slice in python does not implement `__getitem__` and with the introduction of `indices()` it will also no longer be used when we've migrated to `SliceNew` in the stdlib containers. The current implementation is also incorrect when using negative indices anyways.

### DIFF
--- a/stdlib/src/builtin/builtin_slice.mojo
+++ b/stdlib/src/builtin/builtin_slice.mojo
@@ -332,18 +332,6 @@ struct SliceNew(Stringable, EqualityComparable):
 
         return len(range(self.start.value(), self.end.value(), self.step))
 
-    @always_inline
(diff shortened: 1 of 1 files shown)


JoeLoser (approved):

JoeLoser:
!sync

PR #2965, opened 2024-06-06
[stdlib] Make `InlineArray` call its elements' destructor

gabrieldemarmiesse:
Fix this issue: 
*  https://github.com/modularml/mojo/issues/2869

A little explanation of what is going on here, this should make the review easier.

### Modifications to `InlineArray`
`InlineArray` made use of the `@value` decorator, which generates, in addition to a constructor, the following methods: `__moveinit__`, `__copyinit__` and `__del__`. While it's easy to understand what is generated when working with attributes that are values, it's less clear what is happenning when working with plain MLIR types like we do here.

The next step is to manually define those to make sure they do what we want. **We will always assume that the `InlineArray` is completely initialized**. Thus when copying, moving or deleting, we must copy, move and delete each element in the `InlineArray` too.

To make the implementation easier, instead of using an `UnsafePointer` and its methods to control the lifecycle of the values (`move_from_pointee` & friends), we fill the array with `UnsafeMaybeUninitialized`. This struct has methods that allows us to control the lifecycle without necessarily work with pointers.

### What do we do with `InlineArray[...](unsafe_uninitialized=True)`?

This method is a massive footgun. This is because throughout the methods of `InlineArray`, we always assume that the values are initialized. So doing any moving, copying or deleting with uninitialized memory is UB. 

This is not much of an issue for very trivial types where moving and copying are just copying bits and the destructor does nothing. This can still cause UB if the user uses some uninitialized memory but it's fine if the caller is careful. For example, in the stdlib, we use `unsafe_uninitialized` to store some `UInt8` here and there and it's fine. 

As such, the constructor was kept, but an extensive docstring was added to explain in which situation it is okay to use it.

### How `InlineArray` is now used by `InlineList`

`InlineList` wants to carefully manage the lifecycle of each element in the `InlineArray` while still making the `InlineArray` believe that it owns the lifecycle of its elements. In those kind of situation, `UnsafeMaybeUninitialized` is the right struct to use. 

So we use an `InlineArray[UnsafeMaybeUninitialized[ElementType]]` inside `InlineList`. I know it can be confusing since `InlineArray` itself also uses `UnsafeMaybeUninitialized` but this shouldn't be visible from outside `InlineArray` so it should be fine. Basically it disables all lifecycle methods, and `__del__` is a no-op. 

This allows `InlineList` to manually control the lifecycle of each elements which has an index below the size (above the size of the `InlineList`, it's truly uninitialized memory).

### Some guideline for the code review:

I advise to read the files in this order:

* The files `stdlib/test/test_utils/types.mojo` and `stdlib/test/collections/test_inline_list.mojo` have been refactored to make the tests of `InlineArray` use the `ValueDestructorRecorder`. This could have been a separate PR but I was too lazy to split this PR even further.
* The file `stdlib/test/utils/test_tuple.mojo` has a new test that ensure that all the elements in the `InlineArray` have their destructors called.
* The file `stdlib/src/utils/static_tuple.mojo` has all the changes that makes `InlineArray` use `UnsafeMaybeUninitialized` internally. While we could have technically use pointers, it makes writing code easier when controlling lifecycle.
* Finally, we switch the internal type of `InlineList` in `stdlib/src/collections/inline_list.mojo`. We now use `InlineArray[UnsafeMaybeUninitialized[ElementType]]` which allows the `InlineList` to control exactly the lifecycle. I repeat here: yes in can be confusing because overall, we use two `UnsafeMaybeUninitialized` on top of each other when using `InlineArray[UnsafeMaybeUninitialized[ElementType]]`, but the code is still readable, because one "belongs" to `InlineArray` and one "belongs" to `InlineList.`


gabrieldemarmiesse:
@ConnorGray it's ready for review, please take a look when you have the time :)

rd4com:
good job @gabrieldemarmiesse ,
do you think it would be a proper opportunity to add a taking move to `InlineList`?

```mojo
fn __moveinit__(inout self, owned other: Self):
        self._size = other._size
        other.size = 0 # taking move (see ._del)
        self._array = other._array^
 ```

**Comment by gabrieldemarmiesse** [comment]
@rd4com I think it can be added in another PR. Furthermore, the `__moveinit__` is not as trivial as written because only some elements in the underlying array must be moved (the one below the `self.size`)

**Comment by ConnorGray** [comment]
!sync

File: stdlib/src/utils/static_tuple.mojo [hunk 1]
```diff
@@ -257,7 +257,11 @@ struct InlineArray[
 
     # Fields
     alias type = __mlir_type[
-        `!pop.array<`, size.value, `, `, Self.ElementType, `>`
+        `!pop.array<`,
+        size.value,
+        `, `,
+        UnsafeMaybeUninitialized[Self.ElementType],
+        `>`,
```
ConnorGray, line 264: I think I understand the intention here, but I think this may not actually be required.

I think the state we want to aim for is:

* `InlineArray.__init__(unsafe_uninitialized)` is replaced with:

    ```mojo
    @staticmethod
    fn unsafe_uninitialized() -> InlineArray[UnsafeMaybeUninitialized[T]]:
        ...
    ```

* All the remaining `InlineArray.__init__(...)` constructors should (and currently do) always fully initialize the `InlineArray`. 
  
   (This is okay to do without making the `pop.array` storage type use `UnsafeMaybeUninitialized` because the compiler knows that inside an `__init__` method `self` and the fields of `self` won't be fully initialized anyway.

With this approach, all the methods off `InlineArray` will be able to assume one of the following two things is true:

1. The current `InlineArray` is fully initialized
2. The current `InlineArray` may not be fully initialized, but its element types are definitely wrapped in `UnsafeMaybeUninitialized[..]`, so that's okay.

We'd then also want to add this constructor for the programmer to "assert" that an uninitialized `InlineArray` is now fully initialized.

```
struct InlineArray[T]:
    fn __init__(
        inout self,
        owned unsafe_assume_initialized: InlineArray[UnsafeMaybeUninitialized[T]]
    ):
        # transmute the argument `InlineArray` into one that is now considered fully initialized
        ...
```

@gabrieldemarmiesse  What do you think of this?

I think it would eliminate the need for some of the changes that had to be made in the current iteration of this PR throughout the `InlineArray` implementation.

stdlib/src/collections/inline_list.mojo
```diff
@@ -245,5 +246,5 @@ struct InlineList[ElementType: CollectionElementNew, capacity: Int = 16](Sized):
             value: The value to append.
         """
         debug_assert(self._size < capacity, "List is full.")
-        self._array[self._size] = value^
+        self._array[self._size] = UnsafeMaybeUninitialized[ElementType](value^)
```
ConnorGray, line 249: **Question** Does this trigger the `abort()` in `UnsafeMaybeUninitialized.__moveinit__()`?

I wonder if this would be better:

```suggestion
        self._array[self._size].write(value^)
```

stdlib/src/collections/inline_list.mojo
```diff
@@ -20,6 +20,7 @@ from collections import InlineList
 """
```
ConnorGray, line 21: I love how clean the changes are in this file with the introduction of `UnsafeMaybeUninitialized` 🙂

ConnorGray:
Thank you for putting together such an excellent explanation for this work @gabrieldemarmiesse. This was a treat to review 🙂 

I left a suggestion boils down to basically I think we can keep the implementation of `InlineArray` itself mostly unchanged, but other than that this PR is straightforward and looks great 🔥 

Thank you again for pushing on this work Gabriel, this will make a huge difference in how precisely we can model the invariants of our collection types.

stdlib/src/collections/inline_list.mojo
gabrieldemarmiesse: The `InlineList` is very well tested. The `append` function too. So if this were to trigger an abort(), we would have seen it in the CI. For example there: https://github.com/modularml/mojo/blob/nightly/stdlib/test/collections/test_inline_list.mojo#L49

I believe there is no `__moveinit__` called on the `UnsafeMaybeUninitialized` because in Mojo, the `__init__` method always build the result "in-place", it's similar to return value optimization, thus there is no move.

The suggestion you made is shorter, so sure, I don't mind using it.

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: Using `UnsafeMaybeUninitialized` inside `InlineArray` is indeed not necessary, but it was intentional to modify the struct like this. The idea was to move away from pointers and use only `UnsafeMaybeUninitialized` which has a restricted use compared to pointers (notably, no pointer arithmetic). We still want to carefully control the lifecycle, and `UnsafeMaybeUninitialized` allows it.

But sure I can change it back and use plain pointers again, though I believe the code is less readable.

ConnorGray: I think a pointer-based implementation is still my personal preference, since it limits the number of types involved. But I see the merits of your argument too, and I don't want to hold up this excellent work on a small difference in perspective like that 🙂

So I'm going to resync this and merge it, and let this shake out—we can always revisit and cleanup later if there's a readability or maintainability problem in practice.

And I agree, we can / should add `@staticmethod fn unsafe_uninitialized()` in a separate PR.

ConnorGray:
!sync

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: I don't mind reverting one day to pointers if we see that this is causing some issues down the road :)

ConnorGray:
Re-opening to try to merge this internally again.

ConnorGray:
!sync

gabrieldemarmiesse:
> Re-opening to try to merge this internally again.

I saw that it got reverted for performance reasons. If you can pinpoint the reason or have tools to help pinpoint the issue, I'd be happy to know, just to satisfy my curiosity

gabrieldemarmiesse:
@ConnorGray what's the status on this pull request? Should I revert the change to `InlineArray` and use bare pointers instead of `UnsafeMaybeUninitialized` or is it something that you are already working on internally?

stdlib/src/utils/static_tuple.mojo
gabrieldemarmiesse: I'll revert it to see if it fixes the IR

gabrieldemarmiesse:
@ConnorGray I reverted the changes in `InlineArray` that may have impacted the IR generation. Let me know if that works better now :)

JoeLoser:
!sync

stdlib/src/collections/inline_list.mojo
```diff
@@ -20,6 +20,7 @@ from collections import InlineList
...
 
 from sys.intrinsics import _type_is_eq
+from memory.unsafe import UnsafeMaybeUninitialized
```
JoeLoser, line 23: **Suggestion** Shouldn't this import be `from memory.maybe_uninitialized import UnsafeMaybeUninitialized`?  The type isn't in `memory.unsafe` and we don't hard error on missing imports.

JoeLoser: I just fixed this up internally FYI.

gabrieldemarmiesse: Yeah, imports are still a bit funky

PR #2968, opened 2024-06-06
Use exec zsh instead of source ~/.zshrc

camachomarco:
Zsh will otherwise throw an error.


gabrieldemarmiesse:
Can you give the error that you got?

gabrieldemarmiesse:
It seems you are using https://github.com/romkatv/zsh4humans which is deprecated and somewhat unmaintained. If a user is using a regular zsh install, then this error will never pop up. I would recommend using a common and maintained variant of zsh to avoid the error. Executing `zsh` reloads another session, which is not what users want most of the time. 

I don't think we can justify changing a working worflow to adapt it for a minority of users who happened to be using unmaintained software.

camachomarco:
Will look into using a newer version of ZSH to make sure the error doesn't happen then. Thank you for your recommendation.

gabrieldemarmiesse:
No worries! I hope that people who had a similar problem will find your PR through google :)

PR #2969, opened 2024-06-06
[docs] Fix docstring in `swap.mojo`

Brian-M-J:
The code says that `T` is constrained to Movable types, so the docstring should also reflect this. It only uses the `^` operator in its body.


gabrieldemarmiesse (approved):
Looks good!

JoeLoser (approved):

JoeLoser:
!sync

PR #2971, opened 2024-06-06
[stdlib] Give the `CollectionElementNew` trait to `UnsafeMaybeUninitalized`

gabrieldemarmiesse:
This is part of the PR https://github.com/modularml/mojo/pull/2965

@ConnorGray I think you're the right person for a review :)


stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -38,6 +38,17 @@ struct UnsafeMaybeUninitialized[ElementType: CollectionElement](
...
+    fn __init__(inout self, *, other: Self):
+        """Leave this memory location uninitialized. No copy is actually performed here.
+
+        If you wish to actually copy the internal value, use `copy_from()` instead.
+
+        Args:
+            other: The object to copy.
+        """
+        self = Self()
+
```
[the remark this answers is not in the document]
gabrieldemarmiesse: The vision I have with `UnsafeMaybeUninitialized` is to be able to handle all lifecycle operations offered by `UnsafePointer.xxx_pointee_xxx(...)` without needing to use any pointer. It's easier to manipulate I believe as there are less concepts to manipulate here. We one with one struct, there is no `.alloc`, no possible pointer arithmetic, etc... I believe it's a more user-friendly entrypoint to users who want to manage memory themselves.

As such, the struct should be able to call any `UnsafePointer.xxx_pointee_xxx(...)`  method, removing the need for pointers.

Anyway that's my vision for this struct, but it's slighly unrelated to you question, I just wanted to communicate it.

To answer your question, I think you are right. Any lifecycle call should be very explicit about what it does, there is no need to even have traits on this types (not copyable, no movable, etc) I think it's just that at the moment, we're very limited about conditional conformance, this is why those methods are defined but we just abort if called. We could even do a "constraint[False]" to make it crash at compile time. I'll change it and make it crash at compile time if called.

gabrieldemarmiesse:
@ConnorGray  all done. I also think that it should be possible in a future PR to remove the requirements for the type that can be put inside an `UnsafeMaybeUninitialized`.

gabrieldemarmiesse:
Summary of the changes:
1) `UnsafeMaybeUninitialized` now takes `AnyType` as element. Added very specific traits to each method to allow this.
2) `UnsafeMaybeUninitialized` now has the `CollectionElementNew` trait, but none of the methods of this trait can actually be called because of ambiguity, and will fail at compile time.
3) Switch from `abort` to `constrained` for a faster fail. It's a nice QoL feature.
3) Removed `inout` from `initialize_pointee_explicit_copy` as we are not mutating the pointer but the value it points to. This is because, without this change, we can't do `self.unsafe_ptr().initialize_pointee_explicit_copy(...)` because it's mutating an rvalue, which is not allowed by the compiler (rightfully or not, I don't have an opinion on this).
4) Moved ValueDestructorRecorder in the file `stdlib/test/test_utils/types.mojo` where it can be used in other test files, it's a very useful struct.

@ConnorGray  it's ready for review, for real this time :)

ConnorGray:
!sync

stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -38,14 +36,42 @@ struct UnsafeMaybeUninitialized[ElementType: CollectionElement](
...
+            (
+                "You should never call the explicit copy constructor of"
+                " UnsafeMaybeUninitialized because it's ambiguous to copy"
+                " possibly uninitialized memory. Use"
+                " `UnsafeMaybeUninitialized.copy_from()` instead if you want to"
+                " trigger an explicit copy of the content of"
+                " UnsafeMaybeUninitialized. It has very specific semantics."
+            ),
+        ]()
+        self = Self()
```
ConnorGray, line 57: I think making this a compile time failure might prevent uses like `InlineArray[UnsafeMaybeUninitialized[T]]`, because that type would instantiate a call to this copy constructor.

The goal would be to avoid _dynamically_ doing a copy operation, but we need the compiler to accept code that "looks like" it could do a copy, because its up to the programming to make sure no actual copy gets attempted.

stdlib/src/memory/unsafe_pointer.mojo
```diff
@@ -343,7 +343,7 @@ struct UnsafePointer[
     @always_inline
     fn initialize_pointee_explicit_copy[
         T2: ExplicitlyCopyable
-    ](inout self: UnsafePointer[T, address_space], value: T2):
+    ](self: UnsafePointer[T, address_space], value: T2):
```
ConnorGray, line 346: 👍 , this makes sense.

ConnorGray:
Hey Gabriel, thanks for your work on this, I think this all looks good—I like that change to make this work with `AnyType`.

Re: the `constrainted[False]` change, I think that will make `InlineArray[UnsafeMaybeUninitialized[T]]` fail to compile. Considering fixing `InlineArray` / `InlineList` was the motivation for this type, I think we might to stick with `abort` for the time being.

We could investigate switching this to `constrained[False]` maybe in a different PR. (Or maybe even removing the `__init__(inout self, Self)` and `__moveinit__` constructors entirely?)

ConnorGray:
Awesome, thank you for the changes Gabriel! This looks good, I'll move forward with merging :)

ConnorGray:
!sync

Issue #2942, opened 2024-06-03
[stdlib] Remove `simd.splat`


JoeLoser:
Since `SIMD[…].splat(x)` is the same as `SIMD[…](x)`, we should be able to just remove the `splat` method entirely and update the callers to just use the constructor/init on `SIMD`.

devesh-2002:
May I take this issue?

JoeLoser:
> May I take this issue?

Of course, go for it. If you run into any issues, feel free to reach out!

martinvuyk:
Is there a difference in allocation between these two?
```mojo
struct Item[T: DType]:
  vec: SIMD[T, 64]

  fn something1(self):
    self.vec = SIMD[T, 64](1)

  fn something2(self):
    self.vec = self.vec.splat(1)
```
If there is no difference then the change seems fine to me, otherwise I think it should stay as a means to splat any value accros a SIMD vector without allocating a new one. Or at least have SIMD.splat be inplace

JoeLoser:
@adakkak do you have thoughts here?

LJ-9801:
@JoeLoser @martinvuyk In SSE these would equal

```c
// let's say we are using float32 with simd size of 4
// here, 1 is broadcasted to all 4 float32
_m128 simd4 = _mm_set1_ps(1) 
// ==  SIMD[float, 4].splat(1)
// ==  SIMD[float, 4](1)

// expressing this in terms of union
union simd_floatx4 {
    _m128 simd;
    float f[4];
}
```

I don't think there are any other intrinsic that distinguish those two (creating simd with a value or filling it with a value). Ultimately I think they do the same thing: give me 16 bytes of stack memory and fill some values in it.

**PS:** I think I used some `.splat()` in the bit module so whoever works on this might need to change those as well.

abduld:
`SIMD[T, 64](1)` and `SIMD[T, 64].splat(1)` generate the same assembly. So removing `.splat` is fine

PR #2990, opened 2024-06-08
[stdlib] Removed SIMD.splat

devesh-2002:
Issue #2942,
I have removed the .splat method from SIMD. Please look into it, and let me know if there are any changes needed.

### DIFF
--- a/docs/manual/parameters/index.ipynb
+++ b/docs/manual/parameters/index.ipynb
@@ -283,7 +283,7 @@
     "can call the `splat()` method like this:\n",
     "\n",
     "```mojo\n",
-    "GenericArray[Float64](8, 0)\n",
+    "GenericArray[Float64].splat(8, 0)\n",
     "```\n",
     "\n",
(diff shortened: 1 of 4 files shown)


JoeLoser:
!sync

docs/manual/parameters/index.ipynb
```diff
@@ -282,7 +283,7 @@
     "can call the `splat()` method like this:\n",
     "\n",
     "```mojo\n",
-    "GenericArray[Float64].splat(8, 0)\n",
+    "GenericArray[Float64](8, 0)\n",
```
arthurevans, line 286:
```suggestion
    "GenericArray[Float64].splat(8, 0)\n",
```

arthurevans: We need to revert this line, this is not being called on SIMD; this is calling a static method declared above on LL277-278 demonstrating the use of the `Self` type.

JoeLoser:
!sync

PR #2991, opened 2024-06-08
[docs] Explicitly provide Int parameter when calling repeat in docs example

sbrunk:
I encountered a compile error while going through the examples to learn Mojo.

```mojo
fn repeat[MsgType: Stringable, count: Int](msg: MsgType):
    @parameter
    for i in range(count):
        print(msg)

repeat[3](3)
```

It looks like the compiler currently can't infer the parameter correctly otherwise and tries to use the `3` for `MsgType`: `cannot pass 'IntLiteral' value, expected 'Stringable' in call parameter`

Adding the parameter explictly makes the example compile again:
```mojo
repeat[Int, 3](3)
```

Note that it works fine if we change `MsgType` to be an infer-only parameter, so that could be an alternative. But infer-only parameters are only introduced a bit below in the docs.

```mojo
fn repeat[MsgType: Stringable, //, count: Int](msg: MsgType):
    @parameter
    for i in range(count):
        print(msg)

repeat[3](3)
```

### DIFF
--- a/docs/manual/parameters/index.ipynb
+++ b/docs/manual/parameters/index.ipynb
@@ -158,7 +158,7 @@
     "    for i in range(count):\n",
     "        print(msg)\n",
     "\n",
-    "repeat[3](3)"
+    "repeat[Int, 3](3)"
    ]
   },
   {
--- a/docs/manual/parameters/index.ipynb
+++ b/docs/manual/parameters/index.ipynb
@@ -42,7 +42,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 25,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -67,7 +67,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 26,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [
     {
@@ -139,16 +139,15 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 27,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "3\n",
-      "3\n",
-      "3\n"
+      "42\n",
+      "42\n"
      ]
     }
    ],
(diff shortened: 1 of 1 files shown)


soraros:
Let's opt for infer-only as that's the "more correct" typing of this function.

jackos:
Thanks heaps for raising @sbrunk, like @soraros said this is a good opportunity to explain the new `//` inferred-only syntax in the notebook. If you could edit your change to look like this it would be much appreciated:

```mojo
fn repeat[MsgType: Stringable, count: Int](msg: MsgType):
    @parameter
    for i in range(count):
        print(msg)

# Must use keyword parameter for `count`
repeat[count=2](42)
```
This updated function takes any `Stringable` type, so you can pass it an `Int`,
`String`, or `Bool` value.

You can't pass the `count` as a positional keyword without also specifying `MsgType`.
If you put `//` after `MsgType` it declares it as always being inferred by the argument.
Now you can pass the following parameter `count` positionally:

```mojo
fn repeat[MsgType: Stringable, //, count: Int](msg: MsgType):
    @parameter
    for i in range(count):
        print(msg)

# MsgType is always inferred, so first positional keyword `2` is passed to `count`
repeat[2](42)
```

Thanks very much for raising.

jackos (approved):
Thanks heaps for raising! I added fixes for a few other cells that also weren't compiling, and a description for the new inferred-only syntax with `//`

jackos:
!sync

jackos:
Cheers @sbrunk no worries, inferred-only syntax was only just added in yesterday's release!

arthurevans:
Thanks for the fix @sbrunk and @jackos. That was my bad--I think I switched the parameter order to match our style guide and didn't re-run the cell.

I think we might want rethink the examples. My idea here was to provide the simplest possible example of a generic function, but if we have to introduce parameter inference and infer-only parameters at the same time, it kind of muddies the waters. 

Alternately, we could link forward to those sections (https://docs.modular.com/mojo/manual/parameters/#parameter-inference, https://docs.modular.com/mojo/manual/parameters/#infer-only-parameters). I'm just not sure if it's too much. I know a lot of folks have trouble with the generics syntax. 

WDYT?

jackos:
Hey @arthurevans yeah sounds good, can remove the infer-only part here and just leave it as a keyword-only parameter

PR #2994, opened 2024-06-08
[stdlib] Use infer-only in `sort.mojo`

soraros:


stdlib/src/builtin/sort.mojo
```diff
@@ -282,45 +281,40 @@ fn sort[type: DType](inout buff: Pointer[Scalar[type]], len: Int):
...
-    fn _less_than_equal[ty: AnyTrivialRegType](lhs: ty, rhs: ty) -> Bool:
-        return rebind[Scalar[type]](lhs) <= rebind[Scalar[type]](rhs)
+    fn _less_than_equal(lhs: Scalar[type], rhs: Scalar[type]) -> Bool:
+        return lhs <= rhs
 
-    _quicksort[Scalar[type], _less_than_equal](buff, len)
+    _quicksort_reg[_less_than_equal](buff, len)
 
 
 fn sort(inout list: List[Int]):
```
mzaks, line 290: Could you delete this API as well?
It is covered by the generic `sort[type: ComparableCollectionElement, //](inout list: List[type]):`
And I think there should be no downsides if we remove this very specialised one.

soraros: It's slightly different from the generic `sort` as it uses the `_quicksort_reg` variant which contains optimisations for register types. I'm not sure if that's important or not. Shall we decide that in a later PR?

JoeLoser (approved):
Nice cleanup, thank you!

JoeLoser:
!sync

soraros:
@JoeLoser Is this blocked by something?

JoeLoser:
> @JoeLoser Is this blocked by something?

Likely, yes. When I imported it, it was hitting compiler bugs with inferred parameters in internal code. We had a company onsite all week this week, so I haven't dug into it. Will do next week and circle back.

soraros:
@JoeLoser `sort` is seeing a lot of traffic from the team these days, maybe I should try this PR on a later date?

JoeLoser:
> @JoeLoser `sort` is seeing a lot of traffic from the team these days, maybe I should try this PR on a later date?

Let's revisit this soon.  @Dan13llljws is actively working in this area and upgrading `sort` to get off of `Pointer` and needing to write some memory-only adaptor types for the comparison to avoid a function subtyping issue in the compiler.  That may actually unblock this infer-only work, so let's circle back soon (early next week likely).

soraros:
@JoeLoser I guess this is not needed anymore after 755f0de027f1bba0ca0c1aff90c32cff16cd19a2?

JoeLoser:
> @JoeLoser I guess this is not needed anymore after [755f0de](https://github.com/modularml/mojo/commit/755f0de027f1bba0ca0c1aff90c32cff16cd19a2)?

Agreed.  Looks like @Dan13llljws just did the infer-only work as he was upgrading these APIs.  Sorry we couldn't land this originally, but glad with the new state now!

PR #2998, opened 2024-06-09
[stdlib] Add method `List.format_to()`

gabrieldemarmiesse:
Going through the code of `utils/_format.mojo` I see a lot of overlap between this functionality and the file-like objects in Python. I think that in the future we can replace the formatter by a File-like object without any issue. For this reason, I changed the function `write_to` to be a method of `Formatter`, so that in the future, when we replace `Formatter` by a `FileLike` trait (or similar), the `write` method is already used.

I think it would be good to flush out the `io` python module. This would give us access to structs and traits that would have made the code in `_format.mojo` unnecessary. Writing to buffers / files is very common in python, there should be no need to re-invent it.

@ConnorGray for review


ConnorGray:
!sync

ConnorGray:
!sync

ConnorGray:
!sync

PR #3007, opened 2024-06-11
[stdlib] Specify alignment in UnsafePointer.alloc

### LINKED ISSUE #3006: [Feature Request] Specify alignment in UnsafeAlignment.alloc
### Review Mojo's priorities

### What is your request?

It looks like the new `UnsafePointer` type doesn't have a way of specifying `alignment` like `LegacyPointer` does in its `alloc` function

UnsafePointer: https://github.com/modularml/mojo/blob/652485ceb9332885a7537760dcc949bfe8b1e5a0/stdlib/src/memory/unsafe_pointer.mojo#L132
LegacyPointer: https://github.com/modularml/mojo/blob/nightly/stdlib/src/memory/unsafe.mojo#L383

I would like to migrate from LegacyPointer to UnsafePointer in a [lib](https://github.com/mojo-data/arrow.mojo/blob/dbb8c69f654a5e39932a77083b0b99068c5c9a95/arrow/arrow.mojo#L8) I'm working on that [needs alignment to be specified](https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding)

### Any other details?

_No response_

sa-:
Fixes https://github.com/modularml/mojo/issues/3006

Adds a function parameter in `UnsafePointer.alloc` to specify alignment at compile time.

Although I'm not sure how to test it because I couldn't find a way to convert `ptr.address` to an `Int`, but I'm open to ideas

Edit: Tests are now included


stdlib/test/memory/test_unsafepointer.mojo
```diff
@@ -169,6 +169,13 @@ def test_unsafepointer_address_space():
     p2.free()
 
 
+def test_unsafepointer_aligned_alloc():
+    var ptr = UnsafePointer[UInt8].alloc[alignment=64](1)
+    # Not sure how to test this. Maybe:
+    # assert_true(ptr.address % (64*8) == 0)
```
sa-, line 175: Small correction, this would just be `address % 64 == 0` instead of 64*8`

soraros:
I think you could add a parameter with default value to `alloc`.

sa-:
@soraros good point, I've done that now

stdlib/src/memory/unsafe_pointer.mojo
```diff
@@ -129,8 +129,11 @@ struct UnsafePointer[
 
     @staticmethod
     @always_inline
-    fn alloc(count: Int) -> Self:
-        """Allocate an array with default alignment.
+    fn alloc[alignment: Optional[Int] = None](count: Int) -> Self:
```
soraros, line 132:
```suggestion
    fn alloc[alignment: Int = alignof[T]()](count: Int) -> Self:
```

sa-:
Okay I've now added tests that work

stdlib/src/memory/unsafe_pointer.mojo
```diff
@@ -129,8 +129,11 @@ struct UnsafePointer[
 
     @staticmethod
     @always_inline
-    fn alloc(count: Int) -> Self:
-        """Allocate an array with default alignment.
+    fn alloc[alignment: Int = alignof[T]()](count: Int) -> Self:
+        """Allocate an array with specified or default alignment.
+
+        Parameters:
+            alignment: (Optional) The alignment in bytes of the allocated memory.
```
ConnorGray, line 136: **Suggestion** Remove "(Optional)" now that this is no longer literally an `Optional`.

ConnorGray:
!sync

stdlib/src/memory/unsafe_pointer.mojo
sa-: While the type isn't `Optional` it is optional for the user to specify it

I believe removing it would convey that the user is required to specify alignment

Though I'm open to removing it if you think we should after considering this

ConnorGray: Hi @sa-, your reasoning makes sense. This comes down to a matter of consistent style within stdlib code, and optional arguments and parameters are something we don't call out explicitly in general.

E.g. see this example of what we do for `print`:

CleanShot 2024-06-13 at 08 55 50@2x

sa-: Sounds good, i've removed it

ConnorGray: Thanks @sa-! 🙂

ConnorGray:
!sync

stdlib/src/memory/unsafe_pointer.mojo
sa-: Thanks to you as well for your time

ConnorGray:
!sync

PR #3013, opened 2024-06-11
[stdlib] Rename `_get_reference_unsafe` to `unsafe_get` in `InlineArray`

gabrieldemarmiesse:
And also add autoderef.

The renaming is to have something consistent with `List` which has the methods `unsafe_get` and `unsafe_get`.


JoeLoser (approved):

JoeLoser:
!sync

PR #3015, opened 2024-06-11
[stdlib] Add `<`, `<=`, `>`, `>=` to `UInt`

gabrieldemarmiesse:
We use here the unsigned variants of the comparison predicates. See https://mlir.llvm.org/docs/Dialects/IndexOps/#indexcmp-indexcmpop for more info.


JoeLoser (commented):
Nice!  Should we add `==` and `!=` before this, or are you planning on doing it after?  Either way, when all said and done, we should declare conformance to `Comparable` when we're done with the 6 methods. 😄

gabrieldemarmiesse:
Yep I was planning on doing it in another PR :)

stdlib/test/builtin/test_uint.mojo
```diff
@@ -12,7 +12,136 @@
...
+from testing import assert_equal, assert_true, assert_false
+
+# TODO: compute those with ** when IntLiteral supports __pow__()
+alias ABOVE_32_BITS = 4398046511104  # 2**42
+alias CLOSE_TO_UINTMAX = 18446744073709551605  # 2**64 - 10
+alias UINTMAX = 18446744073709551615  # 2**64 - 1
+
+
+def test_uint_lower_than():
+    assert_true(UInt(1) < UInt(2))
```
ConnorGray, line 24:
```suggestion
    assert_true(UInt.__lt__(UInt(1), UInt(2))
```

**Request** Here and below, could we change these tests to name the dunder method they're testing explicitly?

Writing the test using the dunder method explicitly means we can be very confident future regressions won't be masked by the presence of a "r" variant of these methods, or implicit conversion shennigans.

This is consistent with work @laszlokindrat has been doing very recently to change the existing `Int` tests to use this style.

ConnorGray:
This looks good, thank you Gabriel! I did leave one comment asking that we harmonize these new tests with a change in style that Laszlo has been pushing on recently (his changes should be visible in tonights nightly).

ConnorGray:
Thanks for making that update! Awesome work as always. I'll proceed with syncing.

ConnorGray:
!sync

ConnorGray:
!sync

ConnorGray:
@gabrieldemarmiesse I wonder if you'd be open to adding an item for this change in changelog.md, in the New section?

I've been thinking that it might make sense for contributors, especially folks who've been around a while, to have some additionally ownership over representing features they've added in the changelog.

This could both help with the maintainer burden of writing these changelog entries, and help folks making contributions promote what they're doing🙂

It might not be something we expect every contributor to do, just the ones who've got a certain number of merged PRs and experience in this project.

Curious to hear if this is something you'd be interested in.

gabrieldemarmiesse:
@ConnorGray Sure thing, sounds good to me. Can it be in a separate PR? 

I would tend to prefer updating the changelog once with all the new features I added during the month (for exemple,`UInt`, `UnsafeMaybeUninitialized`, `List.unsafe_get()`, `List.unsafe_set()`  rather than doing a back and forth and modify the changelog at each PR. Would that be all right by you? That's of course assuming that the releases happen at the end of each month.

ConnorGray:
It looks like these got added in the same patch Joe mentioned in #3079 was done internally in the last couple days. So the changelog ask is moot 🙃 

I'm sorry for the conflict in work here Gabriel; we'll try to do better in the future to make sure folks aren't accidentally duplicating work. It can be a bit difficult to coordinate, and its part of the growing pains we're going through.

PR #3019, opened 2024-06-12
[stdlib] Implement `__str__` and `__repr__` for `Set`

bgreni:
Implement `__str__` and `__repr__` using `format_to` pattern.

### DIFF
--- a/stdlib/test/collections/test_set.mojo
+++ b/stdlib/test/collections/test_set.mojo
@@ -15,11 +15,11 @@
 from collections import Set
 
 from testing import assert_false, assert_raises, assert_true
+from testing import assert_equal
 
 
-fn assert_equal[T: EqualityComparable](lhs: T, rhs: T) raises:
-    if not lhs == rhs:
-        raise Error("AssertionError: values not equal, can't stringify :(")
+fn assert_equal_set(lhs: Set[Int], rhs: Set[Int]) raises:
+    assert_equal(lhs.__str__(), rhs.__str__())
 
 
 def test_set_construction():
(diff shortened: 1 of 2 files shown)


stdlib/test/collections/test_set.mojo
```diff
@@ -484,6 +485,14 @@ def test_clear():
     assert_true(len(set5) == 0)
 
 
+def test_set_str():
+    var a = Set[Int](1, 2, 3)
+    AE(a.__str__(), "{1, 2, 3}")
+    AE(a.__repr__(), "{1, 2, 3}")
+    var b = Set[String]("a", "b")
+    AE(b.__str__(), "{'a', 'b'}")
```
gabrieldemarmiesse, line 493: Nice PR, thanks! Can you add a test case for an empty set? It seems like a common edge case.

gabrieldemarmiesse (commented):
Looks good! I think we just need to add another test case and we should be good to go :)

ConnorGray:
!sync

stdlib/test/collections/test_set.mojo
```diff
@@ -15,6 +15,7 @@
 from collections import Set
 
 from testing import assert_false, assert_raises, assert_true
+from testing import assert_equal as AE
```
ConnorGray, line 18: **Suggestion** I think it would be better for consistency and readability if we used this as `assert_equal` instead of using a custom alias. I understand the desire for brevity in a common function like this, but I think we should weight consistently higher.

bgreni: I actually did that since this file defines it's own `assert_equal` that drops the `Stringable` requirement which I don't personally like but I didn't want to go changing all the tests

ConnorGray: Ah I see, thanks for clarifying. That's not ideal, but I won't hold this PR up on that. Thanks for adding this!

ConnorGray:
!sync

stdlib/test/collections/test_set.mojo
bgreni: On second thought I've gone and just renamed the local function to make it more clear in the future.

ConnorGray:
!sync

PR #3022, opened 2024-06-12
[stdlib] benchmark sort scalar list

mzaks:
This PR adds a benchmark for sorting a list of scalars.

The benchmarks are performed on the public `sort` and internal `_insertion_sort` and `_small_sort` functions. 
The intention of the benchmark are following:
1. Evaluate results on tiny lists (size 2 to 5) and observe how much faster the `_small_sort` function is compared to `_insertion_sort`.
2. Evaluate result on tiny lists (size 2 to 5) and observe how much overhead does the `sort` function produce compared to internal functions
3. Evaluate results on small lists (size 10 to 100) and observe how much overhead does the `sort` function produce compared to calling the internal insertion sort function

My observations running the benchmarks on an Apple M1:
1. `_small_sort` is sometimes ~10% faster
2. `sort` function performs 5 - 6x slower than internal functions
3. `sort` function performs 3 - 100x slower than internal insertion sort

Suggestions:
- We should run the benchmark on other architectures to identify if my observations are consistent.
- Consider removing the `_small_sort` function if it does really provided almost no benefits compared to insertion sort.
- Expose insertion sort as top level function, as it has some nice characteristics which users might want to utilise directly.
- Consider simplifying the `_quicksort` function and maybe provide a parameter, which defines upper bound of the list size to be sorted with insertion sort.
- Consider adding similar parameter to `sort` function.


Dan13llljws:
I ran the benchmark on Apple M3 and I got see that `sort` performs 3-4x worse than direct calls to internal functions.

To reduce overhead, I added a function `_delegate_small_sort()` and in `sort` I have:
```mojo
    if len <= 5:
        _delegate_small_sort[Scalar[type], _less_than_equal](buff, len)
        return
    if len < 32:
        _insertion_sort[Scalar[type], _less_than_equal](buff, 0, len)
        return
    else:
        _quicksort[Scalar[type], _less_than_equal](buff, len)
```
This makes `sort` as fast as `_small_sort` for $N <= 5$. (I also see no benefit for using small sort instead of insertion sort).
This also makes `sort` as fast as `_insertion_sort` for $N < 32$, and when $N = 32$ I observed that `sort` starts to perform 9-25x worse than direct calls to `_insertion_sort`, which strongly indicates that the overhead in `_quicksort` is BAD.

Then I rewrote `_quicksort` to use direct recursion instead of a stack. This improve the runtime to be only 2-10x worse than direct calls.

Then I rewrote `_partition` as follows:
```mojo
    if start == end:
        return end

    var left = start
    var right = end - 2
    var pivot_value = array[end - 1]

    while True:
        while cmp_fn(array[left], pivot_value):
            left += 1
        while left < right and not cmp_fn(array[right], pivot_value):
            right -= 1
        if left >= right:
            swap(array[left], array[end - 1])
            return left
        swap(array[left], array[right])
        left += 1
        right -= 1
```
This makes `sort` as run about the same speed as `_insertion_sort`. When I extended your benchmark for list of size 1000, I actually see `sort` running in around 0.9x the time of `_insertion_sort` (But I was expecting much better results). Then instead picking the last element as the pivot. I decided to pick a random one:
```mojo 
    var pivot = int(random.random_si64(start, end - 1))
    var pivot_value = array[pivot]
```
which makes `sort` run in around 0.7x the time of `_insertion_sort`.

**Conclusion** 
- Maybe we should remove `_small_sort` network for now as it doesn't provide any value.
- Rewrite `_partition`.
- Either improve the overhead for stack (which is currently implemented using a `List`), or use direct recursion.

mzaks:
Would be great if someone could run the benchmark on Intel/AMD and other architectures.

Dan13llljws:
BTW, I just noticed that the `random_scalar_list` function returns only lists of 0 and 1s. Does this happen to anyone else?

stdlib/benchmarks/builtin/bench_sort_scalars.mojo
```diff
@@ -0,0 +1,196 @@
...
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+# RUN: %mojo
```
Dan13llljws, line 14:
```suggestion
# RUN: %mojo %s
```

stdlib/benchmarks/builtin/bench_sort_scalars.mojo
```diff
@@ -0,0 +1,196 @@
...
+
+fn random_scalar_list[
+    dt: DType
+](size: Int, max: Scalar[dt] = Scalar[dt].MAX) -> List[Scalar[dt]]:
+    var result = List[SIMD[dt, 1]](size)
+    for _ in range(size):
+
+        @parameter
+        if dt.is_integral() and dt.is_signed():
+            result.append(random_si64(0, Int64(max)).cast[dt]())
```
Dan13llljws, line 40:
```suggestion
            result.append(random_si64(0, max.cast[DType.int64]()).cast[dt]())
```

stdlib/benchmarks/builtin/bench_sort_scalars.mojo
```diff
@@ -0,0 +1,196 @@
...
+        elif dt.is_integral() and dt.is_unsigned():
+            result.append(random_ui64(0, UInt64(max)).cast[dt]())
+        else:
+            result.append(random_float64(0, Float64(max)).cast[dt]())
```
Dan13llljws, line 44:
```suggestion
            result.append(random_float64(0, max.cast[DType.float64]()).cast[dt]())
```

Dan13llljws:
It will also be great if you can make everything in one chunk of tests run on same set of data. Currently, you are generating a new set of list for every iteration.

mzaks:
> It will also be great if you can make everything in one chunk of tests run on same set of data. Currently, you are generating a new set of list for every iteration.

Thanks for this and the other suggestions, I implemented them in one go, please have a look.

stdlib/benchmarks/builtin/bench_sort_scalars.mojo
```diff
@@ -0,0 +1,238 @@
...
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+# RUN: %mojo %s
```
JoeLoser, line 14:
```suggestion
# RUN: %mojo %s -t
```

Do you mind adding the `-t`?  This option gets parsed by the benchmarking package internally to run it in "test mode" to speed up CI.  This will be clearer once we can open source the `benchmark` module in the future.

stdlib/benchmarks/builtin/bench_sort_scalars.mojo
```diff
@@ -0,0 +1,238 @@
...
+    DType.uint16,  # DType.int16, DType.float16,
+    DType.uint32,  # DType.int32, DType.float32,
+    DType.uint64,  # DType.int64, DType.float64,
+)
+
+
+fn random_scalar_list[
+    dt: DType
+](size: Int, max: Scalar[dt] = Scalar[dt].MAX) -> List[Scalar[dt]]:
+    var result = List[SIMD[dt, 1]](size)
```
JoeLoser, line 35:
```suggestion
    var result = List[Scalar[dt]](size)
```

JoeLoser (approved):

stdlib/benchmarks/builtin/bench_sort_scalars.mojo
```diff
@@ -0,0 +1,238 @@
...
+            )
+            m.bench_function[bench_insertion_sort](
+                BenchId(
+                    "bench_ins_sort_random_list_"
+                    + str(count)
+                    + "_type_"
+                    + str(dt)
+                )
+            )
+
```
Dan13llljws, line 159:
```suggestion
            _ = list^
```
We need to manually extend the lifetime of `list`. Otherwise this will not be captured correctly in the parameter closure `call_fn`.

stdlib/benchmarks/builtin/bench_sort_scalars.mojo
```diff
@@ -0,0 +1,238 @@
...
```
Dan13llljws, line 226:
```suggestion
            _ = list^
```
Same thing here.

Dan13llljws:
!sync

Dan13llljws:
Hey @mzaks do you mind fixing the merge conflict? We are ready to import this PR :)

Dan13llljws:
@mzaks Can you reformat `bench_sort_scalars.mojo` with `mojo format`?

mzaks:
> @mzaks Can you reformat `bench_sort_scalars.mojo` with `mojo format`?

@Dan13llljws all problems should be resolved now.

Dan13llljws:
!sync

Issue #2948, opened 2024-06-04
[stdlib] Use `normalize_index` throughout the standard library


ConnorGray:
#2677 added a new `normalized_index[..](..)` helper function to the standard library, used to perform bounds checking on indexes into collection types. This replaces boilerplate bounds checking and negative index normalization logic like:

```mojo
        debug_assert(-size <= int(index) < size, "Index must be within bounds.")
        var normalized_idx = int(index)
        if normalized_idx < 0:
            normalized_idx += size
```

with a function call:

```mojo
        var normalized_index = normalize_index["InlineArray"](index, self[])
```

That PR updated only `InlineArray` to use the new helper function.

We should update other collections in the standard library to use this function as well, including:

* [ ] `List.__setitem__`
* [ ] `List.insert`
* [ ] `List.pop`
* [ ] `List.__getitem__`
* [ ] `List.__get_ref`
* [ ] `InlinedFixedVector.__getitem__`
* [ ] `InlinedFixedVector.__setitem__`
* [ ] `InlineArray.__getitem__[index: ..]` — will require calling `normalize_index` at compile time (e.g. `alias normalize_idx = normalize_index["InlineArray"](i, size=size)`

**Please limit the number of call-sites updated to not more than 4-5 per PR, to make reviewing easier.**

@gabrieldemarmiesse @laszlokindrat FYI

gabrieldemarmiesse:
For anyone wanting to do this ticket, you can look at this diff for a good example of what to do: https://github.com/modularml/mojo/pull/2677/files (file stdlib/src/utils/static_tuple.mojo) and you can also use the `Indexer` trait which is the right trait to work with when using indexes. 

`Intable` or `Int` are not recommended when indexing something because 
1) Int is very restrictive, we might want to index with types like `UInt32` or `Int64`
2) Intable is too broad and allows floats to be used for indexing which is obviously not correct since we can't do `my_list[4.5]`

StandinKP:
I can take this up

laszlokindrat:
> For anyone wanting to do this ticket, you can look at this diff for a good example of what to do: https://github.com/modularml/mojo/pull/2677/files (file stdlib/src/utils/static_tuple.mojo) and you can also use the `Indexer` trait which is the right trait to work with when using indexes.
> 
> `Intable` or `Int` are not recommended when indexing something because

Actually, `Int` is implicitly convertible from `Indexer`, and it is currently the recommended way to define `__getitem__` and other methods that take indices.

[...]

PR #3024, opened 2024-06-12
[stdlib] `List` uses `normalize_index`

MatveyF:
Using `normalize_index` in `List` to reduce boilerplate bounds checking. 

Related to #2948

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -20,6 +20,7 @@ from collections import List
 """
 
 
+from collections._index_normalization import normalize_index
 from memory import UnsafePointer, Reference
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
@MatveyF Thank you for the pull request, that's a very nice cleanup! I think that we are hitting an abort at compile-time (yeah that's pretty bad, there must be a bug somewhere in our existing codebase, I think it's related to https://github.com/modularml/mojo/issues/2751).

Can you open one pull request per method so that we can pinpoint better where the bug is coming from? Sorry, I know it's annoying but that would help us a lot!

gabrieldemarmiesse:
Many thanks! I think you can close this one :)

Issue #2975, opened 2024-06-06
[Feature Request] polymorphic string functions, like `isdigit()`


dimitrilw:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Right now, `isdigit()` (found [here](https://github.com/modularml/mojo/blob/bf73717d79fbb79b4b2bf586b3a40072308b6184/stdlib/src/builtin/string.mojo#L236)) expects to receive a numeric value. I.e., the two noted lines fail:

<img width="560" alt="image" src="https://github.com/modularml/mojo/assets/5898931/b5d4d11d-8fe4-4cd6-8515-062786b8aea2">

But if I pass `ord(<value>)`, then it works:

<img width="269" alt="image" src="https://github.com/modularml/mojo/assets/5898931/517d176f-a2f0-44fa-8d0c-5cd28687f72c">

In Python, the `str.isdigit()` method is called from the string itself. More, it does not require advance alteration of the input.

<img width="155" alt="image" src="https://github.com/modularml/mojo/assets/5898931/ad0d0f33-a68f-455d-97f0-3d8467fb016e">

Perhaps Mojo's `isdigit()` function should be polymorphic to take different inputs. For example, in Python, it can assess a full string of characters to confirm all are digits:

<img width="170" alt="image" src="https://github.com/modularml/mojo/assets/5898931/dac68395-4675-4585-a9db-ebc7e5ee8355">

I'm thinking that the Mojo function should perform similarly. Not necessarily as a String.method(), but rather just something that can take different inputs, to include something like `isdigit("1x3")`, which would evaluate as `False`.

If Modular peeps agree, then I will gladly work to make this real and make a PR.

Note, the same could be said about other functions: `islower`, `isupper`, etc.

Thoughts?

### What is your motivation for this change?

Pythonic style to simplify onboarding Py-Peeps. Also, easier-to-read code; I believe `isdigit("x")` is more intuitive than `isdigit(ord("x"))`.

dimitrilw:
I have a fork that adds:
- `isupper(String)`
- `islower(String)`
- `String.isupper()`
- `String.islower()`

It follows the same pattern as bgreni's `isdigit` additions.

[...]

PR #3027, opened 2024-06-12
[stdlib] Add `String.isdigit`

bgreni:
Partially addresses #2975

- Added convenience `String` overload `isdigit` function
- Add `isdigit` member to `String` that checks if all characters in a string are digits as python has


JoeLoser (approved):
Great, thank you!

JoeLoser:
!sync

PR #3029, opened 2024-06-12
[stdlib] `List`'s `pop()` uses `normalize_index()`

MatveyF:
Using `normalize_index()` in `List`'s `pop()` to reduce boilerplate bounds checking.

Note:
Initially opened #3024 but due to a bug was asked to open separate PRs for each method.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -20,6 +20,7 @@ from collections import List
 """
 
 
+from collections._index_normalization import normalize_index
 from memory import UnsafePointer, Reference
 from sys.intrinsics import _type_is_eq
 from .optional import Optional
@@ -519,12 +520,7 @@ struct List[T: CollectionElement](CollectionElement, Sized, Boolable):
         Returns:
             The popped value.
         """
-        debug_assert(-len(self) <= i < len(self), "pop index out of range")
-
-        var normalized_idx = i
-        if i < 0:
-            normalized_idx += len(self)
-
+        var normalized_idx = normalize_index["List"](i, self)
         var ret_val = (self.data + normalized_idx).take_pointee()
         for j in range(normalized_idx + 1, self.size):
             (self.data + j).move_pointee_into(self.data + j - 1)


gabrieldemarmiesse (approved):

ConnorGray:
!sync

ConnorGray:
Hi @MatveyF, thank you for this contribution! I wanted to give a quick update on the reason for my delay in merging this PR.

Although this change is relatively straightforward, it changes the semantics of `List.pop()` in an important way: currently, `List.pop()` does not do bounds checking in release builds: if you `.pop()` an empty list, you get undefined behavior. Out of bounds reads like that are also a problem currently when indexing into a `List`, e.g. reading the 4th element from a length-3 list.

We want to make the `List` implementation safer by default, but we want to do that in a way that does not introduce bounds checking in the optimized-for-low-overhead behavior `List` currently has for some of our internal users. We think we have a plan to do that, but it will take a little bit of time for it to come together. (The basic idea is to parameterize `List` on its "safety level"; the exact details of that still need to be worked out.)

Until then, I'm going to keep this PR on the back-burner—its absolutely the direction we want to go, but we're not quite there yet.

gabrieldemarmiesse:
Just wanted to say that there has been previous work on this area, see:
* [`List.unsafe_get()`](https://github.com/modularml/mojo/blob/nightly/stdlib/src/collections/list.mojo#L771)
* [`List.unsafe_set()`](https://github.com/modularml/mojo/blob/nightly/stdlib/src/collections/list.mojo#L803)
* [`InlineArray.unsafe_get()`](https://github.com/modularml/mojo/pull/3013)

So if we wish to have something more fine-grained and explicit we could also have the option of `List.unsafe_pop()` or parametrize the `pop` with `fn pop[unsafe=False](idx: Int) -> T:`. Those are options to consider. 

We can also mix both parametrization. Something like:

```mojo
struct List[unsafe_list: Bool = False, T: CollectionElement]:
    fn pop[unsafe=Self.unsafe_list](idx: Int) -> T:
        ....
```
Which can allow the user to choose the safety level at both the struct level or the method level. I'm unsure if the compiler allows this currently though. But it could bring the best of both worlds.

MatveyF:
Hi @ConnorGray, thank you for your detailed response, I appreciate the explanation.

Is there anything that I can assist with now? I can try implementing `unsafe_pop()` or parametrizing current `pop()` method as suggested by @gabrieldemarmiesse. Or should I wait for the internal changes to be made?

JoeLoser:
> Hi @ConnorGray, thank you for your detailed response, I appreciate the explanation.
> 
> Is there anything that I can assist with now? I can try implementing `unsafe_pop()` or parametrizing current `pop()` method as suggested by @gabrieldemarmiesse. Or should I wait for the internal changes to be made?

Right now, I'd wait for internal changes to be made.  We've been toying around with the idea of parameterizing the unsafe behavior at the type level, e.g.

```
struct List[T: CollectionElement, Safe = True]:
   ...

using UnsafeList = List[T, Safe = False]
```

JoeLoser:
Thanks for the contribution! Our team is working on some broader changes to bounds checking which may also touch on negative indexing and avoiding runtime costs associated with this where desired.

So, for now, to help clean up the pull request review queue, instead of this hanging out for a while like it has, I'm going to close this pull request and we can see where things settle in the coming weeks to see if this is still an issue. Again, I appreciate your enthusiasm in helping make the standard library better! 🔥

PR #3030, opened 2024-06-12
[stdlib] `List`'s `__getitem__()` uses `normalize_index()`

MatveyF:
Using `normalize_index()` in `List`'s `__getitem__()` to reduce boilerplate bounds checking.

Note:
Initially opened #3024 but due to a bug was asked to open separate PRs for each method.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -20,6 +20,7 @@ from collections import List
 """
 
 
+from collections._index_normalization import normalize_index
 from memory import UnsafePointer, Reference
 from sys.intrinsics import _type_is_eq
 from .optional import Optional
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Mhhhh it seems to be the `__getitem__` which causes the issue. We have a similar problem here: https://github.com/modularml/mojo/pull/2945

JoeLoser:
Thanks for the contribution! Our team is working on some broader changes to bounds checking which may also touch on negative indexing and avoiding runtime costs associated with this where desired.

So, for now, to help clean up the pull request review queue, instead of this hanging out for a while like it has, I'm going to close this pull request and we can see where things settle in the coming weeks to see if this is still an issue. Again, I appreciate your enthusiasm in helping make the standard library better! 🔥

PR #3036, opened 2024-06-13
[stdlib] Add explicit copy constructor to `Optional`

gabrieldemarmiesse:
We can remove `__copyinit__` in a later PR.

This will be useful to use `Optional` in collections as they often do copies in methods.


ConnorGray:
!sync

ConnorGray:
!sync

PR #3037, opened 2024-06-13
[stdlib] Add explicit copy constructor to `Reference`

gabrieldemarmiesse:
This will be useful to use Reference in other collections as a copy can be done for some operations, like reallocating for example.


soraros:
Do we need this for a trivial reg-type like `Ref`?

gabrieldemarmiesse:
That's a good question. In the future, most collections will require elements to be explicitly-copyable, so if we want to use Reference in those collections, we need Reference to have the ExplicitlyCopyable trait.

soraros:
Or have implicit-copyable types derive `copy` automatically when we have advanced trait.

gabrieldemarmiesse:
I agree, this PR is a way to make progress right now. I don't think we'll get advanced traits anytime soon.

ConnorGray:
Yep, ideally this all gets less verbose in the future. But in the mean time we need this to be a separate initializers; this is all an incremental step, but not the final destination we want to be at.

ConnorGray:
!sync

PR #3039, opened 2024-06-13
[sdlib] Add `clamp` to math module

bgreni:
- Add `clamp` to the math module
- Replace private `_clip` function in `List` with `clamp`

### DIFF
--- a/stdlib/src/builtin/_math.mojo
+++ b/stdlib/src/builtin/_math.mojo
@@ -197,3 +197,66 @@ trait Truncable:
     # associated types.
     fn __trunc__(self) -> Self:
         ...
+
+
+# ===----------------------------------------------------------------------=== #
+# clamp
+# ===----------------------------------------------------------------------=== #
+
+
+fn clamp(value: Int, lower: Int, upper: Int) -> Int:
+    """Clamp value to the range of [lower, upper].
+
+    If value is less lower, or greater than upper, it becomes the respective
+    value, else value is returned.
+
+    If lower > upper, then the two are swapped.
+
+    Args:
+        value: The integer to clamp.
+        lower: The lower bound of the range.
+        upper: Then upper bound of the range.
+
+    Returns:
+        A value in the range [lower, upper].
+    """
+    var lo = lower
+    var hi = upper
+    if lo > hi:
+        swap(lo, hi)
+    return max(lo, min(value, hi))
+
+
+fn clamp[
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse (commented):
I'm not in favor or against, just adding info here. In the Python stdlib, it's been discussed to add such a function in [this mailing list](https://mail.python.org/archives/list/python-ideas@python.org/thread/EHZAXE4P2VOT3CE4H6SNDPDASW7H2CGS/#EHZAXE4P2VOT3CE4H6SNDPDASW7H2CGS) the TLDR of the mailing list is that there are many different implementations of `clamp` in the wild and they slightly differ on the way `NaN` are handled. In the end it was not added to the stdlib due to the lack of concensus, mostly about NaN.

Maybe we should add information about how this new function handles `NaN` in the docstring? Adding a test with `NaN` would be good too as we should not change the way we handle them as time goes by.

soraros:
We should also document how the `max > min` case is handled. There is another way out: *define* `clamp` as composition of `min` and `max` (both [`stablehlo`](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#clamp) and XLA took this approach), so it's not implementation detail anymore.

bgreni:
> Maybe we should add information about how this new function handles NaN in the docstring?

Since a `clamp` method already exists on `SIMD`, we'll have to just go with whatever behaviour it has?

> We should also document how the max > min case is handled. 

Out of curiosity I checked the `std::clamp` documentation and of course it's undefined behaviour...
Personally I wouldn't mind simply swapping `lower` and `upper` in that case?

laszlokindrat:
Thanks for the patch! Is there a direct motivation for this change? As pointed out, Python's `math` module doesn't have a `clamp` method, so we don't need this for compatibility. There is also a `SIMD.clamp` method already, so I'm not sure we should bring this in yet.

bgreni:
@laszlokindrat I noticed we had a private function for this in `List` and figured we might as well add it to the math module as a public function where it's properly tested and accounted for, and while it isn't included in Python, the C++ people might appreciate it's inclusion in this form as a free function like `std::clamp`.

laszlokindrat:
At this point, I would like to wait with this patch. The `_math` module is a temporary solution, and the utility of a `clamp` method is not super high without it being generic. Let's circle back to this once we clean up these math modules, maybe in a month or two?

PR #3053, opened 2024-06-15
[stdlib] Move `hash.mojo` off `unroll`

soraros:
As part of migrating off of `unroll` in favor of `@parameter for`, apply the use of `@parameter for` in `hash.mojo`.

### DIFF
--- a/stdlib/src/builtin/hash.mojo
+++ b/stdlib/src/builtin/hash.mojo
@@ -174,13 +174,12 @@ fn _hash_simd[type: DType, size: Int](data: SIMD[type, size]) -> Int:
     var final_data = bitcast[int_type, 1](hash_data[0]).cast[DType.uint64]()
 
     @parameter
-    fn hash_value[i: Int]():
+    for i in range(1, size):
         final_data = _ankerl_hash_update(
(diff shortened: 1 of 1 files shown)


stdlib/src/builtin/hash.mojo
```diff
@@ -174,13 +174,12 @@ fn _hash_simd[type: DType, size: Int](data: SIMD[type, size]) -> Int:
     var final_data = bitcast[int_type, 1](hash_data[0]).cast[DType.uint64]()
 
     @parameter
-    fn hash_value[i: Int]():
+    for i in range(1, size):
```
abduld, line 177: shouldn't this be 
```suggestion
    for i in range(size-1):
```

abduld (commented):
Thanks for the contribution.

Please add a PR description

stdlib/src/builtin/hash.mojo
soraros: I changed the use site from `i + 1` to `i`.

JoeLoser (approved):

JoeLoser:
!sync

PR #3057, opened 2024-06-15
[stdlib] fix `math.gcd` for negative numbers

soraros:
`gcd(0, -2)` should be `2`. I also simplified the implementation a bit, might be unnecessary though. CC @abduld

### DIFF
--- a/stdlib/src/math/math.mojo
+++ b/stdlib/src/math/math.mojo
@@ -1786,34 +1786,25 @@ fn gcd(m: Int, n: Int, /) -> Int:
     Returns:
         The greatest common divisor of the two integers.
     """
-    if m == 0 or n == 0:
-        return max(m, n)
+    var u = abs(m)
+    var v = abs(n)
+    if u == 0:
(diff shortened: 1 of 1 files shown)


abduld (approved):
lgtm. Thanks

soraros:
Rebased on the latest nightly. The tests are removed in c52b4d6, I don't know where to put a `gcd(-2, 0)` test.

JoeLoser:
@soraros I tried adding a test case internally for `gcd(2, -4)` to assert it is `2` in your branch.  That test case fails — do you mind taking a look into making that pass?

JoeLoser:
> @JoeLoser It does seems to work though: [Mojo playground](https://gist.github.com/modularbot/89aec9fa4ac6371b4d5d78d44bd6b1c5). It also works for me locally.

Oops, you're totally right. This was a mistake on my end; sorry for the noise.  I just added the relevant tests internally in the synced PR and this will land soon.  Thanks!

PR #3061, opened 2024-06-16
[stdlib] add `__str__` and `__repr__` for `Optional`

bgreni:
Add `format_to` `__str__` and `__repr__`

### DIFF
--- a/stdlib/src/collections/optional.mojo
+++ b/stdlib/src/collections/optional.mojo
@@ -164,6 +164,60 @@ struct Optional[T: CollectionElement](
         """
         return not self
 
+    fn __str__[
+        U: RepresentableCollectionElement, //
+    ](self: Optional[U]) -> String:
+        """Return the string representation of the value of the Optional.
+
+        Parameters:
+            U: The type of the elements in the list. Must implement the
+              traits `Representable` and `CollectionElement`.
+
+        Returns:
+            A string represenation of the Optional.
+        """
+        var output = String()
+        var writer = output._unsafe_to_formatter()
+        self.format_to(writer)
+        return output
+
+    # TODO: Include the Parameter type in the string as well.
+    fn __repr__[
+        U: RepresentableCollectionElement, //
+    ](self: Optional[U]) -> String:
+        """Returns the verbose string representation of the Optional.
+
+        Parameters:
+            U: The type of the elements in the list. Must implement the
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/optional.mojo
```diff
@@ -154,6 +154,56 @@ struct Optional[T: CollectionElement](CollectionElement, Boolable):
...
+
+        Parameters:
+            U: The type of the elements in the list. Must implement the
+              traits `Representable` and `CollectionElement`.
+
+        Args:
+            writer: The formatter to write to.
+        """
+        if self:
+            write_to(writer, repr(self.value()))
```
gabrieldemarmiesse, line 203: Here `repr(self.value())` can be a very big `String`. If we know that `self.value()` has a `format_to` method, would it be possible to use it to write directly to the `Formatter`?

You can imagine that `self.value()` is a List with 10 000 elements for example.

gabrieldemarmiesse (commented):
Nice addition. I think we can be more efficient here by skipping a temporary allocation of a possibly big string.

stdlib/src/collections/optional.mojo
bgreni: I like the idea, but wouldn't that require use to generally shift away from `RepresentableCollectionElement` in these scenarios to something equivalent to `Formattable & CollectionElement`?

ConnorGray:
!sync

ConnorGray:
It looks like I can't sync this because of the merge conflict. @bgreni could you rebase and then I'll merge this? Thanks! 🙂

stdlib/src/collections/optional.mojo
ConnorGray: I agree with Gabriel that we could make this more efficient, though we can do that in a follow-up PR if you'd like 🙂

bgreni: Might as well do it here. I'm on vacation at the moment but I can look into it on Monday!

bgreni:
> It looks like I can't sync this because of the merge conflict. @bgreni could you rebase and then I'll merge this? Thanks! 🙂

Done now, apologies for the delay!

ConnorGray:
> Done now, apologies for the delay!

No worries at all, thank you for fixing the conflict! I'll sync this in now.

ConnorGray:
!sync

PR #3064, opened 2024-06-17
[stdlib] Make `Slice` `Representable` and some other housekeeping.

bgreni:
- Make `Slice` `Representable`
- add `format_to` to `Slice`
- Make `Slice` string representation match Python
  - Including making `step` Optional for that purpose
- Remove `__getitem__`, which was done previously for `SliceNew` in #2963

### DIFF
--- a/stdlib/src/builtin/builtin_slice.mojo
+++ b/stdlib/src/builtin/builtin_slice.mojo
@@ -27,7 +27,7 @@ fn _compare_optional(x: OptionalReg[Int], y: OptionalReg[Int]) -> Bool:
 
 
 @register_passable("trivial")
-struct Slice(Stringable, EqualityComparable):
+struct Slice(Stringable, EqualityComparable, Representable, Formattable):
     """Represents a slice expression.
 
     Objects of this type are generated when slice syntax is used within square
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -46,7 +46,7 @@ struct Slice(Stringable, EqualityComparable):
     """The starting index of the slice."""
     var end: OptionalReg[Int]
     """The end index of the slice."""
-    var step: Int
```
ConnorGray, line 49: **Question** I wonder if it would be possible to separate out the change to use `Optional` for the `step` into a follow-up PR?

The performance characteristics of these slice types are somewhat sensitive, so it would be nice to de-risk the valuable formatting improvements here, so that we can evaluate the `step` type change on its own.

Things like that will also get easier as our benchmarking story improves 🙂

ConnorGray:
My apologies for the delay in reviewing this PR. This is a great quality of life improvement to have, but as mentioned in my comment in-line, I think it might be better if we split the change to `Slice.step` into a separate PR.

Thank you for this @bgreni 🙂

stdlib/src/builtin/builtin_slice.mojo
bgreni: I think that's fair, I can certainly do that

ConnorGray:
Thank you for those changes @bgreni! 😄 

I'll sync this in now 🔥

ConnorGray:
!sync

PR #3046, opened 2024-06-14
[docs] Fix typo in death.ipynb and traits.ipynb

codingonion:
Fix typo in death.ipynb and traits.ipynb

PR #3067, opened 2024-06-17
Fixed range index

ratulb:
Fixed typo in range index.


MatveyF:
I think this is a duplicate of https://github.com/modularml/mojo/pull/3046

arthurevans:
Thanks for this! As noted above, this was also fixed in #3046 which has just merged, so I'm going to close this PR.

PR #3071, opened 2024-06-17
[Stdlib] Speedup `Dict` (changing modulus to bitshifting)

rd4com:
Hello,

it could be a nice improvement, around +80% here (Ubuntu);

Hard to tell without feedbacks, here is the benchmark used:

```mojo
from time import now
from random import *
from sys.param_env import is_defined
from __dict_simd import Dict2

alias iteration_size = 128
def main():
    var result: Int=0
    var start = now()
    var stop = now()

    
    small = Dict2[Int,Int]()
    start = now()
    for x in range(100):
            small[i]=i
            result += small[i]
    stop = now()
    print(stop-start, result)

    result = 0

    small2 = Dict[Int,Int]()
    start = now()
    for x in range(100):
            small2[i]=i
            result += small2[i]
    stop = now()
    print(stop-start, result)
```
486225 812800
807652 812800

---

Because the dict always augment the reserved size by `<<=1`,

it is possible to use `&=(self.reserved-1)` instead of modulus

:partying_face: hope you got the same improvements as here

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -271,30 +271,30 @@ struct _DictIndex:
     fn get_index(self, reserved: Int, slot: Int) -> Int:
         if reserved <= 128:
             var data = self.data.bitcast[DType.int8]()
-            return int(Scalar.load(data, slot % reserved))
+            return int(Scalar.load(data, slot & (reserved - 1)))
         elif reserved <= 2**16 - 2:
             var data = self.data.bitcast[DType.int16]()
-            return int(Scalar.load(data, slot % reserved))
+            return int(Scalar.load(data, slot & (reserved - 1)))
         elif reserved <= 2**32 - 2:
             var data = self.data.bitcast[DType.int32]()
-            return int(Scalar.load(data, slot % reserved))
+            return int(Scalar.load(data, slot & (reserved - 1)))
         else:
             var data = self.data.bitcast[DType.int64]()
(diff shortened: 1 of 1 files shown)


bethebunny (approved):
Nice! The fact that this wasn't done automatically makes me wonder if the subtraction can also be optimized out. I'll follow up with some of the compiler folks to understand why this case isn't optimized or if we can somehow tell the compiler that `reserved` is guaranteed to be a multiple of 2. As far as I'm concerned, `%` is the correct way to spell this and reads much more clearly as to the intent, but obviously this optimization is the core intention behind always keeping a power-of-2 reserved size ;)

JoeLoser:
!sync

bethebunny:
Following up, the failure to optimize here is because we used a signed int for `reserved` rather than an unsigned type. That's also a reasonable followup.

bethebunny:
Consensus seems to be "relying on specific codegen optimizations for performance critical things is an antipattern", so +1 on merging this and we don't need any followups!

JoeLoser:
!sync

gabrieldemarmiesse:
Hi, a warning here, a PR made previously make the capacity of the dict being decided by the underlying `List`, see https://github.com/modularml/mojo/pull/2905. While `List` always grows by a power a two, that doesn't cause any issue. But if in the future, we change the capacity growing logic in `List`, the capacity of the Dict might not be a power of two anymore. 

Should we remove the optimization where the dict uses all the capacity of the list to make sure we always work with powers of two?

PR #3072, opened 2024-06-18
[stdlib] Add more `String` property check methods

bgreni:
Add `String` convenience overloads to `isupper` `islower` `_isspace` and `isprintable`

### DIFF
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -537,6 +537,22 @@ fn isdigit(c: String) -> Bool:
 # ===----------------------------------------------------------------------=== #
 
 
+@always_inline
+fn isupper(c: String) -> Bool:
+    """Determines whether the given character is an uppercase character.
+
+    This currently only respects the default "C" locale, i.e. returns True iff
+    the character specified is one of "ABCDEFGHIJKLMNOPQRSTUVWXYZ".
+
+    Args:
+        c: The character to check.
+
+    Returns:
+        True if the character is uppercase.
+    """
+    return isupper(ord(c))
+
+
 fn isupper(c: UInt8) -> Bool:
     """Determines whether the given character is an uppercase character.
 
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/string.mojo
```diff
@@ -527,6 +527,22 @@ fn isdigit(c: String) -> Bool:
 # ===----------------------------------------------------------------------=== #
 
 
+@always_inline
+fn isupper(c: String) -> Bool:
```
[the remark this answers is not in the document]
bgreni: hmm yeah good point, I suppose we should have both for all of these as we've done for `isdigit`

ConnorGray:
!sync

stdlib/src/builtin/string.mojo
```diff
@@ -563,6 +579,22 @@ fn _is_ascii_uppercase(c: UInt8) -> Bool:
 # ===----------------------------------------------------------------------=== #
 
 
+@always_inline
+fn islower(c: String) -> Bool:
+    """Determines whether the given character is an lowercase character.
+
+    This currently only respects the default "C" locale, i.e. returns True iff
```
gryznar, line 586: Two "f" in `if`

stdlib/src/builtin/string.mojo
```diff
@@ -537,6 +537,22 @@ fn isdigit(c: String) -> Bool:
 # ===----------------------------------------------------------------------=== #
 
 
+@always_inline
+fn isupper(c: String) -> Bool:
+    """Determines whether the given character is an uppercase character.
+
+    This currently only respects the default "C" locale, i.e. returns True iff
```
gryznar, line 544: typo, two "f"

stdlib/src/builtin/string.mojo
```diff
@@ -589,6 +621,23 @@ fn _is_ascii_lowercase(c: UInt8) -> Bool:
...
+
+    This only respects the default "C" locale, i.e. returns True only if the
+    character specified is one of " \\t\\n\\r\\f\\v". For semantics similar
+    to Python, use `String.isspace()`.
+
+    Args:
+        c: The character to check.
+
+    Returns:
+        True iff the character is one of the whitespace characters listed above.
```
gryznar, line 636: typo

stdlib/src/builtin/string.mojo
ConnorGray: It's a little esoteric, but "iff" is jargon that means "if and only if": https://www.merriam-webster.com/dictionary/iff

Perhaps we should spell this out fully, since understandably many folks may not know this term and it does look a lot like a typo.

(@arthurevans @scottamain FYI for style guide considerations.)

gryznar: Ok, I see. I left only this comment, to reduce noise

PR #3078, opened 2024-06-19
[docs] Correct "Python integration" hyperlink in manual

ScottBrenner:
Seem to be getting a 308 response code from the `/index` path hyperlinked here - https://docs.modular.com/mojo/manual/python/index - which inconsistently redirects me to https://docs.modular.com/mojo/manual/python, but sometimes does not and results in this Page Not Found error, is anyone else able to reproduce this?

image

Proposing changing the hyperlink to the redirected page directly.

### DIFF
--- a/docs/manual/index.md
+++ b/docs/manual/index.md
@@ -60,7 +60,7 @@ feedback](/mojo/community).
 
 - **Python**
 
-  - [Python integration](/mojo/manual/python)
+  - [Python integration](/mojo/manual/python/)
   - [Python types](/mojo/manual/python/types)
 
 - **Tools**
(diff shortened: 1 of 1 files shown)


docs/manual/index.md
```diff
@@ -60,7 +60,7 @@ feedback](/mojo/community).
 
 - **Python**
 
-  - [Python integration](/mojo/manual/python/index)
+  - [Python integration](/mojo/manual/python)
```
arthurevans, line 63:
```suggestion
  - [Python integration](/mojo/manual/python/)
```

arthurevans: We should retain the trailing slash here. Our site appears to accept both, but this is consistent with the link in the left nav.

arthurevans:
!sync

arthurevans (approved):
Thanks for the fix!

PR #3084, opened 2024-06-20
[stdlib] Add `Span.copy_from` method

lsh:
This PR adds a `copy_from` method to `Span`, which is meant to allow users to easily copy each element from one span to another.

### DIFF
--- a/stdlib/src/utils/span.mojo
+++ b/stdlib/src/utils/span.mojo
@@ -224,3 +224,20 @@ struct Span[
         """
 
         return self._data
+
+    @always_inline
+    fn copy_from[
+        lifetime: MutableLifetime, //
+    ](ref [_]self: Span[T, lifetime], ref [_]other: Span[T, _]):
+        """
+        Performs an element wise copy from all elements of `other` into all elements of `self`.
+
+        Parameters:
+            lifetime: The inferred mutable lifetime of the data within the Span.
+
+        Args:
+            other: The Span to copy all elements from.
+        """
+        debug_assert(len(self) == len(other), "Spans must be of equal length")
+        for i in range(len(self)):
+            self[i] = other[i]
--- a/stdlib/test/utils/test_span.mojo
+++ b/stdlib/test/utils/test_span.mojo
@@ -143,6 +143,17 @@ def test_span_slice():
     # assert_equal(res[1], 1)
 
 
+def test_copy_from():
+    var a = List[Int](0, 1, 2, 3)
+    var b = List[Int](4, 5, 6, 7, 8, 9, 10)
+    var s = Span(a)
+    var s2 = Span(b)
+    s.copy_from(s2[: len(a)])
+    for i in range(len(a)):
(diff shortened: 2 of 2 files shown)


stdlib/src/utils/span.mojo
```diff
@@ -224,3 +224,20 @@ struct Span[
...
+        """
+        Performs an element wise copy from all elements of `other` into all elements of `self`.
+
+        Parameters:
+            lifetime: The inferred mutable lifetime of the data within the Span.
+
+        Args:
+            other: The Span to copy all elements from.
+        """
+        debug_assert(len(self) == len(other), "Spans must be of equal length")
```
lsh, line 241: Should this `raise` instead of being a `debug_assert`?

stdlib/src/utils/span.mojo
```diff
@@ -224,3 +224,20 @@ struct Span[
...
+        for i in range(len(self)):
```
lsh, line 242: Should I be doing a `memcpy` instead of iterating in a loop?

ConnorGray: You're doing the right thing🙂 A `memcpy` would not be semantically valid for many types.

Conceptually and literally you're moving these types in memory, so calling `__moveinit__` here is the right thing to do.

For example, consider a type whose `__moveinit__` constructor contains custom logic when moving the type in memory, e.g. to fixup a self-referential pointer. Failing to call `T.__moveinit__` when moving it in memory would break the invariants of that type.

Though, all that being said, using `memcpy()` as an _optimization_ if and only if `T: <TriviallyCopyable>` would be a great thing to do, but we can not currently express that constraint properly in Mojo yet. (Mojo does not have an equivalent to the Rust [`Copy`](https://doc.rust-lang.org/stable/std/marker/trait.Copy.html) marker trait. `AnyTrivialRegType` is _similar_ but not quite usable at the moment for specialization in the same way.)

stdlib/src/utils/span.mojo
```diff
@@ -224,3 +224,20 @@ struct Span[
         """
 
         return self._data
+
+    @always_inline
+    fn copy_from[
+        lifetime: MutableLifetime, //
+    ](ref [_]self: Span[T, lifetime], ref [_]other: Span[T, _]):
```
ConnorGray, line 231:
```suggestion
    ](inout self: Span[T, lifetime], other: Span[T, _]):
```

**Question** Does this work if we leave out these `ref` specifications?

`inout self` is more accurate than `ref [_] self` since `self` _must_ be mutable for this method to work.

And likewise `borrowed other` (leaving the `borrowed` implicit) is valid too because this method never mutates `other`.

ConnorGray:
Thanks for contributing this, this is a great quality of life feature😄

I had a minor suggestion about style, but otherwise this LGTM.

ConnorGray:
!sync

ConnorGray (approved):

PR #3086, opened 2024-06-20
[stdlib] Add `OptionalReg.or_else()`

martinvuyk:
Add `OptionalReg.or_else()`


JoeLoser:
Thanks for the contribution!  This is a fine thing to do, but do note `OptionalReg` won't be around for much longer.  It's only used in a handful of places internally that want to move to using `Optional`.  So I'm not too interested in adding more of an API surface area to something that we want to get rid of.  In any case, if you'd like to see this move forward, please add some unit tests 😄

PR #3094, opened 2024-06-21
[stdlib] Use `ValueDestructorRecorder` in `test_inline_list.mojo`

gabrieldemarmiesse:
This PR is a small piece of #2965 . I'm trying to make #2965 smaller to make it easier to work on.

In itself, this PR is a refactoring. The goal is to unify the "destructor counters" that are present in the unit tests and use only the one in `stdlib/test/test_utils/types.mojo`. The goal of those structs is to record the number of `__del__` call being made to make sure the lifecycle of elements in a collection is correct. To make it usable with `InlineList`, I had to add the explicit copy constructor.

@ConnorGray  since you already did a review of #2965

### DIFF
--- a/stdlib/test/collections/test_inline_list.mojo
+++ b/stdlib/test/collections/test_inline_list.mojo
@@ -14,7 +14,7 @@
 
 from collections import InlineList, Set
 
-from test_utils import MoveCounter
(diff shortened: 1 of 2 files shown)


ConnorGray:
!sync

PR #3101, opened 2024-06-22
[Stdlib] Add `PythonObject.__contains__`

rd4com:
Simple and small implementation,

Improving the `🐍 Python` experience !

Example usage:

```mojo
x = PythonObject([1,2,3])
if 1 in x: 
   print("1 in x")
```

That method should be replaced by a `c-python` function if possible.


JoeLoser (commented):
Great!  Do you mind adding some unit tests please?

rd4com:
### 🆕 unit tests

:+1: yep it is important, should there be more ?


Python throws an error if the object is not iterable,

But mojo gets an `internal error: PyObject_GetAttrString failed` from it.

Current PR have a debug_assert when we have the error and returns False.

Is it possible to make `__contains__` re-raise the same error as python would have thrown ?

JoeLoser:
@rd4com sorry this one fell off my queue, I will look at this on Monday and your comment about the re-raising an error and see what we can do about that.

JoeLoser:
!sync

rd4com:
Yes, it works better, but it is not yet the same error as python:

```
Unhandled exception caught during execution: 'int' object has no attribute '__contains__'
```

python
```
TypeError: argument of type 'int' is not iterable
```

It could be because we are using `__getitem__`, see:

[Documentation: datamodel](https://docs.python.org/3/reference/datamodel.html#object.__contains__)

We might need to use something like `object.__contains__`,

Any idea ?

rd4com:
Nice :+1: , we partially got this with last commit, it behaves more like python:

in python
```python
class MyClass:
    x = 1
print(123 in MyClass())
```
> TypeError: argument of type 'MyClass' is not iterable

in python

```python
class MyClass:
    x = 1
    def __contains__(self, rhs):
        return True
print(123 in MyClass())
```
> True

in mojo
```mojo
from python import Python
def main():
    m = Python.import_module("myclass")
    x = m.MyClass()
    print(1 in x)
```
> True

if we remove the `__contains__` dunder in myclass.py

> Unhandled exception caught during execution: 'MyClass' object is not iterable


The difficulty is now to implement the last step of the "in" membership operator test.
(which is using `__getitem__`)

1. Any suggestion on how to create a test that uses the class created above in mojo?
(to test classes with import_module or not in `test_python_object.mojo`)
2. Should we return a PythonObject of class bool?

stdlib/src/python/_cpython.mojo
```diff
@@ -396,6 +396,16 @@ struct CPython:
         self._inc_total_rc()
         return r
 
+    fn PyObject_HasAttrString(
+        inout self,
+        obj: PyObjectPtr,
+        name: StringRef,
+    ) -> Int:
+        var r = self.lib.get_function[
+            fn (PyObjectPtr, DTypePointer[DType.uint8]) -> Int
```
JoeLoser, line 405: **Suggestion** Use `UnsafePointer` as we're phasing `DTypePointer` out.

stdlib/src/python/object.mojo
```diff
@@ -1102,6 +1102,28 @@ struct PythonObject(
         """
         return self._call_zero_arg_method("__invert__")
 
+    fn __contains__(self, rhs: PythonObject) raises -> Bool:
```
JoeLoser, line 1105: To clarify, in one of your comments, you're asking about the return type here, correct?  E.g. `Bool` vs `PythonObject`?  I'm fine with `Bool` as-is.  What do you think?

JoeLoser:
> Nice 👍 , we partially got this with last commit, it behaves more like python:
> 
> in python
> 
> ```python
> class MyClass:
>     x = 1
> ```
> 
> ```python
> print(123 in MyClass())
> ```
> 
> > TypeError: argument of type 'MyClass' is not iterable
> 
> in python
> 
> ```python
> class MyClass:
>     x = 1
>     def __contains__(self, rhs):
>         return True
> ```
> 
> ```python
> print(123 in MyClass())
> ```
> 
> > True
> 
> in mojo
> 
> ```mojo
> from python import Python
> def main():
>     m = Python.import_module("myclass")
>     x = m.MyClass()
>     print(1 in x)
> ```
> 
> > True
> 
> if we remove the `__contains__` dunder in myclass.py
> 
> > Unhandled exception caught during execution: 'MyClass' object is not iterable
> 
>  
> 
> The difficulty is now to implement the last step of the "in" membership operator test. (which is using `__getitem__`)
> 
> 1. Any suggestion on how to create a test that uses the class created above in mojo?
>    (to test classes with import_module or not in `test_python_object.mojo`)
> 2. Should we return a PythonObject of class bool?

For (1), are you asking about a user-defined type that implements `__iter__` but not `__contains__`?    Or are you asking about a type that defines neither `__iter__` or `__contains`, but _does_ define `__getitem__`?  This last one is the case of

> Lastly, the old-style iteration protocol is tried: if a class defines [__getitem__()](https://docs.python.org/3/reference/datamodel.html#object.__getitem__), x in y is True if and only if there is a non-negative integer index i such that x is y[i] or x == y[i], and no lower integer index raises the [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError) exception. (If any other exception is raised, it is as if [in](https://docs.python.org/3/reference/expressions.html#in) raised that exception).

from https://docs.python.org/3/reference/expressions.html#membership-test-details

stdlib/src/python/object.mojo
rd4com: Yes, the return type of the function :+1: 

It should be good, `Bool` can probably implicitly convert to `PythonObject` if user do chaining.

(but it is something to keep in mind for the future)

rd4com:
Yes, we need to test the whole function (combinations of instances that: `__getitem__` `__contains__`, `__iter__`).
For that, we need a way to create an instance of a class as `PythonObject`:
`Python.import_module` `test_classes_for_contains` for example.

(In a way that does not pollute the python environment for other tests of `PythonObject`)

stdlib/src/python/object.mojo
JoeLoser: Exactly, I had the composability question in mind, and I think we're fine with `Bool` for now.

JoeLoser:
> Yes, we need to test the whole function (combinations of instances that: `__getitem__` `__contains__`, `__iter__`). For that, we need a way to create an instance of a class as `PythonObject`: `Python.import_module` `test_classes_for_contains` for example.
> 
> (In a way that does not pollute the python environment for other tests of `PythonObject`)

One naive approach is just to have the different instances we want to test in their own test file with their own Python environment/CPython interpreter running.  It's far from great, but an option.

Could we leverage context managers and have a context managed CPython interpret environment for each of these to avoid test pollution?

rd4com:
That is a great idea, context managers are good for that :+1: 

The difficulty is that `Python()` returns `_get_global_python_itf().cpython()`:

```mojo
from python import Python
def main():
    A = Python()
    B = Python()
    A.eval("x = 123")

    print(A.evaluate("x")) 
    print(B.evaluate("x"))
```

The `PythonObject` returned by `import_module` is there:
```python
from python import  Python
def main():
    A = Python()
    B = Python()
    m = A.import_module("myclass")
    print(A.import_module("sys").modules.keys()) #myclass
    print(B.import_module("sys").modules.keys()) #myclass
```

So if we import for testing it would contaminate the environment,
maybe multiples test files is the good option,
what do you think ?

(we should test assert that the environment stays clean with multiple test files too)

rd4com:
Hello, two more tests to test the testing environment:
- `stdlib/test/python/test_testing_evironement_is_clean_A.mojo`
  1. sets a global variable
  2. import my_module.py
  3. assert that `my_module` is in `import_module("sys").modules.keys()`
- `stdlib/test/python/test_testing_evironement_is_clean_B.mojo`
  1. assert that the global variable raises (not exist)
  2. assert that `my_module` is not in `import_module("sys").modules.keys()`
  
For this to work, we need to be sure that the tests are done in that order. (A then B)

What do you think ?

JoeLoser:
> Hello, two more tests to test the testing environment:
> 
> * `stdlib/test/python/test_testing_evironement_is_clean_A.mojo`
>   
>   * sets a global variable
>   * import my_module.py
>   * assert that `my_module` is in `import_module("sys").modules.keys()`
> * `stdlib/test/python/test_testing_evironement_is_clean_B.mojo`
>   
>   * assert that the global variable raises (not exist)
>   * assert that `my_module` is not in `import_module("sys").modules.keys()`
> 
> For this to work, we need to be sure that the tests are done in that order. (A then B)
> 
> What do you think ?
> 
>  
> 
> In addition:
> 
> * The test for the dunder is moved into `test_python_object_dunder_contains.mojo`
> * It imports `module_for_test_python_object_dunder_contains.py`
>   
>   * Class_no_iterable_no_contains
>   * Class_no_iterable_but_contains
>   * Class_iterable_no_contains

I don't know if we have a way to force the testing order of A and then B — either in lit, or how we run the tests internally in our Bazel environment (CC: @keith here).  Perhaps just omit those tests for now but keep the others so we can rebase and land this?

rd4com:
Hello,
just implemented the changes :+1:  :
- ✅ reverted the `A` `B` 
- ✅ rebase

JoeLoser:
!sync

stdlib/src/python/python_object.mojo
```diff
@@ -1325,6 +1325,27 @@ struct PythonObject(
...
+
+        Calls the underlying object's `__contains__` method.
+
+        Args:
+            rhs: Right hand value.
+
+        Returns:
+            True if rhs is in self.
+        """
+        # TODO: replace/optimize with c-python function.
```
JoeLoser, line 1339: **Suggestion** Do you mind filing some follow-up GitHub issues in case you (or others) are interesting in tackling these.

rd4com: Yep, good idea !
will do it now and resolve it with the link here
(ping to @JoeLoser)

PR #3106, opened 2024-06-22
parallelize over rows, cols - not rows,rows

jesper-olsen:
Parallelize over rows, cols - not rows,rows. Rows & cols happen to be the same value in this example.


mikowals:
This PR is based on. a misunderstanding of what `parallelize[...](rows, rows)` does.  The original code is correct as it loops `rows` times using `rows` number of workers.  See [documentation of parallelize](https://docs.modular.com/mojo/stdlib/algorithm/functional/parallelize).

PR #3110, opened 2024-06-23
[stdlib] Add `fill` method to `Span`

lsh:
This PR tried to round out `Span` a little more by adding a `fill` method. Precedents are [std::fill](https://en.cppreference.com/w/cpp/algorithm/fill) and Rust's [slice fill](https://doc.rust-lang.org/std/primitive.slice.html#method.fill).

### DIFF
--- a/stdlib/src/utils/span.mojo
+++ b/stdlib/src/utils/span.mojo
@@ -241,3 +241,17 @@ struct Span[
         debug_assert(len(self) == len(other), "Spans must be of equal length")
         for i in range(len(self)):
             self[i] = other[i]
+
+    @always_inline
+    fn fill[lifetime: MutableLifetime](self: Span[T, lifetime], value: T):
+        """
+        Fill the memory that a span references with a given value.
+
+        Parameters:
+            lifetime: The inferred mutable lifetime of the data within the Span.
+
+        Args:
+            value: The value to assign to each element.
+        """
+        for element in self:
+            element[] = value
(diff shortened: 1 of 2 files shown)


stdlib/src/utils/span.mojo
```diff
@@ -241,3 +241,17 @@ struct Span[
         debug_assert(len(self) == len(other), "Spans must be of equal length")
         for i in range(len(self)):
             self[i] = other[i]
+
+    @always_inline
+    fn fill[lifetime: MutableLifetime](self: Span[T, lifetime], value: T):
```
ConnorGray, line 246: **Question** Does this method still compile if written as `inout self`, without the explicit lifetime parameter?

If so, I suggest we do that to keep the signature simple in cases where that's possible.

lsh: It doesn't work. I think the nuance here is that `inout` refers to immutable of the container, but not the data that it points to. So:
```
fn subslice[lifetime: ImmutableLifetime](inout self: Span[T, lifetime], value: T):
    self = self[0:3]
```
works just fine, since it makes no assumptions about the mutability of the data itself, just the value stored in the container type.

ConnorGray: Ah, that makes perfect sense, thanks for the explanation. This was just me misreading / misunderstanding the signature meaning—sorry for the noise.

ConnorGray:
!sync

PR #3114, opened 2024-06-24
[stdlib] Add comp time SIMD range constructor

### LINKED ISSUE #3103: [Feature Request] [stdlib] [proposal] Add comp time `SIMD` range constructor
### Review Mojo's priorities

### What is your request?
```mojo
@staticmethod
fn from_range[start: Int, end: Int, step: Int = 1]() -> Self:
    """Construct from a compile time range using the index as value.

    Parameters:
        start: The start of the range.
        end: The end of the range.
        step: The step of the range.

    Constraints:
        The range must be equal to or less than SIMD size.

    Examples:
    ```mojo
    print(SIMD[DType.uint8, 4].from_range[0, 4]()) # [0, 1, 2, 3]
    print(SIMD[DType.uint8, 4].from_range[3, -1, -1]()) # [3, 2, 1, 0]
    ```
    .
    """

    constrained[
        abs((start - end) // step) <= size,
        "The range must be equal to or less than SIMD size.",
    ]()
    var vec = Self()
    var idx = 0

    @parameter
    for i in range(start, end, step):
        vec[idx] = i
        idx += 1
    return vec
```
test:
```mojo
def test_range_constructor():
    alias INum = SIMD[DType.uint8, 4]
    assert_equal(INum.from_range[0, 4](), INum(0, 1, 2, 3))
    assert_equal(INum.from_range[3, -1, -1](), INum(3, 2, 1, 0))
    alias FNum = SIMD[DType.float16, 4]
    assert_equal(FNum.from_range[0, 4](), FNum(0, 1, 2, 3))
    assert_equal(FNum.from_range[3, -1, -1](), FNum(3, 2, 1, 0))
```

To be able to implement a `SIMD.index()` function, it would be needed to be able to contruct a range for the SIMD vec
for example if a generic version for `SIMD.index()` is wanted:
```mojo
fn index(self, value: Scalar[T]) -> Optional[Int]:
    alias indices = SIMD[DType.uint16].from_range[size -1, -1, -1]()
    var idx = ((self == value).cast[DType.uint16]() * indices).reduce_max()
    if idx == 0:
        return None
    return size - int(idx)
```

To be able to construct a reversed function for example shuffle could take in a compile time range
```mojo
fn reverse(inout self):
    self = self.shuffle[range(size -1, -1, -1)]()
```
or using the range constructor
```mojo
fn reverse(inout self):
    self = self.shuffle[Self.from_range[size -1, -1, -1]()]()
```

### Any other details?

_No response_

martinvuyk:
Add comp time SIMD range constructor. Closes #3103


stdlib/src/builtin/simd.mojo
```diff
@@ -469,6 +469,42 @@ struct SIMD[type: DType, size: Int = simdwidthof[type]()](
...
+        Examples:
+        ```mojo
+        print(SIMD[DType.uint8, 4].from_range[0, 4]()) # [0, 1, 2, 3]
+        print(SIMD[DType.uint8, 4].from_range[3, -1, -1]()) # [3, 2, 1, 0]
+        ```
+        .
+        """
+
+        constrained[
+            abs((start - end) // step) <= size,
```
soraros, line 496: Maybe:
```suggestion
            len(range(start, end, step)) == size,
```

soraros:
Actually, can we simply take a `_Range` after #2949?

martinvuyk:
> Actually, can we simply take a `_Range` after #2949?

That will be nice, though it still isn't merged so I guess this will have to wait until then

PR #3119, opened 2024-06-25
[stdlib] Add a `unsafe_assume_initialized` constructor to `InlineArray`

gabrieldemarmiesse:
This is a follow-up on https://github.com/modularml/mojo/pull/2965#discussion_r1648153449

The idea here is to guide the user in the unsafe lands of programming. Such that even when manipulating uninitialized memory, the assumptions about the data are clear.

The next step is to give a new constructor so that doing 
```
InlineArray[UnsafeMaybeUninitialized[String], 3](unsafe_uninitialized=True)
```
become a bit more user-friendly. We'd like to make it impossible to use `unsafe_uninitialized` without working with `UnsafeMaybeUninitialized`. This is for a next PR.

@ConnorGray for review as you were the one to follow the work in this area.


stdlib/src/utils/static_tuple.mojo
```diff
@@ -287,6 +287,39 @@ struct InlineArray[
...
+        It means that when copying, moving and destroying the new `InlineArray`, the copy constructor,
+        move constructor and destructor of the elements will be called.
+
+        If the elements of the input array are not initialized, the behavior is undefined.
+
+        There is only one situation when it is somewhat correct to use this constructor and the input array
+        do not have initialized elements. This is when the input array elements' type
+        has trivial copy constructor, move constructor and destructor (like integers or floats).
+        In this case, the values in the array will be random (since it's uninitialized memory)
+        and it's up to the user to handle them accordingly.
```
ConnorGray, line 310: I think this paragraph is not quite correct. The issue is subtle, but the basic problem is that _even if_ `ElementType` is valid for _every possible bit pattern_, its still undefined behavior to attempt a read of uninitialized memory, even if every possible value of that uninitialized memory would be a valid instance of the type.

This "poison" value for uninitialized data is why the [`std::mem::uninitialized()`](https://doc.rust-lang.org/stable/std/mem/fn.uninitialized.html#:~:text=The%20reason%20for%20deprecation%20is%20that%20the%20function%20basically%20cannot%20be%20used%20correctly) function was deprecated in favor of `MaybeUninit` in Rust: there is no way to use it correctly, even if the programmer only uses it for "trivial" types. The undefined-behavior-causing optimizations enabled for the compiler based on the assumption that uninitialized memory is never read are indifferent to whether that memory is a trivial type or not.

So I think its important we document clearly that callers of this function MUST take care that every element HAS been written to before doing this "assume its initialized" transmutation.

gabrieldemarmiesse: Okay I think I understand better now, it's true that I didn't take into account all kinds of optimizations the compiler could perform. Thanks for teaching me all this! I'll fix the PR.

gabrieldemarmiesse: @ConnorGray  it's been fixed, let me know if this docstring works for you :)

ConnorGray: No problem at all, thank you so much for pushing forward with such energy on this work!🙂  These are some solid foundations 🧱 you're building for folks wanting to write low-level abstractions in Mojo 🔥

ConnorGray:
The changes look good, thanks Gabriel for the edits. Syncing in now.

ConnorGray:
!sync

ConnorGray:
!sync

gabrieldemarmiesse:
With the recent revert of https://github.com/modularml/mojo/pull/2965, now this PR depends on https://github.com/modularml/mojo/pull/2965 which should be merged first.

JoeLoser:
https://github.com/modularml/mojo/pull/2965 just landed yesterday.  I had to parameterize the behavior of running destructors or not and default it to `False` to avoid some internal IR and PTX regressions.  That's fine though as we can explicitly specify `True` to run the destructors in your stack of PRs.

Feel free to ping after this is rebased and we can keep iterating on the SSO stack.

gabrieldemarmiesse:
@JoeLoser this PR is up to date with nightly again. It's ready for review.

JoeLoser:
!sync

JoeLoser:
!sync

PR #3120, opened 2024-06-25
[stdlib] Pythonic string isupper and islower functions and methods pr

dimitrilw:
Adds functions `isupper()` and `islower()` to handle single ASCII characters. Adds methods `String.isupper()` and `String.islower()` to match Pythonic behavior. Adds tests for functions and methods.

NOTE:
Does not yet handle UTF-8. ASCII only. Much like Mojo's case conversions currently only handle ASCII.

Also, added note to existing in-code TODO about need to update for UTF-8.

Tests pass:

image

Pythonic behavior, for ASCII only:

image

image


stdlib/test/builtin/test_string.mojo
```diff
@@ -833,28 +833,56 @@ fn test_splitlines() raises:
...
     assert_false(isupper(ord("!")))
+    assert_false(isupper("!"))
     assert_false(isupper(ord("0")))
+    assert_false(isupper("0"))
+
+    assert_true(String("ABC").isupper())
+    assert_false(String("abc").isupper())
+    assert_false(String("AbC").isupper())
+    assert_true(String("ABC123").isupper())
+    assert_false(String("1!").islower())
```
laszlokindrat, line 858:
```suggestion
    assert_false(String("1!").isupper())
```

laszlokindrat (commented):
Thanks for the patch! I think some overlapping functionality has been delivered in https://github.com/modularml/mojo/pull/3072. Could you please rebase?

dimitrilw:
@laszlokindrat , I have a question regarding the methods added in #3072 . 

In Python, `str.isupper()` and `str.islower()` resolve:
False if no alphabetic characters in string, else
True if all alphabetic characters meet the criteria.

In current Mojo implementation, they resolve:
True only if all characters in string meet the criteria.

image

Does Mojo seek Pythonic behavior? Or is this variance desired?

If Mojo seeks more Pythonic behavior, then the accepted code from #3072 need refactor & I will continue with this PR to fix the variance. If divergence from Python's behavior is desired, then this PR can be discarded.

bgreni:
> In Python, `str.isupper()` and `str.islower()` resolve: False if no alphabetic characters in string, else True if all alphabetic characters meet the criteria.
> 
> In current Mojo implementation, they resolve: True only if all characters in string meet the criteria.

The variance wasn't intentional. We should probably match the Python behaviour, although personally I'm not a fan of it.

laszlokindrat:
> In Python, `str.isupper()` and `str.islower()` resolve:
False if no alphabetic characters in string, else
True if all alphabetic characters meet the criteria.

@dimitrilw Thanks, good point! I was in this neighborhood for a different reason, so I will have a patch for this very soon. Please circle back once it lands, and feel free to open further patches if you find any issues. One particular area that needs love is Unicode support for these methods; we would highly appreciate your contributions!

dimitrilw:
I agree with you, @bgreni , re the behavior being "odd". The original Py logic is that they are more of a check of this:
> Does `some_str.isupper() == some_str.upper()`?

Perhaps your methods could be `is_all_upper()`?

And I can add Pythonic `isupper/lower` methods. I will make PR later today.

Regarding Unicode support: thar be dragons. In a former life, I did a LOT of unicode stuff related to international data exchanges and also the py 2.x -> py 3 conversions. It was always far messier than expected, especially if you ever explore point-code concepts, like what you see in Zalgo text. But, I'm a sucker for pain. I'll need to dust off the whole "what defines upper/lower in Unicode" concept. When it comes to Western languages, it is similar to ASCII (very fixed characters). When it comes to other languages, it can get less clear. I'll refresh on the comp-sci definition of Unicode upper/lower; I'm rusty. 😄

dimitrilw:
Oh! Re-reading your message, @laszlokindrat , I see that you will be making patch since you are already editing around there. Sounds good. After you patch, could you LMK? I'll rebase this PR and at least push the tests into place. I have a few more I should add, to ensure Py behavior.

Sorry for cluttering this thread; I'm multi-tasking right now, and like any multi-tasking moment, I'm misreading everything (both here, and in work stuff). 😂

laszlokindrat:
@dimitrilw I just landed my fixes, they should be published in the next nightly. I'm closing this now, but looking forward to any fixes/improvements you might have!

PR #3122, opened 2024-06-25
[stdlib] Use `InlineArray[UnsafeMaybeUninitialized[T], size]()`

gabrieldemarmiesse:
This requires https://github.com/modularml/mojo/pull/3119 to be merged first.

Note that I would like this method to be in the future `UnlineArray[UnsafeMaybeUninitialized[...], ...]()` in the future, but we currently have no way to verify that the generic type of the elements is `UnsafeMaybeUninitialized` so that will have to wait.

I'm not sure if this PR is fully correct. Is it okay to work with pointers on uninitialized data like this? Maybe the code was already wrong to begin with?

@ConnorGray for review as you are the one following the work in this area.


ConnorGray:
Mojo recently gained support for custom `self` types in `__init__` methods, so I wonder if we could write this now as:

```
struct InlineArray[..](..):

    fn __init__[T: CollectionElement](
        inout self: InlineArray[UnsafeMaybeUninitialized[T]]
    ):
        ...
```

thus enforcing the requirement that the element type is `UnsafeMaybeUninitialized[T]`, and avoiding the need for the static method.

We don't even need the method itself to be "unsafe", because its not unsafe to create uninitialized instances of `UnsafeMaybeUninitialized` (though, after the array is created, it is unsafe to interact with those elements).

gabrieldemarmiesse:
>  I wonder if we could write this now as...

It's possible (though the overload was sensitive to the order of declaration of the different `__init__` methods). I changed the PR to make use of this pattern. Feel free to take a look after https://github.com/modularml/mojo/pull/3119 is merged.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -264,43 +264,75 @@ struct InlineArray[
...
+            unsafe_assume_initialized[i].unsafe_ptr().move_pointee_into(
+                self.unsafe_ptr() + i
+            )
 
-        Note that this is highly unsafe and should be used with caution.
+    fn __init__[
+        ThisElementType: CollectionElementNew, this_size: Int
+    ](
+        inout self: InlineArray[
+            UnsafeMaybeUninitialized[ThisElementType], this_size
```
soraros, line 296: I think you can simply use `Self.size` here.

soraros:
Maybe the constructor from `Uninit[T]` could have been a conditional method on `Array[Uninit[T]]` types:

```mojo
struct Array[...]:
  fn unsafe_unwrap_uninitiated[T: ...](owned self: Array[Uninit[T], size]) -> Array[T, size]:
  ...
```

So we can use them like:
```mojo
var u = Array[Uninit[Int], 10]()
modify(u)
var a = u.unsafe_unwrap_uninitiated()
use(a)
```

gabrieldemarmiesse:
> Maybe the constructor from Uninit[T] could have been a conditional method on Array[Uninit[T]] types.

Indeed that's a possibility. I don't have a strong opinion on it as the two ways of doing it are similar.

soraros:
I also don't have a vary strong opinion, but using a method does have two advantages:

- Works better with method chaining
- This way, both constructing and unwrapping are only available on `Array[Uninit[T]]`. Ideally, they should be a type extension using conditional conformance. But you are right, we will need a new constructor anyways, making this point kinda moot. What about a constructor from `pop.array` instead?

Anyways, your PR your decision.

gabrieldemarmiesse:
Thanks for the suggestion. I don't see a strong reason to do additional work since the compiler prevents a misuse anyway as we don't have more compile-time checks with a method. If a maintainer ask me to change the code I'll do it though.

stdlib/src/utils/static_tuple.mojo
```diff
@@ -264,43 +264,74 @@ struct InlineArray[
...
-        constrained[
-            False,
-            (
-                "Initialize with either a variadic list of arguments, a default"
-                " fill element or pass the keyword argument"
-                " 'unsafe_uninitialized'."
-            ),
-        ]()
+
         self._array = __mlir_op.`kgen.undef`[_type = Self.type]()
```
soraros, line 285: Is this a temporary way to implement this? This function should just be a transmute (unsafe type cast).

gabrieldemarmiesse: I think in the future we may be able to avoid the copy, but at the moment, it's the only way possible as far as I know.

ConnorGray:
I've just merged #3119, so I think we're finally ready to move forward with this PR🙂  Thanks for your patience Gabriel as we've worked this through some internal hurdles, its really appreciated.

gabrieldemarmiesse:
@ConnorGray I merged nightly into this branch. It's ready for review. Since I stumbled upon a few compiler bug, I removed the `@value` and changed the order of the `__init__` methods. I'll do whatever is needed to please our overlord the Mojo compiler.

JoeLoser (commented):
!sync

JoeLoser:
!sync

gabrieldemarmiesse:

PR #3128, opened 2024-06-26
[Stdlib] Speedup `Dict` (reuse removed _DictIndex slots (not entries))

rd4com:
Hello, `🐍 python` went trough it's adventure of improving its dictionary,

there are many Python conference ▶️ videos that talks about that,

the speakers explains so well what is a dictionary and what is important, 

that i decided to learn and try to work a little bit on that.

&nbsp;

### ♻️ Problem solved:

if a key is popped, the slot `hash%self.reserved()` is marked as 🗑️ removed in `self._index`,
if it is reinserted, its hash will need to be 🔂 perturb until we find a slot marked with empty.

it means that inserting and popping the same key X times in a row makes its probe time x times longer.

current PR let us re-use ♾️ removed slots to create the shortest probe sequence possible.
(this does not affect the order of the dict, just its probe sequence)

It will still look until it find an empty slot, to make sure the key does not exist,
but if it encounted a removed slot on the way, it will reuse it.

That way, the slot will always be the nearest amount of perturb away.

&nbsp;

### ❄️ Another change is in `_maybe_resize()`,

We just append `self._reserved()` times `None` to the `List`, which makes it double.

Then we call `self._compact` which recalculate the indexes and compact it !

It accomplish the same as before but faster because:

`_compact` uses `_find_empty_index` instead of `_find_index` (which do an equality on the keys)

We can save a few performance on that check for equality, 

because the dict already said that every entry it contains deserves a unique existence;

&nbsp;

### 💡 Doing the compaction is a very costly thing

maybe we should just check if it is needed, once we doubled the size of the dict ?

&nbsp;

Remember, the first compaction check happens only if we don't need to resize (`not self._over_load_factor()`).

Once we are in the resize logic, we still don't know if we need to compact,

this could save us from moving many items in the list;

If no compaction is needed, we could just re-index each entries.

&nbsp;

*( If the user only popped one key in the middle of many entries,
should we move everything just for that ? )*

&nbsp;

I think that the `resize*2` logic need one new step:
 1. double the size by appending `None`
 2. check if we need to compact (calculated using the new `self._reserved()`)
   2.a yes, then call `self._compact` which do all the work for us
   2.b no, just assign empty slots to each live entries

(The current PR include that change ✅)

What do you think ?

# Benchmark 1, which is focus on the feature

94x speedup improvement on this particular benchmark 
(This benchmark is design for theses new features)

A:22730192154
B:239400760

```mojo
from time import now

def main():
    var result: Int=0
    var dict_size = 1<<12
    var start = now()
    var stop = now()
    
    result = 0

    small2 = Dict[Int,Int]()
    start = now()
    for x in range(200):
        for i in range(dict_size):
            small2[i]=i
            result+= small2[i]
            result+=small2.pop(0)
            small2[0] = 0
    stop = now()
    print(stop-start, result)
```

# Benchmark 2, which is focus on just resizing by 2

+= 1.6x (difficult to say without feedbacks)

A:1163417881
B:698271306

```mojo
from time import now

def main():
    var result: Int=0
    var dict_size = 1<<16
    var start = now()
    var stop = now()
    
    result = 0

    start = now()
    for x in range(50):
        small2 = Dict[Int,Int]()
        for i in range(dict_size):
            small2[i+x]=i+x
            result+= small2[i+x]
    stop = now()
    print(stop-start, result)
```

# Benchmark 3

from 1.3 to 2X

```mojo
from time import now

def main():
    var result: Int=0
    var dict_size = 1<<8
    var start = now()
    var stop = now()
    
    result = 0

    small2 = Dict[Int,Int]()
    start = now()
    for x in range(8):
        for i in range(dict_size):
            # grow the dictionary up to its new size
            small2[i]=i
        for y in range(100):
            for i in range(dict_size>>2):
                # pop one fourth of it 
                _=small2.pop(i)
            for i in range(dict_size>>2):
                # assign it back
                small2[i]=i
            for i in range(dict_size>>2):
                result+=small2[i]
        # double the size of the dictionary
        dict_size *= 2
    stop = now()
    print(stop-start, result)
```

# Benchmark 4, resizing for small dictionaries

1.39x speedup

A:745216
B:533981

```mojo
from time import now

def main():
    var result: Int=0
    var dict_size = 1<<8
    var start = now()
    var stop = now()
    
    result = 0

    start = now()
    for x in range(16):
        small2 = Dict[Int,Int]()
        for i in range(dict_size):
            small2[i]=i
            result += small2[i]
    stop = now()
    print(stop-start, result)
```

The dictionary start with a capacity of 8 entries,
it and doubles after some insertions until 512.

Each time it reaches 2/3 of its current capacity, it doubles.
(this check is done after each insertions)

(This is what `_maybe_resize` does)

---

It is difficult to say how better or not it is,

hopefully, somebody will be able to `⏱️ benchmark this well` and decide ?

&nbsp;

I'd like to finish this long PR description with a good job for our `Dict`,

it is probably the most favourite type of users, 

let's continue !

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -931,12 +931,16 @@ struct Dict[K: KeyElement, V: CollectionElement](
         # Return (found, slot, index)
         var slot = hash & (self._reserved() - 1)
         var perturb = bitcast[DType.uint64](Int64(hash))
+        var removed_slot = OptionalReg[Int](None)
         while True:
             var index = self._get_index(slot)
             if index == Self.EMPTY:
+                if removed_slot:
+                    return (False, removed_slot.value(), self._n_entries)
                 return (False, slot, self._n_entries)
             elif index == Self.REMOVED:
-                pass
+                if not removed_slot:
+                    removed_slot = slot
             else:
                 var entry = self._entries.__get_ref(index)
                 debug_assert(entry[].__bool__(), "entry in index must be full")
@@ -974,7 +978,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
             while not self._entries.__get_ref(right)[]:
                 right += 1
                 debug_assert(right < self._reserved(), "Invalid dict state")
-            var entry = self._entries.__get_ref(right)
+            var entry = Reference(self._entries[right])
             debug_assert(entry[].__bool__(), "Logic error")
             var slot = self._find_empty_index(entry[].value().hash)
             self._set_index(slot, left)
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -931,7 +931,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
         # Return (found, slot, index)
         var slot = hash & (self._reserved() - 1)
         var perturb = bitcast[DType.uint64](Int64(hash))
-        var removed_slot = OptionalReg[Int](None)
+        var removed_slot = Optional[Int](None)
         while True:
             var index = self._get_index(slot)
             if index == Self.EMPTY:


stdlib/src/collections/dict.mojo
```diff
@@ -911,15 +911,19 @@ struct Dict[K: KeyElement, V: CollectionElement](
                 return slot
             self._next_index_slot(slot, perturb)
 
-    fn _find_index(self, hash: Int, key: K) -> (Bool, Int, Int):
+    fn _find_index(inout self, hash: Int, key: K) -> (Bool, Int, Int):
```
bethebunny, line 914: This shouldn't need to change to `inout`

stdlib/src/collections/dict.mojo
```diff
@@ -911,15 +911,19 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
+        var removed_slot = OptionalReg[Int](None)
         while True:
             var index = self._get_index(slot)
             if index == Self.EMPTY:
+                if removed_slot:
+                    return (False, removed_slot.value(), self._n_entries)
                 return (False, slot, self._n_entries)
             elif index == Self.REMOVED:
+                if not removed_slot: removed_slot = slot
                 pass
```
bethebunny, line 927: Remove `pass` now that this block does something

stdlib/src/collections/dict.mojo
```diff
@@ -911,15 +911,19 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
         var perturb = bitcast[DType.uint64](Int64(hash))
+        var removed_slot = OptionalReg[Int](None)
         while True:
             var index = self._get_index(slot)
             if index == Self.EMPTY:
+                if removed_slot:
+                    return (False, removed_slot.value(), self._n_entries)
                 return (False, slot, self._n_entries)
             elif index == Self.REMOVED:
+                if not removed_slot: removed_slot = slot
```
bethebunny, line 926: style
```suggestion
                if not removed_slot:
                    removed_slot = slot
```

stdlib/src/collections/dict.mojo
```diff
@@ -958,7 +967,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
             while not self._entries.__get_ref(right)[]:
                 right += 1
                 debug_assert(right < self._reserved(), "Invalid dict state")
-            var entry = self._entries.__get_ref(right)
+            var entry = Reference(self._entries[right])
```
bethebunny, line 970: This change isn't related, and it's also going to make a copy the way `List.__getitem__` is currently implemented. I think `List` will be fixed soon :tm: but let's not touch this for now.

stdlib/src/collections/dict.mojo
```diff
@@ -939,17 +943,22 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
-        self._entries = self._new_entries(_reserved)
-        self._index = _DictIndex(self._reserved())
+        var _reserved = self._reserved()
 
-        for i in range(len(old_entries)):
-            var entry = old_entries.__get_ref(i)
-            if entry[]:
-                self._insert(entry[].unsafe_take())
+        for i in range(_reserved):
+            self._entries.append(None)
```
bethebunny, line 949: This feels off for a couple reasons.
- Why are we doubling the size when we're over the compaction factor, but not over the load factor? That's the wrong behavior.
- As far as I can tell, in the resize case, this will optimistically be the same performance, assuming that `List` always doubles in size which is exactly what we want to do here; however if that weren't the case this would actually do extra copies of `None` values. Is there any benefit I'm missing? If not this is just harder to understand to me vs. manually creating and copying the new list.

bethebunny (changes requested):
Thank you! We had this behavior at some point but it got lost xD Very glad to see it coming back, and it's awesome to see the speedups you're seeing :D

bethebunny:
Ahhh, something important I missed in my first pass.
By doing the resize _before_ checking the compact factor, you will _never actually compact_. So effectively some of the speedup you're seeing is a _performance_ speedup at the expense of using a bunch of extra memory: as the entries fill up with dead entries, rather than clean up, we just allocate more memory.

I think we should limit this PR scope to just re-using the removed slots.

However, 2 higher level thoughts to pull some of the other ideas forward:
- Should we consider resizing before compacting, as this PR currently does, given the performance speedup?
    - The answer probably depends somewhat on workloads
    - My intuition is no, I already feel weird that we never shrink the dict if it gets small enough
    - Definitely not obvious to me though
- If `_find_empty_index` is indeed substantially faster than `_insert`, we should refactor the resizing logic to use it

stdlib/src/collections/dict.mojo
ConnorGray: Just wanted to follow-up here: @bethebunny was totally right that up until _very recently_ this change would have done a copy, but since `List.__getitem__` was changed to return a `ref` just a couple days ago, this change to use `Reference(..)` is the right one 🙂 

So carry on @rd4com, your line is correct: `Reference(self._entries[right])`.

rd4com:
Thanks :raised_hands:  for taking the time to review this, hope you like the ideas;
It is both fun and useful to try to make the `Dict` faster for everybody.
I'll implement the changes you've requested, and remove the ones you suggested to remove.

Here is just a little note , 
to explain why there is no extra memory for the performance; 

&nbsp;

But it is not obvious to me neither, it is a big system to think about;
We're fighting against complexity here, but i think we're winning

Here is my understanding:
- compacting never reduces `self.size`
- `self._over_load_factor` only uses `self.size`
- doubling is only determined by `self._over_load_factor`

So we are not trade-offing memory for speed, which is important :+1: 

&nbsp;

Also it's important to notice that line 1 of `_maybe_resize()` did not change,
we still do the check for compaction when not overloaded.

if the user keep doing `Dict.pop`, then `insert`, it will compact :+1: .

&nbsp;

One of the downside of delaying compaction when overloaded is that `__iter__` is slower;
(more `None` to skip)
Also that more `None` values will be moved at once, in a single `_compact`.
(no sure if its always good, could be less smoothy ?)

&nbsp;

Compacting is directly influenced by how much a user `pop` values,
`pop` Nonify an entry in the `List`, we can't use that position again.
(it would turn the ordered dict into an unordered dict)

So compaction move the `None` value to the right, at the end of `self._n_entry`.
(or it move the values to the left of the `None`, same logic)

&nbsp;

🔹 Insertions increments `self._n_entries` and `self.size`
🔹 pop decrement `self.size`

`self._n_entries` is where we can always `_insert` without changing the order of the dict.
(it is always >= `self.size`)

&nbsp;

Compact move all the Nones between entries just after the last one, 
once it's done it change the value of `self._n_entries` to `self.size`.
It also recalculate the probe sequence for all entries, making the dict faster.
(removing extra perturb steps for passing over a _removed slot)

Doubling do exactly the same, this PR just don't move values there.
(we just create the new probe sequence filled with _empty, 
and get an empty slot for each non-none key-value pair in the `List`)

The reason for not moving values is that we received a bunch of `None` values!
theses are to the right of `self._n_entries` so we can use them to `_insert`;

Once we reach the point at which we cannot `_insert` anymore,
the `None` values are moved back to the right by `_compact`,
and we can continue to `_insert` without changing the dict order;

&nbsp;

Delaying moving `None` values somewhere else in the same `List` should not affect memory
(We have a `List` of `Optional`, which is a `Variant`, its size is its largest type)
(`print(sizeof[Optional[SIMD[DType.uint8,32]]]() == 40)`)

&nbsp;

Maybe it explains better than the first description lol, that was the reasoning behind the PR

&nbsp;

The second check for compaction, just after `append(None)` could be removed,
it is there just in case, but should never happen :+1: 

&nbsp;

For compacting after resize:

In Nightly ,the `Dict` currently compact on resize*2,
it is appending non-None entries to a new `List` with `_insert`,
in the context of a new `self._index=_DictIndex(self._reserved())`
(exactly what `_compact` does but in a different manner)

There must be a good reason for compacting on resize !

(the PR don't compact there, because new `None` can be used)

&nbsp;

💡 for the `Dict` that down-resize (`/=2`) suggestion

If the `Dict` needs to divide its size by two,
we could just call a new `_under_load_factor()` on `pop`,
and by comparing `self.size` < `1/3 of self._reserved()`,
we could just removes all the `None`, recalculate the _index, 
and finaly `pop` the top half of the `List`?
(maybe appending to a new one or `_realloc` to make sure its capacity is smaller)

stdlib/src/collections/dict.mojo
rd4com: Hello,

The first check for compact factor is done on line 1 of _maybe_resize;
it is the same as before (only done when were not overloaded, and returns).

If it does not return, we double the size,
the second check for compact factor, after the resize, can be removed.
it should never get done, and is there just in case.
(could be turned into a debug_assert)

Even a full list of size self._reserved(), doubled,
would not be at compact factor of self._reserved()*2 :

size 8, _n_entries 8, reserved 8 (impossible to reach, would been doubled !)

doubled reserved 16:
(8 * 4) > (16 * 3) 
(32 > 48) == False

This is why when we are overloaded, we are technically never overcompact,
because we have to double the size anyway.
So it does not compact on overload, taking into consideration the new reserved.

(Part of the performance come from not compacting on overload)

&nbsp;

For the list replacing, 
`_new_entries` first create a list full of none and return it,
then every existing entries are moved in to replace some.
(Self.reserved * 2 moves of none, +  self.size move of non-none)

here, `List._realloc` move half the new size with previous values, and we add none.
(Self.reserved * moves of none and non-none, +  Self.reserved move of none)

(without replacing values)

So both implementation archeive 2*Self.reserved, but PR don't replace values.
(`List.append` initialize self._reserved() items with none, without replacing)

Not sure if it is better or not, but not `List.__del__` too !

rd4com:
### 🆕 changes requested in the reviews were done 🥳!
🔥 The 94x speedup still hold (maybe a slighly less with changes)👍

The other ones (Benchmark 2 and 4) not, they depended on other changes.

But the decision to keep this PR to that one is correct 👍. 

It is better to keep them separated as they can be reverted in an easy manner.

Should we talk about the two higher level 💡 thoughts in another PR!?

Would you like to tackle the `_insert`/`_find_empty_index` first ?

I'd like to add more tests to to this PR too!

(ping to @bethebunny)

stdlib/src/collections/dict.mojo
```diff
@@ -931,12 +931,16 @@ struct Dict[K: KeyElement, V: CollectionElement](
         # Return (found, slot, index)
         var slot = hash & (self._reserved() - 1)
         var perturb = bitcast[DType.uint64](Int64(hash))
+        var removed_slot = OptionalReg[Int](None)
```
JoeLoser, line 934:
```suggestion
        var removed_slot = Optional[Int](None)
```

JoeLoser:
!sync

JoeLoser:
ping @bethebunny due to blocking request for changes

JoeLoser:
@rd4com do you mind rebasing so we can sync and merge this? Thanks!

PR #3133, opened 2024-06-28
[stdlib] Add `Dict._resize_down` and `_under_load_factor()`

rd4com:
Hello, 

this is a PR to solve a problem mentioned by @bethebunny 
(in #3128)

Good little feature to have,

it probably should not be a default and be parametrized,

because of the cost in performance of halving memory.

### 💡 for the Dictionary that not down-resize

Did an implementation of it, maybe you'll like it ! 

It really grows and ungrows dynamically,

at each `pop`, if `self.size < 1/3 of self._reserved()`,

the `Dict` halves it's memory with a smaller list capacity.

(so the RAM usage is dynamic and change on the size)

When pop is used on all elements of a dict of 256 elements,

the transition goes from 512, 256, 128, 64, 32, 16, 8, 4

That way, it doubles on 2/3 and halves on 1/3

&nbsp;

The ratio is quite simple and it seem to do the job:
(same as for resizing up (below 1/3 of `self._reserved()`))

0.33 * 256 == 84.48 <- we are below (we can resize down)
0.66 * 128 == 84.48 <- we are below (no need to resize up again)

0.66 * 256 == 168.96 <- we are above (we can resize up)
0.33 * 512 == 168.96 < we are above (no need to resize down again)

#### Example of a `Dict` cycling trough grow and ungrow:

```mojo
from time import now
def main():
    var result: Int=0
    var dict_size = 256
    var start = now()
    var stop = now()
    
    result = 0

    small2 = Dict[Int,Int]()
    start = now()
    for y in range(2):
        for i in range(dict_size):
            # grow the dictionary up to its full size
            print("before",small2._reserved(), len(small2))
            small2[i]=i
            print("after",small2._reserved(), len(small2))

        for i in range(dict_size):
            # ungrow the dictionary
            print("before",small2._reserved(), len(small2))
            result += small2.pop(i)
            print("after",small2._reserved(), len(small2))
    stop = now()
    print(stop-start, result)
```

The result is correct (`65280`):

```python
x = 0
for y in range(2):
    for i in range(256):
        x+=i
print(x) #65280
```

&nbsp;

Maybe that kind feature need to be parametrized, so users can choose?
(resizing down reduce the performance, so may not be a good default)

- `DictConfig` struct, as default `Dict` parameter is a possibility
  - we might end up in a parametrization challenge later
- another `Dict` type is also a possibility (`BaloonDictionary` like) ?

I suggest the second one would be better, but struggle for naming here.

&nbsp;

If reviewers likes it, maybe we can bring it to life with more reviews ?

(There are more methods where the logic need to be done, and tests)

&nbsp;

Note that the hard work have been done by people working on `Dict`,

this pr just add a halving part, which use the same ratio as the doubling.

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -1016,6 +1016,11 @@ struct Dict[K: KeyElement, V: CollectionElement](
                     return (True, slot, index)
             self._next_index_slot(slot, perturb)
 
+    fn _under_load_factor(self) -> Bool:
+        if self._reserved() == Self._initial_reservation:
+            return False
+        return 3 * self.size < self._reserved()
+
     fn _over_load_factor(self) -> Bool:
         return 3 * self.size > 2 * self._reserved()
 
@@ -1023,21 +1028,34 @@ struct Dict[K: KeyElement, V: CollectionElement](
         return 4 * self._n_entries > 3 * self._reserved()
 
     fn _maybe_resize(inout self):
-        if not self._over_load_factor():
-            if self._over_compact_factor():
-                self._compact()
-            return
-        var _reserved = self._reserved() * 2
-        self.size = 0
-        self._n_entries = 0
-        var old_entries = self._entries^
-        self._entries = self._new_entries(_reserved)
-        self._index = _DictIndex(self._reserved())
+        if self._over_load_factor():
+            self._resize(self._reserved() << 1)
+        elif self._under_load_factor():
+            self._resize(self._reserved() >> 1)
+        elif self._over_compact_factor():
+            self._compact()
 
-        for i in range(len(old_entries)):
-            var entry = old_entries[i]
-            if entry:
-                self._insert[safe_context=True](entry.unsafe_take())
+    fn _resize(inout self, _new_reserved: Int):
+        """Not for resizing, used internally to manage the capacity.
+
+        The `Dict` load balancer uses that method for balancing memory usage.
+        
+        """
+        debug_assert(_new_reserved%2==0, "_new_reserved is not a power of two")
+        var old_entries = self._entries^
+        self._entries = self._new_entries(_new_reserved)
+        self._index = _DictIndex(_new_reserved)
+        self.size  = 0
+        for i in range(self._n_entries):
+            var entry = Reference(old_entries[i])
+            if entry[]:
+                self._set_index(
+                    self._find_empty_index(entry[].value().hash),
+                    self.size
+                )
+                self._entries[self.size] = entry[].unsafe_take()
+                self.size +=1
+        self._n_entries = self.size
 
     fn _compact(inout self):
         self._index = _DictIndex(self._reserved())
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -846,7 +846,9 @@ struct Dict[K: KeyElement, V: CollectionElement](
             var entry_value = entry[].unsafe_take()
             entry[] = None
             self.size -= 1
-            return entry_value^.reap_value()
+            var tmp = entry_value^.reap_value()
+            self._maybe_resize()
+            return tmp^
         raise "KeyError"
 
     fn popitem(inout self) raises -> DictEntry[K, V]:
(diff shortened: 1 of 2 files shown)


stdlib/src/collections/dict.mojo
```diff
@@ -790,7 +790,16 @@ struct Dict[K: KeyElement, V: CollectionElement](
             var entry_value = entry[].unsafe_take()
             entry[] = None
             self.size -= 1
-            return entry_value.value^
+            var tmp = entry_value.value^ #TODO: not to merge with current PR
+            #It is necessary for def test_pop_default()
```
gryznar, line 794: Missing space after hash :)

bethebunny: Why not just return `entry_value.value^` directly?

stdlib/src/collections/dict.mojo
```diff
@@ -973,6 +986,24 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
+        var new_i = 0
+        for i in range(self._n_entries):
+            if _previous_list[i]: 
+                self._set_index(
+                    self._find_empty_index(_previous_list[i].value().hash),
+                    new_i
+                )
+                self._entries[new_i]=_previous_list[i].unsafe_take()
+                new_i += 1
+        self._n_entries = self.size
```
bethebunny, line 1005: Is there enough overlap with `_maybe_resize` to reuse?

rd4com: :+1:  Good idea, yes they are unified in new commits

stdlib/src/collections/dict.mojo
rd4com: :+1: Thanks @bethebunny , @gryznar, 

It is related to this issue: 
- https://github.com/modularml/mojo/pull/2796

The test `test_pop_default()` would fail,

I think it has to do with values that needs to be fully initialized to be ready for the `__del__` ,

but we move the value out anyway so maybe it has to do with something else ?

martinvuyk:
> Maybe that kind feature need to be parametrized, so users can choose?

Hi, yes It would be very interesting to be able to use std Dict in a memory constrained environment like a microcontroller. Maybe something like a `memory_constrained: Bool = False` parameter. But it can also be that we end up developing a specialized version for such environments as an extension to the stdlib :man_shrugging: (see proposal #3232)

JoeLoser (approved):
Sorry for leaving this one hanging, @rd4com.  The PR LGTM, do you mind rebasing so we can land this?

I'd hold off on the `DictConfig` idea or separate dict type for now.

rd4com:
Hello, all good :+1: , 
we also have an extra speedup by revisiting this! :partying_face: 

Added more test and one with the test_utils for `__del__` counts,
also fixed a small thing in `_find_slot` by using a `Pointer` (it was doing `__copyinit__` of `dict_entry`),
it results in a 2.27 speedup depending on the type for all methods (`insert`, `pop`, `__getitem__`, ..)
so it should now be even faster than before lol
(there was no `Pointer/Reference` when `Dict` got created so that's not a bug but integrating the feature)
(shout out to clattner who put us on the path to also focus on integration in last community meeting :+1: )

#### small benchmark (1.7x):
```mojo
from time import now
from random import *
from collections import Dict

alias iteration_size = 1024
def main():
    var result: Int=0
    var start = now()
    var stop = now()

    
    small = Dict[Int,Int]()
    start = now()
    for x in range(100):
            small[i]=i
            result += small[i]
    stop = now()
	_ = small
    print(stop-start, result)
```

#### 🔥 6.5x speedup
```mojo
from time import now
from random import *
from sys.param_env import is_defined
from collections import Dict

def main():
    var dict_size = 1<<12
    var result = 0

    var start = now()
    var stop = now()
    
    var x = Dict[String, String]()
     
    for i in range(dict_size):
        var val = str(i)
        x[val] = val

    start = now()
    for i in x.keys():
        result += len(x[i[]])
    stop = now() 
    
    print(result, stop-start)
```

#### 2x for `**kwargs`
```mojo
from time import now

fn sum_args(**kwargs: Int) raises ->Int:
    var res = 0
    for k in kwargs:
        res+=(kwargs[k[]])
    return res

def main():
    var iterations = 1<<10
    var result = 0

    var start = now()
    var stop = now()
    
        
    start = now()
    for i in range(iterations):
        result += sum_args(a=i, b=i+1, c=i+2)
    stop = now() 
    
    print(result, stop-start)
```

It is difficult to measure but `Pointer` should be faster anyway,
please do some benchmarks if needed :+1: 

There is also another idea to speedup `**kwargs` by using `Dict[UInt, V]`,
and just `rebind[UInt](int("arg1".unsafe_ptr()))` !
(but not sure if it is good to have a pointer as `K` or not)

NathanSWard:
Closing this PR as Dict was reworking to use a swiss-table implementation so its performance is considerably better.
Please reopen if you think there is still room for improvement with the new implementation, thanks!

PR #3135, opened 2024-06-28
[stdlib] Add `CollectionElementNew` to a few structs, part 1

gabrieldemarmiesse:
@ConnorGray who is following this work.

When switching all collection to `CollectionElementNew` that requires that many structs used throughout the stdlib implement this trait too.

This PR modifies `DType`, `Error`, `_NoneMarker`, `_ObjectImpl` and `StringLiteral`


ConnorGray:
!sync

PR #3138, opened 2024-06-28
[stdlib] Add the trait `CollectionElementNew` to `List`

gabrieldemarmiesse:
@ConnorGray for review

At first I wanted to make a second constructor but it was not possible because of this bug:
* https://github.com/modularml/mojo/issues/3137

so in the end, I replaced the existing one. Hopefully end users won't get too many surprises with that, as it can change the type of the result when doing things like `List(List[Int]())` since it uses then the constructor with variadic arguments.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -77,7 +77,9 @@ struct _ListIter[
             return self.index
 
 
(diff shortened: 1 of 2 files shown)


ConnorGray:
!sync

PR #3139, opened 2024-06-28
[stdlib] Add `CollectionElementNew` to a few structs, part 2

gabrieldemarmiesse:
@ConnorGray for the review

That should make the final transition of our collections to CollectionElementNew easier. I can see that the trait was needed for those structs. See the global PR here: https://github.com/modularml/mojo/pull/3134

This PR does not depend on part 1, it can be merged independently

This PR adds the explicit copy constructor to `Arc`, `UnsafePointer`, `DTypePointer` and `LegacyPointer`

### DIFF
--- a/stdlib/src/memory/arc.mojo
+++ b/stdlib/src/memory/arc.mojo
@@ -70,7 +70,7 @@ struct _ArcInner[T: Movable]:
 
 
 @register_passable
-struct Arc[T: Movable](CollectionElement):
+struct Arc[T: Movable](CollectionElement, CollectionElementNew):
     """Atomic reference-counted pointer.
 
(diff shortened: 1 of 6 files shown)


stdlib/src/memory/arc.mojo
```diff
@@ -101,6 +101,15 @@ struct Arc[T: Movable](CollectionElement):
             value^
         )
 
+    fn __init__(inout self, *, other: Self):
+        """Copy the object.
```
ConnorGray, line 105: **Suggestion** For `Arc`, what "the object" refers to could be somewhat ambiguous.

We should expand on this description to clarify that this is a cheap reference count increment copy, not a "deep" copy of the inner object.

ConnorGray:
!sync

PR #3140, opened 2024-06-28
[stdlib] Add `CollectionElementNew` to a few structs, part 3

gabrieldemarmiesse:
@ConnorGray for review

As I worked on removing `CollectionElement`, I noticed that those structs needed the new trait in the migration.

Those are pretty trivial and the new constructor will soon be used left and right so there is no test, but if you believe that's needed, just ask.

This PR does not depend on part 3, it can be merged independently

This PR adds the explicit copy constructor to `StringRef`, `_FixedString`, `InlineString`, `DLHandle`, `PythonObject` and `Path`

### DIFF
--- a/stdlib/src/python/object.mojo
+++ b/stdlib/src/python/object.mojo
@@ -119,6 +119,14 @@ struct PythonObject(
         """Initialize the object with a `None` value."""
         self.__init__(None)
 
+    fn __init__(inout self, *, other: Self):
+        """Copy the object.
+
+        Args:
+            other: The value to copy.
+        """
+        self = other
+
     fn __init__(inout self, ptr: PyObjectPtr):
         """Initialize the object with a `PyObjectPtr` value.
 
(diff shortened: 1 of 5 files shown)


ConnorGray:
!sync

stdlib/src/python/object.mojo
```diff
@@ -119,6 +119,14 @@ struct PythonObject(
         """Initialize the object with a `None` value."""
         self.__init__(None)
 
+    fn __init__(inout self, *, other: Self):
```
ConnorGray, line 122: **Nit** I think you accidentally omitted the explicit conformance to `CollectionElementNew` for `PythonObject`—you added it for the other types.

Not a major problem, I'll sync this in anyway. But wouldn't hurt if you wanted to drop that in as an add-on aside to any future PRs you do 🙂

gabrieldemarmiesse: Yeah I tried, but I hit a compiler bug when doing so. It's hard to do a minimal repro. It's really weird

ConnorGray: Huh, that does sound odd. Even if you can't get a minimal repro, if you can trigger it again easily or see it again, can you put it in a comment here? Then I'll at least know if I run into it that it's the same thing you saw.

PR #3143, opened 2024-06-28
[stdlib] Add CollectionElementNew to a few structs, part 4

gabrieldemarmiesse:
@ConnorGray for review

Adding the explicit copy constructor is needed for when collections will require all their elements to have an explicit copy constructor.

This PR contains the modification for `_ImmutableString`, `_ObjectImpl`, `_FormatCurlyEntry`, `OwnedKwargsDict` and `_NoneType`


ConnorGray:
!sync

PR #3153, opened 2024-06-30
[stdlib] Make trait `CollectionElement` inherit from `ExplicitlyCopyable`

gabrieldemarmiesse:
Waiting for the new nightly to be released so I can rebase this PR and shrink the diff

List of issues I have currently:
### 1 - Passing functions by parameter gives cryptic error message
Here is a minimal reproducer:
```mojo
fn foo(inout buff: UnsafePointer[Int], len: Int):

    @parameter
    fn _less_than_equal(lhs: Int, rhs: Int) -> Bool:
        return lhs <= rhs

    bar[type=Int, cmp_fn=_less_than_equal](buff, len)

@always_inline
fn bar[
    type: CollectionElement, cmp_fn: fn (type, type) capturing -> Bool
](array: UnsafePointer[type], size: Int):
    ...

```
The error message is 
```
invalid call to 'bar': callee parameter 'cmp_fn' has 'fn(Int, Int) capturing -> Bool' type, but value has type 'fn(lhs: Int, rhs: Int) capturing -> Bool'
```
The error message is not clear at all. What can we do to fix it? This issue prevents me from migrating over `sort.mojo`.

### 2 - The `None` constructor of `Optional` cannot be used anymore. We must use `Optional[...]()` or `Optional[...](_NoneType)`

It's still present in the code, but if one tries to use it, then the compiler says "cannot bind MLIR type None to trait CollectionElement". And the error message is often at the wrong place in the code (compiler bug?)

### 3 - Using `Dict[Self, Self]` in `PythonObject` gives `recursive reference to declaration`

So I had to comment out the method. I don't see how to reintroduce it.


stdlib/src/builtin/format_int.mojo
```diff
@@ -330,7 +332,7 @@ fn _try_write_int[
             len=1,
         )
         fmt.write_str(zero)
-        return
```
gabrieldemarmiesse, line 333: Here is an example of the None constructor of Optional not working anymore.

stdlib/src/python/object.mojo
```diff
@@ -295,18 +302,19 @@ struct PythonObject(
...
+    #
+    #    Args:
+    #        value: The dictionary value.
+    #    """
+    #    var cpython = _get_global_python_itf().cpython()
+    #    self.py_object = cpython.PyDict_New()
+    #    for entry in value.items():
+    #        var result = cpython.PyDict_SetItem(
+    #            self.py_object, entry[].key.py_object, entry[].value.py_object
+    #        )
```
gabrieldemarmiesse, line 317: Uncommenting this gives "recursive reference to declaration"

gabrieldemarmiesse:
@ConnorGray would that help if I rebase this PR or do you have an internal branch for this that you're taking care or? (in which case I can just close this PR then)

stdlib/test/collections/test_optional.mojo
```diff
@@ -151,10 +151,12 @@ def test_optional_equality():
     assert_true(o == 10)
     assert_true(o != 11)
     assert_true(o != n)
-    assert_true(o != None)
+    # TODO: Find a way to allow "o != None" again.
+    assert_true(o.__ne__(None))
```
gabrieldemarmiesse, line 155: in the meantime we can recommend `is None` which works well.

gabrieldemarmiesse:
@ConnorGray you can take a look at what this PR to get an idea of what it takes to make this work. But I'll likely split it further.

stdlib/test/collections/test_optional.mojo
ConnorGray: I think this will be fixed when we change the compiler to emit `NoneType()` when it sees `None`. (Currently it emits the `#kgen.none` raw MLIR type, which means there's always one extra "hop" of conversion where `NoneType` is currently used.)

stdlib/src/collections/optional.mojo
```diff
@@ -193,6 +204,17 @@ struct Optional[T: CollectionElement](
             return False
         return not rhs
 
+    fn __ne__(self, rhs: NoneType) -> Bool:
```
ConnorGray, line 207: **Suggestion** I think `opt == None` will work if you change this signature to:

```mojo
    # TODO(MSTDL-715):
    #   This initializer should not be necessary, we should need
    #   only the initilaizer from a `NoneType`.
    fn __ne__(self, rhs: NoneType._mlir_type):
```

(See the similar comment and pattern on `Optional.__init__(inout self, value: NoneType._mlir_type)`.)

stdlib/test/test_utils/types.mojo
```diff
@@ -108,14 +108,20 @@ struct MoveCounter[T: CollectionElementNew](
         self.move_count = existing.move_count
 
 
+# TODO: Pass directly a UnsafePointer when possible, instead of an Int.
+# Otherwise we get "argument #2 cannot be converted from 'UnsafePointer[List[Int], 0, 0]' to 'UnsafePointer[List[Int], 0, 0]'"
+# This bug also appears if we pass the list as borrow and grab the address in the constructor, in
+# which case we get "argument #2 cannot be converted from 'List[Int]' to 'List[Int]'"
```
ConnorGray, line 114: **Question** Does this work if you use `rebind[UnsafePointer[List[Int]]](ptr)` wherever the pointer is passed? That might be a simpler workaround (if it does work).

gabrieldemarmiesse: I'm not sure I understand where the rebind should go, but I tried this:
```mojo
@value
struct ValueDestructorRecorder(CollectionElement):
    var value: Int
    var destructor_counter: UnsafePointer[List[Int]]

    fn __init__(inout self, value: Int, destructor_counter: UnsafePointer[List[Int]]):
        self.value = value
        self.destructor_counter = rebind[UnsafePointer[List[Int]]](destructor_counter)

    fn __init__(inout self, *, other: Self):
        self.value = other.value
        self.destructor_counter = other.destructor_counter

    fn __del__(owned self):
        self.destructor_counter[].append(self.value)
```
And I get the following error:

```
/projects/open_source/mojo/stdlib/test/memory/test_maybe_uninitialized.mojo:26:32: error: no matching function in initialization
        ValueDestructorRecorder(
        ~~~~~~~~~~~~~~~~~~~~~~~^
/projects/open_source/mojo/stdlib/test/test_utils/types.mojo:120:8: note: candidate not viable: argument #2 cannot be converted from 'UnsafePointer[List[Int], 0, 0]' to 'UnsafePointer[List[Int], 0, 0]'
    fn __init__(inout self, value: Int, destructor_counter: UnsafePointer[List[Int]]):
       ^
/projects/open_source/mojo/stdlib/test/test_utils/types.mojo:124:8: note: candidate not viable: missing 1 required keyword-only argument: 'other'
    fn __init__(inout self, *, other: Self):
       ^
```

stdlib/src/collections/optional.mojo
gabrieldemarmiesse: I just tried and got the following error:

```
/projects/open_source/mojo/stdlib/test/collections/test_optional.mojo:148:5: error: cannot bind MLIR type 'None' to trait 'CollectionElement'
def test_optional_equality():
    ^
/projects/open_source/mojo/stdlib/src/builtin/value.mojo:148:8: note: MLIR type cannot satisfy required trait function here
    fn __init__(inout self, *, other: Self):
       ^
/projects/open_source/mojo/stdlib/test/collections/test_optional.mojo:148:5: error: cannot bind MLIR type 'None' to trait 'CollectionElement'
def test_optional_equality():
    ^
/projects/open_source/mojo/stdlib/src/builtin/value.mojo:148:8: note: MLIR type cannot satisfy required trait function here
    fn __init__(inout self, *, other: Self):
       ^
```
Since I get the same error when I didn't do any modifications to `Optional` I believe that it's a bug in the overload resolution.

ConnorGray:
!sync

stdlib/src/python/object.mojo
```diff
@@ -313,18 +313,18 @@ struct PythonObject(
...
+    #    """Initialize the object from a dictionary of PythonObjects.
+    #
+    #    Args:
+    #        value: The dictionary value.
+    #    """
+    #    var cpython = _get_global_python_itf().cpython()
+    #    self.py_object = cpython.PyDict_New()
+    #    for entry in value.items():
+    #        var result = cpython.PyDict_SetItem(
+    #            self.py_object, entry[].key.py_object, entry[].value.py_object
```
gryznar, line 326: Should this code be commented without todo / fixme etc?

stdlib/src/python/python.mojo
```diff
@@ -206,14 +206,14 @@ struct Python:
...
-        """
-        return PythonObject(Dict[PythonObject, PythonObject]())
+    # @staticmethod
+    # fn dict() -> PythonObject:
+    #    """Construct an empty Python dictionary.
+    #
+    #    Returns:
+    #        The constructed empty Python dictionary.
+    #    """
+    #    return PythonObject(Dict[PythonObject, PythonObject]())
```
gryznar, line 216: Same question, is commenting that intentional?

ConnorGray:
!sync

gabrieldemarmiesse:
@ConnorGray I found a workaround for the `Dict[PythonOject, PythonObject]`. Let me know what you think :)

JoeLoser:
!sync

gabrieldemarmiesse:
Feel free to ping me when you have time for a review and want me to rebase the PR

gabrieldemarmiesse:

PR #3154, opened 2024-06-30
[stdlib] Disallow implicit float to int conversions in SIMD types

msaelices:
Disallow this kind of implicit conversion:

```mojo
var x: Int64 = 1.679  # x will be 1, so we lose 0.679
var y: Int64 = Float64(2.123)  # y will be 2
```

Related to issue https://github.com/modularml/mojo/issues/3149

Note for the reviewer: I am not sure if you want to generalize the `_simd_construction_checks()` functions by adding the `value` types as well and be able to make checks on them, comparing `type` with `value_type`.


stdlib/test/builtin/test_file.mojo
```diff
@@ -64,7 +64,7 @@ def test_file_read_bytes_multi():
         assert_equal(string2, "dolor ")
 
         # Read where N is greater than the number of bytes in the file.
-        var s: String = f.read(1e9)
```
laszlokindrat, line 67: Why is this change needed?

laszlokindrat: oh, because this is a FloatLiteral, nvm

laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

PR #3159, opened 2024-07-01
[stdlib] Make `object` `Representable` and `Formattable`

bgreni:
Make `object` `Representable` and `Formattable`.

### DIFF
--- a/stdlib/src/builtin/object.mojo
+++ b/stdlib/src/builtin/object.mojo
@@ -250,7 +250,13 @@ struct _Function(CollectionElement, CollectionElementNew):
         )
 
 
-struct _ObjectImpl(CollectionElement, CollectionElementNew, Stringable):
+struct _ObjectImpl(
+    CollectionElement,
+    CollectionElementNew,
+    Stringable,
+    Representable,
+    Formattable,
+):
     """This class is the underlying implementation of the value of an `object`.
     It is a variant of primitive types and pointers to implementations of more
     complex types.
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/object.mojo
```diff
@@ -557,18 +563,22 @@ struct _ObjectImpl(CollectionElement, CollectionElementNew, Stringable):
         else:
             lhs = lhs.convert_bool_to_int()
 
-    fn __str__(self) -> String:
+    fn format_to(self, inout writer: Formatter):
         """Returns the name (in lowercase) of the specific object type."""
```
gryznar, line 567: This docstring is no longer valid

stdlib/src/builtin/object.mojo
```diff
@@ -577,32 +587,47 @@ struct _ObjectImpl(CollectionElement, CollectionElementNew, Stringable):
...
                 + str(object(entry[].value.copy()))
             )
             print_sep = True
-        res += "}"
-        return res
+        writer.write("}")
+        return
+
+    fn __repr__(self) -> String:
+        """Returns the name (in lowercase) of the specific object type."""
```
gryznar, line 622: (in lowercase)? `None` is represented as `"None"` and not `"none"`. The same applies to `fn __str__`

stdlib/src/builtin/object.mojo
```diff
@@ -577,32 +587,47 @@ struct _ObjectImpl(CollectionElement, CollectionElementNew, Stringable):
...
+        return self.__str__()
+
+    fn __str__(self) -> String:
+        """Returns the name (in lowercase) of the specific object type."""
+        var output = String("")
+        var writer = output._unsafe_to_formatter()
+        self.format_to(writer)
+        return output
```
ConnorGray, line 630: **Suggestion** We can avoid the need to touch the internal formatter functions if we change this to:

```mojo
fn __str__(self) -> String:
    return String.format_sequence(self)
```

ConnorGray:
Thanks for this improvement @bgreni! 🙂 

I left a couple of minor pieces of feedback, and I also agree with @gryznar's suggestions about a couple of doc strings.

Once those are taken care of, I'll proceed with syncing this in 🙂

stdlib/src/builtin/object.mojo
```diff
@@ -943,6 +979,16 @@ struct object(IntableRaising, ImplicitlyBoolable, Stringable):
         """
         return str(self._value)
```
soraros, line 980: I think we also need to move this `String.format_sequence`?
```mojo
fn __str__(self) -> String:
  return String.format_sequence(self)
```

ConnorGray:
!sync

Issue #3145, opened 2024-06-28
[stdlib] Add missing comparison operator tests


laszlokindrat:
The following types are missing tests for the operators required by the `Comparable` trait (`__le__`, `__ge__`, etc.): 

- [x] `Int`: done in https://github.com/modularml/mojo/pull/3161
- [x] `IntLiteral`: done in https://github.com/modularml/mojo/pull/3214
- [x] `FloatLiteral`: done in https://github.com/modularml/mojo/pull/3216
- [x] `SIMD`: done in https://github.com/modularml/mojo/pull/3221

We should also ensure they have tests for `__eq__` and `__ne__`. Keep in mind that dunder methods should be tested directly (i.e. not through operator syntax sugar) for these basic numeric types to avoid unintended implicit conversions.:

vguerra:
hello @laszlokindrat , just wanted to let you know that I pushed a couple of PRs to address the tasks in this issue: #3214 , #3216 and #3221

laszlokindrat:
This is now complete thanks to @MatveyF, @vguerra, and @martinvuyk! 🎉

martinvuyk:
I actually had nothing to do with this lol, just pushed a commit that referenced this during a rebase. Anyhow great work :D

PR #3161, opened 2024-07-01
[stdlib] `Int` comparison operator unit tests

MatveyF:
Added unit tests for `__lt__`, `__le__`, `__eq__`, `__ne__`, `__gt__`, and `__ge__` dunders of `Int`.

Please double check the tests in case I missed any corner cases. Other unit tests did not test interactions with other types so I did not do that either. 

Related to #3145

### DIFF
--- a/stdlib/test/builtin/test_int.mojo
+++ b/stdlib/test/builtin/test_int.mojo
@@ -195,48 +195,32 @@ def test_int_uint():
     assert_equal(0, int(u2))
 
 
-def test_lt():
+def test_comparison():
     assert_true(Int(5).__lt__(Int(10)))
     assert_true(Int(-10).__lt__(Int(-5)))
     assert_false(Int(0).__lt__(Int(0)))
     assert_false(Int(10).__lt__(Int(5)))
 
-
-def test_le():
     assert_true(Int(5).__le__(Int(10)))
     assert_true(Int(-10).__le__(Int(-5)))
     assert_true(Int(0).__le__(Int(0)))
     assert_false(Int(10).__le__(Int(5)))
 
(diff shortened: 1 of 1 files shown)


stdlib/test/builtin/test_int.mojo
```diff
@@ -195,6 +195,54 @@ def test_int_uint():
...
+
+
+def test_eq():
+    assert_true(Int(5).__eq__(Int(5)))
+    assert_true(Int(0).__eq__(Int(0)))
+    assert_false(Int(0).__eq__(Int(1)))
+    assert_false(Int(5).__eq__(Int(10)))
+
+    var a: Int = 5
+    assert_true(a.__eq__(a))
```
laszlokindrat, line 219: This case (and the one below for `__ne__`) is not really needed, could you please drop them?

stdlib/test/builtin/test_int.mojo
```diff
@@ -195,6 +195,54 @@ def test_int_uint():
     assert_equal(0, int(u2))
 
 
+def test_lt():
```
laszlokindrat, line 198: Instead of separate functions, could you please just put these into a single `test_comparisons`? We don't have a consistent strategy for testing these, so for now I would like to reduce the number of test functions.

laszlokindrat (approved):
Thanks for the patch! I have few nits, but LGTM otherwise. Could you please rebase on the latest nightly?

MatveyF:
Hi @laszlokindrat! Thank you for your review, I address your suggestions and rebase the branch later today

laszlokindrat:
!sync

Issue #3157, opened 2024-06-30
[Feature Request] Simple float conversion


RichardCoppin:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Currently `float()` behaves weirdly in VSCode IDE. It seems to be identified as a keyword by the LSP (I assume), but then when you try to use it, we get the error `use of unknown declaration 'float'`:
![image](https://github.com/modularml/mojo/assets/59029926/78872e29-a61f-4e8a-9c88-1b0e36a54c1a)

We also do not have a way to denote a class which can be converted to a floating point value.

Could we not just have something similar to the integer class:

```mojo

trait Floatable:
    """ `Floatable` trait describes a type that can be converted into an floating point representation.
   
    An struct will need to implement the `__float__(self) -> FLoat64` method to be considered 'Floatable'
    """


    fn __float__(self) -> Float64:
        """Get the float representation of the struct.

        Returns:
            A Float64 value.
        """
        ...

fn float(value: Int) -> Float64:
    return Float64(value)

fn float(value: Float64) -> Float64:
    return value

fn float[T: Floatable](value: T) -> Float64:
    return value.__float__()
```

### What is your motivation for this change?

Floating point values seem to be a bit of a sharp edge in Mojo at the moment.

The `float` token is also exhibiting contradictory behavior. VSCode identifies it as a keyword, but the compiler does not have a definition of it.

When constructing your own struct, which should be convertible into float, there is no simple way of casting to floating point:

```mojo
struct Floatable_Struct():
    var value: Float64

    fn __init__(inout self, value: Float64):
        self.value = value

    fn __float__(self) -> Float64:
        return Float64(self.value)
```
To extract the floating point value of the struct, one would need to do something like:
```mojo
fn main()
    var flt_st: Floatable_Struct = Floatable_Struct(0)
    print(Float64(flt_st.__float__())
```
Trying to do a mere conversion without the clunky dunder call, just results in an error:
![image](https://github.com/modularml/mojo/assets/59029926/3ccad7c0-5985-4e10-9d56-46967609c136)

Yet with the above
```mojo
struct Floatable_Struct(Floatable):
    var value: Float64

    fn __init__(inout self, value: Float64):
        self.value = value

    fn __float__(self) -> Float64:
        return Float64(self.value)
```
I can just call :
```mojo
[...]

PR #3163, opened 2024-07-01
[stdlib] Add `Floatable` trait

bgreni:
Begins #3157

Add `Floatable` and `FloatableRaising` traits to allow casting compliant types to `Float64`.

### DIFF
--- a/stdlib/src/builtin/bool.mojo
+++ b/stdlib/src/builtin/bool.mojo
@@ -108,6 +108,7 @@ struct Bool(
     Representable,
     Stringable,
     Formattable,
+    Floatable,
 ):
     """The primitive Bool scalar value used in Mojo."""
 
@@ -247,6 +248,15 @@ struct Bool(
             1 if the Bool is True, 0 otherwise.
         """
         return _select_register_value(self.value, Int(1), Int(0))
+
+    @always_inline("nodebug")
+    fn __float__(self) -> Float64:
+        """Convert this Bool to a float.
+
+        Returns:
+            1.0 if True else 0.0 otherwise.
+        """
+        return _select_register_value(self.value, Float64(1.0), Float64(0.0))
 
     @always_inline("nodebug")
     fn __index__(self) -> Int:
--- a/stdlib/src/builtin/floatable.mojo
+++ b/stdlib/src/builtin/floatable.mojo
@@ -1,0 +1,151 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+"""Implements the `Floatable` and `FloatableRaising` traits.
+
+These are Mojo built-ins, so you don't need to import them.
+"""
+
+
+trait Floatable:
+    """The `Floatable` trait describes a type that can be converted to a Float.
+
+    Any type that conforms to `Floatable` works with the built-in `float`
+    function.
+
+    This trait requires the type to implement the `__float__()` method.
+
+    For example:
+
+    ```mojo
+    @value
+    struct Foo(Floatable):
+        var i: Float64
+
+        fn __float__(self) -> Float64:
+            return self.i
+    ```
+
+    A `Foo` can now be converted to a `Float64` using `float`:
+
+    ```mojo
+    var f = float(Foo(5.5))
+    ```
+
+    **Note:** If the `__float__()` method can raise an error, use
+    the [`FloatableRaising`](/mojo/stdlib/builtin/floatable/floatableraising)
+    trait instead.
+    """
+
+    fn __float__(self) -> Float64:
+        """Get the float point representation of the value.
+
+        Returns:
+            The float point representation of the value.
+        """
+        ...
+
+
+trait FloatableRaising:
+    """The `FloatableRaising` trait describes a type that can be converted to a
+    Float, but the conversion might raise an error (e.g.: a string).
+
+    Any type that conforms to `FloatableRaising` works with the built-in `float`
+    function.
+
+    This trait requires the type to implement the `__float__()` method, which
+    can raise an error.
+
+    For example:
+
+    ```mojo
+    @value
+    struct MaybeFloat(FloatableRasing):
+        var value: Variant[Float64, NoneType]
+
+        fn __float__(self) raises -> Float64:
+            if self.value.isa[NoneType]():
+                raise "Float expected"
+            return self.value[Float64]
+    ```
+
+    A `MaybeFloat` can now be converted to `Float64` using `float`:
+
+    ```mojo
+    try:
+        print(float(MaybeFloat(4.6)))
+    except:
+        print("error occured")
+    ```
+    """
+
+    fn __float__(self) raises -> Float64:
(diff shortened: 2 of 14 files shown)


stdlib/src/builtin/floatable.mojo
```diff
@@ -0,0 +1,134 @@
...
+
+        Returns:
+            The float point representation of the value.
+        """
+        ...
+
+
+trait FloatableRaising:
+    """The `FloatableRaising` trait describes a type that can be converted to a Float,
+    but the conversion might raise an error (eg: a string).
```
laszlokindrat, line 60: nit:
```suggestion
    but the conversion might raise an error (e.g.: a string).
```

stdlib/src/builtin/floatable.mojo
```diff
@@ -0,0 +1,134 @@
...
+
+    Any type that conforms to `FloatableRaising` works with the built-in `float` function.
```
laszlokindrat, line 62: nit: Please ensure docstrings conform to 80 char line length.

laszlokindrat (commented):
Makes sense to me, thanks! Could you please
- add a changelog entry;
- make at least one type conform to `Floatable`;
- and add tests?

laszlokindrat:
@bgreni just a gentle ping, looks like you resolved my comments but maybe forgot to push? Also, don't forget to rebase :)

bgreni:
> @bgreni just a gentle ping, looks like you resolved my comments but maybe forgot to push? Also, don't forget to rebase :)

Yes I was close to making a push last night before I was pulled away by something else, I'll have it done soon!

bgreni:
@laszlokindrat I was trying to implement `Floatable` on `Int` but the compiler is yelling gibberish at me when I do it for some reason? 

```
Included from /Users/bgreni/Coding/mojo/stdlib/src/builtin/__init__.mojo:1:
/Users/bgreni/Coding/mojo/stdlib/src/builtin/simd.mojo:83:17: error: recursive reference to declaration
alias Float64 = Scalar[DType.float64]
                ^
Included from /Users/bgreni/Coding/mojo/stdlib/src/builtin/__init__.mojo:1:
/Users/bgreni/Coding/mojo/stdlib/src/builtin/simd.mojo:62:15: note: previously used here
alias UInt8 = Scalar[DType.uint8]
              ^
Included from /Users/bgreni/Coding/mojo/stdlib/src/builtin/__init__.mojo:1:
/Users/bgreni/Coding/mojo/stdlib/src/builtin/string.mojo:795:22: error: 'String' value has no attribute '_buffer_type'
    var _buffer: Self._buffer_type
```

laszlokindrat:
> _buffer_type



> @laszlokindrat I was trying to implement `Floatable` on `Int` but the compiler is yelling gibberish at me when I do it for some reason?
> 
> ```
> Included from /Users/bgreni/Coding/mojo/stdlib/src/builtin/__init__.mojo:1:
> /Users/bgreni/Coding/mojo/stdlib/src/builtin/simd.mojo:83:17: error: recursive reference to declaration
> alias Float64 = Scalar[DType.float64]
>                 ^
> Included from /Users/bgreni/Coding/mojo/stdlib/src/builtin/__init__.mojo:1:
> /Users/bgreni/Coding/mojo/stdlib/src/builtin/simd.mojo:62:15: note: previously used here
> alias UInt8 = Scalar[DType.uint8]
>               ^
> Included from /Users/bgreni/Coding/mojo/stdlib/src/builtin/__init__.mojo:1:
> /Users/bgreni/Coding/mojo/stdlib/src/builtin/string.mojo:795:22: error: 'String' value has no attribute '_buffer_type'
>     var _buffer: Self._buffer_type
> ```

yeah, this makes no sense, did something go wrong with the rebase maybe?

stdlib/test/builtin/test_bool.mojo
```diff
@@ -145,6 +145,11 @@ def test_comparisons():
     assert_true(True <= True)
 
 
+def test_float_conversion():
+    assert_equal(float(True), 1.0)
+    assert_equal(float(False), 0.0)
```
laszlokindrat, line 150: please test `__float__` directly to avoid accidental conversion for these primitive types

stdlib/test/builtin/test_float_literal.mojo
```diff
@@ -192,6 +192,11 @@ def test_abs():
     assert_equal(FloatLiteral.__abs__(neg_inf), inf)
 
 
+def test_float_conversion():
+    assert_equal(float(4.5), 4.5)
+    assert_equal(float(0.0), 0.0)
```
laszlokindrat, line 197: See me other comment; this might not actually test `FloatLiteral.__float__` due to implicit conversion and materialization.

stdlib/src/builtin/bool.mojo
```diff
@@ -232,6 +233,17 @@ struct Bool(
...
+    @always_inline("nodebug")
+    fn __float__(self) -> Float64:
+        """Convert this Bool to a float.
+
+        Returns:
+            1.0 if True else 0.0 otherwise.
+        """
+        return __mlir_op.`pop.select`[_type=Float64](
+            self.value, Float64(1.0), Float64(0.0)
+        )
```
laszlokindrat, line 245: Could it be that this causes the recursive reference somehow?

laszlokindrat (approved):
Thank you! Do you know any type that could conform to `FloatableRaising`?

bgreni:
> Thank you! Do you know any type that could conform to `FloatableRaising`?

Actually upon remembering that `atof` has already been implemented, making `String` conformant to `FloatableRaising` is trivial!

stdlib/src/builtin/bool.mojo
bgreni: I don't think so because `Int` was the first thing I tried to implement it on, so the only other changes were adding the trait itself at the time.

bgreni:
@laszlokindrat I've added `FloatableRaising` to `String` and `StringLiteral`. `Int` is still the only one that's given me that compiler error.

laszlokindrat:
> `Int` is still the only one that's given me that compiler error.

I think this is because `Floatable` depends on `Float64`, which is an alias for `SIMD`, and `SIMD` depends on `Int` through the `size` parameter, hence the dependency cycle. I don't have a solution though :(

bgreni:
> I think this is because `Floatable` depends on `Float64`, which is an alias for `SIMD`, and `SIMD` depends on `Int` through the `size` parameter, hence the dependency cycle. I don't have a solution though :(

Oooooooh yikes ok. Well for now I've added a special overload to `float` for `Int`. Are there any plans for a standalone `Float` type equivalent to `Int` that we could use instead of `Float64` in the future?

laszlokindrat:
> Are there any plans for a standalone `Float` type equivalent to `Int` that we could use instead of `Float64` in the future?

We actually discussed this recently, and decided not to do it, because it would require us to pick a representation (e.g. `f64`), and unlike `Int`, it doesn't seem to make sense to make this a platform dependent size. That being said, this dependency cycle problem is on mind, so we might circle back to it.

stdlib/src/builtin/bool.mojo
JoeLoser: We can fix this up after it lands, but we can use the new `_select_register_value` I added at https://github.com/modularml/mojo/blob/965c59196977aedac475cc6367071fbb3de6c824/stdlib/src/utils/_select.mojo

JoeLoser (approved):

laszlokindrat:
@bgreni looks like CI is red now, is this a recent change?

bgreni:
> @bgreni looks like CI is red now, is this a recent change?

The same example is failing on my other PR so I think it's independent of these changes?

JoeLoser:
> > @bgreni looks like CI is red now, is this a recent change?
> 
> The same example is failing on my other PR so I think it's independent of these changes?

Yeah, `nbody` example is messed up right now and marked unsupported internally which will make its way out in the latest nightly.  It's fine to proceed with this PR @laszlokindrat.

laszlokindrat:
!sync

PR #3166, opened 2024-07-01
[stdlib] Be explicit about `PythonObject` conversions

gabrieldemarmiesse:
@ConnorGray for review


stdlib/src/python/object.mojo
```diff
@@ -101,7 +101,6 @@ struct _PyIter(Sized):
 
 @register_passable
 struct PythonObject(
-    CollectionElement,
```
gabrieldemarmiesse, line 104: Not needed since there is the KeyElement trait.

JoeLoser (approved):

JoeLoser:
!sync

PR #3171, opened 2024-07-03
[Stdlib] Add `Dict.__init__` overload with `power_of_two_initial_capacity`

rd4com:
Hello, here is a new `__init__` that let user specify the initial reservation.

Skipping the un-necessary growth phases when the final size is known to be big.

&nbsp;

2.2x speedup on insertion, lookup should be faster aswel (during overcapacity).

1056082202       **PR**
35184367894528

2388001454
35184367894528

```mojo
from time import now
def main():
    var dict_size = 1<<23
    var start = now()
    var stop = now()
    start = now()
    x = Dict[Int,Int](power_of_two_initial_capacity = 1<<24)
    for i in range(dict_size):
        x[i] = i
    stop = now()
    
    print(stop-start)
    var result = 0
    for i in range(len(x)):
        result+=x[i]
    print(result)
```

We started with 24 and not 23 because the dict grows at 2/3 of capacity,
but reserving 23 is a huge speedup too.

&nbsp;

💡 Could this ameliorate variadic keyword arguments ?
If mojo knows there are 8 arguments passed,
maybe reserve 16 (and skip the dict growth phases).

&nbsp;

@gabrieldemarmiesse  suggested a `power_of_two` type in a PR,
thought about it too during this one ! :+1:

### DIFF
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -34,6 +34,7 @@ See the `Dict` docs for more details.
 from builtin.value import StringableCollectionElement
 
 from .optional import Optional
+from bit import is_power_of_two
 
 
 trait KeyElement(CollectionElement, Hashable, EqualityComparable):
@@ -459,6 +460,11 @@ struct Dict[K: KeyElement, V: CollectionElement](
         ```
 
         """
+        debug_assert(
+            bit.is_power_of_two(power_of_two_initial_capacity)
+            and power_of_two_initial_capacity >= 8,
+            "power_of_two_initial_capacity need to be >=8 and a power of two",
+        )
         self.size = 0
         self._n_entries = 0
         self._entries = Self._new_entries(power_of_two_initial_capacity)
--- a/stdlib/src/collections/dict.mojo
+++ b/stdlib/src/collections/dict.mojo
@@ -444,6 +444,26 @@ struct Dict[K: KeyElement, V: CollectionElement](
         self._entries = Self._new_entries(Self._initial_reservation)
         self._index = _DictIndex(len(self._entries))
 
+    @always_inline
+    fn __init__(inout self, *, power_of_two_initial_capacity: Int):
+        """Initialize an empty dictiontary with a pre-reserved initial capacity.
+
+        Args:
+            power_of_two_initial_capacity: At least 8, has to be a power of two.
+
+        Example usage:
+
+        ```mojo
+        var x = Dict[Int,Int](power_of_two_initial_capacity = 1024)
+        # Insert (2/3 of 1024) entries without reallocation.
+        ```
+
+        """
+        self.size = 0
+        self._n_entries = 0
+        self._entries = Self._new_entries(power_of_two_initial_capacity)
+        self._index = _DictIndex(len(self._entries))
+
     # TODO: add @property when Mojo supports it to make
     # it possible to do `self._reserved`.
     @always_inline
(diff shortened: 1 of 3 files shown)


stdlib/src/collections/dict.mojo
```diff
@@ -444,6 +444,26 @@ struct Dict[K: KeyElement, V: CollectionElement](
         self._entries = Self._new_entries(Self._initial_reservation)
         self._index = _DictIndex(len(self._entries))
 
+    @always_inline
+    fn __init__(inout self, *, power_of_two_initial_capacity: Int):
+        """Initialize an empty dictiontary with a pre-reserved initial capacity.
+
+        Args:
+            power_of_two_initial_capacity: At least 8, has to be a power of two.
```
JoeLoser, line 452: **Suggestion** Add a `debug_assert` to check this invariant.  Another approach would be to round up the initial capacity provided to the nearest power of two to retain the class invariants.

rd4com: Yes, commit [f7c4d94](https://github.com/modularml/mojo/pull/3171/commits/f7c4d9498c8b8c4ec34312f1ab5432c816e04387) brings a `debug_assert` :+1:

rd4com:
#### 🔁 Updates:
- `debug_assert` for `is_power_of_two` and `>= 8`
- `def test_init_inital_capacity()` in `test_dict.mojo`

#### ➕ In addition to the last review:

We might need another overload for rounded initialization, 
as it could affect performance, what do you think ?

I'd propose a chained solution in the form of:

```
var dictionary = Dict[Int,Int](
   power_of_two_initial_capacity = Dict.required_capacity_for_n_entries(600)
)
```

It would return `1024` (because `Dict` grows on `2/3` capacity)

#### 🗒️ Design notes
The keyword argument `power_of_two_initial_capacity` is used,
so that it naturally brings to the attention of the user the power of two invariant.

JoeLoser (approved):

JoeLoser:
> #### 🔁 Updates:
> * `debug_assert` for `is_power_of_two` and `>= 8`
> * `def test_init_inital_capacity()` in `test_dict.mojo`
> 
> #### ➕ In addition to the last review:
> We might need another overload for rounded initialization, as it could affect performance, what do you think ?
> 
> I'd propose a chained solution in the form of:
> 
> ```mojo
> var dictionary = Dict[Int,Int](
>    power_of_two_initial_capacity = Dict.required_capacity_for_n_entries(600)
> )
> # Dict.effective_capacity_for(entries = 600)
> # Dict.effective_capacity_for[entries = 600]()
> ```
> 
> It would return `1024` (because `Dict` grows on `2/3` capacity)
> 
> #### 🗒️ Design notes
> The keyword argument `power_of_two_initial_capacity` is used, so that by typing "power_of_two", a user can learn the invariant.
> 
> We could introduce the concept of effective capactity for user-friendlyness, as capacity does not really mean that it is capable of having `x` items.
> 
> ```mojo
> var dictionary = Dict[Int,Int](
>    initial_effective_capacity = 600 # it does the math of 2/3
> )
> ```
> 
> ```mojo
> # performance
> var dictionary = Dict[Int,Int](
>    power_of_two_initial_capacity = Dict.effective_capacity_for[600]()
> )
> ```

Let's keep it as-is now and see how our collections grow over time with power-of-two ideas for performance.  I think it's more user-friendly to round up the provided capacity to the nearest power of two, but let's lazily evaluate that.  Thanks!

JoeLoser:
!sync

Issue #3065, opened 2024-06-17
[BUG] Unsigned integer casting overflowing as if signed when using `int()` or `UInt32()`


fnands:
### Bug description

Migrating this here after a bit of discussion in Discord.

It seems like casting to unsigned integers actually just casts to signed integers, but has different behaviour in different cases.

I.e. I would expect that if I took a `UInt32` value, set it to `129`, cast it to `UInt8` and then call `int()`, I would still get 129.

However, this seem to end up as `4294967169`, i.e. `2**32 - 127`

If I instead call `UInt32()` I get `1`, so it overflows as `2**32 + 1`? Not sure about the last one.

So despite the fact that no overflow is expected (`129 < 255`), it seems to overflow, i.e. act like an signed integer.

The same goes if I cast to `UInt16`, i.e. casting `32768` (`2**16/2 + 1`) to `UInt16` and then calling `int()` I get `4294934528` i.e. `2**32 - 2**16/2`

So it looks like the cast to `UInt8` and `UInt16` actually perform casts to `Int8` and `Int16`, and then overflow.

Even more confusingly, this behaviour is different based on where the original `UInt32` is from, but behaviour differs when initialising using `var` and `alias` (see below).

### Steps to reproduce

Most minimal example

```mojo
def main():
 
    var b: UInt32 = 129
    print(b.cast[DType.uint8]())
    print(int(b.cast[DType.uint8]()))
    print(UInt32(b.cast[DType.uint8]()))
```

Produces:

```
129
-127
1
```

Doesn't have the same effect in the REPL, but the REPL does seem to have a clue:

```mojo
  1> var a: UInt32 = 129 
  2.  
(SIMD[uint32, 1]) a = {
  (ui32) [0] = 129
}
  2> var b = a.cast[DType.uint8]() 
  3.  
(SIMD[uint8, 1]) b = {
  (ui8) [0] = -127
}
```

A little bit less of a minimal example:

```mojo
fn fill_super_minimal() -> List[UInt32]:
    var table = List[UInt32](capacity=4)
    table.append(129)
 
    # This overflows with alias, but not var
    table.append(int(table[0].cast[DType.uint8]()))
 
    # This overflows in both cases, as if uint8 overflowed.
    var a: UInt32 = 129
    table.append(int(a.cast[DType.uint8]()))
 
    # This overflows as if uint32 overflows
    var b: UInt32 = 129
    table.append(UInt32(b.cast[DType.uint8]()))
 
    return table
 
 
[...]

PR #3172, opened 2024-07-03
[stdlib] Fix SIMD uint cast

martinvuyk:
Fix issue #3065 by adding some manual bit padding when converting to uint and forcing the bits to be interpreted as the target type, and slicing when `target.bitwidth() < type.bitwidth()`

Added a bitwidth constraint blocked by issue #933 since the conversion when `target.bitwidth() > type.bitwidth()` uses llvm interleave intrinsic and merging this would cause conflicts for code using `SIMD.cast()` at compile time.

The same constraint also blocks slicing that for some reason doesn't work at compile time when bitcasting the result down to the new type width `failed to fold operation pop.bitcast`


soraros:
Does this work when `type` is floating point etc.?

martinvuyk:
> Does this work when `type` is floating point etc.?

It doesn't work at all lol,  but no the main issue is `uint -> int` and `int -> uint` casting

soraros:
Shouldn't you also add `type.is_integral()` etc. then?

stdlib/src/builtin/simd.mojo
```diff
@@ -1483,6 +1483,38 @@ struct SIMD[type: DType, size: Int](
...
+                alias zero_x4 = zero_x2.join(zero_x2)
+
+                @parameter
+                if target.bitwidth() // type.bitwidth() == 2:
+                    return bitcast[target, size](zero_x1.interleave(self))
+                elif target.bitwidth() // type.bitwidth() == 4:
+                    return bitcast[target, size](
+                        zero_x2.interleave(zero_x1.interleave(self))
+                    )
+                elif target.bitwidth() // type.bitwidth() == 8:
```
gryznar, line 1503: In every of theses 3 branches, you are calculating `target.bitwidth() // type.bitwidth()`. Separate variable?

laszlokindrat:
Thanks for the patch! I'm currently working on several conversion related issues with SIMD, so I would like to wait with this patch. Is this blocking some other work you're doing ATM?

martinvuyk:
> Is this blocking some other work you're doing ATM?

Nope, just wanted to add some ideas in case it inspired someone who actually understands MLIR :D

laszlokindrat:
Since #3065 has been fixed, this should not be needed anymore. Please open a new ticket if you run into any other problems!

martinvuyk:
this is a very pretty solution to look at :)
```mojo
return self.cast[_uint_type_of_width[int_width]()]().__int__()
```

PR #3179, opened 2024-07-05
[stdlib] Add `SIMD.count` for `Scalar` rhs

rd4com:
Hello,

here is a method that do a `SIMD` count:

```mojo
var B = SIMD[DType.bool, 32](True)
assert_equal(B.count(True), 32)

for i in range(32):
    if i%2 == 0:
        B[i] = False
    else:
        B[i] = True

assert_equal(B.count(True), 16)
assert_equal(B.count(False), 16)
```
The implementation looks like this:
```mojo
(self == value).cast[DType.uint8]().reduce_add().__int__()
```
And we do a `@parameter if` check to make sure `self.size` is below `256` :+1:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -289,9 +289,9 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   Example usage:
 
   ```mojo
-  var B = SIMD[DType.int64, 4](1)
-  B[0] = 0
-  print(B.count(1)) # 3
+  var vector_of_four_int64 = SIMD[DType.int64, 4](1)
+  vector_of_four_int64[0] = 0
+  print(vector_of_four_int64.count(1)) # 3
   ```
 
 ### 🦋 Changed
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -283,6 +283,17 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   # pw_gecos='System Administrator', pw_dir='/var/root', pw_shell='/bin/zsh')
   ```
 
+- Added `SIMD.count` method for `Scalar` rhs.
+  ([PR #3179](https://github.com/modularml/mojo/pull/3179) by [@rd4com](https://github.com/rd4com))
+
+  Example usage:
+
+  ```mojo
+  var B = SIMD[DType.int64, 4](1)
+  B[0] = 0
+  print(B.count(1)) # 3
+  ```
+
 ### 🦋 Changed
 
 - The pointer aliasing semantics of Mojo have changed. Initially, Mojo adopted a
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -289,9 +289,7 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   Example usage:
 
   ```mojo
-  var vector_of_four_int64 = SIMD[DType.int64, 4](1)
-  vector_of_four_int64[0] = 0
-  print(vector_of_four_int64.count(1)) # 3
+  print(SIMD[DType.uint8, 4](1, 1, 1, 0).count(1)) # 3
   ```
 
 ### 🦋 Changed
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -284,7 +284,8 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   ```
 
 - Added `SIMD.count` method for `Scalar` rhs.
-  ([PR #3179](https://github.com/modularml/mojo/pull/3179) by [@rd4com](https://github.com/rd4com))
+  ([PR #3179](https://github.com/modularml/mojo/pull/3179)
+  by [@rd4com](https://github.com/rd4com))
 
   Example usage:
 
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2161,7 +2161,7 @@ struct SIMD[type: DType, size: Int](
     fn count(self, value: Scalar[type]) -> Int:
         """Counts the number of occurrences of a value in the vector.
 
-        Example:
+        Examples:
 
         ```mojo
         print(SIMD[DType.uint8, 4](1, 1, 1, 0).count(1)) # 3
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2161,10 +2161,9 @@ struct SIMD[type: DType, size: Int](
     fn count(self, value: Scalar[type]) -> Int:
         """Counts the number of occurrences of a value in the vector.
 
-        Examples:
+        Example:
 
         ```mojo
-        var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
         print(SIMD[DType.uint8, 4](1, 1, 1, 0).count(1)) # 3
         ```
 
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2158,6 +2158,30 @@ struct SIMD[type: DType, size: Int](
         constrained[type.is_numeric(), "the SIMD type must be numeric"]()
         return __mlir_op.`pop.max`(self.value, other.value)
 
+    fn count(self, value: Scalar[type]) -> Int:
+        """Count the occurences the value in the vector.
+
+        Example usage:
+
+        ```mojo
+        var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
+        thirty_two_uint8[0] = 0
+        print(thirty_two_uint8.count(1)) # 31
+        ```
+
+        Args:
+            value: The value.
+
+        Returns:
+            The amount of time the value occurs in the vector.
+        """
+
+        @parameter
+        if self.size > 255:
+            constrained[False, "size > 255"]()
+            # adding 256 occurences of a value would wrap to 0
+        return (self == value).cast[DType.uint8]().reduce_add().__int__()
+
     # ===------------------------------------------------------------------=== #
     # Reduce operations
     # ===------------------------------------------------------------------=== #
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2178,7 +2178,7 @@ struct SIMD[type: DType, size: Int](
 
         constrained[self.size < 256, "size must be less than 256"]()
         # adding 256 occurences of a value would wrap to 0
-        return (self == value).cast[DType.uint8]().reduce_add().__int__()
+        return int((self == value).cast[DType.uint8]().reduce_add())
 
     # ===------------------------------------------------------------------=== #
     # Reduce operations
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2165,7 +2165,6 @@ struct SIMD[type: DType, size: Int](
 
         ```mojo
         var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
-        thirty_two_uint8[0] = 0
         print(SIMD[DType.uint8, 4](1, 1, 1, 0).count(1)) # 3
         ```
 
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2176,9 +2176,8 @@ struct SIMD[type: DType, size: Int](
             The number of occurrences of the value in the vector.
         """
 
-        @parameter
         constrained[self.size < 256, "size must be less than 256"]()
-            # adding 256 occurences of a value would wrap to 0
+        # adding 256 occurences of a value would wrap to 0
         return (self == value).cast[DType.uint8]().reduce_add().__int__()
 
     # ===------------------------------------------------------------------=== #
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2159,7 +2159,7 @@ struct SIMD[type: DType, size: Int](
         return __mlir_op.`pop.max`(self.value, other.value)
 
     fn count(self, value: Scalar[type]) -> Int:
-        """Count the occurences the value in the vector.
+        """Counts the number of occurrences of a value in the vector.
 
         Example usage:
 
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2174,8 +2174,9 @@ struct SIMD[type: DType, size: Int](
             The number of occurrences of the value in the vector.
         """
 
-        constrained[self.size < 256, "size must be less than 256"]()
-        # adding 256 occurences of a value would wrap to 0
+        @parameter
+        if size > 255:
+            return int((self == value).cast[DType.uint16]().reduce_add())
         return int((self == value).cast[DType.uint8]().reduce_add())
 
     # ===------------------------------------------------------------------=== #
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2161,7 +2161,7 @@ struct SIMD[type: DType, size: Int](
     fn count(self, value: Scalar[type]) -> Int:
         """Counts the number of occurrences of a value in the vector.
 
-        Example usage:
+        Examples:
 
         ```mojo
         var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2177,8 +2177,7 @@ struct SIMD[type: DType, size: Int](
         """
 
         @parameter
-        if self.size > 255:
-            constrained[False, "size > 255"]()
+        constrained[self.size < 256, "size must be less than 256"]()
             # adding 256 occurences of a value would wrap to 0
         return (self == value).cast[DType.uint8]().reduce_add().__int__()
 
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -2166,7 +2166,7 @@ struct SIMD[type: DType, size: Int](
         ```mojo
         var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
         thirty_two_uint8[0] = 0
-        print(thirty_two_uint8.count(1)) # 31
+        print(SIMD[DType.uint8, 4](1, 1, 1, 0).count(1)) # 3
         ```
 
         Args:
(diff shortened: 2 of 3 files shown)


gryznar:
How about naming this method just "count"? Will be more intuitive and compatible with Python naming

rd4com:
:+1: @gryznar , 
the implementation does not support vectors of sizes `256`,
(not sure if need to add a `#TODO: support vectors >= 256` ?)

I'd rather let somebody who understand how to optimize it do that,
instead of just using an `uint16` ?

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
...
+
+        Returns:
+            The amount of time the value occurs in the vector.
+        """
+
+        @parameter
+        if self.size > 255:
+            constrained[False, "size > 255"]()
+            # adding 256 occurences of a value would wrap to 0
+        return (self == value).cast[DType.uint8]().reduce_add().__int__()
```
martinvuyk, line 2176:
```suggestion
        return int((self == value).cast[DType.uint8]().reduce_add())
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
...
+        Args:
+            value: The value.
+
+        Returns:
+            The amount of time the value occurs in the vector.
+        """
+
+        @parameter
+        if self.size > 255:
+            constrained[False, "size > 255"]()
```
martinvuyk, line 2174:
```suggestion
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
...
+            # adding 256 occurences of a value would wrap to 0
```
martinvuyk, line 2175:
```suggestion
            # adding 256 occurences of a value would wrap to 0
            return int((self == value).cast[DType.uint16]().reduce_add())
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
         constrained[type.is_numeric(), "the SIMD type must be numeric"]()
         return __mlir_op.`pop.max`(self.value, other.value)
 
+    fn count(self, value: Scalar[type]) -> Int:
+        """Count the occurences the value in the vector.
+
+        Example usage:
```
martinvuyk, line 2157:
```suggestion
        Examples:
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
...
+
+        ```mojo
+        var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
+        thirty_two_uint8[0] = 0
+        print(thirty_two_uint8.count(1)) # 31
```
martinvuyk, line 2162:
```suggestion
        print(SIMD[DType.uint8, 4](1, 1, 1, 0).count(1)) # 3
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
...
         return __mlir_op.`pop.max`(self.value, other.value)
 
+    fn count(self, value: Scalar[type]) -> Int:
+        """Count the occurences the value in the vector.
+
+        Example usage:
+
+        ```mojo
+        var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
+        thirty_two_uint8[0] = 0
```
martinvuyk, line 2161:
```suggestion
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
         constrained[type.is_numeric(), "the SIMD type must be numeric"]()
         return __mlir_op.`pop.max`(self.value, other.value)
 
+    fn count(self, value: Scalar[type]) -> Int:
+        """Count the occurences the value in the vector.
+
+        Example usage:
+
+        ```mojo
+        var thirty_two_uint8 = SIMD[DType.uint8, 32](1)
```
martinvuyk, line 2160:
```suggestion
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
...
+        thirty_two_uint8[0] = 0
+        print(thirty_two_uint8.count(1)) # 31
+        ```
+
+        Args:
+            value: The value.
+
+        Returns:
+            The amount of time the value occurs in the vector.
```
martinvuyk, line 2169:
```suggestion
            The number of occurrences of the value in the vector.
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2151,6 +2151,30 @@ struct SIMD[type: DType, size: Int](
         constrained[type.is_numeric(), "the SIMD type must be numeric"]()
         return __mlir_op.`pop.max`(self.value, other.value)
 
+    fn count(self, value: Scalar[type]) -> Int:
+        """Count the occurences the value in the vector.
```
martinvuyk, line 2155:
```suggestion
        """Counts the number of occurrences of a value in the vector.
```

martinvuyk (commented):
Hi, I don't see why we would need to limit this to only 256 items :)

docs/changelog.md
```diff
@@ -255,6 +255,17 @@ what we publish.
   # pw_gecos='System Administrator', pw_dir='/var/root', pw_shell='/bin/zsh')
   ```
 
+- Added `SIMD.count` method for `Scalar` rhs.
+  ([PR #3179](https://github.com/modularml/mojo/pull/3179) by [@rd4com](https://github.com/rd4com))
+
+  Example usage:
+
+  ```mojo
+  var B = SIMD[DType.int64, 4](1)
```
gryznar, line 264: It will look better IMHO with some more meaningful name than just "B". It is also a instance like, so snake_case here

soraros:
I think the chance of dealing with super wide SIMD vectors is slim; you can just cast to `uint16` as a fallback.

docs/changelog.md
soraros: Just write `v` or `a` here for conformance of the naming guidelines, I guess.

stdlib/test/builtin/test_simd.mojo
```diff
@@ -1545,6 +1545,46 @@ def test_contains():
...
+    var B = SIMD[DType.bool, 32](True)
+    assert_equal(B.count(True), 32)
+    B[0] = False
+    assert_equal(B.count(True), 31)
+
+    for i in range(32):
+        if i % 2 == 0:
+            B[i] = False
+        else:
+            B[i] = True
```
soraros, line 1563:
```suggestion
        B[I] = i % 2
```

docs/changelog.md
rd4com: @gryznar , @soraros , fixed with new commit ([2adda8f](https://github.com/modularml/mojo/pull/3179/commits/2adda8f34908e51f32d58941093e66fc753d29ae))

rd4com:
@martinvuyk please bring thoses wonderful optimisations in new `PR`,
as with many changes suggested in small pieces, 
we cant see the current changes clearly.

If reviewer insist on this being done, we can do it here, 
but suggest it in one review block please :+1:

stdlib/test/builtin/test_simd.mojo
```diff
@@ -1545,6 +1545,43 @@ def test_contains():
     assert_false(0 in y or 5 in y)
 
 
+def test_count():
```
gryznar, line 1548: Sorry for split review, but I don't spot that earlier. "A", "B", "C" etc. should be lowercase to follow naming convention

rd4com: Yes, sorry for the convention on naming things, i'll improve as time goes on :+1: 

Next PR reflect the convention !

Please no sorry for split-reviews, and do as much reviews as needed, 

the review process is what refines things so it is fantastic :+1: 

Last message just suggested @martinvuyk to make code suggestions in one block,

as it is difficult to imagine what the final result of merging all small code suggestions looks.

martinvuyk:
> but code-suggest it in one review block please 👍
> (as it is difficult to imagine what the final result of merging all small code suggestions looks)

yeah sorry I don't know how to make them in one block in a way that can be committed directly in GitHub, I could write them as a comment and not a suggestion for change ;) thanks for the feedback

stdlib/src/builtin/simd.mojo
```diff
@@ -2158,6 +2158,30 @@ struct SIMD[type: DType, size: Int](
...
+        Args:
+            value: The value.
+
+        Returns:
+            The number of occurrences of the value in the vector.
+        """
+
+        @parameter
+        if self.size > 255:
+            constrained[False, "size > 255"]()
```
laszlokindrat, line 2181: Does this not work for some reason?
```suggestion
        constrained[self.size < 256, "size must be less than 256"]()
```

laszlokindrat (approved):
Nice, thanks!

stdlib/src/builtin/simd.mojo
rd4com: Hello @laszlokindrat, yes it works :+1:

stdlib/src/builtin/simd.mojo
gryznar: Seems that this was resolved, but not introduced

stdlib/src/builtin/simd.mojo
gryznar: Same here

stdlib/src/builtin/simd.mojo
gryznar: Same

stdlib/src/builtin/simd.mojo
gryznar: Resolved, not introduced

gryznar:
I see that some changes from @martinvuyk was marked as resolved, but they were not introduced. Is this intentional?

stdlib/src/builtin/simd.mojo
gryznar: Same

rd4com:
Ok i merged all of theses small suggestions and it does not work anymore as they are outdated.

Its not even picking the uint16 part which is the one we need the most.

i'll force push the clean starting point, can you remake suggestion please @martinvuyk ?

stdlib/src/builtin/simd.mojo
gryznar: The last one, I think

docs/changelog.md
```diff
@@ -283,6 +283,17 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
...
 
+- Added `SIMD.count` method for `Scalar` rhs.
+  ([PR #3179](https://github.com/modularml/mojo/pull/3179) by [@rd4com](https://github.com/rd4com))
+
+  Example usage:
+
+  ```mojo
+  var vector_of_four_int64 = SIMD[DType.int64, 4](1)
+  vector_of_four_int64[0] = 0
+  print(vector_of_four_int64.count(1)) # 3
```
gryznar, line 294: Hmm, maybe use here the same example what's below? Instead of these 3 lines above:
`print(SIMD[DType.uint8, 4](1, 1, 1, 0).count(1)) # 3`

stdlib/src/builtin/simd.mojo
rd4com: We got this one :+1:

stdlib/src/builtin/simd.mojo
rd4com: We got this one :+1:

stdlib/src/builtin/simd.mojo
rd4com: We got this one :+1:

stdlib/src/builtin/simd.mojo
rd4com: We got this one :+1:  (with a modification that removes the plurality `s` from `Examples` )

stdlib/src/builtin/simd.mojo
rd4com: We got this one :+1:

docs/changelog.md
rd4com: Yes @gryznar , new commit implement this suggestion :+1: (https://github.com/modularml/mojo/pull/3179/commits/87761d025c82d991e55fba7f1ce3cf8c1bbcff1c)

gryznar (approved):

rd4com:
@gryznar Fantastic, it's a collaborative effort, waiting for @martinvuyk :+1: 

Do you think that this optimization would be faster than `uint16` for vectors of size exactly `256` ?:

```mojo
@parameter
if self.size == 256:
    var compared = self == rhs
    var tmp = compare[255].__int__()
    compared[255] = False
    tmp += compared.cast[DType.uint8]().reduce_add().__int__()
    return tmp
```

gryznar:
Hmm, I don't know. Maybe it is better question to Laszlo or Connor from Modular?

stdlib/src/builtin/simd.mojo
```diff
@@ -2158,6 +2158,26 @@ struct SIMD[type: DType, size: Int](
         constrained[type.is_numeric(), "the SIMD type must be numeric"]()
         return __mlir_op.`pop.max`(self.value, other.value)
 
+    fn count(self, value: Scalar[type]) -> Int:
+        """Counts the number of occurrences of a value in the vector.
+
+        Example:
```
martinvuyk, line 2164: Mojo standard is to keep docstring section names plural
```suggestion
        Examples:
```

stdlib/src/builtin/simd.mojo
```diff
@@ -2158,6 +2158,26 @@ struct SIMD[type: DType, size: Int](
...
+
+        Args:
+            value: The value.
+
+        Returns:
+            The number of occurrences of the value in the vector.
+        """
+
+        constrained[self.size < 256, "size must be less than 256"]()
+        # adding 256 occurences of a value would wrap to 0
```
martinvuyk, line 2178: There is no need to limit ourselves to 255 items
```suggestion
        @parameter
        if size > 255:
            return int((self == value).cast[DType.uint16]().reduce_add())
```

docs/changelog.md
```diff
@@ -283,6 +283,15 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   # pw_gecos='System Administrator', pw_dir='/var/root', pw_shell='/bin/zsh')
   ```
 
+- Added `SIMD.count` method for `Scalar` rhs.
+  ([PR #3179](https://github.com/modularml/mojo/pull/3179) by [@rd4com](https://github.com/rd4com))
```
martinvuyk, line 287: markdownlint might give problems with this one
```suggestion
  ([PR #3179](https://github.com/modularml/mojo/pull/3179)
  by [@rd4com](https://github.com/rd4com))
```

martinvuyk (commented):
> Do you think that this optimization would be faster than uint16 for vectors of size exactly 256 ?:

I think it would be faster to just split the vector and reduce add `while size > 255`, but it might be overkill as I don't know the real performance difference for reduce_add between i16 and i8 in SIMD ops.

ConnorGray:
> Hmm, I don't know. Maybe it is better question to Laszlo or Connor from Modular?

I think we'd need a benchmark to say for certain.

 Since this is a new feature, I suggest we start with a simple but perhaps not maximally optimal implementation, and leave optimization work to future PRs where the discussion can be focused on performance. In this initial PR we should stay focused on correctness, quality, and documentation 🙂

rd4com:
@martinvuyk, thanks for the fast response and sorry for re-iterations!

With new commit, there is now a test for `SIMD`sizes `>= 256` :+1: 

ℹ️ new test uses `size` above `simdwidthof[DT]` (please make sure it is ok)

rd4com:
@ConnorGray , @laszlokindrat , this constrained failed to work:
```mojo
constrained[
    size <= SIMD[DType.uint16, 1].MAX_FINITE.__int__(),
    "size is above maximum finite value of uint16"
]()
```
It would run the error, even with `x = SIMD[DType.uint8, 32]`

laszlokindrat:
> @ConnorGray , @laszlokindrat , this constrained failed to work:
> 
> ```mojo
> constrained[
>     size <= SIMD[DType.uint16, 1].MAX_FINITE.__int__(),
>     "size is above maximum finite value of uint16"
> ]()
> ```
> 
> It would run the error, even with `x = SIMD[DType.uint8, 32]`

Any chance you could provide a standalone repro along with the expected and actual behaviors, and file this? I can take a look.

laszlokindrat (approved):
Nice!

laszlokindrat:
!sync

rd4com:
@laszlokindrat, 
Yes, i'm preparing the bug report and will ping you from there :+1: 
(should be ready in less than 20 minutes)

laszlokindrat:
@rd4com Is there a particular use case for this? Seems like we might not be able to take this without some justification.

rd4com:
@laszlokindrat , if there is no need, no merge :+1: 

Most `SIMD` users can probably do the `==`, `cast`, `reduce_add`.
So it's mainly a new step for new-comers to progressively learn and use SIMD.

Interesting use-case:
close the gap between high-level/low-level by having both worlds have common methods.
(`List.count`, `SIMD.count`,...)

just proposing features, but never sure if they useful enough,
so it is all good if the PR is not merged :+1:

rd4com:
@laszlokindrat 
Don't merge, there is a nasty bug with `DType.bool` for `SIMD.load`: 
- https://github.com/modularml/mojo/issues/3229

Did already implement the `List.count` for `Scalar[DT]` as a fallback to this `PR`, 
but it uses `SIMD.load` too!

rd4com:
#### 🔁 Update: this pr seem to not be affected by the bug

Possibly because it does not use `SIMD.load`

```mojo
for o in range(32):
    var B = SIMD[DType.bool, 32](True)
    assert_equal(B.count(True), 32)
    for i in range(32):
        if i==o: B[i] = False
        else: B[i] = True
    assert_equal(B[o], False)
    for i in range(32):
        if i==o: assert_equal(B[i], False)
        else: assert_equal(B[i], True)
```

```mojo
for o in range(32):
    var B = SIMD[DType.bool, 32](False)
    assert_equal(B.count(False), 32)
    for i in range(32):
        if i==o: B[i] = True
        else: B[i] = False
    assert_equal(B[o], True)
    for i in range(32):
        if i==o: assert_equal(B[i], True)
        else: assert_equal(B[i], False)
```

laszlokindrat:
> Interesting use-case:
close the gap between high-level/low-level by having both worlds have common methods.
(List.count, SIMD.count,...)

For the record, I completely agree with this, but unfortunately we can't currently take this without a concrete use case. I really appreciate your work, though, and look forward to more of your patches!

Issue #2639, opened 2024-05-13
[Feature Request] Add `stol` function


YichengDWu:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

`atol` is in the std but not `stol`, a similar to `std::stol` in cpp.

### What is your motivation for this change?

More functionalities to parser strings.

gxyd:
Can I maybe try to work on this issue?

soraros:
@gxyd You can just go ahead!

bgreni:
Since `atol` in mojo already accepts an optional `base` argument why would we have need for a second function in this case?

YichengDWu:
`atol` recognizes `1234` and `stol` regonize `1234asda`.

bgreni:
Ah I see the distinction between the behaviours now thank you.

PR #3180, opened 2024-07-05
[stdlib]  Fix atol parsing for prefixed integer literals with leading underscores

jjvraw:
This addresses #3182 to correctly parse prefixed string literals. Changes include:
- Bug-fix
- Added appropriate test cases to verify the new behaviour

When the base 2, 8, 16 are set or inferred (base 0), for a string prefixed appropriately, leading underscores should be allowed as per Python's integer literal grammar: 

```
integer      ::=  decinteger | bininteger | octinteger | hexinteger
decinteger   ::=  nonzerodigit (["_"] digit)* | "0"+ (["_"] "0")*
bininteger   ::=  "0" ("b" | "B") (["_"] bindigit)+
octinteger   ::=  "0" ("o" | "O") (["_"] octdigit)+
hexinteger   ::=  "0" ("x" | "X") (["_"] hexdigit)+
nonzerodigit ::=  "1"..."9"
digit        ::=  "0"..."9"
bindigit     ::=  "0" | "1"
octdigit     ::=  "0"..."7"
hexdigit     ::=  digit | "a"..."f" | "A"..."F"
```

Currently, this such a string passed to `atol` raises. The changes in this PR fixes this issue. 

I should note that I have another PR (#3207) to refactor, and clean up, `atol` for reusability to introduce `stol` as per request #2639. If this PR is merged, I will update #3207, which will require a minor tweak.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -471,6 +471,13 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
 
 - `SIMD.load/store` now supports `UnsafePointer` overloads.
 
+- The `atol` function now correctly supports leading underscores,
+  (e.g.`atol("0x_ff", 0)`), when the appropriate base is specified or inferred
+  (base 0). non-base-10 integer literals as per Python's [Integer Literals](\
+  <https://docs.python.org/3/reference/lexical_analysis.html#integers>).
+  ([PR #3180](https://github.com/modularml/mojo/pull/3180)
+  by [@jjvraw](https://github.com/jjvraw))
+
 ### ❌ Removed
 
 - It is no longer possible to cast (implicitly or explicitly) from `Reference`
--- a/stdlib/src/builtin/string.mojo
+++ b/stdlib/src/builtin/string.mojo
@@ -228,6 +228,7 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
     var ord_letter_max = (-1, -1)
     var result = 0
     var is_negative: Bool = False
+    var has_prefix: Bool = False
     var start: Int = 0
     var str_len = len(str_ref)
     var buff = str_ref.unsafe_ptr()
@@ -250,14 +251,17 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
             str_ref[start + 1] == "b" or str_ref[start + 1] == "B"
         ):
             start += 2
+            has_prefix = True
         elif base == 8 and (
             str_ref[start + 1] == "o" or str_ref[start + 1] == "O"
         ):
             start += 2
+            has_prefix = True
         elif base == 16 and (
             str_ref[start + 1] == "x" or str_ref[start + 1] == "X"
         ):
             start += 2
+            has_prefix = True
 
     alias ord_0 = ord("0")
     # FIXME:
@@ -269,6 +273,7 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
         var real_base_new_start = _identify_base(str_ref, start)
         real_base = real_base_new_start[0]
         start = real_base_new_start[1]
+        has_prefix = real_base != 10
         if real_base == -1:
             raise Error(_atol_error(base, str_ref))
     else:
@@ -285,10 +290,9 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
 
     var found_valid_chars_after_start = False
     var has_space_after_number = False
-    # single underscores are only allowed between digits
-    # starting "was_last_digit_undescore" to true such that
-    # if the first digit is an undesrcore an error is raised
-    var was_last_digit_undescore = True
+    # Prefixed integer literals with real_base 2, 8, 16 may begin with leading
+    # underscores under the conditions they have a prefix
+    var was_last_digit_undescore = not (real_base in (2, 8, 16) and has_prefix)
     for pos in range(start, str_len):
         var ord_current = int(buff[pos])
         if ord_current == ord_underscore:
--- a/stdlib/test/builtin/test_string.mojo
+++ b/stdlib/test/builtin/test_string.mojo
@@ -365,6 +365,9 @@ def test_atol():
     assert_equal(10, atol("0o12", 8))
     assert_equal(10, atol("0O12", 8))
     assert_equal(35, atol("Z", 36))
+    assert_equal(255, atol("0x_00_ff", 16))
+    assert_equal(18, atol("0b0001_0010", 2))
+    assert_equal(18, atol("0b_000_1001_0", 2))
 
     # Negative cases
     with assert_raises(
@@ -398,11 +401,36 @@ def test_atol():
     ):
         _ = atol("5", 5)
 
+    with assert_raises(
+        contains="String is not convertible to integer with base 10: '0x_ff'"
+    ):
+        _ = atol("0x_ff")
+
+    with assert_raises(
+        contains="String is not convertible to integer with base 3: '_12'"
+    ):
+        _ = atol("_12", 3)
+
     with assert_raises(contains="Base must be >= 2 and <= 36, or 0."):
         _ = atol("0", 1)
 
     with assert_raises(contains="Base must be >= 2 and <= 36, or 0."):
         _ = atol("0", 37)
+
+    with assert_raises(
+        contains="String is not convertible to integer with base 16: '_ff'"
+    ):
+        _ = atol("_ff", base=16)
+
+    with assert_raises(
+        contains="String is not convertible to integer with base 2: '  _01'"
+    ):
+        _ = atol("  _01", base=2)
+
+    with assert_raises(
+        contains="String is not convertible to integer with base 10: '0x_ff'"
+    ):
+        _ = atol("0x_ff")
 
     with assert_raises(
         contains="String is not convertible to integer with base 10: ''"
@@ -433,6 +461,14 @@ def test_atol_base_0():
 
     assert_equal(0, atol("0X0", base=0))
 
+    assert_equal(255, atol("0x_00_ff", base=0))
+
+    assert_equal(18, atol("0b_0001_0010", base=0))
+    assert_equal(18, atol("0b000_1001_0", base=0))
+
+    assert_equal(10, atol("0o_000_12", base=0))
+    assert_equal(10, atol("0o00_12", base=0))
+
     with assert_raises(
         contains="String is not convertible to integer with base 0: '  0x'"
     ):
@@ -452,11 +488,6 @@ def test_atol_base_0():
         contains="String is not convertible to integer with base 0: '0r100'"
     ):
         _ = atol("0r100", base=0)
-
-    with assert_raises(
-        contains="String is not convertible to integer with base 0: '0b_0'"
-    ):
-        _ = atol("0b_0", base=0)
 
     with assert_raises(
         contains="String is not convertible to integer with base 0: '0xf__f'"


martinvuyk:
Thanks for taking the time to do this, could you add some of the suggestions and tests for the other PR here? I see the same behavior missing (underscore and 0 at the beginning of non base 10 numbers)

stdlib/src/builtin/string.mojo
```diff
@@ -381,19 +420,30 @@ fn _identify_base(str_ref: StringRef, start: Int) -> Tuple[Int, Int]:
 fn atol(str: String, base: Int = 10) raises -> Int:
     """Parses and returns the given string as an integer in the given base.
 
-    For example, `atol("19")` returns `19`. If base is 0 the the string is
-    parsed as an Integer literal, see: https://docs.python.org/3/reference/lexical_analysis.html#integers.
+    For example, `atol("32")` returns `32`, and `atol("FF", 16)` returns `255`. If base is set to 0, the string is parsed as an Integer literal, with the following considerations:
```
martinvuyk, line 423: I'm not sure how many characters these are here since I don't have my IDE ruler, but mojo goes for 80 characters. You can set it up in vscode or whatever IDE you use if you wish. But this and other docstring lines are definitely above 80

stdlib/src/builtin/string.mojo
```diff
@@ -312,29 +289,91 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
...
+        str_ref: The input string.
+        str_len: Length of the input string.
+        base: The specified base.
+    Returns:
+        Updated position after the prefix, if applicable.
+    """
+    var start = pos
+    if str_ref[start] == "0" and start + 1 < str_len:
+        if base == 2 and (
+            str_ref[start + 1] == "b" or str_ref[start + 1] == "B"
```
martinvuyk, line 362: `str_ref[start + 1]` could be used as a variable before the if and spare many lines

stdlib/src/builtin/string.mojo
```diff
@@ -312,29 +286,86 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
...
+    Args:
+        pos: Current position in the string.
+        str_ref: The input string.
+        str_len: Length of the input string.
+        base: The specified base.
+    Returns:
+        Updated position after the prefix, if applicable.
+    """
+    var start = pos
+    var prefix_char = str_ref[start + 1]
```
martinvuyk, line 357: I meant before the inner if, this will try to access the pointer before checking `start + 1 < str_len`

stdlib/src/builtin/string.mojo
```diff
@@ -381,19 +412,32 @@ fn _identify_base(str_ref: StringRef, start: Int) -> Tuple[Int, Int]:
...
 
     Returns:
         An integer value that represents the string, or otherwise raises.
+
+    Notes:
+        This follows [Python's integer literals](\
+        https://docs.python.org/3/reference/lexical_analysis.html#integers).
+
+    See Also:
+        :func:`stol`: A similar function that returns both the parsed integer and the remaining unparsed string.
```
martinvuyk, line 440:
```suggestion
    - function `stol`: A similar function that returns both the parsed
        integer and the remaining unparsed string.
```
This is sphinx formatting right? Mojo follows [Google's Python convention for docstrings](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods).

Btw. could you also add some newlines between `(function summary) Args:` and `(end of args) Returns:` etc.?

I'm sorry for the nit picking but we have to keep docstrings consistent

martinvuyk (approved):
LGTM!
btw. loved this refactor
```mojo
    var start = pos
    if start + 1 < str_len:
        var prefix_char = str_ref[start + 1]
        if str_ref[start] == "0" and (
            (base == 2 and (prefix_char == "b" or prefix_char == "B"))
            or (base == 8 and (prefix_char == "o" or prefix_char == "O"))
            or (base == 16 and (prefix_char == "x" or prefix_char == "X"))
        ):
            start += 2
    return start
```

stdlib/src/builtin/string.mojo
jjvraw: Sure :) 

The single newline seems to be consistent with others, no?

stdlib/src/builtin/string.mojo
```diff
@@ -284,20 +261,17 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
 
     var found_valid_chars_after_start = False
     var has_space_after_number = False
-    # single underscores are only allowed between digits
-    # starting "was_last_digit_undescore" to true such that
-    # if the first digit is an undesrcore an error is raised
-    var was_last_digit_undescore = True
+    var was_last_digit_underscore = True if real_base == 10 else False
```
soraros, line 264:
```suggestion
    var was_last_digit_underscore = real_base == 10
```

jjvraw:
Added @martinvuyk @soraros as co-authors.

stdlib/src/builtin/string.mojo
```diff
@@ -312,29 +286,86 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
...
     if is_negative:
         result = -result
     return result
 
 
-fn _atol_error(base: Int, str_ref: StringRef) -> String:
+@always_inline
+fn _trim_and_handle_sign(str_ref: StringRef, str_len: Int) -> (Int, Bool):
+    """Trims leading whitespace and handles the sign of the number in the string.
+    Args:
```
martinvuyk, line 314:
```suggestion

    Args:
```

stdlib/src/builtin/string.mojo
```diff
@@ -312,29 +286,86 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
...
+        str_ref: A StringRef containing the number to parse.
+        str_len: The length of the string.
+    Returns:
```
martinvuyk, line 317:
```suggestion

    Returns:
```

stdlib/src/builtin/string.mojo
```diff
@@ -312,29 +286,86 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
...
+
+
+@always_inline
+fn _handle_base_prefix(
+    pos: Int, str_ref: StringRef, str_len: Int, base: Int
+) -> Int:
+    """Adjusts the starting position if a valid base prefix is present.
+    Handles "0b"/"0B" for base 2, "0o"/"0O" for base 8, and "0x"/"0X" for base 16.
+    Only adjusts if the base matches the prefix.
+    Args:
```
martinvuyk, line 348:
```suggestion

    Args:
```

stdlib/src/builtin/string.mojo
```diff
@@ -312,29 +286,86 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
...
+        pos: Current position in the string.
+        str_ref: The input string.
+        str_len: Length of the input string.
+        base: The specified base.
+    Returns:
```
martinvuyk, line 353:
```suggestion

    Returns:
```

jjvraw:
Thanks @martinvuyk :) But single newline seems to be consistent, no? 
https://github.com/modularml/mojo/blob/8bd1dbdf26c70c634768bfd4c014537f6fdb0fb2/stdlib/src/builtin/string.mojo#L32-L44

jjvraw:
I have repurposed this PR to focus on the mentioned issue.

stdlib/src/builtin/string.mojo
```diff
@@ -284,10 +284,7 @@ fn _atol(str_ref: StringRef, base: Int = 10) raises -> Int:
 
     var found_valid_chars_after_start = False
     var has_space_after_number = False
-    # single underscores are only allowed between digits
-    # starting "was_last_digit_undescore" to true such that
-    # if the first digit is an undesrcore an error is raised
-    var was_last_digit_undescore = True
+    var was_last_digit_undescore = real_base == 10
```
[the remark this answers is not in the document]
martinvuyk: You're right, I hadn't seen `"Base must be >= 2 and <= 36, or 0."` so that means this could receive other bases that have no prefix. So yeah totally, leading underscore should fail
Python:
```text
print(int("_123", 30))
ValueError: invalid literal for int() with base 30: '_123'
```
so it should be 
```mojo
real_base in (2, 8, 16)
```
nice catch, maybe we could add some tests with assert raises (and also passing ones) with bases other than the classics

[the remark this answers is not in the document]
jjvraw: Will for sure add some test cases with the next commit. We will have to add a check for whether there exists a prefix as well though. 

So something along the lines of `real_base in (2, 8, 16) and has_prefix`, as we could have `atol("_ff", 16)` which should raise as invalid. 

Was playing around with `(real_base & 0b11010) != real_base and has_prefix` which would've have been fun... 10 and 18 breaks it though.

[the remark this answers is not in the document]
martinvuyk: > So something along the lines of real_base in (2, 8, 16) and has_prefix, as we could have atol("_ff", 16) which should raise as invalid.

yeah I guess `str_ref[start] == "0"` can be taken out of the if to a has_prefix var, nice catch again :D

> Was playing around with (real_base & 0b11010) != real_base and has_prefix which would've have been fun...

Niiice, maybe this will work?
```mojo
from bit import is_power_of_two
...
# first character as underscore is only valid for bases 2, 8, 16 with prefix
var was_last_digit_undescore = is_power_of_two(real_base & 0b11010) and has_prefix:
    ...
```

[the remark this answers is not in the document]
jjvraw: Unfortunately, 10 and 18 breaks this as well. I like the thought though.

We can use a mask,  `bitmask = (1 << 2) | (1 << 8) | (1 << 16)` and then `not ((1 << real_base) & bitmask) and has_prefix`. 

This seems to work. But it might be overcomplicated for a simple value range check.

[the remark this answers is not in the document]
martinvuyk: > This seems to work. But it might be overcomplicated for a simple value range check.

yeah I think this was a fun exercise lol but let's just stick with tuple contains

just out of curiosity, does this work ?
```mojo
var mask = (0b11010 * (real_base < 17 and has_prefix).cast[DType.uint8]())
var was_last_digit_undescore = not is_power_of_two(real_base & mask)
```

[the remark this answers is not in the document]
soraros: > yeah I think this was a fun exercise lol but let's just stick with tuple contains

I agree. At least we can defer further optimisation to a later PR should we find it necessary.

jjvraw:
This will conflict with PR #3207, but we can make the necessary changes and clean up in that PR as well.

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
jjvraw: @martinvuyk, I couldn't get it to run. Cannot call `cast` on a Boolean.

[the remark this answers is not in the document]
martinvuyk: Then it will probably work by just putting `int(real_base < 17 and has_prefix)`

docs/changelog.md
```diff
@@ -419,6 +419,10 @@ what we publish.
 - `memcmp`, `memset` and `memset_zero` no longer take in `LegacyPointer`,
   instead, use `UnsafePointer`.
 
+- The `atol` function now correctly supports leading underscores,
+  (e.g.`atol("0x_ff", 0)`), when the appropriate base is specified or inferred
+  (base 0). non-base-10 integer literals as per Python's [Integer Literals](https://docs.python.org/3/reference/lexical_analysis.html#integers).
```
martinvuyk, line 424:
```suggestion
  (base 0). non-base-10 integer literals as per Python's [Integer Literals](\
  https://docs.python.org/3/reference/lexical_analysis.html#integers).
  [PR #3180](https://github.com/modularml/mojo/pull/3180)
   by [@jjvraw](https://github.com/jjvraw)
```

stdlib/src/builtin/string.mojo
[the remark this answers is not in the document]
jjvraw: Same issue @martinvuyk, but now for `Int`.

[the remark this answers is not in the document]
martinvuyk: yeah but removing the `.cast` since it is a SIMD method

JoeLoser:
!sync

JoeLoser:
!sync

PR #3187, opened 2024-07-07
[docs] Fix typo in changelog.md

Brian-M-J:
Change "complexify" to "complicate"


JoeLoser (commented):
@arthurevans thoughts here?  "complexify" is indeed a word, so I don't feel strongly here.

arthurevans (approved):
As @JoeLoser pointed out (to my surprise), the OED says that complexify is actually a word. It also says that it basically means the same thing as complicate, which is 100 times as common in English. 

so unless I'm missing an important distinction here that's specific to programming language or compiler design, for example, I'd take the more common word.

JoeLoser:
!sync

Brian-M-J:
> As @JoeLoser pointed out (to my surprise), the OED says that complexify is actually a word.

This is a surprise to me as well.

PR #3195, opened 2024-07-09
[stdlib] Enable equality comparisons for `List[T] == List[T]`

kszucs:
Noticed that this feature has been missing, so I implemented `List.__eq__` and `List.__ne__`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -16,6 +16,11 @@ what we publish.
 
 ### ⭐️ New
 
+- `List[T]` values are now equality comparable with `==` and `!=` when `T` is
+  equality comparable.
+  ([PR 3195#](https://github.com/modularml/mojo/pull/3195) by
+  [@kszucs](https://github.com/kszucs))
+
 - `__setitem__` now works with variadic argument lists such as:
 
   ```mojo
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -197,6 +197,62 @@ struct List[T: CollectionElement](
     # Operator dunders
     # ===-------------------------------------------------------------------===#
 
+    fn __eq__[
+        U: EqualityComparableCollectionElement
+    ](self: List[U], other: List[U]) -> Bool:
+        """Checks if two lists are equal.
+
+        Examples:
+        ```mojo
+        var x = List[Int](1, 2, 3)
+        var y = List[Int](1, 2, 3)
+        if x == y: print("x and y are equal")
+        ```
+
+        Parameters:
+            U: The type of the elements in the list. Must implement the
+               traits `EqualityComparable` and `CollectionElement`.
+
+        Args:
+            other: The list to compare with.
+
+        Returns:
+            True if the lists are equal, False otherwise.
+        """
+        if len(self) != len(other):
+            return False
+        var index = 0
+        for i in self:
+            if i[] != other[index]:
+                return False
+            index += 1
+        return True
+
+    fn __ne__[
+        U: EqualityComparableCollectionElement
+    ](self: List[U], other: List[U]) -> Bool:
+        """Checks if two lists are not equal.
+
+        Examples:
+
+        ```mojo
+        var x = List[Int](1, 2, 3)
+        var y = List[Int](1, 2, 4)
+        if x != y: print("x and y are not equal")
+        ```
+
+        Parameters:
+            U: The type of the elements in the list. Must implement the
+               traits `EqualityComparable` and `CollectionElement`.
+
+        Args:
+            other: The list to compare with.
+
+        Returns:
+            True if the lists are not equal, False otherwise.
+        """
+        return not (self == other)
+
     fn __contains__[
         U: ComparableCollectionElement
     ](self: List[U], value: U) -> Bool:
--- a/stdlib/test/collections/test_list.mojo
+++ b/stdlib/test/collections/test_list.mojo
@@ -816,6 +816,29 @@ def test_list_contains():
     # assert_equal(List(0,1) in y,False)
 
 
+def test_list_eq_ne():
+    var l1 = List[Int](1, 2, 3)
+    var l2 = List[Int](1, 2, 3)
+    assert_true(l1 == l2)
+    assert_false(l1 != l2)
+
+    var l3 = List[Int](1, 2, 3, 4)
+    assert_false(l1 == l3)
+    assert_true(l1 != l3)
+
+    var l4 = List[Int]()
+    var l5 = List[Int]()
+    assert_true(l4 == l5)
+    assert_true(l1 != l4)
+
+    var l6 = List[String]("a", "b", "c")
+    var l7 = List[String]("a", "b", "c")
+    var l8 = List[String]("a", "b")
+    assert_true(l6 == l7)
+    assert_false(l6 != l7)
+    assert_false(l6 == l8)
+
+
 def test_list_init_span():
     var l = List[String]("a", "bb", "cc", "def")
     var sp = Span(l)


stdlib/src/collections/list.mojo
```diff
@@ -197,6 +197,56 @@ struct List[T: CollectionElement](
     # Operator dunders
     # ===-------------------------------------------------------------------===#
 
+    fn __eq__[T2: ComparableCollectionElement](self: List[T], other: List[T2]) -> Bool:
```
kszucs, line 200: `ComparableCollectionElement` seems to require more than just `__eq__` and `__ne__` but it is the existing composition with `CollectionElement` hence I am using it.

stdlib/src/collections/list.mojo
```diff
@@ -197,6 +197,56 @@ struct List[T: CollectionElement](
...
+            other: The list to compare with.
+
+        Returns:
+            True if the lists are equal, False otherwise.
+        """
+        constrained[_type_is_eq[T, T2](), "other type is not self.T"]()
+        if len(self) != len(other):
+            return False
+        var index = 0
+        for i in self:
```
kszucs, line 222: Ideally it would be better to use `range(len(self))` here, but the iterator here returns with a `Reference` rather than a `ref` for each element which I can rebind to `Reference[T2]`. 

Perhaps there is a nicer way, but I am just not aware of that at the moment.

stdlib/test/collections/test_list.mojo
```diff
@@ -816,6 +816,18 @@ def test_list_contains():
...
+
+def test_list_eq_ne():
+    var l1 = List[Int](1, 2, 3)
+    var l2 = List[Int](1, 2, 3)
+    assert_true(l1 == l2)
+    assert_false(l1 != l2)
+
+    var l3 = List[Int](1, 2, 3, 4)
+    assert_false(l1 == l3)
+    assert_true(l1 != l3)
```
kszucs, line 828: Let me know what other use cases should I add here.

stdlib/src/collections/list.mojo
```diff
@@ -197,6 +197,60 @@ struct List[T: CollectionElement](
     # Operator dunders
     # ===-------------------------------------------------------------------===#
 
+    fn __eq__[
```
martinvuyk, line 200: Question:
I'm not very familiar with conditional conformance yet
```mojo
fn __eq__[T2: ComparableCollectionElement](self: List[T], other: List[T2])
```
can't they just be set to be the same type? or does it not work?
```mojo
fn __eq__[C: ComparableCollectionElement](self: List[C], other: List[C])
```

stdlib/src/collections/list.mojo
```diff
@@ -197,6 +197,60 @@ struct List[T: CollectionElement](
...
+        T2: ComparableCollectionElement
+    ](self: List[T], other: List[T2]) -> Bool:
+        """Checks if two lists are equal.
+
```
martinvuyk, line 204:
```suggestion

    Examples:
    
```

stdlib/src/collections/list.mojo
martinvuyk: does it work if you try this ?
```mojo
for i in range(len(self)):
    var left = rebind[T2](self.unsafe_get(i))
    var right = other.unsafe_get(i)
```

martinvuyk (commented):
Hi! Thanks for the PR

stdlib/src/collections/list.mojo
soraros: I think `fn __eq__[T: ComparableCollectionElement](self: List[T], other: List[T]) -> Bool` should work, and we don't need to use the rebind trick.

soraros: If we define `__eq__` as follows
```mojo
fn __eq__[T: C...](self: List[T], other: List[T]) -> Bool:
```
then `a.__eq__(b)` would work, while `a == b` would throw the following error
```
invalid call to `__eq__`: could not deduce parameter `T` of parent struct `List`
```

stdlib/src/collections/list.mojo
soraros: @martinvuyk This wouldn't work if we can't use the one-typed formation of conditional conformance because `rebind` only works on reg-types.

kszucs:
It doesn't look robust enough, the nested case is failing:

```mojo
    var nl1 = List(List(1, 2), List(3, 4))
    var nl2 = List(List(1, 2), List(3, 4))
    assert_true(nl1 == nl2)
```

```
/mojo/stdlib/test/collections/test_list.mojo:831:21: error: invalid call to '__eq__': could not deduce parameter 'T' of parent struct 'List'
    assert_true(nl1 == nl2)
                ~~~~^~~~~~
/mojo/stdlib/test/collections/test_list.mojo:1:1: note:  struct declared here
```

stdlib/src/collections/list.mojo
JoeLoser: I was actually exploring changing List and Dict iterator's `__next__` to return a `ref` yesterday.  We can't quite do that yet.  In the `List` case, it actually introduces extra copies.  Note that, independent of the `ref` vs `Reference` stuff, we may be able to remove some crazy rebinds that happen in `__eq__`, `__ne__`, and `__contains__` with conditional conformance - we should explore that.  In today's nightly release, you'll see I changed `List.__contain__` to remove the rebind and just be:

```
 fn __contains__[U: ComparableCollectionElement](self: List[U], value: U) -> Bool:
   for i in self:
       if i[] == value:
           return True
    return False
```

stdlib/src/collections/list.mojo
JoeLoser: @lattner is this something we can fix parameter inference wise so we can retain the nice sugar syntax of `==` without dropping back to `__eq__` manually?  I also ran into this yesterday when looking at removing some rebinds we were talking about.

stdlib/test/collections/test_list.mojo
JoeLoser: Can you please add some tests with:
1. An empty list
2. Lists with memory-only types (such as `String`)

soraros:
Yea, we don't have complete support for conditional conformance, so things don't compose nicely.

stdlib/src/collections/list.mojo
soraros: FYI, I raised #3199 for this.

stdlib/test/collections/test_list.mojo
kszucs: Going to add those cases.

stdlib/src/collections/list.mojo
kszucs: Thanks, going to rebase on it then.

stdlib/src/collections/list.mojo
JoeLoser: Using `==` instead of `__eq__` should work fine now since https://github.com/modularml/mojo/issues/3199 is closed now - see the discussion there.

stdlib/src/collections/list.mojo
kszucs: Updated along with using `EqualityComparableCollectionElement`, it is much nicer now, thanks!

stdlib/src/collections/list.mojo
kszucs: Updated.

stdlib/src/collections/list.mojo
kszucs: Now using `EqualityComparableCollectionElement`.

JoeLoser (approved):
Great work, thank you! Want to add a changelog entry for this work?

kszucs:
@JoeLoser added it to the changelog.

JoeLoser (approved):
!sync

docs/changelog.md
```diff
@@ -16,6 +16,9 @@ what we publish.
 
 ### ⭐️ New
 
+- `List[T]` values are now equality comparable with `==` and `!=` when `T` is
```
JoeLoser, line 19: Don't forget to credit yourself!  See the parts below in the changelog like

```
([PR 2910#](https://github.com/modularml/mojo/pull/2910) by
  [@msaelices](https://github.com/msaelices))
```

for example.

kszucs: One second.

kszucs: Done.

JoeLoser:
!sync

PR #2109, opened 2024-04-01
[mojo-stdlib] Improve ord and chr functions to accept any character (not only ASCII)

mzaks:
This can be seen as a first step towards https://github.com/modularml/mojo/issues/1616

PR #3196, opened 2024-07-09
[stdlib] Add is_valid_utf8 and test

martinvuyk:
Verify that the bytes are valid UTF-8.

This algorithm is good enough since it will be decently fast for languages with many ASCII characters, in case a faster algorithm is desired, there are projects like https://github.com/cyb70289/utf8 . Some benchmarking would be needed to assess the actual effects though.

Moved the `_utf8_byte_type` implementation from `builtin.string` to `utils.string_slice` and refactored it to be usable with any size of SIMD vector


stdlib/test/utils/test_string_slice.mojo
```diff
@@ -175,10 +176,97 @@ fn test_slice_bool() raises:
...
+    Lorem Ipsum段落的纸张，从而广泛普及了它的使用。最近，计算机桌面出版软件
+    למה אנו משתמשים בזה?
+    זוהי עובדה מבוססת שדעתו של הקורא תהיה מוסחת על ידי טקטס קריא כאשר הוא יביט בפריסתו. המטרה בשימוש
+     ב- Lorem Ipsum הוא שיש לו פחות או יותר תפוצה של אותיות, בניגוד למלל ' יסוי 
+    יסוי  יסוי', ונותן חזות קריאה יותר.הרבה הוצאות מחשבים ועורכי דפי אינטרנט משתמשים כיום ב-
+    Lorem Ipsum כטקסט ברירת המחדל שלהם, וחיפוש של 'lorem ipsum' יחשוף אתרים רבים בראשית
+    דרכם.גרסאות רבות נוצרו במהלך השנים, לעתים בשגגה 
+    Lorem Ipsum е едноставен модел на текст кој се користел во печатарската 
+    индустрија.
+    Lorem Ipsum - це текст-"риба", що використовується в друкарстві та дизайні.
```
laszlokindrat, line 199: 🇺🇦

stdlib/test/utils/test_string_slice.mojo
```diff
@@ -175,10 +176,97 @@ fn test_slice_bool() raises:
...
+        List[UInt8](0x00, 0x00, 0x00, 0xF1, 0x80, 0x00),
+        List[UInt8](0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xF1),
+        List[UInt8](0x00, 0x00, 0x00, 0x00, 0xF1, 0x00, 0x80, 0x80),
+        List[UInt8](0x00, 0x00, 0xF1, 0x80, 0xC2, 0x80, 0x00),
+        List[UInt8](0x00, 0x00, 0xF0, 0x80, 0x80, 0x80),
+    )
+    for item in negative:
+        assert_false(is_valid_utf8[8](item[].unsafe_ptr(), len(item[])))
+        assert_false(is_valid_utf8[4](item[].unsafe_ptr(), len(item[])))
+    _ = negative
```
laszlokindrat, line 262: Is there an issue with lifetimes here that makes this necessary?

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,129 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
 """An immutable static string slice."""
 
 
+fn is_valid_utf8[
+    width: Int = 8
```
laszlokindrat, line 30: Q: did you do any benchmarking to determine that this should be 8 and not 4 or 16, or is this just an educated guess?

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,129 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
 """An immutable static string slice."""
 
 
+fn is_valid_utf8[
```
laszlokindrat, line 29: Q: Where do you plan to use this? I'm asking because I want to be careful with adding public APIs, and if this is going to be mostly a string implementation detail, I'd rather call it `_is_valid_utf8` (especially since its signature is very unsafe).

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,129 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
...
+            return True
+        elif b0 == 0xED and not (UInt8(0x80) <= b1 <= UInt8(0x9F)):
+            return True
+        elif b0 == 0xF0 and not (UInt8(0x90) <= b1 <= UInt8(0xBF)):
+            return True
+        elif b0 == 0xF4 and not (UInt8(0x80) <= b1 <= UInt8(0x8F)):
+            return True
+        return False
+
+    var idx = 0
```
laszlokindrat, line 84: Could you please provide some comments explaining this algorithm? Also, consider better variable names, e.g. `n2_m` below is not very meaningful.

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,129 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
...
+
+        if invalid_special(d[0], d[1]):
+            return False
+        elif first_byte_type == 2 and d[0] < UInt8(0b1100_0010):
+            return False
+        idx += int(first_byte_type)
+        iter_len -= int(first_byte_type)
+
+    @parameter
+    fn invalid[amnt: Int](i: Int) -> Bool:
```
laszlokindrat, line 123: Q: What is `amnt`?

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,129 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
...
+        @parameter
+        for j in range(1, amnt + 1):
+            if countl_zero(~(ptr[i + j] & UInt8(0b1111_0000))) != 1:
+                return True
+        return invalid_special(ptr[i], ptr[i + 1])
+
+    idx = length - iter_len
+    while idx < length:
+        var val = ptr[idx]
+        var byte_type = countl_zero(~(val & UInt8(0b1111_0000)))
```
laszlokindrat, line 136: Can you use the `_utf8_byte_type` helper?

laszlokindrat (commented):
Thanks! This could definitely be useful! Could you please share some of your motivation for introducing this? What are you planning to build on top of it?

stdlib/test/utils/test_string_slice.mojo
martinvuyk: dunno just in case lol, this has bit me so many times

stdlib/src/utils/string_slice.mojo
martinvuyk: Educated guess since: 
- the bigger the width the more bytes can be discarded at once when they are all ASCII
- 8x8 = 64 bits which fits into any SIMD engine. 
- I thought of adding 16 but I think I read somewhere that most small strings fall into 10 - 24 bytes long (might be wrong though and could make this bigger, will google some more later)

stdlib/src/utils/string_slice.mojo
martinvuyk: I still need to use the countl_zero directly since it can operate on SIMD and not only single byte

stdlib/src/utils/string_slice.mojo
martinvuyk: amnt of continuation bytes to check, I'll change to a better name

stdlib/src/utils/string_slice.mojo
laszlokindrat: ok, thanks!

stdlib/test/utils/test_string_slice.mojo
laszlokindrat: Could you please remove these, and see?

martinvuyk:
Brought the `_utf8_byte_type` helper into string_slice.mojo which might break other PRs, and also refactored it to be usable with any size of SIMD. 

Refactored the while loop to be able to compute the remainder as well as simd width and hardcoded width combinations depending on the target architecture. Also changed variable names and left a couple of comments on what each section does.

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,155 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
...
+                return 0
+        var byte_types = _utf8_byte_type(d)
+        var first_byte_type = byte_types[0]
+        if first_byte_type == 0:
+            for i in range(1, width):
+                if byte_types[i] != 0:
+                    idx += i
+                    iter_len -= i
+                    break
+            continue
```
laszlokindrat, line 85: This seems to be doing some extra work, and it's also a bit confusing. Why not use `comp` (which should probably be renamed to `is_ascii` or something) to determine the first non-ascii byte, and jump there accordingly?

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,155 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
...
+
+        # byte_type has to match against the amount of continuation bytes
+        alias vec_t = SIMD[DType.uint8, 4]
```
laszlokindrat, line 88: style nit: PascalCase for type parameters: https://github.com/modularml/mojo/blob/nightly/stdlib/docs/style-guide.md#identifier-naming-conventions
```suggestion
        alias VecType = SIMD[DType.uint8, 4]
```

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,155 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
...
+    U+0000..U+007F     | 00..7F     |             |            |             |
+    U+0080..U+07FF     | C2..DF     | 80..BF      |            |             |
+    U+0800..U+0FFF     | E0         | ***A0***..BF| 80..BF     |             |
+    U+1000..U+CFFF     | E1..EC     | 80..BF      | 80..BF     |             |
+    U+D000..U+D7FF     | ED         | 80..***9F***| 80..BF     |             |
+    U+E000..U+FFFF     | EE..EF     | 80..BF      | 80..BF     |             |
+    U+10000..U+3FFFF   | F0         | ***90***..BF| 80..BF     | 80..BF      |
+    U+40000..U+FFFFF   | F1..F3     | 80..BF      | 80..BF     | 80..BF      |
+    U+100000..U+10FFFF | F4         | 80..***8F***| 80..BF     | 80..BF      |
+    .
```
laszlokindrat, line 150: nit: Please move this before the `Args:`.

laszlokindrat (approved):
Nice, thank you! I have a couple requests mostly related to making the code more readable, but otherwise it looks great!

stdlib/src/utils/string_slice.mojo
```diff
@@ -26,6 +26,155 @@ alias StaticString = StringSlice[ImmutableStaticLifetime]
 """An immutable static string slice."""
 
 
+fn _utf8_byte_type[
+    size: Int
+](b: SIMD[DType.uint8, size]) -> SIMD[DType.uint8, size]:
```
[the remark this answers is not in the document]
martinvuyk: Niice :+1:

stdlib/src/utils/string_slice.mojo
martinvuyk: You're totally right, will do

soraros:
Question: is this somehow related to the algorithm used in `ord` and `chr` (#2109)?

soraros:
> We could also add tests for `ord()` and `chr()` with this function (I have a strong suspicion that `chr()` is not taking all unicode special ranges into account).

I should maybe write a proof for `chr` as well like I did [here](https://github.com/modularml/mojo/pull/2109#issuecomment-2031021790).

martinvuyk:
> I should maybe write a proof for `chr` as well like I did [here](https://github.com/modularml/mojo/pull/2109#issuecomment-2031021790).

That z3 library looks pretty cool. But the `chr` function should have debug_assert (or maybe even raise) with _is_valid_utf8 since I don't think it currently checks unicode valid ranges. And a proof like you did only checks if the values were bitshifted correctly (I think)

See the exceptions in the second byte in bold:

[Table 3-7 page 94](http://www.unicode.org/versions/Unicode6.0.0/ch03.pdf).
Well-Formed UTF-8 Byte Sequences

Code Points        | First Byte | Second Byte | Third Byte | Fourth Byte |
:----------        | :--------- | :---------- | :--------- | :---------- |
U+0000..U+007F     | 00..7F     |             |            |             |
U+0080..U+07FF     | C2..DF     | 80..BF      |            |             |
U+0800..U+0FFF     | E0         | ***A0***..BF| 80..BF     |             |
U+1000..U+CFFF     | E1..EC     | 80..BF      | 80..BF     |             |
U+D000..U+D7FF     | ED         | 80..***9F***| 80..BF     |             |
U+E000..U+FFFF     | EE..EF     | 80..BF      | 80..BF     |             |
U+10000..U+3FFFF   | F0         | ***90***..BF| 80..BF     | 80..BF      |
U+40000..U+FFFFF   | F1..F3     | 80..BF      | 80..BF     | 80..BF      |
U+100000..U+10FFFF | F4         | 80..***8F***| 80..BF     | 80..BF      |

stdlib/src/utils/string_slice.mojo
laszlokindrat: I see, I never used a footnote in a docstring, I will see if it breaks anything on our end.

laszlokindrat:
!sync

Issue #328, opened 2023-06-05
[BUG] Mojo `object` integer representation is not arbitrary precision


MeioNeiz:
### Bug description

```shell
def factorial(n):
    if n == 0 or n == 1:
        return 1
    else:
        return n * factorial(n-1)

print(factorial(20)) # Expect 2432902008176640000
2432902008176640000

print(factorial(21)) # Expect 51090942171709440000
-4249290049419214848

# After 21 the results are incorecct both positive and negative integers

print(factorial(65))
-9223372036854775808

print(factorial(66))
0
```

### Steps to reproduce

```shell

def factorial(n):
    if n == 0 or n == 1:
        return 1
    else:
        return n * factorial(n-1)

print(factorial(20)) # Expect 2432902008176640000
2432902008176640000

print(factorial(21)) # Expect 51090942171709440000
-4249290049419214848
```

### System information

```shell
11:hugetlb:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod9d85d3cb_0174_44b8_b579_f639d1b70891.slice/cri-containerd-7b611459c9a2b310ad42d2f516052b8f7ea3da1ac7e0c199fb9c68c8e178738b.scope
10:memory:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod9d85d3cb_0174_44b8_b579_f639d1b70891.slice/cri-containerd-7b611459c9a2b310ad42d2f516052b8f7ea3da1ac7e0c199fb9c68c8e178738b.scope
9:cpu,cpuacct:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod9d85d3cb_0174_44b8_b579_f639d1b70891.slice/cri-containerd-7b611459c9a2b310ad42d2f516052b8f7ea3da1ac7e0c199fb9c68c8e178738b.scope
8:perf_event:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod9d85d3cb_0174_44b8_b579_f639d1b70891.slice/cri-containerd-7b611459c9a2b310ad42d2f516052b8f7ea3da1ac7e0c199fb9c68c8e178738b.scope
7:devices:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod9d85d3cb_0174_44b8_b579_f639d1b70891.slice/cri-containerd-7b611459c9a2b310ad42d2f516052b8f7ea3da1ac7e0c199fb9c68c8e178738b.scope
[...]

PR #3202, opened 2024-07-09
[Proposition] Add an arbitrary precision integer type to the stdlib

oskgo:
We propose standardizing an arbitrary precision runtime signed integer type, `BigInt`, with the aim of using this to build an arbitrary precision integer variant of `object`.

The structure of the proposition is modeled after the Rust RFC template, which I find quite good.

cc #328

Signed-off-by: Oskar Goldhahn <oskar@goldhahn.info>

### DIFF
--- a/proposals/arbitrary-precision-int.md
+++ b/proposals/arbitrary-precision-int.md
@@ -1,0 +1,30 @@
+# Summary
+
+We propose standardizing an arbitrary precision runtime signed integer type, `BigInt`, with the aim of using this to build an arbitrary precision integer variant of `object`.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


laszlokindrat:
Thank you for your proposal! We discussed this with the team, and decided that at this point we would like to postpone work related to arbitrary precision integer arithmetic. While I agree that https://github.com/modularml/mojo/issues/328 is annoying from the Python compatibility perspective, it points beyond arbitrary precision integer arithmetic and the relationship between `object`, `PythonObject`, `def`, etc. are evolving quickly right now.

A multiprecision library in Mojo sounds really interesting, and it's definitely something we expect in the future, but it is too early for the language, especially given that using Python interop is a viable alternative that exists today. At the same time, we would love to see an external project for BigInt (and would be happy to showcase it in our newsletter), or maybe a wrapper library for GMP.

PR #3203, opened 2024-07-09
[stdlib] Simplify several methods in the `UnsafePointer` struct:

soraros:
- Change the `offset` method to take `Int` instead of `Intable`. `UnsafePointer` is itself `Intable`, and being able to offset a pointer by another pointer is undesirable.
- Clean up `initialize_pointee_explicit_copy` by utilizing the latest conditional conformance support.
- Change the parameter name for `bitcast` to improve consistency.
- Modify several methods to infer parameters automatically; relying on the default parameter instead of passing it manually.

### DIFF
--- a/stdlib/src/memory/unsafe_pointer.mojo
+++ b/stdlib/src/memory/unsafe_pointer.mojo
@@ -200,11 +200,8 @@ struct UnsafePointer[
         )
 
     @always_inline
-    fn offset[T: Intable](self, idx: T) -> Self:
+    fn offset(self, idx: Int) -> Self:
         """Returns a new pointer shifted by the specified offset.
-
-        Parameters:
-            T: The Intable type of the offset.
 
         Args:
             idx: The offset of the new pointer.
@@ -212,7 +209,7 @@ struct UnsafePointer[
         Returns:
             The new constructed DTypePointer.
         """
-        return __mlir_op.`pop.offset`(self.address, int(idx).value)
+        return __mlir_op.`pop.offset`(self.address, idx.value)
 
     @always_inline
     fn __getitem__(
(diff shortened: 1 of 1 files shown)


stdlib/src/memory/unsafe_pointer.mojo
```diff
@@ -250,7 +247,7 @@ struct UnsafePointer[
         Returns:
             An offset pointer.
         """
-        return self + (-offset)
+        return self.offset(-offset)
```
ConnorGray, line 250: **Question** Does it work if we keep this line as `self + (-offset)`?

I understand this change is a bit of a simplification because it avoids an extra "hop" through `__add__` before going to `offset`. But for style and consistency reasons, representing pointer offsetting using `+` is preferred.

ConnorGray:
!sync

stdlib/src/memory/unsafe_pointer.mojo
soraros: Yea, it works. I simply didn't know that `__add__` is preferred. Will revert this change.

ConnorGray: No worries at all :)

(The fact that `UnsafePointer.offset` exists at the moment is a slight temporary state; it will likely get removed in favor of having only one way (`UnsafePointer.__add__`) to do offsetting.)

ConnorGray:
!sync

stdlib/src/memory/unsafe_pointer.mojo
soraros: Just a thought: why don't we simply remove `__add__` and `__sub__`, and only leave `offset` which is easier in method chaining (`(p + n)[]` looks bad)?

Issue #1134, opened 2023-10-21
[Feature Request] Add mmap module


tairov:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Implement `mmap` module natively in Mojo 
link: https://man7.org/linux/man-pages/man2/mmap.2.html

### What is your motivation for this change?

When implementing multiple solutions on Mojo, we reached a stage where huge files needed to be loaded into memory (for example, open source llm models).

For these purposes, it would be really great to have the mmap module implemented natively in Mojo.

jackos:
@abduld fyi 

Relevant blog post: https://justine.lol/mmap/

tairov:
> @abduld fyi
> 
> Relevant blog post: https://justine.lol/mmap/

Epic story how they get it working in C++, I didn't know such a simple and obvious thing might be relevantly complicated to implement within cpp/stl. 
For example on C , the mmap is trivial -- https://github.com/karpathy/llama2.c/pull/50/files 
On Mojo it should be straightforward I hope, since we have full control on pointers. Currently I'm doing "pseude-mmap", when weights are loaded into memory one time (file file IO), and then I'm carefully set up pointers.

ihnorton:
Calling mmap through a function pointer (can't use `external_call` yet, AFAICT) seems to crash: https://gist.github.com/ihnorton/a49c8165f86489def9031583f3ed1b20

JoeLoser:
Thanks for the links and discussion here, everyone.  I should have a chance to look into this at the end of this week.

tairov:
> Calling mmap through a function pointer (can't use `external_call` yet, AFAICT) seems to crash: https://gist.github.com/ihnorton/a49c8165f86489def9031583f3ed1b20

What a pity.. I was betting high on this

deepankarsharma:
Looks like I was able to get mmap to work based on @ihnorton excellent post above in https://github.com/modularml/mojo/issues/1134#issuecomment-1811770296

`from sys import ffi
from memory import unsafe

struct MapOpt:
    alias MAP_SHARED = 0x01
    alias MAP_PRIVATE = 0x02

struct Prot:
    alias PROT_NONE = 0x0
    alias PROT_READ = 0x1
    alias PROT_WRITE = 0x2
    alias PROT_EXEC = 0x4

alias c_void = UInt8
[...]

PR #3218, opened 2024-07-11
[stdlib] Proposal for mmap

KCaverly:
A few people have been discussing adding a `mmap` module directly to the stdlib.
This has been welcomed by the core team @JoeLoser, in #1134 

This is an initial proposal to introduce such a module, for working with memory mapped files in Mojo.
It has taken inspiration from [memmap2-rs](https://github.com/RazrFalcon/memmap2-rs) and the [mmap module in Python](https://docs.python.org/3/library/mmap.html).

A brief proof of concept is provided, with a working example, of how we can construct a mmap object, read bytes from it, and close it. 

This is my first time contributing to Mojo, so please let me know if there is anything I can improve.
If we are comfortable with the proposal, I would be happy to drive implementation and port some of my existing work over to the stdlib.

### DIFF
--- a/proposals/mmap_proposal.md
+++ b/proposals/mmap_proposal.md
@@ -23,88 +23,88 @@ A `Mmap` object, containing the functionality for construction/destruction and i
 An initial implementation of `Mmap` in a read-only context can be used as follows. Note, the API has been meant to provide similarity with the existing `FileHandle` API.
 
 ```mojo
-var length: Int32 = 24;
-var fd: FileDescriptor = open("./test.txt", "r");
+    var length: Int32 = 24;
+    var fd: FileDescriptor = open("./test.txt", "r");
 
-# Initialize a mmap object for the particular file, with read and shared
-# permissions
-var mmap = Mmap(fd=fd, length=length, offset=0, mode=MmapMode.READ)
+    # Initialize a mmap object for the particular file, with read and shared
+    # permissions
+    var mmap = Mmap(fd=fd, length=length, offset=0, mode=MmapMode.READ)
 
-# Print out the entire file as a String
-print(String(mmap.read_bytes()))
+    # Print out the entire file as a String
+    print(String(mmap.read_bytes()))
 
-# Memory mapped files must be closed once they are finished
-mmap.close()
+    # Memory mapped files must be closed once they are finished
+    mmap.close()
 ```
 
 Whereas an initial implementation of `Mmap` in a read/write context can be used as follows:
 
 ```mojo
-var length: Int32 = 24;
-var fd: FileDescriptor = open("./test.txt", "r+")
+    var length: Int32 = 24;
+    var fd: FileDescriptor = open("./test.txt", "r+")
 
-# Initialize a mmap object for the particular file, with read/write and shared permissions
-var mmap = Mmap(fd=fd, length=length, offset=0, mode=MmapMode.WRITE)
+    # Initialize a mmap object for the particular file, with read/write and shared permissions
+    var mmap = Mmap(fd=fd, length=length, offset=0, mode=MmapMode.WRITE)
 
-# Print out the entire file as a String
-print(String(mmap.read_bytes()))
+    # Print out the entire file as a String
+    print(String(mmap.read_bytes()))
 
-# Write bytes to mmap
-mmap.write("hello world")
+    # Write bytes to mmap
+    mmap.write("hello world")
 
-# Memory mapped files must be closed once they are finished
-mmap.close()
+    # Memory mapped files must be closed once they are finished
+    mmap.close()
 ```
 ### Initial API
 
 An initial API for `MmapMode` and `Mmap` could look like the below. 
 
 ```mojo
-MmapMode:
-	alias READ: String = "READ"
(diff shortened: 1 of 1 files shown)


JoeLoser:
Thanks for the proposal!  As you know, we're +1 for introducing mmap things to the library.  We'll review it in detail at the team's design discussion this coming Monday and respond to the PR with our notes from the discussion and go from there.

martinvuyk:
Hi! 
I'm not very familiar with Python's or any other language's memory mapping implementations, so question: 
can't we fold this to being a specialization of `FileHandle` ? 
You can see for example in PR #3257 I tried setting up a `FileHandle` iterator that returns `StringSlice`s split by newlines, but it needs memory mapping to work efficiently.

Theoretically we could offer simply `open[memory_mapped: Bool = False]()` parameter and hide this complexity from the user (we could also offer an API to get the underlying pointer with conditional conformance). Another thing is that we wouldn't be duplicating APIs from `FileHandle` for `Mmap`

proposals/mmap_proposal.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Initial API
+
+An initial API for `MmapMode` and `Mmap` could look like the below.
+
+```mojo
+    struct MmapMode:
+        alias READ: String = "READ"
+        alias WRITE: String = "WRITE"
+        alias EXEC: String = "EXEC"
```
ConnorGray, line 80: **Question** Would it make sense to add a `MmapMode.NONE` constant here as well, so callers can specify memory regions that should not be touched? (Useful for things like debugging techniques.)

ConnorGray, line 80: **Suggestions** Could these constants be plain numeric literals instead? That would be more in keeping with how we model `enum`-like values elsewhere in the stdlib, see for example how the `DType` type and named values are defined.

proposals/mmap_proposal.md
```diff
@@ -0,0 +1,342 @@
...
+    fn _get_prot(mode: String) raises -> Int32:
+        """Helper method to translate between `MmapMode` and OS accepted prot values"""
+        ...
+
+    @staticmethod
+    fn _get_flags(mode: String) raises -> Int32:
+        """Helper method to translate between `MmapMode` and OS accepted flags values"""
+        ...
+
+    fn __init__(inout self, fd: FileDescriptor, length: Int32, offset: Int32, mode: String, address: UnsafePointer[UInt8]) raises:
```
ConnorGray, line 98: **Suggestion** Would it make sense (here and below) for this to be `length: UInt32`, since negative lengths semantically are not valid?

proposals/mmap_proposal.md
```diff
@@ -0,0 +1,342 @@
...
+        """Given a FileDescriptor, an expected length and offset, mmap configuration, and an initial address, initialize a `mmap` object"""
+        ...
+
+    fn __init__(inout self, length: Int32, offset: Int32, mode: String, address: UnsafePointer[UInt8]) raises:
```
ConnorGray, line 102:
```suggestion
    fn __init__(inout self, length: Int32, offset: Int32, mode: MmapMode, address: UnsafePointer[UInt8]) raises:
```

**Suggestion** Since we have the `MmapMode` struct, should we use it here (and elsewhere) instead of having this argument be "stringly typed"? 🙂

ConnorGray (approved):
Thanks for this 🔥  contribution @KCaverly! 😄

The stdlib team had a chance to review this proposal earlier this week, and I'm sharing some initial feedback 🙂 

I left a few minor comments we had about the proposed design of the API. I know though this is just an initial proposal, so I'd expect further feedback would follow once an implementation PR is ready for review.

Please let me know if you have any questions or concerns about our feedback!

We're excited to get `mmap` support into the standard library, and it's awesome that you're enthusiastic to work on this 😄

proposals/mmap_proposal.md
JoeLoser: We also discussed needing these to be numerics since they're bit flags in the `mmap` POSIX call:

> It is either PROT_NONE or the bitwise OR of one or more of the
       following flags:

from https://man7.org/linux/man-pages/man2/mmap.2.html.

jgsimard:
@JoeLoser any updates on this ?

JoeLoser:
> @JoeLoser any updates on this ?

Hi @jgsimard - we would still welcome someone to pick up this proposal, respond to our feedback, and then review it again. If you're open to implementing this, even better! Given that folks can just use external_call to call mmap and friends right now, this still is not close to the highest priorities for the team at this time.

martinvuyk:
> > @JoeLoser any updates on this ?
> 
> Hi @jgsimard - we would still welcome someone to pick up this proposal, respond to our feedback, and then review it again. If you're open to implementing this, even better! Given that folks can just use external_call to call mmap and friends right now, this still is not close to the highest priorities for the team at this time.

@JoeLoser I think that if we implement @owenhilyard 's IO Engine approach we won't need to implement this module separately? i.e. if we abstract away that opening a file can mean opening it mapped in memory or using a cursor, by using a unified interface, I think this complexity all boils away

Issue #3148, opened 2024-06-29
[Feature Request] Make print default to `__repr__` not `format_to`


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

I didn't find any issue or PR regarding [this commit](https://github.com/modularml/mojo/commit/4eed372586ddce66b1610f2c4b6864a97e7125ce) by @ConnorGray so i thought I'd open one for discussion.

I do like adding a better performance alternative, but I think `print()` should default to using the struct's `__repr__` method just like Python since most developers are accustomed to just build that method and go for `print(object)`. I don't know how we would deal with two overloads when a struct implements both `__repr__` and `format_to` though (Variant type?).

### What is your motivation for this change?

`print()` requiring `format_to` breaks away from how Python works by default.

### Any other details?

An idea would be to slightly change `__repr__(self) -> String` signature to `__repr__(self, inout writer: Formatter)` or somehow accept both signatures and use Formatter when available

PR #3224, opened 2024-07-11
[stdlib] Simplify and pythonify `Formattable`

gabrieldemarmiesse:
### Goals

The goal of this PR (and the future ones) is to simplify the concepts and usage of `Formatter`. The `Formatter` concept is completely new to python users and Mojo users, and could be expressed by composing existing concepts. 

Basically, using `Formatter` is a way to write an utf8 representation of an object to a pointer. We can also express its use in this way, by declaring the following functions and traits (half of them already exist)

```mojo
fn str[T: Stringable](value: T) -> String: ...
fn str[T: Formattable](value: T, *, inout writer: Formatter): ...

fn repr[T: Representable](value: T) -> String: ...
fn repr[T: RepresentableToBuffer](value: T, *, inout writer: Formatter): ...

trait Stringable:
    fn __str__(self) -> String: ...
trait StringableToBuffer:
    fn __str__(self, *, inout writer: Formatter): ....

trait Representable:
    fn __repr__(self) -> String: ....
trait RepresentableToBuffer:
    fn __repr__(self, *, inout writer: Formatter): ....
```

This reduces the number of new functions and methods users need to learn.

### Next steps

The next big step, would be to replace `Formatter` by a "file-like" trait. Python users know file-like objects and what they are used for. They will understand the advantages of writing progressively to a buffer compared to concatenating many possibly big strings. It's not uncommon for Python users to use [`StringIO`](https://docs.python.org/3/library/io.html#io.StringIO) to write to a string when using functions accepting file-like objects.

With this trait for the methods described above, we could easily write to a string with a `StringIO` struct, or write to a file, or a socket, or whatever we want.

Much further down the line, to avoid forcing users to define two `__str__` and two `__repr__` methods every time, we could easily autogenerate the missing methods.

### Precedent

There is precedent here, as the function `print` in python can take a `file` keyword argument to store the result. We could follow this convention. [Reference](https://docs.python.org/3/library/functions.html#print)

Having two different functions (one to return a string and one to output to a file) is common in python. See `json.dump()` which writes to a file and `json.dumps()` which returns a string. [Reference](https://docs.python.org/3/library/json.html#json.dump)

### Conclusion

I believe those API changes can make Python and Mojo users feel right at home and will make those improvements as intuitive and easy to explain as [the ones made to hashing](https://github.com/modularml/mojo/blob/main/proposals/improved-hash-module.md).

@ConnorGray since we already talked about this offline.


soraros:
I think, conceptually, `Hasher` and `Formatter` are similar, and they are new to Python users in the exact same way. `Representable` is like `Hashable` with a different default hasher comparing with `Stringable`; different output channel is like different hash state/type. If we can come up with an approach to ease Python users into hasher based `Hashable`, we could do the same for `Formattable`.

yinonburgansky:
`str` and `repr` are usually pure functions without side effects, unlike `print` which has side effects, 
I think using:
```python
writer.write(something)
writer.write(something, repr=True)  #  or writer.write_repr(something)
writer.write(number, use_scientific_notation=True)  # support parametric kwargs when available
```
makes more sense than:
```python
str(something, writer=writer)
repr(something, writer=writer)
```
I think it's ok for dunder methods, since usually they are not be called manually.
This will also reduce memory allocations, e.g. in List repr.

gabrieldemarmiesse:
That's great feedback thanks! I worry that the proposed approach is not user-extensible. 

Let's say w want, in an external library, to add `__jsonstr__()` method to specify how object should be serialized to json, we cannot change the standard library and the file-like trait to add `writer.write(something, json_str=True)`. But it's possible in our library to create a function `json_str(obj, writer=...)` and `json_str(obj) -> String`, which would then call `obj.__jsonstr__()`.

`pprint(something, underscore_numbers=True)` has a special case for `int` and `float` and I don't think we want that. If a user creates a custom `Number` struct, this won't work.

We probably don't want to bikeshed too much here, but this is the main reason `str` is completely independant from `repr` in this implementation. It gives room to other future methods in user-defined code and doesn't special case existing structs in the stdlib, which is one of the stated goals of Mojo (anything can be redefined by a user).

yinonburgansky:
> I worry that the proposed approach is not user-extensible.

In python `**kwargs` allow you to pass any keyword arguments without knowing them ahead of time
```python
class Formatter:
    def write(self, object, **kwargs):
        object.format_to(self, **kwargs)

class LibObject:
    def format_to(writer, *, json_str=False, **_kwargs):
        if json_str:
            ...
        else
            ...
```
I think in mojo with parametric kwargs this can be done at compile time, avoiding a branch.

> We probably don't want to bikeshed too much here.

Sorry, I won't pursue this any further; I just wanted to address your point.

gabrieldemarmiesse:
> In python **kwargs allow you to pass any keyword arguments without knowing them ahead of time

That's a good point. It's not possible to call a function and passing `**kwargs` at runtime in Mojo yet. It's only allowed in the signature currently.

I'm totally open to revise this API in the future when Mojo supports `**kwargs` in function bodies :)

remorses:
I would prefer removing the concept of Formattable altogether and only keep Stringable, the formatter can be used as an internal concept to make writing `__str__` easier. 

If Formattable is needed to have a "lazy" version of `__str__` that does not buffer the string then you could later add a `LazyStringable` that adds generator support to `__str__`

```mojo
@value
struct Point(LazyStringable):
    var x: Int
    var y: Int

    fn __str__(self) -> Generator[StringRef]:
        yield "Point("
        yield self.x
        yield ", "
        yield self.y
        yield ")"
```

But currently you could easily remove the Formattable concept from the std and it would make things more clear.

martinvuyk:
> I would prefer removing the concept of Formattable altogether and only keep Stringable

+1 on finding a way to fold the concept of Formattable into another one, but I think it should be Representable. The Generator approach is very interesting though not achievable with current Mojo features :(

I found the Formatter concept to be quite confusing, as far as I've understood the code it is just a roundabout way for asking a pointer to a string representation of an object, and concatenating it in a more efficient manner.  

I like this approach :+1: (it's what I proposed in #3148 as well :D )
```mojo
trait Representable:
    fn __repr__(self) -> String: ....
trait RepresentableToBuffer:
    fn __repr__(self, *, inout writer: Formatter): ....
```

And since json was brought up, I think the interface `fn __repr__(self, *, inout writer: Formatter): ....` is very easily extensible for how you want to controll that representation in case you want an object to be represented as json. 

Another benefit is that binary serialization protocols could also (I think) be integrated into this interface in the future for sending to a file descriptor directly (not sure how exactly, but I'm sure it's achievable).

mzaks:
Reading through the comments I have a feeling that there is a lot of confusion going on in the community :)

I would like to try clarify some things.

The main point is about call-return semantics vs. data-flow semantics.

A method like `__hash__`, `__str__` or `__repr__` and even `__eq__` can be implemented with call-return or data-flow semantics in mind.

What's the difference?

With call-return, as the name suggested, the method returns the result. In case of `__str__` and `__repr__` it generates a sequence of bytes and returns it as a String (a data structure which without small string optimisation, allocates memory on the heap). It doesn't seem to be a problem at the first glance as, you might think, we need that string anyways, but think about what it means if you have a type which composes multiple other types. For example:

```Mojo
struct Person:
  var name: String
  var birthday: Date
  var gender: Gender
  var address: PostalAddress
  var friends: List[Reference[Person]]

  fn __str__(self) -> String:
    return "name: " + self.name.__str__() + ", " + 
        "birthday: " + self.birthday.__str__() + ", " + 
        "gender: " + selg.gender.__str__() + ", " + 
        "address: " + selg.address.__str__() + ", " + 
        "friends: " + len(self.freinds).__str__() 
```

`Person` contains of fields of type `String`, `Date`, `Gender`, `PostalAddress` and sorry for being evil a `List` of Person reference (I know it is currently not a valid Mojo, but it should be possible to do something similar in the future). In order to create a string for a person we want to call the `__str__` method on the other types, which means all the other types generate transient string instances which have a single purpose in life, to be copied into a bigger string which represents the final person string. So we have a cascade of heap allocations and mem copies.

Now here comes the data-flow. With dataflow we do not return a value, but we employ something similar to visitor pattern. The method accepts an instance which "collects" the result.

```Mojo
struct Person:
  var name: String
  var birthday: Date
  var gender: Gender
  var address: PostalAddress
  var friends: List[Reference[Person]]

  fn __str__(self, *, inout writer: Formatter):
    "name: ".__str__(writer)
    ", ".__str__(writer)
    self.name.__str__(writer)
    "birthday: ".__str__(writer)
    self.birthday.__str__(writer)
    ", ".__str__(writer)
    "gender: ".__str__(writer)
    self.gender.__str__(writer)
    ", ".__str__(writer)
    "friends: ".__str__(writer)
```

With this approach we avoid cascading allocations and copies, only values which need to be converted to a string are converted and directly stored in the final buffer. We can go even one step further, by generalising the `Formatter`. As creators of the Person struct we don't care if the Formatter will write the bytes stream we provide into memory, standard out, file or socket. The Formatter is a visitor, we give it data! A user of our struct should also not use the `__str__` method directly, but rather through a function like `str(person)`, `print(person)`, `file(person)`, `socket(person)`. Those functions will instantiate the formatter they need, in case of `str` it is a formatter which writes into a buffer, in case of `print` the formatter writes to standard out, etc... 

PS: The above implementation of `fn __str__(self, *, inout writer: Formatter):` is verbose but IMHO simpl to understand how things should work. It can be in a less verbose way by calling `writer.write("name: ", self.name, ", " , "birthday: ", self.birthday, ...)` or even employing some compiler magic with f-strings: `writer.write(f"name: {self.name}, birthday: {self.birthday} ...")`

Overall I think even though the concept is new to Python developers the API of dunder methods which employ data-flow semantics will dictate the solution, hence I think keeping both, call-return and data-flow functions is rather confusing.

remorses:
@mzaks why can't `__str__` return a StringSlice instead of String? This way you wouldn't need to copy and allocate the strings again, right?

mzaks:
> @mzaks why can't `__str__` return a StringSlice instead of String? This way you wouldn't need to copy and allocate the strings again, right? I guess you would need to pass the lifetime manually when calling `__str__` but it should not be a problem because you usually would use `str` instead

What would the `StringSlice` be for a `PostalAddress` type? It is a combination of its fields with some "templating" (e.g. street, city, zip code, country, etc...). So it is a new string which combines other strings if you are working with call-return. With data-flow you can feed the writer with slices inside the `PostalAddress` and avoid creating transient strings.

remorses:
@mzaks you are right, it would be cool to use generators for this use case in the future

JoeLoser:
Thanks for opening this for discussion, @gabrieldemarmiesse.  Given @jackos's recent work on changing the `Formattable` trait, I'm going to close this PR, but feel free to propose any changes to the `Writable` traits and we can discuss and iterate on those.

PR #3228, opened 2024-07-12
[stdlib] Move `find()` from `StringRef` to `StringSlice`

gabrieldemarmiesse:
We are just moving the StringRef implementation of `find`. We'll delete `StringRef.find()` later on. The base implementation is here: https://github.com/modularml/mojo/blob/nightly/stdlib/src/utils/stringref.mojo#L497

`String.find()` is now plugged to `StringSlice.find()`

In general we are trying to move away from `StringRef` and at some point maybe delete it in favor of `StringSlice` which is safer, both in terms of memory and in terms of utf-8 validity.


stdlib/src/utils/stringref.mojo
```diff
@@ -494,6 +494,8 @@ struct StringRef(
 
         return res
 
+    # TODO: remove this method later on when nothing depends on it anymore.
+    # it has already been transferred to `StringSlice`
```
gabrieldemarmiesse, line 498:
```suggestion
    # It has already been copied to `StringSlice`.
```

stdlib/src/utils/stringref.mojo
```diff
@@ -494,6 +494,8 @@ struct StringRef(
 
         return res
 
+    # TODO: remove this method later on when nothing depends on it anymore.
+    # It has already been copied to `StringSlice`.
     fn find(self, substr: StringRef, start: Int = 0) -> Int:
```
ConnorGray, line 499: **Suggestion** Instead of leaving this definition behind, could we go one step farther, and have this definition call into `StringSlice.find()`?

I think that would look something like:

```mojo
# Methods on struct StringRef
fn find(self, substr: StringRef, start: Int = 0) -> Int:
    var self_str = self.unsafe_as_string_slice()
    var substr_str = substr.unsafe_as_string_slice()

    return self_str.find(substr, start)
```

where `StringRef.unsafe_as_string_slice()` would be a new method with the ~definition:

```mojo
fn unsafe_as_string_slice[
    lifetime: AnyLifetime = ImmutableStaticLifetime
](self) -> StringSlice[lifetime]:
    """Returns a string slice to this StringRef data.
    
    Safety:
        * This `StringRef` must remain valid for the duration that the resulting `StringSlice` is used.
           This might require manual lifetime extension.
        * This `StringRef` must contain valid UTF-8 encoded data.
    """
    return StringSlice[lifetime](
        unsafe_from_utf8_strref=substr
    )
```

stdlib/src/utils/string_slice.mojo
```diff
@@ -493,3 +493,77 @@ struct StringSlice[
         without the string getting deallocated early.
         """
         pass
+
+    fn _from_start(self, start: Int) -> Self:
+        """Gets the StringRef pointing to the substring after the specified slice start position.
```
ConnorGray, line 498:
```suggestion
        """Gets the `StringSlice` pointing to the substring after the specified slice start position.
```

stdlib/src/utils/stringref.mojo
gabrieldemarmiesse: I tried and it's not possible. It's called at compile time and the new find method in StringSlice is not compile-time friendly for unknown reasons.

JoeLoser: Makes sense, do you mind filing a GitHub issue for the follow-up refactoring so we don't forget to circle back here?

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
Looks good.  @gabrieldemarmiesse do you mind resolving conflicts in `stdlib/src/utils/string_slice.mojo` and then I'll sync this?  Thanks!

stdlib/src/utils/stringref.mojo
gabrieldemarmiesse: Done: https://github.com/modularml/mojo/issues/3294

gabrieldemarmiesse:
@JoeLoser I merged nightly into this branch, it should be all good now :)

JoeLoser:
!sync

Issue #3200, opened 2024-07-09
[Docs] Contributor dev env setup process could be improved


haifeng-jin:
### Where is the problem?

https://github.com/modularml/mojo/blob/main/CONTRIBUTING.md

### What can we do better?

As a first-time contributor setting up my dev env, I ran into a few issues that could be improved from both the repo setup and the documentation.

## Issues

1. Installing the test dependencies of LLVM version 17 was not straightforward. I had to follow the `.yml` file specifying the CI workflow to understand the dependency version. I had to copy the bash commands line-by-line to run them in my terminal.
2. It is unclear how to run a single or all test(s) locally. I encountered a few failed tests while running the `run-tests.sh` and found no documentation on running a single test.

## Solution
1. Extract the two `.sh` scripts (`install-build-tools-linux.sh` and `install-build-tools-macos.sh` under `stdlib/scripts`) to install test dependencies like LLVM with the specified version on MacOS and Linux. Let the CI call these scripts to install the dependencies.
2. Add a new section in `CONTRIBUTING.md` describing how to install the testing dependency and run all the tests and a single test.

cc @scottamain @jackos 

### Anything else?

I am happy to create some PRs to fix it.
I also have a small article summarizing how to polish contributor experience for open-source software. Just in case you are interested : ) [Link](https://betterprogramming.pub/the-essentials-of-a-contributor-friendly-open-source-project-b3fbd0fc36a7)

JoeLoser:
Thankd for filing the issue! We would be happy to accept PRs for both issues you mentioned.

haifeng-jin:
@JoeLoser Thanks! I will have a PR ready soon.

haifeng-jin:
The scripts has been separated out from the CI yml file now.
I am working on updating the contributing guide to include how to run tests locally.

ematejska:
@haifeng-jin Hi there, are you still working on the contributing guide?

haifeng-jin:
Hi @ematejska, Sorry that I am not working on it right now.
Feel free to close the issue or assign it to others.

scottamain:
This documentation and some of the process may have changed a lot since this issue was filed. If you still have any issues, please do file a new issue and offer your suggestions. Thank you!

PR #3234, opened 2024-07-13
[GitHub Actions] Extract install build tools into separate scripts

haifeng-jin:
Related to #3200.

Putting the install scripts as separate files would allow the users to easily install the dependencies.

The setup instructions will be added to the contributing guide later.

### DIFF
--- a/stdlib/scripts/install-build-tools-linux.sh
+++ b/stdlib/scripts/install-build-tools-linux.sh
@@ -1,0 +1,17 @@
+LLVM_VERSION=17
+wget https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+sudo ./llvm.sh $LLVM_VERSION
+rm llvm.sh
+
+# Make common LLVM binaries (including FileCheck) in our PATH so they work when used in an unversioned context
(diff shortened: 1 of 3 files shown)


stdlib/scripts/install-build-tools-linux.sh
```diff
@@ -0,0 +1,17 @@
+LLVM_VERSION=17
```
JoeLoser, line 1: **Suggestion** (Here and in the other script), I recommend adding

```
#!/usr/bin/env bash
```

in case users want to invoke these directly.  Make sure they have user executable bit set too (`chmod +x stdlib/scripts/install-build-tools-linux.sh`).

JoeLoser (approved):
Looks great, thank you!

stdlib/scripts/install-build-tools-linux.sh
JoeLoser: Also, adding `set -euo pipefail` in both scripts would be good for those that invoke the scripts directly too.

haifeng-jin: Done. Thanks!

JoeLoser (approved):

JoeLoser:
!sync

PR #3236, opened 2024-07-13
[stdlib] Remove incorrect alignment constraint

YichengDWu:
I believe the constraint 
```mojo
constrained[
            sizeof_t % alignment == 0, "size must be a multiple of alignment"
        ]()
```
is a misunderstanding of that `size_t*count` (size of bytes) must be a multiple of `alignment`.

This is not compile time checkable since `count` is a run-time value.

The following should be possible
```mojo
var ptr = UnsafePointer[Int64].alloc[alignment=64](8)
```

Notably, `DTypePointer` doesn't have this constraint.


Dan13llljws:
!sync

laszlokindrat (approved):
Makes sense to me, thanks!

PR #3239, opened 2024-07-13
[stdlib] Add utf8 safeguards, fix `chr` method, add unicode and utf16 parsing for `String`

### LINKED ISSUE #2842: [BUG] [stdlib] Unicode escape secuence is interpreted as utf8 and misrepresented.
### Bug description

```mojo
print(ord("\x85")) # prints 5
print(ord(String(List[UInt8](0x78, 0x38, 0x35, 0)))) # prints 120
```
`\x85` or `120 56 53` or `0x78 0x38 0x35` is NEXT LINE (NEL) character

### Steps to reproduce

### System information

```shell
`mojo 2024.5.2605`
```

martinvuyk:
Add utf8 safeguards, the second of many steps to fix [#2842](https://github.com/modularml/mojo/issues/2842#issuecomment-2171732716)

`fn chr(c: Int) -> String` function now returns a replacement character (�)
if the Unicode codepoint is invalid.

Added `String.from_unicode(values: List[Int]) -> String` and
`String.from_utf16(values: List[UInt16]) -> String` functions that return a String
containing the concatenated characters. If a Unicode codepoint
is invalid, the parsed String has a replacement character (�) in that index.


stdlib/test/builtin/test_string.mojo
```diff
@@ -305,13 +305,21 @@ def test_ord():
     assert_equal(ord("🔥".as_string_slice()), 128293)
 
 
-def test_chr():
+def test_unicode():
```
mzaks, line 308: I think I would rather have 3 tests as now the test is covering `char`, `String.from_unicode` and `String.from_utf16`

martinvuyk: The undelying function is the same for all 3 `_shift_unicode_to_utf8` which I pulled out of the `chr` implementation, but yeah I can separate them

JoeLoser:
@martinvuyk do you mean for this to be a draft still, or still pursuing this?  Happy to review it after it's rebased.

martinvuyk:
@JoeLoser I'm still mulling this one over.

Python's behavior is quite varied around this functionality:
> class str(object=b'', encoding='utf-8', errors='strict'):
>   [...]
>   If at least one of encoding or errors is given, object should be a [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object) (e.g. [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) or [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray)). In this case, if object is a [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) (or [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray)) object, then str(bytes, encoding, errors) is equivalent to [bytes.decode(encoding, errors)](https://docs.python.org/3/library/stdtypes.html#bytes.decode). Otherwise, the bytes object underlying the buffer object is obtained before calling [bytes.decode()](https://docs.python.org/3/library/stdtypes.html#bytes.decode). See [Binary Sequence Types — bytes, bytearray, memoryview](https://docs.python.org/3/library/stdtypes.html#binaryseq) and [Buffer Protocol](https://docs.python.org/3/c-api/buffer.html#bufferobjects) for information on buffer objects.

Some things hold me back on implementing these fully:
- Every possible datatype that can have those encodings needs to be taken into account
- Every possible string combination to express the [standard encodings](https://docs.python.org/3/library/codecs.html#standard-encodings) needs to be taken into account
- Every kind of code-path to decode according to the errors argument needs to be implemented
> errors controls how decoding errors are handled. If 'strict' (the default), a [UnicodeError](https://docs.python.org/3/library/exceptions.html#UnicodeError) exception is raised. Other possible values are 'ignore', 'replace', and any other name registered via [codecs.register_error()](https://docs.python.org/3/library/codecs.html#codecs.register_error). See [Error Handlers](https://docs.python.org/3/library/codecs.html#error-handlers) for details.

I'm leaning towards creating `Span.decode(self) -> String` and `String.encode(self) -> Span[Byte]` since it seems like we are slowly consolidating towards `Span[Byte]` being the equivalent of `bytes`. But there is a glaring issue which is the memory leakage of newly allocated `Span`s which have no ownership of their data and to me is a massive foot-gun (`String.encode()` for a non utf8 encoding would mean allocating a new buffer). `Span` would need a flexible pointer which can be or not be owned for this not to happen, this was one big motivation for `FlexiblePointer` in proposal #3728 which we could still implement independent of the trait proposed there (which Owen convinced me is too constraining). And if we do give it a special pointer and make it able to own its data or not, I'd like us to rename `Span` to `Buffer` since I think it will be more fitting.

Another possibility is going for `String.decode(List[Byte]) -> String` and `String.encode(self) -> List[Byte]` which would behave similarly to Python's  `bytes.decode()` and `string.encode(self) -> bytes`. But we would be leaving a lot of performance on the table, namely:
- Having decode return a new String would be wasteful for utf8 encoded buffers (which is most of text out there) for buffers that are passed as owned and own their data (which happens most of the time when receiving data over the wire and decoding) 
- `String.encode()` would need to consume the string if we want to give the pointer to the `List` without wasting copies
- Having both decode and encode require a `List[Byte]` is not as scalable as using `Span[Byte]` (if we make them able to own their data)

I can make a proposal for the special pointer and renaming of `Span` -> `Buffer`, or we can go with the `List[Byte]` approach WDYT?

martinvuyk:
split off into  #5255 and  #5258

PR #3242, opened 2024-07-14
[stdlib] Add min and max to Int

lsh:
Since `SIMD` already has a `min` and `max` function on it, it would make sense to add the same for `Int`. Also, since `Int` is basically just a thin wrapper around `__mlir_type.index`, it should probably reveal the rest of the `index` dialect ops eventually.


laszlokindrat:
Thanks for the patch! Given that the builtin `max` function provides the same functionality, I don't think we should do this. I sympathize with your point about consistency with `SIMD`, but I think we should rethink whether the `SIMD.max` method is worth keeping. I will look into this.

PR #3251, opened 2024-07-15
[stdlib] `__contains__` for ListLiteral

jjvraw:
An implementation of the `__contains__` function for `ListLiteral` as per issue #2658. 

The function parameterises the `EqualityComparable` trait, this maintains consistency with `Tuple.__contains__` behaviour (in nightly).

### DIFF
--- a/stdlib/test/builtin/test_list.mojo
+++ b/stdlib/test/builtin/test_list.mojo
@@ -12,7 +12,7 @@
 # ===----------------------------------------------------------------------=== #
 # RUN: %mojo %s
 
-from testing import assert_equal
+from testing import assert_equal, assert_true, assert_false
 
 
 fn test_list() raises:
@@ -39,7 +39,41 @@ fn test_repr_list() raises:
     assert_equal(empty.__repr__(), "[]")
 
 
+fn test_contains() raises:
+    # List
+    var l = List[String]("Hello", ",", "World", "!")
+    assert_true("Hello" in l)
+    assert_true(l.__contains__(String(",")))
+    assert_true("World" in l)
+    assert_true("!" in l)
+    assert_false("Mojo" in l)
+    assert_false(l.__contains__("hello"))
+    assert_false("" in l or l.__contains__(""))
+
+    # ListLiteral
+    var h = [1, False, String("Mojo")]
+    assert_true(1 in h)
+    assert_true(h.__contains__(1))
+    assert_false(True in h)
+    assert_false(h.__contains__(True))
+    assert_false(0 in h)
+    assert_true(False in h)
+    assert_false("Mojo" in h)
+    assert_true(String("Mojo") in h)
+    assert_false(String("") in h)
+    assert_false("" in h)
+
+    # TODO:
+    # Reevaluate the strict type checking behaviour in ListLiteral.__contains__.
+    # Consider aligning with the behavior with List.__contains__ when possible
+    # or a feasible workaround is identified.
+    # For instance, consider the following:
+    assert_true("Hello" in l and String("Hello") in l)
+    assert_true("Mojo" not in h and String("Mojo") in h)
+
+
 def main():
     test_list()
     test_variadic_list()
     test_repr_list()
+    test_contains()
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -297,6 +297,10 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   # Insert (2/3 of 1024) entries
   ```
 
+- `ListLiteral` now supports `__contains__`.
+  ([PR #3251](https://github.com/modularml/mojo/pull/3251) by
+  [@jjvraw](https://github.com/jjvraw))
+
 ### 🦋 Changed
 
 - The pointer aliasing semantics of Mojo have changed. Initially, Mojo adopted a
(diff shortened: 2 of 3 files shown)


stdlib/test/builtin/test_list.mojo
```diff
@@ -39,7 +39,28 @@ fn test_repr_list() raises:
     assert_equal(empty.__repr__(), "[]")
 
 
+fn test_equality() raises:
```
JoeLoser, line 42:
```suggestion
fn test_contains() raises:
```

I think calling this `test_contains` is more explicit than "equality" as we're not testing `__eq__` and `__ne__` on `ListLiteral` or `List` here, for example.

stdlib/test/builtin/test_list.mojo
```diff
@@ -39,7 +39,28 @@ fn test_repr_list() raises:
...
+    var p = List(2, 3, 5, 7)
```
JoeLoser, line 43: **Suggestion** I recommend also testing the `List.__contains__` with memory-only types, such as `String`.  Of note, this is homogenous whereas, the `ListLiteral` test case is a collection of heterogenous elements (as expected).

JoeLoser (approved):
Nice!

stdlib/test/builtin/test_list.mojo
jjvraw: On this note, shall I move the `List.__contains__` testing to [`collections/test_list.mojo`](https://github.com/modularml/mojo/blob/nightly/stdlib/test/builtin/test_list.mojo) rather?

JoeLoser: Well, the `List` already has tests for `contains` - see https://github.com/modularml/mojo/blob/a78d7548ebebfc074d237c21da29e027c89fd3f1/stdlib/test/collections/test_list.mojo#L792.  I'd recommend we beefen up those tests to include `List[String]` so we test a memory-only type.

I'd keep the `ListLiteral.__contains__` tests here in `stdlib/test/builtin/test_list.mojo`.  How does that sound?

jjvraw: Happy :)

stdlib/test/builtin/test_list.mojo
```diff
@@ -39,7 +39,39 @@ fn test_repr_list() raises:
...
 
+fn test_contains() raises:
+    # List
+    var p = List(2, 3, 5, 7)
+    assert_true(2 in p and 3 in p)
+    assert_true(p.__contains__(5) and p.__contains__(7))
+    assert_false(1 in p and 4 in p)
+    assert_false(p.__contains__(6) and p.__contains__(8))
+
+    var l = List[String]("Hello", ",", "World", "!")
```
jjvraw, line 50: Should `List[String].__contains__` handle `StringLiteral` differently from `ListLiteral.__contains__`? Under the current implementation of `List.__contains__`, the following would hold:

```mojo
# Where `l` is defined as a List and `h` as a ListLiteral

assert_true("Hello" in l and String("Hello") in l)

assert_true("Mojo" not in h and String("Mojo") in h)
```

The behaviour is not exactly aligned between the two. Should the inherent nature of ListLiteral as a heterogeneous list result in a different __contains__ method compared to List[String]?

martinvuyk: Hi, thanks for this contribution !

I think to achieve that generically:
- on the side of ListLiteral: we'd need to somehow check if `Ts[i]` has a constructor that takes in `T` which I think isn't possible. 
- on the side of List: we'd need to "turn off" implicit casting of other types to the `EqualityComparableCollectionElement` which I also think is not possible currently.

Hacky solution ?: add a `StringableCollectionElement` constraint and use `str()` to compare everything. Kill performance, be hated by the next generation, but hey it works lol

just kidding, TLDR:  I think it's not achievable with current Mojo features :( , and I think accepting implicit casting is more pythonic

jjvraw: Thanks for the reply! :) My concern was that the implementations between `List` and `ListLiteral` aren't directly aligned.  

I read through a few issues and it seems that strict type checking in `ListLiteral.__contains__` is somewhat necessary given the current state of Mojo.

I suppose this is a design choice moving forward: whether to keep strict comparisons in the heterogeneous-like `ListLiteral` or adopt a more Pythonic behavior like `List`. For the latter, I can leave a TODO?

martinvuyk: > I can leave a TODO?

great idea to avoid this popping up mysteriously as a bug later on, maybe even a `# FIXME` with the example you gave

jjvraw: Okay, let's wait for @JoeLoser's feedback

JoeLoser: I think this is expected, although unfortunate.  Feel free to put a `FIXME` or `TODO` in the `ListLiteral` test case you mentioned.  Thanks for digging into this!

stdlib/test/builtin/test_list.mojo
```diff
@@ -39,7 +39,47 @@ fn test_repr_list() raises:
     assert_equal(empty.__repr__(), "[]")
 
 
+fn test_contains() raises:
+    # List
```
JoeLoser, line 43: **Suggestion** You can drop these list specific tests since they're tested elsewhere

JoeLoser:
!sync

PR #3261, opened 2024-07-16
[stdlib] Fix argument convention for `iota`

soraros:
- The `List` arguments should be `inout`.
- Improved the docstinrg a bit.
- Make `DType` parameter infer-only.


laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

PR #3264, opened 2024-07-17
[stdlib] Support creating nested python objects using list literals

kszucs:
Previously creating python objects from list literals was not possible making the python experience less convenient. Noticed this while I wanted to create a pyarrow schema from mojo. Instead of the following:

```mojo
    var pa = Python.import_module("pyarrow")
    var int_field = pa.field("int_field", pa.int32())
    var string_field = pa.field("string_field", pa.string())
    var schema = pa.schema([int_field, string_field])
```

I had to append the fields to an empty schema:

```mojo
    var schema = pa.schema([])
    schema = schema.append(some_int)
    schema = schema.append(some_string)
```

This change allows creating a `PythonObject` from both `ListLiteral`s and `Tuple`s containing `PythonObject` values.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -16,6 +16,27 @@ what we publish.
 
 ### ⭐️ New
 
+- Creating nested `PythonObject` from a list or tuple of python objects is
+  possible now:
+
+  ```mojo
+  var np = Python.import_module("numpy")
+  var a = np.array([1, 2, 3])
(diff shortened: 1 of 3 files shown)


JoeLoser (approved):
Nice, thank you for the contribution!

JoeLoser (commented):
!sync

JoeLoser:
!sync

PR #3266, opened 2024-07-17
[stdlib] Implement `collections.Counter` - Part 3

msaelices:
Support for Python's [collections.Counter](https://docs.python.org/3/library/collections.html#collections.Counter) - Part 3

Pending to match Python signature: `__or__`, `__ior__` and `__neg__`


soraros:
- `__or__` should also strip nonpositive counts (according to Python)
- `__or__` and `__ior__` probably shouldn't raise

laszlokindrat (commented):
Thanks! Generally looks good, but as @soraros pointed out, it would be better if these methods didn't raise. If they have to, could you please describe what they raise in their docstring?

About the behavior of `__or__`: it would be better to follow Python's semantics, but if there is a good reason, we can also deviate. Either way, please document the behavior with respect to elements with zero counts.

msaelices:
> Thanks! Generally looks good, but as @soraros pointed out, it would be better if these methods didn't raise. If they have to, could you please describe what they raise in their docstring?

I tried to avoid the `raises` declaration by this tricks: https://github.com/msaelices/mojo/commit/0e1958a6f4a8e8aee42c4bf6adcda9f3ad164f48

Actually, it will never raise anything, as the errors that can potentially be raised are `KeyErrors` and we are just accessing the keys retrieved from the `.keys()` method, so it will never raise but compiler does not know.

Not sure if the fix is too tricky.

> About the behavior of `__or__`: it would be better to follow Python's semantics, but if there is a good reason, we can also deviate. Either way, please document the behavior with respect to elements with zero counts.

Fixed here: https://github.com/msaelices/mojo/commit/8923e19bb3339515849b68b84eb447a9e4ee0fd1

msaelices:
Could you PTAL?

laszlokindrat (approved):
LGTM, thanks!

laszlokindrat:
!sync

PR #3268, opened 2024-07-17
[CI] Use build scripts from CI workflows

JoeLoser:
In https://github.com/modularml/mojo/pull/3234, when synced internally, only the `scripts` were added.  The modifications to the GitHub workflow didn't land in https://github.com/modularml/mojo/commit/b939aeab913b2a83f3343cec0c87717eb3f6ca1a since the `.github` directory isn't managed by Copybara currently.

So, this PR does the `.github` workflow cleanup part of https://github.com/modularml/mojo/pull/3234 now that https://github.com/modularml/mojo/commit/b939aeab913b2a83f3343cec0c87717eb3f6ca1a has landed.

### DIFF
--- a/.github/workflows/standard_library_tests_and_examples.yml
+++ b/.github/workflows/standard_library_tests_and_examples.yml
@@ -38,7 +38,6 @@ jobs:
         shell: bash
     env:
       DEBIAN_FRONTEND: noninteractive
-      LLVM_VERSION: 17
 
     steps:
       - name: Checkout repo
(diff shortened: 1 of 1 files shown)


ConnorGray (approved):
Makes sense, thanks for doing this manually. Hopefully we'll catch it next time.

JoeLoser:
> Makes sense, thanks for doing this manually. Hopefully we'll catch it next time.

I have an open Linear issue internally to sync the `.github` directory with Copybara FYI.

Issue #3254, opened 2024-07-16
[Feature Request] Add benchmark for `Dict` and explore swiss table algorithm with SIMD lookup


jon-chuang:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

While optimizing for small dictionaries is fine, I also think it might be nice to support cases which may involve large dicts.

For good perf on these, swiss table is a decent choice, implemented for Rust's HashMap and abseil's flat_hash_map. It utilizes SIMD to generate a bitmask from 7bit key prefixes for further probing of key exact match when doing a linear scan of a bucket.

### What is your motivation for this change?

perf

### Any other details?

This will likely be a last-mile issue to solve perf-wise

Let's see some benchmark numbers first

martinvuyk:
Hi, interesting algorithm. But wouldn't it be simpler to just keep a B-tree index (of the first/last 7 hash bits) if the data becomes really big (often used in DBs) ? Though this would introduce much more memory overhead, and inserts would be kinda expensive. 

Maybe we could build it as a separate experimental type first and try it out. 

@mzaks @bethebunny you guys have been working with Dict a lot (I think),  thoughts?

JoeLoser:
Benchmarking dict and understanding the tradeoffs of performance, memory usage, and various internal metadata-tracking algorithms would all be good things to do.  Feel free to work on this!  We're not actively working in this area at the moment.

mzaks:
Current implementation of the dict in Mojo standard library is based on Pythons implementation. It implements a pseudo random probing algorithm and stores key value pairs together. This doesn't allow SIMD  comparison, but I think this is still not the main bottleneck, main bottleneck is the hash function. Hence I wrote a proposal https://github.com/modularml/mojo/blob/main/proposals/improved-hash-module.md and there are a few PRs to address it, which are currently blocked by some compiler issues see https://github.com/modularml/mojo/pull/2891. 

In the meantime we can implement Swiss table like dict and you can also have a look at https://github.com/mzaks/compact-dict I did a short presentation about it here https://youtu.be/3FKSlhZNdL0?si=fEshjEgpNvVUAdBu (I am the second presenter at 13:20) I would be happy if someone wants to take the time and add a SIMD based probing to it. Currently it works with simple linear probing.

soraros:
[...]

PR #3269, opened 2024-07-18
[stdlib] Better `Dict` benchmarks

martinvuyk:
Better `Dict` benchmarks
one of the steps for #3254


stdlib/benchmarks/collections/bench_dict.mojo
```diff
@@ -116,17 +86,43 @@ fn bench_dict_large_lookup(inout b: Bencher) raises:
...
+        m.bench_function[bench_dict_lookup[size]](
+            BenchId("bench_dict_lookup[" + str(size) + "]")
+        )
+
     m.dump_report()
+
+    @parameter
+    for i in range(len(sizes)):
+        alias size = sizes.get[i, Int]()
+        var mem_s = sizeof[Dict[Int, Int]]() * size
```
martinvuyk, line 125: Here we would actually need to know how many bytes are allocated in all internal fields (there could be pointers, non linearly growing data w/ regards to size, etc.). So we will likely need to add some kind of trait that returns the size of allocated bytes for a type and implement it for all types whose memory footprint is to be benchmarked.
Something like:
```mojo
trait MemoryMeasurable:
    fn _total_bytes_used(self):
        """Calculate total bytes used by the instance including `sizeof[Self]()`."""
        ...

struct Dict:
    ...
    fn _total_bytes_used(self):
        # subtract sizeof type because sizeof[Self]() already includes that (?)
        var entries = self._entries._total_bytes_used() - sizeof[V]
        var index = self._index._total_bytes_used() - sizeof[K]
        return sizeof[Self]() + entries + index
```

rd4com: Hello, very cool, here is a review for that part of the PR.

For the size that a Dict uses (in byte), you'd have to use a formula:

```mojo
alias elements = 1000
var tmp_reserved = upper_nearest_power_of_two(elements) #1024
#the resize up if overloaded
if (elements*3)>(tmp_reserved*2): tmp_reserved *= 2

var base = tmp_reserved

base *= sizeof[Optional[DictEntry[K,V]]]() #allocated self._entries
#self._entries and self._index structs + _reserved:Int, _n_entries: Int, size:Int
base += sizeof[Dict[K,V]]() 

#the allocated memory by index table:
if tmp_reserved <= 128:
    base+= sizeof[Scalar[DType.int8]]()*tmp_reserved
elif tmp_reserved <= 2**16 - 2:
    base+= sizeof[Scalar[DType.int16]]()*tmp_reserved
elif ... #see the DictIndex.__init__() for more elifs

print(base, " bytes")
```

Important note:
if `V` or `K` of `Dict[K,V]` is a type that allocate, it would not be reflected.
(a List for example, have a known inlined size, and heap allocations on the pointer)

As you pointed out,

maybe we need a new method+trait on some types that says how much memory an instance currently uses.
(`type.total_memory_used()` or something alike)

Using conditional conformance, we could assert the total memory usage of a type instance,
even if it is composed of types that allocates but conform to the trait.

It might not be a priority but an idea to keep in mind for the future.

rd4com:
Hello @martinvuyk , good job with last commits.

Here is another review for the implementation of last review.

&nbsp;

Since the new function use an already populated dict:

```mojo
fn total_bytes_used(items: Dict[Int, Int]) -> Int:
```

The resizes occured already.

&nbsp;

So you can remove the:

```mojo
var reserved = bit_ceil(len(items))
if (len(items) * 3) > (reserved * 2):
    reserved *= 2
```

in favour of:

```
var reserved = items._reserved()
```

&nbsp;

In addition, the size of the dict structure is fixed.

So changing this:
```mojo
amnt_bytes += sizeof[Dict[Int, Int]]() - entry_size
```
to this:
```mojo
amnt_bytes += sizeof[Dict[Int, Int]]()
```
(The size of the dict struct is just two integers, an list struct and a pointer for _index)

&nbsp;

Other than that, the allocated part is added gracefully :+1:

JoeLoser:
!sync

martinvuyk:
Hi @JoeLoser is there a reason this can't get merged internally?

JoeLoser:
> Hi @JoeLoser is there a reason this can't get merged internally?

Looks like it failed the CI on auto-merge when I last imported it and the CI logs have expired.  Let me resync it and take a look.  Thanks for the ping!

JoeLoser:
!sync

JoeLoser:
The benchmark is failing in CI for some simple reasons now (needing to import `Optional`, and the `BenchConfig` keyword argument is no longer valid).  I'll take a look at these soon.

PR #3275, opened 2024-07-20
[stdlib] Add more constants, and tweak descriptions.

helehex:
Not sure if it's better to just use the `nan[DType]()` function, but I find it convenient to have a FloatLiteral one.

If anyone wants me to add the twelfth root of 2, I am more than happy todo so.

### DIFF
--- a/stdlib/src/math/constants.mojo
+++ b/stdlib/src/math/constants.mojo
@@ -10,7 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
-"""Defines math utilities.
+"""Defines mathematical constants, accurate to 64 decimal places.
 
 You can import these APIs from the `math` package. For example:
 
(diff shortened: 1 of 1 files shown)


helehex:
It's probably smart to have `tau = 2 * pi` instead of writing it all out, to at least ensure consistency between them.. but I'm not sure if it matters in practice. Let me know I guess.

soraros:
I actually wonder if it makes more sense to add these alias to `math`.

soraros:
`utils.numerics.nan` etc. are also exported through `math`.

helehex:
oh yeah, I was going to change the sub-module docstring to something like:
`"""Defines mathematical constants, accurate to 64 decimal digits."""`

helehex:
Don't worry @soraros, I have a feeling it will all work out in the end :]

helehex:
Also, `pi*2 == tau` is still `True`, so that's always good.

helehex:
I just remembered, you can also do things like `Float64.Max` to get inf. Maybe there's too many ways to do a simple thing.

JoeLoser:
Thanks for the contribution, but we're unfortunately not accepting contributions to the `math` module at this time.  As such, I'm going to close this PR. Sorry!

PR #3276, opened 2024-07-20
[stdlib] Make `FileDescriptor` conform to `CollectionElement` trait

martinvuyk:
Make `FileDescriptor` conform to `CollectionElement` trait

Add `Arc.count()` method that returns the current inner refcount value


martinvuyk:
Hi @ConnorGray , is there any way we could merge this today? I'd like to start testing my attempt at building a socket package this weekend and I need an `Arc[FileDescriptor]` to  build it safely. I can implement it differently for the time being, just asking in case it's possible :D

soraros:
FYI, 22060ea recently landed.

PR #3283, opened 2024-07-22
[stdlib] Remove `CollectionElementNew` and use only `CollectionElement`

gabrieldemarmiesse:
After the work in https://github.com/modularml/mojo/pull/3282 is merged, we end up with the two traits `CollectionElement` and `CollectionElementNew` inheriting from the same two traits `Movable` and `ExplicitlyCopyable`. Since they're identical, we can remove one of them (the `CollectionElementNew`). 

Sometime, we remove the trait entirely in structs definitions because `KeyElement` is already present and `KeyElement` inherit from `CollectionElement`. The compiler warns us when such a situation happens with the message `struct xxx already inherit from CollectionElement`.


gabrieldemarmiesse:

PR #3284, opened 2024-07-22
[stdlib] Make implicit copy stricter for a few types (part 1)

gabrieldemarmiesse:
This PR depends on https://github.com/modularml/mojo/pull/3283

* `Variant` is not implicitly copyable anymore. I wanted to enable it only if all types of the variant have the trait `Copyable` but I couldn't find the syntax to do so.
* `Optional` Is implicitly copyable only if T is `Copyable`. Which I believe is reasonnable since it implies a cheap copy. It uses conditional conformance. I added the corresponding tests.
* `_FormatCurlyEntry` is not implicitly copyable anymore, which is good since the variant within can contain a `String` which is expensive to copy.
* `InlineString` is not implicitly copyable anymore since that can mean copying a whole `String`.


gabrieldemarmiesse:

PR #3286, opened 2024-07-22
[docs] Add `str()` in `print_many()` in Functions

shayan-shams:
Resolves `error: invalid call to 'print_string': argument #0 cannot be converted from 'T' to 'String' `


JoeLoser (approved):

JoeLoser:
!sync

PR #3298, opened 2024-07-24
[stdlib] Add equality methods to `Reference`.

helehex:
Does anyone know if there are any checks to take care of some equality/inequality cases at compile time? Like, by comparing the lifetimes.

### DIFF
--- a/stdlib/src/memory/reference.mojo
+++ b/stdlib/src/memory/reference.mojo
@@ -332,3 +332,31 @@ struct Reference[
             The MLIR reference for the Mojo compiler to use.
         """
         return __get_litref_as_mvalue(self.value)
+
+    @always_inline("nodebug")
+    fn __eq__(self, rhs: Reference[type, _, address_space]) -> Bool:
+        """Returns True if the two pointers are equal.
+
+        Args:
+            rhs: The value of the other pointer.
+
+        Returns:
+            True if the two pointers are equal and False otherwise.
+        """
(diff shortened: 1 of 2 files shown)


stdlib/src/memory/reference.mojo
```diff
@@ -332,3 +332,31 @@ struct Reference[
...
+
+        Args:
+            rhs: The value of the other pointer.
+
+        Returns:
+            True if the two pointers are equal and False otherwise.
+        """
+        return UnsafePointer(
+            __mlir_op.`lit.ref.to_pointer`(self.value)
+        ) == __mlir_op.`lit.ref.to_pointer`(rhs.value)
```
soraros, line 348: Maybe wrap both with `UnsafePointer` instead of relying on implicit conversion.

helehex: it looks so nice though

helehex: I'll consider it. There might be a better way anyways, if not soon.

martinvuyk, line 348: How about this ?
```suggestion
        alias op = __mlir_op.`lit.ref.to_pointer`
        return UnsafePointer(op(self.value)) == UnsafePointer(op(rhs.value))
```

helehex: I am not strictly opposed to it

Issue #3126, opened 2024-06-26
[BUG] `List` doesn't work at compile time.


soraros:
### Bug description

As title. At least `List.__getitem__` doesn't work.

### Steps to reproduce

```mojo
fn main():
    alias l = List[Int](1, 2, 3)
    print(l[0])  # prints 0
```

### System information

```shell
Mojo 2024.6.2614 (366c690a) on Docker, Intel Mac
```

DWSimmons:
I went back in time (modular clean; modular install nightly/mojo --install-version) and found that this bug does not exist in 2024.6.2305 but does exist in 2024.6.2505.

Additionally, there is some odd - to me as I am truly a rookie with extra time on his hands - behavior.  If some additional light could be shed on the following, I would greatly appreciate it.

As expected: 
In 2024.6.2305 through to today (Note use of `var`)
```
fn main():
    var l = List[Int](1, 2, 3)
    print(l[0])  # prints 0
```
In 2024.6.2305 `alias` behaves as expected
```
fn main():
    alias l = List[Int](1, 2, 3)
    print(l[0])  # prints 1
```
However, in 2024.6.2505
```
fn main():
    alias l = List[Int](1, 2, 3)
    print(l[0])  # prints 0
```

DWSimmons:
Possibly related to  https://github.com/modularml/mojo/pull/2847 ?

mikowals:
There is [some history](https://github.com/modularml/mojo/pull/2294#issuecomment-2065916684) of compile time issues as collections switched to returning references.  But there is at least one [use of alias List  in the stdlib](https://github.com/modularml/mojo/blob/main/stdlib/src/builtin/dtype.mojo#L748-L771) so that problem seems to have been solved or occurs much more narrowly than in [previous implementations](https://github.com/modularml/mojo/issues/2543).

If this is a recent breakage as mentioned above then it is probably [this commit](https://github.com/modularml/mojo/commit/620b26091e3eb4a47b1d6a7a3cad7ba0e959d95f).   My PR (#2847 ) was merged afterward just for the changes to tests if I understood the comment there correctly.

gabrieldemarmiesse:
I think we need way more "compile time tests" because those regressions/bugs are really frequent. Not many structs are currently tested at both compile-time and runtime. This could be a very good "good first issue".

ematejska:
[...]

PR #3066, opened 2024-06-17
[stdlib] Cleanup `polynomial.mojo`, improve docstring

soraros:
- Small simplification to the function `_horner_evaluate`.
- Improve its docstring to match the argument names.
- ~~Finish off removing `unroll` by cleaning up the imports.~~

PR #3301, opened 2024-07-24
[stdlib] Simplify `_horner_evaluate`

soraros:
Reapply #3066 after #3126.

### DIFF
--- a/stdlib/src/math/polynomial.mojo
+++ b/stdlib/src/math/polynomial.mojo
@@ -81,15 +81,10 @@ fn _horner_evaluate[
     Returns:
         The polynomial specified by the coefficients evaluated at value x.
     """
-    alias num_coefficients = len(coefficients)
-    alias c_last = coefficients[num_coefficients - 1]
-    alias c_second_from_last = coefficients[num_coefficients - 2]
-
-    var result = x.fma(c_last, c_second_from_last)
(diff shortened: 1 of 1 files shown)


stdlib/src/math/polynomial.mojo
```diff
@@ -81,15 +81,10 @@ fn _horner_evaluate[
     Returns:
         The polynomial specified by the coefficients evaluated at value x.
     """
-    alias num_coefficients = len(coefficients)
-    alias c_last = coefficients[num_coefficients - 1]
-    alias c_second_from_last = coefficients[num_coefficients - 2]
-
-    var result = x.fma(c_last, c_second_from_last)
+    var result = x.fma(coefficients[-1], coefficients[-2])
```
laszlokindrat, line 84: While this is equivalent in terms of the result it yields, because the indexing happens in a non-parameter context, this will potentially materialize the `coefficients` vector and indexing could happen at runtime.

We currently don't have a way to mark functions as "free of side effects" or "can be safely hoisted to parameter domain", so the compiler cannot assume that `List.__getitem__` would not, for example, print to stdout. So users have to explicitly mark compile-time computations as such, using `alias`.

laszlokindrat:
Thanks @soraros! While I agree that this change would be nice to make, it can cause potential perf degradation so we can't take it at this point.

PR #3306, opened 2024-07-25
[stdlib] Add `has_fma` function to `sys` package

leandrolcampos:
This PR adds a new function `has_fma` to the `sys` package. This function checks if the host system supports Fused Multiply-Add (FMA) instructions.

**Justification**:

The `has_fma` function is important for developers implementing math functions. Numerical algorithms can adopt different strategies depending on whether the target architecture supports FMA or not. For example, the implementation of the `log10` function in the LLVM libc library considers the presence of FMA support.

**Note on Unit Testing:**

I have not added a unit test for this function because I am unsure how to implement it. I searched for unit tests for similar functions, such as `has_avx`, but could not find any. If anyone has suggestions on how to write a unit test for this function, they are welcome.


laszlokindrat:
> I have not added a unit test for this function because I am unsure how to implement it. I searched for unit tests for similar functions, such as `has_avx`, but could not find any. If anyone has suggestions on how to write a unit test for this function, they are welcome.

Excellent question! It is indeed difficult to unit test these, but you can at least add tests that prevent accidental deletion/changes of these:
```mojo
from sys import has_fma, has_amx

fn test_target_has_feature():
    _ : Bool = has_fma()
    _ : Bool = has_amx()
    ...
```

Issue #2867, opened 2024-05-28
[stdlib] Improvements to `pow` and `Powable`


laszlokindrat:
In https://github.com/modularml/mojo/commit/3539c86d4c38a037e253c5c240061004328cc066, the `pow` function was moved from `math` to the builtins. There are a couple of TODOs in that patch:
- [ ] Add a non-trivial but also not too complicated example in the `Powable` trait's docstring
- [x] Improve and extends tests for `SIMD.__pow__`: done in https://github.com/modularml/mojo/pull/3315
- [ ] [NEEDS DISCUSSION] Implement `__pow__` for `IntLiteral`. Before going forward with this, we should have a discussion on how we handle negative exponents, since this will have an effect on the API. Either way, the implementation should use [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring).

There are also further potential improvements:
- [ ] Seems like `Int` might be missing tests for `__pow__`.
- [ ] Python's `pow` accepts a third `mod` argument to `pow`. We should do the same, and modify the `Powable` trait accordingly. Something similar is being done for `round` and could serve as guidance: https://github.com/modularml/mojo/pull/2752
- [ ] [NEEDS DISCUSSION] We might want a separate trait and `pow` overload for modeling things that can have `Int` powers only. The problem of negative exponents will also come up here, so we should think about how we approach this.

If someone is interested in tackling these, please do one of these per PR, to ensure the changes are limited in scope and can be reviewed quickly. Also, please don't start work on the items marked with `[NEEDS DISCUSSION]` until we align on the approach.

conorbronsdon:
Status update: `pow()` was made builtin and `Powable` trait added in [v24.4](https://docs.modular.com/mojo/changelog/#v244-2024-06-07). Some specific improvements from the original request may remain.

Issue #2855, opened 2024-05-27
[BUG] [stdlib] Weird bug with `SIMD.__mod__`


martinvuyk:
### Bug description

using this code
```mojo
fn main():
    alias val1 = SIMD[DType.uint8, 4](2, 2, 2, 0)
    alias val2 = SIMD[DType.uint8, 4](2, 2, 2, 0)
    alias val3 = SIMD[DType.uint8, 4](2, 2, 2, 0)
    alias val4 = SIMD[DType.uint8, 4](2, 2, 2, 0)
    # mod with others
    print(val1 % val2)
    print(val1 % val3)
    print(val1 % val4)
    print(val2 % val3)
    print(val2 % val4)
    print(val3 % val4)
    # self mod
    print(val1 % val1)
    print(val2 % val2)
    print(val3 % val3)
    print(val4 % val4)
```

outputs:
```text
[85, 85, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
[0, 0, 0, 0]
```

### Steps to reproduce

- Include relevant code snippet or link to code that did not work as expected.
- If applicable, add screenshots to help explain the problem.
- If using the Playground, name the pre-existing notebook that failed and the steps that led to failure.
- Include anything else that might help us debug the issue.

### System information

```shell
- What OS did you do install Mojo on ?
- Provide version information for Mojo by pasting the output of `mojo -v`
`mojo 2024.5.2605`
- Provide Modular CLI version by pasting the output of `modular -v`
```

JoeLoser:
Can you be a bit clearer on your idea of expected behavior so we can be aligned on what the claimed bug is?<br><br>Re the "is this UB?", won't this hit [https://github.com/modularml/mojo/blob/d509360f80c6d883bbdbabd18f6839692e868c99/stdlib/src/builtin/simd.mojo#L760](https://github.com/modularml/mojo/blob/d509360f80c6d883bbdbabd18f6839692e868c99/stdlib/src/builtin/simd.mojo#L760)?  A lot of functions in the stdlib suffer from this problem, but because of their pervasive use, they don't raise and instead just "return the wrong result".  This is terrible for many reasons.  Until we have parametric raises, I think we should change these sort of things to `abort`.  What do you think?

[...]

PR #3307, opened 2024-07-25
[stdlib] Introduce modular exponentation via `pow(Int, Int, Int)`

jjvraw:
Partially addresses Issue #2867. 

This PR implements modular exponentiation, as suggested,
> Python's pow accepts a third mod argument to pow. We should do the same, and modify the Powable trait accordingly.

Since modulo arithmetic is defined for integers, I believe the implementation should be part of `Int`, as opposed to modifying `Powable`. Consequently, `pow(Int, Int, Int)` can wrap `Int.__pow__(Self, Self, Self)`.

Modular exponentiation uses the square-and-multiply algorithm. For negative exponents, it requires the calculation of a multiplicative inverse, which is performed using the extended Euclidean algorithm and implemented in `Int.mod_inverse`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -350,6 +350,17 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   # "Mojo 'Mojo'"
   ```
 
+- Modular exponentation for is now supported via `math.pow(Int, Int, Int)`
+  ([PR #3307](https://github.com/modularml/mojo/pull/3307) by
+  [@jjvraw](https://github.com/jjvraw))
+
+  Example:
+
+  ```mojo
+  pow(2, 4, 5) # 1
+  pow(2, -1, 7) # 4
+  ```
+
 ### 🦋 Changed
 
 - The pointer aliasing semantics of Mojo have changed. Initially, Mojo adopted a
--- a/stdlib/src/builtin/int.mojo
+++ b/stdlib/src/builtin/int.mojo
@@ -649,6 +649,56 @@ struct Int(
         return res
 
     @always_inline("nodebug")
+    fn __pow__(self, exp: Self, mod: Self) -> Self:
+        """Returns `self` raised to `exp` power, modulo `mod`.
+
+        This method uses the square-and-multiply algorithm to compute
+        modular exponentiation.
+
+        Args:
+            exp: The exponent. Can be positive or negative.
+            mod: The modulus. Must be greater than 1.
+
+        Returns:
+            The result of `self` raised to `exp` power mod `mod`.
+
+        Note:
+            - If mod <= 1, returns 0 to indicate an error.
+            - For negative exponents, it computes the modular multiplicative
+              inverse of self before exponentiation. If no inverse exists,
+              returns 0.
+        """
+        if mod == 0:
+            # Modulus must be greater than 1.
+            # This should raise an exception.
+            debug_assert(False, "Modulus must be greater than 1.")
+            return 0
+
+        if mod == 1:
+            return 0
+
+        var base: Int = self % mod
+        var e = exp
+
+        if e < 0:
+            base = self._mod_inverse(mod)
+            if not base:
+                # Inverse of `base` mod `mod` does not exist
+                # This should raise an exception.
+                return 0
+            e = -e
+
+        var res: Int = 1
+
+        while e > 0:
+            if e & 1 == 1:
+                res = (res * base) % mod
+            base = (base * base) % mod
+            e >>= 1
+
+        return res
+
+    @always_inline("nodebug")
     fn __lshift__(self, rhs: Int) -> Int:
         """Return `self << rhs`.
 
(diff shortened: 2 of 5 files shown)


jjvraw:
Note worthy:
Python returns `ValueError`'s for the following cases in which this implementation will return `0`.

```python
>>> pow(2, 4, 0)
ValueError: pow() 3rd argument cannot be 0
```

```python
>>> pow(2, -1, 4)
ValueError: base is not invertible for the given modulus
```

stdlib/src/builtin/int.mojo
```diff
@@ -648,6 +648,55 @@ struct Int(
...
+
+        Note:
+            - If mod <= 1, returns 0 to indicate an error.
+            - For negative exponents, it computes the modular multiplicative
+              inverse of self before exponentiation. If no inverse exists,
+              returns 0.
+        """
+        if mod == 0:
+            # Modulus must be greater than 1.
+            # This should raise an exception.
```
martinvuyk, line 673: Hi, @JoeLoser sorry to pull you in here, but I think this falls into the same category as issue #2855 about undefined math operations leading to undefined behavior. I think we should define a standard way to deal with these soon or we might end up with a bunch of these edge cases hidden away.

docs/changelog.md
```diff
@@ -350,6 +350,20 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
...
+
+  Example:
+
+  ```mojo
+  pow(2, 4, 5)
+  # 1
+    
+  pow(2, -1, 7)
+  # 4
+  ```
```
martinvuyk, line 365: nit:
```suggestion
  Examples:

  ```mojo
  pow(2, 4, 5) # 1
  pow(2, -1, 7) # 4
  ```
```

stdlib/src/builtin/int.mojo
```diff
@@ -1099,6 +1148,51 @@ struct Int(
...
+
+        Returns:
+            The modular multiplicative inverse of `self` modulo `m`, if it exists.
+            Returns 0 if no inverse exists or if an error occurs.
+
+        Note:
+            - If `self` and `m` are not coprime, no modular inverse exists,
+              and the method returns 0.
+        """
+        if m <= 1:
```
martinvuyk, line 1169: In Python: 
- `pow(1, 1, 0)` raises `ValueError: pow() 3rd argument cannot be 0`
- `pow(0, 1, 1)` returns `0`. 
- `pow(0, 1, -1)` returns `0`. (definitely not valid math)
- `pow(50, 1, -50)` returns `0`. (definitely not valid math)

I think the mod inverse of 0 doesn't exist either since any number mod by itself is 0. But I think we should keep to Pyhton's behavior.

If we take a `UInt` into this function it might also lead to -50 for example being cast to its positive equivalent (18446744073709551566) and giving different results than Python which is sad.

And as for 0 that raises in Python,  I personally have an inclination towards aborting for cases of undefined Math to not mark the function as `raises` instead of raising. But it needs to be decided by the community/Mojo team how we deal with these cases as a whole.

stdlib/src/builtin/math.mojo
```diff
@@ -309,6 +309,22 @@ fn pow(base: SIMD, exp: Int) -> __type_of(base):
     return base.__pow__(exp)
 
 
+fn pow(base: Int, exp: Int, mod: Int) -> Int:
+    """Compute modular exponentiation of an integer raised to a power, modulo
+       another integer.
```
martinvuyk, line 314: micro-nit (sorry XD)
```suggestion
    another integer.
```

martinvuyk (approved):
Hi @jjvraw nice to see you here again! It looks good to me :D

stdlib/src/builtin/int.mojo
laszlokindrat: +1, but at a minimum we should add a `debug_assert` here for now.

stdlib/src/builtin/int.mojo
```diff
@@ -1099,6 +1148,51 @@ struct Int(
...
+        var x = 0
+        var y = 1
+        var u = 1
+        var v = 0
+
+        while a != 0:
+            var q = b // a
+            x, u = u, x - q * u
+            y, v = v, y - q * v
+            b, a = a, b - q * a
```
laszlokindrat, line 1185: style: Here and elsewhere: could you please choose more expressive variable names?

stdlib/src/builtin/int.mojo
```diff
@@ -1099,6 +1148,51 @@ struct Int(
     # Methods
     # ===-------------------------------------------------------------------===#
 
+    fn mod_inverse(self, m: Self) -> Self:
```
laszlokindrat, line 1151: I think we should not expose this as a stable API:
```suggestion
    fn _mod_inverse(self, m: Self) -> Self:
```

stdlib/src/builtin/int.mojo
jjvraw: Thank you for the review Martin and Laszlo :) I was actually busy writing a discussion regarding this, as this seems to be a lingering around the community. 

I am too for exception raises when it comes to undefined behaviour. Joe mentioned parametric raises in one of your PRs or issues, Martin, which I believe would be a strategy. This also seems to allow oppurtunity for good documentation and a standard to follow. 

I had been following the style of `__pow__(self, exp: Self)`, of which returns zero on exception. Shall I go ahead with the `debug_assert`? Both here and on the coprime issue, for now? Or should this be put on hold until a discussion?

laszlokindrat: It's unclear when we will have parametric raises, and until then, the precedent dictates that we prioritize performance for these cases, and use `debug_assert`. So yes, please go ahead, but also feel free to start/contribute to discussion regarding this, we love having community discussions!

docs/changelog.md
```diff
@@ -350,6 +350,17 @@ future and `StringSlice.__len__` now does return the Unicode codepoints length.
   # "Mojo 'Mojo'"
   ```
 
+- Modular exponentation is now supported via `math.pow`
```
laszlokindrat, line 353: Please mention that this is the case for `Int` only (for now).

laszlokindrat:
@abduld can you PTAL?

laszlokindrat:
Pinging @abduld again: any concerns with this?

jjvraw:
Does this [commit](https://github.com/modularml/mojo/commit/2ade90314254c9d1799e4c8ee081cc6ff16556aa) perhaps have any influence on the status of this PR?

laszlokindrat:
@JoeLoser This PR doesn't actually modify the math module, since `pow` is a builtin. How should we proceed?

JoeLoser:
> @JoeLoser This PR doesn't actually modify the math module, since `pow` is a builtin. How should we proceed?

Let's move forward with this.  Thanks for the ping.

JoeLoser:
!sync

JoeLoser:
> Does this [commit](https://github.com/modularml/mojo/commit/2ade90314254c9d1799e4c8ee081cc6ff16556aa) perhaps have any influence on the status of this PR?

After some discussion internally, we're going to hold off and not accept this PR as is.  Rationale:
- While it doesn't make big changes to the `math` module, it would create a bad precedent to accept this given that we've documented that we don't welcome changes to the `math` module right now
- The current implementation is not very performant.

We're happy to spend engineering time implementing a more performant version if an internal use case for it comes up.  But until then, I'd recommend implementing this functionality as a wrapper/free function in your codebase if needed.

I appreciate your enthusiasm and I'm sorry we can't accept this PR as is.

jjvraw:
Sure, no problem. This PR was directed towards #2867, perhaps this issue should be updated now

PR #3315, opened 2024-07-26
[stdlib] Extend tests for `SIMD.__pow__`

jjvraw:
Partially addresses Issue #2867

### DIFF
--- a/stdlib/test/builtin/test_simd.mojo
+++ b/stdlib/test/builtin/test_simd.mojo
@@ -1257,26 +1257,149 @@ def test_reduce_bit_count():
 
 
 def test_pow():
+    alias nan = FloatLiteral.nan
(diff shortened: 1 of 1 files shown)


laszlokindrat (commented):
Thanks, generally looks great, and welcome to Mojo! :) Looks like you have a lot of tests for the `__pow__` overload that takes an `Int`, but no unit tests where the power is another non-scalar SIMD vector. Could you please add some, or maybe modify some of the existing ones? Also, it would be good to have some tests involving `neg_zero`.

jjvraw:
Hi :) 

Added a few additional test cases, and removed a few unnecessary ones. Happy to adjust further if needed.

laszlokindrat (approved):
Great, thank you!

laszlokindrat:
!sync

PR #3323, opened 2024-07-27
[stdlib] Add `has_fma` function to `sys` package

leandrolcampos:
This PR adds a new function `has_fma` to the `sys` package. This function checks if the host system supports Fused Multiply-Add (FMA) instructions.

**Justification**:

The `has_fma` function is important for developers implementing math functions. Numerical algorithms can adopt different strategies depending on whether the target architecture supports FMA or not. For example, the implementation of the `log10` function in the LLVM libc library considers the presence of FMA support.

**Note on Unit Testing:**

~~I have not added a unit test for this function because I am unsure how to implement it. I searched for unit tests for similar functions, such as `has_avx`, but could not find any. If anyone has suggestions on how to write a unit test for this function, they are welcome.~~

As suggested by @laszlokindrat [here](https://github.com/modularml/mojo/pull/3306#issuecomment-2254066284), I added a basic test for this function to ensures it exists and returns a boolable value. This test prevents accidental deletion or changes but does not validate the correctness of the `has_fma` output. If anyone has suggestions on how to write a better test for this function, they are welcome.

**Replacement**

This PR replaces PR #3306.

### DIFF
--- a/stdlib/src/sys/__init__.mojo
+++ b/stdlib/src/sys/__init__.mojo
@@ -22,6 +22,7 @@ from .info import (
     has_avx,
     has_avx2,
     has_avx512f,
+    has_fma,
     has_intel_amx,
     has_neon,
     has_neon_int8_dotprod,
--- a/stdlib/src/sys/info.mojo
+++ b/stdlib/src/sys/info.mojo
@@ -108,6 +108,22 @@ fn has_avx512f() -> Bool:
 
 
 @always_inline("nodebug")
+fn has_fma() -> Bool:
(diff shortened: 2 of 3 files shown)


leandrolcampos:
Hi, @laszlokindrat.

I had to replace the old PR (#3306) with this one because it seems there was an issue when I synced my local branch with `upstream/nightly`.

I added a test for `has_fma` (and similar functions) as you suggested.

JoeLoser (approved):

JoeLoser:
!sync

PR #3327, opened 2024-07-29
[stdlib] Introduce `random.shuffle` for `List`

jjvraw:
Introduce `random.shuffle` for `List`. Implementation follows the Fisher-Yates shuffle.


jjvraw:
For concerns regarding the Fisher-Yates algorithm, one can refer to a corresponding [issue](https://github.com/python/cpython/issues/108598) in the CPython repository.  In summary, the Python community appears to be satisfied with this implementation.

stdlib/src/random/random.mojo
```diff
@@ -201,3 +201,28 @@ fn randn[
     for i in range(size):
         ptr[i] = randn_float64(mean, variance).cast[type]()
     return
+
+
+fn shuffle[T: CollectionElement](inout list_in: List[T]):
```
soraros, line 206:
```suggestion
fn shuffle[T: CollectionElement, //](inout list: List[T]):
```

stdlib/src/random/random.mojo
```diff
@@ -201,3 +201,28 @@ fn randn[
...
+    Args:
+        list_in: The list to modify.
+
+    Parameters:
+        T: The type of element in the List.
+    """
+
+    var length = len(list_in)
+
+    for i in range(length - 1, 0, -1):
```
soraros, line 220:
```suggestion
    for i in reversed(range(length)):
```

stdlib/src/random/random.mojo
```diff
@@ -201,3 +201,28 @@ fn randn[
...
+        var j: Int = random_ui64(0, i).cast[DType.int64]().value
+
+        var i_ptr = list_in.data.offset(i)
```
soraros, line 223: I think the `offset` method is introduced for compatibility, we'd better use `+` instead.

stdlib/src/random/random.mojo
```diff
@@ -201,3 +201,28 @@ fn randn[
...
+        var j_ptr = list_in.data.offset(j)
+
+        var tmp = i_ptr.take_pointee()
```
soraros, line 226: Can we use `swap` here?

stdlib/src/random/random.mojo
jjvraw: Ah, I was wondering the discrepancy between `+` and `offset`. Thank you.

soraros:
Can we set a seed to make the test deterministic?

stdlib/src/random/random.mojo
soraros: I actually prefer `offset`, but Connor suggested `+` in this [comment](https://github.com/modularml/mojo/pull/3203#discussion_r1672630293).

stdlib/src/random/random.mojo
```diff
@@ -201,3 +201,28 @@ fn randn[
...
+        list_in: The list to modify.
+
+    Parameters:
+        T: The type of element in the List.
+    """
+
+    var length = len(list_in)
+
+    for i in range(length - 1, 0, -1):
+        var j: Int = random_ui64(0, i).cast[DType.int64]().value
```
soraros, line 221: Can we simply write the following?
```suggestion
        var j = int(random_ui64(0, i))
```

stdlib/src/random/random.mojo
jjvraw: Why inferred here?

stdlib/src/random/random.mojo
jjvraw: Unfortunately not.

stdlib/src/random/random.mojo
soraros: IIUC, this is the direction the stdlib is moving towards. We won't need to manually specify this type parameter anyways.

stdlib/src/random/random.mojo
soraros: What about `swap(i_ptr[], j_ptr[])`?

soraros: Actually, `swap(list.unsafe_get(i), list.unsafe_get(j))` could be even better.

jjvraw:
Are you suggesting the same seed for all tests? Is the concern about potential failing assertions due to randomness? 

It might be beneficial to introduce some seeded tests, or perhaps modify some existing ones to seeded. Currently we assert that a successfully shuffled list relies on an __extremely__ low chance of a shuffled list resulting in the same order to assert success. There is more confidence that a failed test case indicates an ineffective shuffle. Alternatively, using property tests, followed by sorting and checking for equality with the original list, might be more reliable. What do you think?

stdlib/src/random/random.mojo
jjvraw: Very nice!

soraros:
I don't have a very strong opinion about this, and if we were indeed to move to explicit seeding, we definitely don't have to do it in this PR. Just taking the chance to bring it up.

My point is mainly: if the tests are deterministic, we can simplify them into something like this:
```mojo
seed()
var l = List(1, 2, 3, 4)
shuffle(l)
assert_equal(l, List(2, 3, 1, 4))  # we know beforehand the result we will get
``

File: stdlib/src/random/random.mojo [hunk 6]
```diff
@@ -201,3 +201,19 @@ fn randn[
...
+
+    Args:
+        list: The list to modify.
+
+    Parameters:
+        T: The type of element in the List.
+    """
+    for i in reversed(range(len(list))):
+        var j = int(random_ui64(0, i))
+        swap(list.unsafe_get(i), list.unsafe_get(j))
```
JoeLoser, line 219: **Question** Should we use the new `List.swap_elements` function instead for exclusivity checking purposes?

stdlib/test/random/test_random.mojo
```diff
@@ -73,6 +80,94 @@ def test_seed():
...
+    var a = L_i(1, 2, 3, 4)
+    var b = L_i(1, 2, 3, 4)
+    var c = L_s("Random", "shuffle", "in", "Mojo")
+    var d = L_s("Random", "shuffle", "in", "Mojo")
+
+    shuffle(b)
+    assert_equal(len(a), len(b))
+    assert_true(a != b)
+    for i in range(len(b)):
+        assert_true(a.__contains__(b[i]))
```
JoeLoser, line 98: **Question** Can we safely use the `in` sugar syntax since this test isn't interested in testing dunder contains directly?

stdlib/src/random/random.mojo
jjvraw: Sounds good, missed the introduction of this.

JoeLoser (approved):

JoeLoser:
!sync

PR #3345, opened 2024-08-01
[stdlib] Remove `String` constructor from `PythonObject`

soraros:
It's removed to avoid implicit conversion from e.g. `Int` through `PythonObject`.


JoeLoser:
!sync

JoeLoser (approved):

PR #3347, opened 2024-08-01
[stdlib] Use infer only in `bit`

soraros:
Also changed parameter names to be more uniform.


JoeLoser (approved):
Looks great, thanks!

JoeLoser:
!sync

PR #3348, opened 2024-08-01
[stdlib] Refactor `format_int.mojo`

helehex:
Refactor `format_int.mojo` to use Formatter, and take care of corresponding TODO's.

### DIFF
--- a/stdlib/src/builtin/format_int.mojo
+++ b/stdlib/src/builtin/format_int.mojo
@@ -18,7 +18,9 @@ These are Mojo built-ins, so you don't need to import them.
 
 from os import abort
 from collections import List, Optional, InlineArray
-from utils import StringSlice, StaticString
+from collections.string import _calc_format_buffer_size
+from utils import StringSlice, StaticString, Variant
+
 
 alias _DEFAULT_DIGIT_CHARS = "0123456789abcdefghijklmnopqrstuvwxyz"
 
@@ -28,7 +30,37 @@ alias _DEFAULT_DIGIT_CHARS = "0123456789abcdefghijklmnopqrstuvwxyz"
 # ===----------------------------------------------------------------------===#
 
 
-fn bin(num: Scalar, /, *, prefix: StaticString = "0b") -> String:
+fn bin[
+    WriterType: Writer, //
+](value: Scalar, /, *, inout writer: WriterType, prefix: StaticString = "0b"):
+    """Writes the binary string representation an integral value to a formatter.
+
+    ```mojo
+    print(bin(123))
+    print(bin(-123))
+    ```
+    ```plaintext
+    '0b1111011'
+    '-0b1111011'
+    ```
+
+    Parameters:
+        WriterType: The type of the `writer` argument.
+
+    Args:
+        value: An integral scalar value.
+        writer: The formatter to write to.
+        prefix: The prefix of the formatted int.
+    """
+
+    @parameter
+    if value.type is DType.bool:
+        _write_int(writer, value.cast[DType.int8](), 2, prefix=prefix)
+    else:
+        _write_int(writer, value, 2, prefix=prefix)
+
+
+fn bin(value: Scalar, /, *, prefix: StaticString = "0b") -> String:
     """Return the binary string representation an integral value.
 
     ```mojo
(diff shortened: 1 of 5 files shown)


helehex:
Oh I forgot about something.

helehex:
Ok, I added those tests back in.

stdlib/src/builtin/object.mojo
```diff
@@ -591,9 +591,7 @@ struct _ObjectImpl(
             )
             return
         if self.is_func():
-            writer.write(
-                "Function at address " + hex(int(self.get_as_func().value))
-            )
```
ConnorGray, line 596: **Question** Is there a way we might be able to keep the `hex(..)` formatting here?

ConnorGray: **Followup** Never mind, I see you've changed the formatting of `UnsafePointer` to use `hex` by default as well.

That seems reasonable to me, though it would be great to call that out in the description as well :)

ConnorGray:
!sync

JoeLoser:
FYI this is still hitting some compiler bugs internally in the elaborator, so added the "blocked" label on this PR for now.

JoeLoser:
@helehex do you mind rebasing this since @jackos recently overhauled the writer/formatter stuff and we can see what else remains here?  Thanks!

helehex:
yup i'll get this working again today.

stdlib/src/builtin/format_int.mojo
```diff
@@ -332,8 +425,6 @@ fn _try_write_int[
         )
         writer.write(zero)
 
-        return None
```
soraros, line 335: I think it should now be `return` since it's an early return.

stdlib/src/builtin/object.mojo
```diff
@@ -572,29 +572,25 @@ struct _ObjectImpl(
...
-                + str(
-                    StringRef(
-                        self.get_as_string().data, self.get_as_string().length
-                    )
-                )
-                + "'"
+                "'",
+                StringRef(
+                    self.get_as_string().data, self.get_as_string().length
+                ),
```
soraros, line 588:
```suggestion
                self.get_as_string(),
```

stdlib/src/builtin/format_int.mojo
helehex: Can you explain further? there's nothing after the return, and the function itself doesn't return, so it should be unnecessary right?

stdlib/src/builtin/format_int.mojo
```diff
@@ -28,7 +28,37 @@ alias _DEFAULT_DIGIT_CHARS = "0123456789abcdefghijklmnopqrstuvwxyz"
 # ===----------------------------------------------------------------------===#
 
 
-fn bin(num: Scalar, /, *, prefix: StaticString = "0b") -> String:
+fn bin[
+    WriterType: Writer, //
+](value: Scalar, /, *, inout writer: WriterType, prefix: StaticString = "0b"):
```
soraros, line 33: Nit: make `writer` the last arg so it's consistent with `print`.

stdlib/src/builtin/format_int.mojo
helehex: oh i notice that the original does return... i don't remember changing that, maybe it was something upstream. I'm guessing theres a reason why it returns `Optional[Error]` instead of raising then?

soraros: I think `Optional[Error]` is like `Result[Void]`.

stdlib/src/builtin/object.mojo
helehex: I was also thinking about that, should of tried it. I wonder why there was so much indirection before...

helehex: Tests failed, maybe it was the formatting, hang on

soraros: It's really that `get_as_string` doesn't return a `String` but `_ImmutableString`.

helehex: Oh yeah, it's because it return an `_ImmutableString`

helehex: So the original should be fine and not do an extra allocation.

stdlib/src/builtin/format_int.mojo
helehex: hmm I might like that as well, hang on.

helehex:
Is this still blocked? maybe because of the `Optional[Error]` thing?

stdlib/src/builtin/format_int.mojo
helehex: Mojo doesn't let us put a required keyword-only argument after an optional keyword-only argument

JoeLoser:
> Is this still blocked? maybe because of the `Optional[Error]` thing?

I'll re-sync and circle back.

JoeLoser:
!sync

JoeLoser:
This is still crashing internally.  Can you add a test for `print(some_str)` where `some_str` is meant to be an alias/compile-time thing (the crash is in failing to internally comp-time eval the call when doing the `write_to` functions like you modified in this PR)?  It'd be good to expose the failure in the OSS tests which I think we ought to be able to do without too much effort.

helehex:
yeah i think i see what you mean

helehex:
`_snprintf()` also fails running at comp time (or maybe one of the comp time allocation issues) which causes other problems.

helehex:
hmm not sure how that `_try...` got flipped around either.

helehex:
Seems like ctime str was already broken in the current mojo version for everything i changed (SIMD, object, UnsafePointer)

helehex:
Is the plan to stop using `_snprintf()` eventually?

JoeLoser (approved):

JoeLoser:
> Seems like ctime str was already broken in the current mojo version for everything i changed (SIMD, object, UnsafePointer), because of the issue with `_snprintf`.

Right, because external calls can't be comp-time interpreted right now.  Thanks for digging into it.  This is a known shortcoming right now.

JoeLoser:
@ConnorGray or myself would be happy to review this PR again to see if the crashes have gone away.  You should be able to test this yourself now too since we have Bazel in place and have open sourced the whole kernel library.  Feel free to give it a shot and open up a new PR when you're ready.

PR #3354, opened 2024-08-03
[stdlib] Add `FloatLiteral.{exp, log, pow}`

helehex:
Add methods: 
- `FloatLiteral.__exp__()`
- `FloatLiteral.__log__()`
- `FloatLiteral.__pow__(IntLiteral)`
- `FloatLiteral.__pow__(FloatLiteral)`

### DIFF
--- a/stdlib/src/builtin/float_literal.mojo
+++ b/stdlib/src/builtin/float_literal.mojo
@@ -435,7 +435,95 @@ struct FloatLiteral(
         """
         return rhs // self
 
-    # TODO - maybe __pow__?
+    @always_inline("nodebug")
+    fn __exp__[atol: FloatLiteral = 0.1e-20](self) -> FloatLiteral:
+        """Calculates exponential of `self`.
+
+        Parameters:
+            atol: The absolute tolerance for selecting the desired precision.
+
+        Returns:
+            `e` raised to the power of `self`.
+        """
+        if self.is_nan():
+            return self
+        var result: FloatLiteral = 1.0
+        var count: IntLiteral = 1
+        var term: FloatLiteral = self
+        while abs(term) > atol:
+            result += term
+            count += 1
+            term *= self / count
+        return result
+
+    @always_inline("nodebug")
+    fn __log__[atol: FloatLiteral = 0.1e-20](self) -> FloatLiteral:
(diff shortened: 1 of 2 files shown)


helehex:
The default `atol` gives about ~20 iterations for each `exp` and `pow`. It could probably be improved.

helehex:
Oh, let me restate the tests to match the others in that file, I like them better.

JoeLoser:
Thanks for the contribution!  Do you mind sharing a summary/motivation for the corresponding change?  WDYT of this @jackos?

helehex:
I saw there was a TODO about *maybe* adding these, so I decided to go ahead and implement it

stdlib/src/builtin/float_literal.mojo
```diff
@@ -435,7 +435,95 @@ struct FloatLiteral(
         """
         return rhs // self
 
-    # TODO - maybe __pow__?
+    @always_inline("nodebug")
+    fn __exp__[atol: FloatLiteral = 0.1e-20](self) -> FloatLiteral:
```
lsh, line 439: Thanks for this work. A small nit is that `atol` is a function that acts on numbers, so I think we should spell out the whole word `absolute_tolerance`.

helehex:
I'll give some more context about the impl instead of falling back to the todo.

Some downsides:
Writing math with literals like `10.0**5` could be more expensive than you expect/need, since calls to dunder methods are avoided in user code, you wont specify the precision anyways, so you could say that in most cases you would want the current behavior which returns a `Float64` (even when run at compile time).

Some upsides:
It allows `FloatLiteral` conformance to `Powable`, and makes it more self-consistent with other ops like `FloatLiteral.__add__` by returning `Self`. This allows you to keep the result as a `FloatLiteral` when working with them at compile time, which is other wise annoying to do, and you end up having to write you're own (like i have done).

stdlib/src/builtin/float_literal.mojo
helehex: yeah i agree

JoeLoser:
Given @lattner's recent work on literals and `@always_inline("builtin")`, I don't think we want to pursue this anymore.  Thanks for the thought though!

PR #3358, opened 2024-08-05
[stdlib] Added example to ```Powable``` docstring

jlofti:
Hello,

Wanted to address #2867 point number one. I added an example of demonstrating how ```__pow__``` could be implemented on a ```Rational``` struct.

Thanks!


stdlib/src/builtin/math.mojo
```diff
@@ -262,7 +262,17 @@ trait Powable:
...
+    ```mojo
+    @value
+    struct Rational(Powable):
+        var numerator: Float64
+        var denominator: Float64
+
+        fn __pow__(self,exp: Self)  -> Self:
+            var exp_value = exp.numerator / exp.denominator
+            return Self(pow(self.numerator,exp_value), pow(self.denominator,exp_value))
+    ```
```
jjvraw, line 275: Might be worth giving an example of how this enables the exponentiation operator, `**`. This would make it consistent with the others :) Reference here https://github.com/modularml/mojo/blob/1f9ccc7cc07fa1d70c02ba6df3bc7b92f54c85fb/stdlib/src/builtin/value.mojo#L19-L48 

Also, a minor formatting nit: it would be clearer to space the parameters.

jlofti: Thanks for the feedback! Should all be addressed.

jlofti: Hi was also wondering, this is my first time contributing here. Should I mark this as resolved or do reviewers mark? Thanks!

JoeLoser: Looks good to me, thanks! We don't have anything formal yet around "who resolves the conversations" on the PR.  I've synced this PR internally and it'll land here soon.

JoeLoser (approved):

JoeLoser:
Thanks for the contribution! 🎉 

I've synced this PR internally and it'll land in today's nightly release! 🔥

JoeLoser:
!sync

PR #3368, opened 2024-08-07
[stdlib] Improve `InlineArray` Initialization

helehex:
Add `DefaultableCollectionElementNew` trait.
Add default initialization to `InlineArray`
Change to using a parameter to specify uninit data in `InlineArray`.
Moved explicit copy initializer to avoid bug causing the type to not conform to `ExplicitlyCopyable`.

It seems better to use a parameter for marking the data to be left uninitialized, and `self.__init__[False]()` captures the unsafeness well.
Also since we have a defaultable trait, you don't really have to worry about the default initializer anymore as long as your element types have a default initializer, or you want more control.

### DIFF
--- a/stdlib/src/collections/inline_array.mojo
+++ b/stdlib/src/collections/inline_array.mojo
@@ -10,7 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
-"""Defines the `InlineArray` type.
+"""Implements InlineArray, a statically-sized uniform container.
 
 You can import these APIs from the `collections` package. For example:
 
@@ -23,8 +23,9 @@ from collections._index_normalization import normalize_index
 from sys.intrinsics import _type_is_eq
 from memory.maybe_uninitialized import UnsafeMaybeUninitialized
 
+
 # ===----------------------------------------------------------------------===#
-# Array
+# Inline Array
 # ===----------------------------------------------------------------------===#
 
 
@@ -36,7 +37,7 @@ fn _inline_array_construction_checks[size: Int]():
     Parameters:
       size: The number of elements.
     """
-    constrained[size > 0, "number of elements in `InlineArray` must be > 0"]()
+    constrained[size >= 0, "number of elements in `InlineArray` must be >= 0"]()
 
 
 @value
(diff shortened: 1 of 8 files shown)


gabrieldemarmiesse:
It seems this pull request is in direct conflict with https://github.com/modularml/mojo/pull/3122 . The general idea is that it can never be correct to use an uninitialized `InlineArray` unless you are using it with `UnsafeMaybeUninitialized`. See those comments which explains it well: 
* https://github.com/modularml/mojo/issues/2869#issuecomment-2137962617
* https://github.com/modularml/mojo/pull/2965#discussion_r1648153449
* https://github.com/modularml/mojo/pull/3119#discussion_r1653289653

Once  https://github.com/modularml/mojo/pull/3122 is merged, we can maybe add the `Defaultable` trait from your PR, but since it's going to be a conditional trait (the element type **must** be `UnsafeMaybeUninitialized`), I'm not sure the compiler is able to do this yet.

helehex:
Why does it seems that way? shouldn't both work fine?

helehex:
The pr also adds a line to the docstring that says you must manually initialize the elements when using the unsafe constructor

helehex:
Also, doing `var a = InlineArray[NonDefaultableType, x]()` gives you an error in the editor with this change, because you *must* provide the parameter in order to initialize the array with uninitialized data.

helehex:
I'll check it out later and see if there's a problem, I'm not familiar with the maybe uninitialized pattern, so i'll have to explore it a bit to understand.

helehex:
My bad, not sure how i didn't test that good enough, but the empty initializer is not fully constrained to give an error in the editor. 
I'll update the pr to be just for the `Defaultable` parts.

helehex:
I deprecated the default initializer when the element types are non-defaultable, to give a hint to the user to use something else.

stdlib/src/collections/inline_array.mojo
```diff
@@ -10,7 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
-"""Defines the `InlineArray` type.
+"""Implements InlineArray, a statically-sized uniform container.
```
gryznar, line 13: I don't know if it was intentional, but bacticks \` are missing

ConnorGray:
@helehex this PR has gotten a bit out of date with changes over the last few months. Are there changes here you'd still like to see us incorporate? I'm uncertain how much of this PR would be done differently with the language now supporting things like trait compositions and `__list_literal__` initializers.

I'm going to close this PR to simplify our queue, but we're happen to reopen if you're still wanting to push on it 🙂

PR #3372, opened 2024-08-09
[stdlib] assertion overloads for `List`

jjvraw:
Add `assert_equal` and `assert_not_equal` overload for `List` comparisons, a nice-to-have in a few places. Most probably a temporary solution under the current trait system.

### DIFF
--- a/stdlib/test/testing/test_assertion.mojo
+++ b/stdlib/test/testing/test_assertion.mojo
@@ -28,6 +28,28 @@ from builtin._location import _SourceLocation
 from python import PythonObject
 
 
+def test_assert_messages():
+    try:
+        assert_true(False)
+    except e:
+        assert_true("test_assertion.mojo:33:20: AssertionError:" in str(e))
+
+    try:
+        assert_false(True)
+    except e:
+        assert_true("test_assertion.mojo:38:21: AssertionError:" in str(e))
+
+    try:
+        assert_equal(1, 0)
+    except e:
+        assert_true("test_assertion.mojo:43:21: AssertionError:" in str(e))
+
+    try:
+        assert_not_equal(0, 0)
+    except e:
+        assert_true("test_assertion.mojo:48:25: AssertionError:" in str(e))
+
+
 @value
 struct DummyStruct:
     var value: Int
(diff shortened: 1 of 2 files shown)


stdlib/test/testing/test_assertion.mojo
```diff
@@ -28,6 +28,28 @@ from builtin._location import _SourceLocation
...
+
+    try:
+        assert_equal(1, 0)
+    except e:
+        assert_true("test_assertion.mojo:43:21: AssertionError:" in str(e))
+
+    try:
+        assert_not_equal(0, 0)
+    except e:
+        assert_true("test_assertion.mojo:48:25: AssertionError:" in str(e))
```
jjvraw, line 50: Moved here to remove dependency on line numbers and future changes

helehex:
Thanks, I just had to add something like this to one of my repos, except for `List[List[Int]]`. I can see this being useful for the time being. It would simplify tests in `/test_list.mojo`
One suggestion: add a TODO to remove this once we have associated types or more powerful conditional conformance that can infer conformance based on methods which use a inferred parameter to conditionally conform to the trait.

helehex:
I wish someone would just add the str overload back in. It's easier to track a single TODO on an overload, than to track everywhere where you are unable to use a method because of lacking features.

helehex (approved):
Thanks, I just added something very similar to one of my repos, except for `List[List[Int]]`. I can see this being useful for the time being. It would simplify tests in `/test_list.mojo`

One suggestion:
Add a TODO to remove this once we have associated types or more powerful conditional conformance that can infer conformance based on methods which use a inferred parameter to conditionally conform to the trait.

jjvraw:
Cool. There are two PRs atm that would benefit from this as well. 

> Add a TODO to remove...

There's a comment right above the other overloads, will modify that.

> repos, except for `List[List[Int]]`

I actually opened a discussion relevant to this issue, #3357, if you have any opinions :)

helehex:
Oh yeah, i remember a comment in there somewhere, i guess my PR which removes the need for the SIMD overload hasn't gotten merged yet.

JoeLoser:
!sync

PR #3373, opened 2024-08-11
[stdlib] Fix TODO item about using os.path.split() in tempfile test

msaelices:
Fix pending TODO item about using `os.path.split()` in a `tempfile` test


JoeLoser (approved):
Thanks!

JoeLoser:
!sync

PR #3375, opened 2024-08-11
[stdlib] Fix TODO in `/io.mojo`.

helehex:
Maybe there's an internal reason this couldn't be done, but i'll open it so it can at least be tracked.


JoeLoser:
!sync

PR #3378, opened 2024-08-13
[stdlib] Add proposal for hardware time stamp counter support

owenhilyard:
Hardware time stamp counters (TSCs) are present on most modern architectures and offer a cheaper way to get timestamps than other methods. This is useful for benchmarking, logging, and more exotic usecases.

This proposal includes an API which should meet most needs, as well as how to handle the absence of a reliable TSC.

### DIFF
--- a/examples/life/magic.lock
+++ b/examples/life/magic.lock
@@ -17,7 +17,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aom-3.9.1-hac33072_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/attr-2.5.1-h166bdaf_1.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/noarch/attrs-24.3.0-pyh71513ae_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-auth-0.8.0-hb921021_15.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-auth-0.8.0-h205f482_16.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-cal-0.8.1-h1a47875_3.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-common-0.10.6-hb9d3cd8_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-compression-0.3.0-h4e1184b_5.conda
@@ -26,10 +26,10 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-io-0.15.3-h831e299_5.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-mqtt-0.11.0-h11f4f37_12.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-s3-0.7.7-hf454442_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-sdkutils-0.2.1-h4e1184b_4.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-sdkutils-0.2.2-h4e1184b_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-checksums-0.2.2-h4e1184b_4.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-crt-cpp-0.29.7-hd92328a_7.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-sdk-cpp-1.11.458-hc430e4a_4.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-crt-cpp-0.29.8-h8570fcd_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-sdk-cpp-1.11.458-h7001638_5.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/azure-core-cpp-1.14.0-h5cfcd09_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/azure-identity-cpp-1.10.0-h113e628_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/azure-storage-blobs-cpp-12.13.0-h3cf044e_1.conda
@@ -99,10 +99,10 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/ld_impl_linux-64-2.43-h712a8e2_2.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/lerc-4.0.0-h27087fc_0.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libabseil-20240722.0-cxx17_hbbce691_4.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-18.1.0-hd595efa_7_cpu.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-acero-18.1.0-hcb10f89_7_cpu.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-dataset-18.1.0-hcb10f89_7_cpu.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-substrait-18.1.0-h08228c5_7_cpu.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-18.1.0-h9d9f30d_8_cpu.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-acero-18.1.0-hcb10f89_8_cpu.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-dataset-18.1.0-hcb10f89_8_cpu.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libarrow-substrait-18.1.0-h08228c5_8_cpu.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libasprintf-0.22.5-he8f35ee_3.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libasprintf-devel-0.22.5-he8f35ee_3.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libavif16-1.1.1-h1909e37_2.conda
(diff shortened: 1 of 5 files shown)


owenhilyard:
For clarity, this is submitted as a proposal because implementing it will add code that isn't portable (I have fallbacks planned) and thus violates the "just send a PR" part of the contributor guidelines.

JoeLoser:
Thanks for the proposal!  Sorry for the delay in reviewing this so far.  We briefly talked async about this internally, but @ConnorGray let's add this to the design discussion meeting agenda for today if possible.

owenhilyard:
@JoeLoser 
I have a good chunk of the implementation written now. What's currently blocking me updating the proposal with an implementation is that I'm not sure how to write a sound implementation of cpuid for detecting the frequency on Intel Xeons. sys._assembly.inlined_assembly doesn't appear to have a good way to support multi-output instructions (since cpuid has 4 outputs), or perhaps I'm using it wrong. I see a lot of ICEs when experimenting so I know it's not well traveled ground. AMD does have a reliable method (hidden behind far too much digging), which is that `/sys/devices/system/cpu/cpu0/acpi_cppc/nominal_freq` gives you the correct frequency on Linux, so I'm going to move that up to the default way to get it for AMD, meaning that Intel CPUs would be the only ones falling back to calibrating against the OS timer. 

I also don't have a great way to test on ARM aside from bringing an old Jetson Nano back from the dead, so I'd appreciate some help there. That implementation is at least straightforward, so I expect it to work properly with only minor fixes. 

What I need from modular is:
1. A version of clang's `__cpuid` I can call into. If you can hand me the 4 registers I can do the rest but from experimenting today neither Clang nor LLVM IR inline asm seem to work, either causing ICEs or giving me zeroed registers.
2. Where do you want arch specific code put? Are sys._arch.x86_64 and sys._arch.arm ok? 
3. Do you want the perf_counter_ns implementation for CUDA read grouped under this API? It's also a TSC, but I can't easily provide a converstion to real timestamps for it without doing "NTP over PCIe" to calibrate the clocks, which would make it the most expensive timer to initialize by far. The benefit would be that once initialized the timestamp could be trivially shared between kernels, meaning it's a one-time cost in well-written programs. The downside is that these timestamps will be rough at best. 
4. What kinds of warnings you want me to cover this with since it's not reliable on many laptops and might not be fully reliable on some types of VMs (AWS and Azure work fine, VMWare requires special handling, others unknown). 
5. Whether you want me to add a "magic define" which makes perf_counter_ns use the TSC functions with a reference point set by the compiler before `main()`. If that is desirable, I'll offer `__MODULAR_MOJO_STDLIB_TSC_PERF_COUNTER` as a a potential name for "use the TSC api", and `__MODULAR_MOJO_STDLIB_TSC_PERF_COUNTER_FORCE_UNSAFE` for "use it even if the kernel says it's unreliable" given that they should be sufficiently namespaced to avoid conflicts and sufficiently scary to keep people away from them if they don't know what they're doing. I'll document that magic define alongside all of the warnings about what needs to be done to make it reliable. At the moment, I think that doing this would require either doing very odd things or compiler support since I don't think Mojo has linker attributes anywhere.

proposals/time-stamp-counter-support.md
```diff
@@ -0,0 +1,274 @@
...
+to run?" or for places where a timestamp that will only be compared to the same
+clock needs to be acquired cheaply, such as in a hot loop. They are also
+monotonic, meaning that they are unaffected by changes to system time such as
+leap seconds. They are widely supported, with more recent architectures making
+their usage easier (ex: x86 requires lookup tables to get the frequency, ARM
+and RISC-V make you read a register).
+
+### Why use Time Stamp Counters directly?
+
+TSCs give you a much cheaper measurment of time than with most other methods.
```
JoeLoser, line 20:
```suggestion
TSCs give you a much cheaper measurement of time than with most other methods.
```

proposals/time-stamp-counter-support.md
```diff
@@ -0,0 +1,274 @@
...
+        """
+        ...
+
+    fn get_raw(borrowed self) -> UInt64:
+        """
+        Access the raw, architecture dependent value of the TSC. 
+        """
+        return self._tsc_value
+
+struct Tsc:
```
JoeLoser, line 159: **Suggestion** Let's spell this out fully for readability sake: `TimestampCounter` (from internal review)

proposals/time-stamp-counter-support.md
```diff
@@ -0,0 +1,274 @@
...
+    (typically the platform epoch). Used to aid in converting TscTimestamp 
+    instances to timestamps
+
+    start_tsc_timestamp: A tsc timestamp taken right after the timestamp 
+    syscall call returned. 
+
+    tsc_is_reliable: Whether the TSC is usable as a time source. If false, a monotonic but fast to query OS clock is used.
+    """
+
+    var freq: TscFrequency
```
JoeLoser, line 182: **Suggestion** Have you considered lifting the frequency and `tsc_is_reliable` out of this struct?  This will be the de-facto type passed around, so we'd want to be mindful of size and layout with respect to properties that conceptually are static for a system and don't differ between two different timestamp instances.

proposals/time-stamp-counter-support.md
```diff
@@ -0,0 +1,274 @@
...
+
+    fn get_tsc_timestamp(borrowed self) -> TscTimestamp:
+        """
+        Read from the TSC (or fallback clock). 
+        """
+        pass
+
+    fn now(borrowed self) -> Time:
+        """
+        To be added once the standard library decides how to handle time.
```
JoeLoser, line 205: **Question** Have you given much thought on how this work ties into a `chrono` module and what that may look like with regards to this lower-level pre-requisite work?

proposals/time-stamp-counter-support.md
```diff
@@ -0,0 +1,274 @@
...
+### High precisions latency measurements
+
+In networked applications it is often desirable to obtain as precise as
+possible of a latency measurement. This is used for setting service level
+agreements (SLAs) as well as for reasoning about the correctness of some
+more exotic systems such as Google's Spanner database. The latency test
+harnesses for these systems often have hundreds or thousands of messages
+in flight at a time and want to make getting a timestamp as cheap as possible.
+
+## Why the Standard Library?
```
JoeLoser, line 262: +1 for putting it in Mojo standard library.  There's some prior art with Folly for example [here](https://github.com/facebook/folly/blob/9c9430777ac2041a3d9faa9d87862f1e35e42d19/folly/chrono/Hardware.h)

proposals/time-stamp-counter-support.md
owenhilyard: The frequency is the most expensive value to determine on some architectures:

| Arch | Cost |
|--------|--------|
| aarch64 | MSR Read |
| RISC-V | MSR Read |
| AMD x86-64 | At least 1 syscall, parsing an int from a string (Linux) |
| Intel x86-64 | Traversal through CPUID and a variety of lookup tables |

There are two ways I can lift it out to a global. The first is to just store one, which will break ancient x86 CPUs (pre-
Nehalem on Intel, and I'm fairly sure pre-Zen for AMD), due to drift between cores/sockets, on ARM the architecture requires a single clock, so that's fine, and RISC-V is similar for most implementations. The other option is to store a frequency for each CPU core on x86, which also means every Mojo program has to move itself between CPU cores during runtime init. That is not very desirable, so if dropping very old CPUs is fine I'll go with using a global variable for frequency, a base TSC timestamp and a WALL time. 

If I do that and have the runtime do the reliability check before main (to allow for forcing it off using an environment variable if someone's environment doesn't have good TSC), then these can all become free-standing functions and the `Tsc` struct goes away.

proposals/time-stamp-counter-support.md
owenhilyard: My plan was to store a unix epoch timestamp in nanoseconds (ideally in the TAI timezone) to use to convert TSC timestamps to real time. There's not a lot of value vs calling `clock_gettime()` on Linux there unless you already have TSC readings you want to convert to real time since VDSO makes `clock_gettime()` cheap, but I'd need to check the MacOS and Windows behaviors to see if this would be useful for timekeeping. 

Due to the number of hardware/firmware/hypervisor bugs that TSCs seem to have, I don't think backing a `chrono` module with them by default is a good idea, but allowing a conversion to a type from `chrono` if the user asks for it should be fine.

owenhilyard:
@JoeLoser Sorry about the delay, you got back to me while I was at a conference and I've been catching up on work since then. 

Getting builtins for `__get_cpuid_max`, `__get_cpuid`, and `__get_cpuid_count` should cover most usecases in this area, and enable runtime feature detection in std. I can port over the constants. It might be that I'm using the inline asm incorrectly and that it is doable in pure Mojo as well. 

Let me know if dropping support for very old x86 in the name of making the `Tsc` struct go away is fine.

proposals/time-stamp-counter-support.md
owenhilyard: ACK

proposals/time-stamp-counter-support.md
owenhilyard: ACK

owenhilyard:
@JoeLoser  Per the newly ratified RVA23 profile for RISC-V, time stamp counters will also be available in the RVA23U64 profile, which is intended as a baseline for higher-powered devices like servers. It looks like this should be widely portable since ARM also forces a TSC for their server chips. 

https://drive.google.com/file/d/12QKRm92cLcEk8-5J9NI91m0fAQOxqNAq/view

proposals/time-stamp-counter-support.md
JoeLoser: It's fine to drop very old CPUs, so I'd go recommend we go that direction!

Can you help me understand more about needing to do the reliability check all **before main**?  Why does it need to happen before main and not, say, at the very beginning?

JoeLoser:
> Let me know if dropping support for very old x86

Dropping support for very old CPUs is fine - I just responded to the other comment.  If you need specific builtins from us, do you mind filing a follow-up GH issue and we can take a look?  I imagine most of these are just light wrapper around some LLVM code that already exists upstream.

owenhilyard:
> > Let me know if dropping support for very old x86
> 
> Dropping support for very old CPUs is fine - I just responded to the other comment. If you need specific builtins from us, do you mind filing a follow-up GH issue and we can take a look? I imagine most of these are just light wrapper around some LLVM code that already exists upstream.

They are, I'll open up an issue.

proposals/time-stamp-counter-support.md
owenhilyard: > Can you help me understand more about needing to do the reliability check all **before main**? Why does it need to happen before main and not, say, at the very beginning?

The idea was to set the "is_tsc_reliable" flag before the user has a chance to run code that would read it to avoid races. The main downside of this approach is that under some circumstances I have to fall back to "calibrate using the system timer" to determine what the frequency of the TSC is. That involves a sleep (probably up to 100ms) and some syscalls. 

I could also it so that there's a global tsc_frequency (assuming that `KGEN_CompilerRT_GetGlobalOrNull` is fairly cheap) that is left as 0 if the TSC is uninitalized and 1 if the TSC is unreliable (since a 1 Hz timer is useless anyway) and initialize it on first use along with providing a memory fence, then this should be fine if it's only used when converting a TSC timestamp to real time. 

```mojo
fn get_tsc_freq_hz() -> UInt64:
    var tsc_freq = _get_global_or_null["__tsc_frequency"]().bitcast[UInt64]()
    if tsc_freq != UnsafePointer[UInt64]():
        if tsc_freq[] > 1:
           return tsc_freq[]
        else:
          mfence()
          if tsc_freq[] > 1:
            return tsc_freq[]
          else:
              return init_tsc()
  else:
    return init_tsc()
```

If you want accelerators supported by this, then I'll stick the current target arch name on the end of `__tsc_frequency`.

PR #3381, opened 2024-08-15
[stdlib] Add `sqrt(IntLiteral) -> IntLiteral`.

helehex:
I know there are better methods, but it would be nice to have something.
I'm not sure there's a nice way to track compile time benchmarks in the stdlib right now, so that might have to wait.

### DIFF
--- a/stdlib/src/builtin/int_literal.mojo
+++ b/stdlib/src/builtin/int_literal.mojo
@@ -248,6 +248,23 @@ struct IntLiteral(
     # TODO: implement __pow__
 
     @always_inline("nodebug")
+    fn __sqrt__(self) -> Self:
+        """Return the square root of `self`, rounded down if necessary. Negative values return themselves.
+
+        Returns:
+            The square root of `self`.
+        """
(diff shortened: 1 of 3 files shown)


helehex:
I move the tests into `/test_intliteral.mojo` instead of `/test_math.mojo`

JoeLoser:
Can you help provide some context as to why we'd want to use a `__sqrt__` dunder method?  There's no Python precedence for this as @laszlokindrat kindly pointed out to me offline.  _If_ we were to define a dunder method (although it opens the question of do we even want to go down that route for ones Python doesn't define), we should be intentional about it, and likely introduce a corresponding trait for it.

Given the existing `math.sqrt(Int)` function, I could imagine putting an `IntLiteral` overload next to it.  What do you think, @abduld?

helehex:
if we change to using `isqrt` to mean integer square root this should probably be rethought as well.

helehex:
Oh yeah, having it be a method in `IntLiteral` makes the tests more explicit too

helehex:
Closing until we have better benchmarks.

PR #3382, opened 2024-08-15
[stdlib] Add `sqrt(FloatLiteral) -> FloatLiteral` overload.

helehex:
Add `FloatLiteral.__sqrt__()`
Add `FloatLiteral.__recip_sqrt__()`
Add Newtons method for solving it.

I hope this doesn't explode anyone's compile times, but since i did #3381 I figured I should add this as well.

### DIFF
--- a/stdlib/src/builtin/float_literal.mojo
+++ b/stdlib/src/builtin/float_literal.mojo
@@ -16,6 +16,7 @@ These are Mojo built-ins, so you don't need to import them.
 """
 
 from builtin._math import Ceilable, CeilDivable, Floorable, Truncable
+from math.solver import newtons_method
 
 # ===----------------------------------------------------------------------===#
 # FloatLiteral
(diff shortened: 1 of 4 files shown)


helehex:
There's worse things you could already do at compile time, right? It'll be finee...

helehex:
If it is a problem, I could get rid of the `sqrt()` overload, but keep the `FloatLiteral__sqrt__()` method so we still have something. Then we can improve the implementation over time. same for #3381

JoeLoser (commented):
Similar comments about the use of special dunder methods as my comments from https://github.com/modularml/mojo/pull/3381#issuecomment-2305320989.  @abduld how do you feel about introducing a Newton solver in the math module?  I'm not so sure we want to go down that route right now in the standard library.

helehex:
The dunder would gives a nice handle for a sqrtable trait, and yeah I'm not sure about the newton solver either, but it's what I had on hand, and I realized that having a intliteral sqrt and not a floatliteral sqrt might cause issues, so I just did both

helehex:
Also, i know someone was asking about floatliter/intliteral square root in one of the community meetings, and since i already made them for one of my projects, i figured i'd try adding them. (which is another reason this would be nice; a sqrtable trait would avoid conflics with builtin sqrt that i've have to work around in that project)

helehex:
Closing until we have better benchmarks

PR #3386, opened 2024-08-16
[proposal] Some Mojo Classes proposal fixes

msaelices:
- Fix the `mro()` sample implementation
- Fix `__new__` signature


JoeLoser:
Thanks for the fixes, but we've actually decided to remove this proposal from the public for now since there's no current plans of working on this in the short-term.  The design of classes will be different as the language evolves.

PR #3387, opened 2024-08-16
[stdlib] Make `NoneType` Formattable, Stringable and Representable.

helehex:
Since currently, you can't do `str(NoneType())`

### DIFF
--- a/stdlib/src/builtin/none.mojo
+++ b/stdlib/src/builtin/none.mojo
@@ -18,7 +18,13 @@ These are Mojo built-ins, so you don't need to import them.
 
 @value
 @register_passable("trivial")
-struct NoneType(CollectionElement):
+struct NoneType(
+    CollectionElement,
+    CollectionElementNew,
+    Formattable,
+    Representable,
+    Stringable,
+):
     """Represents the absence of a value."""
 
     alias _mlir_type = __mlir_type.`!kgen.none`
@@ -26,10 +32,12 @@ struct NoneType(CollectionElement):
 
     var _value: Self._mlir_type
 
+    @always_inline
     fn __init__(inout self):
         """Construct an instance of the `None` type."""
         self._value = None
 
+    @always_inline
     fn __init__(inout self, *, other: Self):
         """Explicit copy constructor.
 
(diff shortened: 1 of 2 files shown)


soraros:
After this we can finally remove the `_NoneType` from optional.

stdlib/src/builtin/none.mojo
```diff
@@ -18,7 +18,7 @@ These are Mojo built-ins, so you don't need to import them.
 
 @value
 @register_passable("trivial")
-struct NoneType(CollectionElement):
+struct NoneType(CollectionElement, Formattable, Representable, Stringable):
```
soraros, line 21: Maybe also add conformance to `CollectionElementNew`.

JoeLoser:
!sync

stdlib/src/builtin/none.mojo
```diff
@@ -37,3 +37,28 @@ struct NoneType(CollectionElement):
...
+    fn __str__(self) -> String:
+        """Returns the string representation of `None`.
+
+        Returns: `"None"`.
+        """
+        return "None"
+
+    @no_inline
+    fn __repr__(self) -> String:
+        """Return the string representation of `None`.
```
gryznar, line 51: Inconsistency. Above is Returns (with -s)

stdlib/src/builtin/none.mojo
```diff
@@ -37,3 +37,28 @@ struct NoneType(CollectionElement):
...
+
+        Returns: `"None"`.
+        """
+        return "None"
+
+    @no_inline
+    fn format_to(self, inout writer: Formatter):
```
JoeLoser, line 58: **Suggestion** Can we write a test for this?  I think this will currently fail still in `test/utils/test_format.mojo`:

```
ef test_formatter_of_none():
   var x = None
   var fmt = Formatter(x)
   fmt.write(x)
   assert_equal(x, "None")
```

I think it will fail to construct the formatter.

helehex:
Thanks guys, I will definitely get to the suggestions here, just feels kinda weird right now with the tests failing, and the version mismatch problem. Hopefully those things get sorted out soon because I couldn't even run the stdlib tests locally without downgrading my mojo version last time I tried.

helehex:
There's a test that's failing even if I override the version check, so I cant really make sure it's correct. Hopefully it's all good though.

helehex:
Btw @soraros @JoeLoser, Maybe you know this already, but using `NoneType` instead of  `_NoneType` breaks something in `Set`. 
Maybe because `NoneType` is `register_passable`?

helehex:
Oups, Joe is on vacation

helehex:
They're both zero size, so i dont think that would be the issue..

soraros:
I also tried removing `_NoneType` and it was trigging some compiler bugs. So maybe we should do that on a later date.

stdlib/test/builtin/test_none.mojo
```diff
@@ -12,11 +12,28 @@
...
+    test_format_to()
     test_type_from_none()
 
 
+def test_str():
+    assert_equal(NoneType().__str__(), "None")
+
+
+def test_repr():
+    assert_equal(NoneType().__str__(), "None")
```
JoeLoser, line 30: **Question** Shouldn't this be calling `repr` dunder?

JoeLoser:
!sync

stdlib/test/builtin/test_none.mojo
helehex: Yes it should, good catch and thanks for fixing it for me.
I still have to set up my stdlib development with magic properly.

PR #3388, opened 2024-08-16
[stdlib] Implicit conversion from `None` to `UnsafePointer`

helehex:
None is sort of like null in other languages, so this is helpful as a consistent way to say a pointer is null, without having to write out the full type.

### DIFF
--- a/stdlib/src/memory/unsafe_pointer.mojo
+++ b/stdlib/src/memory/unsafe_pointer.mojo
@@ -92,8 +92,12 @@ struct UnsafePointer[
     # ===-------------------------------------------------------------------===#
 
     @always_inline
-    fn __init__(inout self):
-        """Create a null pointer."""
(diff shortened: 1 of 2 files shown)


JoeLoser:
I'm not sure we want to do this - WDYT @ConnorGray?

JoeLoser:
After discussing some more with the team, here's some take-aways for why we're not looking to accept this change as is:
- Avoiding implicit conversions in low-level code (like `UnsafePointer`) is good practice, so introducing an implicit conversion here isn't ideal since you want to ensure you really understanding everything that is going on.
- We could imagine something akin to `nullptr` potentially which is `UnsafePointer[T]()` or a `nullptr_t` so folks can use it without knowing the `T` in the given context.

To help clean up the pull request queue, I'm going to close this.

PR #3392, opened 2024-08-17
[stdlib] Addition of a struct to read from stdin and an input function

thatstoasty:
Resolves https://github.com/modularml/mojo/issues/167

This is my attempt at recreating the Python input function and some of the stdin class functionality. Pretty straightforward, I'm using `getdelim` to read from stdin until the delimiter or EOF is reached. 

A few questions come to mind:

- I pretty much duplicated the `_fdopen` struct as part of the `stdin` struct, should I instead modify the `_fdopen` struct so mode isn't fixed to "a"? It should be reusable if that's updated. I didn't see any usage of the STDOUT and STDERR aliases that are part of it either.
- What should we do about testing logic that relies on the stdin stream? The tests I added are going to fail in automation because there's no input, but I was able to run them on my mac and in a linux container. I was also able to run one test at a time using FileCheck to echo some text into stdin, but of course that was consumed by the first test case which lead to the subsequent tests failing.


jackos (approved):
Fantastic this is great thanks @thatstoasty

I fixed up the test using echo in the run command as you hinted, one test is fine for now to make sure `_fdopen` is behaving as expected with stdin.

We should rework things in the future to move all file handle related functionality onto the `FileHandle` struct, and allow writing/reading directly from streams using `sys.stdin`, `sys.stdout` and `sys.stderr` as it behaves in Python, but we can do this later. For now this is functional and working great thanks.

jackos:
!sync

PR #3422, opened 2024-08-26
[stdlib] Atof bug: Fix shift overflow

yardenlaif:
Hi! I fixed the overflow that caused numbers such as `3.5e21` not to be parsed correctly. This is a partial fix for: https://github.com/modularml/mojo/issues/3419.

Signed-off-by: Yarden Laifenfeld <ylaifenfeld@gmail.com>


yardenlaif:
I'm assuming you want floats with big exponents (>16) to be treated the same by the compiler and the stdlib, so I'm not sure how to fix the bug with: `47421763.548648646474532187448684`.
The stdlib currently parses this to `47421763.548648641`, while the compiler parses it to `47421763.548648648`.

gabrieldemarmiesse:
@yardenlaif we should round to the nearest value representable by a float64. You can use `FloatLiterals`, which have infinite precision to verify which number is the closest:

```mojo
def main():
    alias x: FloatLiteral = 47421763.548648646474532187448684
    alias stdlib: FloatLiteral = 47421763.548648641
    alias compiler: FloatLiteral = 47421763.548648648

    alias x_minus_stdlib: FloatLiteral = x - stdlib
    print("x_minus_stdlib", x_minus_stdlib)

    alias x_minus_compiler: FloatLiteral = x - compiler
    print("x_minus_compiler", x_minus_compiler)

# Output:
# x_minus_stdlib 5.474532187448684e-09
# x_minus_compiler -1.525467812551316e-09
```
So the compiler produces a number closer to the original number, thus the compiler is right here.

Also I'm afraid the way we currently made the `atof` function is not great notably because we loose precision by using so many iteration on the floating point values. Thus it produces wrong or imprecise numbers when the exponents are very high or very low. I'm working on the implementation of the paper "Number Parsing at a Gigabyte per Second" by Daniel Lemire, which is the state of the art on the subject, and should solve a bunch of problems at once. PR is coming soon.

So I don't believe we can just "patch" the existing `atof` function, it has to be remade from scratch to be correct.

JoeLoser:
> @yardenlaif we should round to the nearest value representable by a float64. You can use `FloatLiterals`, which have infinite precision to verify which number is the closest:
> 
> ```mojo
> def main():
>     alias x: FloatLiteral = 47421763.548648646474532187448684
>     alias stdlib: FloatLiteral = 47421763.548648641
>     alias compiler: FloatLiteral = 47421763.548648648
> 
>     alias x_minus_stdlib: FloatLiteral = x - stdlib
>     print("x_minus_stdlib", x_minus_stdlib)
> 
>     alias x_minus_compiler: FloatLiteral = x - compiler
>     print("x_minus_compiler", x_minus_compiler)
> 
> # Output:
> # x_minus_stdlib 5.474532187448684e-09
> # x_minus_compiler -1.525467812551316e-09
> ```
> 
> So the compiler produces a number closer to the original number, thus the compiler is right here.
> 
> Also I'm afraid the way we currently made the `atof` function is not great notably because we loose precision by using so many iteration on the floating point values. Thus it produces wrong or imprecise numbers when the exponents are very high or very low. I'm working on the implementation of the paper "Number Parsing at a Gigabyte per Second" by Daniel Lemire, which is the state of the art on the subject, and should solve a bunch of problems at once. The PR is here: #3434
> 
> So I don't believe we can just "patch" the existing `atof` function, it has to be remade from scratch to be correct.

I agree with Gabriel here.  Thank you for starting on this @yardenlaif, but let's move forward with https://github.com/modularml/mojo/pull/3434.  Thanks @gabrieldemarmiesse and others!  As such, I'll close this PR.

PR #3439, opened 2024-08-31
[stdlib] Fix `String.isdigit()` when the string is empty

msaelices:
Corner case that was not working correctly as `String("").isdigit()` is `False`


soraros:
Maybe you could also update the doctoring to match python:

```
A string is a digit string if all characters in the string are digits and there
is at least one character in the string.
```

msaelices:
Sure: https://github.com/msaelices/mojo/commit/29eb124e0d19402d37541c06c0dcf6c69ca73fca

> Maybe you could also update the doctoring to match python:
> 
> ```
> A string is a digit string if all characters in the string are digits and there
> is at least one character in the string.
> ```

Sure. Done: https://github.com/msaelices/mojo/commit/29eb124e0d19402d37541c06c0dcf6c69ca73fca

JoeLoser (approved):

JoeLoser:
!sync

PR #3445, opened 2024-09-03
[stdlib] Fix soundness issues in InlinedFixedVector on copy/del

szbergeron:
`InlinedFixedVector` had apparent double free (on multiple call of `_del_old()`) that was easily accidentally invoked when `__copyinit__()` is used to create two instances.

This:
a) Moves to using `__del__()` for destruction, ensuring `dynamic_data` is zeroed to avoid further accidents (cheap, free if MLIR sees the write going into a black hole)
b) Switches `__copyinit__()` to use an explicit copy constructor, as mentioned in the TODO
c) Adds a `__moveinit__()` to allow for perf-identical behavior in additional remaining situations displaced by removing the original `__copyinit__()` (though non-owning slices are necessary for replacing all possible uses as before)


stdlib/test/collections/test_vector.mojo
```diff
@@ -18,6 +18,23 @@ from test_utils import MoveCounter
...
 
+def test_inlined_fixed_vector_moves():
+    var v1 = InlinedFixedVector[Int, 5](10)
+    var v2 = InlinedFixedVector[Int, 5](10)
+
+    # do one within the smallvec
+    v2[3] = 99
+    v1[3] = 42
+
+    # plus one within the dynarry
```
gryznar, line 29: typo probably. `dynarry` -> `dynarray`

szbergeron:
This also adds some tests for the various move constructors, but "instrumented" testing is difficult as InlinedFixedVector doesn't do type destruction (can't check as in ObservableDel). Do we want to add destructor support to this, or leave it as-is (with trivial types only)?

stdlib/src/collections/vector.mojo
```diff
@@ -125,37 +125,47 @@ struct InlinedFixedVector[
...
         """
-        var res = Self(self.capacity)
-        for i in range(len(self)):
-            res.append(self[i])
-        return res
+        Destructor.
+        """
+        if self.dynamic_data:
+            self.dynamic_data.free()
+            self.dynamic_data = UnsafePointer[type]()
```
JoeLoser, line 168: **Question** Can you help me understand the motivation of nulling out this field?  For comparison, we don't do this in `List.__del__` at https://github.com/modularml/mojo/blob/80a9de7261fbe850fa5ceb37c1489052f18364ac/stdlib/src/collections/list.mojo#L211

JoeLoser (approved):
LGTM, but see my comment about whether we need/want to null out the dynamic data member.

JoeLoser:
!sync

stdlib/src/collections/vector.mojo
szbergeron: This isn't necessary in "correct" code, but it _does_ prevent even incorrect usage of things like MaybeUninit double destruction from causing issues since it turns a double free into a noop. Just mitigates damage in those scenarios with ~no cost

szbergeron: (I'm more than happy to cut that line since that's probably an unnecessary amount of paranoia)

owenhilyard:
@JoeLoser 
> Let's leave support for just trivial types only and we can lazily add support for memory-only types later if needed. It may be moot as mentioned offline in Discord once we have small buffer optimization in `List` itself.

I have a few usecases for a collection like inline array which can't be grown after allocation, specifically in combination with a length field to be used in IO buffers. It makes packet processing and similar operations much more prefetcher friendly to have the metadata and the buffer as part of single allocation. The more restrictive API of InlineArray is good for this use-case since buffers of often allocated in pools. InlineArray is also useful as a building block in BTrees and slab lists.

PR #3449, opened 2024-09-04
[stdlib] Fix typo in inlined_assembly 9-arg function arguments

tboerstad:
I discovered this copy-paste typo, arg3 repeated twice for 9 arg assembly generation, while poking around the stdlib. 

Unfortunately I don't know how to test this.
Tests passed both before and after this change, so I assume this code path is not tested on my setup


laszlokindrat (approved):
Nice catch, haha! Thanks!

PR #3450, opened 2024-09-04
[stdlib] Implement Weak[T] refs for Arc[T]

szbergeron:
This implements a Weak[T] non-owning ref for Arc[T] that should allow for more ergonomic implementation of cyclic data structures without having to fall back to UnsafePointer[T] or Optional[Arc[T]] + complex directed drop impls.

This increases the footprint of _ArcInner by (minimally) 8 bytes. If we wish to cut this down, specialization and conditional inclusion of the field (once implemented) may allow for Arc[T, Kind=StrongOnly] constructs.


szbergeron:
Needs some fixes after parameterizing Arc for Weak, but recent compiler changes have fixed the parameter propagation/inference issues so it's just a matter of fixing the broken ref count.

stdlib/src/memory/arc.mojo
```diff
@@ -45,32 +45,231 @@ print(Arc(String("ok"))[])
...
+            # if this is the last weak, and there
+            # are no other strongs, then we can dealloc the inner
+            return last_any
+        else:
+            #TODO: statically ensure unreachable?
+            return False
+
+
+@register_passable
+struct Weak[T: Movable, enable_weak: Bool = True](CollectionElement, CollectionElementNew):
```
ConnorGray, line 193: **Suggestion** I think in principle it should be possible for us to remove this `enable_weak` parameter from `Weak` itself.

The logic being that an instance of `Weak` can only have been derived from an `Arc[T, True]`, so the parameter would always have to be `True` anyway.

stdlib/src/memory/arc.mojo
```diff
@@ -45,32 +45,231 @@ print(Arc(String("ok"))[])
...
+    such that users can easily avoid memory leaks.
+
+    Parameters:
+        T: The type of the value that this Weak may be upgraded into an Arc[T] of.
+    """
+
+    alias _inner_type = _ArcInner[T]
+    var _inner: UnsafePointer[Self._inner_type]
+
+    fn __init__(inout self, strong: Arc[T, enable_weak]):
```
ConnorGray, line 207: **Suggestion** Likewise, we can change this initializer so that `Weak[T]` can only be constructed frfom an `Arc[T, True]`:

```preview
          fn __init__(inout self, strong: Arc[T, True]):
```

This would let you remove the `constrained[enable_weak, "..."]()` below as well 🙂

stdlib/src/memory/arc.mojo
```diff
@@ -45,32 +45,231 @@ print(Arc(String("ok"))[])
...
+        """Implicit copy of Weak[T],
+        simply delegates to the regular explicit copy constructor.
+
+        Args:
+            other: The Weak[T] to (implicitly) copy.
+        """
+        other._inner[].add_weak()
+        self._inner = other._inner
+
+    fn upgrade(inout self) -> Optional[Arc[T]]:
```
ConnorGray, line 249: **Suggestion** I think this needs to encode that this is an `Arc` containing a weak count:

```suggestion
    fn upgrade(inout self) -> Optional[Arc[T, True]]:
```

stdlib/src/memory/arc.mojo
```diff
@@ -126,11 +335,12 @@ struct Arc[T: Movable](CollectionElement, CollectionElementNew):
...
         Decrement the ref count for the reference. If there are no more
-        references, delete the object and free its memory."""
-        if self._inner[].drop_ref():
-            # Call inner destructor, then free the memory.
+        references, free its memory."""
+        var should_del = self._inner[].drop_ref()
+        if should_del:
             (self._inner).destroy_pointee()
             self._inner.free()
+        self._inner = UnsafePointer[Self._inner_type]()
```
ConnorGray, line 343: **Request** This "nulling out" of fields is not required within `__del__()` methods. After the `__del__()` completes, this object instance is no longer alive, so there's no code that could observe `self._inner` one way or the other 🙂

stdlib/src/memory/arc.mojo
```diff
@@ -45,32 +45,231 @@ print(Arc(String("ok"))[])
...
 
-struct _ArcInner[T: Movable]:
+struct _ArcInner[T: Movable, enable_weak: Bool = False]:
     var refcount: Atomic[DType.uint64]
-    var payload: T
+    var payload: UnsafeMaybeUninitialized[T]
+    
+    alias _weak_refcount_type = __mlir_type[
+        `!pop.array<`, Int(enable_weak).value, `, `, Atomic[DType.int64], `>`
+    ]
```
ConnorGray, line 62: **Comment** This is clever using the boolean to make this a zero-sized field, neat trick! 🙂

stdlib/src/memory/arc.mojo
```diff
@@ -45,32 +45,231 @@ print(Arc(String("ok"))[])
...
+            return UnsafePointer[Atomic[DType.int64]]()
+
+    fn _destroy_value(inout self):
+        """
+        Run the destructor for the payload,
+        with the assumption that it has not been run yet
+        """
+        self.payload.assume_initialized_destroy()
+
+    fn try_upgrade_weak(inout self) -> Bool:
```
ConnorGray, line 100: **Suggestion** I think this could use the "custom self types" / "conditional conformance" trick to be statically constrained to:

```suggestion
fn try_upgrade_weak(inout self: Arc[T, True]) -> Bool
```

and then the `@parameter if enable_weak` statement below could be simplified.

ConnorGray (commented):
Hey @szbergeron, thanks for tackling this! 🙂 This will be a nice QoL thing for folks doing self-referential data structures in Mojo.

I've left a few initial comments on the implementation strategy, particularly about how we might be able to make stronger guarantees around the `enable_weak: Bool` parameter using the type system.

Let me know if my suggestions are workable! 🙂 

Thanks again for your contribution, awesome stuff 🔥

stdlib/src/memory/arc.mojo
szbergeron: I think this is an artifact of having written the first pass when cond-conf was...rather "newly minted" and had some rather squirrely behavior. I can see if I can pull it out now without knocking over the jenga tower.

stdlib/src/os/atomic.mojo
```diff
@@ -24,7 +24,7 @@ from memory import UnsafePointer, bitcast
 from sys.info import triple_is_nvidia_cuda
 
 
-struct Atomic[type: DType]:
+struct Atomic[type: DType](ExplicitlyCopyable):
```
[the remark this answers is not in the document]
szbergeron: This was to allow me to construct the [0|1] length array for weak. I wanted to revisit the specifics, since having an array of atomics can be useful in a lot of contexts but there isn't currently a good way of constructing one without arcane MaybeUninit incantations. I do have some questions about the atomic type that can be had out-of-thread too, that (I think) make Arc technically unsound in a single threaded context and relatively impossible to use in a multi-threaded one

stdlib/src/memory/arc.mojo
```diff
@@ -152,7 +362,27 @@ struct Arc[T: Movable](CollectionElement, CollectionElementNew):
...
+        constrained[enable_weak, "A typical Arc[T] does not support downgrading." + 
+            "If you want to enable downgrading of this Arc[T], convert it to an Arc[T, enable_weak = True]"]()
+        
+        @parameter
+        if enable_weak:
+            return Weak[T, enable_weak](self)
+        else:
+            constrained[False, "A typical Arc[T] does not support downgrading." + 
+                "If you want to enable downgrading of this Arc[T], convert it to an Arc[T, enable_weak = True]"]()
+            return abort[Weak[T, enable_weak]]("unreachable")
```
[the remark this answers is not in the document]
szbergeron: I like the simplification, but is there a good way with that change of presenting specific, actionable instructions to the user that what they are trying to do can only be done if they make their Arc `enable_weak = True`?

JoeLoser:
@szbergeron going to close this for now, but if it becomes a priority in the future - let's talk internally.

PR #3453, opened 2024-09-05
[stdlib] Use ownership ops in `Uninit` to remove pointer indirection.

helehex:
The ownership ops provide a more clear implementation than `pop.array`, and allows the pointer indirection to be removed.
This makes the inner field of `UnsafeMaybeUninitialized` just a `ElementType` instead of a `pop.array<...>`

Because we get an error when marking subobjects as destroyed, ~~i used a capturing function within the types destructor to make sure the underlying values destructor isn't called.~~ **edit**: I introduced two helper functions `_forget()` and `_uninit()`, which provide a very clear implementation.

I had to bump up the move counter check in one of the tests, because of how the implementation changed, but i'm pretty sure the actual move gets optimized away, even if the move counter doesn't.

### DIFF
--- a/stdlib/src/collections/inline_array.mojo
+++ b/stdlib/src/collections/inline_array.mojo
@@ -128,15 +128,11 @@ struct InlineArray[
             unsafe_assume_initialized: The array of `UnsafeMaybeUninitialized` elements.
         """
 
-        self._array = __mlir_op.`kgen.param.constant`[
-            _type = Self.type,
-            value = __mlir_attr[`#kgen.unknown : `, Self.type],
-        ]()
-
-        for i in range(Self.size):
-            unsafe_assume_initialized[i].unsafe_ptr().move_pointee_into(
-                self.unsafe_ptr() + i
-            )
+        self = (
+            UnsafePointer.address_of(unsafe_assume_initialized)
+            .bitcast[Self]()
+            .take_pointee()
+        )
 
     @always_inline
     fn __init__(inout self, fill: Self.ElementType):
--- a/stdlib/src/memory/maybe_uninitialized.mojo
+++ b/stdlib/src/memory/maybe_uninitialized.mojo
@@ -15,6 +15,18 @@ from os import abort
 from builtin._documentation import doc_private
 
 
+@always_inline
+fn _forget[T: AnyType](owned value: T):
+    # Avoids running the destructor.
+    __mlir_op.`lit.ownership.mark_destroyed`(__get_mvalue_as_litref(value))
+
+
+@always_inline
+fn _uninit[T: AnyType]() -> T as value:
+    # Returns uninitialized data.
+    __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(value))
+
+
 struct UnsafeMaybeUninitialized[ElementType: AnyType](CollectionElementNew):
     """A memory location that may or may not be initialized.
 
@@ -31,17 +43,14 @@ struct UnsafeMaybeUninitialized[ElementType: AnyType](CollectionElementNew):
         ElementType: The type of the element to store.
     """
 
-    alias type = __mlir_type[`!pop.array<1, `, Self.ElementType, `>`]
-    var _array: Self.type
+    var _data: ElementType
 
     @always_inline
     fn __init__(inout self):
         """The memory is now considered uninitialized."""
-        self._array = __mlir_op.`kgen.param.constant`[
-            _type = Self.type,
-            value = __mlir_attr[`#kgen.unknown : `, Self.type],
-        ]()
+        self._data = _uninit[ElementType]()
 
+    # TODO: allows conformance to CollectionElement for use in collections
     @doc_private
     @always_inline
     fn __init__(inout self, *, other: Self):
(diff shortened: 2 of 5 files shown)


stdlib/test/memory/test_maybe_uninitialized.mojo
```diff
@@ -89,7 +89,7 @@ def test_maybe_uninitialized_move_from_pointer():
     _ = a^
 
     # a is uninitialized now. Thankfully, we're working with trivial types
-    assert_equal(b.assume_initialized().move_count, 1)
+    assert_equal(b.assume_initialized().move_count, 2)
```
gabrieldemarmiesse, line 92: This is quite strange indeed. Can you add a unit test with a non-movable type and see if it works correctly?

gabrieldemarmiesse (changes requested):

stdlib/test/memory/test_maybe_uninitialized.mojo
helehex: Once I figure out stdlib dev with magic, i'll add it.

stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -250,4 +267,12 @@ struct UnsafeMaybeUninitialized[ElementType: AnyType](CollectionElementNew):
         If the memory was initialized, the caller should
         use `assume_initialized_destroy` before.
         """
-        pass
+
+        @parameter
+        @always_inline
+        fn _drop[T: AnyType](owned val: T):
```
JoeLoser, line 273: **Question** What's the value in wrapping this MLIR op call in an inner function instead of just inlining it?

stdlib/test/memory/test_maybe_uninitialized.mojo
JoeLoser: @helehex did you dig into this a bit more?  Seems like a regression and/or a compiler bug.  If it's a bug, I'm sure @lattner would like to hear about it.

stdlib/src/memory/maybe_uninitialized.mojo
JoeLoser: Ah, I see now from your comment in the PR description - can you check if this is required still?  @lattner has improved this subobject destruction area recently in `CheckLifetimes`, so it may not be needed anymore.

helehex: I'll add this to my todo list tomorrow.

For some reason i cant comment on your review below, but the reason the move count goes up by one is because I'm doing two moves instead of one now. The actual moves should get optimized away, but it cant optimize the counter increment because that would be incorrect. (i'll explore cleaning it up some more though and getting rid of the extra move)

stdlib/test/memory/test_maybe_uninitialized.mojo
helehex: The reason the move count goes up by one is because I'm doing two moves instead of one now. The actual moves should get optimized away, but it cant optimize the counter increment because that would be incorrect. (i'll explore cleaning it up some more though and getting rid of the extra move)

stdlib/src/memory/maybe_uninitialized.mojo
```diff
@@ -165,7 +181,8 @@ struct UnsafeMaybeUninitialized[ElementType: AnyType](CollectionElementNew):
         Args:
             other: The object to move.
         """
-        self.move_from(other.unsafe_ptr())
+        self = UnsafeMaybeUninitialized[MovableType](other._data^)
+        other = UnsafeMaybeUninitialized[MovableType]()
```
helehex, line 186: I'm seeing that a bunch of the things that seemed wrong to me before, have been fixed.
This is now giving me:
`field 'other._data' destroyed out of the middle of a value, preventing the overall value from being destroyed`
Which is kind of what i expect, so thats good. I just have to think of a better way to do it now

helehex: i tried it like this originally to avoid having `lit.ownership.mark_initialized()` everywhere, but maybe it makes more sense to do that anyways..

stdlib/src/memory/maybe_uninitialized.mojo
helehex: moving it out of the closure still complains about not being able to destroy subobjects, but maybe there's a better way

helehex: but i can mark subobjects initialized now, which helps

helehex: i'm cooking

helehex:
I do have a question for @gabrieldemarmiesse, i noticed that you made an `__init__()` for a movable element type which does the same thing as `write()`. I dont see how having both would be necessary, since i dont believe it's a matter of conformance to some trait, do you more info on that decision, or a preference on which one to remove?

helehex:
I'm not sure it makes sense to have an `unsafe_ptr()` method anymore, but i'll push my changes and you guys can take a look.

helehex:
I added two private helper functions: `_forget()`, and `_uninit`, if you take a look at where they are being used, hopefully it's a very clear way of doing it.

helehex:
the diff is also in the negative now, and i didn't remove any functionality, so that's always good

stdlib/src/collections/inline_array.mojo
```diff
@@ -122,12 +122,11 @@ struct InlineArray[
...
-
-        for i in range(Self.size):
-            unsafe_assume_initialized[i].unsafe_ptr().move_pointee_into(
-                self.unsafe_ptr() + i
-            )
+        self = (
+            UnsafePointer.address_of(unsafe_assume_initialized)
+            .bitcast[Self]()
+            .take_pointee()
+        )
```
helehex, line 130: I had to change this initializer to be a direct cast and move, rather than moving each element individually.
Seems like it's correct, but maybe something to check into.

stdlib/test/memory/test_maybe_uninitialized.mojo
helehex: Oh
Now it might be because it goes through `take_pointee()` first, then gets moved into the destination variable.

helehex: Oh yeah, a non-movable test... i should try that

helehex: Wait, if it's non-movable then you cant use `move_from`... could you maybe clarify what the test should look like?

helehex: @JoeLoser Maybe `UnsafePointer.take_pointee()` should be using a named return slot?

PR #3461, opened 2024-09-09
[stdlib] Add negative case for `String.split()`

jjvraw:
**Repurposed PR.** 

Originally addressed significant inefficiency in `String.split()`. It appears the issue was rather related to 5b026f4.

### DIFF
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -1545,7 +1545,7 @@ struct String(
         var items = 0
         var sep_len = sep.byte_length()
         if sep_len == 0:
-            raise Error("Seperator cannot be empty.")
+            raise Error("Separator cannot be empty.")
         if str_byte_len < 0:
             output.append("")
 
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -1522,6 +1522,9 @@ struct String(
         Returns:
             A List of Strings containing the input split by the separator.
 
+        Raises:
+            If the separator is empty.
+
         Examples:
 
         ```mojo
@@ -1542,7 +1545,7 @@ struct String(
         var items = 0
         var sep_len = sep.byte_length()
         if sep_len == 0:
-            raise Error("ValueError: empty separator")
+            raise Error("Seperator cannot be empty.")
         if str_byte_len < 0:
             output.append("")
 
--- a/stdlib/test/collections/test_string.mojo
+++ b/stdlib/test/collections/test_string.mojo
@@ -819,6 +819,9 @@ def test_split():
     assert_equal(res6[3], "сит")
     assert_equal(res6[4], "амет")
 
+    with assert_raises(contains="Seperator cannot be empty."):
+        _ = String("1, 2, 3").split("")
+
 
 def test_splitlines():
     # Test with no line breaks
--- a/stdlib/test/collections/test_string.mojo
+++ b/stdlib/test/collections/test_string.mojo
@@ -819,7 +819,7 @@ def test_split():
     assert_equal(res6[3], "сит")
     assert_equal(res6[4], "амет")
 
-    with assert_raises(contains="Seperator cannot be empty."):
+    with assert_raises(contains="Separator cannot be empty."):
         _ = String("1, 2, 3").split("")
 
 


gabrieldemarmiesse:
I'll just add here that it's related to https://github.com/modularml/mojo/issues/3460

stdlib/src/collections/string.mojo
```diff
@@ -1514,6 +1514,9 @@ struct String(
         Returns:
             A List of Strings containing the input split by the separator.
 
+        Raises:
+            If the seperator is empty.
```
msaelices, line 1518: NIT: Small typo:
```suggestion
            If the separator is empty.
```

soraros:
I fail to understand exactly which part contribute to the slowness.

stdlib/src/utils/string_slice.mojo
```diff
@@ -739,6 +739,77 @@ struct StringSlice[
...
+            var found_at = str_len
+            var curr_ptr = ptr.offset(current_offset)
+
+            for i in range(current_offset, str_len):
+                if i + sep_len > str_len:
+                    break
+
+                if memcmp(ptr.offset(i), sep_ptr, sep_len) == 0:
+                    found_at = i
+                    break
```
soraros, line 784:
```suggestion
            loc = _memmem(curr_ptr, str_len - current_offset, sep_ptr, sep_len)
            if loc:
                found_at = int(loc) - int(ptr)
```

jjvraw:
My assumption is that given Mojo's value semantics and no GC, the constant creation of temporary `StringSlice` objects through the `_from_start` (called in `find`), and frequent conversion from `StringSlice` to `String` , suggests significant overhead from object initialisation and destruction, per iteration. The frequent object creations/destructions, seem to incur costs in terms of memory allocations and deallocation through their lifecycle methods.

This is the main difference with the new implementation, directly working with pointers and avoiding any object creation other than when appending to the out list.

I agree that simply inlining `_memmem` shouldn't cause this much speedup, so this is the reasoning I am leaning towards.

soraros:
That's the thing, both `StringSlice` and `Span` live on the stack, and are essentially registers passable modulo lifetime. Bound check can be the source of some overhead, but I think it shouldn't be this large. I would be equally surprised if `string_view` is this much slower than `*char`, for instance.

msaelices:
> That's the thing, both `StringSlice` and `Span` live on the stack, and are essentially registers passable modulo lifetime. Bound check can be the source of some overhead, but I think it shouldn't be this large. I would be equally surprised if `string_view` is this much slower than `*char`, for instance.

I think this could mean that there is some issue in the `StringSlice` and `Span` structs we would need to fix to really optimize them but, in the meantime, this PR is a vast and legit optimization we could add to the `nightly` branch, right? Maybe I am missing something.

soraros:
@msaelices Just to be clear, I'm fully supportive of having this PR merged. I just wanted to take the opportunity to raise my concerns.

msaelices:
> @msaelices Just to be clear, I'm fully supportive of having this PR merged. I just wanted to take the opportunity to raise my concerns.

It's clear that there is something odd, so it's a legit concern, of course. Thanks for the clarification 

> 
> Actually, I think the problem is resolved by [5b026f4](https://github.com/modularml/mojo/commit/5b026f4e946fba3e6af02383bd2a112ed49cbaa5)?

Ohh! Great fix! I will recheck my CSV benchmarks. Maybe we can beat Python now?

jjvraw:
Yes, it has! Thanks for pointing out @soraros 

> So, in practice, we see a lot of calls to `String.format` (which was
called from inside `debug_assert` which does not-so-small amount of work to
construct a `String` that may never used at runtime in a program.

This makes more sense. I guess the two instances of string construction from the `debug_assert`s in `_from_start`, per iteration, were causing a not-so-small amount of work... On this note, I wonder which, if any, profiling was used.

I will drop the refactor and repurpose this PR to include just the negative case :)

stdlib/src/collections/string.mojo
```diff
@@ -1542,7 +1545,7 @@ struct String(
         var items = 0
         var sep_len = sep.byte_length()
         if sep_len == 0:
-            raise Error("ValueError: empty separator")
+            raise Error("Seperator cannot be empty.")
```
JoeLoser, line 1548:
```suggestion
            raise Error("Separator cannot be empty.")
```

stdlib/test/collections/test_string.mojo
```diff
@@ -819,6 +819,9 @@ def test_split():
     assert_equal(res6[3], "сит")
     assert_equal(res6[4], "амет")
 
+    with assert_raises(contains="Seperator cannot be empty."):
```
JoeLoser, line 822:
```suggestion
    with assert_raises(contains="Separator cannot be empty."):
```

JoeLoser (approved):

jjvraw:
These failures seem to be related to the latest nightly.

JoeLoser:
!sync

PR #3436, opened 2024-08-31
[stdlib] Faster hashing logic in string-keyed dicts (~2x speed-up)

msaelices:
The idea is not to use the SIMD hashing algorithm when the Dict keys are strings, usually small ones

## Benchmark 1

Same benchmark used in this other optimization: https://github.com/modularml/mojo/pull/3071

```mojo
from collections import Dict
from time import now
from random import *

alias iteration_size = 128 #2048
def main():
    var result: Int=0
    var start = now()
    var stop = now()

    small2 = Dict[Int,Int]()
    start = now()
    for x in range(100):
        for i in range(iteration_size):
            small2[i]=i
        for i in range(iteration_size):
            result += small2[i]
    stop = now()
    print("Int dicts:", stop-start, "ns", result, "rows")

    small3 = Dict[String,String]()
    start = now()
    for x in range(100):
        for i in range(iteration_size):
            small3[str(i)]=str(i)
        for i in range(iteration_size):
            result += len(small3[str(i)])
    stop = now()
    print("String dicts:", stop-start, "ns", result, "rows")
 ```

### Results (lower is better)

**Before** :
```
Int dicts: 5420564 ns 209612800 rows
String dicts: 273638909 ns 210321000 rows
```

**After**:
```
Int dicts: 3266663 ns 209612800 rows
String dicts: 126893375 ns 210321000 rows
```
This is roughly **x2 speed-up** in `Dict[String, X]`.

### Results (the last column is timing, lower is better)

**Before**:
```csv
version,n_wds,n_keys,the,sec
13638,1944,236,0.0083167180000000007
13638,1944,236,0.008287684
13638,1944,236,0.0084045869999999998
13638,1944,236,0.0083934049999999996
13638,1944,236,0.0083934230000000006
13638,1944,236,0.0084422609999999995
13638,1944,236,0.0084124249999999994
13638,1944,236,0.0084025450000000008
13638,1944,236,0.0084028699999999998
13638,1944,236,0.0083881839999999999
```

**After**:
```csv
version,n_wds,n_keys,the,sec
13638,1943,236,0.0027207440000000002
13638,1943,236,0.0026398810000000002
13638,1943,236,0.0026396229999999998
13638,1943,236,0.002676248
13638,1943,236,0.0027269859999999998
13638,1943,236,0.0027547840000000001
13638,1943,236,0.0027066960000000002
13638,1943,236,0.0027737180000000001
13638,1943,236,0.0028008479999999999
13638,1943,236,0.0029300910000000001
[...]

PR #3427, opened 2024-08-28
[stdlib] Faster hashing logic in string-keyed dicts (~x2 speed-up)

msaelices:
Faster `String.__hash__()` algorithm by not using the builtin `hash` function, which uses SIMD underneath, and somehow it's slowing the hash logic down (maybe because unneeded data conversion or `memcpy`). The hashing logic was slowing down Mojo string-keyed Dicts compared with Python, and with this optimization it is way faster now.

I'm now using a very simplified DJBX33A algorithm version only for the `String` struct. If the reviewer thinks the approach is valid, I can refine it.

See https://github.com/modularml/mojo/discussions/1747

**Note**: there is a less disruptive approach here: https://github.com/modularml/mojo/pull/3436. It should have the same x2 outcome but, unfortunately, it's failing when compiling, maybe because of this compiler issue https://github.com/modularml/mojo/issues/3437

## Benchmark 1

Same benchmark used in this other optimization: https://github.com/modularml/mojo/pull/3071

```mojo
from collections import Dict
from time import now
from random import *

alias iteration_size = 128 #2048
def main():
    var result: Int=0
    var start = now()
    var stop = now()

    small2 = Dict[Int,Int]()
    start = now()
    for x in range(100):
        for i in range(iteration_size):
            small2[i]=i
        for i in range(iteration_size):
            result += small2[i]
    stop = now()
    print(stop-start, result)

    small3 = Dict[String,String]()
    start = now()
    for x in range(100):
        for i in range(iteration_size):
            small3[str(i)]=str(i)
        for i in range(iteration_size):
            result += len(small3[str(i)])
    stop = now()
 ```

### Results (lower is better)

**Before** :
```
187688 812800
17102347 840200
```

**After**:
```
147619 812800
8381432 840200
```
This is roughly **x2 speed-up** in `Dict[String, X]`.

### Results (the last column is timing, lower is better)

**Before**:
```csv
version,n_wds,n_keys,the,sec
13638,1944,236,0.0083167180000000007
13638,1944,236,0.008287684
13638,1944,236,0.0084045869999999998
13638,1944,236,0.0083934049999999996
13638,1944,236,0.0083934230000000006
13638,1944,236,0.0084422609999999995
13638,1944,236,0.0084124249999999994
13638,1944,236,0.0084025450000000008
13638,1944,236,0.0084028699999999998
13638,1944,236,0.0083881839999999999
```

**After**:
```csv
version,n_wds,n_keys,the,sec
[...]

PR #3476, opened 2024-09-13
[stdlib] Add a new hash function based on AHash algorithm

mzaks:
This PR introduces a new hash function based on [AHash](https://github.com/tkaitchuck/aHash/tree/master) algorithm. The goal is to provide a playground for std lib devs to consider replacing current hash function.

### DIFF
--- a/stdlib/benchmarks/builtin/bench_hash.mojo
+++ b/stdlib/benchmarks/builtin/bench_hash.mojo
@@ -665,7 +665,7 @@ struct AHasher:
         self.extra_keys = U128(pi_key[2], pi_key[3])
 
     @always_inline
-    fn large_update(inout self, new_data: U128):
+    fn _large_update(inout self, new_data: U128):
         """Update the buffer value with new data.
 
         Args:
--- a/stdlib/benchmarks/builtin/bench_hash.mojo
+++ b/stdlib/benchmarks/builtin/bench_hash.mojo
@@ -706,7 +706,7 @@ struct AHasher:
                     var block = data.offset(offset).bitcast[
                         DType.uint64
                     ]().load[width=2]()
-                    self.large_update(block)
+                    self._large_update(block)
                     offset += 16
             else:
                 var a = data.bitcast[DType.uint64]().load()
--- a/stdlib/benchmarks/builtin/bench_hash.mojo
+++ b/stdlib/benchmarks/builtin/bench_hash.mojo
@@ -1,0 +1,872 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+# RUN: %mojo-no-debug %s -t
+
+from benchmark import Bench, BenchConfig, Bencher, BenchId
+
+# Source: https://www.101languages.net/arabic/most-common-arabic-words/
+alias words_ar = """
+لا, من, هذا, أن, في, أنا, على, ما, هل,
+ يا, و, لقد, ذلك, ماذا, أنت, هنا, لم, إلى, نعم, كان, هو, ان, هذه, هناك, عن, فى, كل, ليس, فقط, كنت, الآن, يجب, انا,
+ لك, مع, شيء, لكن, لن, الذي, حسنا, كيف, سوف, هيا, نحن, إنه, ـ, أجل, لماذا, إذا, عندما, انه, كذلك, لي, الى, بعد, انت,
+ هي, أين, أنه, كانت, حتى, أي, إنها, أعرف, قد, قبل, تلك, الأمر, بعض, أو, مثل, أريد, رجل, لو, أعتقد, ربما, أيها, بخير,
+ يكون, عليك, جيد, أنك, شخص, إن, التي, ولكن, أليس, علي, أحد, به, الوقت, يمكن, انها, اليوم, شئ, تعرف, تريد, صحيح, أكثر,
+ تكون, لست, كما, أستطيع, منذ, جدا, سيدي, يمكنك, لذا, واحد, لديك, يبدو, أوه, كلا, الرجل, لدي, تفعل, غير, عليه, اذا,
+ آخر, حدث, مرة, شكرا, لدينا, الناس, يوجد, له, مكان, سيد, سيكون, أعلم, رائع, مرحبا, آسف, بهذا, وقت, اللعنة, كم,
+ ليست, أفضل, بها, معك, أنها, الذى, الكثير, قلت, بك, يحدث, الان, يكن, يوم, وأنا, واحدة, بي, أخرى, ولا, علينا,
+ أبي, بأن, ثم, تعال, هكذا, يمكنني, هم, ألا, بالطبع, أنني, المكان, بذلك, معي, لهذا, ها, شىء, انك, إلهي, تستطيع,
+ العمل, العالم, الحقيقة, الليلة, بالتأكيد, حقا, تعلم, أمي, الطريق, حال, لى, لها, الأن, هؤلاء, فعل, توقف,
+ عمل, حول, لنا, خلال, اعتقد, السيد, انظر, منك, أى, أفعل, فعلت, لأن, إذن, قال, الجميع, تم, الجحيم, هى, فيه,
+ جيدة, عنه, بشكل, بما, تقول, لديه, ثانية, لذلك, أكون, دعنا, ايها, المال, يمكننا, الذهاب, متى, تعتقد,
+ اريد, عليها, أذهب, ستكون, فضلك, بدون, أرجوك, التى, شيئا, نذهب, لكي, نفسك, بنا, اين, وأنت, لكم, اي,
+ بين, إنهم, أرى, المنزل, بحق, كنا, عند, أم, منه, نفس, اذهب, حيث, مجرد, أقول, تبدو, الحياة, أيضا,
+ تحت, الأشياء, معه, يريد, أننا, أنظر, لما, اعرف, إلي, ثلاثة, انتظر, الرجال, الذين, حصلت, أني,
+ سعيد, لابد, عزيزتي, الشيء,
+  فكرة, انهم, الله, الباب, سيدى, دائما, رأيت, مشكلة, استطيع, تكن, تذهب, ليلة, شيئ, أظن, طوال,
+  جميل, وهو, الشرطة, او, دولار, السيارة, وهذا, كبير, مني, بسرعة, النار, الأمور, سمعت, أشعر, يعرف, 
+  أعني, لدى, بهذه, أحب, سنوات, بأس, الأفضل, بالنسبة,
+   أنتم, عظيم, يقول, جميلة, جون, جاك, بسبب, الوحيد, أمر, بل, بالفعل, الشخص, الي, دعني, خارج, اجل, الخير, ــ,
+   حالك, للغاية, فحسب, كانوا, أردت, فتاة, بشأن, يعني, كبيرة, ترى, آسفة, دقيقة, أنهم, يستطيع, احد, بأنك, تعمل,
+   تريدين, فيها, اليس, رائعة, رجال, نوع, حياتي, الأرض, البيت, قتل, اوه, والآن, مات, بكل, تعرفين, أحتاج, نستطيع,
+   جديد, صباح, ألم, عيد, منها, يعمل, الموت, إليك, جميع, لأنه, لحظة, لكني, الامر, عشر, لكنه, بحاجة, بأنه, أتمنى,
+   إليه, عنك, الفتاة, لهم, بالضبط, سأكون, اعلم, اللعين, رقم, طريق, منهم, المدينة, الحب, لنذهب, خذ, أكن, فوق,
+   عزيزي, دون, الـ, صغيرة, الرئيس, تتحدث, ترجمة, صديقي, فقد, الصغير, ولم, ساعة, يفعل, غرفة, وماذا, المرة, قام,
+   إلا, عام, هذة, متأكد, دقائق, سيارة, فعله, سعيدة, مما, ومن, معنا, سبب, سأذهب, الطريقة, الأطفال, سنة, بينما,
+   يرام, السبب, أننى, أول, اى, أريدك, قمت, الأولى, المدرسة, ذهبت, لطيف, نفسي, الا, الجنس, أية, أقصد, غريب, نفعل,
+   الصباح, حالة, المزيد, أبدا, مهما, اسمع, لأنك, أحاول, وقد, ايضا, أحبك, اكثر, فرصة, رأيك, افعل, الحصول, صغير,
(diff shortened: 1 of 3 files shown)


stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+alias U256 = SIMD[DType.uint64, 4]
+alias U128 = SIMD[DType.uint64, 2]
+alias MULTIPLE = 6364136223846793005
+alias ROT = 23
+
+
+@always_inline
+fn _folded_multiply(lhs: UInt64, rhs: UInt64) -> UInt64:
+    """A fast function to emulate a folded multiply of two 64 bit uints.
+    Used becuase we don't have UInt128 type.
```
msaelices, line 26: NIT: Small typo:
```suggestion
    Used because we don't have UInt128 type.
```

stdlib/benchmarks/builtin/bench_hash.mojo
```diff
@@ -0,0 +1,872 @@
...
+    References:
+
+    - [Reference Implementation in Rust](https://github.com/tkaitchuck/aHash)
+
+    ```mojo
+    from random import rand
+    var n = 64
+    var rand_bytes = UnsafePointer[UInt8].alloc(n)
+    rand(rand_bytes, n)
+    _ = hash(rand_bytes, n)
```
msaelices, line 734: NIT: 
```suggestion
    _ = ahash(rand_bytes, n)
```

stdlib/benchmarks/builtin/bench_hash.mojo
```diff
@@ -0,0 +1,872 @@
...
+            0x1319_8A2E_0370_7344,
+            0xA409_3822_299F_31D0,
+            0x082E_FA98_EC4E_6C89,
+        )
+        self.buffer = pi_key[0]
+        self.pad = pi_key[1]
+        self.extra_keys = U128(pi_key[2], pi_key[3])
+
+    @always_inline
+    fn large_update(inout self, new_data: U128):
```
msaelices, line 668: @mzaks, probably this is intended to be internal, so not part of the API, so, why not starting with `_`?

```suggestion
    fn _large_update(inout self, new_data: U128):
```

stdlib/benchmarks/builtin/bench_hash.mojo
```diff
@@ -0,0 +1,872 @@
...
+                var tail = data.offset(length - 16).bitcast[
+                    DType.uint64
+                ]().load[width=2]()
+                self.large_update(tail)
+                var offset = 0
+                while length - offset > 16:
+                    var block = data.offset(offset).bitcast[
+                        DType.uint64
+                    ]().load[width=2]()
+                    self.large_update(block)
```
msaelices, line 709: NIT:
```suggestion
                    self._large_update(block)
```

stdlib/benchmarks/builtin/bench_hash.mojo
```diff
@@ -0,0 +1,872 @@
...
+                    offset += 16
+            else:
+                var a = data.bitcast[DType.uint64]().load()
+                var b = data.offset(length - 8).bitcast[DType.uint64]().load()
+                self.large_update(U128(a, b))
```
msaelices, line 714:
```suggestion
                self._large_update(U128(a, b))
```

stdlib/benchmarks/builtin/bench_hash.mojo
```diff
@@ -0,0 +1,872 @@
...
+
+        Returns:
+            Final hash value.
+        """
+        var rot = self.buffer & 63
+        var folded = _folded_multiply(self.buffer, self.pad)
+        return (folded << rot) | (folded >> (64 - rot))
+
+    @always_inline
+    fn write(inout self, data: UnsafePointer[UInt8], length: Int):
```
msaelices, line 690: @mzaks, this is not following your proposal about [improve-hash-module](https://github.com/modularml/mojo/blob/main/proposals/improved-hash-module.md). Why not being consistent with it by just renaming it?

```suggestion
    fn update(inout self, data: UnsafePointer[UInt8], length: Int):
```

Or `_update_with_bytes()`

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+        n: The length of the byte array.
+
+    Returns:
+        A 64-bit integer hash. This hash is _not_ suitable for
+        cryptographic purposes, but will have good low-bit
+        hash collision statistical properties for common data structures.
+    """
+
+    var hasher = AHasher(key)
+    hasher.write(bytes, n)
```
msaelices, line 178: NIT:
```suggestion
    hasher.update(bytes, n)
```

stdlib/test/builtin/test__hash.mojo
```diff
@@ -0,0 +1,803 @@
...
+    assert_fill_factor["RU"](words, len(words) // 13, 1.0)
+
+    assert_fill_factor_old_hash["RU"](words, len(words), 0.18)
+
+
+def main():
+    test_hash_byte_array()
+    test_avalanche()
+    test_trailing_zeros()
+    test_fill_factor()
```
msaelices, line 803: That's awesome! Can I assume this is consistent with the Rust tests, so we can trust the algorithm by comparison with Rust?

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+    Args:
+        lhs: 64 bit uint.
+        rhs: 64 bit uint.
+
+    Returns:
+        A value which is similar in its bitpattern to result of a folded multply.
+    """
+    var b1 = lhs * byte_swap(rhs)
+    var b2 = byte_swap(lhs) * (~rhs)
+    return b1 ^ byte_swap(b2)
```
msaelices, line 37: :1st_place_medal:

msaelices (approved):
Great job! I've just added some NIT-pick comments, but this looks amazing

JoeLoser:
!sync

martinvuyk:
Hey guys, I'm a bit late to the party  since this is already imported internally. But what if we create a hashlib package like [python's](https://github.com/python/cpython/blob/main/Lib/hashlib.py) for this algo and what @msaelices is also doing with the DJBX33A algo in #3436 and #3427, that way we can also start unifying the interfaces

mzaks:
> Hey guys, I'm a bit late to the party  since this is already imported internally. But what if we create a hashlib package like [python's](https://github.com/python/cpython/blob/main/Lib/hashlib.py) for this algo and what @msaelices is also doing with the DJBX33A algo in #3436 and #3427, that way we can also start unifying the interfaces

Python hashlib is for cryptographic hash functions. I am generally all in for including multiple hash functions to std lib, but it only starts making sense when we can finally implement the Hasher trait and allow users to inject algorithm they want to use with the hash function. It might also not be in the interest of the std lib dev team to maintain multiple hash functions.

stdlib/test/builtin/test__hash.mojo
```diff
@@ -0,0 +1,803 @@
...
+    for w in words:
+        var h = old_hash(w[].unsafe_ptr(), w[].byte_length())
+        buckets[h % num_buckets] += 1
+    var unfilled = 0
+    for v in buckets:
+        if v[] == 0:
+            unfilled += 1
+
+    var fill_factor = 1 - unfilled / num_buckets
+    assert_true(
```
JoeLoser, line 731: FYI I'm seeing this fail internally on a m7i AWS VM with error:

```
stdlib/test/builtin/test__hash.mojo:731:16: AssertionError: Fill factor for AR is 0.59550286040551736, provided lower boound was 0.62
```

Should we adjust the bound, or is there an error elsewhere?

stdlib/test/builtin/test__hash.mojo
```diff
@@ -0,0 +1,803 @@
...
+        fill_factor >= lower_bound,
+        str("Fill factor for {} is {}, provided lower boound was {}").format(
```
JoeLoser, line 733:
```suggestion
        str("Fill factor for {} is {}, provided lower bound was {}").format(
```

stdlib/test/builtin/test__hash.mojo
JoeLoser: This would be good to investigate, I agree 👍

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
+# ===----------------------------------------------------------------------=== #
```
JoeLoser, line 1: As we start work to make the generic `HasherTrait` and have multiple hash functions to allow specifying them by the user, it may make sense to make a `hash` top-level module in the `stdlib` and home this new hash type here rather than in `builtin`.  This is still possible to re-export our eventual default hash type/functions now that we have the `prelude` module.

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+        A value which is similar in its bitpattern to result of a folded multply.
+    """
+    var b1 = lhs * byte_swap(rhs)
+    var b2 = byte_swap(lhs) * (~rhs)
+    return b1 ^ byte_swap(b2)
+
+
+@always_inline
+fn read_small(data: UnsafePointer[UInt8], length: Int) -> U128:
+    """Produce a SIMD[DType.uint64, 2] value from data which is smaller or equal 8 bytes.
```
JoeLoser, line 42:
```suggestion
    """Produce a `SIMD[DType.uint64, 2]` value from data which is smaller than or equal to `8` bytes.
```

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+            return U128(0, 0)
+
+
+struct AHasher:
+    var buffer: UInt64
+    var pad: UInt64
+    var extra_keys: U128
+
+    fn __init__(inout self, key: U256):
+        """Initialise the hasher with a key.
```
JoeLoser, line 79:
```suggestion
        """Initialize the hasher with a key.
```

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+
+        Args:
+            key: Modifier for the computaiton of final hash value.
```
JoeLoser, line 82:
```suggestion
            key: Modifier for the computation of the final hash value.
```

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+
+        Returns:
+            Final hash value.
+        """
+        var rot = self.buffer & 63
+        var folded = _folded_multiply(self.buffer, self.pad)
+        return (folded << rot) | (folded >> (64 - rot))
+
+    @always_inline
+    fn write(inout self, data: UnsafePointer[UInt8], length: Int):
```
JoeLoser, line 117: **Question** Have you consider changing this public API to work on a `Span` type instead of ptr + length constituents directly?  We'd pull them (the pointer and length) out in the body of this function where needed.

stdlib/src/builtin/_hash.mojo
JoeLoser: +1 to `update` per https://github.com/modularml/mojo/blob/nightly/proposals/improved-hash-module.md#introduce-a-hasher-trait anyways.  This is all subject to change as we don't have the trait yet, so just something to consider.  What do you think, @mzaks?

stdlib/benchmarks/builtin/bench_hash.mojo
```diff
@@ -0,0 +1,872 @@
...
+        var h = ahash(s.unsafe_ptr(), s.byte_length())
+        benchmark.keep(h)
+
+    b.iter[call_fn]()
+
+
+# ===----------------------------------------------------------------------===#
+# Benchmark Main
+# ===----------------------------------------------------------------------===#
+def main():
```
JoeLoser, line 811: Awesome work with adding the extensive benchmarks here! 🎉 🔥

stdlib/benchmarks/builtin/bench_hash.mojo
```diff
@@ -0,0 +1,872 @@
...
+            var w1 = w[].strip()
+            for w in list:
+                var w2 = w[].strip()
+                result.append(w1 + " " + w2)
+    except:
+        pass
+    return result
+
+
+# ===----------------------------------------------------------------------===#
```
JoeLoser, line 583: **Question** Instead of copying the impl here, can we do something like

```
from builtin._hash.mojo import AHash
```

stdlib/src/builtin/_hash.mojo
```diff
@@ -0,0 +1,179 @@
...
+            0x1319_8A2E_0370_7344,
+            0xA409_3822_299F_31D0,
+            0x082E_FA98_EC4E_6C89,
+        )
+        self.buffer = pi_key[0]
+        self.pad = pi_key[1]
+        self.extra_keys = U128(pi_key[2], pi_key[3])
+
+    @always_inline
+    fn large_update(inout self, new_data: U128):
```
JoeLoser, line 95: **Question** Is this intended to be public, or should we prefix with `_`?  See https://github.com/modularml/mojo/pull/3476/files#r1765504783

stdlib/test/builtin/test__hash.mojo
mzaks: Ah yes, the results of old hash are arch dependent (because of difference in SIMD width). I already observed some flotation here in CI and corrected for it. @JoeLoser I will change the value to be `0.59`

stdlib/benchmarks/builtin/bench_hash.mojo
mzaks: Hm, I just rebased and wanted to try this out, but it seems like I can't run benchmarks at all. I get many errors like this one: 
```
******************** TEST 'Mojo Standard Library Benchmarks :: utils/bench_formatter.mojo' FAILED ********************
Exit Code: 1

Command Output (stderr):
--
RUN: at line 14: mojo /Users/mzaks/dev/mojo/stdlib/benchmarks/utils/bench_formatter.mojo -t
+ mojo /Users/mzaks/dev/mojo/stdlib/benchmarks/utils/bench_formatter.mojo -t
/Users/mzaks/dev/mojo/stdlib/benchmarks/utils/bench_formatter.mojo:17:6: error: unable to locate module 'benchmark'
from benchmark import Bench, BenchConfig, Bencher, BenchId, Unit, keep, run
     ^
```

stdlib/src/builtin/_hash.mojo
mzaks: I like the idea. Should it be done in this PR or should we make a separate PR for it?

JoeLoser: Totally fine to do it as a follow up.

JoeLoser:
!sync

soraros:
You can use the `folded_mul` [here](https://github.com/modularml/mojo/pull/3434#issuecomment-2384556591).

JoeLoser:
!sync

stdlib/benchmarks/builtin/bench_hash.mojo
JoeLoser: Unfortunately, as a stdlib developer, that's a rough edge right now.  This is because, what you'd like to do, is have the `benchmark.mojopkg` get rebuilt (it depends on other parts of the stdlib that you may have modified, for example).  However, since its sources aren't available, it'll pick up the pre-built `benchmark.mojopkg` that ships with the Mojo nightly.

I've fixed up your benchmarks and removed the duplication of `AHash` and friends to just benchmark the old vs. new hash functions.  It should land today and we can iterate on future improvements (relocating the hash functions, defining the hasher trait, experimenting with replacing the existing hash function, changing the impl of `folded_multiply` to use the one from @soraros, etc.).

JoeLoser:
@mzaks FYI I have an internal PR up for review for introducing a "hashlib" top-level module that puts ahash and the current hash, hasher, etc. into there - along with moving its tests.

PR #3486, opened 2024-09-15
[stdlib] Make `UInt` `Hashable`

hayduker:
### Problem
Calling `hash` on a `UInt` value doesn't compile, e.g.

```mojo
def main():
    var x: UInt = 1234
    hash(x)
```

results in 

```
error: no matching function in call to 'hash'
    hash(u)
    ~~~~^~~
```

even though it does compile for `UInt8`, `UInt16`, `UInt32`, and `UInt64`. This means using `UInt` as a key to `Dict` doesn't compile either, for example

```mojo
from collections import Dict

def main():
    var d = Dict[UInt, String]()
```

results in

```
error: 'Dict' parameter #0 has 'KeyElement' type, but value has type 'AnyStruct[UInt]'
    var d = Dict[UInt, String]()
                 ^~~~
```

### Fix
This _does_ work for `Int8` through `Int64` as well as `Int` itself, so I essentially copied over the definition of `Int.__hash__` with the appropriate change to the underlying `DType` passed to `_hash_simd`, under the assumption it should work the same for `UInt`. Added a (very simple) test similar to that for `String.__hash__` too.

This seemed simple enough to go ahead and open a PR for, but it's my first contribution so if I need to open an issue first or am missing something obvious, please let me know!


JoeLoser (approved):
Looks good, thank you! This was just an oversight when we introduced the `UInt` type fairly recently.

JoeLoser:
Do you mind rebasing to resolve the merge conflicts, please?

dcsmith1212:
Closing this PR and opening a new one with my changes because I royally borked the rebase somehow.

PR #3495, opened 2024-09-18
[stdlib] Modify `String.__repr__` to handle multi-byte UTF-8 characters

### LINKED ISSUE #3456: [BUG] `__repr__` returns bytes soup for non ASCII, but valid UTF-8 chars
### Bug description

I did not expect non ASCII but valid UTF-8 strings to be returned as bytes soup. If this is expected behaviour, please delete this issue.

### Steps to reproduce

```Mojo
print("hello 🔥!".__repr__()) 
# prints: 'hello \xf0\x9f\x94\xa5!'
```

### System information

```shell
macOS 14.5
mojo 2024.9.805 (d7ccdc12)
modular 0.8.0 (39a426b5)
```

hayduker:
This PR fixes #3456 by modifying `String.__repr__` to handle multi-byte UTF-8 characters. The original code iterated over the string's internal bytes buffer and handled each one individually, causing the multi-byte characters to get emitted separately.

I took advantage of `String.__iter__` since it already has the logic internally to slice the string up into its UTF-8 characters, but I do have another implementation that still iterates over the byte buffer and pulls out the characters directly, if that solution is preferred for whatever reason.

The old definition for the module function `string.ascii` used to just call `String.__repr__` directly, so I copied the old definition of `String.__repr__` to `string.ascii` so that it still works the same.

### DIFF
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -202,7 +202,6 @@ fn _repr_ascii(c: UInt8) -> String:
             return hex(uc, prefix=r"\x")
 
 
-# TODO: This is currently the same as repr, should change with unicode strings
 @always_inline
 fn ascii(value: String) -> String:
     """Get the ASCII representation of the object.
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
Looks good, thanks for the contribution!

JoeLoser:
!sync

JoeLoser:
!sync

PR #3496, opened 2024-09-18
[stdlib] Add full unicode support for character casing functions

mzaks:
The code I used to generate the lookup tables can be found here https://gist.github.com/mzaks/bbadaeebcf81a5200021af041568b26b


stdlib/src/collections/_unicode.mojo
```diff
@@ -0,0 +1,265 @@
...
+
+    output[output_offset] = 0
+    var list = List[UInt8](
+        unsafe_pointer=output, size=(output_offset + 1), capacity=capacity
+    )
+    return String(list)
+
+
+fn to_uppercase(s: String) -> String:
+    """Returns a new string with all characters converted to upercase.
```
martinvuyk, line 209:
```suggestion
    """Returns a new string with all characters converted to uppercase.
```

stdlib/src/collections/_unicode.mojo
```diff
@@ -0,0 +1,265 @@
...
+
+    Args:
+        s: Input string.
+
+    Returns:
+        A new string where cased letters have been converted to upercase.
```
martinvuyk, line 215:
```suggestion
        A new string where cased letters have been converted to uppercase.
```

stdlib/src/collections/_unicode.mojo
```diff
@@ -0,0 +1,265 @@
...
+    var num_bytes_marker = UInt32(0xFF) << (8 - num_bytes)
+    p[0] = (((rune >> shift) & mask) | num_bytes_marker).cast[DType.uint8]()
+    for i in range(1, num_bytes):
+        shift -= 6
+        p[i] = (((rune >> shift) & 0b00111111) | 0b10000000).cast[DType.uint8]()
+    return num_bytes
+
+
+fn to_lowercase(s: String) -> String:
+    """Returns a new string with all characters converted to upercase.
```
martinvuyk, line 164:
```suggestion
    """Returns a new string with all characters converted to uppercase.
```

stdlib/src/collections/_unicode.mojo
```diff
@@ -0,0 +1,265 @@
...
+
+    var num_bytes = _utf8_len(rune)
+    var shift = 6 * (num_bytes - 1)
+    var mask = UInt32(0xFF) >> (num_bytes + 1)
+    var num_bytes_marker = UInt32(0xFF) << (8 - num_bytes)
+    p[0] = (((rune >> shift) & mask) | num_bytes_marker).cast[DType.uint8]()
+    for i in range(1, num_bytes):
+        shift -= 6
+        p[i] = (((rune >> shift) & 0b00111111) | 0b10000000).cast[DType.uint8]()
+    return num_bytes
```
martinvuyk, line 160: This code logic is used in many other places. I'll open a PR to fix `chr` implementation which I put in #3239 and ended up mixed with other problems and getting forgotten.

martinvuyk (commented):
Hi @mzaks this is great! . Just a bit of a nitpick: I think `_unicode.mojo` and `_unicode_lookups.mojo` should be in the utils package

JoeLoser:
!sync

mzaks:
> Hi @mzaks this is great! . Just a bit of a nitpick: I think `_unicode.mojo` and `_unicode_lookups.mojo` should be in the utils package

Very good point. Just moved them to utils.

JoeLoser:
!sync

mzaks:
I tried a performance improvement based on static B+Tree described here: https://en.algorithmica.org/hpc/data-structures/s-tree/

It looks promising, but I think it will be better as a separate PR. 
@JoeLoser what do you think?

martinvuyk:
> https://en.algorithmica.org/hpc/data-structures/s-tree/

what an awesome book, I have new reading material for a couple of months...

> I tried a performance improvement based on static B+Tree

I think in this case it might or it might not be better. The lists are static and known at compile time yet the binary search is generic. I think you could just hardcode values and ranges here, though I'm not sure how that much branching would affect performance. But there are some quite big ranges is some places e.g.:
```mojo
alias has_uppercase_mapping = List[UInt32, hint_trivial_type=True](
    ...
    0x007A,  # LATIN SMALL LETTER Z z
    0x00B5,  # MICRO SIGN µ
    0x00E0,  # LATIN SMALL LETTER A WITH GRAVE à
    ...
    0x0586,  # ARMENIAN SMALL LETTER FEH ֆ
    0x10D0,  # GEORGIAN LETTER AN ა
    ...
    0x2D2D,  # GEORGIAN SMALL LETTER AEN ⴭ
    0xA641,  # CYRILLIC SMALL LETTER ZEMLYA ꙁ
    ...
)
```
And if we are going to nitpick performance, I think ASCII letters should be prioritized and not use a generic algorithm for something that is realistically mostly going to be used for the first X amount of letters (we could use an ASCII optimized version and have another Unicode one, at the cost of checking the first UTF8 byte and the branch that follows)

JoeLoser:
!sync

JoeLoser:
> FYI I'm chasing after the issues internally that this PR is hitting, I'll have an update soon.

All merged now @mzaks!  Thanks for the contribution 🚀

JoeLoser:
> I tried a performance improvement based on static B+Tree described here: https://en.algorithmica.org/hpc/data-structures/s-tree/
> 
> It looks promising, but I think it will be better as a separate PR. @JoeLoser what do you think?

Agree re separate PR if you think there's perf optimizations to be had here. Always like keeping it simpler as we're getting started.

mzaks:
> > I tried a performance improvement based on static B+Tree described here: https://en.algorithmica.org/hpc/data-structures/s-tree/
> > It looks promising, but I think it will be better as a separate PR. @JoeLoser what do you think?
> 
> Agree re separate PR if you think there's perf optimizations to be had here. Always like keeping it simpler as we're getting started.

Yes I was also thinking along the same line, this PR is quite simple and provides a correct implementation of casing. I will create follow up PRs to introduce performance relevant changes.

PR #3497, opened 2024-09-18
[stdlib] Simplify `Int.__floordiv__` and friends

soraros:


stdlib/src/builtin/int.mojo
```diff
@@ -592,14 +592,11 @@ struct Int(
...
         """
         # This should raise an exception
-        var denominator = select(rhs == 0, 1, rhs)
-        var div: Int = self._positive_div(denominator)
-
+        var denom = select(rhs == 0, 1, rhs)
+        var div = self._positive_div(denom)
         var mod = self - div * rhs
-        var divMod = select(((rhs < 0) ^ (self < 0)) & mod, div - 1, div)
-        div = select(self > 0 & rhs > 0, div, divMod)
```
soraros, line 600: The first branch was never taken anyways, since
```agda
  self > 0 & rhs > 0
=
  self > (0 & rhs) > 0
=
  self > 0 > 0
=
  False
```

JoeLoser:
!sync

stdlib/src/builtin/int.mojo
```diff
@@ -630,17 +622,9 @@ struct Int(
...
-        if rhs == 0:
-            return 0, 0
-        var div: Int = self._positive_div(rhs)
-        if rhs > 0 & self > 0:
-            return div, self._positive_rem(rhs)
-        var mod = self - div * rhs
-        if ((rhs < 0) ^ (self < 0)) & mod:
-            return div - 1, mod + rhs
-        return div, mod
+        return self // rhs, self % rhs
```
JoeLoser, line 627: **Question** By recomputing things, won't this regress performance a bit?

soraros: Both `//` and `%` are marked as `always_inline`; I just assume that the compiler can see through the and do CSE well.

soraros: As a ref point, `UInt` also does this.

soraros: I checked and they generates identical assembly.

soraros:
@JoeLoser Gentle ping.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
Hey @soraros, I'd like to land this and there's some internal feedback:

> The point here is to duplicate to avoid some repeat computation.  For example,` var rem = self._positive_rem(denom)` might now be computed twice. Can we combine `div` and `mod` here to make sure we are not computing the same thing multiple times? 

I know you checked the assembly already.

soraros:
@JoeLoser Ok, I will inline those into `__divmod__` and dedup manually.

JoeLoser:
!sync

soraros:
Finally.

PR #3505, opened 2024-09-19
[Docs] Typo: "callee" should be "caller" 

gregsgit:
I think "callee" should be "caller" in "any living variable from the callee is not affected by the function."

Signed-off-by: Greg Sullivan <gregs@sulliwood.org>


scottamain (approved):
Thank you!

PR #3508, opened 2024-09-19
[stdlib] Implement difference between pointers

msaelices:
- [stdlib] Implement difference between pointers, so we have completed the pointers arithmetic we can do in C
- [docs] Document pointer arithmetic in pointers document

To complete the pointers arithmetic that we can do it in C

```
❯ mojo
  1> s = 100 
  2.  
(Int) s = 100
  2> p = UnsafePointer.address_of(s) 
  3.  
(UnsafePointer[Int, 0, 0, alignof[::AnyType,__mlir_type.!kgen.target]() if triple_is_nvidia_cuda() else 1]) p = {
  (pointer<_stdlib::_builtin::_int::_Int>) address = 0x000071858be84008
}
  3> p2 = p + 1 
  4.  
(UnsafePointer[Int, 0, 0, alignof[::AnyType,__mlir_type.!kgen.target]() if triple_is_nvidia_cuda() else 1]) p2 = {
  (pointer<_stdlib::_builtin::_int::_Int>) address = 0x000071858be84010
}
  4> offset = p2 - p 
  5.  
(Int) offset = 1  
```


JoeLoser (commented):
Can you add a unit test, please?

msaelices:
> Can you add a unit test, please?

Sure. Done: https://github.com/msaelices/mojo/commit/19830ecaf39eede047daa7e4d40ad42794931ea7

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
From some feedback internally on the PR:

The `sizeof[Int]()` might not be the same as `sizeof[Ptr]()` in Mojo so we need to make sure the code is not truncating the ptr types and doing the difference computation.

JoeLoser:
Thanks for the contribution here.  There's some ongoing compiler work to make this easier in the stdlib/more robust.  So, I'm going to close this PR and we'll implement this functionality once the compiler work is done.  In this case, I expect we'll have a `pop.pointer.diff` operation that we can call directly from the stdlib.  Appreciate your patience here 👍

PR #3509, opened 2024-09-19
[stdlib] Add `Arc.count()`

martinvuyk:
Add `Arc.count()`

### DIFF
--- a/stdlib/src/memory/arc.mojo
+++ b/stdlib/src/memory/arc.mojo
@@ -68,10 +68,6 @@ struct _ArcInner[T: Movable]:
         hits zero."""
         return self.refcount.fetch_sub(1) == 1
 
-    fn count(self) -> Int64:
-        """Count the amount of current references."""
-        return self.refcount.load()
-
 
 @register_passable
 struct Arc[T: Movable](CollectionElement, CollectionElementNew):
@@ -173,4 +169,4 @@ struct Arc[T: Movable](CollectionElement, CollectionElementNew):
         Returns:
             The current amount of references to the pointee.
         """
-        return self._inner[].count()
+        return self._inner[].refcount.load()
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):

JoeLoser:
!sync

martinvuyk:
Now that I moved  the implementation it fails because `load()` uses `fetch_add` which is a mutating method, no idea how it works when doing `self._inner[].refcount.load()`... might be a hack bypassing the borrow checker?

martinvuyk:
Oh so `UnsafePointer` let's you just change an alias attribute without having to do `inout self`... damn that's dangerous. But I guess it's needed :man_shrugging:

stdlib/src/memory/arc.mojo
```diff
@@ -162,3 +162,11 @@ struct Arc[T: Movable](CollectionElement, CollectionElementNew):
         """
         # TODO: consider removing this method.
         return UnsafePointer.address_of(self._inner[].payload)
+
+    fn count(self) -> Int64:
```
JoeLoser, line 166: Thoughts on `UInt` as the return type instead?  We know the number of references will always be non-negative.

martinvuyk: Conceptually I agree, but the datatype used in `_ArcInner` is a `DType.int64`, and I don't know if anything funny may happen during conversion. Do I just change it?

JoeLoser: Let's chase that type down after I land this, how's that sound?  It predates us having a `UInt` type, so that's why the inner type is the way it is I suspect.

martinvuyk: I changed it to UInt64 in the latest push hehe, theoretically this type  is just for internal counting... Do I revert the  change or do we see if it lands?

JoeLoser: !sync

JoeLoser:
!sync

PR #3513, opened 2024-09-21
[stdlib] Implement Python-like base64 encoding/decoding

hayduker:
This PR implements a few features to make Mojo's base64 encoding and decoding conform more closely to [Python's implementation](https://docs.python.org/3/library/base64.html#base64.b64encode):
- `b64decode` now ignores whitespace in the input, fixing #3446.
- Both `b64encode` and `b64decode` implement an optional `altchars` argument which allows the caller to substitute two characters for the default `+/`, useful for URL-safe encoding.
- `b64decode` implements an optional `validate` argument which, when `True`, will cause the function to raise an `Error` if the input contains any characters outside the alphabet. When `False` (the default), these characters are ignored like whitespace.
- `b64decode` now raises an `Error` if the input's length (after whitespace / invalid character removal) isn't divisible by 4.

Note about my comment about "leaky abstraction" in `b64encode`: I tried using `String.unsafe_ptr()` instead of `String._buffer` directly, since this was the function used for `StringLiteral` originally, but it does not return the correct underlying `UInt`s corresponding to the characters in the string. I ran into this issue switching from using a `StringLiteral` alias to `String` variable for the alphabet, since I need to modify the alphabet according to the `altchars` argument.

I'm not sure if this is a bug or just my misunderstanding `UnsafePointer`. Here is a small example that exhibits what I mean:

```mojo
def main():
    var l = "ABC"
    var lp = l.unsafe_ptr()
    print(lp[0])  # returns 65

    var s = String("ABC")
    var sp = s.unsafe_ptr()
    print(sp[0])  # returns 0
```

If this is a bug and I should open an issue for it, let me know.

### DIFF
--- a/stdlib/src/base64/base64.mojo
+++ b/stdlib/src/base64/base64.mojo
@@ -168,10 +168,10 @@ fn b64decode(
     var specials = _validate_altchars(altchars)
     base64_alphabet += specials
 
-    # Step 1: Remove ASCII whitespace from data
+    # Remove ASCII whitespace from data
     var encoded = _remove_whitespace(s)
 
-    # Step 3: Check for invalid characters
+    # Check for invalid characters
     var valid_chars = base64_alphabet + "="
     for char in encoded:
         if char not in valid_chars:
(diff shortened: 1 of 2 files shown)


soraros:
FYI, there is already #3443.

hayduker:
Oh cool, not sure how I missed that. Closing this.

PR #3519, opened 2024-09-21
[tools] Initial support for nix shell/develop

ThomasMader:
Nix shell/develop provides similar functionality to 'magic shell' but works with the Nix package manager (https://nixos.org/download/) and is generally used to manage an entire operating system.
Nix does not follow the FHS standard which means that magic does not work on systems using Nix out of the box and providing such a developer environment makes it easier for people using Nix to use and contribute to the software provided by Modular. On a system with a default Nix installation the shell can be entered by going to the mojo repository path and executing 'nix-shell'.
With flake support enabled it also works with 'nix develop'.

The Nix package manager is used in the NixOS Linux distribution and enables the configuration of the entire system declaratively in a functional language also called Nix.
But it can also be used on any other Linux distribution and macOS with the same guarantees as on NixOS with the exception that it uses the base system on the lowest layer in this case instead of it's own kernel.
With https://github.com/NixOS/nixpkgs there exists a repository for packaging software which is also a possible future path with the Modular products but currently I believe that it's not yet feasible.
I thought that the mojo repository is for now the best place to centralize the support for a Nix dev shell even though I realize that this is not something of high importance for Modular.

Nix might also be of interest because it could potentially also being used to build containers (https://nix.dev/tutorials/nixos/building-and-running-docker-images.html) and an increasing number of repositories provide a flake.nix file in the repository to package the current upstream version directly in comparison to the nixpkgs approach.

I know the chances are slim that this gets merged but I thought I will try.

I tested the shell on NixOS but was not able to test on macOS because I don't have access to it.

Anybody interested in trying it out can do so by installing the Nix package manager by following the instructions on https://nixos.org/download and running 'nix-shell' in the mojo repository directory with the nix files present.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,77 @@
+{
+  "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1696426674,
+        "narHash": "sha256-kvjfFW7WAETZlt09AgDn1MrtKzP7t90Vf7vypd3OL1U=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "0f9255e01c2351cc7d116c072cb317785dd33b33",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1726560853,
+        "narHash": "sha256-X6rJYSESBVr3hBoH0WbKE5KvhPU5bloyZ2L4K60/fPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "c1dfcf08411b08f6b8615f7d8971a2bfa81d5e8a",
+        "type": "github"
+      },
+      "original": {
+        "id": "flake-utils",
+        "type": "indirect"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1726755586,
+        "narHash": "sha256-PmUr/2GQGvFTIJ6/Tvsins7Q43KTMvMFhvG6oaYK+Wk=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "c04d5652cfa9742b1d519688f65d1bbccea9eb7e",
+        "type": "github"
+      },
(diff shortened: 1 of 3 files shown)


JoeLoser:
While I'd love to support Nix, I don't have any experience in this area, and so am reluctant to have our team maintain support for this right now.  Do you have any thoughts here, @zbowling?

noverby:
For anyone interested, I've successfully packaged Mojo for Nix without the need of emulating FHS! Check it out at:
[Mojo Nix Package](https://github.com/noverby/noverby/blob/main/packages/mojo.nix)

ThomasMader:
@noverby Thanks for notifying in here and doing that work.

A few questions to your solution.

Is there any reason you named it mojo instead of max, modular-max or something alike?
You seemed to have packaged everything not just the mojo parts.

Do you plan on upstreaming it into nixpkgs?

I was not able to test it out yet but since you didn't seem to include any patching for magic I wonder if all the dependency stuff works out like that.
I always imagined that a solution similar to the one used for Rust cargo in nixpkgs needs to be implemented.

What happens when the self update of magic is executed?

noverby:
@ThomasMader 

> Is there any reason you named it mojo instead of max, modular-max or something alike?

My primary interest was Mojo, so I named it after that. I could package Max also, but then it should be in a separate package.

> Do you plan on upstreaming it into nixpkgs?

I don't think the current license of Mojo/Max allow them to be included in Nixpkgs. 
If I'm mistaken, and it is allowed, then I would upstream it.

> What happens when the self update of magic is executed?

My Mojo package works without Magic. I could potentially create a tool that could read the Magic lockfile and install the dependencies using Nix.

ThomasMader:
> I don't think the current license of Mojo/Max allow them to be included in Nixpkgs.
If I'm mistaken, and it is allowed, then I would upstream it.

I hoped this has changed with the latest changes to the license but reading the version from April 12th, 2025 makes it clear that it's not allowed to redistribute outside of your own application.
Looks to me like we have to wait for the open source release for the upstream packaging to be allowed.

ThomasMader:
I close this PR because it's outdated.
I will also delete the source branch but the state of this PR is preserved in https://github.com/ThomasMader/modular/commit/955c9cc4f334d0643a82b1fc001e1c690ba74cb8.

That commit is included in branch https://github.com/ThomasMader/modular/tree/NixDevShell in which I updated the flake and I will continue to use it for future updates.

PR #3521, opened 2024-09-21
[stlib] String conversion of list literals and lists with representable items

msaelices:
To improve the ergonomics of the Mojo language for Python developers:

  ```mojo
  l = [1, False, "Mojo is awesome 🔥"]
  print(l)
  # [1, False, 'Mojo is awesome 🔥']
  l = List(1, 2, 3)
  print(str(l))
  # [1, 2, 3]
  ```


stdlib/test/builtin/test_str.mojo
```diff
@@ -16,7 +16,10 @@ from testing import assert_equal
 
 
 def test_str_none():
-    assert_equal(str(None), "None")
+    # TODO(#3393): Change to str(None) when MLIR types do not confuse overload resolution.
```
JoeLoser, line 19: **Question** Is this still valid since we have `NoneType` in the `stdlib` now?  We could give the conformance of `RepresentableCollectionElement` to it and the previous code may "just work"?

stdlib/src/builtin/builtin_list.mojo
```diff
@@ -80,6 +85,74 @@ struct ListLiteral[*Ts: CollectionElement](Sized, CollectionElement):
...
+
+        @parameter
+        for i in range(len(VariadicList(Ts))):
+            alias T = Ts[i]
+            var s: String
+            if i > 0:
+                writer.write(", ")
+
+            @parameter
+            if _type_is_eq[T, PythonObject]():
```
JoeLoser, line 136: @jackos would this sort of thing able to be improved with your in-flight PR for writer?  I don't love explicitly enumerating a bunch of types that we know are convertible to writable since we just store a `Tuple[*Ts]` and don't know if the `Ts` are writable themselves.

JoeLoser (commented):
This is great, thank you!

PR #3524, opened 2024-09-22
[stdlib] Create Box[] type as a new smart pointer

szbergeron:
This introduces a new Boxed[] type as an alternative to UnsafePointer[] to explicitly heap allocate large or slow to move/slow to copy types.


stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,97 @@
...
+    fn __del__(owned self: Boxed[T, AddressSpace.GENERIC]):
+        """Destroy the Boxed[]."""
+        # check that inner is non-null to accomodate into_inner and other
+        # consuming end states
+        if self._inner:
+            (self._inner).destroy_pointee()
+            self._inner.free()
+            self._inner = UnsafePointer[T, AddressSpace.GENERIC]()
+
+    fn into_inner[T: Movable](owned self: Boxed[T, AddressSpace.GENERIC]) -> T:
```
martinvuyk, line 96:
```suggestion
    fn __getitem__[T: Movable](owned self: Boxed[T, AddressSpace.GENERIC]) -> T:
```
I think this aligns a bit more to `Reference` syntax i.e. `something_to_deref[]`

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,112 @@
...
+        Parameters:
+            T: The type of the data to store.
+
+        Args:
+            copy_value: The value to explicitly copy into the Boxed[].
+        """
+        self._inner = UnsafePointer[T, address_space].alloc(1)
+        self._inner.init_pointee_explicit_copy(copy_value)
+
+    fn __init__[T: ExplicitlyCopyable](inout self: Boxed[T, AddressSpace.GENERIC], *, copy_box: Boxed[T, AddressSpace.GENERIC]):
```
martinvuyk, line 50: these are way over 80 characters, you can setup a ruler in your IDE for the docstrings, but the code should be auto formatted by mojo format when saving a file (If you set it up)

martinvuyk (commented):
Hi, thanks for the contribution. 

I'm not sure what the stdlib team's plans are with Reference and ownership of the underlying data, so this needs their input.

Personally, I like Rust's name `Box` more than `Boxed`. But the name I think does not do justice to what it really is: an `OwnedReference`

stdlib/src/memory/boxed.mojo
szbergeron: into_inner was something I made explicitly separate from the regular `__getitem__`, it has consume semantics instead. There is still a `__getitem__` to get a ref to the data above this.

stdlib/src/memory/boxed.mojo
szbergeron: I just need to spend a few minutes fixing the hook again \:)

stdlib/src/memory/boxed.mojo
martinvuyk: Oh sorry didn't read with attention. This might make more sense as a `take()` function then

szbergeron: I considered it, I'm not sure the semantics I'd associate with "take" make as much sense here, since it doesn't actually leave behind a "null box"--it fully consumes the original box, and requires moving it into the method. Maybe a `consume()`? I just took the Rust nomenclature for it and figured it was at least a reasonable first cut.

helehex: I'd recommend calling the type `OwnedPointer` or just `Owned`, and then call the function `take()`, `take_data()`, `unbox()`, but naming isn't too important right now, we can figure it out later.

helehex: Also, we might want to have an `unsafe_ptr()` method, (even though you can just access the field) it would follow the current stdlib conventions

gryznar:
> box

+1 for `Box`. As a noun it sounds better and name is compatible with current naming convention

gryznar:
> Hi, thanks for the contribution.
> 
> I'm not sure what the stdlib team's plans are with Reference and ownership of the underlying data, so this needs their input.
> 
> Personally, I like Rust's name `Box` more than `Boxed`. But the name I think does not do justice to what it really is: an `OwnedReference`

+1 for `Box`. As a noun it sounds better and name is compatible with current naming convention

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+    Parameters:
+        T: The type to be stored in the Boxed[].
+        address_space: The address space that the allocation behind this box resides in.
+    """
+
+    var _inner: UnsafePointer[T, address_space]
+
+    fn __init__[
+        T: Movable
+    ](inout self: Boxed[T, AddressSpace.GENERIC], owned value: T):
```
msaelices, line 33: @szbergeron why not relying on the `address_space` param, which is `AddressSpace.GENERIC` by default?

```suggestion
    ](inout self: Boxed[T, address_space], owned value: T):
```

stdlib/test/memory/test_boxed.mojo
```diff
@@ -0,0 +1,129 @@
...
+
+def main():
+    test_basic_ref()
+    test_box_copy_constructor()
+    test_moving_constructor()
+    test_copying_constructor()
+    test_basic_ref_mutate()
+    test_basic_del()
+    test_take()
+    test_moveinit()
```
msaelices, line 129: Great work! :tada:

stdlib/src/memory/boxed.mojo
szbergeron: This is waiting on design decisions regarding custom allocators. It's unclear what could substitute for `malloc` as a sane default for allocating in any addrspace other than GENERIC. I did want to at least stub that out, but a long discussion in https://discord.com/channels/1087530497313357884/1285357816030298112 mostly arrived at the conclusion that

1. Small allocations with Boxed[] don't make that much sense on things like GPUs (no good way of doing a traditional stack)
2. Custom allocators themselves need a bunch of design work to find the right abstraction
3. It might not make sense to have such an open constructor, instead of treating it has relatively hidden such that "deeper" libraries can return already constructed ones that nicely follow some rules for more efficient allocation (I could see Boxed[] being a relatively generic "this is a handle on some data that can be released using an allocator api")

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+from memory import UnsafePointer, stack_allocation, memcpy
+
+
+struct Boxed[T: AnyType, address_space: AddressSpace = AddressSpace.GENERIC]:
+    """A safe, owning, smart pointer.
+
+    This smart pointer is designed for cases where there is clear ownership
+    of the underlying data, and restricts access to it through the lifetime
+    system such that no more than one mutable alias for the underlying data
+    may exist. Consider Boxed[T] over UnsafePointer[T] where possible.
```
JoeLoser, line 22: I think the statement

> Consider Boxed[T] over UnsafePointer[T] where possible.

is a bit broad.  I'd recommend omitting it for now.  `Box` is a very specific use of `UnsafePointer`; I could see it generally replacing a pattern of `UnsafePointer[T].alloc(1)`.

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+from memory import UnsafePointer, stack_allocation, memcpy
+
+
+struct Boxed[T: AnyType, address_space: AddressSpace = AddressSpace.GENERIC]:
```
JoeLoser, line 16: **Suggestion** As mentioned as a point in the summarized review: let's drop this `address_space` parameter for now and just fix the held data (`_inner`) to use `AddressSpace.GENERIC` for now.  We can evaluate this lazily/add it when we have a clear use case for other address spaces.  You really want finer control over the allocations/deallocations, which this address space — in its current form, won't help you with.  We should wait until we have a proper design for custom allocators.

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+        """Construct a new Boxed[] by explicitly copying the value from another Boxed[].
+
+        Parameters:
+            T: The type of the data to store.
+
+        Args:
+            copy_box: The Boxed[] to copy.
+        """
+        self.__init__(copy_value=copy_box[])
+
```
JoeLoser, line 75: **Suggestion** Let's also add a constructor for `T: Copyable` (i.e. when `T` is implicitly copyable).  Right now `ExplicitCopyable` is in a half state (and a lot of the stdlib types aren't there yet), so to make `Box` useful out of the box (no pun intended 😆 ), let's add this additional constructor.

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+        Args:
+            existing: The value to move.
+        """
+        self._inner = existing._inner
+        existing._inner = UnsafePointer[T, address_space]()
+
+    fn __getitem__(
+        ref [_, address_space._value.value]self
+    ) -> ref [__lifetime_of(self._inner), address_space._value.value] T:
```
JoeLoser, line 87: @lattner will the lifetime system "do the right thing" here?  Is `__lifetime_of(self._inner)` the same as `__lifetime_of(self)` here?  How should we think about tying the return type of public APIs with a lifetime associated with a field (and a private field at that!) from a style perspective?

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+            ),
+        )
+
+        return self._inner[]
+
+    fn __del__(owned self: Boxed[T, AddressSpace.GENERIC]):
+        """Destroy the Boxed[]."""
+        # check that inner is non-null to accomodate take() and other
+        # consuming end states
+        if self._inner:
```
JoeLoser, line 114: **Suggestion** Remove this runtime branch - see my comment in the `take` method about using `lit.ownership.mark_destroyed`.

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+                since this consuming operation only makes sense for types that you want to avoid copying.
+                For types that are Copy or ExplicitlyCopy but are not Movable, you can copy them through
+                `__getitem__` as in `var v = some_box_var[]`.
+
+        Returns:
+            The data that is (was) backing the Boxed[].
+        """
+        var r = self._inner.take_pointee()
+        self._inner.free()
+        self._inner = UnsafePointer[T, AddressSpace.GENERIC]()
```
JoeLoser, line 141: **Suggestion** Instead of this, let's invoke MLIR directly to help the lifetime system out in the compiler so we can avoid the runtime branch in the destructor as a result of this.  Specifically,

remove `self._inner = UnsafePointer[T, AddressSpace.GENERIC]()` in favor of something like `__mlir_op.`lit.ownership.mark_destroyed`(__get_mvalue_as_litref(self))`.  You can see us employ this sort of trickery in `dict` [here](https://github.com/modularml/mojo/blob/d1b9b3419aa3dc6c66e628cce40684e9cfc2f987/stdlib/src/collections/dict.mojo#L228) for example.

stdlib/test/memory/test_boxed.mojo
```diff
@@ -0,0 +1,129 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo --debug-level full %s
```
JoeLoser, line 13:
```suggestion
# RUN: %mojo  %s
```

JoeLoser (commented):
Hey there, thanks for the contribution!  This is great to see coming together.  Internally, the team discussed this pull request at length in a full API design review this afternoon.  Here is some summarized feedback along with specific inline comments.

1. We prefer the name `Box` over `Boxed`.  It's shorter and the `ed` suffix doesn't provide a lot of additional value.  This is especially important for such common vocabulary types like `Box`.
2. Remove the `address_space` parameter on `Box` and just fix it to generic address space for the `UnsafePointer`.  Rationale: it's meaningless to provide this at the moment and only provides _confusion_ to users I'd argue.  In the future, once we design custom allocators, we can focus on this (perhaps address space will be a property on custom allocators anyways).  In short, let's lazily evaluate and add it later.
3. Let's have the invariant that `Box[T]` always has an initialized instance of `T`.  This will help us avoid the runtime branch in the destructor.  I'll comment inline on using `lit.ownership.mark_destroyed` in the consuming `take` method so we can avoid the destructor being run later (which is the need for the `if` check).

stdlib/src/memory/boxed.mojo
```diff
@@ -0,0 +1,143 @@
...
+        return self._inner[]
+
+    fn __del__(owned self: Boxed[T, AddressSpace.GENERIC]):
+        """Destroy the Boxed[]."""
+        # check that inner is non-null to accomodate take() and other
+        # consuming end states
+        if self._inner:
+            (self._inner).destroy_pointee()
+            self._inner.free()
+            self._inner = UnsafePointer[T, AddressSpace.GENERIC]()
```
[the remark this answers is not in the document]
szbergeron: Nope, just me keeping my head straight \:) I'll take that out

helehex:
Thanks for the `unsafe_ptr()` method!

gryznar:
@szbergeron title can also be adjusted from `Boxed` to `Box` :)

JoeLoser:
In 65008da7c2750172c75bb0a83957f26184f515a0, I think you meant to rename the test file rather than remove it. 😄

stdlib/src/memory/boxed.mojo
szbergeron: I'm not quite sure how I'd want to disambiguate the copying from the moving `__init__`, as I want people to reach for the moving `__init__` "by default" but also don't want the implicit copying constructor to require a whole song and dance just to call it.`Box(copy_value=foo)` is currently the explicit one, I could change that to `Box(explicit_copy_value=foo)` but that verges on mocking the user for daring to not `move`. Is there prior work for this that we'd like to conform to?

stdlib/test/memory/test_boxed.mojo
szbergeron: For types that may have debug asserts within them, do we not want to test with those all enabled?

JoeLoser:
Re the CI flakiness/failures, rebase now that https://github.com/modularml/mojo/pull/3545 just landed which should help with the stability.  Sorry about that.

stdlib/test/memory/test_boxed.mojo
JoeLoser: As surprising as it may sound, the `--debug-level full` is rather around creating the debug info in the IR.  We have some additional testing for this internally that parameterizes how `%mojo` expands, similar to how we do this in the open source CI for running with/without assertions.  See https://github.com/modularml/mojo/blob/936018ed2980488cd0e10da9298279e01c563137/.github/workflows/standard_library_tests_and_examples.yml#L31 for example - it gets propagated at https://github.com/modularml/mojo/blob/936018ed2980488cd0e10da9298279e01c563137/stdlib/test/lit.cfg.py#L73.

stdlib/src/memory/box.mojo
```diff
@@ -0,0 +1,128 @@
...
+
+        Args:
+            existing: The value to move.
+        """
+        self._inner = existing._inner
+        existing._inner = UnsafePointer[T]()
+
+    fn __getitem__(
+        ref [_, AddressSpace.GENERIC._value.value]self
+    ) -> ref [__lifetime_of(self._inner), AddressSpace.GENERIC._value.value] T:
```
JoeLoser, line 85: I spoke with @lattner today about https://github.com/modularml/mojo/pull/3524#discussion_r1772307879.  The TL;DR is that `__lifetime_of(self._inner)` won't work as you'd expect (modeling the lifetime of the inner data) today.  Can you please change it to `__lifetime_of(self)`?  Chris has some ideas around this topic of indirect lifetimes that will get implemented at some point, but for now, just use `__lifetime_of(self)`.

stdlib/src/memory/box.mojo
```diff
@@ -0,0 +1,128 @@
...
+        Parameters:
+            T: The type of the data to store.
+
+        Args:
+            copy_value: The value to explicitly copy into the Box[].
+        """
+        self._inner = UnsafePointer[T].alloc(1)
+        self._inner.init_pointee_explicit_copy(copy_value)
+
+    # TODO: disambiguation and other niceties
```
JoeLoser, line 54: Totally fine to deal with this in a follow-up, I think we can disambiguate by providing a phony parameter (e.g. `U: NoneType = None`) to deal with overload fitness but still allowing this.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
Do you mind adding a changelog entry for this great new contribution so new users know about this?  You'll want to add it [here](https://github.com/modularml/mojo/blob/936018ed2980488cd0e10da9298279e01c563137/docs/changelog.md#%EF%B8%8F-new).  Thanks!

stdlib/src/memory/box.mojo
szbergeron: Oh, ok! I feel like I changed it at some point to deal with the `address_space` thing not propagating correctly. Lemme fix that up.

szbergeron:
Do y'all want the lifetime fixes and copying constructor pushed to this branch? Can do that now if so, just don't want to confuse Copybara.

JoeLoser:
> Do y'all want the lifetime fixes and copying constructor pushed to this branch? Can do that now if so, just don't want to confuse Copybara.

Yep, feel free to push directly to this. I can resync at anytime, not a biggie.

JoeLoser:
!sync

szbergeron:
@JoeLoser for potentially composing this nicely with Optional, is there anything in the pipeline that could take advantage of somehow proving that _inner is never null? Ideally that could allow something similar to the "Option<Box<T>> is exactly sizeof a pointer" optimization from elsewhere

PR #3525, opened 2024-09-22
[stdlib] Add `Optional` unwrap using `__getitem__ `

martinvuyk:
Add `Optional` unwrap using `__getitem__ `

### DIFF
--- a/mojo/stdlib/stdlib/collections/optional.mojo
+++ b/mojo/stdlib/stdlib/collections/optional.mojo
@@ -256,6 +256,20 @@ struct Optional[T: Copyable & Movable](
         """
         return not self
 
+    @always_inline
+    fn __getitem__(ref self) raises -> ref [self._value] T:
+        """Retrieve a reference to the value inside the `Optional`.
+
+        Returns:
+            A reference to the value inside the `Optional`.
+
+        Raises:
+            On empty `Optional`.
+        """
+        if not self:
+            raise Error(".value() on empty Optional")
+        return self.value()
+
     fn __str__[
         U: Copyable & Movable & Representable, //
     ](self: Optional[U]) -> String:
--- a/mojo/stdlib/test/collections/test_optional.mojo
+++ b/mojo/stdlib/test/collections/test_optional.mojo
@@ -170,6 +170,15 @@ def test_optional_copied():
     assert_equal(opt_owned.value(), String("foo"))
 
 
+def test_optional_unwrap():
+    var a = Optional(123)
+    assert_true(a)
+    assert_equal(123, a[])
+    a = Optional[Int](None)
+    with assert_raises(contains="on empty Optional"):
+        _ = a[]
+
+
 def main():
     test_basic()
     test_optional_reg_basic()
@@ -182,3 +191,4 @@ def main():
     test_optional_str_repr()
     test_optional_equality()
     test_optional_copied()
+    test_optional_unwrap()


helehex:
Typically in the stdlib, `__getitem__()` returns a ref, not an owned value. I'm ok with this and might even prefer it, but we might want to consider if it's worth keeping it consistent throughout the stdlib, or where to break from it (maybe this is a case?)

martinvuyk:
Yeah I also think this breaks away from convention but in this case I think it makes sense since if you call a function and want to unwrap it doing `some_func()^` feels weird and `some_func()[][]` does not spark joy. Saying `some_func()[]` returns a reference that gets auto dereferenced also feels convoluted IMO. And logically speaking if you are unwrapping an Optional you most probably want to consume it taking ownership of the value not a reference of it :man_shrugging: . But honestly idc if it's a reference or a value as long as I can just do `some_func()[]` irresponsibly in raising functions, also no idea what kind of performance difference there might be.

stdlib/src/collections/optional.mojo
```diff
@@ -379,6 +379,21 @@ struct Optional[T: CollectionElement](
...
+
+        Returns:
+            The value inside the Optional.
+
+        Raises:
+            ValueError on empty Optional.
+        """
+
+        if not self:
+            raise Error("ValueError: take() on empty Optional")
```
gryznar: To the stdlib team: Shall we have the same error system as Python has?
All ValueError, TypeError, etc? Python errors are kind of (expensive) control flow. I am not sure if it is the greatest idea here to treat unwrapping as recoverable

gryznar: In message `take()` is quite missleading, as underneath `unsafe_take()` is used. Maybe better would be sth like "unwrap empty Optonal"?

stdlib/src/collections/optional.mojo
```diff
@@ -379,6 +379,21 @@ struct Optional[T: CollectionElement](
             return self._value[T]
         return default
 
+    fn __getitem__(owned self) raises -> T:
```
lsh: I would rather this not be consuming (i.e return `ref T`). If one needs to `take` the value, they can.

lsh (changes requested):

stdlib/src/collections/optional.mojo
martinvuyk: True I think this PR was way back before we had `ref [_]`

martinvuyk: I also made it use `.value()` now because you're right that it might be best; just thinking about people using this the same as a pointer and referencing the value several times and it unexpectedly exploding in their faces is not the best

mojo/stdlib/stdlib/collections/optional.mojo
```diff
@@ -256,6 +256,18 @@ struct Optional[T: Copyable & Movable](
...
+    fn __getitem__(ref self) raises -> ref [self._value] T:
+        """Retrieve a reference to the value inside the `Optional`.
+
+        Returns:
+            A reference to the value inside the `Optional`.
+
+        Raises:
+            On empty `Optional`.
+        """
+        return self.value()
```
lsh, line 271: I think this needs a branch and to raise if None

lsh (changes requested):

mojo/stdlib/stdlib/collections/optional.mojo
martinvuyk: yeah forgot about it

lsh:
!sync

Issue #3526, opened 2024-09-22
[BUG] Fix `String.find()` family of functions to work by unicode codepoints


martinvuyk:
### Bug description

Fix `String.find()` family of functions to work by unicode codepoints

Indirectly related to comment/issue [#3246](https://github.com/modularml/mojo/issues/3246#issuecomment-2246123195)

### Steps to reproduce

...

### System information

```shell
...
```

PR #3528, opened 2024-09-22
[stdlib] Fix `String.split()` implementations

### LINKED ISSUE #3635: [Feature Request] Make `String.split` with `sep` non-raising
### Review Mojo's priorities

### What is your request?

As title.

It's quite annoying to use in non-raising contexts.

### Any other details?

We could/should:
1. Change `raise Error("Separator cannot be empty.")` to `abort(...)` in the short term
2. Adopt the semantics of Rust when `sep == ""`: `"abc".split(sep="") == ["", "a", "b", "c", ""]`

It would be different from Python, but in this particular case, I don't think it's a bad thing.

martinvuyk:
#### Main issue

Fix `String.split()` implementations to use a generic implementation and without assuming that indexing is by byte offset. Some important optimizations were added by parametrizing and avoiding slicing with numeric tricks.

#### Changes in behavior

This PR changes `split("")` behavior to be non-raising and return the separated unicode characters analogous to when the whole string has the separator at start, end, and in between every character. Closes #3635

#### Benchmark results:

CPU: Intel® Core™ i7-7700HQ

improvement metric: markdown percentage improvement (`(old_value - new_value) / old_value`)

Average improvement for split with a sequence: 91.2486% . In orders of magnitude, this is a 11x improvement
Average improvement for split on any whitespace: 99.9975% . In orders of magnitude, this is a 40k x improvement

|Name                          | old_value (ms)           | new_value (ms)      | improvement|
|:-----------------------------|---------------------:|----------------:|------:|
|`bench_string_split[1000000]` | 5.6970877871796 | 0.503624904955618 | 0.912486298354641|
|`bench_string_split_none[1000000]` | 147157.6808882 | 3.78480079182398 | 0.999975287456096|

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_string.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_string.mojo
@@ -17,6 +17,7 @@
 
 from collections.string import String
 from collections.string._utf8 import _is_valid_utf8
+from collections.string.string_slice import _split
 from memory import stack_allocation
 from os import abort
 from pathlib import _dir_of_current_file
@@ -106,18 +107,20 @@ fn bench_string_split[
     filename: StaticString = "UN_charter_EN",
     sequence: Optional[StaticString] = None,
 ](mut b: Bencher) raises:
-    var items = make_string[length](filename + ".txt")
+    var items = (
+        make_string[length](filename + ".txt").as_string_slice().get_immutable()
+    )
 
     @always_inline
     @parameter
     fn call_fn() raises:
-        var res: List[String]
+        var res: List[__type_of(items)]
 
         @parameter
         if sequence:
-            res = items.split(sequence.value())
+            res = _split[has_maxsplit=False](items, sequence.value(), -1)
         else:
-            res = items.split()
+            res = _split[has_maxsplit=False](items, None, -1)
         keep(res.data)
 
     b.iter[call_fn]()
--- a/mojo/stdlib/stdlib/collections/string/string.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string.mojo
@@ -1321,8 +1321,7 @@ struct String(
         """
         return self.as_string_slice().isspace()
 
-    # TODO(MSTDL-590): String.split() should return `StringSlice`s.
-    fn split(self, sep: StringSlice, maxsplit: Int = -1) raises -> List[String]:
+    fn split(self, sep: StringSlice, maxsplit: Int = -1) -> List[String]:
         """Split the string by a separator.
 
         Args:
@@ -1333,9 +1332,6 @@ struct String(
         Returns:
             A List of Strings containing the input split by the separator.
 
-        Raises:
-            If the separator is empty.
-
         Examples:
 
         ```mojo
@@ -1345,6 +1341,8 @@ struct String(
         _ = String("hello,,world").split(",") # ["hello", "", "world"]
         # Splitting with maxsplit
         _ = String("1,2,3").split(",", 1) # ['1', '2,3']
+        # Splitting with an empty separator
+        _ = StringSlice("123").split("") # ["", "1", "2", "3", ""]
         ```
         .
         """
--- a/mojo/stdlib/stdlib/collections/string/string_slice.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string_slice.mojo
@@ -1161,139 +1161,6 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
             res.append_byte(self_ptr[i])
         return res^
 
-    fn split(
-        self, sep: StringSlice, maxsplit: Int = -1
-    ) raises -> List[Self.Immutable]:
-        """Split the string by a separator.
-
-        Args:
-            sep: The string to split on.
-            maxsplit: The maximum amount of items to split from String.
-                Defaults to unlimited.
-
-        Returns:
-            A List of Strings containing the input split by the separator.
-
-        Raises:
-            If the separator is empty.
-
-        Examples:
-
-        ```mojo
-        # Splitting a space
-        _ = StringSlice("hello world").split(" ") # ["hello", "world"]
-        # Splitting adjacent separators
-        _ = StringSlice("hello,,world").split(",") # ["hello", "", "world"]
-        # Splitting with maxsplit
-        _ = StringSlice("1,2,3").split(",", 1) # ['1', '2,3']
-        ```
-        """
-        var output = List[Self.Immutable]()
-
-        var str_byte_len = self.byte_length() - 1
-        var lhs = 0
-        var items = 0
-        var sep_len = sep.byte_length()
-        if sep_len == 0:
-            raise Error("Separator cannot be empty.")
-        if str_byte_len < 0:
-            output.append(rebind[Self.Immutable](StaticString("")))
-
-        while lhs <= str_byte_len:
-            var rhs = self.find(sep, lhs)
-            if rhs == -1:
-                output.append(rebind[Self.Immutable](self[lhs:]))
-                break
-
-            if maxsplit > -1:
-                if items == maxsplit:
-                    output.append(rebind[Self.Immutable](self[lhs:]))
-                    break
-                items += 1
-
-            output.append(rebind[Self.Immutable](self[lhs:rhs]))
-            lhs = rhs + sep_len
-
-        if self.endswith(sep) and (len(output) <= maxsplit or maxsplit == -1):
-            output.append(rebind[Self.Immutable](StaticString("")))
-
-        return output^
-
-    fn split(
-        self, sep: NoneType = None, maxsplit: Int = -1
-    ) -> List[Self.Immutable]:
-        """Split the string by every Whitespace separator.
-
-        Args:
-            sep: None.
-            maxsplit: The maximum amount of items to split from String. Defaults
-                to unlimited.
-
-        Returns:
-            A List of Strings containing the input split by the separator.
-
-        Examples:
-
-        ```mojo
-        # Splitting an empty string or filled with whitespaces
-        _ = StringSlice("      ").split() # []
-        _ = StringSlice("").split() # []
-
-        # Splitting a string with leading, trailing, and middle whitespaces
-        _ = StringSlice("      hello    world     ").split() # ["hello", "world"]
-        # Splitting adjacent universal newlines:
-        _ = StringSlice(
-            "hello \\t\\n\\v\\f\\r\\x1c\\x1d\\x1e\\x85\\u2028\\u2029world"
-        ).split()  # ["hello", "world"]
-        ```
-        """
-
-        return self._split_whitespace()
-
-    fn _split_whitespace(self, maxsplit: Int = -1) -> List[Self.Immutable]:
-        fn num_bytes(b: UInt8) -> Int:
-            var flipped = ~b
-            return Int(count_leading_zeros(flipped) + (flipped >> 7))
-
-        var output = List[Self.Immutable]()
-        var str_byte_len = self.byte_length() - 1
-        var lhs = 0
-        var items = 0
-        while lhs <= str_byte_len:
-            # Python adds all "whitespace chars" as one separator if no
-            # separator was specified.
-            for s in self[lhs:].codepoint_slices():
-                if not s.isspace():
-                    break
-                lhs += s.byte_length()
-            # If it went until the end of the String, then it should be sliced
-            # up until the original start of the whitespace which was already
-            # appended.
-            if lhs - 1 == str_byte_len:
-                break
-            elif lhs == str_byte_len:
-                # If the last char is not whitespace
-                output.append(rebind[Self.Immutable](self[str_byte_len:]))
-                break
-            var rhs = lhs + num_bytes(self.unsafe_ptr()[lhs])
-            for s in self[
-                lhs + num_bytes(self.unsafe_ptr()[lhs]) :
-            ].codepoint_slices():
-                if s.isspace():
-                    break
-                rhs += s.byte_length()
-
-            if maxsplit > -1:
-                if items == maxsplit:
-                    output.append(rebind[Self.Immutable](self[lhs:]))
-                    break
-                items += 1
-
-            output.append(rebind[Self.Immutable](self[lhs:rhs]))
-            lhs = rhs
-
-        return output^
-
     @always_inline
     fn strip(self, chars: StringSlice) -> Self:
         """Return a copy of the string with leading and trailing characters
@@ -1910,12 +1777,16 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
 
         return Int(loc) - Int(self.unsafe_ptr())
 
-    fn isspace(self) -> Bool:
+    fn isspace[single_character: Bool = False](self) -> Bool:
         """Determines whether every character in the given StringSlice is a
         python whitespace String. This corresponds to Python's
         [universal separators](
         https://docs.python.org/3/library/stdtypes.html#str.splitlines):
          `" \\t\\n\\v\\f\\r\\x1c\\x1d\\x1e\\x85\\u2028\\u2029"`.
+
+        Parameters:
+            single_character: Whether to evaluate the stringslice as a single
+                unicode character (avoids overhead when already iterating).
 
         Returns:
             True if the whole StringSlice is made up of whitespace characters
@@ -1926,7 +1797,7 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
         Check if a string contains only whitespace:
 
         ```mojo
-        from testing import assert_true, assert_false
+        %# from testing import assert_true, assert_false
 
         # An empty string is not considered to contain only whitespace chars:
         assert_false(StringSlice("").isspace())
@@ -1940,14 +1811,110 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
         ```
         """
 
-        if self.byte_length() == 0:
+        fn _is_space_char(s: StringSlice) -> Bool:
+            # sorry for readability, but this has less overhead than memcmp
+            # highly performance sensitive code, benchmark before touching
+            alias ` ` = UInt8(ord(" "))
+            alias `\t` = UInt8(ord("\t"))
+            alias `\n` = UInt8(ord("\n"))
+            alias `\r` = UInt8(ord("\r"))
+            alias `\f` = UInt8(ord("\f"))
+            alias `\v` = UInt8(ord("\v"))
+            alias `\x1c` = UInt8(ord("\x1c"))
+            alias `\x1d` = UInt8(ord("\x1d"))
+            alias `\x1e` = UInt8(ord("\x1e"))
+
+            var no_null_len = s.byte_length()
+            var ptr = s.unsafe_ptr()
+            if likely(no_null_len == 1):
+                var c = ptr[0]
+                return (
+                    c == ` `
+                    or c == `\t`
+                    or c == `\n`
+                    or c == `\r`
+                    or c == `\f`
+                    or c == `\v`
+                    or c == `\x1c`
+                    or c == `\x1d`
+                    or c == `\x1e`
+                )
+            elif no_null_len == 2:
+                return ptr[0] == 0xC2 and ptr[1] == 0x85  # next_line: \x85
+            elif no_null_len == 3:
+                # unicode line sep or paragraph sep: \u2028 , \u2029
+                var last_byte = ptr[2] == 0xA8 or ptr[2] == 0xA9
+                return ptr[0] == 0xE2 and ptr[1] == 0x80 and last_byte
             return False
 
-        for s in self.codepoints():
-            if not s.is_python_space():
-                return False
+        @parameter
+        if single_character:
+            return _is_space_char(self)
+        else:
+            for s in self.codepoint_slices():
+                if not _is_space_char(s):
+                    return False
+            return self.byte_length() != 0
 
-        return True
+    fn split(
+        self, sep: StringSlice, maxsplit: Int = -1
+    ) -> List[Self.Immutable]:
+        """Split the string by a separator.
+
+        Args:
+            sep: The string to split on.
+            maxsplit: The maximum amount of items to split from String.
+                Defaults to unlimited.
+
+        Returns:
+            A List of Strings containing the input split by the separator.
+
+        Examples:
+
+        ```mojo
+        # Splitting a space
+        _ = StringSlice("hello world").split(" ") # ["hello", "world"]
+        # Splitting adjacent separators
+        _ = StringSlice("hello,,world").split(",") # ["hello", "", "world"]
+        # Splitting with maxsplit
+        _ = StringSlice("1,2,3").split(",", 1) # ['1', '2,3']
+        # Splitting with an empty separator
+        _ = StringSlice("123").split("") # ["", "1", "2", "3", ""]
+        ```
+        .
+        """
+        return _split[has_maxsplit=True](self.get_immutable(), sep, maxsplit)
+
+    fn split(
+        self, sep: NoneType = None, maxsplit: Int = -1
+    ) -> List[Self.Immutable]:
+        """Split the string by every Whitespace separator.
+
+        Args:
+            sep: None.
+            maxsplit: The maximum amount of items to split from String. Defaults
+                to unlimited.
+
+        Returns:
+            A List of Strings containing the input split by the separator.
+
+        Examples:
+
+        ```mojo
+        # Splitting an empty string or filled with whitespaces
+        _ = StringSlice("      ").split() # []
+        _ = StringSlice("").split() # []
+
+        # Splitting a string with leading, trailing, and middle whitespaces
+        _ = StringSlice("      hello    world     ").split() # ["hello", "world"]
+        # Splitting adjacent universal newlines:
+        _ = StringSlice(
+            "hello \\t\\n\\v\\f\\r\\x1c\\x1d\\x1e\\x85\\u2028\\u2029world"
+        ).split()  # ["hello", "world"]
+        ```
+        .
+        """
+        return _split[has_maxsplit=True](self.get_immutable(), sep, maxsplit)
 
     fn isnewline[single_character: Bool = False](self) -> Bool:
         """Determines whether every character in the given StringSlice is a
@@ -2297,10 +2264,10 @@ fn get_static_string[
 
 
 fn _to_string_list[
-    O: ImmutableOrigin, //,
+    O: Origin, //,
     T: Copyable & Movable,  # TODO(MOCO-1446): Make `T` parameter inferred
     len_fn: fn (T) -> Int,
-    unsafe_ptr_fn: fn (T) -> UnsafePointer[Byte, mut=False, origin=O],
+    unsafe_ptr_fn: fn (T) -> UnsafePointer[Byte, mut = O.mut, origin=O],
 ](items: List[T]) -> List[String]:
     var i_len = len(items)
 
@@ -2315,9 +2282,7 @@ fn _to_string_list[
 
 
 @always_inline
-fn _to_string_list[
-    O: ImmutableOrigin, //
-](items: List[StringSlice[O]]) -> List[String]:
+fn _to_string_list[O: Origin](items: List[StringSlice[O]]) -> List[String]:
     """Create a list of Strings **copying** the existing data.
 
     Parameters:
@@ -2332,7 +2297,7 @@ fn _to_string_list[
 
     fn unsafe_ptr_fn(
         v: StringSlice[O],
-    ) -> UnsafePointer[Byte, mut=False, origin=O]:
+    ) -> UnsafePointer[Byte, mut = O.mut, origin=O]:
         return v.unsafe_ptr()
 
     fn len_fn(v: StringSlice[O]) -> Int:
@@ -2342,9 +2307,7 @@ fn _to_string_list[
 
 
 @always_inline
-fn _to_string_list[
-    O: ImmutableOrigin, //
-](items: List[Span[Byte, O]]) -> List[String]:
+fn _to_string_list[O: Origin](items: List[Span[Byte, O]]) -> List[String]:
     """Create a list of Strings **copying** the existing data.
 
     Parameters:
@@ -2359,7 +2322,7 @@ fn _to_string_list[
 
     fn unsafe_ptr_fn(
         v: Span[Byte, O]
-    ) -> UnsafePointer[Byte, mut=False, origin=O]:
+    ) -> UnsafePointer[Byte, mut = O.mut, origin=O]:
         return v.unsafe_ptr()
 
     fn len_fn(v: Span[Byte, O]) -> Int:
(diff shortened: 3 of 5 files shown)


stdlib/src/collections/string.mojo
```diff
@@ -1565,25 +1565,34 @@ struct String(
             raise Error("Separator cannot be empty.")
         if str_byte_len < 0:
             output.append("")
+        var ptr = self.unsafe_ptr()
+        alias SelfSlice = StringSlice[__lifetime_of(self)]
+
+        @always_inline("nodebug")
+        fn build_slice(s_ptr: UnsafePointer[UInt8], length: Int) -> SelfSlice:
```
msaelices, line 1572: NIT: Just for convention, I think it's better to prefix the function with `_` to improve readability, implying that it's not a public function imported from somewhere:
```suggestion
        fn _build_slice(s_ptr: UnsafePointer[UInt8], length: Int) -> SelfSlice:
```

stdlib/src/collections/string.mojo
```diff
@@ -1565,25 +1565,34 @@ struct String(
...
+            return SelfSlice(unsafe_from_utf8_ptr=s_ptr, len=length)
 
         while lhs <= str_byte_len:
+            # FIXME(#3295): this will fail when find is changed to use unicode codepoints
             rhs = self.find(sep, lhs)
             if rhs == -1:
-                output.append(self[lhs:])
+                var s = String(build_slice(ptr + lhs, str_byte_len - lhs + 1))
```
msaelices, line 1579: NIT:
```suggestion
                var s = String(_build_slice(ptr + lhs, str_byte_len - lhs + 1))
```

stdlib/src/collections/string.mojo
```diff
@@ -1565,25 +1565,34 @@ struct String(
...
+                output.append(s^)
                 break
 
             if maxsplit > -1:
                 if items == maxsplit:
-                    output.append(self[lhs:])
+                    var s = build_slice(ptr + lhs, str_byte_len - lhs + 1)
```
msaelices, line 1585: NIT:
```suggestion
                    var s = _build_slice(ptr + lhs, str_byte_len - lhs + 1)
```

stdlib/src/collections/string.mojo
```diff
@@ -1565,25 +1565,34 @@ struct String(
...
+                    output.append(String(s^))
                     break
                 items += 1
 
-            output.append(self[lhs:rhs])
+            output.append(String(build_slice(ptr + lhs, rhs - lhs)))
```
msaelices, line 1590: NIT:
```suggestion
            output.append(String(_build_slice(ptr + lhs, rhs - lhs)))
```

stdlib/src/collections/string.mojo
```diff
@@ -1565,25 +1565,34 @@ struct String(
...
             lhs = rhs + sep_len
 
         if self.endswith(sep) and (len(output) <= maxsplit or maxsplit == -1):
             output.append("")
-        return output
+        return output^
```
msaelices, line 1595: :1st_place_medal:

stdlib/src/collections/string.mojo
```diff
@@ -1613,47 +1622,50 @@ struct String(
...
         var output = List[String]()
         var str_byte_len = self.byte_length() - 1
         var lhs = 0
         var rhs = 0
         var items = 0
+        var ptr = self.unsafe_ptr()
+        alias SelfSlice = StringSlice[__lifetime_of(self)]
+
+        @always_inline("nodebug")
+        fn build_slice(s_ptr: UnsafePointer[UInt8], length: Int) -> SelfSlice:
```
msaelices, line 1634: NIT:

```suggestion
        fn _build_slice(s_ptr: UnsafePointer[UInt8], length: Int) -> SelfSlice:
```

stdlib/src/collections/string.mojo
```diff
@@ -1613,47 +1622,50 @@ struct String(
...
+            return SelfSlice(unsafe_from_utf8_ptr=s_ptr, len=length)
+
         while lhs <= str_byte_len:
             # Python adds all "whitespace chars" as one separator
             # if no separator was specified
-            for s in self[lhs:]:
-                if not str(s).isspace():  # TODO: with StringSlice.isspace()
+            for s in build_slice(ptr + lhs, str_byte_len - lhs):
```
msaelices, line 1640: NIT:
```suggestion
            for s in _build_slice(ptr + lhs, str_byte_len - lhs):
```

stdlib/src/collections/string.mojo
```diff
@@ -1613,47 +1622,50 @@ struct String(
...
-            elif lhs == str_byte_len:
-                # if the last char is not whitespace
-                output.append(self[str_byte_len])
-                break
-            rhs = lhs + num_bytes(self.unsafe_ptr()[lhs])
-            for s in self[lhs + num_bytes(self.unsafe_ptr()[lhs]) :]:
-                if str(s).isspace():  # TODO: with StringSlice.isspace()
+            var b_len = _utf8_first_byte_sequence_length(ptr[lhs])
+            rhs = lhs + b_len
+            for s in build_slice(ptr + rhs, str_byte_len - rhs):
```
msaelices, line 1651: NIT:
```suggestion
            for s in _build_slice(ptr + rhs, str_byte_len - rhs):
```

stdlib/src/collections/string.mojo
```diff
@@ -1613,47 +1622,50 @@ struct String(
...
+                if s.isspace():
                     break
                 rhs += s.byte_length()
+            if rhs == str_byte_len:
+                rhs += 1
 
             if maxsplit > -1:
                 if items == maxsplit:
-                    output.append(self[lhs:])
+                    var s = String(build_slice(ptr + lhs, str_byte_len - lhs))
```
msaelices, line 1660: NIT:
```suggestion
                    var s = String(_build_slice(ptr + lhs, str_byte_len - lhs))
```

stdlib/src/collections/string.mojo
```diff
@@ -1613,47 +1622,50 @@ struct String(
...
+                    output.append(s^)
                     break
                 items += 1
 
-            output.append(self[lhs:rhs])
+            output.append(String(build_slice(ptr + lhs, rhs - lhs)))
```
msaelices, line 1665: NIT:
```suggestion
            output.append(String(_build_slice(ptr + lhs, rhs - lhs)))
```

msaelices (approved):
Great job. I've just add a NIT-pick suggestion.

Also, is it possible to add a unit test?

stdlib/src/collections/string.mojo
martinvuyk: It's still defined withing the scope of the function, so I didn't see a need to make it's name "private" since no other func can access it. But I'll change them if you want to :man_shrugging:

martinvuyk:
> Great job. I've just add a NIT-pick suggestion.
> 
> Also, is it possible to add a unit test?

Hi, thanks for the review. Any type of test in mind that [split tescases](https://github.com/modularml/mojo/blob/4fb9c7a68743ae88a1c4fd31cef6d1175b9ffd0e/stdlib/test/collections/test_string.mojo#L685) don't  cover ?

.gitignore
```diff
@@ -16,6 +16,8 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+.magic/
```
JoeLoser, line 19: **Suggestion** Can you pull this into a separate PR that I can instant approve/merge, please?  Otherwise, when I sync this PR internally, since `.gitignore` is not managed by Copybara, it will get undone during then nightly export.

stdlib/src/collections/string.mojo
```diff
@@ -1613,47 +1622,50 @@ struct String(
...
         var output = List[String]()
         var str_byte_len = self.byte_length() - 1
         var lhs = 0
         var rhs = 0
         var items = 0
+        var ptr = self.unsafe_ptr()
+        alias SelfSlice = StringSlice[__lifetime_of(self)]
+
+        @always_inline("nodebug")
+        fn _build_slice(s_ptr: UnsafePointer[UInt8], length: Int) -> SelfSlice:
```
JoeLoser, line 1634: **Question** Have you considered having this private function work on a `Span` instead of continuing with ptr + length?  I haven't checked to see if `StringSlice` has a constructor from a `Span`, but if not, we could pull the ptr + length out of the span when needed.  It helps tighten up the types until we absolutely need to work with ptr + length bare.  WDYT?

stdlib/src/utils/string_slice.mojo
```diff
@@ -64,6 +65,15 @@ fn _shift_unicode_to_utf8(ptr: UnsafePointer[UInt8], c: Int, num_bytes: Int):
         ptr[i] = ((c >> shift) & 0b0011_1111) | 0b1000_0000
 
 
+@always_inline("nodebug")
+fn _utf8_first_byte_sequence_length(b: UInt8) -> Int:
+    """Get the length of the sequence starting with given byte. Do note that
```
JoeLoser, line 70: **Question** Should we add a `debug_assert` that the given byte is not a continuation byte then?

msaelices:
> > Great job. I've just add a NIT-pick suggestion.
> > Also, is it possible to add a unit test?
> 
> Hi, thanks for the review. Any type of test in mind that [split tescases](https://github.com/modularml/mojo/blob/4fb9c7a68743ae88a1c4fd31cef6d1175b9ffd0e/stdlib/test/collections/test_string.mojo#L685) don't cover ?

I thought this check would be broken in `nightly` but I am glad that it's not:

```diff
❯ git diff
diff --git a/stdlib/test/collections/test_string.mojo b/stdlib/test/collections/test_string.mojo
index a664d321..1b53c039 100644
--- a/stdlib/test/collections/test_string.mojo
+++ b/stdlib/test/collections/test_string.mojo
@@ -216,6 +216,7 @@ def test_string_join():
     assert_equal(s5, "1")
 
 
+
 def test_string_literal_join():
     var s2 = ",".join(List[UInt8](1, 2, 3))
     assert_equal(s2, "1,2,3")
@@ -824,6 +825,11 @@ def test_split():
     assert_equal(res6[2], "долор")
     assert_equal(res6[3], "сит")
     assert_equal(res6[4], "амет")
+    var res7 = in6.split("м")
+    assert_equal(res7[0], "Лоре")
+    assert_equal(res7[1], " ипсу")
+    assert_equal(res7[2], " долор сит а")
+    assert_equal(res7[3], "ет")
```

BTW, I still think it's a good test to add.

.gitignore
martinvuyk: done. see #3531

stdlib/src/collections/string.mojo
martinvuyk: yes it has a constructor, will do. It looks like it will simplify things :D

martinvuyk: Only problem that I now realize is that using span for indexing will hurt performance because of index normalization. I'll revert to using pointers for indexing since this function is pretty performance sensitive. There is also the fact that doing `String(StringSlice(Span(ptr, len)))` seems a bit redundant when there is `String(StringSlice(ptr, len))`.

martinvuyk:
> I thought this check would be broken in `nightly` but I am glad that it's not:
> 
> ```diff
> ❯ git diff
> diff --git a/stdlib/test/collections/test_string.mojo b/stdlib/test/collections/test_string.mojo
> index a664d321..b7a85c6c 100644
> --- a/stdlib/test/collections/test_string.mojo
> +++ b/stdlib/test/collections/test_string.mojo
> @@ -824,6 +824,11 @@ def test_split():
>      assert_equal(res6[2], "долор")
>      assert_equal(res6[3], "сит")
>      assert_equal(res6[4], "амет")
> +    var res7 = in6.split("м")
> +    assert_equal(res7[0], "Лоре")
> +    assert_equal(res7[1], " ипсу")
> +    assert_equal(res7[2], " долор сит а")
> +    assert_equal(res7[3], "ет")
> ```
> 
> BTW, I still think it's a good test to add.

I'm not understanding, so the lines from `var res7 ...` onwards didn't get merged in another PR and you'd like me to add them here?

stdlib/src/utils/string_slice.mojo
JoeLoser: I see the conversation here resolved, but didn't see a `debug_assert` added.  Is that intentional?

martinvuyk: Yeah I'm just waiting for @msaelices response to add everything in 1 commit

msaelices:
> > I thought this check would be broken in `nightly` but I am glad that it's not:
> > ```diff
> > ❯ git diff
> > diff --git a/stdlib/test/collections/test_string.mojo b/stdlib/test/collections/test_string.mojo
> > index a664d321..b7a85c6c 100644
> > --- a/stdlib/test/collections/test_string.mojo
> > +++ b/stdlib/test/collections/test_string.mojo
> > @@ -824,6 +824,11 @@ def test_split():
> >      assert_equal(res6[2], "долор")
> >      assert_equal(res6[3], "сит")
> >      assert_equal(res6[4], "амет")
> > +    var res7 = in6.split("м")
> > +    assert_equal(res7[0], "Лоре")
> > +    assert_equal(res7[1], " ипсу")
> > +    assert_equal(res7[2], " долор сит а")
> > +    assert_equal(res7[3], "ет")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > BTW, I still think it's a good test to add.
> 
> I'm not understanding, so the lines from `var res7 ...` onwards didn't get merged in another PR and you'd like me to add them here?

It's just a diff if you want to complete it with more test. LGTM anyways so don't worry. Thanks for that contribution 🥇

stdlib/src/builtin/string_literal.mojo
```diff
@@ -389,19 +390,35 @@ struct StringLiteral(
...
 
-        # Does NOT include the NUL terminator.
         return Span[UInt8, __origin_of(self)](
-            unsafe_ptr=self.unsafe_ptr(),
-            len=self.byte_length(),
+            unsafe_ptr=self.unsafe_ptr(), len=self.byte_length()
+        )
+
+    @always_inline
+    fn as_bytes_read[O: ImmutableOrigin](ref [O]self) -> Span[UInt8, O]:
```
soraros, line 407: I think these can be written as `return self.as_bytes().get_immutable()`.

martinvuyk: I applied your suggestion to `String`, I didn't find any other way to get the trait to work for `StringLiteral` and `StringSlice` since they both give a "non self related origin" with their "as_bytes()". I'm open to suggestions though since I was just tinkering trying to get everything to behave nicely. 

I've also now removed the "as_bytes(ref [_]self)" from `StringLiteral` because when I removed the method that returns `StaticConstantOrigin` the ref method gives `error: cannot form reference to borrowed register value`

soraros: Thanks for investigating! I don't know what's best/correct way to use them either. We can get the part where you are sure working well in first, and iterate. WDYT?

soraros:
```mojo
trait AsBytesWrite:
  fn as_bytes_write[O: MutableOrigin](ref [O]self) -> Span[Byte, O]: ...

trait AsBytesRead:
  fn as_bytes_read[O: ImmutableOrigin](ref [O]self) -> Span[Byte, O]: ...

trait Stringlike(AsBytesRead):
  # don't need `byte_length` and `unsafe_ptr` since we can get them from `as_bytes_[r|w]()`
  fn find[T: Stringlike, //](self, substr: T, start: Int = 0) -> Int: ...
  fn __iter__(ref [_]self) -> _StringSliceIter[__origin_of(self)]: ...
```

martinvuyk:
> ```mojo
> trait Stringlike(AsBytesRead):
>  # don't need `byte_length` and `unsafe_ptr` since we can get them from `as_bytes_[r|w]()`
>  fn find[T: Stringlike, //](self, substr: T, start: Int = 0) -> Int: ...
>  fn __iter__(ref [_]self) -> _StringSliceIter[__origin_of(self)]: ...
> ```

This is a great idea, we'll eventually be able to remove the `unsafe_ptr()` method if we want to with this and leave that to `Span` :muscle: . I also removed `byte_length()` from the trait as you suggested but I don't know if that one or `unsafe_ptr()` will be removed any time soon since they are heavily used. But this is a great step toward unifying stuff :)

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
FYI this is hitting a parser crash on some internal code, so it'll take some time to dig into it and root cause and explore workarounds.  I won't have time for that today, but just letting you know.

martinvuyk:
@JoeLoser Added some better tests and found the crashes. I fixed it with some very shady use of rebind, not sure if making `StringSlice.split()` return a `List[StringSlice[ImmutableAnyOrigin]]` is safe at all memory-wise. We can split it off for another pr and let `StringSlice.split()` still return a `List[String]` for now WDYT?

JoeLoser:
!sync

martinvuyk:
@JoeLoser fixed all conflicts.

Also took out a part of the split implementation to a small public function (in `string_slice.mojo`) to edit in the future that handles all string list building, which once we have more fancy memory allocation mechanisms will maybe make a perf difference (right now it just overallocates to powers of 2):
```mojo
fn to_string[
    O: ImmutableOrigin, //
](items: List[StringSlice[O]]) -> List[String]:
    """Create a list of Strings copying the existing data. This overallocates
    in most cases in exchange for better speed overall.
    ...
    """
    ...
```
(There is an overload for `List[Span[Byte]]` as well)

we are now ~1.2x slower than python (still need some better benchmarking to be sure)

martinvuyk:
Update after a very deep dive into the Python C macros (again) and going over piece by piece of the code I wrote, I fixed very tiny overheads that added up to the point of now being around as fast as Python :tada: (still need better benchmarks). The function `to_string_list()` now simply allocates the given size of each string object since it seems that trying to do page alignment was counterproductive.

stdlib/src/utils/string_slice.mojo
```diff
@@ -978,14 +1155,256 @@ trait Stringlike:
...
+    T1: Stringlike,
+    O: ImmutableOrigin, //,
+    has_maxsplit: Bool,
+](ref [O]src_str: T0, sep: T1, maxsplit: Int) -> List[Span[Byte, O]] as output:
+    sep_len = len(sep.as_bytes_read())
+    if sep_len == 0:
+        iterator = src_str.__iter__()
+        i_len = len(iterator) + 2
+        out_ptr = UnsafePointer[Span[Byte, O]].alloc(i_len)
+        out_ptr.init_pointee_move(rebind[Span[Byte, O]]("".as_bytes_read()))
```
soraros, line 1320: I think it should be roughly it should take the value `src_str.as_bytes()[:0]` so the `ptr` still points to the original string.

martinvuyk: great idea, thanks

stdlib/src/utils/string_slice.mojo
```diff
@@ -1317,15 +1317,13 @@ fn _split_impl[
...
+        out_ptr.init_pointee_move(src_str.as_bytes_read()[:0])
         i = 1
         for s in iterator:
             sp = rebind[Span[Byte, O]](s.as_bytes_read())
             (out_ptr + i).init_pointee_move(sp^)
             i += 1
-        (out_ptr + i).init_pointee_move(
-            rebind[Span[Byte, O]]("".as_bytes_read())
-        )
+        (out_ptr + i).init_pointee_move(src_str.as_bytes_read()[:0])
```
soraros, line 1326: This should technically point to the end of the span (something like `[len(..) - 1:]`)?

And since `Span` is trivial, you can simply write `out_ptr[i] = rebind[Span[Byte, O]](s.as_bytes_read())`.

martinvuyk: thanks for the sugestions, went for this:
```mojo
out_ptr[0] = src_str.as_bytes_read()[0:0]
i = 1
for s in iterator:
    out_ptr[i] = rebind[Span[Byte, O]](s.as_bytes_read())
    i += 1
out_ptr[i] = src_str.as_bytes_read()[-1:-1]
```

martinvuyk:
@JoeLoser fixed all merge conflicts and added benchmark results to PR description. The large diff is because of the magic.lock file

stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -261,30 +261,30 @@ def main():
             alias old = old_chars.get[j, StringLiteral]()
             alias new = new_chars.get[j, StringLiteral]()
             suffix = "[" + str(length) + "]"  # "(" + fname + ")"
-            m.bench_function[bench_string_count[length, fname, old]](
-                BenchId("bench_string_count" + suffix)
-            )
+            # m.bench_function[bench_string_count[length, fname, old]](
```
JoeLoser, line 264: **Question** Did you mean to leave these commented out?

martinvuyk: oops :laughing:

martinvuyk:
@JoeLoser I managed to unify what was `as_bytes_read()` and `as_bytes_write()` into a parametrized `as_bytes()` that also works for `StringSlice`, I updated the PR description. This will allow some very cool generic algorithms on any type that implements the trait :)

stdlib/src/utils/span.mojo
```diff
@@ -343,7 +347,7 @@ struct Span[
         for element in self:
             element[] = value
 
-    fn get_immutable(self) -> Span[T, _lit_mut_cast[origin, False].result]:
+    fn read(self) -> Span[T, _lit_mut_cast[origin, False].result]:
```
JoeLoser, line 350: **Question** Can you help me understand how the name `read()` is better/more favorable than the previous `get_immutable`?

stdlib/src/utils/string_slice.mojo
```diff
@@ -298,7 +299,10 @@ struct StringSlice[is_mutable: Bool, //, origin: Origin[is_mutable].type,](
         Safety:
             `unsafe_from_utf8` MUST be valid UTF-8 encoded data.
         """
-
+        # FIXME: this is giving a lot of problems at comp time
```
JoeLoser, line 302: **Question** Can you file a bug with some more details, please and link to it from here?

stdlib/src/utils/string_slice.mojo
```diff
@@ -640,16 +658,28 @@ struct StringSlice[is_mutable: Bool, //, origin: Origin[is_mutable].type,](
...
 
         Returns:
-            A slice containing the underlying sequence of encoded bytes.
+            A contiguous slice pointing to bytes.
+
+        Notes:
+            This does not include the trailing null terminator.
         """
-        return self._slice
+        return rebind[Span[Byte, origin]](self._slice)
```
JoeLoser, line 679: **Question** Why is the rebind necessary instead over just the span constructor?

stdlib/src/utils/string_slice.mojo
```diff
@@ -371,11 +375,7 @@ struct StringSlice[is_mutable: Bool, //, origin: Origin[is_mutable].type,](
         Args:
             value: The string value.
         """
-
-        debug_assert(
-            _is_valid_utf8(value.as_bytes()), "value is not valid utf8"
-        )
-        self = StringSlice[O](unsafe_from_utf8=value.as_bytes())
+        self = StringSlice[O](unsafe_from_utf8=value.as_bytes[False, O]())
```
JoeLoser, line 378: **Suggestion** Here and below it may be nicer to say `is_mutable=False` to help clarify what's going on here.  Thoughts?

stdlib/src/utils/string_slice.mojo
```diff
@@ -1108,6 +1291,150 @@ fn _to_string_list[
...
+    @parameter
+    if has_sep:
+        return _to_string_list(
+            _split_impl[has_maxsplit](src_str, sep.value(), maxsplit)
+        )
+    else:
+        return _to_string_list(_split_impl[has_maxsplit](src_str, maxsplit))
+
+
+fn _split_sl[
```
JoeLoser, line 1310: **Suggestion** Spell out `sl` here (presumably abbreviated as `StringSlice`) for clarity for future readers.

stdlib/src/utils/string_slice.mojo
```diff
@@ -1108,6 +1291,150 @@ fn _to_string_list[
...
+    has_sep: Bool,
+    T: Stringlike = StringLiteral,
+](src_str: StringSlice[O], sep: Optional[T], maxsplit: Int) -> List[
+    StringSlice[O]
+] as output:
+    var items: List[Span[Byte, O]]
+
+    @parameter
+    if has_sep:
+        items = rebind[__type_of(items)](
```
JoeLoser, line 1322: **Question** Why is the `rebind` needed?

stdlib/src/utils/string_slice.mojo
```diff
@@ -1108,6 +1291,150 @@ fn _to_string_list[
...
+            _split_impl[has_maxsplit](src_str, sep.value(), maxsplit)
+        )
+    else:
+        items = rebind[__type_of(items)](
+            _split_impl[has_maxsplit](src_str, maxsplit)
+        )
+
+    i_len = len(items)
```
JoeLoser, line 1330:
```suggestion
    var i_len = len(items)
```

And below/elsewhere.

stdlib/src/utils/string_slice.mojo
```diff
@@ -1108,6 +1291,150 @@ fn _to_string_list[
...
+    @parameter
+    if has_maxsplit:
+        amnt = maxsplit + 1 if maxsplit < prealloc else prealloc
+    output = __type_of(output)(capacity=amnt)
+    str_byte_len = len(src_str.as_bytes[False, O]())
+    lhs = 0
+    rhs = 0
+    items = 0
+    ptr = src_str.as_bytes[False, O]().unsafe_ptr()
+    # str_span = src_str.as_bytes[False, O]() # FIXME(#3295)
```
JoeLoser, line 1375: **Question** Is that GH issue link right?  #3295 refers to https://github.com/modularml/mojo/issues/3296 which is closed.  Ditto a few lines below in the next FIXME.

stdlib/test/utils/test_string_slice.mojo
```diff
@@ -487,6 +487,103 @@ def test_splitlines():
         _assert_equal(s.splitlines(keepends=True), items)
 
 
+def test_split():
+    alias S = StringSlice[StaticConstantOrigin]
+
+    fn st(value: StringLiteral) -> StringSlice[ImmutableAnyOrigin]:
+        return rebind[StringSlice[ImmutableAnyOrigin]](StringSlice(value))
+
+    # FIXME: remove once StringSlice conforms to TestableCollectionElement
```
JoeLoser, line 496: **Question** Is there anything blocking us from doing this for `StringSlice` today?

JoeLoser (commented):
This is a huge amount of work - nice job @martinvuyk!  The benchmark improvements look great.

I left a few comments/questions (mostly non-blocking).  Would love to be able to land this big PR soon and we can also address some of the things as follow-ups to avoid PR churn/merge conflicts as I know you are working on other things in this area.

stdlib/src/utils/span.mojo
martinvuyk: Just a matter of personal preference, I think it's shorter and self-explanatory. And it was really only used in 1 place so i figured nobody is using the API too much :man_shrugging: . Can revert it if you wish

stdlib/src/utils/string_slice.mojo
martinvuyk: I'll add a mention for issue #3706

stdlib/src/utils/string_slice.mojo
martinvuyk: I don't see why we'd need to build a new instance here. AFAIK rebind is just telling the compiler "trust me it's this type". I'm mostly just trying stuff out, WDYT?

stdlib/src/utils/string_slice.mojo
martinvuyk: I think it is a bit self-evident with `O: ImmutableOrigin`

stdlib/src/utils/string_slice.mojo
JoeLoser: Ah, I see.  The origins may be different, which is why the rebind is needed.  Makes sense 👍 

I agree now we don't need an explicit constructor call.

stdlib/src/utils/string_slice.mojo
martinvuyk: No idea :man_shrugging: but it won't assign otherwise

stdlib/src/utils/string_slice.mojo
martinvuyk: Sorry, meant to link to issue #3526 and mean solve with #3548

stdlib/test/utils/test_string_slice.mojo
martinvuyk: Didn't want to extend this PR too much. But there is probably nothing blocking. I can open another PR if I have time this weekend.

stdlib/src/builtin/string_literal.mojo
```diff
@@ -356,23 +361,43 @@ struct StringLiteral(
...
+        Parameters:
+            is_mutable: Whether the result will be mutable.
+            origin: The origin of the data.
 
         Returns:
-            A reversed iterator over the string.
+            A reversed iterator of references to the string unicode characters.
         """
-        return _StringSliceIter[StaticConstantOrigin, False](
+        constrained[not is_mutable, "StringLiteral can't be mutated"]()
```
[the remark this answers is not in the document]
martinvuyk: I don't see how we'd do that given that the origin parameter in `UnsafePointer` is declared `origin: Origin[True].type = MutableAnyOrigin`. I guess We'd need to add `is_mutable = True, origin: Origin[is_mutable].type = _lit_mut_cast[MutableAnyOrigin, is_mutable].result`. That way it would either be `MutableAnyOrigin` or `ImmutableAnyOrigin` and once you bitcast the pointer to `StaticConstantOrigin` it wouldn't allow mutation (?).

I'd like to look into it in another PR but how would I test that it's carrying the correct origin?

JoeLoser:
FYI this is crashing several things internally in the parser with mutability, so we'll have to dig into that next week sometime. 😞

martinvuyk:
Ok :disappointed:

martinvuyk:
@JoeLoser I removed all uses of generic origin functions and went straight for unsafe pointer and byte length for the internal split implementation. Would you mind testing this so that we can merge it before the next stable release? (if you haven't already branched from nightly at this point)

JoeLoser:
> @JoeLoser I removed all uses of generic origin functions and went straight for unsafe pointer and byte length for the internal split implementation. Would you mind testing this so that we can merge it before the next stable release? (if you haven't already branched from nightly at this point)

Happy to see how this looks internally — do you mind rebasing to fix the conflicts so I can sync it?  Otherwise the sync won't work.

stdlib/src/utils/string_slice.mojo
```diff
@@ -1423,7 +1422,7 @@ fn _split_impl[
     if sep_len == 0:
         var ptr = src_str.unsafe_ptr()
         var iterator = _StringSliceIter[
-            _lit_mut_cast[__origin_of(src_str), False].result
+            _lit_mut_cast[False, __origin_of(src_str)].result
```
soraros, line 1425: You could use `ImmutableOrigin.cast_from[__origin_of(src_str)].result`.

martinvuyk: Turns out I could've just used the `O: ImmutableOrigin` param :v

martinvuyk:
@JoeLoser fixed all conflicts.

@ConnorGray regarding commit 6acac62142cd570cbda3e9ed667acc74ff5f34d7, `Stringlike` is intended for later building more generics on top of it. The uses that are currently in nightly are fine to lower to a concrete type, but once we have parametric traits I'd like to use this trait for encodings (an other things). There is also a solid use case that you can see in this PR which is a generic `.find[T: Stringlike](self, sub: T)` which allows some very ergonomic code on the user side without implicit casting. The other use case is that I'd like external libraries to be able to conform to this trait and let the stdlib integrate nicely with their custom string types. There are many other functions where I'd like to introduce this trait: `.join[T: Stringlike](items: List[T])`, `strip`, `endswith`, `count`, `replace`, and basically every string function which takes in another string. This will also allow us to avoid allocating for non-owning/immutable concrete types interacting with an owning `String`.

martinvuyk:
@JoeLoser can we merge this? 5 months of keeping this on top of the tree and with so many changes to string stuff is laborious.

@ConnorGray I'm un-deprecating `__iter__` here because it is needed. Here you can also see many examples where the performance of building a `Codpoint` (encoding from utf8 to utf32) would be unnecessarily expensive.

My plan is to then on top of this, similar to  #3858, build an iterator taking this code. Then just adding a layer which assembles it into a list.

stdlib/src/collections/string/string.mojo
```diff
@@ -1486,31 +1537,18 @@ struct String(
 
         ```mojo
         # Splitting an empty string or filled with whitespaces
-        _ = String("      ").split() # []
-        _ = String("").split() # []
-
+        _ = "      ".split() # []
+        _ = "".split() # []
```
ConnorGray, line 1541: **Question** Given that these are examples on the `String` type, I wonder if it wouldn't be better to leave them as-is, instead of changing to use string literals?

martinvuyk: IMO it's a bit more visual noise. And I don't know what the plans are for `StringLiteral` materialization into `String`. But IMO it's easier documentation to read . I can change it back if you wish.

ConnorGray: I don't entirely disagree, I do think some of these docs are a bit redundant given that `String` / `StringLiteral` / `StringSlice` have essentially the same behavior. And long-term I think you're right re. materialization, and perhaps eventually delegation in Mojo.

But given that these are docs on `String`, I think we should show `String` explicitly. At the very list, it shrinks the size of the diff of this PR, which helps make review a little easier 🙂

stdlib/test/collections/string/test_string_slice.mojo
```diff
@@ -592,15 +592,7 @@ def test_count_utf8_continuation_bytes():
...
-    # empty separators default to whitespace
-    var d0 = StringSlice("hello world").split()
-    assert_true(len(d0) == 2)
-    assert_true(d0[0] == "hello")
-    assert_true(d0[1] == "world")
-    var d = StringSlice("hello \t\n\n\v\fworld").split(sep="\n")
-    assert_true(len(d) == 3)
-    assert_true(d[0] == "hello \t" and d[1] == "" and d[2] == "\v\fworld")
-
+    alias L = List[StringSlice[StaticConstantOrigin]]
```
ConnorGray, line 595: This change to just use `List` literals here to simplify these tests is great! 🙂

**Suggestion** To try to help shrink this PR and make review easier, I wonder if we could front-load these testing cleanups in a separate PR? That PR could use:


```mojo
alias L = List[String]
```

and then in this PR the diff here would just be changing `String` to `StaticString`.

stdlib/src/collections/string/string.mojo
martinvuyk: I also made them this way because I didn't want to have to change the documentation between all the 3 types. If I change the examples for `String`, for consistency I'd have to also change the ones for `StringSlice` which is very verbose IMO.

mojo/stdlib/test/pathlib/test_pathlib.mojo
```diff
@@ -144,15 +144,6 @@ def test_home():
...
     # Match Python behavior allowing `home()` to overwrite existing path.
     assert_equal(user_path, Path("test").home())
-    # Tests with empty "HOME" and "USERPROFILE"
-    set_home(Path(""))
-    if os_is_windows():
-        # Don't expand on windows if home isn't set
-        assert_equal(Path("~"), Path.home())
-    else:
-        # Test fallback to `/etc/passwd` works on linux
-        assert_true(len(Path.home().path) > 1)
```
martinvuyk: This is a flaky test that I don't know how it didn't get caught before. In Ubuntu 22.04 this works fine, in 24.04 it doesn't. MacOS defaults to `User/runner` instead of the empty string. We shouldn't be concerned with and chase after default OS behavior in this test.

JoeLoser: Can you please split this test fix out to a separate PR and we can land it separately?

martinvuyk:
@JoeLoser @ConnorGray this PR is now independent of the others, it just changes stringslice to use the faster implementation without adding all overloads etc.

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
@martinvuyk can you please rebase this PR to resolve conflicts so I can land this to unblock https://github.com/modular/modular/pull/3894?

martinvuyk:
@JoeLoser done. FYI they aren't dependent on one another, I just wanted you to prioritize this one that I've been working on for far longer and want to see it released

JoeLoser:
!sync

PR #3529, opened 2024-09-23
[stdlib] Add `_count_utf8_continuation_bytes()`

martinvuyk:
Add `_count_utf8_continuation_bytes()`


soraros:
Can we have it take a byte `Span`?

stdlib/src/utils/string_slice.mojo
```diff
@@ -31,6 +31,33 @@ alias StaticString = StringSlice[StaticConstantLifetime]
...
+            for _ in range(rest // s):
+                var vec = (ptr + processed).load[DType.uint8, width=s]()
+                var comp = (vec & 0b1100_0000) == 0b1000_0000
+                amnt += int(comp.cast[DType.uint8]().reduce_add())
+                processed += s
+
+    for i in range(num_bytes - processed):
+        amnt += int((ptr[processed + i] & 0b1100_0000) == 0b1000_0000)
+
+    _ = span
```
soraros, line 57: I think this is needed.

stdlib/src/utils/string_slice.mojo
```diff
@@ -144,10 +171,11 @@ struct _StringSliceIter[
         self.index = 0 if forward else length
         self.ptr = unsafe_pointer
         self.length = length
-        self.continuation_bytes = 0
-        for i in range(length):
-            if _utf8_byte_type(unsafe_pointer[i]) == 1:
-                self.continuation_bytes += 1
+        self.continuation_bytes = _count_utf8_continuation_bytes(
```
soraros, line 174: I think you can write `_count_utf8_continuation_bytes(self._slice)`?

martinvuyk: I could change the input from ptr and len to span but it just gives the work to all places where a `_StringSliceIter` is instantiated, and I tried and it ends up with an ugly lifetime mess for something that is internal and should be simple. And this PR would become bigger.

stdlib/src/utils/string_slice.mojo
martinvuyk: yes it is ?

soraros: Sorry, typo.

stdlib/src/utils/string_slice.mojo
soraros: I'm fine as is! Plus we can always land and iterate.

stdlib/src/utils/string_slice.mojo
martinvuyk: do you mean the for loop for the remaining bytes or the span explicit discard?

- for the for loop my thoughts were: 
The other option is to build a `SIMD[DType.uint8, 8]` and progressively load the pointer, and then do the `&`, `!=`, `reduce_and()`. This could be wasted operations vs just processing the remaining slots. I've seen in some algorithms for things like these (using SIMD) that the performance for the last < 8 bytes are sometimes just better by doing index access than vector ops. Might be wrong here since I haven't really measured the perf. diff. (since here we do 2 operations not just 1 so there might be wasted cycles or the pipeline scheduler could do it's job well :man_shrugging: )

- for the explicit discard of span:
the lifetime of the span outside the function is unknown and the pointer is used inside this func, so I'd rather ensure the owner of the data doesn't dealloc the pointer before the func has counted everything. Again, might be wrong since I don't understand lifetimes all that well.

WDYT?

soraros: I meant the `_ = span`. It will be kept alive for the duration of the function.

martinvuyk: But isn't that because it has `ImmutableLifetime`? I also see that in all cases some indexing of the span is still done, which besides hurting performance because of index normalization it also means that the lifetime is ensured (?). 

PS: I'll delete the discard, but I still want to understand

soraros: A borrowed argument (unlike a `var`) is kept alive for the function call. See also `Formatter.write_str` and `_quicksort_partition_right`.

stdlib/src/builtin/string_literal.mojo
```diff
@@ -287,6 +287,18 @@ struct StringLiteral(
             unsafe_pointer=self.unsafe_ptr(), length=self.byte_length()
         )
 
+    fn __reversed__(
+        ref [_]self,
```
soraros, line 291: I think `self` can be a borrowed argument.

martinvuyk: Both `String` and `StringLiteral` `__iter__` funcs have `ref [_]self` as signature (String has it's `__reversed__` func the same)

martinvuyk: There, removed them and nothing exploded :D

soraros: I think `StringLiteral.__iter__` can use `borrowed self` since SL doesn't support mutable iterations.

stdlib/test/collections/test_string.mojo
```diff
@@ -1335,17 +1335,20 @@ def test_string_iter():
...
             var byte_len = v.byte_length()
-            assert_equal(item[byte_idx : byte_idx + byte_len], v)
+            for i in range(byte_len):
+                assert_equal(ptr[byte_idx + i], v.unsafe_ptr()[i])
             byte_idx += byte_len
-            utf8_sequence_len += 1
-        assert_equal(utf8_sequence_len, utf8_sequence_lengths[item_idx])
+            amnt_characters += 1
+
+        assert_equal(amnt_characters, items_amount_characters[item_idx])
```
martinvuyk, line 1351: This is unrelated to the main topic of this PR, I just realized this is another place where indexing was assuming byte offset so I fixed it and also renamed some variables with unclear names.

PR #3531, opened 2024-09-23
[stdlib] Add .magic/ and magic.lock to .gitignore

martinvuyk:
Add .magic/ and magic.lock to .gitignore


martinvuyk:
My `magic.lock` file is currently 7631 lines long, every time someone installs in any different platform this would get bigger and bigger...

the `pixi.toml`/`magic.toml` file can have fixed versions and ranges, but gives some leeway when trying to build a big package if you tried to fix versions, channels, build_id, licence, and the whole dependency tree it would be practically impossible sometimes or the file would take gigabytes for projects that depend on a lot.

JoeLoser:
> I'm under the impression that lock files are meant to be tracked. Maybe instead of using `COMPATIBLE_COMPILER_VERSION`, the team should just publish `magic.lock`?

@zbowling I think this is something you've briefly mentioned in the past.  Can you chime in here / what are your thoughts?

jackos (approved):

jackos:
Thanks @martinvuyk just merging this directly as the file isn't upstream

jackos:
@soraros we don't have this in place at the moment, but @zbowling has been looking into a way to update `pixi.toml` with the new version of mojo/max on each release. But until we have that we can't track lockfiles, otherwise it will lock to the old version of mojo/max

PR #3539, opened 2024-09-23
[stdlib] Fix `String.format()` to use byte indexing, refactor for performance and readability, and prepare for full format spec support

### LINKED ISSUE #3296: [BUG] [stdlib] Fix `String.format()` to use explicit byte indexing and not assume `String.__getitem__` indexes directly to byte offset.
### Bug description

Fix `String.format()` to use explicit byte indexing and not assume `String.__getitem__` indexes directly to byte offset.

Python's String indexes by unicode codepoints (UTF-32) our String implementation currently uses UTF-8 and it's `__getitem__` returns by byte offset. We need to be compliant with Python's indexing using unicode codepoints and `String.format()` needs to be fixed.

This is related to PR #3295 and issue #3246

@rd4com I saw it's code you commited, do you wanna take this or should I ?

### Steps to reproduce

### System information

```shell
```

martinvuyk:
Fix `String.format()` to use byte indexing, refactor for performance and readability, and prepare for full format spec support. Closes #3296


stdlib/src/collections/string.mojo
```diff
@@ -2207,6 +2202,25 @@ struct String(
         var result = String(buffer)
         return result^
 
+    @always_inline("nodebug")
+    fn _slice(
```
martinvuyk, line 2206: We could expose this in `StringSlice` as `fn unsafe_slice()` and wrap it here

martinvuyk:
Hi @mzaks (while debugging for this PR I came accross `/test/hashlib/test_ahash.mojo`) I changed the line to alloc 8 `UInt8` [here](https://github.com/modularml/mojo/blob/18830ba255ccd4ff4a127d636ac0c058460458f1/stdlib/test/hashlib/test_ahash.mojo#L675) since it was using an `UnsafePointer` without freeing and the amount of bits is small so stack alloc made sense. The test `test_trailing_zeros` is failing for [with macos-14 and assertions=0](https://github.com/modularml/mojo/actions/runs/11223515251/job/31198320030#logs) and not for `assertions=1` :thinking: , diff between hashes is 18 instead of `diff > 19` like it says in the assert, should I lower the bar? 

It is still very weird that how the memory is allocated affects the results of this :/ or is it because the algorithm uses the address of the pointer and the stack allocations are closer than heap?

mzaks:
> Hi @mzaks (while debugging for this PR I came accross `/test/hashlib/test_ahash.mojo`) I changed the line to alloc 8 `UInt8` [here](https://github.com/modularml/mojo/blob/18830ba255ccd4ff4a127d636ac0c058460458f1/stdlib/test/hashlib/test_ahash.mojo#L675) since it was using an `UnsafePointer` without freeing and the amount of bits is small so stack alloc made sense. The test `test_trailing_zeros` is failing for [with macos-14 and assertions=0](https://github.com/modularml/mojo/actions/runs/11223515251/job/31198320030#logs) and not for `assertions=1` :thinking: , diff between hashes is 18 instead of `diff > 19` like it says in the assert, should I lower the bar? 
> 
> 
> 
> It is still very weird that how the memory is allocated affects the results of this :/ or is it because the algorithm uses the address of the pointer and the stack allocations are closer than heap?

Ah, there is a bug in the test. We need to add memsetzero call. Otherwise it is UB what is in this memory region. I will definitely fix it in the next PR, but you can also fix it in yours.

martinvuyk:
> Ah, there is a bug in the test. We need to add memsetzero call. Otherwise it is UB what is in this memory region. I will definitely fix it in the next PR, but you can also fix it in yours.

Ooh, ok. I'll add it here for now but I'm not sure when the team will merge this. I also did stack alloc for a ptr in `test_avalanche()` and added a space to the `split(", ")` in case you want to incorporate those changes :)

JoeLoser:
!sync

martinvuyk:
@JoeLoser made the `Stringlike` trait public to avoid conflicts with PR #3528 , are you ok with it being public for this PR ? I "introduce it officially" in that other PR with many methods, here it just has `unsafe_ptr()` and `byte_length()` methods

JoeLoser:
!sync

JoeLoser:
> @JoeLoser made the `Stringlike` trait public to avoid conflicts with PR #3528 , are you ok with it being public for this PR ? I "introduce it officially" in that other PR with many methods, here it just has `unsafe_ptr()` and `byte_length()` methods

Works for me.  Like everything in the stdlib, we'll keep iterating as we go ❤️ 

Thanks for your great contribution here in continuing on the format spec work!  This is a huge basis for what will eventually back the nice format strings everyone wants 😄

PR #3545, opened 2024-09-24
[CI] Pin `magic` to 0.2.3 to avoid timeouts

JoeLoser:
There's issues with HTTP timeouts/client certificate errors in the latest version of `magic` (0.3.0 at the time of writing) due to the uv/python package solver.  Workaround these to improve the stability of our CI by just downgrading `magic` to a known, previous version, `0.2.3`.


JoeLoser:
Discussed with Frank internally and this is passing — we also made a similar change the other day for MAX nightlies internally.  Going to merge this.

PR #3546, opened 2024-09-24
[stdlib] Optimize `_StringSliceIter` to not have branching in forward iteration

martinvuyk:
Optimize `_StringSliceIter` to not have branching in forward iteration.

This also fixes an error in the iterator's `__len__` logic which assumed it would be called only at the beginning of the iteration


JoeLoser (commented):
For the part about

> This also fixes an error in the iterator's __len__ logic which assumed it would be called only at the beginning of the iteration

Can we add a unit test case that gets fixed as a result of your changes?

JoeLoser (approved):

JoeLoser:
!sync

PR #3548, opened 2024-09-25
[stdlib] Add `Span.find()` and `rfind()`

martinvuyk:
Add `Span.find()` and `rfind()`. This will be useful to look for raw data, not only string processing.

### DIFF
--- a/mojo/stdlib/std/memory/span.mojo
+++ b/mojo/stdlib/std/memory/span.mojo
@@ -34,6 +34,12 @@ from std.hashlib import Hasher
 from std.builtin.device_passable import DevicePassable
 from std.compile import get_type_name
 import std.format._utils as fmt
+from std.collections.string.string_slice import (
+    _memmem,
+    _memchr,
+    _memrmem,
+    _memrchr,
+)
 
 
 # ===-----------------------------------------------------------------------===#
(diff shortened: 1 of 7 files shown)


stdlib/src/utils/span.mojo
```diff
@@ -335,3 +338,194 @@ struct Span[
         return Span[T, _lit_mut_cast[lifetime, False].result](
             unsafe_ptr=self._data, len=self._len
         )
+
+    fn find[
```
JoeLoser, line 342: **Question** It seems a bit weird from an API design perspective to have this "find" function in `span`, can you help me understand why we'd want this?  It feels more coupled to string algorithms, and `Span` isn't just for a view over string data.

martinvuyk: It can be used if you want to find occurrences of a specific scalar value sequence. My main thought was that we'll lose the ability to do that once `StringSlice.find()` is fixed to use unicode codepoints and it introduces quite some overhead for algorithms that are faster using raw bytes. We could also make this private WDYT?

martinvuyk: @JoeLoser gentle ping and also to add something, since it seems like we are moving toward unifying Python's `bytes` into `Span[UInt8]`in PR #3636, request #3634 and many other places. Python's [bytes.find](https://docs.python.org/3/library/stdtypes.html#bytes.find) is actually a method, and many other things that we put in `StringSlice` and `StringRef` are actually [operations from `bytes`](https://docs.python.org/3/library/stdtypes.html#bytes-and-bytearray-operations).

msaelices: I think this is a great change for overall performance when manipulating slices. It opens a lot of possibilities.

martinvuyk:
@JoeLoser CI is failing because of unrelated reasons and ignore the diff numbers, it's because of the magic.lock file. 

I refactored `normalize_index` and used it here. I also refactored `memrmem` and `memrchr` to be vectorized like their counterparts. These are algorithms that apply not only to string data, you can use them to find anything (DB indexes etc.). I had tried to make a similar algorithm for Array in my repo but failed to make it as efficient as these, kudos to whoever implemented them. Can we please have this? They will be used a lot. I've also found that `.replace()` and `.count()` will use this function besides `.split()`, `.splitlines()`, `List.index()`, and others.

martinvuyk:
@JoeLoser would it be possible to land this before the next stable release? I'd like to use this to be able to bypass string's `.find()` for algorithms which work on raw bytes. And using that bypass, I'd like to make the breaking change to switch over to full unicode:
- `String.__len__()` should return unicode length and not byte length
- `String.find()` should work on unicode offsets
- `String.__getitem__()` should work on unicode offsets

I would like to make the change as soon as possible because it will break a lot of projects, and many more are coming because the community is growing.

stdlib/src/utils/span.mojo
owenhilyard: @JoeLoser per discussion on discord, `Span[T].find` is intended to be `memchr` or `memmem` as appropriate, but that the subspan of bytes must be aligned to `alignof[T]()`. We can probably expand it to any `T: Trivial` once we have a `Trivial` trait. 

The proper string version likely needs be specialized for UTF-8 comparisons. @martinvuyk and I have discussed an "AsciiString" which can more aggressively optimize for using byte-oriented intrinsics since UTF-8 is annoying to do with SIMD.

martinvuyk:
@JoeLoser we need this to be able to support unicode indexing while also keeping the ability to look for raw bytes.

gryznar:
Shall we follow Python convention? `rfind` is not obvious at first glance without diving into documentation

martinvuyk:
> Shall we follow Python convention? rfind is not obvious at first glance without diving into documentation

I have no problem going for `find_from_right` or something like that. But let's leave it for another PR

gryznar:
This question is to @JoeLoser if stdlib team is open to cleaning this up in the nearest future, step by step

mojo/stdlib/std/memory/span.mojo
```diff
@@ -626,3 +632,165 @@ struct Span[
...
+            if from_left:
+                return 0
+            else:
+                return _len
+
+        var n_s: Int
+
+        # _memXXX implementations already handle when haystack_len == 0
+        @parameter
+        if unsafe_dont_normalize:
```
JoeLoser: **Question** Can we use `normalize_index` with the appropriate parameter here to simplify this calling code?

mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -626,3 +632,165 @@ struct Span[
...
+            n_s = v * Int(v < _len and v > 0) + _len * Int(v >= _len)
+        var s_ptr = self.unsafe_ptr()
+        var haystack = __type_of(self)(
+            ptr=s_ptr + n_s, length=_len - n_s
+        ).get_immutable()
+        var loc: UnsafePointer[Scalar[D]]
+
+        @parameter
+        if from_left and not single_value:
+            loc = _memmem(
```
JoeLoser: **Comment** Can't wait to land your other PRs that change these APIs to use a safer `Span` soon.

mojo/stdlib/test/collections/string/test_string.mojo
```diff
@@ -631,6 +635,24 @@ def test_find():
...
+    assert_equal(0, (String("0") * 100).find("00"))
+    assert_equal(9, String.DIGITS.find("9"))
+    assert_equal(25, String.ASCII_LETTERS.find("z"))
+    assert_equal(25, String.ASCII_LETTERS.find("zA"))
+    assert_equal(25, String.ASCII_LETTERS.find("zABCD"))
+    for i in range(2, 50):
+        assert_equal(51, (String.ASCII_LETTERS * i).find("Za"))
+
+    # Special characters.
+    # FIXME: once find works by unicode codepoints
```
JoeLoser: **Suggestion** Link to the appropriate GitHub issue, please.

mojo/stdlib/test/collections/string/test_string.mojo
```diff
@@ -674,12 +696,25 @@ def test_rfind():
...
+    assert_equal(9, String.DIGITS.rfind("9"))
+    assert_equal(25, String.ASCII_LETTERS.rfind("z"))
+    assert_equal(25, String.ASCII_LETTERS.rfind("zA"))
+    assert_equal(25, String.ASCII_LETTERS.rfind("zABCD"))
+    for i in range(2, 50):
+        assert_equal(0, ("123" + String.ASCII_LETTERS * i).rfind("123"))
 
     # Special characters.
-    # TODO(#26444): Support unicode strings.
+    # FIXME: once find works by unicode codepoints
```
JoeLoser: **Suggestion** Ditto here for linking to the appropriate GitHub issue.

JoeLoser (approved):
Generally looks good, just needs a rebase.  We're +1 for adding these algorithms to `Span` as discussed in https://github.com/modular/modular/issues/4758 - we discussed this issue as a team the other day.

Thanks, Martin!

mojo/stdlib/std/memory/span.mojo
martinvuyk: The logic is very different, python's find clamps indices to accept any range. I opened  #3783 a very long time ago because there are several functions like this with different clamping or normalization criteria, some also accept or reject containers with 0 length etc.

We could just use `normalize_index` though, WDYT?

gryznar: Should we strictly follow Python here? With extensions, I suppose that implementation in Mojo could be different

JoeLoser:
!sync

JoeLoser:
!sync

martinvuyk:
```mojo
struct Span:
  comptime _UnsafePointerType = NonNullUnsafePointer[T, origin]

def _memmem[
    dtype: DType, //
](
    haystack_span: Span[mut=False, Scalar[dtype], ...],
    needle_span: Span[mut=False, Scalar[dtype], ...],
) -> type_of(haystack_span.unsafe_ptr()): ...
```
This might cause a problem @NathanSWard, memmem returns a null pointer when not found. I'm not sure if you want all `.unsafe_ptr()` to return a non null pointer in the future, but this is a case that needs to be nullable. Also I'm not 100% sure that we won't ever build a `Span` of length 0 with a null pointer

NathanSWard:
> This might cause a problem @NathanSWard, memmem returns a null pointer when not found.

In this case, `memmem` should return an `Optional[NonNullUnsafePointer]`. I can go through and update this internally. 

---

Ah wait, I think this is actually okay as is.  As long as the methods use the haystack length and don't depend on the nullability of the input spans pointer this is fine.

martinvuyk:
> > This might cause a problem @NathanSWard, memmem returns a null pointer when not found.
> 
> In this case, `memmem` should return an `Optional[NonNullUnsafePointer]`. I can go through and update this internally.
> 
> Ah wait, I think this is actually okay as is. As long as the methods use the haystack length and don't depend on the nullability of the input spans pointer this is fine.

I actually do need to null the pointer / ask whether it's found.
I'm conceptually doing `return Int(loc) - Int(s_ptr) if loc else -1` with some bit tricks doing `return ((Int(loc) - Int(s_ptr) + 1) & splat(Bool(loc))) - 1`

This needs to be as fast as possible because find is used in a lot of hot loops, so it's not about making it the safest internally (by using something like optional)

NathanSWard:
> I actually do need to null the pointer / ask whether it's found.

Yes, you can still do that. However, you need to use `Optional[NonNullUnsafePointer]`. the size and layout of `Optional[NNUP]` is equal to that of `NNUP`. If you want to do that address tricks that's fine too since a `None` of `Optional[NNUP]` is 0 (since the None state is represented as NULL).

The thing you _cannot do_ is rely on `Span.unsafe_ptr()` returning `NULL` when `len(span) == 0`.

PR #3160, opened 2024-07-01
[stdlib] Make `Slice` step `Optional`

bgreni:
Follow up from https://github.com/modularml/mojo/pull/3064#issuecomment-2195753384

Make `step` member of `Slice` `Optional` in order to better match the behaviour/semantics of the Python slice.

PR #3549, opened 2024-09-25
[stdlib] Mojo-Python slice interoperability

jjvraw:
Leveraging Mojo's `Slice` for more natural interop between `PythonObject`s, enabling something like below:

```mojo
a = PythonObject([1, 2, 3, 4])
a_reversed = a[::-1]
```

- We leave C Python to handle slice parameters. 
- Negative cases such as `b[1.3:10]` or `b["1":10]` are handled by parser which would normally throw a `TypeError` in Python.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -187,11 +187,18 @@ what we publish.
 - Support for multi-dimensional indexing for `PythonObject`
   ([PR #3583](https://github.com/modularml/mojo/pull/3583) by [@jjvraw](https://github.com/jjvraw)).
 
+- Support for multi-dimensional indexing and slicing for `PythonObject`
+  (PRs  [#3549](https://github.com/modularml/mojo/pull/3549),
+  [#3583](https://github.com/modularml/mojo/pull/3583) by [@jjvraw](https://github.com/jjvraw)).
+
     ```mojo
     var np = Python.import_module("numpy")
-    var a = np.array(PythonObject([1,2,3,1,2,3])).reshape(2,3)
+    var a = np.array(PythonObject([1,2,3,4,5,6])).reshape(2,3)
     print((a[0, 1])) # 2
-    ```
+    print((a[1][::-1])) # [6 5 4]
+   ```
+
+  Note, that the syntax, `a[1, ::-1]`, is currently not supported.
 
 - [`Arc`](/mojo/stdlib/memory/arc/Arc) now implements
   [`Identifiable`](/mojo/stdlib/builtin/identifiable/Identifiable), and can be
--- a/stdlib/src/python/_cpython.mojo
+++ b/stdlib/src/python/_cpython.mojo
@@ -1755,6 +1755,33 @@ struct CPython:
         self._inc_total_rc()
         return r
 
+    fn PySlice_FromSlice(inout self, slice: Slice) -> PyObjectPtr:
+        # Convert Mojo Slice to Python slice parameters
+        # Note: Deliberately avoid using `span.indices()` here and instead pass
+        # the Slice parameters directly to Python. Python's C implementation
+        # already handles such conditions, allowing Python to apply its own slice
+        # handling.
+        var py_start = self.Py_None()
+        var py_stop = self.Py_None()
+        var py_step = self.Py_None()
+
+        if slice.start:
+            py_start = self.PyLong_FromSsize_t(c_ssize_t(slice.start.value()))
+        if slice.end:
+            py_stop = self.PyLong_FromSsize_t(c_ssize_t(slice.end.value()))
+        if slice.end:
+            py_step = self.PyLong_FromSsize_t(c_ssize_t(slice.step.value()))
+
+        var py_slice = self.PySlice_New(py_start, py_stop, py_step)
+
+        if py_start != self.Py_None():
+            self.Py_DecRef(py_start)
+        if py_stop != self.Py_None():
+            self.Py_DecRef(py_stop)
+        self.Py_DecRef(py_step)
+
+        return py_slice
+
     # const char *PyUnicode_AsUTF8AndSize(PyObject *unicode, Py_ssize_t *size)
     fn PyUnicode_AsUTF8AndSize(inout self, py_object: PyObjectPtr) -> StringRef:
         """See https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize.
@@ -1913,3 +1940,31 @@ struct CPython:
             fn (PyObjectPtr) -> c_int
         ]("PySequence_Check")(obj)
         return follows_seq_protocol != 0
+
+    # ===-------------------------------------------------------------------===#
+    # Python Slice Creation
+    # ===-------------------------------------------------------------------===#
+
+    # PyObject *PySlice_New(PyObject *start, PyObject *stop, PyObject *step)
+    # ref: https://docs.python.org/3/c-api/slice.html#c.PySlice_New
+    fn PySlice_New(
+        inout self, start: PyObjectPtr, stop: PyObjectPtr, step: PyObjectPtr
+    ) -> PyObjectPtr:
+        var r = self.lib.get_function[
+            fn (PyObjectPtr, PyObjectPtr, PyObjectPtr) -> PyObjectPtr
+        ]("PySlice_New")(start, stop, step)
+
+        self.log(
+            r._get_ptr_as_int(),
+            " NEWREF PySlice_New, refcnt:",
+            self._Py_REFCNT(r),
+            ", start:",
+            start._get_ptr_as_int(),
+            ", stop:",
+            stop._get_ptr_as_int(),
+            ", step:",
+            step._get_ptr_as_int(),
+        )
+
+        self._inc_total_rc()
+        return r
--- a/stdlib/test/python/test_python_object.mojo
+++ b/stdlib/test/python/test_python_object.mojo
@@ -517,6 +517,94 @@ def test_setitem_raises():
         d[[1, 2, 3]] = 5
 
 
+fn test_py_slice() raises:
+    var a = PythonObject([1, 2, 3, 4, 5])
+    assert_equal("[2, 3]", str(a[1:3]))
+    assert_equal("[1, 2, 3, 4, 5]", str(a[:]))
+    assert_equal("[1, 2, 3]", str(a[:3]))
+    assert_equal("[3, 4, 5]", str(a[2:]))
+    assert_equal("[1, 3, 5]", str(a[::2]))
+    assert_equal("[2, 4]", str(a[1::2]))
+    assert_equal("[4, 5]", str(a[-2:]))
+    assert_equal("[1, 2, 3]", str(a[:-2]))
+    assert_equal("[5, 4, 3, 2, 1]", str(a[::-1]))
+    assert_equal("[1, 2, 3, 4, 5]", str(a[-10:10]))  # out of bounds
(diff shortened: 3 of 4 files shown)


stdlib/test/python/test_python_object.mojo
```diff
@@ -409,6 +409,58 @@ fn test_none() raises:
...
+    assert_equal("[4, 5]", str(a[-2:]))
+    assert_equal("[1, 2, 3]", str(a[:-2]))
+    assert_equal("[5, 4, 3, 2, 1]", str(a[::-1]))
+    assert_equal("[1, 2, 3, 4, 5]", str(a[-10:10]))  # out of bounds
+    assert_equal("[1, 2, 3, 4, 5]", str(a[::]))
+    assert_equal("[1, 2, 3, 4, 5]", str(a[:100]))
+    assert_equal("[]", str(a[5:]))
+    assert_equal("[5, 4, 3, 2]", str(a[:-5:-1]))
+
+    var b = Python.evaluate("[i for i in range(1000)]")
```
martinvuyk, line 429: I had no idea we could already do this :exploding_head:

stdlib/test/python/test_python_object.mojo
```diff
@@ -409,6 +409,58 @@ fn test_none() raises:
...
+    assert_equal("[]", str(empty[1:2:3]))
+
+    var d = Python.dict()
+    d["a"] = 1
+    d["b"] = 2
+    with assert_raises(contains="slice(1, 3, 1)"):
+        _ = d[1:3]
+
+    var custom = Python.evaluate(
+        "type('CustomSliceable', (), {'__getitem__': lambda self, key: key})()"
```
martinvuyk, line 459: I've done and seen some shady metaclass manipulation in Python but I've never seen this manual use of `type` :rofl:

docs/changelog.md
```diff
@@ -120,6 +120,15 @@ what we publish.
   determining a default SDK to use. The user can select the default SDK to use
   with the `Mojo: Select the default MAX SDK` command.
 
+- Mojo-Python slice interoperability, enabling something like below:
+
+  ```mojo
+  a = PythonObject([1, 2, ,3, 4])
+  a_reversed = a[::-1]
+  ```
+
```
martinvuyk, line 129: nit (changelog file is very long already):
```suggestion
  ```mojo
  print(PythonObject([1, 2, ,3, 4])[::-1]) # [4, 3, 2, 1]
  ```
```

martinvuyk (commented):
Hi @jjvraw this is awesome, thank you for tackling it!

stdlib/test/python/test_python_object.mojo
jjvraw: _shady_ indeed: [SO post](https://stackoverflow.com/questions/15247075/how-can-i-dynamically-create-derived-classes-from-a-base-class)

jjvraw:
Going to mark as draft, and wait for `PyObject_GetItem` of #3583 to be merged.

JoeLoser:
> Going to mark as draft, and wait for `PyObject_GetItem` of #3583 to be merged.

https://github.com/modularml/mojo/pull/3583 has been merged, so feel free to rebase whenever it's convenient for you.

jjvraw:
> > Going to mark as draft, and wait for `PyObject_GetItem` of #3583 to be merged.
> 
> #3583 has been merged, so feel free to rebase whenever it's convenient for you.

Made some additional changes. Since the compiler doesn't implicitly handle Mojo-Python conversion for slice, and the syntax `[0:1, 3:]` is probably not recognised, this gives the user some play with creating slice `PythonObjects`s to comma seperate indices. 

Can always revert these, however.

stdlib/test/python/test_python_object.mojo
```diff
@@ -485,6 +485,93 @@ fn test_getitem_raises() raises:
...
+    with_2d = Python.evaluate(
+        """type('With2D', (), {
+            '__init__': lambda self: setattr(self, 'data', [[1, 2, 3], [4, 5, 6], [7, 8, 9]]),
+            '__getitem__': lambda self, key: (
+                [row[key[1]] for row in self.data[key[0]]] if isinstance(key, tuple) and all(isinstance(k, slice) for k in key)
+                else (self.data[key[0]][key[1]] if isinstance(key, tuple)
+                else self.data[key])
+            )
+        })()"""
+    )
```
jjvraw, line 552: Also, would integrating something like [`stdlib/test/python/my_module.py`](https://github.com/modularml/mojo/blob/nightly/stdlib/test/python/my_module.py) be preferred over this..?  @JoeLoser

stdlib/src/python/_cpython.mojo
```diff
@@ -1115,6 +1115,30 @@ struct CPython:
...
+            start._get_ptr_as_int(),
+            ", stop:",
+            stop._get_ptr_as_int(),
+            ", step:",
+            step._get_ptr_as_int(),
+        )
+
+        self._inc_total_rc()
+        return r
+
```
ConnorGray, line 1141: **Suggestion** Could we put this method in a new section with its own top-level comment header? This is currently in a section intended to store `PyObject_*` methods. (Though I understand the impulse to co-locate this near `PyObject_CallObject`.)

stdlib/src/python/_cpython.mojo
```diff
@@ -1393,6 +1420,31 @@ struct CPython:
...
+        var py_slice = self.PySlice_New(py_start, py_stop, py_step)
+
+        if py_start != self.Py_None():
+            self.Py_DecRef(py_start)
+        if py_stop != self.Py_None():
+            self.Py_DecRef(py_stop)
+        self.Py_DecRef(py_step)
+
+        return py_slice
+
```
ConnorGray, line 1447: **Suggestion** This file (`_cpython.mojo`) is mostly intended to be low level wrappers around functions in the CPython CAPI. Since `PySlice_FromSlice` isn't a CPython function, I think the logic from this method might fit better as an initializer on `PythonObject`, e.g.:

```mojo
struct PythonObject(..):
    # ...

    fn __init__(inout self, slice: Slice):
        ...
```

Additionally, instead of calling low-level methods like `PyLong_FromLong()` or `Py_None()` directly, the code might be improved by using the existing higher level safe wrappers for those functions (provided by `PythonObject` as well). So the final code might look something like:

```mojo
    fn __init__(inout self, slice: Slice) -> PyObjectPtr:
        # Convert Mojo Slice to Python slice parameters
        # Note: Deliberately avoid using `span.indices()` here and instead pass
        # the Slice parameters directly to Python. Python's C implementation
        # already handles such conditions, allowing Python to apply its own slice
        # handling.
        var py_start = PythonObject(None)
        var py_stop = PythonObject(None)
        var py_step = PythonObject(slice.step)

        if slice.start:
            py_start = PythonObject(slice.start.value())
        if slice.end:
            py_stop = PythonObject(slice.end.value())

        var py_slice = cpython.PySlice_New(py_start, py_stop, py_step)

        return PythonObject(py_slice)
```

This has the advantage of allowing us to omit the manual ref count management logic as well 🙂 

Let me know what you think, or if there's a subtlety I missed about how you're using that means this might not be feasible for your use-case.

-----

For context, very roughly speaking, there are currently three files in the stdlib pertaining to interop, roughly in order from low-level to high-level:

* `_cpython.mojo` — low-level, "raw" bindings to the CPython API. In some future world, these might even become auto-generated bindings. They should have minimal logic.

* `python_object.mojo` — defines `PythonObject`, which interfaces between Mojo types and valid Python object values.

* `python.mojo` — high-level, safe wrappers around functionality from the CPython API.

ConnorGray:
Hey @jjvraw thanks for tackling this, and all your related work in this area recently 🙂

I did a quick pass and made a couple of suggestions for some possible improvements. Let me know what you think, or if there's anything I might have missed.

Thanks again for the 🔥 contributions, this is great polish work that will make interop that much more pleasant and intuitive to use 🙂

stdlib/src/python/_cpython.mojo
jjvraw: Hi, thank you for the time and feedback! :) 

> Let me know what you think, or if there's a subtlety I missed about how you're using that means this might not be feasible for your use-case.

I originally wanted to avoid `PythonObject` within `__getitem__`, as i noticed marginal gains. This was the intent of `PySlice_FromSlice`, to construct a `PyObjectPtr`. However, I overlooked the fact that the other `From*` functions are actually part of the C API.

I agree with your point about maintaining the minimalism of `_cpython.mojo`. What do you think of refactoring `PySlice_FromSlice` not inline of `__init__` but rather as a helper within `python_object.mojo`/`python.mojo`? This way we can maintain `PyObjectPtr` within `__getitem__`. 

> Additionally, instead of calling low-level methods like PyLong_FromLong() or Py_None() directly, the code might be improved by using the existing higher level safe wrappers for those functions (provided by PythonObject as well). So the final code might look something like:

On this note, I noticed what seems to be lifetime issues. To apply this suggestion, I would have to do something like the following:
```mojo
  ...
        
  cpython = _get_global_python_itf().cpython()
  var py_slice = cpython.PySlice_New(
    py_start.unsafe_as_py_object_ptr(),
    py_stop.unsafe_as_py_object_ptr(),
    py_step.unsafe_as_py_object_ptr()
  )
    
  _ = py_start^    
  _ = py_stop^    
  _ = py_step^    
  
  return py_slice
```

This might be worth exploring, following 400866c0abc2d0deba91645a53a86242e9028709

jjvraw:
> Hey @jjvraw thanks for tackling this, and all your related work in this area recently 🙂
> 
> I did a quick pass and made a couple of suggestions for some possible improvements. Let me know what you think, or if there's anything I might have missed.
> 
> Thanks again for the 🔥 contributions, this is great polish work that will make interop that much more pleasant and intuitive to use 🙂

Hey Connor. No problem :)

Please see my comment, and changes of the last commit 39a9e23ab91bf79ce81dd7e0b9c7d9a521dc6e9b. Note that I don't have a strong opinion on these, and happy to apply changes.

JoeLoser:
!sync

jjvraw:
A few conflicts came up with #3160 and #3639

JoeLoser:
> Not sure if this has been worked on internally or not, but a few conflicts came up with #3160 and #3639. Have pushed the changes required.

It was failing CI, but I'm taking a look today at it.  Thanks for the ping - just re-synced.

stdlib/test/python/test_python_object.mojo
JoeLoser: I think refactoring these sort of things to a separate test file (class with methods like you have here) may improve readability over raw strings.  Feel free to defer to a follow-up though :)

stdlib/test/python/test_python_object.mojo
```diff
@@ -517,6 +517,94 @@ def test_setitem_raises():
...
+    assert_equal("(4, 3, 2)", str(t[3:0:-1]))
+
+    var empty = PythonObject([])
+    assert_equal("[]", str(empty[:]))
+    assert_equal("[]", str(empty[1:2:3]))
+
+    var d = Python.dict()
+    d["a"] = 1
+    d["b"] = 2
+    with assert_raises(contains="slice(1, 3, None)"):
```
JoeLoser, line 563: FYI this fails internally in CI since it says `unhashable type: 'slice'`.  I haven't looked closely as to why it fails internally but passes in OSS CI (perhaps different CPython versions as we're using 3.11 internally it looks like primarily).  I disabled the test so I can merge this, but we should look into it as a follow-up.

PR #3210, opened 2024-07-10
[stdlib] Modernise `Span.__[eq/ne]__` with latest conditional conformance support

soraros:

PR #3553, opened 2024-09-26
[stdlib] Use the new mem type `rebind` in stdlib

soraros:


rd4com:
`Rebind` on `Reference` was really usefull (`trivial` type), 
we used it before to change a type, by changing the type of the `Reference`,
but i think it does copy on memory-only types:

```mojo
@value
struct MyStruct:
    var x:Int
    fn __copyinit__(inout self, other: Self):
        self.x = other.x
        print("copyinit")
    fn __moveinit__(inout self, owned other: Self):
        self.x = other.x
        print("moveinit")
    fn __del__(owned self): print("del")

@value
struct MyStruct2:
    var x:Int
    fn __copyinit__(inout self, other: Self):
        self.x = other.x
        print("copyinit")
    fn __moveinit__(inout self, owned other: Self):
        self.x = other.x
        print("moveinit")
    fn __del__(owned self): print("del")

fn main():
    a = MyStruct(1)
    b = rebind[MyStruct2](a)
    print(b.x)


```
#### output:
```
   copyinit
   del
   1
   del
```

I think that this is the same thing that's blocking us from returning `ref` instead of `Reference` in iterators,

tried one time to make it for `List`, 

and it would error because types not implementing `Copyable`

(not sure if it is a bug or a feature)

**Waiting for feedback to create a bug report or not**

soraros:
The copyinit happens on `b = rebind[MyStruct2](a)`. I believe the code I touched doesn't do this:
- `get` returns a `ref T`
- `rebind[T](a).__eq__(b)` borrows `self`
- The code regrading `StaticIntTuple` was returning `Int`, so copy is fine

rd4com:
Yep that make sense :+1:

JoeLoser (commented):
Nice cleanup! Thank you!

JoeLoser:
!sync

stdlib/src/collections/inline_array.mojo
```diff
@@ -326,7 +326,7 @@ struct InlineArray[
     @always_inline
     fn __contains__[
         T: EqualityComparableCollectionElement, //
-    ](self, value: T) -> Bool:
+    ](self: InlineArray[T, size], value: T) -> Bool:
```
lattner, line 329: This is a great way to get rid of the `constrained` check later, which shifts the error message from the elaborator (which produces terrible stack traces) up to parser time.  Could we use this in more places?

lattner: @soraros curious your view on this, I'm a fan :-)

soraros: Indeed we can, and I did some in #3210 in the past.

However, we can run into problems:
1. Heterogeneous container, example: [`Tuple.__getitem__`](https://github.com/modularml/mojo/blob/c687893d202d59f1556d1cb370a0e9f37d1984cf/stdlib/src/builtin/tuple.mojo#L195-L227)
2. Trait bound mismatch, example: `InlineList[T]` takes `T: CollectionElementNew` but nobody got around to add a `EqualityComparableCollectionElementNew` (only have `EqualityComparableCollectionElement` which is slightly different) to put on its `count` function. It is only temporary since we are in the process of moving towards `CollectionElementNew`. All that is to say that I really want anonymous trait bound (`T: A & B`).
3. We are missing "half" of the support for conditional conformance: we don't have conditional extension. This one covers the majority of cases where `constrained` is needed in the stdlib.
```mojo
trait Stringable:
  fn __str__(self) -> String: ...

struct List[T]:  # List itself doesn't conform to Stringable
  fn __str__[T: Stringable](self: List[T]) -> String: ...
```

I think there aren't many cases (if any at all module 2. above) where we can use CC to remove `constrained` or `_type_is_eq`.

PR #3557, opened 2024-09-27
[stdlib] Prevent crash for non-subscriptable `PythonObject`s

jjvraw:
Add a null check for `callable_obj` in `PythonObject.__getitem__` method to prevent crashes for non-subscriptable objects. 

**NOTE:**
This results in a different error than Python's `'<object>' is not subscriptable`:
- We raise an `AttributeError` at method lookup, while Python raises a `TypeError`.
    - Due to call of `PyObject_GetAttrString` as apposed to `PyObect_GetItem`

We may want to consider handling for non-variadic arguments, leveraging [`PyObject_GetItem`](https://docs.python.org/3/c-api/object.html#c.PyObject_GetItem). It's part of the stable ABI and may improve Python interop, perhaps some performance gain aswell.

### DIFF
--- a/stdlib/test/python/test_python_object.mojo
+++ b/stdlib/test/python/test_python_object.mojo
@@ -409,6 +409,50 @@ fn test_none() raises:
     assert_true(n is None)
 
 
+fn test_getitem_raises() raises:
+    var a = PythonObject(2)
+    with assert_raises(contains="'int' object has no attribute '__getitem__'"):
+        _ = a[0]
+
+    var b = PythonObject(2.2)
+    with assert_raises(
+        contains="'float' object has no attribute '__getitem__'"
+    ):
+        _ = b[0]
+
+    var c = PythonObject(True)
+    with assert_raises(contains="'bool' object has no attribute '__getitem__'"):
+        _ = c[0]
+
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
Nice improvement, thank you!

JoeLoser:
!sync

stdlib/test/python/test_python_object.mojo
```diff
@@ -409,6 +409,50 @@ fn test_none() raises:
...
+        _ = d[0]
+
+    var with_get = Python.evaluate(
+        "type('WithGetItem', (), {'__getitem__': lambda self, key: f\"Key:"
+        ' {key}"})()'
+    )
+    assert_equal("Key: 0", str(with_get[0]))
+
+    var without_get = Python.evaluate(
+        "type('WithOutGetItem', (), {'__str__': \"SomeString\"})()"
```
msaelices, line 440: NIT: Not affecting the test but to be precise, `__str__` is a callable:
```suggestion
        "type('WithOutGetItem', (), {'__str__': lambda self: \"SomeString\"})()"
```

msaelices (approved):
Great fix!

PR #3561, opened 2024-09-28
[stdlib] Transition `InlinedFixedVecor.type` to `CollectionElement`

rd4com:
This commit is a possible transition for `InlinedFixedVector`,

changing it's `type` parameter from `AnyTrivialRegType` to `CollectionElement`.

It is for a feature request by @joeloser:  

https://github.com/modularml/mojo/issues/3478

---

Not sure if it is exactly what you had in mind @joeloser,

but here is a possible implementation :+1:

### DIFF
--- a/stdlib/src/collections/vector.mojo
+++ b/stdlib/src/collections/vector.mojo
@@ -19,10 +19,10 @@ from collections import InlinedFixedVector
 ```
 """
 
+from memory.maybe_uninitialized import UnsafeMaybeUninitialized
 from memory import Reference, UnsafePointer, memcpy
+from collections import inline_array
 from sys import sizeof
-
-from utils import StaticTuple
 
 # ===----------------------------------------------------------------------===#
 # _VecIter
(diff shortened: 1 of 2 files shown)


szbergeron:
Would it be possible to use conditional conformance within `__del__` to only even emit the destruction code if the type isn't trivial? I know one of the wants from this type was originally that it be almost entirely trivial to move/copy (and would be if it doesn't overflow the inline vec)

JoeLoser (commented):
Looks great!  Just needs a rebase and a minor organization comment I made wrt to the tests.

Thanks for tackling this!

rd4com:
:facepalm: , this one got lost, never had a notification for this!
would you like me to try to revive this in same or other way ?
(ping to @JoeLoser)

JoeLoser:
> 🤦 , this one got lost, never had a notification for this! would you like me to try to revive this in same or other way ? (ping to @JoeLoser)

No worries, the InlineFixedVector type got removed so we can close this PR now. Thanks for working initially on this though!

PR #3565, opened 2024-09-29
[stdlib] Add `array_tester` with `Tester` and `Testee` for lists

rd4com:
Introducing two new types for testing lists and datastructures.
(`Tester` and `Testee`)

`Testee` can be used as the element type of an array or a similar datastructure.
`Tester` holds `Int` fields to count the amount of moves, copy, init, del.

Multiple instances of `Testee` contains an `UnsafePointer` to an instance of `Tester`.
Assertions can be created on the current amount of `init`, `copy`, `del`, `moves` a list does.

Additionally, `Testee` have an `Int` field to store a value.
That value can be used to check the equality of the elements of two lists. 

Example:
```mojo
def main():
    var _tester = Tester()
    var x = List[Testee]()
    for i in range(10):
        x.append(Testee(_tester, i))
    for i in x: print(i[].value)
    assert_equal(_tester.oninit, 10)
    assert_equal(_tester.ondelete, 0)
    assert_equal(_tester.oncopy, 0)
    y = x[0]
    assert_equal(y.value, 0)
    assert_equal(_tester.oncopy, 1)
    _ = x.pop()
    assert_equal(_tester.ondelete, 2)
    __type_of(x).__del__(x^)
    assert_equal(_tester.ondelete, 11)
    assert_equal(_tester.oncopy, 1)
```

### DIFF
--- a/stdlib/src/utils/array_tester.mojo
+++ b/stdlib/src/utils/array_tester.mojo
@@ -1,0 +1,131 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Defines `Testee` and `Tester`. 
+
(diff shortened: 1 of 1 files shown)


JoeLoser (commented):
Thanks for the contribution!  Can this be actively applied to any existing stdlib tests, or what's the motivation compared to some of the other test utils we have?

rd4com:
Hello @JoeLoser , closing this :+1: 

DelRecorder/DelCounter are the best ones and can do it  !

This is still available for thinking if needed,
we should be able to do better than this pr with lists of __call_locations.

For counting, pointers are really good,
because some datastructures could have `__copyinit__` just while moving values,
if the count was a field and not a pointer, the value count is del.

I think that counter for this should have two fields,
but it is not possible to increment a copy_done field on existing because it is borrowed.

Issue #3538, opened 2024-09-23
[BUG] Constraint in `rotate_bits_left` is given in bytes instead of bits


samufi:
### Bug description

The functions `rotate_bits_left` and `rotate_bits_right` take the desired shift of bit positions as parameter. This parameter is constrained to the size of the considered type (which makes sense). However, the size of the type is determined via the `sizeof` function, which returns the size in byte count not in bit count. Therefore, a shift of more bits than the type has bytes fails. 

For example, `Int` typically has 64 bits and 8 bytes. The bits should be shiftable by 63 positions, but they can only be shifted by 7 positions.

### Steps to reproduce

```mojo
from bit import rotate_bits_left

def main():
    var a: Int = 0
    _ = rotate_bits_left[1](a) # works
    _ = rotate_bits_left[2](a) # works
    _ = rotate_bits_left[3](a) # works
    _ = rotate_bits_left[4](a) # works
    _ = rotate_bits_left[5](a) # works
    _ = rotate_bits_left[6](a) # works
    _ = rotate_bits_left[7](a) # works
    _ = rotate_bits_left[8](a) # fails
```

The error message reads

```
note: constraint failed: Constraints: -sizeof[Int]() <= shift < sizeof[Int]()
```

### System information

```shell
- WSL on Windows 10
- mojo 24.5.0 (e8aacb95)
- magic 0.2.3
```

yichi170:
I believe it could be a good first issue. I'd love to help and submit a fix if it's available

JoeLoser:
> I believe it could be a good first issue. I'd love to help and submit a fix if it's available

Please go ahead! I'll happily review a PR. Thanks!

PR #3569, opened 2024-09-30
[stdlib] fix: set constraint in `rotate_bits_*` function to bit size

yichi170:
Set constraint to bit size instead of byte size in `rotate_bits_left` and `rotate_bits_right`

Fixes: #3538


JoeLoser (approved):
Thanks!  I do idly wonder whether it's a good idea to allow negative shifts.

JoeLoser:
!sync

PR #3577, opened 2024-09-30
[stdlib] Refactor `memset()` to be generic for all trivial types

martinvuyk:
Refactor `memset` to be generic over scalars and trivial types to be filled with scalar values of the same bitwidth.

### DIFF
--- a/mojo/stdlib/stdlib/memory/memory.mojo
+++ b/mojo/stdlib/stdlib/memory/memory.mojo
@@ -283,40 +283,80 @@ fn memcpy[
 
 @always_inline("nodebug")
 fn _memset_impl[
-    address_space: AddressSpace
+    dtype: DType, address_space: AddressSpace, //
 ](
-    ptr: UnsafePointer[Byte, address_space=address_space],
-    value: Byte,
+    ptr: UnsafePointer[Scalar[dtype], address_space=address_space],
+    value: Scalar[dtype],
     count: Int,
 ):
-    @parameter
-    fn fill[width: Int](offset: Int):
-        ptr.store(offset, SIMD[DType.uint8, width](value))
+    alias width = simdwidthof[Scalar[dtype]]()
+    var vector_end = align_down(count, width)
 
-    alias simd_width = simdwidthof[Byte]()
-    vectorize[fill, simd_width](count)
+    for i in range(0, vector_end, width):
+        ptr.store(i, SIMD[dtype, width](value))
+
+    for i in range(vector_end, count):
+        ptr.store(i, value)
 
 
 @always_inline
 fn memset[
-    type: AnyType, address_space: AddressSpace
-](
-    ptr: UnsafePointer[type, address_space=address_space],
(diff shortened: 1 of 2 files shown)


soraros:
Thanks for the contribution and congrats on the performance improvements! However, I do wonder if `memset` is right place to add such features. It has the semantics of copying raw memory, and `init_pointee_copy` is not that.

martinvuyk:
Hi @soraros the implementation in `_memset_impl` also just copies a value on a vector and then stores it in the memory address range:
```mojo
alias simd_width = simdwidthof[Scalar[type]]()
var vector_end = _align_down(count, simd_width)

for i in range(0, vector_end, simd_width):
    ptr.store(i, SIMD[type, simd_width](value))

for i in range(vector_end, count):
    ptr.store(i, value)
```

And the original signature of `memset` also made me consider doing this as I interpreted it like it is supposed to be generic:
```mojo
fn memset[
    type: AnyType, address_space: AddressSpace
](ptr: UnsafePointer[type, address_space], value: UInt8, count: Int):
  ...
```
the only thing this new version assumes differently than the original is that you are able to copy a value of the same type as the pointer into that address range. 

IMO these two branches are doing the same thing, copy a thing `count` times from the pointer start
```mojo
if dt is not DType.invalid:
    var p = ptr.bitcast[Scalar[dt]]()
    _memset_impl[dt](p, rebind[Scalar[dt]](value), count)
else:
    for i in range(count):
        (ptr + i).init_pointee_copy(value)
```

Even `memcpy` also does copies onto vectors before storing in an address range, so there is no "raw" bit-mapping on an electrical level either:
```mojo
# Copy in 32-byte chunks.
alias chunk_size = 32
var vector_end = _align_down(n, chunk_size)
for i in range(0, vector_end, chunk_size):
    dest_ptr.store(i, src_ptr.load[width=chunk_size](i))
for i in range(vector_end, n):
    dest_ptr.store(i, src_ptr.load(i))
```

JoeLoser:
!sync

JoeLoser:
FYI we'll want to do some benchmarking here (both in comp and runtime) with this change internally, so that's why it has not been merged internally yet.  Thanks for your patience and contribution! 🔥

martinvuyk:
@jackos to reduce the scope of this PR I reverted the changes to make memset generic over memory only types (I have a better idea to not bloat compile times). I also removed the `DType.get_dtype()` to its own PR #3810. This PR now only makes `memset` generic over scalars and trivial types to be filled with scalar values of the same bitwidth.

JoeLoser:
!sync

JoeLoser:
Just coming back to this as I'm going through old PRs.  Can you please rebase this and spruce up the PR description for the motivation and then I'll have someone from the team take a look?  Thanks.

martinvuyk:
@JoeLoser I rebased, but I can't think of more motivation for the PR that it already has in the description :man_shrugging:

JoeLoser:
!sync

PR #3587, opened 2024-10-01
[stdlib] Update gitignore for magic

jackos:
Copybara doesn't bring these files across, so raising a PR directly to this repo


jackos:
Already raised in: https://github.com/modularml/mojo/pull/3531

PR #3588, opened 2024-10-01
[stdlib] Clean up `memory.unsafe`

soraros:
- Make more things infer-only
- Remove unnecessary overload
- Rename the `bitcast` overload that performs the 'movemask' operation to `pack_mask`. This change is intended to prevent issues where the wrong overload might be selected, and since there is an implicit conversion from scalar to simd at return, the user won't get a type mismatch error to warn them about that. @JoeLoser, do you think this change is a good idea? The function in NumPy is called `packbits`, maybe that's a better name.

### DIFF
--- a/stdlib/src/memory/unsafe.mojo
+++ b/stdlib/src/memory/unsafe.mojo
@@ -28,18 +28,21 @@ from sys import bitwidthof
 
 @always_inline("nodebug")
 fn bitcast[
-    new_type: DType, new_width: Int, src_type: DType, src_width: Int
-](val: SIMD[src_type, src_width]) -> SIMD[new_type, new_width]:
+    type: DType,
+    width: Int, //,
+    new_type: DType,
+    new_width: Int = width,
+](val: SIMD[type, width]) -> SIMD[new_type, new_width]:
     """Bitcasts a SIMD value to another SIMD value.
 
     Constraints:
         The bitwidth of the two types must be the same.
 
     Parameters:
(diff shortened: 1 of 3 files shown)


stdlib/src/memory/unsafe.mojo
```diff
@@ -65,45 +68,31 @@ fn bitcast[
...
+        return DType.uint32
+    else:
+        return DType.uint64
 
 
 @always_inline("nodebug")
-fn bitcast[
-    new_type: DType, src_width: Int
-](val: SIMD[DType.bool, src_width]) -> Scalar[new_type]:
+fn pack_mask[
```
JoeLoser, line 83: +1 to the rename

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
> @JoeLoser Do we need to add changelog entries for this and perhaps the `to_bits` changes in `SIMD`? The changes might be too trivial to warrant it, but people could encounter issues with the `_floats_to_bits` name change.

Wouldn't hurt, I agree. Want to write something up?  If not, I can next week.

soraros:
@JoeLoser I guess it's the best if you could do is since both PRs are now merged. Thanks in advance!

PR #3589, opened 2024-10-01
[stdlib] Use magic to run stdlib tests in CI

jackos:


stdlib/COMPATIBLE_COMPILER_VERSION
```diff
@@ -1 +1 @@
-24.3.0
+24.5.0
```
JoeLoser, line 1: FYI @patrickdoc as this was missed in the recent releases

JoeLoser (approved):

JoeLoser:
Thanks for fixing this up, Jack!  Much appreciated.

PR #3591, opened 2024-10-02
[stdlib] Add `String` `__add__` with `StringSlice` and `StringLiteral` & optimize to not use `List.resize()`

martinvuyk:
Add `String` `__add__` with `StringSlice` and `StringLiteral` & optimize to not use `List.resize()`


stdlib/src/collections/string.mojo
```diff
@@ -1090,19 +1090,34 @@ struct String(
...
+    fn __add__(self, other: StringSlice) -> String:
+        """Creates a string by appending a string slice at the end.
+
+        Args:
+            other: The string slice to append.
+
+        Returns:
+            The new constructed string.
+        """
+        if not self:
```
JoeLoser, line 1109: **Suggestion** The logic here, in `__radd__`, and in the above `__add__` is all similar (get a pointer to the underlying data, do some memcpy of data, etc.).  Can we factor that into a private helper function and use it in these places to avoid duplication?

JoeLoser (approved):
LGTM once conflicts are addressed and then I can sync it.  Thanks!

martinvuyk:
@JoeLoser fixed :)

JoeLoser:
!sync

PR #3583, opened 2024-10-01
[stdlib] Speedup (~1.4x) on `PythonObject.__getitem__`

jjvraw:
Noticing consistent ~1.4x speedup for basic benchmarks on my machine (M1), example:

```mojo
for i in range(iterations):
    var idx_a = random_si64(0, 999)
    var idx_b = random_si64(0, 9999)
    var idx_c = random_si64(0, 99999)
        
    _ = a[idx_a]
    _ = b[idx_b]
    _ = c[idx_c]
```

This essentially free speedup comes from removing any overhead associated with `PyObject_GetAttr`, by using [`PyObject_GetItem`](https://docs.python.org/3/c-api/object.html#c.PyObject_GetItem) directly. Also noticed minimising the use of  `PythonObject` wrappers, and working directly with `PyObjectPtr` attributes contributed to the speedup as well.

In addition, multi-dimensional indexing into Python objects are now possible, such as below

```mojo
var np = Python.import_module("numpy")
var a = np.array(PythonObject([1,2,3,1,2,3])).reshape(2,3)
print((a[0, 1]))
```

Previously, this would raise an `expected 1 argument, got 2` exception.

Lastly, error messages are also now consistent to Python, for non-subscriptable objects (mentioned in #3557).

PR #3595, opened 2024-10-03
[stdlib] Integrate `PyObject_SetItem` for speedup and error handling

jjvraw:
Following the trend of:
- #3549 
- #3583 

Again, noticing ~1.4x speedup on basic benchmarks:
```mojo
for _ in range(iterations):
    var idx_a = random_si64(0, 999)
    var idx_b = random_si64(0, 9999)
    var idx_c = random_si64(0, 99999)
    
    a[idx_a] = idx_a
    b[idx_b] = idx_b
    c[idx_c] = idx_c 
```

and improved error handling, preventing crashes.

### DIFF
--- a/stdlib/src/python/_cpython.mojo
+++ b/stdlib/src/python/_cpython.mojo
@@ -1009,6 +1009,26 @@ struct CPython:
         self._inc_total_rc()
         return r
 
+    fn PyObject_SetItem(
+        inout self, obj: PyObjectPtr, key: PyObjectPtr, value: PyObjectPtr
+    ) -> Int:
+        var r = self.lib.get_function[
+            fn (PyObjectPtr, PyObjectPtr, PyObjectPtr) -> Int
+        ]("PyObject_SetItem")(obj, key, value)
+
+        self.log(
+            "PyObject_SetItem result:",
+            r,
+            ", key:",
+            key._get_ptr_as_int(),
+            ", value:",
+            value._get_ptr_as_int(),
+            ", parent obj:",
+            obj._get_ptr_as_int(),
+        )
+
+        return r
+
     fn PyObject_GetAttrString(
         inout self,
         obj: PyObjectPtr,
--- a/stdlib/src/python/_cpython.mojo
+++ b/stdlib/src/python/_cpython.mojo
@@ -1011,9 +1011,9 @@ struct CPython:
 
     fn PyObject_SetItem(
         inout self, obj: PyObjectPtr, key: PyObjectPtr, value: PyObjectPtr
-    ) -> Int:
+    ) -> c_int:
         var r = self.lib.get_function[
-            fn (PyObjectPtr, PyObjectPtr, PyObjectPtr) -> Int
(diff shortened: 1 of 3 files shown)


stdlib/src/python/_cpython.mojo
```diff
@@ -1009,6 +1009,26 @@ struct CPython:
         self._inc_total_rc()
         return r
 
+    fn PyObject_SetItem(
+        inout self, obj: PyObjectPtr, key: PyObjectPtr, value: PyObjectPtr
+    ) -> Int:
+        var r = self.lib.get_function[
+            fn (PyObjectPtr, PyObjectPtr, PyObjectPtr) -> Int
```
JoeLoser, line 1016:
```suggestion
            fn (PyObjectPtr, PyObjectPtr, PyObjectPtr) -> c_int
```

Otherwise, this is incorrect from an ABI perspective (Int is 8 byte but the function in CPython is `int`, which is 4 bytes.  If you'd like to fix this here and for the other function pointers that run afoul of this problem, that would be great! 

These are incorrectly using Int, which is 64 bit, whereas c_int is almost always 32 bits on modern platforms we support.

This hasn't been a major problem, particularly for return values, because Int is typically a small-valued error status code stored in a little-endian register upon return.

JoeLoser (approved):
Looks great, thank you!  See my comment about the ABI issue with the function pointer.

stdlib/src/python/_cpython.mojo
jjvraw: Thank you for making me aware of this. Will update this addition and open a separate PR for the others

JoeLoser: That would be very much appreciated, thank you!  We're looking all the rough edges of FFI right now, including the ABI struggles, as we've been working in this neighborhood with CPython APIs recently (@ConnorGray has gotten burned a time or two I'd say lately 😆 ).

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

stdlib/src/python/_cpython.mojo
soraros: I think we should put a link to the Python doc in all these functions in `CPython` so people can check for API/ABI mismatch more easily.
```mojo
# Ref: https://docs.python.org/3/c-api/object.html#c.PyObject_SetItem
```

soraros: > Regarding consistency, is there any preference for matching the return types? ...
> ... put a link to the Python doc in all these functions in `CPython` ...
@jjvraw Let's create an issue to track these so Conner can let us know what's the best way forward.

jjvraw: I'm actually of the opinion that in this case, there is no benefit in the type cast. At least for how `PyObject_IsTrue` is currently being used in the lib now.

However, given the points Joe made and that is seems to be an internal goal at the moment, it might be better for Joe, or someone from the team, to open an issue

PR #3597, opened 2024-10-03
[stdlib] Make `StringSlice` `CollectionElement`

martinvuyk:
Make `StringSlice` `CollectionElement`


soraros:
Can we simply use `@value` here?

martinvuyk:
HI @soraros. I'm seeing that `Span` uses `@value` but it's components are just an `UnsafePointer` and an `Int`, so if we used `@register_passable("trivial")` here it would need to be changed as well. BTW I also still don't quite understand the difference between `@register_passable("trivial")` and plain `@register_passable`. WDTY?

soraros:
@martinvuyk I'm not sure either. Let's wait for the team to respond.

JoeLoser (approved):

martinvuyk:
@JoeLoser fixed the merge conflict but the CI keeps failing on `error: cannot open input file '/home/runner/work/mojo/mojo/stdlib/scripts/check-licenses.mojo` for this and other PRs

JoeLoser:
!sync

PR #3604, opened 2024-10-04
[stdlib] Adopt AHasher to Hasher trait

mzaks:


stdlib/src/hashlib/_ahash.mojo
```diff
@@ -71,17 +72,13 @@ fn read_small(data: UnsafePointer[UInt8], length: Int) -> U128:
...
 
-    fn __init__(inout self, key: U256):
-        """Initialize the hasher with a key.
-
-        Args:
-            key: Modifier for the computation of the final hash value.
-        """
+    fn __init__(inout self):
+        """Initialize the hasher."""
         var pi_key = key ^ U256(
```
JoeLoser, line 82: **Question** Can this be an `alias` since the `key` is a parameter now?

stdlib/src/hashlib/_ahash.mojo
```diff
@@ -175,6 +205,6 @@ fn hash[
         hash collision statistical properties for common data structures.
     """
 
-    var hasher = AHasher(key)
-    hasher.write(bytes, n)
-    return UInt(int(hasher.finish()))
+    var hasher = AHasher[key]()
+    hasher._update_with_bytes(bytes, n)
+    return UInt(int(hasher^.finish()))
```
JoeLoser, line 210: **Question** Why do we need the `int(...)`?  Also, should we change the return type of this function to be `UInt64` (i.e. match the return type of the `finish()` API)?

JoeLoser (approved):
Looks great, just a few small questions!

stdlib/src/hashlib/_ahash.mojo
soraros: > Why do we need the `int(...)`?

There is no `__uint__`, so `finish(): UInt64` need to roundtrip to `index` through `int` first.

JoeLoser:
!sync

stdlib/src/hashlib/_ahash.mojo
mzaks: Yeah, this function is temporary it will be replaced with the `_hash_with_hasher` function currently located in `_hasher` module. So I think now it is best to just change the type to be `UInt64`

JoeLoser:
!sync

PR #3621, opened 2024-10-06
Filehandle as a copiable and movable @value

msaelices:
`FileHandle` should be copyable and movable and should be able to be passed by value or add them to lists, tuples, etc. This allows for more ergonomic and efficient file handling in Mojo.


docs/changelog.md
```diff
@@ -16,6 +16,10 @@ what we publish.
 
 ### ⭐️ New
 
+- `FileHandle` are now copyable and movable and can be passed by value or add
```
gryznar, line 19: *is

gabrieldemarmiesse:
I'm wondering what happens if I have a file handle `f1`, then `f1` is copied into `f2`, then `f1` is deleted. As I understand it, the file is going to be closed, that's how I understand the `__del__` implementation. Then it means `f2` isn't usable anymore. So maybe file handles should be movable but not copyables to avoid footguns?

I tried it in python and it's not possible to do a deepcopy of a file-handle, suggesting that it's not copyable in python either:

```
>>> f1 = open("./hello", "w+")
>>> import copy
>>>
>>> f2 = copy.deepcopy(f1)
  File "C:\Users\yolo\miniconda3\lib\copy.py", line 161, in deepcopy
    rv = reductor(4)
TypeError: cannot pickle '_io.TextIOWrapper' object
```

martinvuyk:
> So maybe file handles should be movable but not copyables to avoid footguns?

We could also replace:
```mojo
var handle: OpaquePointer
"""The underlying pointer to the file handle."""
```
with
```mojo
var handle: Arc[OpaquePointer]
"""The underlying pointer to the file handle."""
```

And then the desctructor would be 
```mojo
fn __del__(owned self):
    """Closes the file handle if it is the last reference to it."""
    try:
        if self.handle.count() == 1:
            self.close()
    except:
        pass
```

But this might need a lot of refactoring (no idea if there will be a perf. hit) and making it `Movable` only might be "conceptually safer"

msaelices:
> > So maybe file handles should be movable but not copyables to avoid footguns?

Ohhh! I don't know why I missed this case. You are totally right! Thanks for the heads up.

> 
> We could also replace:
> 
> ```mojo
> var handle: OpaquePointer
> """The underlying pointer to the file handle."""
> ```
> 
> with
> 
> ```mojo
> var handle: Arc[OpaquePointer]
> """The underlying pointer to the file handle."""
> ```
> 
> And then the desctructor would be
> 
> ```mojo
> fn __del__(owned self):
>     """Closes the file handle if it is the last reference to it."""
>     try:
>         if self.handle.count() == 1:
>             self.close()
>     except:
>         pass
> ```
> 
> But this might need a lot of refactoring (no idea if there will be a perf. hit) and making it `Movable` only might be "conceptually safer"

Great advice. I'm not sure if it's worth it but I followed your suggestion and refactored `FileHandle` to wrap the actual handle with `Arc`: https://github.com/msaelices/mojo/commit/856b0328c6ca6095deeacdc092c2d146f388a385

gryznar:
If it is not copyable in Python, what are the advantages of adding this to Mojo, while it is required to opaque it with `Arc` which may reduce overall performance?

gryznar (commented):
If it is not copyable in Python, what are the advantages of adding this to Mojo, while it is required to opaque it with `Arc` which may reduce overall performance (for moving `Arc` is not required)?

docs/changelog.md
gryznar: Gentle ping, still actual :)

stdlib/src/builtin/file.mojo
```diff
@@ -67,15 +68,16 @@ struct _OwnedStringRef(Boolable):
         return self.length != 0
 
 
+@value
 struct FileHandle:
     """File handle to an opened file."""
 
-    var handle: OpaquePointer
+    var handle: Arc[OpaquePointer]
```
lsh, line 75: I don't like the idea of making this choice for users rather than them making the choice themselves. If a user does not need to pay the cost of an `Arc`, then they probably won't want to. I do think making it `Movable` has value though.

lsh (changes requested):

PR #3632, opened 2024-10-10
[stdlib] Refactor `external_call()` and inline `_LITRefPackHelper` into `VariadicPack`

martinvuyk:
Refactor `external_call()` and inline most of `_LITRefPackHelper` into `VariadicPack`, leave `_LITRefPackHelper` itself intact since I found no way to inline the `AddressSpace` parameter into `VariadicPack`.


stdlib/src/builtin/builtin_list.mojo
```diff
@@ -463,23 +463,30 @@ struct VariadicListMem[
...
 
 
 alias _AnyTypeMetaType = __mlir_type[`!lit.anytrait<`, AnyType, `>`]
 
 
-@value
-struct _LITRefPackHelper[
-    is_mutable: Bool, //,
-    lifetime: Lifetime[is_mutable].type,
-    address_space: __mlir_type.index,
```
soraros, line 477: Is this of any significance?

martinvuyk: Only usage I could find was to do `get_loaded_kgen_pack(self) -> Self._loaded_kgen_pack_type` which is just a subset of what VariadicPack's functionality should be since it is also a `!lit.ref.pack<:variadic` wrapper. That's why I inlined it :man_shrugging:

stdlib/src/sys/ffi.mojo
```diff
@@ -383,39 +419,93 @@ fn _get_dylib_function[
...
+        __mlir_op.`pop.external_call`[func = callee.value, _type=None](p)
         return rebind[type](None)
     else:
-        return __mlir_op.`pop.external_call`[func = callee.value, _type=type](
-            loaded_pack
-        )
+        return __mlir_op.`pop.external_call`[func = callee.value, _type=type](p)
+
+
+# FIXME: no one should be having to do manual pop.external_call for extra_args
```
soraros, line 468: I don't think we need to include this now (or in the future when unpacking syntax becomes available).

JoeLoser:
!sync

stdlib/src/builtin/builtin_list.mojo
JoeLoser: We do want to rely on different address spaces with variadics.  Currently, there's a known bug in the compiler lowering for variadics with non-generic address spaces that we ran into a few months ago internally.  I'd recommend we preserve the address space for now since we do intend to use it.

Of note, I don't think we have any non-generic address space tests (since it mostly comes up with GPU things) and the GPU module is not open source.  We can't easily run those tests in the open source repo right now.

martinvuyk: Ok, do I just set it up like it was for `_LITRefPackHelper` in its params and its `_mlir_value` ? Kind of a dumb question but I don't really understand MLIR yet, it is a bit too abstract for me sometimes

JoeLoser:
CC: @ConnorGray as you have an internal PR up for consolidating `_LITRefPackHelper` and `VariadicPack` from other work you were d oing.

ConnorGray:
!sync

martinvuyk:
@ConnorGray fixed all conflicts, this became mostly a cleanup PR

ConnorGray:
!sync

JoeLoser:
@martinvuyk do you mind rebasing this on top of @ConnorGray's recent changes in this area/the latest nightly?  We'd love to incorporate the doc improvements here.

martinvuyk:
@JoeLoser done

PR #3639, opened 2024-10-10
[stdlib] Clean up `CPython` primitive type conversion

soraros:
- Align types with the corresponding C API.
- Add function signatures and URLs to the CPython API documentation.
- Reorder methods in the following order: `bool`, `int`, `float`, `str`. Within each type, place `PyT_FromX` before `PyT_AsX`.
- Remove redundant overloads.

CC: @ConnorGray

### DIFF
--- a/stdlib/src/python/python_object.mojo
+++ b/stdlib/src/python/python_object.mojo
@@ -331,7 +331,7 @@ struct PythonObject(
         Args:
             none: None.
         """
-        self = Self(none=NoneType(none))
+        self = Self(none=NoneType())
 
     fn __init__(inout self, none: NoneType):
         """Initialize a none value object from a `None` literal.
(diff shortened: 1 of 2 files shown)


JoeLoser:
!sync

JoeLoser:
!sync

stdlib/src/python/python_object.mojo
```diff
@@ -1412,8 +1398,26 @@ struct PythonObject(
...
+        return cpython.PyLong_AsSsize_t(self.py_object)
+
+    fn __float__(self) -> Float64:
+        """Returns a float representation of the object.
+
+        Returns:
+            A floating point value that represents this object.
+        """
+        cpython = _get_global_python_itf().cpython()
+        return cpython.PyFloat_AsDouble(self.py_object)
```
JoeLoser, line 1411: **Suggestion** The [docs](https://docs.python.org/3/c-api/float.html#c.PyFloat_AsDouble) show this method may fail. Want to handle this failure handling in a follow-up? The lack-of-failure-handling was pre-existing, so going to merge this as is for now.

soraros: Indeed we need to add error check to all the python to mojo conversion functions.

PR #3650, opened 2024-10-12
[stdlib] Span slicing with negative step

### LINKED ISSUE #3649: [BUG] Span reversed slicing are not giving the right values
### Bug description

The construct `String(...).as_bytes_span()[::-1]` does not return the expected iterator 

### Steps to reproduce

image

### System information

```shell
mojo 2024.10.1105 (e911bf68)
```

```
modular 0.9.3 (020e342b)
```

```
      Pixi version: 0.29.0
          Platform: linux-64
  Virtual packages: __unix=0=0
                  : __linux=6.8.0=0
                  : __glibc=2.35=0
                  : __cuda=12.2=0
                  : __archspec=1=skylake
         Cache dir: /home/msaelices/.cache/rattler/cache
      Auth storage: /home/msaelices/.rattler/credentials.json

Project
------------
              Name: aoc2023
           Version: 0.1.0
     Manifest file: /home/msaelices/src/my-repos/advent-of-code/aoc2023/mojoproject.toml
      Last updated: 12-10-2024 12:44:31

Environments
------------
       Environment: default
          Features: default
          Channels: conda-forge, https://conda.modular.com/max-nightly
  Dependency count: 1
      Dependencies: max
  Target platforms: linux-64
```
```

### PR DESCRIPTION
**Pull request by msaelices**
Fixes https://github.com/modularml/mojo/issues/3649

```mojo
  1> s = String("01234")  
(String) s = "01234"
  2> for i in s.as_bytes_span()[::-1]: 
  3.   print(chr(int(i[]))) 
  4.    
4
3
2
1
0
  5>  
```


stdlib/src/utils/span.mojo
```diff
@@ -180,9 +180,18 @@ struct Span[
...
+        if step < 0:
+            step = -step
+            var new_len = (start - end + step - 1) // step
+            var dest_ptr = UnsafePointer[T].alloc(new_len)
+            i = 0
+            while start > end:
+                dest_ptr[i] = self._data[start]
+                start -= step
+                i += 1
+            return Span[T, lifetime](unsafe_ptr=dest_ptr, len=new_len)
```
msaelices, line 193: How can we make sure the `dest_ptr` allocated memory is destroyed when the span is deleted?

martinvuyk: Hmmm, not sure since I don't really understand lifetimes yet. But I think building a `var items = List[T](unsafe_ptr=dest_ptr, size=new_len, capacity=new_len)` and returning `Span[T, lifetime](items^)` will work?

Either way I think this should probably return an iterator in the future (once we have them), it doesn't make much sense to allocate for this since it is a read only view on the data.

msaelices:
> * `Span` suppose to be a non-owning view, I don't think this is the right fix.
> * Actually, I'm not even sure a span with only a length is capable of representing slicing with negative step at all.

Yes, it's a little bit weird. However, in Python you can slice a slice, like this:
```python
>>> l = list(range(10))[2:5][::-1]
>>> l
[4, 3, 2]
>>> 
```

How Mojo could support this kind of operations without the support of negative steps?

bgreni:
> * Actually, I'm not even sure a span with only a length is capable of representing slicing with negative step at all.

I made a change previously to kinda hack it by giving `Span` a `Slice` member here #2950, but apparently it caused issues and was reverted internally? It feels like something we _should_ be able to do though.

soraros:
#2950 essentially demonstrates the points I made above: `Span` itself is indeed incapable of representing stride. I believe it's still desirable for `Span` to remain a simple pointer-length pair. Perhaps we could introduce a `StridedSpan`?

martinvuyk:
What do you think of this proposal #3653 ?

bgreni:
> #2950 essentially demonstrates the points I made above: `Span` itself is indeed incapable of representing stride. I believe it's still desirable for `Span` to remain a simple pointer-length pair. Perhaps we could introduce a `StridedSpan`?

That's reasonable, I wish we had some context for why that change didn't work internally though.

JoeLoser:
@lsh curious to get your thoughts here

JoeLoser:
> > #2950 essentially demonstrates the points I made above: `Span` itself is indeed incapable of representing stride. I believe it's still desirable for `Span` to remain a simple pointer-length pair. Perhaps we could introduce a `StridedSpan`?
> 
> That's reasonable, I wish we had some context for why that change didn't work internally though.

The internal revert just said

> Causing deterministic failures in pre and post submit.

I wonder if that's still the case/whether there was a real bug in the code? 🤔

bgreni:
> The internal revert just said
> 
> > Causing deterministic failures in pre and post submit.
> 
> I wonder if that's still the case/whether there was a real bug in the code? 🤔

I wonder if we could revisit such a change. For instance I think having `fn __getitem__(self, span: Slice)` return a `StringSlice` for `String` would be intuitive, but it can't work unless we have this sort of functionality.

JoeLoser (commented):
!sync

JoeLoser:
!sync

PR #3654, opened 2024-10-12
[stdlib] Implement `sub_with_sat` using LLVM intrinsic

soraros:


stdlib/src/utils/_utf8_validation.mojo
```diff
@@ -112,22 +111,22 @@ fn validate_chunk[
     alias v80 = SIMD[DType.uint8, simd_size](0x80)
     alias third_byte = 0b11100000 - 0x80
```
mzaks, line 112: Does it make sense to keep number representation in base 2 (`0b11100000 - 0b10000000`)? I think it might be a bit easier  to read. I just made a hex to binary conversions in my head :).

JoeLoser (approved):
Nice use of LLVM intrinsic here 👍

JoeLoser:
!sync

PR #3658, opened 2024-10-13
[stdlib] Mark assume as has_side_effect=False

soraros:
~~Also mark `assume` as side-effect free.~~


soraros:
Actually, will `assume` get DCEed if we set `has_side_effect` to `False`?

JoeLoser:
@jackos do you mind taking a look at this?  This is modeled after https://en.cppreference.com/w/cpp/language/attributes/assume.

jackos:
Hi @soraros I've updated externally to keep the `has_side_effect=False` for `assume` as that's the right thing to do, but removed `assume` from `debug_assert` as it's always DCE'd.

jackos (approved):
Hi @soraros I've updated externally to keep the has_side_effect=False for assume as that's the right thing to do, but removed assume from debug_assert as it's always DCE'd.

PR #3659, opened 2024-10-13
[stdlib] Add `expect` intrinsic

soraros:
- Add `llvm.expect` intrinsic to stdlib
- Refector `likely` and `unlikely` intrinsics to use `expect`
- Mark `likely` and `unlikely` intrinsics as side-effect free

### DIFF
--- a/stdlib/src/sys/intrinsics.mojo
+++ b/stdlib/src/sys/intrinsics.mojo
@@ -1505,6 +1505,34 @@ struct _RegisterPackType[*a: AnyTrivialRegType]:
 
 
 # ===----------------------------------------------------------------------=== #
+# expect
+# ===----------------------------------------------------------------------=== #
+
+
+@always_inline("nodebug")
+fn expect[T: AnyTrivialRegType, //, expected_val: T](val: T) -> T:
+    """Provides information about expected (the most probable) value of `val`,
+    which can be used by optimizers.
+
+    Constraints:
+        Only work with integer types.
+
+    Parameters:
+        T: The type of the input value.
+        expected_val: The expected value of `val`.
+
+    Args:
+        val: The input value.
+
+    Returns:
+        The input value.
+    """
+    return llvm_intrinsic["llvm.expect", T, has_side_effect=False](
+        val, expected_val
+    )
+
+
+# ===----------------------------------------------------------------------=== #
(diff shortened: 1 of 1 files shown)


stdlib/src/sys/intrinsics.mojo
```diff
@@ -1504,6 +1504,34 @@ struct _RegisterPackType[*a: AnyTrivialRegType]:
...
+        expected_val: The expected value of `val`.
+
+    Args:
+        val: The input value.
+
+    Returns:
+        The input value.
+    """
+    return llvm_intrinsic["llvm.expect", T, has_side_effect=False](
+        val, expected_val
```
martinvuyk, line 1531: The times I've built functions that take infer only params and try to use positional params i.e. `expect[True](some_expression)` it leads to it trying to set the infer only param as that value, so you end up having to use kwarg. So in order to avoid the kwarg being long, this seems a bit better to me
```suggestion
fn expect[T: AnyTrivialRegType, //, value: T](val: T) -> T:
    """Provides information about expected (the most probable) value of `val`,
    which can be used by optimizers.

    Constraints:
        Only work with integer types.

    Parameters:
        T: The type of the input value.
        value: The expected value of `val`.

    Args:
        val: The input value.

    Returns:
        The input value.
    """
    return llvm_intrinsic["llvm.expect", T, has_side_effect=False](
        val, value
```

stdlib/src/sys/intrinsics.mojo
```diff
@@ -1504,6 +1504,34 @@ struct _RegisterPackType[*a: AnyTrivialRegType]:
         return __mlir_op.`kgen.pack.extract`[index = i.value](self.storage)
 
 
+# ===----------------------------------------------------------------------=== #
+# expect
+# ===----------------------------------------------------------------------=== #
+
+
+@always_inline("nodebug")
+fn expect[T: AnyTrivialRegType, //, expected_val: T](val: T) -> T:
```
martinvuyk, line 1513: Deleted a previous comment about this, I thought the positional params issue was not resolved and it seems like it is if it's passing the tests :+1: solid addition

JoeLoser (approved):
LGTM, thanks!

JoeLoser:
!sync

Issue #3661, opened 2024-10-13
[Docs] Align `_cpython.mojo` docs with the C API


helehex:
### Where is the problem?

https://github.com/modularml/mojo/blob/main/stdlib/src/python/_cpython.mojo

### What can we do better?

The `_cpython` module provides bindings to CPython, so it should reflect and wrap it in a consistent way, with corresponding documentation. Currently some things are missing though.

This involves:
- Links to the official documentation, or docstrings which accurately describe the function
- Documenting cases where a function raises

Maybe this could fall under bindgen and be done automatically:
We could put the ref comments in and later pull in the documentation automatically and put it in as a docstring.
However, the documentation does not seem very gen-friendly.

### Anything else?

This was talked about in the public stdlib discussion [here](https://docs.google.com/document/d/1VKLr0-RXpbuuYd-iW4_u5zGyO990pd9Gr8bYd84VV-o/edit?tab=t.0#heading=h.v0f2z3w9btop))
@soraros started work on cleaning it up [here](https://github.com/modularml/mojo/pull/3639)

soraros:
We should also use types (`c_char` etc.) from `ffi` in the `get_function` calls.

soraros:
@laszlokindrat FYI, there's this tracking issue for the PRs I've been making.

JoeLoser:
The team (with help from many others in the community including Sora) has improved the docs for internal Python bindings related functionality including [https://github.com/modular/modular/blob/07edbaba710845f0ddc5598a8278f4778b120790/mojo/stdlib/stdlib/python/\_cpython.mojo](https://github.com/modular/modular/blob/07edbaba710845f0ddc5598a8278f4778b120790/mojo/stdlib/stdlib/python/\_cpython.mojo).<br><br>Is there something more specific you'd like to see here?  If so, let's create a new GitHub issue and narrowly scope it please.  Thanks!

soraros:
We've just completed the first pass of improving the documentation and hardening the ABI for the `_cpython` module. I believe the original issue can now be considered resolved.

laszlokindrat:
Thanks @soraros!

PR #3664, opened 2024-10-13
[stdlib] Document CPython bindings - Round 1

msaelices:
- [X] Fix pre-commit hook, broken after renaming check licenses script
- [X] Document PyGILState_STATE
- [X] Rename PyKeyValuePair to PyKeysValuePair to match CPython
- [X] Add docstrings to all the structs in cpython.mojo
- [X] Add single docstring regarding C-API docs to all the functions

Related issue: https://github.com/modularml/mojo/issues/3661


stdlib/src/python/_cpython.mojo
```diff
@@ -869,12 +1029,14 @@ struct CPython:
         return create_module_fn(module_def_ptr, module_api_version)
 
     # int PyModule_AddFunctions(PyObject *module, PyMethodDef *functions)
```
soraros, line 1031: Maybe we can move the comments like this (signature) to above the `get_function` call?

msaelices: Sure, done: https://github.com/msaelices/mojo/commit/6e5b526ed1ae5a21fe996c4f281928ef2ca9ae3e

.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
         stages: [commit]
       - id: check-license
         name: check-license
-        entry: mojo stdlib/scripts/check-licenses.mojo
+        entry: mojo stdlib/scripts/check_licenses.mojo
```
JoeLoser, line 18: Thanks for fixing this - I rolled it into https://github.com/modularml/mojo/pull/3667 which I can merge now to unblock nightly branch.  Note this file isn't managed by copybara.

stdlib/src/python/_cpython.mojo
```diff
@@ -79,7 +79,16 @@ alias newfunc = fn (PyObjectPtr, PyObjectPtr, PyObjectPtr) -> PyObjectPtr
 @value
 @register_passable("trivial")
 struct PyGILState_STATE:
+    """Represents the state of the Python Global Interpreter Lock (GIL).
+
+    This struct is used to store and manage the state of the GIL,
```
JoeLoser, line 84: 👍 currently our use of CPython APIs aren't GIL-aware (basically, Mojo holds the GIL exclusively which isn't good).  If you'd like to start this work, ping me.  I've fixed some of the recent ABI uses in this area, and for a minimal working start, we can essentially do the lock/unlocking at enter/exit of each CPython API we wrap.  A context manager used internally would look more aesthetic rather than pushing it onto callers, and we can talk about optimizing GIL lock/unlocking in the future (I have ideas here too).

JoeLoser (approved):
This is great, thanks so much for cleaning up and documenting this area!  Much appreciated as we're actively working in this neighborhood currently 😄

JoeLoser:
!sync

stdlib/src/python/_cpython.mojo
soraros: Yea, we started this documentation everything effort because I want to make changes like that. Please let us (the community) know what else we can do in #3661.

soraros (approved):
Great work, thanks a lot! I want to add signature to *every* `to_function` calls so we can easily check for API compliance. Let's do that in a subsequent PR.

Issue #3634, opened 2024-10-10
[Feature Request] Introduce `alias Byte = UInt8`


soraros:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

As title.

### What is your motivation for this change?

It helps document the intent, especially in low-level code (e.g., `memcpy`), where we treat the underlying type as just a pack of bits rather than a numerical type, similar to the distinction between `i1` and `bool`.

We *could* introduce a nominal type `Byte` which implicitly converts to `UInt8`, but that feels more complex than necessary.

### Any other details?

N/A

martinvuyk:
I think it's also important to note that in Python byte is it's own object type with it's own type of representation (prints with `b'{value_here}'`) and [many other byte related things](https://docs.python.org/3/library/stdtypes.html#binary-sequence-types-bytes-bytearray-memoryview).

Also, `Span[Uint8]` could be aliased to `ByteArray` and `List[UInt8]` to `Bytes` since their APIs are very similar to [byte operations](https://docs.python.org/3/library/stdtypes.html#bytes-and-bytearray-operations)

(btw typo in the title)

JoeLoser:
Have you seen [https://github.com/modularml/mojo/pull/2096#issuecomment-2067515392](https://github.com/modularml/mojo/pull/2096#issuecomment-2067515392) and our feedback there?  I'd be open to someone adding a lightweight alias in the meantime to improve readability/express intent better at call sites.

soraros:
> Have you seen [#2096 (comment)](https://github.com/modularml/mojo/pull/2096#issuecomment-2067515392) and our feedback there? 

Yea, I think `Bytes` needs way more justification and design.

> I'd be open to someone adding a lightweight alias in the meantime to improve readability/express intent better at call sites.

Let's move forward with simply `alias Byte = UInt8` then. I think it's a good first issue.

JoeLoser:
Sounds good, just added "good first issue".  Whoever takes this, feel free to add the alias and we can incrementally apply it where we're working with "Bytes".

PR #3670, opened 2024-10-14
[stdlib] Introduce `Byte`

soraros:
And use it in `memory.mojo`. Related to #3634.

### DIFF
--- a/stdlib/src/memory/memory.mojo
+++ b/stdlib/src/memory/memory.mojo
@@ -50,9 +50,10 @@ fn _align_down(value: Int, alignment: Int) -> Int:
 
 
 @always_inline
-fn _memcmp_impl_unconstrained[
+fn _memcmp_impl[
     type: DType
 ](s1: UnsafePointer[Scalar[type], _], s2: __type_of(s1), count: Int) -> Int:
+    constrained[type.is_unsigned(), "the input dtype must be unsigned"]()
     alias simd_width = simdwidthof[type]()
     if count < simd_width:
         for i in range(count):
@@ -94,14 +95,6 @@ fn _memcmp_impl_unconstrained[
 
 
 @always_inline
-fn _memcmp_impl[
-    type: DType
-](s1: UnsafePointer[Scalar[type], _], s2: __type_of(s1), count: Int) -> Int:
-    constrained[type.is_integral(), "the input dtype must be integral"]()
-    return _memcmp_impl_unconstrained(s1, s2, count)
-
-
-@always_inline
 fn memcmp[
     type: AnyType, address_space: AddressSpace
 ](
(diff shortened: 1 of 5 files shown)


JoeLoser (approved):

JoeLoser:
LGTM once tests are passing

stdlib/test/memory/test_memory.mojo
```diff
@@ -125,16 +125,16 @@ def test_memcmp():
...
     p2.store(120)
 
-    var c = memcmp(p1, p2, 1)
-    assert_equal(c, -1, "-120 is smaller than 120")
+    c = memcmp(p1, p2, 1)
+    assert_equal(c, 1)
 
     c = memcmp(p2, p1, 1)
-    assert_equal(c, 1, "120 is bigger than -120")
+    assert_equal(c, -1)
```
soraros, line 137: @JoeLoser I think this small change actually revealed a bug. `memcpy` should treat the underlying buffer as `unsigned char`. [ref](https://cplusplus.com/reference/cstring/memcmp/).

JoeLoser: Yeah, I think we discovered this on a PR a few months ago that `memcpy` was always working with signed int8 which was recipe for error.

JoeLoser:
!sync

stdlib/test/memory/test_memory.mojo
soraros: Sorry I made a typo. `memcpy` using `Int8` is fine (at least from the ABI point of view). It's actually that `memcmp` was also using `Int8`, which is a logical error.

stdlib/src/memory/memory.mojo
```diff
@@ -50,9 +50,10 @@ fn _align_down(value: Int, alignment: Int) -> Int:
 
 
 @always_inline
-fn _memcmp_impl_unconstrained[
+fn _memcmp_impl[
     type: DType
 ](s1: UnsafePointer[Scalar[type], _], s2: __type_of(s1), count: Int) -> Int:
+    constrained[type.is_unsigned(), "the input dtype must be unsigned"]()
```
soraros, line 56: Since we use `SIMD.__gt__` etc. in this function, if we don't restrict to `unsigned`, the result will be wrong.

soraros:
@JoeLoser This is a minor change, but I think we should add a changelog entry for it so people are aware and can start using it.

JoeLoser:
> @JoeLoser This is a minor change, but I think we should add a changelog entry for it so people are aware and can start using it.

Agreed, feel free to add a changelog entry to promote its use! 👍

soraros:
@JoeLoser I asked you since this PR is already merged. Is it possible that you add it internally?

stdlib/src/memory/memory.mojo
```diff
@@ -128,14 +121,14 @@ fn memcmp[
...
     @parameter
-    if sizeof[type]() >= sizeof[DType.int32]():
+    if sizeof[type]() >= sizeof[UInt32]():
         return _memcmp_impl(
-            s1.bitcast[Int32](),
-            s2.bitcast[Int32](),
-            byte_count // sizeof[DType.int32](),
+            s1.bitcast[UInt32](),
+            s2.bitcast[UInt32](),
+            byte_count // sizeof[UInt32](),
```
soraros, line 128: This is very unsafe and wrong.

JoeLoser:
> @JoeLoser I asked you since this PR is already merged. Is it possible that you add it internally?

Oh I see, yeah, I just opened a PR internally for a changelog entry - no biggie 👍

PR #3679, opened 2024-10-16
[stdlib] Fix example in `/floatable.mojo`

helehex:
Not sure if I should import `Variant` and use it unqualified.

### DIFF
--- a/stdlib/src/builtin/floatable.mojo
+++ b/stdlib/src/builtin/floatable.mojo
@@ -69,8 +69,10 @@ trait FloatableRaising:
     For example:
 
     ```mojo
+    from stdlib.utils import Variant
+
     @value
-    struct MaybeFloat(FloatableRasing):
+    struct MaybeFloat(FloatableRaising):
         var value: Variant[Float64, NoneType]
 
         fn __float__(self) raises -> Float64:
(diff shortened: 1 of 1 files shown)


stdlib/src/builtin/floatable.mojo
```diff
@@ -70,8 +70,8 @@ trait FloatableRaising:
 
     ```mojo
     @value
-    struct MaybeFloat(FloatableRasing):
-        var value: Variant[Float64, NoneType]
+    struct MaybeFloat(FloatableRaising):
+        var value: stdlib.utils.Variant[Float64, NoneType]
```
JoeLoser, line 74: Can you just say

```
from utils import Variant
```
after the mojo doc block and use `Variant` unqualified here, please?

helehex: Sure, I'm gonna look at more examples, so I'll do that going forward.

JoeLoser: That'd be great, thanks! I think of examples as "what would we like users to write/model code after?"  In that lens, we'd want them to import said type if it's not already imported in the prelude, and then use it unqualified.

stdlib/src/builtin/floatable.mojo
```diff
@@ -69,8 +69,10 @@ trait FloatableRaising:
     For example:
 
     ```mojo
+    from stdlib.utils import Variant
```
JoeLoser, line 72:
```suggestion
    from utils import Variant
```

JoeLoser (approved):

JoeLoser:
!sync

PR #3681, opened 2024-10-16
[stdlib] Fix `/_math.mojo` examples.

helehex:
FYI: `Truncable` is not in `math`, so that one still fails, not sure what's going on there.


helehex:
I guess `bulitin/_math.mojo` is imported in the `math/math.mojo`, but not `builtin/math.mojo`?

helehex:
So many different maths, i feel like there's a better way

JoeLoser (approved):

JoeLoser:
!sync

PR #3685, opened 2024-10-17
[stdlib] Fix `Roundable` example

helehex:
I left the `Absable` one (even though it's broken), since it will probably change with associated aliases.


JoeLoser (approved):

JoeLoser:
!sync

PR #3686, opened 2024-10-17
[stdlib] Refactor ASCII related things in `string.mojo`

martinvuyk:
Improvements:
- Refactor ASCII related things in `string.mojo`
- This fixes the `ascii` function to work like python's for multi byte characters.
- Add a `_repr() -> String` function which is much faster than the current one


stdlib/src/collections/string.mojo
```diff
@@ -29,88 +29,189 @@ from sys.intrinsics import _type_is_eq
...
+    ```mojo
+    print(ord("a"), ord("€")) # 97 8364
+    ```
+    .
+    """
 
-fn ord(s: StringSlice) -> Int:
-    """Returns an integer that represents the given one-character string.
+    # FIXME(#933): llvm intrinsic can't recognize !pop.scalar<ui8> value when
+    # trying to fold ctlz at comp time
```
martinvuyk, line 80: Sorry to ping you @lattner, but issue #933 is popping up again and I had to make this ugly workaround :cry:

martinvuyk:
Leaving as draft until PR #3792 is merged

PR #3692, opened 2024-10-18
Update variables.ipynb remove extra "from"

sahilmob:
From appeared twice under the Type annotations section.


Brian-M-J:
This issue does not exist in the nightly branch. See [here](https://github.com/modularml/mojo/blob/nightly/docs/manual/variables.ipynb?short_path=e4f676b#L174)

PR #3694, opened 2024-10-18
[stdlib] Introduce fixed-width `bin` and `hex`

### LINKED ISSUE #3671: [Feature Request] Introduce `hex` and `bin` functions that produce fixed-width string output
### Review Mojo's priorities

### What is your request?

As title.

In Python, the `bin` and `hex` functions return variable-length binary/hexadecimal string representations for `int` values. Python also follows a convention where negative numbers are represented as `-0bXXXX`, with `XXXX` being the binary/hexadecimal representation of the corresponding positive number. This convention primarily makes sense when working with arbitrarily large integers.

Mojo, on the other hand, uses machine integer types, which have well-defined binary/hexadecimal representations (e.g., two's complement for negative numbers). Introducing fixed-width variants of `bin` and `hex` functions would not only enhance their utility in low-level programming but also offer a potential performance improvement since the buffer size can be predetermined.

### Any other details?

N/A

soraros:
fix #3671

### DIFF
--- a/mojo/stdlib/src/bit/bit.mojo
+++ b/mojo/stdlib/src/bit/bit.mojo
@@ -19,8 +19,10 @@ from bit import count_leading_zeros
 ```
 """
 
+from memory import bitcast, unpack_bits
 from sys import llvm_intrinsic, sizeof
 from sys.info import bitwidthof
+from utils import Span
 from utils._select import _select_register_value as select
 
 # ===-----------------------------------------------------------------------===#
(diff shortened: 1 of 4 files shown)


stdlib/src/bit/bit.mojo
```diff
@@ -654,3 +656,80 @@ fn rotate_bits_right[
...
+    bytes = unpack_bits(r).cast[DType.uint8]()
+    s.unsafe_ptr().store(bytes + `0`)
+
+
+# fmt: off
+alias _table = SIMD[DType.uint8, 16](
+    ord("0"), ord("1"), ord("2"), ord("3"), ord("4"), ord("5"), ord("6"), ord("7"),
+    ord("8"), ord("9"), ord("a"), ord("b"), ord("c"), ord("d"), ord("e"), ord("f"),
+)
+# fmt: on
```
martinvuyk, line 704: TIL, neat 0.0

stdlib/src/memory/unsafe.mojo
```diff
@@ -119,3 +119,33 @@ fn bitcast[
     return __mlir_op.`pop.bitcast`[
         _type = __mlir_type[`!pop.scalar<`, new_type.value, `>`]
     ](val.value)
+
+
+@always_inline("nodebug")
+fn unpack_bits[
```
martinvuyk, line 125: :fire:

martinvuyk (commented):
Nice, though maybe add some explicit mention in the docstrings that this deviates from Python's format spec and how (examples would work best here IMO). I think it might confuse ppl otherwise why there are different versions in separate modules

soraros:
@martinvuyk Writing examples in the docstring is a good idea.

JoeLoser:
!sync

soraros:
@JoeLoser I still need to improve the doc string, add tests and maybe a changelog entry for this one.

JoeLoser (commented):
Generally looks good, but I would like to see some unit tests.

JoeLoser:
@soraros is this something you'd like to pursue still?  Do you have any thoughts on this @laszlokindrat?

mojo/stdlib/src/memory/unsafe.mojo
```diff
@@ -140,3 +140,33 @@ fn pack_bits[
     return __mlir_op.`pop.bitcast`[
         _type = __mlir_type[`!pop.scalar<`, new_type.value, `>`]
     ](val.value)
+
+
+@always_inline("nodebug")
+fn unpack_bits[
```
laszlokindrat, line 146: It feels like this should be either in `bit.mojo` or a staticmethod on `SIMD`. WDYT?

mojo/stdlib/src/bit/bit.mojo
```diff
@@ -611,3 +613,73 @@ fn rotate_bits_right[
...
             x, x, __type_of(x)(shift)
         )
+
+
+# ===----------------------------------------------------------------------===#
+# bin and hex
+# ===----------------------------------------------------------------------===#
+
+
+fn bin[dtype: DType, //](x: Scalar[dtype]) -> String:
```
laszlokindrat, line 623: I think we have these utilities in `format_int.mojo` already. I don't know if that's the best place for them, though.

PR #3697, opened 2024-10-20
[stdlib] Refactor and optimize `String.splitlines()`

martinvuyk:
Refactor and optimize `String.splitlines()`, add method `isnewline()` to `StringSlice` only since it is not a Python API but very necessary when going over string views to get better perf. 

Added a method `_to_string_list()` which is very fast in allocating list of strings from either `List[Span[Byte]]` or `List[StringSlice]`

Benchmark results:

CPU: Intel® Core™ i7-7700HQ

improvement metric: markdown percentage improvement (`(old_value - new_value) / old_value`)

Average improvement: 95% . In orders of magnitude, this is a 20x improvement

|Name                          | old_value (ms)           | new_value (ms)      | improvement|
|:-----------------------------|---------------------:|----------------:|------:|
|`bench_string_splitlines[10]` | 0.000940701744458119 | 0.000105671880950995 | 88.77%|
|`bench_string_splitlines[30]` | 0.0027447983499448 | 0.000178265230791744 | 93.51%|
|`bench_string_splitlines[50]` | 0.00458368162716199 | 0.000258061944157181 | 94.37%|
|`bench_string_splitlines[100]` | 0.00918959318972136 | 0.000439200495137525 | 95.22%|
|`bench_string_splitlines[1000]` | 0.0918725769310776 | 0.00332629453828916 | 96.38%|
|`bench_string_splitlines[10000]` | 0.923161754994739 | 0.0309016497554505 | 96.65%|
|`bench_string_splitlines[100000]` | 9.5575084572 | 0.327944576673714 | 96.57%|
|`bench_string_splitlines[1000000]` | 87.4676443712821 | 3.33141099393327 | 96.19%|

### Giving some context to the numbers
At an average 5 letters per word and 300 words per page (in the English language):
- 10: 2 words
- 30: 6 words
- 50: 10 words
- 100: 20 words
- 1000: ~ 1/2 page (200 words)
- 10_000: ~ 7 pages (2k words)
- 100_000: ~ 67 pages (20k words)
- 1_000_000: ~ 667 pages (200k words)


gryznar:
If it is not a Python API, so why it is named `isnewline` instead of `is_new_line`?

martinvuyk:
Mainly to keep consistency with `isspace()` which is the C standard name. `isnewline()` actually exists in windows and apple platforms, but surprisingly not in the C standard, I think it's because newline func (in C) is just homemade every time.

On a sidenote (a rant, really), it surprises me how every single one of these functions is frustrating. In the apple (swift) docs case it is missing `\x1c`, `\x1d`, `\x1e` which is ASCII yet they support unicode sequences, I would need to run that function to see if it is even implemented. But I have no idea if these 3 bytes are so often ignored because of being legacy :man_shrugging: 

(windows documentation doesn't even list which sequences are newlines)

So overall I'm trying to follow a nonexistent/set a standard, and I think keeping names similar will help with that

stdlib/src/utils/string_slice.mojo
```diff
@@ -301,14 +262,12 @@ struct StringSlice[is_mutable: Bool, //, origin: Origin[is_mutable].type,](
         #   StringLiteral is guaranteed to use UTF-8 encoding.
         # FIXME(MSTDL-160):
         #   Ensure StringLiteral _actually_ always uses UTF-8 encoding.
-        # TODO(#933): use when llvm intrinsics can be used at compile time
+        # FIXME: this gets practically stuck at compile time
```
JoeLoser, line 265: **Suggestion** Can you please file a new GitHub issue for the compiler team to look into and extend @lattner's work for running the llvm intrinsics at comp time?  Thanks!

stdlib/src/utils/string_slice.mojo
```diff
@@ -905,55 +881,107 @@ struct StringSlice[is_mutable: Bool, //, origin: Origin[is_mutable].type,](
...
+                if not _is_newline_char(s):
+                    return False
+            return self.byte_length() != 0
+
+    fn splitlines[
+        O: ImmutableOrigin, //
+    ](self: StringSlice[O], keepends: Bool = False) -> List[StringSlice[O]]:
         """Split the string at line boundaries. This corresponds to Python's
-        [universal newlines](
+        [universal newlines:](
```
JoeLoser, line 939: **Question** Did you mean to have the `:` in the hyperlink?

stdlib/src/utils/string_slice.mojo
```diff
@@ -987,6 +1015,77 @@ trait Stringlike:
...
+        og_ptr = unsafe_ptr_fn(i_ptr[i])
+        memcpy(p, og_ptr, og_len)
+        p[og_len] = 0  # null terminator
+        buf = String._buffer_type(unsafe_pointer=p, size=f_len, capacity=f_len)
+        (out_ptr + i).init_pointee_move(String(buf^))
+    return List[String](unsafe_pointer=out_ptr, size=i_len, capacity=i_len)
+
+
+@always_inline
+fn to_string_list[
```
JoeLoser, line 1040: **Question** Even though `String` and `StringLiteral` are using this, do we want this to be a public API?  I'd prefer to keep it private for now.

stdlib/test/builtin/test_string_literal.mojo
```diff
@@ -378,63 +378,41 @@ def test_split():
...
-    assert_equal(res10[1], "world\f")
-    assert_equal(res10[2], "mojo\x1c")
-    assert_equal(res10[3], "language\x1d")
+    assert_equal("".splitlines(), L())
+    # test \v \f \x1c \x1d
+    s2 = "hello\vworld\fmojo\x1clanguage\x1d"
+    assert_equal(s2.splitlines(), hello_mojo)
+    assert_equal(
+        s2.splitlines(keepends=True),
+        L("hello\v", "world\f", "mojo\x1c", "language\x1d"),
```
JoeLoser, line 406: **Question** Is this output expected?  Notably, in Python this would be

```
['hello\x0b', 'world\x0c', 'mojo\x1c', 'language\x1d']
```

JoeLoser (commented):
Generally looks good - left a few comments.  Thanks for continuing to work on improving all-things-String! 🔥

stdlib/src/utils/string_slice.mojo
martinvuyk: Nope, it renders like this:
image

stdlib/src/utils/string_slice.mojo
martinvuyk: Ok, I'll change it to private, though I think this will be useful since nobody is going to bother writing such a helper using `UnsafePointer` and doing manual `(out_ptr + i).init_pointee_move(String(buf^))` and `memcpy`

stdlib/src/utils/string_slice.mojo
martinvuyk: opened issue #3706

stdlib/test/builtin/test_string_literal.mojo
martinvuyk: Interesting observation, but it turns out they are equivalent since the escaped values are converted to hex:
```python
res = "hello\vworld\fmojo\x1clanguage\x1d".splitlines(keepends=True)
expected_res = ["hello\v", "world\f", "mojo\x1c", "language\x1d"]
print(res == expected_res)
print([r[-1] for r in res])
```
output
```text
True
['\x0b', '\x0c', '\x1c', '\x1d']
```

martinvuyk:
@JoeLoser I added a script to run the benchmarks sequentially with pure mojo without lit since there is a problem with it linking the benchmark module. I'm not sure if the benchmark results above are correct since I don't know if this is linking correctly to the new version I built on this branch

I edited the benchmark code to import from the package name
```mojo
from stdlib.collections.string import String
from stdlib.utils._utf8_validation import _is_valid_utf8
```
and the script `stdlib/scripts/run-benchmarks-sequentially.sh` is identical to the original but to run does this:
```bash
# Run the benchmarks
for f in $(find $BENCHMARK_PATH | grep -E ".mojo|.🔥"); do 
  echo "-------------------- BENCHMARK START ${f} --------------------"
  mojo run -I build/ $f
  echo "-------------------- BENCHMARK END ${f} --------------------"
done
```

JoeLoser:
!sync

martinvuyk:
@JoeLoser I'm making some better benchmark code (and it turns out my code is actually a regression in perf), give me some time before merging I still want to work on this and be sure about the numbers

stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -16,8 +16,9 @@ from benchmark import Bench, BenchConfig, Bencher, BenchId, Unit, keep, run
 from random import random_si64, seed
 from pathlib import _dir_of_current_file
 from collections import Optional
-from utils._utf8_validation import _is_valid_utf8
 from os import abort
+from stdlib.collections.string import String
```
JoeLoser, line 20: I've reverted the explicit `stdlib.` in the import internally - it's not needed.  When you say `from collections.string import String` for example, the compiler implicitly looks within the `stdlib` mojopkg first in the compiler's name lookup.  In other words, all imports implicitly check stdlib first, priority wise essentially.

stdlib/scripts/run-benchmarks-sequentially.sh
```diff
@@ -0,0 +1,39 @@
...
+BENCHMARK_PATH="${REPO_ROOT}/stdlib/benchmarks"
+if [[ $# -gt 0 ]]; then
+  # If an argument is provided, use it as the specific file or directory.
+  BENCHMARK_PATH=$1
+fi
+
+# Run the benchmarks
+for f in $(find $BENCHMARK_PATH | grep -E ".mojo|.🔥"); do 
+  echo "-------------------- BENCHMARK START ${f} --------------------"
+  mojo run -I build/ $f
```
JoeLoser, line 37: This is fine to JIT them (i.e. what `mojo run` does), did you verify whether `-I build` will pick up your changes in local `stdlib.mojopkg`?  I think this may "mostly work" modulo the fact that the `benchmark.mojopkg` will still be the one picked up from your installation/not your local build and so _that package_ won't take into account your recent modifications to the library.  This is inevitable right now since the source is not shipped for the `benchmark` module, and thus cannot be re-compiled on an external developer's machine using the new `stdlib` sources.  In most cases, this will be "fine" except for breaking changes (e.g. an interface changed and you're effectively mixing ABIs at that point via different mojopkgs).

martinvuyk:
@JoeLoser I'm getting some very weird results:

(commenting out all other benchmarks except `bench_splitlines`)
improvement metric: markdown percentage improvement (old_value - new_value)/old_value

Measurement 1: 
1. checkout nightly branch `upstream/nightly`, paste the string benchmark from this branch on there and run `magic run mojo run ./stdlib/benchmarks/collections/bench_string.mojo`.
2. delete changes to nightly and checkout current branch. Run `magic run ./stdlib/scripts/run-benchmarks-sequentially.sh ./stdlib/benchmarks/collections/bench_string.mojo`
3. Calculate the improvement, I get average around -2%

Measurement 2: 
1. checkout current branch. Run `magic run mojo run ./stdlib/benchmarks/collections/bench_string.mojo`
2. checkout current branch. Run `magic run ./stdlib/scripts/run-benchmarks-sequentially.sh ./stdlib/benchmarks/collections/bench_string.mojo`
3. Calculate the markdown improvement, I get average around +14%

JoeLoser:
!sync

JoeLoser:
> @JoeLoser I'm getting some very weird results:
> 
> (commenting out all other benchmarks except `bench_splitlines`)
> 
> Measurement methodology:
> 
> 1. checkout nightly branch `upstream/nightly` and package the stdlib
> 2. checkout the current branch
> 3. Run `magic run mojo run ./stdlib/benchmarks/collections/bench_string.mojo`.
> 4. Run `magic run ./stdlib/scripts/run-benchmarks-sequentially.sh ./stdlib/benchmarks/collections/bench_string.mojo`
> 5. Calculate the improvement, I get average around -1%
> 
> I've updated the PR description with the numbers. But looking at the code and what I changed, it makes no sense to me that this would cause a regression.

Before we dive into specifics, the `bench_string` doesn't even successfully compile at the moment due to

```
var res = _is_valid_utf8(items.unsafe_ptr(), length)
```

on line 214.  I'd recommend fixing that before we dive into the oddities you point out.

martinvuyk:
@JoeLoser added the change, but I can't compile it since the linker uses the existing stdlib instead of the packaged one. I tried using `-I ./build/` to try to take that as stdlib but it doesn't work :man_shrugging:

martinvuyk:
@JoeLoser I kind of fixed the config issue for the lit benchmark, it was still trying to import from the old way of doing things instead of the magic env.

It now resolves the path perfectly but it still fails.

So what I did to be able to run the benchmarks for now is to just copy the contents of the magic env into my build directory. But when I run the `run-benchmarks.sh` it blazes through the benchmark and shows no results and a message as if it were a test.

I tried following the trail of lit's config class but it leads nowhere, raw and dynamic and I have no clue what each attribute does. There is no documentation nor types in the code and it is very hard to follow. I'm getting very annoyed at this point.

For some reason the benchmark functions aren't running at all

In the benchmark main:
```mojo
print("start showing results")
for k_v in results.items():
    print(k_v[].key, k_v[].value[0], sep=",")
raise Error("show the damn results")
```
output:
```text
Exit Code: 1

Command Output (stdout):
--
start showing results
Unhandled exception caught during execution: show the damn results
```

JoeLoser:
> @JoeLoser I kind of fixed the config issue for the lit benchmark, it was still trying to import from the old way of doing things instead of the magic env. It now resolves the path perfectly.
> 
> But when I run the `run-benchmarks.sh` it blazes through the benchmark and shows no results and a message as if it were a test.
> 
> I tried following the trail of lit's config class but it leads nowhere, raw and dynamic and I have no clue what each attribute does. There is no documentation nor types in the code and it is very hard to follow.
> 
> I added a raise at the end to show results but for some reason the benchmark functions aren't running at all:
> 
> In the benchmark main:
> 
> ```mojo
> print("results:")
> for k_v in results.items():
>     print(k_v[].key, k_v[].value[0], sep=",")
> raise Error("what")
> ```
> 
> output:
> 
> ```
> Exit Code: 1
> 
> Command Output (stdout):
> --
> results:
> Unhandled exception caught during execution: what
> ```

It's because when you're running it through `lit`, it's just doing compilation essentially because it'll process the `RUN` directive in `bench_string.mojo`:

```
# RUN: %mojo-no-debug %s -t
```

The `-t` is interpreted by the benchmark module which is "test mode" or something of the like (compile, no execution).  If you just run the file using `mojo` directly, without the `-t`, it will actually execute the functions/benchmark it.  Does that make sense?  This is done this way because internally in CI, we don't run the stdlib benchmarks on _every PR_ (we have a scheduled benchmarking with bisection tools, etc.).

martinvuyk:
Got it, they are now running. I'll update the PR with the numbers later. I'll also add a comment bellow that line indicating that you need to remove the `-t` to benchmark changes on a branch, maybe also to the readme.

JoeLoser:
!sync

Issue #1310, opened 2023-11-19
[BUG]: Mojo permits the use of any constructor for implicit conversions


soraros:
### Bug description

As title. Maybe it's working as intended, but I consider it a bug, for Mojo is not JavaScript. Is the behaviour there to circumvent the traitless-ness of `print`?

### Steps to reproduce

I think the following code shouldn't type check.

```mojo
fn main():
  let s: String = 1
```

### System information

```shell
Mojo 0.5.0 on Docker, Intel Mac
```

guidorice:
Not sure it's implicit, I think this is just how the `stdlib` String is implemented. It has a constructor that takes an int, and a dozen of other constructors.

> Constructs a string representing an integer value.
https://docs.modular.com/mojo/stdlib/builtin/string.html#init__

guidorice:
If you try the opposite, you can generate a compiler error:

`let n: Int = "42"`

> cannot implicitly convert 'StringLiteral' value to 'Int' in 'let' initializer

guidorice:
I was thinking that it is simply because String has a constructor that takes an Int value- however , now I am not so sure about how implicit conversions really work, and could not find in the programming manual where this is explained. Good question 👍🏽

helehex:
I like it, but an `@explicit` decorator would be nice

helehex:
also, if you want to control this in the mean time, I've been wrapping the argument in a single tuple and it usually works

ivellapillil:
Implicit conversions are going to be headscratchers in the future. At least marking a constructor as @implicit would document that such a conversation is taking place.

helehex:
@nonmaterializable is even impliciter

helehex:
i wonder if we'll ever get the implicitest conversion

lattner:
Agree, we need to make implicit conversions be opt-in with an `@implicit` sort of decorator or something

guidorice:
@soraros I think that's not so much about implicit conversions. 

"The overload resolution logic filters for candidates according to the following rules, in order of precedence" seems well documented here: https://docs.modular.com/mojo/programming-manual.html#overloading-on-parameters

guidorice:
It does see counter-intuitive,  that the numeric types don't win in that overload. But the docs say the precedence is:

> 1. Candidates with the minimal number of implicit conversions (in both arguments and parameters).

JoeLoser:
[...]

PR #3698, opened 2024-10-21
[stdlib] Make `StringRef` ctor from `UnsafePointer[Byte]` keyword only

soraros:
This issues was reported by `@aurelian`(@diocletiann) on Discord.
This prevents accidental `UnsafePointer[Byte]` to `String` implicit conversion through `String` ctor from `StringRef`.

@JoeLoser Should we make the constructor from pointer+length keyword only as well, following `String`?

### DIFF
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -956,7 +956,7 @@ struct String(
             buff: The buffer. This should have an existing terminator.
         """
 
-        return String(buff, len(StringRef(buff)) + 1)
+        return String(buff, len(StringRef(ptr=buff)) + 1)
 
     @staticmethod
     fn _from_bytes(owned buff: Self._buffer_type) -> String:
--- a/stdlib/src/os/env.mojo
+++ b/stdlib/src/os/env.mojo
@@ -74,4 +74,4 @@ fn getenv(name: String, default: String = "") -> String:
     var ptr = external_call["getenv", UnsafePointer[UInt8]](name.unsafe_ptr())
     if not ptr:
         return default
-    return String(StringRef(ptr))
+    return String(StringRef(ptr=ptr))
--- a/stdlib/src/pathlib/path.mojo
+++ b/stdlib/src/pathlib/path.mojo
@@ -46,7 +46,7 @@ fn cwd() raises -> Path:
     if res == UnsafePointer[c_char]():
         raise Error("unable to query the current directory")
 
-    return String(StringRef(buf))
+    return String(StringRef(ptr=buf))
 
 
 @always_inline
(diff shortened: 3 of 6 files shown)


gryznar:
I am curious why implicit conversions are present in Mojo at all? It seems that it is source of a lot of bugs

soraros:
@gryznar Even though I'm not a fan of implicit conversions, they seem to be necessary to model things like literals and mimic Python behaviour in the dynamic portion. However, I do agree that we should be able to control them, and only opt-in to implicit conversion when creating a constructor (#1310).

JoeLoser:
> @JoeLoser Should we make the constructor from pointer+length keyword only as well, following String?

Can you elaborate here?  For `String`, it's implicit I think - see [this constructor](https://github.com/modularml/mojo/blob/92e2230a832ecda26dfb9132f7b8d970a7a8aafc/stdlib/src/collections/string.mojo#L821).

soraros:
@JoeLoser I was always under the impression that it worked this way. I must have mixed it up with another struct.

That said, the unsafe constructors from pointers are really a mess, and I find myself having to look up the keyword names every time. Could we adopt a convention like `fn __init__(inout self, *, ptr: UnsafePointer, len: Int)`?

- `String`: `fn __init__(inout self, ptr: UnsafePointer[UInt8], len: Int):`
- `StringRef`: `fn __init__(inout self, ptr: UnsafePointer[c_char], len: Int):`
- `StringSlice`: ` fn __init__(inout self, *, unsafe_from_utf8_ptr: UnsafePointer[UInt8], len: Int):`
- `Span`: `fn __init__(inout self, *, unsafe_ptr: UnsafePointer[T], len: Int):`
- `List`: `fn __init__(inout self, *, unsafe_pointer: UnsafePointer[T], size: Int, capacity: Int):`

JoeLoser:
> @JoeLoser I was always under the impression that it worked this way. I must have mixed it up with another struct.
> 
> That said, the unsafe constructors from pointers are really a mess, and I find myself having to look up the keyword names every time. Could we adopt a convention like `fn __init__(inout self, *, ptr: UnsafePointer, len: Int)`?
> 
> * `String`: `fn __init__(inout self, ptr: UnsafePointer[UInt8], len: Int):`
> * `StringRef`: `fn __init__(inout self, ptr: UnsafePointer[c_char], len: Int):`
> * `StringSlice`: ` fn __init__(inout self, *, unsafe_from_utf8_ptr: UnsafePointer[UInt8], len: Int):`
> * `Span`: `fn __init__(inout self, *, unsafe_ptr: UnsafePointer[T], len: Int):`
> * `List`: `fn __init__(inout self, *, unsafe_pointer: UnsafePointer[T], size: Int, capacity: Int):`

Agreed, they're wildly inconsistent.  +1 to your proposed convention - feel free to open a PR.  I'm also +1 for making them all (including `String` kw-only for the ptr + len constructor until we fix implicit constructor issue you linked).

soraros:
@JoeLoser I think this one is ready. The CI failure is caused by some Conda issues.

JoeLoser:
!sync

JoeLoser (approved):

JoeLoser:
Looks good still, we'll just need to fix up our internal usage before this lands/merges internally.  I'm on PTO tomorrow, but can look on Monday likely.

PR #3700, opened 2024-10-22
[stdlib] Fix iterator ergonomics

martinvuyk:
Fix iterator ergonomics. Add `iter() -> Iterator` and `next[T](Iterator[T]) -> T` for most common iterators, add missing `reversed() -> Iterator[forward=False]` for Stringlike types


soraros:
I think the name is dictated by the compiler, thus can't be changed.

martinvuyk:
Hi @soraros yep, was just about to @JoeLoser . IMO it needs to be changed as it deviates from other such dunders e.g. `is_not` and overall snake case conventions. I can revert the change if deemed unnecessary :man_shrugging:

JoeLoser:
> Hi @soraros yep, was just about to @JoeLoser . IMO it needs to be changed as it deviates from other such dunders e.g. `is_not` and overall snake case conventions. I can revert the change if deemed unnecessary 🤷‍♂️

Yeah, its name is coupled in the compiler, so we'd need to change it both in the library and compiler. I'm not wedded to the name that @Mogball chose here, but I'll let him chime in as well. CC: @ConnorGray in case you have opinions too.

Mogball:
This function is temporary anyways, so the name doesn't really matter

JoeLoser (commented):
Instead of making everything implement dunder bool, did you consider changing `iter` function to just directly call `self.__hasmore__()` in the `debug_assert` usage?

martinvuyk:
If `__hasmore__` is to be temporary then I think it's all the more a reason to have every Iterator implement a `__bool__` that returns whether it has more items. It seems a bit more idiomatic to me (weird to even write that for such a young language but you get my meaning).

I'm not so sure about implicit conversion and such, but to me this looks prettier:
```mojo
fn example_optional_wrapper[O: ImmutableOrigin]() -> Optional[StringSlice[O]]:
    iterator = iter("hello world")
    return next(iterator) if iterator else None
```

gryznar:
IMO `iterator.__has_more__()` much better shows the intent than `bool(iterator)`

martinvuyk:
@gryznar does this seem readable to you?
```mojo
    fn read(
        self, start: UInt = 0
    ) -> (UInt, JsonType, Self._Sp, UInt) as output:
        """Read the next `JsonType` **skipping over whitespace**.

        Args:
            start: The absolute offset in bytes to start reading a valid
                `JsonType`.

        Returns:
            The absolute offset, its `JsonType`, the `Span` containing it, and
            the start of the next item (does not check bounds).
        """

        alias `{` = UInt8(ord("{"))
        alias `}` = UInt8(ord("}"))
        alias `[` = UInt8(ord("["))
        alias `]` = UInt8(ord("]"))
        alias `,` = UInt8(ord(","))
        alias `"` = UInt8(ord('"'))
        alias ` ` = UInt8(ord(" "))
        alias `\n` = UInt8(ord("\n"))
        alias `\t` = UInt8(ord("\t"))
        alias `\r` = UInt8(ord("\r"))

        invalid = start, JsonType(JsonType.invalid), self._buffer, start

        ptr = self._buffer.unsafe_ptr()
        debug_assert(
            start <= len(self._buffer), "start is bigger than buffer length"
        )
        iterator = StringSlice[origin](
            unsafe_from_utf8_ptr=ptr + start, len=len(self._buffer) - start
        ).__iter__()
        iter_len = len(iterator)
        if iter_len == 0:
            output = invalid
            return
        char = iterator.__next__()
        while (
            char.byte_length() > 0
            and Self.isspace(char.unsafe_ptr()[0])
            and iterator.__hasmore__()
        ):
            char = iterator.__next__()
        if not char:
            output = invalid
            return
        char_p = char.unsafe_ptr()
        b0_char = char_p[0]
        if b0_char == `{`:
            if not iterator.__hasmore__():
                output = invalid
                return
            char = iterator.__next__()
            char_p = char.unsafe_ptr()
            b0_char = char_p[0]
            if Self.isspace(b0_char):
                if not iterator.__hasmore__():
                    output = invalid
                    return
                char = iterator.__next__()
                char_p = char.unsafe_ptr()
                b0_char = char_p[0]

            if b0_char == `}`:
                ...  # TODO: object end
            elif b0_char == `,`:
                ...  # TODO: next item
        elif b0_char == `[`:
            if not iterator.__hasmore__():
                output = invalid
                return
            char = iterator.__next__()
            char_p = char.unsafe_ptr()
            b0_char = char_p[0]
            if Self.isspace(b0_char):
                if not iterator.__hasmore__():
                    output = invalid
                    return
                char = iterator.__next__()
                char_p = char.unsafe_ptr()
                b0_char = char_p[0]

            if b0_char == `]`:
                ...  # TODO: array end
            elif b0_char == `,`:
                ...  # TODO: next item
        elif b0_char == `"`:
            ...  # TODO: string
        # TODO: value
        # TODO: escaped_value
        # TODO: parametrized support for NaN
        # TODO: parametrized support for Infinity
        # TODO: parametrized support for -Infinity
        output = invalid
```
yes this very much something I threw together without much thought and can do capturing functions to prettify some of the stuff, but it is still very ugly even adding `iter` and `next` builtins

gryznar:
@martinvuyk if in case above most code operates on dunders, it is more consistent to do the same with `__has_more__`

martinvuyk:
This PR is exactly about the ergonomics of iterator related operations... dunders aren't meant to be used directly.

gryznar:
Yeah, sure, but this is for Modular team to make decision here :) Your proposition `bool(iterator)` deviates from Python and that's why this is not the best approach for me:
image

JoeLoser:
!sync

JoeLoser:
From some feedback internally: can you please split this PR into smaller chunks to make it easier to review closely internally?  Thank you!

For example, one starting PR could be adding dunder `bool` to the types that delegates to the `has_next` calls.  I'm not convinced yet we want to do that, but it would allow that part of the (broader) set of changes to be discussed independently.

martinvuyk:
@JoeLoser done, removed the bool dunders. Once/if we have default trait implementations we can circle back to this as it's literally 2 extra lines of code (the dunders could've just been deleted then :man_shrugging: ). The rest is just `next` and `iterator` implementations for each specific iterator, most of the PR is just docstrings.

stdlib/src/builtin/reversed.mojo
```diff
@@ -173,3 +164,21 @@ fn reversed[
     return _DictEntryIter[K, V, dict_origin, False](
         src[]._reserved() - 1, 0, src
     )
+
+
+@always_inline
+fn reversed[
```
JoeLoser, line 170: **Suggestion** Should we add some unit tests to ensure this works as expected?

martinvuyk: Modified the `String`, `StringLiteral`, and `StringSlice` tests to use the new builtins

martinvuyk:
@JoeLoser gentle ping, and also to ask if you want me to split this PR. IMHO it makes no difference since this barely changes any functionality, it's just wrappers and docstrings. But if it will be faster for you guys just let me know and I'll do it.

PR #3714, opened 2024-10-26
[stdlib] Add `_ConcatStr` tool to concatenate Writable items lazily

martinvuyk:
Add `_ConcatStr` tool to concatenate `Writable` items lazily.


martinvuyk:
This will be indirectly replaced by #3738. Might revive this if I find a way to concatenate/store `VariadicPack`s

PR #3716, opened 2024-10-26
[stdlib] Add a bunch of new features to object (dict, tuple, `__iter__`, `structs`!)

rd4com:
Hello, it would be really nice to use our object a lot more,
it is the default type in def functions and makes programming easy,
hope this pr is good, it has a dict implementation for both K and V of object,
really looking forward to use the features of the pr everyday:

- simplify a lot by removing structs that wrapped other structs
    uses Arc in in variant for ref count (list, dict, attributes, string)
- add dict and tuple with ref count ❤️‍🔥
- implement __repr__ for `RefCountDict` and `RefCountTuple`
- add __hash__ for Self
- implement __eq__ for `RefCountAttrsDict`, `RefCountDict` and `RefCountTuple`
- add `_ObjectImpl.ref_count()` for `RefCounted*` structs
- add __contains__ for list, dict and tuple
- add `WrappedStruct`
- add pop for list and dict
- add __len__ for dict and tuple
- add int("1") conversion to int
- str(x) returns "value" instead of  " 'value' "
- add object(List(Attr("values",[1, 2])))

Can anybody check a benchmark for the matmul example ?
because changes might be slower or faster (not tested)

With this that now works:
```mojo
    a = object.dict()
    a["one"] = 1
    a[2] = "two"
    assert_equal(a["one"], 1)
    assert_equal(a[2], "two")
```
We can then implement type(x) is type(y) !
json could also be an exciting next step,
then we could have a hybrid between dynamic and typed:
`a.typed[Int]() += SIMD[Int, 4]().reduce_add()`

Hope you like this pr, if it's not good enough no worries lol 
(no merge :+1:  or we could even refine with huge reviews if needed) 

usecase is to transition from 
`def event(inout self, props: PythonObject)`
to
`def event(inout self, props)`
in a web-framework i'm trying to create


stdlib/src/builtin/object.mojo
```diff
@@ -1086,7 +1033,10 @@ struct object(
             True if the object is less than the right hard argument.
         """
         if self._value.is_str() and rhs._value.is_str():
-            return self._string_compare(rhs) < 0
+            return self._value.get_as_string().__lt__(
+                rhs._value.get_as_string()
+            )
```
martinvuyk, line 1038:
```suggestion
            return self._value.get_as_string() < rhs._value.get_as_string()
```

stdlib/src/builtin/object.mojo
```diff
@@ -1109,7 +1059,10 @@ struct object(
             True if the object is less than or equal to the right hard argument.
         """
         if self._value.is_str() and rhs._value.is_str():
-            return self._string_compare(rhs) <= 0
+            return self._value.get_as_string().__le__(
+                rhs._value.get_as_string()
+            )
```
martinvuyk, line 1064:
```suggestion
            return self._value.get_as_string() <= rhs._value.get_as_string()
```

stdlib/src/builtin/object.mojo
```diff
@@ -1131,20 +1084,25 @@ struct object(
...
+            if self._value.is_list() and rhs._value.is_list():
+                return self._list_compare(rhs) == 0
+
+            @always_inline
+            fn bool_fn(lhs: Bool, rhs: Bool) -> Bool:
+                return lhs == rhs
+
+            return Self._comparison_op[Float64.__eq__, Int64.__eq__, bool_fn](
+                self, rhs
+            ).__bool__()
```
martinvuyk, line 1099:
```suggestion
            c = Self._comparison_op[Float64.__eq__, Int64.__eq__, bool_fn](self, rhs)
            return bool(c)
```

stdlib/src/builtin/object.mojo
```diff
@@ -1154,18 +1112,12 @@ struct object(
...
-
-        @always_inline
-        fn bool_fn(lhs: Bool, rhs: Bool) -> Bool:
-            return lhs != rhs
-
-        return Self._comparison_op[Float64.__ne__, Int64.__ne__, bool_fn](
-            self, rhs
-        )
+        try:
+            return bool(not self.__eq__(rhs))
```
martinvuyk, line 1116:
```suggestion
            return not (self == rhs)
```

stdlib/src/builtin/object.mojo
```diff
@@ -1793,10 +1735,12 @@ struct object(
             or dictionary value.
         """
         if self._value.is_str():
-            return self._value.get_as_string().length
+            return self._value.get_as_string().__len__()
```
martinvuyk, line 1738:
```suggestion
            return len(self._value.get_as_string())
```

stdlib/src/builtin/object.mojo
```diff
@@ -1979,3 +1926,113 @@ struct object(
...
+        if "two" in a:
+            print("two is in a")
+        ```
+
+        Returns:
+            A `Bool` (`True` or `False`).
+        """
+        if self._value.is_list():
+            for v in self._value.get_as_list().impl[]:
+                if object(v[]).__eq__(value):
```
martinvuyk, line 1948:
```suggestion
                if object(v[]) == value:
```

stdlib/src/builtin/object.mojo
```diff
@@ -1979,3 +1926,113 @@ struct object(
...
+                    return True
+            return False
+        if self._value.is_dict():
+            return self._value.get_as_dict().impl[].__contains__(value)
```
martinvuyk, line 1952:
```suggestion
            return value in self._value.get_as_dict().impl[]
```

stdlib/src/builtin/object.mojo
```diff
@@ -1979,3 +1926,113 @@ struct object(
...
+        x["two"] = 2
+        y = x.pop("two")
+        print(y == 2)
+        ```
+        Note: implemented for list and dictionary.
+        """
+        # TODO: argument with None as default for pop()
+        if self._value.is_list():
+            if value._value.is_int():
+                tmp_i = value._value.get_as_int().__int__()
```
martinvuyk, line 1977:
```suggestion
                tmp_i = int(value._value.get_as_int())
```

martinvuyk (commented):
Hi @rd4com this looks good. Left some nitpicks on the usages of dunders that can be changed to builtin functions.

stdlib/src/builtin/object.mojo
```diff
@@ -293,52 +213,22 @@ struct _ObjectImpl(
...
     alias list: Int = 5
     """Type discriminator indicating a list."""
-    alias dict: Int = 8  # TODO
+    alias dict: Int = 8
     """Type discriminator indicating a dictionary."""
     alias function: Int = 6
     """Type discriminator indicating a function."""
     alias obj: Int = 7
     """Type discriminator indicating an object."""
+    alias tuple: Int = 8
```
JoeLoser, line 222: **Question** Why is it allowed to have `dict` and `tuple` to have the same value for the type discriminator?  Is that intended?

stdlib/src/builtin/object.mojo
```diff
@@ -490,9 +382,26 @@ struct _ObjectImpl(
...
+        if self.is_dict():
+            return "dict"
+        if self.is_tuple():
+            return "tuple"
         debug_assert(self.is_obj(), "expected a generic object")
         return "obj"
 
+    def ref_count(ref [_]self) -> Int:
+        if self.is_dict():
+            return self.value[_RefCountedDict].impl.count().__int__()
```
JoeLoser, line 394: **Suggestion** Any reason not to use `int(...)` rather than calling the dunder directly?

stdlib/src/builtin/object.mojo
```diff
@@ -1178,15 +1105,11 @@ struct object(
             True if the left hand value is greater.
```
JoeLoser, line 1105: **Question** Since you changed `eq` and `ne` to return `Bool`, should we have `gt` to return `Bool` too?  Then `object` could model `Comparable` trait, for example.

stdlib/src/builtin/object.mojo
```diff
@@ -1790,13 +1706,19 @@ struct object(
 
         Returns:
             The length of the string value or the number of elements in the list
-            or dictionary value.
+            ,dictionary or tuple value.
```
JoeLoser, line 1709:
```suggestion
            , dictionary or tuple value.
```

stdlib/src/builtin/object.mojo
rd4com: Nope (just a habit), fixed with next commit :+1:

stdlib/src/builtin/object.mojo
rd4com: Nope, it's fixed with next commit :+1: 
(we should also be able to get this from variant and simplify by following the variant.Ts order)

stdlib/src/builtin/object.mojo
rd4com: Yep it is a good idea,
would you mind if we leave this open for another day ? 
because it requires doing something like list_compare for tuple
(it is using a `List` too :+1: )
(need to think about how to make list compare be usable by both)

rd4com:
Hello @JoeLoser , last review is implemented !
Note that with this draft that uses *ref*, users can do:
```mojo
my_list._value.get_as_list()[x._value.get_as_int()]._value.get_as_list()[y._value.get_as_int()] = z
#List.__getitem__[0]().__getitem__[1]() = z
```
It is super important, because the lifetime/origin is propagated through,
list, dict, tuple, attrdict, are implemented as List[object],Dict.., so we can enjoy a ref to an object !
We might need our arc to share the PythonObject ref_count so objects can live in both world at the same time,
and the mojo side would be able to use a ref to an arc innervalue :+1: 
(ping to @ConnorGray )

Last commit also have `__iter__` for both list and dict keys ❤️‍🔥
(it works by storing the iterator in a variant)

stdlib/src/builtin/object.mojo
rd4com: Implemented in last commit :+1: ,
but keep in mind that this could not be good, because if the user don't specify an `__eq__` attribute,
we would need a way to raise an error

rd4com:
Hello, closing the PR,
it was great to work on such a cool thing   :+1: ,
in the future, with new features,
we'll probably have the best thing ever anyway :partying_face:

Issue #3563, opened 2024-09-29
[Feature Request] Use `@doc_private` on initializers from MLIR types to hide them from the API docs


soraros:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

As title. Suggested by @lattner on Discord.

More concretely, we should change
```mojo
fn __init__(inout self, value: Self._mlir_type): ...
```
to
```mojo
fn __init__(inout self, *, _mlir: Self._mlir_type): ...
```

### What is your motivation for this change?

So we can hide them in docgen or maybe intellisense to avoid, well, this:
<img width="855" alt="Screenshot 2024-09-29 at 06 16 12" src="https://github.com/user-attachments/assets/0dd48f0b-e5d2-4ebe-8fa2-1bbd3198bc75">

### Any other details?

We lose implicit conversion from MLIR values this way, I don't know how important that is.

lattner:
+1 thank you for filing this! This would be a great improvement to make MLIR stuff an implementation detail.  We should also look for the `var value : Self._mlir_type` pattern and upgrade it to `_value` for similar reasons.

soraros:
@lattner Actually, we already have a decorator [`@doc_private`](https://github.com/modularml/mojo/blob/nightly/stdlib/src/builtin/_documentation.mojo) to do this, let's just put them on all the MLIR constructors.

I don't think making the arguments keyword only is that good though. Having implicit conversion in this case can be quite useful.

lattner:
Oh nice, I didn't realize we had `@doc_private`.  I agree, that seems like a good solution!

JoeLoser:
Yeah, we should more liberally apply `@doc_private` for things like this.

JoeLoser:
A simple grep like `git grep 'mlir' stdlib | grep -i 'fn __init__' should find all the ones to go chase down if anyone is interested.

miguelcsx:
Hi @JoeLoser , I started working on this issue, could you assign it to me? thanks :rocket:

JoeLoser:
> Hi @JoeLoser , I started working on this issue, could you assign it to me? thanks 🚀

Done, thanks for starting on this!

JoeLoser:
I think this is done everywhere, but we can always reopen if there's more.

PR #3717, opened 2024-10-27
[stdlib] Rename `UInt.value` to `UInt._value`.

helehex:
Part of making mlir fields private. There was probably a better way to do all this.

### DIFF
--- a/stdlib/src/builtin/int.mojo
+++ b/stdlib/src/builtin/int.mojo
@@ -268,7 +268,7 @@ fn int(value: UInt) -> Int:
     Returns:
         The integral representation of the value.
     """
-    return value.value
+    return value._value
 
 
 # ===----------------------------------------------------------------------=== #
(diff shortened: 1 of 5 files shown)


soraros:
Maybe mentioning that it's part of #3563. Can we also introduce the `_mlir_type` alias or do you prefer doing that in a separate PR?

stdlib/src/builtin/int.mojo
```diff
@@ -410,7 +410,7 @@ struct Int(
         Args:
             value: The init value.
         """
-        self = Self(value.value)
+        self = Self(value._value)
```
soraros, line 413:
```suggestion
        self._value = value._value
```

NathanSWard:
`Uint` recently has been removed and replaced with an alias for `SIMD` so now the underlying value it the "private" field of the `SIMD` struct :) 
Closing this PR in light of this change.

PR #3724, opened 2024-10-29
[tests] Remove COMPATIBLE_COMPILER_VERSION

Ahajha:
This will be replaced with lockfiles soon. This will be merged after that change.

I have horrifically broken git. Going to scrap my 3 line change and try again.

PR #3727, opened 2024-10-31
[CI] Use lockfiles in CI

Ahajha:
(Re-attempt at #3724)

Lockfiles are now committed to the repo, no need for the `COMPATIBLE_COMPILER_VERSION` file anymore. The check for the contents of the file has already been deleted.


JoeLoser (approved):
LGTM but you'll need to also git rm COMPATIBLE_COMPILER_VERSION internally if you haven't already. Otherwise, it will reappear in the next outbound nightly sync due to it being managed by Copybara.

Ahajha:
`COMPATIBLE_COMPILER_VERSION` isn't in the internal repo, it's exclusive to this one, and ignored by copybara. Regardless, if it does show back up it doesn't hurt anything.

PR #3729, opened 2024-11-01
[stdlib] Deque - part 2 - add operator dunders and remaining traits

avitkauskas:
This is a second part of Deque implementation adding the operator dunders and all remaining traits.

### DIFF
--- a/stdlib/src/collections/deque.mojo
+++ b/stdlib/src/collections/deque.mojo
@@ -758,5 +758,5 @@ struct _DequeIter[
             return self.index
 
     @always_inline
-    fn __hasmore__(self) -> Bool:
+    fn __has_next__(self) -> Bool:
         return self.__len__() > 0
--- a/stdlib/src/collections/deque.mojo
+++ b/stdlib/src/collections/deque.mojo
@@ -207,6 +207,157 @@ struct Deque[ElementType: CollectionElement](
         self._data.free()
 
     # ===-------------------------------------------------------------------===#
+    # Operator dunders
+    # ===-------------------------------------------------------------------===#
+
+    fn __add__(self, other: Self) -> Self:
+        """Concatenates self with other and returns the result as a new deque.
+
+        Args:
+            other: Deque whose elements will be appended to the elements of self.
+
+        Returns:
+            The newly created deque with the properties of `self`.
+        """
+        new = Self(other=self)
+        for element in other:
+            new.append(element[])
+        return new^
+
+    fn __iadd__(inout self, other: Self):
+        """Appends the elements of other deque into self.
+
+        Args:
+            other: Deque whose elements will be appended to self.
+        """
+        for element in other:
+            self.append(element[])
+
+    fn __mul__(self, n: Int) -> Self:
+        """Concatenates `n` deques of `self` and returns a new deque.
+
+        Args:
+            n: The multiplier number.
+
+        Returns:
+            The new deque.
+        """
+        if n <= 0:
+            return Self(
+                capacity=self._min_capacity,
+                min_capacity=self._min_capacity,
+                maxlen=self._maxlen,
+                shrink=self._shrink,
+            )
+        new = Self(other=self)
+        for _ in range(n - 1):
+            for element in self:
+                new.append(element[])
+        return new^
+
+    fn __imul__(inout self, n: Int):
+        """Concatenates self `n` times in place.
+
+        Args:
+            n: The multiplier number.
+        """
+        if n <= 0:
+            self.clear()
+            return
+
+        orig = Self(other=self)
+        for _ in range(n - 1):
+            for element in orig:
+                self.append(element[])
+
+    fn __eq__[
+        EqualityElementType: EqualityComparableCollectionElement, //
+    ](
+        self: Deque[EqualityElementType], other: Deque[EqualityElementType]
+    ) -> Bool:
+        """Checks if two deques are equal.
+
+        Parameters:
+            EqualityElementType: The type of the elements in the deque.
+                Must implement the trait `EqualityComparableCollectionElement`.
+
+        Args:
+            other: The deque to compare with.
+
+        Returns:
+            `True` if the deques are equal, `False` otherwise.
+        """
+        if len(self) != len(other):
+            return False
+
+        for i in range(len(self)):
+            offset_self = self._physical_index(self._head + i)
+            offset_other = other._physical_index(other._head + i)
+            if (self._data + offset_self)[] != (other._data + offset_other)[]:
+                return False
+        return True
+
+    fn __ne__[
+        EqualityElementType: EqualityComparableCollectionElement, //
+    ](
+        self: Deque[EqualityElementType], other: Deque[EqualityElementType]
+    ) -> Bool:
+        """Checks if two deques are not equal.
+
+        Parameters:
+            EqualityElementType: The type of the elements in the deque.
+                Must implement the trait `EqualityComparableCollectionElement`.
+
+        Args:
+            other: The deque to compare with.
+
+        Returns:
+            `True` if the deques are not equal, `False` otherwise.
+        """
+        return not (self == other)
+
+    fn __contains__[
+        EqualityElementType: EqualityComparableCollectionElement, //
+    ](self: Deque[EqualityElementType], value: EqualityElementType) -> Bool:
+        """Verify if a given value is present in the deque.
+
+        Parameters:
+            EqualityElementType: The type of the elements in the deque.
+                Must implement the trait `EqualityComparableCollectionElement`.
+
+        Args:
+            value: The value to find.
+
+        Returns:
+            True if the value is contained in the deque, False otherwise.
+        """
+        for i in range(len(self)):
+            offset = self._physical_index(self._head + i)
+            if (self._data + offset)[] == value:
+                return True
+        return False
+
+    fn __iter__(
+        ref [_]self,
+    ) -> _DequeIter[ElementType, __origin_of(self)]:
+        """Iterates over elements of the deque, returning the references.
+
+        Returns:
+            An iterator of the references to the deque elements.
+        """
+        return _DequeIter(0, Pointer.address_of(self))
+
+    fn __reversed__(
+        ref [_]self,
+    ) -> _DequeIter[ElementType, __origin_of(self), False]:
+        """Iterate backwards over the deque, returning the references.
+
+        Returns:
+            A reversed iterator of the references to the deque elements.
+        """
+        return _DequeIter[forward=False](len(self), Pointer.address_of(self))
+
+    # ===-------------------------------------------------------------------===#
     # Trait implementations
     # ===-------------------------------------------------------------------===#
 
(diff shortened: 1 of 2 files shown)


stdlib/test/collections/test_deque.mojo
```diff
@@ -355,6 +554,112 @@ fn test_getitem() raises:
...
+fn test_setitem() raises:
+    q = Deque(1, 2)
+    assert_equal(q[0], 1)
+
+    q[0] = 3
+    assert_equal(q[0], 3)
+
+    q[-1] = 4
+    assert_equal(q[1], 4)
+
```
JoeLoser, line 566: **Question** It may be good to add a test for "index normalization" that we do in other collections, or explicitly note that we don't do this for deque.  What do you think?  Concretely, this would be adding a test case for a negative index (not `-1`), e.g. `-2` in this case, or `-3`, etc.

stdlib/test/collections/test_deque.mojo
```diff
@@ -355,6 +554,112 @@ fn test_getitem() raises:
...
+fn test_ne() raises:
+    q = Deque[Int](1, 2, 3)
+    p = Deque[Int](3, 2, 1)
+
+    assert_true(q != p)
+
+    q.appendleft(0)
+    p.append(0)
+    assert_true(q != p)
+
```
JoeLoser, line 588: **Question** What's the behavior/do you mind adding a test for when the element types differ but the values are the same?  Does that even compile currently?  E.g.

```
q = Deque[Int](1, 2, 3)
p = Deque[UInt](1, 2, 3)
assert_true(q != p) # what's the behavior here?
```

JoeLoser (commented):
Looks great, thank you for continuing this work!  I left a few minor comments, but happy to see this moving along.  When they're addressed, happy to sync this internally and land it.

Feel free to start adding a changelog entry if you'd like for your `deque` work, or would you prefer until it's fully finished from your perspective?

stdlib/src/collections/deque.mojo
```diff
@@ -206,6 +206,157 @@ struct Deque[ElementType: CollectionElement](
...
+        return new^
+
+    fn __iadd__(inout self, other: Self):
+        """Appends the elements of other deque into self.
+
+        Args:
+            other: Deque whose elements will be appended to self.
+        """
+        for element in other:
+            self.append(element[])
```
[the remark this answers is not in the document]
JoeLoser: Makes sense, I'm good with that 👍

stdlib/src/collections/deque.mojo
```diff
@@ -274,6 +506,37 @@ struct Deque[ElementType: CollectionElement](
         if self._head == self._tail:
             self._realloc(self._capacity << 1)
 
+    fn appendleft(inout self, owned value: ElementType):
```
gryznar, line 509: Why it is called `appendleft` and not `appendleft`? Mojo style guide points that snake_case should be used :)

stdlib/src/collections/deque.mojo
```diff
@@ -306,6 +569,40 @@ struct Deque[ElementType: CollectionElement](
             (src + i).move_pointee_into(self._data + self._tail)
             self._tail = self._physical_index(self._tail + 1)
 
+    fn extendleft(inout self, owned values: List[ElementType]):
```
gryznar, line 572: Same as above

stdlib/src/collections/deque.mojo
avitkauskas: This is how it named in Python: https://docs.python.org/3/library/collections.html#collections.deque
Do we want to follow Python API, or should we follow the Mojo guides?
Rust and C++ call it `push_front` and `push_back`, so lots of choices if we do not want to follow Python :)

stdlib/test/collections/test_deque.mojo
avitkauskas: Can you please clarify this. We do not want to allow indexes out of range in Deque (Python raises IndexError in these cases). I use `debug_assert()` for that in the code (same as I see it used in the List implementation). Though I do not fully understand how this should work in Mojo, as the documentation says it's `noop` in production builds, then actually it means `undefined behaviour` for the indexes out of range in production builds. Not good in my point of view. Actually, if I try this code
```python
    q = Deque(1, 2)
    e = q[-3]
    print(e)
```
with `mojo run file.mojo` it gives `5359310824`, and if I try `mojo run -g err.mojo` it gives me `1`. Why?
Why does the debug_assert not fail in debug compile with `-g`?
```
debug_assert(
   -len(self) <= normalized_idx < len(self),
    "index: ",
    normalized_idx,
    " is out of bounds for `Deque` of size: ",
    len(self),
)
```
In nightly tests ` _ = q[-3]` does fail, but I cannot test it with `assert_raises()`.

So, I can add a test case with `-2` if it helps, but not with `-3` as `debug_assert()` should 'somehow' save me from these out of range cases, but how? So what should be the desired behaviour in stdlib anyway?

stdlib/test/collections/test_deque.mojo
avitkauskas: No, it does not compile. Neither this compiles with `List`.
```
err.mojo:7:13: error: invalid call to '__eq__': could not deduce positional-only parameter #1 of callee '__eq__'
    print(q == p)
          ~~^~~~
err.mojo:7:16: note: failed to infer parameter #1, parameter inferred to two different values: 'Int' and 'UInt'
    print(q == p)
               ^
deque.mojo:267:8: note: function declared here
        EqualityElementType: EqualityComparableCollectionElement, //
       ^
```
As both arguments of `__eq__` are defined being of type `EqualityElementType`, I suppose they are expected to be of exactly the same type.

stdlib/src/collections/deque.mojo
```diff
@@ -253,6 +404,87 @@ struct Deque[ElementType: CollectionElement](
...
+        The elements' type must implement the `__repr__()` method for this to work.
+
+        Parameters:
+            RepresentableElementType: The type of the elements in the deque.
+                Must implement the trait `RepresentableCollectionElement`.
+
+        Returns:
+            A string representation of the deque.
+        """
+        output = String()
```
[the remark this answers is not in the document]
avitkauskas: This does not compile. `String` does not have a `write` method in my Mojo API documentation :)
My implementation follows the way it is done in `List` in nighly now.

avitkauskas:
Please give me feedback on my comments to move forward. I cannot make any changes without your further guidance right now. Conserning the changelog entry, I will add it together with the last part that should follow after this PR.

stdlib/src/collections/deque.mojo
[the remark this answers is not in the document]
avitkauskas: These are the errors I get if I try `return String.write(self)`:
```
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/deque.mojo:456:28: error: no matching function in call to 'write'
        return String.write(self)
               ~~~~~~~~~~~~^~~~~~
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/string.mojo:854:8: note: candidate not viable: argument #0 must be mutable in order to pass to a mutating argument
    fn write[*Ts: Writable](inout self, *args: *Ts):
       ^
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/string.mojo:872:8: note: candidate not viable: could not deduce parameter 'Ts' of callee 'write'
    fn write[
       ^
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/deque.mojo:456:29: note: failed to infer parameter 'Ts', argument type 'Deque[RepresentableElementType]' does not conform to trait 'Writable'
        return String.write(self)
                            ^~~~
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/string.mojo:908:8: note: candidate not viable: failed to infer implicit parameter 'elt_is_mutable' of argument 'args' type 'VariadicPack'
    fn write[
       ^
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/deque.mojo:456:28: note: failed to infer parameter #1, parameter isn't used in any argument
        return String.write(self)
               ~~~~~~~~~~~~^~~~~~
```

stdlib/src/collections/deque.mojo
```diff
@@ -253,6 +404,87 @@ struct Deque[ElementType: CollectionElement](
...
+        The elements' type must implement the `__repr__()` for this to work.
+
+        Parameters:
+            RepresentableElementType: The type of the elements in the deque.
+                Must implement the trait `RepresentableCollectionElement`.
+
+        Returns:
+            A string representation of the deque.
+        """
+        return self.__str__()
```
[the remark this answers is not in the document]
avitkauskas: These are the errors if I try `return str(self)`:
```
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/deque.mojo:487:19: error: no matching function in call to 'str'
        return str(self)
               ~~~^~~~~~
Included from /Users/alvis/repos/mojo/stdlib/src/builtin/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/builtin/str.mojo:134:4: note: candidate not viable: could not deduce parameter 'T' of callee 'str'
fn str[T: Stringable](value: T) -> String:
   ^
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/deque.mojo:487:20: note: failed to infer parameter 'T', argument type 'Deque[RepresentableElementType]' does not conform to trait 'Stringable'
        return str(self)
                   ^~~~
Included from /Users/alvis/repos/mojo/stdlib/src/builtin/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/builtin/str.mojo:150:4: note: candidate not viable: argument #0 cannot be converted from 'Deque[RepresentableElementType]' to 'None'
fn str(value: None) -> String:
   ^
Included from /Users/alvis/repos/mojo/stdlib/src/builtin/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/builtin/str.mojo:163:4: note: candidate not viable: could not deduce parameter 'T' of callee 'str'
fn str[T: StringableRaising](value: T) raises -> String:
   ^
Included from /Users/alvis/repos/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/alvis/repos/mojo/stdlib/src/collections/deque.mojo:487:20: note: failed to infer parameter 'T', argument type 'Deque[RepresentableElementType]' does not conform to trait 'StringableRaising'
        return str(self)
                   ^~~~
```

stdlib/test/collections/test_deque.mojo
JoeLoser: Great, that's exactly what I expected to happen - thanks for confirming.

JoeLoser:
!sync

stdlib/src/collections/deque.mojo
JoeLoser: Let's follow the Python APIs for the time being and we can discuss this in our next team design meeting (CC: @ConnorGray).

stdlib/test/collections/test_deque.mojo
JoeLoser: Unfortunately, the name `debug_assert` has nothing to do with "debug mode".  To control the behavior of `debug_assert`, you'd need to specify `-D ASSERT=all` for example to enable assertions.

Right now, the testing for various `debug_assert` isn't ideal, but doable - see https://github.com/modularml/mojo/blob/ce76c9e42165bd2099a8ca5f461d7090b9e5ef9c/stdlib/test/builtin/test_range_out_of_bounds.mojo#L36 for example (note the test is using the `not --crash` in its `RUN` directive).

More broadly speaking, what our plan is with index normalization for various collections is something we need to figure out, so sticking with what you have is totally fine.  Right now, due to the viralness of raises, we don't want to raise for out of bound accesses like in Python.  Otherwise, it pushes all downstream callers to be in a potentially-raising context which won't be nice in the current state of Mojo.

avitkauskas: I tried to test the index out of range (to include it in the next PR part).
I included these lines in the test file:
```
# REQUIRES: has_not
# RUN: not --crash mojo -D ASSERT=all %s 2>&1
```
But I got an error:
```
FAIL: Mojo Standard Library :: collections/test_deque.mojo (1 of 1)
******************** TEST 'Mojo Standard Library :: collections/test_deque.mojo' FAILED ********************
Exit Code: 1

Command Output (stderr):
--
RUN: at line 14: not --crash mojo -D ASSERT=all /Users/alvis/repos/mojo/stdlib/test/collections/test_deque.mojo 2>&1
+ not --crash mojo -D ASSERT=all /Users/alvis/repos/mojo/stdlib/test/collections/test_deque.mojo
--
```
What else am I missing?

PR #3734, opened 2024-11-03
[stdlib] Fix allow immutable `UnsafePointer` and adjust `.unsafe_ptr()` methods

martinvuyk:
Fix allow immutable `UnsafePointer` and adjust `.unsafe_ptr()` methods.

### DIFF
--- a/stdlib/src/builtin/string_literal.mojo
+++ b/stdlib/src/builtin/string_literal.mojo
@@ -23,6 +23,7 @@ from hashlib._hasher import _HashableWithHasher, _Hasher
 from utils import StringRef, Span, StringSlice, StaticString
 from utils import Writable, Writer
 from utils._visualizers import lldb_formatter_wrapping_type
+from builtin.builtin_list import _lit_mut_cast
 
 from utils.string_slice import (
     _StringSliceIter,
@@ -407,23 +408,29 @@ struct StringLiteral(
         return __mlir_op.`pop.string.size`(self.value)
 
     @always_inline("nodebug")
-    # FIXME(MSTDL-956): This should return a pointer with StaticConstantOrigin.
-    fn unsafe_ptr(self) -> UnsafePointer[UInt8]:
+    fn unsafe_ptr[
+        is_mutable: Bool = False,
+        origin: Origin[is_mutable]
+        .type = _lit_mut_cast[StaticConstantOrigin, is_mutable]
+        .result,
+    ](self) -> UnsafePointer[Byte, is_mutable=is_mutable, origin=origin]:
         """Get raw pointer to the underlying data.
 
         Returns:
             The raw pointer to the data.
         """
+        constrained[not is_mutable, "StringLiteral can't be mutated"]()
         var ptr = UnsafePointer(__mlir_op.`pop.string.address`(self.value))
 
         # TODO(MSTDL-555):
         #   Remove bitcast after changing pop.string.address
         #   return type.
-        return ptr.bitcast[UInt8]()
+        return ptr.bitcast[Byte, is_mutable=is_mutable, origin=origin]()
 
     @always_inline
-    # FIXME(MSTDL-956): This should return a pointer with StaticConstantOrigin.
-    fn unsafe_cstr_ptr(self) -> UnsafePointer[c_char]:
+    fn unsafe_cstr_ptr(
+        self,
+    ) -> UnsafePointer[c_char, is_mutable=False, origin=StaticConstantOrigin]:
         """Retrieves a C-string-compatible pointer to the underlying memory.
 
         The returned pointer is guaranteed to be NUL terminated, and not null.
--- a/stdlib/src/memory/unsafe_pointer.mojo
+++ b/stdlib/src/memory/unsafe_pointer.mojo
@@ -28,7 +28,7 @@ from sys.intrinsics import (
     strided_load,
     strided_store,
 )
-
+from builtin.builtin_list import _lit_mut_cast
 from bit import is_power_of_two
 from memory.memory import _free, _malloc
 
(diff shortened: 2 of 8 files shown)


martinvuyk:
@JoeLoser All the tests are passing but the packaging of `test_utils` is giving problems, the linker keeps grabbing nightly stdlib even though the script sets `MODULAR_MOJO_NIGHTLY_IMPORT_PATH=$BUILD_DIR` and I've also tried adding `mojo package "${TEST_UTILS_PATH}" -I ${BUILD_DIR}`.

stdlib/src/builtin/builtin_list.mojo
```diff
@@ -290,6 +290,12 @@ struct _lit_mut_cast[
     ]
 
 
+struct _is_mutable_origin[
```
JoeLoser, line 293: **Question** Can you help me understand why this is needed?  You should be able to directly access the `is_mutable` from the call sites without any issues.

stdlib/src/builtin/string_literal.mojo
```diff
@@ -407,23 +408,33 @@ struct StringLiteral(
         return __mlir_op.`pop.string.size`(self.value)
 
     @always_inline("nodebug")
-    # FIXME(MSTDL-956): This should return a pointer with StaticConstantOrigin.
-    fn unsafe_ptr(self) -> UnsafePointer[UInt8]:
+    fn unsafe_ptr[
+        is_mutable: Bool = False,
+        origin: Origin[is_mutable]
+        .type = _lit_mut_cast[StaticConstantOrigin, is_mutable]
```
JoeLoser, line 414: Nice to see this get fixed up with respect to the right origin - thank you!

stdlib/src/builtin/string_literal.mojo
```diff
@@ -407,23 +408,33 @@ struct StringLiteral(
...
         """Get raw pointer to the underlying data.
 
+        Parameters:
+            is_mutable: Whether the origin is mutable.
+            origin: The origin of the data.
+
         Returns:
             The raw pointer to the data.
         """
+        constrained[not is_mutable, "StringLiteral can't be mutated"]()
```
JoeLoser, line 426: **Question** Given that `is_mutable` is a parameter but it's invalid for users to pass `True` here to this method, do we need the parameter at all, or should we just hard-code `is_mutable = False` in the return type, etc like you do in `unsafe_cstr_ptr()` below?

stdlib/src/memory/unsafe_pointer.mojo
```diff
@@ -290,7 +294,7 @@ struct UnsafePointer[
     # TODO: replace with a safe model that checks the body of the method for
     # accesses to the origin.
     @__unsafe_disable_nested_origin_exclusivity
-    @always_inline("nodebug")
+    @always_inline
```
JoeLoser, line 297: **Question** Why change these?  I think we want to leave these alone as they won't benefit much to users in seeing this in a stack frame in our debugger.

stdlib/test/collections/test_string.mojo
```diff
@@ -1253,15 +1253,6 @@ def test_string_iter():
         concat += v
     assert_equal(321, atol(concat))
 
-    for v in vs:
```
JoeLoser, line 1256: **Question** Why remove this test?

JoeLoser (commented):
This is **so exciting** to see coming together! I appreciate you tackling this! 🎉

stdlib/src/builtin/string_literal.mojo
martinvuyk: Because otherwise the trait won't be adhered to 
```mojo
trait Stringlike(...):
    ...
    fn unsafe_ptr[
        is_mutable: Bool, origin: Origin[is_mutable].type
    ](self) -> UnsafePointer[Byte, is_mutable=is_mutable, origin=origin]:
    ...
```

stdlib/test/collections/test_string.mojo
martinvuyk: It was actually a test that shouldn't be possible. The compiler won't let you do this now
```mojo
for v in vs: # immutable ref iterator
    v.unsafe_ptr()[] = ord("1") # this is  mutating the origin
```

stdlib/src/builtin/builtin_list.mojo
martinvuyk: Not on type signatures where all you have is `__origin_of()` and you want to extract the mutability of that:
```mojo
    @staticmethod
    @always_inline("nodebug")
    fn address_of(
        ref [_, address_space._value.value]arg: type
    ) -> UnsafePointer[
        type,
        address_space,
        is_mutable = _is_mutable_origin[__origin_of(arg)].result,
        origin = __origin_of(arg),
    ] as result:
        """Gets the address of the argument.

        Args:
            arg: The value to get the address of.

        Returns:
            An UnsafePointer which contains the address of the argument.
        """
        return __type_of(result)(
            __mlir_op.`lit.ref.to_pointer`(__get_mvalue_as_litref(arg))
        )
```
Though if there is a cleaner way to extract an attribute from `__origin_of()` I'm all ears

stdlib/src/memory/unsafe_pointer.mojo
```diff
@@ -624,6 +628,7 @@ struct UnsafePointer[
             offset: The offset to store to.
             val: The value to store.
         """
+        constrained[is_mutable, _must_be_mut_err]()
```
soraros, line 631: Can we turn these into conditional conformance?

martinvuyk: Theoretically the compiler won't even let you mutate the origin. I added them for convenience so that users get a clear error message. Conditional Conformance doesn't give you that, only a cryptic error that some parameters can't be inferred.

JoeLoser:
!sync

JoeLoser:
@martinvuyk quick update here — @ConnorGray has been pushing a lot this week on several improvements to origins and working with them.  He's been iterating on your imported PR and it's passing CI internally right now.

We're going to land this work _after_ the branch cut for our stable release which is coming very soon to avoid any potential mishaps with exclusivity checking and not putting ourself in a halfway spot of origin handling in the stdlib.  But there's nothing needed from you at this point, and we really appreciate you pushing on the initial exploration of this work!  It's been super helpful and we look forward to landing your work soon and continuing to push origins throughout the library 🎉 🚀

martinvuyk:
Hi @JoeLoser thanks for keeping me updated, and thanks @ConnorGray for working on it :D

> We're going to land this work after the branch cut for our stable release which is coming very soon to avoid any potential mishaps with exclusivity checking and not putting ourself in a halfway spot of origin handling in the stdlib

That sounds wise :+1:

PR #3735, opened 2024-11-03
[stdlib] Add `os.path.expandvars` and respective tests

thatstoasty:
This PR adds the `expandvars` function to the `os.path` module.

The documentation for contributing references installing `lit` into our local Python environment, but with `magic` we have a Python environment already defined, so I added it as a dependency in the `pixi.toml` file. On a similar note, I added `magic run` to pre-commit hooks so `magic` is used for the Python and Mojo environments.

Any feedback on the changes or on the implementation would be welcome!


magic.lock
```diff
@@ -119,6 +119,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libxcrypt-4.4.36-hd590300_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libxml2-2.13.4-h064dc61_2.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libzlib-1.3.1-hb9d3cd8_2.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/lit-19.1.3-pyhd8ed1ab_0.conda
```
JoeLoser, line 122: @Ahajha can you comment here how we're managing the magic lock files here internally wrt this?  Given the migration to magic, I agree with @thatstoasty that we could just specify this as part of the dependencies in the `pixi.toml`.

stdlib/src/os/path/path.mojo
```diff
@@ -390,3 +391,156 @@ def split[PathLike: os.PathLike, //](path: PathLike) -> (String, String):
...
+    var bytes = path_str.as_bytes()
+    var buf = String()
+
+    # Byte scanning should be fine, ${} is ASCII.
+    i = 0
+    j = 0
+    while j < len(bytes):
+        if bytes[j] == ord("$") and j + 1 < len(bytes):
+            if not buf:
+                buf._buffer.reserve(new_capacity=2 * len(bytes))
```
JoeLoser, line 516: **Suggestion** This sort of suggests the public API of `String` ought to have a "reserve" like functionality to avoid needing to access the internal `_buffer` data member.  Do you mind filing a GitHub issue for this?

stdlib/test/os/path/test_expandvars.mojo
```diff
@@ -0,0 +1,80 @@
...
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo %s
+
+import os
+from os.path import expandvars
+from testing import assert_equal
+
+
+def test_expansion():
+    _ = os.setenv("TEST_VAR", "World")
```
JoeLoser, line 21: **Suggestion** It may be nice to unset these variables explicitly at the end of each function to avoid state carrying over to the next set of tests within this file.  I think it would fall under the "element of least surprise" sort of thing - WDYT?  You could also introduce a local context manager whos `__exit__` function unsets the env variable if that's any cleaner.

JoeLoser (approved):
Great addition, @toiletsandpaper.  LGTM, appreciate the easy-to-read test cases 👍

JoeLoser:
!sync

stdlib/src/os/path/path.mojo
```diff
@@ -390,3 +391,156 @@ def split[PathLike: os.PathLike, //](path: PathLike) -> (String, String):
...
+        ord("2"),
+        ord("3"),
+        ord("4"),
+        ord("5"),
+        ord("6"),
+        ord("7"),
+        ord("8"),
+        ord("9"),
+    )
+    if int(byte) in shell_variables:
```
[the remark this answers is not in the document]
thatstoasty: Yeah definitely, I'll make that change! Slipped my mind that contains is already returning a Boolean.

stdlib/src/os/path/path.mojo
thatstoasty: I initially started out with a `List[Byte]` buffer and constructed a `String` from that at the end. But, writing directly to a `String` seemed more straightforward to understand. I think my need here would be solved with some sort of StringBuilder or being able to call `write` or `write_bytes` on `List[Byte]`. It could be handy to be able to allocate capacity ahead of time, I'll open up a Github issue and see the consensus is.

In the meantime, should I avoid using the internal `_buffer` for the time being?

stdlib/test/os/path/test_expandvars.mojo
thatstoasty: Done 👍🏾

We didn't have `os.unsetenv` yet, so I added that function with a test. If it's preferable to wrap up the `external_call` part of that function and move that over into `sys._libc` I can do that as well.

magic.lock
Ahajha: I think this should be fine. 2 points:
1: This would allow a follow-up cleanup where we could remove the section of CI that installs `lit` through llvm.sh.
2: Internally, the lockfiles are wiped and then recreated each night (rather than updated in-place), so the only reason to update them here is to satisfy the `--frozen` flag, as `pixi.toml` is essentially ignored in CI here. (There might be a different flag that might be better for this use case, where it will check both?) After a night this should sync internally + regenerate, and then everything should be fine.

Verbal approval, from the CI side of things this should be a-ok.

@thatstoasty Thanks for improving the state of the world :)

thatstoasty: Great! I handle a lot of the CI for my teams at work, so I know how fickle internal processes can be 🙂. Glad to see that the change might make its way downstream to make contributing a little easier!

stdlib/src/os/env.mojo
```diff
@@ -51,6 +51,26 @@ fn setenv(name: String, value: String, overwrite: Bool = True) -> Bool:
     return status == 0
 
 
+fn unsetenv(name: String) -> Bool:
+    """Unsets an environment variable.
+
+    Constraints:
+        The function only works on macOS or Linux and returns False otherwise.
```
JoeLoser, line 58: **Question** Any particular reason not to just use `constrained` instead (hard error in other words) instead of just gracefully returning `False` on Windows?

stdlib/src/os/env.mojo
```diff
@@ -51,6 +51,26 @@ fn setenv(name: String, value: String, overwrite: Bool = True) -> Bool:
...
+        name: The name of the environment variable.
+
+    Returns:
+        True if unsetting the variable succeeded. Otherwise, False is returned.
+    """
+    alias os_is_supported = os_is_linux() or os_is_macos()
+    if not os_is_supported:
+        return False
+
+    var status = external_call["unsetenv", Int32](name.unsafe_ptr())
```
JoeLoser, line 70:
```suggestion
    var status = external_call["unsetenv", c_int](name.unsafe_ptr())
```

from `sys.ffi` to be explicit (we'd like to be consistent about doing this with other `external_call`s).

stdlib/src/os/path/path.mojo
JoeLoser: Thanks for the GH issue!  You can keep using `_buffer` for now as there are other users we'd need to fix up in the future too.  Appreciate it.

JoeLoser (approved):
LGTM! Just a minor remaining question/comment and then we can land this!

stdlib/src/os/env.mojo
thatstoasty: No concerns here, I was following the same pattern as `getenv` and `setenv`. I'll open up a separate PR to make them consistent with using `constrained` and `c_int`. For now I've updated only `unsetenv`.

stdlib/src/os/env.mojo
thatstoasty: Updated!

stdlib/src/os/path/path.mojo
```diff
@@ -390,3 +391,154 @@ def split[PathLike: os.PathLike, //](path: PathLike) -> (String, String):
...
+        ord("2"),
+        ord("3"),
+        ord("4"),
+        ord("5"),
+        ord("6"),
+        ord("7"),
+        ord("8"),
+        ord("9"),
+    )
+    return int(byte) in shell_variables
```
martinvuyk, line 429:
```suggestion
    alias `0` = Byte(ord("0"))
    alias `9` = Byte(ord("9"))
    alias shell_variables = SIMD[DType.uint8, 8](
        Byte(ord("!")),
        Byte(ord("!")),
        Byte(ord("#")),
        Byte(ord("$")),
        Byte(ord("*"),
        Byte(ord("-")),
        Byte(ord("?")),
        Byte(ord("@")),
    )
    return `0` <= byte <= `9` or byte in shell_variables
```

martinvuyk (commented):
Hi @thatstoasty nice work. I left some ideas in case you want to apply them though some might take time and it's fine to leave them as is or do a followup (they're a bit of an overkill).

stdlib/src/os/path/path.mojo
```diff
@@ -390,3 +391,154 @@ def split[PathLike: os.PathLike, //](path: PathLike) -> (String, String):
...
+        or ord("0") <= b
+        and b <= ord("9")
+        or ord("a") <= b
+        and b <= ord("z")
+        or ord("A") <= b
+        and b <= ord("Z")
+    )
+
+
+fn _parse_variable_name(bytes: Span[Byte]) -> Tuple[String, Int]:
```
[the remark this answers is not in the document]
thatstoasty: That's a good point, we don't need to allocate a new string *every* time. Seems like it's a relatively small change as well, I'll just need to adjust the `${}` or `${` case since it was returning a new hardcoded string. I'll take a look into making this change alongside the suggestions immediately below this.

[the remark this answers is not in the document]
thatstoasty: I've combined the cases into one branch, and I switched over to returning an immutable StringSlice.

stdlib/src/os/path/path.mojo
```diff
@@ -390,3 +391,154 @@ def split[PathLike: os.PathLike, //](path: PathLike) -> (String, String):
...
+            and bytes[2] == ord("}")
+        ):
+            return String(StringSlice(unsafe_from_utf8=bytes[1:2])), 3
+
+        # Scan until the closing brace or the end of the bytes.
+        var i = 1
+        while i < len(bytes):
+            if bytes[i] == ord("}"):
+                if i == 1:
+                    return String("${}"), 2
```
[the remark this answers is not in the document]
thatstoasty: I was able to squash the cases into one:
 
 ```python
# Invalid syntax (`${}` or `${`) or $ was not followed by a name; write as is.
if name.startswith("{") or name == "":
    buf.write_bytes(bytes[j : j + length + 1])
```

`if length <= 2:` did not work for one letter environment variables like `$a`.

File: stdlib/src/os/path/path.mojo (hunk 7 above)
Comment (thatstoasty) [reply stdlib/src/os/path/path.mojo:429]: I'm seeing some compilation issues using SIMD, so I'm going to leave it as it for now. `Byte(ord(...))` was adding visual noise IMO.

File: stdlib/src/os/path/path.mojo [hunk 10]
```diff
@@ -390,3 +391,154 @@ def split[PathLike: os.PathLike, //](path: PathLike) -> (String, String):
...
+    var b = int(byte)
+    return (
+        b == ord("_")
+        or ord("0") <= b
+        and b <= ord("9")
+        or ord("a") <= b
+        and b <= ord("z")
+        or ord("A") <= b
+        and b <= ord("Z")
+    )
```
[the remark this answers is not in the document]
thatstoasty: Whenever I see backtick variables, it gets a bit difficult to differentiate between a variable name and a value. So I'm going to leave it as is for now

thatstoasty:
@JoeLoser I think I'm good on additional changes to the code, unless you have some more suggestions!

JoeLoser:
!sync

PR #3736, opened 2024-11-03
[stdlib] Faster repr() logic for String, StringRef, UInt, DType and others

msaelices:
- [X] Implement `StringRef.__repr__()`
- [X] Faster `path.stat()` implementation using Writer
- [X] Faster `UInt.__repr__()`
- [X] Faster `Error.__repr__()`
- [X] Faster `DType.__repr__()`


stdlib/src/os/fstat.mojo
```diff
@@ -150,30 +150,45 @@ struct stat_result(Stringable):
...
-        res += ", st_atime=" + str(self.st_atimespec)
-        res += ", st_mtime=" + str(self.st_mtimespec)
-        res += ", st_ctime=" + str(self.st_ctimespec)
-        res += ", st_birthtime=" + str(self.st_birthtimespec)
-        res += ", st_blocks=" + str(self.st_blocks)
-        res += ", st_blksize=" + str(self.st_blksize)
-        res += ", st_rdev=" + str(self.st_rdev)
-        res += ", st_flags=" + str(self.st_flags)
-        return res + ")"
+        var output = String()
```
JoeLoser, line 189: **Suggestion** I think this can just be `return String.write(self)`

JoeLoser (commented):
Do you mind looking into the pathlib test failure as a result of this change?  See [this CI run](https://github.com/modularml/mojo/actions/runs/11655733986/job/32497162367?pr=3736) for example.

stdlib/src/os/fstat.mojo
msaelices: Thanks. Done

msaelices:
> Do you mind looking into the pathlib test failure as a result of this change? See [this CI run](https://github.com/modularml/mojo/actions/runs/11655733986/job/32497162367?pr=3736) for example.

I think the check is flaky as it depends on the OS, and I don't have the MacOS to test it. I've just removed off-topic checks

msaelices:
@JoeLoser could you please TAL?

JoeLoser:
> @JoeLoser could you please TAL?

Yep, CI is failing though - do you mind looking into that first?

stdlib/src/collections/string.mojo
```diff
@@ -686,6 +686,63 @@ fn isprintable(c: UInt8) -> Bool:
     return ord_space <= int(c) <= ord_tilde
 
 
+# ===----------------------------------------------------------------------=== #
+# String Representation
+# ===----------------------------------------------------------------------=== #
+
+
+struct StringRepr(Writable):
```
JoeLoser, line 694: **Question** Should we make this private or do you expect users to interact with this?  I'd default to private for now.

stdlib/src/os/fstat.mojo
```diff
@@ -46,7 +46,7 @@ fn _constrain_unix():
 
 
 @value
-struct stat_result(Stringable):
+struct stat_result(Stringable, Writable):
```
JoeLoser, line 49: **Suggestion** Split this change to `stat_result` to be `Writable` along with its tests to a separate PR, please.

stdlib/src/collections/string.mojo
```diff
@@ -686,6 +686,63 @@ fn isprintable(c: UInt8) -> Bool:
...
+
+        Parameters:
+            W: The type of the writer.
+
+        Args:
+            writer: The writer to write the string representation to.
+        """
+        var quote = "'"
+        if len(self._value) > 0 and self._value[0] == "'":
+            quote = '"'
```
martinvuyk, line 722: this is not how python works
```python
>>> print(repr("hi"))
'hi'
>>> print(repr("'hi'"))
"'hi'"
>>> print(repr("a 'hi' a"))
"a 'hi' a"
```

stdlib/src/collections/string.mojo
```diff
@@ -686,6 +686,63 @@ fn isprintable(c: UInt8) -> Bool:
...
+                var codepoint = ord(s)
+                if isprintable(codepoint):
+                    writer.write(s)
+                elif codepoint < 0x10:
+                    writer.write(hex(codepoint, prefix=r"\x0"))
+                elif codepoint < 0x20 or codepoint == 0x7F:
+                    writer.write(hex(codepoint, prefix=r"\x"))
+                else:  # multi-byte character
+                    writer.write(s)
+        writer.write(quote)
```
martinvuyk, line 743: This isn't really much faster. It still uses write `hex()` and doesn't alloc before doing everything

martinvuyk (changes requested):
Hi, I've been working on faster repr and ascii logic for about 3 weeks now, see PR #3686. This does not add much speed really. 

I think you can still merge the other improvements you made on this PR. But see the [_repr function](https://github.com/modularml/mojo/pull/3686/files#diff-53bade406e0c0f5616a9bbee7766e2f85b9ab8cbf1acd7441d3e52aef555465cR261) that I've been working on, there are several other things that need to exist as helpers to make this much faster.

stdlib/src/collections/string.mojo
```diff
@@ -686,6 +686,63 @@ fn isprintable(c: UInt8) -> Bool:
...
+
+    fn __init__(inout self, value: String):
+        """Constructs a string representation of the string.
+
+        Args:
+            value: The string to represent.
+        """
+        self._value = StringSlice[ImmutableAnyOrigin](
+            unsafe_from_utf8_ptr=value.unsafe_ptr(),
+            len=len(value),
```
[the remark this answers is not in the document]
msaelices: Fixed here: https://github.com/msaelices/mojo/commit/3ab5ef09411b75bb7b69e8f2e7992dc12127d447

stdlib/src/os/fstat.mojo
msaelices: The goal of this commit was to have a faster representation of `path.stat()` which uses `stat_result` underneath, which was already representable but slow, so to me was in the same scope of "Faster repr() logic" https://github.com/msaelices/mojo/commit/950d36adbe5823c5611038669e14c2a9dd00f594

Do you still want me to split those changes into a new PR?

stdlib/src/collections/string.mojo
msaelices: I don't find the issue here. This is a comparison:
image

msaelices:
@martinvuyk, @JoeLoser could you please TAL?

stdlib/src/collections/string.mojo
msaelices: Yes, it was not a great optimization. I only mapped the top-level constructor to use `.write()` but did not spend the needed work on making it way faster.

stdlib/src/collections/string.mojo
martinvuyk: The current nightly implementation does this:
```mojo
var use_dquote = False
for s in self:
    use_dquote = use_dquote or (s == "'")
```
which checks if any character in the string is `'`, and that makes the `__repr__` change to using double quotes. The logic above only checks if the first character is `'` and would fail in the 3rd example.

Not sure where the command line execution you're showing here comes from, but if it's from the results of this code then I must've misread the logic :smile:

stdlib/src/os/fstat.mojo
```diff
@@ -150,30 +150,43 @@ struct stat_result(Stringable):
...
+        writer.write(", st_size=", self.st_size)
+        writer.write(", st_atime=", str(self.st_atimespec))
+        writer.write(", st_mtime=", str(self.st_mtimespec))
+        writer.write(", st_ctime=", str(self.st_ctimespec))
+        writer.write(", st_birthtime=", str(self.st_birthtimespec))
+        writer.write(", st_blocks=", self.st_blocks)
+        writer.write(", st_blksize=", self.st_blksize)
+        writer.write(", st_rdev=", self.st_rdev)
+        writer.write(", st_flags=", self.st_flags)
+        writer.write(")")
```
martinvuyk, line 180: Suggestion: 
The `String.write_bytes()` reserves the amount of bytes to join the next item every time, in the future we should be able to constrain traits and know if they implement for example `as_bytes()` and be able to reserve the total amount of bytes beforehand. So this suggestion doesn't make a difference currently but maybe in the future: 

these can all go in the same function call separated by commas. They will be formatted similarly by the Mojo formatter so it won't make much difference in the aesthetics other than the name of the field won't be directly to the left, still it's just a matter of personal preference and not performance currently :man_shrugging: .

martinvuyk (approved):
LGTM

stdlib/src/os/fstat.mojo
JoeLoser: I understand the scope of the commit, but in the future, I'd recommend teasing apart the two ideas (applying `Writable` to a struct and making `repr()` faster for a few types).  They're orthogonal things in my mind and could have been two separate PRs.  It's fine for this one though, but something to consider in the future 😄

JoeLoser (approved):

PR #3751, opened 2024-11-06
[stdlib] Addition of a new constructor and `reserve` method to the `String` struct

thatstoasty:
Adds a public API for users to allocate capacity through String, instead of having to reach into the private internal `._buffer` attribute.

Satisfies: https://github.com/modularml/mojo/issues/3738

NOTE: Waiting for #3735 to go through so the `pixi.toml` changes make it through.

### DIFF
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -766,6 +766,15 @@ struct String(
         """Construct an uninitialized string."""
         self._buffer = Self._buffer_type()
 
+    @always_inline
+    fn __init__(inout self, *, capacity: Int):
+        """Construct an uninitialized string with the given capacity.
+
+        Args:
+            capacity: The capacity of the string.
+        """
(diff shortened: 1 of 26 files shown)


stdlib/src/collections/string.mojo
```diff
@@ -2296,6 +2305,17 @@ struct String(
...
 
+    fn reserve(inout self, new_capacity: Int):
+        """Reserves the requested capacity.
+
+        If the current capacity is greater or equal, this is a no-op.
+        Otherwise, the storage is reallocated and the date is moved.
+
+        Args:
+            new_capacity: The new capacity.
+        """
```
martinvuyk, line 2316: nit:
```suggestion
        Args:
            new_capacity: The new capacity.

        Notes:
            If the current capacity is greater or equal, this is a no-op.
            Otherwise, the storage is reallocated and the data is moved.
        """
```

thatstoasty:
Commits are borked here, opening a new PR

Issue #3710, opened 2024-10-25
[BUG] Mojo frees memory while reference to it is still in use


avitkauskas:
### Bug description

You cannot take the address of a List `data` into a variable and use it as a reference later, because Mojo frees the memory allocated to `data` by destructuring the List as it reckons it's not being used beyond the point of the last syntactic mention of the List variable.

### Steps to reproduce

```mojo
fn main():
    a = List(1, 2, 3)
    b = List[Int](capacity=3)

    p = a.data
    for i in range(len(a)):
        (p + i).move_pointee_into(b.data + i)
    b.size = 3

    # if you uncomment the line below, everything works fine
    # _ = a.size

    print(b.__str__())
    # => [4462641216, 2, 3]
```

### System information

```shell
macOS 15.0.1
mojo 2024.10.2406 (54f70ee8)
magic 0.3.1 - (based on pixi 0.29.0)
```

soraros:
This is the expected behaviour [ref](https://docs.modular.com/mojo/manual/lifecycle/death#default-destruction-behavior).

`a.data` returns a `UnsafePointer` that doesn't keep `a` alive.

avitkauskas:
So maybe ASAP destruction is not such a great idea? Maybe scope-based destruction makes better sense? Because cases like this are quite surprising IMHO.

10 days ago I was wondering about another case on Discord. And Chris Lattner noticed that and fixed the case immediatelly. Thanks to him! But I can just show you what the case was:
```mojo
from memory import UnsafePointer

struct A:
    var data: UnsafePointer[Int]

    fn __init__(inout self):
        self.data = UnsafePointer[Int].alloc(1)
        self.data.init_pointee_copy(1)

    fn __del__(owned self):
        self.data.free()
```
Create a file `test_struct.mojo`
```
from testing import assert_equal
from destruct import A

fn test_destruct() raises:
    a = A()
    assert_equal(a.data[0], 1)
    # _ = a.data
```
And test was failing unless you uncommented the last line.

The case of the current issue is a bit different, and I understand it when you say it's expected. But still, I find it surprising and inconvenient.

avitkauskas:
In this case, we could solve the problem just using `a.data` directly:
```mojo
for i in range(len(a)):
    (a.data + i).move_pointee_into(b.data + i)
```

But let's say I want to move elements from the end of `a` to the front of `b`. Then:
```mojo
scr = a.data + len(a) - 1
for i in range(len(a)):
[...]

Issue #3580, opened 2024-09-30
[BUG] `UnsafePointer` doesn't extend lifetime for long enough


soraros:
### Bug description

As title. CC @lattner. Originally reported [here](https://github.com/modularml/mojo/issues/3559#issuecomment-2381194721).

A variant of #3559. The problem persists if `print` is non-variadic.

### Steps to reproduce

```mojo
from memory import UnsafePointer

fn print_str(x: String):
  print(x)

@export
fn main():
  s = String("test")
  p = UnsafePointer.address_of(s)
  print_str(p[])  # prints nothing
```

I checked the IR, and it seems that `@llvm.lifetime.end` is still emitted before the `print_str` call.

### System information

```shell
Mojo 2024.9.3005 (882b4490) on Docker, Intel Mac
```

lattner:
This works now

PR #3754, opened 2024-11-07
[stdlib] [proposal] Add `del` builtin function

martinvuyk:
Add `del` builtin function. I don't like doing `_ = value^`, if there is not going to be support on the language side then at least this function makes the intention of deletion clear. And until we have the `take`/`consume` keyword, the user will have to explicitly transfer `^` the value into the function (AFAIK the compiler will transfer anyway if no further uses are made).

The Mojo formatter seems to not like the definition of a function name equal to the `del` Python keyword and removes parenthesis. Not sure if you guys even want this function or will add support for it in the future. But can this issue be solved downstream? (I imagine you guys just forked black formatter).

If support for this is added on the compiler side later on, it should work the same anyway. Python also accepts:
```python
>>> a, b = "asd", 123
>>> del(a, b)
>>> print(a, b)
NameError: name 'a' is not defined
```

I'll be bold and @Mogball , is support for `del` on the compiler side planned?

### DIFF
--- a/stdlib/src/base64/_b64encode.mojo
+++ b/stdlib/src/base64/_b64encode.mojo
@@ -205,7 +205,7 @@ fn store_incomplete_simd[
     ]()
 
     memcpy(dest=pointer, src=tmp_buffer_pointer, count=nb_of_elements_to_store)
-    _ = simd_vector  # We make it live long enough
+    del simd_vector^  # We make it live long enough
 
 
 # TODO: Use Span instead of List as input when Span is easier to use
--- a/stdlib/src/builtin/_pybind.mojo
+++ b/stdlib/src/builtin/_pybind.mojo
@@ -59,7 +59,7 @@ fn fail_initialization(owned err: Error) -> PythonObject:
         error_type,
         err.unsafe_cstr_ptr(),
     )
-    _ = err^
+    del err^
     return PythonObject(PyObjectPtr())
 
 
--- a/stdlib/src/builtin/anytype.mojo
+++ b/stdlib/src/builtin/anytype.mojo
@@ -65,3 +65,12 @@ trait AnyType:
         end of this function.
         """
         ...
+
+
+# TODO: this needs to take/consume the value, not copy implicitly
+fn del(owned args: VariadicPack[element_trait=AnyType]):
+    __type_of(args).__del__(args^)
+
+# TODO: this needs to take/consume the value, not copy implicitly
+fn del[*T: AnyType](owned *args: *T):
+    del(args)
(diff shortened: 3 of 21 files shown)


stdlib/src/builtin/anytype.mojo
```diff
@@ -65,3 +65,12 @@ trait AnyType:
...
         ...
+
+
+# TODO: this needs to take/consume the value, not copy implicitly
+fn del(owned args: VariadicPack[element_trait=AnyType]):
+    __type_of(args).__del__(args^)
+
+# TODO: this needs to take/consume the value, not copy implicitly
+fn del[*T: AnyType](owned *args: *T):
+    del(args)
```
lattner, line 76: Why are you passing on the variadic pack here?  you can just use "pass" and the compiler will call the dtors on all the elements for you.

stdlib/src/base64/_b64encode.mojo
```diff
@@ -205,7 +205,7 @@ fn store_incomplete_simd[
     ]()
 
     memcpy(dest=pointer, src=tmp_buffer_pointer, count=nb_of_elements_to_store)
-    _ = simd_vector  # We make it live long enough
+    del simd_vector^  # We make it live long enough
```
lattner, line 208: I don't htink this is necessary, have you tried just removing the `_ = simd_vector`?  I fixed lifetime extension of unsafe pointers.

stdlib/src/builtin/_pybind.mojo
```diff
@@ -59,7 +59,7 @@ fn fail_initialization(owned err: Error) -> PythonObject:
         error_type,
         err.unsafe_cstr_ptr(),
     )
-    _ = err^
+    del err^
```
lattner, line 62: Similarly, I don't think this manual lifetime extension is necessary, can we just remove it?

stdlib/src/collections/string.mojo
```diff
@@ -1416,7 +1416,7 @@ struct String(
             result.write(a)
 
         elems.each[add_elt]()
-        _ = is_first
+        del is_first^
```
lattner, line 1419: Some of these are needed because there are bugs in the closure handling stuff.

lattner (changes requested):

lattner:
I don't think that this is the right approach.  You can already spell this in mojo with `x.^__del__()`, but we're on a path to fix the bugs that require manual lifetime extension -- we don't want to provide syntax for the bad thing.

martinvuyk:
Hi @lattner 

ok so the approach will be full auto ASAP destruction then? (and hence no `del` keyword)

> I fixed lifetime extension of unsafe pointers.

sadly there are still some bugs remaining (#3580, #3710), though I think with PR #3734 such foot-guns will be harder to trigger

lattner:
Let's work through them, thank you for pushing on this (still getting caught up)

PR #3755, opened 2024-11-07
[stdlib] Addition of a new constructor and reserve method to the String struct

thatstoasty:
Adds a public API for users to allocate capacity through String, instead of having to reach into the private internal `._buffer` attribute.

Satisfies: https://github.com/modularml/mojo/issues/3738


thatstoasty:
@martinvuyk I made your suggested change to the docstring of the reserve function!

JoeLoser (approved):

JoeLoser:
!sync

PR #3766, opened 2024-11-11
[stdlib] Add Libc and tests

martinvuyk:
### Add Libc and tests
This will pave the way for better cross platform support.

Tests for the most used libc functionality were added. Some trivial functions were left as `# TODO` for later testing. `ioctl` was deactivated as it needs to be tested thoroughly.

On the networking side, I have functional socket code but it needs a counterpart to be used or use async code, so I left it "untested" but it already works. The next stage after this would be to bring in the socket package I've been working on for the last couple of months, but I'd rather wait for full async Mojo since many APIs would benefit from being async first.

I can keep developing this further in my repo, but I think this will benefit the stdlib greatly by centralizing the implementation and testing. And also the whole ecosystem will find it useful to have an already "cross platform" Libc wrapper. These tests also introduce a very easy way to figure out platform support problems.


JoeLoser:
Hey Martin, thanks so much for pushing on the `libc` work!  It's something the team has talked about briefly in the past (whether to include libc bindings in the stdlib or not).  I'll add it again to your team's discussion this upcoming Monday and we'll circle back here based on what we discuss.

I'm sure @lsh has some thoughts on this space too 😄

owenhilyard:
My main suggestion is to try to make the structure mirror the structure of POSIX, to make it easier to find things. Also, please remember /etc/protocols, there's a lot of useful things in there like PIM, mptcp (Linux only), IPv6 with no next header (nice for routers), kernel-managed TLS, and DCCP.

stdlib/src/sys/ffi/utils.mojo
```diff
@@ -17,8 +17,8 @@ from memory import UnsafePointer
 
 from utils import StringRef
 
-from .info import os_is_linux, os_is_windows, is_64bit, os_is_macos
-from .intrinsics import _mlirtype_is_eq
+from sys.info import os_is_linux, os_is_windows, is_64bit, os_is_macos
+from sys.intrinsics import _mlirtype_is_eq
 
 from sys._libc import dlerror, dlopen, dlclose, dlsym
```
msaelices, line 23: @martinvuyk what about the `sys._libc` module? are they going to merge at some point?

stdlib/src/sys/ffi/c/types.mojo
```diff
@@ -0,0 +1,275 @@
...
+    """Type: `void`."""
+    alias NULL = UnsafePointer[Self.void]()
+    """Constant: NULL pointer."""
+    alias ptr_addr = Int
+    """Type: A Pointer Address."""
+
+
+alias size_t = UInt
+"""Type: `size_t`."""
+alias ssize_t = Int
```
msaelices, line 69: There are a couple of similar aliases in the `sys._libc` module.

stdlib/src/sys/ffi/c/libc.mojo
```diff
@@ -0,0 +1,2498 @@
...
+        """Construct a Libc instance using the default dylib location for the
+        given OS."""
+
+        @parameter
+        if os_is_windows():
+            self._lib = DLHandle("msvcrt")
+        elif os_is_macos():
+            self._lib = DLHandle("libc.dylib")
+        else:
+            self._lib = DLHandle("libc.so.6")
```
msaelices, line 99: :1st_place_medal:

stdlib/src/sys/ffi/c/libc.mojo
```diff
@@ -0,0 +1,2498 @@
...
+            The number of bytes printed excluding the terminating null byte.
+            Otherwise a negative value and `errno` is set.
+
+        Notes:
+            [Reference](https://man7.org/linux/man-pages/man3/printf.3p.html).
+        """
+        var a = args.get_loaded_kgen_pack()
+
+        @parameter
+        if is_nvidia_gpu():
```
msaelices, line 842: Could just `is_gpu()` work?

stdlib/src/sys/ffi/utils.mojo
martinvuyk: Hi @msaelices, yes I imagine they will merge at some point. There is also some future cleanup to switch `c_int` to use `C.int` and other such variables (if they decide to merge this PR at all).

stdlib/src/sys/ffi/c/libc.mojo
martinvuyk: I saw this workaround in a stdlib function, I don't know if it's some NVIDIA particular thing or if it's that the API for GPUs needs to be different :man_shrugging:. This `vprintf` signature is actually non standard AFAIK since it doesn't use the [`va_list`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdarg.h.html) type, but it could also be that the underlying `VariadicPack` uses such a structure internally and when a Pointer traverses though it it works fine, when I tested `vprintf` on Linux with the same setup it did not work :man_shrugging:.

martinvuyk:
@JoeLoser any changes in opinion regarding this idea? I discovered cdb9f5527d67322e4f8fb0a038d009cc4b1fd927 way back when I wrote the tests for this. We could reduce the scope of this and use only what is needed for the stdlib. And I think allowing to dynamically link to libc might be interesting for a lot of folks

PR #3768, opened 2024-11-12
[stdlib] Implement `Int.from_bytes()` and `Int.as_bytes()`

msaelices:


msaelices:
Closed in favor of https://github.com/modularml/mojo/pull/3795

This PR ends up in a mess because of conflicts with the nightly branch

PR #3773, opened 2024-11-15
[stdlib] Fix unsafe `String.write_bytes()`

martinvuyk:
Fix an unsafe assumption that the span is null terminated.

### DIFF
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -853,6 +853,9 @@ struct String(
             bytes: The byte span to write to this String. Must NOT be
                 null terminated.
         """
+        # TODO: maybe check byte by byte that there is no 0 ?
+        # need Span.count() to be faster
+        debug_assert(bytes[-1] != 0, "byte Span must not be null terminated")
         self._iadd[False](bytes)
 
     fn write[*Ts: Writable](inout self, *args: *Ts):
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -847,14 +847,13 @@ struct String(
     # ===------------------------------------------------------------------=== #
 
     fn write_bytes(inout self, bytes: Span[Byte, _]):
-        """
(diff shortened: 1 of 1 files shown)


JoeLoser (commented):
Thanks for the fix!  A few questions:

1. Is there a unit test we can add to expose this bug?
2. Should we add a `debug_assert` to ensure the argument to `write_bytes` is not null terminated per the doc string?

martinvuyk:
I added a debug assert for the null termination, still there can be the case of data with a 0 somewhere in the middle but I'll leave a `# TODO` until PR #3686 gets merged which can count faster. Do you think it will be necessary to ensure no zeroes in the middle ? it might slow down the whole CI since string writing is something that gets used a lot

JoeLoser (approved):

JoeLoser:
LGTM when CI is passing

JoeLoser:
> I added a debug assert for the null termination, still there can be the case of data with a 0 somewhere in the middle but I'll leave a `# TODO` until PR #3686 gets merged which can count faster. Do you think it will be necessary to ensure no zeroes in the middle ? it might slow down the whole CI since string writing is something that gets used a lot.
> 
> > Is there a unit test we can add to expose this bug?
> 
> I don't think there is one, since this was a call to a private method

I think what you have is fine wrt the `debug_assert` safety check - thanks for adding that! Makes sense re the unit test.

martinvuyk:
> I don't think there is one, since this was a call to a private method

I mean theoretically we could, just need to find a way for the allocation to be just on the limit of a page and try to access that to get a segfault :laughing:

martinvuyk:
@JoeLoser one problem, in Python unlike C you migh actually have a string `"\x00"` which would be `List[Byte](0,  0)` and this debug assert would complain.
```python
>>> print(repr("\0"), bytes("\0", "utf8"))
'\x00' b'\x00'
```
luckily we had tests that caught this

JoeLoser:
!sync

PR #3778, opened 2024-11-17
Remove the move for `_Global`, (`T:Movable`->`T:AnyType`)

rd4com:
Hello, not sure if it is what you meant for the todo,
but here is a possible solution :+1: 
~(Note: we could also use `-> out as MyType` if needed)~ 
(Edit: new out return arg convention, no `inout` for init)

(ping to @ConnorGray )


soraros:
@rd4com Is this still relevant?

rd4com:
Hello, yes tests still green :+1:

soraros:
You might want to update the issue description then.

PR #3782, opened 2024-11-18
[stdlib] [NFC] Move Formatter structs to their own module. Add `TODO`s and observations

martinvuyk:
Move Formatter structs to their own module. Add `TODO`s and observations.

Motivation:
These implementations will get more and more complex as time goes on. They are extensive and specific enough that they should be in their own module. Eventually they should become public structs, but that remains in the future until they are feature complete.


JoeLoser (approved):
LGTM, thanks!  In the future, like you say, it may make sense to up-level the formatting things into their own top-level module to avoid cluttering up utils/it becoming even more of a kitchen sink of things.  Totally fine for now though, of course. 😄

PR #3793, opened 2024-11-21
[docs] Stdlib insider documentation

owenhilyard:
People working on the standard library need to have some more information about the API contracts and behavior of the runtime and compiler builtins in order to be able to write correct and performant code in the stdlib. This is not intended to be high effort documentation or acknowledgement that a feature is stable, but simply communication of enough information for those of us without access to the compiler source and autogenerated mlir documentation to work with them correctly.

Pinging @lattner as requested.


stdlib/docs/internal/README.md
```diff
@@ -0,0 +1,16 @@
+## WARNING
```
msaelices, line 1: @owenhilyard, not sure if "internal" is the best name for me it could be consider as a barrier to potential contributors. Even the plural form "internals" is clearer to me that we are documenting things happening underneath.

Anyways, I'm not a native English speaker, so maybe I'm misleading here.

owenhilyard: Good point. I'm trying to aggressively ward people away from using this unless they have no other choice since we want to keep the MLIR APIs and not technically stable parts of the language flexible, and I think that `internals` may not express what I want. Can I get some general feedback from readers on which of the following directory names seems like it wouldn't stop you from reading it as a prospective stdlib contributor but would stop you from using it at work (or other ideas)?

* internal
* internals
* stdlib_contributor
* unstable
* stdlib_only
* language_internals

lattner:
Awesome, thank you for putting this together Owen. 

For context, this came out of a conversation we had where there is a bunch of archana (much of which is unstable, eg mlir details) that is exposed through the stdlib.  We'd like to rationalize and rein some of it in, but it will never go away, and having some reference is useful for contributors.

@JoeLoser do you have any concerns with this?  If not, plz approve and we can merge and improve in-tree.

proposals/stdlib-insider-docs.md
```diff
@@ -0,0 +1,136 @@
...
+employees have to catch misuses. These language-internal dialects are, like
+Mojo itself, subject to enhancements, breaking changes, and even complete
+removal. This presents a problem since the stdlib is correctness-critical code,
+and when people who don't understand the API contract of a construct use it
+in correctness-critical code, issues are bound to happen.
+
+## Proposal
+
+In order to help address this, I propose the creation of a "stdlib insider"
+document, which contains information on MLIR operations/types, WIP features and
```
lattner, line 27: In addition to mlir stuff, I think it would also be great to have notes on "these are projects in progress", e.g. like the changes to implicit copyability.

owenhilyard: I agree that having some kind of work tracking for stdlib would be useful. We have a lot of in-progress work which make decent first issues (ex: converting collections to ExplicitlyCopyable) that could be better highlighted since I think some people want to contribute and then see us arguing about allocator APIs, async executors, and linear types and think that's were the bar to contribute is. It would also enable us to track what things we want to do that are blocked or functionally blocked (essentially a rewrite) on new language features, so I'll try to merge that with the list of language feature desires by the external stdlib contributors. 

@JoeLoser Would you want something like this tracked in github projects (so you can do a one-way sync to your internal work tracking) or do you want to keep it in some markdown files? If we do markdown, I think we can use something like `stdlib/docs/insider/work_tracking` to store them. The main downside of using github is likely a bunch of front-loaded work to load everything into there and some maintenance work to keep it up to date unless there's a way to give some of us permissions to add and edit projects/tasks but nothing else in the repo. I'm open to other ideas as well.

proposals/stdlib-insider-docs.md
```diff
@@ -0,0 +1,136 @@
...
+return from the current function?).
+
+### `!lit.origin.set`
+
+This is used in Coroutines (presumably to store capture origins), but it looks
+like it may be useful for storing collections of references.
+
+### The backing allocator(s) for `pop.aligned_alloc` and `pop.global_alloc`
+
+Time spent in GDB has led me to believe this is tcmalloc, but knowing for
```
JoeLoser, line 132: Yes, currently it's `tcmalloc`.

JoeLoser (approved):
Thanks for writing this up @owenhilyard and discussing this previously with @lattner.

I'm happy to approve this and I'll work on adding some external-facing documentation here on the internals.

JoeLoser:
> Awesome, thank you for putting this together Owen.
> 
> For context, this came out of a conversation we had where there is a bunch of archana (much of which is unstable, eg mlir details) that is exposed through the stdlib. We'd like to rationalize and rein some of it in, but it will never go away, and having some reference is useful for contributors.
> 
> @JoeLoser do you have any concerns with this? If not, plz approve and we can merge and improve in-tree.

I do agree now is a good time to do this.  Early on when we open-sourced the stdlib, this was a topic of conversation and we decided to not hide the internal dialects (given their heavy use at the time) but we also weren't quite ready to document things widely.

I just approved the PR and we'll work on adding docs here.  Thanks, all!

proposals/stdlib-insider-docs.md
JoeLoser: I'm +1 for updating our stdlib projects for both:

1. In-progress work (e.g. making things explicit by removing our incremental @implicit decorators we added recently).
2. Back-burnered work (perhaps things blocked by compiler bugs) — this would be things like SBO, implicit copyability (some of that may be in a much better state now with explicit constructor control though).

We'll also update the roadmap and make it clearer which areas we're looking for help in from contributors so we can work more effectively together on bigger projects.

As for the mechanism of how we'll track the work (GH projects or Markdown), let me think about that a bit more.  Earlier this year, we switched off of GitHub and over to Linear internally for tracking issues.  I'm not sure how well GitHub _projects_ will lend itself to that, so I'll need to look into how to best organize things internally/externally without duplicating too many things.

JoeLoser:
!sync

stdlib/docs/internal/README.md
JoeLoser: I don't feel strongly.  Any of them are fine and we can always easily rename the directory.

JoeLoser:
!sync

nmsmith:
Just a small note: "insider docs" sounds like some kind of exclusive club, or something nefarious, like "insider trading". "Contributor docs" would sound less hostile and exclusionary.

owenhilyard:
@nmsmith 
> Just a small note: "insider docs" sounds like some kind of exclusive club, or something nefarious, like "insider trading". "Contributor docs" would sound less hostile and exclusionary.

I'm trying to walk the line between being welcoming to new contributors and making people shy away from using this in non-stdlib projects, since anyone who builds any kind of substantial body of code on top of the docs proposed for this is likely to get stuck on old versions of the language unless they are already a stdlib contributor and are keeping up with changes to the language.

PR #3795, opened 2024-11-22
[stdlib] Implement `Int.from_bytes()` and `Int.as_bytes()`

msaelices:


stdlib/src/builtin/int.mojo
```diff
@@ -1194,6 +1195,64 @@ struct Int(
 
         writer.write(self)
 
+    @staticmethod
+    fn from_bytes[
+        type: DType, big_endian: Bool = False
```
lsh, line 1200: I'm in favor of adding a `from_bytes` and `to_bytes` method, but I'm not a fan of using a boolean to represent the endianness here. If we wanted to be Pythonic, we could use `StringLiteral` in the parameter and constrain it to be `"little"` or `"big"`, or we could make an `Endian` wrapper struct which we can parameterize other methods on in the future. WDYT @JoeLoser?

I also don't think we want to add a `DType` parameter here. We can just get the `sizeof[Int]()`, or perhaps add a helper to get the `DType` equivalent to `Int`.

lsh (changes requested):

stdlib/src/builtin/int.mojo
msaelices: I was adding these methods because they are very convenient for networking protocols. For example I am needing these for a pure Mojo implementation of websockets.

If we don't have the `DType` and `big_endian` parameters, and we get an array of 60 bytes from some stream, how can we do the following cases:

1. I want to get 10 UInt64 integers (10x6=60)
2. I want to get 10 Int64 integers
3. I want to get 30 UInt16 integers (30x2=60)
4. I want to get 30 Int16 integers

Python implementation supports both, but as arguments, it's slower as we cannot do it in comptime. But if we are implementing something IO-related, like WebSockets, we already know the endianness of the target we want to read from or write to, so it's faster if we use parameters.

lsh: If anything, these comments make me think that `from_bytes` and `to_bytes` belongs on `SIMD`, not `Int`. 

> Python implementation of int.as_bytes() have the length and byteorder arguments, and int.from_bytes() assumes we always have a 4 bytes integers and only have the byteorder one, but as arguments (run-time) which is slower and in Mojo we can do it comptime.

I agree :) that's whyi still said it should be a parameter, but my comments were questioning what the parameter should look like, because I'm not sold on `Bool`.

martinvuyk: Yeah if we follow Python's lead, `bytes` is binary `Byte` and can be encoded in `UInt16` etc. We could also hold off on doing it like Python and let this function be inferred and have `bytes: Span[Scalar[D]]`.

I think `type` shouldn't be used as prolifically as it is for parameters since it's the name of Python's `type` builtin function that I think we'll eventually have in Mojo as well.
```suggestion
        D: DType, big_endian: Bool = False
```

---
> questioning what the parameter should look like, because I'm not sold on Bool.

As for this, I personally never liked `sys.byteorder` returning `"big"` or `"little"` because there is no third option, so a boolean makes sense to me :man_shrugging:. But compatibility may be a solid argument in favor of it.

stdlib/src/builtin/int.mojo
```diff
@@ -1212,6 +1213,64 @@ struct Int(
...
+        var ptr: UnsafePointer[Byte] = UnsafePointer.address_of(bytes[0])
+        var type_ptr: UnsafePointer[Scalar[type]] = ptr.bitcast[Scalar[type]]()
+        var value = type_ptr[]
+
+        @parameter
+        if is_big_endian() and not big_endian:
+            value = byte_swap(value)
+        elif not is_big_endian() and big_endian:
+            value = byte_swap(value)
+        return int(value)
```
martinvuyk, line 1243:
```suggestion
        var ptr = bytes.unsafe_ptr().bitcast[Scalar[type]]()

        @parameter
        if is_big_endian() and not big_endian:
            return byte_swap(ptr[])
        elif not is_big_endian() and big_endian:
            return byte_swap(ptr[])
        else:
            return int(ptr[])
```

stdlib/src/builtin/int.mojo
```diff
@@ -1212,6 +1213,64 @@ struct Int(
...
+            value = byte_swap(value)
+
+        var ptr: UnsafePointer[Scalar[type]] = UnsafePointer.address_of(value)
+        var byte_ptr: UnsafePointer[Byte] = ptr.bitcast[Byte]()
+        var list = List[Byte](capacity=type_len)
+
+        # TODO: Maybe this can be a List.extend(ptr, count) method
+        memcpy(list.unsafe_ptr(), byte_ptr, type_len)
+        list.size = type_len
+
```
martinvuyk, line 1271: I don't think we should add more `UnsafePointer` APIs than absolutely necessary
```suggestion
        var ptr =  UnsafePointer.address_of(value)
        var list = List[Byte](capacity=type_len)

        memcpy(list.unsafe_ptr(), ptr.bitcast[Byte](), type_len)
        list.size = type_len
```

stdlib/test/builtin/test_int.mojo
```diff
@@ -19,6 +19,8 @@ from testing import assert_equal, assert_true, assert_false, assert_raises
 from python import PythonObject
 from memory import UnsafePointer
 
+alias Bytes = List[Byte]
```
martinvuyk, line 22: Suggestion: This definition is not decided over yet. could you take this into the scope of your test?

martinvuyk (commented):
Hi @msaelices, cool to know someone else is also pushing things forward for binary serialization.

I have two main opinions:
- I think we should let the `DType` be inferred and have the `Span` be of that `DType`. I know Python treats everything like `bytes` but I think it's something that will bite us in the long term. We have a statically typed language and we can make use of that. So IMO we should hold off on making everything revolve around `Byte` and having to `if-else` on every binary interface until we are absolutely sure we want to go that way.
- I think we can hold off on controlling endianness in the parameter until users need it.

stdlib/src/builtin/int.mojo
msaelices: We are already diverging from the Python `int.from_bytes()` and `int.to_bytes()`. As a Python developer, I personally don't mind making a slight adaptation (e.g., `big` -> `True` and `small` -> `False`). The real issue would be if there is no equivalent implementation in Mojo's stdlib; you would need to implement the logic yourself.

msaelices:
> Hi @msaelices, cool to know someone else is also pushing things forward for binary serialization.
> 
> I have two main opinions:
> 
> * I think we should let the `DType` be inferred and have the `Span` be of that `DType`. I know Python treats everything like `bytes` but I think it's something that will bite us in the long term. We have a statically typed language and we can make use of that. So IMO we should hold off on making everything revolve around `Byte` and having to `if-else` on every binary interface until we are absolutely sure we want to go that way.

Great suggestion. I did it her: https://github.com/msaelices/mojo/commit/63fa5b45a16e8e3215873e0e7c322f5cb7bfaa4e

Unfortunately, the tests do not work because this kind of errors: 
```
/home/msaelices/src/mojo/stdlib/test/builtin/test_int.mojo:252:62: error: invalid call to 'from_bytes': failed to infer implicit parameter 'is_mutable' of argument 'bytes' type 'Span'            
    assert_equal(Int.from_bytes[DType.int16, big_endian=True](Bytes(0, 16)), 16)  
```
My concern is that we will have to complicate the currently complicated signature by explicitly pass the `is_mutable` or `origin` param. Do you have any idea?


> * I think we can hold off on controlling endianness in the parameter until users need it.

martinvuyk:
> Unfortunately, the tests do not work because this kind of errors:
> 
> ```
> /home/msaelices/src/mojo/stdlib/test/builtin/test_int.mojo:252:62: error: invalid call to 'from_bytes': failed to infer implicit parameter 'is_mutable' of argument 'bytes' type 'Span'            
>     assert_equal(Int.from_bytes[DType.int16, big_endian=True](Bytes(0, 16)), 16)  
> ```
> 
> My concern is that we will have to complicate the currently complicated signature by explicitly pass the `is_mutable` or `origin` param. Do you have any idea?
> 
> >

Hmm yet another place where origins get annoying :roll_eyes:. This seems like a complete bug, it should work:
```mojo
struct Span:
    ...
    fn __init__(out self, ref [origin]list: List[T, *_]): ...
```
You could try something like forcing the casting of the origin:
```mojo
assert_equal(Int.from_bytes[DType.int16, big_endian=True](Span(Bytes(0, 16)).get_immutable()), 16)  
```
but this should definitely not be necessary.

msaelices:
> > Unfortunately, the tests do not work because this kind of errors:
> > ```
> > /home/msaelices/src/mojo/stdlib/test/builtin/test_int.mojo:252:62: error: invalid call to 'from_bytes': failed to infer implicit parameter 'is_mutable' of argument 'bytes' type 'Span'            
> >     assert_equal(Int.from_bytes[DType.int16, big_endian=True](Bytes(0, 16)), 16)  
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My concern is that we will have to complicate the currently complicated signature by explicitly pass the `is_mutable` or `origin` param. Do you have any idea?
> > >
> 
> Hmm yet another place where origins get annoying 🙄. This seems like a complete bug, it should work:
> 
> ```mojo
> struct Span:
>     ...
>     fn __init__(out self, ref [origin]list: List[T, *_]): ...
> ```
> 
> You could try something like forcing the casting of the origin:
> 
> ```mojo
> assert_equal(Int.from_bytes[DType.int16, big_endian=True](Span(Bytes(0, 16)).get_immutable()), 16)  
> ```
> 
> but this should definitely not be necessary.

Maybe this change in the compiler will help here: https://github.com/modularml/mojo/commit/3c4f57c9225d8f7634fb4df7e0bde7004f7c189b

msaelices:
> > > Unfortunately, the tests do not work because this kind of errors:
> > > ```
> > > /home/msaelices/src/mojo/stdlib/test/builtin/test_int.mojo:252:62: error: invalid call to 'from_bytes': failed to infer implicit parameter 'is_mutable' of argument 'bytes' type 'Span'            
> > >     assert_equal(Int.from_bytes[DType.int16, big_endian=True](Bytes(0, 16)), 16)  
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > My concern is that we will have to complicate the currently complicated signature by explicitly pass the `is_mutable` or `origin` param. Do you have any idea?
> > > >
> > 
> > 
> > Hmm yet another place where origins get annoying 🙄. This seems like a complete bug, it should work:
> > ```mojo
> > struct Span:
> >     ...
> >     fn __init__(out self, ref [origin]list: List[T, *_]): ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You could try something like forcing the casting of the origin:
> > ```mojo
> > assert_equal(Int.from_bytes[DType.int16, big_endian=True](Span(Bytes(0, 16)).get_immutable()), 16)  
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but this should definitely not be necessary.
> 
> Maybe this change in the compiler will help here: [3c4f57c](https://github.com/modularml/mojo/commit/3c4f57c9225d8f7634fb4df7e0bde7004f7c189b)
> 
> Will check it probably today.

I could not get this working even after the support in the compiler of setting the implicit `is_mutable` parameter :(

I am afraid that we need to revert some of the changes from [6d395a1c691000ac06fa1bcbd96305d6359d1784] make it work. Done here: https://github.com/msaelices/mojo/commit/b5e09c5c528d4b8ee8fe574f983bc44c340a62ef

stdlib/test/builtin/test_int.mojo
msaelices: Sure! Done: https://github.com/msaelices/mojo/commit/7766ea3487b6859d13d92f4e73507f3f65d5c83d

stdlib/src/builtin/int.mojo
msaelices: Agreed. Changed here: https://github.com/msaelices/mojo/commit/30027dffadc04f55616bbc49f4c111ff3f24d50e

stdlib/src/builtin/int.mojo
```diff
@@ -1212,6 +1217,65 @@ struct Int(
 
         writer.write(self)
 
+    @staticmethod
+    fn from_bytes[
```
lsh, line 1221: I don't think my comment before was ever properly addressed. I think this method should be on `SIMD`, not `Int`. That way the `DType` makes more sense. We can think more about what it would look like on `Int` later, and doing `int(UInt64.from_bytes(span))` doesn't seem like too much of a hit to me.

lsh (changes requested):

stdlib/src/builtin/int.mojo
msaelices: Sorry, I missed it. Done:
https://github.com/msaelices/mojo/commit/776b4dfd36427d90eeeba05ebf69cbe4bd7d4c0e
https://github.com/msaelices/mojo/commit/265b6481a6af040058f3c3da76ed7c17cf4ae5c5

lsh: Sorry for all the change requests, but to be clear I am comfortable merging these changes on `SIMD`, but not on `Int`. I think the `SIMD` API makes a ton of sense, but the `int` API would need some more thought since taking a `DType` feels wrong.

stdlib/src/builtin/simd.mojo
```diff
@@ -1879,6 +1880,63 @@ struct SIMD[type: DType, size: Int](
...
+        var value = ptr[]
+
+        @parameter
+        if is_big_endian() and not big_endian:
+            value = byte_swap(value)
+        elif not is_big_endian() and big_endian:
+            value = byte_swap(value)
+        return value
+
+    fn as_bytes[big_endian: Bool = False](self) -> List[Byte]:
```
lsh, line 1913: **thought**: maybe this should return `InlineArray[Byte, type.sizeof()]`? That way we don't allocate

stdlib/src/builtin/int.mojo
msaelices: Yes, actually the whole point of this PR was to help Python developers who need to migrate Python code using [int.from_bytes()](https://docs.python.org/3/library/stdtypes.html#int.from_bytes) and [int.to_bytes()](https://docs.python.org/3/library/stdtypes.html#int.to_bytes). So, to me, it's easier if we have it on the `Int` struct too, and iterate it.

stdlib/src/builtin/simd.mojo
msaelices: Good suggestion. I tried it in this commit: https://github.com/msaelices/mojo/commit/8caad4136b92231e9541b94076ab618871c2148c

Unfortunately, it's breaking the test with a weird compiler error, which is a Mojo bug IMO:
image

@lsh, do you have any ideas? If not, we can revert the commit, add a TODO line for future improvement, and use the `List[Byte]` in the meantime.

stdlib/src/builtin/int.mojo
lsh: I disagree with having it on the Int struct too because we would either:

* Break code users relied on
* Never get around to revising it and have a substandard API

I'm also not compelled by the idea that the goal is to mirror a Python API considering your earlier argument in favor of breaking away from the Python API using a parameter. As I mentioned before, I think `int(UInt64.from_bytes(my_span))` is fine for now

stdlib/src/builtin/simd.mojo
martinvuyk: It's not a compiler bug, it says so in the error. You need to initialize the array with a value to fill the whole array or pass it an argument to unsafely initialize it. Set it to 0 and it should be good to go.

msaelices: Sorry, the `call expansion failed` confused me. I need to read the errors better. 

Fixed here: https://github.com/msaelices/mojo/commit/dd2981411720f5dcf0cb7acf36b6200e4c3d4f11

Thanks

stdlib/src/builtin/int.mojo
```diff
@@ -1212,6 +1213,37 @@ struct Int(
...
+        D: DType, big_endian: Bool = False
```
lsh, line 1218: As mentioned above, I'm still opposed to adding a `DType` parameter to any `from_bytes`/`to_bytes` method for `Int`. We could either split these out into a new PR to nail down that API, or we can remove them.

stdlib/src/builtin/simd.mojo
```diff
@@ -1879,6 +1881,60 @@ struct SIMD[type: DType, size: Int](
...
+        @parameter
+        if is_big_endian() and not big_endian:
+            value = byte_swap(value)
+        elif not is_big_endian() and big_endian:
+            value = byte_swap(value)
+        return value
+
+    fn as_bytes[
+        big_endian: Bool = False
+    ](self) -> InlineArray[Byte, Self.type_len]:
```
lsh, line 1913: Nice! Glad this is working :)

stdlib/src/builtin/simd.mojo
```diff
@@ -1879,6 +1881,60 @@ struct SIMD[type: DType, size: Int](
 
         return bitcast[_integral_type_of[type](), size](self).cast[int_dtype]()
 
+    @staticmethod
+    fn from_bytes[
+        big_endian: Bool = False
+    ](bytes: InlineArray[Byte, Self.type_len]) raises -> Scalar[type]:
```
lsh, line 1887: Does this need to be `raises` anymore?

lsh (changes requested):

stdlib/src/builtin/int.mojo
msaelices: @lsh I agree. As it's also in all the SIMD scalar types, we don't need the `DType` parameter here. Good suggestion.

Fixed here: https://github.com/msaelices/mojo/commit/98c29fcdd490866e5e1f2ca4a85e201f6804bc85

stdlib/src/builtin/simd.mojo
msaelices: Good catch! Removed here: https://github.com/msaelices/mojo/commit/40b9fc016ce7ebf7246a0e0c7db08aea00c445e4

lsh:
!sync

lsh:
!sync

Issue #3229, opened 2024-07-12
[BUG] `SIMD.store/load` for `DType.bool`


rd4com:
### Bug description

Hello, after implementing `List.__contains__` for `Scalar[DT]`,
tests failed with `DType.bool` !

Minimal repro:
```mojo
alias width = simdwidthof[DType.bool]()
var x = DTypePointer[DType.bool].alloc(width)
alias T = SIMD[DType.bool, width]
T.store(x, T(True))
print(T.load(x)) #32 elements True
x[0] = False
print(T.load(x)) #8 firsts elements False
x.free()
```
Note: it seem to work correctly with others `DType`

Important: if we initialized with `T.store(x, T(False))` instead,
and do `x[0] = True`, it seem to work correctly, but it's not:
assigning to offset 1 results in item 8 being changed

This test shows different results:
```mojo
for i in range(width): 
    print(x[i])`        # see first four bytes
```

Possible fix:
It seem to think that the size of the type is 8 bytes
(debugging with `DTypePointer.__setitem__` and `__getitem__`)

(ping to @laszlokindrat)

### Steps to reproduce

Nothing special, the steps are in the description :+1: 

### System information

```shell
- Ubuntu 23.10
- mojo 2024.7.1105 (211527ca)
- modular 0.7.2 (d0adc668)
```

laszlokindrat:
@rd4com Would you mind trying to reproduce this on the latest nightly? There have been several patches over the past two months that addressed potentially related issues, so this might no longer be a problem.

rd4com:
@laszlokindrat , yes good job with all of this :+1: :heart_on_fire: 

It could be just a small thing, but the bug is still there:
```mojo
def main():
    alias width = 32
    var x = UnsafePointer[Scalar[DType.bool]].alloc(32)
    alias T = SIMD[DType.bool, width]
    x.store(0,T(True))
    print(x.load[width=32]()) #32 elements True
    x[0] = False
    print(x.load[width=32]()) #8 firsts elements False
    x.free()
```
With `DType.uint8`, it seem to work great (no bug) :+1:

helehex:
also see #2813

laszlokindrat:
This still reproduces:
```mojo
def main():
    alias width = 32
    var x = UnsafePointer[Scalar[DType.bool]].alloc(width)
    x.store(0, SIMD[DType.bool, width](fill=True))
    print(x.load[width=width]()) #32 elements True
    x[0] = False
[...]

PR #3800, opened 2024-11-22
[stdlib] Fix inconsistency with `UnsafePointer[SIMD[bool]]`

### LINKED ISSUE #2813: [BUG] `DTypePointer[bool]` packs bits inconsistently

helehex:
Fixes #2813 and #3229
This was originally fixed with `DTypePointer`, but regressed after it got removed.
I'm still curious about whether original behavior is intended or not, and what we want to do about the inconsistency (if anything).

`SIMD[bool]` does seem to be packed in memory, and I'm guessing that just carries over to `pop.store`/`pop.load`:
```mojo
var a = SIMD[DType.bool, 4](True)
var ptr = UnsafePointer.address_of(a)
print(ptr.bitcast[UInt8]().load())
# prints 15
```
Because of that, this fix may have unwanted consequences.

The solution in this PR is the same as before: it forces simd bools to be represented as bytes when doing stores/loads.
Doing it this way keeps simd bools consistent with the other dtypes when doing stores/loads of varying widths.

Maybe there's a way we could pack bools in a width-independent way, but i cant see that working with pointer offset.
we might want to have a separate method, or a parameter which keeps the original behavior.

Another thing i noticed, `bitwidthof[SIMD[DType.bool, 4]]()` returns `32`, so I'm really not sure what's intended here.

### DIFF
--- a/stdlib/src/memory/unsafe_pointer.mojo
+++ b/stdlib/src/memory/unsafe_pointer.mojo
@@ -479,6 +479,23 @@ struct UnsafePointer[
             "both volatile and invariant cannot be set at the same time",
         ]()
 
+        # bool load unpacks bits differently depending on the width.
+        # This causes issues when doing stores and loads of different widths.
+        # Cast between uint8 to keep a consistent representation in memory.
+        # TODO: decide how/whether to pack SIMD[bool] bits in memory
+        @parameter
+        if type == DType.bool:
+            return (
+                self.bitcast[UInt8]()
+                .load[
+                    width=width,
+                    alignment=alignment,
+                    volatile=volatile,
+                    invariant=invariant,
+                ]()
+                .cast[type]()
+            )
+
         @parameter
         if is_nvidia_gpu() and sizeof[type]() == 1 and alignment == 1:
             # LLVM lowering to PTX incorrectly vectorizes loads for 1-byte types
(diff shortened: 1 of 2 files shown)


JoeLoser (commented):
@abduld curious to get your thoughts here.

owenhilyard:
I'm not a fan of silently bit-packing bools, mostly due to the same reasons as why `std::vector<bool>` is considered a mistake. Could we get another DType for `packed_bool` or similar if it's built into the language and requires powers of 2 >= 8 (for byte alignment) widths in SIMD? 

The other place my mind goes is to move towards `Int[signed: Bool, width: Index]` since LLVM already has support which Zig and C23's _BitInt have tested. This means we have bool as an llvm `i1` for `BitBool` and keep `Bool` as byte sized, and people who want packing behavior can use `BitBool` and the compiler can do its best to bit-pack `BitBool` items in a struct. We can add implicit constructors to convert between them. This also helps `bitwidthof`, so that generic code which does checks does the right thing.

PR #3804, opened 2024-11-24
[stdlib] Faster variadic arguments `StringLiteral.join()` when arguments are strings - WIP

msaelices:
Note: It does not work yet as unpacked arguments are not supported yet by the compiler.


soraros:
@msaelices Is this still relevant?

JoeLoser:
Can/should we close this in favor of https://github.com/modular/modular/pull/4562?

PR #3807, opened 2024-11-25
[stdlib] [proposal] Add `DBuffer` type

### LINKED ISSUE #3797: [Feature Request] [proposal] Evolve `Span` into `Buffer`
### Review Mojo's priorities

### What is your request?

# Evolving `Span` into `Buffer`

The current `Span` is a struct which is defined as "A non owning view of
contiguous data".

`Span` is slowly becoming the de-facto way to exchange fat pointers over binary
and any other collectable types. `Span[Byte]` is converging into Python's
`bytes`.

`Span` currently allocates in its `__getitem__(self, slc: Slice) -> Self` if the
step is negative, however if the programmer doesn't read the docstring where it
warns that and the underlying `UnsafePointer` isn't taken ownership of, it will
lead to a memory leak. I see more of these cases eventually coming up if we keep
developing `Span`'s API further. This opens the door to massive leaks.

Thusly, I propose we let `Span` also own its data. And fitting to the change of
function, also changing its name to `Buffer` to reflect that the data can be
owned and mutated.

## How?

This part of the proposal can change and morph into something completely
different.

One way would be to change the underlying Pointer to one that can be or not be
owned. 
- If done at a parametrix level, it will lead to conflicts for methods which
  might return owned or non owned `Buffer`s.
- If done by adding a field to a Pointer, it will create struct padding issues.

Another way is to store the information of ownership in the upper bit of the
`Buffer`'s `_len` field, and thus avoiding adding struct padding overhead. The
current proposal will be limited to storing a single bit: `self_is_owner`. Which
in the `Buffer`'s destructor would fork it into `self._data.free()` or doing
nothing in case the `Buffer` doesn't own its data.

## Alternatives

We could also build an independent type `View`, or let `Span` stay and remove
the functions that allocate to a new `Buffer` type and lean towards it for most things.

## Future ideas

What we can build on top of this is a way to safely deal with pointer
allocations and make people feel comfortable using them directly. We could make
`Buffer.alloc()` a function that does `memset_zero()` to safely initialize the
memory.

We could develop `Buffer`'s API further:
```mojo
fn main():
    buf = Buffer[Byte].alloc(3)
    buf[0] = ord("h")
    buf[1] = ord("i")
    buf[2] = 0
    print(String(Buffer(other=buf))) # copies
    print(String(buf^)) # zero copy
    # buf does not do self._data.free() inside the String constructor because
    # .steal_data() sets the self_is_owner bit to False.
```

---
CC: @owenhilyard this is a bit of an evolved version of what was proposed in #3728
CC: @JoeLoser this is enables [#3239 (comment)](https://github.com/modularml/mojo/pull/3239#issuecomment-2493746583)

martinvuyk:
# Add `DBuffer` type
Closes #3797

This is a proof of concept and a potential approach to add  the type. Another suggested approach was using `Variant[List[T], Span[T]]` but because of complications with the current origin system and the amount of branching that `Variant` introduces, the current approach is to just use 1 bit from the `_len` field to store an ownership flag.

This can potentially become a more generic version of Python's `bytes`/`bytearray`. Many APIs will benefit from being able to dynamically take ownership of data instead of copying, while also allowing mutable or immutable `Span`s as an input. This also allows abstracting over `InlineArray` by using `DBuffer`'s constructor. Any stack allocated pointer can also be used.

This type can also be potentially phagocytized by `List`, becoming it's underlying mechanism to signal ownership. And as such allow shallow copies of `List` in its `__copyinit__` and deep copies in its `List(other=other_list)` constructor. This is an alternative that seems very compelling to me since it will give "free" (at the cost of the flag handling) performance improvements to many programs. One thing that holds this back is the current clunkyness of parametrized origins.

```mojo
struct DBuffer[
    is_mutable: Bool, //,
    T: CollectionElement,
    origin: Origin[is_mutable].type,
](CollectionElementNew):
    """A potentially owning view of contiguous data.

    Parameters:
        is_mutable: Whether the DBuffer is mutable.
        T: The type of the elements in the DBuffer.
        origin: The origin of the DBuffer.

    Examples:

    ```mojo
    %# from collections.dbuffer import DBuffer
    fn parse(
        owned buf: DBuffer[Byte], encoding: String = "utf-8"
    ) raises -> String:
        if encoding == "utf-16":
            ...
        elif encoding == "utf-8":
            debug_assert(
                len(buf) > 0 and buf[-1] == 0,
                "parser expects null terminated data"
            )
            return String(ptr=buf.steal_data(), length=len(buf))
        else:
            raise Error("Unsupported encoding")

    fn main() raises:
        l1 = List[Byte](ord("h"), ord("i"), 0)
        # l1 gets implicitly built into a DBuffer that doesn't own the data.
        # Since the DBuffer doesn't own the data, the method steal_data()
        # makes a copy of the data to pass as owned to the String constructor
        print(parse(l1)) # hi
        # Passing an owned DBuffer makes the .steal_data() inside the parse
        # method not allocate
        print(parse(DBuffer[origin=MutableAnyOrigin].own(l1^))) # hi
        # the compiler won't let you use l1 beyond this point
    ```
    """

    alias _intwidth = bitwidthof[Int]()
    alias _shift = Self._intwidth - 1
    alias _len_mask = 0x7F_FF if Self._intwidth == 32 else 0x7F_FF_FF_FF
    alias _max_length = 2**Self._shift
    var _data: UnsafePointer[T]
    var _len: UInt

    ...
```

PS: I wanted to name this type `Buffer`, but a `Buffer` type already exists

### DIFF
--- a/stdlib/src/collections/dbuffer.mojo
+++ b/stdlib/src/collections/dbuffer.mojo
@@ -191,6 +191,9 @@ struct DBuffer[
         Args:
             list: The list to which the DBuffer refers.
 
+        Returns:
+            The owned DBuffer with the data.
+
         Examples:
 
         ```mojo
--- a/stdlib/src/collections/dbuffer.mojo
+++ b/stdlib/src/collections/dbuffer.mojo
@@ -1,0 +1,477 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Defines the DBuffer type."""
+
+
+from builtin.builtin_list import _lit_mut_cast
+from collections import InlineArray, List
+from memory import UnsafePointer
+from sys.info import bitwidthof
+from utils import Span
+
+
+@value
+struct _DBufferIter[
+    is_mutable: Bool, //,
+    T: CollectionElement,
+    origin: Origin[is_mutable].type,
+    forward: Bool = True,
+]:
+    """Iterator for DBuffer.
+
+    Parameters:
+        is_mutable: Whether the reference to the DBuffer is mutable.
+        T: The type of the elements in the DBuffer.
+        origin: The origin of the DBuffer.
+        forward: The iteration direction. `False` is backwards.
+    """
+
+    var index: Int
+    var src: DBuffer[T, origin]
+
+    @always_inline
+    fn __iter__(self) -> Self:
+        return self
+
+    @always_inline
+    fn __next__(inout self) -> ref [origin] T:
+        @parameter
+        if forward:
+            self.index += 1
+            return self.src[self.index - 1]
+        else:
+            self.index -= 1
+            return self.src[self.index]
+
+    @always_inline
+    fn __has_next__(self) -> Bool:
+        return self.__len__() > 0
+
+    @always_inline
+    fn __len__(self) -> Int:
+        @parameter
+        if forward:
+            return len(self.src) - self.index
+        else:
+            return self.index
+
+
+# TODO: decide whether DBuffer will be the entrypoint API for Python-like bytes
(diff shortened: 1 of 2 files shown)


lsh:
Closing for now based on my comment on #3797 about deferring the proposal.

martinvuyk:
@gabrieldemarmiesse today I finally remembered to read some of your SBO work. I think this approach will interest you. It kind of achieves the same thing at runtime. We wouldn't get the same comptime guarantees, but we also get the perf improvements of less allocations. `String` and any other type could pass arbitrary stack pointers, and the only price paid is the bit flag handling (no struct padding or workarounds for the compiler issues). One big issue though would be having to add a parametrized origin to List. WDYT?

Issue #3581, opened 2024-10-01
[Feature Request] `is_trivial`


rd4com:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

It would be nice to have a function that check if an `AnyRegType` became `AnyType`:

```mojo
fn is_trivial[T:AnyTrivialRegType]()->Bool:
    return True
fn is_trivial[T:AnyType]()->Bool:
    return False

def main():
    print(is_trivial[Int]()) #not yet
```

### What is your motivation for this change?

To simplify `List` that currently take 2 parameters: `type` and `is_trivial_hint`,

it could also be useful in `InlinedFixedVector.__del__`.

JoeLoser:
Unfortunately, I don't think it's quite that simple as you've shown in your example.  In the compiler, we know if the type is "trivial", but we don't have sufficient information to do this in the library.  Reassigning to the compiler team, but do know I requested this a month or two back during Gabriel's string optimization work.

JoeLoser:
Just like all types inherit from `AnyType`, I think you'd want types (in the stdlib and user defined types) to implicitly conform to `Trivial` for example based on the triviality bit known in the front end for a given type.

NathanSWard:
This should be fixed as of the most recent Mojo version. We now have a `TrivialRegisterType` trait which you can check with `conforms_to`.
```
fn is_trivial[T: AnyType]() -> Bool:
    return conforms_to(T, TrivialRegisterType)
```

PR #3809, opened 2024-11-25
[stdlib] Integrate `hint_trivial_type` in `List.__copyinit__`

rd4com:
Hello, 
`List.__copyinit__ ` needed to implement `@parameter if hint_trivial_type`,
so here is a PR :+1: !

Because `String.__copyinit__` does `List.__copyinit__`,
and that it's buffer uses `hint_trivial_type`,
there is a nice speedup from it!

:partying_face:  nice speedup for:
  - `List[_, hint_trivial_type=True].__copyinit__ `
  - `String.__copyinit__` (57x, increase with length)
	(`var buffer: List[UInt8, hint_trivial_type=True]`)
  - `fn(**kwargs)` should be faster too !
	(see PR #3133, because it currently does `__copyinit__` of `DictEntry[K,V]`)
	(`_find_slot` just need to use `Pointer` :+1: )
  - More types that uses `hint_trivial_type=True` ?

I need to learn the `Benchmark`, but here is a simple one (`benchmarks/collections/bench_list`)
(please somebody help try to convert it to `Benchmark`)

Is it possible to check if it is faster or that the benchmark is good ?

The benchmark for `Int` is less than for `UInt8`, is it a bug ?
- (`4512122÷720588 = 6.2x`) for size `4096`
- (`5425619÷102102 = 53x`) for size `4096`

#### results PR:

```
iterations:  1024
list_hint_trivial_type
	 1		33281
	 2		43820
	 4		41106
	 8		28076
	 16		26829
	 32		27688
	 128	35084
	 256	46950
	 512	73477
	 1024	115724
	 2048	339350
	 4096	720588

string_copyinit
	 1		26295
	 2		28299
	 4		39899
	 8		40835
	 16		32544
	 32		27912
	 128	30006
	 256	31930
	 512	32685
	 1024	39246
	 2048	59733
	 4096	102102
```

#### results currently:

```
iterations:  1024
list_hint_trivial_type
	 1		43318
	 2		29643
	 4		30210
	 8		32667
	 16		42152
	 32		58145
	 128	157062
	 256	325332
	 512	534089
	 1024	1036117
	 2048	2079208
	 4096	4512122

string_copyinit
	 1		28839
	 2 	 	30548
	 4 	 	33308
	 8 	 	37221
	 16		47267
	 32		67287
	 128	190275
	 256	347621
	 512 	714950
	 1024 	1327402
	 2048 	2785542
	 4096	5425619
```

### DIFF
--- a/stdlib/benchmarks/collections/bench_list.mojo
+++ b/stdlib/benchmarks/collections/bench_list.mojo
@@ -15,74 +15,81 @@
 # the -t flag. Remember to replace it again before pushing any code.
 
 from benchmark import Bench, BenchConfig, Bencher, BenchId, Unit, keep, run
-from memory import UnsafePointer
 from random import seed
 from collections import List
-from collections import Dict
-from time import now
+from random import *
 
 
-def benchmark_list_hint_trivial_type_int[length: Int, iterations: Int]() -> Int:
-    var start = now()
-    var stop = now()
-    alias size = length
+# ===-----------------------------------------------------------------------===#
+# Benchmark Data
+# ===-----------------------------------------------------------------------===#
+fn make_list[
+    size: Int, DT: DType, is_trivial: Bool
+]() -> List[Scalar[DT], is_trivial]:
+    alias scalar_t = Scalar[DT]
+    var d = List[Scalar[DT], is_trivial](capacity=size)
+    rand[DT](
+        d.unsafe_ptr(),
+        size,
+        min=scalar_t.MIN.cast[DType.float64](),
+        max=scalar_t.MAX.cast[DType.float64](),
+    )
+    d.size = size
+    return d
 
-    var items = List[Int, True]()
-    for i in range(size):
-        items.append(i)
 
-    start = now()
-    for iter in range(iterations):
-        var items2 = items
-        keep(items2.data)
-    stop = now()
+# ===-----------------------------------------------------------------------===#
+# Benchmark `List[DT, True].__copyinit__`
+# ===-----------------------------------------------------------------------===#
+
+
+@parameter
+fn bench_list_copyinit[
+    size: Int, DT: DType, is_trivial: Bool
+](mut b: Bencher) raises:
+    var items = make_list[size, DT, is_trivial]()
+    var result = List[Scalar[DT], is_trivial]()
+    var res = 0
+
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        result = items
+        res += len(result)
+        keep(result.data)
+        keep(items.data)
+
+    b.iter[call_fn]()
+    print(res)
+    keep(bool(items))
+    keep(bool(result))
+    keep(result.data)
     keep(items.data)
-    return stop - start
-
-
-def benchmark_string_copyinit__[length: Int, iterations: Int]() -> Int:
-    var start = now()
-    var stop = now()
-
-    var x: String = ""
-    for l in range(length):
-        x += str(l)[0]
-
-    start = now()
-    for iter in range(iterations):
-        var y: String
-        String.__copyinit__(y, x)
-        keep(y._buffer.data)
-    stop = now()
-    keep(x._buffer.data)
-    return stop - start
 
 
 def main():
     seed()
 
-    alias iterations = 1 << 10
-
-    alias result_type = Dict[String, Int]
-    var results = Dict[String, result_type]()
-    results["list_hint_trivial_type"] = result_type()
-    results["string_copyinit"] = result_type()
-
+    var m = Bench(
+        BenchConfig(
+            num_repetitions=1,
+            max_runtime_secs=0.5,
+            min_runtime_secs=0.25,
+            min_warmuptime_secs=0,  # 0.25
+        )
+    )
     alias lengths = (1, 2, 4, 8, 16, 32, 128, 256, 512, 1024, 2048, 4096)
 
     @parameter
     for i in range(len(lengths)):
         alias length = lengths.get[i, Int]()
-        results["list_hint_trivial_type"][
-            str(length)
-        ] = benchmark_list_hint_trivial_type_int[length, iterations]()
-        results["string_copyinit"][str(length)] = benchmark_string_copyinit__[
-            length, iterations
-        ]()
+        m.bench_function[bench_list_copyinit[length, DType.uint8, True]](
+            BenchId("List[Scalar[DT], True].__copyinit__ [" + str(length) + "]")
+        )
+        m.bench_function[bench_list_copyinit[length, DType.uint8, False]](
+            BenchId(
+                "List[Scalar[DT], False].__copyinit__ [" + str(length) + "]"
+            )
+        )
 
-    print("iterations: ", iterations)
-    for benchmark in results:
-        print(benchmark[])
-        for result in results[benchmark[]]:
-            print("\t", result[], "\t", results[benchmark[]][result[]])
-        print()
+    m.dump_report()
--- a/stdlib/benchmarks/collections/bench_list.mojo
+++ b/stdlib/benchmarks/collections/bench_list.mojo
@@ -1,0 +1,88 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


martinvuyk:
Hi @rd4com, I don't think this needed a benchmark it's a bit overkill but nice :rofl:.

FYI, I just opened PR #3810 to merge a change from another PR which got stuck long ago, that is useful for this kind of optimization. I was thinking of tackling these kind of wholes later on, I completely forgot we had the hint parameter :laughing:. If we combine that with your request #3581 we will be able to optimize this automatically without the hint :rocket:.

rd4com:
@martinvuyk , great work on #3810 !
It can give the `List` the speedup even `if hint_trivial_type == False` :partying_face: 
So users could get the speedup with a simple `List[Float64]`! (no `hint_trivial_type`)
(but not `Int` yet, maybe need a temporary thing for it)

```mojo
@parameter
if hint_trivial_type or get_dtype[T]!=DType.invalid:
	memcpy(self.data, other.data, 1)
else:
	for i in range(len(existing)):
		...
```

For the `is_trivial[T]`, it is probably somewhere in the future as trait (chat `#nightly`),
So let's just wait for features to naturally grow,
and auto-optimize for basic things like `List[Int]` and `dtype` how we currently can :+1: 
(that way users get the speedup before even learning trivial type  :partying_face: )

martinvuyk:
> It can give the `List` the speedup even `if hint_trivial_type == False` 🥳 So users could get the speedup with a simple `List[Float64]`! (no `hint_trivial_type`) (but not `Int` yet, maybe need a temporary thing for it)

Yep, I'm pretty stoked too :smile:

 

> ```mojo
> @parameter
> if hint_trivial_type or get_dtype[T]!=DType.invalid:
> ```

I was thinking more along the lines of having a private alias inside `List` that has its trivial `DType` or `DType.invalid`
```mojo
fn _get_same_bitwidth_dtype[T: AnyType, hint_trivial_type: Bool]() -> DType:
    @parameter
    if not hint_trivial_type:
        return DType.invalid

    alias width = bitwidthof[T]()
    
    @parameter
    if width == 64:
        return DType.uint64
    elif width == 32:
        return DType.uint32
    elif width == 16:
        return DType.uint16
    elif width == 8:
        return DType.uint8
    else:
        return DType.invalid


struct List[T: CollectionElement, hint_trivial_type: Bool = False]:
    alias _D = (
        DType.get_dtype[T]() if DType.is_scalar[T]()
        else _get_same_bitwidth_dtype[T, hint_trivial_type]()
    )
```

This way the compiler runs the logic only once and not on every function call, and we'll reduce compile times.

> (but not Int yet, maybe need a temporary thing for it)

We can totally add a temporary hack for it inside the `_get_same_bitwidth_dtype` function using `_type_is_eq` until we have the `is_trivial[T: AnyType]() -> Bool` function

rd4com:
❤️‍🔥 ✅ auto-optimize `hint_trivial_type`
	(speedup for `List[Int]`, `List[UInt8]`, `List[Scalar]`)

@martinvuyk , it works!
(See `# FIXME: Use dtypes checker of PR #3810 (DType.get_dtype[T]()`)

But if we change `hint_trivial_type=False` to `hint_trivial_type=_trivial_default[T]()`,
there is a test error:
```
test/test_utils/types.mojo:159:8: note: candidate not viable: argument #2 cannot be converted from 'UnsafePointer[List[Int], alignment=1]' to 'UnsafePointer[List[Int, False]]'
struct ValueDestructorRecorder(ExplicitlyCopyable):
```

(ping to @joeloser)

martinvuyk:
@rd4com that's cool though I didn't mean to do it in this PR, it might extend it more since I think this concept alone needs some independent review, for example:
- I think the alias should be private
- The alias should be a `DType` for later rebinding after the checking e.g. `@parameter if Self._D is not DType.invalid: memcpy(self.unsafe_ptr().bitcast[Scalar[Self._D]](), ...) ...`
- IMO the dtype getter should just be a helper func in the `list.mojo` file (maybe we'll take it to `Span` later on)

> test/test_utils/types.mojo:159:8: note: candidate not viable: argument #2 cannot be converted from 'UnsafePointer[List[Int], alignment=1]' to 'UnsafePointer[List[Int, False]]'

this seems like a compiler frontend bug (the function is not getting executed for one version but it is for the other)

Until PR #3810 gets merged I think its perfectly fine to keep using `hint_trivial_type` as the parameter, the switch to the `DType` logic should be straightforward

JoeLoser:
@rd4com this is great!  To keep things self-contained, can we keep this PR focused on the original commit of eb869a44f1568d83999a806e4fd7f56415286e02 and land that first?  Then we can look at https://github.com/modularml/mojo/pull/3810 and integrate that in with your future commits here to extending the optimization to other known trivial types.

rd4com:
Thanks, yep :+1: 

Just rebased and split the commits for another PR !

(ping to @JoeLoser)

JoeLoser:
!sync

JoeLoser:
> Thanks, yep 👍
> 
> Just rebased and split the commits for another PR !
> 
> (ping to @JoeLoser)

Thanks!  I have an open PR split off from this for adding a test + the optimization for not destroying trivial elements when the hint is provided.  I'll need to do a few things to this imported PR before we can land it:
- Measure compile time impact on e2e models (should be OK, but good to double check) for the runtime perf optimization in trivial element case.

stdlib/benchmarks/collections/bench_list.mojo
```diff
@@ -0,0 +1,88 @@
...
+from benchmark import Bench, BenchConfig, Bencher, BenchId, Unit, keep, run
+from memory import UnsafePointer
+from random import seed
+from collections import List
+from collections import Dict
+from time import now
+
+
+def benchmark_list_hint_trivial_type_int[length: Int, iterations: Int]() -> Int:
+    var start = now()
```
JoeLoser, line 26: **Suggestion** Instead of doing this stopwatch sort of thing, can we use the `benchmark` module as intended?  You can take a look at the other benchmarks for how this sort of thing can be done.  Thanks!

JoeLoser (changes requested):
Formally requesting changes until https://github.com/modularml/mojo/pull/3809#discussion_r1874191642 is addressed.  I'll do some internal benchmarking with compile-times on e2e models like llama as well.

rd4com:
Hello, 
sorry for late commit, changes implemented :+1: 
- ✅ [stdlib] Add parametrized `benchmarks` for `List[Scalar[DT], is_trivial].__copyinit__`

Again, still learning `Benchmark` (please have a look because it is slower than initial one)

(ping to @JoeLoser)

stdlib/benchmarks/collections/bench_list.mojo
```diff
@@ -0,0 +1,95 @@
...
+    b.iter[call_fn]()
+    print(res)
+    keep(bool(items))
+    keep(bool(result))
+    keep(result.data)
+    keep(items.data)
+
+
+def main():
+    seed()
```
JoeLoser, line 71: **Question** Why call `seed()`?

rd4com: Yes let's remove it, i usually sum the results and check if results are all the same to make sure it worked,
but we now have `Benchmark.keep` :+1:

stdlib/benchmarks/collections/bench_list.mojo
```diff
@@ -0,0 +1,93 @@
...
+    keep(items.data)
+
+
+def main():
+    var m = Bench(
+        BenchConfig(
+            num_repetitions=1,
+            max_runtime_secs=0.5,
+            min_runtime_secs=0.25,
+            min_warmuptime_secs=0,  # FIXME: adjust the values
```
JoeLoser, line 76: **Question** Any harm in using the default `BenchConfig` values?

JoeLoser:
!sync

stdlib/benchmarks/collections/bench_list.mojo
rd4com: It takes too much time on my small laptop with defaults
(please replace it with what makes sense)

Would you prefer a commit with defaults ?

rd4com:
Hello @JoeLoser , 
does this PR need more changes (tag is: `waiting-reponse`) ?

JoeLoser:
Hi @rd4com, I just reworked this PR internally to be a NFC and just add tests + benchmarks.  Note that @lattner recently already improved using `hint_trivial_type` in more places, including this one in his overhaul of String recently for SSO and such.

rd4com:
> Hi @rd4com, I just reworked this PR internally to be a NFC and just add tests + benchmarks. Note that @lattner recently already improved using `hint_trivial_type` in more places, including this one in his overhaul of String recently for SSO and such.

Yep, that's great, good job, all good then ?
(Should i close this pr ?)

PR #3810, opened 2024-11-25
[stdlib] Add `DType.get_dtype()`

martinvuyk:
Add `DType.get_dtype()`.

This will enable a lot of vector based optimizations for more generic function signatures.

This is a split-off from PR #3577

Examples:
```mojo
fn memset[type: Copyable](ptr: UnsafePointer[type], value: type, count: Int):
    alias dt = DType.get_dtype[type]()

    @parameter
    if dt is not DType.invalid:
        var p = ptr.bitcast[Scalar[dt]]()
        _memset_impl[dt](p, rebind[Scalar[dt]](value), count)
    else:
        for i in range(count):
            (ptr + i).init_pointee_copy(value)
```

### DIFF
--- a/mojo/stdlib/stdlib/builtin/dtype.mojo
+++ b/mojo/stdlib/stdlib/builtin/dtype.mojo
@@ -19,6 +19,8 @@ These are Mojo built-ins, so you don't need to import them.
 from hashlib._hasher import _HashableWithHasher, _Hasher
 from os import abort
 from sys import bitwidthof, os_is_windows, sizeof
+from sys.intrinsics import _type_is_eq
+
 
 alias _mIsSigned = UInt8(1)
 alias _mIsInteger = UInt8(1 << 7)
@@ -209,7 +211,7 @@ struct DType(
             str: The name of the DType.
         """
         if str.startswith("DType."):
-            return Self._from_str(String(str).removeprefix("DType."))
+            return Self._from_str(str.removeprefix("DType."))
         elif str == "bool":
             return DType.bool
         elif str == "index":
@@ -255,10 +257,12 @@ struct DType(
             return DType.bfloat16
         elif str == "float16":
             return DType.float16
+
         elif str == "float32":
             return DType.float32
         elif str == "tensor_float32":
             return DType.tensor_float32
+
         elif str == "float64":
             return DType.float64
 
@@ -319,23 +323,25 @@ struct DType(
 
         elif self is DType.float8_e3m4:
             return writer.write("float8_e3m4")
-        elif self is DType.float8_e5m2:
-            return writer.write("float8_e5m2")
-        elif self is DType.float8_e5m2fnuz:
-            return writer.write("float8_e5m2fnuz")
         elif self is DType.float8_e4m3fn:
             return writer.write("float8_e4m3fn")
         elif self is DType.float8_e4m3fnuz:
             return writer.write("float8_e4m3fnuz")
+        elif self is DType.float8_e5m2:
+            return writer.write("float8_e5m2")
+        elif self is DType.float8_e5m2fnuz:
+            return writer.write("float8_e5m2fnuz")
 
         elif self is DType.bfloat16:
             return writer.write("bfloat16")
         elif self is DType.float16:
             return writer.write("float16")
+
         elif self is DType.float32:
             return writer.write("float32")
         elif self is DType.tensor_float32:
             return writer.write("tensor_float32")
+
         elif self is DType.float64:
             return writer.write("float64")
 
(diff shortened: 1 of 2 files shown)


stdlib/src/builtin/dtype.mojo
```diff
@@ -617,6 +619,67 @@ struct DType(
...
+    fn get_dtype[T: AnyType, size: Int = 1]() -> DType:
+        """Get the `DType` if the given Type is a `SIMD[_, size]` of a `DType`.
+        Parameters:
+            T: AnyType.
+            size: The SIMD size to compare against.
+        Returns:
+            The `DType` if matched, otherwise `DType.invalid`.
+        """
+
+        if _type_is_eq[T, SIMD[DType.bool, size]]():
```
martinvuyk, line 636: I don't know compiler internals, but I figured adding a `@parameter` would be redundant after the first execution (?). Should I add it anyway?

JoeLoser: Let's add it anyways as I'm not sure.

Re the implementation itself, is there a way we can leverage the `element_type` alias from SIMD to simplify this?  Conceptually, we want to just retrieve `T.element_type` for SIMD types and return `DType.invalid` otherwise.

martinvuyk: To do that we'd somehow need to rebind the `AnyType` to a `SIMD`, which I don't think the compiler will allow when the type is not `SIMD`. I think we'd need a version of `_type_is_eq` that works for unbound parameters like what @rd4com showed on Discord: `_type_is_eq[SIMD[_, _]]()` and then we'd need an "unbound rebind" where we'd do `dtype = rebind[SIMD[_, _]](value).element_type`, but I think this would need some very heavy involvement on the compiler side.

I had another idea that we could also add another method `.get_simd() -> (DType, Int)` that finds the vector characteristics by brute force
```mojo
@parameter
for i in range(len(dtypes)):
    @parameter
    for j in range(len(sizes)):
        alias size = sizes.get[j, Int]()

        @parameter
        if sizeof[T]() // size != 1:
            continue

        alias dt = dtypes.get[i, DType]()

        @parameter
        if _type_is_eq[T, SIMD[dt, size]](value)
            return dt, size
return DType.invalid, 1
```
This seems like it would make compile times suffer, but it would actually only need `alias sizes = (1, 2, 4, 8)` which would then work for only one size `sizeof[T]() // size != 1`, and then it just unrolls the comparison for every dtype. It's still expensive if not used properly, but I think we could execute it only once and save the results as aliases inside the struct like:
```mojo
alias _res = DType.get_simd[T]()
alias _D = Self._res[0]
alias _size = Self._res[1]
alias _SIMD = SIMD[Self._D, Self._size]
alias _SpanT = Span[Self._SIMD, _]
``` 
We can leave this for another PR if you'd like. I think the current `.get_dtype()` implementation is good for most scalar use cases or those where one knows the size of the `SIMD` vectors. WDYT?

martinvuyk: @JoeLoser gentle ping. I think we won't be able to implement what you're suggesting without having reflection first. So could we move forward with this PR?

lattner:
The _from_str fixes seem "obviously good" to me, the extension I have less insight into though

martinvuyk:
@JoeLoser now that I'm going through the stdlib searching for places where we should use `UInt` instead of `Int` I came across this:
```mojo
        # TODO: Simplify this with a UInt type.
        debug_assert(
            -self._len <= Int(idx) < self._len, "index must be within bounds"
        )
        # TODO(MSTDL-1086): optimize away SIMD/UInt normalization check
        var offset = Int(idx)
        if offset < 0:
            offset += len(self)
        return self._data[offset]
```
I already changed it to:
```mojo
        var offset = UInt(idx.__index__())
        # TODO(MSTDL-1086): optimize away SIMD normalization check
        if not _type_is_eq[I, UInt]():
            if Int(idx) < 0:
                offset = len(self) + Int(idx)
        debug_assert(offset < self._len, "index must be within bounds")
        return self._data[offset]
```

MSTDL-1086 would be solved by doing:
```mojo
        var offset = UInt(idx.__index__())
        alias dt = DType.get_dtype[I]()

        @parameter
        if not _type_is_eq[I, UInt]() and not dt.is_unsigned():
            if Int(idx) < 0:
                offset = len(self) + Int(idx)
        debug_assert(offset < self._len, "index must be within bounds")
        return self._data[offset]
```

JoeLoser:
!sync

mojo/stdlib/test/builtin/test_dtype.mojo
```diff
@@ -100,10 +100,34 @@ def test_from_str():
         assert_equal(DType._from_str(String(dtypes[i])), dtypes[i])
 
 
+def test_get_dtype():
+    def _test[D: DType]():
+        assert_equal(DType.get_dtype[Scalar[D]](), D)
+
+        @parameter
+        for i in range(32):
+            assert_equal(DType.get_dtype[SIMD[D, i], i](), D)
```
soraros: This won't work when we have `requires`. Maybe take the extra effort to write the test for `[1, 2, 4, ..., 32]`.

martinvuyk: why wouldn't this work?

soraros: `SIMD[dt, 3]` will not type check. It's further down the line, we don't need to fix it now though.

martinvuyk: ahaaa, true lol. I'll just add a check for power of two and continue lol

soraros: sneakyyyyyyy. `2**i` would have been better.

mojo/stdlib/stdlib/builtin/dtype.mojo
```diff
@@ -847,6 +849,91 @@ struct DType(
...
+        elif _type_is_eq[T, SIMD[DType.int128, size]]():
+            return DType.int128
+        elif _type_is_eq[T, SIMD[DType.uint128, size]]():
+            return DType.uint128
+        elif _type_is_eq[T, SIMD[DType.int256, size]]():
+            return DType.int256
+        elif _type_is_eq[T, SIMD[DType.uint256, size]]():
+            return DType.uint256
+
+        elif _type_is_eq[T, SIMD[DType.float8_e3m4, size]]():
```
soraros, line 905: Absolute nit, don't have to do it if you don't feel like it: could you reorder the (floating) types to be alphabetical to match the `__mlir_type` method? I'd like all the code branching on `dtype` to have a common shape.

mojo/stdlib/test/builtin/test_dtype.mojo
martinvuyk: > 2**i would have been better.

will do :+1:

mojo/stdlib/stdlib/builtin/dtype.mojo
martinvuyk: I also fixed it in other places now

JoeLoser:
!sync

Issue #3811, opened 2024-11-25
[Feature Request] [stdlib] [proposal] Remove all uses of `type` as a type parameter.


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

`type` is the name of a Python builtin function. It is likely that in the future we will have a similar function to do analogous things. However, there are many places throughout the stdlib where `type: DType` or `type: AnyType` are used. I propose we change all of them to `D: DType` and `T: AnyType` to follow the same parameter naming convention as the rest of the stdlib.

### What is your motivation for this change?

...

### Any other details?

This is a good first issue for people looking to contribute

gryznar:
Why this can behave as a soft keyword like current?

martinvuyk:
Because types should be uppercase anyway. And the word type will become a function. If you use the constructor syntax on a `type: AnyType` parameter, it won't work for the global builtin function. `Type` is also not an option because it might become a struct/compiler builtin (see proposal #3756). `type: DType` also makes no sense because it's a specific subset of types, namely `DType`'s fields.

ConnorGray:
Hey folks, I just wanted to communicate what the Mojo standard library team discussed re. this proposal:

* We are +1 to renaming `type: DType` parameters to `dtype: DType`, for easier at-a-glance readability that that identifier refers to a `DType` value. We would accept PRs making this change.

* We want to postpone work on renaming `type: Trait` parameters. We acknowledge the potential for naming clashes if/when Mojo's reflection capabilities improve and a builtin-in `type()` function is added. However, we feel we can "cross that bridge when we come to it" and do the rename lazily, and only when necessary.

   * We want to avoid PRs "merely" renaming `type: Trait` to `T: Trait` or `ElementType: Trait`, to preserve PR review capacity for feature and fix PRs. However, we would accept PRs that happen to use `T` or `ElementType`  (instead of `type`) as type parameter names in _new_ functionality.

As always, please let me know if you have any questions or concerns🙂

martinvuyk:
Superseded by #4215

PR #3812, opened 2024-11-26
[stdlib] [NFC] Rename `SIMD.type` to `SIMD.D`

martinvuyk:
Rename `SIMD.type` to `SIMD.D`. Part of #3811

### DIFF
--- a/stdlib/src/builtin/simd.mojo
+++ b/stdlib/src/builtin/simd.mojo
@@ -115,43 +115,41 @@ alias Byte = UInt8
 
 
 @always_inline("nodebug")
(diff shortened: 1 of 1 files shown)


gryznar:
-1. This look strange

martinvuyk:
`size` is a variable
`D` is a Type
I don't see your point

JoeLoser:
I understand the intent, but let's hold off on this PR until we discuss https://github.com/modularml/mojo/issues/3811 and whether it's something we want to do.  I don't see a need to worry about this _now_.  Let's lazily evaluate this when we get a `type` function and the like.

Issue #3511, opened 2024-09-20
[BUG] The field storing the number of items in a List or Dict should be named "length", not "size"


nmsmith:
### Bug description

The field `List.size` should probably be renamed to `List.length`, because:
- That's what Python's `len()` function returns (it returns the # of elements in any collection)
- "length" is used in everyday English when talking about lists

A similar reasoning applies to `Dict.size`. The expression `len(my_dict)` obtains the number of elements, so we should probably be referring to the number of elements as the "length" of the dictionary.

Also, the term "size" is more often used to refer to memory consumption, e.g. C++'s `sizeof` operator returns the number of _bytes_ that an object consumes. It would be good to avoid this conflation.

### Steps to reproduce

N/A

lattner:
+1, this makes sense to me.

JoeLoser:
> +1, this makes sense to me. @JoeLoser wdyt?

+1 to renaming the field names here for consistency (using `length` instead of `size`).  We should also signal that they are meant to be private as well by prefixing these with an `_`.  It would be bad for users to do something like

```mojo
var xs = List[String]()
xs.size = 10 # oops

# When we run the destructor for `xs` here, we'd try to free uninitialized memory
```

Our team can do all this internally and fix up all the uses.

nmsmith:
Also, if we're renaming the word "size", it might also make sense to rename the Sized trait (that includes the __len__ method) to something different. Notably, in Rust, the Sized trait means something [completely different](https://doc.rust-lang.org/std/marker/trait.Sized.html). Unfortunately there isn't really an English adjective that means "has a length", but the name `Countable` seems very accurate.

(I'll open a different issue for this if desired.)

JoeLoser:
We definitely had similar thoughts since the "able" suffix naming doesn't quite work with the (currently) named `Size`, e.g. `Lengthable` isn't great.  Mind opening a separate GH issue for that and we can discuss with the community, please?

nmsmith:
That said, given Mojo already invented the adjectives Intable and Stringable, maybe Lengthable would be fine. 🤷‍♀️

soraros:
Let's use `_len`: it's short, less specific (can be applied to types where a literal "length" doesn't make sense, but still maintains parity with `len`), and somewhat "ugly", indicating its internal nature.

> the name `Countable` seems very accurate.
+1

nmsmith:
For discussing the `Sized` trait, I opened #3514.

lattner:
[...]

PR #3814, opened 2024-11-27
[stdlib] Rename `List.size` to `List._len` and refactor usage of the field to use the public API

martinvuyk:
### Rename `List.size` to `List._len` and refactor usage of the field to use the public API.

Details:
- Part of #3511
- To prepare for any future implementation details than can change under the hood for `List`.


martinvuyk:
I'll leave this as draft until a decision is made of whether to deprecate accessing the fields directly. Since this implementation will have issues when trying to access the field with an immediate operation e.g. `some_list.length += 1`, which was solved only for `some_list.size += 1` by making `size` the default `getattr` field name

JoeLoser (approved):

JoeLoser:
> Personal opinion: I'd like it if we could just deprecate letting people modify the length variable directly (removing setters and getters), and make them use the public API.

Let's do that and remove the setters/getters.

Can you please also add a changelog entry so people understand their compiler errors as a result of this change if they are accessing the length field directly of `List`?

stdlib/src/collections/list.mojo
```diff
@@ -467,38 +468,90 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
...
+
+        Parameters:
+            D: The DType.
+
+        Args:
+            value: The value to append.
+        """
+        self.reserve(len(self) + len(value))
+        memcpy(self.data + len(self), value.unsafe_ptr(), len(value))
+        self._len += len(value)
```
ConnorGray, line 554: **Suggestion** I think it would be better if we leave these new `append()` overloads to a separate PR. I know it can feel like a small thing to add QoI "goodies" to a PR like this, but it can make the PR significantly harder to review when its changing multiple things at once 😌 

Layering multiple changes in one PR can feel more efficient since it means more changes can merge "all at once", but I think it might lower overall throughput since each PR takes more brain 🧠 to review.

For some additional context, when we're reviewing community PRs, it's often in between other unrelated internal work. A small PR can often be reviewed in a few small gaps throughout the day, since small means minimum context needs to get loaded into "working memory" 🧠, means more potential time slots are big enough to make progress on the review.

On the other hand, a larger PR might not fit as easily, which unfortunately means longer reviews and longer delays. It can be especially inefficient if that delay holds up reviews of subsequent smaller PRs that might have been able to be reviewed earlier if they weren't "stuck" behind a larger PR.

We love your work Martin ♥️ (and the work of all of our community members), and we'll always do our best to review your PRs as efficiently and quickly as we can while maintaining our review quality standards. I just wanted to give some context and suggestions on how we might all benefit from small, targeted PRs 🙂

martinvuyk: ok, will do. I added these here because splitting them off would mean making this PR a draft until the other one is merged, since the functionality needs to edit `List.size`.

PS: just tell me when you'd like something to be split off, I also want to help you guys makes the whole endeavor faster. I'm still learning to contribute to OSS, I'm used to having merge access and making monster changes :rofl:.

ConnorGray: Thanks Martin 🙂

> I added these here because splitting them off would mean making this PR a draft until the other one is merged, since the functionality needs to edit List.size.

Am I understanding correctly that's because of visibility? If so, I think that other PR could be decoupled from this one by mutating `list.size` / `list._len` directly. `_len` makes a field "notionally" / "conventionally" private, but its still okay in special circumstances to mutate it directly 🙂  It just should be done with care.

> PS: just tell me when you'd like something to be split off, I also want to help you guys makes the whole endeavor faster. I'm still learning to contribute to OSS, I'm used to having merge access and making monster changes 🤣.

Thanks Martin, we're all learning together here 🙂  It's been an adjustment for me as well; in the past I've mostly worked on codebases where I was the sole contributor and could make sweeping changes quickly. Here's to us learning to make smaller changes and learning how to use the process most effectively 🍻 🙂 

I think a good rule of thumb is PRs should generally try to do only one thing, e.g. fix a bug, perform a refactoring, or add a new feature, but maybe not multiple in the same PR. That helps with reading the PR. E.g. if I know a PR is only supposed to be doing a refactoring, I'll read it with a different mindset than if I know its adding new (potentially unproved) code.

stdlib/src/base64/_b64encode.mojo
```diff
@@ -268,12 +250,7 @@ fn b64encode_with_buffers(
         ](
             nb_of_elements_to_load
         )
-        store_incomplete_simd(
-            result.unsafe_ptr() + len(result),
-            result_vector_with_equals,
-            nb_of_elements_to_store,
-        )
-        result.size += nb_of_elements_to_store
```
ConnorGray, line 276: Related to [our discussion about PR contents](https://github.com/modularml/mojo/pull/3814#discussion_r1878848233), I see that the changes in this file are a refactoring to use the new `append()` methods we talked about.

If this refactoring was a small part of the PR adding those new `append()` methods, I think that would be okay. Even though techincally that would be a PR both adding a new feature and doing a refactoring in the same PR, that can be okay as long as the overall size is still small. The "do one thing" idea is just a rule of thumb 🙂

docs/changelog.md
```diff
@@ -118,6 +118,7 @@ what we publish.
 - `StringRef` is being deprecated. Use `StringSlice` instead.
   - Changed `sys.argv()` to return list of `StringSlice`.
   - removed `StringRef.startswith()` and `StringRef.endswith()`
+- Direct access to `List.size` has been removed. Use the public API instead.
```
arthurevans, line 121: If people are using `size`, it may be because they're not aware of what other API they should be using. Can we list replacements here (or at least examples)? I think the main ones are `len(my_list)` and `resize()`?
```suggestion
- Direct access to `List.size` has been removed. Use the public API instead—for
  example, [`len(list)`] and [`list.resize()`](/mojo/stdlib/collections/list/List#resize)
```

arthurevans:
Not sure what the status of this PR is, but left a suggestion for the changelog entry just in case.

martinvuyk:
Hi @arthurevans thanks for the input, you're totally right that this needed some examples :)

ConnorGray:
!sync

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
Quick update here: I've fixed (almost) all of our internal use so this should land soon internally.  I'm OOO tomorrow, so if it doesn't land tonight/soon, it will land Friday.  Thanks, Martin!

PR #3822, opened 2024-11-28
[stdlib] [NFC] Fix revert python format with ruff inside stdlib folder

martinvuyk:
Revert d6fdf0cac5a987c030f5bc4da0f0bd9ba0ff112e , which used ruff formatter that is not the same as the mojo black formatter fork (I assume that's what was forked). It was making the public CI fail.

CC @keith


keith:
thanks for cc'ing me! I submitted https://github.com/modularml/mojo/pull/3832 to use ruff on CI as well

PR #3823, opened 2024-11-28
[stdlib] Fix allow implicit casting `MutableOrigin` -> `ImmutableOrigin` for `Span` and `StringSlice`

martinvuyk:
Fix allow implicit casting `MutableOrigin` -> `ImmutableOrigin` for `Span` and `StringSlice` by adding implicit constructors

### DIFF
--- a/mojo/stdlib/src/memory/span.mojo
+++ b/mojo/stdlib/src/memory/span.mojo
@@ -118,7 +118,12 @@ struct Span[
         alignment: The minimum alignment of the underlying pointer known statically.
     """
 
-    # Field
+    # Aliases
+    alias Mutable = Span[T, MutableOrigin.cast_from[origin].result]
+    """The mutable version of the `Span`."""
+    alias Immutable = Span[T, ImmutableOrigin.cast_from[origin].result]
+    """The immutable version of the `Span`."""
+    # Fields
     var _data: UnsafePointer[
         T,
         mut=mut,
@@ -131,6 +136,20 @@ struct Span[
     # ===------------------------------------------------------------------===#
     # Life cycle methods
     # ===------------------------------------------------------------------===#
+
+    @doc_private
+    @implicit
+    @always_inline("nodebug")
+    fn __init__(
+        other: Span[T, _],
+        out self: Span[T, ImmutableOrigin.cast_from[other.origin].result],
+    ):
+        """Implicitly cast the mutable origin of self to an immutable one.
+
+        Args:
+            other: The Span to cast.
+        """
+        self = rebind[__type_of(self)](other)
 
     @always_inline("builtin")
     fn __init__(
@@ -345,6 +364,15 @@ struct Span[
     # ===------------------------------------------------------------------===#
 
     @always_inline
+    fn get_immutable(self) -> Self.Immutable:
+        """Return an immutable version of this `Span`.
+
+        Returns:
+            An immutable version of the same `Span`.
+        """
+        return rebind[Self.Immutable](self)
+
+    @always_inline
     fn unsafe_ptr(
         self,
     ) -> UnsafePointer[
@@ -470,27 +498,6 @@ struct Span[
         """
         for element in self:
             element[] = value
-
-    fn get_immutable(
-        self,
-    ) -> Span[
-        T,
-        ImmutableOrigin.cast_from[origin].result,
-        address_space=address_space,
-        alignment=alignment,
-    ]:
-        """
-        Return an immutable version of this span.
-
-        Returns:
-            A span covering the same elements, but without mutability.
-        """
-        return Span[
-            T,
-            ImmutableOrigin.cast_from[origin].result,
-            address_space=address_space,
-            alignment=alignment,
-        ](ptr=self._data, length=self._len)
 
     fn swap_elements(
         self: Span[mut=True, T, alignment=alignment], a: UInt, b: UInt
--- a/mojo/stdlib/src/builtin/sort.mojo
+++ b/mojo/stdlib/src/builtin/sort.mojo
@@ -216,11 +216,10 @@ fn _quicksort[
 
     # Work with an immutable span so we don't run into exclusivity problems with
     # the List[Span].
-    var imm_span = span.get_immutable()
-    alias ImmSpan = __type_of(imm_span)
+    alias ImmSpan = span.Immutable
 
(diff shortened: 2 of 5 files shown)


nmsmith:
Small note: If we end up modelling origin mutation as an effect that functions perform, rather than today's approach where structs and functions are parameterized by `is_mutable` parameters, this PR would end up being redundant, because "mutable slices" and "immutable slices" (etc.) would have the same type. (This idea has been in discussion for a while.)

That's not to say this PR shouldn't be accepted. Maybe this is solving some near-term problems that people have been experiencing. I'm out of the loop on this.

martinvuyk:
Hi @nmsmith, I would love for this PR to become redundant. Parametric origins are a pain. This is mostly to facilitate work in function signatures and at call-site to allow using both types interchangeably on both ends and letting implicit construction do the job of casting

stdlib/src/utils/span.mojo
```diff
@@ -114,6 +118,28 @@ struct Span[
...
+    @doc_private
+    @implicit
+    @always_inline("nodebug")
+    fn __init__(out self: Self.mut, other: Self.immut):
+        """Implicitly cast the immmutable origin of self to a mutable one.
+
+        Args:
+            other: The Span to cast.
+        """
+        self = rebind[__type_of(self)](other)
```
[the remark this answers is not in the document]
martinvuyk: I didn't put much thought into it either. Now that you mention it it seems like a possible foot-gun.

The only use case I can think of is when trying to work around the compiler's restriction to having 2 mutable references to the same origin inside a function. An example (for `Pointer`) would be sorting algorithms's inner functions where they'd need 2 references to be compared and then swapped, and if they pointed to the same memory location the compiler would enforce that they should be immutable refs. Currently the sort module just uses `UnsafePointer` to avoid such restrictions.

I'm still not sure though whether we should offer this constructor even as an explicit one, it seems like encouraging function signature "contract breaking".

The more I think about it the less I like the idea, I'll revert the changes to leave only `Immutable` -> `Mutable` and if we find a solid use case we could always implement the counter direction.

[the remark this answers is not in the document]
ConnorGray: I agree that we could conceivable want an "unsafe" way to case an immutable origin to a mutable origin (we have that already in a less ergonomic form in `_lit_mut_cast`), but I agree its a special case, and we could leave it out of this PR for the time being.

Thanks for removing those for the time being Martin — I'll give this another look now! 🙂

stdlib/src/builtin/type_aliases.mojo
```diff
@@ -56,26 +56,42 @@ struct Origin[is_mutable: Bool]:
...
     ]
+    """The underlying MLIR type."""
 
     # ===-------------------------------------------------------------------===#
     # Fields
     # ===-------------------------------------------------------------------===#
 
-    var _mlir_origin: Self.type
+    var value: Self.type
+    """The underlying MLIR value."""
```
ConnorGray, line 66: I know we've used the convention of having a `value` field in other structs that wrap MLIR values, but IMO I think we'd be better moving away from that. (`value` isn't very readable at a glance since its such a generic term.)

My preference is that we leave this field name as `_mlir_origin` for the time being. I'm open to reconsidering that though, perhaps as part of a holistic effort to revist all the MLIR-wrapping types in the stdlib to be more consistent about this in general.

stdlib/src/os/path/path.mojo
```diff
@@ -563,7 +564,7 @@ fn expandvars[PathLike: os.PathLike, //](path: PathLike) -> String:
         The expanded path.
     """
     var path_str = path.__fspath__()
-    var bytes = path_str.as_bytes().get_immutable()
+    var bytes = __type_of(path_str.as_bytes()).immut(path_str.as_bytes())
```
ConnorGray, line 567: **Question** Is there a way we might avoid the duplication here? I can see the potential value in having aliases for mutable/immutable variants of the current `Span`/`StringSlice` type, but in a case like this, a "normal" method call seems more readable at a glance.

stdlib/src/builtin/type_aliases.mojo
martinvuyk: I can revert the change for this PR np. It was part of my attempt to do #3833, I don't have any string opinion on the name of the field itself either.

stdlib/src/os/path/path.mojo
martinvuyk: True, I removed the function mostly because I disliked the name. We could also change the method name to something like `.to_immut()` or directly `.immut()` (or leave that for another PR) WDYT?

mojo/stdlib/src/memory/span.mojo
```diff
@@ -471,23 +487,11 @@ struct Span[
...
-    ) -> Span[
-        T,
-        ImmutableOrigin.cast_from[origin].result,
-        address_space=address_space,
-        alignment=alignment,
-    ]:
-        """
-        Return an immutable version of this span.
+    @always_inline
+    fn immut(self) -> Self.ImmutSelf:
```
JoeLoser: Can we keep the old name of the get_immutable function for the purposes of this PR, please?

JoeLoser (commented):
@lsh do you mind taking a look at this? This came up during your talk the other day.

mojo/stdlib/src/memory/span.mojo
martinvuyk: ok

martinvuyk:
@JoeLoser @lsh I could delete the `get_immutable()` in this PR since it seems like it is not needed anymore. But I'll leave it to reduce the scope of the PR. For now the casting to immutable seems to work flawlessly :tada:.

JoeLoser:
> @JoeLoser @lsh I could delete the `get_immutable()` in this PR since it seems like it is not needed anymore. But I'll leave it to reduce the scope of the PR. For now the casting to immutable seems to work flawlessly 🎉.

Nice work!  Like you said, let's leave `get_immutable` around for the purposes of this PR so we can fix our own internal use and then remove it.  @lsh mentioned he would give a review of this PR today - thank you for working on this!!!

martinvuyk:
@lsh did the same name change here

lsh:
!sync

lsh:
@martinvuyk this PR now has a conflict, would you mind resolving it?

martinvuyk:
@lsh done

lsh:
!sync

PR #3824, opened 2024-11-28
[stdlib] fix: remove `DictEntry.__copyinit__` in `Dict` methods (`Pointer`, speed up)

rd4com:
Hello, this probably fixes some bug reports :partying_face: 
it is also now faster because we use `Pointer` and it removes `__copyinit__`

Many methods like `__getitem__` uses `_find_index`,
it was doing `__copyinit__` on entries while iterating

The result is also a speedup, especially on types that allocate memory,
best speedup is probably for `V:Arc` :+1: 

It should also speedup `**kwargs`!

### DIFF
--- a/stdlib/test/collections/test_dict.mojo
+++ b/stdlib/test/collections/test_dict.mojo
@@ -605,7 +605,6 @@ fn test_dict_setdefault() raises:
 
 def test_dict_entries_del_count():
     alias dict_type = Dict[Int, ValueDestructorRecorder]
-    var iter = 0
     for entries_to_insert in range(16):
         var x = List[Int]()
         var y = dict_type()
@@ -637,8 +636,6 @@ def test_dict_entries_del_count():
         __type_of(y).__del__(y^)
         assert_equal(len(x), entries_to_insert)
         assert_equal(result, expected_result)
-        iter += 1
-    assert_equal(iter, 16)
 
 
 def main():
--- a/stdlib/test/collections/test_dict.mojo
+++ b/stdlib/test/collections/test_dict.mojo
@@ -15,8 +15,9 @@
 from collections import Dict, KeyElement, Optional
 from collections.dict import OwnedKwargsDict
 
-from test_utils import CopyCounter
+from test_utils import CopyCounter, ValueDestructorRecorder
 from testing import assert_equal, assert_false, assert_raises, assert_true
+from memory import UnsafePointer
 
 
 def test_dict_construction():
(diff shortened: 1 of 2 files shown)


rd4com:
#### 🔥 benchmarks

Theses are some small simple benchmarks with `Time.now()`,
speedups increases with the size of the dictionary ❤️‍🔥

```
690781 3264000 (1.13x)
885908 6528000 (1.19x)
2173504 6593800 (3.36x)
1645963 6659600 (2.32x)
316705 6725400 (7.3x)
```

```mojo
from time import now
from collections import Dict

alias iteration_size = 1<<8
def main():
    var result: Int=0
    var start = now()
    var stop = now()

    
    small = Dict[Int, Int]()
    start = now()
    for x in range(100):
            small[i]=i
            result += small.pop(i)
    stop = now()
    print(stop-start, result)
    
    start = now()
    for x in range(100):
        small = Dict[Int, Int]()
            small[i]=i
            result += small[i]
    stop = now()
    print(stop-start, result)
    
    start = now()
    for x in range(100):
        small2 = Dict[Int, String]()
        @parameter
            alias tmp = str(i)
            small2[i]=tmp
            result += len(small2[i])
    stop = now()
    print(stop-start, result)
    
    small2 = Dict[Int, String]()
    start = now()
    for x in range(100):
        @parameter
            alias tmp = str(i)
            small2[i]=tmp
            result += len(small2.pop(i))
    stop = now()
    print(stop-start, result)
    
    small2 = Dict[Int, String]()
    @parameter
        alias tmp = str(i)
        small2[i]=tmp
    start = now()
    for x in range(100):
            result += len(small2[i])
    stop = now()
    print(stop-start, result)
```

stdlib/test/collections/test_dict.mojo
```diff
@@ -602,6 +603,44 @@ fn test_dict_setdefault() raises:
...
+            y[i] = ValueDestructorRecorder(i, UnsafePointer.address_of(x))
+        assert_equal(len(x), 0)
+        for i in range(entries_to_insert):
+            result += y.pop(i).value
+        assert_equal(len(x), entries_to_insert)
+        __type_of(y).__del__(y^)
+        assert_equal(len(x), entries_to_insert)
+        assert_equal(result, expected_result)
+        iter += 1
+    assert_equal(iter, 16)
```
JoeLoser, line 641: **Question** What's the value in bumping `iter` and then checking the value here?  `assert_*` are "hard errors" and so the stop would executing if any of them failed before we exhausted all the iterations in the fixed for loop.

JoeLoser (approved):
Nice work!

stdlib/test/collections/test_dict.mojo
rd4com: Yep sorry it was after doing `@parameter loops` for another PR
(to test if the range iterate all the alias types)

Should it be removed for this one ?

rd4com:
✅ Changes implemented
- [review] remove `iter` for `test_dict_entries_del_count()`
- rebase

rd4com:
Hello,
closing this PR which was split into:
- #4162 

We got a more recent one by bgreni :+1:

PR #3832, opened 2024-12-02
[CI] Format python with ruff

keith:
Ruff is a faster and more modern python formatter. This swaps out using
`mojo format` for python with ruff.

### DIFF
--- a/stdlib/test/lit.cfg.py
+++ b/stdlib/test/lit.cfg.py
@@ -11,6 +11,8 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 
+# ruff: noqa
+
 import os
 import shutil
 from pathlib import Path
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
         name: mojo-format
         entry: mojo format
         language: system
-        files: '\.(mojo|🔥|py)$'
+        files: '\.(mojo|🔥)$'
         stages: [commit]
       - id: check-docstrings
         name: check-docstrings
@@ -24,3 +24,9 @@ repos:
     hooks:
     - id: markdownlint
       args: ['--config', 'stdlib/scripts/.markdownlint.yaml', '--fix']
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.8.0
+    hooks:
+      - id: ruff
+        args: [--fix]
+      - id: ruff-format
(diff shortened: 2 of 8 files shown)


stdlib/test/lit.cfg.py
```diff
@@ -48,9 +50,7 @@ def has_not():
     config.test_source_root = os.path.dirname(__file__)
 
     # test_exec_root: The root path where tests should be run.
-    config.test_exec_root = os.path.join(
-        config.modular_obj_root, "mojo-stdlib", "test"
-    )
+    config.test_exec_root = os.path.join(config.modular_obj_root, "mojo-stdlib", "test")
```
martinvuyk, line 53: This is above 80 characters for example

martinvuyk (commented):
@keith I think this will cause conflicts for every PR until `mojo format` behaves the same. I normally run `mojo format ./stdlib/` for example. Are there plans to change mojo format to behave like ruff instead of black formatter?

stdlib/test/lit.cfg.py
keith: fixed line length setting

martinvuyk:
@keith the tool having a static config option already sold me on wanting to use it :smile: (although now I read that black formatter also has that, I had no idea). Is there another option to avoid the difference in newline in PR [#3822 (file)](https://github.com/modularml/mojo/pull/3822/files) ?

keith:
I think we might just want to stop `mojo format` from running on python, but I'm hoping @JoeLoser can provide an opinion here

martinvuyk:
> I think we might just want to stop `mojo format` from running on python, but I'm hoping @JoeLoser can provide an opinion here

Just an idea: add a flag that is default mojo only where you can also make mojo format run on any extension e.g. `mojo format -e py -e ipynb`

walter-erquinigo:
Hi! I'm the maintainer of `mojo format` and we never really intended to have `mojo format` be used on python files =P 
So, the change that we'll do is to prevent it from touching any non-mojo file. Hopefully I'll land that this week.

JoeLoser (approved):
LGTM but you'll need to split this across two PRs since the `.pre-commit-config.yaml` isn't managed by Copybara whereas the rest of the files are I believe.  Also, in a follow-up PR, can you add Ruff to the contributing docs for formatting Python-related parts to avoid future confusion?

patrickdoc:
!sync (for testing)

patrickdoc:
!sync (testing 2)

walter-erquinigo:
https://github.com/modularml/modular/pull/52586 should fix the mblack part

patrickdoc:
!sync (testing repo move)

patrickdoc:
!sync (one more time)

ematejska:
@keith Please re-open off of a cloned repo's branch going forward if desirable.

PR #3835, opened 2024-12-04
[stdlib] Add `ListLiteral.__getitem__`

rd4com:
Hello,
just a small pr to add `__getitem__` to `ListLiteral` :+1: 

For example:
```mojo
def main():
    x = [0, 1.0, "two"]
    print(
        x[0] == 0,
        x[1] == 1.0,
        x[2] == "two"
    )
```

Does it needs the `@always_inline("nodebug")` like `Tuple` ?


JoeLoser:
The `ListLiteral` type as-before was removed recently, so closing this.

Issue #3653, opened 2024-10-12
[Feature Request] [stdlib] [proposal] Have all `fn __getitem__(self, span: Slice) -> Self` return `Iterator[Self]` instead


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Once we have Iterators designed or right now using each type's specialized iterator, change all `fn __getitem__(self, span: Slice) -> Self` to return the iterator and add constructors that take that iterator so that implicit conversion is possible.

### What is your motivation for this change?

Te idea came to me while reading PR #3650 which allocates a buffer to return a negatively stepped slice over a readonly view of data that the `Span` is built upon which doesn't make much sense but there is really no alternative since the function should return an instance of `Self`.

Iterating with a particular step and in any particular direction is much simpler and doable without copying buffers. There is also the added benefit that when people do:
```mojo
var string = String("some string")
print(string[4:]) # no copy since it would just be written to the file stream
var amount_of_s = 0
for s in string[4:]: # no allocation needed here
    if s == "s":
        amount_of_s += 1
```

when building an instance of the type with that iterator it would only allocate when the type is an owning type or when `step != 1` for non-owning types
```mojo
var string = String("some string")
var s2 = string[4:] # no copy
var s3 = String(s2) # copies all values, since abs(step) == 1 copies directly
var s4 = StringSlice(s2) # no copy
# can use s4 methods which are mostly the same as the String API
# but without paying the cost of allocation
var s5 = String(string[4::2]) # copies all values by iterating
var s6 = StringSlice(string[4::-1]) # copies reversing with SIMD (once it's implemented in List)

var items = List[UInt8](1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
var i1 = items[1::2] # iterator over even numbers
var i2 = List[UInt8](i1) # should use SIMD strided load
```

This will also (I think) benefit performance in the case of `String` slicing which currently does not do it by unicode codepoints but will in the future, which needs to count all previous codepoints and then from then count until  `start - end` of the slice is reached, allocate a buffer for the new string + new null terminator, and return. This iterator approach would mean just counting until the beginning of the slice and returning an iterator with the given `start - end` length and the given step.

### Any other details?

[...]

PR #3839, opened 2024-12-05
[stdlib] Add `List.map(fn(mut T)->None)`

rd4com:
Hello, the pr is an implementation for `map`,
we could have used parameters, but it is for a first pythonic experience :snake:
(progressive exposure to complexity and build on current python skills already learned)

- 🆕 `map`
  Apply a function to every elements of `self` and returns a new `List`
  ```mojo
  fn MyFunc(mut e: Int):
      e+=1
  
  var MyList = List(0, 1, 2).map(MyFunc)
  
  print(
      MyList[0] == 1,
      MyList[1] == 2,
      MyList[2] == 3,
  )
  ```

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -930,6 +930,35 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         """
         return self.data
 
+    fn map(ref self, func: fn (mut T) -> None) -> Self:
+        """Map the values of the list into a new list trough a function.
+
+        Args:
+            func: The function used on every elements to create the new list.
+
+        Returns:
+            A new `List` created by calling `func` on every elements of `self`.
+
+        For example:
+        ```mojo
+        fn MyFunc(mut e: Int):
+            e+=1
+
+        var MyList = List(0, 1, 2).map(MyFunc)
+
+        print(
+            MyList[0] == 1,
+            MyList[1] == 2,
+            MyList[2] == 3,
+        )
+        ```.
+
+        """
+        var tmp = self
+        for i in tmp:
+            func(i[])
+        return tmp
+
 
 fn _clip(value: Int, start: Int, end: Int) -> Int:
     return max(start, min(value, end))
--- a/stdlib/test/collections/test_list.mojo
+++ b/stdlib/test/collections/test_list.mojo
@@ -924,6 +924,21 @@ def test_list_repr():
     assert_equal(empty.__repr__(), "[]")
 
 
+def test_list_map():
+    fn MyFunc(mut e: Int):
+        e += 1
+
+    var lst = List(0, 1, 2).map(MyFunc)
+    for e in range(len(lst)):
+        assert_equal(lst[e], e + 1)
+
+    lst = List(0, 1, 2)
+    var lst2 = lst.map(MyFunc)
+    for e in range(len(lst)):
+        assert_equal(lst[e], e)
+        assert_equal(lst2[e], e + 1)
+
+
 # ===-------------------------------------------------------------------===#
 # main
 # ===-------------------------------------------------------------------===#
@@ -962,3 +977,4 @@ def main():
     test_indexing()
     test_list_dtor()
     test_list_repr()
+    test_list_map()


owenhilyard:
In order to get feature parity with python, we probably want lazy iterators. My vote would be for copying the Rust API where you need to do a `.collect[collection: FromIterator]()`, and then you can implement that trait for all collections. This also means the slightly more pythonic version (`List([1,2,3,4].map(lambda x: x ** 2))`) will work fine. The other reason I want laziness is to allow for the compiler to do iterator fusion, which lets a long series of maps, filters, and other iterator ops compile down into the same code as the equivalent for loop.

Could we do something like making that return a `Map[origin, _ListIter, fn]` and which itself is an iterator (which may be doable to implement using aliases in traits)? At the moment, I don't see a good way to abstract over owned vs read/mut cleanly unless we force read/mut to `Pointer`, which may be sub-optimal but start a conversation about first class references. I think we 100% want to use parameters so that the API can be compiled more efficiently, because if we do it using runtime function pointers we end up leaning on different parts of LLVM which I don't think are quite as guaranteed to work (leading to fragile optimizations).

martinvuyk:
I think it's a good idea to deviate form Python and add `map`, `filter`, `sum`, `join`, `reduce` to `List` (one of the first things I did for my `Array` [implementation](https://github.com/martinvuyk/forge-tools/blob/main/src/forge_tools/collections/array.mojo)). 

> we probably want lazy iterators

+1 I think it's worth making it an iterator return and not a copy of the data.

We should also aim to implement [Python's builtin map function](https://docs.python.org/3/library/functions.html#map) (also `sum`) at some point. These methods are very good because they're generic, but the API is not at all achievable until we have the `Iterator` traits and APIs defined.

> My vote would be for copying the Rust API where you need to do a .collect[collection: FromIterator]()

I think this will dovetail nicely with #3653, which would mean for people who want to just declare the variable type they can, and those who want to use functional style collect also can :)

---

The approach I'll propose can become ugly very fast once we do it for many types, but I think it's worth trying to implement it and testing it out:

What if we implement one specific type of iterator for each op, for this PR the case would be `_ListMapIter` that yields the elements. We could then do the same for `sum`, `filter`, etc. (or just implement `map` and test it out first)

Or, we could just let the function return a new List and come back to iterate and fix it once we have Iterators better defined. WDYT?

owenhilyard:
> We should also aim to implement [Python's builtin map function](https://docs.python.org/3/library/functions.html#map) (also `sum`) at some point. These methods are very good because they're generic, but the API is not at all achievable until we have the `Iterator` traits and APIs defined.

I agree, the builtin map can take an iterable (remember that iterators are iterable and the `.iter()` is a noop) 
 
> The approach I'll propose can become ugly very fast once we do it for many types, but I think it's worth trying to implement it and testing it out:
> 
> What if we implement one specific type of iterator for each op, for this PR the case would be `_ListMapIter` that yields the elements. We could then do the same for `sum`, `filter`, etc. (or just implement `map` and test it out first)

I'd rather have a generic iterator, so that `List[T].map` returns `IterMap[ListIter[T], map_fn]`. If we really need the extra performance, we can also specialize map on a bunch of things like the C++ ranges concepts, for instance contiguous buffers, and potentially have traits for iterators that can efficiently yield SIMD chunks. Ideally, I'd like something like Rust's `impl Iterator<Output=T>`, where we can make very light guarantees about the behavior to the user but the compiler can see it's actually an `_ListIter[T]`, which implements `Iterator[T] + FixedSizeSIMDIterator[T] + ContigiousBufferIterator[T] + ...` and use that when determining which specializations to choose.

stdlib/src/collections/list.mojo
```diff
@@ -930,6 +930,35 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         """
         return self.data
 
+    fn map(ref self, func: fn (mut T) -> None) -> Self:
```
lsh, line 933: **thought**: The API you're describing seems more like `for_each` than `map`.
`fn(mut T) -> None` is a common `for_each` signature, whereas `map` tends to be `fn[U: CollectionElement](T) -> U`.

Examples:
* Rust [for_each](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each) and [map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map)
* Swift [for_each](https://developer.apple.com/documentation/swift/array/foreach(_:)) and [map](https://developer.apple.com/documentation/swift/sequence/map(_:))
* C++ [ranges::for_each](https://en.cppreference.com/w/cpp/algorithm/ranges/for_each) and [ranges::transform](https://en.cppreference.com/w/cpp/algorithm/ranges/transform)

martinvuyk: Yes I think it was just a mistake in assuming the output will be the same type as the input collection type

lsh:
@owenhilyard I don't want to get _too_ into the weeds on this and derail the main topic of this PR, but I tend to think the cleanest API would be a constructor for `List` from `Iterator`. Due to the cultural linkage to Python, there will probably be a skew towards in favor of using comprehensions over `map`/`filter`. I also personally think `List[Int](x ** 2 for x in [1,2,3] if x <= 4)` looks nicer than `List[Int](1,2,3).map[lambda x: x ** 2]().filter[lambda x: x <= 4]()`, but that's just a personal preference without taking this to the team.

stdlib/src/collections/list.mojo
lsh: Not exactly. `for_each` tends to mutate in place, while `map` tends to create a new allocation. Users associate both APIs with those characteristics in my opinion.

martinvuyk: I meant the OP meant map as in map a func to return a new instance with copied values changed, just that he forgot the use case where the collection type of the result might be different.

personal note: I'd actually like us to call `for_each`  `apply` because it's a name used in the pandas and numpy ecosystem and is IMO much more intuitive (see PR #3877)

owenhilyard:
> @owenhilyard I don't want to get _too_ into the weeds on this and derail the main topic of this PR, but I tend to think the cleanest API would be a constructor for `List` from `Iterator`.

+1 to `List` from `Iterator`, I think it's useful to have both that and a `.collect()` postfix option. 

> Due to the cultural linkage to Python, there will probably be a skew in favor of using comprehensions over `map`/`filter`. I also personally think `List[Int](x ** 2 for x in [1,2,3] if x <= 4)` looks nicer than `List[Int](1,2,3).map[lambda x: x ** 2]().filter[lambda x: x <= 4]()`, but that's just a personal preference without taking this to the team.

I agree short list comprehensions are cleaner, but I find they get messy once you want to do `[expensive_function(j[1]) for j in [get_filter_key(i) for i in foo] if filter_function(j[0])]` or similar longer chains. Python's lambda syntax does make it worse, and I hope we can find an alternative since I don't think I've ever talked to someone who likes it.

rd4com:
The `for_each` suggestion makes sense @lsh  :+1:  
after checking python, maybe `map` could be a function to match python, wdyt?
```python
items = [1, 2, 3, 4, 5]
inc = list(map(lambda x: x+1, items))
```
That way, there is more time to think about the chaining thing ?

owenhilyard:
> The `for_each` suggestion makes sense @lsh 👍 after checking python, maybe `map` could be a function to match python, wdyt?
> 
> ```python
> items = [1, 2, 3, 4, 5]
> inc = list(map(lambda x: x+1, items))
> ```
> 
> That way, there is more time to think about the chaining thing ?

I prefer to do method chaining, since it means that I'm not reading my code inside-out. In terms of this particular PR, I agree it should be `for_each`, `map` is a transformation function.

martinvuyk:
> I'd actually like us to call `for_each` `apply` because it's a name used in the pandas and numpy ecosystem and is IMO much more intuitive (see PR https://github.com/modularml/mojo/pull/3877)

I want to repeat this point, I strongly dislike this being called `for_each`. Either call it `map` and fix the `func: fn[T2](T1) -> T2` or consider using `apply` and make the operation inplace. I use numpy and pandas quite heavily, I've used Javascript a bunch as well, and I very much prefer thinking of applying a function than `forEach`. There is also the not-trivial problem that the abstraction of `for_each` falls appart when dealing with matrices, tensors, and vectors; with `apply` it's easy to think about axes. It is also a word that is much more used in the ML space.

owenhilyard:
> I want to repeat this point, I strongly dislike this being called `for_each`. Either call it `map` and fix the `func: fn[T2](T1) -> T2` or consider using `apply` and make the operation inplace. I use numpy and pandas quite heavily, I've used Javascript a bunch as well, and I very much prefer thinking of applying a function than `forEach`. There is also the not-trivial problem that the abstraction of `for_each` falls appart when dealing with matrices, tensors, and vectors; with `apply` it's easy to think about axes. It is also a word that is much more used in the ML space.

`for_each` or `foreach` has very old roots in FP, I think more people are going to find issues with `apply` than `for_each`.

martinvuyk:
> `for_each` or `foreach` has very old roots in FP, I think more people are going to find issues with `apply` than `for_each`.

I'd argue that many more people use Python (pandas, numpy, ML frameworks) and R ([it's a builtin](https://www.geeksforgeeks.org/apply-lapply-sapply-and-tapply-in-r/)) than Haskell...

owenhilyard:
> > `for_each` or `foreach` has very old roots in FP, I think more people are going to find issues with `apply` than `for_each`.
> 
> I'd argue that many more people use Python (pandas, numpy, ML frameworks) and R ([it's a builtin](https://www.geeksforgeeks.org/apply-lapply-sapply-and-tapply-in-r/)) than Haskell...

The other side of the fence is JS, TS, Rust, C++, Java, C#, most FP languages, Scala, Kotlin, and Swift, among others. Ocaml uses `for_all`. Ruby uses `.each`.

martinvuyk:
> > > `for_each` or `foreach` has very old roots in FP, I think more people are going to find issues with `apply` than `for_each`.
> > 
> > 
> > I'd argue that many more people use Python (pandas, numpy, ML frameworks) and R ([it's a builtin](https://www.geeksforgeeks.org/apply-lapply-sapply-and-tapply-in-r/)) than Haskell...
> 
> The other side of the fence is JS, TS, Rust, C++, Java, C#, most FP languages, Scala, Kotlin, and Swift, among others. Ocaml uses `for_all`. Ruby uses `.each`.

Ok that's a long list :laughing:. But I still stand by my opinion, I think `for_each` just comes from the procedural mentality of for loops and it does not adjust well to tensors. I also don't agree with it returning a copy of the data and then modifying it, it should be inplace. With `map` and `apply` we have much better readability IMO:
```mojo
fn add_two(value: Int) -> Int:
    return value + 2

fn bigger_smaller_or_eq(value: Int) -> String:
    return "bigger" if value > 4 else "smaller" if value != 4 else "equal"

fn main():
    items = List[Int](1, 2, 3)
    print(items.map(add_two)) # 3 4 5
    items.apply(add_two)
    print(items) # 3 4 5
    # items.apply(bigger_or_smaller) # won't compile
    print(items.map(bigger_smaller_or_eq)) # smaller equal bigger
```

rd4com:
Hello @lsh ,
do we close this and wait for iterator traits/generics ?

rd4com:
Hello @lsh ,
closing this PR, generic iterators are better :+1:

PR #3825, opened 2024-11-29
[stdlib] Micro-optimize string `splitlines` and `isnewline`

martinvuyk:
## Micro-optimize string `splitlines` and `isnewline`.

Benchmark results:

CPU: Intel® Core™ i7-7700HQ

improvement metric: markdown percentage improvement (`(old_value - new_value) / old_value`)

Average improvement: 1.2% . Many of the discrepancies are likely because of runtime (nano-second scale) pipelining (and other) differences during testing. The most robust estimates are the biggest tests which show around 4.2% improvement.

|Name                          | old_value (ms)           | new_value (ms)      | improvement|
|:-----------------------------|---------------------:|----------------:|------:|
|`bench_string_splitlines[10]` | 0.000100497059839818 | 0.00010261460376 | -2.11% |
|`bench_string_splitlines[30]` | 0.000171998352418428 | 0.000174028055158599 | -1.18% |
|`bench_string_splitlines[50]` | 0.000264958575880379 | 0.000251598931395228 | 5.04% |
|`bench_string_splitlines[100]` | 0.000417350538881514 | 0.000415764808815687 | 0.38% |
|`bench_string_splitlines[1000]` | 0.00320600819966129 | 0.00322517645847026 | -0.60% |
|`bench_string_splitlines[10000]` | 0.0302790238107332 | 0.0305872863513719 | -1.02% |
|`bench_string_splitlines[100000]` | 0.319274428592796 | 0.304513179279761 | 4.62% |
|`bench_string_splitlines[1000000]` | 3.24098357344348 | 3.10220707742267 | 4.28% |

PR #3843, opened 2024-12-07
[stdlib] [NFC] Fix docstrings for `string.strip()` function family

martinvuyk:
Fix docstrings for `string.strip()` function family.

Was deleted in b0094d3d9dfc6e38154393862098c22a0c1dec89. @lsh I really don't want that documentation deleted, this is very often not included and caused me quite a couple of headaches in finding the correct information and backward compatibility reasons for using those characters. And we should be explicit that this method does not yet support full Python spaces until we can run the methods with `.isspace()` at compile time.

### DIFF
--- a/stdlib/src/builtin/string_literal.mojo
+++ b/stdlib/src/builtin/string_literal.mojo
@@ -889,8 +889,9 @@ struct StringLiteral(
         return str(self).islower()
 
     fn strip(self) -> String:
-        """Return a copy of the string literal with leading and trailing whitespaces
-        removed.
+        """Return a copy of the string literal with leading and trailing
+        whitespaces removed. This only takes ASCII whitespace into account:
+        `" \\t\\n\\v\\f\\r\\x1c\\x1d\\x1e"`.
 
         Returns:
             A string with no leading or trailing whitespaces.
@@ -922,7 +923,9 @@ struct StringLiteral(
         return str(self).rstrip(chars)
 
     fn rstrip(self) -> String:
-        """Return a copy of the string with trailing whitespaces removed.
+        """Return a copy of the string with trailing whitespaces removed. This
+        only takes ASCII whitespace into account:
+        `" \\t\\n\\v\\f\\r\\x1c\\x1d\\x1e"`.
 
         Returns:
             A copy of the string with no trailing whitespaces.
@@ -941,7 +944,9 @@ struct StringLiteral(
         return str(self).lstrip(chars)
 
     fn lstrip(self) -> String:
-        """Return a copy of the string with leading whitespaces removed.
+        """Return a copy of the string with leading whitespaces removed. This
+        only takes ASCII whitespace into account:
+        `" \\t\\n\\v\\f\\r\\x1c\\x1d\\x1e"`.
 
         Returns:
             A copy of the string with no leading whitespaces.
--- a/stdlib/src/collections/string.mojo
+++ b/stdlib/src/collections/string.mojo
@@ -1919,6 +1919,7 @@ struct String(
         """Return a copy of the string with leading and trailing characters
         removed.
 
+
         Args:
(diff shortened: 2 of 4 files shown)


owenhilyard:
+1, knowing this doesn't handle all the various whitespace-like characters in unicode is important information. 

Does it make sense to have `.strip()` handle unicode and have a separate `.strip_ascii()` which can be more heavily vectorized for people who know they only have ascii text and want performance?

martinvuyk:
> Does it make sense to have `.strip()` handle unicode and have a separate `.strip_ascii()` which can be more heavily vectorized for people who know they only have ascii text and want performance?

@owenhilyard  We could add such a helper function inside `string_slice.mojo`, but I've found that for several cases having a sequence of scalar operations to check the value range is faster. I imagine it's because it's more easily pipelineable in the ALUs than the latency of doing `any(asci_whitespace_vec == byte_val)`. I haven't benchmarked the vectorized approach for #3528 but I think it might give the same results as in #3825 where it seems like the amount of branching makes the latency of the ops be expensive on branch prediction fail or because of the amount of instructions when implementing a branchless version (I'm still waiting for #3528 to be merged to experiment further, `split()` might have a bigger proportion of separator/text than `splitlines()` so the results might be different).

Both the vectorized and scalar versions are quite trivial to implement so I think people would reimplement them anyway.

I was also thinking that one day we might implement (or an external library for that matter) an `ASCIIString` which has all of the optimizations.

owenhilyard:
@martinvuyk Given that we're likely going to eventually have hand-written versions of functions like this per arch, I wonder if it makes sense to have small and large string versions of the code, instead of trying to make one algorithm work for someone who just mmaped a 20 GB csv and someone else who wants to split a single line of 4 comma delimited floats. We can try to make a set of mostly branchless loops for various string sizes, and potentially different versions for SIMD implementations with masking like AVX-512 and SVE.

For instance, on ASCII text you could vectorize each individual compare and OR them together (ex: '\t' <= chunk <=  '\r' or chunk == ' ') to make a mask, invert the mask, then use `llvm.experimental.vector.compress` and a bitcount (ex: popcnt) on the mask to figure out what your store width should be back into the string buffer (either the source string or an output buffer).

martinvuyk:
> Given that we're likely going to eventually have hand-written versions of functions like this per arch, I wonder if it makes sense to have small and large string versions of the code, instead of trying to make one algorithm work for someone who just mmaped a 20 GB csv and someone else who wants to split a single line of 4 comma delimited floats.

@owenhilyard  For very large mmaped files, I was thinking of eventually implementing a lazy version of the algorithm that works on `_StringSliceIter` and returns another iterator (I also want to do the same for splitlines). 

The main problem of making this vectorized is the unicode separators (and the damned `\r\n`), unless we have UTF-32 encoding I don't know how to vectorize the iteration and checking for UTF-8. For checking whether they are separators I implemented a branchless version but it was 30% slower (using SIMD ops, might have to test again with sequential ones), to iterate you can have different length of characters so there will be a checking length of the last char and masking cost at each step. There might be a point at which the cost is worth it, but I'd need to invest quite some time into development + benchmarking to be sure (time which I'd prefer investing into vectorizing `List` once #3810 lands, and many other string methods which are suboptimal).

> For instance, on ASCII text you could vectorize each individual compare and OR them together (ex: '\t' <= chunk <= '\r' or chunk == ' ') to make a mask, invert the mask, then use llvm.experimental.vector.compress and a bitcount (ex: popcnt) on the mask to figure out what your store width should be back into the string buffer (either the source string or an output buffer).

Wow, yeah that might be useful for csv parsers that know that the separators can be any ASCII whitespace char. The `split(sep: String)` algorithm is already vectorized. I think that can be left for an external library to implement though (unless we go all in on implemeting an `ASCIIString` type)

JoeLoser (approved):

PR #3844, opened 2024-12-07
[docs] Prevent unwanted Latex interpretation in `Path` docs

jjvraw:
As title, see [docs](https://docs.modular.com/mojo/stdlib/pathlib/path/Path#home).


JoeLoser (approved):
Good spot, thank you for the fix!

JoeLoser:
!sync

PR #3851, opened 2024-12-09
[stdlib] Deque - don't use non-self out args

avitkauskas:
As the original author of the Deque, I would like to propose to not use the `out` args instead of `-> ElementType as element` in `pop()` and `popleft()` methods.

The named return variable was just a simple convenience in these functions and not needed at all.

In general, having something in the args list of the function that you do not pass to the function at the call site looks weird (except for constructors). In my opinion, it makes the function signature just less clear to read and the return type of the function unnecessarily hidden inside the args list.

So, let's just use normal return statements here and have a clear function signatures.

### DIFF
--- a/stdlib/src/collections/deque.mojo
+++ b/stdlib/src/collections/deque.mojo
@@ -794,7 +794,7 @@ struct Deque[ElementType: CollectionElement](
 
         return (self._data + self._head)[]
 
-    fn pop(mut self, out element: ElementType) raises:
+    fn pop(mut self) raises -> ElementType:
         """Removes and returns the element from the right side of the deque.
 
         Returns:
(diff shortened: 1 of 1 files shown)


avitkauskas:
Pre-commit fails because of reformating of `stdlib/test/python/my_module.py` file - nothing to do with this commit.
Should I reformat it and push again anyway?

JoeLoser:
> Pre-commit fails because of reformating of `stdlib/test/python/my_module.py` file - nothing to do with this commit. Should I reformat it and push again anyway?

Don't worry about this right now. @keith and @walter-erquinigo are looking into it.  See https://github.com/modularml/mojo/pull/3832.

JoeLoser (approved):

PR #3852, opened 2024-12-09
[stdlib] Add vectorized `Span` iterator

martinvuyk:
Add vectorized `Span` iterator

Examples:

```mojo
var iterator = iter(Span(List[UInt8](1, 2, 3, 4, 5, 6, 7, 8, 9, 10)))
var has_a_10 = False

for v in iterator.vectorized():
    has_a_10 |= any(v == 10)

for s in iterator:
    has_a_10 |= s[] == 10

print(has_a_10) # True
```


martinvuyk:
I think this might be better as a `List[T1].map(func: fn (T1) -> T2) -> Iterator[T2]` API

PR #3854, opened 2024-12-10
[stdlib] Add `List[Scalar[D]]` append `SIMD` and `Span[Scalar[D]]`

martinvuyk:
Add `List[Scalar[D]]` append `SIMD` and `Span[Scalar[D]]`

split off from #3814. This is needed to enable efficient appending of scalar value sequences to a `List` without having to resort to `UnsafePointer` manually.

### DIFF
--- a/stdlib/src/collections/list.mojo
+++ b/stdlib/src/collections/list.mojo
@@ -493,38 +493,42 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         self.capacity = new_capacity
 
     fn append(mut self, owned value: T):
-        """Appends a value to this list. If there is no capacity left, resizes
-        to twice the current capacity. Except for 0 capacity where it sets 1.
+        """Appends a value to this list.
 
         Args:
             value: The value to append.
+
+        Notes:
+            If there is no capacity left, resizes to twice the current capacity.
+            Except for 0 capacity where it sets 1.
         """
-        if len(self) >= self.capacity:
+        if self.size >= self.capacity:
             self._realloc(self.capacity * 2 + int(self.capacity == 0))
-        (self.data + len(self)).init_pointee_move(value^)
+        (self.data + self.size).init_pointee_move(value^)
         self.size += 1
 
     fn append[
         D: DType, //
     ](mut self: List[Scalar[D], *_, **_], value: SIMD[D, _]):
-        """Appends a vector to this list. If there is no capacity left, resizes
-        to `len(self) + value.size`.
+        """Appends a vector to this list.
 
         Parameters:
             D: The DType.
 
         Args:
             value: The value to append.
+
+        Notes:
+            If there is no capacity left, resizes to `len(self) + value.size`.
         """
-        self.reserve(len(self) + value.size)
-        (self.data + len(self)).store(value)
+        self.reserve(self.size + value.size)
+        (self.data + self.size).store(value)
         self.size += value.size
 
     fn append[
         D: DType, //
     ](mut self: List[Scalar[D], *_, **_], value: SIMD[D, _], count: Int):
-        """Appends a vector to this list. If there is no capacity left, resizes
-        to `len(self) + count`.
+        """Appends a vector to this list.
 
         Parameters:
             D: The DType.
@@ -532,26 +536,31 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         Args:
             value: The value to append.
             count: The ammount of items to append.
+
+        Notes:
+            If there is no capacity left, resizes to `len(self) + count`.
         """
-        self.reserve(len(self) + count)
+        self.reserve(self.size + count)
         var v_ptr = UnsafePointer.address_of(value).bitcast[Scalar[D]]()
-        memcpy(self.data + len(self), v_ptr, count)
+        memcpy(self.data + self.size, v_ptr, count)
         self.size += count
 
     fn append[
         D: DType, //
     ](mut self: List[Scalar[D], *_, **_], value: Span[Scalar[D]]):
-        """Appends a Span to this list. If there is no capacity left, resizes
-        to `len(self) + len(value)`.
+        """Appends a Span to this list.
 
         Parameters:
             D: The DType.
 
         Args:
             value: The value to append.
+
+        Notes:
+            If there is no capacity left, resizes to `len(self) + len(value)`.
         """
-        self.reserve(len(self) + len(value))
-        memcpy(self.data + len(self), value.unsafe_ptr(), len(value))
+        self.reserve(self.size + len(value))
+        memcpy(self.data + self.size, value.unsafe_ptr(), len(value))
         self.size += len(value)
 
     fn insert(mut self, i: Int, owned value: T):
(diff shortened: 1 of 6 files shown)


stdlib/src/collections/list.mojo
```diff
@@ -493,16 +493,67 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         self.capacity = new_capacity
 
     fn append(mut self, owned value: T):
-        """Appends a value to this list.
+        """Appends a value to this list. If there is no capacity left, resizes
+        to twice the current capacity. Except for 0 capacity where it sets 1.
```
ConnorGray, line 497: **Suggestion** I think it might be more readable at-a-glance if we split this description across multiple lines so that the single-line summary stays short and sweet. The allocation strategy could instead be the next paragraph in the docs 🙂 

I.e.

```mojo
"""Appends a value to this list.

If there is no capacity left, resizes to twice the current
capacity. Except for 0 capacity where it sets 1.

...
```

stdlib/src/collections/list.mojo
```diff
@@ -493,16 +493,67 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
...
+        if len(self) >= self.capacity:
+            self._realloc(self.capacity * 2 + int(self.capacity == 0))
+        (self.data + len(self)).init_pointee_move(value^)
         self.size += 1
 
+    fn append[
+        D: DType, //
+    ](mut self: List[Scalar[D], *_, **_], value: SIMD[D, _]):
+        """Appends a vector to this list. If there is no capacity left, resizes
+        to `len(self) + value.size`.
```
ConnorGray, line 511: **Suggestion** Same suggestion here re. splitting this up so that the first line of the description stays short and simple.

stdlib/src/collections/list.mojo
```diff
@@ -493,16 +493,67 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
...
 
         Args:
             value: The value to append.
         """
-        if self.size >= self.capacity:
-            self._realloc(max(1, self.capacity * 2))
-        (self.data + self.size).init_pointee_move(value^)
+        if len(self) >= self.capacity:
+            self._realloc(self.capacity * 2 + int(self.capacity == 0))
+        (self.data + len(self)).init_pointee_move(value^)
```
ConnorGray, line 504: **Suggestion** I suggest we keep this as `self.size` (and eventually `self._len`) instead of changing to `len(self)`. It's a minor difference, but it keeps the IR as simple as possible (a field access instead of a function call that will need to be inlined) and makes it clear to readers that no logic is being run to get the length. (Which isn't always true, e.g. `String.len` has logic to adjust the actual length by 1 to account for a null terminator.)

stdlib/src/collections/list.mojo
martinvuyk: will do, but I'd rather we make it the norm to send it to it's own section:
```
Notes:
    ...
```

stdlib/src/utils/inline_string.mojo
```diff
@@ -147,28 +147,15 @@ struct InlineString(Sized, Stringable, CollectionElement, CollectionElementNew):
...
-
-            # Record that we've initialized `total_len` count of elements
-            # in `buffer`
-            buffer.size = total_len
-
-            # Add the NUL byte
-            buffer.append(0)
-
+            buffer.append(str_slice.as_bytes())
+            buffer.append(0)  # Add the NUL byte
```
ConnorGray, line 158: Nice! This is a great cleanup. In theory this might be slightly less efficient since the `append()` will need to perform redundant bounds-checking. But I think that's a small cost that we shouldn't worry too much about optimizing yet.

stdlib/src/collections/list.mojo
```diff
@@ -497,12 +497,72 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
...
+        Args:
+            value: The value to append.
+            count: The ammount of items to append.
+
+        Notes:
+            If there is no capacity left, resizes to `len(self) + count`.
+        """
+        self.reserve(self.size + count)
+        var v_ptr = UnsafePointer.address_of(value).bitcast[Scalar[D]]()
+        memcpy(self.data + self.size, v_ptr, count)
```
ConnorGray, line 545: **Suggestion** For this method to be safe, I think we need a check that `count` is not larger than the SIMD width of `value`. Otherwise, someone might be able to read more bytes from `value` than it actually contains, causing UB.

```suggestion
        memcpy(self.data + self.size, v_ptr, count)
        debug_assert(count <= value.size, "cannot append more elements than appear in SIMD vector")
```

stdlib/src/collections/list.mojo
```diff
@@ -497,12 +497,72 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
...
         Args:
             value: The value to append.
+
+        Notes:
+            If there is no capacity left, resizes to twice the current capacity.
+            Except for 0 capacity where it sets 1.
         """
         if self.size >= self.capacity:
-            self._realloc(max(1, self.capacity * 2))
+            self._realloc(self.capacity * 2 + int(self.capacity == 0))
```
ConnorGray, line 506: **Comment** Nice, this is a clever little optimization! 🧠

stdlib/src/utils/inline_string.mojo
martinvuyk: > In theory this might be slightly less efficient since the append() will need to perform redundant bounds-checking.

We could add a `unsafe_no_checks` parameter to all the append functions at some point. That way we'd stop having to resort to `UnsafePointer` so much for such scenarios.

stdlib/src/collections/list.mojo
martinvuyk: Great catch

ConnorGray:
!sync

ConnorGray:
Awesome, thanks for adding those Martin! 🙂  I'll sync this in.

stdlib/src/collections/list.mojo
martinvuyk: Hi @soraros. Looking at the assembly it might seem so.

```assembly
.Lf$local:
	leaq	(%rdi,%rdi), %rax ; %rax = %rdi * %rdi
	cmpq	$1, %rax ; compare and set the flags register
	adcq	$0, %rax ; %rax = %rax + $0 + carry_flag
	retq
```
```assembly
.Lg$local:
	xorl	%eax, %eax ; set %eax to zero
	testq	%rdi, %rdi ; set the flags to 1 if %rdi is zero, no-op
	sete	%al ; %al = int(flags == 1)
	leaq	(%rax,%rdi,2), %rax ; %rax = %rax + %rdi * 2
	retq
```
I'm not sure which one is faster just by looking at this. The lower one seems to involve more registers and ops, but I don't know how this would look pipelined at runtime. I think the `testq` and `sete` are practically no-ops in this case, but I'm not sure since I don't program in assembly much

It might be fixed by doing 
```mojo
var double = UInt(self.capacity) * 2
self._realloc(double + int(double == 0))
```

ConnorGray:
!sync

PR #3856, opened 2024-12-10
[stdlib] [NFC] Simplify `String` `List[Byte]` constructor

martinvuyk:
Simplify `String` `List[Byte]` constructor

### DIFF
--- a/mojo/stdlib/src/collections/list.mojo
+++ b/mojo/stdlib/src/collections/list.mojo
@@ -1078,19 +1078,6 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
 
         return self.data + self._len
 
-    fn _cast_hint_trivial_type[
-        hint_trivial_type: Bool
-    ](owned self) -> List[T, hint_trivial_type]:
-        var size = self._len
-        var capacity = self.capacity
-
-        # TODO: Why doesn't `__disable_del self` work here?
-        var data = self.steal_data()
-
-        return List[T, hint_trivial_type](
-            ptr=data, length=size, capacity=capacity
-        )
-
 
 fn _clip(value: Int, start: Int, end: Int) -> Int:
     return max(start, min(value, end))
--- a/mojo/stdlib/src/collections/string/string.mojo
+++ b/mojo/stdlib/src/collections/string/string.mojo
@@ -718,27 +718,26 @@ struct String(
 
     @always_inline
     fn __init__(out self, *, owned buffer: List[UInt8, *_]):
-        """Construct a string from a buffer of bytes without copying the
-        allocated data.
-
-        The buffer must be terminated with a null byte:
-
-        ```mojo
-        var buf = List[UInt8]()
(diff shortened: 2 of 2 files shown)


stdlib/src/collections/string.mojo
```diff
@@ -784,68 +784,25 @@ struct String(
...
-        The buffer must be terminated with a null byte:
-
-        ```mojo
-        var buf = List[UInt8]()
-        buf.append(ord('H'))
-        buf.append(ord('i'))
-        buf.append(0)
-        var hi = String(buf)
-        ```
+    fn __init__(out self, impl: List[Byte, *_]):
```
lsh, line 787: Why is it no longer owned? This would force making a new allocation.

martinvuyk: AFAIK not if you do `String(some_list^)`

martinvuyk: I can also just add it now :man_shrugging:

ConnorGray:
!sync

ConnorGray:
Hey Martin, I'm sorry to step on your toes here, but I just wanted to let you know I implemented a very similar refactoring that just merged internally.

I was in the area removing `@implicit` from these constructors, and noticed the same duplication you fixed here.

I'm sorry again about that, I'll keep trying to do better on noticing ahead of time when I'm overlapping a community PR😌

martinvuyk:
@ConnorGray don't worry it's no biggie. I don't know why I merged it now if you already imported it lol, but the CI is failing now. I think it's because of removing the `@implicit` from the `String` from `StringSlice` constructor. PR  #3922 will fix this by making `String.startswith()` use `StringSlice`.

ConnorGray:
Thanks for understanding Martin, and thank you for bringing that issue to my attention. I just merged your PR with the fix internally, so that should go out with the next nightly :)

ConnorGray:
Given the above, discussion I'm going to close this PR. But feel free to re-open if I missed something or if there were pieces here you still want to see merged :)

martinvuyk:
@ConnorGray  I can't reopen the PR and I would like to. Seeing commit 4c2c201e4eb6d2d5d35dad8171e8c06c75d39d21 I don't understand why you built a whole method when rebind does the trick. There is no need to move any data if parameters are just a type system "interpretation" (casting is free): `self._buffer = rebind[Self._buffer_type](buffer)`.

stdlib/src/collections/string/string.mojo
```diff
@@ -681,28 +681,25 @@ struct String(
...
+
+        Examples:
 
         ```mojo
-        var buf = List[UInt8]()
-        buf.append(ord('H'))
-        buf.append(ord('i'))
-        buf.append(0)
-        var hi = String(buffer=buf)
+        print(String(List[Byte](ord('h'), ord('i'), 0))) # hi
```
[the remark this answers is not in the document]
martinvuyk: Ok yeah you're right it's a bit dizzying

PR #3858, opened 2024-12-10
[stdlib] Add `_SplitlinesIter`

martinvuyk:
Add `_SplitlinesIter`

This allows lazy iteration over generic line boundaries.

### Benchmark results

CPU: Intel® Core™ i7-7700HQ

|  | old value (ms) | new value (ms) | percentage | magnitude|
|-- | -- | -- | -- | --|
|bench_string_splitlines[10] | 6.832 | 7.279 | -0.065 | 0.939|
|bench_string_splitlines[30] | 5.169 | 5.336 | -0.032 | 0.969|
|bench_string_splitlines[50] | 4.991 | 4.850 | 0.028 | 1.029|
|bench_string_splitlines[100] | 4.855 | 4.709 | 0.030 | 1.031|
|bench_string_splitlines[1000] | 4.353 | 4.708 | -0.081 | 0.925|
|bench_string_splitlines[10000] | 4.273 | 4.254 | 0.004 | 1.004|
|bench_string_splitlines[100000] | 4.289 | 4.327 | -0.009 | 0.991|
|bench_string_splitlines[1000000] | 4.261 | 4.242 | 0.004 | 1.004|
| average  | -  | -  | -0.015 | 0.986

Interpretation:

I had to undo some of the performance gains from  #4683 which stored the previous starting byte in a variable. It allowed the algorithm to just compare from two variables instead of dereferencing a pointer.

But since this is a left to right iterator, it can't store a variable and continue on to the next loop, it would have to call `__next__` recursively. This means the previous byte would have to be stored as a struct variable, and the `__has_next__` method would've had to check the edge-case where a sequence ends with `\r\n`; and that hits performance worse than dereferencing the next byte.

IMO this performance hit is worth the ability to give users the power of lazy iteration, like running the algorithm x times instead of for the whole string when not necessary.

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/string_slice.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string_slice.mojo
@@ -80,6 +80,80 @@ alias StaticString = StringSlice[StaticConstantOrigin]
 """An immutable static string slice."""
 
 
+@fieldwise_init
+struct _SplitlinesIter[
+    is_mutable: Bool, //,
+    origin: Origin[is_mutable],
+    *,
+    keep_ends: Bool = False,
+    forward: Bool = True,
+](Copyable, Movable):
+    """Iterator for `StringSlice` over unicode line breaks.
+
+    Parameters:
+        is_mutable: Whether the slice is mutable.
+        origin: The origin of the underlying string data.
+        keep_ends: Whether to keep the line breaks in the resulting slices.
+        forward: The iteration direction. `False` is backwards.
+    """
+
+    var _slice: StringSlice[origin]
+
+    fn __iter__(self) -> Self:
+        return self
+
+    fn __next__(mut self) -> StringSlice[origin]:
+        alias `\r` = UInt8(ord("\r"))
+        alias `\n` = UInt8(ord("\n"))
+
+        # highly performance sensitive code, benchmark before touching
+        @parameter
+        if forward:
+            var ptr = self._slice.unsafe_ptr()
+            var length = UInt(self._slice.byte_length())
+            var line_end = UInt(0)
+            var is_new_line = False
+            var b0 = Byte(0)
+            var char_len = UInt(0)
+
+            while not is_new_line and line_end < length:
+                b0 = ptr[line_end]
+                char_len = _utf8_first_byte_sequence_length(b0)
+                debug_assert(
+                    line_end + char_len <= length,
+                    "corrupted sequence causing unsafe memory access",
+                )
+                # percentage-wise a newline is uncommon compared to a normal byte
+                is_new_line = unlikely(
+                    _is_newline_char_utf8(ptr, line_end, b0, char_len)
+                )
+                line_end += char_len
+
+            var str_len = line_end
+            var is_r = unlikely(b0 == `\r`)
+            var may_be_r_n = is_r and likely(line_end < length)
+            var is_r_n = UInt(unlikely(may_be_r_n and ptr[line_end] == `\n`))
+            line_end += is_r_n
+
+            @parameter
+            if keep_ends:
+                str_len += is_r_n
+            else:
+                str_len -= UInt(splat(likely(is_new_line))) & char_len
+
+            self._slice = StringSlice[origin](
+                ptr=ptr + line_end, length=length - line_end
+            )
+            return StringSlice[origin](ptr=ptr, length=str_len)
+        else:
+            constrained[False, "reversed splitlines is not yet implemented"]()
+            return abort[StringSlice[origin]]()
+
+    @always_inline
+    fn __has_next__(self) -> Bool:
+        return self._slice.byte_length() > 0
+
+
 struct CodepointSliceIter[
     mut: Bool, //,
     origin: Origin[mut],
@@ -305,6 +379,25 @@ struct CodepointSliceIter[
             self._slice._slice._len -= slice_len
 
         return result
+
+    fn splitlines[
+        keep_ends: Bool = False
+    ](var self: CodepointSliceIter[forward=True]) -> _SplitlinesIter[
+        origin, keep_ends=keep_ends, forward=True
+    ]:
+        """Split the string at line boundaries. This corresponds to Python's
+        [universal newlines:](
+        https://docs.python.org/3/library/stdtypes.html#str.splitlines)
+        `"\\r\\n"` and `"\\t\\n\\v\\f\\r\\x1c\\x1d\\x1e\\x85\\u2028\\u2029"`.
+
+        Parameters:
+            keep_ends: Whether to keep the line breaks in the resulting slices.
+
+        Returns:
+            An iterator of StringSlices over the input split by line boundaries.
+        """
+        alias S = _SplitlinesIter[origin, keep_ends=keep_ends, forward=True]
+        return S(rebind[StringSlice[S.origin]](self._slice))
 
 
 struct CodepointsIter[mut: Bool, //, origin: Origin[mut]](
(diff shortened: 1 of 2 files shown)


stdlib/src/utils/string_slice.mojo
```diff
@@ -166,6 +166,79 @@ fn _memrmem[
     return UnsafePointer[Scalar[type]]()
 
 
+@value
+struct _SplitlinesIter[
```
gryznar, line 170: This would look much better as `_SplitLinesIter` ("l" in "lines" uppercased)

martinvuyk:
Hi @ConnorGray this kind of iterator is what I had in mind for `StringSliceIter` to be able to evolve into. Then also have a grapheme iterator, `UTF32Iter`/`UnicodeIter` (currently called `CharIter` which IMO is confusing), etc. Using UTF-8 with a view is simply much more efficient than building a `Char` which needs to copy and transform data into a UTF-32 format for every iteration.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2017,72 +2110,13 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
-                        line_start = line_end
-                        continue
-                var s = Self.Immutable(ptr=ptr + line_start, length=str_len)
-                output.append(s)
-                line_start = line_end
-
+        res = List[Self.Immutable](capacity=128)  # guessing
         if keepends:
-            _splitlines[keep=True]()
+            for s in self.codepoint_slices().splitlines[keep_ends=True]():
```
JoeLoser, line 2179: Since both of these codepaths are the same, should we just do

```
for s in self.codepoint_slices().splitlines[keep_ends=keepend]():
    res.append(s)
```

WDYT?

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,80 @@ alias StaticString = StringSlice[StaticConstantOrigin]
 """An immutable static string slice."""
 
 
+@fieldwise_init
+struct _SplitlinesIter[
+    is_mutable: Bool, //,
+    origin: Origin[is_mutable],
```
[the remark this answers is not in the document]
JoeLoser: No, this is fine as-is.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,80 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+    *,
+    keep_ends: Bool = False,
+    forward: Bool = True,
+](Copyable, Movable):
```
JoeLoser, line 90: **Suggestion** In the future, it would be good to have this guy conform to the recent `Iterator` traits and implements `bounds` function (that @NathanSWard added recently).

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2068,7 +2161,7 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
                 offset += b_len
             return length != 0
 
-    fn splitlines(self, keepends: Bool = False) -> List[Self.Immutable]:
+    fn splitlines(self, keepends: Bool = False, out res: List[Self.Immutable]):
```
JoeLoser, line 2164: **Question** Any reason to use an `out` parameter here instead of just keeping the return type as-is?  We're just using it as a function local and assigning into it.  We slightly prefer the "Create a function local variable for the output, assign into it, and return it" style for ease of readability.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,80 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+        keep_ends: Whether to keep the line breaks in the resulting slices.
+        forward: The iteration direction. `False` is backwards.
+    """
+
+    var _slice: StringSlice[origin]
+
+    fn __iter__(self) -> Self:
+        return self
+
+    fn __next__(mut self) -> StringSlice[origin]:
```
barcharcraz, line 105: Have you benchmarked this against the (much simpler) approach of just doing a byte-by-byte search for the first newline and checking if there's a carriage return beforehand?

Particularly since that's basically "free" to vectorize

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I want to, I did so in the followup PR  #5398, though I can do it here as well

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I did so in the followup PR  #5398. This one is much older and I forgot to add them here

mojo/stdlib/stdlib/collections/string/string_slice.mojo
JoeLoser: Oh OK, doing that in #5398 is fine - we were just reviewing this one as a team first so we can then proceed with #5398. Thanks!

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I like how it helps avoid a return line and it doesn't require a move (though the compiler most probably optimizes it away)

JoeLoser:
@martinvuyk note that you'll need to fix the CI failures due to some implicit/explicit copying things that have changed in the past few months. PTAL!

mojo/stdlib/stdlib/collections/string/string_slice.mojo
JoeLoser: The compiler should optimize the move in this case ("NRVO"), so no need to use an `out` parameter.  The team would prefer creating a local and returning it here to improve readability.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: There are more kinds of newlines than just LF. You can also see lower down in the diff that my previous approach did something similar to what you're referring to, and I explicitly mentioned the sacrifice in the PR description. Sadly there is no way to vectorize this due to unicode line breaks

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -306,6 +380,25 @@ struct CodepointSliceIter[
 
         return result
 
+    fn splitlines[
```
barcharcraz, line 383: why is this on CodePointSliceIter? It should be on StringSlice just like the list version. one of them can be renamed to avoid conflict (or we could just have only the lazy version, collecting the iterator into a list is easy enough).

martinvuyk:
> @martinvuyk note that you'll need to fix the CI failures due to some implicit/explicit copying things that have changed in the past few months. PTAL!

@JoeLoser though PR  #5398 is much bigger it also solves most issues you pointed out, and is just a direct replacement of returning a list to returning the iterator. Do you want me to bring the implementation here? I would rather just directly change the splitlines function to return an iterator than do this intermediary step. I originally did this because I wasn't sure you guys would want me to change the splitlines API that much (remember that this PR was started on december 10th 2024)

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: This PR was started in december of last year, a lot has changed with regards to python compatibility goals. Basically it was to not change the splitlines API from returning a list to an iterator

JoeLoser:
> > @martinvuyk note that you'll need to fix the CI failures due to some implicit/explicit copying things that have changed in the past few months. PTAL!
> 
> @JoeLoser though PR #5398 is much bigger it also solves most issues you pointed out, and is just a direct replacement of returning a list to returning the iterator. Do you want me to bring the implementation here? I would rather just directly change the splitlines function to return an iterator than do this intermediary step. I originally did this because I wasn't sure you guys would want me to change the splitlines API that much (remember that this PR was started on december 10th 2024)

@martinvuyk thanks for the context. I chatted with @barcharcraz and @NathanSWard .  Let's just review #5398 as a whole and close this PR.  Feel free to address any relevant outstanding comments in terms of that PR then. Thanks!

PR #3859, opened 2024-12-11
[stdlib] Vectorize ASCII helper functions

martinvuyk:
Vectorize ASCII helper functions.

This is a split off from #3686. The vectorized helper implementations are in preparation for it and future ASCII String optimizations.

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/_ascii.mojo
+++ b/mojo/stdlib/stdlib/collections/string/_ascii.mojo
@@ -1,0 +1,225 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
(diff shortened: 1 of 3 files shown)


stdlib/src/collections/string/_ascii.mojo
```diff
@@ -0,0 +1,211 @@
...
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+# ===----------------------------------------------------------------------=== #
+# isdigit
+# ===----------------------------------------------------------------------=== #
+
+
+@always_inline
+fn _isdigit_vec[w: Int](v: SIMD[DType.uint8, w]) -> SIMD[DType.bool, w]:
```
gryznar, line 20: This partially flatcase private function looks strange, while others are fully snake_case

martinvuyk: Because the function name comes from C. And the python function also has that name. And this is a private function anyway...

PR #3860, opened 2024-12-11
[stdlib] Refactor and optimize `StringSlice.replace()`

martinvuyk:
Refactor and optimize `StringSlice.replace()`

CC: @JoeLoser another place where we need #3548

## Benchmark Results

CPU: Intel® Core™ i7-7700HQ

|| old value (ms) | new value (ms) | percentage | magnitude|
|-- | -- | -- | -- | -- |
|bench_string_replace[10] | 7.76 | 5.82 | 0.25 | 1.33|
|bench_string_replace[30] | 16.61 | 6.36 | 0.62 | 2.61|
|bench_string_replace[50] | 14.57 | 4.02 | 0.72 | 3.62|
|bench_string_replace[100] | 10.23 | 2.48 | 0.76 | 4.13|
|bench_string_replace[1000] | 9.98 | 2.22 | 0.78 | 4.49|
|Average improvement | - |- | 0.63 | 3.24|

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_string.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_string.mojo
@@ -210,8 +210,9 @@ def bench_string_replace[
 
     @always_inline
     def call_fn() unified {read}:
-        var res = black_box(items).replace(black_box(old), black_box(new))
-        keep(res)
+        for _ in range(10**6 // length):  # for stability with small lengths
+            var res = black_box(items).replace(black_box(old), black_box(new))
+            keep(res)
 
     b.iter(call_fn)
 
--- a/mojo/stdlib/benchmarks/collections/bench_string.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_string.mojo
@@ -210,8 +210,9 @@ def bench_string_replace[
 
     @always_inline
     def call_fn() {read}:
-        var res = black_box(items).replace(black_box(old), black_box(new))
-        keep(res)
+        for _ in range(10**6 // length):  # for stability with small lengths
+            var res = black_box(items).replace(black_box(old), black_box(new))
+            keep(res)
 
     b.iter(call_fn)
 
--- a/mojo/stdlib/std/collections/string/string.mojo
+++ b/mojo/stdlib/std/collections/string/string.mojo
@@ -1335,14 +1335,15 @@ struct String(
         """Return the number of non-overlapping occurrences of substring
         `substr` in the string.
 
-        If sub is empty, returns the number of empty strings between characters
-        which is the length of the string plus one.
-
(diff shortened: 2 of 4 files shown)


martinvuyk:
Hi @ConnorGray this is why I wanted to have better iterator ergonomics in #3700:
```mojo
s = String("123")
for c in s:
    ...
```
now gives the warning: `warning: Use 'str.chars()' or 'str.char_slices()' instead.`
I strongly dislike this. It should be an option to transform the `_StringSliceIter.as_chars()` or something along those lines (e.g. having `CharIter(iter("123"))`). Because this absolutely kills the pythonic default which is simple and makes it feel like writing Rust.

stdlib/src/collections/string/string_slice.mojo
```diff
@@ -863,7 +863,6 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
             self.unsafe_ptr(), rhs.unsafe_ptr(), min(len1, len2)
         )
 
-    @deprecated("Use `str.codepoints()` or `str.codepoint_slices()` instead.")
```
gryznar, line 866: Same as above. Pr does not need this change

stdlib/src/collections/string/string.mojo
```diff
@@ -1057,7 +1057,6 @@ struct String(
         """
         self._iadd[False](other.as_bytes())
 
-    @deprecated("Use `str.codepoints()` or `str.codepoint_slices()` instead.")
```
gryznar, line 1060: It serms, that this change is done  "by the way", and is out of scope of this pr

martinvuyk:
I did use `__iter__()` in the code for this PR. I will remove the deprecation in favor of using `codepoint_slices()`

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -1120,66 +1120,75 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
         """
         return rebind[Self.Immutable](self)
 
-    fn replace(self, old: StringSlice, new: StringSlice) -> String:
+    fn replace[
+        *, buffer_size: Int = STACK_BUFFER_BYTES
```
laszlokindrat: Do we really need to expose this in the API? I think it's fine if we just use `STACK_BUFFER_BYTES`.

martinvuyk: ok

martinvuyk:
@NathanSWard this fixed the bug you found in https://github.com/modular/modular/commit/cd8d28b5e813fbd0408abf6fe7514457e104d7b3 (and other bugs)

Issue #3756, opened 2024-11-07
[Feature Request] [mojo-lang] More meta-programming tooling


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Several requests to make this possible:
- optional: `@static_arguments` syntax sugar for functions which take in compile time parameters as arguments.
- `__type__() -> AnyType`: dunder that gets added by codegen when on argument type signatures
- `__trait__() -> AnyTrait`: dunder that gets added by codegen when on parameter type signatures
- `_implements[type, trait]()` and `rebind_trait[type, trait]()` as asked for in #3630
- optional: `__trait_of()` that returns the whole type's method signatures as a trait as asked for in #3047.

```mojo
alias AnyTrait = __mlir_type[`!lit.anytrait<`, AnyType, `>`]


@value
struct Type[type_def: AnyType, traits: AnyTrait = __trait_of(type_def)]:
    """This represents a Type which implements a composition of traits.
    
    Parameters:
        type_def: The struct type definition.
        traits: The composition of traits the type conforms to.
    """
    var value: type_def

    @staticmethod
    fn __type__() -> AnyType:
        return Self.type_def

    @staticmethod
    fn __trait__() -> AnyTrait:
        return Self.traits

    @static_arguments
    @staticmethod
    fn __is__[T2: AnyType](other: Type[T2]) -> Bool:
        return _type_is_eq[T, T2]()

    @static_arguments
    @staticmethod
    fn rebind(new_type: AnyType) -> __type_of(Type[new_type, Self.traits]):
        return Type[rebind[new_type](Self.type_def), Self.traits]

    @static_arguments
    @staticmethod
    fn rebind(
        new_trait: AnyTrait
    ) -> __type_of(Type[rebind_trait[new_trait](Self.type_def), new_trait]):
        return Type[rebind_trait[new_trait](Self.type_def), new_trait]

    @static_arguments
    @staticmethod
    fn implements(some_trait: AnyTrait) -> Bool:
        return _implements[Self.type_def, some_trait]()


@static_arguments
fn type[T: AnyType, //](v: T) -> __type_of(Type[__type_of(v)]):
    return Type[__type_of(v)]


@static_arguments
fn isinstance[
    T0: AnyType, T1: AnyType, //
](v: T0, some_type: __type_of(T1)) -> Bool:
[...]

PR #3861, opened 2024-12-11
[stdlib] Reintroduce `Stringlike` trait and use it for `Stringlike.find()`

martinvuyk:
Reintroduce `Stringlike` trait and use it for `Stringlike.find()`

This allows some ergonomic code and not needing to allocate for many cases like `String("123").find("1")` which would previously cast `"1"` to a `String`.


martinvuyk:
Hi @JoeLoser and @ConnorGray sorry if I'm drowning you in PRs I took your advice quite literally, they are mostly small and independent.

In this one I'm coming across the same problem as in #3677 where I tried to build a generic function that returns a `Span` for non-owning types as well as owning ones. Now that we have `UnsafePointer` carrying the origin of `String`, `StringLiteral` and `StringSlice` the same problem arises, which is we need an API that works for all types not just those that can return `__origin_of(self)`.

One of the ideas I had was adding a method `__origin__(self) -> Origin`, but the compiler doesn't seem to like using dynamic values (this was one of the motivations for #3756)

We could make `String`, `StringLiteral` and `StringSlice` return a `MutableAnyOrigin` pointer for now, or try to use this PR as an introduction for whatever fix we can find. WDYT?

ConnorGray:
Hey @martinvuyk, thanks for breaking down the PRs, they're looking great, no need to apologize at all😄

With regards to this PR specifically I think I see what you're aiming for in trying to build an abstraction over all "string like" collections (`StringLiteral`, `String`, `StringSlice`). Modeling that abstraction with a trait is one possible approach. But I wonder if using `StringSlice` as the "bottom" or "common denominator" type might work as well.

To illustrate what I mean, let's have a look at the contract of the `Stringlike` trait you're proposing: 

```mojo
trait Stringlike(CollectionElement, CollectionElementNew):
    alias is_mutable: Bool
    alias origin: Origin[is_mutable]

    fn byte_length(self) -> Int:
        ...

    fn unsafe_ptr(self) -> UnsafePointer[Byte, is_mutable=is_mutable, origin=origin]:
        ...
```

If we compare that to `StringSlice`:

```mojo
struct StringSlice[
    mut: Bool,
    origin: Origin[mut]
]:
    fn byte_length(self) -> Int:
        ...

    fn unsafe_ptr(self) -> UnsafePointer[Byte, mut=mut, origin=origin]:
        ...
```

we see that they're essentially identical. I think that this means that a function accepting an `arg: StringSlice` is roughly an equivalently powerful abstraction as accepting an `[T: Stringlike] (arg: T)`. Additionally, it means that any type that could implement `Stringlike` could just as easily convert itself into a concrete instance of `StringSlice`.

All else being equal then, I think the simpler design choice to use a concrete type (`StringSlice`) is likely easier to learn, easier to use, and easier to maintain. Additionally, it has the benefit of avoiding "bloat" in the compiled binary—instead of compiling separate versions of `find()` (one for `StringSlice`, one for `StringLiteral`, one for `String`, etc.) for each string-like type, we'll have just one (for `StringSlice`).

For an example, consider this these two different signatures: 

```mojo
    fn find[T: Stringlike, //](self, substr: T, start: Int = 0) -> Int:
    fn find(self, substr: StringSlice, start: Int = 0) -> Int:
```

For each signature, the function body can do essentially the same things: get the byte length, and work with a pointer to the string data. But the `StringSlice` version I would argue has all the advantages of the `Stringlike` abstraction, while also being a bit simpler since it can avoid depending on generics.

I'm curious to hear your thoughts on this as well 🙂  — this is just my first impression after seeing how this abstraction is planned to be used initially, but perhaps there's advantages I've missed, or a future state you're building towards that I might not see yet 😌

ConnorGray:
I just wanted to add a quick note for context: Us adding an implicit conversion from `String` / `StringLiteral` / etc. to `StringSlice` is similar in purpose to what Rust calls "deref coercion" (though not the same semantics), whereby with minimal syntax you can "coerce" a Rust `String` type down to a `&str` just by referencing:

```rust
fn main() {
    let string = String::from("foo");
    
    let s: &str = &string;
}
```

martinvuyk:
> trait Stringlike: # TODO: Not sure how to write the `origin` correctly today off the top of my head fn as_string_slice(self) -> StringSlice: ... struct StringSlice: fn __init__[T: Stringlike](out self, stringlike: T): self = stringlike.as_string_slice()

This is a good approach IMO. I actually already tried to implement something similar. 

I implemented the current `fn __init__[O: ImmutableOrigin, //](mut self: StringSlice[O], ref [O]value: String):`, then I tried to make the origin inferred `fn __init__(mut self: StringSlice[__origin_of(value)], ref value: String):` but we still need the arguments to be ordered.

Problem is that `StringLiteral` is very funky at compile time. And every time I tried to use the generic `Stringlike` type it beautifully exploded unless I used `_type_is_eq` and sent it elsewhere.

So I guess for now we could just use `.as_string_slice()` as it currently is which is only for owning types, + a specialized constructor only for `StringLiteral` which we already have.

This is OK for now, but it still constraints the types to be owning (to be able to implement the trait and carry the correct origin), in the case of strings I don't think it's really a problem. But for `.as_bytes()` I think it will be in the future.

martinvuyk:
```mojo
trait StringLike(CollectionElement, CollectionElementNew):
    """Trait intended to be used as a generic entrypoint for all String-like
    types."""
    ...

trait StringOwnerLike(StringLike):
    fn as_bytes(ref self) -> Span[Byte, __origin_of(self)]:
        """Returns a contiguous slice of the bytes owned by this string.

        Returns:
            A contiguous slice pointing to the bytes owned by this string.

        Notes:
            This does not include the trailing null terminator.
        """
        ...

    fn as_string_slice(ref self) -> StringSlice[__origin_of(self)]:
        """Returns a string slice of the data owned by this string.

        Returns:
            A string slice pointing to the data owned by this string.
        """
        ...

trait StringSliceLike(StringLike):
    alias mut: Bool
    """The mutability of the origin."""
    alias origin: Origin[mut]
    """The origin of the data."""

    fn as_bytes(self) -> Span[Byte, origin]:
        """Returns a contiguous slice of the bytes.

        Returns:
            A contiguous slice pointing to the bytes.

        Notes:
            This does not include the trailing null terminator.
        """
        ...

    fn as_string_slice(self) -> StringSlice[origin]:
        """Returns a string slice of the data.

        Returns:
            A string slice pointing to the data.
        """
        ...

trait StringLiteralLike(StringLike):
    fn as_bytes(self) -> Span[Byte, StaticConstantOrigin]:
        """Returns a contiguous slice of the bytes.

        Returns:
            A contiguous slice pointing to the bytes.

        Notes:
            This does not include the trailing null terminator.
        """
        ...

    fn as_string_slice(self) -> StringSlice[StaticConstantOrigin]:
        """Returns a string slice of the data.

        Returns:
            A string slice pointing to the data.
        """
        ...
```

@ConnorGray this isn't necessarily the end version, but this is what I mean by saying that we can't yet generalize the origin that the non-owning types return when calling methods like `as_bytes()` and `as_string_slice()`. We could also just have 3 traits and always 3 functions for cases where we need to generalize a lot until we have tools to check which subset of traits a type implements.

All 3 of these are types that someone else might implement.
- `StringOwnerLike`/`AsBytesOwnerLike`: if somebody implements another string type specialized for their use case. 
- `StringSliceLike`/`AsBytesSliceLike`: I'm currently playing around with some json parsing and validation ideas, this applies to many other cases.
- `StringLiteralLike`/`AsBytesStaticLike`: If we eventually allow embedding custom binary data into the Mojo binary itself, someone might have their own binary data which they wish to traverse in a certain way. It would have a `StaticConstantOrigin`

PR #3863, opened 2024-12-11
[stdlib] Refactor `String` from `StringSlice` constructor

martinvuyk:
Refactor `String` `StringSlice` constructor


stdlib/src/collections/string.mojo
```diff
@@ -887,24 +887,19 @@ struct String(
...
-        # +1 for null terminator, initialized to 0
-        buffer.resize(length + 1, 0)
-        memcpy(
-            dest=buffer.data,
-            src=str_slice.as_bytes().unsafe_ptr(),
-            count=length,
-        )
-        self = Self(buffer^)
+        var length = str_slice.byte_length()
+        var ptr = UnsafePointer[Byte].alloc(length + 1)
```
lsh, line 899: This mostly looks good! :) I think we should add back the commend about `+ 1 for the null terminator` since it's good to have magic numbers addressed by comments.

martinvuyk: IMO it's redundant given it's stated in `.byte_length()`'s docstrings but ok

martinvuyk:
@lsh FYI something happened with modularbotn in commit 181f899f2f39b5b358808c115c20df876aeff3d1 and this didn't get closed after merging, the PR wasn't linked and it has no authorship info

PR #3870, opened 2024-12-11
[stdlib] Remove `String` `StringRef` dependencies

martinvuyk:
Remove `String` `StringRef` dependencies


JoeLoser (approved):
Nice, thank you!

JoeLoser:
!sync

stdlib/src/collections/string.mojo
```diff
@@ -1038,37 +1039,6 @@ struct String(
...
-        """Construct a string from a sequence of bytes.
-
-        This does no validation that the given bytes are valid in any specific
-        String encoding.
-
-        Args:
-            buff: The buffer. This should have an existing terminator.
-        """
-
-        return String(ptr=buff, length=len(StringRef(ptr=buff)) + 1)
```
JoeLoser, line 1053: **Suggestion** Can you just use `StringSlice` here to replace the `StringRef`?  We don't want to delete these `_from_bytes` methods quite yet as we use them internally in some MAX code.

martinvuyk: I just added them again, since we'd need to reimplement `strlen` if we wanted to avoid using `StringRef` here.

JoeLoser:
!sync

stdlib/src/collections/string.mojo
JoeLoser: 👍 thanks!

martinvuyk:
@soraros didn't see the edited lines after merging, seems like @lsh already merged the changes I was doing here

PR #3873, opened 2024-12-12
[stdlib] Add `_write_hex` util

martinvuyk:
Add `_write_hex` util.

This is a split-off from #3686. It is needed for that and any implementation that needs to write a hexadecimal value to a pointer as fast as possible.

Co-author: @soraros with his implementation from #3694

### DIFF
--- a/mojo/stdlib/stdlib/utils/write.mojo
+++ b/mojo/stdlib/stdlib/utils/write.mojo
@@ -12,10 +12,11 @@
 # ===----------------------------------------------------------------------=== #
 """Establishes the contract between `Writer` and `Writable` types."""
 
+from bit import byte_swap
 from collections import InlineArray
 from sys.info import is_amd_gpu, is_gpu, is_nvidia_gpu
 
-from memory import Span, UnsafePointer, memcpy
+from memory import Span, UnsafePointer, memcpy, bitcast
 
 # ===-----------------------------------------------------------------------===#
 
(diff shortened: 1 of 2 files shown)


stdlib/src/utils/write.mojo
```diff
@@ -382,3 +382,99 @@ fn write_buffered[
...
+    alias U8 = DType.uint8
+    return (
+        `0`
+        + b
+        + (((b <= 9).cast[I8]() - 1) & (`a` - `9` - 1).cast[I8]()).cast[U8]()
+    )
+
+
+@always_inline
+fn _write_hex[amnt_hex_bytes: Int](p: UnsafePointer[Byte], decimal: Int):
```
JoeLoser:
```suggestion
fn _write_hex[number_hex_bytes: Int](p: UnsafePointer[Byte], decimal: Int):
```

Let's not abbreviate `amnt` in favor of spelling it out fully (I also suggest `number` to replace `amnt`, but don't feel strongly if you prefer `amount`, it just doesn't read as well).

soraros:
I wonder whether we could have this function take a `Writable` argument.

martinvuyk:
I don't think so, this is supposed to be the low-level util for writing hex. A method that used this could look like:
```mojo
result = String(capacity=some_known_amount_of_hex_bytes)
for value in ...:
    _write_hex[...](res.unsafe_ptr, value)
```
or for an optimized version of `repr` that I'm working on:
```mojo
    var nonprintable_python = span.count[func=_nonprintable_python]()
    var hex_prefix = 3 * nonprintable_python  # \xHH
    var length = span_len + hex_prefix + 2  # for the quotes
```
and an example use is in the body of the `ascii` implementation that I'm also working on:
```mojo
    while v_idx < span_len:
        var b0 = v_ptr[v_idx]
        use_dquote = use_dquote or (b0 == `'`)
        var seq_len = _utf8_first_byte_sequence_length(b0)
        var b1 = v_ptr[v_idx + int(seq_len > 1)]
        var is_2byte_short = seq_len == 2 and b0 <= 0xC3
        if isprintable(b0):
            b_ptr[b_idx] = b0
            b_idx += 1
        elif seq_len == 1 or is_2byte_short:
            codepoint = int(b0)
            if is_2byte_short:
                codepoint = _ord[2](b0, b1)
            _write_hex[2](b_ptr + b_idx, codepoint)
            b_idx += 4
        elif seq_len < 4:
            var codepoint: Int
            if seq_len == 2:
                codepoint = _ord[2](b0, b1)
            else:
                codepoint = _ord[3](b0, b1, v_ptr[v_idx + 2])
            _write_hex[4](b_ptr + b_idx, codepoint)
            b_idx += 6
        else:
            codepoint = _ord[4](b0, b1, v_ptr[v_idx + 2], v_ptr[v_idx + 3])
            _write_hex[8](b_ptr + b_idx, codepoint)
            b_idx += 10
        v_idx += seq_len
```

or for a write_to method:
```mojo
struct Codepoint:
    fn write_to[W: Writer](self, mut writer: W):
        amnt_hex_bytes = _amnt_hex_bytes(Int(self))
        data = InlineArray[Byte, 10](uninitialized=True)
        ptr = data.unsafe_ptr()
        if amnt_hex_bytes == 2:
            _write_hex[2](ptr, Int(self))
        ...
        writer.write_bytes(Span(ptr=ptr, length=amnt_hex_bytes))
```

mojo/stdlib/stdlib/utils/write.mojo
```diff
@@ -533,3 +534,104 @@ struct WritableVariadicPack[
...
+    _write_hex[8](ptr, ord("🔥"))
+    assert_equal(r"\\U0001f525", S(ptr=ptr, length=10))
+    memset_zero(ptr, len(items))
+    _write_hex[4](ptr, ord("你"))
+    assert_equal(r"\\u4f60", S(ptr=ptr, length=6))
+    memset_zero(ptr, len(items))
+    _write_hex[2](ptr, ord("Ö"))
+    assert_equal(r"\\xd6", S(ptr=ptr, length=4))
+    ```
+    .
```
JoeLoser:
```suggestion
```
The Mojo tooling bug is fixed, we can drop this.

JoeLoser (approved):

JoeLoser:
!sync

mojo/stdlib/stdlib/utils/write.mojo
```diff
@@ -533,3 +534,103 @@ struct WritableVariadicPack[
...
+    assert_equal("4f60", S(ptr=ptr, length=4))
+    memset_zero(ptr, len(items))
+    _hex_digits_to_hex_chars(ptr, UInt8(ord("Ö")))
+    assert_equal("d6", S(ptr=ptr, length=2))
+    ```
+    .
+    """
+
+    alias size = decimal.dtype.sizeof()
+    var data: SIMD[DType.uint8, size]
```
soraros, line 581: It's not super important, but could we still call this `bytes`? In this context, we typically use the terms "byte" and "nibble" as they are parallel.

martinvuyk: :man_shrugging:

PR #3877, opened 2024-12-13
[stdlib] Add vectorized `Span.apply()`

martinvuyk:
Add vectorized `Span.apply()`

### DIFF
--- a/mojo/stdlib/stdlib/memory/span.mojo
+++ b/mojo/stdlib/stdlib/memory/span.mojo
@@ -645,3 +645,74 @@ struct Span[
             var value = ptr[middle + 1]
             (ptr + middle - 1).move_pointee_into(ptr + middle + 1)
             (ptr + middle - 1).init_pointee_move(value)
+
+    fn apply[
+        dtype: DType,
+        O: MutableOrigin, //,
+        func: fn[w: Int] (SIMD[dtype, w]) capturing -> SIMD[dtype, w],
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/memory/span.mojo
```diff
@@ -473,3 +473,74 @@ struct Span[
             address_space=address_space,
             alignment=alignment,
         ](ptr=self._data, length=self._len)
+
+    fn apply[
+        D: DType,
+        O: MutableOrigin, //,
+        func: fn[w: Int] (SIMD[D, w]) -> SIMD[D, w],
+    ](mut self: Span[Scalar[D], O]):
```
lsh: Question: why is this `mut self`? With respect to a Span, `mut self` implies a need to adjust the container length rather than mutate the underlying data (which is reflected in the origin)

martinvuyk: true true true, that was a mistake

lsh: Do you want me to fix this before merging, or will you?

martinvuyk: I forgot about it lol, hold on

martinvuyk: fixed

lsh:
!sync

lsh:
Can you rebase this? :)

lsh:
!sync

PR #3885, opened 2024-12-14
[docs] update broken link to nightly branch

miguelcsx:
- The link to the example of implementing operators for a custom type was broken, as it pointed to the `main` branch.
- Updated the URL to point to the correct `nightly` branch, ensuring the link now leads to the correct example: https://github.com/modularml/mojo/tree/nightly/examples/operators
- The change was made on the page: https://docs.modular-dev.com/nightly/mojo/manual/operators/#an-example-of-implementing-operators-for-a-custom-type


arthurevans:
Sorry for the delay on this. This link was designed to work correctly when 24.6 was released (and it does work now).

But now that we have nightly docs on the public site, we need to improve our tooling so we can have the nightly docs link to the nightly branch of the repo, and automatically switch that link when we build the release docs. So I think we do want to link to the nightly branch here.

!sync

PR #3886, opened 2024-12-16
[stdlib] Add bit-mask utils

martinvuyk:
Add bit-mask utils. Making them private for now while we figure out whether centralizing bit mask tricks is best or not
 and finding a good API for it

This is a split off from PR  https://github.com/modular/modular/pull/3528. It enables many branchless algorithms which work with bitwise operations instead of multiplication.

### DIFF
--- a/mojo/stdlib/stdlib/bit/_mask.mojo
+++ b/mojo/stdlib/stdlib/bit/_mask.mojo
@@ -1,0 +1,97 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2024, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Provides functions for bit masks.
+
+You can import these APIs from the `bit` package. For example:
+
+```mojo
+from bit.mask import is_negative
+```
+"""
+
+from sys.info import bitwidthof
+
+
+@always_inline
+fn is_negative(value: Int) -> Int:
+    """Get a bitmask of whether the value is negative.
+
+    Args:
+        value: The value to check.
+
+    Returns:
+        A bitmask filled with `1` if the value is negative, filled with `0`
+        otherwise.
+    """
+    return Int(is_negative(Scalar[DType.index](value)))
+
+
+@always_inline
+fn is_negative[dtype: DType, //](value: SIMD[dtype, _]) -> __type_of(value):
+    """Get a bitmask of whether the value is negative.
+
+    Parameters:
+        dtype: The DType.
+
+    Args:
+        value: The value to check.
+
+    Returns:
+        A bitmask filled with `1` if the value is negative, filled with `0`
+        otherwise.
+    """
+    constrained[
+        dtype.is_integral() and dtype.is_signed(),
+        "This function is for signed integral types.",
+    ]()
+    return value >> (bitwidthof[dtype]() - 1)
+
+
+@always_inline
+fn is_true[
+    dtype: DType, size: Int = 1
+](value: SIMD[DType.bool, size]) -> SIMD[dtype, size]:
+    """Get a bitmask of whether the value is `True`.
+
+    Parameters:
+        dtype: The DType.
+        size: The size of the SIMD vector.
+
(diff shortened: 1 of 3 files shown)


JoeLoser (commented):
Do you still want to pursue this now that we have a Bitset?

martinvuyk:
@JoeLoser I do since this is a bit different, it is mostly to replace magic looking code like the one in PR #3528
```mojo
rhs = src_str.find(sep, lhs)
# if not found go to the end
rhs += -Int(rhs == -1) & (str_byte_len + 1)
```
becomes
```mojo
rhs += BitMask.is_true(rhs == -1) & (str_byte_len + 1)
```
and it should return the type that was inputted not create a `BitSet` with `UInt64` in cases where masks of smaller types are also desired

martinvuyk:
@JoeLoser I don't know why the `magic.lock` is an addition in this PR now

soraros:
I think you could achieve these with `memory.pack_bits`, `SIMD.to_bits/cast`.

martinvuyk:
> I think you could achieve these with `memory.pack_bits`, `SIMD.to_bits/cast`.

@soraros It is my intention to add that kind of functionality later on:
```mojo
struct BitMask:
    fn truth_mask[dtype: DType](values: SIMD[Bool]) -> Scalar[new_type]:
        return pack_bits(values)   
```
I think it will be helpful to have it centralized on a type like this

And I am using `SIMD.cast` here, `SIMD.to_bits` I'm not so sure of but we could do it later on to add support for floats.

And FYI `BitMask.is_true[DType.int8](SIMD[DType.bool, 2](False, True)) == SIMD[DType.int8](0, -1)` so there is no intention of bit packing

soraros:
I agree that we want a better API in the future (that's why I introduced `to_bits` in the first place, #3680), but in the short term, I think we need/want to improve the foundation a bit (e.g., #4526).

martinvuyk:
I think we can do both at the same time. My goal with this type is to centralize all those tricks and functions we have all over the stdlib and make them more readable at a glance.

To give a bit convoluted example:
```mojo
var values = SIMD[DType.int8, 4](0b1, 0b10, 0b11, 0b100)
var condition = (values > 0).reduce_and()
var bits_to_extract = SIMD[DType.int8, 4](0, 0, 0, 1)
var result_1 = (-(condition.cast[DType.int8]()) & (pack_bits[DType.int8](bits_to_extract) & values))
print(result_1) # SIMD[DType.int8, 4](1, 0, 1, 0)

var result_2 = (BitMask.is_true[DType.int8](condition) & (BitMask.truth_mask[DType.int8](bits_to_extract) & values))
print(result_2) # SIMD[DType.int8, 4](1, 0, 1, 0)
```
I find `result_2` much more self-explanatory and readable

soraros:
I understand your concerns, and I’m not satisfied with the ergonomics of bit-trick usage in the standard library either. API design is challenging, especially when we expose things publicly from day one, as in this PR. Here’s my proposal:

- Introduce these as private helper functions (or encapsulate them in a struct) in #3528, so we can evaluate their benefits.
- Refactor the surrounding code to use those helpers, ensuring they’re the right building blocks for a SWAR API.
- Once they’ve been validated, move them back into `bit`, `SIMD`, or wherever they belong, crystallised into a clean, public API.

martinvuyk:
@soraros good points, would you agree if I just make the type `_BitMask`? or would you rather I refactor the methods into standalone functions?

It is actually a very good idea to keep them private first, I was thinking to also experiment and seeing whether it is worth it adding other helpers like
```mojo
fn max_zero_or(value: Int) -> Int:
    """Equivalent to doing `max(0, value)`"""
    return is_negative(-value) & value
```

soraros:
> @soraros good points, would you agree if I just make the type `_BitMask`? or would you rather I refactor the methods into standalone functions?

I don’t have a preference here. Let’s summon @JoeLoser to have the final say.

martinvuyk:
I just refactored them to be standalone functions, it's simpler

JoeLoser:
> I just refactored them to be standalone functions, it's simpler

That is fine.  I'll sync this in to see what others think.

JoeLoser:
!sync

mojo/stdlib/stdlib/bit/_mask.mojo
```diff
@@ -0,0 +1,97 @@
...
+        size: The size of the SIMD vector.
+
+    Args:
+        value: The value to check.
+
+    Returns:
+        A bitmask filled with `1` if the value is `False`, filled with `0`
+        otherwise.
+    """
+    return is_true[dtype](~value)
```
JoeLoser, line 97:
```suggestion
    return is_true(~value)
```

martinvuyk: the dtype of the return type can't be inferred, it has to be defined

JoeLoser:
!sync

PR #3898, opened 2024-12-19
[stdlib] Add `assert_equal(List[StringSlice], List[StringSlice])`

martinvuyk:
Add `assert_equal(List[StringSlice], List[StringSlice])`. Ideally we should be able to generalize over this at some point, it still needs language features.


ConnorGray:
!sync

PR #3901, opened 2024-12-19
[stdlib] Explicitly add the inferred parameter dtype in `min/max`

soraros:
This allows us to disambiguate and use them in higher-order functions like `f[min[dtype=...]]()`.


JoeLoser (approved):

JoeLoser:
!sync

PR #3907, opened 2024-12-23
[docs] Fix incorrect implicit type conversion output in variables.mdx documentation. 

el2e10:
…ntation

- The "Implicit type conversion" section shows the output of implicitly converting an Integer to Float64 as an Integer. It should have been a Float. Updated it to 1.0 instead of 1
- Also added a print statement in that code snippet.


arthurevans (approved):
Good catch--thanks!

arthurevans:
!sync

Issue #3908, opened 2024-12-23
[BUG] `input()` causes mojo to crash when user sends `ctrl-d` with no input


mahiro21h:
### Bug description

sending `ctrl-d` to `input()` when it asks for input causes mojo to crash. this doesn't happen if you _do_ provide some input before pressing `ctrl-d`.

also, this is probably not relevant but im including it just in case, i noticed that `input()` behaved differently compared to python's `input()` when the user presses `ctrl-d` after providing some input (the programs' output is identical).

in this python program, `input()` returns the input after the user presses `ctrl-d` 3 times:
```python
name = input("Who are you? ")
greeting = "Hi, " + name + "!"
print(greeting)
```
output:
`Who are you? mahiroHi, mahir!`

whereas in the mojo program from [the mojo manual](https://docs.modular.com/mojo/manual/get-started/#3-create-and-use-variables), for example, `input()` will return after the users press `ctrl-d` 2 times.
output:
`Who are you? mahiroHi, mahir!`

### Steps to reproduce

- write any program that calls `input()`
- run the program with `mojo run program_name.mojo`
- press `ctrl-d` when the program asks for input

crash backtrace of the program linked earlier that asks the user for their name to greet them:
```
(conway-gol) conway-gol -> mojo run main.mojo 
Who are you? Please submit a bug report to https://github.com/modularml/mojo/issues and include the crash backtrace along with all the relevant source codes.
Stack dump:
0.	Program arguments: mojo run main.mojo
#0 0x00006144c6a616bb llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) Signals.cpp:0:0
#1 0x00006144c6a5f619 llvm::sys::RunSignalHandlers() Signals.cpp:0:0
#2 0x00006144c6a61d3d SignalHandler(int) Signals.cpp:0:0
#3 0x00007005824d81d0 (/usr/lib/libc.so.6+0x3d1d0)
#4 0x00007004e8001e85 
mojo crashed!
Please file a bug report.
Segmentation fault (core dumped)
```

### System information

```shell
OS: archlinux
`mojo -v`: mojo 24.6.0 (4487cd6e)
`magic -V`: magic 0.5.1 - (based on pixi 0.37.0)
```

PR #3919, opened 2024-12-29
[stdlib] Fix `input()` segfaults on EOF

mahiro21h:
pressing `ctrl-d` with no input when `input()` is called causes mojo to crash because `read_until_delimiter()` doesn't check the return value of the C function `getdelim()`. it assumes `getdelim()` always succeeds and so, in the case of an error, it blindly creates a `StringRef` with its length set to the return value - 1 (so the length is -2 in this case). this `StringRef` is then passed to `String()` which in turn passes the `StringRef` to `memcpy()` with a count of -2 and ultimately crashing mojo.

this pr adds a check in `read_until_delimiter()` to check if `getdelim()` failed and raise an error if it does, along with a test to ensure `read_until_delimiter()` continues to behave as it should.

related issue: #3908
closes: #3908


JoeLoser (approved):
Thanks for the fix!

JoeLoser:
!sync

PR #3941, opened 2025-01-12
[stdlib] Use named output for _ListIter __next__() method

bgreni:
- Add a trivial optimization to use a named output for the `__next__()` method in `_ListIter` to save a subtraction operation on every iteration


JoeLoser (approved):
That's a clever use of out parameters here, nice find!

JoeLoser:
!sync

bgreni:
Thank you! I didn't really get them at first, but I'm becoming quite fond of them!

PR #3942, opened 2025-01-13
[stdlib] Enhance Linked List

owenhilyard:
Adds a `LinkedList` type which implements most of the API surface of `List` (also grabbing the tests). This should help with filling in basic data structures in the stdlib.

No iterator because I ran into some nasty issues with origins and I'm not sure how far I can push casting them before I get to UB. 

There's also an instance of `list.__getitem__(0).__getitem__(1).value = "Mojo"` in test_linked_list.mojo since I couldn't find another way to avoid the copy. If someone has a solution please let me know.

### DIFF
--- a/stdlib/src/collections/linked_list.mojo
+++ b/stdlib/src/collections/linked_list.mojo
@@ -17,25 +17,29 @@ from collections._index_normalization import normalize_index
 
 
 @value
-struct Node[ElementType: WritableCollectionElement]:
+struct Node[
+    ElementType: CollectionElement,
+]:
     """A node in a linked list data structure.
 
     Parameters:
         ElementType: The type of element stored in the node.
     """
 
+    alias NodePointer = UnsafePointer[Self]
+
     var value: ElementType
     """The value stored in this node."""
-    var prev: UnsafePointer[Node[ElementType]]
+    var prev: Self.NodePointer
     """The previous node in the list."""
-    var next: UnsafePointer[Node[ElementType]]
+    var next: Self.NodePointer
     """The next node in the list."""
 
     fn __init__(
         out self,
         owned value: ElementType,
-        prev: Optional[UnsafePointer[Node[ElementType]]],
-        next: Optional[UnsafePointer[Node[ElementType]]],
+        prev: Optional[Self.NodePointer],
+        next: Optional[Self.NodePointer],
     ):
         """Initialize a new Node with the given value and optional prev/next
         pointers.
@@ -49,19 +53,29 @@ struct Node[ElementType: WritableCollectionElement]:
         self.prev = prev.value() if prev else __type_of(self.prev)()
         self.next = next.value() if next else __type_of(self.next)()
 
-    fn __str__(self) -> String:
+    fn __str__[
+        ElementType: WritableCollectionElement
+    ](self: Node[ElementType]) -> String:
         """Convert this node's value to a string representation.
+
+        Parameters:
+            ElementType: Used to conditionally enable this function if
+              `ElementType` is `Writable`.
 
         Returns:
             String representation of the node's value.
         """
-        return String.write(self)
+        return String.write(self.value)
 
     @no_inline
-    fn write_to[W: Writer](self, mut writer: W):
+    fn write_to[
+        ElementType: WritableCollectionElement, W: Writer
+    ](self: Node[ElementType], mut writer: W):
         """Write this node's value to the given writer.
 
         Parameters:
+            ElementType: Used to conditionally enable this function if
+              `ElementType` is `Writable`.
             W: The type of writer to write the value to.
 
         Args:
@@ -70,7 +84,9 @@ struct Node[ElementType: WritableCollectionElement]:
         writer.write(self.value)
 
 
-struct LinkedList[ElementType: WritableCollectionElement]:
+struct LinkedList[
+    ElementType: CollectionElement,
+]:
     """A doubly-linked list implementation.
 
     A doubly-linked list is a data structure where each element points to both
@@ -79,12 +95,14 @@ struct LinkedList[ElementType: WritableCollectionElement]:
 
     Parameters:
         ElementType: The type of elements stored in the list. Must implement
-            WritableCollectionElement.
+         CollectionElement.
     """
 
-    var _head: UnsafePointer[Node[ElementType]]
+    alias NodePointer = UnsafePointer[Node[ElementType]]
+
+    var _head: Self.NodePointer
     """The first node in the list."""
-    var _tail: UnsafePointer[Node[ElementType]]
+    var _tail: Self.NodePointer
     """The last node in the list."""
     var _size: Int
     """The number of elements in the list."""
@@ -104,20 +122,36 @@ struct LinkedList[ElementType: WritableCollectionElement]:
         self = Self(elements=elements^)
 
     fn __init__(out self, *, owned elements: VariadicListMem[ElementType, _]):
-        """Initialize a linked list with the given elements.
+        """
+        Construct a list from a `VariadicListMem`.
 
         Args:
-            elements: Variable number of elements to initialize the list with.
+            elements: The elements to add to the list.
         """
         self = Self()
 
-        for elem in elements:
-            self.append(elem[])
+        var length = len(elements)
+
+        for i in range(length):
+            var src = UnsafePointer.address_of(elements[i])
+            var node = Self.NodePointer.alloc(1)
+            var dst = UnsafePointer.address_of(node[].value)
+            src.move_pointee_into(dst)
+            node[].next = Self.NodePointer()
+            node[].prev = self._tail
+            if not self._tail:
+                self._head = node
+                self._tail = node
+            else:
+                self._tail[].next = node
+                self._tail = node
 
         # Do not destroy the elements when their backing storage goes away.
         __mlir_op.`lit.ownership.mark_destroyed`(
             __get_mvalue_as_litref(elements)
         )
+
+        self._size = length
 
     fn __copyinit__(mut self, read other: Self):
         """Initialize this list as a copy of another list.
@@ -125,9 +159,7 @@ struct LinkedList[ElementType: WritableCollectionElement]:
         Args:
             other: The list to copy from.
         """
-        self._head = other._head
-        self._tail = other._tail
-        self._size = other._size
+        self = other.copy()
 
     fn __moveinit__(mut self, owned other: Self):
         """Initialize this list by moving elements from another list.
@@ -157,10 +189,12 @@ struct LinkedList[ElementType: WritableCollectionElement]:
         Args:
             value: The value to append.
         """
-        var node = Node(value^, self._tail, None)
-        var addr = UnsafePointer[__type_of(node)].alloc(1)
-        addr.init_pointee_move(node)
-        if self:
+        var addr = Self.NodePointer.alloc(1)
+        var value_ptr = UnsafePointer.address_of(addr[].value)
+        value_ptr.init_pointee_move(value^)
+        addr[].prev = self._tail
+        addr[].next = Self.NodePointer()
+        if self._tail:
             self._tail[].next = addr
         else:
             self._head = addr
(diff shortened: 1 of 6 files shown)


stdlib/src/collections/linked_list.mojo
```diff
@@ -0,0 +1,657 @@
...
+                return Self.PointerT()
+            else:
+                # print(current)
+                # var c = self._head
+                # for i in range(len(self)):
+                #     print(c)
+                #     c = c[].next
+                return current
+
+    fn __getitem__[I: Indexer](ref self, read idx: I) raises -> ref [self] T:
```
bgreni, line 264: I thought it was decided a while ago to just use `Int` here and let the implicit conversion do the work?

stdlib/src/collections/linked_list.mojo
```diff
@@ -0,0 +1,657 @@
...
+        return self._length
+
+    fn empty(self) -> Bool:
+        """
+        Whether the list is empty.
+
+        Returns:
+            Whether the list is empty or not.
+        """
+        return self._head == Self.PointerT()
```
bgreni, line 385: Why not use `return not self._head` instead?

stdlib/src/collections/linked_list.mojo
owenhilyard: I was following what `List` does. If we want to establish that as a convention, I think that it probably should go in the stdlib docs.

bgreni: Ah I see now it was recently changed _back_ to this way. That's fine I preferred it anyways

owenhilyard:
Move to draft due to existing linked list. I'll circle around after I've merged the two implementations.

lattner:
> There's also an instance of list.__getitem__(0).__getitem__(1).value = "Mojo" in test_linked_list.mojo since I couldn't find another way to avoid the copy. If someone has a solution please let me know.

That sounds super weird.  I'm not sure what is going on, but if you could reduce this down to a small testcase, I would be happy to take a look at some point.

owenhilyard:
@lattner not sure if you want one without the attached data structure, but if I replace stdlib/test/collections/linked_list.mojo:464 with `list[0][1].value = "Mojo"`, I get `error: invalid call to '__setitem__': invalid use of mutating method on rvalue of type 'LinkedList[CopyCountedStruct]'`. There isn't a ton of code under that that isn't normal linked list things, but if that's not minimal enough I can try to reduce it more.

owenhilyard:
cc @abduld as the original author.

JoeLoser:
@lsh I can't seem to assign you to this PR after the GitHub organization move (think there's something up here, FYI @dlumma).  In any case, do you mind taking a look this?  Thanks!

stdlib/test/collections/test_linked_list.mojo
```diff
@@ -101,12 +102,487 @@ def test_setitem():
...
+    assert_equal(l[2], 3)
+
+
+# ===-------------------------------------------------------------------===#
+# LinkedList dtor tests
+# ===-------------------------------------------------------------------===#
+var g_dtor_count: Int = 0
+
+
+struct DtorCounter(CollectionElement, Writable):
```
JoeLoser, line 540: **Question** Thoughts on consolidating this into `test_utils/types.mojo`?  Looks like the `test_list.mojo` has one of these `DtorCounter` types too.

stdlib/src/collections/linked_list.mojo
```diff
@@ -26,26 +26,44 @@ trait WritableCollectionElement(CollectionElement, Writable):
...
+struct Node[
+    ElementType: CollectionElement,
+]:
     """A node in a linked list data structure.
 
     Parameters:
         ElementType: The type of element stored in the node.
     """
 
+    alias NodePointer = UnsafePointer[Self]
```
JoeLoser, line 53:
```suggestion
    alias _NodePointer = UnsafePointer[Self]
```

stdlib/src/collections/linked_list.mojo
```diff
@@ -26,26 +26,44 @@ trait WritableCollectionElement(CollectionElement, Writable):
...
+
     var value: ElementType
```
JoeLoser, line 55: **Suggestion** In a follow-up PR, want to make these private to clean up the data structure?

stdlib/src/collections/linked_list.mojo
```diff
@@ -89,12 +119,14 @@ struct LinkedList[ElementType: WritableCollectionElement]:
 
     Parameters:
         ElementType: The type of elements stored in the list. Must implement
-            WritableCollectionElement.
+         CollectionElement.
     """
 
-    var _head: UnsafePointer[Node[ElementType]]
+    alias NodePointer = UnsafePointer[Node[ElementType]]
```
JoeLoser, line 125:
```suggestion
    alias _NodePointer = UnsafePointer[Node[ElementType]]
```

stdlib/src/collections/linked_list.mojo
```diff
@@ -114,30 +146,44 @@ struct LinkedList[ElementType: WritableCollectionElement]:
...
+        var length = len(elements)
+
+        for i in range(length):
+            var src = UnsafePointer.address_of(elements[i])
+            var node = Self.NodePointer.alloc(1)
+            var dst = UnsafePointer.address_of(node[].value)
+            src.move_pointee_into(dst)
+            node[].next = Self.NodePointer()
+            node[].prev = self._tail
+            if not self._tail:
```
JoeLoser, line 166: **Suggestion** If you swap the conditions, we can stay in the positive boolean checks to avoid a check, e.g.

```
if self._tail:
  self.tail[].next = node
  self._tail = node
else:
  self._head = node
  self._tail = node
```

stdlib/src/collections/linked_list.mojo
```diff
@@ -205,20 +253,138 @@ struct LinkedList[ElementType: WritableCollectionElement]:
...
         """
         var elem = self._tail
+        if not elem:
+            raise "Pop on empty list."
+
         var value = elem[].value
         self._tail = elem[].prev
         self._size -= 1
         if self._size == 0:
             self._head = __type_of(self._head)()
```
JoeLoser, line 270:
```suggestion
            self._head = Self.NodePointer()
```
for consistency with below (now that we have the alias, thanks!).

stdlib/src/collections/linked_list.mojo
```diff
@@ -205,20 +253,138 @@ struct LinkedList[ElementType: WritableCollectionElement]:
...
+        Args:
+            i: The index of the element to get.
+
+        Returns:
+            Ownership of the indicated element.
+        """
+        var current = self._get_node_ptr(Int(i))
+
+        if not current:
+            raise "Invalid index for pop"
```
JoeLoser, line 293: **Suggestion** Use `String.format(...)` to include the index provided in the `Error`.

JoeLoser (commented):
Just did a quick pass of this, looking good!  Left some comments and @lsh will also take a look.

Thanks for working on improving this a ton from the basic implementation that was started prior to your work!  Sorry that we already had a PR in-flight starting the linked list type.

stdlib/test/collections/test_linked_list.mojo
owenhilyard: > Thoughts on consolidating this into `test_utils/types.mojo`? Looks like the `test_list.mojo` has one of these `DtorCounter` types too.

I must have missed that. I started out writing my implementation's tests by copying the entire `test_list.mojo` file and never got around to de-duplicating.

owenhilyard:
@JoeLoser 
> Sorry that we already had a PR in-flight starting the linked list type.

That's fine, I think this shows a need for external stdlib contributors to be able to do some kind of work tracking and for there to be some visibility into non-sensitive work done internally before it hits nightly. Even on the level of a discord channel to toss "I'm working on a linked list" messages into.

stdlib/src/collections/linked_list.mojo
```diff
@@ -232,6 +398,180 @@ struct LinkedList[ElementType: WritableCollectionElement]:
...
+            if next:
+                next[].prev = node
+            current[].next = node
+            if node[].next == Self.NodePointer():
+                self._tail = node
+            if node[].prev == Self.NodePointer():
+                self._head = node
+            self._size += 1
+        else:
+            raise "index out of bounds"
```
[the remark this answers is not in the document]
owenhilyard: Done

stdlib/src/collections/linked_list.mojo
owenhilyard: implemented

stdlib/src/collections/linked_list.mojo
owenhilyard: done

stdlib/test/collections/test_linked_list.mojo
```diff
@@ -101,12 +102,487 @@ def test_setitem():
...
+    var list = LinkedList[Int]()
+    # Test pop with index
+    for i in range(6):
+        list.append(i)
+
+    assert_equal(6, len(list))
+
+    # try popping from index 3 for 3 times
+    for i in range(3, 6):
+        assert_equal[Int](i, list.pop(3))
```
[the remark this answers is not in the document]
owenhilyard: I was getting type errors before, but that seems to be resolved now.

stdlib/test/collections/test_linked_list.mojo
```diff
@@ -101,12 +102,487 @@ def test_setitem():
...
+    v1.insert(len(v1), 1)
+    v1.insert(len(v1), 3)
+    v1.insert(1, 2)
+
+    assert_equal(len(v1), 3)
+    assert_equal(v1[0], 1)
+    assert_equal(v1[1], 2)
+    assert_equal(v1[2], 3)
+
+    print(v1.__str__())
```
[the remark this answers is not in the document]
owenhilyard: Whops, removed.

stdlib/src/collections/linked_list.mojo
```diff
@@ -205,20 +253,138 @@ struct LinkedList[ElementType: WritableCollectionElement]:
...
+
+            # Aside from T, destructor is trivial
+            __mlir_op.`lit.ownership.mark_destroyed`(
+                __get_mvalue_as_litref(node)
+            )
+            current.free()
+            self._size -= 1
+            return data^
+
+    fn pop_if_present(mut self) -> Optional[ElementType]:
```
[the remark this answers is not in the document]
lsh: I think `try` has an association with exceptions. Perhaps `maybe_pop` or `pop_opt`?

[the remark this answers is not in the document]
bgreni: `try_*` does follow the convention in Rust for this sort of thing so at least for me it feels familiar?

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.try_reserve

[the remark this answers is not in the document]
owenhilyard: Rust doesn't really have a name for this kind of thing, since normally everything except for indexing always uses optional/result. However, it felt like a large break with Mojo convention to make the raises variant `try_pop` and the optional one `pop`.

stdlib/src/collections/linked_list.mojo
```diff
@@ -232,6 +398,180 @@ struct LinkedList[ElementType: WritableCollectionElement]:
             curr = curr[].next
         return new^
 
+    fn insert(mut self, owned idx: Int, owned elem: ElementType) raises:
+        """
+        Insert an element `elem` into the list at index `idx`.
```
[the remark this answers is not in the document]
owenhilyard: I'll make it abort on OOM and leave a TODO for later. It won't actually trigger because tcmalloc does not function properly without overcommit, which makes your first notice of OOM the kernel killing your process. 

As an aside, this is also why I need the ability to fully replace tcmalloc or the runtime needs to use an allocator not built on the assumption that getting OOM killed is acceptable. Databases are 1) not supposed to randomly crash and 2) usually the reason the system is low on memory and in a position to give some of it back to keep the system alive.

owenhilyard:
I'm not going to have time to finish up the last few bits of work and deal with the merge conflict today, and potentially not tomorrow.

stdlib/src/collections/linked_list.mojo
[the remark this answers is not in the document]
lsh: No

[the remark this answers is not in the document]
owenhilyard: > I spent some time thinking about this last night, and I'm not sure we even want a `pop(mut self) -> Optional[T]`. I think instead it makes sense to have a `get(ref self, idx) -> Optional[Pointer[T, __origin_of(self)]]`, and that way `my_list.get(0)` would return the `Optional`. I don't think we need mirror `Optional` and `raises` APIs for every method. WDYT?

We want some form of element removal besides "remove the whole list".

I agree that `get` should probably return an optional reference, but I think we want a version which preserves ease of use for things which don't want to unwrap every single item. `try_get` could return a reference and raise to help with cases where you know it's in bounds. Ideally, I'd like some kind of "IndexedCollection" API that unifies `List`, `LinkedList`, slab lists (Linked list of arrays), and the various trees you might use for `O(log(n))` random access.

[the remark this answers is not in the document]
owenhilyard: @lsh @soraros 

I think we want the following options for list-like collections:

* `fn get[I: Indexer](ref self, idx: I) raises -> ref [self] T`
* `fn get_option[I: Indexer](ref self, idx: I) -> Optional[Pointer[T, __origin_of(self)]]`
* `fn pop[I: Indexer](mut self, idx: I) raises -> T`
* `fn pop_option[I: Indexer](mut self, idx: I) -> Optional[T]`
* `fn remove[I: Indexer](ref self, idx: I) raises`
* `fn remove_option[I: Indexer](ref self, idx: I) -> Optional[None]` # this one for API consistency

I think that provides coverage for most things people will want, and helps to work around the ergonomics issues. I'm not sure on the function names, so we can discuss that. My hope would be that #3946 might be able to help abstract over this, so we don't need yet another "abort" version.

JoeLoser:
!sync

stdlib/src/collections/linked_list.mojo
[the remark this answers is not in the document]
lsh: We talked a little about this during one of our team meetings, and for now for this PR we think `maybe_pop` is the name we'd prefer. Thanks for your work on this!

owenhilyard:
@JoeLoser I pushed one last change I had held back, since it looks like you want to push the discussion on the "standard list-like API" for later and I was waiting for that to resolve.

JoeLoser:
!sync

JoeLoser:
CC: @lsh for final review/sync etc

PR #3943, opened 2025-01-13
[stdlib] Make atomics sharable

owenhilyard:
Prior to this, most atomic operations required a mutable reference, which meant you had statically guarenteed mutual exclusion, which means you don't need atomics. This fixes those operations to only require an immutable reference, allowing them to be used in multiple threads.


JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

PR #3945, opened 2025-01-13
[proposal] Structured Async for Mojo

owenhilyard:
Proposes to add structured async to Mojo, following in the the Rust tradition of async since Mojo has the ability to fix many of the issues with Rust's async, some of which are ecosystem inflicted and some of which are inflicted by old decisions made about the language (such as that any value may be leaked). I think that this interface offers a lot more flexibility than the current one does for high performance code, while providing a better path for gradual evolution. It does have some language-level dependences, namely unions, and requires the introduction of `Send` and `Sync` traits, which are used to control data movement between threads.

### DIFF
--- a/proposals/mojo-structured-async.md
+++ b/proposals/mojo-structured-async.md
@@ -559,7 +559,7 @@ async fn evil_fanout(data: List[Foo]):
 
 As you can see, entirely by accident, Mojo has solved one of the largest
 ergonomics issues with Rust async, that being the proliferation of `'static`
-bounds, entirely by accident. With the introduction of linear types, simply
+bounds. With the introduction of linear types, simply
 making the `ChildTask` struct linear stops all of the issues that made this API
 impossible to do soundly in Rust. In fact, members of the Rust community have
 suggested adding linear types to Rust explicitly to solve this issue. All of the
@@ -617,7 +617,7 @@ analysis in the compiler when assembling the type. I personally think that a
 fairly large amount of compile-time overhead is acceptable for this feature
 since it should be in the "you only pay for it if you use it" category, and
 because it should have runtime performance benefits. Another downside of this
-fusion is that it  and it prevents recursive coroutines without adding
+fusion is that it prevents recursive coroutines without adding
 indirection. For the most part, the only solutions I've seen to this issue
 either involve forcing indirection (C++) or stackful coroutines (Go), neither of
 which I am particularly fond of due to the overhead, pointer chasing for C++,
--- a/proposals/mojo-structured-async.md
+++ b/proposals/mojo-structured-async.md
@@ -1,0 +1,665 @@
+# Mojo Structured Async
+
+Owen Hilyard, Created December 17, 2024
+
+## Background (Async/Await)
+
+Async/Await is one of the programming paradigms developed in service of the
+[C10K](https://en.wikipedia.org/wiki/C10k_problem) class of problems, where a
+program needs to handle some large number of concurrent connections. Originally
+this was around 10 thousand, now commonly 10 million or 100 million, but
+generally it is large enough such that blocking IO is no longer sufficient. It
+provides a mechanism through which the compiler can automate the creation of a
+state machine, a formerly manual process, and as a result allows a user to write
+sequential code that is broken up by "yield points", where the compiler may
+suspend the task while IO is completed. However, as typically implemented this
+results in a variety of issues.
+
+## Motivation
+
+Async/await has a few issues for high performance code as normally implemented,
+but I think that Mojo can address those problems. Before I try to introduce
+solutions, let's get on the same page as to the problems. The first one, which
+prevented the adoption of this paradigm in many domains, is the lack of control
+over memory. This is caused by heap allocations being made on behalf of the
+developer, potentially inside of a hot loop, with no way to handle the failure
+to allocate the coroutine frame. Many coroutines are also unable to be allocated
+on the stack, due to the requirement to fully type erase coroutines to run them
+in a generic executor, which is another reason for the heap allocations. Rust's
+coroutine model helped to solve some of these issues, since it does not
+inherently require type erasure or heap allocation. It allows for storing
+futures in caller-allocated storage, giving the programmer control over the
+placement of futures, and imposes no requirements on an async executor, allowing
+for single-future executors such as `futures::block_on` from the `futures`
+crate, or custom executors which execute futures from a closed set by storing
+the in-progress futures as a sum type, or more general executors which perform
+type erasure, such as `tokio` and `glommio`. However, in doing so Rust has
+uncovered contention between the classic work stealing scheduler, thread safety,
+and user ergonomics. Additionally, the general `async`/`await` model has also
+been criticized for causing function coloring and viral annotations, causing
+ecosystems to frequently drift towards forcing beginners to learn
+`async`/`await` and multithreading before performing simple tasks such as
+performing a database query due to the requirement for authors to duplicate
+large parts of their libraries if they want to support both `sync` and `async`
+code. After substantial discussions with other members of the community, I think
+that the model presented here will help increase the expressiveness desired by
+performance-sensitive and correctness-critical users, decrease the
+implementation burden on library authors, make the Mojo ecosystem more friendly
+to new users and be reasonable to implement inside of the Mojo compiler.
+
+## Background: Rust Futures
+
+```rust
+pub trait Future {
+    type Output;
+
+    // Required method
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output>;
+}
+```
+
+Rust treats [`Future`](https://doc.rust-lang.org/std/future/trait.Future.html)
+as a trait. You may notice that Rust requires `self: Pin<&mut Self>`. This is
+because Rust's futures are also almost always self-referential, and Rust does
+not have move constructors, so
+[`Pin`](https://doc.rust-lang.org/std/pin/struct.Pin.html) is needed to make
+moving them around safe by giving them a chance to do pointer fixups before they
+are invoked. We don't need that, so we can toss it out unless we want to discuss
+whether it's a good idea to delay fixups until the last possible second to defer
+work. Note that for any coroutine allocated on the heap (and not moved), `Pin`
+is pure overhead. `Context` mostly serves to store a `Waker` which is used to
+notify the async executor when a task is ready to be executed, primarily for
+things like epoll where the executor isn't getting information from the OS about
+when IO is done. The `Waker` type is implemented using a data pointer and a
+vtable pointer, something which can likely be made generic so it can be
+devirtualized where possible.
+
+The executor calls `poll` on the pinned future, providing a context struct, and
+gets back a `Poll` sum type, which is either a `Poll::Pending`, meaning the
+future is still working but has given control back to the executor and doesn't
+need to run until the `Waker` signals the executor, or a `Poll::Ready(T)`,
+meaning that the coroutine is finished. For Mojo, we can expand this `Poll` enum
+to allow `Poll::Pending` to have data attached to it, allowing the
+implementation of async generators.
+
+However, the most important part of this is that `Future` is a normal trait, and
+can be attached to normal objects. This means that a user can write their own
+futures timeouts (ex: a future that returns Ready if polled more than 1 second
+after it was created), without much difficulty. This also makes it easier to add
+extra capabilities on top of futures in user-space, for instance, a future may
+be possible to run on both the CPU and the GPU if there is support from the
+executor, allowing a program to take advantage of wherever there is free
+compute. It also means there isn't any need for a full coroutine to manage
+something like an epoll io request, you just implement `Future`.
+
+This also allows storing the coroutine frame wherever you want. If you want to
+make a vtable, this single-function trait turns into exactly what the current
+Mojo `Coroutine` type looks like right now, a data pointer and a function
+pointer, and the compiler-generated coroutines, which know the layout of the
+vtable, can manipulate the vtable as needed to keep the same performance.
+However, it leaves the ability for someone who can beat the compiler to
+hand-write their own state machine in the good old "switch over a state tag"
+style and still run in the normal async executor using this single entry point.
+
+However, they have an issue as implemented in executors like `tokio`.
+
+## Foundational Async
+
+Since you have work stealing, any suspended task can be moved to another thread.
+This means you get a bunch of rules about what can and cannot be carried across
+suspend points, chief among them being locks and other synchronization
+primitives, which happen to occur more frequently in async code than sync code.
+In Rust, this leads to `Send + Sync + 'static` bounds on most async API
+boundaries. `Send` means a type is safe to send between threads, and `Sync`
+means that an immutable reference to the type is safe to send between threads.
+`'static` means that the value lives forever, since the stack frame the
+coroutine is temporarily using could go away before before new future runs. When
+combined, this means that you suddenly end up with a lot of `Arc<Mutex<T>>`
+floating around for your resources, even if you know that they live long enough.
+There is a solution to this, but let's lay a foundation first.
+
+This means we have an executor handle interface (or one hidden behind functions
+which access static state), which looks something like this:
+
+```mojo
+trait Waker:
+    # Wake up the associated coroutine
+    fn wake(ref self):
+        ...
+
+# Holds ways for the coroutine to talk to the executor and schedulers.
+# We can add things to this in the future via extension traits as needed
+# to keep the coroutine trait signature intact.
+trait CoroutineContext:
+    alias WakerType: Waker
+    
+    fn get_waker(ref self) -> ref [self] Self.WakerType:
+        ...
+
+# Coroutine becomes a trait. Anyone who wants to hand-roll a state 
+# machine can. The coroutine was already more or less a vtable anyway.
+trait Coroutine:
+    # This is likely to be a tagged union, but for "coroutines" which
+    # hot run and are done, this can just be the return type.
+    # We can create a version of the trait for "yields nothing until
+    # it returns", and another for returning a tagged union. 
+    alias RunReturnType: AnyType
+    
+    # State machine transition. Raises if the coroutine is finished executing.
+    fn run[C: CoroutineContext](mut self, mut context: CoroutineContext) raises -> RunReturnType
+        ...
+
+# Swappable async schedulers. If I have strong opinions about how scheduling
+# works, or so I can have sum types as instead of boxing my coroutines, I
+# can make my own. 
+trait AsyncScheduler:
+    # Note the lack of Send/Sync bounds. This is intentional since all 
+    # executors should be capable of handling local-only tasks. This 
+    # frees us from many of the issues Rust has. 
+    # This should be parametric.
+    alias AsyncSchedulerTaskType: Coroutine
+
+    fn get_next_task(mut self) -> Optional[Self.AsyncSchedulerTaskType]
+        ...
+
+trait AsyncTaskSpawner:
+    # This should also be parametric
+    alias AsyncTaskSpawnerTaskType: Coroutine
+
+    fn spawn_task(mut self, owned co: Self.AsyncTaskSpawnerTaskType):
+        ...
+
+trait AsyncExecutor(AsyncScheduler, AsyncTaskSpawner):
+    # More parametric
+    alias TaskType: Coroutine + Movable
+    alias AsyncSchedulerTaskType = Self.TaskType
+    alias AsyncTaskSpawnerTaskType = Self.TaskType
+
+    # default impl
+    fn run_next_task(mut self, out ret: Self.TaskType.RunReturnType):
+        var task = self.get_next_task()
+        if task:
+            out = Optional[Self.TaskType.RunReturnType](task.run())
+            if not task.is_done():
+                self.spawn_task(task)
+            else:
+                # end linear type
+                destroy task^
+        else:
+            out = Optional[Self.TaskType.RunReturnType]()
+```
+
+This interface lacks error handling, since I think that another proposal I have
+in the works will help with that, and if that proposal or some other form of
+"generic over `raises`" is not accepted, then all of the functions need to
+raise. The goal of this interface is to provide the **MINIMUM** traits that are
+needed to represent an async executor. I don't think there's a reasonable
+version of async/await where you don't have the ability to grab the next task to
+run, and the ability to spawn a task. Remember that a coroutine which does not
+return values can return a zero-sized value, such as an empty struct. Ideally,
+most of these associated types will move into parameters on the trait so that
+the trait can be implemented multiple times for a single executor (for instance
+one which type erases). This is what I anticipate most coroutines will do, but
+this allows for "run this single async function" async executors to avoid
+indirection and still have return value placement optimizations. Of course, I
+would go define a few other options very quickly:
+
+```mojo
+# Don't move the coroutine
+trait RefAsyncScheduler(AsyncScheduler):
+    alias RefAsyncSchedulerTaskType: Coroutine
+
+    fn get_next_task_ref(mut self) -> \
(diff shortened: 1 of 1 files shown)


szbergeron:
Carrying some external discussion in for context on Waker--

The core motivation is that when building something of this sort, you want to spend as little time processing useless data as possible compared to processing useful data. This is simple in theory, but in practice the actual "knowledge" of what can vs can't be effectively progressed is sparse, disaggregated, and rarely has a clear way to even collect that state outside of just "poking" everything that "wants" to make progress.

Wakers partially solve this on their own. For many "boundaries" (things like channels, queues, timers), wakers can be more or less tossed over the barrier to be collected by the other side. When the "other side" is itself administrated by a coroutine registered with an active executor (especially the same executor) the system nicely passes control away from something that can't progress to something else that can, and doesn't waste more time on something that can't progress until _It_ can.

This works very nicely for "closed" systems that can be reduced almost entirely to a single computation, and where there aren't other computational priorities that exist outside of the executor. Unfortunately, the outside world also exists. The most performant forms of I/O nowadays are not interrupt driven, and do not have a way to directly signal in and poke a waker. Also, many systems are going to have multiple priorities competing for attention, and they may not (for various architectural reasons) exist within the same executor _even though they reside within the same process_.

This motivates another cut along which to aggregate/concentrate useful "threads to pull": subsystems. You may have many places that you need to wait for an operation to complete within io_uring, or for some condition to occur in some region of shared memory. This requires busy polling, and has no straightforward (naive) way for a waker to be used to drive a computation. If this is reduced to every future that is waiting on an operation busy polling, we end up with an excess of duplicated computation being done with no progress (bad!).

What's the alternative? Busy polling must occur, but can be moderated and deduplicated. It may or may not even need to happen within the same executor, but it must happen _somewhere_. If we create a way for subsystems to be statically registered, or even dynamically registered with some prioritization flag for executors to poll them with priority more closely matching their utility (how many coroutines depend on them progressing), we can then treat them almost the same way as any other async "barrier" (such as channels, async mutexes, timers). This way, a coroutine corresponding to each subsystem can itself collect wakers from other coroutines that would otherwise busy poll on their own, and itself act as a form of scheduler, as alluded to in the proposal.

The implementation specifics have a bunch of intertwined tradeoffs, but as far as we can tell these broad strokes are the limit of how minimal the overall type structure can be made without necessarily sacrificing significant performance to duplicated computation (no-progress polling).

owenhilyard:
Continuing on from what @szbergeron mentioned, there are a lot of IO mechanisms which are completion based. These busy-polled "subsystem" futures, which ideally can be spawned in a way that the executor is made aware of them as special, can help to de-duplicate a lot of that polling since most of these mechanisms deliver the results through some kind of queue. For epoll-like things, you still want a more central system place to handle polling the eventfd and waking things up. The current API doesn't really have a good way to support this kind of flexibility, so it all but guarantees we have the same executor lock-in issues that Go has and Rust has with Tokio, and doesn't leave room for libraries to experiment with different designs that the stdlib executor might benefit from or for high performance applications to have an executor which meets their own needs.

lattner:
This is a super interesting proposal, thank you for writing this up.  It is a bit over my head, but all of the framing sounds right to me and I think you're tackling the right problems.  I'm really happy you're seeing how the individual ingredients (move ctors, linear types, safe references) come together to build something special!  I think the major conceptual concern I have (mentioned below a couple times) is whether "pluggable executors" are actually a desirable thing?


Doing a quick review of the document:

> We don't need that, so we can toss it out unless we want to discuss whether it's a good idea to delay fixups until the last possible second to defer work.

Let's toss it out - simpler is better, and we've planned to not need this.

> These are concerns I don't think we want in the language, so my solution to this problem is to define a portable, extensible interface, and then provide a default "good enough" implementation hidden behind that interface which can either be swapped (at compile time or runtime) for a newer, better executor (ex: doing IO with epoll -> io_uring -> DPDK + io_uring or io_uring + SPDK -> DPDK + SPDK) or upgraded over time.
...
> > allowing for ecosystem interoperability and preventing "tokio lock-in" where substantial portions of the ecosystem are tied to a single executor. 

I'm not an expert on many things in this space, but your argument to push this into the library aligns with my general instincts.  That said, I'm not sure how important it is in this case.  Rust took a pretty unopinionated view on concurrency runtimes, enabling tokio and others to experiment and build different worlds.  This seemed like a good idea - don't have to make a binding decision early, but then fragmented the software ecosystem, because it turns out we all really just want one /good/ concurrency runtime to build on top of.  

Swift learned from this and did its own concurrency engine, adding a bunch of bells and whistles that UI programmers need (e.g. QoS features), and a nice cancelation approach.  Have you looked at its runtime and how its approach to structured concurrency? 

Other notes: +1 for building on top of io_uring as default on linux though, it seems like clearly the right low-level thing to build on.  As you say, other platforms will have other things and we should have some amount of implementation flexibility to provide a portable layer over them. I find "entirely by accident" to be funny :), +1 for connecting to linear types!

> There are a few downsides to this. First, we would need trait objects or hand-written vtables before this could be used for a general purpose async.

I don't think that's a downside, just a dependence.  I think we should take the long view and build the right thing even if it takes time.  If this is the right design, then let's get existentials / trait objects done and then build on top of it.

>  I personally think that a fairly large amount of compile-time overhead is acceptable for this feature since it should be in the "you only pay for it if you use it" category,

Wouldn't this affect /every/ async function?  On the one hand, yes you're only paying for it there and async makes you opt into it, but for certain code this will be pretty common.

> There is also the "trait explosion" issue, but I don't think that's avoidable with a capability-based API. 

This is the thing I'm most concerned about with this proposal: it isn't really the number of traits that I'm worried about, but the amount of synthesized code. What is the impact on code size, compile time, and other dimensions as a consequence for this flexibility?

> bit of a bump in user-facing complexity over Go's "make everything async" approach. 

We've already accepted the "what is your functions color" approach of Python here.  Let's not worry about go specifically.

I agree with your framing of accidental vs inherent complexity!

-Chris

owenhilyard:
> This is a super interesting proposal, thank you for writing this up. It is a bit over my head, but all of the framing sounds right to me and I think you're tackling the right problems. I'm really happy you're seeing how the individual ingredients (move ctors, linear types, safe references) come together to build something special! I think the major conceptual concern I have (mentioned below a couple times) is whether "pluggable executors" are actually a desirable thing?
>
> > These are concerns I don't think we want in the language, so my solution to this problem is to define a portable, extensible interface, and then provide a default "good enough" implementation hidden behind that interface which can either be swapped (at compile time or runtime) for a newer, better executor (ex: doing IO with epoll -> io_uring -> DPDK + io_uring or io_uring + SPDK -> DPDK + SPDK) or upgraded over time.
> > ...
> > > allowing for ecosystem interoperability and preventing "tokio lock-in" where substantial portions of the ecosystem are tied to a single executor.
> 
> I'm not an expert on many things in this space, but your argument to push this into the library aligns with my general instincts. That said, I'm not sure how important it is in this case. Rust took a pretty unopinionated view on concurrency runtimes, enabling tokio and others to experiment and build different worlds. This seemed like a good idea - don't have to make a binding decision early, but then fragmented the software ecosystem, because it turns out we all really just want one /good/ concurrency runtime to build on top of.

The danger of only having one is that it needs to serve every need in the ecosystem. Everything from running on a tiny RISC-V core with 64K of dedicated memory which acts as a management processor for an accelerator to manage the running kernels and check the command queue occasionally, all the way up to 32 socket servers where it is expected to handle tens of millions of network requests per second.  It also means factoring in things like Intel's DLB fixed-function accelerator, which offloads much of the queue management of a simple actor framework to hardware and can be used to aid in scheduling. 

I think it should be possible to define a robust enough interface that almost all use-cases fit behind it, meaning only highly specialized libraries which need to use more peculiar executor features not contained in the standard library interface would be executor specific. 

> Swift learned from this and did its own concurrency engine, adding a bunch of bells and whistles that UI programmers need (e.g. QoS features), and a nice cancelation approach. Have you looked at its runtime and how its approach to structured concurrency?

I haven't had chance to look at Swift. Cancellation is something I was putting off until later, but I'm interested to see solutions in that area and I'll take a look. 

> Other notes: +1 for building on top of io_uring as default on linux though, it seems like clearly the right low-level thing to build on. As you say, other platforms will have other things and we should have some amount of implementation flexibility to provide a portable layer over them. 

I think that io_uring is a sensible default, but depending on how much chasing "state of the art" is desirable, XDP sockets and/or [DPDK](https://www.dpdk.org/) may be helpful as enhanced capabilities for networking, since io_uring has continued to give performance reports in terms of "% of DPDK performance". If the goal is for Mojo/MAX to be able to be in the same order of magnitude of performance as RDMA when it doesn't have RDMA-capable hardware, DPDK is probably the best path forwards for that. XDP sockets are a step back from DPDK which are more portable than DPDK but tend to be a bit slower since you can't use the NIC's hardware offloads as effectively. io_uring is close enough to SPDK, DPDK's sister library for storage that I don't think it's worthwhile to have wide support for SPDK.  

> I find "entirely by accident" to be funny :)

Oops, I doubled that up during an editing pass at some point and missed it. 

> > I personally think that a fairly large amount of compile-time overhead is acceptable for this feature since it should be in the "you only pay for it if you use it" category,
> 
> Wouldn't this affect /every/ async function? On the one hand, yes you're only paying for it there and async makes you opt into it, but for certain code this will be pretty common.

It does affect every async function, and may force a topological order (following the call graph) onto determining the layout of coroutine frames in order to place them inside of each other. Rust does do a form of this, and I'm not aware of performance complaints with that part of the compiler, so there may be prior art to draw from. 

> > There is also the "trait explosion" issue, but I don't think that's avoidable with a capability-based API.
> 
> This is the thing I'm most concerned about with this proposal: it isn't really the number of traits that I'm worried about, but the amount of synthesized code. What is the impact on code size, compile time, and other dimensions as a consequence for this flexibility?

My understanding, which may be incorrect, is that the amount of generated code is very similar to C++'s approach, which is most similar to what Mojo has now. What may occur is a few functions get inlined, resulting in a few large jump tables instead of a lot of smaller ones. 

I think that the current approach in Mojo may be slightly better for compile time, because it will enable more parallelism, but I haven't spent enough time inside of a compiler to quantify it and I think you (@lattner) are in a far better position to evaluate that. I know there will be a non-zero compile-time cost, but there's also a chance that the ability to do some of the things this enables may have knock-on effects which help in other areas, such as reducing the input size to some items further along in the compilation process via reducing the number of coroutines they need to consider. I've done my best to stick reasonably close to proven approaches as far as things which I think might cause compile-time issues, but I haven't spent enough time writing compilers to know that I caught everything. 

This flexibility also has some issues for recursion, since doing recursion in this model, aside from tail recursion, requires the compiler to automatically add indirection to prevent infinite-size structs. It can also cause issues since it may lead to users accidentally moving or copying multi-kilobyte structs.

lattner:
I'd suggest separating two different things:
1) The ability for an app to provide its own executor engine.
2) The ability for a platform to define its own executor.

The first is the problem that rust fell into.  It has a lot of challenges when two libraries assume different executors.  The later is something that is obviously general goodness (io_uring doesn't exist on windows or your_kernel_driver or a gpu).

lattner:
> I think that the current approach in Mojo may be slightly better for compile time, because it will enable more parallelism, but I haven't spent enough time inside of a compiler to quantify it and I think you (@lattner) are in a far better position to evaluate that.

Actually, I'm not really the best to eval this.  I do have some context in the current impl, but a lot is changing and I'm not on top of everythign.  We're focused on getting other fundamentals nailed down (eg reference system, closures, etc) which this builds on top of though, so when this comes into focus we can better evaluate it.

owenhilyard:
> I'd suggest separating two different things:
> 
>     1. The ability for an app to provide its own executor engine.
> 
>     2. The ability for a platform to define its own executor.
> 
> 
> The first is the problem that rust fell into. It has a lot of challenges when two libraries assume different executors. The later is something that is obviously general goodness (io_uring doesn't exist on windows or your_kernel_driver or a gpu).

So, you would make platforms something like x86_64-unknown-linux-gnu-dpdk-io_uring (meaning x86 linux with DPDK for networking and io_uring for storage)? One of the issues with this is that [DPDK does actually work with GPUs](https://www.youtube.com/watch?v=_-_Wlxjq0co), but it requires host-side setup since the NIC needs to be told to enable P2P as well. In theory, given a network card which supports peer to peer and a GPU which also does, this is a portable feature, although it's not widely supported right now. 

I personally think that some of the issues stem from the lack of a base interface that all executors implement. If there's a basic interface for sleeping, spawning a new task, UDP sockets, TCP sockets, synchronization primitives and common file operations, I think most libraries will be portable. I think Rust started on this far too late, and that we can side-step the problem of "what if this isn't the right interface for this operation?" by heavily decomposing the traits (ex: 1-2 functions in most of the traits in the interface), and encouraging libraries to only depend on the interface they need, instead of a whole executor. This should minimize the amount of incompatibility, and let the stdlib executor focus on the common 90% of cases while still leaving the door open for "I need hard real time with no heap allocations" executors for people with more specialized needs. My hope is that if the stdlib sets an example, other executors will be more willing to split out functionality in a similar way, especially if the "So you want to write an executor?" docs include the reasons why it's desirable to decompose like this. 

The other reason I want to push for custom executors is that I see devices getting smarter, and I can easily see a future in which the primary purpose of a CPU is to schedule an array of accelerators (already somewhat the case in ML clusters). If we go with a single stdlib-owned executor, instead of the "executor subsystem" approach, where a few basic ones like io_uring and kqueue are provided by the stdlib, there will need to be handling for every single device MAX supports which has some kind of "async kernel" support. I think that may threaten the ability for a device vendor to do independent device bring-up for MAX and provide the thing to manage async ops on their hardware as a library, which could lead to forks of MAX for specific hardware, similar to vendor forks of pytorch. I don't think this is desirable, so I want to provide a way to avoid that.

owenhilyard:
> > I think that the current approach in Mojo may be slightly better for compile time, because it will enable more parallelism, but I haven't spent enough time inside of a compiler to quantify it and I think you (@lattner) are in a far better position to evaluate that.
> 
> Actually, I'm not really the best to eval this. I do have some context in the current impl, but a lot is changing and I'm not on top of everythign. We're focused on getting other fundamentals nailed down (eg reference system, closures, etc) which this builds on top of though, so when this comes into focus we can better evaluate it.

Ok, let's circle back around to cost evaluation once the ground work is in place.

npuichigo:
Can we leverage the P2300 proposal from C++ community? https://github.com/NVIDIA/stdexec

owenhilyard:
> Can we leverage the P2300 proposal from C++ community? https://github.com/NVIDIA/stdexec

My general feeling on P2300 is that it specifies more details than I think is wise. Now, because it's C++ with multiple proposals it must do that, but it paints C++ further into a corner over time. 

This proposal is almost more about how the compiler transformations of async to coroutines happens than executors. For executors, I handle more of the IO side in #4728.

YkdWaWEzVmphR1Z1:
@owenhilyard I have concerns about adopting the Rust approach, which forces everything to be a state machine.

Stackless coroutines share the same stack but their state across await points has to be stored in a structure. 
This means that coroutines can become quite large, leading to a large number of memcpy operations.
Even worse is the fact that executors have to store futures in memory when spawn is called. `Async-task`, a library for executors used by smol and other thread-per-core executors, allocates large futures (>=2048B) on the heap.
Therefore, if you have to use spawn, you effectively lose the memory savings of stackless coroutines. If you don't  use spawn, executors cannot efficiently schedule your futures.
LLVM cannot change the memory layout of structures, which in turn means that the Mojo compiler is solely responsible for optimizing the size of coroutines.

The fact that the compiler completely transforms your code also means that stackless coroutines are more difficult to debug.

You are proposing a minimal interface with another layer of indirection.
I propose a maximum interface where all necessary I/O operations are performed directly on the executor.
In case of IoUring, the executor would have access to the submission queue and could efficiently call io_uring_enter.
Removing this layer of indirection solves all these issues.

I prefer an approach that focuses on the operating system and system calls, ensuring that optimal system calls are generated first. After that the overhead from branch misses, cache references, instructions, cache misses and code size should be evaluated. 

I do not consider it necessary to thoughtlessly define interfaces that favor stackless coroutines and harm the ecosystem, but only benefit low-performance hardware such as a Cortex-M with 512 KB RAM.

owenhilyard:
@bGVia3VjaGVu 

> This means that coroutines can become quite large

I agree

> leading to a large number of memcpy operations. 

I don't follow this. You don't have to store large coroutines on the heap, you can just stick them behind an `OwnedPointer` (`std::unique_ptr` in C++ terms). However, some kinds of coroutines are relatively information sparse, and as a result you the developer can decide that a coroutine that really only stores a pointer and a few integers can be stack allocated just fine. You can also store large coroutines on the stack if you want to, although yes that will lead to memcpys and is generally inadvisable unless you have a very good reason. The compiler can be tuned to re-use memory locations or insert padding in order to get rid of shifting data around, at the cost of slightly increased memory usage. This essentially turns the struct into a stack. 

> Even worse is the fact that executors have to store futures in memory when spawn is called. 

Outside of very, very niche cases where you can fit a future inside of registers, I don't see how you'd avoid using memory of some sort to store a coroutine. 

> `Async-task`, a library for executors used by smol and other thread-per-core executors, allocates large futures (>=2048B) on the heap. Therefore, if you have to use spawn, you effectively lose the memory savings of stackless coroutines. 

You don't, because much of the memory savings comes from two places:

1. Only keeping around the information you will need in the future. 
2. Knowing exactly how much state you need at any given time. 

Stackful coroutines don't let you "pop" from the middle of the stack if you don't need that information any more. They also tend to want you to allocate fixed-sized blocks of memory, such as in the segemented stacks of Golang. This can help in some cases, such as when using an arena allocator for the blocks of stack, but you can get similar benefits out of adding a bit of indirection into stackless coroutines to make them form a singly linked list of roughly fixed-size blocks of memory. However, if you have a large number of similar coroutines, as is common in some kinds of systems (CDNs, NGFWs, Databases, etc), then you can pool those and get exact sizing benefits. 

> If you don't use spawn, executors cannot efficiently schedule your futures. 

That depends on how the executor is built. Having a "small' executor which just shoves everything onto the stack to let you do a little bit of IO multiplexing in an otherwise synchronous program isn't uncommon, and that can trivially be done just with references. For example, an executor could take some number of futures as part of its constructor and tie it's lifetime to the lifetime of the futures, ensuring that all of them remain valid allocations past when the executor is done running all of them to completion. 

> LLVM cannot change the memory layout of structures, which in turn means that the Mojo compiler is solely responsible for optimizing the size of coroutines.

I don't see this as an issue. By having the Mojo compiler do it, we can surface options to control various tradeoffs such as memory usage vs amount of memcpy vs indirection. Once Mojo gets reflection, we could even let users write their own coroutine optimizers. 

> The fact that the compiler completely transforms your code also means that stackless coroutines are more difficult to debug.

I consider this more of a tooling issue. If you compile C, C++ or Rust code with -O3, you run into similar problems where the compiler eliminates too many values. Debuggers like GDB or LLDB were not really designed to deal with async of any form, stackless or otherwise, and even stackful coroutines often break debuggers due to many people using `-fomit-frame-pointer`. I also have some ideas on how to structure IO such that one could swap the IO to a "sync" mode by using effect generics and get something that looks a lot more like normal sync IO debugging, which should at least help with logic errors. With logic errors and data races covered, the main remaining thing is deadlocks. There are a lot of options we can borrow from database theory to help detect and deal with deadlocks (transaction managers have to deal with them all the time), although they come with some overheads so they would need to be optional. 

> You are proposing a minimal interface with another layer of indirection. I propose a maximum interface where all necessary I/O operations are performed directly on the executor. In case of IoUring, the executor would have access to the submission queue and could efficiently call io_uring_enter. Removing this layer of indirection solves all these issues.

I'm proposing to use generics to solve this, which, while a layer of indirection, won't have any more performance penalty than the API requires. See #4728 for more of my thoughts on this. I want to expose the full power of the underlying IO APIs, but I want to encourage portability where possible. Golang already got caught in a bad place where it was too tightly bound to epoll and it doesn't even directly expose the primitives of the API. I agree the executor should have direct access to the submission queue, and in fact an io_uring IO Engine would likely be a wrapper around the sq, the cq and some book-keeping data structures. I have a few reasons for wanting that layer of indirection in place.

* Older kernels where io_uring doesn't yet fully replace the epoll APIs
* Some organizations consider io_uring "too radical and too new" to be considered secure yet, and will remain with epoll for a while
* io_uring forces you to do book-keeping that the kernel does for you with epoll. This makes porting applications harder, and for some applications the performance hit is worth the simplification. 
* io_uring is not actually the fastest API for ethernet networking (DPDK easily takes the win here) and it has higher CPU requirements for large piles of flash memory (like a JBOF) than SPDK. It also can't do Infiniband, Omnipath or Ultra Ethernet very well right now.
* Depending on how "high level" the IO interface grows to, something like MPI might eventually fit under it as an adapter, letting Mojo use specialized supercomputer interconnects transparently, or make use of RDMA/RoCE if it's available to the user. 
* I don't think that io_uring will be the final form of IO APIs. I was at Supercomputing 2025 last week (part of why it took me so long to respond), and several vendors had "computational storage devices", which enable offloading some parts of an IO request to an SSD. For example, the SSD might be able to deliver only those rows of a CSV file which have a particular value in a particular column, massively reducing the IO bandwidth requirements. Similarly on the network side, NVIDIA SHARP seems to only be the beginning of the network getting smarter, and NICs are already far more capable than io_uring exposes (ex: You can make most modern NICs parse RPCs for you and shard requests among CPU cores based on something like the requesting user ID). 
* It is actually very useful to be able to take the performance hit and serialize everything so that you have an easy flow of data to inspect while debugging. 
* Over-tuning on io_uring means that macOS and Windows will be problematic to get working at full speed. While the main APIs will be bent towards SOTA, which is on Linux, I'd like to at least beat Boost ASIO on Windows. 
* That indirection also provides a great place to mock things for testing, allowing users to test under absurd rates of packet loss, data corruption, or simply seeing what happens if DNS fails.

I appreciate a direct to the API interface a lot, just look at my work with DPDK, which has almost totally abandoned traditional IO APIs in favor of more performance, but Mojo gives us the tools to get at that performance in a way which is more usable to normal developers, possibly at a slight performance penalty. Of course you can just write a specialized executor which only does exactly the kinds of IO you want and which is heavily co-designed with your program. 

> I prefer an approach that focuses on the operating system and system calls, ensuring that optimal system calls are generated first. After that the overhead from branch misses, cache references, instructions, cache misses and code size should be evaluated.

I'm put Rust coroutines in the hot path of a system that does 70 million RPS on a single core, and in plenty of other places in the multi-million RPS per core range with more significant work per request. They work fine. If anything, I would say that the overhead of switching the stack pointer and potential costs of hot loop allocations in a segmented stack make stackful coroutines non-viable for high performance networking and storage operations if you're pushing for SOTA performance. Using state machines in the form of stackless coroutines means that you can store the state of the operation stream in a lookup table and can likely operate via references after you prefetch the relevant parts of the coroutine into L1d. If you have a lot of state in flight, you have a lot of state in flight and the language can't do a ton about that. 

Stackless coroutines are essentially a compiler-assisted version of the "while loop, switch, and array of state structs" event loops common in many C and C++ programs when paired with a more specialized executor. Unlike Rust, we don't need to hide what a coroutine is because we can actually name the type, so a lot of the "magic" of Rust's version can be thrown out. 

> I do not consider it necessary to thoughtlessly define interfaces that favor stackless coroutines and harm the ecosystem, but only benefit low-performance hardware such as a Cortex-M with 512 KB RAM.

I have tried quite hard to use stackful coroutines in performance sensitive contexts due to the nicer debugging, but I have found them lacking due to a combination of the overhead of moving the stack pointer around (which confuses many optimizers), the issues around segmented stacks or the problems with fixed-sized stacks, and the problems they cause with security features like x86's shadow stack. I am very much a "big hardware" person, and the support for tiny processors is a neat benefit but not really something I care about deeply here. I'm advocating for stackless coroutines because they start to have very large benefits when you have more than a few million tasks in flight at a time, and when you want to be able to retire tasks quickly. 

The main tradeoffs I see are that `alloca` with a non-comptime-known size is problematic for stackless coroutines, and that recursion causes problems unless indirection is added. Guaranteed tail call optimizations help with recursion, since most recursive algorithms should tail call anyway. `alloca` is rare enough and causes enough performance problems that I don't feel horrible about discouraging it.

PR #3951, opened 2025-01-16
[stdlib] Add `stol`

jjvraw:
Introduces `stol` for _more_ flexible string-to-integer parsing (issue #2639).  

Implementation and logic closely follows that of [`atol`](https://docs.modular.com/mojo/stdlib/collections/string/atol/). The function parses integer literals with base handling (2-36), supports base prefixes, and stops at the first invalid character. Unlike `atol`, it returns a tuple (`parsed_int`,  `remaining_string`), allowing partial parsing without raising errors.


jjvraw:
Hi @JoeLoser 

Opened a new PR for this. Too many conflicts to resolve when trying to resolve the DCO issues. 

Question: Since `stol` is a more permissible `atol`, should we avoid raising for an empty string, this seems sensible?

jjvraw:
Hi @JoeLoser. Following the conversation in #4663, is `stol` something you still want to see included in the stdlib?


CC: @martinvuyk

martinvuyk:
I think this is a great addition to the stdlib, IMO we should try to figure out how to integrate it with the "pythonic" API later on. Something like:
```mojo
int_value, str_rest = String("123 something").to_int_and_str()
int_value = Int.__init__[partially_invalid=True]("123 something")
```
But i think this algo is very useful for dirty input

PR #3952, opened 2025-01-16
[stdlib] Use out param for __next__() in more iterators

bgreni:
Apply the change from #3941 to other iterators where it applies.


JoeLoser (approved):

JoeLoser:
!sync

PR #3959, opened 2025-01-18
[stdlib] Use likely in debug_assert

yinonburgansky:
This might allow the compiler to optimize the common path where asserts are not failing.


jackos:
Hi @yinonburgansky thanks for the PR. This can optimize away error checking code, and it appears to be doing so in some of the failed tests. These debug_assert are off for release builds, and so they're already folded away for performance purposes, as such it would be undesirable to potentially optimize error checks away in debug builds. Appreciate the PR though thanks.

yinonburgansky:
> This can optimize away error checking code, and it appears to be doing so in some of the failed tests. 

@jackos This is a compiler bug if it optimizes it away completely. it uses `llvm.expect` which tells the compiler it's most probably going to be `True` so it can arrange the ASM code to make the likely branch faster at the expense of possibly making the unlikely branch slower.

This is different from llvm.assume which tells the compiler to assume it is always True and thus can be optimized away completely.

I think some fails are because it can't run llvm.intrinsic at compile time.

PR #3966, opened 2025-01-24
[stdlib] `SIMD.from_bytes()` and `SIMD.as_bytes()`

msaelices:


docs/changelog.md
```diff
@@ -100,6 +100,10 @@ what we publish.
 
 ### Standard library changes
 
+- New `SIMD.from_bytes()` and `SIMD.as_bytes()` functions to convert a list of bytes
+  to an scalars and vice versa, accepting the endianess as an argument. Similar
```
owenhilyard, line 104: `to an scalars` -> `to scalars` or `to a list of scalars`

stdlib/src/builtin/simd.mojo
```diff
@@ -1956,6 +1959,60 @@ struct SIMD[type: DType, size: Int](
 
         return bitcast[_integral_type_of[type](), size](self).cast[int_dtype]()
 
+    @staticmethod
+    fn from_bytes[
+        big_endian: Bool = False
```
owenhilyard, line 1964: This should probably default to `is_big_endian()`.

stdlib/src/builtin/simd.mojo
```diff
@@ -1956,6 +1959,60 @@ struct SIMD[type: DType, size: Int](
...
+
+        @parameter
+        if is_big_endian() and not big_endian:
+            value = byte_swap(value)
+        elif not is_big_endian() and big_endian:
+            value = byte_swap(value)
+        return value
+
+    fn as_bytes[
+        big_endian: Bool = False
```
owenhilyard, line 1990: This should probably default to is_big_endian().

stdlib/src/builtin/simd.mojo
```diff
@@ -1956,6 +1959,60 @@ struct SIMD[type: DType, size: Int](
 
         return bitcast[_integral_type_of[type](), size](self).cast[int_dtype]()
 
+    @staticmethod
+    fn from_bytes[
+        big_endian: Bool = False
+    ](bytes: InlineArray[Byte, Self.type_len]) -> Scalar[type]:
```
owenhilyard, line 1965: Do we want this to be an `InlineArray`? Those are annoying to construct from a List or other type, thoughts on a `Span` variant?

stdlib/src/builtin/simd.mojo
```diff
@@ -1956,6 +1959,60 @@ struct SIMD[type: DType, size: Int](
...
+        var value = self
+
+        @parameter
+        if is_big_endian() and not big_endian:
+            value = byte_swap(value)
+        elif not is_big_endian() and big_endian:
+            value = byte_swap(value)
+
+        var ptr = UnsafePointer.address_of(value)
+        var array = InlineArray[Byte, Self.type_len](fill=0)
```
owenhilyard, line 2009: It might be better to leave this uninitalized so that LLVM can clearly see it can remove the zeroing.

JoeLoser:
@lsh do you mind TAL at this since you reviewed the original one before it got reverted internally?  Thanks!

stdlib/src/builtin/simd.mojo
lsh: +1, this is should be `sys.info.is_big_endian()`

stdlib/src/builtin/simd.mojo
lsh: That seems like more of an optimization, I'm not going to block on it right now

stdlib/src/builtin/simd.mojo
lsh: That can be a follow up, for now the stricter one is fine with me

stdlib/src/builtin/simd.mojo
```diff
@@ -268,6 +269,8 @@ struct SIMD[type: DType, size: Int](
     alias _Mask = SIMD[DType.bool, size]
 
     alias element_type = type
+    alias type_len = type.sizeof()
```
lsh, line 272: **quibble**: I think this alias makes the methods more confusing rather than more clear.

stdlib/src/builtin/simd.mojo
JoeLoser: +1 to `Span` for the API type here.

docs/changelog.md
msaelices: Good catch! Fixed: https://github.com/msaelices/mojo/commit/6a6ad7678fd2dda050037af8ff7930f6b999a073

stdlib/src/builtin/simd.mojo
msaelices: Good suggestion! Done: https://github.com/msaelices/mojo/commit/5852cb9e1e0a9b5012093ab6364c57127e299767

stdlib/src/builtin/simd.mojo
msaelices: Done: https://github.com/msaelices/mojo/commit/5852cb9e1e0a9b5012093ab6364c57127e299767

stdlib/src/builtin/simd.mojo
msaelices: Done: https://github.com/msaelices/mojo/commit/513af8abce1edea1d131cf0fe79bda4b9b02ebf3

stdlib/src/builtin/simd.mojo
msaelices: How can we approach this? The Python `int.from_bytes()` allows an arbitrary large input because its output is an infinite sized int, like so:
```python
>>> int.from_bytes(b'12321323123212133211213',"big")
4712049637989858926791184397276765665808674789488996659
```
If we allow an arbitrary large input, should the output stop being an scalar but a SIMD with n-value, being n power or two?

owenhilyard: I would only consider the first `Self.type_len` bytes, ignoring everything else, and erroring if not enough bytes are present.

msaelices: Let me try this and open a new PR if I feel useful. I don't like the current hergonomics using InlineArray

PR #3986, opened 2025-02-05
[proposal] Abstracting over `async` (the `yields` effect)

nmsmith:
The proposal document can be found [here](https://github.com/nmsmith/mojo/blob/yields-effect/proposals/yields-effect.md).

I am proposing that we replace `async` with a `yields T` syntax. This allows us to abstract over whether a function is `async`, and more generally, it allows us to abstract over whether traits and even entire libraries are `async`. This solves the function/trait/library duplication problem that arises in languages that have async/await.

This can be seen as a counter-proposal to @owenhilyard's [recent proposal](https://github.com/modular/mojo/pull/3946) on effect handlers. We are both attempting to solve the same problem, and my proposal is heavily inspired by Owen's. The main difference between our proposals is that I have taken the `raises Never` trick that @lattner presented [here](https://github.com/modular/mojo/pull/3946#issuecomment-2601176112), and adapted it to `async`. The end result is really exciting, in my opinion.

I am keen to hear everybody's thoughts. 🙂

**UPDATE July 2025:** I've been reflecting on this design for the last 6 months, and I think we can find a design that's a lot simpler. For details, please read [my post later in this thread.](https://github.com/modular/modular/pull/3986#issuecomment-3102049612)

### DIFF
--- a/proposals/yields-effect.md
+++ b/proposals/yields-effect.md
@@ -1,0 +1,588 @@
+## Abstract
+
+In this document, I propose some changes to how stackless coroutines are
+modelled in Mojo. In particular, I propose that we:
+
+- Replace `async` with a `yields T` syntax, that allows us to abstract over
+  whether a function/trait/library contains suspension points. A function that
+  is not a coroutine (i.e. does not suspend) is equivalent to a function
+  declared as `yields Never`. This feature is extremely powerful. For example,
+  it allows us to write higher-order functions that don't care whether their
+  function argument is a coroutine. Also, it allows us to abstract over whether
+  an I/O library uses non-blocking I/O.
+- Remove the requirement that the `await` keyword be used to wait for a
+  coroutine. This turns out to be a necessary consequence of abstracting over
+  suspension.
+
+In this model, whether a function is a coroutine is relegated to an **effect**
+appearing at the end of a function's signature, very similar to `raises`. Like
+`raises`, the effect is viral by default, so we still have "colored functions".
+Despite this, the design that I present resolves the vast majority of the
+complaints that people have about function coloring. In particular, it prevents
+the need to duplicate functions/traits/libraries into sync versions and async
+versions.
+
+All of this is achieved without performance tradeoffs. In fact, readers
+interested in maximizing the performance of Mojo's coroutines may be interested
+in reading the section towards the end of this document wherein I propose that
+Mojo coroutines be non-allocating. This not a novel idea—Rust coroutines work
+this way—but it may nevertheless be interesting to Modular, since coroutines in
+today's Mojo are heap-allocated by default.
+
+This proposal was inspired by discussions with @owenhilyard and @lattner. Prior
+art is discussed in Appendix B.
+
+## Motivation
+
+Mojo requires a concurrency model that maximizes performance. We need to support
+the concurrent execution of millions of tasks that are waiting on I/O or other
+shared resources, while minimizing memory consumption and compute. The canonical
+use case is a high-throughput server.
+
+To achieve this level of performance, Mojo's concurrency model needs to be built
+upon stackless coroutines. For various reasons, the performance of this model is
+unbeatable. For starters, the memory requirements of a stackless coroutine can
+be calculated at compile time, which allows the coroutine's execution state to
+be stored in a small, fixed-size memory block. Furthermore, Rust has
+demonstrated—in contrast to C++—that coroutines don't need to allocate
+additional memory during their execution. Altogether, this means that stackless
+coroutines can be invoked, suspended, and resumed with very low overhead, and
+memory consumption is minimized. This is exactly what Mojo needs.
+
+Stackless coroutines are normally exposed to programmers through the `async` and
+`await` keywords. Declaring a function to be `async` declares it to be a
+coroutine, and `await` "waits for" a coroutine to finish. Fundamentally, a
+coroutine can only be awaited by another coroutine, unless you're willing to
+block an OS thread while you wait. This is what gives rise to the widely known
+"function coloring" problem, wherein the `async` and `await` keywords are viral
+throughout codebases.
+
+There are several distinct issues that come up in discussions about function
+coloring. The first issue is that it can be annoying to write `async` and
+`await` all over a codebase, especially in the case where you need to add I/O to
+a function that was previously "pure".
+
+It's worth assessing the `async` and `await` keywords separately. For example,
+Kotlin has shown that writing `await` at every call site is
+[unnecessary](https://kotlinlang.org/docs/composing-suspending-functions.html),
+so it's fair to criticize the burden of needing to write `await` every time you
+invoke a coroutine. It's syntactically noisy, it disrupts function chaining, and
+it bloats the size of diffs whenever a function is changed from sync to async,
+or vice versa. The main utility of `await` is to make suspension points stand
+out in the source code. This is analogous to how Swift uses the `try` keyword to
+make exception-driven control flow stand out.
+
+For the most part, the `async` keyword (or Kotlin's `suspend`) subsumes the
+utility of `await`. Like `await`, it allows programmers and compilers to
+identify whether a function call can suspend, by observing its signature.
+Combined with lifetime analysis, this enables reasoning about what data will
+needs to be retained while a coroutine is suspended. In turn, this allows the
+programmer to reason about the memory consumption of suspended coroutines, and
+how to reduce it. The `async` keyword also allows the programmer to check
+whether they are holding any locks across a suspension point, which can
+sometimes be a major performance bottleneck. Finally, if Mojo is to offer
+allocation-free coroutines (like Rust), the compiler needs to prevent a
+coroutine from calling itself recursively, except in cases where the programmer
+has explicitly allocated memory for each recursive call. For the compiler to
+check this, it needs to know whether a recursive function is `async`.
+
+Another major issue that arises in "two-color" languages is the inability to
+abstract over whether a function is `async`. The simplest examples are the
+standard `map` and `fold` functions from functional programming. `map` is a
+higher-order function that takes a data structure and a function pointer, and
+applies it to every element of the data structure. If `map` were a method on
+Mojo's `List` type, its signature would be as follows:
+`fn map[f: fn (Int) -> Int](self) -> Self:`
+
+The problem is that we haven't allowed `f` to be a coroutine. One way to resolve
+this would be to add a second overload:
+`async fn map[f: async fn (Int) -> Int](self) -> Self:`
+
+Unfortunately, this requires the implementation of `map` to be duplicated, which
+would be a maintenance nightmare. Ideally, we want the ability to define `map`
+in such a way that it becomes a coroutine whenever it is passed a coroutine as a
+parameter. A similar problem occurs with traits: many languages that have
+traits/protocols offer sync and async versions of key traits. For example, Rust
+offers both `Iterator` and `AsyncIterator`. And at a larger scale: many
+languages duplicate **entire libraries**, to offer a sync version and async
+version. Altogether, it's clear that Mojo desperately needs a way to abstract
+over the "asyncness" of functions, traits, and libraries. **The purpose of this
+document is to present a means of abstracting over "asyncness" using Mojo's
+parameter system.**
+
+In summary:
+
+- To maximize the performance of Mojo's concurrency model, coroutines must be
+  explicitly declared as `async`, or something equivalent to that.
+- To avoid a situation where functions/traits/libraries are split into sync and
+  async versions, Mojo needs a means of abstracting over "asyncness".
+
+## My proposal
+
+### Introduce the syntax `fn foo() yields T`
+
+I propose we introduce a keyword `yields`, which works very similarly to
+`raises`. A function that suspends must be annotated as `yields`, as in "yields
+control". For example:
+
+`fn read_file(path: String) yields -> String:`
+
+This syntax would replace the use of `async fn` for declaring Mojo-native
+coroutines, but we can keep `async fn` around for Python integration. The
+primary reason for introducing a new syntax is that we want `yields` to be able
+to take a type:
+
+`fn read_file(path: String) yields Blah -> String:`
+
+This syntax has several uses. Firstly, it allows us to declare the types of the
+values yielded by **generators**:
+
+`fn prime_numbers() yields Int64:`
+
+Generators and coroutines can make use of the same compiler transform, where a
+seemingly sequential function definition is transformed into a state machine
+that `yields` periodically. That's why we're discussing generators here.
+
+In Python, the type that a generator yields can be declared by writing the
+generator-function's return type as `-> Generator[int, None, None]`. However,
+this also exposes the concrete type of the generator, and forces the generator's
+state to be heap-allocated. For performance reasons, this is not ideal for Mojo.
+In comparison, the syntax `yields Int64` reveals nothing about the generator's
+implementation—it only describes an aspect of its interface.
+
+The focus of this proposal is the use of coroutines for asynchronous I/O, so I
+will say no more about generators. For the I/O use case, the purpose of allowing
+`yields` to take a type is that we can interpret `yields Never`—where `Never` is
+an uninstantiatable type—as meaning that **a function is NOT a coroutine.** For
+example, the following signature describes a function that reads a file without
+yielding control of the OS thread:
+
+`fn read_file(path: String) yields Never -> String:`
+
+This is equivalent to omitting `yields` entirely:
+
+`fn read_file(path: String) -> String:`
+
+With this syntax, we have the ability to write functions that are **generic**
+over whether an argument is a coroutine:
+
+`fn higher_order[T: AnyType](f: fn() yields T) yields T:`
+
+This is a really neat trick. Chris
+[proposed](https://github.com/modular/mojo/pull/3946#issuecomment-2601176112)
+doing this for `raises`; I've merely adapted the trick to `yields`. We can even
+write functions that are generic over `yields` and `raises` simultaneously:
+
+```plaintext
+fn higher_order[
+    T1: AnyType, T2: AnyType
+](f: fn() yields T1 raises T2) yields T1 raises T2:
+```
+
+### Use the `if`-operator to _programmatically_ configure whether a function is a coroutine
+
+The coolest thing about this design is that it allows us to use Mojo's parameter
+system to **programmatically control** whether a function is a coroutine:
+
+`fn read_file(path: String) yields Blah if config['async'] else Never -> String:`
+
+Here, I'm using Mojo's ternary `if`-operator. `config['async']` is a
+compile-time Boolean. For the moment, let's not worry about where this value
+comes from, or what its name is. Owen
+[proposed](https://github.com/modular/mojo/pull/3946) passing the value around
+as an implicit parameter, but maybe it could be set using compiler flags, or
+maybe Mojo could allow modules to be provided with arbitrary parameters when
+they are imported. The choice that we make here is orthogonal to the rest of the
+design that I'm presenting, so we can shelve this concern for the moment.
+
+The syntax `yields Blah if config['async'] else Never` is verbose, so let's
+introduce some sugar:
+
+1. The first argument of the conditional (`Blah`) can be omitted, in which case
+   it defaults to whatever the default value for `yields` is. (Perhaps `None`.)
+2. The third argument of the conditional (`Never`) can be omitted, in which case
+   it defaults to `Never`.
+
+If we drop both of these arguments from the signature of `read_file`, it
+becomes:
+
+`fn read_file(path: String) yields if config['async'] -> String:`
+
+This is highly readable, in my opinion! This signature communicates very plainly
+that `read_file` will yield (be a coroutine) only if `config['async']` has been
+set. This is peak simplicity: I can't imagine "programmable async" being any
+simpler than this.
+
+Using Mojo's "associated aliases" feature, we can even define traits in such a
+way that their implementers can choose whether the trait methods are coroutines
+or not:
+
+```plaintext
+trait Iterator[T: AnyType]:
+    alias yields_: Bool
+    fn __next__(mut self) raises StopIteration, yields if yields_ -> T: ...
+
+fn get_integer[IterType: Iterator[Int]](mut iterator: IterType) yields if IterType.yields_ -> Int:
+    try:
+        return next(iterator)     # Later on, I'll explain why we're not using 'await' here.
+    except:
+        return 0
+```
+
+In Python, Rust, Swift, and all other languages with `async`/`await` syntax, the
+standard library traits—including `Iterator`—are duplicated into "sync" and
+"async" versions. More broadly, many different APIs have undergone "mitosis",
+ultimately splitting the languages' libraries into sync and async versions. This
+issue has generated a lot of negative sentiment in languages with
+`async`/`await`. As shown above, the `yields if <cond>` syntax allows us to
+avoid this problem. **This would be an enormous win for Mojo.**
+
+As a final note: the `yields if <cond>` syntax allows Mojo's I/O libraries to be
+generic over whether the I/O is blocking or non-blocking. There are a few
+reasons why somebody may want blocking I/O:
+
+1. For "gradual disclosure of complexity", when teaching people programming.
+2. To avoid the nuisance of declaring the `yields` effect when writing simple
+   scripts, such as a script that reads a file, does some calculations, and then
+   writes to a file.
+3. To achieve better performance when writing a program where the number of
+   concurrent tasks is always less than the number of cores. (Such programs
+   don't benefit from yielding control of a core.)
+4. To define an I/O library that is usable from other languages—such as
+   C—without those languages needing to know about coroutines or executors.
+
+In summary: Being able to configure whether a function is a coroutine or not is
+**very** useful, and can be achieved by combining the `yields T` effect with the
+ternary `if`-operator.
+
+### Let the `if`-operator be used for `raises`
+
+We should allow the `if`-operator to be used for `raises`, in addition to
+`yields`. This would have a lot of utility. For example, it would provide a
+mechanism for configuring whether functions that allocate memory should raise an
+exception on allocation failure:
+`fn append(mut self: List[T], owned value: T) raises if config['allocation exceptions']:`
+
+Another use case for `raises if <cond>` is to configure whether a long-running
+task should respond to a cancellation request from its executor:
+`fn long_running_task() yields, raises Cancelled if config['cancellable']:`
+
+This feature is **_desperately_** needed in languages with `async`/`await`. As a
+concrete example, in the Swift community there have been
+[debates](https://forums.swift.org/t/asyncsequences-and-cooperative-task-cancellation/62657)
+about whether "asynchronous sequences" should respond to cancellation requests,
+and if so, whether Swift needs a parallel universe of "asynchronous sequences
+that don't cancel themselves". All of this complexity can be avoided by allowing
+the end user to configure whether they want a function to raise cancellation
+exceptions. This is made trivial by the `raises if <cond>` syntax.
+
+### Drop the requirement that coroutines be `await`ed
+
+Let's think some more about `read_file`:
+
+`fn read_file(path: String) yields if config['async'] -> String:`
+
+As mentioned earlier, the `yields if <cond>` syntax allows people to use this
+function both as a coroutine (with non-blocking I/O), and as a blocking
+function, depending on their needs. Let's suppose that `read_file` is used by
+another function `foo` which also wants to abstract over its coroutine status:
+
+```plaintext
+fn foo(path: String) yields if config['async']:
+    data = await read_file(path)
+    print(data)
+```
+
+Here, we're awaiting the result of `read_file`, and then printing it. In Mojo,
+the expression `await read_file(path)` desugars to
+`read_file(path).__await__()`. In other words, `await` does not have any magical
+properties—it's just a function call.
+
+Unfortunately, this design is incompatible with the semantics of
+`yields if config['async']`. When `config['async']` is false, the definitions of
+`read_file` and `foo` reduce to:
+
+```plaintext
+fn read_file(path: String) -> String:
+    ...
+fn foo(path: String):
+    data = await read_file(path)
+    print(data)
+```
+
+`read_file` becomes a function that returns a `String`. In turn, this means that
+`await read_file(path)` is awaiting a string. This is nonsense! A string can't
+be awaited—it doesn't have an `__await__` method. And in fact, if
+`String.__await__` was defined, the situation would be even worse: the
+synchronous and asynchronous versions of `foo` would be awaiting different
+things!
+
+None of this is acceptable. To make the `yield if <cond>` feature behave
+properly, I can only see two options:
+
+1. Introduce some compiler magic to **pretend you didn't write `await`** when
+   `<cond>` is set to false.
+2. Change the semantics of coroutines, so that `read_file(path)` awaits the
+   coroutine, rather than returning an instance of `Coroutine`. In other words,
+   we eliminate the need to write `await`.
+
+If we chose the second option, our program would be written as follows:
+
+```plaintext
+fn read_file(path: String) yields if config['async'] -> String:
+    ...
+fn foo(path: String) yields if config['async']:
+    var data: String = read_file(path)           # no 'await'
+    print(data)
+```
+
+**Note:** We would still use the `await` keyword for other purposes. I'm not
+saying we should remove the keyword from Mojo.
+
+In my opinion, this model is the right choice, for many different reasons:
+
+1. This model ensures that the `await` keyword isn't treated specially by the
+   compiler—it's just sugar for `__await__()`.
+2. This model makes function chaining more ergonomic: we can write
+   `read_file(path).split('\n')` instead of
+   `(await read_file(path)).split(' \n')`.
+3. This model makes **refactoring** a program more ergonomic: when you add the
+   `yields` effect to a function, you no longer need to go and add the `await`
+   keyword to every call site. This reduces the syntactic burden of function
+   coloring. (That said, you might need to add the `yields` effect to the
+   function's callers, if they don't already have it.)
+4. This model eliminates the need for the following Python features:
+   `async for`, `async with`, `__aiter__`, `__anext__`, `__aenter__`, and
+   `__aexit__`. These features **ONLY** exist so that the compiler can
+   [insert some calls to 'await'](https://snarky.ca/unravelling-the-async-with-statement/)
+   into the desugaring of `for` and `with` statements. We'll still need these
+   features for backwards compatibility with Python coroutines, but we wouldn't
+   need to teach all of this stuff to people writing pure Mojo. In the long
+   term, this would make Mojo a simpler, more approachable language.
+5. If we were to stick with `await read_file(...)`, programmers would need to do
+   mental gymnastics to determine the expression's meaning in the case where
+   `read_file` has been specialized to a synchronous function. Remember: after
+   specialization, the signature of `read_file` becomes
+   `fn read_file(path: String) -> String`, so the meaning of
+   `await read_file(...)` becomes truly baffling, especially if the IDE
+   _displays_ the type of `read_file` as being `fn(String) -> String` when you
+   hover over it!
+6. In contrast, if we drop the requirement to use `await`, a coroutine can be
+   explained as simply "a function that has a `yields` effect". It would have
+   all of the affordances of an ordinary function, including the property that
+   `f()` evaluates the function. This is a simpler model for learners—especially
+   people who are being taught about asynchronous I/O for the first time. We can
+   explain that functions that yield propagate a `yields` effect, analogous to
+   how functions that raise propagate the `raises` effect. **With this model, we
+   don't need to expose learners to a `Coroutine` object**, and we don't need to
+   explain why `read_file(path)` doesn't _actually_ read the file. Instead,
+   function calls continue to work the same, regardless of whether a function
+   contains `yields` in its signature. In short: this model is great for
+   learnability.
+
+Coincidentally, this model is similar to
+[Kotlin's](https://kotlinlang.org/docs/composing-suspending-functions.html).
+Kotlin has the syntax `suspend fun`, which is the equivalent of Mojo's
+`async fn`, and yet it doesn't require `await` to be used at call sites.
+
+Some people might feel uncomfortable with removing `await`, on the thesis that
+it helps document where all of the suspension points are in a function body.
+Personally, I don't think this is a big deal, for the following reasons:
+
+1. In IDEs, we can highlight the identifiers of coroutines differently, e.g.
+   give them a distinct color. This would make it easy to tell which function
+   calls can suspend, and which can't.
+2. The `yields` effect can be understood as being similar to `raises`, in the
+   sense that the effect is visible in function signatures, but not at the call
+   site. In practice, the invisibility of exceptions in Mojo isn't a big deal.
+   And in Python, programmers make do with far less information, because there
+   isn't an explicit `raises` effect. Finally, Swift tried making exceptions
+   explicit at call sites via the `try` keyword ("marked propagation"), and
+   people have mixed feelings about that. In summary: if Mojo doesn't need
+   "marked exception points", then I doubt it needs "marked suspension points".
+   (Especially if we have the aforementioned IDE highlighting.)
+3. In languages where programs are mostly single-threaded (such as JavaScript
+   and Python), `await` points are used to reason about the moments at which
+   "concurrency happens". This doesn't apply to Mojo. Mojo programs are
+   typically (highly) multi-threaded, so concurrency happens _everywhere_. Said
+   another way: in Mojo, `await` points are not synchronization points. They are
+   just points at which the current thread—one of many—might switch to another
+   task.
+4. In languages with green threads, such as Go and Java, suspension points are
+   not explicit. I haven't seen Go or Java users complain about this. To the
+   contrary, they normally talk about the lack of function coloring (i.e.
+   explicit suspension) as the primary _advantage_ of green threads.
+
+Given all of the above, it doesn't seem like we'd be losing much by having
+suspension points be a transparent effect, like exceptions. A dedicated IDE
+color for functions that `yield` would likely be sufficient. That's already a
+step beyond what Go offers!
+
+In summary: To be generic over whether a function `yields`, it seems that we
+need to eliminate `await` at call sites. This may take some getting used to, but
+beyond that, it's not a big deal.
+
+### Coroutines should expose an associated `Task` type
+
+As previously mentioned, the plan is for `read_file(path)` to denote a `String`,
+rather than a `Coroutine`/`Task`. But we still need the latter type (let's call
+it `Task`) for modelling the execution state of a coroutine. An instance of
+`Task` models a unit of work that can be carried out by an executor, potentially
+in parallel with other work. Given that `read_file(path)` returns a `String`,
+we'll need to find a different syntax for constructing a `Task`.
+
+If Mojo is to offer non-allocating coroutines—as Rust does—we will actually need
+a distinct `Task` type for each distinct coroutine. Therefore, it makes sense
+for `Task` to be an **associated type** of a coroutine. For example, the task
+type associated with the `read_file` coroutine would be `read_file.Task`. Once
+we have this, the evaluation of `read_file(path)` can be packaged up into a
+`Task` by passing the arguments to the `Task` constructor, instead of passing
+them to the coroutine: `var task: read_file.Task = read_file.Task(path)`
+
+`read_file.Task` would be value-semantic and its state would be stored inline.
+This means the state of `task` would be stored in the enclosing function's stack
+frame. This avoids the need to allocate memory. Several such tasks can be
+constructed, and then `await`ed simultaneously. The executor can then run the
+tasks concurrently, while the parent task is paused. When they are finished, the
+parent task can be resumed. No memory needs to be allocated at any point in this
+process, except when constructing the `String` that `read_file` returns.
+Alternatively, tasks can be "detached" from their parent task by submitting them
+directly to an executor, e.g. `executor.submit(task^)`. This would return a
+`Future` that can be `await`ed.
+
+**Note:** In Python, a `Task` object is reference-semantic, and points to a task
+that has already been submitted for execution. In the above design, that's
+equivalent to implicitly passing every task to `executor.submit` as soon as it
+is created. This is not the right default for Mojo, for two reasons:
+
+1. `executor.submit` needs to allocate memory to store the task. This has
+   runtime overhead.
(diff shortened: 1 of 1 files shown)


martinvuyk:
> I have proposed that we replace async with a yields T syntax, that allows us to abstract over whether a function/trait/library contains suspension points. A function that is not a coroutine (i.e. does not suspend) is equivalent to a function declared as yields Never.

@nmsmith I love this. Making it a function effect and being able to parametrize it would be awesome.

> ```mojo
> trait Iterator[T: AnyType]:
>    alias yields_: Bool
>    fn __next__(mut self) raises StopIteration, yields if yields_ -> T: ...
> ```

I'm personally not that fond of raising iterators or this conditional yield logic. I also don't think inserting comptime logic inside type signatures is a great idea.

I think I'd rather see something along the lines of:
```mojo
trait Iterator[T: AnyType, generator: Awaitable]:
  fn __next__(mut self) yields generator -> T: ...
  # is the same as
  fn __next__(mut self) -> generator[T]: ...
  fn __has_next__(self) -> Bool: ...
```
We could try to make coroutine have a "false async mode". And then we try to add some kind of compiler decorator which sets the path to materialization
```mojo
@nonmaterializable
struct Coroutine[T: AnyType, sync: Bool, auto_deref: Bool = True]:
    fn __init__(out self: Coroutine[_, sync=False], handle: AnyCoroutine): ...
    fn __init__(out self: Coroutine[_, sync=True], handle: AnyFunction): ...

  @materialization_entrypoint(auto_deref=auto_deref)
  fn __await__(owned self) -> T: ...
```
What this would allow is an "auto-deref" but for the await keyword
```mojo
fn read_file[sync: Bool]() -> _FileGenerator[sync]:  ...
# or the alternative syntax
fn read_file[sync: Bool]() yields _FileGenerator[sync] -> List[Byte]:  ...

fn foo(path: String):
    var data: List[Byte]
    data = read_file[sync=False](path)
    data = read_file[sync=True](path)
```

IMO we should try to achieve something in this direction (even if the details of what I'm proposing here don't add up), where we don't end up having arbitrarily complex code executing inside type signatures.

As for @owenhilyard 's #3946 and also trying the same solution for `raises`. I don't know if my mini-proposal of using `nonmaterializable` and having a `NoError` error type as a workaround is possible. I'll leave it as food for thought.

nmsmith:
Thanks for the feedback @martinvuyk.

> I'm personally not that fond of raising iterators

Acknowledged. In my proposal, I just replicated how iterators work in Python. It was not my intention to propose that Mojo iterators raise exceptions. This is a discussion worth having, but it's outside the scope of this proposal.

> we should try to achieve something [...], where we don't end up having arbitrarily complex code executing inside type signatures.

Mojo's metaprogramming model has always offered the ability to execute arbitrarily complex code at compile-time. This is a feature, not a bug. Also, I've only made use of a simple ternary-`if` operation, so I'm not sure what you're concerned about here.

> What this would allow is an "auto-deref" but for the await keyword.

I'm not a fan of this idea. It sounds very complicated/magical.

owenhilyard:
I think that this causes some awkward syntax, since we want the ability to have a coroutine who's component functions return `Poll[T]`, where `Poll` is a sum type. There isn't a great way to specify "this will yield one sum type variant when you should keep polling, and another when it is finished". It leads to something like `fn foo() yields Poll[T] -> Poll[T]`. While I'm fine with that as a desugaring for `async fn foo() -> T`, I think that people will get very annoyed with that very quickly if they have to write it by hand. I think that, for functions which act as generators, it's a reasonable syntax since you do need the ability to specify return and yield types separately. For infinite generators, such as your prime numbers example, I would do `fn prime_numbers() yields Int64 -> Never`, so the compiler is informed that it is expected to never return. Creating this may necessitate something like Rust's `loop` keyword, signifying an infinite loop, with the yields inside of it.

I like the idea of using ternaries to control the values yielded, returned and raised, but I think I want to try to take it further if the compiler team is willing. Ideally, this shouldn't be special cased and there should be some way to get a token from the compiler which represents a type (which can later be used in reflection, but for now may as well be opaque), and make any expression which returns that mlir type legal in those positions, so you can have a function which figures out what to do instead of having massive nested ternaries.  

I don't know if there is a "sane default" for yields except for `None` or `Never`, meaning it doesn't happen. 

I strongly prefer "awaiting something without __await__ does nothing", since I think it's easier to explain and makes the code more unified. The option of making awaits implicit introduces a lot of headaches for things like timeouts, where you want a future that will be done when time is up and another to do actual IO. 

As far as function chaining, I agree that prefix await is a bit messy, and would prefer postfix await. 

> In IDEs, we can highlight the identifiers of coroutines differently

But github, discord, discourse and other places with regex-based highlighting cannot do that. 

> In languages with green threads, such as Go and Java, suspension points are not explicit. I haven't seen Go or Java users complain about this. To the contrary, they normally talk about the lack of function coloring (i.e. explicit suspension) as the primary advantage of green threads.

The reason that Go and Java can do this is because they have a runtime which can carry a lot of weight for them. Go uses what I call "the paint bucket approach to function coloring", and makes everything async. The compiler is technically free to suspend at any function call, even if you are holding locks that should be released quickly. There are substantial costs which come during performance analysis, and they result in users who want to write fast code having to do deep audits of the codebase and learn arcane rules for when the compiler suspends functions. You will also note that `go` doesn't have a way to get values out of a greenthread except via a channel. This causes a lot of headaches when you want to send out multiple requests and then gather results from all of them. 

> But we still need the latter type (let's call it Task) for modelling the execution state of a coroutine.

This is the coroutine frame. 

> Alternatively, tasks can be "detached" from their parent task by submitting them directly to an executor, e.g. executor.submit(task^). This would return a Future that can be awaited.

I disagree here, I would have normal coroutines used in async return an (excuse my Rust syntax) `impl Future<Output = T>`, which is effectively an alias for `impl Coroutine<Yield = Poll<T>, Return=Poll<T>>`. When you `spawn` a task, that is detaching it from you, such that it survives you being cancelled. Unless the return value is stored inline in the coroutine frame, I'm not sure we could support futures being returned from that, instead I would rather have a `JoinHandle` type, which explictly has that extra allocation to hold things in.

martinvuyk:
> > we should try to achieve something [...], where we don't end up having arbitrarily complex code executing inside type signatures.
> 
> Mojo's metaprogramming model has always offered the ability to execute arbitrarily complex code at compile-time. This is a feature, not a bug. Also, I've only made use of a simple ternary-`if` operation, so I'm not sure what you're concerned about here.

In a way yes, but it will make function signatures get even more bloated (any function can be placed inside the ternary if else condition). Also AFAIK (I'm no compiler person) it's one thing to add code to check the default values for a function argument or parameter, and another for the type itself to be computed (AFAIK that is the purpose of Traits). If we were to allow inserting dynamic code in places where there are normally traits (or function effects), I think it's easier to read if we let a type have a function that returns the end result according to the parameters. 

e.g. adapting from #3756
```mojo
@value
struct Generator_1or2[parameter1: Bool, parameter2: Int]:
    @staticmethod
    fn __type__() -> AnyType:
        return Generator1 if Self.parameter1 else Generator2

    @staticmethod
    fn __trait__() -> AnyTrait:
        return SomeIntParamTrait[Self.parameter2 if Self.parameter1 else  0]
```
And each function gets added by code-gen (for types that have them) when on one such part of a type signature (parameter, type, or function effect space).

A crazy idea: extend this to include `__yields__`, `__raises__`, `__captures__`, and we have ourselves some quite readable advanced meta-programming.


> > What this would allow is an "auto-deref" but for the await keyword.
> 
> I'm not a fan of this idea. It sounds very complicated/magical.

We already use materialization for types like `IntLiteral` (and implicit constructors) when being passed to a function that expects an `Int`, this is just a reversed "implicit destructor" where the type gets consumed and controls to what it materializes to. The parameter "auto_deref" could also be called "nonmaterializable_at_comptime" or basically setting up an implicit barrier that anytime the type is instanciated it is consumed right after exiting from the function where it was created. It would be a "ghost return type".

The main goal of that workaround is to not have to make conditionals on when each effect gets applied and simply say that there is a "NoneEffect", so that the functions actually always return the same types parametrized for different cases.

nmsmith:
Thanks for the feedback Owen! I'll reply to all of your main points.

> I think that this causes some awkward syntax, since we want the ability to have a coroutine who's component functions return `Poll[T]`, where `Poll` is a sum type. There isn't a great way to specify "this will yield one sum type variant when you should keep polling, and another when it is finished". It leads to something like `fn foo() yields Poll[T] -> Poll[T]`.

I don't see why `Poll` needs to be explicit in the function signature. (For other readers: [Poll is a Rust enum](https://doc.rust-lang.org/std/task/enum.Poll.html).) Let's consider the signature `fn foo() yields A -> B`. If you want the executor to receive a sum type of either `A` or `B`, we can surely make this happen without the sum type needing to show up in the signature. Isn't this phenomenon already _implied_ by the fact that a coroutine may either yield or return, depending on the control flow? To me, it makes more sense for the executor to receive a `Variant[A, B]`.

You seem to have made the assumption that the type appearing after `yields` must match the type that shows up in the signature of Rust's [Future.poll](https://doc.rust-lang.org/std/future/trait.Future.html) method. I don't think we need to impose that constraint.

>  Ideally, [the ternary operator] shouldn't be special cased and there should be some way to get a token from the compiler which represents a type, and make any expression which returns that mlir type legal [after `yields` and `raises`].

Yes, it makes sense for Mojo to eventually have this kind of capability. That's outside the scope of this proposal though. 🙂

> making awaits implicit introduces a lot of headaches for things like timeouts, where you want a future that will be done when time is up and another to do actual IO.

I don't understand the issue here. Setting a timeout for an I/O operation could be as concise as:
`var data: String = await read_file.Task(path).timeout(1000)`

...which would raise an exception if the operation times out.

Alternatively, if you want to do that trick where you "race" two tasks—one of which is a timeout—that might look like:
`var data: Variant[String, None] = await race(read_file.Task(path), sleep.Task(1000))`

The only difference from the conventional async/await model is that if you don't want to execute the coroutine immediately, you need to add `.Task` to construct an explicit task.

> But github, discord, discourse and other places with regex-based highlighting cannot do that.

Yes, that's true. That's a legitimate downside of relegating suspension to a function effect. But given that the `raises` effect is also implicit in these places—and `raises` changes control flow—I'm not sure that this is going to be a big deal. In situations where a programmer needs to study a program's behavior carefully, they need a proper IDE regardless. Trying to make reading code on Github a pleasant experience is already a lost cause.

> I would have normal coroutines used in async return an (excuse my Rust syntax) `impl Future<Output = T>`. [...] When you spawn a task, that is detaching it from you, such that it survives you being cancelled. Unless the return value is stored inline in the coroutine frame, I'm not sure we could support futures being returned from that, instead I would rather have a JoinHandle type, which explicitly has that extra allocation to hold things in.

All of this boils down to a disagreement on what a "Future" is. You're using the Rust definition, whereas I was using the C++ definition. In C++ a future is a pointer. If you use this mental model, then I think that section of my proposal makes more sense.

Also, I don't think a `Task` should be automatically detached from the task that spawned it. That prevents us from modelling **subtasks**, i.e. tasks whose lifetime is constrained by a parent task—and which have access to the parent task's state. It's very important that we have something like this. It's what "structured concurrency" is based on.

melodyogonna:
I've gone back and forth with @owenhilyard about the merit of this functionality on Discord, I've been convinced it is massively useful - if we're successful in making the Mojo community adopt async genericity from the get-go.

I, however, have issues with the syntax and semantics proposed here. 
1. I disagree with both removing the requirement to await coroutines and with having the compiler await them.
2. The syntax, while being useful for something the compiler can lower to, is not something people should write every time.
3. Environment, runtime-specific, compatibility concerns are brushed over.

I have a suggestion/proposal/improvement based on this that I'd like to submit for evaluation.

What if we introduce the concept of "partial async" or something similar, which describes a function signature that can be satisfied in either an async or non-async form?

## Details
I suggest we add the ability to denote the asynchronicity of a function is optional - `async?`
Which we can use as following:
```python
trait SomeTrait:
    async? fn method_name() -> Int: ...
```

This trait can be satisfied by async and non-async functions. Functions can also be defined to be partially async:
```python
async? fn partially_async() -> Int: ...
```

The idea here is that functions which are explicitly declared to be partially async can be assumed to be safe to call in both sync and non-sync manner. However, functions that are explicitly fully async - `async fn fully_async(): ...` are designed to be used only in async environments. Likewise, functions without any effect are not async in any manner. The functions that are fully async or not async at all can be lowered into the syntax proposed here or the one Owen proposed.

### Coroutines and Tasks
When a function is partially async, the compiler can decide whether it is a coroutine or not based on how it is called. 

Tasks depend on the execution environment, there needs to be a guarantee that they're being created with a coroutine. I propose we cast a partially async function into fully async based on a model similar to type narrowing. Given a partially async function, we can verify it is async:
```python
@paremeter
if is_async(partial_async_function):
    x = partial_async_function() # creates a coroutine
    task = loop.create_task(x)
    value = await task
else:
    value = partial_async_function()
```  
see - https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards

### Summary
1. 'await' can be used in async and partially async functions.
2. 'await' can be used for async and partially async functions.
3. Actions that require explicit coroutines can not happen with partially async functions unless it has been guaranteed to the compiler that it returns a coroutine.
4. 'await' can not be used in non-async functions.

I'm hoping we can have the sort of uniform generality we have with refs now.

owenhilyard:
> Tasks depend on the execution environment, there needs to be a guarantee that they're being created with a coroutine. I propose we cast a partially async function into fully async based on a model similar to type narrowing. Given a partially async function, we can verify it is async:
> 
> ```python
> @paremeter
> if is_async(partial_async_function):
>     x = partial_async_function() # creates a coroutine
>     task = loop.create_task(x)
>     value = await task
> else:
>     value = partial_async_function()
> ```

I have a version of this in my effect handlers proposal, but you put it inside of the function.

```
async? fn partial_async_function(...):
    # common setup

    if async:
        ...
    else:
        ...
```

nmsmith:
Hi @melodyogonna, thanks for your suggestion. I'll share my feedback below.

In the [proposal document](https://github.com/nmsmith/mojo/blob/yields-effect/proposals/yields-effect.md), I walk through a ton of different use cases for the `yields T` feature. From what I can tell, your `async?` feature doesn't fulfill all of those use cases. If you think it does, I would find it helpful to see the use cases addressed systematically. In the meantime, I am skeptical that your counter-proposal would be sufficient.

My second concern is that Mojo is going to have `raises T` at some point, and I've demonstrated the utility of `raises if <cond>`. If Mojo were to already offer that feature, then the cost of adding `yields if <cond>` would be very small (in terms of compiler + conceptual complexity). Certainly, the cost would be smaller than an `async?` feature that has a novel syntax and semantics. And `raises T` is undoubtedly more appropriate than `raises?`.

nmsmith:
I do find at least one aspect of your design intriguing, @melodyogonna. Your `is_async(f)` syntax inspired me to investigate what would happen if we make the types that a function `yields` and `raises` **associated types** of the function's type. With such a feature, we can dramatically simplify the following signature from my proposal:

`fn higher_order[T1: AnyType, T2: AnyType](f: fn () yields T1 raises T2) yields T1 raises T2:`

Here's the idea:
- When using `fn` to declare a function type, the function is assumed to be non-yielding and non-raising. (As in today's Mojo.)
- When using `any_fn` to declare a function type, the type of value that a function yields and raises becomes an **associated type** of the function type. (Please ignore the keyword's name. If we agree that want this feature, we can bikeshed it later.)

This allows us to shorten the signature of `higher_order` significantly:
`fn higher_order(f: any_fn ()) yields f.YieldType, raises f.ExceptionType:`

This also resolves a syntactic issue concerning function types. For the sake of readability, ideally effect declarations would be comma-separated, e.g. `fn foo() yields X, raises Y, some_other_effect`. But we can't include those commas in function types, because that would mess up the parsing of argument lists. By turning effects into associated types of a function type, e.g. `f.ExceptionType`, this problem is neatly resolved.

martinvuyk:
> This allows us to shorten the signature of `higher_order` significantly:
> `fn higher_order(f: any_fn ()) yields f.YieldType, raises f.ExceptionType:`

+1 on this. Also, if we extend it for docstrings overloading it would be great
```mojo
fn timeit(f: any_fn ()) yields f.YieldType, raises f.ExceptionType:
  __doc__.summary = "Time any function."
  __doc__.args = "f: Any function."
  __doc__.raises = f.__doc__.raises
  __doc__.yields = f.__doc__.yields
```

ivellapillil:
FWIW, Nim has effect system which seems to work pretty well. They use braces as syntax which we can also adopt if needed, and AFICS, is extensible to custom effects as and when needed.

https://nim-lang.org/docs/manual.html#effect-system

melodyogonna:
> Hi @melodyogonna, thanks for your suggestion. I'll share my feedback below.
> 
> In the [proposal document](https://github.com/nmsmith/mojo/blob/yields-effect/proposals/yields-effect.md), I walk through a ton of different use cases for the `yields T` feature. From what I can tell, your `async?` feature doesn't fulfill all of those use cases. If you think it does, I would find it helpful to see the use cases addressed systematically. In the meantime, I am skeptical that your counter-proposal would be sufficient.

While reasoning about my proposal yesterday I noticed a soundness problem. Specifically - calling a partially async function in sync context. If we assume that an API that accepts a partial async function dynamically (e.g trait) can be passed both a synchronous and an asynchronous function, and that coroutines must remain awaitable, then if the partial async function is called it'll just return a coroutine that does nothing. After thinking through different scenarios, it seems your idea to remove the necessity of making coroutines awaitable is the simplest solution. However, the functionality of lazy coroutines is very helpful, it is something we want. So perhaps we can have that option but in fully async environment. To that end, perhaps instead of having only eager coroutines or only lazy coroutines, we can support both?

martinvuyk:
> However, the functionality of lazy coroutines is very helpful, it is something we want. So perhaps we can have that option but in fully async environment. To that end, perhaps instead of having only eager coroutines or only lazy coroutines, we can support both?

Quoting myself:

> We could try to make coroutine have a "false async mode". And then we try to add some kind of compiler decorator which sets the path to materialization
> 
> ```mojo
> @nonmaterializable
> struct Coroutine[T: AnyType, sync: Bool, auto_deref: Bool = True]:
>     fn __init__(out self: Coroutine[_, sync=False], handle: AnyCoroutine): ...
>     fn __init__(out self: Coroutine[_, sync=True], handle: AnyFunction): ...
> 
>   @materialization_entrypoint(auto_deref=auto_deref)
>   fn __await__(owned self) -> T: ...
> ```

if the auto_deref parameter is there, it would mean we could control when coroutines are auto-awaited and when not.

This doesn't need to look exactly like what I proposed or even use materialization, but it would be necessary to sometimes switch no manual await

melodyogonna:
@nmsmith
Another thing. I'm trying to make my model work with yours, however, it's being done with so many different assumptions that it's probably better as separate proposal.

Here are my thinking models.
1. Let's not mix generators and coroutines, let's make coroutines its own type. Also, coroutines can not "yield" a different type compared to what the function returns.
if we try to translate this to your proposal.
```python
async? def function() -> Int: ...
```
would look something like:
```python
def function() yields -> Int: ...
```
which could be further be assumed to mean:
```python
def function() -> Int | Coroutine[1]: ...
```
2. Let's maintain the rule in Python that prevents generators from being used in a coroutine environment. And extend it so generators can't be used in a partially async function.
3. Whether the environment is async is determined by the function declaration rather than some compile-time parameter. so
```python
fn read_file(path: String) yields if config['async'] -> String:
```
becomes
```python
async? read_file(path: String) -> String:
```
when a function is fully async then the config is.
```python
async read_file(path: String) -> String:
```
is desugared to something akin to... I don't really know how to represent it with your model. But if we go with Owen's model
```python
async=async fn read[async: Bool = True](): ...
```
if partial async function is called, the environment it is executing in should be propagated by the function that called it.
4. We can then tie this down with your eager coroutine idea.  In a partial async function, coroutines are eager and always desugar to `coroutine.__await__` even if called without `await`. in a fully non-async environment, `await` can not be used, so coroutines always desugar to common `coroutine.__await__` like in your model. 
However, in a fully async environment, coroutine can be lazy.


Comparing this to current reference semantics.

`async? def` = `ref[_]` - the environment will be inferred
`def` = `ref[Origin[False]]` - the environment is fully sync. This will be propagated through all the call stack
`async def` = `ref[Origin[True]` - the environment is fully async. This will be propagated all the way through

5. I'm basically trying to maintain the current semantics of fully async and fully non-async execution in Python. So you can't call a fully async function in a fully sync environment. Instead, I introduce a bridge, which combines your ideas and Owen's in some way. If Mojo community is bootstrapped using this, we can have functions that can be called in both sync and async environments.

melodyogonna:
@nmsmith
Another thing. I'm trying to make my model work with yours, however, it's being done with so many different assumptions that it's probably better as separate proposal.

Here are my thinking models.
1. Let's not mix generators and coroutines, let's make coroutines its own type. Also, coroutines can not "yield" a different type compared to what the function returns.
if we try to translate this to your proposal.
```python
async? def function() -> Int: ...
```
would look something like:
```python
def function() yields -> Int: ...
```
which could be further be assumed to mean:
```python
def function() -> Int | Coroutine[1]: ...
```
2. Let's maintain the rule in Python that prevents generators from being used in a coroutine environment. And extend it so generators can't be used in a partially async function.
3. Whether the environment is async is determined by the function declaration rather than some compile-time parameter. so
```python
fn read_file(path: String) yields if config['async'] -> String:
```
becomes
```python
async? read_file(path: String) -> String:
```
when a function is fully async then the config is.
```python
async read_file(path: String) -> String:
```
is desugared to something akin to... I don't really know how to represent it with your model. But if we go with Owen's model
```python
async=async fn read[async: Bool = True](): ...
```
if partial async function is called, the environment it is executing in should be propagated by the function that called it.
4. We can then tie this down with your eager coroutine idea.  In a partial async function, coroutines are eager and always desugar to `coroutine.__await__` even if called without `await`. in a fully non-async environment, `await` can not be used, so coroutines always desugar to common `coroutine.__await__` like in your model. 
However, in a fully async environment, coroutine can be lazy.


Comparing this to current reference semantics.

`async? def` = `ref[_]` - the environment will be inferred
`def` = `ref[Origin[False]]` - the environment is fully sync. This will be propagated through all the call stack
`async def` = `ref[Origin[True]` - the environment is fully async. This will be propagated all the way through

5. I'm basically trying to maintain the current semantics of fully async and fully non-async execution in Python. So you can't call a fully async function in a fully sync environment. Instead, I introduce a bridge, which combines your ideas and Owen's in some way. If Mojo community is bootstrapped using this, we can have functions that can be called in both sync and async environments.

melodyogonna:
> Quoting myself:
> 
> > We could try to make coroutine have a "false async mode". And then we try to add some kind of compiler decorator which sets the path to materialization
> > ```mojo
> > @nonmaterializable
> > struct Coroutine[T: AnyType, sync: Bool, auto_deref: Bool = True]:
> >     fn __init__(out self: Coroutine[_, sync=False], handle: AnyCoroutine): ...
> >     fn __init__(out self: Coroutine[_, sync=True], handle: AnyFunction): ...
> > 
> >   @materialization_entrypoint(auto_deref=auto_deref)
> >   fn __await__(owned self) -> T: ...
> > ```
> 
> if the auto_deref parameter is there, it would mean we could control when coroutines are auto-awaited and when not.
> 
> This doesn't need to look exactly like what I proposed or even use materialization, but it would be necessary to sometimes switch no manual await

This is great.

nmsmith:
@ivellapillil, I just looked into Nim's effect-tracking system, and I've noticed that the creator of Nim has described the system as a [mistake](https://forum.nim-lang.org/t/9280) and as having a [bad cost-benefit tradeoff](https://github.com/nim-lang/RFCs/issues/404). So that's an immediate red flag.

That said, the curly-brace syntax that Nim uses in its function signatures has inspired me to consider whether Mojo effects should be bundled into a single data structure, rather than listed separately. (Note: Nim doesn't do this. Nim's `{.foo.}` braces are not a data structure.) By bundling effects together, it would be easier to write functions and traits that are generic over _arbitrary_ effects.

As a thought experiment, let's consider what would happen if we used Python's set literal syntax to bundle effects together:
`fn foo() does {Exception[DivideByZero], Yield}:`

The `does` keyword is just a placeholder, so let's avoid bikeshedding it for the moment. Also, let's gloss over what exactly `Exception` and `Yield` are. Maybe they are types; maybe they are some other kind of compile-time value.

If effects are bundled into a single value, and we have the `any_fn` keyword (proposed in my last post), the syntax for declaring that higher-order function is generic over _all_ effects would reduce to:
`fn higher_order(f: any_fn ()) does f.effects:`

The compiler would need to have special logic for reasoning about the `f.effects` data structure. If it's possible for `f.effects` to include `Exception` or `Yield`, then the compiler must treat calls to `f` as altering control flow, etc.

Come to think of it, `any_fn` plays a very similar role to the `async? fn` syntax that Owen and Melody have been talking about. (And that Rust's keyword-generics initiative has proposed.) Here's an idea: let's consider having an `effectful` keyword (the name is a placeholder), that generalizes over all possible effects a function might have, including `async`. With this keyword, the signature of `higher_order` would be written:
`fn higher_order(f: effectful fn ()) does f.effects:`

`effectful` could also be used in traits, to declare that a method may have arbitrary effects:
```
trait Iterator[T: AnyType]:
    effectful fn __next__(mut self) does Exception[StopIteration] -> T: ...
```

Compare this to the definition of `Iterator` that I presented in my proposal document:
```
trait Iterator[T: AnyType]:
    alias yields_: Bool
    fn __next__(mut self) raises StopIteration, yields if yields_ -> T: ...
```

The definition based on `effectful` is both more concise, and more general, because it allows implementations of `__next__` to have arbitrary effects. Here's what a function that takes an `Iterator` would look like:
```
fn get_integer[IterType: Iterator[Int]](mut iterator: IterType) does IterType.__next__.effects -> Int:
    try:
        return next(iterator)
    except StopIteration:
        return 0
```

The appearance of `__next__` in the signature is a bit clunky, so maybe Mojo should have a built-in function named `effects()` that can return the effects associated with either a single function, or the union of all the effects of the methods available on a type:
```
fn get_integer[IterType: Iterator[Int]](mut iterator: IterType) does effects(IterType) -> Int:
    try:
        return next(iterator)
    except StopIteration:
        return 0
```

There's actually a small problem here: `effects(IterType)` would seemingly include the `StopIteration` effect, but we don't want `get_integer` to inherit that effect. We'd need to consider how to clarify that.

## Advantages of this design (vs. my published proposal)

This design would put us in a good position to add **additional kinds of effects** to Mojo in the future. It would also allow us to support **library-defined effects**, which I know is something that @owenhilyard was interested in.

I can already think of another **major** class of effect that we haven't discussed: reading and/or mutating non-local state. Closures often do this:
```
fn main():
    counter = 0
    fn foo(x: Int) mut counter -> String:
        counter += x
        return bar(x)
        
    foo(5)               # Invoke 'foo'
    higher_order(foo)    # Let somebody else invoke 'foo'
```

The fact that `foo` mutates non-local state is **very** important. For one thing, it means `foo` must not be executed on multiple threads concurrently. Despite this, we still want functions like `higher_order` to be able to take a closure as an argument, and inherit its effects. If `foo` can mutate `counter`, then so can `higher_order(foo)`.

This suggests that maybe the `effectful` keyword should prevent a function from being shared between threads. Maybe it can somehow be connected to the`Send` and `Sync` traits that Rust offers, and which have been contemplated for Mojo.

In summary, maybe `raises`, `yields`, and closures can be folded into one language feature: effects. Every effectful function has "associated effects", accessible via the syntax `f.effects`. This gives us the ability to define higher-order functions that inherit the effects of the functions they call. Notably, every generic function is implicitly a higher-order function: for each trait bound on its type parameters, it receives a set of function implementations.

## Disadvantages of this design (vs. my published proposal)

As a consequence of this design being more general, it is also less explicit. In particular, by integrating whether a function `raises` into the `effectful` and `does <effect>` syntaxes, we might be making it harder to reason about the **control flow** of Mojo programs. Whenever a programmer invokes an `effectful` function, they would need to program defensively, on the assumption that the function may raise.

It's not clear whether this would be a problem in practice. Certainly a Python programmer wouldn't be bothered by this, because in Python, _all_ functions can raise. The same is true of most other programming languages. Maybe as long as we make it clear that `effectful` implies `raises`, that will be sufficient.

## Next steps

I think we should develop this design further, so that we can compare the two alternatives:
1. Supporting only two effects: `raises` and `yields`.
2. Supporting a whole menagerie of effects, and offering a way to abstract over **all** effects at once.

This concludes my brain dump. 🙂

ivellapillil:
If we introduce custom effects, we would need custom effect handlers. 

1. Where would those custom effect handlers live? 
2. How would we hand over the function to them? 
3. What all facilities we need so that the handlers can re-write the functions to handle the effects (e.g. machinery for Coroutine). 
4. How would I tell the compiler that an effect is fully handled within a function and therefore should not be propagated (try-except that catches the top most exception). 
5. The handler might need to analyse/operate upon the outer scope in addition to just the function (capture analysis in case of closure)

I think generic **propagation** of effects is easy - but challenges come in when we want to handle them. 

Overall, I suspect that while generic effect handling would be super cool, it would be quite complex to pull off all kinds of effects (basically would need AST transformations, communication with the compiler etc.). At the very least it could slow down compilation.

We might be better to treat raises and async/yield and one or two more effects hard-coded in the compiler.

nmsmith:
@ivellapillil I didn't mean to suggest that Mojo should have algebraic effects with custom handlers, etc. I was thinking a bit smaller: we could let programmers put custom data in the `does` clause, and this data is (forcibly) inherited by callers, etc. But if users don't have a way to stop propagation, maybe this is not sufficient in practice.

> We might be better to treat raises and async/yield and one or two more effects hard-coded in the compiler.

Yeah, this is really the target I was aiming for: abstracting over a bunch of built-in effects. We could leave custom effects as something to investigate in the (distant) future.

owenhilyard:
@nmsmith 
> As a thought experiment, let's consider what would happen if we used Python's set literal syntax to bundle effects together: `fn foo() does {Exception[DivideByZero], Yield}:`

I think you're part of the way to realizing why I proposed the Mojo Parameter store. I wanted key/value pairs with arbitrary Mojo objects, so that `Exception` could be `List[T: Error]`, `Yield` could just be an `AnyType`, etc. So, I would prefer closer to `fn foo() does {exception=[DivideByZero], yield=Never}`. 

> If effects are bundled into a single value, and we have the `any_fn` keyword (proposed in my last post), the syntax for declaring that higher-order function is generic over _all_ effects would reduce to: `fn higher_order(f: any_fn ()) does f.effects:`

I like this idea, but I also want to be able to do `fn higher_order(f: any_fn ()) does {**f.effects, exception=[StopIteration, *f.effects.exception]}`

> The compiler would need to have special logic for reasoning about the `f.effects` data structure. If it's possible for `f.effects` to include `Exception` or `Yield`, then the compiler must treat calls to `f` as altering control flow, etc.

I think the special handling would actually be to keep the data structure strongly typed. `does` just evaluates a comptime expression to produce the data structure.  

> Come to think of it, `any_fn` plays a very similar role to the `async? fn` syntax that Owen and Melody have been talking about. (And that Rust's keyword-generics initiative has proposed.) Here's an idea: let's consider having an `effectful` keyword (the name is a placeholder), that generalizes over all possible effects a function might have, including `async`. With this keyword, the signature of `higher_order` would be written: `fn higher_order(f: effectful fn ()) does f.effects:`
> 
> `effectful` could also be used in traits, to declare that a method may have arbitrary effects:
> 
> ```
> trait Iterator[T: AnyType]:
>     effectful fn __next__(mut self) does Exception[StopIteration] -> T: ...
> ```
> 
> Compare this to the definition of `Iterator` that I presented in my proposal document:
> 
> ```
> trait Iterator[T: AnyType]:
>     alias yields_: Bool
>     fn __next__(mut self) raises StopIteration, yields if yields_ -> T: ...
> ```
> 
> The definition based on `effectful` is both more concise, and more general, because it allows implementations of `__next__` to have arbitrary effects. Here's what a function that takes an `Iterator` would look like:

I think I like the "compile-time data structure" approach more. And, remember that for compile time reasons we want to minimize the amount of information that can "travel" in multiple directions. An iterator may need some input, in the form of whether it's async, whether it can block, etc. 

> ```
> fn get_integer[IterType: Iterator[Int]](mut iterator: IterType) does IterType.__next__.effects -> Int:
>     try:
>         return next(iterator)
>     except StopIteration:
>         return 0
> ```
> 
> The appearance of `__next__` in the signature is a bit clunky, so maybe Mojo should have a built-in function named `effects()` that can return the effects associated with either a single function, or the union of all the effects of the methods available on a type:
>
> ```
> fn get_integer[IterType: Iterator[Int]](mut iterator: IterType) does effects(IterType) -> Int:
>     try:
>         return next(iterator)
>     except StopIteration:
>         return 0
> ```

I think making it clear it's a property of the function is a good thing. `effects(IterType)` doesn't indicate which function the effects are taken from. You can have a single struct with mixed sync and async methods, some of which may raise. 

> There's actually a small problem here: `effects(IterType)` would seemingly include the `StopIteration` effect, but we don't want `get_integer` to inherit that effect. We'd need to consider how to clarify that.
> ## Advantages of this design (vs. my published proposal)
> 
> This design would put us in a good position to add **additional kinds of effects** to Mojo in the future. It would also allow us to support **library-defined effects**, which I know is something that @owenhilyard was interested in.

Library defined effects are something I absolutely want, as a way to propagate "how do I handle this situation?" through the stack. We don't want all of that to live in the compiler, so ideally we do library defined effects and have the standard library define "general" effects for common situations.

> I can already think of another **major** class of effect that we haven't discussed: reading and/or mutating non-local state. Closures often do this:
> 
> ```
> fn main():
>     counter = 0
>     fn foo(x: Int) mut counter -> String:
>         counter += x
>         return bar(x)
>         
>     foo(5)               # Invoke 'foo'
>     higher_order(foo)    # Let somebody else invoke 'foo'
> ```
> 
> The fact that `foo` mutates non-local state is **very** important. For one thing, it means `foo` must not be executed on multiple threads concurrently. Despite this, we still want functions like `higher_order` to be able to take a closure as an argument, and inherit its effects. If `foo` can mutate `counter`, then so can `higher_order(foo)`.

I think turning the origin checker into an effect is something that would need to be discussed a lot more. Effects are opt-in to some degree, and that shouldn't be.
 
> This suggests that maybe the `effectful` keyword should prevent a function from being shared between threads. Maybe it can somehow be connected to the`Send` and `Sync` traits that Rust offers, and which have been contemplated for Mojo.

I think `Send` and `Sync` are the right abstractions for this, since if a capture of a closure is `!Send` or `!Sync`, the closure inherits that, although a closure with a `!Sync` capture becomes `!Send`, since `T: Sync` is really `&T: Send` in Rust. 

> In summary, maybe `raises`, `yields`, and closures can be folded into one language feature: effects. Every effectful function has "associated effects", accessible via the syntax `f.effects`. This gives us the ability to define higher-order functions that inherit the effects of the functions they call. Notably, every generic function is implicitly a higher-order function: for each trait bound on its type parameters, it receives a set of function implementations.

Closures have to have a place to store captured information, which means they need to interact with allocations in a way most functions do not. I'm open to exploring the idea, but that could get messy. 

> Whenever a programmer invokes an `effectful` function, they would need to program defensively, on the assumption that the function may raise.

This is why I prefer the more explicit, data-structure-based method. You can get something like `effectful` with `does {**f.effects}`, but you don't have to do that if you want to be a bit more explicit about what effects you propagate. If you didn't handle a situation when you do that, the user gets a compiler error and knows to not use your library for that.

> It's not clear whether this would be a problem in practice. Certainly a Python programmer wouldn't be bothered by this, because in Python, _all_ functions can raise. The same is true of most other programming languages. Maybe as long as we make it clear that `effectful` implies `raises`, that will be sufficient.

I would much rather not force the cognitive load of defensively programming for things which cannot happen. There are also contexts, such as in some `unsafe` areas, where raising is a very bad idea, for instance if you are inside of signal handler and want to take a callback. 

> ## Next steps
> 
> I think we should develop this design further, so that we can compare the two alternatives:
> 
>     1. Supporting only two effects: `raises` and `yields`.
> 
>     2. Supporting a whole menagerie of effects, and offering a way to abstract over **all** effects at once.

I think we should aim for a general solution, but do a slow rollout. That way, we can try to roll out a full effect system and then if that doesn't work due to compile time or other reasons, we can lock that capability off and make the compiler provide the few effects we are willing to support, instead of allowing library definitions.

nmsmith:
@owenhilyard 

> I wanted key/value pairs with arbitrary Mojo objects, so that `Exception` could be `List[T: Error]`, `Yield` could just be an `AnyType`, etc. So, I would prefer closer to `fn foo() does {exception=[DivideByZero], yield=Never}`.

Hmm.. yes, maybe an effect `Dict` makes more sense than a `Set`. Good point. But note: if we're using Python's syntax for dictionary literals, we'd be writing `{'exception': [DivideByZero], 'yield': Never}`.

> I also want to be able to do `fn higher_order(f: any_fn ()) does {**f.effects, 'exception': [StopIteration, *f.effects.exception]}`

I agree we would need the ability to add effects to an existing dictionary, but I find the above syntax a bit noisy. Maybe we could have an operator for this. Python's `dict` defines an `|` operator, but unfortunately it doesn't do deep merging. We could either make `EffectDict.__or__` do a deep merge, or we could use a different operator, e.g. `+`:
`fn higher_order(f: any_fn ()) does f.effects + {'exception': [StopIteration]}`

> I think the special handling would actually be to keep the data structure strongly typed. `does` just evaluates a comptime expression to produce the data structure.

I think what you mean is that as long as the compiler knows what the value of `e` is in the expression `does e`, then it will be able to tell whether `e` contains a `raises` or `yields` effect and therefore the compiler can do all of the special work it needs to do to manage these effects. I agree with that assessment.

My original comment was moreso asserting that the compiler will need to do "special work" for exceptions and yielding, etc, even if Mojo allows the effect dictionary to contain custom effects. It's also not clear to me how/whether custom effects would be useful in the absence of custom effect _handlers_. And even more importantly: I have no idea what benefits custom effects + custom handlers would bring to Mojo, and whether it is worth the complexity. Personally, I can only see a solid use case for the `raises` and `yields` effects right now.

> I think I like the "compile-time data structure" approach more.

I think you've misunderstood the `effectful fn` syntax. It was meant to be semantically identical to `any_fn`. I just split the keyword into two components (`effectful` and `fn`), to make it syntactically similar to the `async? fn` syntax that has been brought up a few times. By doing this, I've hopefully demonstrated that `async?` is a special case of a more general/more principled feature.

When `effectful fn foo` appears in a trait definition, it's just shorthand for requiring the implementation of `foo` to have an `effects` field that enumerates its effects. That's all it does. In the absence of `effectful` (or a `does` clause), the implementation is required to have **no effects.**

>  `effects(IterType)` doesn't indicate which function the effects are taken from. You can have a single struct with mixed sync and async methods, some of which may raise.

That's true, but my rationale for proposing `effects(MyType)` is to give a function's author a way to declare that the function may invoke _any_ method on the type, and therefore the function inherits _all_ of those method's effects. This is a lot more concise than listing every method the function invokes in its signature. For many functions, that wouldn't be practical.

> Library defined effects are something I absolutely want, as a way to propagate "how do I handle this situation?" through the stack. We don't want all of that to live in the compiler, so ideally we do library defined effects and have the standard library define "general" effects for common situations.

I don't see the use cases for custom effects, so maybe you should put together a separate proposal at some point (built atop the design(s) we've been discussing), that makes a compelling case for them. In the meantime, I'm probably going to operate on the assumption that we only need `raises`, `yields`, and (if I can make it fit in), `read` and `mut` operations upon variables captured by closures.

> I think turning the origin checker into an effect is something that would need to be discussed a lot more.

Yes, whether closure captures can/should be modelled as effects requires a _lot_ more investigation, I agree.

> I would much rather not force the cognitive load of defensively programming for things which cannot happen.

I don't understand what your point is here. If you're invoking a function that has unknown effects, you have no choice but to operate on the assumption that the function may raise or yield. That's an inescapable consequence of writing effect-generic code.

> I think we should aim for a general solution, but do a slow rollout. That way, we can try to roll out a full effect system and then if that doesn't work due to compile time or other reasons, we can lock that capability off.

Or conversely: we can choose a design that is _capable_ of supporting additional effects, but we only put effort into supporting `async` and `raises` at first. This would allow us to get things done quicker. We don't want to be blocked on needing to come up with a full design for custom effects.

jackos:
Hi @nmsmith very cool idea, have added to our agenda to discuss your proposal

nmsmith:

nmsmith:
This proposal is now almost 6 months old, and I've been reflecting on it.

The overall goal is to avoid forcing generic functions and higher-order functions to specify the "color" of trait/function that they accept. For example, we don't want Mojo to have both:
- the traits `Iterator` and `AsyncIterator`
- the functions `map` and `map_async` (and the same for `filter`, `reduce`, etc.)

I think I've figured out a very simple way to achieve this. Certainly, it's much simpler than the `yields T` proposal that constitutes this PR. Here is the idea:
- **We only allow function _implementations_ to be labelled as sync (the implicit default) or `async`.** A declaration that lacks an implementation must remain unlabelled. For example, inside a trait declaration, the `async` keyword is illegal, and irrelevant. (One exception: if a trait method has a default implementation, that can be declared `async`.)
- When implementing a trait method, the implementer has the full freedom to choose whether their implementation is sync or `async`.
- The asyncness of a generic function is not locked in until it has been instantiated. More precisely, a generic function only needs to be declared `async` if any of its concrete/non-parameter dependencies are `async`. (For example, maybe it waits for the outcome of an async HTTP request.) If none of the concrete dependencies are async, the generic function's asyncness depends upon the value of its parameters. For example:
    - `fn map[func: fn (Int) -> Int](items: List[Int]) -> List[Int]` is considered `async` if and only if its parameter `func` is `async`. Only the _caller_ of `map` has this information.
    - `fn get_next[Iter: Iterator[Int]](iter: Iter) raises -> Int: return next(iter)` is considered `async` if and only if the concrete implementation of `Iter.__next__` is `async`.
    - More generally, `fn foo[T: SomeTrait](...): pass` is `async` if and only if at least one of `T`'s implementations of the methods of `SomeTrait` is `async`. Only the _caller_ of `foo` has this information.
    
In this model, only the person who _instantiates_ a generic function knows whether it is `async` or not. If it is, then the enclosing function must also be declared `async`.

The beauty of this model is that **every function that is generic is _also_ generic over asyncness.** The latter property just magically falls out of the fact that a function inherits the asyncness of the functions/trait implementations it is instantiated with.

Another way to think about this proposal: the absence of `async` should be understood to mean "_maybe_ async", unless you are looking at a fully instantiated function implementation.

And of course: I'm still proposing that an `async` function be compiled to a stackless coroutine. Nothing has changed there.

There are some subtleties regarding when to use `await`, but it will be easy to resolve those.

Consider this my new (or alternative) proposal for `async` in Mojo. It's _dramatically_ simpler than my `yields T` proposal. If people think this new idea is promising, I can write it up as a standalone proposal. Alternatively, if you don't think this idea can work, I'd be keen to hear why.

owenhilyard:
> This proposal is now almost 6 months old, and I've been reflecting on it.
> 
> The overall goal is to avoid forcing generic functions and higher-order functions to specify the "color" of trait/function that they accept. For example, we don't want Mojo to have both:
> 
>     * the traits `Iterator` and `AsyncIterator`
> 
>     * the functions `map` and `map_async` (and the same for `filter`, `reduce`, etc.)
> 

Agree

> I think I've figured out a very simple way to achieve this. Certainly, it's much simpler than the `yields T` proposal that constitutes this PR. Here is the idea:
> 
>     * **We only allow function _implementations_ to be labelled as sync (the implicit default) or `async`.** A declaration that lacks an implementation must remain unlabelled. For example, inside a trait declaration, the `async` keyword is illegal, and irrelevant. (One exception: if a trait method has a default implementation, that can be declared `async`.)

There are some things where asyncness or syncness may be fundamental, so I don't agree with this. However, I think that having it be "don't care" by default is fine. 

>     * When implementing a trait method, the implementer has the full freedom to choose whether their implementation is sync or `async`.
> 
>     * The asyncness of a generic function is not locked in until it has been instantiated. A generic function must declare itself to be `async` if it performs I/O directly. But if it doesn't, its asyncness depends upon the value of its type parameters. For example:

There are cases where doing IO might be sync, such as a `List[UInt8]` where IO is actually memcpy. 
      
>       * `fn map[func: fn (Int) -> Int](items: List[Int]) -> List[Int]` is considered `async` if and only if its parameter `func` is `async`. Only the _caller_ of `map` has this information.

I think that we may need to have that be generic and queryable so that users can make decisions about how to invoke that functions alongside other.

>       * `fn get_next[Iter: Iterator[Int]](iter: Iter) raises -> Int: return next(iter)` is considered `async` if and only if the concrete implementation of `Iter.__next__` is `async`.
>       * More generally, `fn foo[T: SomeTrait](...): pass` is `async` if and only if at least one of `T`'s implementations of the methods of `SomeTrait` is `async`. Only the _caller_ of `foo` has this information.
> 
> 
> In this model, only the person who _instantiates_ a generic function knows whether it is `async` or not. If it is, then the enclosing function must also be declared `async`.

I think this leads to the "paint bucket" approach to any functions which take function parameters, and might not be workable on GPUs. 

> The beauty of this model is that **every function that is generic is _also_ generic over asyncness.** The latter property just magically falls out of the fact that a function inherits the asyncness of the functions/trait implementations it is instantiated with.
> 
> Another way to think about this proposal: the absence of `async` should be understood to mean "_maybe_ async", unless you are looking at a fully instantiated function implementation.
> 
> And of course: I'm still proposing that an `async` function be compiled to a stackless coroutine. Nothing has changed there.
> 
> There are some subtleties regarding when to use `await`, but it will be easy to resolve those.
> 
> Consider this my new (or alternative) proposal for `async` in Mojo. It's _dramatically_ simpler than my `yields T` proposal. If people think this new idea is promising, I can write it up as a standalone proposal. Alternatively, if you don't think this idea can work, I'd be keen to hear why.

nmsmith:
> There are some things where asyncness or syncness may be fundamental, so [it should be possible for traits to restrict their methods to `sync` or `async`]. However, I think that having it be "don't care" by default is fine.

Hmm, true, there's possibly value in being able to define traits where the implementation are forced to be sync. FFI comes to mind.

> There are cases where doing IO might be sync

That's true. I misspoke. I've now corrected that sentence.

> I think that we may need to have [asyncness] be generic and queryable so that users can make decisions about how to invoke that function alongside others.

Sure, that makes sense. Just as we have `T.__moveinit__is_trivial_unsafe` for inspecting whether a move constructor is trivial, I can imagine offering `func.is_async` for inspecting whether a function is async.

> I think this leads to the "paint bucket" approach to any functions which take function parameters, and might not be workable on GPUs.

I'm not sure what you mean by a "paint bucket" approach. You mean `async` is contagious? That would be true. For GPUs, I think you're alluding that GPU functions can't be async? If that's true, we can resolve that by offering the aforementioned `sync` keyword. This would force a generic function's instantiation to be `sync`, or else raise a type error:

`sync fn foo[T: SomeTrait](...):`

In the above example, if the caller tried to pass an implementation of `SomeTrait` where one or more methods was `async`, a type error would be raised.

In some cases, it might make more sense to attach the `sync` keyword to the relevant parameters:

`fn foo[T: sync SomeTrait](...):`

Ironing out these details would require a bit of time and attention, but I don't see any roadblocks.

nmsmith:
@JoeLoser Just FYI, I forgot I had an open pull request drawing from my clone of the Modular repo, so I didn't mean to close this thread. Not sure how you (or Denis) would like to keep track of the proposal(s) in this thread going forward (if at all).

PR #3988, opened 2025-02-05
[stdlib] [proposal] String, ASCII, Unicode, UTF, Graphemes

martinvuyk:
This is a proposal on the main goals. Followed by a proposed approach to get there.

I would like us to reach a consensus on this topic given it will affect many projects, and involve a lot of work to fully support.

Everyone involved with strings or that should be part of this conversation I can think of: @JoeLoser @ConnorGray @lsh @jackos @mzaks @bgreni @thatstoasty @leb-kuchen

### DIFF
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -23,11 +23,11 @@ with supporting only ASCII is that other encodings can get corrupted.
 - UTF-16 is 1-2 sets of 16 bits long
 - UTF-32 is 1 set of 32 bits long
 
-#### When slicing by unicode codepoint e.g. "🔥🔥🔥" (\U0001f525\U0001f525\U0001f525)
+### When slicing by unicode codepoint e.g. "🔥🔥🔥" (\U0001f525\U0001f525\U0001f525)
 
-- UTF-8: 12 sets (12 bytes) long. The first byte of each fire can be used to know the
-length, and the next bytes are what is known as a continuation byte. There are
-several approaches to achieve the slicing, they can be explored with
+- UTF-8: 12 sets (12 bytes) long. The first byte of each fire can be used to
+know the length, and the next bytes are what is known as a continuation byte.
+There are several approaches to achieve the slicing, they can be explored with
 benchmarking later on.
 - UTF-16: 6 sets long. It's very similar in procedure to UTF-8.
 - UTF-32: It is fastest since it's direct index access. This is not the case
@@ -47,11 +47,11 @@ to see if it's an extended cluster *.
 exists the posibility of implementing only a subset, but that defeats the
 purpose of going through the hassle of supporting it in the first place.
 
-# Context
+## Context
 
 C, C++, and Rust use UTF-8 with no default support for graphemes.
 
-## Swift
+### Swift
 
 Swift is an interesting case study. For compatibility with Objective-C they went
 for UTF-16. They decided to support grapheme clusters by default. They recently
@@ -60,7 +60,7 @@ that I think inspired our current `Char` type, it is a generic representation of
 a Character in any encoding, that can be from one codepoint up to any grapheme
 cluster.
 
-## Python
+### Python
 
 Python currently uses UTF-32 for its string type. So the slicing and indexing
 is simple and fast (but consumes a lot of memory, they have some tricks to
@@ -68,12 +68,12 @@ reduce it). They do not support graphemes by default. Pypi is implementing a
 UTF-8 version of Python strings, which keeps the length state every x
 characters.
 
-## Mojo
+### Mojo
 
 Mojo aims to be close to Python yet be faster and customizable, taking advantage
 of heterogeneous hardware and modern type system features.
 
-# Value vs. Reference
+## Value vs. Reference
 
 Our current `Char` type uses a u32 as storage, every time an iterator that
 yields `Char`is used, an instance is parsed from the internal UTF-8 encoded
@@ -83,7 +83,9 @@ The default iterator for `String` returns a `StringSlice` which is a view into
 the character in the UTF-8 encoded `StringSlice`. This is much more efficient
 and does not add any complexity into the type system nor developer headspace.
 
-# Now, onto the Proposal
+## Now, onto the Proposal
+
+### Goals
 
 #### Hold off on developing Char further and remove it from stdlib.builtin
 
@@ -108,11 +110,11 @@ which have mostly ASCII characters but still keep full unicode support.
 #### Grapheme support
 Grapheme support should exist but be opt-in due to their high compute cost.
 
-## One concrete (tentative) way forward
+### One concrete (tentative) way forward
 
 With a clear goal in mind, this is a concrete (tentative) way forward.
 
-### Add parameters to String and StringSlice
+#### Add parameters to String and StringSlice
 
 ```mojo
 struct Encoding:
@@ -155,7 +157,7 @@ indexing-agnostic)**.
 
 The implementations can be added with time making liberal use of `constrained`.
 
-### Adapt StringSliceIter
+#### Adapt StringSliceIter
 
 The default iterator should iterate in the way in which the `String` is
 parametrized. The iterator could then have functions which return the other
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -21,15 +21,15 @@ with supporting only ASCII is that other encodings can get corrupted.
 
 - UTF-8 is 1-4 sets of 8 bits long
 - UTF-16 is 1-2 sets of 16 bits long
-- UTF-32 is 1 set 32 bits long
+- UTF-32 is 1 set of 32 bits long
 
 #### When slicing by unicode codepoint e.g. "🔥🔥🔥" (\U0001f525\U0001f525\U0001f525)
 
-- UTF-8: 12 bytes long. The first byte of each fire can be used to know the
+- UTF-8: 12 sets (12 bytes) long. The first byte of each fire can be used to know the
 length, and the next bytes are what is known as a continuation byte. There are
 several approaches to achieve the slicing, they can be explored with
 benchmarking later on.
-- UTF-16: 6 bytes long. It's very similar in procedure to UTF-8.
+- UTF-16: 6 sets long. It's very similar in procedure to UTF-8.
 - UTF-32: It is fastest since it's direct index access. This is not the case
 when supporting graphemes.
 
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -121,6 +121,7 @@ struct Encoding:
     alias UTF8 = 0
     alias UTF16 = 1
     alias UTF32 = 2
+    alias ASCII = 3
 
 struct Indexing:
     alias DIRECT = 0
@@ -179,9 +180,8 @@ for c in iter(data).chars[encoding.UTF32]():
 
 # We could also have lazy iterators which return StringSlices according to the
 # encoding and indexing parameter.
-# In the case of indexing.DIRECT and Encoding.UTF8 or UTF16 unicode paragraph
-# and line separator (\u2028 and \u2029) can be ignored and thus the processing
-# is much faster.
+# In the case of Encoding.ASCII unicode separators (\x85, \u2028, \u2029) can
+# be ignored and thus the processing is much faster.
 for c in iter(data).split():
   ... # StringSlice lazily split by all whitespace
 for c in iter(data).splitlines():
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -1,0 +1,188 @@
+# String, ASCII, Unicode, UTF, Graphemes
+
+The current proposal will attempt to unify the handling of the standards.
+Providing nice ergonomics, as Python-like defaults as possible, and keeping the
+door open for optimizations.
+
+## String
+
+String is currently an owning type which encodes string data in UTF-8, which is
+a variable length encoding format. Data can be 1 byte up to 4 bytes long. ASCII
+text, which is where English text falls, is 1 byte long. As such, the defaults
+should optimize for that character set given it is most typical on the internet
+and on backend systems.
+
+## ASCII
+
+Ascii is pretty optimizable, there are many known tricks. The big problem
+with supporting only ASCII is that other encodings can get corrupted.
+
+## UTF standards
+
+- UTF-8 is 1-4 sets of 8 bits long
+- UTF-16 is 1-2 sets of 16 bits long
+- UTF-32 is 1 set 32 bits long
+
+#### When slicing by unicode codepoint e.g. "🔥🔥🔥" (\U0001f525\U0001f525\U0001f525)
+
+- UTF-8: 12 bytes long. The first byte of each fire can be used to know the
+length, but for slicing it's faster to jump and count the continuation
+bytes before the indices, then shift the index access to the right by that
+amount. There are several other approaches to do achieve the same result, they
+can be explored with benchmarking later on.
+- UTF-16: 6 bytes long. It's very similar in procedure to UTF-8.
+- UTF-32: It is fastest since it's direct index access. This is not the case
+when supporting graphemes.
+
+## Graphemes
+
+Graphemes are an extension which allows unicode codepoints to modify the end
+result through concatenation with other codepoints. Visually they are shown as a
+single unit e.g. é is actually comprised of `e` followed by `´`  (\x65\u0301).
+
+Graphemes are more expensive to slice because one can't use faster algorithms,
+only skipping each unicode codepoint one at a time and checking the last byte
+to see if it's an extended cluster *.
+
+*: extended clusters is the full feature complete grapheme standard. There
+exists the posibility of implementing only a subset, but that defeats the
+purpose of going through the hassle of supporting it in the first place.
+
+# Context
+
+C, C++, and Rust use UTF-8 with no default support for graphemes.
+
+## Swift
+
+Swift is an interesting case study. For compatibility with Objective-C they went
+for UTF-16. They decided to support grapheme clusters by default. They recently
+changed the preferred encoding from UTF-16 to UTF-8. They have a `Character`
+that I think inspired our current `Char` type, it is a generic representation of
+a Character in any encoding, that can be from one codepoint up to any grapheme
+cluster.
+
+## Python
+
+Python currently uses UTF-32 for its string type. So the slicing and indexing
+is simple and fast (but consumes a lot of memory, they have some tricks to
+reduce it). They do not support graphemes by default. Pypi is implementing a
+UTF-8 version of Python strings, which keeps the length state every x
+characters.
+
+## Mojo
+
+Mojo aims to be close to Python yet be faster and customizable, taking advantage
+of heterogeneous hardware and modern type system features.
+
+# Value vs. Reference
+
+Our current `Char` type uses a u32 as storage, every time an iterator that
+yields `Char`is used, an instance is parsed from the internal UTF-8 encoded
+`StringSlice` (into UTF-32).
+
+The default iterator for `String` returns a `StringSlice` which is a view into
+the character in the UTF-8 encoded `StringSlice`. This is much more efficient
+and does not add any complexity into the type system nor developer headspace.
+
+# Now, onto the Proposal
+
+#### Hold off on developing Char further and remove it from stdlib.builtin
+
+`Char` is currently expensive to create and use compared to a `StringSlice`
+which is a view over the original data. There is also the problem that it forces
+UTF-32 on the data, and those transformations are expensive.
+
+We can revisit `Char` later on making it take encoding into account. But the
+current state of the type makes it add more complexity than strictly necessary.
+
+#### Full ASCII optimizations
+If someone wishes to use a `String` as if it's an ASCII-only String, then there
+should either be a parameter that signifies that, or the stdlib/community should
+add an `ASCIIString` type which has all optimizations possible for such
+scenarios.
+
+#### Mostly ASCII optimizations
+Many functions can make use of branch predition, instruction and data
+prefetching, and algorithmic tricks to make processing faster for languages
+which have mostly ASCII characters but still keep full unicode support.
+
+#### Grapheme support
+Grapheme support should exist but be opt-in due to their high compute cost.
+
+## One concrete (tentative) way forward
+
+With a clear goal in mind, this is a concrete (tentative) way forward.
+
+### Add parameters to String and StringSlice
+
+```mojo
+struct Encoding:
+    alias UTF8 = 0
+    alias UTF16 = 1
+    alias UTF32 = 2
+
+struct Indexing:
+    alias DIRECT = 0
+    alias CODEPOINT = 1
+    alias GRAPHEME = 2
+
+alias ASCIIString = String[encoding=Encoding.UTF8, indexing=Indexing.DIRECT]
+
+struct String[
+    encoding: Encoding = Encoding.UTF8,
+    indexing: Indexing = Indexing.CODEPOINT,
+]:
+    ... # data is bitcasted to bigger DTypes when encoding is 16 or 32 bits
+
+struct StringSlice[
+    encoding: Encoding = Encoding.UTF8,
+    indexing: Indexing = Indexing.CODEPOINT,
+]:
+    ... # data is bitcasted to bigger DTypes when encoding is 16 or 32 bits
+```
+
+#### What this requires
+
+First, we can add the parameters and constraint on the supported encodings.
+
+Then, we need to rewrite every function signature that makes two `String`s
+interact with one another, where the code doesn't require both to be the
+defaults.
+
+Many of those functions will need to be branched when the encodings are
+different and allocations for parsing added. But the default same-encoding
+ones will remain the same **(as long as the code is encoding and
+indexing-agnostic)**.
+
+The implementations can be added with time making liberal use of `constrained`.
+
+### Adapt StringSliceIter
+
+The default iterator should iterate in the way in which the `String` is
+parametrized. The iterator could then have functions which return the other
+types of iterators for ergonomics (or have constructors for each) *.
+
+*: this is a workaround until we have generic iterator support.
+
+e.g.
+```mojo
+data = String("123 \n\u2029🔥")
+for c in data:
+  ... # StringSlice in the same encoding and indexing scheme by default
+
+# Once we have Char developed enough
+for c in iter(data).chars[encoding.UTF8]():
+  ... # Char type instances
+for c in iter(data).chars[encoding.UTF32]():
+  ... # Char type instances
+
+# We could also have lazy iterators which return StringSlices according to the
+# encoding and indexing parameter.
+# In the case of indexing.DIRECT and Encoding.UTF8 or UTF16 unicode paragraph
+# and line separator (\u2028 and \u2029) can be ignored and thus the processing
+# is much faster.
+for c in iter(data).split():
+  ... # StringSlice lazily split by all whitespace
+for c in iter(data).splitlines():
+  ... # StringSlice lazily split by all newline characters
+```
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -128,7 +128,7 @@ struct Indexing:
     alias CODEPOINT = 1
     alias GRAPHEME = 2
 
-alias ASCIIString = String[encoding=Encoding.UTF8, indexing=Indexing.DIRECT]
+alias ASCIIString = String[encoding=Encoding.ASCII, indexing=Indexing.DIRECT]
 
 struct String[
     encoding: Encoding = Encoding.UTF8,
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -171,7 +171,7 @@ for c in data:
 
 # Once we have Char developed enough
 for c in iter(data).chars[encoding.UTF8]():
-  ... # Char type instances
+  ... # Char type instances (UTF8 packed in a UInt32 ?)
 for c in iter(data).chars[encoding.UTF32]():
   ... # Char type instances
 
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -26,10 +26,9 @@ with supporting only ASCII is that other encodings can get corrupted.
 #### When slicing by unicode codepoint e.g. "🔥🔥🔥" (\U0001f525\U0001f525\U0001f525)
 
 - UTF-8: 12 bytes long. The first byte of each fire can be used to know the
-length, but for slicing it's faster to jump and count the continuation
-bytes before the indices, then shift the index access to the right by that
-amount. There are several other approaches to do achieve the same result, they
-can be explored with benchmarking later on.
+length, and the next bytes are what is known as a continuation byte. There are
+several approaches to achieve the slicing, they can be explored with
+benchmarking later on.
 - UTF-16: 6 bytes long. It's very similar in procedure to UTF-8.
 - UTF-32: It is fastest since it's direct index access. This is not the case
 when supporting graphemes.
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -1,8 +1,10 @@
 # String, ASCII, Unicode, UTF, Graphemes
 
-Edit 0: 2025-02-05. Created.
-Edit 1: 2025-04-27. Reduced the scope to only affect `StringSlice` and fixed
-some details after feedback.
+- Edit 0: 2025-02-05. Created.
+- Edit 1: 2025-04-27. Reduced the scope to only affect `StringSlice` and fixed
+  some details after feedback.
+- Edit 1: 2025-05-11. Reduced the scope even further and separated the work into
+  several much smaller phases.
 
 The current proposal will attempt to unify the handling of the standards.
 Providing nice ergonomics, as Python-like defaults as possible, and keeping the
@@ -27,15 +29,15 @@ with supporting only ASCII is that other encodings can get corrupted.
 - UTF-16 is 1-2 sets of 16 bits long
 - UTF-32 is 1 set of 32 bits long
 
-### When slicing by unicode codepoint e.g. "🔥🔥🔥" (\U0001f525\U0001f525\U0001f525)
+### When slicing by unicode codepoint e.g. "🔥🔥" (\U0001f525\U0001f525)
 
-- UTF-8: 12 sets (12 bytes) long. The first byte of each fire can be used to
-know the length, and the next bytes are what is known as a continuation byte.
-There are several approaches to achieve the slicing, they can be explored with
-benchmarking later on.
-- UTF-16: 6 sets long. It's very similar in procedure to UTF-8.
+- UTF-8: 8 sets (8 bytes) long. The first byte of each fire can be used to
+  know the length, and the next bytes are what is known as a continuation byte.
+  There are several approaches to achieve the slicing, they can be explored with
+  benchmarking later on.
+- UTF-16: 4 sets long. It's very similar in procedure to UTF-8.
 - UTF-32: It is fastest since it's direct index access. This is not the case
-when supporting graphemes.
+  when supporting graphemes.
 
 ## Graphemes
 
@@ -115,6 +117,7 @@ prefetching, and algorithmic tricks to make processing faster for languages
 which have mostly ASCII characters but still keep full unicode support.
 
 #### Grapheme support
+
 Grapheme support should exist but be opt-in due to their high compute cost.
 
 ### One concrete (tentative) way forward
@@ -128,12 +131,13 @@ their own merit.
 
 ```mojo
 struct Encoding:
-    alias UTF8 = 0
-    alias UTF16 = 1
-    alias UTF32 = 2
-    alias ASCII = 3
-    # maybe in the future we will even implement parsing for http header encoding
-    # alias ISO_8859_1 = 4
+    alias ASCII = 0
+    alias UTF8 = 1
+    # maybe we implement some of these at some point
+    # alias UTF32 = 2
+    # alias UTF16 = 3
+    # alias ISO_8859_1 = 4 # http header encoding
+    # alias UTF8_G = 4 # UTF-8 extension for fast grapheme ops.
 
 struct Indexing:
     alias RAW = 0
@@ -144,10 +148,9 @@ alias ASCIIString = StringSlice[encoding=Encoding.ASCII, indexing=Indexing.RAW]
 
 
 struct StringSlice[
-    encoding: Encoding = Encoding.UTF8,
-    indexing: Indexing = Indexing.RAW,
+    encoding: Encoding = Encoding.UTF8, indexing: Indexing = Indexing.RAW
 ]:
-    ... # data is bitcasted to bigger DTypes when encoding is 16 or 32 bits
+    ... # ptr is bitcasted to bigger DTypes when encoding is 16 or 32 bits
 
 struct String:
     fn as_string_slice(
@@ -156,30 +159,52 @@ struct String:
         ...
 ```
 
+#### Motivation for parameters instead of different types
+
+I am not a particular fan of creating new types because it means:
+- A lot of API docstrings and function duplication
+  - Several basic functions that everyone expects in a pythonic string type will
+    be duplicated for each new type.
+- Developing a whole set of traits to avoid having to allocate when going
+  from one type to the other
+  - Besides being a lot of work to come up with it, developing a set of traits
+    would also involve quite the function signature rewrite for several that
+    work agnostically regardless of encoding or indexing scheme.
+
 #### What this requires
 
-1. Phase 1: Add parameters and constraints.
+1. Phase 1: Add parameters and constraints for `Indexing` only.
+- Add the parameters and constraint on the supported indexing schemes.
+
+2. Phase 2: Add the implementations for each `Indexing` scheme for UTF-8.
+- Add the indexing schemes and fix all functions which assume one particular
+  kind.
+
+3. Phase 3: Add parameters and constraints for `Encoding`.
+- We can start with only UTF-8 (current) and ASCII.
 - Add the parameters and constraint on the supported encodings. This
-will enable progressive implementation for each encoding and indexing scheme
-without breaking any existing code, since our type currently already implements
-UTF-8 encoding and (to my chagrin) raw indexing.
+  will enable progressive implementation for each encoding without breaking any
+  existing code.
 
-2. Phase 2: Add parametric support for other encodings.
-- Rewrite every function signature that uses `StringSlice`s, where the code
-doesn't require them to be the defaults. This will be a lot of work because it
-basically affects every place `StringSlice` is used in a function signature with
-a specified origin or mutability.
+4. Phase 4: Add parsing for other encodings.
+- We will need to add parametrized `StringSlice` explicit constructors that
+  parse `StringSlice`s from another encoding. But the default same-encoding
+  internal methods will remain the same **(as long as the code is encoding and
+  indexing-agnostic)**.
 
-3. Phase 3: Add parsing for other encodings.
-- We will need to add parametrized `StringSlice` constructors that parse
-`StringSlice`s from another encoding. But the default same-encoding internal
-methods will remain the same **(as long as the code is encoding and
-indexing-agnostic)**.
+5. Phase 5: Add utility functions for each encoding.
+- Similarly to `stdlib/collections/string/_utf8.mojo` add functions to implement
+  optimized versions of different functions.
 
-4. Phase 4: Optimize for each encoding.
+6. Phase 6: Optimize for each encoding.
 - Add support for functions like `.isspace()`, `.split()` and `.splitlines()`
-for those encodings.
+  for those encodings.
 
+7. Phase 7: Add parametric support for other encodings.
+- Rewrite every function signature that uses `StringSlice`s, where the code
+  doesn't require them to be the defaults. This will be a lot of work because it
+  basically affects every place `StringSlice` is used in a function signature
+  with a specified origin or mutability.
 
 #### Adapt CodepointIter
 
@@ -229,16 +254,11 @@ ascii_val.uppercase() # incredibly fast due to bitflip trick
 
 ```mojo
 struct StringSlice[
-    encoding: Encoding = Encoding.UTF8,
-    indexing: Indexing = Indexing.RAW,
+    encoding: Encoding = Encoding.UTF8, indexing: Indexing = Indexing.RAW
 ]:
   fn __getitem__(self, slice: Slice) -> Self:
     @parameter
-    if encoding is Encoding.UTF32:
-      return Self(unsafe_from_utf32=rebind[Span[UInt32, origin]](self._slice)[slice])
-    elif encoding is Encoding.UTF16 and indexing is Indexing.RAW:
-      return Self(unsafe_from_utf16=rebind[Span[UInt16, origin]](self._slice)[slice])
-    elif encoding is Encoding.ASCII and indexing is not Indexing.GRAPHEME:
+    if encoding is Encoding.ASCII and indexing is not Indexing.GRAPHEME:
       return Self(unsafe_from_ascii=self._slice[slice])
     elif encoding is Encoding.UTF8 and indexing is Indexing.RAW:
       ... # current implementation
--- a/mojo/proposals/string-ascii-unicode-utf-grapheme.md
+++ b/mojo/proposals/string-ascii-unicode-utf-grapheme.md
@@ -174,37 +174,31 @@ I am not a particular fan of creating new types because it means:
 #### What this requires
 
 1. Phase 1: Add parameters and constraints for `Indexing` only.
-- Add the parameters and constraint on the supported indexing schemes.
-
+  - Add the parameters and constraint on the supported indexing schemes.
 2. Phase 2: Add the implementations for each `Indexing` scheme for UTF-8.
-- Add the indexing schemes and fix all functions which assume one particular
-  kind.
-
+  - Add the indexing schemes and fix all functions which assume one particular
+    kind.
 3. Phase 3: Add parameters and constraints for `Encoding`.
-- We can start with only UTF-8 (current) and ASCII.
-- Add the parameters and constraint on the supported encodings. This
-  will enable progressive implementation for each encoding without breaking any
-  existing code.
-
+  - We can start with only UTF-8 (current) and ASCII.
+  - Add the parameters and constraint on the supported encodings. This
+    will enable progressive implementation for each encoding without breaking
+    any existing code.
 4. Phase 4: Add parsing for other encodings.
-- We will need to add parametrized `StringSlice` explicit constructors that
-  parse `StringSlice`s from another encoding. But the default same-encoding
-  internal methods will remain the same **(as long as the code is encoding and
-  indexing-agnostic)**.
-
+  - We will need to add parametrized `StringSlice` explicit constructors that
+    parse `StringSlice`s from another encoding. But the default same-encoding
+    internal methods will remain the same **(as long as the code is encoding and
+    indexing-agnostic)**.
 5. Phase 5: Add utility functions for each encoding.
-- Similarly to `stdlib/collections/string/_utf8.mojo` add functions to implement
-  optimized versions of different functions.
-
+  - Similarly to `stdlib/collections/string/_utf8.mojo` add functions to
+    implement optimized versions of different functions.
 6. Phase 6: Optimize for each encoding.
-- Add support for functions like `.isspace()`, `.split()` and `.splitlines()`
-  for those encodings.
-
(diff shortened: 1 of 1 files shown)


gryznar:
I like this proposal :)

proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,189 @@
...
+the character in the UTF-8 encoded `StringSlice`. This is much more efficient
+and does not add any complexity into the type system nor developer headspace.
+
+## Now, onto the Proposal
+
+### Goals
+
+#### Hold off on developing Char further and remove it from stdlib.builtin
+
+`Char` is currently expensive to create and use compared to a `StringSlice`
```
gabrieldemarmiesse, line 92: Do we have numbers on this? The creation but also using different methods on it. Let's avoid optimizing things without measuring all the workflows first.

ConnorGray:
Hi Martin, thanks for taking the time to write up a proposal on this topic! 🙂

Apologies for the length of this response—"if I'd had more time, I would have written a shorter letter" and all that😌 

Before responding to your proposal, let me share a bit out where my head is at regarding how we handle these string processing APIs. For some context on my recent work in this area, we had had a discussion internally on the stdlib team and came to a tentative consensus to move forward with the name `Char` / `char_*` for types that operate on Unicode codepoints, with the understanding that we could rename if needed for clarity. It sounds like the current name may not be ideal, and so we're open to renaming it 🙂 

My current thinking is a modification to what @owenhilyard [proposed in Discord here](https://discord.com/channels/1087530497313357884/1224434323193594059/1330191006833971322):

> I'm somewhat torn. For most people, they seem to want to iterate over symbols in text, graphemes. However, that can have a substantial perf cost. `Char` also has potential for confusion with `c_char`, since `c_char` is usually a byte. What if we kept the name of the functions as `.characters()` but make the function return a `Grapheme` iterator, but also provided `.codepoints()` and `.bytes()`? I personally prefer to make the API most people will reach for the most correct in terms of what most people want (iteration over symbols they can see), and then if you have perf issues you can use codepoints or bytes directly.

 More directly, I'm currently thinking should do the following:

* Rename `Char` to `Codepoint` for minimal ambiguity
* Keep `StringSlice.as_bytes() -> Span[Byte]`
* Rename `.chars()` to `StringSlice.codepoints() -> CodepointsIter` (iterator over `Codepoint`)
* Rename `.char_slices()` to `StringSlice.codepoint_slices()` (iterator over single-codepoint `StringSlice` pointers)
* Eventually introduce a grapheme cluster iterator called `.characters()` or `.graphemes()`, along with `Character` (owning) and `CharacterSlice` (view) types.
  - (If we're feeling particularly bold, perhaps we'd call those `Grapheme` and `GraphemeSlice`—but that might be _too_ unfamiliar to folks 🙃.)

Then, eventually, re-add a `StringSlice.__iter__()` that either calls `.codepoint_slices()` or `.graphemes()`, pending discussion about the merits of each of those (performance vs correctness tradeoff, essentially).

---

Now some initial thoughts on your proposal 🙂 

> They [Swift] have a `Character` that I think inspired our current `Char` type, it is a generic representation of a Character in any encoding, that can be from one codepoint up to any grapheme cluster.

Minor clarification: This is correct re. Swift's `Character` containing one or more clustered codepoints, but our `Char` type I added was actually inspired by Rust's `char` type, which is a unsigned 32 bit integer. I hadn't intended for our `Char` to eventually support stored an unbounded amount of codepoints, because that would have meant it would need to allocate, which wasn't optimal for the common cases where iterating over codepoints is sufficient.

To that end, that's partly why I'm in favor of renaming our `Char` type to `Codepoint`. `Char` has the advantage of being "recognizable" as an English-ish word, but I've come around to the opinion that we're better off using API naming that more closely adheres to the underlying model and inherent complexity of UTF-8 and Unicode, instead of trying to mask that complexity by pretending that a "Character" and Unicode scalar value are equivalent.

Using terminology like `Codepoint` might be unfamiliar at first to folks not already knowledgeable about how computers represent and process text. But I think part of Mojo's broader language philosophy is not shying away from and trying to paper over complexity. Instead, I think it's better to both teach and make that complexity feel manageable with APIs that model it well 🙂 

> ## Value vs. Reference
>
> Our current `Char` type uses a u32 as storage, every time an iterator that
yields `Char` is used, an instance is parsed from the internal UTF-8 encoded
`StringSlice` (into UTF-32).
> 
> The default iterator for `String` returns a `StringSlice` which is a view into
> the character in the UTF-8 encoded `StringSlice`. This is much more efficient
> and does not add any complexity into the type system nor developer headspace.

I think both types of iteration (over decoded `u32` codepoints, and over single-codepoint `StringSlice` views) are useful, and supporting both is valuable. I agree that iteration over single-character `StringSlice` is likely the most common for any kind of general-purpose string processing. But for certain specialized use cases, e.g. within parsers, low-level string manipulation algorithms, character encoding converters, etc. being able to compare codepoint values directly can be the most natural way to express the desired logic.

As a simple example, our Unicode case conversion logic maps a single `Char` codepoint to a sequence of replacement `Char` that are the uppercase expression of the original `Char`. That is an algorithm most naturally expressed in terms of single codepoint values, because the Unicode casing tables powering it are based on codepoints, not UTF-8 sequences.

So insofar as your argument is that `Char` should not be used frequently (and even then only be advanced users), I think this is another argument in favor of renaming `Char` to `Codepoint`, to de-emphasize it for use by less knowledgeable users.

* * * 

Regarding the general thrust of the rest of the proposal:

1. I like the idea of having a dedicated `ASCIIString` type that would enable use to perform encoding-specific optimization 😄 
   * I think it would be reasonable to start with a `struct ASCIIString` that is its own dedicated type, and then work later to assess the performance and ergonomics tradeoff in parameterizing `StringSlice` on an `encoding` parameter.
2. I'm less certain that having a parameter for controlling indexing behavior is optimal. 
3. I'm unclear on the semantics of the proposed `iter(data).chars[encoding.UTF8]()` method — could you elaborate that section?
   * My understanding is that the `StringSlice` would have some specific encoding its stored in memory as, so I'm uncertain what "casting" that encoding during iteration would do?

Re. (2): One idea we've been discussed a bit internally (but haven't had time to implement yet) is the idea of using named arguments within `__getitem__` methods, to enable a syntax like:

```mojo
var str = StringSlice("abcef")

var bytes: Span[Byte] = str[byte=2:5]

var cps: StringSlice = str[codepoints=1:3]

var chars: StringSlice = str[chars=1:5] # Grapheme clusters
```
 
I still have some reservations about this, but I like that it strikes a balance between being a relatively terse syntax for indexing, while also staying absolutely clear and readable about what indexing semantics are being used (helping the programmer stay aware of behavior and potentially performance concerns).

I'm unsure though what I think the default should be for `str[1:5]` — codepoints or grapheme clusters, I'm not sure. 

* * *

Apologies again for the long and scattered nature of these comments; I'm erring on the side of leaving any feedback at all, instead of waiting to write something more polished. Thank you again Martin for helping to articulate a possible design direction here 🙂  — like you, I'm excited about the possibilities of using Mojo's 🔥 features to provide flexible and powerful string handling capabilities to our users.

martinvuyk:
Hi @ConnorGray

> Apologies for the length of this response—"if I'd had more time, I would have written a shorter letter" and all that😌

Thanks for writing, I actually like reading through well thought ideas. Your "unordered thoughts" match some of my essays lol

> More directly, I'm currently thinking should do the following:
>   - Rename Char to Codepoint for minimal ambiguity
>   - Keep StringSlice.as_bytes() -> Span[Byte]
>   - Rename .chars() to StringSlice.codepoints() -> CodepointsIter (iterator over Codepoint)

+1 on all 3 of these

>   - Rename .char_slices() to StringSlice.codepoint_slices() (iterator over single-codepoint StringSlice pointers)

This one I'm not so fond of because I like the idea of parametrizing more. I'll expand later

>   - Eventually introduce a grapheme cluster iterator called .characters() or .graphemes(), along with Character (owning) and CharacterSlice (view) types.
>       - (If we're feeling particularly bold, perhaps we'd call those Grapheme and GraphemeSlice—but that might be too unfamiliar to folks 🙃.)

I think we should have the method be `.graphemes()` because IMO `.characters()` is more ambiguous. IMO `Grapheme` and `GraphemeSlice` won't be necessary, when iterating over a `String` or `StringSlice` with `Indexing.GRAPHEME` it would just return a `StringSlice` whose `__len__()` method might return more than 1 when iterating.

> Then, eventually, re-add a `StringSlice.__iter__()` that either calls .codepoint_slices() or .graphemes(), pending discussion about the merits of each of those (performance vs correctness tradeoff, essentially).

I seriously think it is not necessary to deprecate the default iteration over unicode codepoints given it is the IMO sane default  (and practically free using `StringSlice`). It also follows Python's default which IMO should also count. This is also solved by parametrizing `String` and `StringSlice` in a way in which we don't have to make that decision for people.

---

> Using terminology like Codepoint might be unfamiliar at first to folks not already knowledgeable about how computers represent and process text. But I think part of Mojo's broader language philosophy is not shying away from and trying to paper over complexity. Instead, I think it's better to both teach and make that complexity feel manageable with APIs that model it well 🙂

100% Agree on this

> I think both types of iteration (over decoded u32 codepoints, and over single-codepoint StringSlice views) are useful, and supporting both is valuable. 

Also on this, but I just think that the default should be what is most often used. I also think it's quite straightforward to do:
```mojo
a = "123"
for item in a:
    c = Codepoint(item)
    # or for what I've read them being used for very often
    # in our unicode casing code and golang's stdlib
    # some people call them "runes"
    utf32_value, utf8_length = Codepoint.parse_utf32_utf8_length(item)
    ...
```

---

>    - I think it would be reasonable to start with a struct ASCIIString that is its own dedicated type, and then work later to assess the performance and ergonomics tradeoff in parameterizing StringSlice on an encoding parameter.

IMHO this will be much more work than parametrizing and adding `constrained` on every branch that is not utf8 and progressively adding optimizations. Because it means duplicating every API and docstrings since we don't have good inheritance-like mechanisms yet.

>   2. I'm less certain that having a parameter for controlling indexing behavior is optimal.

Quite the opposite, the optimization trickles down to all users and libraries which interact with the generic `String`. The indexing is very easy to change with a rebind or if we add some APIs to go back and forth. `ascii(String("123"))` would return a rebound string which is free. Any function which uses any sort of string manipulation and whose signature accepts a generic `String` would benefit from the perf. gains of e.g. changing to an `ASCIIString` for your specific use case since you know only those sequences will be there. Just as the performance hit of going full grapheme support will only happen where needed (**determined by the end user: the programmer, not us**).

>    3. I'm unclear on the semantics of the proposed `iter(data).chars[encoding.UTF8]()` method — could you elaborate that section?

That section was to propose in the case that we wanted a `Char` type which can cave any one of the 4 encodings underneath*. In the case in which you wanted to iterate over `Char` of a different encoding than the current string for which `iter(some_str)` was called, then you could pass the encoding parameter for the iterator. All 4 encodings can be packed inside a `UInt32` and as such have the same fix-sized stack-allocated variable. And IMO this still makes sense because a character in ascii is just 1 byte, in utf8 1-4, utf16 1-2, utf32 1. The underlying methods could be parametrized for each encoding and constrained where the methods don't make sense for certain cases.

*: This wouldn't include when `Indexing.Grapheme`, but this might actually be more of an argument in favor of renaming the type to `Codepoint` as you proposed.

>   - My understanding is that the StringSlice would have some specific encoding its stored in memory as, so I'm uncertain what "casting" that encoding during iteration would do?

It's just a matter of bitcasting the underlying bytes to the proper encoding datatype. For example, the data for a utf16 is actually stored in a `List[UInt16]` but the `String` type uses a `List[UInt8]` (and a `StringSlice` points to it), so the only necessary step when doing anything is bitcasting the internal pointer e.g. `utf16_buffer = rebind[Span[UInt16, __origin_of(some_str)]](some_str.as_bytes())`. But both `String` and `StringSlice` remain untouched, except inside the body of the methods where the casting needs to happen.

---

> Apologies again for the long and scattered nature of these comments; I'm erring on the side of leaving any feedback at all, instead of waiting to write something more polished. Thank you again Martin for helping to articulate a possible design direction here 🙂 — like you, I'm excited about the possibilities of using Mojo's 🔥 features to provide flexible and powerful string handling capabilities to our users.

Apologies again for being intense sometimes hehe. String handling is what I still love about Python, and I want to make it even better in Mojo :fire:

jackos:
thanks @martinvuyk I've added to next design discussion meeting

leb-kuchen:
The thing is how much memory are you willing to use to get O(n) indexing.
The most efficient solution is probably to store a subset of breakpoints.
However segmentation is not as predictable as encoding, so the efficiency will vary. 
For ASCII I think, it is a good idea, Rust is experimenting with ASCII Char, just because of how painful UTF-8 conversion are. 
But I don't think a separate Grapheme type is a good idea. This would be essentially String.graphemes().next(), i.e. are wrapper around String or StringSlice.

owenhilyard:
Overall, I agree with @ConnorGray. I think that Mojo benefits by being correct by default, and then having the ability to drop down to get more performance after you read about the hazards. In this case, that means defaulting to Graphemes, and giving codepoint and byte access options. People dealing with large bodies of text will need to understand unicode and what possibilities exist for their problem, or they can just throw CPU at the problem. When people think of `Character`, 90% of developers mean Graphemes, so `Char` should be a buffer slice (or just a pointer if we're fine re-validating it on use).

> (If we're feeling particularly bold, perhaps we'd call those Grapheme and GraphemeSlice—but that might be too unfamiliar to folks 🙃.)

We can make an alias, so that people who reach for `CharacterSlice` get the `GraphemeSlice` type back and can either ignore that or look a bit further. 

> I like the idea of having a dedicated ASCIIString type that would enable use to perform encoding-specific optimization 😄 

The reason I think that we may want to be parameterized over encoding is because not only do you want Ascii, but also UTF-16 (JS interop with WASM, Java), but also UCS-2 (Windows). It means the stdlib-side string handling code is going to be gross, but I would rather go for a solution we can adapt than one where we have "oops, we can't talk to that". This is especially true since I've seen some RAG DB research which wants to use UTF-32 as the input encoding since it's easier to only need to do "codepoints -> graphemes" on a GPU. 

> That section was to propose in the case that we wanted a Char type which can cave any one of the 4 encodings underneath*. In the case in which you wanted to iterate over Char of a different encoding than the current string for which iter(some_str) was called, then you could pass the encoding parameter for the iterator. All 4 encodings can be packed inside a UInt32 and as such have the same fix-sized stack-allocated variable. And IMO this still makes sense because a character in ascii is just 1 byte, in utf8 1-4, utf16 1-2, utf32 1. The underlying methods could be parametrized for each encoding and constrained where the methods don't make sense for certain cases.

@martinvuyk these "casts" are non-trivial amounts of compute. There are only a few of these "casts" that are cheap, like ASCII -> UTF-8. There are also issues with characters which are impossible to represent in ASCII. I think having the buffer parameterized and converting the whole thing when you want to is a better idea, since I think most code will take input, make everything a single encoding for internal processing, and then potentially convert back to a desired output encoding. These are also potentially fairly large copies to compute, since UTF-16 and UTF-32 usually take more memory than the equivalent UTF-8. I think this makes sense because the encoding is a property of the data in the buffer, not the function called on the data in the buffer.  

@leb-kuchen 
> The thing is how much memory are you willing to use to get O(n) indexing. The most efficient solution is probably to store a subset of breakpoints. However segmentation is not as predictable as encoding, so the efficiency will vary. For ASCII I think, it is a good idea, Rust is experimenting with ASCII Char, just because of how painful UTF-8 conversion are. But I don't think a separate Grapheme type is a good idea. This would be essentially String.graphemes().next(), i.e. a wrapper around String or StringSlice.

`O(n)` indexing is doable by parsing while we go. If we want better indexing, that's where indexes come into play.

rcghpge:
I think with low-level programming for Mojo you want closer to the metal than even llvm and mlir. What would the concept of passes in this framework look like. Since I was first introduced to Mojo, the impressions I get are you want orders of magnitude in performance.

owenhilyard:
> > `O(n)` indexing is doable by parsing while we go. If we want better indexing, that's where indexes come into play.
> 
> I think it is either O(n) memory or O(n) time. In my opinion it is not worth it and iteration should be faster in 90% of the cases. It would also limit the extensibility of string if this API was introduced.
> 
> I think of the following API
> 
>     * graphemes
> 
>     * is_grapheme_boundary
> 
>     * ceil_grapheme_boundary
> 
>     * floor_grapheme_boundary
> 
> 
> Graphemes are designed for segmentation , and not really for indexing. Is still possible to possible to design an API this way, but it is not where graphemes shine.

My thought is that something like this will happen:

If the iteration isn't in the critical path, then nobody will notice and it will silently do the thing which is correct 99% of the time, which is iterating over visible symbols. Very few people not well versed in unicode expect "👍🏻"  and "👍" result in different numbers of loop iterations.

If it is in the critical path, then it shows up in profiling, and someone googles "Mojo string iteration slow", and lands on a docs page or blog post about how we decided to do graphemes by default, and which tells the user to make use of `.codepoints()` or `.bytes` if they want faster iteration in exchange for having to deal with some parts of unicode themselves.

martinvuyk:
> Overall, I agree with @ConnorGray. I think that Mojo benefits by being correct by default, and then having the ability to drop down to get more performance after you read about the hazards. In this case, that means defaulting to Graphemes, and giving codepoint and byte access options. People dealing with large bodies of text will need to understand unicode and what possibilities exist for their problem, or they can just throw CPU at the problem.

IMO you are underestimating how much code in the wild is a series of small/medium/large text processing, and how little developers care about performance once a project works. (I think) The impact of these defaults will be big not only for large text.

> Mojo benefits by being correct by default

Sidenote on this: graphemes are an extension of the unicode standard, using only codepoints is also correct. And code which is migrated from Python that does arithmetic based on codepoints might have issues with Mojo having different defaults.

> > That section was to propose in the case that we wanted a Char type which can cave any one of the 4 encodings underneath*. In the case in which you wanted to iterate over Char of a different encoding than the current string for which iter(some_str) was called, then you could pass the encoding parameter for the iterator. All 4 encodings can be packed inside a UInt32 and as such have the same fix-sized stack-allocated variable. And IMO this still makes sense because a character in ascii is just 1 byte, in utf8 1-4, utf16 1-2, utf32 1. The underlying methods could be parametrized for each encoding and constrained where the methods don't make sense for certain cases.
> 
> @martinvuyk these "casts" are non-trivial amounts of compute. There are only a few of these "casts" that are cheap, like ASCII -> UTF-8. There are also issues with characters which are impossible to represent in ASCII. I think having the buffer parameterized and converting the whole thing when you want to is a better idea, since I think most code will take input, make everything a single encoding for internal processing, and then potentially convert back to a desired output encoding. These are also potentially fairly large copies to compute, since UTF-16 and UTF-32 usually take more memory than the equivalent UTF-8. I think this makes sense because the encoding is a property of the data in the buffer, not the function called on the data in the buffer.

@owenhilyard I don't mean converting from UTF-8 to these encodings, I mean casting the buffer to e.g. a `Span[UInt16]`, doing your editing, and voila. It's just using a buffer of bytes as all the other encoding storage, not only utf8. It's a way for us to not have to touch `String` or `StringSlice` 's internal data storage format (`List[UInt8]`, `Span[UInt8]`)

---

I'm willing to change my mind on the defaults if we can have some benchmarks which show a relatively low impact for common use-cases. My main focus with this proposal is to have `String` parametrized so that we can have some proper utf16 and utf32 encoding support and not a half-hazard set of APIs, while also allowing us to bring back some cool ASCII and other encoding-specific optimizations.

proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,189 @@
+# String, ASCII, Unicode, UTF, Graphemes
+
+The current proposal will attempt to unify the handling of the standards.
+Providing nice ergonomics, as Python-like defaults as possible, and keeping the
+door open for optimizations.
+
+## String
+
+String is currently an owning type which encodes string data in UTF-8, which is
+a variable length encoding format. Data can be 1 byte up to 4 bytes long. ASCII
```
lattner: What is "Data"? Graphemes can be much larger

proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,189 @@
...
+text, which is where English text falls, is 1 byte long. As such, the defaults
+should optimize for that character set given it is most typical on the internet
```
lattner: Agree that ascii is very common :)

proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Graphemes
+
+Graphemes are an extension which allows unicode codepoints to modify the end
+result through concatenation with other codepoints. Visually they are shown as a
+single unit e.g. é is actually comprised of `e` followed by `´`  (\x65\u0301).
+
+Graphemes are more expensive to slice because one can't use faster algorithms,
+only skipping each unicode codepoint one at a time and checking the last byte
+to see if it's an extended cluster *.
```
lattner: agreed; but graphemes are what most humans care about

proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,189 @@
...
+C, C++, and Rust use UTF-8 with no default support for graphemes.
+
+### Swift
+
+Swift is an interesting case study. For compatibility with Objective-C they went
+for UTF-16. They decided to support grapheme clusters by default. They recently
+changed the preferred encoding from UTF-16 to UTF-8. They have a `Character`
+that I think inspired our current `Char` type, it is a generic representation of
+a Character in any encoding, that can be from one codepoint up to any grapheme
+cluster.
```
lattner: This is confusing a couple things.  utf8 is just a "strictly better than utf16" way of storing the data, but isn't something that clients of the apis should care about.  Swift/Objc were held back by legacy java-era decisions around utf16 which were just legacy and wrong, mojo isn't beholden to that.

IMO. we should be utf8 storage, and grapheme-by-default. Character should be a grapheme.

proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Python
+
+Python currently uses UTF-32 for its string type. So the slicing and indexing
+is simple and fast (but consumes a lot of memory, they have some tricks to
+reduce it). They do not support graphemes by default. Pypi is implementing a
+UTF-8 version of Python strings, which keeps the length state every x
+characters.
```
lattner: wow, yuck, I guess that is far better than python 2

proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,189 @@
...
+### Goals
+
+#### Hold off on developing Char further and remove it from stdlib.builtin
+
+`Char` is currently expensive to create and use compared to a `StringSlice`
+which is a view over the original data. There is also the problem that it forces
+UTF-32 on the data, and those transformations are expensive.
+
+We can revisit `Char` later on making it take encoding into account. But the
+current state of the type makes it add more complexity than strictly necessary.
```
lattner: Char should be like String and SSO.

YkdWaWEzVmphR1Z1: GraphemesIterator.next() would have to do memcpy and the inlined ascii path would contain a function call which is difficult to optimize if Char were a wrapper around String. If CharSlice was introduced, it would make it even more confusing. I guess Char would have convenience methods and statically prove it contains a grapheme cluster. Python uses strings for indexing, so I don't really a downside to using StringSlices.

proposals/string-ascii-unicode-utf-grapheme.md
martinvuyk: Yeah I should've written unicode codepoint which is the base unit of "data" in this case

proposals/string-ascii-unicode-utf-grapheme.md
martinvuyk: Not in every case, when designing a database index or something where the size of the string keys need to be known or at least limited up to X bytes one does care about unicode codepoints or even just bytes (when using pure ASCII). There is also doing a `StringSlice` into an http header for example, where the encoding is ISO-8859-1. IMO we should be able to offer a zero-copy way to read them

The thing with Mojo is that our user-base will be from systems people all the way up to hopefully even web development where graphemes are what is most natural to think about. My main issue with going graphemes by default is that it will force the cost down to people who will not want it. We could keep `String` with such defaults, but maybe we should let `StringSlice` be much more generic over string data.

proposals/string-ascii-unicode-utf-grapheme.md
martinvuyk: > utf8 is just a "strictly better than utf16" way of storing the data

I disagree here. Any language that is normally 2 bytes in utf-8 will be faster in utf-16 (in exchange for more memory/storage). It is probably not very common, but I would like to leave the door open for people to go crazy on optimizations for their use-case. But I also don't have UTF-16 itself very much in mind, UTF-32 and ASCII (and in the future things like ISO-8859-1) optimizations are much more important to me.

> IMO. we should be utf8 storage, and grapheme-by-default. Character should be a grapheme.

I agree that Character should be a grapheme and on using utf-8 by default. But on grapheme indexing by default not so much. I could just relent on that detail though; my main concern is allowing multiple encodings and indexing schemes

proposals/string-ascii-unicode-utf-grapheme.md
martinvuyk: > Char should be like String and SSO.

Makes sense, I imagine that most grapheme clusters are less than 24 bytes long. But for default iteration I would still argue for just using `StringSlice` of the same encoding and indexing scheme (codepoint or grapheme) because there would be no need to build an instance and the pointer indirection cost is amortized because the data is in the cache line during the iteration

YkdWaWEzVmphR1Z1: Many multi-person emojis contain >= 24 bytes, so it is not like that malloc will never called. 
The common case, like map(String) can still be optimized in the back end. Calling mmap and making a syscall can take thousands of cycles, so this should also be considered.

proposals/string-ascii-unicode-utf-grapheme.md
YkdWaWEzVmphR1Z1: I think if were to do graphemes by default, we would need to extend UTF-8. My favorite idea is to follow the x86 approach and to introduce a prefix. 5 to 8 leading ones are all available, e.g.  11111110(0xfe). 
It would indicate, that the next codepoint does not break. 
Thus no table lookup would be required for indexing and iteration. 
However this would make appending more expensive, so I think there should be lazy methods like append_lazy which set a lazy bit.
Also writing to a file would more expensive, since all the prefixes needed to be removed. For that another prefix bit would be introduced. Also "\r\n" would be especially excluded because of how common it is.

mojo/proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,247 @@
...
+for UTF-16. They decided to support grapheme clusters by default. They recently
+changed the preferred encoding from UTF-16 to UTF-8. They have a `Character`
+type which is a generic representation of a Character in any encoding, that can
+be from one codepoint up to any grapheme cluster.
+
+### Python
+
+Python currently uses UTF-32 for its string type. So the slicing and indexing
+is simple and fast (but consumes a lot of memory, they have some tricks to
+reduce it). They do not support graphemes by default. Pypi is implementing a
```
gabrieldemarmiesse: I think you mean pypy, not pypi. Ref: https://pypy.org/posts/2019/03/pypy-v71-released-now-uses-utf-8-451324088028792912.html

martinvuyk: hehe nice catch tks!

proposals/string-ascii-unicode-utf-grapheme.md
martinvuyk: That is an interesting Idea but probably best left for the `String` implementation and potentially a separate new type that adds those optimizations. I reduced the new scope of this proposal to only be about `StringSlice` itself, which is not able to have that kind of mutability. We could eventually implement support for that kind of encoding you mention but it would be an extension of UTF-8 as you pointed out (and thus a different `Encoding`).

YkdWaWEzVmphR1Z1: StringSlice needs to have this, because otherwise slicing is too expensive.
I see no reason why StringSlice can't have this.
You normally don't use mutable StringSlices.
We could call it UTF8-G.
Though I think it is more a internal representation, this representation is not stable across unicode versions. I guess the version code optionally be decoded in the string. Also this representation works with existing utf8 and does not require parameters although it could be done.

martinvuyk: That all constitutes a new type of encoding and an encoding-specific optimization, and could be possibly implemented using the approach of this proposal (adding a new "`Encoding.UTF8_G`" and the optimized implementations). In any case, while it is a possibility, it's a bit out of scope of the discussion.

YkdWaWEzVmphR1Z1: Why is it out of the discussion. What other ways are there to implement grapheme indexing. It is about feasibility.

YkdWaWEzVmphR1Z1:
In my opinion this proposal has many great ideas, but I still remain skeptical towards parameterized encoding. 
I will make a counter proposal which involves a custom encoding ( UTF8-G) and a bit-packed structure at the front.
This should enable fast indexing for graphemes while minimizing overhead for ASCII and other graphemes of length 1.

proposals/string-ascii-unicode-utf-grapheme.md
owenhilyard: I don't think that a new encoding is a good idea, especially for something as widespread as UTF-8. I'm in agreement with @martinvuyk that Character should be a grapheme, since when most people think of characters that's what they mean. We can have it be `alias Character = Grapheme` somewhere in the string handling so anyone who bothers to look a little deeper knows what we did. 

I personally lean towards parameterizing string on the encoding, so that people can have `String[Ascii]` or `String[Latin1]` if they really need indexing speed and know they only have to deal with `sizeof(Character) == 1`. Ideally, I'd also like a way to make any lookup tables or indexes go away since they may not be used in all cases (for example, in one of my projects I only hash strings, I never inspect individual characters).

YkdWaWEzVmphR1Z1: I am generally not a fan of parameterizing on the encoding and especially indexing, since the encodings in the proposal don't handle graphemes efficiently. I think traits and default methods are a much more elegant solution. Parameterizing also comes with a high cost, especially in a low language where people optimize for the encoding.
UTF-8 has won, so I think extending it is the most efficient solution. 
Increasing the memory size by 3 to 4 times to offset a pointer is also not a great solution.
I will cover more regarding graphemes in my proposal.

martinvuyk: I strongly disagree in not using "standard" UTF-8 as the default. If one deals with C, C++, Rust, Python, and other ABIs, then it must be UTF-8 not any extension. That means if we force the extension encoding on every string all ABI compatibility is thrown out the window or we pay the price of the parsing on every interaction. I'm not saying the encoding can't be useful or implemented as a special optimization at some point, just the world uses UTF-8 by default and so should Mojo.

>  Increasing the memory size by 3 to 4 times to offset a pointer is also not a great solution.

At least what I understood about what you described you are talking about adding a header to every codepoint which is potentially 2x size.

> I am generally not a fan of parameterizing on the encoding and especially indexing, since the encodings in the proposal don't handle graphemes efficiently.

That is literally a subset of the proposal. If we can parametrize over encoding and indexing you decide what price to pay and where. If an encoding does not support an indexing type, then it's just a matter of adding it to the constraints of the type's implementation.

>  I think traits and default methods are a much more elegant solution. 

Imagine this:
```mojo
trait Splittable[SubType: AnyType]:
    fn split(self: SubType) -> List[SubType]: ...

# One can't instantiate nor split without grapheme indexing
impl Splittable for StringSlice[Encoding.UTF8_G, Indexing.GRAPHEME]: ...
```
This is completely invented Mojo syntax, but I wouldn't be surprised that it can be possible in the future.

In any case, your "counter proposal" is not countering this one. We are talking about different things. Unless you plan to propose a way to unify different `StringSlice` like types with one trait system, then we aren't really arguing over the same points.

YkdWaWEzVmphR1Z1: UTF8-G is equivalent to UTF-8, unless the string contains a grapheme cluster.
If so, a grapheme cluster contains a G prefix if the prefix bit is set.
"a🏴󠁧󠁢󠁥󠁮󠁧󠁿bc"
| UTF-8    | UTF8-G  |
| -------- | ------- |
| "\\u{61}\\u{1f3f4}\\u{e0067}\\u{e0062}\\u{e0065}\\u{e006e}\\u{e0067}\\u{e007f}\\u{62}\\u{63}"  |   "\\u{61}\\G{7}\\u{1f3f4}\\u{e0067}\\u{e0062}\\u{e0065}\\u{e006e}\\u{e0067}\\u{e007f}\\u{62}\\u{63}"|
|  "\\u{61}\\u{62}\\u{63}"  |  "\\u{61}\\u{62}\\u{63}" |

mzaks: Interesting, how is G{7} represented in the byte stream?

YkdWaWEzVmphR1Z1: "\\xfe\\x87"

mzaks: That would be an invalid sequence though.

YkdWaWEzVmphR1Z1: That's the point, an UTF8-G string which contains the G prefix is invalid UTF-8, but valid UTF8-G.

mzaks: So it's extend the buffer on allocation, allows grapheme based iteration, but no random grapheme indexing and needs cleanup on export. 

Compared to jump points every X graphemes as separate buffer same amount of work on allocation, the size of the buffer will be a bit bigger with jump points. Iteration will be a bit slower with jump points, but they allow random indexing and export does need a cleanup. 

So it depends if your use case is very iteration heavy, then UTF8-G can be a valuable solution, otherwise I am not convinced.

mzaks: Also with UTF8-G the buffer size ist harder to estimate where with jump points you can always have an upper bound estimate by computing (byte_count // jump_point_gap) * index_type_width, where jump_point_gap can be 32 or 64. This way you can have a single allocation data structure. 

With UTF8-G you could have a very rough estimate of (byte_count // min_grapheme_cluster_size) * 2, where the value for min grapheme cluster size can be as small as 3 I think. 

If you don't gracefully allocate memory on init you might end up with full memcopy if the allocation is not enough which can make instance initialization expensive.

YkdWaWEzVmphR1Z1: My idea is to add `Lookup Elimination`.
```mojo
fn append_codepoint(mut self, c: CodePoint):
	@paramter
	if lookup_eliminated:
		self.append_codepoint_raw(c)
	else:
		# existing logic
		

contract LookupEliminiation(String, StringSlice):
	type = optimization
	decorator = lookup_eliminated
	requires not used( __getitem__[::Int] )
	optional requires not used( __len__ )
	optional requires not used( StringIter.__next__ )
	requires not used ( bytes[G-Prefix=True] )
	...
	
	
struct String[Lazy=False]
alias LazyString = String[Lazy=True]

extension String:
	LazyParam = Self.Lazy
	fn __iadd__
	fn __getitem__
	...
```

 > I'd also like a way to make any lookup tables or indexes go away since they may not be used in all cases (for example, in one of my projects I only hash strings, I never inspect individual characters).

Hashing qualifies for `Lookup Elimination`, `G-Prefixes` are not hashed and ideally another invalid byte is used for unique and not the length. For cases where `Lookup Elimination` doesn't apply lazy methods are provided.

martinvuyk: I think this merits its own separate discussion. It is a new encoding, and requires a rewrite of the `String` type. My main focus with this proposal is to allow several use-cases to implement specific algorithms using a parametrized `StringSlice`.

Iteration might be faster with UTF8-G, but for other use-cases interop with C APIs is the most important thing; point being, IMO we should have a `StringSlice` which can generically reference several different types of encodings and indexing schemes for every use-case to be able to have a common set of `StringSlice` APIs.

Yesterday I was reading some Python changelogs and came across an encoding for the first time: Z85. Python 3.13 added support for [`base64.z85encode()`](https://docs.python.org/3/library/base64.html#base64.z85encode) and decode. There are many such interesting encodings that might make more sense for certain use-cases (just like UTF8-G), and my whole point with this proposal is to let them be used with `StringSlice` natively (instead of parsing back and forth to the default string representation (currently UTF-8)).

YkdWaWEzVmphR1Z1: The String representation would look something like this.
```mermaid
   flowchart TD
        A(["String Representation"])
        A --> ptr
        A --> len
        A --> cap
        G{"G-State"}
        GH["G-Heap"]
        GI["G-Inline"]
        U["UTF8-G"]
        GIndex["G-Index"]
        ptr --> Heap{"Heap"}
        cap --> GI
        Heap --> GH
        Heap --> GIndex
        Heap --> U
        G --> G-Prefix
        G --> G-Lazy
        len --> GO["G-Offset"]
        GO --> GIndex
        GH --> G
        GI --> G
```
UTF8-G also supports slicing which other approaches like only storing indexes or pointers do not support. Storing indexes is also more work when pushing and popping which are the most common operations. UTF8-G also support deferring the storing of indexes. For example `G-Index` could be updated during reserve. I don't think time complexity is really relevant. In practice the average amount of cycles and cache usage or eviction is more relevant.

YkdWaWEzVmphR1Z1: > Compared to jump points every X graphemes as separate buffer same amount
>  of work on allocation, the size of the buffer will be a bit bigger with
>  jump points. Iteration will be a bit slower with jump points, but they
> allow random indexing and export does need a cleanup.



# UTF8-G+

## Why UTF-8 is inefficient.

Unicode encoding is incredibly economical with code points. It's like they fear running out of code points. Integers are the most efficient way to encode code points. Instead unicode opted for lookup tables.

I will illustrate this using emojis, because they are the easiest to understand.

1F1E6(A) to 1F1FF(Z) are the number codes for flag emoji. A flag consist of 2 number codes, i.e. 2 code points. This is inefficient from a technical standpoint. At least one bit for even or odd is required. Segmenting backwards requires potentially iterating the entire string. This is overhead that has even to be payed if not using flag emojis. A better approach is to allocate 676(26 \* 26) code points for all possible combinations of country codes. This is what `UTF8-G+` does.

## How UTF8-G+ and UTF32-G+ work.

[RGI Emoji Set](https://unicode.org/Public/emoji/latest/emoji-test.txt)

[Extended Pictographic](https://www.unicode.org/Public/16.0.0/ucd/emoji/emoji-data.txt)

Emoji are not edited by code point and are defined. Nevertheless unicode uses multiple code points for emoji. UTF8-G+ aims to fix this mistake.

UTF8-G+ is a superset of UTF8-G. Since UTF8-G is a superset of UTF-8, UTF8-G+ is also a superset of UTF-8. UTF8-G+ has the same byte length as UTF-8, thus  these encodings can be converted without memmoves. UTF8-G+ uses 5(11111xxx) and 6(111111xx) bytes sequences to represent 32 bits, more than 4 Billion. If a grapheme's UTF-8 length exceeds its UTF8-G+ length, padding bytes (255) can be inserted. Code points are determined through the analysis of graphemes used in practice, for example RGI emoji sequences or occurrences of graphemes in Indian text .This is essentially what Unicode should be doing.

martinvuyk: Please take this to a separate discussion. This proposal doesn't have anything to do with the efficiency or what encoding we should use. It is only about allowing the implementation of different encoding and indexing optimizations for `StringSlice` by parametrizing it.

YkdWaWEzVmphR1Z1: I try to find a technical solution for a difficult issue, because unicode has messed up. 
You like UTF-16, because it takes 33% less space for 0x800 to 0x10000. However UTF-16 also takes 50% more space for 0 to 0x80. Also I don't like UTF-16, because it can't be extended. Your proposal makes my proposal harder to implement. I address arguments made in this discussion. Therefore I don't understand your sentiment.

martinvuyk: I am trying to propose a parametrized version of `StringSlice` that allows people to choose the encoding and indexing mechanism that is best for their use-case. It has nothing to do with what encoding to use beyond saying what defaults make sense to me, and even that may change. What you are proposing, which is implementing a completely new encoding, is orthogonal to this proposal.

mojo/proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,247 @@
...
+is simple and fast (but consumes a lot of memory, they have some tricks to
+reduce it). They do not support graphemes by default. Pypy is implementing a
+UTF-8 version of Python strings, which keeps the length state every x
+characters.
+
+### Mojo
+
+Mojo aims to be close to Python yet be faster and customizable, taking advantage
+of heterogeneous hardware and modern type system features.
+
```
YkdWaWEzVmphR1Z1, line 80: The problem with code point indexing is that it depends on the implementation details of unicode .
`String[::-1]` for example is a useless feature for code points, because it corrupts the input. Flag emojis and `\r\n` are two simple examples. I don't think mojo should support useless features from python, but improving upon it is a great solution in my opinion. I really hate how unicode handles flag emoji, because these are the only used graphemes that require traversing the entire string or keeping state. Limiting the length by ignoring graphemes >= 31 code points won't work here.
The question is, do we hide the complexity and just iterate or try to implement graphemes natively and efficiently.

martinvuyk: > `String[::-1]` for example is a useless feature for code points

I disagree, if you know your string doesn't contain graphemes then it is more efficient. There are some use-cases that still find it useful. Reversing a string is even simpler and faster (can use SIMD) for ASCII text, there are use-cases that need it and will be thankful for perf improvements.

>  The question is, do we hide the complexity and just iterate or try to implement graphemes natively and efficiently.

Again, that is part of a separate discussion of **how**  we implement types like `String` and what encoding and indexing scheme should be the default


My main goal with this proposal is to allow people to choose what features they want and what price they are willing to pay with `StringSlice`, which would handle the encoding-specific implementations of the methods that `String` uses whatever the defaults (we could also parametrize `String` in the future with defaults but that brings a lot of problems)

proposals/string-ascii-unicode-utf-grapheme.md
YkdWaWEzVmphR1Z1: With `UTF8-G+` Char could be optimized to a word, a pointer to the heap for unlikely cases, an 32 bit integer to store an extended code point and null for the discriminant.

proposals/string-ascii-unicode-utf-grapheme.md
owenhilyard: I have to agree with @martinvuyk, parameterizing over encoding seems like the best solution to the problem @leb-kuchen is talking about, but what the list of encodings is should be left to another discussion. We can also have an open set of encodings and a "TextEncoding" trait (or trait family), although that might require a lot of "flags" inside of the trait (or trait family) to get the performance of a closed set back.

mojo/proposals/string-ascii-unicode-utf-grapheme.md
YkdWaWEzVmphR1Z1: > I disagree, if you know your string doesn't contain graphemes then it is more efficient. There are some use-cases that still find it useful. Reversing a string is even simpler and faster (can use SIMD) for ASCII text, there are use-cases that need it and will be thankful for perf improvements.

If the `G-Prefix` is not set,  i.e. the string does not contains multi-codepoint grapheme clusters, then more efficient operations can be used. If you really want this operation, it should be called something like `reverse_codepoints` and not `String[::-1]`. 

`UTF-8` is simply not an efficient encoding for grapheme operations. 
Random access is difficult, who would have thought (https://www.unicode.org/reports/tr29/#Random_Access). There are 3 rules for which 2 codepoints don't provide enough context. 
2 of them are completely unnecessary, of which 1 is completely stupid. 
First try understanding the rules. 
🏴󠁧󠁢󠁥󠁮󠁧󠁿 🇬🇧
England is a region, Great Britain is a country. Of course these have different encodings. 
On of these uses extend rules, because regions don't deserve a rule.
Country flags require context of all preceding characters. 
Why didn't unicode use an existing rule utilizing extend character, it looks so good on paper. 
Why didn't unicode add all recognized countries, decisions are hard. 
Why didn't unicode add all possible countries , there are not enough codepoints. 

I recommend implementing graphemes, I am curious if it changes your opinion. 
If not, why not learn a foreign script (https://www.learnsanskrit.org/guide/devanagari/how-devanagari-works/).

For the the implementation you need:
https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakProperty.txt
https://www.unicode.org/Public/16.0.0/ucd/auxiliary/GraphemeBreakTest.html
https://www.unicode.org/Public/16.0.0/ucd/auxiliary/GraphemeBreakTest.txt
https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt
https://www.unicode.org/Public/16.0.0/ucd/emoji/emoji-data.txt

For optimizing, you probably also need this:
https://www.unicode.org/reports/tr44/#Grapheme_Extend
https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Break_Property_Values
https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedGeneralCategory.txt

proposals/string-ascii-unicode-utf-grapheme.md
YkdWaWEzVmphR1Z1: I think `UTF8-G+` should be the only encoding. Abstraction over other encoding only adds complexity.
While there are minor benefits for special encodings, the tradeoff does not seem worth it.

mojo/proposals/string-ascii-unicode-utf-grapheme.md
```diff
@@ -0,0 +1,261 @@
...
+Swift is an interesting case study. For compatibility with Objective-C they went
+for UTF-16. They decided to support grapheme clusters by default. They recently
+changed the preferred encoding from UTF-16 to UTF-8. They have a `Character`
+type which is a generic representation of a Character in any encoding, that can
+be from one codepoint up to any grapheme cluster.
+
+### Python
+
+Python currently uses UTF-32 for its string type. So the slicing and indexing
```
barcharcraz, line 70: I don't think this is true anymore. They use the format outlined in PEP393, which can be UTF-8, UTF-16, or UTF-32 or some combination (new representations are created as-needed as copies). Also I think it has some way of preserving bogus codepoints.

Python strings are immutable, which makes the PEP393 approach easier.

martinvuyk: It still kind of is UTF-32 (unicode-ish), just with tricks to reduce it to latin-1 or BMP when bigger codepoints aren't present. That is why I phrased it like that. But the indexing and slicing still just follows unicode codepoint logic, not utf-8 nor utf-16.

> UTF-8, UTF-16, or UTF-32 or some combination

AFAIK that is not the case, they are just shorter ranges of the unicode character list and use `UInt8`, `UInt16`, or `UInt32` to store the unicode codepoint values and not encoded ones.

> Python strings are immutable, which makes the PEP393 approach easier.

100% having in-place mutability makes everything more complicated

barcharcraz:
I'm going to be writing up an actual proposal for handling String going into 1.0. Here are my thoughts so far

I'm fairly against making the default iteration order be grapheme clusters, or making the "character" type grapheme clusters for the following reasons (among others):

* If we make "String" be fully based on grapheme clusters (copying swift) then we will need to introduce a new type to deal with filesystem paths. Technically even just enforcing utf-8 would require a new type, but doing things like normalized comparisons will make the issue way worse on all non-apple platforms. Swift uses String for filesystem things, I think they get away with it because all the MacOS filesystem APIs are normalization insensitive.
* grapheme clusters are NOT "user perceived characters" or what the "user expects to see" In particular:
   * they are not appropriate places to break text, while it's probably true that grapheme clusters shouldn't be split by a line break it's not true that grapheme cluster boundaries are always appropriate places to put a break.
   * They aren't appropriate to decide what to delete when you press "backspace" because there are breaks after some combining characters that are written/edited as a unit in certain scripts.
   * they are not appropriate for counting the number of "characters" in a string, because different users have different ideas about what they want counted in different languages/scripts/situations/applications. The clusterization rules can also change from unicode-version to unicode-version so using them for "string length" means your strings will change length with updates to the UCD.
   * they aren't appropriate for moving the cursor about, since often you do, actually, want to edit "between" combining marks (or even expand precomposed characters during cursor movement.
   * see L2/23-140, L2/23-141
* grapheme clusters are only useful when you want a _rough_ estimation of "character" that doesn't need to know about the user's language/situation/font/document/time of day/breakfast preferences/etc; I'm not sure where the idea came from that "grapheme cluster sequence" is the "appropriate" way to think of a string in a modern unicode world came from, but I don't really understand the rationale, beyond looking good in cute blog posts about how "🇦🇮🏳️‍🌈🇯🇵".length is "surprising"


Basically it feels like defaulting to EGCs is making things much more complicated in service of a particular set of operations that are still impossible to do properly even after all your blood, sweat and tears. To do any of this stuff "correctly" you need to know _at least_ the user's language/writing system and details about the active font, and making the core "string" type dependent on either is ..... probably a bad idea.

I'm unsure about unicode normalization-based equality comparison. It's not _as_ bad as grapheme clusters, but there are strings that are equivalent when normalized but very much not equivalent from the perspective of a reader/user.

martinvuyk:
I think we should maybe add graphemes later on but I wouldn't bet on it especially now that extensions are becoming a possibility. I think we should go for codepoint slicing and indexing by default because byte indexing and slicing is **unsafe** for UTF-8 and it should not be exposed by default.

Ideally I'd like this proposal to move forward by having indexing parametrized for `StringSlice` (and let people implement extensions for graphemes). Even better if I would be allowed to add `Encoding` into the mix as well. I think that having a native way to handle different encodings will give big performance and correctness improvements (I've already seen people building with `from_utf8` using http headers which are in `latin-1` for example). We could even get Python <-> Mojo native zero copy `StringSlice`s

> I'm unsure about unicode normalization-based equality comparison. It's not as bad as grapheme clusters, but there are strings that are equivalent when normalized but very much not equivalent from the perspective of a reader/user.

I think we should go for raw byte comparison as we currently are and create methods for different "case insensitive-ish" comparisons (see  #5389). We can have several performance choices for each, but the default should be the fastest and simplest.

PS: @barcharcraz If you're interested you can look at my backlog of PRs, I've been trying to fix a bunch of things that are wrong with our string for a while now

barcharcraz:
> I think we should maybe add graphemes later on but I wouldn't bet on it especially now that extensions are becoming a possibility. I think we should go for codepoint slicing and indexing by default because byte indexing and slicing is **unsafe** for UTF-8 and it should not be exposed by default.

Is byte indexing and slicing _really_ unsafe so long as you don't edit anything and when you actually _look_ at the slice you're able to skip over any bogus stuff at the beginning and end? Maybe it's not _useful_ but I'm not sure how unsafe it is. I'm not really sure String should provide `__getitem__` at all.

> Ideally I'd like this proposal to move forward by having indexing parametrized for `StringSlice` (and let people implement extensions for graphemes). Even better if I would be allowed to add `Encoding` into the mix as well. I think that having a native way to handle different encodings will give big performance and correctness improvements (I've already seen people building with `from_utf8` using http headers which are in `latin-1` for example). We could even get Python <-> Mojo native zero copy `StringSlice`s

I think it makes more sense to have an encoding parameter than it does an indexing parameter. Really the indexing parameter should be on `__getitem__` and not the slice. I think we should have separate indexing methods (or methods returning an iterator) on string slice for each kind of indexing you might want, and then _only_ have `__getitem__` on slices of "bytestring/ascii" encoding. 
> 
> > I'm unsure about unicode normalization-based equality comparison. It's not as bad as grapheme clusters, but there are strings that are equivalent when normalized but very much not equivalent from the perspective of a reader/user.
> 
> I think we should go for raw byte comparison as we currently are and create methods for different "case insensitive-ish" comparisons (see #5389). We can have several performance choices for each, but the default should be the fastest and simplest.
> 
> PS: @barcharcraz If you're interested you can look at my backlog of PRs, I've been trying to fix a bunch of things that are wrong with our string for a while now


Yes, I saw quite a collection, some of which look like good ideas. I'll try and get through them tomorrow (🫰🏼)

martinvuyk:
> Is byte indexing and slicing really unsafe so long as you don't edit anything and when you actually look at the slice you're able to skip over any bogus stuff at the beginning and end?

There is an example in #5281. If you really implement high performance stuff then you will make assumptions based on the first byte or continuation bytes of a sequence. We shouldn't be exposing that by default and hope developers read documentation (highly unlikely IMO).

> I think it makes more sense to have an encoding parameter than it does an indexing parameter. Really the indexing parameter should be on __getitem__ and not the slice. I think we should have separate indexing methods (or methods returning an iterator) on string slice for each kind of indexing you might want, and then only have __getitem__ on slices of "bytestring/ascii" encoding.

If you don't have an indexing parameter then how would functions like `__len__`, `find`, `__getitem__(idx)`, `__getitem__(Slice)` all coordinate in tandem without allowing developer error? It would be very easy for people to do `str_slice[:str_slice.find(";")]` and mistakenly use one indexing scheme for one and another for the other function call. There is a lot of code that would break if one of these methods had a different indexing scheme than the other

Anyway, I don't really care much about the indexing parameter, I proposed it more to have an option for all kinds including graphemes. But the default indexing, slicing, find, and length should be unicode codepoints due to safety and if byte indexing is desired then there is always `.as_bytes()` at hand (or we provide keyword getitem e.g. `[bytes=:3]` and `find[bytes=True](...)` or whatever).

> I think we should have separate indexing methods (or methods returning an iterator)

Me too (see #3653). Especially now that the iterator abstractions are maturing we might start pushing for other patterns like #5398. I'm stoked to see what this all will look like when it comes together :)

PR #3997, opened 2025-02-14
[docs] Add all env. vars. used to pass mojo lib. search paths to tutorial

izo0x90:
The env. var. shown in the tutorial section of the development document for the Stdlib is (at least in some cases) insufficient to point `mojo` to a newly build stdlib.mjpkg. This adds all the same variable that are passed in the `lit` config to hopefully get all edge cases covered and make getting stared easier.

Issue mentiones on Modular Discord:
https://discord.com/channels/1087530497313357884/1151418092052815884/1295522197275934824
https://discord.com/channels/1087530497313357884/1282843332690575444/1282845603017130127

### DIFF
--- a/stdlib/docs/development.md
+++ b/stdlib/docs/development.md
@@ -262,11 +262,14 @@ def main():
     print(all_paths.__str__())
 ```
 
-We also need to set the following environment variable that tells Mojo to
+We also need to set the following environment variables that tells Mojo to
 prioritize imports from the standard library we just built, over the one that
 ships with Mojo:
 
 ```bash
+MODULAR_MOJO_IMPORT_PATH=../build \
+MODULAR_MOJO_MAX_IMPORT_PATH=../build \
+MODULAR_MOJO_MAX_NIGHTLY_IMPORT_PATH=../build \
 MODULAR_MOJO_NIGHTLY_IMPORT_PATH=../build mojo main.mojo
 ```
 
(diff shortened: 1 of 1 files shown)


owenhilyard:
I'm wondering if this makes sense or if it's better to ship the stdlib with a `modular.cfg` in its root which has the import paths specified, if the LSP and compiler can handle that. If they can't, I'd like to suggest an "inherit/override" system, where the stdlib modular.cfg could be something like

```toml
[inherit]
path = "./.magic/envs/default/share/max/modular.cfg"

[mojo-max.import_path]
prepend = true
path = "./build/"
```

JoeLoser:
Gentle ping @Ahajha @zbowling

Ahajha:
Ah this is interesting. We're trying to move away from modular.cfg as a whole, but this is an interesting case. What we have currently I think is okay, things will get simpler when we open-source (or even just more bits from the stdlib like `benchmark`).

That said, the documentation is likely out of date, a lot of the env vars won't do anything now. `MODULAR_MOJO_MAX_IMPORT_PATH` is the only variable to adjust stdlib import paths.

Ahajha:
We actually _had_ a config override system internally until a couple days ago when I deleted it. FWIW, there were exactly 0 call sites outside of tests, so you couldn't have used it anyways.

izo0x90:
@Ahajha Should I just update this PR with only having `MODULAR_MOJO_MAX_IMPORT_PATH` for right now ... so that at least there is a working approach that ppl who are just getting started can look up in the docs ?

Ahajha:
Yea that sounds fine for now. In theory they shouldn't need to mess with the env vars, just running through `magic` should set everything up, but it's a step in the right direction.

izo0x90:
@Ahajha Actually looking at the current state of main looks like you had already made this update to the docs so I can just close this. Thank you !!

PR #3998, opened 2025-02-14
[stdlib] Adds functionality to spawn and manage processes from exec. file

izo0x90:
[Foundation for this PR is set here]( https://github.com/modular/mojo/pull/4017), it adds the needed lowlevel utilities

This PR:
- Adds utility struct. to allow spawning and managing child processes
  from an executable file
  - `run` factory method that creates process and returns "management"
    object
  - Instance methods to send signals to child process, INT, HUP, KILL
- Adds utility struct. to handle Posix pipe interprocess communication

Ex.:
```
import os

fn main():
 p = os.Process.run("ls", ["-lha"])
 p.interrupt()
```

Example usage:
https://github.com/izo0x90/mojo/blob/b99503e7bc5b2f210c957e5cbd8d0fb8b2c9ed1a/examples/process/process_example.mojo#L22-L146


stdlib/src/sys/_libc.mojo
```diff
@@ -106,6 +106,33 @@ fn dup(oldfd: c_int) -> c_int:
...
 
+@always_inline
+fn execvp(file: UnsafePointer[c_char], argv: UnsafePointer[c_str_ptr]) -> c_int:
+    """`execvp` expects that the c_str_ptr array is terminated with a NULL pointer.
+    """
+    return external_call["execvp", c_int](file, argv)
+
+
+@always_inline
+fn fork() -> c_int:
```
owenhilyard, line 117: `fork` is very, very unsound to use in Mojo right now. We probably can safely expose `vfork`, but `fork` allows threading in the absence of a concurrency model, which causes a lot of problems. Could you switch over to that and make this an internal function for Process?

stdlib/src/sys/_libc.mojo
```diff
@@ -106,6 +106,33 @@ fn dup(oldfd: c_int) -> c_int:
...
+    alias INT = 2  # (interrupt)
+    alias QUIT = 3  # (quit)
+    alias ABRT = 6  # (abort)
+    alias KILL = 9  # (non-catchable, non-ignorable kill)
+    alias ALRM = 14  # (alarm clock)
+    alias TERM = 15  # (software termination signal)
+
+
+@always_inline
+fn kill(pid: c_int, sig: c_int):
```
owenhilyard, line 132: This should provide the return code from `kill`.

stdlib/src/sys/ffi.mojo
```diff
@@ -71,6 +71,9 @@ alias c_float = Float32
 alias c_double = Float64
 """C `double` type."""
 
+alias c_str_ptr = UnsafePointer[Int8]
```
owenhilyard, line 74: This is incorrect on arm64 Linux. Please use `UnsafePointer[c_char]`.

stdlib/src/os/os.mojo
```diff
@@ -415,3 +422,94 @@ def removedirs[PathLike: os.PathLike](path: PathLike) -> None:
...
+struct Process:
+    """Create and manage child processes from file executables.
+    TODO: Add windows support.
+    """
+
+    var child_pid: c_int
+    """Child process id."""
+
+    fn __init__(mut self, child_pid: c_int):
+        """Struct to manage meta about child process.
```
owenhilyard, line 441: meta -> metadata

stdlib/src/os/os.mojo
```diff
@@ -415,3 +422,94 @@ def removedirs[PathLike: os.PathLike](path: PathLike) -> None:
...
+
+struct Process:
+    """Create and manage child processes from file executables.
+    TODO: Add windows support.
+    """
+
+    var child_pid: c_int
+    """Child process id."""
+
+    fn __init__(mut self, child_pid: c_int):
```
owenhilyard, line 440: Add something to point users towards `Process.run` to spawn a process.

stdlib/src/os/os.mojo
```diff
@@ -415,3 +422,94 @@ def removedirs[PathLike: os.PathLike](path: PathLike) -> None:
...
+
+
+# ===----------------------------------------------------------------------=== #
+# Process execution
+# ===----------------------------------------------------------------------=== #
+
+
+struct Process:
+    """Create and manage child processes from file executables.
+    TODO: Add windows support.
```
owenhilyard, line 434: Windows support is a ways off, please remove that and we'll update the whole stdlib later.

stdlib/src/os/os.mojo
```diff
@@ -415,3 +422,94 @@ def removedirs[PathLike: os.PathLike](path: PathLike) -> None:
...
             break
         head, tail = os.path.split(head)
+
+
+# ===----------------------------------------------------------------------=== #
+# Process execution
+# ===----------------------------------------------------------------------=== #
+
+
+struct Process:
```
owenhilyard, line 432: Process management probably deserves its own file. Thoughts?

stdlib/test/os/test_process.mojo
```diff
@@ -0,0 +1,45 @@
...
+    )
+
+    # Forking appears to break asserts
+    with assert_raises():
+        _ = Process.run(missing_executable_file, List[String]())
+
+
+def main():
+    test_process_run()
+    # TODO: How can exception being raised on missing file be asserted
```
owenhilyard, line 44: https://docs.modular.com/mojo/stdlib/testing/testing/assert_raises/

owenhilyard (commented):
Good work so far, just a few fixes.

stdlib/src/os/os.mojo
izo0x90: That def. makes sense, I just wasn't sure if I should break out yet just for this ... but I guess eventually there will have to be more process stuff. I'll add `process.mojo` if that is inline with the plan on how to structure the stdlib ?

stdlib/test/os/test_process.mojo
izo0x90: The issue here from what I understand seems to be that I am rasing the exception in the forked child process when we don't find the executable file ... which `try` actually handles gracefully ... so this works fine:
```
try:
  _ = Process.run("file_does_not_exist", List[String]())
except e:
 print(e)
```
But for some reason (I assume related to the raise being in a different process) when try to validate the raise with:
```
with assert_raise():
  _ = Process.run("file_does_not_exist", List[String]())
```
It crashed the test suite with a no such process error. 🤔

owenhilyard: I think that `raises` in a fork pre-exec is unsound, you'll need to check before you fork.

izo0x90: Ok will work on something to do an explicit check before the the fork happens.

izo0x90: Update: Did a little digging on the established/ canonical way to deal with this on posix/ unix systems and implemented it here, everything is working on Macos ... just running into some issues with linux, guessing some inconsistencies between the platform libc calls, working on clearing that up.

izo0x90: @owenhilyard Hey quick question for you, so I did some testing in a linux container. The crashing/ failing tests seem to be related to the test runners on Linux and not the actual functionality. As you can see I can successfully run a process on linux using the Process utilities ( see screen shot below ). 

Do you have any ideas on how to proceed from here, it seems the code being added is functioning as intended ? Thanks again for all the feedback/ help! 🔥 

Screenshot 2025-02-20 at 7 08 42 PM

owenhilyard: Does doing non-trival work in the other threads function correctly? Try to do something like launch a simple max graph from one of them.

JoeLoser:
Exciting to see work on this area!  Feel free to ping me when CI is passing for Linux as well - we'll need to ensure everything works on both Linux and macOS before we can land this.

stdlib/test/os/test_process.mojo
izo0x90: Did some further testing now that I got a min, what appears to be failing on Linux is as follows:

- We issue an `execvp` call to a non existent executable in the forked process
- The forked process signals to the parent through the share pipe that there was an error
- Forked process closes out pipe and exits
- Main process, run function finished running
- Main process that called `Process.run` finishes running, all the library and user code finishes running
- At the end when that process is exiting on Linux, with whatever the runtime does on exit, we get a fatal error from the Mojo runtime
 Crash at end of program:
Screenshot 2025-02-21 at 4 21 31 PM
- However if we just call `exit()` at the end of our program that called `Process.run` everything finishes closing out the program fine and we get no errors from the runtime 🫤 


So literally this causes the runtime to complain and error out:
```
from collections import List
from os import Process
from sys._libc import exit
fn main() raises:
    print("IN MAIN")
    try:
        print("ATTEMPTING EXECUTION")
        _ = Process.run(
            "brokeenn", List[String]("\n\nCORRECTLY RUNNING CHILD PROCESS !")
        )
    except:
        print("GOT THE EXCEPTION")
    print("END MAIN PROCESS")
    print("IF WE CALL `exit()` NO ISSUE. IF WE DO NOT, RUNTIME CRASHES.")
    # exit(1) # NO CALL TO `exit`, WE GET AN ERROR FROM THE RUNTIME
```
But just adding `exit({any_value})` allows everything to run exit the program fine:
```
from collections import List
from os import Process
from sys._libc import exit
fn main() raises:
    print("IN MAIN")
    try:
        print("ATTEMPTING EXECUTION")
        _ = Process.run(
            "brokeenn", List[String]("\n\nCORRECTLY RUNNING CHILD PROCESS !")
        )
    except:
        print("GOT THE EXCEPTION")
    print("END MAIN PROCESS")
    print("IF WE CALL `exit()` NO ISSUE. IF WE DO NOT, RUNTIME CRASHES.")
    exit(1) # WITH A CALL TO EXIT, EVERYTHING FINISHES RUNNING FINE
```

We can observe this issue with:
- `mojo run`
- `mojo build` and then run the binary
- In the test being ran by the test suite
Always once all stdlib and user code has finished running and the runtime is doing cleanup closing out things.

Wonder what next steps can be since we don't have any access to the Mojo runtime codebase ?

@owenhilyard @JoeLoser (Again thanks for the feedback and help!)

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +87,38 @@ struct FileDescriptor(Writer):
                 written,
             )
 
+    @always_inline
+    fn read_bytes(mut self, size: Int) raises -> Span[Byte, MutableAnyOrigin]:
```
ConnorGray, line 91: **Suggestion** Given that this returns owned memory that the user will need to be sure to free, it might be better if this returned a `List[Byte]`, so that the pointer is freed automatically.

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +87,38 @@ struct FileDescriptor(Writer):
...
+
+        @parameter
+        if is_nvidia_gpu():
+            constrained[False, "Nvidia GPU read bytes not implemented"]()
+            return abort[Span[Byte, MutableAnyOrigin]]()
+        elif is_amd_gpu():
+            constrained[False, "AMD GPU read bytes not implemented"]()
+            return abort[Span[Byte, MutableAnyOrigin]]()
+        elif os_is_macos() or os_is_linux():
+            var buf = UnsafePointer[UInt8].alloc(size)
```
ConnorGray, line 110:
```suggestion
            var list = List[Byte](capacity=size)
```

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +87,38 @@ struct FileDescriptor(Writer):
...
+            read = external_call["read", Int](self.value, buf, size)
```
ConnorGray, line 111:
```suggestion
            read = external_call["read", Int](self.value, buf.unsafe_ptr(), size)
```

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +87,38 @@ struct FileDescriptor(Writer):
...
+
+            if read < 0:
+                buf.free()
+                raise Error("Failed to read bytes.")
+
```
ConnorGray, line 116:
```suggestion
list.size = read
```

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +87,38 @@ struct FileDescriptor(Writer):
...
+            return Span[Byte, MutableAnyOrigin](ptr=buf, length=size)
```
ConnorGray, line 117:
```suggestion
            return list^
```

ConnorGray:
Hi @izo0x90, thanks for working on this, this adds a lot of great new APIs! 🔥 🙂

The team just had some discussion about this PR and are generally looking forward to merging this functionality 🙂

I did want to ask if you might be able to split this PR, to make each individual component added easier to review.

For example, you might start with a smaller PR that just contains the new `_libc.mojo` functions and the `FileDescriptor.read_bytes()` method on their own. That would make it easier to read and provide meaningful suggestions for those changes on their own. Then this PR could stay focused to meaty new `process.mojo` file being added here.

Please let me know if you think splitting this up could be feasible, and if I can provide any help with that just let me know :)

Thanks again for contributing to Mojo, we're glad you're hear 😄

stdlib/src/builtin/file_descriptor.mojo
izo0x90: Awesome, done, I actually originally had it as a list and only changed it later to have it match the write method lol. 🔥

stdlib/test/os/test_process.mojo
JoeLoser: I hope to have some time next week to take a look more into this.  The team has been really busy working on some things we're showcasing next week at Nvidia GTC - sorry for the delay!

izo0x90: No worries at all, thank you, let me know what I can do once you guys get a chance to take a look 🔥 ... Best of luck with the showcase !!!!

JoeLoser:
@izo0x90 do you mind rebasing, please? Then I'm happy to sync it back in internally and take a look at your errors on Linux now that GTC is behind us 😄

izo0x90:
@JoeLoser sry for the delay, will make those updates shortly (as soon as I get caught up with some work stuff 😄 ) and ping you when its done. Thanks again !!

izo0x90:
@JoeLoser Rebased on newest main ... tested locally that I can still build the stdlib and replicate the same issue I was seeing before in Linux (ubuntu docker container). Let me know if there is any other way to assist with figuring out the issue or providing further detail in what I am seeing. I can find time and jump on discord or video chat etc. if the issue does not replicate for you (altho it does seem to show up in the linux test that CI run in an actual Linux vm) or anything else. Thanks again for taking a look at this !!🚀

JoeLoser:
!sync

izo0x90:
@JoeLoser Hey Joe, just checking in seeing if there is anything I can work on here. Thanks much!

JoeLoser:
> @JoeLoser Hey Joe, just checking in seeing if there is anything I can work on here. Thanks much!

Hi @izo0x90 — sorry this got held up internally due to some compiler issues preventing this from landing.  Do you mind rebasing this now that we've switched over to using Bazel in the external builds?  Once CI is green in the external build here, I'm happy to sync this in.  Note that since `test_process.mojo` uses `FileCheck`, we'll need to add it to the Bazel list [here](https://github.com/modular/modular/blob/c49ed04653a42d94704f59fe5c8260c70e40a96b/mojo/stdlib/test/BUILD.bazel#L188).  You can remove the `RUN` directives etc at the top of `test_process.mojo` then.

izo0x90:

izo0x90:
@JoeLoser Hey finally got around to this, sry for the delay ... update code and tests to match the new syntax, conventions and I am able to get tests passing locally now. MacOS testes failed on GH but does not seem related to the process module as `test_process.mojo` shows as successful in the logs and many other tests run after. Thanks again for all your help with this.

izo0x90:
@JoeLoser Clean pass when pulling in the newest main changes.

mojo/stdlib/stdlib/os/process.mojo
```diff
@@ -0,0 +1,284 @@
...
+                var arg_count = len(argv)
+                var argv_array_ptr_cstr_ptr = UnsafePointer[
+                    UnsafePointer[c_char, mut=False]
+                ].alloc(arg_count + 2)
+                var offset = 0
+                # Arg 0 in `argv` ptr array should be the file name
+                argv_array_ptr_cstr_ptr[offset] = file_name.unsafe_cstr_ptr()
+                offset += 1
+
+                for var arg in argv:
```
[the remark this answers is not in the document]
izo0x90: We need to copy the strings here because `unsafe_cstr_ptr` is "mutating". Could use `ref` but then we would require the args list to be passed in as `mut`.

mojo/stdlib/stdlib/os/process.mojo
```diff
@@ -0,0 +1,284 @@
...
+                var err_len = exec_err_code.byte_length()
+                var buf = Span[Byte, MutableAnyOrigin](
+                    ptr=UnsafePointer[Byte].alloc(err_len), length=err_len
+                )
+                buf[0] = 0  # Explicitly default to empty C string
+                var bytes_read = pipe.read_bytes(buf)
+                err = StringSlice(unsafe_from_utf8=buf)
+            except e:
+                err = None
+
```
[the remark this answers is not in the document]
izo0x90: +1 to the LLMs lol. No need for an allocation here at all actually, since we are the ones setting the error string which never changes all lengths are know at compile time so we can just use an InlineArray and avoid the heap altogether.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,290 @@
...
+
+    var fd_in: Optional[FileDescriptor]
+    """File descriptor for pipe input."""
+    var fd_out: Optional[FileDescriptor]
+    """File descriptor for pipe output."""
+
+    fn __init__(
+        out self,
+        in_close_on_exec: Bool = False,
+        out_close_on_exec: Bool = False,
```
barcharcraz, line 117: IMO the default should be true.

mojo/stdlib/stdlib/os/process.mojo
```diff
@@ -0,0 +1,290 @@
...
+        if CompilationTarget.is_linux() or CompilationTarget.is_macos():
+            var file_name = String(path.split(sep)[-1])
+            var pipe = Pipe(out_close_on_exec=True)
+            var exec_err_code = StaticString("EXEC_ERR")
+
+            var pid = vfork()
+
+            if pid == 0:
+                # Child process.
+                pipe.set_output_only()
```
barcharcraz: nb: this is only OK because pipe is on the stack. Would prefer we just close the fd and not write onto the stack.

mojo/stdlib/stdlib/os/process.mojo
```diff
@@ -0,0 +1,290 @@
...
+
+                # `argv` ptr array terminates with NULL PTR
+                argv_array_ptr_cstr_ptr[offset] = UnsafePointer[c_char]()
+
+                var path_cptr = path.unsafe_cstr_ptr()
+
+                _ = execvp(path_cptr, argv_array_ptr_cstr_ptr)
+
+                # This will only get reached if exec call fails to replace currently executing code
+                argv_array_ptr_cstr_ptr.free()
```
barcharcraz: yes, and because of that I'm pretty sure the allocation leaks in the parent process. vfork doesn't change the page-table mapping, so the allocation is in BOTH processes and it's freed ONLY in the child by virtue of the exec wiping everything out.

mojo/stdlib/stdlib/os/process.mojo
```diff
@@ -0,0 +1,290 @@
...
+
+                # Canonical fork/ exec error handling pattern of using a pipe that closes on exec is
+                # used to signal error to parent process `https://cr.yp.to/docs/selfpipe.html`
```
barcharcraz: Not sure about macOS but on linux you can use `CLONE_PIDFD` to avoid this race

barcharcraz: and this is another reason to just call `posix_spawn`

mojo/stdlib/stdlib/os/process.mojo
```diff
@@ -0,0 +1,290 @@
...
+                pipe.write_bytes(exec_err_code.as_bytes())
+
+                exit(1)
```
barcharcraz: I think this will call the parent's atexit handlers and flush the parents io buffers

mojo/stdlib/stdlib/os/process.mojo
```diff
@@ -0,0 +1,290 @@
...
+                var bytes_read = pipe.read_bytes(buf)
+                err = StringSlice(unsafe_from_utf8=buf)
+            except e:
+                raise Error(
+                    "Failed to read child process response from pipe, exception"
+                    " was: "
+                    + String(e)
+                )
+
+            if err and len(err.value()) > 0 and err.value() == exec_err_code:
```
barcharcraz: does the child need to be reaped in this case?

barcharcraz:
My general feeling on this is that we should be using `posix_spawn` if possible. vfork is a very fiddly API.

izo0x90:
@barcharcraz 
> My general feeling on this is that we should be using `posix_spawn` if possible. vfork is a very fiddly API.

Thanks for the review much appreciated... will dive into this deeper after work today, but having taken a quick look at `posix_spawn` def. seems like a better option and can let us expose more user control over what fd are brought in the child process etc. eventually. I will reach out once I have those changes implemented on here, thanks again.

Hundo1018:
Hi izo, I'm curious about different between stdlib/os/process and stdlib/subprocess, or just something changes I missing? anyway, great work for the process, it will be very useful!

izo0x90:
> Hi izo, I'm curious about different between stdlib/os/process and stdlib/subprocess, or just something changes I missing? anyway, great work for the process, it will be very useful!

Great question lol ... the general idea here was to provide a way to spawn processes without blocking the parent process ... plan was to work on signal handler next, so we can get status and know when child done running, in a non-blocking way. 

This was before the subprocess module was merged however and I am just now aware of it, and while that runs subprocesses in a blocking manner, still brings into question if what I am doing here is redundant.

@barcharcraz @JoeLoser In light of @Hundo1018 question, and having seen that a subprocess module now exists in main, is the work I am doing here for a non-blocking process execution (and not using clib popen) something I should continue ?

barcharcraz:
I definitely think it's worth having a way to spawn subprocesses without going through the shell. The API in subprocess.mojo is quite bare at the moment.

JoeLoser:
> I definitely think it's worth having a way to spawn subprocesses without going through the shell. The API in subprocess.mojo is quite bare at the moment.

+1, I agree with Charlie here @izo0x90

izo0x90:
@barcharcraz @JoeLoser Here is a cleaned up version using `posix_spanwp` ... 

- implements the polling functionality I had in mind 
- we reap the subprocess/ always wait on exit in the destructor to make sure no zombie pids. 
- Also the Process object now keeps internal state to be able to check if we already retrieved the exit status of the child process before so that there is no chance of targeting another process to which the os might have assigned the pid after our child process exited. (From what I am reading online this is unlikely, but not impossible so worth handling imo)

Let me know what you think whenever you get a chance to take a look, thank you.

izo0x90:
Updated to pass all checks and new Pointers sematics.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+        Args:
+            in_close_on_exec: Close the read side of pipe if an `exec` syscall
+              is issued in the process.
+            out_close_on_exec: Close the write side of pipe if an `exec`
+              syscall is issued in the process.
+
+        Raises:
+            Error: If the pipe could not be created or configured.
+        """
+        var pipe_fds = alloc[c_int](2)
```
barcharcraz: must we malloc these? I think we can use an inline array.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+        if pipe(pipe_fds) < 0:
+            pipe_fds.free()
+            raise Error("Failed to create pipe")
+
+        if in_close_on_exec:
+            if not self._set_close_on_exec(pipe_fds[0]):
```
barcharcraz, line 135: Ideally we'd call `pipe2` to avoid the race where some other thread execs in between the `pipe` and the `fcntl`, but macOS doesn't have pipe2 and avoiding these races is a bit of a sisyphean task.

barcharcraz (commented):
This looks good aside from the allocation in `Pipe` (which honestly is minor, I can correct it while merging if you'd like).

One annoyance is we have no way of detaching a process, but this is complicated and can be left for later. I think it should just be a matter of defining flags for `posix_spawn` but they'll be somewhat platform specific.

barcharcraz:
!sync

izo0x90:
> This looks good aside from the allocation in `Pipe` (which honestly is minor, I can correct it while merging if you'd like).
> 
> One annoyance is we have no way of detaching a process, but this is complicated and can be left for later. I think it should just be a matter of defining flags for `posix_spawn` but they'll be somewhat platform specific.

Yes I left off the flags as that def. looked like another whole can of worms, was hoping something that can be done as follow up as adoption grows loll. 

If you are ok with making those pipe fixes when merging that would awesome and much appreciated 🔥 ... just let me know if there is anything else I can do. Thanks again for all the feedback/ help.

barcharcraz:
> > This looks good aside from the allocation in `Pipe` (which honestly is minor, I can correct it while merging if you'd like).
> > One annoyance is we have no way of detaching a process, but this is complicated and can be left for later. I think it should just be a matter of defining flags for `posix_spawn` but they'll be somewhat platform specific.
> 
> Yes I left off the flags as that def. looked like another whole can of worms, was hoping something that can be done as follow up as adoption grows loll.
> 
> If you are ok with making those pipe fixes when merging that would awesome and much appreciated 🔥 ... just let me know if there is anything else I can do. Thanks again for all the feedback/ help.

yeah, I'll get rid of the allocation. `pipe2` is just a whole can-o-worms.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+        self.child_pid = child_pid
+        self.status = None
+
+    fn __del__(deinit self):
+        """Waits for the process to exit when the `Process` object is destroyed.
+        """
+        try:
+            _ = self.wait()
+        except:
+            # Errors in __del__ should be suppressed.
```
JoeLoser, line 375: **Question** If `wait()` fails (e.g. already called, or some other error) - the child process may become a zombie. Is there anything better we can do here @barcharcraz? Suppressing all errors silently can mask real issues.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+            pass
+
+    fn _kill(mut self, signal: Int) -> Bool:
```
JoeLoser, line 378: This is a TOCTOU (time of check, time of use) race I believe (right @barcharcraz?)

Like your comment says, the process can exit between `poll()` and `kill()`, potentially allowing the PID to be reused for a different process. I'd recommend we just call `kill()` directly and check for `ESRCH` errno. The kernel handles this automatically for us:

```mojo
fn _kill(mut self, signal: Int) -> Bool:
  if kill(self.child_pid, signal) == 0:
      return True
  var err = get_errno()
  return err != ESRCH # False if process doesn't exist, True for other errors
```

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+        if out_close_on_exec:
+            if not self._set_close_on_exec(pipe_fds[1]):
+                pipe_fds.free()
+                raise Error("Failed to configure output pipe close on exec")
+
+        self.fd_in = FileDescriptor(Int(pipe_fds[0]))
+        self.fd_out = FileDescriptor(Int(pipe_fds[1]))
+        pipe_fds.free()
+
+    fn __del__(deinit self):
```
JoeLoser, line 145: I think we have some resource management issues here.  `close()` can fail (e.g. `EINTR`, `EIO`) but the errors are ignored here.  This may be easier to ensure correctness if we initialize `fd_in` and `fd_out` to `None` before any operations that raise in `__init__` for example. Then we could check the return values from `close()` and handle errors appropriately.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+
+            # `argv` ptr array terminates with NULL PTR
+            argv_array_ptr_cstr_ptr[offset] = UnsafePointer[
+                mut=False, c_char, ImmutAnyOrigin
+            ]()
+            var path_cptr = path.unsafe_cstr_ptr()
+
+            var pid: c_pid_t = 0
+
+            var has_error_code = posix_spawnp(
```
JoeLoser, line 559: **Note** `posix_spawnp` call here passes the environment pointer as NULL, meaning the child inherits the parent's environment. Currently we provide no way to customize this: is that intentional?

**Suggestion**: We could add an optional `env: Optional[Dict[String, String]]` argument to `run()` and use it here.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+        var status: c_int = 0
+        var pid = waitpid(
+            self.child_pid, UnsafePointer(to=status), WaitFlags.WNOHANG
+        )
+        var result = self._check_status(pid, status)
+        if result.has_exited():
+            self.status = result
+        return result
+
+    fn wait(mut self) raises -> ProcessStatus:
```
JoeLoser, line 482: **Question** What happens if you call `wait()` multiple times? The code currently caches the status (which is good) but this isn't documented. I'd suggest we document that `wait()` and `poll()` cache results so that subsequent calls return the cached status.

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+                fcntl(fd, FcntlCommands.F_GETFD, 0) | FcntlFDFlags.FD_CLOEXEC,
+            )
+            == 0
+        )
+
+    @always_inline
+    fn set_input_only(mut self):
+        """Close the output descriptor/ channel for this side of the pipe."""
+        if self.fd_out:
+            _ = close(rebind[Int](self.fd_out.value()))
```
JoeLoser, line 166: **Question** Why is the `rebind[Int]` needed here and below in `set_output_only`, etc? By using `rebind`, we're bypassing type checking essentially. Could we make it so that `FileDescriptor` can be passed directly to `close()` perhaps?

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+                )
+
+            argv_array_ptr_cstr_ptr.free()
+
+            return Process(child_pid=pid)
+        else:
+            constrained[
+                False, "Unknown platform process execution not implemented"
+            ]()
+            abort[prefix="ERROR:"](
```
JoeLoser, line 580: **Question** Is this reachable, or is it redundant?

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Implements os methods for dealing with processes.
```
JoeLoser, line 13: It'd be great to add some module/API docstrings for some of the more nuanced things for usage patterns, caveats, etc - especially around resource cleanup.

1. What happens if you call signal methods after the process has exited?
2. Are these methods thread-safe?
3. What's the behavior of calling `wait()` multiple times?
4. Should users always call `wait()` or is relying on `__del__` sufficient?

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+
+            if has_error_code > 0:
+                raise Error(
+                    "Failed to execute "
+                    + path
+                    + ", EINT error code: "
+                    + String(has_error_code)
+                )
+
+            argv_array_ptr_cstr_ptr.free()
```
JoeLoser: **Suggestion** Good that we free the `argv` stuff, but can we document the lifetime requirements of the `String` objects themselves in `argv`?

mojo/stdlib/test/os/test_process_stdout.mojo
```diff
@@ -0,0 +1,53 @@
+# ===----------------------------------------------------------------------=== #
```
JoeLoser, line 1: **Suggestion** I'd like to see some comprehensive tests, especially for `Pipe` which has zero test coverage right now.

Some ideas of tests that are missing right now:

- Exit code verification
- Signal handling (interrupt, hangup, kill)
- Polling vs. waiting
- Multiple wait calls (we have this in an example elsewhere though IIRC)
- Pipe functionality is completely untested right now
- Process that exits immediately
- Process that runs for a long time (this is OK though as you have an example elsewhere)

mojo/stdlib/std/os/process.mojo
```diff
@@ -0,0 +1,441 @@
...
+        var result = self._check_status(pid, status)
+        if result.has_exited():
+            self.status = result
+        else:
+            # This should not be reachable with a blocking waitpid call.
+            raise Error("Blocking waitpid returned without process exiting.")
+        return result
+
+    @staticmethod
+    fn run(var path: String, argv: List[String]) raises -> Process:
```
JoeLoser: **Suggestion** Currently, we don't have a way to capture stdout/stderr from the child, provide stdin to the child, or redirect I/O to files or pipes.  This would all be good essential functionality for a process API. I'd suggest tackling this in a follow-up by adding optional `stdin`, `stdout`, and `stderr` arguments that accept `Pipe` or file descriptors.

JoeLoser (commented):
This is looking good! I left a few comments to check out before we land this.  Feel free to consider some of them as things to tackle as follow-ups - use your best judgement and work with @barcharcraz on it.

Don't forget to also add a [changelog entry here](https://github.com/modular/modular/blob/13736a8166b257e09dadcdb08ff73baf36084de5/mojo/docs/changelog.md#library-changes) so others know about your great improvements!

Thanks again for working on this, very exciting to see!

mojo/stdlib/std/os/process.mojo
barcharcraz: It's actually OK to swallow errors from close. EBADF is basically structurally impossible, historically everybody except HP-UX would close the fd before returning EINTR, creating inconsistency. Now the open group has said the FD _MUST_ still be open if EINTR is returned but EINPROGRESS should indicate the call was interrupted but the fd was closed.

Sadly this change needs to be adopted by linux. Basically everyone still does the linux semantics.

I don't think we need to care about HP-UX, we can assume the sane linux semantics.

For other IO errors there's nothing we can do by the time we get to this close call. If the user wanted to handle them they would have synched the file.


See:
https://lwn.net/Articles/576478/
https://ewontfix.com/4/

mojo/stdlib/std/os/process.mojo
barcharcraz: I think we should have a higher level FD api. For now I'd kinda rather just use c_int directly throughout (Int is actually wrong, anyway).

barcharcraz: Honestly, Pipe kinda _is_ the higher level fd api.

mojo/stdlib/std/os/process.mojo
barcharcraz: The PID won't be reused even if the process has already exited. The process will become a zombie until reaped because we don't change the default signal disposition for sigchld.

The way SIGCHLD works is confusing, if you set the disposition to SIG_IGN then the process isn't zombified at all (and even the poll might poll the wrong process), but if it's SIG_DFL then you _do_ get zombies even though the "default" action is to ignore the signal.

Anyway, since we aren't messing with SIGCHLD we do get zombies. Either we poll and the process has exited, in which case the poll will reap the process and we never kill, or it hasn't and the process isn't reaped. After being killed it will be zombified to be handled by a later `wait`/`pool`/`__del__`

barcharcraz: Turns out it's hard to kill your children :)

mojo/stdlib/std/os/process.mojo
izo0x90: Yes, my thinking here was to add this as a follow up... was't sure what the best way to add a todo was.

mojo/stdlib/std/os/process.mojo
owenhilyard: @barcharcraz Can you include that last comment inline in the function (no need for a docstring) as a note for future readers? If Linux ever updates its behavior (unlikely) we'll need to fix that, and there's a non-zero chance of another POSIX OS (Redox comes to mind) being written to the spec with the now-correct behavior. Having that information there will save a lot of time.

mojo/stdlib/std/os/process.mojo
izo0x90: What is the best way to document that ?

mojo/stdlib/std/os/process.mojo
owenhilyard: Being able to supply `String`/`StringSlice`/`StringLiteral` would be nice as well.

izo0x90: Is there a way to do that in a "Trait" way somehow yet ? Or do we just do a union type of those 3 ?

owenhilyard: I think you'll need to do it with function overloading and wrap the string versions around the FD versions.

mojo/stdlib/std/os/process.mojo
izo0x90: A return is expected out of the function ... so even tho. we have constrained in this code path the compiler still complains about not having a return if we do not explicitly call abort

izo0x90:
@JoeLoser @barcharcraz went ahead and added in passing in env and redirecting stdio for the subprocess as well a test for just about all functionality ... let me know what you think when you get a chance to take a look. Thanks 🔥 !!

izo0x90:
Also:
- Updated allocations to raii pattern (a bit of extra complexity but the if/ try finally check to catch all fail cases looked much worse to me), more robust deallocations for all the different code paths that can fail in the run method.
- Updated Pipe to use InlineArray so no more dynamic alloc there.

barcharcraz:
> @JoeLoser @barcharcraz went ahead and added in passing in env and redirecting stdio for the subprocess as well a test for just about all functionality ... let me know what you think when you get a chance to take a look. Thanks 🔥 !!

Let's review these improvements in another PR. I've merged 68364e43f06de0dac68c584f369cd956f1618cb3 (which is the last of your commits here that I've looked at). You should be able to do something like `git rebase --onto origin/master 68364e43f` to transplant the new commits.

Apologies on the long wait before merging.

PR #3999, opened 2025-02-14
[stdlib docs] Information on Dependent types

owenhilyard:
Chris gave some information I thought should be documented for stdlib developers. 

* `always_inline("builtin")` vs `always_inline("nodebug")`
* Some information on MLIR interpreter limitations.

### DIFF
--- a/stdlib/docs/internal/compiler.md
+++ b/stdlib/docs/internal/compiler.md
@@ -1,0 +1,21 @@
+## WARNING
+
+Everything in this file is subject to revision on any bugfix or security
+update. We (the stdlib team and contributors), reserve the right to remove,
+change the API contracts of, rename, or cause to instantly crash the program,
+any operation described in here. These are **PRIVATE** APIs and implementation
+details for the Mojo stdlib and for MAX to use. **WE WILL CHANGE IT WHENEVER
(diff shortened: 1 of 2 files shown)


ConnorGray:
Hey @owenhilyard, thanks for contributing this! :)

This looks basically good to me, though I wonder if you might consider including footnotes to the places where Chris' mentions the things added here? Even if the best sources at the moment are things Chris has said on Discord or forum posts, I think any "breadcrumbs" we can give future readers are better than none.

owenhilyard:
> Hey @owenhilyard, thanks for contributing this! :)
> 
> This looks basically good to me, though I wonder if you might consider including footnotes to the places where Chris' mentions the things added here? Even if the best sources at the moment are things Chris has said on Discord or forum posts, I think any "breadcrumbs" we can give future readers are better than none.

I've included links to the Discord thread where Chris gave the explanation. I have no idea how stable those are long-term, but we shall see.

ConnorGray:
Thanks Owen! Understood, I don't think it'll be great to rely on Discord as a primary source long-term ;)

But perhaps that will tide us over until the Mojo compiler is open source🙂

ConnorGray:
!sync

lattner:
Nice, thank you for putting this together @owenhilyard !

JoeLoser:
!sync

PR #4005, opened 2025-02-17
[stdlib] Add iterator for LinkedList

bgreni:
There was some previous discussion about whether casting the lifetime of the list elements in this manner was sound, but now that 25.1 is out the door I think we can revisit this?

From my understanding the lifetime of the list elements should be the same as the list itself so it _should_ be fine to do it this way?

### DIFF
--- a/stdlib/src/collections/linked_list.mojo
+++ b/stdlib/src/collections/linked_list.mojo
@@ -86,6 +86,50 @@ struct Node[
         writer.write(self.value)
 
 
+@value
+struct _LinkedListIter[
+    mut: Bool, //,
+    ElementType: CollectionElement,
+    origin: Origin[mut],
+    forward: Bool = True,
+]:
+    var src: Pointer[LinkedList[ElementType], origin]
+    var curr: UnsafePointer[Node[ElementType]]
+
+    # Used to calculate remaining length of iterator in
+    # _LinkedListIter.__len__()
+    var seen: Int
+
+    fn __init__(out self, src: Pointer[LinkedList[ElementType], origin]):
+        self.src = src
+
+        @parameter
+        if forward:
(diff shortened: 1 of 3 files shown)


stdlib/src/collections/linked_list.mojo
```diff
@@ -692,6 +733,30 @@ struct LinkedList[
         """
         return self._size
 
+    fn __iter__(self) -> _LinkedListIter[ElementType, __origin_of(self)]:
+        """Iterate over elements of the list, returning immutable references.
+
+        Time Complexity: O(1)
```
owenhilyard, line 739: Can you clarify the iterator construction is `O(1)` but iteration is `O(n)`, and do the same for `__reversed__`?

owenhilyard:
Other than some doc clarifications, looks good to me.

JoeLoser (approved):
Thanks for the contribution!  LGTM once the doc comments from Owen are addressed.  Please add a changelog entry as well and then I'm happy to sync it in.

JoeLoser:
!sync

stdlib/src/collections/linked_list.mojo
```diff
@@ -86,6 +86,47 @@ struct Node[
...
+@value
+struct _LinkedListIter[
+    mut: Bool, //,
+    ElementType: CollectionElement,
+    origin: Origin[mut],
+    forward: Bool = True,
+]:
+    var src: Pointer[LinkedList[ElementType], origin]
+    var curr: UnsafePointer[Node[ElementType]]
+    var seen: Int
```
JoeLoser, line 98: **Suggestion** From review internally, can you please add a comment for what this variable does/why it's needed?  Thanks.

JoeLoser:
!sync

PR #4015, opened 2025-02-25
[stdlib] Add split overloads

martinvuyk:
Add all split overloads with examples. Part of #3528

cc: @ConnorGray

### DIFF
--- a/mojo/stdlib/stdlib/builtin/string_literal.mojo
+++ b/mojo/stdlib/stdlib/builtin/string_literal.mojo
@@ -697,13 +697,12 @@ struct StringLiteral[value: __mlir_type.`!kgen.string`](
         var result = self
         return result.join(elems)
 
-    fn split(self, sep: StringSlice, maxsplit: Int = -1) -> List[StaticString]:
+    @always_inline
+    fn split(self, sep: StringSlice) -> List[StaticString]:
         """Split the string by a separator.
 
         Args:
             sep: The string to split on.
-            maxsplit: The maximum amount of items to split from String.
-                Defaults to unlimited.
 
         Returns:
             A List of Strings containing the input split by the separator.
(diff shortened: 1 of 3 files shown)


stdlib/src/collections/string/string_slice.mojo
```diff
@@ -1737,6 +1680,100 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
 
         return True
 
+    @always_inline
+    fn split(self, sep: StringSlice, maxsplit: Int) raises -> List[Self]:
```
ConnorGray, line 1684: **Nit** I wonder if these should take `UInt` instead of `Int`, since at this point they should never be negative. I don't think that's worth changing in this PR, since there might be some value to preserving the ability to pass `-1` explicitly. Just food for thought for future ways to potentially tighten up this API.

stdlib/src/collections/string/string_slice.mojo
```diff
@@ -1737,6 +1680,100 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
 
         return True
 
+    @always_inline
```
ConnorGray, line 1683: **Suggestion** I notice these `split()` methods are being added around line ~1683, but the `_split()` and `_split_whitespace()` implementation methods are way up around line ~1000. Should we place these altogether?

ConnorGray:
This looks fantastic and is much easier to keep all the context in my head, making review snappy. Thank you so much for doing that Martin 😄

Other than a question about method placement, this looks ready to go.

ConnorGray:
!sync

stdlib/src/collections/string/string_slice.mojo
martinvuyk: I like the idea. IMO we won't need the ability to pass a negative value anymore once #3528 lands

ConnorGray:
Hmm, it seems this is running into Mojo compiler assertion failures in CI internally.

I'm able to reproduce a crash using `$ mojo package ./stdlib/src/` to build this PR. (Curious that the CI here succeeds? That's odd.). The crash I've been seeing looks like this:

```
1.  Crash resolving decl body at loc("collections/string/string.mojo":1445:8)
    >>     fn split(self, sep: StringSlice, maxsplit: Int) raises -> List[String]:
              ^...................................................................
    >>         """Split the string by a separator.
    >> 
    >>         Args:
    >>             sep: The string to split on.
2.  Crash parsing statement at loc("collections/string/string.mojo":1466:9)
    >>         return _to_string_list(self.as_string_slice().split(sep, maxsplit))
               ^..................................................................<
```

Are you able to call these new `split()` functions in any example programs locally, without running into a compiler crash?

I think this is running into the same parameter inference problem I ran into when I recently did some work on `StringSlice.split()`, which led me to use the explicit parameters you [see in the signature here](https://github.com/modular/mojo/pull/4015/files#diff-f8da59dec02a848d700736433c150c14adfea4c79cb453ded5c247c1c597b654L965):

```mojo
    fn split[
        sep_mut: Bool,
        sep_origin: Origin[sep_mut], //,
    ](
        self,
        sep: StringSlice[sep_origin],
        maxsplit: Int = -1,
    ) raises -> List[
        String
    ]:
```

I notice that you still change `StringSlice.split()` to return a `List[StringSlice]` (spelled as `List[Self]`) in this PR — I'm fairly certain that's the cause of the crash.

Originally I thought this PR was writing out each of these overloads as a way you found to workaround that crash, but it appears that may not have been the goal?

My best guess is that this issue occurs when something goes wrong inferring origin parameters when `StringSlice` appears in both as the `sep: StringSlice` argument and in the return `List[StringSlice[..]]`, but I haven't narrowed down a minimal repro myself.

martinvuyk:
@ConnorGray I cant repro the crash

> I notice that you still change StringSlice.split() to return a List[StringSlice] (spelled as List[Self]) in this PR — I'm fairly certain that's the cause of the crash.

AFAIK the `Self` type has the origin bound defined and is not the same as `sume_argument: StringSlice`

> Originally I thought this PR was writing out each of these overloads as a way you found to workaround that crash, but it appears that may not have been the goal?

I didn't experience any crash :man_shrugging: . The `List[Self]` was  mainly because I found `List[StringSlice[origin]]` verbose and I'm lazy

martinvuyk:
@ConnorGray gentle ping and also to ask whether the crash is still there, and if I can help in any way to repro/fix it

JoeLoser:
!sync

JoeLoser:
This is still crashing parameter inference in the compiler internally.  Perhaps we can help it out by explicitly specifying origins etc here.

soraros:
> This is still crashing parameter inference in the compiler internally. Perhaps we can help it out by explicitly specifying origins etc here.

This is quite unfortunate. Should we add a TODO to indicate future cleanups?

martinvuyk:
> This is quite unfortunate. Should we add a TODO to indicate future cleanups?

Probably, but I'm not even sure what is causing the errors or how to replicate them to open an issue

martinvuyk:
@JoeLoser @ConnorGray  I think I found the issue, could any of you test it?
```mojo
fn _to_string_list[
    O: ImmutableOrigin, //
](items: List[StringSlice[O]]) -> List[String]:
```
Changed it to `O: Origin`

JoeLoser:
!sync

JoeLoser:
!sync

martinvuyk:
@JoeLoser I think I found the issue, when a `StringSlice[MutableOrigin]` is passed to `split()` and it returned a `List[StringSlice[MutableOrigin]]` it allowed writing to an already referenced memory location. So I just made it so that it returns a `List[StringSlice[ImmutableOrigin]]` regardless of the initial origin's mutability.

martinvuyk:
@JoeLoser this PR is ready for testing as well

ConnorGray:
!sync

ConnorGray:
!sync

ConnorGray:
!sync

PR #4016, opened 2025-02-25
[stdlib] Make `String.split()` return `List[StringSlice]`

### LINKED ISSUE #3879: [Feature Request] [proposal] Make `String.split()` and `.splitlines()` return `List[StringSlice]`
### Review Mojo's priorities

### What is your request?

Make `String.split()` and `.splitlines()` return `List[StringSlice]`

`String.split()` and `.splitlines()` are normally used as an intermediary step for other logic, I think once we setup some more implicit casting infrastructure the negative impact to existing code would be very minimal. Especially given that edge cases like:
```python
data = "hello world"
lines = data.split(" ")
for i, l in enumerate(lines):
    l += "!"
    lines[i] = l
```
are quite simple to solve if we have an implicit constructor from `List[String].__init__(self, other: List[StringSlice[_]])` and only the line `var lines: List[String] = data.split(" ")` would need to be changed. This would also allow function signatures that receive `List[String]` to be passed a `List[StringSlice]`.

Other than inplace mutation of a string, most code shouldn't be too impacted by this deviation. If anyone can think of more edge cases please do write them bellow, I wouldn't want us to break every codebase out there because of a couple of "innocent" assumptions, and I'm not sure how "bad" it is to deviate from Python in this manner.

### Any other details?

In some benchmarks I've been doing, our current splitlines implementation is around 2x slower than similar code in Python. If the Mojo code is changed to do `.as_string_slice().splitlines()` it's actually 16% faster than Python's implementation. My hypothesis is that it's because strings in python are copy-on-write, and as such when building a string from another they are virtually the same as a `StringSlice`.

martinvuyk:
Make all split functions return `List[StringSlice]`. Part of #3528. Closes #3879. Closes MSTDL-590.

### DIFF
--- a/max/kernels/src/nvml/nvml.mojo
+++ b/max/kernels/src/nvml/nvml.mojo
@@ -12,6 +12,7 @@
 # ===----------------------------------------------------------------------=== #
 """Implements wrappers around the NVIDIA Management Library (nvml)."""
 
+from collections.string.string_slice import _to_string_list
 from os import abort
 from pathlib import Path
 from sys.ffi import _get_dylib_function as _ffi_get_dylib_function
@@ -382,11 +383,10 @@ struct Device(Writable):
                 fn (UnsafePointer[c_char], UInt32) -> Result,
             ]()(driver_version_buffer, UInt32(max_length))
         )
-        var driver_version_list = String(
-            StaticString(unsafe_from_utf8_ptr=driver_version_buffer)
+        var driver_version_list = StaticString(
+            unsafe_from_utf8_ptr=driver_version_buffer
         ).split(".")
-        var driver_version = DriverVersion(driver_version_list)
-        return driver_version
+        return DriverVersion(to_string_list(driver_version_list))
 
     fn _max_clock(self, clock_type: ClockType) raises -> Int:
         var clock = UInt32()
--- a/max/kernels/src/nvml/nvml.mojo
+++ b/max/kernels/src/nvml/nvml.mojo
@@ -386,7 +386,7 @@ struct Device(Writable):
         var driver_version_list = StaticString(
             unsafe_from_utf8_ptr=driver_version_buffer
         ).split(".")
-        return DriverVersion(to_string_list(driver_version_list))
+        return DriverVersion(_to_string_list(driver_version_list))
 
     fn _max_clock(self, clock_type: ClockType) raises -> Int:
         var clock = UInt32()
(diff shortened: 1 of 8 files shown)


martinvuyk:
@JoeLoser this one is ready for testing as well

JoeLoser:
@ConnorGray do you mind taking a look at this, please?  @kartva you may be interested in this space as well given your recent changes to `String.__getitem__`.

JoeLoser (commented):
!sync

JoeLoser:
!sync

JoeLoser (commented):
Don't forget a changelog entry, please!

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2332,7 +2332,7 @@ fn _to_string_list[
 
 
 @always_inline
-fn _to_string_list[O: Origin](items: List[StringSlice[O]]) -> List[String]:
+fn to_string_list[O: Origin, //](items: List[StringSlice[O]]) -> List[String]:
```
JoeLoser: **Suggestion** I don't see a strong motivation for making this public even yet even though there's a few more places in the stdlib using it for now: it's fine for those places to use this private functions until we have better generic algorithms. I'd suggest keeping it private for this PR.

martinvuyk: ok, will do

martinvuyk: updated

max/kernels/src/nvml/nvml.mojo
```diff
@@ -382,11 +383,10 @@ struct Device(Writable):
...
             ]()(driver_version_buffer, UInt32(max_length))
         )
-        var driver_version_list = String(
-            StaticString(unsafe_from_utf8_ptr=driver_version_buffer)
+        var driver_version_list = StaticString(
+            unsafe_from_utf8_ptr=driver_version_buffer
         ).split(".")
-        var driver_version = DriverVersion(driver_version_list)
-        return driver_version
+        return DriverVersion(to_string_list(driver_version_list))
```
JoeLoser:
```suggestion
        return DriverVersion(_to_string_list(driver_version_list))
```

JoeLoser (approved):

JoeLoser:
!sync

PR #4017, opened 2025-02-25
[stdlib] Hristo/foundations for os process module

izo0x90:
Sets up the foundation for implementing the `os/process` module [PR with process module changes](https://github.com/modular/mojo/pull/3998)
  - Add `read_bytes` capability to `FileDecscriptor`
  - Add file descriptor controls function to Libc. bindings
  - Adds vfork, execvp, kill system call utils. to Mojos cLib binds

### DIFF
--- a/mojo/stdlib/src/builtin/file_descriptor.mojo
+++ b/mojo/stdlib/src/builtin/file_descriptor.mojo
@@ -90,11 +90,10 @@ struct FileDescriptor(Writer):
 
     @always_inline
     fn read_bytes(mut self, buffer: Span[mut=True, Byte]) raises -> UInt:
-        """
-        Read a number of bytes from the file.
+        """Read a number of bytes from the file into a buffer.
 
         Args:
-            buffer: Span[Byte] of length n where to store read bytes. n = number of bytes to read.
+            buffer: A `Span[Byte]` to read bytes into. Read up to `len(buffer)` number of bytes.
 
         Returns:
             Actual number of bytes read.
@@ -108,7 +107,7 @@ struct FileDescriptor(Writer):
 
         @parameter
         if os_is_macos() or os_is_linux():
-            read = external_call["read", c_ssize_t](
+            var read = external_call["read", c_ssize_t](
                 self.value, buffer.unsafe_ptr(), len(buffer)
             )
             if read < 0:
--- a/mojo/stdlib/src/builtin/file_descriptor.mojo
+++ b/mojo/stdlib/src/builtin/file_descriptor.mojo
@@ -89,7 +89,7 @@ struct FileDescriptor(Writer):
             )
 
     @always_inline
-    fn read_bytes(mut self, mut buffer: Span[Byte, _]) raises -> UInt:
+    fn read_bytes(mut self, buffer: Span[mut=True, Byte]) raises -> UInt:
         """
         Read a number of bytes from the file.
 
--- a/mojo/stdlib/src/builtin/file_descriptor.mojo
+++ b/mojo/stdlib/src/builtin/file_descriptor.mojo
@@ -98,6 +98,8 @@ struct FileDescriptor(Writer):
 
         Returns:
             Actual number of bytes read.
+        Notes:
+            [Reference](https://pubs.opengroup.org/onlinepubs/9799919799/functions/read.html).
         """
 
         constrained[
--- a/mojo/stdlib/src/builtin/file_descriptor.mojo
+++ b/mojo/stdlib/src/builtin/file_descriptor.mojo
@@ -26,7 +26,7 @@ f.close()
 from sys import os_is_macos, os_is_linux
 from sys._amdgpu import printf_append_string_n, printf_begin
 from sys.ffi import c_ssize_t, external_call
-from sys.info import is_amd_gpu, is_nvidia_gpu
+from sys.info import is_amd_gpu, is_gpu, is_nvidia_gpu
 
 from builtin.io import _printf
 from builtin.os import abort
@@ -89,33 +89,35 @@ struct FileDescriptor(Writer):
             )
 
     @always_inline
-    fn read_bytes(mut self, size: UInt) raises -> List[Byte]:
+    fn read_bytes(mut self, mut buffer: Span[Byte, _]) raises -> UInt:
         """
         Read a number of bytes from the file.
 
         Args:
-            size: Number of bytes to read.
+            buffer: Span[Byte] of length n where to store read bytes. n = number of bytes to read.
 
         Returns:
-          A list of bytes.
+            Actual number of bytes read.
         """
 
+        constrained[
+            not is_gpu(), "`read_bytes()` is not yet implemented for GPUs."
+        ]()
+
         @parameter
-        if is_nvidia_gpu():
-            constrained[False, "Nvidia GPU read bytes not implemented"]()
-            return abort[List[Byte]]()
-        elif is_amd_gpu():
-            constrained[False, "AMD GPU read bytes not implemented"]()
-            return abort[List[Byte]]()
-        elif os_is_macos() or os_is_linux():
-            var ptr = UnsafePointer[Byte].alloc(size)
-
-            read = external_call["read", c_ssize_t](self.value, ptr, size)
-
-            return List[Byte](ptr=ptr, length=read, capacity=read)
+        if os_is_macos() or os_is_linux():
+            read = external_call["read", c_ssize_t](
+                self.value, buffer.unsafe_ptr(), len(buffer)
+            )
+            if read < 0:
+                raise Error("Failed to read bytes.")
+            return read
         else:
-            constrained[False, "Unknown platform read bytes not implemented"]()
-            return abort[List[Byte]]()
+            constrained[
+                False,
+                "`read_bytes()` is not yet implemented for unknown platform.",
+            ]()
+            return abort[UInt]()
 
     fn write[*Ts: Writable](mut self, *args: *Ts):
         """Write a sequence of Writable arguments to the provided Writer.
--- a/mojo/stdlib/src/builtin/file_descriptor.mojo
+++ b/mojo/stdlib/src/builtin/file_descriptor.mojo
@@ -30,6 +30,7 @@ from sys.info import is_amd_gpu, is_nvidia_gpu
 
 from builtin.io import _printf
 from builtin.os import abort
+from collections import List
 from memory import Span, UnsafePointer
 
 
@@ -88,9 +89,9 @@ struct FileDescriptor(Writer):
             )
 
     @always_inline
-    fn read_bytes(mut self, size: Int) raises -> Span[Byte, MutableAnyOrigin]:
+    fn read_bytes(mut self, size: Int) raises -> List[Byte]:
         """
-        Read a Span of bytes from the file.
+        Read a number of bytes from the file.
 
         Args:
             size: Number of bytes to read.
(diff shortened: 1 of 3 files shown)


stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,36 @@ struct FileDescriptor(Writer):
                 written,
             )
 
+    @always_inline
+    fn read_bytes(mut self, size: Int) raises -> List[Byte]:
```
martinvuyk, line 92:
```suggestion
    fn read_bytes(mut self, size: UInt) raises -> List[Byte]:
```

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,36 @@ struct FileDescriptor(Writer):
...
+        elif is_amd_gpu():
+            constrained[False, "AMD GPU read bytes not implemented"]()
+            return abort[List[Byte]]()
+        elif os_is_macos() or os_is_linux():
+            var list = List[Byte](capacity=size)
+            read = external_call["read", Int](
+                self.value, list.unsafe_ptr(), size
+            )
+            list.size = read
+            return list^
```
martinvuyk, line 116: I'm trying to remove all usages of list.size and rename the field in  #3814

Suggestion:
```suggestion
            var ptr = UnsafePointer[Byte].alloc(size)
            var read = external_call["read", c_ssize_t](self.value, ptr, size)
            return List(ptr=ptr, length=read, capacity=read)
```

izo0x90:
@ConnorGray @JoeLoser @owenhilyard Foundational changes broken off into their own branch/ PR. 😃

martinvuyk (commented):
Hi, just 2 comments :)

izo0x90:
> Hi, just 2 comments :)

Awesome 😃, all set.

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,35 @@ struct FileDescriptor(Writer):
...
+            return abort[List[Byte]]()
+        elif is_amd_gpu():
+            constrained[False, "AMD GPU read bytes not implemented"]()
+            return abort[List[Byte]]()
+        elif os_is_macos() or os_is_linux():
+            var ptr = UnsafePointer[Byte].alloc(size)
+
+            read = external_call["read", c_ssize_t](self.value, ptr, size)
+
+            return List[Byte](ptr=ptr, length=read, capacity=read)
```
ConnorGray, line 115:
```suggestion
            return List[Byte](ptr=ptr, length=read, capacity=size)
```

Minor typo here :)

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,35 @@ struct FileDescriptor(Writer):
...
+        if is_nvidia_gpu():
+            constrained[False, "Nvidia GPU read bytes not implemented"]()
+            return abort[List[Byte]]()
+        elif is_amd_gpu():
+            constrained[False, "AMD GPU read bytes not implemented"]()
+            return abort[List[Byte]]()
+        elif os_is_macos() or os_is_linux():
+            var ptr = UnsafePointer[Byte].alloc(size)
+
+            read = external_call["read", c_ssize_t](self.value, ptr, size)
```
ConnorGray, line 113: **Question** I believe an earlier version of this checked the error condition if `read == -1`; might we still want that, or was it removed intentionally (and if so, why)?

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,35 @@ struct FileDescriptor(Writer):
                 written,
             )
 
+    @always_inline
+    fn read_bytes(mut self, size: UInt) raises -> List[Byte]:
```
owenhilyard, line 92: Are we sure we don't want this to be:
`fn read_bytes(mut self, mut buffer: Span[Byte, _]) raises -> UInt`? 

I'm not sure we want to force allocations for basic IO, and this allows partial reads. You can just resize the list.

martinvuyk: Yeah +1, when doing several reads in a loop this will avoid having many allocs for some custom buffered stuff. Also when you just want to write to some other destination directly

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,35 @@ struct FileDescriptor(Writer):
...
+        Read a number of bytes from the file.
+
+        Args:
+            size: Number of bytes to read.
+
+        Returns:
+          A list of bytes.
+        """
+
+        @parameter
```
JoeLoser, line 103: **Suggestion** We can simplify this to be 

```mojo
constrained[not is_gpu(), "`read_bytes()` is not yet implemented for GPUs."]()
# rest of impl
```

stdlib/src/builtin/file_descriptor.mojo
JoeLoser: +1 to not allocating for users on the return type.  Having this just return a `UInt` allows for partial reads which is important too.  This is how we plan on modeling a `Readable` trait in the future too.

When POSIX `read` fails (when it returns `-1`), we can just raise an error based on the errno value.

stdlib/src/sys/_libc.mojo
```diff
@@ -106,6 +106,67 @@ fn dup(oldfd: c_int) -> c_int:
...
+    return external_call["kill", c_int](pid, sig)
+
+
+@always_inline
+fn pipe(fildes: UnsafePointer[c_int]) -> c_int:
+    return external_call["pipe", c_int](fildes)
+
+
+@always_inline
+fn close(fd: c_int) -> c_int:
```
ConnorGray, line 142:
```suggestion
fn close(fd: c_int) -> c_int:
    """[`close()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/close.html)
    — close a file descriptor.
    """
```

stdlib/src/sys/_libc.mojo
```diff
@@ -106,6 +106,67 @@ fn dup(oldfd: c_int) -> c_int:
...
+    alias F_GETFD: c_int = 1
+    alias F_SETFD: c_int = 2
+
+
+struct FcntlFDFlags:
+    alias FD_CLOEXEC: c_int = 1
+
+
+@always_inline
+fn fcntl[*types: Intable](fd: c_int, cmd: c_int, *args: *types) -> c_int:
```
ConnorGray, line 166:
```suggestion
fn fcntl[*types: Intable](fd: c_int, cmd: c_int, *args: *types) -> c_int:
    """[`fcntl()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fcntl.html)
    — file control.
    """
```

stdlib/src/sys/ffi.mojo
```diff
@@ -71,6 +71,9 @@ alias c_float = Float32
 alias c_double = Float64
 """C `double` type."""
 
+alias c_str_ptr = UnsafePointer[c_char]
+"""C `*char` type"""
```
ConnorGray, line 75: **Suggestion** This is a nice alias, but I think it would be better if we included this in a dedicated PR where we could discuss the usage, semantics, and naming. Once minor concern would be that the other `snake_case` items in this file map directly to C types, but this alias does not (there is no `str_ptr` type in C).

I totally understand wanting to provide friendly names for common things like this, however, for better or worse, C never adopted a standard typedef for this, so for the sake of readability of our FFI function signatures, I think we should tread carefully before adopting an alias for this in our own signatures. I'd love to hash that out in a dedicated PR though where we might be able to find a name and semantics that make this worthwhile 🙂  However, I think it's better to do that "on purpose", rather than having it be an incidental part of this larger PR, if that sounds alright with you 🙂

stdlib/src/sys/_libc.mojo
```diff
@@ -106,6 +106,67 @@ fn dup(oldfd: c_int) -> c_int:
     return external_call[name, c_int](oldfd)
 
 
+@always_inline
+fn execvp(file: UnsafePointer[c_char], argv: UnsafePointer[c_str_ptr]) -> c_int:
+    """`execvp` expects that the c_str_ptr array is terminated with a NULL pointer.
```
ConnorGray, line 111:
```suggestion
    """[`execvp`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/exec.html)
    — execute a file.
    
    Args:
        argv: The c_str_ptr array must be terminated with a NULL pointer.
```

ConnorGray (commented):
Hi @izo0x90, thanks for splitting this off, this is great! 🙂

I've left a few minor comments, mostly suggestions that we include links to the POSIX docs where appropriate. Once those are addressed, I'll proceed with merging this in.

In the meantime, I'll `!sync` this preemptively, just to ensure it passes internal CI.

Thanks again for contributing to Mojo 🔥

ConnorGray:
!sync

stdlib/src/builtin/file_descriptor.mojo
izo0x90: Really good point on not allocating , to avoid n+1 allocing!

stdlib/src/sys/ffi.mojo
izo0x90: Ya that makes perfect sense, if we are breaking/ extending convention traceability is key. Not something I had even thought about so thanks. I will make the changes here to just use the inline type for now and open a separate PR for just this alias change once this PR is settled 🔥

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,37 @@ struct FileDescriptor(Writer):
                 written,
             )
 
+    @always_inline
+    fn read_bytes(mut self, mut buffer: Span[Byte, _]) raises -> UInt:
```
martinvuyk, line 92: This doesn't edit the span but the data inside
```suggestion
    fn read_bytes(mut self, buffer: Span[mut=True, Byte]) raises -> UInt:
```

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,37 @@ struct FileDescriptor(Writer):
...
+        """
+        Read a number of bytes from the file.
+
+        Args:
+            buffer: Span[Byte] of length n where to store read bytes. n = number of bytes to read.
```
martinvuyk, line 97:
```suggestion
        """Read a number of bytes from the file into a buffer.

        Args:
            buffer: A `Span[Byte]` to read bytes into. `len(buffer)` = number of bytes to read.
```

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,37 @@ struct FileDescriptor(Writer):
...
+            Actual number of bytes read.
+        """
+
+        constrained[
+            not is_gpu(), "`read_bytes()` is not yet implemented for GPUs."
+        ]()
+
+        @parameter
+        if os_is_macos() or os_is_linux():
+            read = external_call["read", c_ssize_t](
```
martinvuyk, line 109: nitpick: our convention is to use `var`
```suggestion
            var read = external_call["read", c_ssize_t](
```

stdlib/src/builtin/file_descriptor.mojo
```diff
@@ -85,6 +88,37 @@ struct FileDescriptor(Writer):
...
+    fn read_bytes(mut self, mut buffer: Span[Byte, _]) raises -> UInt:
+        """
+        Read a number of bytes from the file.
+
+        Args:
+            buffer: Span[Byte] of length n where to store read bytes. n = number of bytes to read.
+
+        Returns:
+            Actual number of bytes read.
+        """
```
martinvuyk, line 101:
```suggestion
        Notes:
            [Reference](https://pubs.opengroup.org/onlinepubs/9799919799/functions/read.html).
        """
```

martinvuyk (commented):
LGTM

stdlib/src/builtin/file_descriptor.mojo
ConnorGray: Re. Martin's suggestion, I would clarify that `len(buffer)` isn't an exact or minimum number of bytes to read; instead *up to* that many bytes may be read. Something like:

> buffer: A `Span[Byte]` to read bytes into. The number of bytes read will be no more than `len(buffer)`.

ConnorGray:
Awesome, thanks for incorporating the feedback @izo0x90 🙂 I'll go ahead and sync this in.

ConnorGray:
!sync

JoeLoser:
!sync

PR #4022, opened 2025-02-26
[stdlib][proposal] Duration module proposal

bgreni:
A proposal for a `Duration` struct inspired by `std::chrono::duration` from the C++ stdlib

### DIFF
--- a/proposals/duration_module.md
+++ b/proposals/duration_module.md
@@ -1,0 +1,127 @@
+# Duration Module
+
+Brian Grenier, February 21st, 2025
+
+## Motivation
+
+Representing time durations in programming can be a general hassle at best,
+and a source of bugs at worst. Different functions interpret the same plain
+integer value in different ways, and developers must be careful that
+they calculate the intended time interval properly. Consider the following
+example, where you wish to create a recurring function call that is
+executed every 15 minutes.
+
+```mojo
+alias TIMEOUT_IN_MILLISECONDS = 15 * (60 * 1000)
+struct Job:
+    fn set_timeout(t: Int):
+        """
+        Args:
+            t: The timeout duration in milliseconds
+        """
+        ...
+
+def main():
+    var j = Job()
+    j.set_timeout(TIMEOUT_IN_MILLISECONDS)
+```
+
+Now most programmers have done such calculations frequently enough that we do
+not consider this to be such a terrible burden, but we _can_ do better,
+removing the risk entirely. One piece of the C++ standard library that
+was actually done quite well in my opinion is
+[`std::chrono::duration`](https://en.cppreference.com/w/cpp/chrono/duration).
+By encoding the meaning of the numerical value in the type system, we
+can write much safer,
+and easy to understand APIs.
+
+## Proposal
+
+I propose we follow a similar pattern to `std::chrono::duration`, and
+define a `Duration` struct that takes a `Ratio` parameter to denote its
+relation to the base value of one second.
+
+```mojo
+@value
+@register_passable("trivial")
+struct Ratio[N: UInt, D: UInt = 1]:
+    alias Milli = Ratio[1, 1000]
+    pass
+
+alias Seconds = Duration[Ratio[1](), 's']
+alias Milliseconds = Duration[Milli, 'ms']
+alias Minutes = Duration[Ratio[60](), 'm']
+
+@value
+@register_passable("trivial")
+struct Duration[R: Ratio, postfix: StringLiteral='']:
+    var _value: Int
+
+    fn cast[R: Ratio](self) -> Duration[R]:
+        return Duration[R](0)
+
+    fn write_to[W: Writer](self, mut writer: W):
+        writer.write(self.count(), postfix)
+
+    ...
+```
+
+With these pieces we can now convert the above example into something much
+more ergonomic.
+
+```mojo
+struct Job:
+    fn set_timeout(t: Milliseconds):
+        """
+        Args:
+            t: The timeout duration in milliseconds
+        """
+        ...
+
+def main():
+    var j = Job()
+    j.set_timeout(Minutes(15).cast[Ratio.Milli]())
+
+    sleep(Seconds(10))
+```
+
+## Limitations
+
+I've currently identified two limitations when implementing something
+like this, but I believe they can both be easily remedied when the appropriate
+language features are available.
+
+### Inter-ratio arithmetic
+
+In C++ you can do arithmetic on durations of differing ratios
+because the compiler can determine the higher precision timeframe, which will
+be used as the return type of the operation, and the other value will be cast
+down.
+
+```c++
+#include <chrono>
+
+int main()
+{
+
+    using namespace std::chrono;
+
+    seconds s(10);
+    milliseconds m(100);
+
+    std::cout << m + s; // prints 10100ms
+}
+```
+
+Here `s` will be implicitly cast to `milliseconds`. This is facilitated by
+[`std::common_type`](
+    https://en.cppreference.com/w/cpp/chrono/duration/common_type),
+which as far as I am aware, Mojo currently cannot express. So for now we will
+have to stick to doing arithmetic with matching types.
+
+### Generic representation type
+
+`std::chrono::duration` also makes the representation type of the value
+generic, but since Mojo doesn't currently have something like a `Numeric`
+trait, we would either have to stick to using `Int`, or be generic over
+`Scalar`


martinvuyk:
I think this will add some nice type safety on some otherwise very common mistakes when using APIs.

I still have the whole datetime module on my repo waiting for the stdlib to mature more so that the workload isn't too huge. But my main goal with the design is to bundle datetime as 1 and not a timedelta and date object like Python does (there are many other significant design choice diffs). Separating duration out would be like making an independent timedelta (which in the [python stdlib docstrings](https://github.com/python/cpython/blob/main/Lib/_pydatetime.py#L627) says is a delta between datetimes). But if we don't mix it up with datetime I'm all for it.

bgreni:
> I still have the whole datetime module on my repo waiting for the stdlib to mature more so that the workload isn't too huge. But my main goal with the design is to bundle datetime as 1 and not a timedelta and date object like Python does (there are many other significant design choice diffs). Separating duration out would be like making an independent timedelta (which in the [python stdlib docstrings](https://github.com/python/cpython/blob/main/Lib/_pydatetime.py#L627) says is a delta between datetimes). But if we don't mix it up with datetime I'm all for it.

I think `duration` and `datetime` would exist separately from one another, but that's a very important point to consider thank you for bringing it up, I'm curious what the team has to say on that front.

Julian-J-S:
I agree that having something like "Duration" or "TimeDelta" is very valueable and much nicer and safer than throwing around integers without context.

As for the implementation I personally like to keep it easy and intuitive and close to python instead of fancy rust/c style 😆 
In the end the api should be easy to use and understand.
So I would favor something like this:

```mojo
# Duration object with an underlying unit resolution
d = Duration(minutes=15, seconds=30, unit='ms')

# or parameterized
d = Duration['ms'](minutes=15, seconds=30)

use(d)
```

Instead of:

```mojo
struct Duration[R: Ratio, postfix: StringLiteral='']: ... # user questions: Ratio? postfix? I want a "normal" constructor
struct Ratio[N: UInt, D: UInt = 1]: ... # user questions: Why do I need this?

d = Minutes(15) # user questions: How would I create a "complex" duration of 15 minutes and 30 seconds?

use(d.cast[Ratio.Milli]()) # user questions: What is Ratio.Milli? Why do I need to cast?
```

Is there a reason we need to deviate from python and become more complex/fancy? What are the advantaged for the user?
Just my initial thoughts. Maybe you are completely right and your approach is better 😉

bgreni:
@martinvuyk @Julian-J-S 

Exposing the ratio in some way allows users to define arbitrary time intervals. Like this example taken from the C++ docs.

```C++
#include <chrono>
#include <iostream>
 
using namespace std::chrono_literals;
 
template<typename T1, typename T2>
using mul = std::ratio_multiply<T1, T2>;
 
int main()
{
    using microfortnights = std::chrono::duration<float,
        mul<mul<std::ratio<2>, std::chrono::weeks::period>, std::micro>>;
    using nanocenturies = std::chrono::duration<float,
        mul<mul<std::hecto, std::chrono::years::period>, std::nano>>;
    using fps_24 = std::chrono::duration<double, std::ratio<1, 24>>;
 
    std::cout << "1 second is:\n";
 
    // integer scale conversion with no precision loss: no cast
    std::cout << std::chrono::milliseconds(1s).count() << " milliseconds\n"
              << std::chrono::microseconds(1s).count() << " microseconds\n"
              << std::chrono::nanoseconds(1s).count() << " nanoseconds\n";
 
    // integer scale conversion with precision loss: requires a cast
    std::cout << std::chrono::duration_cast<std::chrono::minutes>(1s).count()
              << " minutes\n";
    // alternative to duration_cast:
    std::cout << 1s / 1min << " minutes\n";
 
    // floating-point scale conversion: no cast
    std::cout << microfortnights(1s).count() << " microfortnights\n"
              << nanocenturies(1s).count() << " nanocenturies\n"
              << fps_24(1s).count() << " frames at 24fps\n";
}
```

Most of the time users are just going to use the provided aliases, and I agree we could namespace common ratios into a `Unit` struct to make it more ergonomic so most users won't be exposed to that concept. Personally I've used `std::chrono::duration` many times, and I had no idea it was implemented this way until I looked to see if I could port it to Mojo.

I think Julian's idea is basically just a datetime, and in that case you should just use that.

> `d = Minutes(15) # user questions: How would I create a "complex" duration of 15 minutes and 30 seconds?`

You would do this `d = Minutes(15) + Seconds(30)`, which I mentioned wouldn't be possible in Mojo at the moment due to
lack of compiler support for determining that the return type of that operation should be a `Seconds` type, so initially you couldn't do that. Which I think is ok for now as the idea here is to provide a type safe mechanism for representing time durations that also only need a 
single number to represent them so it can be register passable and performant.

Julian-J-S:
@bgreni thanks for the explanation. I understand from a technical point of view.

I think we are coming from different perspectives.
I come more from a Data Science/Analytics point of view where a Duration is the difference between two DateTimes where a Duration and also be negative and is like "1day 2hours 3minutes" or "11s 22ms 33ns" (all stored as a single Int with a base unit).
You seem to come more from a Frequency/Refresh-Rate point of view which is quite different I guess.

Nevertheless, I think mojo needs to decide on a certain api-style. Yours is rather close to C++/Rust while I think staying closer to python might be better for users.

Maybe one option is to provide both options as constructor
```mojo
# Duration of 1hour 2 minutes
Duration(
    hours=1,
    minutes=2,
)
# or Duration is terms of frequency: 1/24s
Duration(
    N=1,
    D=24
    base_unit="s",
)
```

On the other hand maybe it would be better to have 2 distinct objects for Duration and Frequency that can be converted into each other 🤔 

```mojo
f = Frequency(
    N=1,
    D=24
    base_unit="s",
)
f.to_duration()
```

martinvuyk:
I'll chime in here again, because I have to say I don't like timedelta nor the concept of mixing different units of measurement in the same object. From experience I can also tell you that it complicates the implementation much more, we would need to add overloads for each parametrized unit of measurement to make building e.g. a `Duration[Unit.Minutes](seconds=30)` impossible.

In my [implementation](https://github.com/martinvuyk/forge-tools/tree/main/src/forge_tools/datetime) of datetime I made subtraction return a datetime, not a timedelta object because it makes no sense. On underflow, the object goes to the end datetime of it's given calendar. 

The `Frequency` idea is quite interesting. I see its usefulness in many hardware-related APIs. It would be interesting once the language permits to just make it inherit duration and add a constructor which takes in Hz .

Julian-J-S:
> In my [implementation](https://github.com/martinvuyk/forge-tools/tree/main/src/forge_tools/datetime) of datetime I made subtraction return a datetime, not a timedelta object because it makes no sense. On underflow, the object goes to the end datetime of it's given calendar.

I cannot follow 😮 
The difference between two points in time (DateTime) should be some sort of Duration/TimeDelta with a direction (positive/negative).
(almost) all languages/frameworks/implmentations have this conecpt because it is absolutely neccessary.
some examples:
- 2025-01-20 minus 2025-01-22
  - DateTime: 9999-xxxxx ??? # underflow to the "end datetime" makes no sense
  - Duration: -48hours (physically stored as Int in some time unit)
- 2025-01-22 minus 2025-01-20
  - DateTime: 1970-xxxxx / 0000-00-02 ??? # makes no sense as a DateTime
  - Duration: 48hours (physically stored as Int in some time unit)

Also a Duration between 2 DateTimes is always "absolute". A single quantifyable number.
If you have the result as a DateTime this is NOT true because for example "1year 3month 2 days" is NOT quantifyable in terms of days/seconds. Years have different length/days, even days can have different length (daylight savings) etc. Therefore modelling the "delta" between two points in time as a DateTime is very problematic und impossible for many use cases 🤔

martinvuyk:
> The difference between two points in time (DateTime) should be some sort of Duration/TimeDelta with a direction (positive/negative).

No difference in time can be absolute. It depends on the calendar it is built upon. There is a reason Python's timedelta only considers up to days. Which won't be true for places where the day doesn't last the same as the current rotation speed of the earth.

> - 2025-01-20 minus 2025-01-22
>    - DateTime: 9999-xxxxx ??? # underflow to the "end datetime" makes no sense
>    - Duration: -48hours (physically stored as Int in some time unit)

It's the same thing as with integers, an underflow is the problem of the programmer because that is not how this type is meant to be used. The dates should be converted to a format like a `Duration` (e.g. seconds since unix epoch) to subtract and have a negative result. My implementation has a `.seconds_since_epoch()` which returns the proleptic Gregorian calendar epoch by default (Python's default).

> - 2025-01-22 minus 2025-01-20
>    - DateTime: 1970-xxxxx / 0000-00-02 ??? # makes no sense as a DateTime
>    - Duration: 48hours (physically stored as Int in some time unit)

Idem here. Convert to seconds and subtract.

The main logical problem is that when subtracting two datetimes the concept is this:
(taken from the tests in my repo)
```mojo
def test_add():
    # using python and unix calendar should have no difference in results
    alias pycal = PythonCalendar
    alias unixcal = UTCCalendar
    alias dt = DateTime[iana=False, pyzoneinfo=False, native=False]
    alias TZ = dt._tz
    tz_0_ = TZ("Etc/UTC", 0, 0)
    tz_1 = TZ("Etc/UTC-1", 1, 0)
    tz1_ = TZ("Etc/UTC+1", 1, 0, -1)

    # test february leapyear
    result = dt(2024, 2, 29, tz=tz_0_, calendar=pycal) + dt(
        0, 0, 1, tz=tz_0_, calendar=pycal
    )
    offset_0 = dt(2024, 3, 1, tz=tz_0_, calendar=unixcal)
    offset_p_1 = dt(2024, 3, 1, hour=1, tz=tz_1, calendar=unixcal)
    offset_n_1 = dt(2024, 2, 29, hour=23, tz=tz1_, calendar=unixcal)
    add_seconds = dt(2024, 2, 29, tz=tz_0_, calendar=unixcal).add(
        seconds=24 * 3600
    )
    assert_equal(result, offset_0)
    assert_equal(result, offset_p_1)
    assert_equal(result, offset_n_1)
    assert_equal(result, add_seconds)
```
```mojo
def test_subtract():
    # using python and unix calendar should have no difference in results
    alias pycal = PythonCalendar
    alias unixcal = UTCCalendar
    alias dt = DateTime[iana=False, pyzoneinfo=False, native=False]
    alias TZ = dt._tz
    tz_0_ = TZ("Etc/UTC", 0, 0)
    tz_1 = TZ("Etc/UTC-1", 1, 0)
    tz1_ = TZ("Etc/UTC+1", 1, 0, -1)

    # test february leapyear
    result = dt(2024, 3, 1, tz=tz_0_, calendar=pycal) - dt(
        0, 0, 1, tz=tz_0_, calendar=pycal
    )
    offset_0 = dt(2024, 2, 29, tz=tz_0_, calendar=unixcal)
    offset_p_1 = dt(2024, 2, 29, hour=1, tz=tz_1, calendar=unixcal)
    offset_n_1 = dt(2024, 2, 28, hour=23, tz=tz1_, calendar=unixcal)
    sub_seconds = dt(2024, 3, 1, tz=tz_0_, calendar=unixcal).subtract(days=1)
    assert_equal(result, offset_0)
    assert_equal(result, offset_p_1)
    assert_equal(result, offset_n_1)
    assert_equal(result, sub_seconds)
```

It might seem quite counterintuitive to allow "weird" start and end date calendars. But using this, I actually managed to build a calendar and datetime type which can represent timedelta (that has months and years as well).

```mojo
fn timedelta[
    dst_storage: ZoneStorageDST = ZoneInfoMem32,
    no_dst_storage: ZoneStorageNoDST = ZoneInfoMem8,
    iana: Bool = True,
    pyzoneinfo: Bool = True,
    native: Bool = False,
](
    years: UInt = 0,
    months: UInt = 0,
    days: UInt = 0,
    hours: UInt = 0,
    minutes: UInt = 0,
    seconds: UInt = 0,
    m_seconds: UInt = 0,
    u_seconds: UInt = 0,
    n_seconds: UInt = 0,
    tz: Optional[
        DateTime[
            dst_storage=dst_storage,
            no_dst_storage=no_dst_storage,
            iana=iana,
            pyzoneinfo=pyzoneinfo,
            native=native,
        ]._tz
    ] = None,
) -> DateTime[
    dst_storage=dst_storage,
    no_dst_storage=no_dst_storage,
    iana=iana,
    pyzoneinfo=pyzoneinfo,
    native=native,
] as output:
    """Return a `DateTime` with `ZeroCalendar`.

    Args:
        years: The years.
        months: The months.
        days: The days.
        hours: The hours.
        minutes: The minutes.
        seconds: The seconds.
        m_seconds: The miliseconds.
        u_seconds: The microseconds.
        n_seconds: The nanoseconds.
        tz: The TimeZone for the timedelta object.

    Returns:
        A `DateTime` with a calendar set to using 0000-00-00 as epoch start.
        Beware this `DateTime` kind should only be used for adding/subtracting
        for instances in the same timezone.
    """
    output = __type_of(output)(
        int(years),
        int(months),
        int(days),
        int(hours),
        int(minutes),
        int(seconds),
        int(m_seconds),
        int(u_seconds),
        int(n_seconds),
        tz,
        ZeroCalendar,
    )
```


We could still make it so that the subtraction/addition of two `DateTime` returns a `ZeroCalendar` based `DateTime`. It actually makes more sense as you pointed out. And if a e.g. `PythonCalendar` based `DateTime` gets subtracted a `ZeroCalendar` based datetime, a `DateTime` of `PythonCalendar` type is returned.

Anyway, this is deviating a lot from the proposal at hand. My design for `DateTime` I know will take a lot of proposals and convincing to merge. But it remains a topic for the future.

bgreni:
@Julian-J-S 

The proposed solution is intended more for the systems developer crowd within Mojo.  While Mojo does try to be pythonic, at the end of the day it _is_ at its core a systems language, and often times we have to invite implementation complexity tradeoffs to conserve both flexibility and performance. And since Mojo does/intends to have the compile time programming features to implement something like this, I think we should.

> Maybe one option is to provide both options as constructor
> 
> ```mojo
> # Duration of 1hour 2 minutes
> Duration(
>     hours=1,
>     minutes=2,
> )
> # or Duration is terms of frequency: 1/24s
> Duration(
>     N=1,
>     D=24
>     base_unit="s",
> )
> ```

This kind of design doesn't solve the problem of building a **type safe and performant** API for representing time intervals. Consider this example.

```mojo
fn wait_minutes(d: Duration):
    ...

fn wait_minutes(d: Minutes):
    ...

...

wait_minutes(Duration(milliseconds=10)) # quietly does nothing, or throws an exception
wait_minutes(Milliseconds(10)) # compiler error
```

With a keyword based approach we lose the ability to catch incompatible durations at compile time, and have to settle for either it being a noop, or throwing a runtime error. While in the proposed approach we can tell at compile time that milliseconds cannot be safely implicitly converted to minutes, and we produce a compile time error. We also lose out on the flexibility of defining arbitrary duration types as I mentioned before, and making it a more complex struct at runtime also means it's likely a non-starter for performance sensitive applications.

owenhilyard:
I wonder if, since Mojo is aiming for some scientific use, it would make more sense to do a unified system for handling units of measure, similar to Rust's [uom](https://docs.rs/uom/latest/uom/). Ideally, we would default to using reasonable-size integers, and then have a parameter which lets you use other integers, floats or arbitrary precision numbers.  Getting all of the SI units in there should be a decent "first issue" project if the inital push has the 7 basic SI units + more detail for time. This would allow you to store things in whatever units you have, until such time as you bother to convert it to proper SI units. 

I think we do need a "Numeric" or "Real Number" trait in Mojo to abstract over combinations of dtype + arbitrary precision numbers. 

Ideally, I would like strings kept as far away from the API as possible. I know it's "pythonic", but that tends to turn into a massive mess and ends up requiring rules around whitespace resolution, instead of just using something like a "unit" enum and a magnitude (number or enum). 

I agree `Ratio` is nice to have, but I think we can generalize even more by making everything else able to use the same type aliases as we use for dealing with various kinds of second. 

Formatting should be a separate concern, formatting time is hard, and we should really leave that to a separate issue. Let's figure out how to store the data first. 

I think that the keyword style is going to have extra overhead, which isn't something HPC users would accept. I think this needs to be as compile-time heavy as we can make it so we can use it for correctness. 

@Julian-J-S I think we should have date handling out of this since that means we need to deal with time zones, which is a mess. 

Similarly, @martinvuyk, let's table calendars for now and stick to SI unit durations.

martinvuyk:
> Similarly, @martinvuyk, let's table calendars for now and stick to SI unit durations.

SI units are based on universal constants so I have np. with using them. Hours and hence days aren't always the same though.

> I wonder if, since Mojo is aiming for some scientific use, it would make more sense to do a unified system for handling units of measure, similar to Rust's [uom](https://docs.rs/uom/latest/uom/).

That project is just... wow

Maybe I'm daydreaming but this would just scratch a scientific itch in me:
```mojo
alias kilo = Magnitude[10**3]()
alias Newtons = (kg * m) / s**2

fn does_material_break(material: Material, incident_force: kilo * Newtons) -> Bool:
    return material.breaking_strength < incident_force
```

If we go in the direction of building a whole unit framework, wouldn't that supersede the present proposal?

owenhilyard:
> > Similarly, @martinvuyk, let's table calendars for now and stick to SI unit durations.
> 
> SI units are based on universal constants so I have np. with using them. Hours and hence days aren't always the same though.

That's part of why I'm fine moving to standardize them immediately, we can't "get them wrong" so long as it's all compile time. 

> > I wonder if, since Mojo is aiming for some scientific use, it would make more sense to do a unified system for handling units of measure, similar to Rust's [uom](https://docs.rs/uom/latest/uom/).
> 
> That project is just... wow
> 
> Maybe I'm daydreaming but this would just scratch a scientific itch in me:
> 
> ```mojo
> alias kilo = Magnitude[10**3]()
> alias Newtons = (kg * m) / s**2
> 
> fn does_material_break(material: Material, incident_force: kilo * Newtons) -> Bool:
>     return material.breaking_strength < incident_force
> ```
> 
> If we go in the direction of building a whole unit framework, wouldn't that supersede the present proposal?

I think it would, but I think for scientific usecases, that would be great. If Mojo can advertise "free" built-in dimensional analysis, that would be amazing for a lot of scientists.

bgreni:
@owenhilyard I hadn't thought of that, but I like the direction you're headed with this. Do you have any concrete ideas for how we would alter the present proposal to enable us to build additional modules on top of the foundational pieces that would be added here? I would be happy to work them in here.

I also yearn for a `Numeric` trait or something similar, but since we likely won't get one for a while maybe we should just be generic over `DType` for now?

owenhilyard:
> @owenhilyard I hadn't thought of that, but I like the direction you're headed with this. Do you have any concrete ideas for how we would alter the present proposal to enable us to build additional modules on top of the foundational pieces that would be added here? I would be happy to work them in here.

I think that taking heavy inspiration from `uom` is a good idea. Most of it is type-level stuff that wasn't possible in Rust for a long time, but should be doable in Mojo. The one place where I would deviate would be that how rounding and conversions are handled isn't the best in all scenarios in `uom`, so we might want to look at a way to set a rounding mode or default conversion direction (ex: towards larger precision for floats). Most of that will be in finding a good way to represent the units. For example, do we decompose Volts into `kg⋅m^2⋅s^{−3}⋅A^{−1}`, or do we try to lift that pattern up to volts? I think @rd4com is the resident non-modular expert on type system things like this, so I'd like him to weigh in.

> I also yearn for a `Numeric` trait or something similar, but since we likely won't get one for a while maybe we should just be generic over `DType` for now?

I think it might make sense to create one that unifies `IntLiteral`, `FloatLiteral` and `SIMD`, and we can look at arbitrary precision later. That should be doable as a simple trait, so long as we can live with `Bool` being numeric. A `Numeric` trait is also a good thing to help abstract over block floats.

bgreni:
> For example, do we decompose Volts into `kg⋅m^2⋅s^{−3}⋅A^{−1}`, or do we try to lift that pattern up to volts?

I think maybe the details of implementing future work is a bit out of scope here, but I'll take a look at how they've implemented their time module to ensure we avoid making any decisions that might block future modules. I also fear the team might be less eager to include all of that into the stdlib as there could be a lot of opinions around how it should be done, whereas time is a bit more cut and dry, and I was hoping we could integrate it into existing stdlib APIs, which we couldn't do if I made this an external lib.

martinvuyk:
> I'll take a look at how they've implemented their time module to ensure we avoid making any decisions that might block future modules. I also fear the team might be less eager to include all of that into the stdlib as there could be a lot of opinions around how it should be done

I think I agree with @bgreni after all, maybe adding SI units is too much. I also think duration handling is something that is very much needed in a lot of code out there, and if we do implement something like uom we could try and migrate over to it.

> I think that taking heavy inspiration from uom is a good idea

And whatever can be learnt from uom we can try and use in the implementation of the Duration module first.

rd4com:
Hello, thanks @owenhilyard , i love  parametrizing but not an expert, 
and how things compose up from that first parametrization :+1:

There is the new dependent type improvements and great `FloatLiteral`:
- https://github.com/modular/max/blob/main/mojo/proposals/always_inline_builtin.md
- https://github.com/modular/max/blob/main/mojo/proposals/fixing-simple-literals.md

The idea of a `Duration` system is great,
we can possibly have something with parameters
(example with `FloatLiteral` or `IntLiteral`)

That way, it is possible to switch to big int later ?

Here is a small draft of how user-friendly this could be with the new improvements:
@bgreni, similar ideas for constants:

```mojo
alias Millisecond = Second/1000
alias Second = Duration[1.0]()
alias Minute = Second *60
alias Hour = Minute * 60

@register_passable("trivial")
struct Duration[value:FloatLiteral]:
    @always_inline("builtin")
    fn __init__(out self: Duration[value]): pass
    @always_inline("builtin")
    fn __add__(
        self, 
        other: Duration, 
        out result: Duration[self.value + other.value]
    ): return __type_of(result)()
    @always_inline("builtin")
    fn __mul__(
        self, 
        other: FloatLiteral, 
        out result: Duration[__type_of(self.value * other)()]
    ): return __type_of(result)()
    @always_inline("builtin")
    fn __truediv__(
        self, 
        other: FloatLiteral, 
        out result: Duration[__type_of(self.value / other)()]
    ): return __type_of(result)()
    @always_inline("builtin")
    fn to_hour(self, out ret: __type_of(FloatLiteral[self.value.value]()/FloatLiteral[Hour.value.value]())):
        return __type_of(ret)()
    fn __eq__(self, other: Duration)->Bool: return self.value==other.value
def main():
    x = Millisecond*500+Second*2
    y = Millisecond*1000 + Second + Minute + Hour
    print(y.value)
    print(y.to_hour())
    print((Millisecond*1000) == Second == (Minute/60) == ((Hour/60)/60))
```

If we add a label to `Duration` into the parameters,
it is possible to do more things too :+1:

bgreni:
> Hello, thanks @owenhilyard , i love parametrizing but not an expert, and how things compose up from that first parametrization 👍
> 
> There is the new dependent type improvements and great `FloatLiteral`:
> 
> * https://github.com/modular/max/blob/main/mojo/proposals/always_inline_builtin.md
> * https://github.com/modular/max/blob/main/mojo/proposals/fixing-simple-literals.md
> 
> The idea of a `Duration` system is great, we can possibly have something with parameters (example with `FloatLiteral` or `IntLiteral`)
> 
> That way, it is possible to switch to big int later ?

I don't think this is the right place for that kind of implementation. In this scenario, how would you constrain that a function arg is some number of seconds? The type safety component of this proposal is lost if we can't do that

(that's actually a really cool code snippet though @rd4com :D)

rd4com:
Sorry @bgreni, that make sense, just trying to help :+1: 
it is possible to downcast to seconds on `__iadd__`,
but there is some non-idealities with the "auto" downcast approach:
if we add `GreatYears(1) + Seconds(1) + Millisecond(1)`, it does not fit into the number,

Now i don't know if the idea have huge times values like this,
there must be a way to make this safe if this is a way to go :+1: 

So this is possible: `d = Minutes(15) + Seconds(30)` but not sure how to make it good !

```mojo
@value
struct Duration[unit:Rationable]:
    var value: Int
    fn __add__(lhs: Duration[unit], rhs: Duration[unit],out ret: Self):
        return Self(lhs.value + rhs.value)
    fn __add__(lhs: Duration[Minutes], rhs: Duration[Seconds],out ret: Duration[Seconds]):
        return __type_of(ret)(lhs.value*60 + rhs.value)
    fn __add__(lhs: Duration[Hours], rhs: Duration[Minutes],out ret: Duration[Minutes]):
        return __type_of(ret)(lhs.value*60 + rhs.value)
```

If we could add an `T` to an exiting `Ts`, we could add numbers in a different way!
An `InlinedArray[UInt64]`, each entries of `UInt64` corresponding to `Ts[x]` (`*Ts: Ratios`)
that way, we can have a value of `UInt64` for GreatYear and another for Millisecond for example:

```mojo
@value
struct Duration[*units:Rationable]:
    var values: InlineArray[UInt64, len(VariadicList(units))]
    fn __init__(out self, owned *args: *units):
        self.values  = __type_of(self.values)(uninitialized=True)
        @parameter
        for i in range(args.__len__()):
            self.values[i] = args[i].get_value()
	#fn __add__(lhs: Duration, rhs: Duration, out: Duration[*lhs.units,*rhs.units]):...
```
Then we can add them up only when finished appending new `T` to `Ts`,
and do some checks to make sure the result is correct, for example.
But that's the difficult part, what is the result type ?:
```mojo
    fn to_unit(self, out res: some_type):
        var tmp = some_type(0) # we need a type that make sense
        @parameter
        for i in range(VariadicList(units).__len__()): #iterate values
            # can check if all the values can fit is some_type here
            tmp += (some_type(self.values[i]) * ( 
		# the ratio for the current value
                FloatLiteral[units[i].N_value]() / FloatLiteral[units[i].D_value]()
            ).__float__())
        res = tmp
```

On the other hand, if a function takes `Duration[Minutes,Seconds]`,
this should work similarely than `Variant` :man_shrugging: !

bgreni:
> Sorry @bgreni, that make sense, just trying to help 👍 it is possible to downcast to seconds on `__iadd__`, but there is some non-idealities with the "auto" downcast approach: if we add `GreatYears(1) + Seconds(1) + Millisecond(1)`, it does not fit into the number,

@rd4com  Of course don't apologize, defending the approach I've taken is all a part of the process!

C++ defines the minimum range that each builtin type must be able to represent, and otherwise the risk of overflow is similar to any other integer arithmetic,  and if the lower bounds are well defined like this then users can make an informed decision about whether it fits their use case.

We don't have the compiler support for the auto downcast behaviour so we'll have lots of time to discuss how to handle that, but maybe in the end we leave it to the user to cast them explicitly and throw an error if overflow occurs. 

image

[https://en.cppreference.com/w/cpp/chrono/duration](https://en.cppreference.com/w/cpp/chrono/duration)

ConnorGray:
Hey folks, the stdlib team had a chance to discuss this proposal recently so I wanted to communicate some of our takeaways from that discussion. I think the bottom line was that, while we think this work is incredibly interesting, we're not sure that we have the bandwidth at this time to properly review contributions in a new complex and nuanced domain like this.

As experience in other ecosystems has shown, datetime, duration, timezone and other time-related libraries often evolve significantly from whatever was initially added to a language's standard library.

I personally think Mojo has HUGE 🔥 potential to provide powerful and ergonomic datetime modeling (given the expressivity of our compile-time programming model), but that same flexibility makes me think this is the kind of domain that is best left to bake via healthy exploration of the design space amongst community libraries for the foreseeable future.

I'm going to close this PR, as I think we're not ready to accept contributions in this area, but I encourage you all to use this thread as you see fit to continue organizing work in this area 🙂

bgreni:
@ConnorGray Thanks for update! I will continue working on this within an external package for now.

ConnorGray:
And I'd be remiss if I didn't say THANK YOU to @bgreni for authoring the initial proposal, and to @Julian-J-S, @martinvuyk, @owenhilyard, and @rd4com and anyone else I may have missed in this thread who's engaged in a very interesting discussion already about possibilities in this area 🙂

As I said, we're simply unable to accept formal contributions in this area at the moment, but I want to encourage y'all to keep working together and designing and prototyping and making libraries; that collaborative work is 🔥 to see 🙂

I'm looking forward to hopefully seeing a community-supported library come out of this, so the entire community can benefit from the work and thought and care y'all are putting into this topic 🙂 Modeling unit systems at compile time is a really exciting design space, so I can't wait to see the clever ideas that emerge 😄

PR #4043, opened 2025-03-01
[stdlib] `List.extend` uses `origin` on its `UnsafePointer` logic.

rd4com:
Hello,
this small commit is to integrate the `origin` parameter into the existing implementation.

(during logic on `UnsafePointer` to the `other` arg)

For example, if `other^.__del__()` is added before the loop, an error comes on `move_pointee_into`:
"uninitialized value other"

(preventing moving of values out of an maybe freed pointer)

The `origin` parameter makes sure that `other` is not `__del__` while

we use the `UnsafePointer` to it.

Not sure if needed here, but it feels like a nice thing ! 👍

### DIFF
--- a/mojo/stdlib/src/collections/list.mojo
+++ b/mojo/stdlib/src/collections/list.mojo
@@ -580,16 +580,20 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         other._len = 0
 
         var dest_ptr = self.data + len(self)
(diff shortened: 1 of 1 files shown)


lsh:
This should be safe either way because variables in arguments live through the duration of the function, but it never hurts to be more careful

lsh:
!sync

PR #3344, opened 2024-07-31
[stdlib] Use `is_power_of_two` from `bit` in `simd.mojo`

soraros:

PR #4045, opened 2025-03-02
[stdlib] Constrain SIMD width to be an integer power of 2 in `UnsafePointer.load()`

christianbator:
### Description
`UnsafePointer.load()` currently only constrains `SIMD` load widths to be a positive integer, and since loading bypasses `_simd_construction_checks()`, the `SIMD` vector can have a non-power of 2 size.

### Example
```mojo

fn main():
    var ptr = UnsafePointer[UInt8].alloc(3)
    ptr[0] = 1
    ptr[1] = 2
    ptr[2] = 3

    var simd = ptr.load[width = 3]()
    print(simd)
    
    # Output:
    # [1, 2, 3] <- SIMD size == 3
```

### Solution
Use `_simd_construction_checks()` in `UnsafePointer.load()`.

### DIFF
--- a/mojo/stdlib/src/memory/unsafe_pointer.mojo
+++ b/mojo/stdlib/src/memory/unsafe_pointer.mojo
@@ -31,6 +31,7 @@ from sys.intrinsics import (
 
 from bit import is_power_of_two
 from memory.memory import _free, _malloc
+from builtin.simd import _simd_construction_checks
 
 # ===----------------------------------------------------------------------=== #
 # UnsafePointer
@@ -501,9 +502,7 @@ struct UnsafePointer[
         Returns:
             The loaded value.
         """
-        constrained[
-            is_power_of_two(width), "width must be a power of 2 integer value"
-        ]()
+        _simd_construction_checks[type, width]()
         constrained[
             alignment > 0, "alignment must be a positive integer value"
         ]()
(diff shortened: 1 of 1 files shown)


soraros:
I tried this in #3344, and apparently the SIMD part is not [desirable](https://github.com/modular/max/pull/3344#issuecomment-2263252643).

christianbator:
> I tried this in #3344, and apparently the SIMD part is not [desirable](https://github.com/modular/max/pull/3344#issuecomment-2263252643).

Ah thanks. I can remove the `SIMD` update, but I still think there should be a check in `UnsafePointer` to avoid non-power of 2 loads, right?

JoeLoser (commented):
This works, but instead, WDYT about calling `_simd_construction_checks[type, width]()` instead? Then we'd get this check and others for free (such as as the DType checks for bf16 support and not being invalid)?

christianbator:
> `_simd_construction_checks[type, width]()`

Yep makes sense - updating

JoeLoser:
!sync

JoeLoser (approved):

JoeLoser:
This is hitting some errors internally since we were relying on this behavior.  We'll dig into them at some point soon — nothing required from you at this time 😄

christianbator:
All good! Yeah I had to change my project too that was relying on it 😅

PR #4046, opened 2025-03-02
[stdlib] Optimize compiler times by prevent huge loop unrolling when filling inline arrays.

msaelices:
The `InlineArray(fill=...)` constructor loops for the initialization of all the pointee items in the array in comptime.

This makes the compiler really slow when the size of the inline array is >2k, and even shows a warning if the size is bigger than 64k:

image

IMO, I don't find useful to have 64k `init_pointee_copy` lines when we can unroll by batches, keeping the compiler fast and the actual runtime still fast.


owenhilyard:
Could we make unroll_threshold into a parameter that's defaulted to something a bit more reasonable, like 64? That way people can customize it, since unrolling 1000 iterations is already icache pollution.

mojo/stdlib/src/collections/inline_array.mojo
```diff
@@ -140,8 +140,21 @@ struct InlineArray[
         _inline_array_construction_checks[size]()
         __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))
 
+        alias unroll_threshold = 1000
+        alias unrolled = size // unroll_threshold
+        alias remainder = size % unroll_threshold
+
         @parameter
-        for i in range(size):
+        for i in range(unrolled):
```
owenhilyard: We probably want to unroll the inner loop, not the outer one.

mojo/stdlib/src/collections/inline_array.mojo
```diff
@@ -140,8 +140,21 @@ struct InlineArray[
         _inline_array_construction_checks[size]()
         __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))
 
+        alias unroll_threshold = 1000
+        alias unrolled = size // unroll_threshold
```
owenhilyard: `math.align_down` makes intent more clear.

msaelices:
> Could we make unroll_threshold into a parameter that's defaulted to something a bit more reasonable, like 64? That way people can customize it, since unrolling 1000 iterations is already icache pollution.

Done: https://github.com/msaelices/mojo/commit/edadb9814bbb5d1ae7ab3ac8c22f2c6cfd4e88e9

mojo/stdlib/src/collections/inline_array.mojo
msaelices: Sure: https://github.com/msaelices/mojo/commit/bb401c4700a294a463bd82ab598ee67f465a1d4a

mojo/stdlib/src/collections/inline_array.mojo
msaelices: Done: https://github.com/msaelices/mojo/commit/bb401c4700a294a463bd82ab598ee67f465a1d4a

msaelices:
@owenhilyard  Thanks for the review. Could you please take another look?

mojo/stdlib/src/collections/inline_array.mojo
```diff
@@ -131,7 +132,7 @@ struct InlineArray[
 
     @always_inline
     @implicit
-    fn __init__(out self, fill: Self.ElementType):
+    fn __init__[batch_size: Int = 100](out self, fill: Self.ElementType):
```
owenhilyard: I think this should be a power of 2 in order to keep the chunks aligned well. 32 or 64 is good enough to not bloat the icache but still be fast, especially since processing the loop variable will happen in parallel with the vector ops on any superscalar processor, which is most of them at this point.

owenhilyard: If it's an inline array of bytes, then 100 will need to do some really odd things with instructions, especially on AVX512 since it will move 64, then 32, then 4.

msaelices: Done: https://github.com/msaelices/mojo/commit/212caea18f10592b61de038ff73df46b9af496ec

owenhilyard (approved):

soraros:
FTR, this is another case for open source [`algorithm.functional`](https://docs.modular.com/mojo/stdlib/algorithm/functional/).

JoeLoser:
@weiweichen @npanchen curious to get your thoughts on this PR when you get a chance

JoeLoser:
> FTR, this is another case for open source [`algorithm.functional`](https://docs.modular.com/mojo/stdlib/algorithm/functional/).

Stay tuned my friend 😄

weiweichen:
> @weiweichen @npanchen curious to get your thoughts on this PR when you get a chance

I don't think I have strong opinion on this, you trade compilation time with run time, and sometimes unroll more gives your more performant code, sometimes it doesn't depends on how much more optimization can be achieved with unrolling more.

npanchen:
> @weiweichen @npanchen curious to get your thoughts on this PR when you get a chance

Specifically to this PR having partial unroll is definitely better than full unroll. I'm curious why 64 has been chosen as that unroll factor seems too large, especially if HW has loop stream detector -like feature. @msaelices ?

But generally having any unrolling at compile time on parameterized `fill` function  is a really tricky question. For example, unroll can be harmful when dtype is `Int8` as it's going to disable `memset` recognition that LLVM has: LoopIdiomPass does expect non-unrolled loop and LLVM has no reroll (yet).
Obviously, there're cases when partial unroll is still beneficial. Full unroll when trip count is small is definitely worth it.

owenhilyard:
> > @weiweichen @npanchen curious to get your thoughts on this PR when you get a chance
> 
> Specifically to this PR having partial unroll is definitely better than full unroll. I'm curious why 64 has been chosen as that unroll factor seems too large, especially if HW has loop stream detector -like feature. @msaelices ?

64 was my choice. It's mostly chosen since 64 bytes is 1 AVX512 operation, meaning fast creation of byte arrays initialized to a single value without relying on the compiler to unroll the loop further. That's balanced against larger types, where 64 instances may be multiple instructions to fill per loop iteration. In that case, even 8-byte pointers, one can still expect 8 items per iteration, which is a bit over optimal on most current CPUs since that gets close to bloating icache, but CPUs are also getting deeper with wider decode. If we could have some kind of `TrivialCopy` marker trait, then it would be more doable to differentiate the case of byte copy initialization and needing to do a full copy constructor. Right now, I'm leaning towards expecting mostly `SIMD` sub-types and pointers in inline arrays. 

> But generally having any unrolling at compile time on parameterized `fill` function is a really tricky question. For example, unroll can be harmful when dtype is `Int8` as it's going to disable `memset` recognition that LLVM has: LoopIdiomPass does expect non-unrolled loop and LLVM has no reroll (yet). Obviously, there're cases when partial unroll is still beneficial. Full unroll when trip count is small is definitely worth it.

Does that pass handle 16, 32 and 64 bit values as well? I think the goal for this should aim to be good general purpose perf, since there is an override for the unroll factor as a parameter. If we trip a byte specific optimization but give up vectorized handling of values of other common widths, then it's going to be codebase specific as to whether this is a performance gain. From some test code in godbolt, it looks like code in this pattern can trip the memset detection in some other way, [at least on clang trunk](https://godbolt.org/z/7GvcasGje), but something, either Clang or LLVM, doesn't seem to bother for lower width operations. 

Overall, I'd prefer to have the unroll control in there even if we set it to 1 by default.

npanchen:
> 64 was my choice. It's mostly chosen since 64 bytes is 1 AVX512 operation, meaning fast creation of byte arrays initialized to a single value without relying on the compiler to unroll the loop further. That's balanced against larger types, where 64 instances may be multiple instructions to fill per loop iteration. In that case, even 8-byte pointers, one can still expect 8 items per iteration, which is a bit over optimal on most current CPUs since that gets close to bloating icache, but CPUs are also getting deeper with wider decode. If we could have some kind of `TrivialCopy` marker trait, then it would be more doable to differentiate the case of byte copy initialization and needing to do a full copy constructor. Right now, I'm leaning towards expecting mostly `SIMD` sub-types and pointers in inline arrays.

I see, so the default value is mainly motivated by AVX512 and I do agree with your thoughts about it too. Right, biggest concern is a zoo of HWs in-order, out-of-order, with SIMD, without SIMD etc.  That's where I actually see that compiler can be helpful, unless entire logic that compiler has is going to be replicated here, i.e. "if itarget is <X>, use that SIMD size and that UF".
IMO, partially "fine" solution would be to have `UF = min-known-max-vector-width / sizeof(elt_size)` and rely on a compiler to SLP and unroll after. Or use SIMD, but rely on unrolling after. That's at least what LLVM currently will do when loop vectorizer is disabled. However, we have no loop unroll in LLVM pipeline.

But not to derail conversation too deep into performance which was not the original motivation of this PR, the other angle to look at whether 64 is good is to check current common sizes for this ctor. At least internally I don't see we use `InlineArray` with more than 64 elements, so 64 from our internal uses is safe and is good for me.

> Does that pass handle 16, 32 and 64 bit values as well? 

The pass can only handle them if all value's bytes are identical as it pattern-matches libc's memset
 
> I think the goal for this should aim to be good general purpose perf, since there is an override for the unroll factor as a parameter. If we trip a byte specific optimization but give up vectorized handling of values of other common widths, then it's going to be codebase specific as to whether this is a performance gain. 

I had to be more specific: I'm not advocating for any changes here to get the best performance for target X, rather than saying that we do leave performance on a table for large trip counts which @msaelices has. And that's regardless if `fill` can be pattern-matched to memset or not: memset has generally more efficient implementation (for example, [avx512](https://github.com/bminor/glibc/blob/master/sysdeps/x86_64/multiarch/memset-avx512-no-vzeroupper.S)), non-memset version might not be optimal for hw with scalable vectors. But in order to make it right, we would need many things that are not high priority now.

> From some test code in godbolt, it looks like code in this pattern can trip the memset detection in some other way, [at least on clang trunk](https://godbolt.org/z/7GvcasGje), but something, either Clang or LLVM, doesn't seem to bother for lower width operations.

Not sure I follow.

> Overall, I'd prefer to have the unroll control in there even if we set it to 1 by default.

I do agree with you.

in order to not get lost in lengthy comment: 64 does look safe for our use cases, so from that point the change does look good to me. @JoeLoser

owenhilyard:
> > 64 was my choice. It's mostly chosen since 64 bytes is 1 AVX512 operation, meaning fast creation of byte arrays initialized to a single value without relying on the compiler to unroll the loop further. That's balanced against larger types, where 64 instances may be multiple instructions to fill per loop iteration. In that case, even 8-byte pointers, one can still expect 8 items per iteration, which is a bit over optimal on most current CPUs since that gets close to bloating icache, but CPUs are also getting deeper with wider decode. If we could have some kind of `TrivialCopy` marker trait, then it would be more doable to differentiate the case of byte copy initialization and needing to do a full copy constructor. Right now, I'm leaning towards expecting mostly `SIMD` sub-types and pointers in inline arrays.
> 
> I see, so the default value is mainly motivated by AVX512 and I do agree with your thoughts about it too. Right, biggest concern is a zoo of HWs in-order, out-of-order, with SIMD, without SIMD etc. That's where I actually see that compiler can be helpful, unless entire logic that compiler has is going to be replicated here, i.e. "if itarget is , use that SIMD size and that UF". IMO, partially "fine" solution would be to have `UF = min-known-max-vector-width / sizeof(elt_size)` and rely on a compiler to SLP and unroll after. Or use SIMD, but rely on unrolling after. That's at least what LLVM currently will do when loop vectorizer is disabled. However, we have no loop unroll in LLVM pipeline.

`min-known-max-vector-width` would be NEON/SSE @ 128 bit. I'm not sure I want to go all the way down to that, especially for x86 where avx2 should be somewhat reliably present at this point. This is where I would really like a `Trivial` trait, so we can specialize on that to handle SIMD, pointers and other small structs, then we could easily back off to something more reasonable like 2 or 4 for larger than pointer structs. We could specialize on SIMD and UnsafePointer for now, and leave the rest for later. 

> But not to derail conversation too deep into performance which was not the original motivation of this PR, the other angle to look at whether 64 is good is to check current common sizes for this ctor. At least internally I don't see we use `InlineArray` with more than 64 elements, so 64 from our internal uses is safe and is good for me.
> 
> > Does that pass handle 16, 32 and 64 bit values as well?
> 
> The pass can only handle them if all value's bytes are identical as it pattern-matches libc's memset

Can you think of an architecture where memset is going to be appreciably faster than a series of vectorized stores which know the correct length at compiler time? I can't think of anything except for very large arrays and DMA accelerators. However, there's a non-zero chance that hits Tensor zeroing on newer Intel Xeons, since most of those should have DSA enabled. Given that it can handle 1, 2, 4 or 8 byte patterns with its fill operation, that's probably not something we want to ignore. Does it make sense to push this into an MLIR dialect and let the compiler figure out when DMA accelerators make sense, or does it make more sense to keep that explicit and we can add a parameter flag later?

> > Overall, I'd prefer to have the unroll control in there even if we set it to 1 by default.
> 
> I do agree with you.
> 
> in order to not get lost in lengthy comment: 64 does look safe for our use cases, so from that point the change does look good to me. @JoeLoser

ack

npanchen:
> `min-known-max-vector-width` would be NEON/SSE @ 128 bit. I'm not sure I want to go all the way down to that, especially for x86 where avx2 should be somewhat reliably present at this point. This is where I would really like a `Trivial` trait, so we can specialize on that to handle SIMD, pointers and other small structs, then we could easily back off to something more reasonable like 2 or 4 for larger than pointer structs. We could specialize on SIMD and UnsafePointer for now, and leave the rest for later.

By min-max I was trying to cover RVV that limits minimal known VLEN. For AVX512 min of 512 is still 512.

> Can you think of an architecture where memset is going to be appreciably faster than a series of vectorized stores which know the correct length at compiler time? 

Even for fixed vectors the common technique in memset is to first (ignoring low trip check) process as big chunk as possible -> reduce VL -> memset -> reduce VL -> memset ... avx512 implementation does exactly this. RVV plans to do that, but it's not there yet (not sure why). Obviously, it depends on a size passed to memset: the large it's, more time memset will expose ILP.
Also RVV _could_ be an example, but mostly because SIMD type requires VL=2^n and cannot be scalable. "Could be", because there's no vector version of memset I found. But if there's one, it will be better than default fixed vectorization which suppose to assume VLEN=128.


> However, there's a non-zero chance that hits Tensor zeroing on newer Intel Xeons, since most of those should have DSA enabled. Given that it can handle 1, 2, 4 or 8 byte patterns with its fill operation, that's probably not something we want to ignore.

Looks like I need to read about it as I no longer follow what x86(intel or amd) has these days.

> Does it make sense to push this into an MLIR dialect and let the compiler figure out when DMA accelerators make sense, or does it make more sense to keep that explicit and we can add a parameter flag later?

At least right now the goal is to keep that explicit. Having more examples where performance or usability are bad will be useful to move it into compiler

owenhilyard:
> > `min-known-max-vector-width` would be NEON/SSE @ 128 bit. I'm not sure I want to go all the way down to that, especially for x86 where avx2 should be somewhat reliably present at this point. This is where I would really like a `Trivial` trait, so we can specialize on that to handle SIMD, pointers and other small structs, then we could easily back off to something more reasonable like 2 or 4 for larger than pointer structs. We could specialize on SIMD and UnsafePointer for now, and leave the rest for later.
> 
> By min-max I was trying to cover RVV that limits minimal known VLEN. For AVX512 min of 512 is still 512.

Aha, that didn't jump to mind for me since most of my time is spent with x86_64 and ARM. 

> > Can you think of an architecture where memset is going to be appreciably faster than a series of vectorized stores which know the correct length at compiler time?
> 
> Even for fixed vectors the common technique in memset is to first (ignoring low trip check) process as big chunk as possible -> reduce VL -> memset -> reduce VL -> memset ... avx512 implementation does exactly this. RVV plans to do that, but it's not there yet (not sure why). Obviously, it depends on a size passed to memset: the large it's, more time memset will expose ILP. Also RVV _could_ be an example, but mostly because SIMD type requires VL=2^n and cannot be scalable. "Could be", because there's no vector version of memset I found. But if there's one, it will be better than default fixed vectorization which suppose to assume VLEN=128.

Well, for masked vector ISAs where you can compute a dynamic mask cheaply, (ex: `TRUE_MASK >> (SIMD_BIT_WIDTH - (remaining_bytes * 8)`), wouldn't it make sense to use that mask even for dynamic memset? uiCA is pointing me to that aligned masked store actually being higher throughput than the unaligned unmasked store: [masked](https://uica.uops.info/?code=%20%20%20%20%20%20%20%20movabs%20%20rax%2C%2068719476735%0D%0A%20%20%20%20%20%20%20%20vpxor%20%20%20xmm0%2C%20xmm0%2C%20xmm0%0D%0A%20%20%20%20%20%20%20%20vmovdqu64%20%20%20%20%20%20%20zmmword%20ptr%20%5Brdi%5D%2C%20zmm0%0D%0A%20%20%20%20%20%20%20%20kmovq%20%20%20k1%2C%20rax%0D%0A%20%20%20%20%20%20%20%20vmovdqu8%20%20%20%20%20%20%20%20zmmword%20ptr%20%5Brdi%20%2B%2064%5D%20%7Bk1%7D%2C%20zmm0&syntax=asIntel&uArchs=ICL&tools=uiCA&tools=llvm&alignment=0) [unaligned](https://uica.uops.info/?code=vxorps%20%20xmm0%2C%20xmm0%2C%20xmm0%0D%0Avmovups%20zmmword%20ptr%20%5Brdi%20%2B%2036%5D%2C%20zmm0%0D%0Avmovups%20zmmword%20ptr%20%5Brdi%5D%2C%20zmm0%0D%0A&syntax=asIntel&uArchs=ICL&tools=uiCA&tools=llvm&alignment=0). For power of 2 batch sizes, calculating the mask is an and, a shift, and a subtraction, but it avoids the unaligned store penalty. 

For unmasked vector ISAs (NEON, x86_64 <AVX512, etc), I agree dropping the vector width is a better idea. My hope was that LLVM would do that somewhat reliably in the drain loop since that is fully unrolled. If it won't, then we can always have a set of `@parameter` ifs to make it do that

> > However, there's a non-zero chance that hits Tensor zeroing on newer Intel Xeons, since most of those should have DSA enabled. Given that it can handle 1, 2, 4 or 8 byte patterns with its fill operation, that's probably not something we want to ignore.
> 
> Looks like I need to read about it as I no longer follow what x86(intel or amd) has these days.

[DPDK's API](https://doc.dpdk.org/api/rte__dmadev_8h.html) is something which enough vendors conform to that it might be worthwhile to standardize on that API. 

There's a talk by one of Intel's technical leads on it, [https://www.youtube.com/watch?v=54TCuqW5iF0](https://www.youtube.com/watch?v=54TCuqW5iF0). The talk is mostly for DPDK, but it is generally applicable as an overview to the kind of device. We probably want to lower the base requirements a bit more, since the DPDK API requires something more capable than normal memset, but this is a useful interface to have Mojo plugged into, especially since most DMA accelerators can do device to device copies. This is also part of my motivation for wanting `async memcpy` and `async memset`. 

> > Does it make sense to push this into an MLIR dialect and let the compiler figure out when DMA accelerators make sense, or does it make more sense to keep that explicit and we can add a parameter flag later?
> 
> At least right now the goal is to keep that explicit. Having more examples where performance or usability are bad will be useful to move it into compiler

It's probably something we can program without compiler assistance, although an `if async` would be helpful for determining when we're in a context where a DMA accelerator would be helpful. In code outside of MAX, it might require plumbing in support for specifying a DMA accelerator at compile time if you expect it to be present on the system, or detecting it with the default build settings. DMA devices mostly operate asynchronously, although a few more exotic ones are accessed via instructions that kick off something in the memory controller.

JoeLoser (approved):
LGTM!  Do you mind expanding on the commit message from the PR discussion?  After that, happy to sync this in internally and land it.

Thank you!

msaelices:
> LGTM! Do you mind expanding on the commit message from the PR discussion? After that, happy to sync this in internally and land it.
> 
> Thank you!

I added it to the docstring and as comments here: https://github.com/msaelices/mojo/commit/720e7a23f59e3c7551f6a501061873b2639e5913

@owenhilyard  Please take a look as I am not an expert on that matter.

mojo/stdlib/src/collections/inline_array.mojo
```diff
@@ -164,22 +165,53 @@ struct InlineArray[
...
         Example:
             ```mojo
             var filled = InlineArray[Int, 5](fill=42)  # [42, 42, 42, 42, 42]
+            
+            # For large arrays, consider adjusting batch_size to balance 
+            # compile time and runtime performance:
+            var large = InlineArray[Int, 10000, batch_size=32](fill=0)
             ```
+            
+        Note:
```
JoeLoser:
```suggestion
        Notes:
        
```

JoeLoser (approved):

JoeLoser:
!sync

npanchen:
> Well, for masked vector ISAs where you can compute a dynamic mask cheaply, (ex: `TRUE_MASK >> (SIMD_BIT_WIDTH - (remaining_bytes * 8)`), wouldn't it make sense to use that mask even for dynamic memset? uiCA is pointing me to that aligned masked store actually being higher throughput than the unaligned unmasked store: [masked](https://uica.uops.info/?code=%20%20%20%20%20%20%20%20movabs%20%20rax%2C%2068719476735%0D%0A%20%20%20%20%20%20%20%20vpxor%20%20%20xmm0%2C%20xmm0%2C%20xmm0%0D%0A%20%20%20%20%20%20%20%20vmovdqu64%20%20%20%20%20%20%20zmmword%20ptr%20%5Brdi%5D%2C%20zmm0%0D%0A%20%20%20%20%20%20%20%20kmovq%20%20%20k1%2C%20rax%0D%0A%20%20%20%20%20%20%20%20vmovdqu8%20%20%20%20%20%20%20%20zmmword%20ptr%20%5Brdi%20%2B%2064%5D%20%7Bk1%7D%2C%20zmm0&syntax=asIntel&uArchs=ICL&tools=uiCA&tools=llvm&alignment=0) [unaligned](https://uica.uops.info/?code=vxorps%20%20xmm0%2C%20xmm0%2C%20xmm0%0D%0Avmovups%20zmmword%20ptr%20%5Brdi%20%2B%2036%5D%2C%20zmm0%0D%0Avmovups%20zmmword%20ptr%20%5Brdi%5D%2C%20zmm0%0D%0A&syntax=asIntel&uArchs=ICL&tools=uiCA&tools=llvm&alignment=0). For power of 2 batch sizes, calculating the mask is an and, a shift, and a subtraction, but it avoids the unaligned store penalty.

If you were referring to RISC-V and RVV, there are couple parameters users can play: LMUL and EVL. In simple: LMUL groups vector registers into a single "virtual" vector register, EVL is "effective vector length", i.e. "dynamic mask" you referred to. That is, in this case it's possible to express mem* algorithms as "use LMUL=8 (8 vregs)) -> `use LMUL=4` -> .... That does look counter-intuitive, but some HWs (for example T-Head) isn't quite optimized and latency for LMUL=8 with EVL=1 or EVL=32 (VLEN=128, EEW=32) is identical. That can be addressed by compiler, but better to be by the library.

> For unmasked vector ISAs (NEON, x86_64 <AVX512, etc), I agree dropping the vector width is a better idea. My hope was that LLVM would do that somewhat reliably in the drain loop since that is fully unrolled. If it won't, then we can always have a set of `@parameter` ifs to make it do that

Compiler tries to find a good balance between performance, code size, compile time and frequency. Since LLVM is mostly for C/C++ and people tend to write mem* themselves. Doing similar mutliversioning for mem* by the compiler will increase codesize, increase compile time and add target-dependent redundancy as libc will have it anyway. So it's decided just to pattern-match and rely on libc implementaiton
 

> There's a talk by one of Intel's technical leads on it, https://www.youtube.com/watch?v=54TCuqW5iF0. The talk is mostly for DPDK, but it is generally applicable as an overview to the kind of device. We probably want to lower the base requirements a bit more, since the DPDK API requires something more capable than normal memset, but this is a useful interface to have Mojo plugged into, especially since most DMA accelerators can do device to device copies. This is also part of my motivation for wanting `async memcpy` and `async memset`.

Thanks, I will take a look

owenhilyard:
> > Well, for masked vector ISAs where you can compute a dynamic mask cheaply, (ex: `TRUE_MASK >> (SIMD_BIT_WIDTH - (remaining_bytes * 8)`), wouldn't it make sense to use that mask even for dynamic memset? uiCA is pointing me to that aligned masked store actually being higher throughput than the unaligned unmasked store: [masked](https://uica.uops.info/?code=%20%20%20%20%20%20%20%20movabs%20%20rax%2C%2068719476735%0D%0A%20%20%20%20%20%20%20%20vpxor%20%20%20xmm0%2C%20xmm0%2C%20xmm0%0D%0A%20%20%20%20%20%20%20%20vmovdqu64%20%20%20%20%20%20%20zmmword%20ptr%20%5Brdi%5D%2C%20zmm0%0D%0A%20%20%20%20%20%20%20%20kmovq%20%20%20k1%2C%20rax%0D%0A%20%20%20%20%20%20%20%20vmovdqu8%20%20%20%20%20%20%20%20zmmword%20ptr%20%5Brdi%20%2B%2064%5D%20%7Bk1%7D%2C%20zmm0&syntax=asIntel&uArchs=ICL&tools=uiCA&tools=llvm&alignment=0) [unaligned](https://uica.uops.info/?code=vxorps%20%20xmm0%2C%20xmm0%2C%20xmm0%0D%0Avmovups%20zmmword%20ptr%20%5Brdi%20%2B%2036%5D%2C%20zmm0%0D%0Avmovups%20zmmword%20ptr%20%5Brdi%5D%2C%20zmm0%0D%0A&syntax=asIntel&uArchs=ICL&tools=uiCA&tools=llvm&alignment=0). For power of 2 batch sizes, calculating the mask is an and, a shift, and a subtraction, but it avoids the unaligned store penalty.
> 
> If you were referring to RISC-V and RVV, there are couple parameters users can play: LMUL and EVL. In simple: LMUL groups vector registers into a single "virtual" vector register, EVL is "effective vector length", i.e. "dynamic mask" you referred to. That is, in this case it's possible to express mem* algorithms as "use LMUL=8 (8 vregs)) -> `use LMUL=4` -> .... That does look counter-intuitive, but some HWs (for example T-Head) isn't quite optimized and latency for LMUL=8 with EVL=1 or EVL=32 (VLEN=128, EEW=32) is identical. That can be addressed by compiler, but better to be by the library.

I don't have any RISC-V hardware, much less RVV, so most of my masked vector ISA experience is with SVE and AVX512, but I'm working on fixing that. Some of my experience might break down a bit when VLEN != DLEN. 

> > For unmasked vector ISAs (NEON, x86_64 <AVX512, etc), I agree dropping the vector width is a better idea. My hope was that LLVM would do that somewhat reliably in the drain loop since that is fully unrolled. If it won't, then we can always have a set of `@parameter` ifs to make it do that
> 
> Compiler tries to find a good balance between performance, code size, compile time and frequency. Since LLVM is mostly for C/C++ and people tend to write mem* themselves. Doing similar mutliversioning for mem* by the compiler will increase codesize, increase compile time and add target-dependent redundancy as libc will have it anyway. So it's decided just to pattern-match and rely on libc implementaiton

That makes sense, although a bit annoying for "bare-metal" code.

PR #4051, opened 2025-03-04
[stdlib] Revert the deletion of `StringLiteral` comptime `__mul__`

martinvuyk:
Revert the deletion of `StringLiteral` comptime `__mul__`. As stated in [this comment](https://github.com/modular/max/commit/4a83bf01c3853cbedecaa2e5d90ffbd3acd1f247#r152875958) for commit 4a83bf01c3853cbedecaa2e5d90ffbd3acd1f247, there does exist a need for that method.

Even if we make `StringLiteral` be non-materializable in the future, the multiplication by a compile time known variable should not allocate on the heap and can remain in parameter space, and then be embedded into the binary. Unless I'm wrong on how `String` builds its values from a `StringLiteral` (which AFAIK is allocating on the heap), I think this will have an impact on compile time.

If we do have a mechanism in the future where we can differentiate on compile time known `Int` vs runtime `Int`, then I guess we could delete this method. If I'm wrong on wanting to revert this, I'd like to know. I'd rather this surface to be talked about than remain in obscurity.

### DIFF
--- a/mojo/stdlib/src/builtin/string_literal.mojo
+++ b/mojo/stdlib/src/builtin/string_literal.mojo
@@ -189,6 +189,34 @@ struct StringLiteral(
         """
         self = self + rhs
 
+    @always_inline("nodebug")
+    fn __mul__(self, n: IntLiteral) -> StringLiteral:
+        """Concatenates the string literal `n` times. Can only be evaluated at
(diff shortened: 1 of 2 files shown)


martinvuyk:
Closing in favor of  #4132

PR #4057, opened 2025-03-05
[stdlib] Add `Variant.is_type_supported` method.

rd4com:
Hello,
this commit adds a method to check if a `Variant` supports a type, for example:
```mojo
def MyFunction(mut arg: Variant):
    if arg.is_type_supported[Float64]():
        arg = Float64(1.5)

def main():
    var x = Variant[Int, Float64](1)
    MyFunction(x)
    if x.isa[Float64]():
        print(x[Float64]) # 1.5
```


gryznar:
What's the point of this method? I don't see it from example

mojo/docs/changelog.md
```diff
@@ -156,6 +156,21 @@ what we publish.
   provide high performance for these types and may have missing operations like
   divide, remainder, etc.
 
+- Add `Variant.is_type_supported` method. ([PR #4057](https://github.com/modular/max/pull/4057))
+  Example:
+  
+  ```mojo
+    def MyFunction(mut arg: Variant):
```
gryznar: Function name in Mojo should be snake_cased. I see also more reasons for having it as `fn`

mojo/stdlib/src/utils/variant.mojo
```diff
@@ -390,3 +391,27 @@ struct Variant[*Ts: CollectionElement](
...
                 return i
         return Self._sentinel
+
+    @staticmethod
+    fn is_type_supported[T: CollectionElement]() -> Bool:
+        """Check if a type can be used by the `Variant`.
+
+        Example:
+        ```mojo
+        def MyFunction(mut arg: Variant):
```
gryznar: Same what above

JoeLoser (commented):
Can you help me understand a bit more about why this is useful?

Note that in C++, we have something similar: for checking if any of the non-active types match the asked type (https://en.cppreference.com/w/cpp/utility/variant/holds_alternative).  In mp11, for type list, we have [`boost::mp11::mp_contains`](https://www.boost.org/doc/libs/1_87_0/libs/mp11/doc/html/mp11.html) if you're familiar with that.

soraros:
It can be used in `_get_ptr`. If the team finds this patch useful, we should refactor that part a bit.

rd4com:
Hello, thanks for the reviews :+1: 

Here is the reasoning:

If arg were `Variant[Int, Bool]`, it would be less useful there,
 because the function only takes `*Ts = (Int, Bool)`
 (the programmer that creates the function knows that `arg` conform to this,
 and just assign an `Int()` to it)
 
If arg is just `Variant`, mojo auto-parametrize the function🔥 :
 
```mojo
 def MyFunction(mut arg: Variant)
``` 
 
and we can take any Variant, like if we would:
 
```mojo
 def MyFunction[*Ts: CollectionElement](mut arg: Variant[*Ts])
```
 
 So from within that function, we have to check if the variant support a type,
 before mutating the arg, for example :+1: 
 
The usecase is pretty narrow,
we might need this later for implicit conversion of `arg:T` to `Self` in `Variant.__init__` too
(currently, every `CollectionElement` implicitly comes in and is constrained during `__init__` )

gryznar:
What's the case for naming function "MyFunction" (like struct) and not "my_function" and using `def` instead of `fn`?

`my_function` in this case is also tatally abstract and does not bring any value for understanding

rd4com:
Yep, thanks for reviewing and working together to refine the pr @gryznar :heart_on_fire: :handshake: ,
just waiting for the overall pr idea to be welcomed first,
so that i don't implement the reviews to close the PR later :+1:

gryznar:
Yeah, I see. Sorry for reminding, but this hurts my eyes 😂

martinvuyk (commented):
Hi, I understand the need for this functionality.

But maybe we can try and make it more obvious at first glance
```mojo
    fn can_be[T: CollectionElement]() -> Bool:
        """Check if a type can be stored inside the `Variant`.
```

But I think this would be even better once we have a mechanism to use `__contains__` in parameter space:
```mojo
if Int in Variant[Int, String]:
    print("can be an Int")
```
it would only need `VariadicPack` to have a `__contains__` for types.

I'll play around with the idea and see if we have the type system features or if I manage to force it.

gryznar:
I like naming and future enhancment from @martinvuyk . In his example it is self explanatory

mojo/stdlib/src/utils/variant.mojo
```diff
@@ -76,6 +76,7 @@ struct Variant[*Ts: CollectionElement](
             - This currently does an extra copy/move until we have origins
             - It also temporarily requires the value to be mutable
         - use `set[T](owned new_value: T)` to reset the variant to a new value
+        - use `is_type_supported[T]` to check if the variant support a type
```
JoeLoser:
```suggestion
        - use `is_type_supported[T]` to check if the variant permits the type `T`
```

mojo/stdlib/src/utils/variant.mojo
JoeLoser:
```suggestion
        def my_function(mut arg: Variant):
```

mojo/stdlib/src/utils/variant.mojo
```diff
@@ -390,3 +391,27 @@ struct Variant[*Ts: CollectionElement](
             if _type_is_eq[Ts[i], T]():
                 return i
         return Self._sentinel
+
+    @staticmethod
+    fn is_type_supported[T: CollectionElement]() -> Bool:
+        """Check if a type can be used by the `Variant`.
+
+        Example:
```
JoeLoser, line 395:
```suggestion
        Example:
        
```

JoeLoser, line 395: **Suggestion** Move this after the API docs for parameter/returns sections for consistency with other Mojo code.

mojo/stdlib/test/utils/test_variant.mojo
```diff
@@ -184,6 +184,18 @@ def test_get_returns_mutable_reference():
     assert_equal(v2[String], "something else")
 
 
+def test_is_type_supported():
+    x = Variant[Float64, Int32](Int32(0))
```
JoeLoser:
```suggestion
    var x = Variant[Float64, Int32](Int32(0))
```
as we still use `var` throughout for now.

JoeLoser (approved):
LGTM once the minor style comments are addressed.  Ping me once they're addressed and I'm happy to sync this in and land it.

Thank you!

mojo/stdlib/src/utils/variant.mojo
gryznar: @JoeLoser I think that `my_function` and `arg` do not provide any value in understanding the purpose of this method. Maybe we should pick more self explanatory example?

mojo/docs/changelog.md
gryznar: @rd4com, still relevant. Maybe we can also choose more self explanatory names for arg and func which illustrate usage example better than `my_function` and `arg`?

mojo/stdlib/src/utils/variant.mojo
rd4com: Hello, implemented the suggestion :+1: 

But had to add:
> For example, the `Variant[Int, Bool]` permits `Int` and `Bool`.

So that it ends with an `.` :+1: 

But it does convey too the idea that you suggested in the other review,
WDYT?

JoeLoser:
!sync

PR #4118, opened 2025-03-06
[stdlib] rename `variance` parameter to `std` (standard deviation) 

illiasheshyn:
addresses the following [issue](https://github.com/modular/max/issues/3976) regarding the argument name in `randn_float64()` as well as `randn`. 

The choice to go with standard deviation over variance was made based on the existing `random.gauss` function from Python's standard library and numpy's `random.normal` which both take mean and standard deviation.

Things to consider
 - the new argument name could arguably be more descriptive.
 - the added test does increase the testing time by 2-3 seconds on my machine.

Changes
- renames the argument for `randn_float64()` and `randn`
- extends the corresponding unit tests

### DIFF
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -327,4 +327,4 @@ ctx.enqueue_function(compiled_func, grid_dim=1, block_dim=1)
 ### 🛠️ Fixed
 
 - [#3976](https://github.com/modular/max/issues/3976) The `variance` argument
-  in `random.randn_float64` and `random.randn` has been renamed to `standard_deviation`.
+  in `random.randn_float64` and `random.randn` has been renamed to `standard_deviation` so that values are drawn from the correct distribution.
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -325,3 +325,6 @@ ctx.enqueue_function(compiled_func, grid_dim=1, block_dim=1)
   capacity constructor, or an `InlineArray`.
 
 ### 🛠️ Fixed
+
+- [#3976](https://github.com/modular/max/issues/3976) The `variance` argument
+  in `random.randn_float64` and `random.randn` has been renamed to `standard_deviation`.
--- a/mojo/stdlib/src/random/random.mojo
+++ b/mojo/stdlib/src/random/random.mojo
@@ -181,19 +181,17 @@ fn rand[
         return
 
 
-fn randn_float64(mean: Float64 = 0.0, variance: Float64 = 1.0) -> Float64:
-    """Returns a random double sampled from a Normal(mean, variance) distribution.
+fn randn_float64(mean: Float64 = 0.0, std: Float64 = 1.0) -> Float64:
+    """Returns a random double sampled from a Normal(mean, std) distribution.
 
     Args:
         mean: Normal distribution mean.
-        variance: Normal distribution variance.
+        std: Normal distribution standard deviation.
 
     Returns:
-        A random float64 sampled from Normal(mean, variance).
+        A random float64 sampled from Normal(mean, std).
     """
-    return external_call["KGEN_CompilerRT_NormalDouble", Float64](
-        mean, variance
-    )
+    return external_call["KGEN_CompilerRT_NormalDouble", Float64](mean, std)
 
 
 fn randn[
@@ -202,9 +200,9 @@ fn randn[
     ptr: UnsafePointer[Scalar[type]],
     size: Int,
     mean: Float64 = 0.0,
-    variance: Float64 = 1.0,
+    std: Float64 = 1.0,
 ):
-    """Fills memory with random values from a Normal(mean, variance) distribution.
+    """Fills memory with random values from a Normal(mean, std) distribution.
 
     Constraints:
         The type should be floating point.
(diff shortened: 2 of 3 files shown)


soraros:
Can we move this test to a separate function (for this is a borderline quality test) and use `seed` to reduce the number of iterations (and thus the run time)?

mojo/stdlib/test/random/test_random.mojo
```diff
@@ -68,8 +68,47 @@ def test_random():
...
+    var mean: Float64 = sum / num_samples
+
+    var sum_sq: Float64 = 0.0
+    for sample in samples:
+        sum_sq += (sample[] - mean) * (sample[] - mean)
+
+    var variance = sum_sq / num_samples
+
+    # Calculate absolute differences (errors)
+    var mean_error = mean if mean > 0 else -mean
```
soraros:
```suggestion
    var mean_error = abs(mean)
```
Same below, it's not slower.

mojo/stdlib/test/random/test_random.mojo
```diff
@@ -68,8 +68,47 @@ def test_random():
             ),
         )
 
-    var random_normal = randn_float64(0, 1)
-    # it's quite hard to verify that the values returned are forming a normal distribution
+    # verify `randn_float64` values are normally distributed
+    var samples = List[Float64]()
```
soraros:
```suggestion
    var samples = List[Float64](capacity=num_samples)
```

mojo/stdlib/test/random/test_random.mojo
```diff
@@ -68,8 +68,47 @@ def test_random():
...
+
+    var sum: Float64 = 0.0
+    for sample in samples:
+        sum += sample[]
+
+    var mean: Float64 = sum / num_samples
+
+    var sum_sq: Float64 = 0.0
+    for sample in samples:
+        sum_sq += (sample[] - mean) * (sample[] - mean)
```
soraros:
```suggestion
        sum_sq += (sample[] - mean)**2
```
The compiler is smart enough to see this.

illiasheshyn:
@soraros thanks for the quick review! Do the changes align with what you had in mind?

mojo/stdlib/src/random/random.mojo
```diff
@@ -181,19 +181,17 @@ fn rand[
...
+        std: Normal distribution standard deviation.
 
     Returns:
-        A random float64 sampled from Normal(mean, variance).
+        A random float64 sampled from Normal(mean, std).
     """
-    return external_call["KGEN_CompilerRT_NormalDouble", Float64](
-        mean, variance
-    )
+    return external_call["KGEN_CompilerRT_NormalDouble", Float64](mean, std)
```
martinvuyk: Uhm, can you really just change that? if the underlying function expects a variance and you give it a standard deviation... you'll get another distribution than what you're expecting. You'd have to square the value for this to make sense.

soraros: I thought the bug was that the underlying function expects a std, and the argument was wrongly named?

martinvuyk: My bad, didn't read the linked issue

mojo/stdlib/src/random/random.mojo
```diff
@@ -181,19 +181,17 @@ fn rand[
         return
 
 
-fn randn_float64(mean: Float64 = 0.0, variance: Float64 = 1.0) -> Float64:
-    """Returns a random double sampled from a Normal(mean, variance) distribution.
+fn randn_float64(mean: Float64 = 0.0, std: Float64 = 1.0) -> Float64:
```
JoeLoser:
```suggestion
fn randn_float64(mean: Float64 = 0.0, standard_deviation: Float64 = 1.0) -> Float64:
```

both here and in the docstrings.

mojo/stdlib/src/random/random.mojo
JoeLoser: Yes, it's fine to change this.  Eventually this just calls https://en.cppreference.com/w/cpp/numeric/random/normal_distribution/normal_distribution which accepts `(mean, stddev)` in its constructor arguments.  This is just an outright bug.

mojo/stdlib/src/random/random.mojo
```diff
@@ -202,9 +200,9 @@ fn randn[
     ptr: UnsafePointer[Scalar[type]],
     size: Int,
     mean: Float64 = 0.0,
-    variance: Float64 = 1.0,
+    std: Float64 = 1.0,
```
JoeLoser:
```suggestion
    standard_deviation: Float64 = 1.0,
```

JoeLoser (approved):
Looks good once the minor naming quibble is addressed.  Do you mind also adding a changelog entry [here](https://github.com/modular/max/blob/9c9c0dd869da2df1f3d86320c21aca982e78d793/mojo/docs/changelog.md#standard-library-changes) as well for this bug fix?

I'll fix the internal parameter name to match in `KGEN_CompilerRT_NormalDouble` when I sync this.

Thanks again for fixing this bug!

illiasheshyn:
Do let me know if the changelog entry needs any tweaks at all 👍

mojo/docs/changelog.md
```diff
@@ -325,3 +325,6 @@ ctx.enqueue_function(compiled_func, grid_dim=1, block_dim=1)
   capacity constructor, or an `InlineArray`.
 
 ### 🛠️ Fixed
+
+- [#3976](https://github.com/modular/max/issues/3976) The `variance` argument
+  in `random.randn_float64` and `random.randn` has been renamed to `standard_deviation`.
```
JoeLoser:
```suggestion
  in `random.randn_float64` and `random.randn` has been renamed to `standard_deviation` so that values are drawn from the correct distribution.
```

JoeLoser (approved):

JoeLoser:
!sync

Issue #4119, opened 2025-03-07
[BUG] Numpy ints no longer cast to Mojo


fnands:
### Actual behavior
Up until at least Mojo 24.5 it was possible to cast a Numpy int to Mojo as: 
```
var np = Python.import_module("numpy")
var py_numpy_int = np.int64(42)
var to_mojo = py_numpy_int.__int__() # or just Int(py_numpy_int)
```

Current behaviour is to either return `-1`, regardless of what the integer is, or error with `Unhandled exception caught during execution: an integer is required` which I ran into when trying to update the image parsing library I created. See [here](https://github.com/fnands/mimage/blob/main/tests/testing_utils.mojo) for the test case that fails. 

### Expected behavior

Expected behavior would be that Numpy integers be correctly cast to Mojo.  

It's still possible to get to a Mojo Int, but have to cast to `Float64` first, i.e. `Int(Float64(py_numpy_int))` returns `42` as expected. 

### Steps to reproduce

A minimum reproducible example is: 

```
from python import Python
from testing import assert_equal

var np = Python.import_module("numpy")
var py_numpy_int = np.int64(1)
var mojo_int = Int(1)
assert_equal(Int(py_numpy_int), mojo_int)
```

### System information

- Provide the system information by running `magic info`.
```
Magic version: 0.7.0
System
------------
       Pixi version: 0.41.3
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.8.0=0
                   : __glibc=2.35=0
                   : __archspec=1=skylake
```
- Provide version information for MAX (includes Mojo) by pasting the output of `magic list max`.
`max         25.2.0.dev2025030705  release  9.7 KiB    conda  max`

fnands:
This seems to be the commit that broke it: 
https://github.com/modular/max/commit/dc1dd21770eb0f3c6de566b86263f7eb84841ba8#diff-96fd78e96c5900407cfea53a0952c7b64ce146b5f16e3feccf8558823b691c14

soraros:
Hmmm, that change should be NFC (no functional change). I this issue might relate to #3510, which I never got around to fix.

jackos:
Thanks @fnands I have a patch up to fix this, will let you know when it lands in this repo

Issue #3510, opened 2024-09-20
[BUG] `PythonObject` doesn't handle large `UInt64` correctly


soraros:
### Bug description

As title.

A Mojo `UInt64` value went through the following to reach Python runtime:
- `PythonObject.__init__[dt: DType](Self, SIMD[dt, 1])`
https://github.com/modularml/mojo/blob/61d8208e3beedb4ec4a35b73c91cc9169fa46591/stdlib/src/python/python_object.mojo#L294-L296
- `CPython.toPython(Self, Int)`
https://github.com/modularml/mojo/blob/61d8208e3beedb4ec4a35b73c91cc9169fa46591/stdlib/src/python/_cpython.mojo#L966-L967
- `CPython.PyLong_FromLong(Self, Int)`
https://github.com/modularml/mojo/blob/61d8208e3beedb4ec4a35b73c91cc9169fa46591/stdlib/src/python/_cpython.mojo#L889-L902
- `PyLong_FromLong` has the signature [`PyObject *PyLong_FromLong(long v)`](https://docs.python.org/3/c-api/long.html#c.PyLong_FromLong)

An unwanted bit cast happened on CPython's API boundary.

We should
- Add a branch in `PythonObject.__init__`
- Add a method calling ~~`PyLong_FromUnsignedLong`~~ `PyLong_FromSize_t` on `struct CPython`
- Wire them up

### Steps to reproduce

```mojo
from python import PythonObject

def main():
  n = UInt64(0xFEDCBA0987654321)
  print(PythonObject(n))  # prints -81986143110479071
```

A more peculiar repro:

```mojo
from python import Python

def main():
  py = Python.import_module("builtins")
  a = UInt64(0x1234567890ABCDEF)
  b = UInt64(0xFEDCBA0987654321)
  # bb = py.int(str(b))      # doesn't crash
  a, bb = a, py.int(str(b))  # only crashes when tuple unpacking is used
  print(a + bb)
```

### System information

```shell
Mojo 2024.9.905 (ae516e17) on Docker, Intel Mac
```

dhruvmalik007:
hi @JoeLoser , here is the code that will act patch for this bug : https://github.com/dhruvmalik007/mojo/commit/bf8d434b67f196437adc91e9b1f28e0fdf6f1e01 , let me know if you need more tests / changes required and I will do the needful in order to create PR and merge . 

Thanks

soraros:
[...]

PR #4120, opened 2025-03-07
[stdlib] Add tests for python and numpy casting

fnands:
This PR is related to Bug #4119. 

This is not a fix, but just adds unit tests that will fail until the bug is resolved. 

Only the test `test_python_numpy_int_to_mojo_int` will currently fail, the others were just added for good measure. 

Happy to fix as well, but would need some pointers for where to start looking (~EDIT: although I guess this might be a language issue not stdlib?~)

### DIFF
--- a/mojo/stdlib/test/python/test_python_to_mojo.mojo
+++ b/mojo/stdlib/test/python/test_python_to_mojo.mojo
@@ -57,8 +57,45 @@ fn test_python_to_string() raises:
     assert_true(String(os.environ).startswith("environ({"))
 
 
+fn test_python_to_int() raises:
+    var py_int = PythonObject(1)
+    var mojo_int = Int(1)
+    assert_equal(Int(py_int), mojo_int)
+
+
+fn test_python_to_float() raises:
+    var py_float = PythonObject(1.0)
+    var mojo_float = Float64(1.0)
+    assert_equal(Float64(py_float), mojo_float)
+
+
+fn test_python_to_bool() raises:
+    var py_bool = PythonObject(True)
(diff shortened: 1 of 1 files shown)


fnands:
It seems that [this commit](https://github.com/modular/max/commit/dc1dd21770eb0f3c6de566b86263f7eb84841ba8#diff-96fd78e96c5900407cfea53a0952c7b64ce146b5f16e3feccf8558823b691c14) introduced this change. 


The casting from Python int to numpy int used to happen with 

```
fn PyLong_AsLong(inout self, py_object: PyObjectPtr) -> Int:
        return self.lib.get_function[fn (PyObjectPtr) -> Int]("PyLong_AsLong")(
            py_object
        )
 ```
 
 Which was replaced by: 
 
 ```
 fn PyLong_FromSize_t(inout self, value: c_size_t) -> PyObjectPtr:
    r = self.lib.get_function[fn (c_size_t) -> PyObjectPtr](
        "PyLong_FromSize_t"
  ```
  
  Reverting to PyLong_AsLong seems to resolve the issue, but I feel like I would need some more context for the change. 
  
  @soraros , maybe you can help?

soraros:
Curious. I think it's a ABI related issue. We might hit an overflow in [`PyLong_AsSsize_t`](https://docs.python.org/3/c-api/long.html#c.PyLong_AsSsize_t)

Possibly related #3510.

JoeLoser:
@jackos just fixed this internally and I'll sync your test additions. Thanks for looking into this!

JoeLoser:
!sync

PR #4125, opened 2025-03-08
[stdlib] Add test for `InlineArray` (`sizeof[T]` by `capacity`).

rd4com:
Hello,
this PR adds a small test for `InlineArray`:

```mojo
(sizeof[ElementType]() * capacity) == sizeof[self._array]()
```

### DIFF
--- a/mojo/stdlib/test/collections/test_inline_array.mojo
+++ b/mojo/stdlib/test/collections/test_inline_array.mojo
@@ -226,22 +226,17 @@ fn test_unsafe_ptr() raises:
         assert_equal(arr[i], ptr[i])
 
 
-def test_sizeof_array[
-    current_type: CollectionElement, capacity: Int
-](fill_value: current_type):
+def test_sizeof_array[current_type: CollectionElement, capacity: Int]():
     """Testing if `sizeof` the array equals capacity * `sizeof` current_type.
-
-    Args:
-        fill_value: The value to initialize the array with.
 
     Parameters:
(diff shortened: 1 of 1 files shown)


mojo/stdlib/test/collections/test_inline_array.mojo
```diff
@@ -225,6 +226,25 @@ fn test_unsafe_ptr() raises:
         assert_equal(arr[i], ptr[i])
 
 
+def test_sizeof_array[
+    current_type: CollectionElement, capacity: Int
+](fill_value: current_type):
```
gryznar: Maybe more reasonable would be to just use these values directly instead of parametrizing test function? Right now, there is not a need for sharing and these parameters may be added if need appears

rd4com: That is a good point :+1: 
there was no need to create an instance with `fill_value` !

gryznar (approved):

soraros:
Which one do we prefer for tests these days? `fn raises` or `def`?

JoeLoser (approved):
LGTM, thanks for the additional testing coverage!

JoeLoser:
!sync

JoeLoser:
> Which one do we prefer for tests these days? `fn raises` or `def`?

In general, we still prefer `def`, but it's easy to refactor all of the tests, so I won't reject a contribution for something minor like that.

PR #4126, opened 2025-03-09
[Docs] use `self.cap` in heap allocated array example

illiasheshyn:
Just some minor neat-picks spotted as I was reading through [Life of a value](https://docs.modular.com/mojo/manual/lifecycle/life) page. The corresponding issue https://github.com/modular/max/issues/4127

**Changes**

- 3 cases of implicit conversion are now mentioned, replaces `both` with `all`
- uses `self.cap` to allocate memory (consistent with the previous `HeapArray` example on the same page)

### DIFF
--- a/mojo/docs/manual/functions.mdx
+++ b/mojo/docs/manual/functions.mdx
@@ -368,8 +368,13 @@ def greet(*names: String):
 Inside the function body, the variadic argument is available as an iterable list
 for ease of use. Currently there are some differences in handling the list
 depending on whether the arguments are register-passable types (such as `Int`)
-or memory-only types (such as `String`). TODO: We hope to remove these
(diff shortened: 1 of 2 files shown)


illiasheshyn:
- added a sneaky change to render the `TODO` as a markdown note in the [Homogeneous variadic arguments](https://docs.modular.com/mojo/manual/functions/#homogeneous-variadic-arguments)

feels wrong to lump unrelated changes under this issue, I'm happy to raise the last commit in a separate issue if needed.

JoeLoser (approved):
Thanks for the doc edits!  No worries about lumping in the minor change in `mojo/docs/manual/functions.mdx` in this case 👍

JoeLoser:
!sync

PR #4132, opened 2025-03-09
[stdlib] Fix `StringLiteral` `__mul__` and `__add__` using `is_compile_time()`

martinvuyk:
Fix `StringLiteral` `__mul__` and `__add__` using `is_compile_time()`


mojo/stdlib/src/builtin/string_literal.mojo
```diff
@@ -113,7 +129,13 @@ struct StringLiteral(
         Returns:
             The string concatenated `n` times.
         """
-        return self.as_string_slice() * n
+        if is_compile_time():
+            var concat = ""
+            for _ in range(n):
+                concat = concat + self
```
soraros, line 135: Can we write this?
```suggestion
                concat += self
```

martinvuyk: not anymore since 39b532ae6a0851ecaf5f42150049d134260cb863

lattner:
FYI, I'm actively working on moving StringLiteral to the same model IntLiteral and FloatLiteral are using, which will define these problems away.

JoeLoser:
> FYI, I'm actively working on moving StringLiteral to the same model IntLiteral and FloatLiteral are using, which will define these problems away.

+1 to this.  Given that you're actively working on this, I'm going to just close this PR for now.

lattner:
I'm sorry I didn't mean to get in the way of incremental progress.  Sorry for stepping on toes here, but yes, I'd like us to fix the root cause :)

martinvuyk:
@lattner no worries, it's best to solve it from the root. I'm stoked to see what will become of `StringLiteral`.

PR #4133, opened 2025-03-09
[stdlib] Add compile time versions of utf8 helpers

martinvuyk:
Add compile time versions of utf8 helpers.

Part 1/2 of #4056

PR #4134, opened 2025-03-09
[stdlib] Add compile and runtime utf8 `debug_assert` validation for `StringSlice`

### LINKED ISSUE #3706: [BUG] [stdlib] Intrinsic using function gets stuck
### Bug description

Intrinsic using function gets stuck, not sure if at compile or runtime since I don't understand how mojo package works exactly. But this definitely only happens for `StringLiteral` which is heavily used at comp time

Killed the test after 114 seconds, but still had to kill mojo process with system monitor.

#### Repro
In nightly:
Add to `/utils/string_slice.mojo`: `from ._utf8_validation import _is_valid_utf8`
Uncomment this [line](https://github.com/modularml/mojo/blob/92e2230a832ecda26dfb9132f7b8d970a7a8aafc/stdlib/src/utils/string_slice.mojo#L307) and try running `magic run ./stdlib/scripts/run-tests.sh ./stdlib/test/utils/test_string_slice.mojo`

CC: @JoeLoser I'm not sure whom to CC and I don't want to be spammy :smile: 

### Steps to reproduce

...

### System information

```shell
...
```

martinvuyk:
Add compile and runtime utf8 `debug_assert` validation for `StringSlice`.

Closes #3706
Closes MSTDL-160
Closes MOCO-1525
Closes MSTDL-160


martinvuyk:
I need help debugging this. I have no idea why it keeps stalling. I've tried using `abort()` without a `StringLiteral` message, tried setting the span to empty and running only the `_is_valid_utf8_comptime`. It is most likely that the function `_is_valid_utf8_comptime` is currently not really for compile time, but I don't know how to figure out if that is true. No idea if `SIMD.reduce()` is compile time executable or not, and I don't know what other functions may be giving problems since they are all either arithmetic or logic.

martinvuyk:
I managed to fix the problems for hidden recursive paths and the intrinsics in `SIMD`. PR  #4133 already adds the `_utf8_byte_type` comptime branch, and PR  #4145 adds the paths for `SIMD.reduce_x` functions.

This is still failing in a way in which it makes me suspect something more to do with the compiler front-end is somehow being affected by this:
```
/stdlib/test/builtin/test_print_stderr.mojo:24:10: 
error: invalid call to 'print': argument #3 cannot be converted
from 'StringLiteral' to 'StringSlice[StaticConstantOrigin]'
```
and a check-file test for `/stdlib/test/builtin/test_print_stderr.mojo` is failing

lattner:
Why doesn't `llvm.vector.reduce.and` work at comptime? Can you file a bug with a small example that fails and cc me?

PR #4135, opened 2025-03-10
[stdlib] Add tests for `Tuple` (`__init__`, `__del__`, `a, b = t^`).

rd4com:
Hello,
this PR adds a few tests for `Tuple`:
    - `__init__`
    - `__del__`
    - `x, y = t^` (test disabled, possibly need a fix)

### DIFF
--- a/mojo/stdlib/test/builtin/test_tuple.mojo
+++ b/mojo/stdlib/test/builtin/test_tuple.mojo
@@ -14,7 +14,9 @@
 
 from sys import os_is_macos
 
-from testing import assert_false, assert_true
+from memory import UnsafePointer
+from testing import assert_false, assert_true, assert_equal
+from test_utils import DelRecorder
 
 
 def test_tuple_contains():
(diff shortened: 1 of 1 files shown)


rd4com:
This pr have a new disabled test for `a, b = a_tuple`,
➡️ there might be `len(tuple)` * `__copyinit__` to save by fixing it (if it is a bug) :+1: 

The test:

`Tuple.__init__(c, x, y)` #initially move values into the tuple (one move for each args)
`a, b = c^`	 #move values into theses (one more move per elements, tuple is consumed)

So this is 2 moves per args, and one `__del__` per arg,

Here we have twice `__del__` per args,

so we could have one `__copyinit__` to save on each args :+1: 

---

For example, `Dict` implementation uses [`Tuple return syntax`](https://docs.modular.com/mojo/changelog/#2023-06-14) a lot:
```mojo
fn ()-> (Bool, UInt64, Int)
```

We currently do `a, b = a_tuple` mostly to return trivial types,

but not sure if the `T.__copyinit__` slow downs trivial type!
(`T` of `Tuple` is `CollectionElement`)

rd4com:
Hello, i think this pr fixes it:
[[mojo-lang] Finish 'LHS = RHS' type inference from LHS to RHS.](https://github.com/modular/modular/commit/2f27688eb7c355cda9b1fabc935156fd81cf3822)

Or at least it creates ways to tackle this one :+1:

rd4com:
This patch helps a lot for this:
https://github.com/modular/modular/commit/d906a55678331315edf7c5e5dfa858a80bb4ebbf
[[mojo-lang] Make `var` declarations in functions more friendly](https://github.com/modular/modular/commit/d906a55678331315edf7c5e5dfa858a80bb4ebbf)

(not tested yet if it fixes the move or not)

mojo/stdlib/test/builtin/test_tuple.mojo
```diff
@@ -112,5 +114,56 @@ def test_tuple_contains():
...
+def test_tuple_assigned_into_multiple_variables():
+    var destructor_counter = List[Int]()
+    var pointer_to_destructor_counter = UnsafePointer.address_of(
+        destructor_counter
+    )
+    var t = (
+        DelRecorder(1, pointer_to_destructor_counter),
+        DelRecorder(2, pointer_to_destructor_counter),
+    )
+    a, b = t^
```
soraros, line 159:
```suggestion
    var a, b = t^
```

PR #4138, opened 2025-03-11
[stdlib][feature request] use `Tuple[K,V]` instead of `DictEntry[K,V]`

illiasheshyn:
Signed-off-by: Illia Sheshyn <187636533+illiasheshyn@users.noreply.github.com>

as per https://github.com/modular/max/issues/2554

Changes
- `Dict` type owns computed key hashes
- removes usages and doc mentions of `DictEntry` type

### DIFF
--- a/mojo/stdlib/src/collections/dict.mojo
+++ b/mojo/stdlib/src/collections/dict.mojo
@@ -68,7 +68,7 @@ struct _DictEntryIter[
     dict_origin: Origin[dict_mutability],
     forward: Bool = True,
 ]:
-    """Iterator over immutable DictEntry references.
+    """Iterator over immutable Tuple references.
 
     Parameters:
         dict_mutability: Whether the reference to the dictionary is mutable.
@@ -95,7 +95,7 @@ struct _DictEntryIter[
     @always_inline
     fn __next__(
         mut self,
-    ) -> Pointer[DictEntry[K, V], __origin_of(self.src[]._entries[0].value())]:
+    ) -> Pointer[Tuple[K, V], __origin_of(self.src[]._entries[0].value())]:
         while True:
             var opt_entry_ref = Pointer.address_of(
                 self.src[]._entries[self.index]
@@ -146,8 +146,8 @@ struct _DictKeyIter[
 
     fn __next__(
         mut self,
-    ) -> Pointer[K, __origin_of(self.iter.__next__()[].key)]:
-        return Pointer.address_of(self.iter.__next__()[].key)
+    ) -> Pointer[K, __origin_of(self.iter.__next__()[][0])]:
+        return Pointer.address_of(self.iter.__next__()[][0])
 
     @always_inline
     fn __has_next__(self) -> Bool:
@@ -196,7 +196,7 @@ struct _DictValueIter[
         # Cast through a pointer to grant additional mutability because
         # _DictEntryIter.next erases it.
         return Self.ref_type.address_of(
-            UnsafePointer.address_of(entry_ref[].value).origin_cast[
+            UnsafePointer.address_of(entry_ref[][1]).origin_cast[
                 origin=dict_origin
             ]()[]
         )
@@ -207,50 +207,6 @@ struct _DictValueIter[
 
     fn __len__(self) -> Int:
         return self.iter.__len__()
-
-
-@value
-struct DictEntry[K: KeyElement, V: CollectionElement](
-    CollectionElement, CollectionElementNew
-):
-    """Store a key-value pair entry inside a dictionary.
-
-    Parameters:
-        K: The key type of the dict. Must be Hashable+EqualityComparable.
-        V: The value type of the dict.
-    """
-
-    var key: K
-    """The unique key for the entry."""
-    var value: V
-    """The value associated with the key."""
-
-    fn __init__(out self, owned key: K, owned value: V):
-        """Create an entry from a key and value, computing the hash.
-
-        Args:
-            key: The key of the entry.
-            value: The value of the entry.
-        """
-        self.key = key^
-        self.value = value^
-
-    fn copy(self) -> Self:
-        """Copy an existing entry.
-
-        Returns:
-            A copy of the value.
-        """
-        return self
-
-    fn reap_value(owned self, out result: V):
-        """Take the value from an owned entry.
-
-        Returns:
-            The value of the entry.
-        """
-        result = self.value^
-        __disable_del self
 
 
 alias _EMPTY = -1
@@ -482,7 +438,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
 
     # We use everything available in the list. Which means that
     # len(self._entries) == self._entries.capacity == self._reserved()
-    var _entries: List[Optional[DictEntry[K, V]]]
+    var _entries: List[Optional[Tuple[K, V]]]
 
     # ===-------------------------------------------------------------------===#
     # Life cycle methods
@@ -493,9 +449,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
         """Initialize an empty dictiontary."""
         self.size = 0
         self._n_entries = 0
-        self._entries = Self._new_list[DictEntry[K, V]](
-            Self._initial_reservation
-        )
+        self._entries = Self._new_list[Tuple[K, V]](Self._initial_reservation)
         self._key_hashes = Self._new_list[UInt64](Self._initial_reservation)
         self._index = _DictIndex(len(self._entries))
 
@@ -523,7 +477,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
         )
         self.size = 0
         self._n_entries = 0
-        self._entries = Self._new_list[DictEntry[K, V]](
+        self._entries = Self._new_list[Tuple[K, V]](
             power_of_two_initial_capacity
         )
         self._key_hashes = Self._new_list[UInt64](power_of_two_initial_capacity)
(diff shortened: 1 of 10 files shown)


mojo/stdlib/src/collections/dict.mojo
```diff
@@ -875,14 +836,17 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
             var entry = Pointer.address_of(self._entries[index])
+            var entry_hash = Pointer.address_of(self._key_hashes[index])
             debug_assert(entry[].__bool__(), "entry in index must be full")
+            debug_assert(entry_hash[].__bool__(), "hash must exist for entry")
             var entry_value = entry[].unsafe_take()
             entry[] = None
+            entry_hash[] = None
             self.size -= 1
-            return entry_value^.reap_value()
+            return entry_value[1]
```
illiasheshyn, line 846: I don't actually think this is right, but I haven't learnt Mojo to the extent of refactoring this bit correctly.

gryznar:
What is the impact of this change in terms of performance?

martinvuyk:
> What is the impact of this change in terms of performance?

I also want to know that. FYI public benchmarking script is currently broken, so you will need to use the trick from #4054 for now

illiasheshyn:
I'll run the benchmarks 👍 Does the benchmarking script run with the wrong stdlib without the patch @martinvuyk ?
 
Also what's the approximate runtime to execute all 9 benchmarks? For reference, I'm on M1.

martinvuyk:
> Does the benchmarking script run with the wrong stdlib without the patch @martinvuyk ?

yes It currently uses the packaged stdlib instead of the in-tree one

> Also what's the approximate runtime to execute all 9 benchmarks? For reference, I'm on M1.

No idea, never timed the dict benchmarks. You can try it with fewer dict sizes first :man_shrugging:

illiasheshyn:
After running the benchmarks I no longer think this is a good idea.

main:

```
-- Testing: 1 tests, 1 workers --
Testing: 
PASS: Mojo Standard Library Benchmarks :: collections/bench_dict.mojo (1 of 1)
Exit Code: 0

Command Output (stdout):
--
Running bench_dict_init
Running bench_dict_insert[10]
Running bench_dict_lookup[10]
Running bench_dict_insert[30]
Running bench_dict_lookup[30]
Running bench_dict_insert[50]
Running bench_dict_lookup[50]
Running bench_dict_insert[100]
Running bench_dict_lookup[100]
Running bench_dict_insert[1000]
Running bench_dict_lookup[1000]
Running bench_dict_insert[10000]
Running bench_dict_lookup[10000]
--------------------------------------------------------------
| name                     | met (ms)              | iters   |
--------------------------------------------------------------
| bench_dict_init          | 0.11664069999999999   | 10000   |
| bench_dict_insert[10]    | 0.0023127957514038847 | 518383  |
| bench_dict_lookup[10]    | 0.0004064344464499991 | 2955126 |
| bench_dict_insert[30]    | 0.0015739644349956289 | 762941  |
| bench_dict_lookup[30]    | 0.0003881338210539945 | 3089200 |
| bench_dict_insert[50]    | 0.0020235444374895606 | 592709  |
| bench_dict_lookup[50]    | 0.0004001546309666075 | 2989052 |
| bench_dict_insert[100]   | 0.002014405277635384  | 596934  |
| bench_dict_lookup[100]   | 0.0004345849029736721 | 2759921 |
| bench_dict_insert[1000]  | 0.0017241858515323965 | 696556  |
| bench_dict_lookup[1000]  | 0.0003855355258063531 | 3112315 |
| bench_dict_insert[10000] | 0.0031414045873729016 | 380610  |
| bench_dict_lookup[10000] | 0.0003457493817437144 | 3410398 |
--------------------------------------------------------------

"bench_dict_memory_size[ 10 ]", 576 ,0
"bench_dict_memory_size[ 30 ]", 2160 ,0
"bench_dict_memory_size[ 50 ]", 4272 ,0
"bench_dict_memory_size[ 100 ]", 8752 ,0
"bench_dict_memory_size[ 1000 ]", 69680 ,0
"bench_dict_memory_size[ 10000 ]", 557104 ,0

--
Command Output (stderr):
--
RUN: at line 13: mojo /Users/illiasheshyn/projects/main/max/mojo/stdlib/benchmarks/collections/bench_dict.mojo
+ mojo /Users/illiasheshyn/projects/main/max/mojo/stdlib/benchmarks/collections/bench_dict.mojo

--

********************
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.. 

Testing Time: 755.38s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```

vs the current branch:

```
-- Testing: 1 tests, 1 workers --
Testing: 
PASS: Mojo Standard Library Benchmarks :: collections/bench_dict.mojo (1 of 1)
Exit Code: 0

Command Output (stdout):
--
Running bench_dict_init
Running bench_dict_insert[10]
Running bench_dict_lookup[10]
Running bench_dict_insert[30]
Running bench_dict_lookup[30]
Running bench_dict_insert[50]
Running bench_dict_lookup[50]
Running bench_dict_insert[100]
Running bench_dict_lookup[100]
Running bench_dict_insert[1000]
Running bench_dict_lookup[1000]
Running bench_dict_insert[10000]
Running bench_dict_lookup[10000]
--------------------------------------------------------------
| name                     | met (ms)              | iters   |
--------------------------------------------------------------
| bench_dict_init          | 0.1753140274050007    | 7079    |
| bench_dict_insert[10]    | 0.0030153959273198573 | 401210  |
| bench_dict_lookup[10]    | 0.0005599123669446356 | 2182738 |
| bench_dict_insert[30]    | 0.0022467145505445105 | 532956  |
| bench_dict_lookup[30]    | 0.0005172997543559624 | 2316767 |
| bench_dict_insert[50]    | 0.002724266163198065  | 436439  |
| bench_dict_lookup[50]    | 0.0005368650139242661 | 2238179 |
| bench_dict_insert[100]   | 0.0026650843007495887 | 446378  |
| bench_dict_lookup[100]   | 0.0005356770767102103 | 2241201 |
| bench_dict_insert[1000]  | 0.0025914859420305596 | 460415  |
| bench_dict_lookup[1000]  | 0.0004987785432849825 | 2404506 |
| bench_dict_insert[10000] | 0.00419460891683742   | 283060  |
| bench_dict_lookup[10000] | 0.0004564153054408435 | 2643818 |
--------------------------------------------------------------

"bench_dict_memory_size[ 10 ]", 472 ,0
"bench_dict_memory_size[ 30 ]", 1672 ,0
"bench_dict_memory_size[ 50 ]", 3272 ,0
"bench_dict_memory_size[ 100 ]", 6728 ,0
"bench_dict_memory_size[ 1000 ]", 53320 ,0
"bench_dict_memory_size[ 10000 ]", 426056 ,0

--
Command Output (stderr):
--
RUN: at line 13: mojo /Users/illiasheshyn/projects/main/max/mojo/stdlib/benchmarks/collections/bench_dict.mojo
+ mojo /Users/illiasheshyn/projects/main/max/mojo/stdlib/benchmarks/collections/bench_dict.mojo

--

********************
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.. 

Testing Time: 789.58s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```

Surfacing the `DictEntry` as a `Tuple` does not seem like a beneficial endeavor wrt runtime performance, however, if the goal is to be compatible with Python then I guess we can do that.

gryznar:
I don't think that sacrificing performance for Python compatibility is the way to go for language which aims to squeeze every bit of performance from hardware. I don't see sense of being compatible with Python but at moderate speed (which this compatibility will cause). In such case why to even bother to choose Mojo if there is already C++ for CPU or CUDA for GPU (not speaking about raw assembly) which do not care about Python ballast and will be way faster?

martinvuyk:
I also don't think it's worth sacrificing performance yet. Our dict is still slower than other implementations.

My hope is that in the future we get to control how a type is unpacked with a dunder method, and that way `DictEntry` can be unpacked with tuple syntax while still having `.key` and `.value` and having nice pythonic syntax.

I also think we could make `_DictEntryIter.__next__()` return a `Tuple[K, V]` without a pointer to be able to be unpacked. But I don't know what the mutability and perf. implications inside the loop itself that change would have. I think someone from Modular would have to chime in on this.

gryznar:
I'd love to see priorities as follow:
1. SOTA performance keeping correctness
2. Safety
3. Language consistency (API not being thorn between C, Python and Mojo itself)
4. Python compatibility

soraros:
I'd like to know why there is any performance difference at all. I thought they would generate identical assembly.

illiasheshyn:
> I'd like to know why there is any performance difference at all. I thought they would generate identical assembly.

My understanding so far is the change in data locality. This code is doing an additional index for the key's hash, which is now in its own array, while the key and hash are very frequently read together.

I'll think about it some more, but the problem that I see would be constructing a tuple for the entry if I were to change how those values are stored within `Dict`

soraros:
I see. I missed the part where the hash is no longer stored in the entry. So this PR is actually two changes combined: moving the hash out of line and converting the struct to a tuple. We need to be careful with the former I think.

illiasheshyn:
> We need to be careful with the former I think.

@soraros yeah, that was just an idea I thought was easy to try out. Now that I see the performance impact I don't want to proceed with this change. Hope that the PR assists future discussions in some way.

Issue #4142, opened 2025-03-12
[Feature Request] [stdlib] Make all SIMD functions be able to be run at compile time


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

As title.

### What is your motivation for this change?

#3706,  #4134, and several others

### Any other details?

We also need a way to test this without bloating our test suite (motivation for #4139)

Issue #4139, opened 2025-03-11
[Feature Request] [tooling] A command/tool/way to `mojo test --comptime some_dir_file_or_function`


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

A way to test if a code-path can be run at compile time.

### What is your motivation for this change?

Since our goal is to have a language that can be run seamlessly in both cases. The ideal would be not to have to go down the route of having to color functions as `static`.

### Any other details?

It would also help in not having to duplicate tests for every case using `alias` as well as `var`.

### Other options
- Having a decorator to mark functions to be also tested at compile time

JoeLoser:
I think there's a slight misunderstanding here with this issue getting closed internally: what I understand you're asking for is a way to avoid large duplication in the stdlib testing suite for verifying something works both at comp time and runtime.  In the current status quo, we'd have to duplicate the existing runtime unit tests and mark things with alias everywhere to ensure they also work at comp time.

Is this a correct understanding, @martinvuyk?

martinvuyk:
@JoeLoser yep, exactly. Ideally the solution would also be usable outside the stdlib for the same purpose, since there are many intrinsics we won't explicitly build workarounds for.

ematejska:
Joe said: "Makes sense, I just reopened the issue internally and we can continue discussing. Thanks!"

ematejska:
test sync

ematejska:
test sync 3

PR #4145, opened 2025-03-13
[stdlib] Add comptime `SIMD` code

martinvuyk:
Add comptime `SIMD` code.

Part of #4142


owenhilyard:
@martinvuyk Does it make more sense to start a `sys.intrinsics` package and move llvm intrinsics over there? That way we can provide comptime fallbacks in one place. Ideally, we could try to script generating Mojo functions from the portable llvm intrinsics then manually add fallbacks, possibly with a custom `llvm-tblgen` backend? That should work for the initial push.

martinvuyk:
@owenhilyard I think it's a good idea for other intrinsics. These ones for `SIMD` I think are too specialized to be worth moving and having to setup all parameters etc. Many can also be replaced by just one compile time branch like those for reduce_max

lattner:
Curiosity - why do we need comptime branches for all these?  I think I fixed a bunch of the problems with llvm intrinsics, we should be able to do floor etc now?

martinvuyk:
@lattner I don't know whether to wait for #4139 or to build the test cases for comptime. I'll leave it as draft for now

soraros:
I think we also want some comptime tests.

martinvuyk:
@soraros that's why I asked for https://github.com/modular/modular/issues/4139

PR #4146, opened 2025-03-13
[stdlib] [NFC] Add `debug_assert(location=)`

martinvuyk:
Add `debug_assert(location=)`

### DIFF
--- a/mojo/stdlib/stdlib/builtin/debug_assert.mojo
+++ b/mojo/stdlib/stdlib/builtin/debug_assert.mojo
@@ -67,7 +67,7 @@ fn debug_assert[
     assert_mode: StaticString = "none",
     cpu_only: Bool = False,
     *Ts: Writable,
-](*messages: *Ts):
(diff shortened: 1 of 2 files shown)


soraros:
If I understand correctly, your implementation choice means that `location: _SourceLocation = __call_location()` doesn't work. Does it evaluate `__call_location` too early?

martinvuyk:
@soraros  no idea I also found it weird, I copied the implementation from what `assert_true` and family use exactly just in case.

soraros:
I tried removing `Optional`, but it resulted in `<unknown location in parameter context>:0:0`. So, I suppose the current approach is the right one.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
I'm interested in landing this and we just need to check it doesn't drastically balloon build times.

martinvuyk:
makes sense :+1:

JoeLoser:
!sync

PR #4151, opened 2025-03-15
[stdlib] Add `append()` and `extend()` `unsafe_no_checks` parameter

martinvuyk:
Add `append()` and `extend()` `unsafe_no_checks` parameter.

This gives the best performance without having to use pointers directly when memory was reserved previously.

Merge after #4156 and #4158

### DIFF
--- a/mojo/stdlib/src/collections/list.mojo
+++ b/mojo/stdlib/src/collections/list.mojo
@@ -522,6 +522,7 @@ struct List[T: Copyable & Movable, hint_trivial_type: Bool = False](
         self.data = new_data
         self.capacity = new_capacity
 
+    # FIXME(#4157): remove overload and set unsafe_no_checks default to false
     fn append(mut self, owned value: T):
         """Appends a value to this list.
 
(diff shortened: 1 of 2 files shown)


soraros:
If we do want this change, I'd like to keep `no_check` keyword only.

mojo/stdlib/src/collections/list.mojo
```diff
@@ -503,11 +504,33 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
             If there is no capacity left, resizes to twice the current capacity.
             Except for 0 capacity where it sets 1.
         """
+
         if self._len >= self.capacity:
             self._realloc(self.capacity * 2 | Int(self.capacity == 0))
         self._unsafe_next_uninit_ptr().init_pointee_move(value^)
         self._len += 1
 
+    fn append[*, unsafe_no_checks: Bool](mut self, owned value: T):
```
JoeLoser, line 542: @ConnorGray WDYT about this?  We've talked about this in the past.

PR #4156, opened 2025-03-15
[stdlib] Remove `List.pop()`'s buffer shrinking (realloc) logic

martinvuyk:
Remove `List.pop()`'s buffer shrinking (realloc) logic

This is a split-off from PR #4151. This change is useful to avoid redundant (and surprising) buffer reallocations.


mojo/stdlib/src/collections/list.mojo
```diff
@@ -658,9 +658,13 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         memcpy(self._unsafe_next_uninit_ptr(), value.unsafe_ptr(), len(value))
         self._len += len(value)
 
-    fn pop(mut self, i: Int = -1) -> T:
+    fn pop[*, realloc: Bool = True](mut self, i: Int) -> T:
         """Pops a value from the list at the given index.
 
+        Parameters:
```
soraros: `... defaults to True`.

mojo/stdlib/src/collections/string/string.mojo
```diff
@@ -1021,7 +1021,7 @@ struct String(
             return
         self._buffer.reserve(self.byte_length() + o_len + 1)
         if len(self._buffer) > 0:
-            _ = self._buffer.pop()
+            _ = self._buffer.pop[realloc=False]()
```
JoeLoser: **Suggestion** It would be good to add a comment here as to why it's safe/why we'd want to avoid the reallocation.  It may not be obvious given we do an `extend` call immediately after.

JoeLoser (commented):
I could imagine why this functionality is useful, albeit it seems narrow.  Do you mind sharing more about your use case?

Please also add some unit tests for these API changes where we verify the capacity of the list when reallocation is disabled.

mojo/stdlib/src/collections/string/string.mojo
martinvuyk: To me it's self-evident with the preceding reserve, but ok I'll add a comment. I also don't see any unsafety here, it's just avoiding downsizing the list which is an extra malloc that is redundant here.

martinvuyk:
> I could imagine why this functionality is useful, albeit it seems narrow. Do you mind sharing more about your use case?

I updated the PR description.

> Please also add some unit tests for these API changes where we verify the capacity of the list when reallocation is disabled.

The reallocating by downsizing the list is an extra feature. That "optimization" should be tested rather than the default logic of just deleting the item from the list and returning it.

JoeLoser (commented):
What do you think about removing the reallocation entirely, unconditionally in `pop`?  From a collection perspective, potentially doing a reallocation when we push elements (via `append`) makes sense, but the reallocation during `pop` is surprising.

We could always add a `shrink_to_fit` API or something of the like if desired, separate of making the reallocation conditional.

martinvuyk:
> What do you think about removing the reallocation entirely, unconditionally in `pop`? From a collection perspective, potentially doing a reallocation when we push elements (via `append`) makes sense, but the reallocation during `pop` is surprising.

I also found it surprising, it required some weird bugs for me to realize it was doing that. In some way though `Dict` behaves similarly. The default in Python is this downsizing so it is probably there because of that. IMO this is a default worth deviating from Python.

> We could always add a shrink_to_fit API or something of the like if desired, separate of making the reallocation conditional.

I think I'd rather have that than the current alternative if you agree with changing the default away from Python. I wrote this like I did because I didn't want to change a default

soraros:
+100 to `shrink_to_fit`. IIRC, the current behaviour of automatic `shrink_to_fit` is also mentioned somewhere in `test_list.mojo`, but it has never been explicitly defined through a proper API.

JoeLoser:
> > What do you think about removing the reallocation entirely, unconditionally in `pop`? From a collection perspective, potentially doing a reallocation when we push elements (via `append`) makes sense, but the reallocation during `pop` is surprising.
> 
> I also found it surprising, it required some weird bugs for me to realize it was doing that. In some way though `Dict` behaves similarly. The default in Python is this downsizing so it is probably there because of that. IMO this is a default worth deviating from Python.
> 
> > We could always add a shrink_to_fit API or something of the like if desired, separate of making the reallocation conditional.
> 
> I think I'd rather have that than the current alternative if you agree with changing the default away from Python. I wrote this like I did because I didn't want to change a default

I agree, I'd rather change the behavior of `pop` to not do the reallocation at all, and then add a `shrink_to_fit` API.  I'm fine deviating from whatever Python chooses to do here.

mojo/stdlib/test/collections/test_list.mojo
```diff
@@ -148,12 +148,15 @@ def test_list_pop():
     # test default index as well
     assert_equal(2, list.pop())
     list.append(2)
-    assert_equal(2, list.pop())
+    list_copy = list.copy()
+    assert_equal(2, list.pop[shrink_to_fit=True]())
+    assert_equal(2, list_copy.pop[shrink_to_fit=False]())
 
     # list should be empty now
     assert_equal(0, len(list))
```
gryznar, line 154: You could also add test for length of `list_copy` to make sure, that nothing has changed after providing this parameter

mojo/stdlib/src/collections/list.mojo
```diff
@@ -658,9 +658,14 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         memcpy(self._unsafe_next_uninit_ptr(), value.unsafe_ptr(), len(value))
         self._len += len(value)
 
-    fn pop(mut self, i: Int = -1) -> T:
+    fn pop[*, shrink_to_fit: Bool = False](mut self, i: Int) -> T:
```
JoeLoser: **Suggestion** I think there may have been a slight misunderstanding — sorry if I did not communicate this well.  What I was suggesting was to have:

```mojo
fn pop[mut self, i: Int) -> T:
   # logic for popping element, but **do not do the relloc/shrink to fit behavior**
```

In other words, remove the
```mojo
        if shrink_to_fit:
            if self._len * 4 < self.capacity and self.capacity > 1:
                self._realloc(self.capacity // 2)
```
completely from the `pop` method.  Instead, offer a separate API method: `shrink_to_fit` which can do this.  We could even have that be a separate PR completely and this PR focused on just removing the reallocation completely from `pop` so it stays focused.  Does that make sense/seem reasonable?  This would remove the complexity of a kw-only parameter to the `pop` methods in favor of just dedicated APIs with no surprises which should achieve your original intent/goal.

mojo/stdlib/src/collections/list.mojo
```diff
@@ -677,11 +677,20 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         for j in range(normalized_idx + 1, self._len):
             (self.data + j).move_pointee_into(self.data + j - 1)
         self._len -= 1
-        if self._len * 4 < self.capacity:
-            if self.capacity > 1:
-                self._realloc(self.capacity // 2)
+
         return ret_val^
 
+    fn pop(mut self) -> T:
```
JoeLoser: **Question** What's the motivation for a separate overload?  What was wrong with the one single `pop` function that defaults the index to pop from to be `-1`?

martinvuyk: Less ops for high perf. I assumed the normalization + for loop + amount of moves is worse when just popping the last item and wanting to invoke its destructor right away/store it in another variable. Though I don't know if our compiler is smart enough to optimize every branch away given the default -1.

JoeLoser: Makes sense and seems reasonable.  Let's push that for a separate PR though as it's unrelated to changing the shrink to fit allocating behavior 😄 .  I'm happy to approve and merge that separate PR quickly.

JoeLoser (approved):
LGTM, thank you!

JoeLoser:
!sync

PR #4161, opened 2025-03-17
[stdlib] Upgrade `Dict.popitem()` to move the entry out.

rd4com:
Hello, 
small pr that removes few `__copyinit__`,
this changes the method to move the entry out.

The method is also upgraded to the new out convention.

Implementation is the same as the `pop()` method :+1:

### DIFF
--- a/mojo/stdlib/src/collections/dict.mojo
+++ b/mojo/stdlib/src/collections/dict.mojo
@@ -882,7 +882,7 @@ struct Dict[K: KeyElement, V: CollectionElement](
             return entry_value^.reap_value()
         raise "KeyError"
 
-    fn popitem(mut self) raises -> DictEntry[K, V]:
+    fn popitem(mut self, out ret: DictEntry[K, V]) raises:
         """Remove and return a (key, value) pair from the dictionary. Pairs are returned in LIFO order.
         popitem() is useful to destructively iterate over a dictionary, as often used in set algorithms.
         If the dictionary is empty, calling popitem() raises a KeyError.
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/collections/dict.mojo
```diff
@@ -896,17 +896,23 @@ struct Dict[K: KeyElement, V: CollectionElement](
...
-            val = Optional(item[].value)
-            break
-
-        if key:
-            _ = self.pop(key.value())
-            return DictEntry[K, V](key.value(), val.value())
+        var tmp_iter = reversed(self.items())
+        if tmp_iter.__has_next__():
+            var ptr = tmp_iter.__next__()
+            tmp_iter^.__del__()
```
JoeLoser, line 902: **Suggestion** It may be good to explain why we're doing this here and how it compares to similar things we do in the `pop` method.  WDYT?  A lot of the logic below is the same as in `pop` except for the `reap_value()`, setting the entry explicitly to `None`, etc.

rd4com: Yep, sorry it is difficult to explain in the description :+1: 

but in a mixture of english and mojo:

`popitem()`, previous implementation:
1. get a pointer to the last `entry`, which is (`key`, `value`, `hash`) :thumbsup
2. `__copyinit__` `entry[].key` and `entry[].value` into two `Optional`
3. `_ = self.pop(key_optional[])`:
  `self.pop(k)` does `hash(k)` (we already have it in `popitem()`)
  `__del__` `key`, and move `value` by returning it to `popitem()`.
4. `__del__` the return of `_ = self.pop()` (the `value`)
then we `__copyinit__` the `key` and `value` from the two `entry` `Optional`, 
to move them into a new `entry` to return it
5. asap `__del__` the two `Optional`  (also on `key` and `value`)

So it does many copy and moves on both key and value,
(and also recompute the hash that we already have)

Here instead of doing `self.pop(k)`,
we only do the part of `self.pop` that we need,
and return the original `entry` with no `__copyinit__` on the way,
by initializing the out convention variable (in one `__moveinit__` :+1: )

The new test with the `ObservableDel` asserts that we do no `__copyinit__`,
theses are good `__copyinit__` tests,
because with each `__copyinit__` follows a `__del__` :+1: 

It seem there is no proper way to explain this in the description,
it took a lot of time to write this explanation, refine it as much,
(was working on it before asked, probably not refined enough),
this is way more difficult than to do the fix,
Should some of this be in the description or can we just leave the link

rd4com: Sorry, i forgot the test, working on it !

rd4com: Ok, that makes sense:
The other parallel PR removes all  `__copyinit__` in `_find_index`,
with an `ObservableDel` test and benchmark :+1: 
(https://github.com/modular/max/pull/4162)

`_find_index` is used in `popitem()` and `pop()`,
can't add the test here because `_find_index` is not fixed in this branch yet,
let's think tomorrow, but did tested theses two prs unsplitted.

rd4com: Got the new test! (see last commit), 2 new tests:
1. temporary: assert that `popitem()` does no `__copyinit__`:
	first we get a `_find_index` `__del__` count in to a variable,
	then set the count to 0, do the `popitem().__del__()`, 
 	assert_equal that it is the count +1 !
2. A future test that don't need `_find_index` count:
	We can switch the first one for this one later,
	it is also one that can be enabled to test both PR togethers

NathanSWard:
Closing this PR as now `Dict.popitem` moves the value out instead of copying it.
Thank you for this PR though :)

PR #4162, opened 2025-03-18
[stdlib] Upgrade `Dict._find_index` to use `Pointer`.

rd4com:
Hello,
this pr upgrade the method to use `Pointer` to iterate `self._entries`. 
(removing a `__copyinit__` on `K` and `V`)

The method is used by almost all methods of `Dict`, 
that should be a great speedup with `K`,`V` that allocates.

(PR is a split of a previous PR :+1: )

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_dict.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_dict.mojo
@@ -97,6 +97,32 @@ fn bench_dict_lookup[size: Int](mut b: Bencher) raises:
 
 
 # ===-----------------------------------------------------------------------===#
+# Benchmark Dict Lookup (types that allocated)
+# ===-----------------------------------------------------------------------===#
+@parameter
+fn bench_dict_lookup_types_that_allocates[size: Int](mut b: Bencher) raises:
+    """Lookup size items."""
+    dict = Dict[String, String]()
+    keys = List[String](capacity=size)
+    for i in range(size):
+        keys.append(String(i))
+    for i in range(size):
+        dict[keys[i]] = keys[i]
+
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        res = 0
+        for i in range(size):
+            res += len(dict[keys[i]])
+        keep(res)
+
+    b.iter[call_fn]()
+    keep(Bool(dict))
+    keep(Bool(keys))
+
+
+# ===-----------------------------------------------------------------------===#
 # Benchmark Dict Memory Footprint
 # ===-----------------------------------------------------------------------===#
 
@@ -138,6 +164,11 @@ def main():
         )
         m.bench_function[bench_dict_lookup[size]](
             BenchId(String("bench_dict_lookup[", size, "]"))
+        )
+        m.bench_function[bench_dict_lookup_types_that_allocates[size]](
+            BenchId(
+                String("bench_dict_lookup_types_that_allocates[", size, "]")
+            )
         )
 
     m.dump_report()
--- a/mojo/stdlib/benchmarks/collections/bench_dict.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_dict.mojo
@@ -102,8 +102,8 @@ fn bench_dict_lookup[size: Int](mut b: Bencher) raises:
 @parameter
 fn bench_dict_lookup_types_that_allocates[size: Int](mut b: Bencher) raises:
     """Lookup size items."""
-    dict = Dict[String, String]()
-    keys = List[String](capacity=size)
+    var dict = Dict[String, String]()
+    var keys = List[String](capacity=size)
     for i in range(size):
         keys.append(String(i))
     for i in range(size):
@@ -112,7 +112,7 @@ fn bench_dict_lookup_types_that_allocates[size: Int](mut b: Bencher) raises:
     @always_inline
     @parameter
     fn call_fn() raises:
-        res = 0
+        var res = 0
         for i in range(size):
             res += len(dict[keys[i]])
         keep(res)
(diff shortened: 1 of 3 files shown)


gryznar:
I am curious about the numbers. Maybe it will be worthy to run benchmarks?

rd4com:
Just run this simple `main.mojo` on the branch:

Around 7x-11x faster for `__getitem__`:
(memory types)

```mojo
from time import monotonic
from collections import Dict
def main():
    x = Dict[String, String]()
    y = List[String]()
    alias size = 100000
    for i in range(size): y.append(String(i))
    for i in range(size): x[y[i]] = y[i]

    res = 0
    stop =monotonic()
    start=monotonic()
    for i in range(size): res+= len(x[y[i]])
    stop = monotonic()
    print(res, stop-start)
    y^.__del__()
    x^.__del__()
```

`99394472`
`8137627`

See this [PR](https://github.com/modular/max/pull/4149 ) to run mojo with current branch :+1: 
(updated `development.md`)

Just added a benchmark to `bench_dict.mojo` (last commit),
maybe we have to wait for that pr that fixes benchmark get merged to run it.

gryznar:
Wow, really impressive :)

mojo/stdlib/src/collections/dict.mojo
```diff
@@ -1043,10 +1043,11 @@ struct Dict[K: KeyElement, V: CollectionElement](
             elif index == Self.REMOVED:
                 pass
             else:
-                var entry = self._entries[index]
-                debug_assert(entry.__bool__(), "entry in index must be full")
-                if hash == entry.value().hash and key == entry.value().key:
-                    return (True, slot, index)
+                var entry = Pointer.address_of(self._entries[index])
+                debug_assert(entry[].__bool__(), "entry in index must be full")
+                if hash == entry[].unsafe_value().hash:
```
soraros, line 1047: Why do we need to rewrite this part (2 `if`s, use `__eq__` explicitly)?

mojo/stdlib/src/collections/dict.mojo
```diff
@@ -1043,10 +1043,11 @@ struct Dict[K: KeyElement, V: CollectionElement](
             elif index == Self.REMOVED:
                 pass
             else:
-                var entry = self._entries[index]
-                debug_assert(entry.__bool__(), "entry in index must be full")
-                if hash == entry.value().hash and key == entry.value().key:
-                    return (True, slot, index)
+                var entry = Pointer.address_of(self._entries[index])
+                debug_assert(entry[].__bool__(), "entry in index must be full")
```
soraros, line 1046: Nit:
```suggestion
                debug_assert(Bool(entry[]), "entry in index must be full")
```

mojo/stdlib/benchmarks/collections/bench_dict.mojo
```diff
@@ -96,6 +96,32 @@ fn bench_dict_lookup[size: Int](mut b: Bencher) raises:
     keep(Bool(items))
 
 
+# ===-----------------------------------------------------------------------===#
+# Benchmark Dict Lookup (types that allocated)
+# ===-----------------------------------------------------------------------===#
+@parameter
+fn bench_dict_lookup_types_that_allocates[size: Int](mut b: Bencher) raises:
+    """Lookup size items."""
+    dict = Dict[String, String]()
```
soraros: I think the stdlib style is using `var` everywhere.

mojo/stdlib/src/collections/dict.mojo
JoeLoser: +1 here, can you elaborate a bit @rd4com?

JoeLoser (approved):
Nice work!  Just had one question/follow-up that @soraros had as well.  Once that's addressed, happy to sync this in and land it.  Thank you!

mojo/stdlib/src/collections/dict.mojo
rd4com: There should not be any difference, as long as the quick check (compare two Int64) is done first.
For example: `debug_assert(is_debug() and huge_function(), " ... " )`
First one get done first, so if `True`, second one not needed !

Why not have the two check with an `and` ? 
`80` cols but can put it back there and `mojo format` if needed. (need answer)

`T.__eq__(a, b)` or `a == b` ?
Well this one is "just to be sure lol",
for example, if there is a bug, or a missing `(`, `)`,
overall this removes the redundency for a bug to exist,
there are some tests created by CL in this way for example,
we can put it back, does works and looks better :+1:  (need answer)

mojo/stdlib/benchmarks/collections/bench_dict.mojo
rd4com: yep, will upgrade to this, thanks

mojo/stdlib/src/collections/dict.mojo
rd4com: Thanks, that makes sense.
There is a lot of `debug_assert(value.__bool__())` for `Dict`,
i'd suggest another PR to replace them all if this is the way to go ?

mojo/stdlib/benchmarks/collections/bench_dict.mojo
rd4com: Hello @soraros , 
last commit implements the review, thanks :+1:

rd4com:
[Test stdlib and examples / with macos-14 and assertions=1 (pull_request)](https://github.com/modular/max/actions/runs/14103019700/job/39503202904?pr=4162)
(Successful in 23m)
It is more than usually, is it normal ?
(ping to @JoeLoser)

rd4com:
Hello @JoeLoser ,
i'll close this PR in favour of @bgreni 's
- https://github.com/modular/max/pull/4356

They essentially the same,  
except the most recent one don't need more work to resolve conflicts.

copy/paste the benchmarks if needed, or pull them from here,
more `__copyinit__` can be removed in `_maybe_resize` and `__init__`

PR #4167, opened 2025-03-19
[stdlib] Add fill ctor and method for `List`

bgreni:
@lsh Since you mentioned you would make an internal ticket for this.

Add a fill constructor and method for `List`.

### DIFF
--- a/mojo/stdlib/src/collections/list.mojo
+++ b/mojo/stdlib/src/collections/list.mojo
@@ -187,6 +187,18 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
         self._len = length
         self.capacity = capacity
 
+    fn __init__(out self, *, fill: T, count: UInt):
+        """Constructs a list containing `count` copies of the `fill` element.
+
+        Args:
+            fill: The value used to fill the list.
+            count: The number of values in the list.
+        """
+        self = Self(capacity=count)
+        for i in range(count):
+            (self.data + i).init_pointee_copy(fill)
+        self._len = count
+
     fn __moveinit__(out self, owned existing: Self):
         """Move data of an existing list into a new one.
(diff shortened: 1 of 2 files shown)


JoeLoser:
@lsh already has an internal PR up from yesterday for this — sorry for any confusion there.  He opted not to have a `fill` method and instead just

```mojo
    fn __init__(out self, *, length: Int, fill: T):
        """Constructs a list with the given capacity.
        Args:
            length: The requested length of the list.
            fill: The element to fill each element of the list.
        """
        self = Self()
        self.resize(length, fill)
```

I'll let him work with you on deciding whether we want a dedicated `fill` method or whether the constructor only suffices.  I think just the constructor is sufficient, WDYT?

gryznar:
I am curious why not to use `UInt` for length?

bgreni:
> @lsh already has an internal PR up from yesterday for this — sorry for any confusion there. He opted not to have a `fill` method and instead just
> 
> ```mojo
>     fn __init__(out self, *, length: Int, fill: T):
>         """Constructs a list with the given capacity.
>         Args:
>             length: The requested length of the list.
>             fill: The element to fill each element of the list.
>         """
>         self = Self()
>         self.resize(length, fill)
> ```
> 

No worries at all, I wasn't sure if he would have the time! I don't have any issue with excluding a `fill` method.

I'm not a huge fan of just calling out to `resize` personally as it involves a couple unnecessary branches, but I might be nitpicking here.

lsh:
@bgreni I'm not particularly concerned in this case. In my opinion, the effort is probably better spent optimizing `resize` to be as fast as possible. Either way, my implementation will be out in the next release and we can increment on it as we go.

@gryznar fair point, I changed it to UInt before merging.

PR #4171, opened 2025-03-20
[stdlib] Fix large inline array fill initializations

bgreni:
#4046 Broke EmberJSON due the indexing being improperly calculated causing a segfault when initializing large arrays.

https://github.com/bgreni/EmberJson/blob/main/emberjson/slow_float_parse.mojo#L247

Simply the logic and also avoid lots of unnecessary arithmetic by incrementing a simple pointer. I have also applied this approach to the variadic initialization constructor.

Also added a test for constructing a filled array larger than the default batch size.

### DIFF
--- a/mojo/stdlib/src/collections/inline_array.mojo
+++ b/mojo/stdlib/src/collections/inline_array.mojo
@@ -201,20 +201,20 @@ struct InlineArray[
 
         alias unroll_end = math.align_down(size, batch_size)
 
-        for i in range(0, unroll_end, batch_size):
+        var ptr = self.unsafe_ptr()
+
+        for _ in range(0, unroll_end, batch_size):
 
             @parameter
-            for j in range(batch_size):
-                var ptr = UnsafePointer.address_of(
-                    self.unsafe_get(i * batch_size + j)
-                )
+            for _ in range(batch_size):
                 ptr.init_pointee_copy(fill)
+                ptr += 1
 
         # Fill the remainder
         @parameter
-        for i in range(unroll_end, size):
-            var ptr = UnsafePointer.address_of(self.unsafe_get(i))
+        for _ in range(unroll_end, size):
             ptr.init_pointee_copy(fill)
+            ptr += 1
 
     @always_inline
     @implicit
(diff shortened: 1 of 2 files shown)


gryznar:
I am curious about benchmarking. Seems to be a clear win, but maybe it is worth checking?

JoeLoser (approved):
Thanks for the fix!  Can you please also add some additional tests for different batch sizes?  We don't have any coverage for that I think.

JoeLoser:
!sync

JoeLoser:
@bgreni I went ahead and synced this so I can cherry pick it into the 25.2 release branch that already got cut to avoid a regression slipping into the release.  We can follow-up with additional testing for non-default batch sizes and large arrays.

bgreni:
> @bgreni I went ahead and synced this so I can cherry pick it into the 25.2 release branch that already got cut to avoid a regression slipping into the release. We can follow-up with additional testing for non-default batch sizes and large arrays.

Sorry I had to step out for a few hours. I can definitely make a follow up with additional testing!

JoeLoser:
> > @bgreni I went ahead and synced this so I can cherry pick it into the 25.2 release branch that already got cut to avoid a regression slipping into the release. We can follow-up with additional testing for non-default batch sizes and large arrays.
> 
> Sorry I had to step out for a few hours. I can definitely make a follow up with additional testing!

No problem at all, that would be great! Just wanted to share context as to why I was syncing and merging even after I left my comment since it was good to merge this into 25.2 today since a new RC was getting cut.

rd4com:
Hello, it is late here but feel like it is important,
there is just a lot of parameters to test,
so here are some tests in a different way:

```mojo
def test_init_fill[size: Int, batch_size:Int, DT:DType](arg: Scalar[DT]):
    var arr = InlineArray[Scalar[DT], size].__init__[batch_size=batch_size](fill=arg)
    assert_equal(arr[0], arg)
    assert_equal(arr[size-1], arg)
    assert_equal(arr[size//3], arg)

def test_init_fill_scalars[*DTs: DType, sizes: List[Int], batch_sizes: List[Int]]():
    @parameter
    for current_batch_size in range(len(batch_sizes)):
        @parameter
        for current_size in range(len(sizes)):
            @parameter
            for current_type in range(len(VariadicList(DTs))):
                test_init_fill[sizes[current_size], batch_sizes[current_batch_size]](
                    Scalar[DTs[current_type]].MAX
                )
                test_init_fill[sizes[current_size], batch_sizes[current_batch_size]](
                    Scalar[DTs[current_type]].MIN
                )

test_init_fill_scalars[
    Int32.type, 
    Int16.type, 
    Int8.type, 
    sizes=List(
        1, 2, 3, 4, 8, 32, 64, 128,
        256, 512, 1023, 1024
    ), 
    batch_sizes=List(
        1, 2, 3, 4, 8, 31, 32, 63, 64, 128,
        #256, 512, 1023, 1024
    )
]()
```
More tests should be created, because not sure with `List` in parameters,
but that is something already at least :+1: 
gtg, just make sure to create more tests !

rd4com:
Need update, for people to be able to adjust `batch_size`:
The doc example for the `__init__` don't work:
```mojo
var a = InlineArray[Int,4 ,batch_size=4](fill=0)
```
this do:
```mojo
var b = InlineArray[Int,4].__init__[batch_size=4](fill=0)
```


ping to @JoeLoser

PR #4173, opened 2025-03-20
[stdlib] Add simple `List.pop()` overload for last item

martinvuyk:
Add simple `List.pop()` overload for last item


JoeLoser (approved):

JoeLoser:
!sync

soraros:
FTR, I checked the generated asm, this overload doesn't seem to change the output much.

PR #4174, opened 2025-03-21
[docs] update benchmarks/README.md

eltociear:
minor fix

### DIFF
--- a/mojo/stdlib/benchmarks/README.md
+++ b/mojo/stdlib/benchmarks/README.md
@@ -41,7 +41,7 @@ benchmarks to get started.
 Note that the `benchmark` package isn't open source yet and we do not currently
 have a mechanism for generating nightly API docs for closed source packages.
 So, we manually provide relatively up-to-date docs for these [here](../../docs/bencher/).
-In the future, we hope to open source the `benchmark` package and and also generate
+In the future, we hope to open source the `benchmark` package and also generate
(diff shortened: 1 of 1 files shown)


mojo/stdlib/benchmarks/README.md
```diff
@@ -41,7 +41,7 @@ benchmarks to get started.
 Note that the `benchmark` package isn't open source yet and we do not currently
```
JoeLoser, line 41: **Suggestion** We can rework this entire paragraph now that we have nightly API docs.  Do you mind rewording this to point at https://docs.modular.com/mojo/stdlib/benchmark/?  We no longer provide the manual bencher docs since we have up-to-date nightly API docs.

JoeLoser (commented):
Thanks for the fix!  This doc is a bit outdated, do you have interest in fixing up the paragraph more broadly given my suggestion?  If not, I'm happy to do it internally as well.

JoeLoser:
!sync

mojo/stdlib/benchmarks/README.md
JoeLoser: I'll go ahead and sync this in to land your PR as-is, and then I'll do the overhaul of the file given things have changed a bit in the past few months since I have a lot of that context. 😄 

Thanks for the contribution!

PR #4181, opened 2025-03-22
[stdlib] `InlineArray` -> fix example, add `debug_assert`.

rd4com:
Hello, small pr to help with fix #4171.

This adds a `debug_assert` to test that all values are initialized,
and also correct the doc example.

### DIFF
--- a/mojo/stdlib/src/collections/inline_array.mojo
+++ b/mojo/stdlib/src/collections/inline_array.mojo
@@ -217,7 +217,7 @@ struct InlineArray[
             ptr.init_pointee_copy(fill)
             ptr += 1
         debug_assert(
-            Int(ptr) == (Int(self.unsafe_ptr().offset(size))),
+            ptr == self.unsafe_ptr().offset(size),
             "error during initialization, please create a bug report",
         )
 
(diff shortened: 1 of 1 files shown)


soraros:
This looks really awful, I wonder whether we could get the old behaviour back.

mojo/stdlib/src/collections/inline_array.mojo
```diff
@@ -215,6 +216,10 @@ struct InlineArray[
         for _ in range(unroll_end, size):
             ptr.init_pointee_copy(fill)
             ptr += 1
+        debug_assert(
+            Int(ptr) == (Int(self.unsafe_ptr().offset(size))),
```
soraros: Nit:
```suggestion
            ptr == self.unsafe_ptr().offset(size),
```

bgreni:
> This looks really awful, I wonder whether we could get the old behaviour back.

You only have to invoke `__init__` directly when you want to override the default batch size so it's probably fine for now until we have better syntax for this sort of thing.

mojo/stdlib/src/collections/inline_array.mojo
```diff
@@ -215,6 +216,10 @@ struct InlineArray[
         for _ in range(unroll_end, size):
             ptr.init_pointee_copy(fill)
             ptr += 1
+        debug_assert(
+            ptr == self.unsafe_ptr().offset(size),
+            "error during initialization, please create a bug report",
```
JoeLoser:
```suggestion
            "error during `InlineArray` initialization, please file a bug report.",
```

JoeLoser (approved):

JoeLoser:
!sync

PR #4191, opened 2025-03-27
[stdlib] Upgrade `Dict.__contains__` to use `get_ptr(k)`.

rd4com:
Hello,

this upgrade that method that returns a `Bool`,
so that it does no `__copyinit__` on an entry value.

replaces:
- `find` (returns an `Optional` to an entry value) with:
- `get_ptr` (returns an `Optional` to a `Pointer` to an entry value)

We then return the result of `.__bool__()` on the `Optional` `Pointer`.

(speedup for `V` of types that allocates)

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_dict.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_dict.mojo
@@ -97,6 +97,27 @@ fn bench_dict_lookup[size: Int](mut b: Bencher) raises:
 
 
 # ===-----------------------------------------------------------------------===#
+# Benchmark Dict __getitem__ K: trivial V:allocated
+# ===-----------------------------------------------------------------------===#
+@parameter
+fn bench_dict_getitem[size: Int](mut b: Bencher) raises:
+    """__getitem__ size new items."""
+    var items = Dict[Int, String]()
+    for i in range(size):
+        items[i] = String(i)
+
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        for key in range(size):
+            var res = Int(random.random_si64(0, size)) in items
+            keep(res)
+
+    b.iter[call_fn]()
+    keep(Bool(items))
+
+
+# ===-----------------------------------------------------------------------===#
 # Benchmark Dict Memory Footprint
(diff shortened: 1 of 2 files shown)


gryznar:
Here also benchmarks will be desired to increase coverage ;)

rd4com:
Thanks for reviewing :+1: , implemented the benchmark in last commit @gryznar :
⏱️ not pr:
```
-------------------------------------------------------------
| name                     | met (ms)              | iters  |
-------------------------------------------------------------
| bench_dict_getitem[10]   | 0.0015975058472214173 | 759164 |
| bench_dict_getitem[30]   | 0.005039061336539594  | 237868 |
| bench_dict_getitem[1000] | 0.19445256099935104   | 6164   |
-------------------------------------------------------------
```
⏱️ pr
```
--------------------------------------------------------------
| name                     | met (ms)              | iters   |
--------------------------------------------------------------
| bench_dict_getitem[10]   | 0.001191833286        | 1000000 |
| bench_dict_getitem[30]   | 0.0035297459894227177 | 340163  |
| bench_dict_getitem[1000] | 0.14234149146018762   | 8314    |
--------------------------------------------------------------
```
That's a `0.19/0.14` speedup,
you'll see most of it shine when pr on `_find_index` get merged :+1: 

Thanks for asking for the benchmark, the speedup is for `V` of memory types!

gryznar:
Coverage in benchmarks is not so good, so thanks for adding ;)

mojo/stdlib/benchmarks/collections/bench_dict.mojo
```diff
@@ -96,6 +96,27 @@ fn bench_dict_lookup[size: Int](mut b: Bencher) raises:
...
+    """__getitem__ size new items."""
+    var items = Dict[Int, String]()
+    for i in range(size):
+        items[i] = String(i)
+
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        for key in range(size):
+            var res = Int(random.random_si64(0, size)) in items
```
soraros, line 113: Can we simply write `key in items` here?

PR #4195, opened 2025-03-28
[stdlib] Add `ParamKVCache` for `Dict`.

rd4com:
🆕 New meta-programmed struct to:

`__getitem__[k:K]()->ref[O]V` for each `K` in `Self.List[K]`.

The struct takes a `Keys: List[K]` as a parameter. 
Pointers to the values of `Keys` are retrieved during `Self.__init__`.

They are stored in an `InlineArray`, logically mapped to the Keys:
    - `for i in range(len(Keys)): value_pointers[i] = dict[Keys[i]]`

The `get_ref()` method takes a `Key: K` as a parameter. 
If that key is not in the `List[K]` struct parameter, an error is raised. 
If it is, the `Pointer[V]` mapped to it on `__init__` is returned.

The cache can be updated with the method `update_cache`.

It is then possible to `__getitem__` a `Dict`,
by passing a key previously cached as a parameter 🎉.

Pointers are cached on `__init__` once, or with `update_cache()`.

### DIFF
--- a/mojo/stdlib/src/collections/dict.mojo
+++ b/mojo/stdlib/src/collections/dict.mojo
@@ -1307,3 +1307,159 @@ struct OwnedKwargsDict[V: CollectionElement](
     @always_inline
     fn _insert(mut self, key: StringLiteral, owned value: V):
         self._insert(String(key), value^)
+
+
+@value
+struct ParamKVCache[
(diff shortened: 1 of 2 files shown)


gryznar:
Why do we here rely on raising instead of using optional? Mimicking Python is not the best way to go IMO

PR #4196, opened 2025-03-29
[tools] Add environment variable for Mojo Dev Container

benz0li:
Due to https://github.com/modular/max/commit/974872977e3138f72c48161094c2c35f6a979699, `MODULAR_MOJO_NIGHTLY_IMPORT_PATH=/opt/modular/lib/mojo ./stdlib/scripts/run-benchmarks.sh` does not work anymore.

### DIFF
--- a/mojo/stdlib/benchmarks/lit.cfg.py
+++ b/mojo/stdlib/benchmarks/lit.cfg.py
@@ -47,7 +47,10 @@ else:
     # This is important since `benchmark` is closed source
     # still right now and is always used by the benchmarks.
     pre_built_packages_path = Path(
-        repo_root / ".magic" / "envs" / "default" / "lib" / "mojo"
+        os.environ.get(
+            "MODULAR_MOJO_PREBUILT_PACKAGES_PATH",
+            repo_root / ".magic" / "envs" / "default" / "lib" / "mojo",
+        )
     )
 
     # The `run-tests.sh` script creates the build directory for you.


benz0li:
Cross reference:

* https://github.com/benz0li/mojo-dev-container/issues/3

benz0li:
ℹ️ This merge request was created using the [Mojo Dev Container](https://github.com/benz0li/mojo-dev-container).

Ahajha:
I would rather not hardcode workarounds for non-Modular code here. This repo is set up to assume use of `magic`.

Ahajha:
I'll give this some thought though. I understand the issue and there should be a way to allow use through things other than `magic`.

benz0li:
> I'll give this some thought though. I understand the issue and there should be a way to allow use through things other than `magic`.

@Ahajha Thank you. Very much appreciated.

benz0li:
I tried choosing a *neutral* name for the environment variable: `MODULAR_MOJO_PREBUILT_PACKAGES_PATH`.

Regarding the Mojo Dev Container: https://github.com/modular/max/blob/main/mojo/stdlib/docs/development.md#dev-container

Ahajha:
The issue is that then it looks like an "official" variable, while it actually has no relation to our code at all.

I do think I've found a reasonable workaround though. In our conda environment, we set `MODULAR_HOME` to `<package root>/share/max`, so rather than hardcoding `.magic/envs/default/lib/mojo`, we can use `MODULAR_HOME/../../lib/mojo`. Then you can override `MODULAR_HOME` to the new `<package root>/share/max`. (I have plans to expose another env var, something like `MODULAR_PACKAGE_ROOT` in the future, but that's for another thing that's very WIP).

I like this solution better since it generalizes over the package itself, and uses existing mechanisms not tied to `magic`.

Ahajha:
I've made a PR for my approach, can you try it out and see if you can make that work? Technically you can set `MODULAR_HOME` to the same value as `MODULAR_MOJO_PREBUILT_PACKAGES_PATH` was, though you should probably set it to `MODULAR_MOJO_PREBUILT_PACKAGES_PATH`/../../share/max to be proper.

benz0li:
> can you try it out and see if you can make that work?

That will work.

The [Mojo Dev Container](https://github.com/benz0li/mojo-dev-container) has `MODULAR_HOME` set to `/opt/modular/share/max`.

So `MODULAR_HOME/../../lib/mojo` resolves to `/opt/modular/lib/mojo`.  

The same I set `MODULAR_MOJO_NIGHTLY_IMPORT_PATH` to.
And the same I would have set `MODULAR_MOJO_PREBUILT_PACKAGES_PATH` to.

benz0li:
Closed in favour of https://github.com/modular/max/pull/4260.

Ahajha:
@benz0li Just a note, you should also now change `MODULAR_MOJO_NIGHTLY_IMPORT_PATH` to `MODULAR_MOJO_MAX_IMPORT_PATH`. All of the `MODULAR_MOJO_*` variables should be `MODULAR_MOJO_MAX_*` instead, and all `NIGHTLY` versions don't do anything anymore.

Though I don't _think_ you'll need either, but if it's still necessary that's fine, that env var isn't going anywhere so it's fine to rely on it.

benz0li:
> @benz0li Just a note, you should also now change `MODULAR_MOJO_NIGHTLY_IMPORT_PATH` to `MODULAR_MOJO_MAX_IMPORT_PATH`. All of the `MODULAR_MOJO_*` variables should be `MODULAR_MOJO_MAX_*` instead, and all `NIGHTLY` versions don't do anything anymore.

Yes.

> Though I don't _think_ you'll need either

As soon as https://github.com/modular/max/pull/4260 is merged, neither is required.

> but if it's still necessary that's fine, that env var isn't going anywhere so it's fine to rely on it.

As soon as https://github.com/modular/max/pull/4260 is merged, I will close (or even delete) https://github.com/benz0li/mojo-dev-container/issues/3.

PR #4197, opened 2025-03-29
[stdlib] Remove redundant call to `constant` ops

soraros:


lattner (approved):
Awesome, this is strictly better - you're right there should be no need to use kgen.param.constant in the standard library - mlirattrs promote from pvalues to rvalues using it anyway.

lattner:
@JoeLoser and @ConnorGray and @lsh  I'm not sure how integration happens here so I won't merge this, but this seems like an obviously great patch :)

JoeLoser:
!sync

JoeLoser:
> @JoeLoser and @ConnorGray and @lsh I'm not sure how integration happens here so I won't merge this, but this seems like an obviously great patch :)

Feel free to just do !sync in the future and then a PR will open internally and you can approve and merge after it passes CI internally.

PR #4201, opened 2025-03-29
[stdlib] Remove redundant use of `kgen.param.constant`

soraros:
Along with #4197, this removes all use of `kgen.param.constant` from the standard library.

CC @lattner.

### DIFF
--- a/mojo/stdlib/src/utils/numerics.mojo
+++ b/mojo/stdlib/src/utils/numerics.mojo
@@ -509,64 +509,36 @@ fn nan[type: DType]() -> Scalar[type]:
 
     @parameter
     if type is DType.float8_e5m2:
-        return rebind[__mlir_type[`!pop.scalar<`, type.value, `>`]](
-            __mlir_op.`kgen.param.constant`[
-                _type = __mlir_type[`!pop.scalar<f8e5m2>`],
-                value = __mlir_attr[`#pop.simd<"nan"> : !pop.scalar<f8e5m2>`],
-            ]()
+        return rebind[Scalar[type]](
+            __mlir_attr.`#pop.simd<"nan"> : !pop.scalar<f8e5m2>`,
         )
     elif type is DType.float8_e5m2fnuz:
-        return rebind[__mlir_type[`!pop.scalar<`, type.value, `>`]](
-            __mlir_op.`kgen.param.constant`[
-                _type = __mlir_type[`!pop.scalar<f8e5m2fnuz>`],
-                value = __mlir_attr[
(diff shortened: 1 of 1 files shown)


lattner (approved):

lattner:
Super awesome, thank you!  @JoeLoser I'm not merging this because I don't know the protocol but LGTM

soraros:
@lattner Maybe you missed this from @JoeLoser: https://github.com/modular/max/pull/4197#issuecomment-2763949497

JoeLoser (approved):

JoeLoser:
Nice clean up, thank you!

JoeLoser:
!sync

mojo/stdlib/src/utils/numerics.mojo
```diff
@@ -509,64 +509,36 @@ fn nan[type: DType]() -> Scalar[type]:
...
     @parameter
     if type is DType.float8_e5m2:
-        return rebind[__mlir_type[`!pop.scalar<`, type.value, `>`]](
-            __mlir_op.`kgen.param.constant`[
-                _type = __mlir_type[`!pop.scalar<f8e5m2>`],
-                value = __mlir_attr[`#pop.simd<"nan"> : !pop.scalar<f8e5m2>`],
-            ]()
+        return rebind[Scalar[type]](
+            __mlir_attr.`#pop.simd<"nan"> : !pop.scalar<f8e5m2>`,
         )
```
lattner, line 514: I wonder if you could further collapse all of these param ifs with something like:

```
  return rebind[Scalar[type]](
            __mlir_attr[`#pop.simd<"nan"> : `, Scalar[type]._mlir_type]
        )
```

This is how I collapsed similar code in `SIMD.init(self, value: FloatLiteral)`

lattner (approved):

mojo/stdlib/src/utils/numerics.mojo
soraros: ~~Indeed let me try that tomorrow.~~

@lattner Actually, now I remember it. I tried this and was greeted with the following error:

```mojo
error: invalid MLIR attribute: SIMD constant requires a concrete type
```

soraros:
@JoeLoser Given how further simplification is not possible at this moment, I think this PR should be read.

JoeLoser:
> @JoeLoser Given how further simplification is not possible at this moment, I think this PR should be read.

Agreed, I just fixed up the merge conflicts from yesterday's `dtype` renames and this is set to merge internally soon.  Thanks!

PR #4204, opened 2025-03-29
[stdlib] Hardening ABI for `CPython.PyDict_*`

soraros:
- Update all `PyDict_*` functions so that their signatures match the corresponding C API. In particular, `PyDict_Next` now takes `Ptr[PyObjectPtr]`, and the redundant `PyKeysValuePair` struct has been removed.
- Make `PyObjectPtr` implement `Stringable` and `EqualityComparable` so that it works with `assert_equal`.
- Add tests.

### DIFF
--- a/mojo/stdlib/stdlib/python/_cpython.mojo
+++ b/mojo/stdlib/stdlib/python/_cpython.mojo
@@ -106,26 +106,16 @@ struct PyThreadState:
 
 @fieldwise_init
 @register_passable("trivial")
-struct PyKeysValuePair:
-    """Represents a key-value pair in a Python dictionary iteration.
-
-    This struct is used to store the result of iterating over a Python dictionary,
-    containing the key, value, current position, and success status of the iteration.
-    """
-
-    var key: PyObjectPtr
-    """The key of the current dictionary item."""
-    var value: PyObjectPtr
-    """The value of the current dictionary item."""
-    var position: c_int
-    """The current position in the dictionary iteration."""
-    var success: Bool
-    """Indicates whether the iteration was successful."""
-
-
-@fieldwise_init
-@register_passable("trivial")
-struct PyObjectPtr(Boolable, Copyable, Defaultable, Intable, Movable, Writable):
+struct PyObjectPtr(
+    Boolable,
+    Copyable,
+    Defaultable,
+    EqualityComparable,
+    Intable,
+    Movable,
+    Stringable,
+    Writable,
+):
     """Equivalent to `PyObject*` in C.
 
     It is crucial that this type has the same size and alignment as `PyObject*`
@@ -203,6 +193,10 @@ struct PyObjectPtr(Boolable, Copyable, Defaultable, Intable, Movable, Writable):
     @always_inline
     fn __int__(self) -> Int:
         return Int(self.unsized_obj_ptr)
+
+    @always_inline
+    fn __str__(self) -> String:
+        return String.write(self)
 
     # ===-------------------------------------------------------------------===#
     # Methods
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/python/_cpython.mojo
```diff
@@ -990,52 +999,58 @@ struct CPython:
...
+
+        [Reference](https://docs.python.org/3/c-api/dict.html#c.PyDict_Check).
         """
 
-        var my_type = self.PyObject_Type(maybe_dict)
-        var my_type_as_int = my_type._get_ptr_as_int()
-        var dict_type = self.PyDict_Type()
-        var result = my_type_as_int == dict_type._get_ptr_as_int()
-        self.Py_DecRef(my_type)
-        return result
```
soraros: Why didn't we use `PyDict_Check`? Was it an optimisation?

mojo/stdlib/src/python/_cpython.mojo
```diff
@@ -990,52 +999,58 @@ struct CPython:
...
-        """[Reference](
-        https://docs.python.org/3/c-api/dict.html#c.PyDict_Type).
+        """This instance of `PyTypeObject` represents the Python dictionary type.
+
+        [Reference](https://docs.python.org/3/c-api/dict.html#c.PyDict_Type).
         """
+
         if self.dict_type.is_null():
+            # PyTypeObject PyDict_Type
             self.dict_type = self.lib.call["PyDict_Type", PyObjectPtr]()
```
soraros: `PyDict_Type` is of `PyTypeObject`, is this correct?

mojo/stdlib/src/python/_cpython.mojo
```diff
@@ -990,52 +999,58 @@ struct CPython:
...
+        """Returns a `PyKeysValuePair`.
+
+        [Reference](https://docs.python.org/3/c-api/dict.html#c.PyDict_Next).
         """
+
         var key = PyObjectPtr()
         var value = PyObjectPtr()
-        var v = p
-        var position = UnsafePointer[Int].address_of(v)
+        var pos: c_ssize_t = 0
```
soraros: Python doc says
> The [`Py_ssize_t`](https://docs.python.org/3/c-api/intro.html#c.Py_ssize_t) referred to by ppos must be initialized to 0 prior to the first call to this function to start the iteration

Why do we pass in `p` at all?

mojo/stdlib/src/python/_cpython.mojo
```diff
@@ -1056,13 +1071,7 @@ struct CPython:
...
         )
 
-        _ = v
-        return PyKeysValuePair(
-            key,
-            value,
-            position.take_pointee(),
-            result == 1,
-        )
+        return PyKeysValuePair(key, value, ppos.take_pointee(), result == 1)
```
soraros: Can we simply use `pos` here instead of `ppos.take_pointee()`?

mojo/stdlib/src/python/_cpython.mojo
```diff
@@ -990,52 +999,58 @@ struct CPython:
...
 
-    # int PyDict_Next(PyObject *p, Py_ssize_t *ppos, PyObject **pkey, PyObject **pvalue)
-    fn PyDict_Next(
-        mut self, dictionary: PyObjectPtr, p: Int
-    ) -> PyKeysValuePair:
-        """[Reference](
-        https://docs.python.org/3/c-api/dict.html#c.PyDict_Next).
+    fn PyDict_Next(mut self, dictionary: PyObjectPtr) -> PyKeysValuePair:
+        """Returns a `PyKeysValuePair`.
+
```
soraros: The `p` argument is removed.

mojo/stdlib/src/python/_cpython.mojo
laszlokindrat: Possibly because it's not part of the stable ABI.

laszlokindrat (commented):
Thanks for working on this! Do you think you could add some tests to lock down and showcase the behavior of these?

laszlokindrat:
@soraros I love to see these fixes, please @ me in any other patch you need a review for, and let me know when this is ready to review again! Thank you!

soraros:
@laszlokindrat I'm not sure what the best way to add tests for these is. We currently don't have **any** tests for `CPython`, and these methods aren't used anywhere. Do you think it's okay if I just leave some TODOs inline?

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

laszlokindrat:
!sync

laszlokindrat:
!sync

laszlokindrat:
!sync

PR #4206, opened 2025-03-29
[stdlib] Add `always_inline` to `Any(arg: SIMD)`, `All(arg: SIMD)`.

rd4com:
Hello,
this function is used in some `SIMD` methods,
(`__floordiv__`, `__ceildiv__`, `mod`, `_pow`)
and all `SIMD` things are inlined,
so just making a PR, but no idea if needed or not!


JoeLoser:
Let's lazily evaluate this, I don't think this is needed right now.  Going to close this PR for now to help clean up the review queue. Thanks, though!

PR #4208, opened 2025-03-29
[stdlib] Make `Prefetch*` flag types explicitly constructible from `Int`

soraros:
So they behave more like enums.


JoeLoser (approved):

JoeLoser:
!sync

PR #4209, opened 2025-03-30
[stdlib] Add test for `FloatLiteral` from `IntLiteral` construction

soraros:
Add test for `FloatLiteral` from `IntLiteral` construction.

### DIFF
--- a/mojo/stdlib/test/builtin/test_float_literal.mojo
+++ b/mojo/stdlib/test/builtin/test_float_literal.mojo
@@ -26,6 +26,11 @@ alias inf = FloatLiteral.infinity
 alias neg_inf = FloatLiteral.negative_infinity
 
 
+def test_init():
+    alias n: IntLiteral[(4).value] = 4
+    assert_equal(4.0, FloatLiteral(n))
+
+
 def test_division():
     assert_equal(FloatLiteral.__truediv__(4.4, 0.5), 8.8)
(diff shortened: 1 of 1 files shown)


mojo/stdlib/src/builtin/float_literal.mojo
```diff
@@ -61,7 +61,7 @@ struct FloatLiteral[value: __mlir_type.`!pop.float_literal`](
         Args:
             value: The IntLiteral value.
         """
-        out = __type_of(result)()
+        result = __type_of(result)()
```
JoeLoser: Do we have pre-existing tests for this? It's not clear how this worked before. If not, can you add some tests, please?

soraros: I actually don't know how to add a good test for this (but I did it anyway). It seems the problem here is that, supposedly, we can even leave the body of the function empty (using `pass`), since there's no field to initialise.

soraros:
Fixed in 80c48c51ab1f1cdaf5c8f2d420c7f23da4940f98, closing.

soraros:
Actually, do we still want the test?

JoeLoser:
> Actually, do we still want the test?

Sure, happy to take the test - thanks!

JoeLoser:
!sync

PR #4210, opened 2025-03-30
[stdlib] Simplify `bit.mojo`

soraros:
Do we even want/need a `bit.bit_not`?


JoeLoser (approved):
Let's keep `bit_not` for now, but I agree, it has little value given that `~val` is well-formed where `val` is a `SIMD` type.

JoeLoser:
!sync

PR #4212, opened 2025-03-30
[stdlib] Fix `SIMD.__init__` for large int literals

### LINKED ISSUE #4049: [BUG] Cant instantiate variable of UInt256 with more than 128 bit value
### Bug description

open-source/mojo/stdlib/stdlib/builtin/_startup.mojo:114:4: error: call expansion failed
open-source/mojo/stdlib/stdlib/builtin/_startup.mojo:45:4: note: function instantiation failed
<unknown>:0: note: call expansion failed
open-source/mojo/stdlib/stdlib/builtin/simd.mojo:425:8: note: function instantiation failed
<unknown>:0: note: integer value 170141183460469231731687303715884105728 requires 129 bits to store, but the destination bit width is only 128 bits wide

### Steps to reproduce

```mojo
fn main():
    alias Uint256 = Scalar[DType.uint256]
    var a: Uint256 = 170141183460469231731687303715884105727
    a += 1
    var b: Uint256 = a #This works 
    var c: Uint256 = 170141183460469231731687303715884105728 # This does not
    print(b, c)
```

### System information

max         25.2.0.dev2025030405

soraros:
Fix #4049. CC @lattner.

I don't know whether using wider types would cause performance issues.

### DIFF
--- a/mojo/stdlib/src/builtin/dtype.mojo
+++ b/mojo/stdlib/src/builtin/dtype.mojo
@@ -19,10 +19,10 @@ from collections import KeyElement
 from hashlib._hasher import _HashableWithHasher, _Hasher
 from sys import bitwidthof, os_is_windows, sizeof
 
-alias _mIsSigned = UInt8(1)
-alias _mIsInteger = UInt8(1 << 7)
-alias _mIsNotInteger = UInt8(~(1 << 7))
-alias _mIsFloat = UInt8(1 << 6)
+alias _mIsSigned = __mlir_attr.`#pop.simd<1> : !pop.scalar<ui8>`
+alias _mIsInteger = __mlir_attr.`#pop.simd<128> : !pop.scalar<ui8>`  # 1 << 7
+alias _mIsNotInteger = __mlir_attr.`#pop.simd<127> : !pop.scalar<ui8>`  # ~(1 << 7)
+alias _mIsFloat = __mlir_attr.`#pop.simd<64> : !pop.scalar<ui8>`  # 1 << 6
 
 
 @value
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/builtin/dtype.mojo
```diff
@@ -20,10 +20,10 @@ from collections.string import StringSlice
 from hashlib._hasher import _HashableWithHasher, _Hasher
 from sys import bitwidthof, os_is_windows, sizeof
 
-alias _mIsSigned = UInt8(1)
-alias _mIsInteger = UInt8(1 << 7)
-alias _mIsNotInteger = UInt8(~(1 << 7))
-alias _mIsFloat = UInt8(1 << 6)
+alias _mIsSigned = __mlir_attr.`#pop.simd<1> : !pop.scalar<ui8>`
```
gryznar, line 22: Could we have this and below in snake_case?

soraros: Yea, I suppose. But I think it's camel case since these were symbols matching their C++ internals?

gryznar: But should it match? I suppose it is not vital here

soraros: In this particular case, I actually think yes, because they are black magic values (bit patterns corresponding to how each `dtype` is modelled in the compiler).

A bit like names in the `_cpython` module.
https://github.com/modular/max/blob/88d37439b83085ad830b9df870146562a7af413b/mojo/stdlib/src/python/_cpython.mojo#L1650

Issue #4215, opened 2025-03-30
[BUG] Change name of the `DType` parameters through out the stdlib to `dtype`


soraros:
This is a tracking issue for the migration towards `dtype` following 08c5a0e0c8cbd194a7c4da6efa9d170a9b5b1e7e.

CC @JoeLoser. I think it's good first issue.

JoeLoser:
+1, thanks for filing this from our discussion. I'd happily welcome PRs helping in this area! @martinvuyk and @gryznar I think you've expressed interest in this in the past if I recall correctly.

PR #4221, opened 2025-03-30
[stdlib] [NFC] Rename `type: DType` parameters to `dtype` in `dtype.mojo`

martinvuyk:
Rename `type: DType` parameters to `dtype` in `dtype.mojo`. Part of #4215

### DIFF
--- a/mojo/stdlib/src/builtin/dtype.mojo
+++ b/mojo/stdlib/src/builtin/dtype.mojo
@@ -182,7 +182,7 @@ struct DType(
     )
     """Represents a special floating point format supported by NVIDIA Tensor
     Cores, with the same range as float32 and reduced precision (>=10 bits).
-    Note that this type is only available on NVIDIA GPUs.
+    Note that this dtype is only available on NVIDIA GPUs.
     """
     alias float64 = DType(__mlir_attr.`#kgen.dtype.constant<f64> : !kgen.dtype`)
     """Represents an IEEE754-2008 `binary64` floating point value."""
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/builtin/dtype.mojo
```diff
@@ -182,7 +182,7 @@ struct DType(
     )
     """Represents a special floating point format supported by NVIDIA Tensor
     Cores, with the same range as float32 and reduced precision (>=10 bits).
-    Note that this type is only available on NVIDIA GPUs.
+    Note that this dtype is only available on NVIDIA GPUs.
```
soraros, line 185: I'm a bit unsure about this one.

martinvuyk: Why? it's a dtype instance:
```mojo
    alias tensor_float32 = DType(
        __mlir_attr.`#kgen.dtype.constant<tf32> : !kgen.dtype`
    )
```

soraros: It sounds like the statement is referring to the type that `dtype` maps to, since the instance itself is available, but GPU support for this datatype is limited to Nvidia. That's just my impression, though – I'll leave it to your judgment.

That said, since I'm not a native speaker, my sense of the language might not be entirely accurate.

martinvuyk: Idk :man_shrugging:  to me it sounds like the datatype is only available for Nvidia GPUs which is what is normally said when talking about ASICs and the datatypes that are implemented in hardware.

JoeLoser (approved):

JoeLoser:
!sync

PR #4243, opened 2025-03-31
[stdlib] Make InlineArray variadic ctor assert msg more descriptive

bgreni:
Ran into this fairly unhelpful debug assert message when creating a larger lookup table with `InlineArray`. I've updated the message to include the expected, and actual size of the variadic list used to initialize it.


JoeLoser (approved):

JoeLoser:
!sync

PR #4244, opened 2025-03-31
[stdlib] Upgrade `Slice.__str__` to new out convention `(+capacity=16)`.

rd4com:
Hello, small PR to upgrade,
this also removes a move on the way,
it is quite nice how the two methods work together like this!

Small capacity of 16 (might be enough),
that way, it won't realloc too many times while building the return.

It should be possible to calculate an optimized capacity,
but that seem like a good base line:

```mojo
slice(1, 2, 3)
```

We might need to adjust the capacity, if `None` step is used a lot!

### DIFF
--- a/mojo/stdlib/src/builtin/builtin_slice.mojo
+++ b/mojo/stdlib/src/builtin/builtin_slice.mojo
@@ -95,15 +95,14 @@ struct Slice(
     # ===-------------------------------------------------------------------===#
 
     @no_inline
-    fn __str__(self) -> String:
+    fn __str__(self, out ret: String):
         """Gets the string representation of the span.
 
         Returns:
             The string representation of the span.
         """
-        var output = String()
-        self.write_to(output)
-        return output
+        ret = String(capacity=16)
+        self.write_to(ret)
 
     @no_inline
     fn __repr__(self) -> String:


mojo/stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -95,24 +95,23 @@ struct Slice(
...
         """
-        var output = String()
-        self.write_to(output)
-        return output
+        ret = String(capacity=16)
+        self.write_to(ret)
 
     @no_inline
-    fn __repr__(self) -> String:
+    fn __repr__(self, out ret: String):
```
JoeLoser: Does the code-gen change at all for this case?  I wouldn't expect any improvement in this case.

rd4com: 👍 Testing the counters: no improvements there:
```mojo
alias c_m =SIMD[DType.uint8, 2]
@value
struct DunderCounters[O:Origin[True]]:
    var x: Pointer[c_m, O]
    fn __copyinit__(out self, other:Self):
        self.x = other.x
        self.x[][0]+=1
    fn __moveinit__(out self, owned other:Self):
        self.x=other.x
        self.x[][1]+=1
@value
struct MyStruct[O:Origin[True]]:
    var x: DunderCounters[O]
    fn __init__(out self, owned x: DunderCounters[O]):
        self.x = x^
    @no_inline
    fn _repr(self)->DunderCounters[O]:
        return self._str()
    @no_inline
    fn _str(self, out ret:DunderCounters[O]):
        ret = self.x

from sys import sizeof
def main():
    value = SIMD[DType.uint8, 2](0,0)
    x = MyStruct(DunderCounters(Pointer.address_of(value)))
    a = x._repr()
    print(value) #[1, 1] (`__copyinit__`, `__moveinit__`) 👍
    a^.__del__()
    x^.__del__()
```
Ok, so no improvements there 👍
Would you like this one to be reverted ?
Does the `@no_inline` affect this ?

JoeLoser: I'd suggest reverting it here since we're not applying it elsewhere in the libraries. You'll get the guaranteed move elision essentially here in the return, so that's why there's no improvement. This is expected from Mojo.

rd4com: Thanks for explaining :bow: , that makes sense !
Revert: Yes that's what my question suggested but i expressed poorly, sorry.
New commit integrate the review :+1: 
(ping to @JoeLoser)

mojo/stdlib/src/builtin/builtin_slice.mojo
```diff
@@ -95,15 +95,14 @@ struct Slice(
...
+    fn __str__(self, out ret: String):
         """Gets the string representation of the span.
 
         Returns:
             The string representation of the span.
         """
-        var output = String()
-        self.write_to(output)
-        return output
+        ret = String(capacity=16)
```
martinvuyk, line 104: I'm all for these kinds of optimizations, but I'd rather see some arithmetic in code (+ comments when it's complicated) to explain the numbers

martinvuyk (commented):
I personally don't like changing this return type without also changing it all around the stdlib, it's inconsistent. The `out` ref AFAIK is meant for functions which have multiple return points, or complicated type signatures, or for unifying `__init__` with other function types.

mojo/stdlib/src/builtin/builtin_slice.mojo
martinvuyk:
```mojo
var amnt_nones = Int(not Bool(self.start)) + Int(not Bool(self.end)) + Int(not Bool(self.step))
alias base_strings = "slice()".byte_length() + 2 * ", ".byte_length()
# typically start end or step number won't be bigger than 3 bytes (999) long each.
var capacity = base_strings + 4 * amnt_nones + 3 * (3 - amnt_nones) + 1 # null terminator
```
or just directly doing a maximum estimate to avoid runtime overhead
```mojo
alias max_likely_capacity = 24 # "slice()" + 2 * ", " + 3 * 4-byte-numbers + null terminator
```

PR #4246, opened 2025-03-31
[stdlib] [NFC] Reorganize UTF-8 utils

martinvuyk:
Reorganize UTF-8 utils into their own files and tests. It is bloating `string_slice.mojo` and `test_string_slice.mojo`, and in the case that  #3988 gets accepted we will need to build similar infrastructure for UTF-16 and UTF-32.

### DIFF
--- a/mojo/stdlib/src/collections/string/string_slice.mojo
+++ b/mojo/stdlib/src/collections/string/string_slice.mojo
@@ -23,7 +23,13 @@ from collections.string import StringSlice
 
 from collections import List, Optional
 from collections.string.format import _CurlyEntryFormattable, _FormatCurlyEntry
-from collections.string._utf8_validation import _is_valid_utf8
+from collections.string._utf8 import (
+    _is_valid_utf8,
+    _count_utf8_continuation_bytes,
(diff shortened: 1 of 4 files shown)


mojo/stdlib/src/collections/string/string_slice.mojo
```diff
@@ -1794,14 +1800,14 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
 
         @parameter
         if single_character:
-            return length != 0 and _is_newline_char[include_r_n=True](
```
JoeLoser, line 1797: **Question** Was the rename from `_is_newline_char` to `_is_newline_char_utf8` intentional here?

JoeLoser (approved):
Thanks for splitting this out from the other PR 👍

JoeLoser:
!sync

mojo/stdlib/src/collections/string/string_slice.mojo
martinvuyk: yep, I renamed the function since it is only for utf-8

PR #4257, opened 2025-04-02
[stdlib] [NFC] [proposal] Add `StringSlice` and `StaticString` to prelude

martinvuyk:
Add `StringSlice` and `StaticString` to prelude.

Since we are starting to use both these types pervasively, I think it makes sense to add them to the prelude. This will also help guide users towards using `StringSlice` more since it is best for read-only/in-place mutation operations.

### DIFF
--- a/mojo/stdlib/src/prelude/__init__.mojo
+++ b/mojo/stdlib/src/prelude/__init__.mojo
@@ -17,14 +17,14 @@
 from collections import KeyElement, List, Optional, InlineArray
 from collections.string import (
     Codepoint,
+    StaticString,
     String,
+    StringSlice,
     ascii,
     atof,
     atol,
     chr,
     ord,
-    StringSlice,
-    StaticString,
 )
 from hashlib.hash import Hashable, hash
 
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/prelude/__init__.mojo
```diff
@@ -15,7 +15,17 @@
...
+    Codepoint,
+    String,
+    ascii,
+    atof,
+    atol,
+    chr,
+    ord,
+    StringSlice,
+    StaticString,
+)
```
soraros: Nit: we seem to follow a type then function ordering.

```suggestion
from collections.string import (
    Codepoint,
    StaticString,
    String,
    StringSlice,
    ascii,
    atof,
    atol,
    chr,
    ord,
)
```

JoeLoser (approved):

JoeLoser (commented):
Can you please add a small changelog entry so other users know?  After that, happy to sync and land this.  Thank you!

martinvuyk:
@JoeLoser done

JoeLoser (approved):

JoeLoser:
!sync

martinvuyk:
@JoeLoser can we merge this before @lattner  fills up the stdlib with imports for `StaticString` ? :laughing:

PR #4260, opened 2025-04-02
[CI] Use MODULAR_HOME to determine package path

Ahajha:
Alternative to https://github.com/modular/max/pull/4196

This lets our CI scripts not assume we're using `magic`, so long as we pass `MODULAR_HOME` (which `magic` does automatically). Then the location of the packaged standard library is computed from that instead.

As an added bonus, finally get rid of `MODULAR_MOJO_IMPORT_PATH`, which didn't do anything. Now we're using `MODULAR_MOJO_MAX_IMPORT_PATH`, the canonical env var to override Mojo stdlib import paths.

### DIFF
--- a/examples/mojo/lit.cfg.py
+++ b/examples/mojo/lit.cfg.py
@@ -46,17 +46,13 @@ config.test_source_root = Path(__file__).parent.resolve()
 # Substitute %mojo for just `mojo` itself.
 config.substitutions.insert(0, ("%mojo", "mojo"))
 
-pre_built_packages_path = os.environ.get(
-    "MODULAR_MOJO_IMPORT_PATH",
-    Path(os.environ["MODULAR_HOME"])
-    / "pkg"
-    / "packages.modular.com_nightly_mojo"
-    / "lib"
-    / "mojo",
-)
+# MODULAR_HOME is at <package root>/share/max by default
+pre_built_packages_path = (
+    Path(os.environ["MODULAR_HOME"]) / ".." / ".." / "lib" / "mojo"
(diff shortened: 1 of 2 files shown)


benz0li:
@Ahajha Thank you!

benz0li:
Resolves https://github.com/benz0li/mojo-dev-container/issues/3.

examples/mojo/lit.cfg.py
```diff
@@ -46,17 +46,13 @@
...
-    "MODULAR_MOJO_IMPORT_PATH",
-    Path(os.environ["MODULAR_HOME"])
-    / "pkg"
-    / "packages.modular.com_nightly_mojo"
-    / "lib"
-    / "mojo",
-)
+# MODULAR_HOME is at <package root>/share/max by default
+pre_built_packages_path = (
+    Path(os.environ["MODULAR_HOME"]) / ".." / ".." / "lib" / "mojo"
```
JoeLoser, line 51: **Suggestion** (Non-blocking) instead of the `/ ".." / ".."`, we could use `.parent.parent` if you'd like.  I find that slightly nicer but up to you.

JoeLoser (approved):

JoeLoser:
!sync

examples/mojo/lit.cfg.py
Ahajha: Oh yea, forgot that existed. I'll add that in before merging.

Ahajha: Added on the internal PR

PR #4262, opened 2025-04-02
[stdlib] [NFC] Fix docstrings in `list.mojo`

martinvuyk:
Fix docstrings in `list.mojo`


JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

soraros:
Funny how we immediately need another update to these docstrings after the list literal change.

martinvuyk:
@soraros yeah lol, but mostly in the examples. But I would leave them as is because they show how to specify the collection type instead of assuming `[1, 2, 3]` defaults to `List[Int]`. Or we could make some nice and "more aesthetic" docs :man_shrugging:

PR #4266, opened 2025-04-02
[stdlib] Inline and improve compile time iota implementation

martinvuyk:
Inline and improve compile time iota implementation

### DIFF
--- a/mojo/stdlib/stdlib/math/math.mojo
+++ b/mojo/stdlib/stdlib/math/math.mojo
@@ -29,6 +29,7 @@ from sys import (
     llvm_intrinsic,
     simdwidthof,
     sizeof,
+    is_compile_time,
 )
 from sys._assembly import inlined_assembly
 from sys.ffi import _external_call_const
@@ -1103,18 +1104,6 @@ fn isclose[
 # ===----------------------------------------------------------------------=== #
 
 
-# TODO: Remove this when `iota` works at compile-time
-fn _compile_time_iota[dtype: DType, width: Int]() -> SIMD[dtype, width]:
-    constrained[
-        dtype.is_integral(),
-        "_compile_time_iota can only be used with integer dtypes.",
-    ]()
-    var a = SIMD[dtype, width](0)
-    for i in range(width):
-        a[i] = i
-    return a
-
-
 @always_inline
 fn iota[
     dtype: DType, width: Int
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/math/math.mojo
```diff
@@ -1096,20 +1083,19 @@ fn iota[
...
-            SIMD[DType.index, simd_width],
-            has_side_effect=False,
-        ]()
-        return it.cast[dtype]() + offset
+
+    alias step_dtype = dtype if dtype.is_integral() else DType.index
+
+    if is_compile_time():
+        var a = SIMD[step_dtype, simd_width](0)
+        for i in range(Scalar[step_dtype](simd_width)):
```
soraros: I think you can just use the `Int` variant here.
```suggestion
        for i in range(simd_width):
```

soraros:
This looks great. Should we keep the comment though? I think `llvm.stepvector` should run at compile time. CC @lattner.

mojo/stdlib/src/math/math.mojo
```diff
@@ -1096,20 +1083,19 @@ fn iota[
...
-            has_side_effect=False,
-        ]()
-        return it.cast[dtype]() + offset
+
+    alias step_dtype = dtype if dtype.is_integral() else DType.index
+
+    if is_compile_time():
+        var a = SIMD[step_dtype, simd_width](0)
+        for i in range(simd_width):
+            a[i] = Scalar[step_dtype](i)
```
soraros: Sorry for another nit. I think `Int` is always implicit convertible to `SIMD[...]`?

martinvuyk: True, tks

mojo/stdlib/stdlib/math/math.mojo
```diff
@@ -1137,20 +1124,19 @@ fn iota[
...
+    if is_compile_time():
+        var a = SIMD[step_dtype, simd_width](0)
+        for i in range(simd_width):
+            a[i] = i
+        return a.cast[dtype]() + offset
+
+    var step = llvm_intrinsic[
+        "llvm.stepvector", SIMD[step_dtype, simd_width], has_side_effect=False
+    ]()
+    return step.cast[dtype]() + offset
```
soraros: Maybe the following? So it's clear how `cast` is shared. Not super important though.
```suggestion
    var step: SIMD[step_dtype, simd_width]
    if is_compile_time():
        step = 0
        for i in range(simd_width):
            step[i] = i
    else:
        step = llvm_intrinsic[
            "llvm.stepvector",
            SIMD[step_dtype, simd_width],
            has_side_effect=False,
        ]()
    return step.cast[dtype]() + offset
```

soraros: Or you could rename `a` into `step`.

JoeLoser:
!sync

mojo/stdlib/stdlib/math/math.mojo
soraros: Haha, I suggested this fro I anticipated https://github.com/modular/modular/commit/a33a525af8016c69a2b2c049ff9db56ed7506a99.

PR #4275, opened 2025-04-03
[stdlib] Add compile time `_validate_utf8`

martinvuyk:
Add compile time `_validate_utf8`. Split off from PR  #4134

Had some issues to work around for llvm intrinsics. I would like to merge this and once the issues are solved we can drop them.

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/_utf8.mojo
+++ b/mojo/stdlib/stdlib/collections/string/_utf8.mojo
@@ -14,7 +14,7 @@
 """Implement UTF-8 utils."""
 
 from base64._b64encode import _sub_with_saturation
-from sys import simd_width_of
+from sys import simd_width_of, is_compile_time
 from sys.intrinsics import likely
 
 from bit import count_leading_zeros
@@ -162,6 +162,44 @@ fn _is_valid_utf8_runtime(span: Span[mut=False, Byte, **_]) -> Bool:
     return all(has_error.eq(0))
 
 
+fn _is_valid_utf8_comptime(span: Span[mut=False, Byte, **_]) -> Bool:
+    var ptr = span.unsafe_ptr()
+    var length = UInt(len(span))
+    var offset = UInt(0)
+
+    while offset < length:
+        var b0 = ptr[offset]
+        var byte_type = _utf8_byte_type(b0)
+        if byte_type == 0:
+            offset += 1
+            continue
+        elif byte_type == 1:
+            return False
+
+        for i in range(1, byte_type):
+            var idx = offset + i
+            if idx >= length or not _is_utf8_continuation_byte(ptr[idx]):
+                return False
+
+        # special unicode ranges
+        var b1 = ptr[offset + 1]
+        if byte_type == 2 and b0 < UInt8(0b1100_0010):
+            return False
+        elif b0 == 0xE0 and b1 < UInt8(0xA0):
+            return False
+        elif b0 == 0xED and b1 > UInt8(0x9F):
+            return False
+        elif b0 == 0xF0 and b1 < UInt8(0x90):
+            return False
+        elif b0 == 0xF4 and b1 > UInt8(0x8F):
+            return False
+
+        offset += UInt(byte_type)
+
+    return True
+
+
+@always_inline("nodebug")
 fn _is_valid_utf8(span: Span[mut=False, Byte, **_]) -> Bool:
     """Verify that the bytes are valid UTF-8.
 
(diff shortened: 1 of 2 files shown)


laszlokindrat:
@martinvuyk Could you please rebase this on the latest nightly?

mojo/stdlib/test/collections/string/test_utf8.mojo
```diff
@@ -81,8 +84,22 @@ alias BAD_SEQUENCES = [
...
+    var runtime = _is_valid_utf8_runtime(span)
+    assert_equal(comptime, runtime)
+    return comptime
+
+
+def validate_utf8(span: Span[Byte]) -> Bool:
+    var comptime = _is_valid_utf8_comptime(span)
+    var runtime = _is_valid_utf8_runtime(span)
+    assert_equal(comptime, runtime)
+    return comptime
```
laszlokindrat, line 100: Can't we just use `_is_valid_utf8` directly in the different contexts?

mojo/stdlib/stdlib/collections/string/_utf8.mojo
```diff
@@ -162,6 +162,117 @@ fn _is_valid_utf8_runtime(span: Span[Byte]) -> Bool:
     return all(has_error == 0)
 
 
+fn _validate_utf8_simd_slice[
```
laszlokindrat: Why can't we reuse this in the runtime implementation?

mojo/stdlib/stdlib/collections/string/_utf8.mojo
```diff
@@ -162,6 +162,117 @@ fn _is_valid_utf8_runtime(span: Span[Byte]) -> Bool:
...
+            return False
+    if iter_len >= 16 and simdwidthof[DType.uint8]() >= 16:
+        if not _validate_utf8_simd_slice[16](ptr, length, iter_len):
+            return False
+    if iter_len >= 8:
+        if not _validate_utf8_simd_slice[8](ptr, length, iter_len):
+            return False
+    if iter_len >= 4:
+        if not _validate_utf8_simd_slice[4](ptr, length, iter_len):
+            return False
```
laszlokindrat: Do we really need a simd-optimized compile-time implementation of this? I'm not sure how we interpret these simd instructions at compile time. It is also taking on a lot of complexity and the it's not clear to me that compile time utf8 validations are really that expensive.

mojo/stdlib/stdlib/collections/string/_utf8.mojo
martinvuyk: Because Gabriel's runtime implementation is much faster. It is based on one of Lemire's algorithms. This comptile time one is just a step by step checking of all cases in the spec, the runtime one uses tables and a bunch of other tricks

mojo/stdlib/stdlib/collections/string/_utf8.mojo
martinvuyk: > I'm not sure how we interpret these simd instructions at compile time.

I had to make a lot of adjustments to make this work (see #4142)

> it's not clear to me that compile time utf8 validations are really that expensive.

If we enable utf8 checking at compile time for debug builds, it could bloat compile times every time a `StringSlice` is built from a `String`. Even if that is not done, I would still prefer to have it be as fast as possible. Anyone that might embed a blob of binary into a Mojo executable with a lot of text information might use this. And there are places in the stdlib where we just build a `String` from a `Span[StaticConstantOrigin]` without checking the underlying utf-8

mojo/stdlib/test/collections/string/test_utf8.mojo
martinvuyk: I prefer to check that both are outputting the same results in the different test cases as well, I caught an issue with the compile time implementation that way before.

mojo/stdlib/stdlib/collections/string/_utf8.mojo
laszlokindrat: There is no benefit of the SIMD optimizations in `_validate_utf8_simd_slice`. I checked, and the interpreter that executes this at compile time doesn't actually use SIMD instructions when it runs. I'm not saying that we shouldn't have compile-time utf8 validation, only that it should be simple. Optimizing the interpreter is something we're working on, but we're far from the point where optimizations like this are worthwhile.

martinvuyk: The SIMD optimizations inside `_validate_utf8_simd_slice` aren't really that complex or different from what I'd do in scalar code. The scalar version would iterate through the 4 following bytes and check the conditions just the same. I can simplify this chain of if-else using a parameter for loop if you wish. But I think scalar code that checks utf-8 compliance will look more or less the same (IMO less understandable in some cases like the `valid_n4 or valid_n3 or valid_n2` section)

laszlokindrat: Please do simplify it; there is no clear benefit to doing simd optimizations for compile-time only code (in fact it might make compile time longer). So the thing to optimize for in this case is code maintainability.

mojo/stdlib/stdlib/collections/string/_utf8.mojo
```diff
@@ -162,6 +162,107 @@ fn _is_valid_utf8_runtime(span: Span[mut=False, Byte, **_]) -> Bool:
...
+    while iter_len > 0 and (iter_len >= width or remainder):
+        var d: SIMD[DType.uint8, width]  # use a vector of the specified width
+
+        @parameter
+        if not remainder:
+            d = ptr.offset(idx).load[width=width]()
+        else:
+            d = SIMD[DType.uint8, width](0)
+            for i in range(iter_len):
+                d[i] = ptr[idx + i]
```
laszlokindrat: This algorithm is still using SIMD optimizations, and isn't the simplest thing possible (recall my earlier feedback that you cannot actually optimize compile-time performance using SIMD optimizations). We're looking for a simple iteration through the bytes, at each checking which type of code point we have, checking the subsequent 1 to 3 bytes (if available), and advancing the iteration accordingly.

martinvuyk: Got it, I'll refactor this sometime this week. rn I've been working for 12 hours straight :melting_face:

laszlokindrat: That's okay, and we appreciate you working on this! Please ping me again when this is ready for a review!

martinvuyk: done :)

laszlokindrat (approved):
Very clean, thank you for working with me on this!

laszlokindrat:
!sync

PR #4277, opened 2025-04-03
[stdlib] Fix `PythonObject` creation from large `UInt64` values

### LINKED ISSUE #3510: [BUG] `PythonObject` doesn't handle large `UInt64` correctly

soraros:
Fix #3510

### DIFF
--- a/mojo/stdlib/test/python/test_python_object.mojo
+++ b/mojo/stdlib/test/python/test_python_object.mojo
@@ -255,6 +255,12 @@ def test_dunder_methods(mut python: Python):
     assert_equal(c, -35)
 
 
+def test_num_conversion() -> None:
+    alias n = UInt64(0xFEDC_BA09_8765_4321)
+    alias n_str = String(n)
+    assert_equal(n_str, String(PythonObject(n)))
+
+
 def test_bool_conversion() -> None:
     var x: PythonObject = 1
     assert_true(x == 0 or x == 1)
@@ -592,6 +598,7 @@ def main():
     var python = Python()
 
     test_dunder_methods(python)
+    test_num_conversion()
     test_bool_conversion()
     test_string_conversions()
     test_len()
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -161,6 +161,9 @@ previously compiler would throw error "cannot fold operation".
 
 ### 🛠️ Fixed
 
+- [#3510](https://github.com/modular/max/issues/3510) - `PythonObject` doesn't
+  handle large `UInt64` correctly.
+
 - [#3847](https://github.com/modular/max/issues/3847) - Count leading zeros
   can't be used on SIMD at compile time.
 
--- a/mojo/stdlib/src/python/python_object.mojo
+++ b/mojo/stdlib/src/python/python_object.mojo
@@ -383,16 +383,19 @@ struct PythonObject(
         Args:
             value: The scalar value.
         """
-        cpython = _get_global_python_itf().cpython()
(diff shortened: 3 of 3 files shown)


mojo/stdlib/test/python/test_python_object.mojo
```diff
@@ -255,6 +255,11 @@ def test_dunder_methods(mut python: Python):
     assert_equal(c, -35)
 
 
+def test_num_conversion() -> None:
+    alias n = UInt64(0xFEDC_BA09_8765_4321)
+    assert_true(String(n), String(PythonObject(n)))
```
JoeLoser: **Question** Do we need to stringify things or does `assert_equal(n, PythonObject(n))` work?  BTW, I think you meant `assert_equal` instead of `assert_true` regardless 😄

JoeLoser (commented):
Don't forget to add a changelog entry under the **Fixed** section mentioning you fixed https://github.com/modular/max/issues/3510 😄

soraros:
@JoeLoser Will definitely add a change log entry when I'm finalising this one.

mojo/stdlib/test/python/test_python_object.mojo
soraros: Good catch! And this is scary. Though I'm not sure `assert_equal(n, PythonObject(n))` makes sense since this suppose to check that we are converting the `n` to `Python` correctly. Maybe I should inline the string result?

soraros: Even round trip to back to `Int` wouldn't work, since this value is too large to fit in a `Int64`.

JoeLoser (approved):
LGTM, thanks for fixing this!

JoeLoser:
!sync

PR #4253, opened 2025-04-02
[stdlib] More `bit.mojo` cleanups

soraros:

Issue #3271, opened 2024-07-18
[Feature Request] Rename `bit_ceil` and `bit_floor` to `closest_pow_2_ge` and `closest_pow_2_le`


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Rename `bit_ceil` and `bit_floor` to `closest_pow_2_ge` and `closest_pow_2_le`.

Similarly as issue #3158, I think some renaming would be fitting here. I know these names come from C++, but some verbosity and clearer use case is better when looking for something and not intuitively understanding what the function name means.

In my particular case, I literally reimplemented `bit_ceil` with bitshifting (100% copied from stack overflow) because I was looking for the closest power of two greater than or equal to a dynamic value. And I was literally writing the code to include in a PR when I read the code for `bit_ceil`

### What is your motivation for this change?

Verbose names in exchange for clarity. Also, extending the naming to include  `closest_pow_2_gt` and  `closest_pow_2_lt` becomes clear and simple.

JoeLoser:
Thanks for filing the issue!  We talked about this internally earlier this week and decided on the names: `next_power_of_two` and `prev_power_of_two`.  This is taking inspiration from the Rust [https://doc.rust-lang.org/std/primitive.usize.html#method.next_power_of_two](https://doc.rust-lang.org/std/primitive.usize.html#method.next_power_of_two).

martinvuyk:
@JoeLoser now that I'm doing some work in the bit package I remembered that this got forgotten, should I open a PR or can this be done internally?

ConnorGray:
@martinvuyk I think we'd be happy to accept a community PR getting this started🙂  From a quick look, most of the uses of this function are in our open-source code, so there wouldn't be much work to do to get this over the finish line internally.

martinvuyk:
This was solved by 98498edd830b4a3b476d6cf89c9f8f9bb599e989 a while ago

PR #4278, opened 2025-04-03
[stdlib] Remove redundant `next_power_of_two()` from `math.mojo` and optimize existing function in `bit.mojo`

martinvuyk:
Optimization of `next_power_of_two`

### Benchmark results
CPU: i7 7th gen
Measured in miliseconds to do 100 million calculations
```text
bench_next_power_of_two_int_v1,204.92954861666666
bench_next_power_of_two_int_v2,117.86505349666666
bench_next_power_of_two_int_v3,111.60907072111111
bench_next_power_of_two_int_v4,105.20949030055556
bench_next_power_of_two_uint_v1,89.40362252435895
bench_next_power_of_two_uint_v2,82.5889815992857
bench_next_power_of_two_uint_v3,85.60044429065934
bench_next_power_of_two_uint_v4,108.61543930999999
```

I decided to go for `..._int_v3` and `..._uint_v2` (which use `SIMD[bool].select`) since they are the most stable in performance between executions, even though the `..._int_v4` and `..._uint_v1` have sometimes given better results.

### DIFF
--- a/mojo/stdlib/benchmarks/bit/bench_bit.mojo
+++ b/mojo/stdlib/benchmarks/bit/bench_bit.mojo
@@ -1,0 +1,107 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License next_pwer_of_two_v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo-no-debug %s -t
+# NOTE: to test changes on the current branch using run-benchmarks.sh, remove
+# the -t flag. Remember to replace it again before pushing any code.
+
+from benchmark import Bench, BenchConfig, Bencher, BenchId, Unit, keep, run
+from bit import bit_width, count_leading_zeros
+from collections import Dict
+from random import random_ui64, seed
+from sys import bitwidthof
+
+
+# ===-----------------------------------------------------------------------===#
+# Benchmarks
+# ===-----------------------------------------------------------------------===#
+
+# ===-----------------------------------------------------------------------===#
+# next_power_of_two
+# ===-----------------------------------------------------------------------===#
+
+
+fn next_power_of_two_v1(val: Int) -> Int:
+    if val <= 1:
+        return 1
+
+    if val.is_power_of_two():
+        return val
+
+    return 1 << bit_width(val - 1)
+
+
+fn next_power_of_two_v2(val: Int) -> Int:
+    if val <= 1:
+        return 1
+
+    return 1 << (bitwidthof[Int]() - count_leading_zeros(val - 1))
+
+
+fn _build_list() -> List[Int]:
+    var values = List[Int](capacity=10_000)
+    for _ in range(10_000):
+        values.append(Int(random_ui64(0, 2 ^ 64 - 1)))
+    return values^
+
+
+var values = _build_list()
+
+
+@parameter
+fn bench_next_power_of_two_v1(mut b: Bencher) raises:
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        for _ in range(10_000):
+            for i in range(len(values)):
+                var result = next_power_of_two_v1(values.unsafe_get(i))
+                keep(result)
+
+    b.iter[call_fn]()
+
+
+@parameter
+fn bench_next_wer_of_two_v2(mut b: Bencher) raises:
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        for _ in range(10_000):
+            for i in range(len(values)):
+                var result = next_wer_of_two_v2(values.unsafe_get(i))
+                keep(result)
+
+    b.iter[call_fn]()
+
+
+# ===-----------------------------------------------------------------------===#
+# Benchmark Main
+# ===-----------------------------------------------------------------------===#
+def main():
+    seed()
+    var m = Bench(BenchConfig(num_repetitions=10))
+    m.bench_function[bench_next_power_of_two_v1](
+        BenchId("bench_next_power_of_two_v1")
+    )
+    m.bench_function[bench_next_power_of_two_v2](
+        BenchId("bench_next_power_of_two_v2")
+    )
+
+    results = Dict[String, (Float64, Int)]()
+    for info in m.info_vec:
+        n = info[].name
+        time = info[].result.mean("ms")
+        avg, amnt = results.get(n, (Float64(0), 0))
+        results[n] = ((avg * amnt + time) / (amnt + 1), amnt + 1)
+    print("")
+    for k_v in results.items():
+        print(k_v[].key, k_v[].value[0], sep=",")
--- a/mojo/stdlib/benchmarks/bit/bench_bit.mojo
+++ b/mojo/stdlib/benchmarks/bit/bench_bit.mojo
@@ -19,6 +19,7 @@ from bit import bit_width, count_leading_zeros
 from collections import Dict
 from random import random_ui64, seed
 from sys import bitwidthof
+from sys.intrinsics import unlikely, likely
 
 
 # ===-----------------------------------------------------------------------===#
@@ -30,7 +31,7 @@ from sys import bitwidthof
 # ===-----------------------------------------------------------------------===#
 
 
-fn next_power_of_two_v1(val: Int) -> Int:
+fn next_power_of_two_int_v1(val: Int) -> Int:
     if val <= 1:
         return 1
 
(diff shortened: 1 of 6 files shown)


mojo/stdlib/src/math/math.mojo
```diff
@@ -2357,26 +2357,6 @@ fn clamp(
...
-    Args:
-        n: The input value.
-
-    Returns:
-        The next power of two greater than or equal to the input.
-    """
-    if n <= 1:
-        return 1
-
-    return 1 << (bitwidthof[Int]() - count_leading_zeros(n - 1))
```
JoeLoser, line 2377: Out of curiosity, did you compare the IR/perf of this vs the implementation in `bit.mojo`?

JoeLoser (commented):
I'm interested in perf difference (if any) between the two implementations.  I think we should just remove it from `math` entirely rather than having the import around in `mojo/stdlib/src/math/__init__.mojo`.

mojo/stdlib/src/math/math.mojo
martinvuyk: nope, abdul's version is probably faster though. I'll trow some benchmark together

martinvuyk:
@JoeLoser moved the implementation, added a benchmark and changed the whole PR description.

mojo/stdlib/benchmarks/bit/bench_bit.mojo
```diff
@@ -0,0 +1,107 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License next_pwer_of_two_v2.0 with LLVM Exceptions:
```
gryznar: Does `next_pwer_of_two` really should be here?

gryznar:
> @JoeLoser moved the implementation, added a benchmark and changed the whole PR description.

Could you please post benchmark results (numbers) as comment / description?

mojo/stdlib/benchmarks/bit/bench_bit.mojo
martinvuyk: yes because it is a bitwise operation, not a mathematical one.

gryznar: So, word "pwer" has a typo

martinvuyk: ok, didn't understand that is what you were hinting at

martinvuyk:
Read the PR description. 2x speedup. 199ms vs 100ms to execute the code, doesn't make much difference to post the exact results for a doubling in perf in a trivial function.

soraros:
Shall we write it in the same branchless fashion like I did in #4253?

martinvuyk:
@soraros that is very interesting. Now I'm wondering whether 
```mojo
val = Scalar[DType.index](int_value)
select(val <= 1, 1, 1 << (bitwidthof[Int]() - count_leading_zeros(val - 1)))
```
is faster or not than using bitwise ops i.e.
```mojo
(1 << (bitwidthof[Int]() - count_leading_zeros(val - 1))) & -Int(val > 1) | Int(val <= 1)
```

I'm also wondering whether we should implement this function for `UInt` as well at some point

soraros:
I *think* they ultimately generate the same asm, the `select` variant is just a bit more readable.

In fact, the `select` variant is equivalent to using `if`, though the generated LLVM IR is quite a bit simpler, which is beneficial, since it's an `@always_inline` function.

martinvuyk:
Nothing better than over-engineering the most trivial of functions for 2 hours. Anyway, it was fun trying and learning some new bitwise tricks. @soraros it seems like the best (most stable) option is to use `SIMD[bool].select`, and when the compiler knows that it's a `UInt` being cast to a `DType.index` the performance improves noticeably (see the numbers in the PR description)

soraros:
I checked the generated asm, the following three variants seem to generate the exact same asm. `ir_utils` from [here](https://gist.github.com/soraros/44d56698cb20a6c5db3160f13ca81675).

```mojo
from sys.info import bitwidthof
from bit import count_leading_zeros, count_trailing_zeros
from utils._select import _select_register_value as select

fn f(val: Int) -> Int:
  if val <= 1:
    return 1
  return 1 << (bitwidthof[Int]() - count_leading_zeros(val - 1))

fn g(val: Int) -> Int:
  alias bitwidth = bitwidthof[Int]()
  return select(val <= 1, 1, 1 << (bitwidth - count_leading_zeros(val - 1)))

fn h(val: Int) -> Int:
  var v = Scalar[DType.index](val)
  return Int(
    (v <= 1)
    .select(1, 1 << (bitwidthof[Int]() - count_leading_zeros(v - 1))))

fn main():
  from ir_utils import dump_ir
  dump_ir[f, "out1"]()
  dump_ir[g, "out2"]()
  dump_ir[h, "out3"]()
```

I personally find `g` more readable. Also, there's a constructor from `scalar<index>` to `Int`, so you could skip the `__index__()`.

martinvuyk:
> I personally find g more readable. Also, there's a constructor from scalar<index> to Int, so you could skip the __index__().

Totally agree, now it looks very pretty :]

mojo/stdlib/src/bit/bit.mojo
```diff
@@ -375,21 +376,37 @@ fn next_power_of_two(val: Int) -> Int:
...
+        The smallest power of 2 that is greater than or equal to the input
+        value.
+
+    Notes:
+        This operation is called `bit_ceil()` in C++.
     """
-    if val <= 1:
-        return 1
+    alias w = bitwidthof[Int]()
+    return Int(select(val <= 1, 1, 1 << (w - count_leading_zeros(val - 1))))
```
soraros:
```suggestion
    return select(val <= 1, 1, 1 << (w - count_leading_zeros(val - 1)))
```

mojo/stdlib/src/bit/bit.mojo
```diff
@@ -375,21 +376,37 @@ fn next_power_of_two(val: Int) -> Int:
...
+    Returns:
+        The smallest power of 2 that is greater than or equal to the input
+        value.
 
-    return 1 << bit_width(val - 1)
+    Notes:
+        This operation is called `bit_ceil()` in C++.
+    """
+    alias w = bitwidthof[Int]()
+    return UInt(select(val == 0, 1, 1 << (w - count_leading_zeros(val - 1))))
```
soraros:
```suggestion
    return select(val == 0, 1, 1 << (w - count_leading_zeros(val - 1)))
```

soraros:
I actually find eyeballing the generated IR and assembly really helpful when comparing different implementations. It can greatly enhance the experience of "over-engineering."

mojo/stdlib/src/bit/bit.mojo
```diff
@@ -375,21 +376,37 @@ fn next_power_of_two(val: Int) -> Int:
...
+    Returns:
+        The smallest power of 2 that is greater than or equal to the input
+        value.
 
-    return 1 << bit_width(val - 1)
+    Notes:
+        This operation is called `bit_ceil()` in C++.
+    """
+    alias w = bitwidthof[Int]()
+    return select(val == 0, 1, 1 << (w - count_leading_zeros(val - 1)))
```
abduld:
```suggestion
    return select(val == 0, 1, 1 << (bitwidthof[Int]() - count_leading_zeros(val - 1)))
```

is more readable

mojo/stdlib/src/utils/_select.mojo
```diff
@@ -33,5 +29,10 @@ fn _select_register_value[
 
     Returns:
         The value selected based on the condition.
+
+    Notes:
```
abduld: can you split this into a different PR

abduld (approved):
lgtm. Thanks a bunch for the contribution and sorry for not searching for the existing implementation. btw. If you want to dump the asm, then you can do 

```mojo
from sys import bitwidthof
from compile import compile_info
from utils._select import _select_register_value as select
from bit import count_leading_zeros


@always_inline
fn next_power_of_two(val: UInt) -> UInt:
    return select(
        val == 0,
        1,
        1 << (bitwidthof[__type_of(val)]() - count_leading_zeros(val - 1)),
    )


def main():
    print(compile_info[next_power_of_two, emission_kind="llvm"]())

    print(compile_info[next_power_of_two, emission_kind="asm"]())

```

mojo/stdlib/src/bit/bit.mojo
martinvuyk: I takes up 5 lines of code after formatting if I do that

Edit: I ended up doing it after pulling the change to not use the constructor around the select and it takes only 3 lines :man_shrugging:

mojo/stdlib/src/utils/_select.mojo
martinvuyk: It's trivial but ok

martinvuyk:
> sorry for not searching for the existing implementation.

no worries I also reimplemented the logic and wanted to commit it to the stdlib when I realized `bit_ceil` was already implemented and in the bit module. I'm not a C++ programmer so I didn't know to look for that function name, that's why I proposed the function name change in #3271

>  If you want to dump the asm, then you can do

I think I'm going to start doing that more often and not eyeballing benchmarks for small functions so much :sweat_smile:

JoeLoser (approved):

JoeLoser:
!sync

Issue #3662, opened 2024-10-13
[BUG] Code using `llvm.assume` can't run at compile time


soraros:
### Bug description

As title. I think we couldn't mark it as `has_side_effect=False`, since it will get DCE'd before even reaching LLVM IR.

Related to #3658.

### Steps to reproduce

```mojo
from sys.intrinsics import assume

fn main():
  alias res = f(10)
  print(res)

fn f(n: Int) -> Int:
  assume(n > 0)  # note: cannot constant fold zero-result LLVM intrinsic: llvm.assume
  return n + 1
```

### System information

```shell
Mojo 2024.10.1305 (c0abd11a) on Docker, Intel Mac
```

unfixa1:
Check the usage of llvm.assume: confirm whether the use of llvm.assume meets the requirements of the compiler and ensure that it is used in a context where constant folding can be performed.
In addition, use [servbay](servbay.com) instead of docker to deploy the development environment to eliminate the problem of environment incompatibility. servbay is easier to use on Mac than docker and has fewer bugs.

PR #4289, opened 2025-04-05
[stdlib] Add compile time branch for llvm expect intrinsic

martinvuyk:
Add compile time branch for llvm expect intrinsic. This is needed to run optimized code at compile time.


soraros:
Ohhhh, #3662 can be fixed this way I think.

soraros:
I think we need to explicitly import `is_compile_time` in `intrinsics.mojo`. 
It’s probably being leaked from some random module at the moment.

martinvuyk:
lol, I thought it was already imported

JoeLoser (approved):

JoeLoser:
!sync

PR #4290, opened 2025-04-05
[stdlib] Make `llvm.assume` no-op at compile time

### LINKED ISSUE #3662: [BUG] Code using `llvm.assume` can't run at compile time

soraros:
Fix #3662.

### DIFF
--- a/mojo/stdlib/src/sys/intrinsics.mojo
+++ b/mojo/stdlib/src/sys/intrinsics.mojo
@@ -887,7 +887,9 @@ fn assume(val: Bool):
     Args:
       val: The input value which is assumed to be `True`.
     """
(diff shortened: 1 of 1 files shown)


JoeLoser (approved):

JoeLoser:
!sync

Issue #4283, opened 2025-04-04
[Feature Request] [stdlib] [proposal] Change all structs that have a `size: Int` parameter to `UInt`


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Change all structs that have a `size: Int` parameter to `UInt`

### What is your motivation for this change?

Defining away many compile time checks. Getting back half the address space.

### Any other details?

This will take more and more work as code in the stdlib assumes `Int` in many functions that use the parameters. The sooner we start this the better.

JoeLoser:
We talked about this in our weekly team meeting yesterday and are +1 for this. To get the ball rolling, I'll go change the compiler to support https://github.com/modular/max/pull/4298 since I had poked at this last year in fact.

martinvuyk:
@JoeLoser FYI the PRs I opened don't need to wait for #4298 since I wrote workarounds for it. I could add followup PRs once it works to fix them (I just built an `Int` with their sizes). I also don't change the return type of any `__len__` function since that would require further rewrites. The only PR that might be tricky is #4296, the rest are pretty small and simple

martinvuyk:
> FYI, this isn't your fault, but I don't agree with this line of changes. We want a single common "currency" type for "integers the size of pointers". That type is called Int for a wide range of reasons. We should us it, instead of UInt. Let's talk offline and get something we can document in the stdlib manual. Swift went through all of this years ago and it was a hard won lesson, but after doing it, it was clearly the right path.

@lattner ref [the comment](https://github.com/modular/modular/pull/4294#issuecomment-2853321259) I figured it's best to bring the discussion here.

For me it's more a matter of not letting things that can't be negative use a possibly negative datatype; like collection lengths, allocation sizes, etc. This avoids having several comptime and runtime checks.

There is also the address space, which is arguably a very unimportant improvement since very few people will actually use collections that big.

Another big motivation are the op-level  optimizations that the compiler can do and the length normalization speedup. If we have every collection returning their length with a datatype that avoids that, it's a "free speedup"

[...]

PR #4291, opened 2025-04-05
[stdlib] Change `List._len` and `List.capacity` to `UInt`

martinvuyk:
Part of #4283

### DIFF
--- a/mojo/stdlib/src/collections/list.mojo
+++ b/mojo/stdlib/src/collections/list.mojo
@@ -54,7 +54,7 @@ struct _ListIter[
 
     alias list_type = List[T, hint_trivial_type]
 
-    var index: Int
+    var index: UInt
     var src: Pointer[Self.list_type, list_origin]
 
     fn __iter__(self) -> Self:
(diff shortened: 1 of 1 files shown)


JoeLoser (approved):
Great cleanup, thank you @martinvuyk!

JoeLoser:
!sync

mojo/stdlib/src/collections/list.mojo
```diff
@@ -901,25 +901,24 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
             A reference to the element at the given index.
         """
 
+        var normalized_idx = UInt(idx.__index__())
```
soraros, line 936: Is this right? I think we still want to handle negative index.

martinvuyk: It is, a couple of lines down:
```mojo
        if not _type_is_eq[I, UInt]():
            if Int(idx) < 0:
                normalized_idx = len(self) + Int(idx)
```
but I think we should probably use `normalize_index` here in the future anyway

soraros: Ah, right. I still find the old implementation a bit cleaner. Casting the underlying `index` to `UInt` gives me chills, even if the value is ultimately overwritten a few lines later. In any case, you can use `index(idx)` instead of `__index__()`.

martinvuyk: In my country we have a saying that is often heard when someone inebriated wants to do something very crazy "who's afraid to die shouldn't be born!" :rofl: . Numbers are just a bag of bits and this implementation is temporary anyway, we should clean this up to use `normalize_index`

PR #4292, opened 2025-04-05
[stdlib] Add `len[T: UIntSized](value: T) -> UInt` builtin

martinvuyk:
Part of #4283


NathanSWard:
Closing this PR as we don't have a `UIntSized` trait anymore.

PR #4293, opened 2025-04-05
[stdlib] Change `UnsafePointer.alloc(count: UInt)`

martinvuyk:
Change `UnsafePointer.alloc(count: UInt)`


NathanSWard:
Closing this PR as we now have the freestanding `alloc` function and are standardizing on `Int` for most stdlib APIs.

None the less, thanks for the PR :)

PR #4294, opened 2025-04-05
[stdlib] Change `Span._len` to `UInt`

martinvuyk:
Part of #4283


lattner:
FYI, this isn't your fault, but I don't agree with this line of changes.  We want a single common "currency" type for "integers the size of pointers".  That type is called Int for a wide range of reasons.  We should us it, instead of UInt.  Let's talk offline and get something we can document in the stdlib manual.  Swift went through all of this years ago and it was a hard won lesson, but after doing it, it was clearly the right path.

PR #4297, opened 2025-04-05
[stdlib] Rename `defined_mode` to `ASSERT_MODE`

soraros:
I personally find it much more readable this way.

Other use follow this convention:
- https://github.com/modular/max/blob/main/mojo/stdlib/test/os/path/test_islink.mojo
- https://github.com/modular/max/blob/main/mojo/stdlib/test/pathlib/test_pathlib.mojo


JoeLoser (approved):

JoeLoser:
!sync

PR #4298, opened 2025-04-05
[stdlib] Add parameter for generator for UInt

martinvuyk:
Part of #4283

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_dict.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_dict.mojo
@@ -84,7 +84,7 @@ fn bench_dict_lookup[size: Int](mut b: Bencher) raises:
         @parameter
         if size < 100:
             for _ in range(closest_divisor):
-                for key in range(Int(100 // closest_divisor)):
+                for key in range(100 // closest_divisor):
                     var res = items[key]
                     keep(res)
         else:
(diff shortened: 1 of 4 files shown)


JoeLoser (commented):
Can you add some simple unit tests to show it works as expected for the various cases, please?

martinvuyk:
@JoeLoser You were right wanting to add tests. It turns out the loop is returning `Int` instead of `UInt`. I imagine that at some point the `UInt` yield type gets lowered to `__mlir_type.index`/changed to `IntLiteral` and the default path for reconstruction is then through `Int`. I think this would dovetail nicely with an idea I've been having: I know it might get nasty, but maybe we can create "conditional materialization" where `IntLiteral` lowers to an `Int` when it is negative and `UInt` otherwise.

dgurchenkov:
This PR would fix the new bug we just got: https://github.com/modular/modular/issues/4517

@martinvuyk @JoeLoser what do you think about landing this PR independently of the others, just to fix the bug?

JoeLoser (approved):
LGTM, thank you!  I'll add the compiler machinery necessary to make this work internally when I sync this PR.

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
@martinvuyk we just pushed this over the finish line in the compiler internally - thanks for your help @dgurchenkov! ❤️

PR #4301, opened 2025-04-06
[stdlib] Change `Deque` fields to use `UInt`

martinvuyk:
Part of #4283

### DIFF
--- a/mojo/stdlib/src/collections/deque.mojo
+++ b/mojo/stdlib/src/collections/deque.mojo
@@ -51,7 +51,7 @@ struct Deque[ElementType: Copyable & Movable](
     # Aliases
     # ===-------------------------------------------------------------------===#
 
-    alias default_capacity: Int = 64
+    alias default_capacity: UInt = 64
     """The default capacity of the deque: must be the power of 2."""
 
     # ===-------------------------------------------------------------------===#
@@ -61,19 +61,19 @@ struct Deque[ElementType: Copyable & Movable](
     var _data: UnsafePointer[ElementType]
     """The underlying storage for the deque."""
 
-    var _head: Int
+    var _head: UInt
     """The index of the head: points the first element of the deque."""
 
-    var _tail: Int
+    var _tail: UInt
     """The index of the tail: points behind the last element of the deque."""
 
-    var _capacity: Int
+    var _capacity: UInt
     """The amount of elements that can fit in the deque without resizing it."""
 
-    var _min_capacity: Int
+    var _min_capacity: UInt
     """The minimum required capacity in the number of elements of the deque."""
 
-    var _maxlen: Int
+    var _maxlen: UInt
     """The maximum number of elements allowed in the deque.
 
     If more elements are pushed, causing the total to exceed this limit,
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/collections/deque.mojo
```diff
@@ -495,7 +492,7 @@ struct Deque[ElementType: CollectionElement](
             value: The value to append.
         """
         # checking for positive _maxlen first is important for speed
-        if self._maxlen > 0 and len(self) == self._maxlen:
+        if Int(self._maxlen) > 0 and len(self) == self._maxlen:
```
soraros: Why do we need the explicit casting here? `UInt` has a implicit constructor from `IntLiteral`.

martinvuyk: It's not about that. The comparison is to check for maximum maxlen or not. I'm changing this part of the code rn anyway. But I'm still trying to understand it

martinvuyk:
@avitkauskas do these changes make sense to you? I've read over the parts I'm changing several times and I think I get the logic, but I don't want to introduce any subtle bugs

avitkauskas:
I was away from Mojo for a while now, but at the time of implementation of Deque we had discussed that `Int` vs `UInt` issue with @JoeLoser and I found my comment about that on Discord now:
https://discord.com/channels/1087530497313357884/1224434323193594059/1302002101177552947

The problem was that Mojo was implicitly casting Int to UInt
```mojo
var a: UInt
var b = -1
a = b
print(a)  #  18446744073709551615

# Imagine I'd accidentaly do
q = Deque[Int](capacity=b)
```

Perhaps it's changed now and that implicit casting is not happening any more? Or there are other reasons to go `UInt`?
As I said, I was not following Mojo recently :(

P.S. Just checked it now. On `mojo 25.3.0.dev2025041105` it's still the same.

martinvuyk:
Lately (November 2024), I've championing for better implicit conversion pathways (see #3753) but there has been some push-back internally for what insofar I've been able to discern as internal disagreement as for what the roadmap towards warning vs. error-based implicit conversion should be. 

```mojo
# Imagine I'd accidentaly do
q = Deque[Int](capacity=b)
```
as for this I'm less concerned because if we change the DType to UInt then the theoretical problem will go away (if you pass in a negative `Int` you should get a compilation error )

avitkauskas:
In C++, this "signed/unsigned nemesis" has been around for ages. So, I was learning from _the Man_ himself.

In The C++ Programming Language, Bjarne Stroustrup writes:

> The unsigned integer types are ideal for uses that treat storage as a bit array. Using an unsigned instead of an int to gain one more bit to represent positive integers is almost never a good idea. Attempts to ensure that some values are positive by declaring variables unsigned will typically be defeated by the implicit conversion rules.

And in Section 4.10, advice item [18] says:

> Avoid unsigned arithmetic.

Mojo Team should know better - I'll trust them to make the smart call here :)

martinvuyk:
Currently the implicit conversion for a compile time known `IntLiteral` to `UInt` fails at compile time for negative values. In the runtime case, when a negative `Int` is converted implicitly you get a `UInt` larger than any value `Int` can represent, but I've wanted this behavior to fail with a `debug_assert` (not sure whether for all compilations or only for debug builds) for a long time.

IMO in this problem space of setting all negative `Int` values as being "unlimited sized `DeQueue`s" I don't see a big difference in saying they turn into very big sized `Deque`s instead. And IMO people who would use such big sized `Deque`s are probably going to take great care in such details.

avitkauskas:
This "signed/unsigned" is a "big can of worms". C++ people have learnt a lot about that already. I think we should learn from them. So, I'm on the side of using unsigned mostly for the "non arithmetic" cases.

Just a few (quite random perhaps) links to illustrate the C++ people opinion:
https://hamstergene.github.io/posts/2021-10-30-do-not-use-unsigned-for-nonnegativity/
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nonnegative

But I will not argue. I'm ok with whatever is taken on this.

PR #4304, opened 2025-04-06
[stdlib] Add `Error.as_string_slice()` and polish

martinvuyk:
Add `Error.as_string_slice()` and polish

### DIFF
--- a/mojo/stdlib/src/builtin/error.mojo
+++ b/mojo/stdlib/src/builtin/error.mojo
@@ -76,7 +76,7 @@ struct Error(
             value: The error message.
         """
         self.data = value.unsafe_ptr()
-        self.loaded_length = len(value)
+        self.loaded_length = value.byte_length()
 
     @implicit
     fn __init__(out self, src: String):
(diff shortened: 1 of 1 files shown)


mojo/stdlib/src/builtin/error.mojo
```diff
@@ -211,17 +199,18 @@ struct Error(
...
-            repr(
-                String(
-                    StringSlice[__origin_of(self)](
-                        unsafe_from_utf8_cstr_ptr=self.unsafe_cstr_ptr()
-                    )
-                )
-            ),
-            ")",
-        )
+        return String("Error(", repr(self.as_string_slice()), ")")
```
soraros, line 201: Is the `repr` necessary (can we make it `"Error('"` etc.)?

martinvuyk: I also don't like building a String unnecessarily, but I don't know whether to remove this. It adds `"` to the end and beginning of the message or `'` if the message contains `"`, it also encodes control characters in the ASCII range. Since it encodes those characters instead of printing, I think it's worth it in cases where you would want to print something that might have been corrupted on the way. But I'm not sure.

soraros: I see. Maybe we should make `repr` return a different `Writable` than `String` in the future.

martinvuyk: Maybe once we have some API for streaming a repr i.e. `__repr__(self, mut writer: Writer)`

laszlokindrat (commented):
Thanks! Could you please fix the conflicts so I can bring this in?

laszlokindrat (approved):

martinvuyk:
@laszlokindrat done :)

laszlokindrat:
!sync

PR #4305, opened 2025-04-06
[stdlib] Change `os.env` functions to use `StringSlice`

martinvuyk:
Change `os.env` functions to use `StringSlice`

### DIFF
--- a/mojo/stdlib/src/os/env.mojo
+++ b/mojo/stdlib/src/os/env.mojo
@@ -51,7 +51,7 @@ fn setenv(
     var status = external_call["setenv", Int32](
         name.unsafe_ptr().bitcast[c_char](),
         value.unsafe_ptr().bitcast[c_char](),
-        Int32(1 if overwrite else 0),
+        Int32(overwrite),
     )
     return status == 0
 
--- a/mojo/stdlib/src/os/env.mojo
+++ b/mojo/stdlib/src/os/env.mojo
@@ -21,7 +21,7 @@ from os import setenv
 
 
 from sys import external_call, os_is_linux, os_is_macos, os_is_windows
-from sys.ffi import c_int
+from sys.ffi import c_int, c_char
 
 from memory import UnsafePointer
 
@@ -49,8 +49,8 @@ fn setenv(
         return False
 
     var status = external_call["setenv", Int32](
-        name.unsafe_ptr().bitcast[c_int](),
-        value.unsafe_ptr().bitcast[c_int](),
+        name.unsafe_ptr().bitcast[c_char](),
+        value.unsafe_ptr().bitcast[c_char](),
         Int32(1 if overwrite else 0),
     )
     return status == 0
@@ -70,7 +70,7 @@ fn unsetenv(name: StringSlice) -> Bool:
     ]()
 
     return (
-        external_call["unsetenv", c_int](name.unsafe_ptr().bitcast[c_int]())
+        external_call["unsetenv", c_int](name.unsafe_ptr().bitcast[c_char]())
         == 0
     )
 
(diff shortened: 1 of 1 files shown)


mojo/stdlib/src/os/env.mojo
```diff
@@ -26,35 +26,37 @@ from sys.ffi import c_int
...
+        The function only works on macOS or Linux and returns False otherwise.
     """
     alias os_is_supported = os_is_linux() or os_is_macos()
     if not os_is_supported:
         return False
 
     var status = external_call["setenv", Int32](
-        name.unsafe_cstr_ptr(),
-        value.unsafe_cstr_ptr(),
+        name.unsafe_ptr().bitcast[c_int](),
```
JoeLoser: **Question** Why `c_int`?  We want these to map to `const char*`: see the [UNIX docs for setenv](https://man7.org/linux/man-pages/man3/setenv.3.html). Ditto elsewhere.

martinvuyk: oh, totally. I don't know why I didn't use c_char

owenhilyard:
env args are mutable, so it's probably best that either the runtime makes a copy at the start of the program or that we do make a copy here due to hazards around C libraries modifying them.

mojo/stdlib/src/os/env.mojo
```diff
@@ -67,32 +69,43 @@ fn unsetenv(name: String) -> Bool:
...
-      The value of the environment variable.
+        The value of the environment variable.
+
+    Constraints:
+        The function only works on macOS or Linux and returns an empty string
+        otherwise.
     """
     alias os_is_supported = os_is_linux() or os_is_macos()
 
     if not os_is_supported:
```
soraros: Not necessarily in this PR, though I think we can make this `@parameter if`.

soraros:
@owenhilyard The `setenv` (C version makes a internal copy unlike `putenv`) and `unsetenv` take a `StringSlice` in Mojo land, and the `getenv` performs a copy on return. I don't see any issues with this approach.

mojo/stdlib/src/os/env.mojo
```diff
@@ -21,40 +21,42 @@ from os import setenv
...
     alias os_is_supported = os_is_linux() or os_is_macos()
     if not os_is_supported:
         return False
 
     var status = external_call["setenv", Int32](
-        name.unsafe_cstr_ptr(),
-        value.unsafe_cstr_ptr(),
+        name.unsafe_ptr().bitcast[c_char](),
+        value.unsafe_ptr().bitcast[c_char](),
         Int32(1 if overwrite else 0),
```
soraros:
```suggestion
        Int32(overwrite),
```

soraros:
Actually, do we want to remove `unsafe_cstr_ptr` (different from `unsafe_c_char_ptr` which is a purely ABI thing)? We would lose an important API contract, namely, null termination, even if only from a documentation perspective.

martinvuyk:
> Actually, do we want to remove `unsafe_cstr_ptr` (different from `unsafe_c_char_ptr` which is a purely ABI thing)? We would lose an important API contract, namely, null termination, even if only from a documentation perspective.

I've wanted to deprecate `unsafe_cstr_ptr` for a long time (remember  #3638 ?)

owenhilyard:
> @owenhilyard The `setenv` (C version makes a internal copy unlike `putenv`) and `unsetenv` take a `StringSlice` in Mojo land, and the `getenv` performs a copy on return. I don't see any issues with this approach.

That's probably fine so long as we provide locking on the Mojo side, since all of the C library functions are thread unsafe.

PR #4306, opened 2025-04-06
[stdlib] [NFC] Replace `_strnlen` with `_unsafe_strlen(ptr, max: UInt = UInt.MAX)`

martinvuyk:
Replace `_strnlen` with `_unsafe_strlen(ptr, max: UInt = UInt.MAX)`

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/string_slice.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string_slice.mojo
@@ -2404,11 +2404,14 @@ fn _to_string_list[O: Origin, //](items: List[Span[Byte, O]]) -> List[String]:
 
 
 @always_inline
-fn _unsafe_strlen(ptr: UnsafePointer[Byte, mut=False, **_]) -> Int:
+fn _unsafe_strlen(
+    ptr: UnsafePointer[Byte, mut=False, **_], max: UInt = UInt.MAX
+) -> UInt:
     """Get the length of a null-terminated string from a pointer.
 
     Args:
         ptr: The null-terminated pointer to the string.
+        max: The maximum size of the string.
 
     Returns:
         The length of the null terminated string without the null terminator.
@@ -2416,10 +2419,10 @@ fn _unsafe_strlen(ptr: UnsafePointer[Byte, mut=False, **_]) -> Int:
     Notes:
         The length does NOT include the null terminator.
     """
-    var len = 0
-    while ptr[len]:
-        len += 1
-    return len
+    var offset = UInt(0)
+    while offset < max and ptr[offset]:
+        offset += 1
+    return offset
 
 
 @always_inline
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/os/os.mojo
```diff
@@ -148,9 +141,9 @@ struct _DirHandle:
             if not ep:
                 break
             var name = ep.take_pointee().name
-            var name_ptr = name.unsafe_ptr()
+            var name_ptr = name.unsafe_ptr().bitcast[UInt8]()
```
soraros: Nit: maybe use the `Byte` alias here? Same below.

mojo/stdlib/src/collections/string/string_slice.mojo
```diff
@@ -2301,22 +2301,39 @@ fn _to_string_list[
...
         The length of the null terminated string without the null terminator.
+
+    Notes:
+        The length does NOT include the null terminator.
     """
-    var len = 0
-    while ptr.load(len):
-        len += 1
-    return len
+    return _strnlen(ptr, UInt.MAX)
```
soraros: Actually, would you consider making `MAX` the default?

martinvuyk: Makes sense!

mojo/stdlib/src/collections/string/string_slice.mojo
```diff
@@ -2325,22 +2325,39 @@ fn _to_string_list[
...
-    Get the length of a null-terminated string from a pointer.
-    Note: the length does NOT include the null terminator.
+    var offset = UInt(0)
+    while offset < max and ptr[offset]:
+        offset += 1
+    return offset
+
+
+@always_inline
+fn _unsafe_strlen(owned ptr: UnsafePointer[Byte]) -> UInt:
```
soraros: Can we get rig of this altogether? In any case:
```suggestion
fn _unsafe_strlen(ptr: UnsafePointer[Byte]) -> UInt:
```

soraros:
I suggest removing the `_strlen` variant and renaming `_strnlen` to `_strlen`. We don’t have to match the C API, and those functions are named differently only because C doesn’t support overloading or default arguments.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2326,22 +2326,25 @@ fn _to_string_list[
     return _to_string_list[items.T, len_fn, unsafe_ptr_fn](items)
 
 
-@always_inline
```
soraros, line 2406: Maybe we still want this?

martinvuyk: I'm not sure but I'll just leave it as is then

kartva (approved):
This PR is unifying two functions into one and generally makes the codebase better. The rename doesn't seem to be high-impact. LGTM!

You will need to rebase; there are a few small merge conflicts.

kartva:
!sync

PR #4307, opened 2025-04-06
[stdlib] Change `_lstat` and `_stat` to use `StringSlice`

martinvuyk:
Change `_lstat` and `_stat` to use `StringSlice`


JoeLoser (approved):
LGTM, thanks!

JoeLoser:
!sync

Issue #4217, opened 2025-03-30
[Feature Request] [stdlib] Un-deprecate `str` builtin


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Why was this deprecated? I understand that in the case of the `int` builtin, it's because it's not really an infinite precision int like Python's. But what is ambiguous or incorrect about using `str`?

### What is your motivation for this change?

Pythonic code. Short and fast to write code (no need to write `String` everywhere). Pretty coloring in IDEs.

### Any other details?
CC: @jackos

soraros:
The `String` type is not exactly like `str` because it is mutable. In this sense, it parallels the distinction between `Int` and `int`. However, the main argument seems to be that we don't want two ways to achieve the exact same thing.

martinvuyk:
If we make the `str` builtin return a `StringSlice` would it qualify as being like Python's ? It would be immutable in length, but mutable in-place

Would these traits make sense?
```mojo
trait SliceableMaybeMut:
  fn __str__(ref self) -> StringSlice[__origin_of(self)]: ...

trait SliceableRead:
  fn __str__(read self) -> StringSlice[__origin_of(self)]: ...

trait SliceableView:
  alias mut: Bool
  alias origin: Origin[mut]
  fn __str__(self) -> StringSlice[origin]: ...

trait SliceableStatic:
  fn __str__(self) -> StringSlice[StaticConstantOrigin]: ...

trait SliceableAlloc:
  fn __str__(self) -> StringSlice[AnonymousOrigin]: ... # allocated inside the body of the func
```

And then having 5 overloads of the `str` builtin in the mean time since we don't yet have powerful enough generics to express this.

soraros:
> If we make the `str` builtin return a `StringSlice` would it qualify as being like Python's ? It would be immutable in length, but mutable in-place

This sounds like a good idea. I'd prefer it over `as_string_slice`.

gryznar:
> If we make the `str` builtin return a `StringSlice` would it qualify as being like Python's ? It would be immutable in length, but mutable in-place

Has it to be like in Python? Recent changes show, that Mojo is no longer following path to just copy Python API. I'd prefer to have good API on its own. Ok, Python did many things well, but in many design decisions fails short.

I see Rust as better inspirer towards good design decisions (better, but still not the best).

[...]

PR #4313, opened 2025-04-08
[stdlib] Change DLHandle constructor to use os.PathLike objects.

thatstoasty:
Changed the `DLHandle` constructor to take types that implement that the `os.PathLike` trait instead of just `String`. This should improve the UX a little bit by supporting `Path` objects.

This adds an allocation of a new `String` from the usage of `PathLike.__fspath__()`. Perhaps `__fspath__` can be updated to return a `StringSlice` instead of a `String`?


martinvuyk:
> This adds the allocation of a new from the usage of `PathLike.__fspath__()`. Perhaps `__fspath__` can be updated to return a `StringSlice` instead of a `String`?

+1. I think this falls in the same category as the discussion I had with @soraros in #4217.

I would argue making `__str__` return a `StringSlice` should be a priority, and then try to take that to other cases like fspath. The only small problem is that we don't have parametric traits yet, so the implementation will get nasty.

BUT, the big problem is that `StringSlice` is not null terminated. And anything to do with C APIs needs to be. So until we have a `StringSliceC` (or something) that is guaranteed to be null terminated, fspath will have to wait

thatstoasty:
@martinvuyk Ah okay, at first I didn't see the vision for bringing back a `str` function but now I see where you're going with it! Agreed, I thought StringSlice already supported cstr pointers, but I see that it doesn't. Makes sense.

JoeLoser (approved):

JoeLoser:
!sync

PR #4316, opened 2025-04-08
[stdlib] Added `os.path.split_extension` and tests.

thatstoasty:
Also added a few extra tests for `splitroot` from the Python stdlib test cases for `PosixPath`.


gryznar:
Same question as with second pr about inconsistency

mojo/stdlib/src/os/path/path.mojo
```diff
@@ -420,6 +420,68 @@ fn basename[PathLike: os.PathLike, //](path: PathLike) -> String:
...
+    var head_end = path.rfind(sep)
+    if alt_sep:
+        head_end = max(head_end, path.rfind(alt_sep))
+
+    var file_end = path.rfind(extension_sep)
+    if file_end > head_end:
+        # skip all leading dots
+        var file_start = head_end + 1
+        while file_start < file_end:
+            if path[file_start].as_string_slice() != extension_sep:
```
martinvuyk, line 457: I didn't know that `StringSlice.__getitem__` allocated. I'm opening up a PR to change that.

martinvuyk (commented):
Hi, in general I kind of agree with the sentiment of wanting to make stuff snake case as much as possible. But I don't know to what extend the community and the Mojo team are willing to deviate from Python. I personally find many function names in the os package very ugly and unintuitive. I think we need to come to a consensus at some point.

gryznar:
I don't see how sticking to Python API will help in getting functionality with leading performance which is crucial for AI / system-level related stuff

thatstoasty:
@JoeLoser Is there any preference internally regarding naming of API that were ported over from Python? I don't think its a big lift for a python developer to update `os.path.splitext` to something like `os.path.split_extension`, would that be an undesirable departure?

JoeLoser:
> @JoeLoser Is there any preference internally regarding naming of API that were ported over from Python? I don't think its a big lift for a python developer to update `os.path.splitext` to something like `os.path.split_extension`, would that be an undesirable departure?

+1, let's go with `split_extension`, please.  Thanks for asking!

JoeLoser (approved):

JoeLoser:
!sync

PR #4317, opened 2025-04-08
[stdlib] Add `os.path.is_absolute` and tests.

thatstoasty:


gryznar:
Why do we want to have language inconsistency where 90% of code is snake_case? Maybe just to use `is_abs()` instead? As I am aware, Mojo does not aim to copy Python

thatstoasty:
@gryznar Personally, I would rather use snake case and abandon the legacy Python naming scheme as well. But AFAIK there is no directive to rename Python apis at this moment. I am following the same naming as Python and porting the same behavior (ideally) as Python for familiarity. If anyone from the stdlib team wants to weigh in here, I'll gladly update the function names :)

gryznar:
So, we should ask Joe Loser as stdlib leader if copying Python is still the case. Personally I see it like big balast which does not allow to have consistent API with good design in mind

JoeLoser (approved):
LGTM. Can you also add a small changelog entry for the new functionality, please?

JoeLoser:
+1 to spelling out the function name, eg is_absolute, for example. We don't need to copy the Python naming.

gryznar:
@thatstoasty please also adjust pr title to recent changes (dot at the end is not needed) ;)

thatstoasty:
@JoeLoser All done! I've added a changelog entry.

JoeLoser (approved):

JoeLoser:
!sync

Issue #3849, opened 2024-12-09
[BUG] File open does not have the append flag


martinvuyk:
### Bug description

File open does not have the append flag.

### Steps to reproduce

This works in python but doesn't in Mojo
```mojo
from os import remove

def main():
    name = "some_file.txt"

    with open(name, "w") as f:
        f.write("new contents 🔥")

    with open(name, "r") as f:
        contents = f.read()
        print(contents)

    with open(name, "a") as f:
        f.write("add another fire 🔥")

    with open(name, "r") as f:
        contents = f.read()
        print(contents)
    
    remove(name)
```

### System information

```shell
- Provide version information for Mojo by pasting the output of `mojo -v`
`mojo 24.6.0.dev2024120805`
```

gustawdaniel:
Right now I am using temporary method:

```mojo
fn append_csv(file_path:String, time_sec:Float64) raises:
    # Try to read existing file
    try:
        var file_r = open(file_path, "r")
        contents = file_r.read()
        file_r.close()
    except:
        # File does not exist yet, start empty
        contents = ""

    # Append new line (always add newline)
    contents += String(time_sec) + "\n"

    # Write back
    var file_w = open(file_path, "w")
    file_w.write(contents)
    file_w.close()
```

to append

martinvuyk:
https://github.com/modular/modular/commit/606b639e7ed1350bf602f80e03528e80b646fb24, thanks @JoeLoser !

Issue #3747, opened 2024-11-05
[Feature Request] [stdlib] [proposal] Define `BinarySerializable`, `BinaryDeserializable`, and `BinaryStreamable` traits


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Define `Serializable`, `Deserializable`, and `Streamable` traits

### What is your motivation for this change?

#3744 and #3745 are reference implementations for `Writer` applied to those collection types. They make no sense as they are because the concept of `Writer` is basically conceptually constrained to String streaming. This needs to scale out further.

## Proposals
I will bunch up many thoughts inside this proposal, they can be implemented or not individually.
#### 1. Rename `Span` to `Buffer`
`Buffer` is a known word and concept, it is the preferred word in C. `Span` is not so clear for non native English speakers, and is not one the first words one thinks of when talking about data stored in memory. `Buffer` is also what best describes this structure since it can also mutate (in the future maybe also own) the data.

#### 2. Prioritize and strengthen generics for working with `Span`
Parametrized origins are very clunky. See #3744 for an example of a method to enable reads or writes to a `Span`.

#### 3. Define a generic trait for serialization
proposed pseudocode (we don't have parametrized traits yet so we can constrain it to `Byte` for now)
(A type could potentially serialize to more than 1 type of Buffer. e.g. String could be encoded to different UTF standards)
```mojo
trait Serializable[D: DType]:
    fn __encode__[
        is_mutable: Bool, origin: Origin[is_mutable].type
    ](self) -> Buffer[Scalar[D], origin]:
        ...
```

#### 4. Define a generic trait for binary deserialization
(A type could potentially deserialize from more than 1 type of Buffer. e.g. String could be decoded from different UTF standards)
```mojo
trait BinaryDeserializableByCopy[D: DType]:
    fn __init__(out self, *, buffer: Buffer[Scalar[D]]):
        ...

# if we do make buffer be able to own the data
trait BinaryDeserializableZeroCopy[D: DType]:
    fn __init__[O: MutableOrigin](
        out self,
        *,
        owned buffer: Buffer[Scalar[D], MutableOrigin],
        self_is_owner: Bool,
    ):
        ...
```

#### 4. Define generic traits for binary streaming

[...]

PR #4318, opened 2025-04-08
[stdlib] Add `write_bytes` method to `pathlib.Path`

thatstoasty:
I updated `write_text` to use `Writable` arguments instead of Stringable to avoid the allocation. I think its in alignment with the general direction of using `Writable` for String streaming.


mojo/stdlib/src/pathlib/path.mojo
```diff
@@ -314,6 +314,17 @@ struct Path(
         """
         return os.path.isfile(self)
 
+    fn open(self, mode: StringSlice) raises -> FileHandle:
+        """Opens the file at the path provided.
+
+        Args:
+            mode: The mode to open the file with. For example: "r", "w", "a".
```
martinvuyk: Sadly, File open doesn't have the append flag yet (see #3849, it's also stated in the function docstrings)
```suggestion
            mode: The mode to open the file with. For example: "r", "w".
```

mojo/stdlib/src/pathlib/path.mojo
```diff
@@ -314,6 +314,17 @@ struct Path(
...
+
+        Returns:
+            The opened file.
+        """
+        return os.open(self, mode)
```
martinvuyk: Question: is this a common pattern in python? I've personally only used the `open` function directly in python, never through the os package. If so, then we need to import `file.open` in `os.__init__` because it is currently not, so this shouldn't actually work.

mojo/stdlib/src/pathlib/path.mojo
```diff
@@ -332,17 +343,26 @@ struct Path(
...
-          stringable: The Stringable type.
+            T: The type of an object conforming to the `Writable` trait.
+
+        Args:
+            value: The value to write.
+        """
+        with open(self, "w") as f:
+            f.write(value)
+
+    fn write_bytes(self, bytes: Span[Byte, _]) raises:
```
martinvuyk: I think this works now, but I'm not sure.
```suggestion
    fn write_bytes(self, bytes: Span[Byte]) raises:
```

martinvuyk (commented):
Hi, thanks for working on this. Just a couple of details

mojo/stdlib/src/pathlib/path.mojo
thatstoasty: Ah yes, it should just be using the built in open function. I'll potentially remove Path.open though. It feels fairly redundant to have path.open() and open(path).

thatstoasty: I removed `open` since I think it doesn't offer much over `open(path)`. So I'll close this conversation and the one above!

gryznar:
Maybe better would be to just rethink whole path (OS) handling design and came with different approach, than replicating Python modules structure? I mean involving asyncness / Mojo parameters etc.

martinvuyk:
> Maybe better would be to just rethink whole path (OS) handling design and came with different approach, than replicating Python modules structure? I mean involving asyncness / Mojo parameters etc.

Where exactly would asyncness be necessary? I can't think of good examples in string handling that would need it.

In the case of using parameters/other runtime optimizations, it's something I've been musing at mentally. The conclusion I came to is that most of these operations don't necessarily require top of the line performance. I could be wrong, but I also think there is a risk of over-engineering in the path package. The only thing I would strongly insist is to use `StringSlice` wherever possible instead of `String` which is actually more similar to Python's strings when slicing.

gryznar:
I mean asyncness for I/O operations. `Path` in Python handles not only pure paths as `PurePath`, but also IO. Pr is for `Path.write_bytes()` which easily can be handled asynchronously

martinvuyk:
That is another topic and the operations themselves take place in file.mojo. We still need some better language features to have flexible streaming traits, etc. Ask Owen if you want your head to explode with all the implications of generic async streaming.

Read this discussion if you're interested: #3747

JoeLoser (approved):
LGTM, thanks for working on this! Want to add a changelog entry?

thatstoasty:
@JoeLoser Done here as well! I've added the changelog entry for the changes in the PR.

JoeLoser (approved):

JoeLoser:
!sync

PR #4319, opened 2025-04-08
[stdlib] Change `StringSlice.__getitem__` to return a `StringSlice`

martinvuyk:
Had to do some black magic to get this to work. But we totally needed to stop allocating so much for such cases.

### DIFF
--- a/mojo/stdlib/src/collections/string/string_slice.mojo
+++ b/mojo/stdlib/src/collections/string/string_slice.mojo
@@ -49,6 +49,7 @@ Example:
 """
 
 from collections import List, Optional
+from collections._index_normalization import normalize_index
 from collections.string._unicode import (
     is_lowercase,
     is_uppercase,
@@ -963,7 +964,7 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
         """
         return CodepointSliceIter[origin, forward=False](self)
 
-    fn __getitem__[I: Indexer](self, idx: I) -> String:
+    fn __getitem__[I: Indexer](self, idx: I) -> Self.Immutable:
         """Gets the character at the specified position.
 
         Parameters:
@@ -975,11 +976,9 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
         Returns:
             A new string containing the character at the specified position.
         """
-        # TODO(#933): implement this for unicode when we support llvm intrinsic evaluation at compile time
-        var buf = String._buffer_type(capacity=1)
-        buf.append(self._slice[idx])
-        buf.append(0)
-        return String(buf^)
+        # TODO(#3526): implement this for unicode
+        var i = normalize_index["StringSlice"](idx, len(self))
+        return Self.Immutable(ptr=self.unsafe_ptr() + i, length=1)
 
     fn __contains__(self, substr: StringSlice) -> Bool:
         """Returns True if the substring is contained within the current string.
--- a/mojo/stdlib/src/testing/testing.mojo
+++ b/mojo/stdlib/src/testing/testing.mojo
@@ -48,10 +48,12 @@ fn _assert_error[T: Writable](msg: T, loc: _SourceLocation) -> String:
 
 @always_inline
 fn assert_true[
-    T: Boolable, //
+    T: Boolable, //, O: ImmutableOrigin = StaticConstantOrigin
 ](
     val: T,
-    msg: String = "condition was unexpectedly False",
+    msg: StringSlice[O] = rebind[StringSlice[O]](
+        StaticString("condition was unexpectedly False")
+    ),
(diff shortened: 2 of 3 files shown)


mojo/stdlib/src/testing/testing.mojo
```diff
@@ -49,17 +49,20 @@ fn _assert_error[T: Writable](msg: T, loc: _SourceLocation) -> String:
 
 @always_inline
 fn assert_true[
-    T: Boolable, //
+    T: Boolable, //, O: ImmutableOrigin = StaticConstantOrigin
```
lsh, line 51: We shouldn't be using default origins. I've been removing them slowly. This is a good reminder that I should probably explain that publicly :)

mojo/stdlib/src/testing/testing.mojo
```diff
@@ -117,11 +120,11 @@ trait TestableCollectionElement(
 
 @always_inline
 fn assert_equal[
-    T: Testable
+    T: Testable, O: ImmutableOrigin = StaticConstantOrigin
 ](
     lhs: T,
     rhs: T,
-    msg: String = "",
```
lsh, line 123: Let's leave optimizing the testing utils for a separate PR.

lsh (changes requested):

mojo/stdlib/src/testing/testing.mojo
martinvuyk: How do you suggest I let people pass a `StringSlice` as `msg` otherwise? And the default still needs to be an empty string. I could also make it `msg: Optional[StringSlice] = None`

mojo/stdlib/src/testing/testing.mojo
martinvuyk: It is not about optimization, this is used in the stdlib and so I had to adjust it. Many tests index into a `StringSlice` and send it as the `msg`

mojo/stdlib/src/testing/testing.mojo
martinvuyk: I tried now to make it take an optional, but the origin cannot be inferred without a stringslice being passed in, as has always been the case:
```mojo
@always_inline
fn assert_true[
    T: Boolable, O: ImmutableOrigin, //
](
    val: T,
    msg: Optional[StringSlice[O]] = None,
    *,
    location: Optional[_SourceLocation] = None,
) raises:
    if not val:
        raise _assert_error(msg.or_else(rebind[StringSlice[O]](StaticString(""))), location.or_else(__call_location()))
```
```text
failed to infer parameter 'O', parameter isn't used in any argument
```
only when setting the default `O: ImmutableOrigin = StaticConstantOrigin` does it work. Which kind of defeats the purpose of using an optional in the first place.

martinvuyk: > We shouldn't be using default origins

I would also like to know why. I don't see the problem with setting something as `StaticConstantOrigin` when there is no other option (if you want to set a compile time constant as a default function argument it literally has that origin). The type system should infer other runtime origins without a problem.

mojo/stdlib/src/collections/string/string_slice.mojo
```diff
@@ -963,7 +964,7 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
         """
         return CodepointSliceIter[origin, forward=False](self)
 
-    fn __getitem__[I: Indexer](self, idx: I) -> String:
+    fn __getitem__[I: Indexer](self, idx: I) -> Self.Immutable:
```
lattner, line 967: Why doesn't this return Self?

lattner:
Martin, Joe pointed me at this PR.  I'm making a ton of changes to improve String and surely conflicted all over this PR.  I'll adopt some of your changes into a separate patch that lands this.  I'm sorry for stepping on your work!

lattner:
Landed and should be in the next update, thank you again.

mojo/stdlib/src/collections/string/string_slice.mojo
martinvuyk: There were some origin conflicts because of 2 mutating references. I didn't chase it down to find the exact places.

martinvuyk:
@lattner this is still happening in the latest nightly:
```mojo
fn __getitem__[I: Indexer](self, idx: I) -> String:
```
I can't reopen this PR

martinvuyk:
@lattner with the new inlining of `String` I don't think this optimization is needed anymore since `StringSlice` would introduce extra pointer indirection. When this supports unicode indexing the 1-4 bytes will always fit in SSO :+1:. The only problem will be with Grapheme clusters, but that can also be potentially solved with parametrized overloads (if  #3988 is accepted)

PR #4323, opened 2025-04-09
[stdlib] Add check for file open mode and raise on error

martinvuyk:
Add check for file open mode and raise on error

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_string.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_string.mojo
@@ -45,7 +45,7 @@ fn make_string[
 
     try:
         directory = _dir_of_current_file() / "data"
-        var f = open(directory / filename, "rb")
+        var f = open(directory / filename, "r")
 
         @parameter
         if length > 0:
--- a/mojo/stdlib/stdlib/builtin/file.mojo
+++ b/mojo/stdlib/stdlib/builtin/file.mojo
@@ -79,9 +79,20 @@ struct FileHandle(Defaultable, Movable, Writer):
         """Construct the FileHandle using the file path and mode.
 
         Args:
-          path: The file path.
-          mode: The mode to open the file in (the mode can be "r" or "w" or "rw").
+            path: The file path.
+            mode: The mode to open the file in: {"r", "w", "rw"}.
+
+        Raises:
+            If file open mode is not one of the supported modes.
+            If there is an error when opening the file.
         """
+        # TODO(#3849): this should support the append flag
+        if not (mode == "r" or mode == "w" or mode == "rw"):
+            raise Error(
+                'ValueError: invalid mode: "',
+                mode,
+                '". Can only be one of: {"r", "w", "rw"}',
+            )
         var err_msg = _OwnedStringRef()
         var handle = external_call[
             "KGEN_CompilerRT_IO_FileOpen", OpaquePointer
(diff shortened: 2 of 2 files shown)


mojo/stdlib/src/builtin/file.mojo
```diff
@@ -80,9 +80,20 @@ struct FileHandle(Writer):
...
+        # TODO(MSTDL-1071):
+        #   Once Mojo supports parametric traits, implement EqualityComparable for
+        #   StringSlice such that string slices with different origin types can be
+        #   compared, then drop this horrible hack.
+        debug_assert(
+            mode == rebind[__type_of(mode)](StaticString("r"))
+            or mode == rebind[__type_of(mode)](StaticString("w"))
+            or mode == rebind[__type_of(mode)](StaticString("rw")),
+            'mode can only be one of: {"r", "w", "rw"}',
+        )
```
laszlokindrat: Q: This function already raises; why not turn this into a proper runtime check?

martinvuyk: Hi laszlo, long time no see :)

I guess we could, didn't occur to me

mojo/stdlib/src/builtin/file.mojo
```diff
@@ -82,9 +82,24 @@ struct FileHandle(Writer):
...
+            path: The file path.
+            mode: The mode to open the file in: {"r", "w", "rw"}.
         """
+        # FIXME(#3849): this should support the append flag
+        # TODO(MSTDL-1071):
+        #   Once Mojo supports parametric traits, implement EqualityComparable for
+        #   StringSlice such that string slices with different origin types can be
+        #   compared, then drop this horrible hack.
+        if not (
+            mode == rebind[__type_of(mode)](StaticString("r"))
```
soraros: Do we need the `rebind` here? I think `__merge_with__` should have fixed this problem.

You can even remove the `StaticString`.

martinvuyk: You're right!!

mojo/stdlib/src/builtin/file.mojo
```diff
@@ -82,9 +82,16 @@ struct FileHandle(Writer):
         """Construct the FileHandle using the file path and mode.
 
         Args:
-          path: The file path.
-          mode: The mode to open the file in (the mode can be "r" or "w" or "rw").
+            path: The file path.
+            mode: The mode to open the file in: {"r", "w", "rw"}.
         """
+        # TODO(#3849): this should support the append flag
+        if not (mode == "r" or mode == "w" or mode == "rw"):
```
laszlokindrat: Does this work by any chance?
```suggestion
        if mode not in ("r", "w", "rw"):
```

laszlokindrat (approved):
Thanks! Could you please rebase so I can bring it in?

mojo/stdlib/src/builtin/file.mojo
martinvuyk: It returns false because the types are unequal in the `Tuple.__contains__()` body:
```mojo
        @parameter
        for i in range(len(VariadicList(element_types))):

            @parameter
            if _type_is_eq[element_types[i], T]():
                if rebind[T](self[i]) == value:
                    return True

        return False
```

martinvuyk:
@laszlokindrat I think somebody turned on a stricter linter and it broke every OS PR, because I can't find the issue in the code I added. I also rebased #4416 and the same thing happened

laszlokindrat:
@martinvuyk Sorry I kinda dropped this... Could you please rebase on the latest so I can bring it in?

martinvuyk:
@laszlokindrat done :)

laszlokindrat:
!sync

PR #4328, opened 2025-04-10
[stdlib] Add implicit casting from `MutableOrigin` -> `ImmutableOrigin` for `Pointer`

martinvuyk:
Sister PR of  #3823

### DIFF
--- a/mojo/stdlib/src/memory/pointer.mojo
+++ b/mojo/stdlib/src/memory/pointer.mojo
@@ -317,6 +317,7 @@ struct Pointer[
         address_space: The address space of the pointee data.
     """
 
+    # Aliases
     alias _mlir_type = __mlir_type[
         `!lit.ref<`,
         type,
@@ -326,13 +327,34 @@ struct Pointer[
         address_space._value.value,
         `>`,
     ]
+    alias _with_origin = Pointer[type, _, address_space]
+    alias _immutable_cast = ImmutableOrigin.cast_from[_]
 
+    alias Mutable = Self._with_origin[MutableOrigin.cast_from[origin].result]
+    """The mutable version of the `Pointer`."""
+    alias Immutable = Self._with_origin[Self._immutable_cast[origin].result]
+    """The immutable version of the `Pointer`."""
+    # Fields
     var _value: Self._mlir_type
     """The underlying MLIR representation."""
 
     # ===------------------------------------------------------------------===#
     # Initializers
     # ===------------------------------------------------------------------===#
+
+    @doc_private
+    @implicit
+    @always_inline("nodebug")
+    fn __init__(
+        other: Self._with_origin[_],
+        out self: Self._with_origin[Self._immutable_cast[other.origin].result],
+    ):
+        """Implicitly cast the mutable origin of self to an immutable one.
+
+        Args:
+            other: The `Pointer` to cast.
+        """
+        self = rebind[__type_of(self)](other)
 
     @doc_private
     @always_inline("nodebug")
@@ -378,23 +400,17 @@ struct Pointer[
         return Self(_mlir_value=self._value)
 
     @always_inline
-    fn get_immutable(
-        self,
-        out result: Pointer[
-            type,
-            ImmutableOrigin.cast_from[origin].result,
-            address_space=address_space,
-        ],
-    ):
+    fn get_immutable(self) -> Self.Immutable:
         """Constructs a new Pointer with the same underlying target
         and an ImmutableOrigin.
 
-        Note that this does **not** copy the underlying data.
-
         Returns:
             A new Pointer with the same target as self and an ImmutableOrigin.
+
+        Notes:
+            This does **not** copy the underlying data.
         """
-        return __type_of(result)(_mlir_value=self._value)
+        return rebind[Self.Immutable](self)
 
     # ===------------------------------------------------------------------===#
     # Operator dunders
--- a/mojo/stdlib/test/memory/test_pointer.mojo
+++ b/mojo/stdlib/test/memory/test_pointer.mojo
@@ -49,14 +49,11 @@ def test_pointer_to():
 # We don't actually need to run this,
 # but Mojo's exclusivity check shouldn't complain
 def test_get_immutable() -> Int:
-    fn foo(x: Pointer[Int], y: Pointer[Int]) -> Int:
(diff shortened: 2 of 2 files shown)


soraros:
Subtyping through implicit conversion, neat.

mojo/stdlib/src/memory/pointer.mojo
```diff
@@ -326,14 +327,34 @@ struct Pointer[
         address_space._value.value,
         `>`,
     ]
-
+    alias _UnboundOrigin = Pointer[type, _, address_space]
+    alias MutSelf = Self._UnboundOrigin[MutableOrigin.cast_from[origin].result]
+    """The mutable version of the `Pointer`."""
+    alias _ImmutCast = ImmutableOrigin.cast_from[_]
+    alias ImmutSelf = Self._UnboundOrigin[Self._ImmutCast[origin].result]
+    """The immutable version of the `Pointer`."""
```
soraros: Nit; rational: make function-like things look functional. I also think we should make `_immut_cast` a file level alias.

```suggestion
    alias _with_origin = Pointer[type, _, address_space]
    alias _immut_cast = ImmutableOrigin.cast_from[_]

    alias MutSelf = Self._with_origin[MutableOrigin.cast_from[origin].result]
    """The mutable version of the `Pointer`."""
    alias ImmutSelf = Self._with_origin[Self._immut_cast[origin].result]
    """The immutable version of the `Pointer`."""
```

soraros: In fact, why do we use `MutableOrigin.cast_from` directly in `MutSelf` but `_ImmutCast` in `ImmutSelf`? I *think* we want both or neither (my preference).

martinvuyk: > rational: make function-like things look functional.

I don't think these are really functional, I'm stating that they are types with unbound parameters. `cast_from` is also not a function but a parametrized struct.

> ```mojo
> alias _with_origin = Pointer[type, _, address_space]
> ```

This one is interesting on the naming choice per se. But it may not make much sense in the init fn:
```mojo
    fn __init__(
        other: Self._with_origin, # what is this?
        #  maybe it would make sense using: Self._with_origin[_]
        out self: Self._with_origin[Self._immut_cast[other.origin].result], # makes sense
    ):
```

> In fact, why do we use MutableOrigin.cast_from directly in MutSelf but _ImmutCast in ImmutSelf? I think we want both or neither (my preference).

I added `_UnboundOrigin` and `_ImmutCast` mostly to avoid going over the character limit and to make reading the function signature easier on the eyes. I don't really care what they are called :man_shrugging:

martinvuyk: I have tried and failed many times to setup some better casting system for `Origin` though, I also wish it were prettier to write.

soraros: I'd still vote for `Self._with_origin[_]`. Anyways, it's your call to make.

> I don't think these are really functional, I'm stating that they are types with unbound parameters.

Which is essentially a type level lambda. This pattern is called a functor in C++ I believe (which I don't approve for I also speak Haskell 🙃).

martinvuyk: > Which is essentially a type level lambda. This pattern is called a functor in C++ I believe (which I don't approve for I also speak Haskell 🙃).

:rofl: fine idk anything about fancy PL nomenclature, I just use them

mojo/stdlib/src/memory/pointer.mojo
```diff
@@ -326,14 +327,35 @@ struct Pointer[
         address_space._value.value,
         `>`,
     ]
-
+    alias _with_origin = Pointer[type, _, address_space]
+    alias _immut_cast = ImmutableOrigin.cast_from[_]
+
+    alias MutSelf = Self._with_origin[MutableOrigin.cast_from[origin].result]
+    """The mutable version of the `Pointer`."""
+    alias ImmutSelf = Self._with_origin[Self._immut_cast[origin].result]
```
lsh: `immut` isn't really a naming convention we use anywhere. We should prefer the full names here `Mutable` and `Immutable`. Since we can access them from `Self.Immutable`, that seems cleanest.

lsh (changes requested):
Thanks for trying this out. It's cool that this works, though I want some naming changes.

martinvuyk:
@lsh  done

mojo/stdlib/src/memory/pointer.mojo
```diff
@@ -326,14 +327,35 @@ struct Pointer[
         address_space._value.value,
         `>`,
     ]
-
+    alias _with_origin = Pointer[type, _, address_space]
+    alias _immut_cast = ImmutableOrigin.cast_from[_]
```
lsh: Same with this, it should be `_immutable_cast`

mojo/stdlib/src/memory/pointer.mojo
```diff
@@ -326,14 +327,35 @@ struct Pointer[
...
     # ===------------------------------------------------------------------===#
     # Initializers
     # ===------------------------------------------------------------------===#
 
+    @doc_private
+    @implicit
+    @always_inline("nodebug")
+    fn __init__(
+        other: Self._with_origin[_],
+        out self: Self._with_origin[Self._immut_cast[other.origin].result],
```
lsh: can `out self: Self.Immutable` work?

soraros: Not really, the return type has to depend on `other` in order to work. Related: https://github.com/modular/max/issues/3845#issuecomment-2791450758

lsh: Ah! Good catch

lsh:
!sync

mojo/stdlib/src/memory/pointer.mojo
```diff
@@ -326,14 +327,35 @@ struct Pointer[
...
+    @implicit
+    @always_inline("nodebug")
+    fn __init__(
+        other: Self._with_origin[_],
+        out self: Self._with_origin[Self._immutable_cast[other.origin].result],
+    ):
+        """Implicitly cast the mutable origin of self to an immutable one.
+
+        Args:
+            other: The Span to cast.
```
gryznar: Span? Not "Pointer"?

martinvuyk: nice catch

lsh:
!sync

PR #4332, opened 2025-04-10
[stdlib] Make parameters of `min`/`max` explicit

soraros:
This allows us to disambiguate and use them in higher-order functions like `f[min[dtype=...]]()`.

I introduced this change in #3901, but @latter inadvertently 'cleaned it up' in a subsequent commit.


JoeLoser (approved):

JoeLoser:
!sync

PR #4341, opened 2025-04-12
[stdlib] Fix and modernise `my_complex.mojo`

soraros:
- `__ne__` and `__bool__` were implemented incorrectly

### DIFF
--- a/examples/mojo/operators/main.mojo
+++ b/examples/mojo/operators/main.mojo
@@ -19,24 +19,24 @@ from my_complex import Complex
 def main():
     # Examples of creating Complex instances
     c1 = Complex(-1.2, 6.5)
-    print(StaticString("c1: Real: {}; Imaginary: {}").format(c1.re, c1.im))
+    print("c1 =", repr(c1))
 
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
@soraros this broke the `test_my_complex.mojo`, so this is being reverted.  It doesn't look too hard to fix, but please run `mojo test test_my_complex.mojo` and then reopen the PR.  In the revert description:

> Reverts https://github.com/modular/max/pull/4341 that was imported. This breaks test_my_complex.mojo. Some of the breakage is trivial (e.g. this PR changed getitem to check its index with constrained, so you need to remove the with assert_raises for out of bounds index since they will not raise). However, there are other unit test failures, e.g. in test_binary_rfloat. Just revert and let the OP figure out the test failures.

We should extend the open source CI to run this example and others too in the future to avoid late breakage.

soraros:
@JoeLoser Opened #4355.

PR #4345, opened 2025-04-14
[stdlib] Improve type refinement for `pack_bits`

soraros:
So the following works without `rebind`:
```mojo
var v: UInt8 = pack_bits(SIMD[DType.bool, 8](1))
```

`_uint` gets inlined even when `n` isn't concrete, cluttering tooltips and types.  
Can call-by-value be triggered only when `n` is concrete?

CC @lattner since it's related to `@always_inline("builtin")`.

### DIFF
--- a/mojo/stdlib/src/memory/unsafe.mojo
+++ b/mojo/stdlib/src/memory/unsafe.mojo
@@ -84,16 +84,19 @@ fn bitcast[
     ](val.value)
 
 
-@always_inline("nodebug")
+@always_inline("builtin")
 fn _uint(n: Int) -> DType:
-    if n == 8:
-        return DType.uint8
-    elif n == 16:
-        return DType.uint16
-    elif n == 32:
-        return DType.uint32
-    else:
-        return DType.uint64
+    # fmt: off
+    return (
+        DType.uint8 if n == 8
+        else DType.uint16 if n == 16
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/memory/unsafe.mojo
```diff
@@ -84,16 +84,19 @@ fn bitcast[
...
-        return DType.uint8
-    elif n == 16:
-        return DType.uint16
-    elif n == 32:
-        return DType.uint32
-    else:
-        return DType.uint64
+    # fmt: off
+    return (
+        DType.uint8 if n == 8
```
soraros, line 91: Using the `if` expression since `hlcf.if` is not supported in `builtin`.

JoeLoser (approved):

JoeLoser:
!sync

lattner:
Thanks @soraros .  Just catching up: the point of always_inline(builtin) isn't to inline everything that "could be constant folded".  It is really about things that are necessary to use in dependent types.  There is longer term work to improve the Mojo intepreter that may make it more powerful - this will be useful for this sort of thing as well as for requires clauses - but the parser-time interpreter will always be limited to not knowing about the target (e.g. pointer sizes) so it will be inherently limited in what it can do.

PR #4353, opened 2025-04-16
[stdlib] Fix sqrt loop range and rewrite to prevent overflow

### LINKED ISSUE #4352: [BUG] math.sqrt[Int] returns incorrect results for large inputs
### Actual behavior
Passing any integer ≥ `2**34` to math.sqrt[Int] always returns `131071` (`2**17 - 1`), regardless of the input.

### Expected behavior
`sqrt(x)` should return the correct integer square root of `x`, even when `x >= 2**34`.

### Steps to reproduce

Run the following code:

```
from math import sqrt

fn main():
    print(sqrt(10**16))
```

It prints 131071, but the correct result should be 100000000.

### System information

- `magic list max`
```
Package     Version  Build    Size       Kind   Source
max         25.2.0   release  9.6 KiB    conda  max
max-core    25.2.0   release  251.7 MiB  conda  max-core
max-python  25.2.0   release  121.4 MiB  conda  max-python
```

shogo314:
Fix https://github.com/modular/max/issues/4352

This PR fixes a bug in the `sqrt` function in the standard library.

### Summary of the problem
The loop previously started at `bitwidthof[Int]() // 4`, which was too small to handle larger integer inputs. This resulted in incorrect square root values in some cases.

### Changes made
- Corrected the loop range to start from `((bitwidthof[Int]() - 1) // 2)` to fully cover the bit width of `Int`.
- Rewrote the condition to avoid intermediate overflow by reformulating the check:
  ```mojo
  if r2 <= x - (r << (p + 1)) - (1 << (2 * p)):

### DIFF
--- a/mojo/stdlib/src/math/math.mojo
+++ b/mojo/stdlib/src/math/math.mojo
@@ -191,10 +191,10 @@ fn sqrt(x: Int) -> Int:
     var r2 = 0
 
     @parameter
-    for p in range(bitwidthof[Int]() // 4, -1, -1):
-        var tr2 = r2 + (r << (p + 1)) + (1 << (p + p))
-        if tr2 <= x:
-            r2 = tr2
+    for p in reversed(range(bitwidthof[Int]() // 2)):
+        var dr2 = (r << (p + 1)) + (1 << (p + p))
+        if r2 <= x - dr2:
+            r2 += dr2
             r |= 1 << p
 
     return r
--- a/mojo/stdlib/test/math/test_math.mojo
+++ b/mojo/stdlib/test/math/test_math.mojo
@@ -184,6 +184,14 @@ alias F64x4 = SIMD[DType.float64, 4]
 
 
 def test_sqrt():
+    assert_equal(sqrt(-1), 0)
+    assert_equal(sqrt(0), 0)
+    assert_equal(sqrt(1), 1)
+    assert_equal(sqrt(2**34 - 1), 2**17 - 1)
+    assert_equal(sqrt(2**34), 2**17)
+    assert_equal(sqrt(10**16), 10**8)
+    assert_equal(sqrt(Int.MAX), 3037000499)
+
     var i = SIMD[DType.index, 4](0, 1, 2, 3)
     assert_equal(sqrt(i**2), i)
     assert_equal(sqrt(64), 8)
--- a/mojo/stdlib/test/math/test_math.mojo
+++ b/mojo/stdlib/test/math/test_math.mojo
@@ -187,6 +187,8 @@ def test_sqrt():
     assert_equal(sqrt(-1), 0)
     assert_equal(sqrt(0), 0)
     assert_equal(sqrt(1), 1)
+    assert_equal(sqrt(63), 7)
+    assert_equal(sqrt(64), 8)
     assert_equal(sqrt(2**34 - 1), 2**17 - 1)
     assert_equal(sqrt(2**34), 2**17)
     assert_equal(sqrt(10**16), 10**8)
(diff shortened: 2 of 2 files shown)


gryznar (commented):
Could you add testcase covering thing that previously did not work?

shogo314:
I've added the file `mojo/stdlib/test/math/test_sqrt.mojo`.
Let me know if there's anything you'd like me to adjust!

The last three test cases are the ones that would fail with the previous implementation.

gryznar:
Thanks for that! I am normal community member, so I cannot sync your pr, but from me is +1

mojo/stdlib/test/math/test_sqrt.mojo
gryznar, line 1: This file should be named IMO: `test_math.mojo` as related src. It is also strange, that math does not have tests

mojo/stdlib/src/math/math.mojo
```diff
@@ -191,10 +191,9 @@ fn sqrt(x: Int) -> Int:
     var r2 = 0
 
     @parameter
-    for p in range(bitwidthof[Int]() // 4, -1, -1):
-        var tr2 = r2 + (r << (p + 1)) + (1 << (p + p))
-        if tr2 <= x:
-            r2 = tr2
+    for p in range((bitwidthof[Int]() - 1) // 2, -1, -1):
```
soraros: Nit:
```suggestion
    for p in reversed(range(bitwidthof[Int]() // 2)):
```

mojo/stdlib/src/math/math.mojo
```diff
@@ -191,10 +191,9 @@ fn sqrt(x: Int) -> Int:
...
+        if r2 <= x - (r << (p + 1)) - (1 << (p + p)):
```
soraros: Do you mind putting a comment above this line stating that it's written this way (with code duplication) to prevent overflow?

shogo314:
I actually agree that using `reversed(range(...))` is clearer!
I had originally written it to match the existing style in nearby code, but I’m happy to change it for readability.

Just to clarify — the duplication of `(r << (p + 1)) + (1 << (p + p))` isn't itself meant to prevent overflow.
The key reason for the current structure is to avoid writing the condition as:
```
r2 + (r << (p + 1)) + (1 << (p + p)) <= x
```
which could overflow even when the logic would otherwise be correct.
If preferred, I can extract the duplicated expression into a temporary variable for clarity:
```
var dr2 = (r << (p + 1)) + (1 << (p + p))
if r2 <= x - dr2:
    r2 += dr2
    r |= 1 << p
```

soraros:
> ... If preferred, I can extract the duplicated expression into a temporary variable for clarity:
> 
> ```
> var dr2 = (r << (p + 1)) + (1 << (p + p))
> if r2 <= x - dr2:
>     r2 += dr2
>     r |= 1 << p
> ```

I think this looks better.

JoeLoser (commented):
!sync

soraros:
@JoeLoser I think it hasn’t been synced back by Copybara?

JoeLoser:
> @JoeLoser I think it hasn’t been synced back by Copybara?

Looks like an error in the bot responding as it did indeed never sync.  However, now that the math tests are open sourced (see https://github.com/modular/max/pull/4452), can you rebase with that and adjust your test, please? 🎉

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

soraros:
@JoeLoser This is not merged into `main` somehow.

PR #4355, opened 2025-04-16
[stdlib] Fix and modernise `my_complex.mojo`

soraros:


examples/mojo/operators/test_my_complex.mojo
```diff
@@ -31,6 +31,18 @@ def test_init():
     assert_equal(im2, c2.im)
 
 
+def test_bool():
```
soraros, line 34: This, along with the changes made in `test_equality`, can catch the original bug.

examples/mojo/operators/test_my_complex.mojo
```diff
@@ -68,18 +79,6 @@ def test_indexing():
...
-
-    with assert_raises(contains=err_msg):
-        _ = c1[2]
-
-    with assert_raises(contains=err_msg):
-        c1[-1] = 1.0
-
-    with assert_raises(contains=err_msg):
-        c1[2] = 1.0
-
```
soraros, line 82: No longer needed, we check the index at comptime now.

JoeLoser:
!sync

JoeLoser (approved):
Thanks for the various test fixes that it exposed, appreciate it!

soraros:
I think we really want to run the example tests on the Github CI though. Not sure how to do that.

JoeLoser:
> I think we really want to run the example tests on the Github CI though. Not sure how to do that.

I'll open up an issue as @Ahajha has opinions on how to do this internally too.

Ahajha:
Thanks! Yea this is something that I knew we had to do but has slipped my attention for quite a while.

arthurevans:
FYI @KenJones-Modular 

I'm not sure that all of these changes are desirable. This is the example code from https://docs.modular.com/mojo/manual/operators/#an-example-of-implementing-operators-for-a-custom-type

The tutorial goes through the process of adding operators to a type, and some of these changes lose that context.

For example, the first line changed:

```
print(StaticString("c1: Real: {}; Imaginary: {}").format(c1.re, c1.im))
```

Is demonstrating accessing the values before `__repr__()` has been implemented.

We probably need to make the README more explicit about the linkage between the example and the tutorial.

soraros:
@arthurevans I see. I'd argue that when printing `c.re` and `c.im`, we should instead write `print("c: Real:", c1.re, "; Imaginary:", c1.im)`, since string interpolation deserves its own chapter.

`__ne__` has a faulty implementation in the tutorial as well.

PR #4356, opened 2025-04-18
[stdlib] Avoid copy when fetching a value using Dict.get_ptr

bgreni:
Discord member reported an unexpected copy and move when using `Dict.get_ptr()`. 

https://discord.com/channels/1087530497313357884/1151418092052815884/1362569663422402630

I'm not sure about the source of the move, but I was able to eliminate the copy.

### DIFF
--- a/mojo/stdlib/test/collections/test_dict.mojo
+++ b/mojo/stdlib/test/collections/test_dict.mojo
@@ -15,7 +15,8 @@
 from collections import Dict, KeyElement, Optional
 from collections.dict import OwnedKwargsDict
 
-from test_utils import CopyCounter
+from test_utils import CopyCounter, AbortOnCopy
+from os import abort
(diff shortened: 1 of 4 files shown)


mojo/stdlib/test/collections/test_dict.mojo
```diff
@@ -619,6 +620,24 @@ def test_compile_time_dict():
         assert_equal(val, i)
 
 
+@value
+struct PlsDontCopyMe(Copyable):
```
soraros: Could you move this type to `test_utils/types.mojo`? ~~And maybe use a less creative name.~~

lattner (approved):
Nice, I'll let the stdlib experts review to verify, but this LGTM, thank you!

JoeLoser (approved):

JoeLoser:
!sync

PR #4363, opened 2025-04-20
[stdlib] Fix game of life examples

soraros:
They don't even compile, since `Tuple`s are not `PythonConvertible`.

Why didn't the CI catch this, though? CC @JoeLoser.

### DIFF
--- a/examples/mojo/life/lifev1.mojo
+++ b/examples/mojo/life/lifev1.mojo
@@ -32,7 +32,9 @@ def run_display(
     pygame.init()
 
     # Create a window and set its title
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
Thanks for the fix!  I'm assuming this isn't tested in our internal packaging CI either which is why it went unnoticed (CC: @Ahajha)

JoeLoser:
!sync

Ahajha:
Ah yea, we test some of these files but not these. We should at least build them in CI.

PR #4365, opened 2025-04-20
Fixing array example compilation issue

ratulb:


ratulb:
Another issue under the section "Automatic parameterization of functions" fixed.

```fn print_params(vec: SIMD[*_]):
    print(vec.type)
    print(vec.size)
```

Changed to:
```fn print_params(vec: SIMD[*_]):
    print(vec.dtype)  # from type to dtype
    print(vec.size)
```

PR #4367, opened 2025-04-20
[stdlib] Clean up `optional.mojo` using trait composition

soraros:


JoeLoser (approved):
LGTM! @zyx-billy will be excited seeing us put the new feature to good use so quickly.

JoeLoser:
!sync

soraros:
@JoeLoser Style-wise, should we continue using `CollectionElement` or `KeyElement`?

PR #4378, opened 2025-04-20
[stdlib] Clean up `int.mojo` using trait composition

soraros:


mojo/stdlib/src/collections/string/format.mojo
```diff
@@ -447,14 +447,12 @@ struct _FormatCurlyEntry(CollectionElement, CollectionElementNew):
...
 
-trait _CurlyEntryFormattable(Stringable, Representable):
-    """This trait is used by the `format()` method to support format specifiers.
-    Currently, it is a composition of both `Stringable` and `Representable`
-    traits i.e. a type to be formatted must implement both. In the future this
-    will be less constrained.
-    """
-
-    ...
+alias _CurlyEntryFormattable = Stringable & Representable
```
soraros, line 450: We don't delete this right away, so the migration can be done gradually.

JoeLoser: Can we mark it as `[[deprecated]]` or is there an old bug where deprecated doesn't work on aliases?

JoeLoser (approved):
LGTM, thanks for continuing to push on this cleanup!

JoeLoser:
!sync

mojo/stdlib/src/collections/string/format.mojo
soraros: I *think* it can be deprecated, but since it's internal and there are some other traits being added just a few lines down, I think it's better to leave this as is for now.

PR #4379, opened 2025-04-20
[stdlib] Make `FileDescriptor` ctor from `Int` explicit

soraros:


mojo/stdlib/src/sys/_io.mojo
```diff
@@ -11,5 +11,6 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 """IO constants and functions."""
-alias stdout: FileDescriptor = 1
-alias stderr: FileDescriptor = 2
+alias stdin = FileDescriptor(0)
```
soraros, line 14: Any reason why we don't have this in the past?

JoeLoser: Probably just never needed it

JoeLoser (approved):

JoeLoser:
!sync

PR #4381, opened 2025-04-21
[stdlib] Add `String(buffer=)` constructor again

martinvuyk:
This constructor avoids allocation when constructing from an owned `List` whose capacity is a multiple of 8.

### DIFF
--- a/mojo/stdlib/src/collections/string/string.mojo
+++ b/mojo/stdlib/src/collections/string/string.mojo
@@ -122,16 +122,14 @@ struct _StringCapacityField:
     # Initialize with a specified capacity.  Note that the provided value may
     # be rounded up, so clients should check the capacity() after construction.
     @always_inline("nodebug")
-    fn __init__(out self, capacity: UInt):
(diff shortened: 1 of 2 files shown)


martinvuyk:
@lattner `List.steal_ptr()` and it's `ptr` and `length` constructor aren't only used in stdlib or MAX code. This breaks some low-level side-projects I have. I also don't like doing mallocs unnecessarily. This removes the ability to implement something clever with pointers and passing it back to the constructor for an stdlib type for any external library author.

YkdWaWEzVmphR1Z1:
I would probably add a NO_SSO parameter for those who don't like generate branches and shifts and want to do unsafe stuff. I think List also should get this for String <> List conversion. Downside is it generates complexity

martinvuyk:
@lattner with this constructor we don't leak any `String` details more than the multiple of 8 requirement (i.e. `String(ptr=, length=)` is no longer necessary) WDYT?

martinvuyk:
we could also do a memcpy into an inline string when the list capacity multiple of 8 is small enough instead of forcing it to be an out-of-line string

PR #4382, opened 2025-04-21
[stdlib] Add back a safer filehandle read function

sstadick:
`FileHandle.read` was removed in 716cd45 for safety. This adds a safer mechanism for reading that has a known size, unlike the previous `read` method that could take a size of `-1` to attempt to read the whole file into a pointer.

If I'm missing the point with why the original was removed, feel free to close this!

To Reviewer:
I switch from `KGEN_CompilerRT_IO_FileReadToAddress` to `KGEN_CompilerRT_IO_FileReadBytes` to match the other read functions. However, I don't know for sure that is safe without seeing the docs for those. Tests pass but I wanted to point that out in case it's an issue.

### DIFF
--- a/mojo/stdlib/src/builtin/file.mojo
+++ b/mojo/stdlib/src/builtin/file.mojo
@@ -185,6 +185,12 @@ struct FileHandle(Writer):
     ](self, buffer: Span[Scalar[dtype], origin]) raises -> Int:
         """Read data from the file into the Span.
 
+        This will read n bytes from the file into the input Span where
+        `0 <= n <= len(buffer)`.
+
+        0 is returned when the file is at EOF, or a 0-sized buffer is
+        passed in.
+
         Parameters:
             dtype: The type that the data will be represented as.
             origin: The origin of the passed in Span.
--- a/mojo/stdlib/src/builtin/file.mojo
+++ b/mojo/stdlib/src/builtin/file.mojo
@@ -182,7 +182,7 @@ struct FileHandle(Writer):
 
     fn read[
         dtype: DType, origin: Origin[True]
-    ](self, buffer: Span[Scalar[dtype], origin]) raises -> Int64:
+    ](self, buffer: Span[Scalar[dtype], origin]) raises -> Int:
         """Read data from the file into the Span.
 
         Parameters:
@@ -229,7 +229,6 @@ struct FileHandle(Writer):
         var twelvth_element = ptr2[1]
         print(eleventh_element, twelvth_element)
         ```
-        .
         """
 
         if not self.handle:
@@ -237,9 +236,7 @@ struct FileHandle(Writer):
 
         var err_msg = _OwnedStringRef()
 
-        var bytes_read = external_call[
-            "KGEN_CompilerRT_IO_FileReadToAddress", Int64
-        ](
+        var bytes_read = external_call["KGEN_CompilerRT_IO_FileReadBytes", Int](
             self.handle,
             buffer.unsafe_ptr(),
             len(buffer) * sizeof[dtype](),
--- a/mojo/stdlib/src/builtin/file.mojo
+++ b/mojo/stdlib/src/builtin/file.mojo
@@ -180,6 +180,77 @@ struct FileHandle(Writer):
         var list = self.read_bytes(size)
         return String(bytes=list)
 
+    fn read[
+        dtype: DType, origin: Origin[True]
+    ](self, buffer: Span[Scalar[dtype], origin]) raises -> Int64:
+        """Read data from the file into the Span.
+
+        Parameters:
+            dtype: The type that the data will be represented as.
+            origin: The origin of the passed in Span.
+
+        Args:
+            buffer: The mutable Span to read data into.
+
+        Returns:
+            The total amount of data that was read in bytes.
+
+        Raises:
+            An error if this file handle is invalid, or if the file read
+            returned a failure.
+
+        Examples:
+
+        ```mojo
+        import os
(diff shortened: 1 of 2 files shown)


mojo/stdlib/src/builtin/file.mojo
```diff
@@ -180,6 +180,75 @@ struct FileHandle(Writer):
...
+        # Allocate and load 8 more elements from file handle seek position
+        var ptr2 = UnsafePointer[Float32].alloc(8)
+        var buffer2 = InlineArray[Float32, size=8](fill=0)
+        var bytes2 = file.read(buffer2)
+
+        var eleventh_element = ptr2[0]
+        var twelvth_element = ptr2[1]
+        print(eleventh_element, twelvth_element)
+        ```
+        .
```
JoeLoser:
```suggestion
```
This can be removed now - the tooling bug requiring this has now been fixed. 🎉

mojo/stdlib/src/builtin/file.mojo
```diff
@@ -180,6 +180,75 @@ struct FileHandle(Writer):
         var list = self.read_bytes(size)
         return String(bytes=list)
 
+    fn read[
+        dtype: DType, origin: Origin[True]
```
JoeLoser, line 184:
```suggestion
        dtype: DType, origin: Origin[True], //
```
Should both of these parameters be infer-only from the buffer argument?

mojo/stdlib/test/builtin/test_file.mojo
```diff
@@ -99,6 +99,42 @@ def test_file_read_context():
...
+        assert_equal(f.read(buffer), 954)
+        assert_equal(buffer[0], 76)  # L
+        assert_equal(buffer[1], 111)  # o
+        assert_equal(buffer[2], 114)  # r
+        assert_equal(buffer[3], 101)  # e
+        assert_equal(buffer[4], 109)  # m
+        assert_equal(buffer[5], 32)  # <space>
+        assert_equal(buffer[56], 10)  # <LF>
+
+    with open(
```
JoeLoser, line 120: **Question** What's the value in this test?  The first test already checks that `read()` returns the number of bytes read.

mojo/stdlib/test/builtin/test_file.mojo
```diff
@@ -99,6 +99,42 @@ def test_file_read_context():
...
+    ) as f:
+        var buffer_30 = InlineArray[UInt8, size=30](fill=0)
+        var buffer_1 = InlineArray[UInt8, size=1](fill=0)
+        var buffer_2 = InlineArray[UInt8, size=2](fill=0)
+        var buffer_100 = InlineArray[UInt8, size=100](fill=0)
+        assert_equal(f.read(buffer_30), 30)
+        assert_equal(f.read(buffer_1), 1)
+        assert_equal(f.read(buffer_2), 2)
+        assert_equal(f.read(buffer_100), 100)
+
```
JoeLoser, line 167: **Question** Is it worth adding a test where `read(buffer, size_more_than_input)`?  E.g. an input buffer of 1000 and `read(buffer, 1024)`, or what's the behavior for that currently?

mojo/stdlib/src/builtin/file.mojo
owenhilyard: I don't think they should be.

mojo/stdlib/test/builtin/test_file.mojo
sstadick: These tests were the tests that were here originally, just adjusted to use span. I agree that they are lacking. I was trying to limit the changes. I can totally make the tests more thorough as suggested in the next comment 👍

mojo/stdlib/test/builtin/test_file.mojo
sstadick: See above, copy paste of the original tests, but I can update them!

JoeLoser:
!sync

JoeLoser:
!sync

mojo/stdlib/test/builtin/test_file.mojo
owenhilyard: That test should probably exist since it's a memory safety test.

mojo/stdlib/test/builtin/test_file.mojo
sstadick: @JoeLoser - clarifying, is that a thumbsup of "yes let's flesh out the tests more in this PR" or "it's fine as is, separate PR for more tests"?

JoeLoser: Follow-up is totally fine, I'm +1 for getting this in as-is now for 25.3, but fast follow with more testing would be good. Thank you!

sstadick: Sounds good, will follow up with a PR. Thank you for looking at this!

sstadick:
WIP of updated tests: https://github.com/sstadick/max/pull/1 (diff is vs this PR)

sstadick:
@JoeLoser I've added to the tests, and better labeled what is there, as well as added a few details to the docs.

JoeLoser:
> @JoeLoser I've added to the tests, and better labeled what is there, as well as added a few details to the docs.

Thanks for doing this!  FYI this made it into the stable branch 25.3 that went out today, but not yet in the nightlies.  In any case, I'll resync this and get it in.  Thanks @sstadick!

JoeLoser:
!sync

sstadick:
I saw it in 25.3, thanks for getting it in that release!

PR #4385, opened 2025-04-21
[stdlib] Make the `dest` pointer type in `memcpy` mutable

soraros:


soraros:
@lattner Why doesn't it catch for instance the following use?

https://github.com/modular/max/blob/e606578390fb2ce84db772a0b3ae29e63d221f1d/mojo/stdlib/src/collections/string/string.mojo#L971-L974

JoeLoser:
@soraros is this still something you'd like to pursue?  If so, can you please share some context/motivation in the PR description, please?

soraros:
@JoeLoser Yes, I still want to pursue this. However, until #4386 is fixed, the compiler can’t warn us if we pass it an immutable pointer, so it’s blocked for the moment. I’ll close this for now and try again later.

PR #4388, opened 2025-04-21
[stdlib] Add `List.__init__[multiple_of: UInt](capacity=)` functionality for power of two multiples

martinvuyk:
This is a fast version for power of two multiples. In the future another version for other numbers could be added if necessary.

### DIFF
--- a/mojo/stdlib/src/collections/list.mojo
+++ b/mojo/stdlib/src/collections/list.mojo
@@ -124,18 +124,38 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
             copy.append(e[])
         return copy^
 
-    fn __init__(out self, *, capacity: Int):
+    fn __init__[*, multiple_of: UInt = 1](out self, *, capacity: UInt):
         """Constructs a list with the given capacity.
+
+        Parameters:
+            multiple_of: The power of two multiple to round the capacity up to.
 
         Args:
             capacity: The requested capacity of the list.
+
+        Notes:
+            Zero is a multiple of every number. If capacity is zero, then the
+            rounded number will be zero.
         """
+        constrained[
+            multiple_of.is_power_of_two(),
+            "multiple_of value must be a power of two",
+        ]()
+        var rounded: UInt
+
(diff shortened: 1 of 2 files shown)


leb-kuchen:
Why not just rely on oom or abort on capacity overflow.

martinvuyk:
> Why not just rely on oom or abort on capacity overflow.

What does that have to do with this?

bgreni:
I'm not sure I understand the point of this? Why not just use a power of two directly in the existing ctor?

martinvuyk:
> I'm not sure I understand the point of this? Why not just use a power of two directly in the existing ctor?

I personally don't want to make decisions for people, especially low-level devs. If they want purely power of 2 capacity then they can set `multiple_of=2`

However, I don't know much about the underlying allocator implementations, so if it makes more sense to force it :man_shrugging:. But I imagine someone reserving 2 buffers of capacity 3 and a buffer of capacity 2 from an arena of total capacity 8 and it doesn't work will find an issue with that decision being made for them.

bgreni:
> > I'm not sure I understand the point of this? Why not just use a power of two directly in the existing ctor?
> 
> 
> 
> I personally don't want to make decisions for people, especially low-level devs. If they want purely power of 2 capacity then they can set `multiple_of=2`
> 
> 
> 
> However, I don't know much about the underlying allocator implementations, so if it makes more sense to force it :man_shrugging:. But I imagine someone reserving 2 buffers of capacity 3 and a buffer of capacity 2 from an arena of total capacity 8 and it doesn't work will find an issue with that decision being made for them.

Sorry I meant more just leave it to the user to supply the correct capacity and leave the ctor as is

martinvuyk:
> Sorry I meant more just leave it to the user to supply the correct capacity and leave the ctor as is

I guess I could spin this out onto another PR. I saw the rounding trick being used in `String` and I actually think it is being used wrong but I haven't looked much into it

martinvuyk:
@bgreni you're right, I'm closing this in favor of  #4416

PR #4389, opened 2025-04-21
[docs] Fix example with Python list in manual

### LINKED ISSUE #4387: [Docs]Documented code sample does not work
### Where is the problem?

https://docs.modular.com/mojo/stdlib/python/python/Python

### What can we do better?

The example code given under the section `import_module` does not work:

from python import Python

# This is equivalent to Python's `import numpy as np`
var np = Python.import_module("numpy")
**a = np.array([1, 2, 3]) # Issue here**

To make it work - we need to change it as follows:

**a = np.array(Python.list(1, 2, 3))**

**mojo 25.3.0.dev2025042005 (1b8a38e8)**

### Anything else?

_No response_

ratulb:
**Issue: Can not convert `ListLiteral[Int, Int, Int]` to `PythonObject`**

Fix: Explicitly wrap values using Python.list(...) to ensure compatibility with Python list.

Fixes https://github.com/modular/max/issues/4387.

### DIFF
--- a/mojo/docs/manual/python/index.mdx
+++ b/mojo/docs/manual/python/index.mdx
@@ -63,7 +63,7 @@ def main():
     np = Python.import_module("numpy")
 
     # Now use numpy as if writing in Python
-    array = np.array([1, 2, 3])
+    array = np.array(Python.list(1, 2, 3))
     print(array)
 ```
 


JoeLoser (approved):
Thanks for the contribution!  This looks good to me and I'm happy to merge this once the commit is signed off (see the DCO CI failure).

JoeLoser:
@ratulb to fix the DCO problem, you can sign off on your commit and amend it using: `git commit -s --amend` and then push to `ratulb:python_list` branch again.  Thanks!

JoeLoser:
!sync

PR #4390, opened 2025-04-21
[stdlib] [NFC] Remove expanduser tests with empty `HOME`

martinvuyk:
This is a flaky test that I don't know how it didn't get caught before. In Ubuntu 22.04 this works fine (returns `/etc/passwd`), in 24.04 it doesn't (returns `/`). MacOS defaults to `User/runner` instead of the empty string. We shouldn't be concerned with and chase after default OS behavior in this test.


JoeLoser (approved):

JoeLoser:
!sync

PR #4391, opened 2025-04-22
[stdlib] Fix incorrect implicit ctor for `UnsafePointer`

soraros:


NathanSWard:
This has been fixed as of `25.7` with the new `UnsafePointer` API! :)

PR #4396, opened 2025-04-23
[stdlib] Chase after `UnsafePointer` origin bugs

### LINKED ISSUE #4386: [BUG] Implicit casting from non-mut `UnsafePointer` to mut ones.
### Bug description

As title. CC @lattner.

### Steps to reproduce

```mojo

fn main():
  var p = UnsafePointer[Int, mut=False]()
  var q: UnsafePointer[Int, mut=True]
  q = p
```

### System information

```
Package     Version               Build    Size       Kind   Source
max         25.3.0.dev2025042105  release  9.7 KiB    conda  max
max-core    25.3.0.dev2025042105  release  228.7 MiB  conda  max-core
max-python  25.3.0.dev2025042105  release  11.6 MiB   conda  max-python
```

martinvuyk:
Part of  #4391 and trying to fix #4386.

This is just a draft PR to find all the bugs, will progressively split off to fix different parts. It is all due to `UnsafePointer` default parameter values. But I think we can fix them all.

I'd appreciate any help

CC: @soraros @ConnorGray @JoeLoser @lsh


lattner:
Thank you so much for doing this Martin!

PR #4397, opened 2025-04-23
[stdlib] Fix immutable pointers implicitly cast to mutable in `string.mojo` and `stringslice.mojo`

martinvuyk:
Part of #4396

### DIFF
--- a/mojo/stdlib/src/collections/string/string.mojo
+++ b/mojo/stdlib/src/collections/string/string.mojo
@@ -347,8 +347,11 @@ struct String(
         Args:
             data: The static constant string to refer to.
         """
-        var length = len(data)
+        var length = data.byte_length()
         var cap_field = _StringCapacityField(static_const_length=length)
+        # NOTE: we can't set set_has_nul_terminator(True) because there is no
+        # guarantee that this wasn't constructed from a
+        # Span[Byte, StaticConstantOrigin] without a nul terminator.
         if cap_field.is_inline():
             # Tell mojo it is ok for this to be uninitialized.
             __mlir_op.`lit.ownership.mark_initialized`(
@@ -375,7 +378,8 @@ struct String(
             data: The static constant string to refer to.
         """
         self = StaticString(data)
-        # Indirect string literals are always nul-terminated by the compiler.
+        # All string literals are nul terminated by the compiler but the inline
+        # String overwrites it at any given point
         if not self._capacity_or_data.is_inline():
             self._capacity_or_data.set_has_nul_terminator(True)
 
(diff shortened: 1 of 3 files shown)


martinvuyk:
@lattner sorry to ping you, but do you have any idea why this might be crashing?

mojo/stdlib/src/collections/string/string.mojo
```diff
@@ -1575,13 +1575,17 @@ struct String(
             The pointer to the underlying memory.
         """
         # The string may be empty, add a nul terminator if so.
-        if not self._capacity_or_data.has_nul_terminator():
+        # Reallocate the string if it is static_constant to ensure mutability.
+        if (
+            not self._capacity_or_data.has_nul_terminator()
+            or self._capacity_or_data.is_static_constant()
```
soraros: I think this probably means there are a thousand and one corner cases waiting to be fixed. We definitely need a lot more tests for our new string.

martinvuyk: Well this was first crashing and then some testcases which had a static string were failing on the null terminator :man_shrugging:

soraros: I think the out-of-line static staring is always null terminated. Do we need a `is_inline` here?

martinvuyk: You made me do a deep dive, and yes all string literals are nul terminated. But the constructor from a `StringLiteral` wasn't setting it to true for inline.

But we still need to check whether it doesn't have a nul terminator or if it's out-of-line static constant because we need to ensure that the mutable pointer is always nul terminated (and mutable unlike staticstrings). And the `unsafe_ptr_mut()` reallocates out of line static strings without copying the nul terminator. So what I ended up doing is adding a check for `not is_inline()` besides checking for static string

soraros: > So what I ended up doing is adding a check for `not is_inline()` besides checking for static string

Sorry I wasn't been very clear. I meant exactly this.

martinvuyk: welp, turns out that inline static strings overwrite the nul terminator in some cases because it is assumed to be ignored. I'll just leave this part as-is

lsh:
!sync

mojo/stdlib/src/collections/string/string.mojo
```diff
@@ -811,7 +815,7 @@ struct String(
         var rhs_len = len(rhs)
 
         var result = String(unsafe_uninit_length=lhs_len + rhs_len)
-        var result_ptr = result.unsafe_ptr()
+        var result_ptr = result.unsafe_ptr_mut()
```
soraros, line 818: Maybe we should introduce a `_unsafe_ptr_mut` that only mut-cast the pointer, to skip the branching.

martinvuyk: We could just add a parameter `is_exclusive_string` in the future to `fn unsafe_ptr_mut(...)`

soraros: Since there's `_make_unique_mutable`, we could call it `is_unique`. And yes, I think we do need that. Even in this patch there are like three cases where we can use it.

martinvuyk: Great idea, I'll do a followup PR after this one lands

PR #4398, opened 2025-04-23
[stdlib] Fix origin mutability of pointers in `memory.mojo` and `qmatmul_k.mojo` and add `NDBuffer[mut=False]` implicit ctor

martinvuyk:
Part of  #4396.

This also adds an implicit constructor from `NDBuffer[mut=_]` -> `NDBuffer[mut=False]` to avoid having to fix ambiguous function signatures all up the call-stack

### DIFF
--- a/bazel/mojo.MODULE.bazel
+++ b/bazel/mojo.MODULE.bazel
@@ -1,6 +1,6 @@
 """Define the Mojo toolchain"""
 
-PACKAGE_VERSION = "25.6.0.dev2025091505"
+PACKAGE_VERSION = "25.6.0.dev2025091606"
 
 BASE_URL = "https://dl.modular.com/public/nightly/python"
 
--- a/examples/custom-graph-module/pixi.lock
+++ b/examples/custom-graph-module/pixi.lock
@@ -30,7 +30,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-checksums-0.2.7-h92c474e_2.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-crt-cpp-0.33.1-hb4fd278_2.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-sdk-cpp-1.11.606-h31ade35_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/azure-core-cpp-1.16.0-h3a458e0_0.conda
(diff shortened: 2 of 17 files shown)


mojo/stdlib/src/memory/memory.mojo
```diff
@@ -46,8 +46,8 @@ from memory.pointer import AddressSpace, _GPUAddressSpace
 fn _memcmp_impl_unconstrained[
     dtype: DType, //
 ](
-    s1: UnsafePointer[Scalar[dtype], **_],
-    s2: UnsafePointer[Scalar[dtype], **_],
+    s1: UnsafePointer[Scalar[dtype], mut=False, **_],
```
lsh: Let's put `mut=False` at the start of the parameter list. I'd like to maintain the flexibility to make `mut` an inferred parameter in the future.

lsh (changes requested):

mojo/stdlib/src/memory/memory.mojo
martinvuyk: I can edit them in a follow up PR. I've already followed this pattern in several PRs for #4396

lsh:
!sync

lsh:
This one has some internal conflicts so can't be merged yet

martinvuyk:
@lsh I tried only touching the absolute minimum necessary to make this work

JoeLoser:
Once the conflicts are resolved, @lsh can take a look again. Thanks, all!

lsh:
!sync

JoeLoser:
!sync

JoeLoser:
!sync

lsh:
!sync

lsh:
This seems to fail ASAN and cause a UAF

martinvuyk:
@lsh any specifics on what line/function this is happening? Or any tips on how I can track this down? I have no experience dealing with such issues.

lsh:
This line seems to be the cause of one issue:
https://github.com/modular/modular/blob/3f6ad8687ae7fb9b1b3c026d2ee6d5a2d5949d26/mojo/stdlib/stdlib/algorithm/reduction.mojo#L323

martinvuyk:
@lsh I changed the `src: NDBuffer[mut=False, *_, **_]` which is where the input is built from and also added the alignment parameter to the `NDBuffer._offset()` return. Could you test this again?

lsh:
!sync

PR #4401, opened 2025-04-23
[stdlib] Fix `UnsafePointer` origins in `format.mojo`

martinvuyk:
Part of #4396


mojo/stdlib/src/collections/string/format.mojo
```diff
@@ -301,11 +305,17 @@ struct _FormatCurlyEntry(CollectionElement, CollectionElementNew):
         mut raised_kwarg_field: Optional[String],
         mut total_estimated_entry_byte_width: Int,
     ) raises -> Bool:
-        alias S = StringSlice[StaticConstantOrigin]
+        alias S = StringSlice[ImmutableAnyOrigin]
```
lsh: Is this still necessary since StringSlice implements `__merge_with__`?

martinvuyk: You're right! that is a nice improvement

lsh:
!sync

PR #4402, opened 2025-04-23
[stdlib] Make `InlineString.unsafe_ptr()` carry the origin

martinvuyk:
Part of #4396


soraros:
I think inline string is deprecated and on its way out, so maybe we don't need this.

JoeLoser:
> I think inline string is deprecated and on its way out, so maybe we don't need this.

Correct, we're planning on removing it within the next few days (CC @ConnorGray).

PR #4412, opened 2025-04-23
[stdlib] Fix `List.unsafe_ptr()` related origin casts

martinvuyk:
Part of #4396

### DIFF
--- a/mojo/stdlib/src/collections/list.mojo
+++ b/mojo/stdlib/src/collections/list.mojo
@@ -1064,24 +1064,22 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
     fn unsafe_ptr(
         ref self,
     ) -> UnsafePointer[
-        T,
-        mut = Origin(__origin_of(self)).mut,
-        origin = __origin_of(self),
+        T, mut = Origin(__origin_of(self)).mut, origin = __origin_of(self)
     ]:
         """Retrieves a pointer to the underlying memory.
 
         Returns:
             The pointer to the underlying memory.
         """
-        return self.data
+        return self.data.origin_cast[
+            mut = Origin(__origin_of(self)).mut, origin = __origin_of(self)
+        ]()
 
     @always_inline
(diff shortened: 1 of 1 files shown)


lsh:
!sync

lsh:
!sync

mojo/stdlib/src/collections/list.mojo
```diff
@@ -1107,7 +1109,9 @@ struct List[T: CollectionElement, hint_trivial_type: Bool = False](
             ),
         )
 
-        return self.data + self._len
+        return (self.data + self._len).origin_cast[
+            mut = Origin(__origin_of(self)).mut, origin = __origin_of(self)
+        ]()
```
soraros, line 1110: Can we simply use `self.unsafe_ptr() + self._len` here?

martinvuyk: I guess you're right

martinvuyk: I got this error so I just reverted:
```terminal
Included from /Users/runner/work/max/max/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/runner/work/max/max/mojo/stdlib/src/collections/list.mojo:1108:34: error: argument of '__add__' call allows writing a memory location previously writable through another aliased argument
        return self.unsafe_ptr() + self._len
               ~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~
Included from /Users/runner/work/max/max/mojo/stdlib/src/collections/__init__.mojo:1:
/Users/runner/work/max/max/mojo/stdlib/src/collections/list.mojo:1108:34: note: 'self_is_origin' memory accessed through reference embedded in value of type 'Int'
        return self.unsafe_ptr() + self._len
                                 ^
```
But I think it should've worked

soraros: Should probably file an issue.

martinvuyk: opened #4454

PR #4414, opened 2025-04-23
[stdlib] Let write methods receive `StringSlice`s

martinvuyk:
This removes the unsafe building of a `StaticString` from a `String` and modifies several writer methods to receive `StringSlice`s with a default of `StaticString()` by making use of a default origin of `origin: ImmutableOrigin = StaticConstantOrigin` and rebinding it to the `StringSlice[origin]`

Closes MOCO-1791

### DIFF
--- a/mojo/stdlib/std/builtin/variadics.mojo
+++ b/mojo/stdlib/std/builtin/variadics.mojo
@@ -1168,17 +1168,13 @@ struct VariadicPack[
         return __mlir_op.`kgen.pack.load`(self.get_as_kgen_pack())
 
     def _write_to[
-        O1: ImmutOrigin,
-        O2: ImmutOrigin,
-        O3: ImmutOrigin,
-        *,
-        is_repr: Bool = False,
+        is_repr: Bool = False
     ](
         self: VariadicPack[_, Writable, ...],
         mut writer: Some[Writer],
-        start: StringSlice[O1] = StaticString(""),
-        end: StringSlice[O2] = StaticString(""),
-        sep: StringSlice[O3] = StaticString(", "),
+        start: StringSlice = "",
+        end: StringSlice = "",
+        sep: StringSlice = ", ",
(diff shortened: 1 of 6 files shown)


laszlokindrat:
@ConnorGray can you PTAL?

martinvuyk:
ping @ConnorGray

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -284,7 +284,14 @@ struct String(
 
     fn __init__[
         *Ts: Writable,
-    ](out self, *args: *Ts, sep: StaticString = "", end: StaticString = ""):
+        O1: ImmutableOrigin = StaticConstantOrigin,
+        O2: ImmutableOrigin = StaticConstantOrigin,
+    ](
+        out self,
+        *args: *Ts,
+        sep: StringSlice[O1] = rebind[StringSlice[O1]](StaticString()),
```
JoeLoser: This LGTM, but can we add a simple unit test that passes `sep` and `end` explicitly as `StringSlice`s to ensure there are no issues?  I understand the previous uses/tests with `StaticString` will work as before, but now that we've loosened the requirements in a sense, I want to ensure we have tests with the lowest requirement as well.

martinvuyk: ok makes sense

martinvuyk: done :)

martinvuyk:
ping @JoeLoser @ConnorGray

martinvuyk:
@NathanSWard I think you'll find these tricks interesting :)

lattner:
FYI, I just landed an improvement to default arguments that will make this far easier.  You can now do things like `arg: StringSlice[origin] = ""`, this will come out in the next nightly

PR #4416, opened 2025-04-23
[stdlib] Add `math.next_multiple_of()` and fix `String` impl

martinvuyk:
Spin off from #4388.

This is useful to centralize the implementation and avoid bugs.

It is a compile time version of `align_up` where the multiple is a parameter.

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/string.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string.mojo
@@ -90,6 +90,7 @@ from sys.info import is_32bit
 from sys.ffi import c_char
 
 from bit import count_leading_zeros
+from math import next_multiple_of
 from memory import memcpy, memset, memcmp
 from python import PythonObject, ConvertibleFromPython, ConvertibleToPython
 
@@ -204,8 +205,9 @@ struct String(
                 __get_mvalue_as_litref(self)
             )
         else:
-            self._capacity_or_data = (capacity + 7) >> 3
-            self._ptr_or_data = Self._alloc(self._capacity_or_data << 3)
+            var mul = next_multiple_of[8](capacity)
+            self._capacity_or_data = mul >> 3
+            self._ptr_or_data = Self._alloc(mul)
             self._len_or_data = 0
             self._set_ref_counted()
 
--- a/mojo/stdlib/stdlib/math/__init__.mojo
+++ b/mojo/stdlib/stdlib/math/__init__.mojo
@@ -68,6 +68,7 @@ from .math import (
     modf,
     recip,
     remainder,
+    next_multiple_of,
     scalb,
     sin,
     sinh,
--- a/mojo/stdlib/stdlib/math/math.mojo
+++ b/mojo/stdlib/stdlib/math/math.mojo
@@ -1330,8 +1330,39 @@ fn align_down(value: UInt, alignment: UInt) -> UInt:
 
 
 # ===----------------------------------------------------------------------=== #
-# align_up
+# next_multiple_of
 # ===----------------------------------------------------------------------=== #
+
+
+@always_inline
+fn next_multiple_of[multiple_of: UInt](value: UInt) -> UInt:
+    """Returns the closest multiple of `multiple_of` that is greater than or
+    equal to value.
+
+    Parameters:
+        multiple_of: Value to round up to.
+
+    Args:
+        value: The value to round up.
+
+    Returns:
+        Closest multiple of `multiple_of` that is greater than or equal to the
+        input value.
+
+    Notes:
+        Zero is a multiple of every number. If value is zero, then the
+        rounded number will be zero.
+    """
+    constrained[multiple_of != 0, "multiple_of must be > 0"]()
+
+    @parameter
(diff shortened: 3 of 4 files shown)


mojo/stdlib/src/collections/string/string.mojo
```diff
@@ -599,7 +600,9 @@ struct _StringCapacityField:
         # use it. We store the capacity with the top 3 bits of the storage used
         # for flags.
         if capacity > Self.NUM_SSO_BYTES or is_compile_time():
-            self._storage = (capacity + 7) >> 3
+            self._storage = (
+                next_multiple_of[8, debug_assert_no_zero=False](capacity) >> 3
+            )
```
martinvuyk: In a way, the original code can never be wrong because the lower 3 bits get swallowed up by the bit-shift by 3 because of the information to be stored there. But if we ever reduce the amount of set bits, this is a sneaky bug waiting to happen IMO and I'd rather this use an extra micro-op in the CPU than developer time.

mojo/stdlib/src/math/math.mojo
```diff
@@ -1300,10 +1300,48 @@ fn align_down(value: UInt, alignment: UInt) -> UInt:
...
+
+    Notes:
+        Zero is a multiple of every number. If value is zero, then the
+        rounded number will be zero.
+    """
+    constrained[multiple_of != 0, "multiple_of must be > 0"]()
+
+    @parameter
+    if debug_assert_no_zero:
+        debug_assert(value != 0, "zero value in nex_multiple_of")
```
laszlokindrat: Q: Why not do this unconditionally?

laszlokindrat:
Thanks for the patch! I could be persuaded about trading off a few cycles for improved readability and robustness, but I'm not sure this helper deserves to be in `math`. Do you know of any other use cases? If not, how would you feel about making it hidden, or just keeping it local to `string.mojo`?

CC: @JoeLoser in case you have opinions.

mojo/stdlib/src/math/math.mojo
martinvuyk: Because there are use-cases where one wouldn't want zero to cause an error, e.g. in an algorithm which reserves size for a certain datastructure:
this is the code in string.mojo
```mojo
    fn __init__(out self, *, out_of_line_capacity: UInt):
        var mul = next_multiple_of[8, debug_assert_no_zero=False](
            out_of_line_capacity
        )
        self._storage = mul >> 3
```
Allocations by 0 should be discouraged, but not technically wrong.

There are still other use-cases where a user of the function might not want it failing for a zero value. It's just a way to not have to make that decision for someone.

martinvuyk:
I would like this to eventually replace `align_up` in every datastructure and algo that needs to be aligned. It is faster and IMO more readable at a glance. We would also be centralizing the impl using the bitwise trick and avoiding bugs in the future.

I also think looking for a function to search for the next multiple of a value is something that everyone searches for when implementing something that requires it. I don't much care whether it's in the bit or math package, I just saw `align_up` there so I figured I'd put it above it.

mojo/stdlib/src/math/math.mojo
laszlokindrat: I find this pattern of triggering a `debug_assert` (which itself is useless outside debug builds) via a flag a bit strange. I would even say that this is another smell that this function is not providing the right abstraction. Since `value == 0` does not cause the function to behave badly, assert doesn't seem like the right thing to do, so I'd much rather just put the responsibility on the caller to know what they are passing.

martinvuyk: Yeah I think you're right, I'll drop the assert

mojo/stdlib/src/math/math.mojo
```diff
@@ -1310,10 +1310,41 @@ fn align_down(value: UInt, alignment: UInt) -> UInt:
 
 
 # ===----------------------------------------------------------------------=== #
-# align_up
+# next_multiple_of
 # ===----------------------------------------------------------------------=== #
 
 
+@always_inline
+fn next_multiple_of[multiple_of: UInt](value: UInt) -> UInt:
```
JoeLoser: @abduld thoughts on this being a public method, especially in the `math` module?  I'd lean towards private for now since it's only used in one place in the `String` constructor.

martinvuyk: My idea was to take this function to replace `align_up` where it's used for statically known values. And to add another method to do the same for `align_down`. Mostly because they are mainly used for power of two values (alignment, vectorization based on simd width, etc.), and we can shave off a little overhead on every computation :man_shrugging:.

In a sense it is just compile time optimized `align_up`, but I also want to rename it to follow the same direction as we did for `next_power_of_two`, so that's why I went for `next_multiple_of`. Maybe I should update the PR description with this thought process if you see fit

laszlokindrat:
@martinvuyk Are you still interested in landing this? If so, could you please rebase on the latest nightly? After that, I request that we make this a hidden function for now, and we can evaluate making it public if it gets sufficient use.

martinvuyk:
superseded by #5242

PR #4430, opened 2025-04-26
[stdlib] Make `Span[Byte]` to `String` ctor keyword only

soraros:
So it doesn't get confused with the generic ctor which means `v.__str__()`


JoeLoser (approved):

JoeLoser:
!sync

PR #4432, opened 2025-04-26
[max-examples] implement custom ops on CPU in image_pipeline 

ljoukov:
Attempt to implement image_pipeline using foreach[] thus allowing both CPU and GPU support and avoiding explicit GPU code

### DIFF
--- a/examples/custom_ops/image_pipeline.py
+++ b/examples/custom_ops/image_pipeline.py
@@ -27,22 +27,17 @@ from max.graph import (
 
 
 def main():
-    if accelerator_count() == 0:
-        raise RuntimeError(
-            "The image_pipeline example currently only runs on GPU"
-        )
-
-    device = Accelerator()
+    device = CPU() if accelerator_count() == 0 else Accelerator()
 
     img = imread("dogs.jpg")
 
-    color_tensor = TensorType(
+    color_tensor_type = TensorType(
         DType.uint8,
         shape=img.shape,
         device=DeviceRef.from_device(device),
     )
 
-    gray_tensor = TensorType(
+    gray_tensor_type = TensorType(
         DType.uint8,
         shape=[img.shape[0], img.shape[1]],
         device=DeviceRef.from_device(device),
@@ -50,7 +45,7 @@ def main():
 
     graph = Graph(
         "image_pipeline",
-        input_types=[color_tensor],
+        input_types=[color_tensor_type],
         custom_extensions=[Path(__file__).parent / "kernels"],
     )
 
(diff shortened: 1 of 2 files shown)


ljoukov:
@jackos this is an attempt following your demo yesterday https://youtu.be/uul6hZ5NXC8?t=4227

This is my first attempt at Mojo, a lot of things are unclear, so apologize if there is something completely off.

lattner:
Awesome @ljoukov !

jackos:
Fantastic thanks @ljoukov! Really like your improvements to the variable/parameter naming and extra comments as well. 

I've sent two codes to the email associated with your github commit for Mojo swag.

If you have any feedback about your experience using Mojo for the first time to implement GPU/CPU functionality, you can reply to my email, which would be much appreciated thanks.

jackos:
!sync

jackos (approved):

jackos:
FYI @ljoukov I made a small fix when merging to our internal repo on the box blur, as the loop on `simd_width` was making it average the pixels in sliding box kind of way and making the result look jagged. So I set `simd_width` to 1 and removed the outer loop.

It could be optimized with SIMD to do multiple pixels at a time with less threads, but it would need to be significantly more complicated. This is what it'll look like when it's merged back to the public repo:

```mojo
@register("blur")
struct Blur:
    @staticmethod
    fn execute[
        target: StaticString,
    ](
        img_out: OutputTensor[type = DType.uint8, rank=2],
        img_in: InputTensor[type = DType.uint8, rank=2],
        blur_size: Int64,
        ctx: DeviceContextPtr,
    ) raises:
        @parameter
        @always_inline
        fn blur_kernel[
            simd_width: Int
        ](idx: IndexList[img_out.rank]) -> SIMD[DType.uint8, simd_width]:
            """
            Computes the blurred value for a SIMD vector of pixels.
            """
            var height = img_in.shape()[0]
            var width = img_in.shape()[1]
            var size = Int(blur_size) # Surrounding pixels to average

            var base_row: Int = idx[0]
            var base_col: Int = idx[1]

            var pix_val_accum = 0
            var pixel_count = 0

            # Iterate over the blur kernel window
            for blur_row_offset in range(-size, size + 1):
                for blur_col_offset in range(-size, size + 1):
                    var cur_row = base_row + blur_row_offset
                    var cur_col = base_col + blur_col_offset

                    # Check if the neighbor pixel is within bounds
                    if 0 <= cur_row < height and 0 <= cur_col < width:
                        pix_val_accum += Int(img_in[cur_row, cur_col])
                        pixel_count += 1

            return (pix_val_accum / pixel_count).clamp(0, 255).cast[DType.uint8]()

        # Apply the kernel to each pixel in the output tensor
        foreach[blur_kernel, target=target, simd_width=1](img_out, ctx)
 ```
 Before:
image


 
 After: 
image

 
 Thanks again for the great PR.

ljoukov:
@jackos awesome catch re simd_width=1 

Should the same be done for greyscale?
foreach[color_to_grayscale, target=target](img_out, ctx)

https://github.com/modular/max/pull/4456

PR #4436, opened 2025-04-26
[stdlib] Fix example using `ListLiteral` in `python.mojo`

soraros:
This has slightly higher priority than the other example fixes, as it could lead to new issues asking why things don't work.


JoeLoser (approved):

JoeLoser:
!sync

PR #4449, opened 2025-04-28
[CI] Don't include the packaged stdlib in the import path

Ahajha:
With the `benchmark` package open-sourced, we no longer need to include the packaged stdlib when we run tests.

### DIFF
--- a/.github/workflows/standard_library_tests_and_examples.yml
+++ b/.github/workflows/standard_library_tests_and_examples.yml
@@ -66,3 +66,4 @@ jobs:
         run: |
           magic run --frozen tests
           magic run --frozen examples
+          magic run --frozen benchmarks
--- a/examples/mojo/lit.cfg.py
+++ b/examples/mojo/lit.cfg.py
@@ -46,8 +46,6 @@ config.test_source_root = Path(__file__).parent.resolve()
 # Substitute %mojo for just `mojo` itself.
 config.substitutions.insert(0, ("%mojo", "mojo"))
 
-os.environ["MODULAR_MOJO_MAX_IMPORT_PATH"] = str(build_root)
-
 # Pass through several environment variables
 # to the underlying subprocesses that run the tests.
 lit.llvm.initialize(lit_config, config)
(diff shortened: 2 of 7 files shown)


JoeLoser (approved):
So happy to remove this grossness now finally! 🎉 

Thanks for doing this

JoeLoser:
BTW, after this, we can even go one step further and run the benchmarks (at least to ensure they compile only since they use `-t` in each benchmark) in the OSS CI.

gabrieldemarmiesse:
After all the open sourcing of the next weeks, will that mean that if the OSS CI is passing for a given PR, then no adjustment will be needed on the internal codebase by the stdlib team? In other words, will all the mojo code of Modular be public at this point?

Ahajha:
!sync

JoeLoser:
> After all the open sourcing of the next weeks, will that mean that if the OSS CI is passing for a given PR, then no adjustment will be needed on the internal codebase by the stdlib team? In other words, will all the mojo code of Modular be public at this point?

For the most part, yes — things are improving. There's an ongoing effort to open source the kernels, which is happening in a few phases:

1. Source availability (rolling out now and over the next few days)
2. Contribution support (coming next)

However, between phases (1) and (2), there won't yet be a way to build the kernels externally, so there won't be open-source CI at the start. There’s also a gap in the testing matrix between our internal GPU testing and what will eventually be available for open-source GPU testing.

TL;DR: It's a significant step forward, but not fully complete yet. Does that make sense? We'll have more details to share soon about the kernel library and related updates.

gabrieldemarmiesse:
Thanks that's already very helpful! At least we'll have way more visibility about what code needs to change even without external CI running for the kernels

JoeLoser:
> Thanks that's already very helpful! At least we'll have way more visibility about what code needs to change even without external CI running for the kernels

Exactly.  Once we keep opening more code so you can even build them locally (even without OSS CI GPUs for kernels), that's a big step forward.  Or even _compiling_ the kernels but not being able to run them (e.g. if you don't have a GPU) can help when making stdlib changes, etc.

PR #4456, opened 2025-04-29
[max-examples] fix image_pipeline grayscale filter: set simd_width=1

ljoukov:
Followup on https://github.com/modular/max/pull/4432#issuecomment-2832943018 by @jackos 

It turns out the grayscale filter has the same issue as blur filter had.

Grayscale *only* output before the change:
dogs_out_pre

Grayscale only output after the simd_width=1:
dogs_out

Overall I'd like to confirm that setting simd_width=1 is what should be done here.
Naively I would think grayscale is point-wise filter and should naturally support SIMD.

### DIFF
--- a/examples/custom_ops/kernels/image_pipeline.mojo
+++ b/examples/custom_ops/kernels/image_pipeline.mojo
@@ -53,7 +53,7 @@ struct Grayscale:
 
             return gray_f32.clamp(0, 255).cast[DType.uint8]()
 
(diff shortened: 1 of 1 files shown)


jackos (approved):
Thanks, the problem is that the data is laid out like: [r, g, b, r, g, b, ...]

Because `simd_width` is not a multiple of 3 you get those artifacts

jackos:
!sync

PR #4460, opened 2025-04-29
[stdlib] Add `String.unsafe_ptr_mut[is_unique_mut_ref]()` parameter and rename internal fns

martinvuyk:
Add `String.unsafe_ptr_mut[is_unique_mut_ref]()` parameter and rename internal fns.

The parameter determines whether the variable is a unique mutable reference to the `String`, avoiding a potential realloc branch. This is useful because this function is called in hot paths where the memory was already previously reserved.

CC: @soraros

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/string.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string.mojo
@@ -391,7 +391,11 @@ struct String(
         """
         var length = len(bytes)
         self = Self(unsafe_uninit_length=length)
-        memcpy(self.unsafe_ptr_mut(), bytes.unsafe_ptr(), length)
+        memcpy(
+            self.unsafe_ptr_mut[is_unique_mut_ref=True](),
+            bytes.unsafe_ptr(),
+            length,
+        )
 
     @no_inline
     fn __init__[T: Stringable](out self, value: T):
@@ -840,7 +844,7 @@ struct String(
         var rhs_len = len(rhs)
 
         var result = String(unsafe_uninit_length=lhs_len + rhs_len)
-        var result_ptr = result.unsafe_ptr_mut()
+        var result_ptr = result.unsafe_ptr_mut[is_unique_mut_ref=True]()
         memcpy(result_ptr, lhs.unsafe_ptr(), lhs_len)
         memcpy(result_ptr + lhs_len, rhs.unsafe_ptr(), rhs_len)
         return result^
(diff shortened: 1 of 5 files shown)


soraros:
CC @lattner.

JoeLoser:
Do you mind rebasing to fix the conflicts, please?

lsh:
Could you fix the conflicts on this, please?

martinvuyk:
@lsh done :)

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1180,18 +1180,26 @@ struct String(
         else:
             return self._ptr_or_data
 
-    fn unsafe_ptr_mut(
-        mut self,
-    ) -> UnsafePointer[Byte, mut=True, origin = __origin_of(self)]:
+    fn unsafe_ptr_mut[
+        *, is_owned: Bool = False
```
lsh: After a little thinking, I'm not sure I like this name. Ownership is represented by origins, this `is_owned` feels like it's trying to remove the usage of origins from the pointer.

lsh: If this was a private method I would feel differently since we could always nitpick this kind of stuff

martinvuyk: I'm not sure ownership is represented by origins, because one can have a mutable reference without owning a buffer. Since String is CoW, this name makes more sense to me. We could rename it to `is_uniquely_owned`, it makes no difference to me beyond verbosity :man_shrugging:

lattner: I don't really understand the goal here (nor the original api that I may be at fault for).  Could you add more motivation to the commit message (the "why" not just the "what")?

Without understanding more detail, it seems like `unsafe_ptr` should return a non-mutable unsafepointer and never force uniqueness.  It seems like `unsafe_ptr_mut` should return a mutable pointer, and therefore /always/ force uniqueness.  You don't want a mutable pointer into something shared by other strings!

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1800,28 +1808,31 @@ struct String(
...
             not self._is_indirect_static_constant()
             and _StringOutOfLineHeader.get(self._ptr_or_data).is_unique()
-        ):
-            return
+        )
 
-        # Otherwise, copy to a new buffer to ensure mutability.
-        self._realloc_mutable(self.byte_length())
+    @always_inline
+    fn _make_owned(mut self):
```
lattner: Per other comments, I don't think that "owned" (which is a keyword btw) is the right word for this.  "unique" was meant to indicate this concept and perhaps isn't ideal, but I'm not sure if "owned" is better.

mojo/stdlib/stdlib/collections/string/string.mojo
martinvuyk: I added a description to the PR now and changed the name of the parameter

mojo/stdlib/stdlib/collections/string/string.mojo
martinvuyk: I renamed them to `is_unique_mut_ref` and `_make_unique_mut_ref`. To me this means the same as `var` and `owned` really, which is saying copy the value to a place where the reference holder is the only one that owns the data.

The only thing I was thinking is that there is really no difference when  a function uses `fn some_func(owned arg: String)` now because of `String` being CoW. So in a way changing the name to `_make_owned()`  to me was implementing the same logic of getting a local copy that owns the data.

lsh:
!sync

lsh:
!sync

PR #4479, opened 2025-05-02
[stdlib] Ensure the type parameters match the return type in `SIMD` conditional constructors

soraros:
Unfortunately, we're forced to choose between allowing `Int32(True)` to have the type `Scalar[.bool]` or having it fail entirely.


JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
!sync

PR #4481, opened 2025-05-03
[stdlib] GPU Info Script, 4090 and 4090m info

owenhilyard:
Adds a script to get info for Nvidia GPUs (it will try AMD too, but that crashes right now). 

I'd appreciate someone more familiar with what the fields in `Info` are supposed to be looking over this, since I tried to match them but I'm not totally sure. The AMD half of this will return the wrong arch name, but that's tolerable for now since the goal is to get most of the info out of the GPU. I'd also ideally like to remove all of the things I hardcoded.

### DIFF
--- a/mojo/stdlib/scripts/gpu_info.mojo
+++ b/mojo/stdlib/scripts/gpu_info.mojo
@@ -1,0 +1,266 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


owenhilyard:
cc: @JoeLoser @BradLarson

JoeLoser:
!sync

JoeLoser:
Thanks for working on this @owenhilyard! I'm going to sync this in so others can take a look and play with it as well to give constructive review feedback and/or fix the hard coded parts. This will be super useful as we introduce support for new GPU architectures.

owenhilyard:
@JoeLoser I would appreciate documentation on how exactly to get MAX to select device info via the name, since right now that's the only way to get at this info struct. I'm sure there's something WIP internally for handling B100/B200 (same compute capability), so hopefully that can be generalized. 

One thing I considered was using the actual result from `ctx.name()` as a "first priority" option for exact matches where we know exactly what to treat a given GPU as, but I have a feeling that touching the `DeviceContext` init logic is a bit of minefield as far as messing up internal code. 

I'd be happy to see the hardcoded parts fixed. I'd also like to request that `Info` get documented a bit better (especially in terms of "What does this mean on AMD?") and the mlir target info get some docs if we're going to try to use this to kick off a big batch of users adding GPU info/targets.

PR #4489, opened 2025-05-05
[stdlib] Reorder `dtype`s in `DType` methods

soraros:
We do this to prevent #4483 from happening again.

- Follows the order: `bool`, `index`, `uint8`, `int8`, ..., `uint256`, `int256`, `float8_xx`, ..., `float64`, `invalid`. The `f8` types are ordered alphabetically.


abduld:
!sync

Issue #4503, opened 2025-05-06
[BUG] `pop.bitcast` doesn't support `!pop.scalar<bool>` to `!pop.scalar<bool>` cast


soraros:
### Bug description

As title.

### Steps to reproduce

```mojo
fn main():
  alias type = DType.bool
  val = Scalar[type](False)
  res = __mlir_op.`pop.bitcast`[_type = Scalar[DType.bool]._mlir_type](val.value)
  # error: MLIR verification error: 
  #   'pop.bitcast' op operand type
  #     '!pop.scalar<bool>'
  #   and result type
  #     '!pop.scalar<bool>' 
  #   are cast incompatible
```

### System information

`mojo 25.4.0.dev2025050605 (b840f403)`

PR #4507, opened 2025-05-07
[stdlib] improve pack_bits constraint error message

sstadick:
`pack_bits` can't convert a SIMD[DType.bool, 1] vector into a valid integral type. This adds an explicity error message when this function is called with a scalar.

Create the error:
```python
    alias c = SIMD[DType.bool, 1](1)
    assert_equal(pack_bits(c), 1)
```

New error message:

```
note: constraint failed: the width of the bool vector must be the same as the bitwidth of the target type. Scalar bool (width=1) is not supported.
```

Related:
- https://github.com/modular/modular/issues/4503
- https://forum.modular.com/t/pack-bits-inside-of-vectorize-fails-constraint-check/1384/3

cc @soraros


abduld:
!sync

PR #4508, opened 2025-05-07
[stdlib] Fix the warp prefix sum algorithm on gpu

kirillbobyrev:
The warp prefix sum (scan) algorithm was [acting up on the GPU](https://forum.modular.com/t/prefix-sum-incorrect-results-with-gpu-warp-prefix-sum-and-gpu-block-prefix-sum/1387).

It turns out that the network was starting the shuffles with offset 2, which made the algorithm skip first round of `shuffle_up` (warp lane `i` -> warp lane `i + 1`) and caused incorrect results.


abduld:
!sync

kirillbobyrev:
(oops, accidentally deleted the branch; reopened the PR)

lattner:
nice work @kirillbobyrev you're on 🔥

PR #4510, opened 2025-05-07
[Magic] Clean up lockfiles from old folders

patrickdoc:
These lockfiles are sticking around even though the associated tomls are gone. Remove them to clean up.


lriggs1311 (approved):

Ahajha:
Hmm, curious why this didn't get caught.

PR #4511, opened 2025-05-07
[stdlib] Small improvements to `BitSet`

soraros:
CC @abduld.

* Replaced `InlineArray.__getitem__` with the unchecked `unsafe_get` to eliminate bounds checks. This avoids significant IR bloat—up to 20× for cases like `BitSet(mask)._words[0]`.
* Removed an unnecessary `String` allocation to skip a `malloc` and reduce memory overhead.

### DIFF
--- a/mojo/stdlib/stdlib/collections/bitset.mojo
+++ b/mojo/stdlib/stdlib/collections/bitset.mojo
@@ -69,14 +69,12 @@ fn _check_index_bounds[operation_name: StaticString](idx: UInt, max_size: Int):
     """
     debug_assert(
         idx < max_size,
-        String(
-            "BitSet index out of bounds when ",
-            operation_name,
-            " bit: ",
-            idx,
-            " >= ",
-            max_size,
-        ),
+        "BitSet index out of bounds when ",
+        operation_name,
+        " bit: ",
+        idx,
+        " >= ",
+        max_size,
     )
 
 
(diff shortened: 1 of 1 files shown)


mojo/stdlib/src/collections/bitset.mojo
```diff
@@ -69,14 +69,12 @@ fn _check_index_bounds[operation_name: StaticString](idx: UInt, max_size: Int):
...
     debug_assert(
         idx < max_size,
-        String(
-            "BitSet index out of bounds when ",
-            operation_name,
-            " bit: ",
-            idx,
-            " >= ",
-            max_size,
-        ),
```
soraros: This will allocate, making it *not* no-op on release build.

abduld:
Thanks @soraros , do you mind adding a descriptive PR message as to what this PR doing (I know but others might not).

gryznar:
Maybe reasonable next step would be to add benchmarks to have some baseline for future changes?

soraros:
@abduld Please let me know if you want the description to be more detailed.

abduld:
The description is now great. Thanks

abduld:
!sync

abduld:
> Maybe reasonable next step would be to add benchmarks to have some baseline for future changes?

A benchmark would be great as well

sstadick:
I'd be happy to add some benchmarks in a follow up PR. I did some [here](https://forum.modular.com/t/pack-bits-inside-of-vectorize-fails-constraint-check/1384/11?u=duck_tape) but they won't translate to the stdlib.

JoeLoser:
!sync

JoeLoser (approved):
LGTM, thanks!

PR #4519, opened 2025-05-08
[stdlib] Clean up `get_accum_type` and improve its docstring

soraros:
I made this patch because the semantics of this function is a bit strange.

### DIFF
--- a/mojo/stdlib/stdlib/utils/numerics.mojo
+++ b/mojo/stdlib/stdlib/utils/numerics.mojo
@@ -959,13 +959,16 @@ fn get_accum_type[
         preferred_accum_type: The preferred dtype for accumulation.
 
     Returns:
-        DType.float32 if dtype is a half-precision float, dtype otherwise.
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/utils/numerics.mojo
```diff
@@ -959,13 +959,16 @@ fn get_accum_type[
...
 
     @parameter
     if dtype.is_float8():
+
+        @parameter
         if preferred_accum_type is DType.float32:
-            return preferred_accum_type
+            return DType.float32
         else:
             return DType.bfloat16
```
soraros, line 973: Should we assert in this branch that `preferred_accum_type` isn’t wider than `bfloat16`? It feels odd if `preferred_accum_type` is `float64` but we don’t enable high-accuracy mode.

JoeLoser: CC: @abduld

JoeLoser:
!sync

PR #4523, opened 2025-05-08
[stdlib] Remove redundant `UnsafePointer.store` overloads

soraros:
This also fixes an issue where overload resolution fails to find a candidate for `p.store[width=width](idx, idx)` when `idx: Int`.


JoeLoser (commented):
!sync

JoeLoser:
!sync

PR #4525, opened 2025-05-08
[stdlib] Keep a global stdin handle to fix #3961.

### LINKED ISSUE #3961: [BUG] the second input() segfaults when using standard input redirection
### Bug description

when using standard input redirection, the first input() works but the second causes a segmentation fault.

### Steps to reproduce

main.mojo
```mojo
def main():
  a = input()
  print(a)
  b = input()
  print(b)
```
input.txt
```
str1
str2
str3
```
`magic run mojo build main.mojo`
`./main < input.txt`
```
str1
Segmentation fault (core dumped)
```

### System information

```shell
/etc/lsb-release

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu 24.04.1 LTS"

mojo 24.6.0 (4487cd6e)

- Provide Magic CLI version by pasting the output of `magic -V` or `magic --version` 

magic 0.6.2 - (based on pixi 0.40.0)

- Optionally, provide more information with `magic info`.
```

helehex:
Keep a global `_fdopen(stdin)` to fix #3961.
Adds a test file for `input()`.
Fixes a bug with last character getting removed when using `input()`.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/io.mojo
+++ b/mojo/stdlib/stdlib/builtin/io.mojo
@@ -24,7 +24,7 @@ from sys import (
     is_compile_time,
     is_gpu,
     is_nvidia_gpu,
-    stdin,
+    STDIN,
     stdout,
 )
 from sys._amdgpu import printf_append_args, printf_append_string_n, printf_begin
@@ -155,7 +155,9 @@ struct _fdopen[mode: StaticString = "a"]:
             raise Error("EOF")
         # Copy the buffer (excluding the delimiter itself) into a Mojo String.
         var s = String(
-            StringSlice[buffer.origin](ptr=buffer, length=bytes_read - 1)
+            StringSlice[buffer.origin](ptr=buffer, length=bytes_read).rstrip(
+                delimiter
+            )
         )
         # Explicitly free the buffer using free() instead of the Mojo allocator.
         libc.free(buffer.bitcast[NoneType]())
@@ -476,7 +478,7 @@ fn input(prompt: String = "") raises -> String:
     """
     if prompt != "":
         print(prompt, end="")
-    return _fdopen["r"](stdin).readline()
+    return STDIN.get_or_create_ptr()[].readline()
 
 
 fn _get_stdout_stream() -> OpaquePointer:
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -205,3 +205,7 @@ mutation.
 
 - [#4954](https://github.com/modular/modular/issues/4954) - `InlineArray`
   now does not call the copy constructor when being moved.
+
+- [#3961](https://github.com/modular/modular/issues/3961) - the second input()
+  when using standard input redirection causes
+  `Unhandled exception caught during execution: EOF`
--- a/mojo/magic.lock
+++ b/mojo/magic.lock
@@ -1,0 +1,1897 @@
+version: 6
+environments:
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    - url: https://conda.modular.com/max-nightly/
+    packages:
+      linux-64:
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-h4bc722e_7.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2025.6.15-hbd8a1cb_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/click-8.2.1-pyh707e725_0.conda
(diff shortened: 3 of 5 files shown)


mojo/stdlib/src/sys/_io.mojo
```diff
@@ -11,6 +11,16 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 """IO constants and functions."""
+from sys.ffi import _Global
+from builtin.io import _fdopen
+
 alias stdin = FileDescriptor(0)
+alias STDIN = _Global["FD_STDIN", _fdopen["r"], _init_stdin]
```
soraros: `STDIN` should probably grouped with `_init_stdin`. Also, why the name mismatch?

helehex: Thanks, I missed that when renaming

mojo/stdlib/stdlib/builtin/io.mojo
```diff
@@ -156,7 +156,9 @@ struct _fdopen[mode: StaticString = "a"]:
             raise Error("EOF")
         # Copy the buffer (excluding the delimiter itself) into a Mojo String.
         var s = String(
-            StringSlice[buffer.origin](ptr=buffer, length=bytes_read - 1)
+            StringSlice[buffer.origin](ptr=buffer, length=bytes_read).rstrip(
+                delimiter
+            )
```
soraros, line 160: Is using `rstrip` a bit wasteful (I do not know)? Can we check `buffer[bytes_read]` directly?

helehex: No I think it's good

soraros: Ok then. Hopefully most of the cost can be constant-folded away.

soraros: Nvm, already made #4548 to clean it up. Feel free to fold it into your current PR though.

helehex: I believe `rstrip()` is the correct thing to use here. The delimiter could be more than one character, and rstrip just re-slices the span which is pretty fast

helehex: maybe `rstrip` could be faster by manually changing the span instead of slicing it, but I would worry about that separately.

helehex: I think it makes sense to use the `c_types` here though.

soraros: The problem being the delimiter **cannot** really be more than on character, for `getdelim` doesn't support it. One might argue that we need to handle longer delimiters, but that should probably be solved higher up in the stack, not in the internal C wrapper type.

helehex: Yeah thats fair. It does currently get debug_asserted to be one char because of `ord()` but it's probably good to have assertions at every level.

mojo/stdlib/test/builtin/test_io_dummy_input.txt
JoeLoser, line 1: This is fine, but can you please put this in a `data` subdirectory so it's clear it's not a test and rather test/data input for a test?  Additionally, please add it to the data list [here](https://github.com/modular/modular/blob/43987188b35dc08323a1c4a3fd6140fb4f81e7f1/mojo/stdlib/test/BUILD.bazel#L3) in the Bazel build.

JoeLoser (commented):
Thanks for the fix!  Can you please expand a bit on the PR description outlining the solution?  Additionally, please add a changelog entry [here](https://github.com/modular/modular/blob/43987188b35dc08323a1c4a3fd6140fb4f81e7f1/mojo/docs/changelog.md#%EF%B8%8F-fixed)

JoeLoser (approved):

JoeLoser:
LGTM, thanks for the fix!

JoeLoser:
!sync

helehex:

mojo/stdlib/test/builtin/test_io.mojo
```diff
@@ -0,0 +1,24 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+# RUN: %mojo %s < %S/data/test_io_dummy_input.txt
```
helehex: Not sure why the bazel ones are failing... Maybe because of this?

helehex:
I removed the test

JoeLoser:
@helehex do you mind rebasing and then I'll sync this internally? Thanks!

JoeLoser:
!sync

JoeLoser:
@helehex one last rebase, please!  And then I can resync and land this internally.  There's one small conflict now that there's a dedicated `io` module.

PR #4530, opened 2025-05-09
[stdlib] Clean up implementation of `vectorize`

soraros:
* Inline `_perfect_vectorized_impl` since it's only used in one place
* Slightly reorganize the code to align the implementations of the two overloads
* Add a `debug_assert` for `size` in dynamic `vectorize`
* Fix docstring

### DIFF
--- a/mojo/stdlib/stdlib/algorithm/functional.mojo
+++ b/mojo/stdlib/stdlib/algorithm/functional.mojo
@@ -107,23 +107,25 @@ fn vectorize[
 
     ```mojo
     from algorithm.functional import vectorize
+    from memory import UnsafePointer
 
     # The amount of elements to loop through
     alias size = 10
     # How many Dtype.int32 elements fit into the SIMD register (4 on 128bit)
-    alias simd_width = simdwidthof[DType.int32]()
+    alias simd_width = simdwidthof[DType.int32]()  # assumed to be 4 in this example
 
     fn main():
         var p = UnsafePointer[Int32].alloc(size)
 
         # @parameter allows the closure to capture the `p` pointer
         @parameter
-        fn closure[simd_width: Int](i: Int):
-            print("storing", simd_width, "els at pos", i)
-            p.store[width=simd_width](i, i)
+        fn closure[width: Int](i: Int):
+            print("storing", width, "els at pos", i)
+            p.store[width=width](i, i)
 
         vectorize[closure, simd_width](size)
-        print(p.load[width=size]())
+        print(p.load[width=simd_width]())
+        print(p.load[width=simd_width](simd_width))
     ```
 
     On a machine with a SIMD register size of 128, this will set 4xInt32 values
@@ -141,7 +143,7 @@ fn vectorize[
     You can also unroll the loop to potentially improve performance at the cost
     of binary size:
 
-    ```mojo
+    ```
     vectorize[closure, width, unroll_factor=2](size)
     ```
 
@@ -149,7 +151,7 @@ fn vectorize[
     fewer arithmetic, comparison, and conditional jump operations. The assembly
     would look like this in psuedocode:
 
-    ```mojo
+    ```
     closure[4](0)
     closure[4](4)
     # Remainder loop won't unroll unless `size` is passed as a parameter
@@ -163,13 +165,26 @@ fn vectorize[
     exponent of 2 (2, 4, 8, 16, ...). The remainder loop will still unroll for
     performance improvements if not an exponent of 2.
     """
+    constrained[simd_width > 0, "simd width must be > 0"]()
+    constrained[unroll_factor > 0, "unroll factor must be > 0"]()
+    debug_assert(size >= 0, "size must be >= 0")
 
-    var vector_end_simd = align_down(UInt(size), UInt(simd_width))
-    _perfect_vectorized_impl[
-        func, simd_width=simd_width, unroll_factor=unroll_factor
-    ](size, vector_end_simd)
+    alias unrolled_simd_width = simd_width * unroll_factor
+    var simd_end = align_down(UInt(size), UInt(simd_width))
+    var unrolled_end = align_down(UInt(size), UInt(unrolled_simd_width))
 
-    for i in range(vector_end_simd, size):
+    for unrolled_idx in range(0, unrolled_end, unrolled_simd_width):
+
+        @parameter
+        for idx in range(unroll_factor):
+            func[simd_width](unrolled_idx + idx * simd_width)
+
+    @parameter
+    if unroll_factor > 1:
+        for simd_idx in range(unrolled_end, simd_end, simd_width):
+            func[simd_width](simd_idx)
+
+    for i in range(simd_end, size):
         func[1](i)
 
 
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/algorithm/functional.mojo
```diff
@@ -163,13 +165,25 @@ fn vectorize[
     exponent of 2 (2, 4, 8, 16, ...). The remainder loop will still unroll for
     performance improvements if not an exponent of 2.
     """
+    constrained[simd_width > 0, "simd width must be > 0"]()
+    constrained[unroll_factor > 0, "unroll factor must be > 0"]()
+
+    alias unrolled_simd_width = simd_width * unroll_factor
+    var simd_end = align_down(UInt(size), UInt(simd_width))
+    var unrolled_end = align_down(UInt(size), UInt(unrolled_simd_width))
```
soraros, line 174: Are these `UInt` significant (for performance reasons maybe?)?

mojo/stdlib/stdlib/algorithm/functional.mojo
```diff
@@ -141,15 +143,15 @@ fn vectorize[
     You can also unroll the loop to potentially improve performance at the cost
     of binary size:
 
-    ```mojo
+    ```python
```
gryznar: python?

soraros: The code is incomplete, so I used `python` to please the LSP while still getting some syntax highlighting.

gryznar: Is this the right way to go? Maybe it is up to LSP to handle such case of incomplete code (LSP issue)? Putting `python` here is confusing now and would be even more if Mojo syntax in future will be enhanced.

Overall I am not convinced with such workaround

soraros: I'm not proud of it either. Still, it's annoying to be greeted with irrelevant errors, especially when the LSP is already painfully slow at times. Maybe we should introduce a `mojo_unchecked` or something. In the short term, though, I'd still recommend using `python` or `plaintext`, if only for the sanity of stdlib contributors.

gryznar: `plaintext` is more correct than `python` at least

soraros: Done. Maybe you could write that feature request?

gryznar: I can

gryznar: @soraros done. Feel free to add additional context: https://github.com/modular/modular/issues/4552

mojo/stdlib/stdlib/algorithm/functional.mojo
martinvuyk: align_down for `UInt`
```asm
	testq	%rsi, %rsi
	je	.LBB0_1
	movq	%rdi, %rax
	orq	%rsi, %rax
	shrq	$32, %rax
	je	.LBB0_3
	movq	%rdi, %rax
	xorl	%edx, %edx
	divq	%rsi
	imulq	%rsi, %rax
	retq
.LBB0_1:
	xorl	%eax, %eax
	imulq	%rsi, %rax
	retq
.LBB0_3:
	movl	%edi, %eax
	xorl	%edx, %edx
	divl	%esi
	imulq	%rsi, %rax
	retq
```
align_down for `Int`
```asm
	cmpq	$1, %rsi
	movq	%rsi, %rcx
	adcq	$0, %rcx
	movq	%rdi, %rax
	orq	%rcx, %rax
	shrq	$32, %rax
	je	.LBB0_1
	movq	%rdi, %rax
	cqto
	idivq	%rcx
	jmp	.LBB0_3
.LBB0_1:
	movl	%edi, %eax
	xorl	%edx, %edx
	divl	%ecx
.LBB0_3:
	xorq	%rsi, %rdi
	sets	%cl
	testq	%rdx, %rdx
	setne	%dl
	andb	%cl, %dl
	movzbl	%dl, %ecx
	subq	%rcx, %rax
	testq	%rsi, %rsi
	cmoveq	%rsi, %rax
	imulq	%rsi, %rax
	retq
```

martinvuyk: Sidenote: I'm adding a faster version of `align_up` for compile time known powers of two in  #4416. Then I'd like to do the same for `align_down`. And then I want to replace all these places where they are used with compile time known values to make them way faster

soraros: If the compiler doesn't already prove that `size > 0` somewhere and cut down the branch. But I guess I could add a `debug_assert` and a `expect` instead.

martinvuyk: I think both these variables can be instantiated as alias instead

soraros: > I think both of these variables can be instantiated as aliases instead

`size` is a `var` in this overload.

> Then I'd like to do the same for `align_down`.

I think LLVM will be able to constant-fold and simplify calls like `align_down(_, n)` when `n` is an alias and a power of two.

I also made #4557.

martinvuyk: > I think LLVM will be able to constant-fold and simplify calls like align_down(_, n) when n is an alias and a power of two.

Will it do this though?
```mojo
    @parameter
    if multiple_of == 1:
        return value
    elif multiple_of.is_power_of_two():
        alias offset = multiple_of - 1
        return (value + offset) & ~offset
    else:
        return ceildiv(value, multiple_of) * multiple_of
```
I guess I will have to look at the asm dump at some point when I have time

> I also made https://github.com/modular/modular/pull/4557.

That is great!

JoeLoser (approved):

JoeLoser:
!sync

mojo/stdlib/stdlib/algorithm/functional.mojo
```diff
@@ -107,23 +107,25 @@ fn vectorize[
 
     ```mojo
     from algorithm.functional import vectorize
+    from memory import UnsafePointer
 
     # The amount of elements to loop through
     alias size = 10
     # How many Dtype.int32 elements fit into the SIMD register (4 on 128bit)
-    alias simd_width = simdwidthof[DType.int32]()
+    alias simd_width = 4
```
JoeLoser: From feedback internally:
> can we keep this as `simdwidthof` instead of a magic number. 128 bits is not the simd width of avx512 for example

JoeLoser (approved):
LGTM @soraros, but can you please provide a detailed commit message rather than leaving it empty?  This came up during internal code review.  This applies to your other open PRs too.  Thank you!

mojo/stdlib/stdlib/algorithm/functional.mojo
soraros: Though using `simdwidthof` means that on some platforms people can't reproduce the result from the example. I can of course revert.

JoeLoser: I agree. I'd leave it as you have it; `simdwidth[DType.int32]()` with a comment.

JoeLoser:
Happy to sync once there's a PR description @soraros 😄

soraros:
@JoeLoser Done. Please take a look. I edited the commit message, but forgot about the PR description.

JoeLoser:
!sync

PR #3083, opened 2024-06-20
[stdlib] Simplify `SIMD.reduce_op` functions

soraros:
- Simplify the signature of `reduce` since it's doesn't have to be generic over `type`
- Introduced a `reduce` overload for non-capturing functions
- Simplify `SIMD.reduce_op` functions using `SIMD.op` methods directly

Issue #4538, opened 2025-05-09
[BUG] Compiler fails to find correct overload with inferred parameter on a generic struct


soraros:
### Bug description

As title. Related to #4199. CC @dgurchenkov.

### Steps to reproduce

Code:
```mojo
@fieldwise_init
struct Ptr[dtype: DType = DType.int32]:
  fn store(self, offset: Int, val: Int):
    pass

  fn store[width: Int, //](self, offset: Int):
    pass

fn main():
  p = Ptr()
  i = 1
  p.store[width=4](i)  # error
```
Error:

```
/workspaces/ubuntu/max/playground/main.mojo:31:19: error: no matching method in call to 'store'
  p.store[width=4](i)
  ~~~~~~~~~~~~~~~~^~~
/workspaces/ubuntu/max/playground/main.mojo:22:6: note: candidate not viable: missing 1 required positional argument: 'val'
  fn store(self, offset: Int, val: Int):
     ^
/workspaces/ubuntu/max/playground/main.mojo:25:6: note: candidate not viable: inferred parameter passed out of order: 'width'
  fn store[width: Int, //](self, offset: Int):
```

### System information

`mojo 25.4.0.dev2025050905 (e3051145)`

VerdagonModular:
Repro'd, thanks for the snippet! This might get fixed soon in some coming refactors in the parameter inference system, fingers crossed. Moving to the backlog to be considered for next cycle, just in case.

soraros:
Another interesting case. Maybe this widening should be encoded as a `__merge_with__` instead of implicit conversion.

```mojo
struct V[dtype: DType, size: Int]:
  fn __init__(out self):
    pass

  @implicit
  fn __init__(out self, data: V[dtype, 1]):
    pass

fn fma[
  dtype: DType, width: Int, //
](a: V[dtype, width], b: V[dtype, width], c: V[dtype, width]):
  pass

fn f[dtype: DType, width: Int]():
  var s = V[dtype, 1]()
  var v = V[dtype, width]()

  fma[width=width](s, v, v)  # error: invalid call to 'fma': failed to infer parameter 'dtype'
  fma[dtype=dtype, width=width](s, v, v)
```

Error:
```
/workspaces/ubuntu/max/playground/bug.mojo:39:19: error: invalid call to 'fma': failed to infer parameter 'dtype'
  fma[width=width](s, v, v)
  ~~~~~~~~~~~~~~~~^~~~~~~~~
[...]

PR #4549, opened 2025-05-10
[stdlib] Make inferred parameters explicit in `math.mojo`

soraros:
So we can pass them partially bound to higher-order functions. Example: #3083

### DIFF
--- a/max/kernels/src/linalg/accumulate.mojo
+++ b/max/kernels/src/linalg/accumulate.mojo
@@ -846,7 +846,7 @@ struct _Accumulator[
                     for i in range(row_start, row_stop):
                         # The following should be lifted to registers and show up as
                         # FMA instructions.
-                        self[i, j] = fma[type, simd_width](
+                        self[i, j] = fma[dtype=type, width=simd_width](
                             a_vecs[i][lane].cast[type](),
                             b_vec.cast[type](),
                             self[i, j],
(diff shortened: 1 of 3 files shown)


max/kernels/src/linalg/accumulate.mojo
```diff
@@ -846,7 +846,7 @@ struct _Accumulator[
                     for i in range(row_start, row_stop):
                         # The following should be lifted to registers and show up as
                         # FMA instructions.
-                        self[i, j] = fma[type, simd_width](
+                        self[i, j] = fma[dtype=type, width=simd_width](
```
soraros, line 849: I have to specify `type` here for otherwise the compiler can't find this overload, feels like a bug. CC @VerdagonModular.

The change I made to `fma` is just marking its parameters as inferred. Maybe another case of #4538?

JoeLoser: Agreed, can you file a GH issue please?

soraros: Added a small repro here: https://github.com/modular/modular/issues/4538#issuecomment-2880728531.

JoeLoser: Awesome, thanks for the small repo! ❤️

JoeLoser (approved):

JoeLoser:
!sync

PR #4555, opened 2025-05-10
[stdlib] Variadic argument in Counter constructor and better typing

msaelices:

### DIFF
--- a/mojo/stdlib/stdlib/collections/counter.mojo
+++ b/mojo/stdlib/stdlib/collections/counter.mojo
@@ -37,7 +37,7 @@ struct Counter[V: KeyElement](Sized, Copyable, Movable, Boolable):
 
     ```mojo
     from collections import Counter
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/collections/counter.mojo
```diff
@@ -56,13 +56,40 @@ struct Counter[V: KeyElement](Sized, Copyable, Movable, Boolable):
...
+        print(c["b"])  # print 2
+        ```
+        """
+        self._data = Dict[V, Int]()
+        for item_ref in values:
+            var item = item_ref[]
+            self._data[item] = self._data.get(item, 0) + 1
+
     @implicit
     fn __init__(out self, items: List[V, *_]):
```
gryznar, line 79: 1. Comment above suggested to replace this constructor with variadic one. Is this still relevant?
2. Do we need `implicit` here?

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

PR #4557, opened 2025-05-11
[stdlib] Add a parameter for inserting `llvm.assume` in `debug_assert`

soraros:
In principle, `debug_assert` enforces an invariant that must hold at runtime. This information can also be valuable to the compiler for optimisation. Therefore, we added a `llvm_assume` parameter to `debug_assert` so it can emit an `llvm.assume` even when error reporting is disabled. We don't enable it by default, however, because in cases like using `select` over a "soft UB" in `prev_power_of_two`, emitting an `llvm.assume` could turn that soft UB into a hard one.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/debug_assert.mojo
+++ b/mojo/stdlib/stdlib/builtin/debug_assert.mojo
@@ -20,7 +20,7 @@ from os import abort
 from sys import is_amd_gpu, is_gpu, is_nvidia_gpu, llvm_intrinsic
 from sys._build import is_debug_build
 from sys.ffi import c_char, c_size_t, c_uint, external_call
-from sys.intrinsics import block_idx, thread_idx
+from sys.intrinsics import block_idx, thread_idx, assume
 from sys.param_env import env_get_string
 
 from builtin._location import __call_location, _SourceLocation
@@ -63,10 +63,12 @@ fn _assert_enabled[assert_mode: StaticString, cpu_only: Bool]() -> Bool:
 @always_inline
 fn debug_assert[
     cond: fn () capturing [_] -> Bool,
+    /,
     write_mode: WRITE_MODE = WRITE_MODE_REG,
     assert_mode: StaticString = "none",
-    cpu_only: Bool = False,
     *Ts: Writable,
+    cpu_only: Bool = False,
+    llvm_assume: Bool = False,
 ](*messages: *Ts, location: Optional[_SourceLocation] = None):
     """Asserts that the condition is true at run time.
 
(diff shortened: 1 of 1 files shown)


JoeLoser:
In the spirit of over communicating, can you please also add a PR description sharing the motivation, details, etc? This will help out other reviewers and be good to have in the Git history as well.

soraros:
@JoeLoser Fixed both the PR description and the commit message. PTAL.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

jackos:
Hi @soraros this was causing issues with some recent changes to `debug_assert` to fix register pressure on GPU.

My understanding of the purpose of when to use `llvm.assume` is when you know a condition is satisfied so the compiler can do optimizations like DCE e.g. if `UnsafePointer` had bounds checking:

```mojo
fn process_array(data: UnsafePointer[Int], size: Int, idx: Int) -> Int:
    # You've already validated idx elsewhere in your program but the compiler doesn't know this
    assume(idx >= 0 and idx < size)
    
    # Now the compiler can eliminate bounds checking and generate more efficient code
    return data[idx] * 2
```

Testing this works according to my understanding, this will still pass even when asserts are on:

```mojo
from sys.intrinsics import assume

fn foo(idx: Int):
    assume(idx > 0)
    debug_assert(idx > 0)
    print(idx)

fn main():
    foo(-1)
```

From all the code paths in the code you've added I don't understand what the purpose is:

- When assertions are enabled: if fails assertion the program aborts, otherwise inserts an assume(True) which doesn't do anything?
- When assertions are disabled: The entire function body was previously getting DCE'd, the PR makes it so the condition still gets evaluated at runtime when `llvm_assume=True` and inserts an `assume(True)` or `assume(False)` which doesn't do anything?

Not sure if my understanding is off or I'm missing anything here?

soraros:
@jackos I think you are right that my current implementation is pointless when assertions are enabled. Should probably be like this:

```mojo
fn debug_assert(...):
  @parameter
  if assert_enabled[...]():
    if cond:
      # you are right that insert `assume` here doesn't make sense
      return
    debug_assert_msg(messages)
  @parameter
  if llvm_assume:  # currently this is a elif
    assume(cond)
```

Though currently, it does insert the `assume(expr)` when assertions are not enabled. You can try `mojo build --emit=llvm` this:

```mojo
fn main():
  print(f(10))

@export
fn f(n: Int) -> Int:
  debug_assert[llvm_assume = True](n > 0)
  if n > -1:
    return 0
  return n
```

The goal of this is basically to turn some of the `debug_assert` into an `assume` without duplicating code.

jackos:
Ok thanks @soraros I understand now, when asserts are enabled it's:
```mojo
debug_assert(cond)
```
And when asserts are disabled it essentially gets transformed to:
```mojo
assume(cond)
```
Testing this it works as expected:
```mojo
fn foo(idx: Int):
    debug_assert[llvm_assume = True](idx > 0)
    if idx > 0:
        print(idx)
    else:
        print("negative")

fn main():
    foo(1)
```
Eliminating the else branch because it inserts:
```
  %2 = icmp sgt i64 %0, 0
  call void @llvm.assume(i1 %2)
```
And essentially is doing `assume(cond)`, it's not just doing `assume(True)`

Will try bringing it back in the internal repo

PR #4562, opened 2025-05-11
[stdlib] Speed-up of `String.join()`

msaelices:
This revisits previous optimizations made on [02a2e4b3b08afb7030f0ff1c0d158cf90b9d9eca]

Somehow, this optimization was removed in later refactorizations, slowing down x4-x6 times the join logic

## Benchmarks

Code:
```mojo
from time import perf_counter_ns as now

fn main():
    l = List[String](capacity=1_000_000)
    for i in range(1_000):
        l.append(String(i))
    start = now()
    s = String("")
    for _ in range(10_000):
        s = String(",").join(l)
    end = now()
    print('Len: ', len(s), 'Time: ', (end - start) / 1_000_000_000, 'seconds')
```

### Before this commit
```
❯ mojo bench_join.mojo 
Len:  3889 Time:  0.186151527 seconds
```

### After this commit
```
❯ mojo bench_join.mojo 
Len:  3889 Time:  0.045179215 seconds
```

### Speed-up

~4 times faster

### Note
This is the first time we achieve more performance than the Python equivalent, probably because of the recent SSO optimizations:

image

## More information

See Discord thread:

https://discord.com/channels/1087530497313357884/1331776837738696828/1331957122409435146


gryznar:
I am curious how this compares with e.g. Rust?

xtxtxtxtxtxtx:
> I am curious how this compares with e.g. Rust?

```rust
use std::time::Instant;

fn main() {
    let mut l = Vec::with_capacity(1_000_000);
    for i in 0..1_000 {
        l.push(i.to_string());
    }

    let start = Instant::now();
    let mut s = String::new();
    for _ in 0..10_000 {
        s = l.join(",");
    }
    let duration = start.elapsed();

    println!(
        "Len: {}, Time: {:.6} seconds",
        s.len(),
        duration.as_secs_f64()
    );
}
```
This code runs with practically no diffrence from Optimized Mojo one.Like less than 1% diff on my machine. IIRC rust does not have SSO so there should be some more optimization remaning.

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1075,14 +1075,22 @@ struct String(
...
-            result.write(elems[i])
-        return result^
+        var sep = StaticString(ptr=self.unsafe_ptr(), length=len(self))
+        var total_bytes = _TotalWritableBytes(elems, sep=sep)
+
+        # Use heap if over the stack buffer size
+        if total_bytes.size + 1 > buffer_size:
+            var buffer = _WriteBufferHeap(total_bytes.size + 1)
+            buffer.write_list(elems, sep=sep)
+            buffer.data[total_bytes.size] = 0
```
soraros: We don't guarantee nul-term any more, so this is not necessary.

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1075,14 +1075,22 @@ struct String(
...
+            return String(
```
soraros: This copies the data, is that intended?

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1075,14 +1075,22 @@ struct String(
...
+                StringSlice(ptr=buffer.data, length=total_bytes.size + 1)
+            )
+        # Use stack otherwise
+        else:
+            var result = String()
+            write_buffered[buffer_size](result, elems, sep=sep)
+            return result
```
soraros: Over all, I think it will look slightly cleaner if you use the guard pattern, namely:
```mojo
if total_bytes.size <= buffer_size:
  var res = String()
  write_buffered(...)
  return res
# the main case here.
var result = ...
...
return String(...)
```

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1075,14 +1075,22 @@ struct String(
...
+        if total_bytes.size + 1 > buffer_size:
+            var buffer = _WriteBufferHeap(total_bytes.size + 1)
+            buffer.write_list(elems, sep=sep)
+            buffer.data[total_bytes.size] = 0
+            return String(
+                StringSlice(ptr=buffer.data, length=total_bytes.size + 1)
+            )
+        # Use stack otherwise
+        else:
+            var result = String()
```
soraros: I'm not sure about the details, but would it be beneficial if we use the `String(cap=...)` variant here?

mojo/stdlib/stdlib/collections/string/string.mojo
msaelices: Good point! Fixed: https://github.com/msaelices/mojo/commit/99957bd7b2d2ea173629036263463336a81db263

mojo/stdlib/stdlib/collections/string/string.mojo
msaelices: I did not find a better way. I've thought on using the `String(unsafe_from_utf8_ptr=)` but it copies internally the data anyways. The issue is that the `String(ptr=, length=)` constructor has been removed.

Do you have a better option?

mojo/stdlib/stdlib/collections/string/string.mojo
msaelices: I agree, I wanted to honour the initial commit by the modular team: https://github.com/modular/modular/commit/02a2e4b3b08afb7030f0ff1c0d158cf90b9d9eca

Let's clean it up.

msaelices: Done: https://github.com/msaelices/mojo/commit/1a190059d587fc3ac43a24d16f9e7eb75ffd285c

mojo/stdlib/stdlib/collections/string/string.mojo
gryznar: Martin had some discussion with Chris on this topic. Maybe there was some consensus

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1075,14 +1075,19 @@ struct String(
...
         """
-        var result = String()
-        if not len(elems):
-            return result^
-        result.write(elems[0])
-        for i in range(1, len(elems)):
-            result.write(self)
-            result.write(elems[i])
-        return result^
+        var sep = StaticString(ptr=self.unsafe_ptr(), length=len(self))
```
soraros:
```suggestion
        var sep = StaticString(ptr=self.unsafe_ptr(), length=self.byte_length())
```

mojo/stdlib/stdlib/collections/string/string.mojo
soraros: CC @martinvuyk.

mojo/stdlib/stdlib/collections/string/string.mojo
msaelices: I thought not because of the `write_buffered` usage which are not adding the chars to the String but to a stack based buffer instead (`_WriteBufferStack`). Am I wrong?

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1075,14 +1075,19 @@ struct String(
...
+        var total_bytes = _TotalWritableBytes(elems, sep=sep)
+
+        if total_bytes.size <= buffer_size:
+            # Use stack if we are under the stack buffer size
+            var result = String()
+            write_buffered[buffer_size](result, elems, sep=sep)
+            return result
+
+        # Use heap otherwise
+        var buffer = _WriteBufferHeap(total_bytes.size + 1)
```
soraros: Why do we need the `+1` here?

mojo/stdlib/stdlib/collections/string/string.mojo
msaelices: Fixed: https://github.com/msaelices/mojo/commit/33baebef5422fe0ac5d600a34039be60f8a1ef94

mojo/stdlib/stdlib/collections/string/string.mojo
msaelices: for allocating the null term. Which is not needed. Fixed: https://github.com/msaelices/mojo/commit/396dce7046d9c3ff412f06e32e41e5715241415a

mojo/stdlib/stdlib/collections/string/string.mojo
soraros: You are right.

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1075,14 +1075,19 @@ struct String(
         Returns:
             The joined string.
         """
-        var result = String()
```
soraros, line 1002: Naive question: what if we simply compute the capacity and retain the old code?

martinvuyk:
Hi @msaelices and everyone else, seems I was summoned lol.

> @soraros: Naive question: what if we simply compute the capacity and retain the old code?

I think this is the path, especially because of the discussion of using the List APIs and building a String out of that with Chris culminated in basically: use String's public API and `memcpy` where necessary.

Heap and stack writing of string could be optimized like:
```mojo
var sep = StaticString(ptr=self.unsafe_ptr(), length=self.byte_length())
var total_bytes = _TotalWritableBytes(elems, sep=sep)
var result = String(capacity=total_bytes)
result.write(elems[0])
for i in range(1, len(elems)):
    result.write(self)
    result.write(elems[i])
return result^
```
I think this implementation should be taken to `StringSlice` and replace the broken one there

msaelices:
> martinvuyk

Not sure if I got your suggestion right. I did this commit https://github.com/msaelices/mojo/commit/62684f4b507d604a8075635434e81be7fca4e773 following your suggestion but the new logic slows it down again x3

We will try to add the stack / heap branching again in StringSlice as I think it will optimize everything again.

gryznar:
@msaelices in your recent commt, you pointed out, that performance get a regression. Could you rerun benchmarks? 

Doyou see  some mitigations to have both correctness (as with `StringSlice`) and performance?

msaelices:
> @msaelices in your recent commt, you pointed out, that performance get a regression. Could you rerun benchmarks?
> 
> Do exist some mitigations to have both correctness (as with `StringSlice`) and performance?

I'm working on it. But this was my previous bench execution:

```
❯ MODULAR_MOJO_MAX_IMPORT_PATH=./build mojo bench_join.mojo 
Len:  3889 Time:  0.169267593 seconds
```
So it went from `0.045179215` seconds to `0.169267593`

msaelices:
> > @msaelices in your recent commt, you pointed out, that performance get a regression. Could you rerun benchmarks?
> > Do exist some mitigations to have both correctness (as with `StringSlice`) and performance?
> 
> I'm working on it. But this was my previous bench execution:
> 
> ```
> ❯ MODULAR_MOJO_MAX_IMPORT_PATH=./build mojo bench_join.mojo 
> Len:  3889 Time:  0.169267593 seconds
> ```
> 
> So it went from `0.045179215` seconds to `0.169267593`

I managed to optimize it again with this change: https://github.com/msaelices/mojo/commit/309dee85ba329f769294b7a6242ccc25096ee2cb

```
❯ MODULAR_MOJO_MAX_IMPORT_PATH=./build mojo bench_join.mojo 
Len:  3889 Time:  0.044271145 seconds
```

gryznar (approved):

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2152,28 +2152,43 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
+        var total_bytes = _TotalWritableBytes(elems, sep=sep).size
+
+        if total_bytes <= buffer_size:
+            # Use stack if we are under the stack buffer size
+            var result = String()
+            write_buffered[buffer_size](result, elems, sep=sep)
+            return result^
+
+        # Use heap otherwise
+        var result = String(capacity=total_bytes)
```
martinvuyk: Does this have any effect on perf?

```suggestion
        var result = String(capacity=total_bytes)

        if total_bytes <= buffer_size:
            # Use stack if we are under the stack buffer size
            write_buffered[buffer_size](result, elems, sep=sep)
            return result^

        # Use heap otherwise
```

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2152,28 +2152,43 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
             T: The type of the elements, must implement the `Copyable`,
                 `Movable` and `Writable` traits.
+            buffer_size: The max size of the stack buffer.
 
         Args:
             elems: The input values.
 
         Returns:
             The joined string.
         """
```
martinvuyk: Nit-pick
```suggestion
        """Joins string elements using the current string as a delimiter.

        Parameters:
            T: The type of the elements, must implement the `Copyable`,
                `Movable` and `Writable` traits.
            buffer_size: The max size of the stack buffer.

        Args:
            elems: The input values.

        Returns:
            The joined string.
 
         Notes:
            Defaults to writing to the stack if total bytes of `elems` is less than
            `buffer_size`, otherwise will allocate once to the heap and write
            directly into that. The `buffer_size` defaults to 4096 bytes to match
            the default page size on arm64 and x86-64, but you can increase
            this if you're joining a very large `List` of elements to write into the
            stack instead of the heap.
        """
```

martinvuyk:
@msaelices I've been thinking, I think we should do this in another way:
```mojo
if len(elems) == 0:
    return String()

var sep = StaticString(ptr=self.unsafe_ptr(), length=self.byte_length())
var total_bytes = _TotalWritableBytes(elems, sep=sep).size
var result = String(capacity=total_bytes)

if result._capacity_or_data.is_inline():
    result.write(elems[0])
    for i in range(1, len(elems)):
        result.write(self, elems[i])
    return result^

var buffer = _WriteBufferStack[buffer_size](result)        

buffer.write(elems[0])
for i in range(1, len(elems)):
    buffer.write(self, elems[i])
buffer.flush()
return result^
```
I think this gives the best perf for all 3 scenarios:
- if it is inline it writes directly to the stack address
- if it fits inside a buffer it gets written in the stack then flushed
- if it doesn't fit inside a buffer it gets processed by chunks (anything to be written to the heap is faster if given a page instead of doing several trips anyway!)

So in a way the original `StringSlice.join()` implementation was faster already than the existing String one

mojo/stdlib/stdlib/collections/string/string.mojo
JoeLoser: @msaelices did you explore this suggestion from @soraros?  I'm curious as well.

msaelices: I did and it was x4 times slower again. I think the right move is to wait for this PR: https://github.com/modular/modular/pull/4573 and then simplify the specific `join()` code once the write function itself is optimized

martinvuyk: > wait for this PR: https://github.com/modular/modular/pull/4573 and then simplify the specific join() code once the write function itself is optimized

That PR is for `String.write(VariadicPack)` not for `String.write(List)` (that isn't implemented yet), which I think will just use this implementation you are writing. You could use the code I wrote in https://github.com/modular/modular/pull/4562#issuecomment-2878160467 here

msaelices:
> * gets processed by chunks

@martinvuyk, you were totally right! your version is the fastest! Thanks so much! 

Done here:  https://github.com/msaelices/mojo/commit/059e7df07df6ebf99bb76387dd3c39cc9f34ad9b

mojo/stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -123,6 +123,32 @@ fn bench_string_split[
...
+        count = 1000
+
+    word_list = List[String](capacity=count)
+    for i in range(count):
+        word_list.append(String(i))
+
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        for _ in range(1_000):
```
martinvuyk, line 148: I don't know how much time this takes but if it is any similar to https://github.com/modular/modular/pull/4573 then you might want to set this to a higher number like `1_000_000` because otherwise the measurements aren't really stable. I normally aim to get the execution time above miliseconds

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1050,39 +1055,25 @@ struct String(
         Returns:
             The joined string.
         """
-        var sep = StaticString(ptr=self.unsafe_ptr(), length=len(self))
+        var sep = StaticString(ptr=self.unsafe_ptr(), length=self.byte_length())
         return String(elems, sep=sep)
 
     fn join[
-        T: Copyable & Movable & Writable, //, buffer_size: Int = 4096
```
martinvuyk: we could still keep this `buffer_size` param and pass it down to the `StringSlice` implementation

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2152,28 +2152,51 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
-        var string = String()
-        var buffer = _WriteBufferStack(string)
-        for i in range(len(elems)):
-            buffer.write(elems[i])
-            if i < len(elems) - 1:
-                buffer.write(self)
+
+        Notes:
+            - Defaults to writing to the stack if total bytes of `elems` is
+            inline, otherwise will process it by chunks.
```
martinvuyk:
```suggestion
            - Defaults to writing directly to the string if the bytes
            fit in an inline `String`, otherwise will process it by chunks.
```

martinvuyk (approved):
LGTM

mojo/stdlib/stdlib/collections/string/string_slice.mojo
msaelices: Fixed: https://github.com/msaelices/mojo/commit/0dc8315533cd4e0bb03cf505d19901d1060867b3

mojo/stdlib/stdlib/collections/string/string.mojo
msaelices: Good idea. Fixed: https://github.com/msaelices/mojo/commit/0dc8315533cd4e0bb03cf505d19901d1060867b3

mojo/stdlib/benchmarks/collections/bench_string.mojo
msaelices: It takes longer than I expected in my local machine, so I let it with just 1_000. We can refine the value later on.

msaelices:

JoeLoser:
@jackos do you mind taking a look at this, please?

msaelices:
> @jackos do you mind taking a look at this, please?

@jackos @JoeLoser have you guys got a chance to take a look at this? Thanks in advance

jackos:

jackos (approved):
Thanks, I bought back the _TotalWritableBytes struct and this is showing over 2x improvment

jackos:
!sync

PR #4563, opened 2025-05-12
[stdlib] Clean up `Int` and `UInt`

soraros:
CC @lattner.

- mark several methods as `"builtin"`; add TODOs where this is not yet possible
- reorder things to minimize differences between the two modules

### DIFF
--- a/mojo/stdlib/stdlib/builtin/uint.mojo
+++ b/mojo/stdlib/stdlib/builtin/uint.mojo
@@ -20,22 +20,29 @@ from hashlib.hash import _hash_simd
 from math import CeilDivable
 from sys import bitwidthof
 
-from documentation import doc_private
+from builtin.math import Absable
 
+from utils import Writable, Writer
+from utils._select import _select_register_value as select
 from utils._visualizers import lldb_formatter_wrapping_type
-from utils.write import Writable
 
 
 @lldb_formatter_wrapping_type
 @register_passable("trivial")
 struct UInt(
-    Indexer,
-    _HashableWithHasher,
-    Stringable,
-    Comparable,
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/builtin/uint.mojo
```diff
@@ -681,6 +697,15 @@ struct UInt(
         """
         return self != 0
 
+    @always_inline("builtin")
+    fn __as_bool__(self) -> Bool:
```
JoeLoser: **Question** Why add this other than because `Int` has it?  I wouldn't suggest drive-by adding things like with this refactorings.

soraros: I suppose `UInt` should be `ImplicitlyBoolable`? Of course I can revert this.

JoeLoser:
!sync

JoeLoser (approved):
Looks good, thank you!

lattner (approved):
nice!

JoeLoser:
@soraros can you please resolve the conflicts? Thanks!

JoeLoser:
!sync

soraros:
@JoeLoser Gentle ping. Is this blocked internally?

JoeLoser:
> @JoeLoser Gentle ping. Is this blocked internally?

Yeah, it's failing a few tests on H100 due to cuda errors about using too many resources. I'd suggest poking at the PR to make this truly NFC in case anything functional change wise sticks out.

soraros:
The only differences I can spot are two `@no_inline`. Shall we try again?

JoeLoser:
!sync

PR #4571, opened 2025-05-13
[CODEOWNERS] Update public CODEOWNERS

JoeLoser:
Update the CODEOWNERS file now that we have a few new GitHub teams available in the public modular/modular repo. Specifically:
- Add the @modular/kernels team for kernels changes.
- Add some refinements for specific parts of the stdlib.
- Add specific tracing codeowners.
- Update a few "examples" code owners to include folks from docs.


JoeLoser:
!sync

Issue #4346, opened 2025-04-14
[Feature Request] Add `.name` syntax for parameter access


soraros:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Introduce a new syntax for parameter access using a prefixed dot — `.name` — within brackets. This syntax would serve as a clear and lightweight way to reference parameters that are otherwise implicit, hidden, or shadowed. The idea is to bring clarity and structure to parameterized types and functions, particularly in more complex metaprogramming scenarios.

Use cases include:

- **Accessing defaulted parameters on type aliases**  
  ```mojo
  Int64[.size=4]
  ```

- **Specifying inferred-only parameters**  
  ```mojo
  f[.dtype=DType.int64]  
  S[.mut=True]
  ```

- **Disambiguating when parameter names are shadowed**  
  ```mojo
  S[.type=..., type=...]()  # the first one a type parameter on the struct and the second a parameter of __init__
  ```

In short: use `.name` to refer to parameters not part of the surface interface — things that are hidden, derived, or structurally inferred.

### Scope

This syntax is local to **bracketed parameter contexts**:

- Type application: `T[.param=...]`
- Generic function calls: `f[.param=...](...)`
- Overloaded dispatch like `__getitem__`: `x[.index]()`

It does **not** apply to regular field or property access. It’s not a general-purpose dot; it’s syntactic sugar for structural access to otherwise inaccessible parameter bindings.

#### 1. Structural clarity for inferred-only parameters

Mojo already distinguishes inferred and explicit parameters structurally, e.g., via `//`. This proposal extends that discipline: inferred-only parameters should be marked distinctly — `.name` is an elegant way to do so. It gives a visual cue that you're reaching into the hidden machinery.

#### 2. Better alias semantics

Suppose we define:

```mojo
alias Scalar = SIMD[_, size=1]
alias Int64 = Scalar[DType.int64]
```

Now: can a user still tweak `size` when working with `Int64`?

With `.name`, the answer is unambiguous — yes:

```mojo
alias VI4 = Int64[.size=4]
```

The design reflects intent: you expose what you want via regular `name=...`, and anything left hidden is still reachable via `.name=...`. You don't leak internals accidentally, but neither do you seal them off entirely.

#### 3. Disambiguating overloaded bracket syntax

[...]

PR #4580, opened 2025-05-14
[stdlib] Add List.__contains__ for distinct StringSlices

### LINKED ISSUE #4579: [BUG] List.__contains__ doen't work when working with distinct StringSlices
### Actual behavior
List.__contains__ expect both types to be the same, in order to compare both values.
It means, if we try to use two distinct StringSlices, with different origins, we can't compare them:

### Expected behavior
We should be able to compare StringSlices, no matter the origin, so we don't need to create a String or create another list just to check if a value exists.

### Steps to reproduce

```mojo
fn main():
    first_string = String("foo")
    first_slice = first_string.as_string_slice()
    lst = List[StringSlice[first_slice.origin]](first_slice)

    second_string = String("bar")
    second_slice = second_string.as_string_slice()
    _val = lst.__contains__(second_slice)  # invalid call to '__contains__': could not deduce positional-only parameter #2 of callee '__contains__'
    # failed to infer parameter #2, parameter inferred to two different values: 'StringSlice[(muttoimm first_string)]' and 'StringSlice[(muttoimm second_string)]'
```

### System information

System
------------
       Pixi version: 0.41.4
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=5.15.167.4=0
                   : __glibc=2.35=0
                   : __cuda=12.6=0
                   : __archspec=1=skylake

Package     Version               Build    Size       Kind   Source
max         25.4.0.dev2025051405  release  10.6 KiB   conda  max
max-core    25.4.0.dev2025051405  release  252.7 MiB  conda  max-core
max-python  25.4.0.dev2025051405  release  12.6 MiB   conda  max-python

IbarraSamuel:
Fixes https://github.com/modular/modular/issues/4579

By adding an overload for List.__contains__ where the types are not the same, cannot be implicitly converted, but can be compared one to the other (StringSlices with distinct origins)

### DIFF
--- a/mojo/stdlib/stdlib/collections/list.mojo
+++ b/mojo/stdlib/stdlib/collections/list.mojo
@@ -308,6 +308,29 @@ struct List[T: Copyable & Movable, hint_trivial_type: Bool = False](
                 return True
         return False
 
+    fn __contains__[
+        origin: Origin, //
+    ](self: List[StringSlice[origin], *_], value: StringSlice[*_]) -> Bool:
+        """Verify if a given StringSlice is present in the list.
+
+        ```mojo
+        var x = List[Int](1,2,3)
+        if 3 in x: print("x contains 3")
+        ```
+        Parameters:
+            origin: The origin of the underlying string data.
+
+        Args:
+            value: The value to find.
+
+        Returns:
+            True if the value is contained in the list, False otherwise.
+        """
+        for i in self:
+            if i[] == value:
+                return True
+        return False
+
     fn __mul__(self, x: Int) -> Self:
         """Multiplies the list by x and returns a new list.
 


bgreni:
This problem is also present for `Span` and therefore any other "view" types we might have. This could easily turn into a game of Whac-a-Mole if we can't find a better solution to this.

mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -308,6 +308,29 @@ struct List[T: Copyable & Movable, hint_trivial_type: Bool = False](
...
 
+    fn __contains__[
+        origin: Origin, //
+    ](self: List[StringSlice[origin], *_], value: StringSlice[*_]) -> Bool:
+        """Verify if a given StringSlice is present in the list.
+
+        ```mojo
+        var x = List[Int](1,2,3)
+        if 3 in x: print("x contains 3")
+        ```
```
martinvuyk, line 319: Just a nit-pick on the ordering, could you take this down below the returns and under an `Examples:` section? Also, your example isn't really using the `StringSlice` type

martinvuyk:
> This problem is also present for `Span` and therefore any other "view" types we might have. This could easily turn into a game of Whac-a-Mole if we can't find a better solution to this.

We could probably build a core trait for all view like types using parametric traits:
```mojo
trait ViewType[mut: Bool, //, origin: Origin[mut]]:
    ...

struct List:
    fn __contains__[
        V1: ViewType & Copyable & Movable & Writable,
        V2: V1[.mut=_, .origin=_] & Copyable & Movable & Writable & Comparable[V1[.mut=_, .origin=_]],
    ](self: List[V1, *_], value: V2): ...
```

mojo/stdlib/stdlib/collections/list.mojo
soraros: And shall we format the code block with `mblack` first?

soraros:
It has the same root cause as #4386, so indeed very bad. Or basically, native `Eq` doesn't cut it.

JoeLoser:
> This problem is also present for `Span` and therefore any other "view" types we might have. This could easily turn into a game of Whac-a-Mole if we can't find a better solution to this.

Right, this is true for _any parametric type_.  We often see this with types carrying an origin like the ones you mentioned. I think this will be solvable with parametric traits which is on the compiler roadmap for Q2 still.

That said, the open question is: do we want to do this now or wait for parametric traits?  We all agree in the long term, it doesn't make sense to have specialization in `List[T]` methods for specific parametric `T`s such as `Span`, `StringSlice`, etc. Right now, in the `StringSlice` case, you'd have to convert to a `String` in order to do the comparison or `__contains__` call - is that right?

soraros:
> Right, this is true for _any parametric type_. We often see this with types carrying an origin like the ones you mentioned. I think this will be solvable with parametric traits which is on the compiler roadmap for Q2 still.

It's not even parametric trait, but trait with partially bound/generic associated type.

> That said, the open question is: do we want to do this now or wait for parametric traits? We all agree in the long term, it doesn't make sense to have specialization in `List[T]` methods for specific parametric `T`s such as `Span`, `StringSlice`, etc. Right now, in the `StringSlice` case, you'd have to convert to a `String` in order to do the comparison or `__contains__` call - is that right?

We’ll need to manually write these to make the standard library actually useful, I think. For instance, due to the mutability/immutability bug, `UnsafePointer` doesn’t show this issue yet, but once that's fixed, its `__eq__` will become immediately useless.

It’s actually a bigger problem: the `Eq`-conforming method `__eq__(Self, Self) -> Bool` isn’t the one we’ll end up using, which creates the illusion of trait conformance.

martinvuyk:
> UnsafePointer doesn’t show this issue yet, but once that's fixed, its __eq__ will become immediately useless

@soraros Can confirm firsthand that it's the case, but it's not that bad. It's just a matter of building an `Int` for comparisons

> It’s actually a bigger problem: the Eq-conforming method __eq__(Self, Self) -> Bool isn’t the one we’ll end up using, which creates the illusion of trait conformance.

yep, but again that is solvable using parametric traits + your proposal https://github.com/modular/modular/issues/4346. That is why I put in my code above:

```mojo
    fn __contains__[
        V1: ViewType & Copyable & Movable & Writable,
        V2: V1[.mut=_, .origin=_] & Copyable & Movable & Writable & Comparable[V1[.mut=_, .origin=_]],
    ](self: List[V1, *_], value: V2): ...
```
basically binding V2 to being the same type as V1 with a different mut and origin and that it has to be comparable to any unbound parametrization of V1

> I think this will be solvable with parametric traits which is on the compiler roadmap for Q2 still. [...] That said, the open question is: do we want to do this now or wait for parametric traits?

@JoeLoser I personally would rather wait, otherwise we will be adding a bunch of patch-work everywhere (like the PR where I try to add all the iterator helper methods). Especially if we will be getting parametric traits this year, I would rather wait and solve this the right way at once

bgreni:
+1 to waiting it out for parametric traits

owenhilyard:
> > UnsafePointer doesn’t show this issue yet, but once that's fixed, its **eq** will become immediately useless
> 
> @soraros Can confirm firsthand that it's the case, but it's not that bad. It's just a matter of building an `Int` for comparisons

Whether this is valid depends on whether we're defining `Int` as `ssize_t` or `intptr_t`. Given that pointer tagging has now been encoded into the ARM ISA, we probably want to consider platforms where `SIZE_MAX != UINTPTR_MAX`. Also, it's technically more correct to use `UInt` for pointers, which is why I'm using the unsigned versions. Pointers should, ideally, own their own comparison logic. 

> > It’s actually a bigger problem: the Eq-conforming method **eq**(Self, Self) -> Bool isn’t the one we’ll end up using, which creates the illusion of trait conformance.
> 
> yep, but again that is solvable using parametric traits + your proposal #4346. That is why I put in my code above:

I agree, `__eq__(Self, Self)` needs to go ASAP. It probably makes sense to sit down and review all of the stdlib traits once we have parametric traits, since a lot of them should be parametric. I'd even go so far as to say that having parametric moveinit and copyinit might make sense for cases where you can init a T from a U (although that can be covered with normal ctors). 

> ```mojo
>     fn __contains__[
>         V1: ViewType & Copyable & Movable & Writable,
>         V2: V1[.mut=_, .origin=_] & Copyable & Movable & Writable & Comparable[V1[.mut=_, .origin=_]],
>     ](self: List[V1, *_], value: V2): ...
> ```
> 
> basically binding V2 to being the same type as V1 with a different mut and origin and that it has to be comparable to any unbound parametrization of V1
> 
> > I think this will be solvable with parametric traits which is on the compiler roadmap for Q2 still. [...] That said, the open question is: do we want to do this now or wait for parametric traits?
> 
> @JoeLoser I personally would rather wait, otherwise we will be adding a bunch of patch-work everywhere (like the PR where I try to add all the iterator helper methods). Especially if we will be getting parametric traits this year, I would rather wait and solve this the right way at once

I agree that we should wait. `requires` + parametric traits are two of the big asks from external stdlib contributors to the compiler team, and they will enable a lot of API work.

Now, on a `ViewType` abstraction, I think that also needs to wait for parametric traits, but that having one would be valuable. However, I think that it might make sense to have that trait be something on top of `RandomAccessIterator`.

IbarraSamuel:
@martinvuyk, @soraros We might not need to wait for parametric Traits. If you quick check this:

```mojo

fn main():
    first_string = String("foo")
    _first_slice = first_string.as_string_slice()

    second_string = String("bar")
    _second_slice = second_string.as_string_slice()

    types_are_equal = _type_is_eq[
        __type_of(_first_slice), __type_of(_second_slice)
    ]()

    print(types_are_equal)
```

You will see that using `_type_is_eq`, we can compile time know if it's the value type is the same as the type of the elements in the list, and then do the comparison with a rebind. Should not be a problem since it's just a comparison and it's already how tuple __contains__ work, so this should do the work:
```mojo
    fn __contains__[
        U: EqualityComparable & Copyable & Movable,
        E: EqualityComparable, //,
    ](self: List[U, *_], value: E) -> Bool:
        @parameter
        if _type_is_eq[U, E]():
            value_rebind = rebind[U](value)
            for i in self:
                if i[] == value_rebind:
                    return True
        return False
```

The big problem I can see here is that you don't have "type checked" information, people may be trying to compare if there is an Int within a list of Strings of something like this. So **I don't know if it's a good idea**.

With the `requires` clause, it should do the work perfectly, I guess the LSP will run the requires conditions and tell people to not use incorrect types.

Don't know if we should add Copyable & Movable trait bounds to E, maybe yes, if at the end both types (U and E) should be the same type, but is not really needed for this method (and sometimes, types can have conditional conformance based on some parameters, and a type that could be comparable doesn't require to be Copyable or Movable).

Another option is to eliminate all inneccesary trait bounds, but it requires to do the `_type_is_eq` within the loop. I don't think is a good idea.

```mojo
    fn __contains__[
        E: EqualityComparable, //,
    ](self, value: E) -> Bool:
        for i in self:

            @parameter
            if _type_is_eq[T, E]():
                if value == rebind[E](i[]):
                    return True
        return False
```

bgreni:
> The big problem I can see here is that you don't have "type checked" information, people may be trying to compare if there is an Int within a list of strings of something like this.

Yeah that's a non-starter unfortunately.

soraros:
Default to heterogeneous comparison can be very dangerous, and it's even worse when the language has implicit conversion.

```mojo
a = StringSlice("test")
print(a in ("test", "a", "b"))  # prints False
```

IbarraSamuel:
Totally agree, this one can wait. A for loop is an easy workaround for now

JoeLoser:
Thanks all for the discussion.  We're all definitely eager to play with parametric traits.  In light of this, I'd suggest we close this PR and wait for parametric traits to solve this properly.  Client code can be verbose in the meantime, but that is a fine short-term trade-off.

Issue #3235, opened 2024-07-13
[Feature Request] [LSP] [docstrings] Better Markdown support for arbitrary section headers and correct `<ul>` rendering


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Overall the experience with docstrings is great. Though I think arbitrary headers should be rendered in the same size and bold text as standard ones, and `-` should always be rendered as `<ul>` after all section headers

A classic Python docstring section header that is not in the "standard" Mojo headers:
```mojo
fn _utf8_byte_type(b: SIMD[DType.uint8, _], /) -> __type_of(b):
    """UTF-8 byte type.

    Returns:
        The byte type.

    Notes:
        - 0 -> ASCII byte.
        - 1 -> continuation byte.
        - 2 -> start of 2 byte long sequence.
        - 3 -> start of 3 byte long sequence.
        - 4 -> start of 4 byte long sequence.
    """
```
![image](https://github.com/user-attachments/assets/b294a1f3-b4a6-4ac3-b48e-778a0034c27a)

it does work if you add a newline, but it renders weirdly
```mojo
fn _utf8_byte_type(b: SIMD[DType.uint8, _], /) -> __type_of(b):
    """UTF-8 byte type.

    Returns:
        The byte type.

    Notes:

        - 0 -> ASCII byte.
        - 1 -> continuation byte.
        - 2 -> start of 2 byte long sequence.
        - 3 -> start of 3 byte long sequence.
        - 4 -> start of 4 byte long sequence.
    """
```
![image](https://github.com/user-attachments/assets/464c0f25-3447-4cf8-b306-7b60183825b2)

And something uglier, it renders better but wrong size font
```mojo
fn _utf8_byte_type(b: SIMD[DType.uint8, _], /) -> __type_of(b):
    """UTF-8 byte type.

    Returns:
        The byte type.

    #### Notes:
    ```markdown
        - 0 -> ASCII byte.
        - 1 -> continuation byte.
        - 2 -> start of 2 byte long sequence.
        - 3 -> start of 3 byte long sequence.
        - 4 -> start of 4 byte long sequence.
    ```
    """
```
![image](https://github.com/user-attachments/assets/15c76e17-48f6-44ab-a9a3-a30893ff423b)


And another option that works well but breaks away from the docstring aesthetic
```mojo
[...]

PR #4605, opened 2025-05-17
[stdlib] [NFC] Fix typo in `DLHandle` docstrings

martinvuyk:
Fix typo in `DLHandle` docstrings


mojo/stdlib/stdlib/sys/ffi.mojo
```diff
@@ -178,11 +178,12 @@ struct DLHandle(Copyable, Movable, ExplicitlyCopyable, Boolable):
         """Initialize a dynamic library handle to all global symbols in the
         current process.
 
-        On POXIX-compatible operating systems, this performs
-        `dlopen(nullptr, flags)`.
-
         Args:
             flags: The flags to load the dynamic library.
+
+        Notes:
```
JoeLoser, line 184: @arthurevans @ConnorGray can you remind me on where we stand now with the `Notes` section?  Are we moving away from it in favor of the content being in the body?

ConnorGray: My preference is towards putting this kind of content in the body, but @arthurevans has the final say 🙂

martinvuyk: FWIW I like the notes section best because it forces some mental order. Also, disallowing writing text before specifying the args and parameters (and also the specific ordering of those sections) should IMO be more stringently enforced. I opened issue https://github.com/modular/modular/issues/3235 in july of 2024 because I knew this was only going to get worse because every person has different preferences and I'd rather find some common ground than have the docs be inconsistent and hard to read.

Section headers like `Safety`, `Notes`, and others should be part of our specification and have a logical ordering. Also, the markdown renderer enforces some ordering anyway, all the standard ones get pushed to the top and the text down at the bottom.

arthurevans: Sorry for the delay. When we discussed this (the issue of non-standard headers) in a small group, the majority seemed to be in favor of keeping the list of defined headers small, including just the ones that are linked to the actual signature (Constraints, Parameters, Args, Returns, Raises). Perhaps we should take that discussion up again with a larger group. I'm not opposed to expanding the list, but we'd need to have a really good definition of what goes into each header, or we'll end up with more disorder. 

Notes, in particular I had recommended moving away from because there's no clear definition for when something should be in a Note as opposed to just part of the description. And we've accumulated a number of docs with Notes sections that just repeat information from the description or the constraints, for example. That's really what I'd like to avoid.

This isn't an example of one of these problems, so let's not sweat it.

arthurevans (approved):

arthurevans:
!sync

arthurevans:
!sync

arthurevans:
!sync

PR #4608, opened 2025-05-17
[stdlib] Implement asymmetrical version of `math.isclose`

soraros:
Todo:
- [x] Fix docstring
- [x] Add tests
- [x] Add a change log entry?

### DIFF
--- a/mojo/stdlib/stdlib/math/math.mojo
+++ b/mojo/stdlib/stdlib/math/math.mojo
@@ -1053,10 +1053,12 @@ fn tanh[
 # ===----------------------------------------------------------------------=== #
 
 
-# TODO: control symmetric behavior with flag so we can be compatible with Python
 @always_inline
 fn isclose[
-    dtype: DType, width: Int
+    dtype: DType,
+    width: Int,
+    *,
+    symmetrical: Bool = True,
 ](
     a: SIMD[dtype, width],
     b: SIMD[dtype, width],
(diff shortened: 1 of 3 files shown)


tboerstad:
Looks promising, thank you!

Not for this PR, but would be nice if the default atol and rtol were dtype dependent

soraros:
@tboerstad Actually, should we have `atol` and `rtol` also parametrised by `width` so we could do `isclose(..., atol = SIMD[f64, 2](1e-4, 1e-5))`?

mojo/stdlib/stdlib/math/math.mojo
```diff
@@ -1059,6 +1062,9 @@ fn isclose[
     Parameters:
```
soraros, line 1086: I don't think this is correct. `math.isclose` does use the symmetric formula: https://docs.python.org/3/library/math.html#math.isclose. I think the original author meant "unlike NumPy's `isclose`".

tboerstad:
> @tboerstad Actually, should we have `atol` and `rtol` also parametrised by `width` so we could do `isclose(..., atol = SIMD[f64, 2](1e-4, 1e-5))`?

Not sure, but for the workloads I'm involved in (GenAI) this is not needed. 
Typically I check for closeness of two tensors, and use the same threshold for all values. 

Having sensible default for dtype is something that I would benefit from immediately, as I expect better (lower) tols as the precision of the dtype increases (bfloat16 worse than float32 which is again worse than float64 as an example)

soraros:
> Not sure, but for the workloads I'm involved in (GenAI) this is not needed.

Then let's add lazily.

I will finish up this one, and make another PR to fix the default values.

soraros:
@tboerstad, I think this is ready, PTAL.

tboerstad:
!sync

mojo/stdlib/stdlib/math/math.mojo
```diff
@@ -1046,44 +1048,57 @@ fn isclose[
...
+        atol: Absolute tolerance.
+        rtol: Relative tolerance.
+        equal_nan: If true, treat NaNs as equal (default: false).
 
     Returns:
-        A boolean vector where a and b are equal within the specified tolerance.
+        A boolean vector where `a` and `b` are equal within the given tolerance.
     """
 
-    constrained[
```
tboerstad, line 1093: Was it intentional to remove this constraint? 
I think we should either:
1. Raise an error for integral values, or
2. Support atol/rtol for integral values (preferred). 

If I understand correctly, the new change would provide false for `isclose(int(2),3,atol=10)` which is confusing

mojo/stdlib/test/math/test_math.mojo
```diff
@@ -87,33 +87,99 @@ def test_copysign():
...
+    alias T = SIMD[dtype, 2]
+
+    alias atol = 1e-8
+    alias rtol = 1e-5
+
+    alias inf_ = inf[dtype]()
+    alias nan_ = nan[dtype]()
+    alias v = T(0.1, -0.2)
+
+    fn formula[symm: Bool](a: T, atol: T, rtol: T) -> T:
```
tboerstad: Nit:
```suggestion
    fn create_edge_val[symm: Bool](a: T, atol: T, rtol: T) -> T:
    """Creates a value at the tolerance boundary that should be considered close to `a`."""
```
Could we change the name and/or add a docstring to make it obvious what the formula does?

mojo/stdlib/test/math/test_math.mojo
```diff
@@ -87,33 +87,99 @@ def test_copysign():
...
+        @parameter
+        if symm:
+            return a + min(atol, a * rtol)
```
tboerstad: This looks wrong for negative a?
I haven't quite wrapped my head around it, but something like this might be more correct?
```suggestion
            return a + max(atol, abs(a) * rtol)
```

tboerstad:
Thanks a lot, LGTM barring the integral change

mojo/stdlib/stdlib/math/math.mojo
soraros: I don't think I actually changed any semantics here. The constraint was `dtype is DType.bool or dtype.is_numeric()`, which basically means `dtype is not invalid`. So this change is rather "NFC".

As it stands (even before this patch), `isclose` ignores `atol` and `rtol` and behaves like `__eq__` for boolean and integer types. We *could* tighten the constraint so that it rejects them entirely. What do you think?

tboerstad: Ok, I'm convinced. I can't find any counter examples which would fail the constraint (even DType.index is integral) so agree that the initial removing the constraint is fine. (One could argue to keep it for defensive programming etc, but that sounds premature).

I do support tightening the constraint, and saying something like `isclose is not yet supported for integral and boolean types`. Can you do that?
Then for some future PR, we'll implement similar support for integral as Python does (cast to float iiuc).

I do find Python's support (shown below) for bool to be confusing, but currently we want to behave similar as Python. 

Screenshot 2025-05-30 at 8 58 27 PM

soraros: I think we should tighten the constraints while we still can, namely, now. I'm not even sure we want to support booleans. Python's `math.isclose` accepts `bool` by implicitly casting to `float`, which makes sense in that context, but I see it more as a side effect of Python's weakly typed nature. It's not behaviour I think we should aim to replicate.

tboerstad: Makes sense, let's tighten constraints for this PR. 

I suggest only floating point types for now, since isclose() is currently not implemented correctly for integeral types.

mojo/stdlib/test/math/test_math.mojo
```diff
@@ -87,33 +87,99 @@ def test_copysign():
...
+        (T(inf_, -inf_), T(inf_, -inf_)),
+        (T(1), edge_val[symm](1, atol, 0)),
+        (edge_val[symm](v, atol, 0), v),
+    ]
+
+    @parameter
+    if not symm:
+        all_close += [
+            (edge_val[symm](v, 0, rtol), v),
+            (edge_val[symm](v, atol, rtol), v),
```
soraros, line 126: We can't do this for `symm` because, with floating-point arithmetic, `a - (a - b) == b` doesn't hold.

soraros:
@tboerstad, I removed support for boolean and integer types and fixed the tests. I also [contributed the same implementation](https://github.com/jax-ml/jax/pull/29153) to JAX, so we're effectively using their testing infrastructure to further ensure correctness.

mojo/stdlib/stdlib/math/math.mojo
```diff
@@ -1065,44 +1067,55 @@ fn isclose[
...
-        return a == b
-    var both_nan = isnan(a) & isnan(b)
-    var both_finite = isfinite(a) & isfinite(b)
-    var in_range = abs(a - b) <= max(T(atol), T(rtol) * max(abs(a), abs(b)))
-    return (a == b) | (both_nan & equal_nan) | (both_finite & in_range)
+    if symmetrical:
+        check_fin = isfinite(a) & isfinite(b)
+        in_range = abs(a - b) <= max(T(atol), T(rtol) * max(abs(a), abs(b)))
+    else:
+        check_fin = isfinite(b)
```
soraros, line 1116: This matches NumPy 2.* when `atol` or `rtol` is `inf`.

tboerstad:
!sync

Issue #2860, opened 2024-05-27
[BUG] Implicit conversion from `Boolable` to `Bool` lead to surprising behaviour


soraros:
### Bug description

As title. I think implicit conversion should be used *much* more sparingly.

@lattner @laszlokindrat, since you made the relevant changes.

### Steps to reproduce

```mojo
fn main():
  print(UInt8('test') == 1) # prints True
```

This is because `SIMD` has a constructor from `Bool`, and `String`/`StringLiteral` is `Boolable`. It's easy enough to find out *how* this works, but I find it difficult to justify *why* it should work at all in the first place.

### System information

```shell
N/A
```

laszlokindrat:
I agree that this behavior might be surprising, but I can't see the harm yet. At the same time, implicit conversion to `Bool` and `Int` is often useful and we don't always want to type `Boolable`/`Indexer`. This does make me wonder, though, if we are missing a different abstraction: similarly to how `Int` has `Intable` and `Indexer`, maybe we need two different traits for `Bool` as well?

To mitigate this particular case, we could just add an always-failing constructor from `String`/`StringLiteral` to SIMD.

bgreni:
> I agree that this behavior might be surprising, but I can't see the harm yet.

This could be annoying/cause bugs if I someone passes a string that "looks" like an integer since they might expect it to match the behaviour of `int`.

```mojo
print(Int8('100')) # prints 1
print(int('100')) # prints 100
```

I believe these are the kinds of footguns we should aim to avoid in an nascent language like Mojo.

laszlokindrat:
> > I agree that this behavior might be surprising, but I can't see the harm yet.
> 
> This could be annoying/cause bugs if I someone passes a string that "looks" like an integer since they might expect it to match the behaviour of `int`.
> 
> ```mojo
> print(Int8('100')) # prints 1
> print(int('100')) # prints 100
> ```
> 
> I believe these are the kinds of footguns we should aim to avoid in an nascent language like Mojo.

Again, this can be mitigated by an always-failing String constructor on SIMD and other numeric types. Once we have more powerful traits (e.g. conditional conformance or parametric traits), we could also just prevent non-bool SIMD types from being implicitly constructed from bool. That being said, we are working on improving the implicit conversion story in Mojo, so  there might be functionality we can implement to help control these situations better.

JoeLoser:
[...]

PR #4619, opened 2025-05-17
[stdlib] Minor cleanups to `Int` and `UInt`

soraros:
- ~~Make `UInt` conform to `ImplicitlyBoolable`.~~
- Remove `.copy` from `Int`, as `@register_passable("trivial")` already synthesizes one.
- Tweak the docstring slightly.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/uint.mojo
+++ b/mojo/stdlib/stdlib/builtin/uint.mojo
@@ -33,6 +33,7 @@ struct UInt(
     Absable,
     Boolable,
     CeilDivable,
+    Copyable,
     Movable,
     Comparable,
     ExplicitlyCopyable,
@@ -46,21 +47,19 @@ struct UInt(
 ):
     """This type represents an unsigned integer.
 
-    An unsigned integer represents a positive integral number.
-
     The size of this unsigned integer is platform-dependent.
 
     If you wish to use a fixed size unsigned integer, consider using
     `UInt8`, `UInt16`, `UInt32`, or `UInt64`.
     """
 
-    alias BITWIDTH: Int = bitwidthof[DType.index]()
+    alias BITWIDTH = Int(bitwidthof[DType.index]())
     """The bit width of the integer type."""
 
-    alias MAX: UInt = (1 << Self.BITWIDTH) - 1
+    alias MAX = UInt((1 << Self.BITWIDTH) - 1)
     """Returns the maximum integer value."""
 
-    alias MIN: UInt = 0
+    alias MIN = UInt(0)
     """Returns the minimum value of type."""
 
     var value: __mlir_type.index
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/builtin/uint.mojo
```diff
@@ -687,6 +686,15 @@ struct UInt(
         """
         return self != 0
 
+    @always_inline("builtin")
+    fn __as_bool__(self) -> Bool:
```
JoeLoser: Thoughts on not doing this and removing it from `Int` too?  I don't think we need to mimic Python here for this.

mojo/stdlib/stdlib/builtin/uint.mojo
```diff
@@ -57,10 +56,10 @@ struct UInt(
     alias BITWIDTH: Int = bitwidthof[DType.index]()
```
JoeLoser: **Question** Shouldn't we do the same with `BITWIDTH` then for consistency?

mojo/stdlib/stdlib/builtin/uint.mojo
soraros: I think moving off `ImplicitlyBoolable` is a good idea, but I think those decisions were made in #2860?

mojo/stdlib/stdlib/builtin/uint.mojo
soraros: Done. Thanks. For the record, these changes were made in preparation for #4621.

mojo/stdlib/stdlib/builtin/uint.mojo
JoeLoser: Yeah, but back then, we didn't have @implicit constructor control.  Now that we have that, do we still need `__as_bool__` for `Int` and `UInt`?  Note: I haven't poked at this, but seems fruitful to go look at.

JoeLoser: Maybe punt on this particular change for this PR to unblock exploring https://github.com/modular/modular/issues/4621?

soraros: > Yeah, but back then, we didn't have `@implicit` constructor control. Now that we have that, do we still need `__as_bool__` for `Int` and `UInt`? Note: I haven't poked at this, but seems fruitful to go look at.

That’s a very good point. To be absolutely clear, do you mean that we don't need any implicit conversion from `UInt`/`Int` to `Bool` at all, or simply that we shouldn't implement it using the `ImplicitlyBoolable` trait (instead by adding two constructors to `Bool`)? If we still want this kind of implicit conversion, the trait approach seems like the more general solution.

For this PR, I will remove this change anyways (less implicit is better).

JoeLoser (approved):

JoeLoser:
!sync

PR #4626, opened 2025-05-18
[stdlib] Optimize base 16 and 64 encode and decode

martinvuyk:
Optimize base 16 and 64 encode and decode by using `String` directly and fixing uses of `StringSlice` and parsing where using raw bytes is faster


JoeLoser (approved):
Nice cleanup, thank you! 👍

JoeLoser:
!sync

PR #4628, opened 2025-05-18
[stdlib] `math.factorial()` should raise if overflow

msaelices:
This is the current behavior when we pass n>20 in the `math.factorial()` function:
```
> x = math.factorial(21) 
(Int) x = 0
```

As `factorial(21)` would overflow in Int64, I understand that it's not implemented for >20. But IMO it should raise an error to prevent mistakes in Mojo algorithms.


soraros:
I think this is probably not a change we want right now, since everything else is noexcept in `math`.

msaelices:
> I think this is probably not a change we want right now, since everything else is noexcept in `math`.

Is there any reason for not raising any error in the math module?

I think we should avoid code smells in the code, as they are prone to errors. I don't mind to handle the rest of the cases in `math` library

abduld:
While in general I think this is a good idea, I want to make sure that nothing in the math package raises atm. Is there is a specific use case this is addressing?

msaelices:
> While in general I think this is a good idea, I want to make sure that nothing in the math package raises atm. Is there is a specific use case this is addressing?

Not really, I think it could cause non-alerted bugs. We can close it if you think it's better not raising.

JoeLoser:
> > While in general I think this is a good idea, I want to make sure that nothing in the math package raises atm. Is there is a specific use case this is addressing?
> 
> Not really, I think it could cause non-alerted bugs. We can close it if you think it's better not raising.

Until we have parametric raises, I don't think we should proceed with this right now.  Closing as such.  Thanks all!

PR #4641, opened 2025-05-20
[stdlib] Resolve a stale FIXME in `simd.mojo` by fixing the underlying issue

soraros:
- We can run `FPUtils.exponent_mask` at comptime now.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/simd.mojo
+++ b/mojo/stdlib/stdlib/builtin/simd.mojo
@@ -3593,14 +3593,12 @@ fn _floor(x: SIMD) -> __type_of(x):
     alias bitwidth = bitwidthof[x.dtype]()
     alias exponent_width = FPUtils[x.dtype].exponent_width()
     alias mantissa_width = FPUtils[x.dtype].mantissa_width()
-    # FIXME: GH issue #3613
-    # alias mask = FPUtils[x.type].exponent_mask()
-    alias mask = (1 << exponent_width) - 1
+    alias mask = FPUtils[x.dtype].exponent_mask()
     alias bias = FPUtils[x.dtype].exponent_bias()
     alias shift_factor = bitwidth - exponent_width - 1
 
-    bits = x.to_bits()
-    var e = ((bits >> mantissa_width) & mask) - bias
+    var bits = x.to_bits()
+    var e = ((bits & mask) >> mantissa_width) - bias
     bits = (e < shift_factor).select(
         bits & ~((1 << (shift_factor - e)) - 1),
         bits,


JoeLoser (approved):
LGTM, thanks!

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
FYI this is hitting a few errors in tests internally

patrickdoc:
@soraros , apologies for the issues. It's fixed on main and I bumped your branch, but looks like there are a few real test failures to triage now.

soraros:
@patrickdoc No problem and thanks for the CI fix!

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -3593,14 +3593,12 @@ fn _floor(x: SIMD) -> __type_of(x):
...
-    # alias mask = FPUtils[x.type].exponent_mask()
-    alias mask = (1 << exponent_width) - 1
+    alias mask = FPUtils[x.dtype].exponent_mask()
     alias bias = FPUtils[x.dtype].exponent_bias()
     alias shift_factor = bitwidth - exponent_width - 1
 
-    bits = x.to_bits()
-    var e = ((bits >> mantissa_width) & mask) - bias
+    var bits = x.to_bits()
+    var e = ((bits & mask) >> mantissa_width) - bias
```
soraros, line 3601:
```mojo
alias old_mask = (1 << exponent_width) - 1 
alias new_mask = FPUtils[x.dtype].exponent_mask()
old_mask == new_mask >> mantissa_width
```

JoeLoser: Given that this only failed in internal builds, can you please check test coverage here so without the fix, it would fail said test in the OSS build?

soraros: Without this fix, OSS did fail: https://github.com/modular/modular/pull/4641#issuecomment-2895393458.

The issue was that the commented-out code isn't equivalent to our current implementation, and I submitted the patch without running the tests locally, assuming they were the same. Of course, we could simply remove the FIXME comment and retain the old implementation. Which option do you prefer?

JoeLoser: Ah, I see - thanks for the info.  New implementation looks good.  Did you happen to compare the IR by chance?

JoeLoser:
!sync

mojo/stdlib/stdlib/builtin/simd.mojo
soraros: FYI, given that `mask`, `old_mask`, and `mantissa_width` are all comptime constants, the expressions `(bits >> mantissa_width) & old_mask` and `(bits & mask) >> mantissa_width` are only one single constant fold apart and indeed generate identical optimised LLVM IR.

PR #4645, opened 2025-05-20
[mojo-stdlib] Speed up atof by first checking that the string is ASCII

gabrieldemarmiesse:
We get a ~4x speedup on the atof function with this.

I wonder if we can can know somehow that a string/stringslice is ascii faster by storing it on the struct. It's a common fast path that allows a bunch of important optimization.

@martinvuyk  it seems you're the new strings expert, let me know what you think :) 

Seems related to:
*  https://github.com/modular/modular/pull/3859

Ignore the temporary @no_inline, they're here to make `perf` output useful stuff


martinvuyk:
Hi @gabrieldemarmiesse yes I very much am preparing to enable this kind of optimization with #3859, and a conditional vectorized apply for span (PR  #3877). I wrote a sample in this [forum post](https://forum.modular.com/t/how-to-uppercase-and-lowercase-ascii-strings-using-simd-in-mojo/301)

My hope is that  #3988 gets approved and we can do this automatically when the `StringSlice` is parametrized as being ASCII only:
```mojo
fn _atof(x: StringSlice[encoding=Encoding.ASCII]): ...
```
and every other function that receives a generic `StringSlice` can also reap the benefits from the specification that it is only ASCII (trickles down to `.removeprefix()`, etc. automatically)

martinvuyk:
> We get a ~4x speedup on the atof function with this.

If that is the case, then I wonder if it isn't best to get a mutable `StringSlice` and have a `lower(data: StringSlice[mut=True])` do the mutation in-place. Because if most of the time is spent doing stuff to strings, then removing one alloc should be pretty powerful (when passed a mutable `StringSlice`, we could add an overload that does the allocation for the user or force them to do the copy)

I'm not sure if mutating in-place for lowering is achievable for unicode (would depend if all character uppercase and lowercase codepoints are the same byte width in utf-8), otherwise this would be yet another optimization for ASCII only

But, It can also be that most of the strings passed to `atof` fall under the SSO 24 byte limit, and as such I'm not sure if the allocation is all that important.

martinvuyk (commented):
As for the PR itself I would recommend keeping all the methods as private helpers inside `parsing_floats.mojo` for now since there are many things still changing in `String` and `StringSlice`. And taking a look at the [forum post](https://forum.modular.com/t/how-to-uppercase-and-lowercase-ascii-strings-using-simd-in-mojo/301) I mentioned earlier and adapting it for this.

PS: I'm still reading the atof code bit by bit, great work :rocket:

laszlokindrat:
@gabrieldemarmiesse This looks exciting, do you plan to finish it up?

gabrieldemarmiesse:
Thanks I'll look at it again. Many changes landed in String since then, I'll see how it changes the performance

gabrieldemarmiesse:
Closing this PR as I'm not going to have the time to work on it

PR #4650, opened 2025-05-21
[stdlib] Fix LinkedList docstrings: update time complexity and correct "head" to "tail"

shogo314:
This PR improves the accuracy of the LinkedList documentation by making two key fixes:

1. Time Complexity Update
   - Changed several member function docstrings that incorrectly stated O(1) time complexity to O(n),
     reflecting the actual traversal cost in these methods.

2. Correcting Terminology
   - Fixed the maybe_pop function docstring to correctly refer to the "tail" of the list instead of the "head".

Please review and provide feedback on the updated complexity notation;  
future improvements might specify it as O(min(i, n - i)) if agreed.

Note that the current time complexity notation "O(n) in len(self)" was kept to match the existing style,  
but I personally think that "O(n), where n = len(self)" or simply "O(len(self))" would be clearer and more accurate.  
I hope we can discuss and possibly improve this notation in the future.

### DIFF
--- a/mojo/stdlib/stdlib/collections/linked_list.mojo
+++ b/mojo/stdlib/stdlib/collections/linked_list.mojo
@@ -373,10 +373,10 @@ struct LinkedList[
         raise String("Invalid index for pop: {}").format(Int(i))
 
     fn maybe_pop(mut self) -> Optional[ElementType]:
-        """Removes the head of the list and returns it, if it exists.
+        """Removes the tail of the list and returns it, if it exists.
 
         Returns:
-            The head of the list, if it was present.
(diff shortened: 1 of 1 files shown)


JoeLoser:
@tboerstad this would be a great first PR to take a look at as you're learning the external contributor processes if you're interested 👍

soraros:
I think you could simply also apply the improvements regarding the use of Big-O notations.

tboerstad:
Thanks for the fixes @shogo314.
Agree with matching the existing style O(n) style, can potentially fix/discuss in another PR/issue.

tboerstad:
!sync

PR #4653, opened 2025-05-21
[stdlib] Optimize unicode decoding

martinvuyk:
Add branchy unicode decoding and make it the default. We could make it not be the default, but I think that most branch predictors should make it be actually more performant than the branchless one.

## Benchmark results

CPU: Intel® Core™ i7-7700HQ

Interpretation: English text gets an average 15% improvement but the cost is paid by shorter strings being slower for multi-byte languages due to branch prediction misses, but that cost is amortized and is less than 1% at 100 bytes and from 10k bytes onward is actually sped up up to 20% at 1M bytes. So most probably for those languages setting `optimize_ascii=False` will make all the difference to avoid the `likely()` intrinsic being used and letting the branch predictor do its job.

### For english text

|-| old value (ms) | new value (ms) | percentage | magnitude|
|-- | -- | -- | -- | --|
|bench_write_utf8[10] | 6.66 | 6.32 | 0.05 | 1.05|
|bench_write_utf8[30] | 7.67 | 6.16 | 0.20 | 1.24|
|bench_write_utf8[50] | 4.45 | 4.22 | 0.05 | 1.06|
|bench_write_utf8[100] | 4.64 | 4.37 | 0.06 | 1.06|
|bench_write_utf8[1000] | 9.52 | 6.10 | 0.36 | 1.56|
|bench_write_utf8[10000] | 5.19 | 3.72 | 0.28 | 1.40|
|bench_write_utf8[100000] | 7.20 | 6.22 | 0.14 | 1.16|
|bench_write_utf8[1000000] | 4.12 | 3.79 | 0.08 | 1.09|
|average improvement |-|-| 0.15 | 1.20|

### For the average of english, spanish, arabic, russian, and mandarin

|- | old value (ms) | new value (ms) | percentage | magnitude|
|-- | -- | -- | -- | --| 
|bench_write_utf8[10] | 2.93 | 3.61 | -0.23 | 0.81|
|bench_write_utf8[30] | 2.75 | 3.62 | -0.32 | 0.76|
|bench_write_utf8[50] | 2.75 | 3.22 | -0.17 | 0.85|
|bench_write_utf8[100] | 2.68 | 2.70 | -0.01 | 0.99|
|bench_write_utf8[1000] | 2.63 | 2.64 | 0.00 | 1.00|
|bench_write_utf8[10000] | 2.66 | 2.66 | 0.00 | 1.00|
|bench_write_utf8[100000] | 3.30 | 2.75 | 0.17 | 1.20|
|bench_write_utf8[1000000] | 5.34 | 4.18 | 0.22 | 1.28|
|average improvement |-|-| -0.04 | 0.99


bgreni:
We should probably do some benchmarking on this before potentially needlessly increasing the testing surface area of this function

martinvuyk:
@bgreni @soraros having a hunch and then proving it with benchmarks feels good :tada:. To me the sacrifice of making short multi-byte language strings slower is worth it, WDYT?

abduld:
!sync

abduld:
!sync

abduld:
!sync

PR #4655, opened 2025-05-21
[DO NOT MERGE] Trying to understand the need for FLAG_IS_STATIC_CONSTANT

gabrieldemarmiesse:
Signed-off-by: gabrieldemarmiesse <gabrieldemarmiesse@gmail.com>

Do not merge, I'm just trying things out.

I summon @martinvuyk to understand why this change is not valid!


mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -335,11 +321,7 @@ struct String(
     @always_inline("nodebug")
     fn __init__(out self):
         """Construct an empty string."""
-        self._ptr_or_data = UnsafePointer[UInt8]()
-        self._len_or_data = 0
-        # Note: we treat a nul pointer as a "static constant" pointer because
-        # in the out-of-line form, it has no ARC header.
-        self._capacity_or_data = _StringCapacityField(static_const_length=0)
+        self = Self(capacity=0)
```
martinvuyk, line 326: I think you should leave this as-is because otherwise you force an allocation for the `_StringOutOfLineHeader`
```mojo
struct String:
    fn __init__(out self, *, capacity: Int):
        var cap_field = _StringCapacityField(capacity)
        if cap_field.is_inline(): ...
        else:
            self._len_or_data = 0
            self._ptr_or_data = _StringOutOfLineHeader.alloc(
                cap_field.get_capacity()
            )
            self._capacity_or_data = cap_field

struct _StringOutOfLineHeader:
    fn alloc(capacity: Int) -> UnsafePointer[UInt8]:
        """Allocate space for a new out-of-line string buffer."""
        var ptr = UnsafePointer[UInt8].alloc(capacity + Self._self_size)

        # Initialize the header.
        __get_address_as_uninit_lvalue(
            ptr.bitcast[_StringOutOfLineHeader]().address
        ) = _StringOutOfLineHeader()

        # Return a pointer to right after the header, which is where the string
        # data will be stored.
        return ptr + sizeof[Self]()
```
And your `_is_static_constant` method will already handle the null pointer

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -129,10 +129,8 @@ struct _StringCapacityField:
     # When FLAG_HAS_NUL_TERMINATOR is set, the byte past the end of the string
     # is known to be an accessible 'nul' terminator.
     alias FLAG_HAS_NUL_TERMINATOR = UInt(1) << (UInt.BITWIDTH - 3)
-    # When FLAG_IS_STATIC_CONSTANT is set, the _data pointer is to static
-    # constant string.  capacity() will always be zero for these strings, so any
-    # attempt to append will reallocate.
-    alias FLAG_IS_STATIC_CONSTANT = UInt(1) << (UInt.BITWIDTH - 2)
+    # UNUSED_FLAG is not used, but might be used in the future.
+    alias UNUSED_FLAG = UInt(1) << (UInt.BITWIDTH - 2)
```
martinvuyk, line 133: We could just delete the flag, but add the documentation about the capacity being zero to force realloc for static strings to the struct docstrings

martinvuyk (commented):
I am summoned :laughing: 

NICE! this will totally work :rocket:

mojo/stdlib/stdlib/collections/string/string.mojo
gabrieldemarmiesse: Unless i'm mistaken, `self = Self(capacity=0)` won't force an allocation of _StringOutOfLineHeader since we're going in the if branch `if cap_field.is_inline():`. Maybe i'm misunderstanding how it works? Isn't a 0 capacity supposed to have `is_inline()` being `True`?

mojo/stdlib/stdlib/collections/string/string.mojo
gabrieldemarmiesse: This bit is the only piece of data not used, let's be specific about it. I'll add a comment to the `__init__` of static string

gabrieldemarmiesse:
Thanks @martinvuyk  for the review :)

I'd like to use this bit to add a hint about the ASCIIness of the string. Similar to the flag hinting about the string being zero-terminated.
* 1 means "we know the string is ASCII"
* 0 means "We have no idea if the string is ASCII or not, read it if you want to know"
It would make optimizations for ASCII happen more often. We could transfer this flag to `StringSlice` by using pointer tagging.

A motivating real-world use case would be https://github.com/modular/modular/pull/4645
It would make many `String` methods orders of magnitude faster in general, since ASCII strings are so common in the wild and so easy to process with SIMD.
Checking UTF8 is a common first step to create a `String` and a good place to check ASCIIness of the string since we fully read it.

@lattner is this something you would be interested in? I'm asking because you worked heavily on the String struct and there might be some context I don't have.

mojo/stdlib/stdlib/collections/string/string.mojo
soraros: I think this fix is good. `String(cap=0)` will create an inline string with capacity 23, so it append won't automatically trigger alloc. It even cuts down a `llvm.memset`.

martinvuyk: ooh yeah you are right sorry, my head was going in the out-of-line direction

martinvuyk: ohh yeah you are right, sorry my head was going down the out-of-line path for everything

martinvuyk:
> Checking UTF8 is a common first step to create a String and a good place to check ASCIIness of the string since we fully read it.

I like the idea, but I am not sure if it is for the best to reimplement memcpy while checking each byte, might as well also count unicode length etc. Also, I'm working on adding utf8 validation but only in debug not at runtime. Iterating over string data every time it is getting built will kill performance.

But I seriously think  #3988 is the more elegant and generic approach, the constructor for `StringSlice[Encoding.ASCII].__init__(StringSlice[Encoding.UTF8])` could raise when the string is not ASCII. And this same mechanic can apply for other encoding types where applicable. I would rather not hack together bit setting, especially given any `StringSlice[mut=True]` could overwrite any data in the string. We should **not** be mixing together several encodings into the type at runtime IMHO.

I would seriously prefer going through the pain of creating a separate `ASCIIString` type which I'd really rather not. IMO parametrizing `StringSlice` is the best way because it is scalable for many other encodings and the optimizations trickle down for all.

gabrieldemarmiesse:
Since Chris manifested the desire to remove the flag, I'll make a new clean PR to make reviewing easier. Closing in favor of 
* #4661

PR #4657, opened 2025-05-21
[CI] Fix skipping autotune tests in Bazel

JoeLoser:
https://github.com/modular/modular/commit/997204763fd77afc64eac0d82e716a3766711a28 broken the open source Bazel build due to directory moves.  Adjust the Bazel skip filters to account for the new directory location of the `autotune_tests`.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -47,4 +47,4 @@ jobs:
 
       - name: Build and Test
         run: |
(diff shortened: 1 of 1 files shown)


JoeLoser:
!sync

JoeLoser:
As discussed offline, we need to figure out how to best avoid drift with the `requirements.txt` we use internally in Bazel builds vs externally. We kept it simple intentionally to start, but need to handle `pyyaml` now.

PR #4665, opened 2025-05-22
[stdlib] [NFC] Fix details in `Optional` docstrings

martinvuyk:
Fix details in `Optional` docstrings and use `.unsafe_value()` in a place where a check was already made before


JoeLoser (approved):

JoeLoser:
!sync

PR #4666, opened 2025-05-22
[stdlib] Change default from `Representable` to `Writable` for several collections

martinvuyk:
Change default from `Representable` to `Writable` for several collections


martinvuyk:
nvm this needs a bigger refactor for the `write_to` method to pass down that the format spec is for `repr` and not `str`. I opened proposal #4667

Issue #3630, opened 2024-10-09
[Feature Request] [stdlib] [sys] [instrinsics] Equivalent of `fn _type_is_eq[t1: AnyType, t2: AnyType]() -> Bool` but for checking conformance to a trait `fn _implements[t1: AnyType, t2: AnyTrait]() -> Bool` + `rebind_trait[t1: AnyType, t2: AnyTrait]()`


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Equivalent of `fn _type_is_eq[t1: AnyType, t2: AnyType]() -> Bool` but for checking conformance to a trait `fn _implements[t1: AnyType, t2: AnyTrait]() -> Bool` + `rebind_trait[t1: AnyType, t2: AnyTrait]()`. 

Related (it is one alternative solution) to issue #3403

Also, I'd like to ask for both `_type_is_eq` and `_implements` to become public APIs, since they will be used somewhat often for generic programming outside the stdlib.

### What is your motivation for this change?

Dynamically/parametrically giving different code paths according to the trait that a type implements. Many functions in Python allow arguments that raise at runtime if not compatible with the given types, and such functions are many times impossible to replicate without the dynamism of `AnyType` and progressively constraining or straight up throwing the type system away and using `PythonObject`

martinvuyk:
Officially supported 🎉 https://github.com/modular/modular/commit/6400edb2080e569a55763d14b4cc34fe0efa224b

PR #4669, opened 2025-05-22
[stdlib] [NFC] Remove `Optional` imports

martinvuyk:
Remove `Optional` imports. It is already imported in the prelude

### DIFF
--- a/mojo/stdlib/stdlib/algorithm/reduction.mojo
+++ b/mojo/stdlib/stdlib/algorithm/reduction.mojo
@@ -19,7 +19,6 @@ from algorithm import map_reduce
 ```
 """
 
-from collections import OptionalReg
 from collections.string import StaticString
 from math import align_down, ceildiv, iota
 from os import abort
--- a/mojo/stdlib/test/collections/test_optional.mojo
+++ b/mojo/stdlib/test/collections/test_optional.mojo
@@ -12,7 +12,7 @@
 # ===----------------------------------------------------------------------=== #
 # RUN: %mojo %s
 
-from collections import Optional, OptionalReg
+from collections import OptionalReg
 
 from testing import *
 
--- a/mojo/stdlib/stdlib/builtin/_pybind.mojo
+++ b/mojo/stdlib/stdlib/builtin/_pybind.mojo
@@ -11,7 +11,7 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 
-from collections import Optional
+
 from sys import alignof, sizeof
 
 import python._cpython as cp
--- a/mojo/stdlib/stdlib/builtin/builtin_slice.mojo
+++ b/mojo/stdlib/stdlib/builtin/builtin_slice.mojo
@@ -15,8 +15,6 @@
 These are Mojo built-ins, so you don't need to import them.
 """
 
-from collections import Optional
-
 
 struct Slice(
     Stringable,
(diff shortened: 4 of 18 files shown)


mojo/stdlib/stdlib/algorithm/reduction.mojo
```diff
@@ -19,7 +19,6 @@ from algorithm import map_reduce
 ```
 """
 
-from collections import OptionalReg
```
JoeLoser, line 22: **Suggestion** Please revert. `OptionalReg` is not in the prelude which is intentional.

mojo/stdlib/stdlib/collections/__init__.mojo
```diff
@@ -19,5 +19,5 @@ from .dict import Dict, KeyElement, OwnedKwargsDict
 from .inline_array import InlineArray
 from .linked_list import LinkedList
 from .list import List
-from .optional import Optional, OptionalReg
+from .optional import Optional
```
JoeLoser: **Suggestion** Please revert this. It's still desirable to export `OptionalReg` here, but it is not exported in the `prelude`, for contrast.

mojo/stdlib/test/collections/test_optional.mojo
```diff
@@ -12,7 +12,7 @@
 # ===----------------------------------------------------------------------=== #
 # RUN: %mojo %s
 
-from collections import Optional, OptionalReg
+from collections.optional import Optional, OptionalReg
```
JoeLoser: **Question** Why make this change?

mojo/stdlib/stdlib/algorithm/reduction.mojo
martinvuyk: It is not used in the file

mojo/stdlib/test/collections/test_optional.mojo
martinvuyk: I removed `OptionalReg` from the `collections.__init__` since I thought it is only intended for some low-level use-cases

mojo/stdlib/stdlib/collections/__init__.mojo
martinvuyk: I thought the intention is to deprecate the type or that it is only necessary for some very low-level optimizations

JoeLoser: It will take a fair bit of time before we can do that though: it's used (and needed) in the kernel library since we don't have a way to optimize `Optional` for register passable only types.  In either case, it's on different timelines for removal, so please don't change behavior here.

mojo/stdlib/test/collections/test_optional.mojo
JoeLoser: As mentioned at https://github.com/modular/modular/pull/4669#discussion_r2103383400, let's revert that.  It's why the build is failing too for Bazel.

martinvuyk:
@JoeLoser reverted. I wonder if having #3581 or an `_implements[AnyTrivialRegType](self) -> Bool` (#3630) would solve this by letting us specialize the implementation

JoeLoser:
> @JoeLoser reverted. I wonder if having #3581 or an `_implements[AnyTrivialRegType](self) -> Bool` (#3630) would solve this by letting us specialize the implementation

Unfortunately it won't help the `Optional` case and we would need even more language things to get rid of `OptionalReg`.

JoeLoser (approved):

JoeLoser:
!sync

PR #4670, opened 2025-05-22
[stdlib] [NFC] Remove `List` imports

martinvuyk:
Remove `List` imports. It is already imported in the prelude


JoeLoser (approved):
LGTM but please expand on the PR description mentioning how `List` is already exported in the Prelude, for example.  This will help give contexts on the "why" for a change and how this is a NFC.

JoeLoser:
!sync

PR #4672, opened 2025-05-22
[stdlib] [NFC] Remove `Dict` imports

martinvuyk:
Remove `Dict` imports. It is already imported in the prelude


JoeLoser (approved):
LGTM but please expand on the PR description mentioning how these are already exported in the `prelude`, for example. This will help give contexts on the "why" for a change and how this is a NFC.

JoeLoser:
!sync

PR #4683, opened 2025-05-24
[stdlib] Optimize `.splitlines()` using new `splat` bit mask util

martinvuyk:
Optimize `.splitlines()` using new `splat` bit mask util

### Benchmark results

CPU: Intel® Core™ i7-7700HQ

|  | old value (ms) | new value (ms) | percentage | magnitude
|-- | --:| --:| --:| --:|
|bench_string_splitlines[10] | 9.93 | 8.76 | 0.12 | 1.13|
|bench_string_splitlines[30] | 6.46 | 5.00 | 0.23 | 1.29|
|bench_string_splitlines[50] | 4.91 | 4.60 | 0.06 | 1.07|
|bench_string_splitlines[100] | 4.49 | 3.85 | 0.14 | 1.17|
|bench_string_splitlines[1000] | 3.98 | 3.57 | 0.10 | 1.12|
|bench_string_splitlines[10000] | 3.67 | 3.36 | 0.08 | 1.09|
|bench_string_splitlines[100000] | 3.67 | 3.34 | 0.09 | 1.10|
|bench_string_splitlines[1000000] | 3.67 | 3.42 | 0.07 | 1.07|
|average improvement |   |   | 0.11 | 1.13|

### DIFF
--- a/mojo/stdlib/stdlib/bit/_mask.mojo
+++ b/mojo/stdlib/stdlib/bit/_mask.mojo
@@ -82,3 +82,17 @@ fn splat[
         otherwise.
     """
     return (-(value.cast[DType.int8]())).cast[dtype]()
+
+
+@always_inline
+fn splat(value: Bool) -> Int:
+    """Get a bitmask of whether the value is `True`.
+
+    Args:
+        value: The value to check.
+
+    Returns:
+        A bitmask filled with `1` if the value is `True`, filled with `0`
+        otherwise.
+    """
+    return Int(splat[DType.index](Scalar[DType.bool](value)))
--- a/mojo/stdlib/stdlib/collections/string/string_slice.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string_slice.mojo
@@ -70,6 +70,7 @@ from sys.ffi import c_char
 from sys.intrinsics import likely, unlikely
 
 from bit import count_trailing_zeros
+from bit._mask import splat, is_negative
 from memory import Span, memcmp, memcpy, pack_bits
 from memory.memory import _memcmp_impl_unconstrained
 from python import Python, ConvertibleToPython, PythonObject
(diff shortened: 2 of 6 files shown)


mojo/stdlib/stdlib/bit/_mask.mojo
```diff
@@ -77,6 +77,20 @@ fn is_true[
     return (-(value.cast[DType.int8]())).cast[dtype]()
 
 
+@always_inline
+fn is_true(value: Bool) -> Int:
```
soraros: It's safer to use an unsigned type for the bitmask, or you risk signed extension when casting to a larger type (casting an `Int` to `Int64` adds leading ones; [example](https://github.com/modular/modular/pull/4816#discussion_r2157261515)).

martinvuyk: Yes but that is the whole point of this helper, to be a `-1` mask when true and a `0` when false. It is made to be used for a bitwise `&` operation. I added the `Int` overload mostly for convenience

soraros: ~We could still return `UInt(Int(-1))` right?~

Actually, if we want to return all ones, signed extension has the correct semantics. Please add a note for this fact.

martinvuyk: yes, but the problem with that would be when then using the bitmask with e.g. a `UInt8` I'm not sure how the overflow from `UInt(Int(-1))` would behave. Whereas AFAIK `Int8(Int(-1))` is still `-1` -> `UInt8(255)`

laszlokindrat: Not directly related to this PR, but I find these names really confusing. `is_true` is not an expressive function name to begin with, and here it has some pretty special semantics. IIUC this is basically a "bitsplat"?

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2000,39 +1996,60 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
 
-            var str_len = eol_start - offset + Int(keepends) * eol_length
-            var s = StringSlice[O](ptr=ptr + offset, length=str_len)
-            output.append(s)
-            offset = eol_start + eol_length
+        @always_inline
+        @parameter
+        fn _splitlines[keep: Bool]():
+            while line_start < length:
+                var line_end = line_start
```
laszlokindrat, line 2033: Can you please add some comments, and/or an overall explanation of this algorithm?

mojo/stdlib/stdlib/bit/_mask.mojo
martinvuyk: > is_true is not an expressive function name

The original implementation before making them standalone functions (see  #3886) was `BitMask.is_true(...)` which is a bit more expressive than standalone functions now that you point it out. The main point is that it returns a bitmask of 1 when true, 0 otherwise

martinvuyk: I renamed the overload in line with 7ed18e324e8b98fccd08009646a4377dbcc16cd2

laszlokindrat (approved):
Very nice, thank you for working on this!

laszlokindrat:
!sync

PR #4700, opened 2025-05-26
[stdlib] Allow `Dict.__getitem__` to return mutable references

### LINKED ISSUE #4695: [BUG] Dict.__getitem__ always returns immutable references
### Actual behavior
When accessing elements of a dict, immutable references are returned regardless of whether the dict itself is mutable or not.

### Expected behavior
If the dictionary is mutable, it should provide mutable references to its values.

### Steps to reproduce

Consider the following code

```mojo
d = {1: 1}
ptr = Pointer(to=d[1])
ptr[] = 2 # <-- does not work; "error: expression must be mutable in assignment"
```

### System information

Mojo 25.4.0.dev2025052405 (ec62f670)

samufi:
Mutable references are necessary if a mutable pointer into the  dict needs to be created.

This fixes #4695.

Signed-off-by: Samuel M. Fischer <samufi@web.de>

### DIFF
--- a/mojo/stdlib/stdlib/collections/dict.mojo
+++ b/mojo/stdlib/stdlib/collections/dict.mojo
@@ -598,7 +598,7 @@ struct Dict[K: KeyElement, V: Copyable & Movable](
     # ===-------------------------------------------------------------------===#
 
     fn __getitem__(
-        self, key: K
+        ref self, key: K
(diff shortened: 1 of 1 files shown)


samufi:
This PR is blocked by #4705. If that issue is fixed, this PR might be completely out of date. I therefore close it for now.

PR #4715, opened 2025-05-29
[stdlib] Included intrincics for tanh

simveit:
According to [PTX isa doc](https://docs.nvidia.com/cuda/parallel-thread-execution/#half-precision-floating-point-instructions-tanh) implemented half floating point instructions for `tanh` following `exp2` implementation


abduld:
nice

abduld:
!sync

PR #4727, opened 2025-06-01
[CI] Convert CI to use pixi

Ahajha:
The main build workflow is going away soon, but to aid in the `magic` decommission let's move the workflow to use `pixi` instead.


Ahajha:
lol i work here.

PR #4728, opened 2025-06-01
[stdlib] [proposal] Binary IO API

owenhilyard:
This proposal covers binary IO via an "IO Engine" concept, designed to be highly modular and composable while allowing a mix of state of the art/near future IO APIs with older, more familiar APIs. Although this proposal does provide example APIs, an additional aspect up for discussion is whether this design philosophy is desirable. The primary goal is to allow Mojo to adopt new IO APIs as they are developed, instead of providing a central abstraction which Mojo is chained to (see C++ iostreams), while still allowing abstraction over commonalities. This is designed to function even in the cases where some capabilities are absent, such as GPUs often lacking access to a full network stack. It does somewhat increase the implementation burden of the standard library, but does so for what I think are justifiable reasons.

### DIFF
--- a/mojo/proposals/binary-io-api.md
+++ b/mojo/proposals/binary-io-api.md
@@ -1,0 +1,1067 @@
+# Binary IO API
+
+Author: Owen Hilyard
+
+Date: May 25th, 2025
+
+Status: Proposal
+
+## Scope
+
+This proposal addresses binary IO to both network sockets and files.
+
+### Goals
+
+#### Design for Performance First
+
+Mojo already makes compute fast, so let's also make sure that we can keep that
+compute fed. When possible, we should prefer offering more performance, since it
+should always be possible to offer APIs which offer more ergonomics at the cost
+of performance, but an ergonomics-first design risks making decisions that
+render the API unusable to users who require performance. This includes
+considering and making accommodations for state of the art (SOTA) APIs which may
+not be suitable for all use-cases.
+
+#### Let the user have opinions on how IO is done (`sync`/`async` agnostic where possible)
+
+Some users may have security concerns about some APIs (`io_uring`) or not be
+willing to make tradeoffs as aggressively as some APIs would like (DPDK/SPDK).
+Others might just want to read in a config file and can't do much else until
+that happens, so just do sync IO. The user might also want to bring their own
+APIs, for instance for accessing NVMe drives or the network from a GPU or other
+accelerator. We should strive to accommodate all of these use cases, and not
+pick one as the default. In the Rust ecosystem, there have been problems with
+`sync` and `async`, where users complain that they are forced into async IO
+while still learning the basics of the language. I think that we should strive
+to make it easy for users to choose the right API for their needs, rather than
+trying to guess what they want. When possible, we should try to provide tools to
+library authors to make their libraries as flexible as possible.
+
+#### Make it extensible
+
+Nobody can predict the future. I, the author, have done my best here, but we're
+seeing IO modes which were inconceivable a decade or two ago. Users must be able
+to bring their own IO APIs, and device vendors need to be able to expose
+device-specific APIs. The APIs should make many small promises, ensuring that if
+two pieces of functionality are bundled at the lowest level, it is because they
+cannot exist without each-other, no matter how odd them not existing together
+may seem. Mojo's trait system provides the ability to express these constraints
+in a logical way, and to help us avoid breaking changes down the road. Not all
+files may be possible to close, not all sockets support both send and receive,
+so we break things up. If the user has a new type of IO, they can break it up
+and put it in a library, or offer it to the standard library.
+
+#### Expose alternatives to users
+
+Linux does, in fact, have network protocols other than UDP and TCP. There are
+also categories for reliably-delivered messages, reliably delivered messages
+with a fixed max length (with more perf in exchange), and a way to add
+congestion control to UDP. There's even
+[TIPC](https://docs.kernel.org/networking/tipc.html), which provides a
+substantial amount of the cluster communication services that Kubernetes does,
+including service discovery, failure detection, pub/sub, globally broadcast
+messages, and much more. Normally, users would reach for libraries for these
+things which may be provided as OS-level services. Showing users that these
+options exist undeniably increases complexity, but it also allows users to
+consider what they really need, instead of reaching for the same thing they
+always do.
+
+### Non-Goals
+
+#### Familiar APIs
+
+As will be discussed further on, the classical POSIX APIs have some fairly
+substantial issues with them, in part due to needing copies, and in part due to
+being synchronous. IO is increasingly moving in the direction of the "IO Engine"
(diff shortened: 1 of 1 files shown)


owenhilyard:

mojo/proposals/binary-io-api.md
```diff
@@ -0,0 +1,1067 @@
...
+on what APIs are useful.
+
+### IO APIs
+
+#### CXL
+
+Compute Express Link (CXL) is a new standard for devices with cache, memory and
+compute, which allows them to, for lack of a better description, join in on CPU
+cache coherence protocols. This works by switching the PCIe link into a new mode
+after some negoiation, making it lower latency and capable of carrying cache
```
msaelices: NIT:
```suggestion
after some negotiation, making it lower latency and capable of carrying cache
```

mojo/proposals/binary-io-api.md
```diff
@@ -0,0 +1,1067 @@
...
+operations over to IoRing.
+
+#### POSIX IO APIs (macOS, BSD and Linux)
+
+These are the basic APIs we all know. `read(2)`, `write(2)`, `open(2)`, etc for
+disk IO, and `socket(2)`, `send(2)` and `recv(2)` for networking. They are all
+synchronous APIs, and for a long time they have shaped the *nix community's
+vision of what an async API should look like. As we have discovered, these APIs
+are not ideal for modern hardware, and we should try to avoid them where
+possible. All of the them force a doubling of the memory bandwidth consumption
```
msaelices: NIT:
```suggestion
possible. All of them force a doubling of the memory bandwidth consumption
```

mojo/proposals/binary-io-api.md
```diff
@@ -0,0 +1,1067 @@
...
+    list, a slab list, a tree, or some exotic data structure.
+    """
+    pass
+```
+
+In order to handle a variety of IO APIs, the first thing we need to abandon is
+the idea that buffers are always contiguous chunks of bytes with arbitrary
+alignment. This base trait acts as a marker trait, and is intended to be used
+primarily to help make error messages better. For example, "Given `$TYPE`, which
+implements `IoBuffer`, but is not a `ContiguiousIoBuffer` as required by this IO
```
msaelices: NIT:
```suggestion
implements `IoBuffer`, but is not a `ContiguousIoBuffer` as required by this IO
```

owenhilyard:
@msaelices Thanks for the nits, I gave it an editing pass with fresh eyes.

mojo/proposals/binary-io-api.md
```diff
@@ -0,0 +1,1067 @@
...
+sensible fallback for `io_uring`. Unlike most of the other IO APIs this is a
+readiness-based IO API. This means that the kernel will not do the IO for you,
+and you still have to issue syscalls, it will only do its best to tell you when
+the IO is non-blocking. Epoll also has poor support for async disk IO, meaning
+that a fallback API using epoll would likely need a threadpool to hand disk IO
+off to, which is what both Rust's Tokio, Golang, and libuv currently do.
+
+#### `kqueue` (macOS/BSD)
+
+`kqueue` is the `io_uring` equivalent on BSD systems. It has some disadvantages,
```
toffaletti, line 234: Did this mean to say kqueue is the epoll equivalent?

owenhilyard: `kqueue` is a completion-based IO API, like `io_uring`. It's much slower and still requires you to make `read` or `write` syscalls, but I personally place it closer to `io_uring` in terms of API design than `epoll`. If a BSD gets something more like `io_uring` in terms of talking to the kernel via shared memory, then I think that noting the difference in performance would be more important.

lsh:
I'm still mulling over other parts of this document, but I think `Optional[Origin]` is probably not correct and will result in **many** ergonomics problems. In any case where `None` can be used, an empty origin (`__origin_of()`) or `MutableAnyOrigin` is probably more correct.

owenhilyard:
> I'm still mulling over other parts of this document, but I think `Optional[Origin]` is probably not correct and will result in **many** ergonomics problems. In any case where `None` can be used, an empty origin (`__origin_of()`) or `MutableAnyOrigin` is probably more correct.

This and other issues has me considering whether an "OwnedOrigin" would be useful as an abstraction tool. 

ack on the ergonomics issues, I'm happy to rewrite that parameter. I'm trying to indicate "may be owned or borrowed".

lsh:
I think you should consider this the same way `Cow`'s lifetime works in Rust. It's optional in the sense that it _can_ be owned. Though I'm also starting to wonder how necessary the origin is for the trait's you're describing, since I don't really see it used in a ton of the functions. But I'll save some of that for a more in depth pass.

owenhilyard:
> I think you should consider this the same way `Cow`'s lifetime works in Rust. It's optional in the sense that it _can_ be owned. Though I'm also starting to wonder how necessary the origin is for the trait's you're describing, since I don't really see it used in a ton of the functions. But I'll save some of that for a more in depth pass.

The origin is there to facilitate zero copy. In most cases, you will "own" the buffer you get back from a `read`, but if you were to do a `read` from a span, then it's effectively creating a new span, which means you'd need an origin. With `write`, it's used to constrain buffering so that the IO engine can't "use after free" by reusing a borrowed buffer, for IO engines which can deal with `Span[UInt8]`. It's also because without that getting handed in, there's no way to provide an origin to those buffers or somehow constrain it. 

Rust's `Cow` is a good example of why we need to support both owned buffers and borrowed ones, and for some IO engines the `IoBuffer` instance might look like Rust's `Cow`, especially for something like an actor system where the data might come from the network but it might also be a "send message" transparently converted into a "receive message" function call for two actors on the same server.

lsh:
As an update: @owenhilyard and I discussed this proposal a little bit offline. We're interested, but remain cautious, and think it would be best to see a more fleshed out prototype using the features currently available in Mojo. The proposed API is a major departure from what most users expect, so having a working prototype makes it easier to evaluate.

owenhilyard:
From that same discussion, I think that waiting for parametric traits may be best since while it is possible to implement this API without them, it substantially reduces the flexibility of the implementation and produces a much worse user experience. Since that particular feature is theoretically going to be available in the near future, I will wait to start development work. However, I'd like to encourage continued discussion.

lsh:
I personally am still skeptical that we actually want an origin on the trait rather than just on the types implementing it, but maybe with a more fleshed out example the use of those origins will be more clear.

owenhilyard:
@lsh Andrew Kelley (of Zig) and I seem to have had similar ideas at similar times. Zig's new "colorless" IO interface looks a lot like this proposal, so if you don't think this is urgent I suggest we table the discussion for now and let the Zig community try it out. I'll also go write a small object store in Zig and see what I think Mojo could improve on an API front since this is a much lower effort way to test out the idea than a full impl in Mojo, and I should be able to translate a lot of what I do back to Mojo fairly easily since Mojo's type system is sort-of the other side of the coin for Zig's type system.

PR #4741, opened 2025-06-04
[stdlib] Add proposal for precondition assisted execution

akneni:
This is a more formal proposal of the idea that was proposed in [this discussion](https://forum.modular.com/t/precondition-based-optimization-library-design-proposal/1568) on the modular forum. 

Signed-off-by: Anish Kanthamneni <akneni@gmail.com>


martinvuyk:
I agree with wanting to set a standard, but not necessarily with your examples.

I think `StringSlice` should be parametrized on encoding (see proposal #3988, which also includes ASCII and other encoding-specific speedups), maybe we could keep `String` as UTF-8 only and add these parameters to its methods and have it rebind the `StringSlice` to an ASCII one. So yeah in a way I agree these parameters might be needed if we keep `String` as UTF-8 only.

And the alignment parameter is part of `UnsafePointer` and `Span` already so I don't really see a need for specifying it.

But in general :+1: to standardizing these kinds of parameters.

gryznar:
@martinvuyk I think the better example here is with `get` and `unsafe_get` on `List` iirc. Instead of 2 methods, we can just have one with `unsafe_assume_in_bounds=False` (or sth similar)

akneni:
Thanks for the feedback. I'll leave the string example there, but I'll replace the alignment example with the list one.

akneni:

gryznar:
I like this proposition. It is simple, straightforward with a solution for having unsafe hatches for performance critical parts

soraros:
@akneni Do you mind fixing the link to the discussion in the PR description?

akneni:
@soraros Fixed!

gryznar:
Why this pr was closed? I see it as very valuable proposition

PR #4734, opened 2025-06-03
[stdlib] Modernise `PyObjectPtr`

soraros:
- Make it `Boolable`, `Intable`, and `Writable`.
- Remove `_get_ptr_as_int`.

PR #4744, opened 2025-06-05
[stdlib] Remove `PyObjectPtr.is_null()` in favour of using `not ptr`

soraros:
Follow-up to #4734.

The rationale is to eliminate a redundant API and make the codebase more consistent with standard Python conventions.


soraros:
@JoeLoser This is a rather trivial patch, PTAL.

JoeLoser (approved):
Thanks for the cleanup, LGTM!

JoeLoser:
!sync

PR #4747, opened 2025-06-05
[stdlib] Simplify more code using `var`

soraros:
Migrate all usages to the new `var (name,)+ name =` pattern.

### DIFF
--- a/max/kernels/src/kv_cache/types.mojo
+++ b/max/kernels/src/kv_cache/types.mojo
@@ -450,7 +450,7 @@ struct PagedKVCache[
             "KVCache head_dim_idx is out of range",
         )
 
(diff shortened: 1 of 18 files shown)


mojo/stdlib/test/builtin/test_uint.mojo
```diff
@@ -174,7 +174,7 @@ def test_mod():
 
 
 def test_divmod():
-    var a, b = divmod(UInt(7), UInt(3))
+    a, b = divmod(UInt(7), UInt(3))
```
JoeLoser: **Suggestion** Please revert this one so we explicitly use `var` still.

JoeLoser (approved):
Much nicer, thank you!  Just one small comment.

JoeLoser:
!sync

PR #4753, opened 2025-06-06
[kernel] Fix `kl_div` for SIMD inputs

soraros:


JoeLoser (commented):
Should we add a test exposing this bug, please?

soraros:
Where should I put the tests?

JoeLoser:
> Where should I put the tests?

@fricc33 @abduld @modular/kernels do you have a preference on test location given this is an internal util?

abduld:
We could just create a test/internal_utils package

soraros:
Actually, there is already a [`test_measure.mojo`](https://github.com/modular/modular/blob/main/max/kernels/test/internal_utils/test_measure.mojo). I don’t know how I missed that.

max/kernels/test/internal_utils/test_measure.mojo
```diff
@@ -64,6 +64,28 @@ fn test_correlation() raises:
     x.free()
 
 
+fn test_kl_div() raises:
+    alias dtype = DType.float32
+    alias out_dtype = DType.float64
+    alias len = 10
+
+    var a = UnsafePointer[Scalar[dtype]].alloc(len)
```
JoeLoser: **Suggestion** Should we just use `InlineArray` instead and then can avoid the `.free()` calls?

JoeLoser (approved):

max/kernels/test/internal_utils/test_measure.mojo
soraros: Good idea, let me try that.

JoeLoser (approved):

JoeLoser:
!sync

max/kernels/test/internal_utils/test_measure.mojo
```diff
@@ -64,6 +64,25 @@ fn test_correlation() raises:
...
+fn test_kl_div() raises:
+    alias dtype = DType.float32
+    alias out_dtype = DType.float64
+    alias len = 10
+
+    var a = InlineArray[Scalar[dtype], len](uninitialized=True)
+    var b = InlineArray[Scalar[dtype], len](uninitialized=True)
+    for i in range(len):
+        a[i] = Scalar[dtype](0.01 * i)
+        b[i] = Scalar[dtype](0.1 * i)
```
tboerstad, line 76:
```suggestion
        a[i] = Scalar[dtype](1.0 / len)
        b[i] = Scalar[dtype](2.0 * (i + 1) / (len * (len + 1)))
```
NIT:
I would expect the inputs of KL Div to individually sum to 1.0, here's two which do, 
Since it's floating point math they won't really, but it's less confusing to me.

If it's a hassle to compute a new reference value, feel free to skip this comment

soraros: Ah, I don't know how I missed this review. I can make the changes if needed still. What do you think?

Issue #4754, opened 2025-06-06
[cleanup][stdlib] Move the entire codebase to use `List.unsafe_ptr` instead of accessing `.data` directly


soraros:
### What is your request?

As title. CC @JoeLoser.

### What is your motivation for this change?

So we can change the internal of `List` more easily, enabling work like #2825.

### Any other details?

I think it’s best to do this in steps:

- Rename the field to `_data` (`_` prefix for private fields is good practice).
- Use compilation errors to guide the migration.

This work is likely blocked by #4739, which I discovered during my own migration attempt.

not-heavychevy:
Hello! Can I take up this issue?

Issue #4739, opened 2025-06-04
[BUG] Compiler doesn't correctly handle the `ref` convention with an `AddressSpace` parameter


soraros:
### Bug description

As title. CC @lattner.

### Steps to reproduce

The error message also got cut off.

```mojo
from memory import UnsafePointer

struct S:
  fn unsafe_ptr(ref self) -> UnsafePointer[Int]:
    return {}

fn f[a: AddressSpace](ref[_, a] s: S):
  s.unsafe_ptr()  # error: invalid call to 'unsafe_ptr':
```

### System information

`Mojo 25.4.0.dev2025060220 (d9421ccb)`

The error message:
```
gh_4739.mojo:8:15: error: invalid call to 'unsafe_ptr':
  s.unsafe_ptr()  # error: invalid call to 'unsafe_ptr':
  ~~~~~~~~~~~~^~
gh_4739.mojo:4:6: note: function declared here
  fn unsafe_ptr(ref self) -> UnsafePointer[Int]:
     ^
mojo: error: failed to parse the provided Mojo source module
```

martinvuyk:
@soraros I think it's just because the self argument can't be inferred since it is using the default address space.

This works:
```mojo
from memory import UnsafePointer


struct S:
    fn unsafe_ptr[
        o: Origin, a: AddressSpace
    ](ref [o, a]self) -> UnsafePointer[
        Int, mut = o.mut, origin=o, address_space=a
    ]:
        return {}


fn f[a: AddressSpace](ref [_, a]s: S):
    s.unsafe_ptr()  # warning: value is unused
```

soraros:
That "fix" solves the immediate issue in the small repro, but this bug doesn't exist in a vacuum. Specifically, `S` is actually a reduced ~~`Span`~~`List` which carries an `address_space`. We should at least expect a better error message from the compiler.

Maybe this means we want a `List.unsafe_ptr` that's generic over `AddressSpace`, but I think that hints at a usability problem in and of itself.

Of maybe we should use a refined `Self` here:
https://github.com/modular/modular/blob/cc483c8dfa959127034f865f59607b61cfd92239/mojo/stdlib/stdlib/memory/span.mojo#L165-L174

martinvuyk:
I agree it's a usability problem but to me it's the same as the usability problem of carrying an origin in the parameters, now we would just have to add AddressSpace to that list

soraros:
The more complete repro:

```mojo
from memory import UnsafePointer

[...]

PR #4756, opened 2025-06-06
[stdlib] Fix `fn unsafe_ptr(ref self) -> ...` methods to handle address space

martinvuyk:
Fix `fn unsafe_ptr(ref self) -> ...` methods to handle address space. Part of #4754, workaround for #4739

CC: @soraros


lsh:
!sync

Issue #4759, opened 2025-06-06
[BUG] Compiler doesn't catch incorrectly declared `alias` while LSP does


soraros:
### Bug description

As title. CC @zyx-billy.

### Steps to reproduce

LSP complains, while CI passes, why?

https://github.com/modular/modular/blob/cc483c8dfa959127034f865f59607b61cfd92239/max/kernels/src/layout/layout_tensor.mojo#L6840

### System information

`Mojo 25.4.0.dev2025060513 (2db7011b)`

VerdagonModular:
I'm not seeing an error on that line, what does the LSP say?

soraros:
@VerdagonModular, then it's even worse now. I've since removed the `alias` in #4760. The original problem was that an alias declaration, which doesn't even type-check, was included in the OSS codebase that's supposedly CI-checked. I suspect it's either a CI problem or that, since this `alias` was DCEd, and somehow slipped through.

VerdagonModular:
Got it, that does sound rather worrisome. I'll keep an eye out, and next time I'll try to get to the bug sooner so I can see it live. I'll close this for now since I can't repro, thanks for putting this on my radar!

PR #4760, opened 2025-06-06
[kernel] Remove redundant `alias` from `layout_tensor`

soraros:
This line doesn't even type check. I'm not sure why the compiler didn't catch it.

Related compiler bug filed at #4759.


JoeLoser (approved):

JoeLoser:
FYI @modular/kernels @fricc33 @abduld

JoeLoser:
!sync

PR #4769, opened 2025-06-07
[stdlib] Introduce fnv1a hasher

mzaks:
As requested in PR #3701 lets first introduce the fnv1a hash function, and then we can try migration to the hasher based hashing.


mzaks:
"I have read the CLA Document and I hereby sign the CLA"

mzaks:
recheck

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
```diff
@@ -0,0 +1,83 @@
...
+
+        Args:
+            value: Value used for update.
+        """
+
+        @parameter
+        if value.dtype.is_floating_point():
+            v64 = value.to_bits().cast[DType.uint64]()
+        else:
+            v64 = value.cast[DType.uint64]()
```
soraros: I think this is no longer sufficient, since we have `uint128` and wider integer types.

soraros: I'm fine with adding a `constrained` and a TODO though.

soraros: We could also restrict this function's API to only accept types with `bit_width < 64`, leaving bitcasting to the caller.

mzaks: Oh, I see. No need for constrained. I will think about a better solution.

JoeLoser:
!sync

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
mzaks: I fixed it for fnv1 and also for the ahash. its a bit more involved now

JoeLoser:
!sync

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
```diff
@@ -0,0 +1,97 @@
...
+            value: Value used for update.
+        """
+
+        # number of rounds a single vector value will contribute to a hash
+        # values smaller than 8 bytes contribute only once
+        # values which are multiple of 8 bytes contribute multiple times
+        # e.g. int128 is 16 bytes long and evaluates to 2 rounds
+        alias rounds = (value.dtype.sizeof() >> 3) + (
+            (value.dtype.sizeof() & 7) > 0
+        )
```
soraros:
```suggestion
        alias rounds = value.dtype.sizeof() // 8 + (
            value.dtype.sizeof() % 8 > 0
        )
```

mojo/stdlib/stdlib/hashlib/_ahash.mojo
```diff
@@ -171,20 +171,42 @@ struct AHasher[key: U256](_Hasher):
...
-                self._large_update(U128(v64[i], v64[i + 1]))
+            if v64.size == 1:
+                self._update(v64[0])
+            else:
+
+                @parameter
+                for i in range(0, v64.size, 2):
+                    self._large_update(U128(v64[i], v64[i + 1]))
+        else:
+            # vector values will contribute to hash in multiple rounds
```
soraros, line 202: Let add a `constrained[width > 64]()` here just to clarify the logic.

soraros (commented):
Looks good to me, with one small nit: please follow the stdlib style guide by always using `var` and pre-declaring variables that are initialised later.

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
soraros: I think it's still ok, I could add support for `to_bits()` when `dtype = bool` to remove a branch.

mojo/stdlib/stdlib/hashlib/_ahash.mojo
mzaks: I think it is a bit redundant, but it does not heart :)

mzaks: Added the constraint below, with another condition that the value should be integral.

mzaks: BTW, I see we have tons of GPU specific floating point values are any of them bigger than 8 bytes? @JoeLoser

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
mzaks: IMHO `dtype = bool` doe snot need special handling it will cast to UInt64 just fine, floating point numbers need `to_bits()` as they do logical casting otherwise

mojo/stdlib/stdlib/hashlib/_ahash.mojo
mzaks: Ah I checked the docs, seems to be all good for now. I see Swift also has Float80, but for now we are good :)

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
mzaks: @soraros or do you mean the cast internally having a branch and we could remove it when using `to_bits()` for `bool` as well?

gryznar:
Maybe it will be valuable to set up (or extend) in current or future pr, some bechmarking infrastracture to measure hashing performance?

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
soraros: I will do that in a later refactor, just letting you know. So we can write simply `v.to_bits().cast[...]()` in the future.

mzaks:
> Maybe it will be valuable to set up (or extend) in current or future pr, some bechmarking infrastracture to measure hashing performance?

I think it is a great idea, but let's do it in a separate PR as this one is already quite large.

mojo/stdlib/stdlib/hashlib/_fnv1a.mojo
```diff
@@ -0,0 +1,98 @@
...
+        mut self, data: UnsafePointer[UInt8, mut=False, **_], length: Int
+    ):
+        """Consume provided data to update the internal buffer.
+
+        Args:
+            data: Pointer to the byte array.
+            length: The length of the byte array.
+        """
+        for i in range(length):
+            self._value ^= data[i].cast[DType.uint64]()
```
JoeLoser, line 45: **Quibble** The wikipedia page linked does the multiplication first before the XOR, but it doesn't matter much.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
FYI this got merged internally yesterday.  @patrickdoc why did copybara fail to put the "merged internally" label here?

PR #4770, opened 2025-06-08
[stdlib] Refactor `float32`-to-`bfloat16` conversion function

soraros:
Make use of the SIMD-aware `to_bits` and `from_bits` functions instead of relying on the compiler to pattern-match the unrolled loop.


JoeLoser:
Generally LGTM, can someone from @modular/kernels also take a look, please?

laszlokindrat (approved):
LGTM, let's keep an eye open for perf regressions though, just in case.

laszlokindrat:
!sync

PR #4772, opened 2025-06-08
[stdlib] Refactor SIMD ctors so they generate less IR

soraros:
We skip ctor calls when possible so we don't run `_simd_construction_checks` a thousand time.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/simd.mojo
+++ b/mojo/stdlib/stdlib/builtin/simd.mojo
@@ -564,13 +564,14 @@ struct SIMD[dtype: DType, size: Int](
         var s = __mlir_op.`pop.cast_from_builtin`[
             _type = __mlir_type.`!pop.scalar<bool>`
         ](value.value)
-        alias simd_bool = SIMD[DType.bool, size]._mlir_type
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -231,10 +231,7 @@ fn _simd_construction_checks[dtype: DType, size: Int]():
 
 @always_inline("nodebug")
 fn _unchecked_zero[dtype: DType, size: Int]() -> SIMD[dtype, size]:
-    var zero = __mlir_op.`pop.cast`[_type = Scalar[dtype]._mlir_type](
-        __mlir_attr.`#pop.simd<0> : !pop.scalar<index>`
-    )
-    return Scalar[dtype](zero)
```
soraros, line 237: This is not really *unchecked* since `Scalar[dtype]` to `SIMD[dtype, size]` conversion runs `_simd_construction_checks`.

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

Issue #4586, opened 2025-05-15
[Feature Request] Migrate off of `@value`


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Top level issue to keep track of the migration.

Many types can just replace
```mojo
@value
struct SomeType: ...
```
by
```mojo
@fieldwise_init
struct SomeType: ...
```
And others need the move-init or copy-init functions as well
```mojo
@fieldwise_init
struct SomeType(Copyable, Movable): ...
```

### What is your motivation for this change?

First commit is 7e621db98eb168df31915879516cbc45bd73f609 by @lattner 

### Any other details?

Good first issue for anyone looking to contribute

bgreni:
Oh wow I didn't even realize the `Copyable` and `Movable` auto-generation was already working!

lattner:
Thank you for filing this, I'd love help with this!

lattner:
For largely-mechanical patches like this, I think it would be ok to do fewer but bigger patches.

JoeLoser:
Only 322 uses remaining thanks to some recent PRs.  Would love more help from folks on this!  

```
git grep '@value' mojo/stdlib/ | wc -l
322
```

pbanavara:
This looks fixed.

`
git grep '@value' mojo/stdlib/ | wc -l
0
`

soraros:
FYI @JoeLoser.

JoeLoser:
This is now complete across stdlib and kernel libraries.  Thanks everyone for your help here! ❤️

PR #4785, opened 2025-06-10
[stdlib] Remove usage of @value from max

bgreni:
Opening a draft to reflect my progress on this as I'm chipping away at it

Related to #4586

### DIFF
--- a/examples/custom_ops/benchmarks.mojo
+++ b/examples/custom_ops/benchmarks.mojo
@@ -49,13 +49,13 @@ from utils import IndexList
 
 
 # Wrap a ManagedTensorSlice and DeviceBuffer as an owning Tensor
-@value
+@fieldwise_init
 struct Tensor[
     dtype: DType,
     rank: Int, //,
     io_spec: IOSpec,
     static_spec: StaticTensorSpec[dtype, rank],
-]:
+](Copyable, Movable):
     alias size = Int(static_spec.shape.product())
 
     var slice: ManagedTensorSlice[io_spec=io_spec, static_spec=static_spec]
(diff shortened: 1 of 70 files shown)


JoeLoser (approved):
LGTM so far!  Let me know when you're ready for me to take a look and/or sync it in to merge internally.

bgreni:
@JoeLoser That should be everything left in the max directory. I also didn't add `@fieldwise_init` to any structs that had explicitly declared ctors and built successfully without it

JoeLoser (approved):

JoeLoser:
!sync

examples/custom_ops/benchmarks.mojo
```diff
@@ -49,7 +49,7 @@ from utils import IndexList
 
 
 # Wrap a ManagedTensorSlice and DeviceBuffer as an owning Tensor
-@value
+@fieldwise_init
```
JoeLoser, line 52: This one won't quite work I think:

```mojo
open-source/max/examples/custom_ops/benchmarks.mojo:86:20: error: 'Tensor[io_spec, static_spec]' is not copyable because it has no '__copyinit__'
            return self
```

bgreni: Huh I though that would be included when I ran `./bazelw build //...` looks like I missed the traits

JoeLoser:
!sync

examples/custom_ops/benchmarks.mojo
JoeLoser: I suspect it's getting skipped in the OSS build since you won't have access to these (internal) SDK targets [here](https://github.com/modular/modular/blob/3f6ad8687ae7fb9b1b3c026d2ee6d5a2d5949d26/examples/custom_ops/custom_op_example.bzl#L33).

FYI @keith @patrickdoc

PR #4786, opened 2025-06-10
[stdlib] Introduce `SIMD._refine` to manually refine types in generic code

soraros:
This replaces explicit calls to `rebind`, parallels `SIMD.cast` nicely, and improves readability in method chains.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/simd.mojo
+++ b/mojo/stdlib/stdlib/builtin/simd.mojo
@@ -746,9 +746,9 @@ struct SIMD[dtype: DType, size: Int](
 
         @parameter
         if dtype is DType.bool:
-            return (rebind[Self._Mask](self) & rebind[Self._Mask](rhs)).cast[
-                dtype
-            ]()
+            return (
+                self._refine[DType.bool]() & rhs._refine[DType.bool]()
+            )._refine[dtype]()
         return __mlir_op.`pop.mul`(self.value, rhs.value)
 
     @always_inline("nodebug")
@@ -1109,7 +1109,7 @@ struct SIMD[dtype: DType, size: Int](
 
         @parameter
         if dtype is DType.bool:
-            return rebind[Self](self.select(Self(False), Self(True)))
+            return self.select(False, True)._refine[dtype]()
         else:
             return self ^ -1
 
@@ -1481,7 +1481,7 @@ struct SIMD[dtype: DType, size: Int](
             A PythonObject representing the value.
         """
         constrained[size == 1, "only works with scalar values"]()
-        return PythonObject(rebind[Scalar[dtype]](self))
+        return PythonObject(self._refine[size=1]())
 
     @always_inline("nodebug")
     fn __len__(self) -> Int:
@@ -1511,7 +1511,7 @@ struct SIMD[dtype: DType, size: Int](
                 " instead."
             ),
         ]()
-        return rebind[Scalar[DType.bool]](self.cast[DType.bool]()).value
+        return self._refine[size=1]().cast[DType.bool]().value
 
     @always_inline("nodebug")
     fn __int__(self) -> Int:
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -1732,6 +1732,21 @@ struct SIMD[dtype: DType, size: Int](
     # Methods
     # ===------------------------------------------------------------------=== #
 
+    @always_inline("nodebug")
+    fn refine[
```
JoeLoser: **Question** Should this be private instead?

soraros: I kinda want people to use it in kernel code as well.

JoeLoser: Do you have a good sense as to how much code this would clean up in the kernels directory?  Even just rough order?

soraros: I think there are a few hundred occurrences in `kernels` and a few dozen in `stdlib`. `MOGGKernelAPI.mojo` alone has 82 instances of `rebind[SIMD]`. That said, I'm fine with keeping this private.

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -746,9 +746,9 @@ struct SIMD[dtype: DType, size: Int](
 
         @parameter
         if dtype is DType.bool:
-            return (rebind[Self._Mask](self) & rebind[Self._Mask](rhs)).cast[
-                dtype
-            ]()
+            return (
+                self.refine[DType.bool]() & rhs.refine[DType.bool]()
+            ).refine[dtype]()
```
martinvuyk: Question:
is `rebind` really the same as `cast`? Because currently `cast` is `@always_inline("nodebug")` but could it potentially become a `@always_inline("builtin")` ? And can `rebind` also do it in the future?

soraros: Of course, `rebind` is not the same as `cast`. In this particular case, we’re using `cast` as a no-op, which isn't good and should have been `rebind` all along.

Moreover, I don't think `cast` can ever be a `"builtin"` when it's doing work, and some code paths have huge implementation in Mojo.

mojo/stdlib/stdlib/builtin/simd.mojo
soraros: Let's start with a private `_refine` then.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
From feedback internally from @abduld: perhaps call this `_rebind`?

soraros:
This is now private API, so maybe we (I) shouldn't worry about the naming too much (and just rename it), though I think `refine` is the better name for this method, as it describes what it does (in fact, the docstring of `rebind` also describes its effect as "manually *refine* the type"). I want this method to feel less magical than the free function `rebind` and more like a checked no-op `cast`. WDYT?

JoeLoser:
I'm seeing some failures internally on GPU test:

```
open-source/max/max/kernels/test/gpu/basics/test_print.mojo:1:1: error: failed to run the pass manager for offload functions
^
open-source/max/max/kernels/test/gpu/basics/test_print.mojo:65:69: error: call expansion failed
            var c = value.cast[DType.bfloat16]().cast[DType.float64]()
                                                                    ^
open-source/max/mojo/stdlib/stdlib/builtin/simd.mojo:1752:8: note: function instantiation failed
open-source/max/mojo/stdlib/stdlib/builtin/simd.mojo:1804:66: note: call expansion failed
open-source/max/mojo/stdlib/stdlib/builtin/simd.mojo:1749:9: note: error: rebind input type '!pop.scalar<f32>' does not match result type '!pop.scalar<f64>'
```

JoeLoser:
!sync

Issue #4758, opened 2025-06-06
[Feature Request] [proposal] Centralize algorithms around `Span` and add high perf. utils to it


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

I would like us to transform `Span` into an auto-vectorized beast that makes users not need to dip down into `UnsafePointer` and `SIMD`.

This would require `Span` to implement specialized methods like `fn reverse[dtype: DType, //](self: Span[mut=True, Scalar[dtype]]): ...`

PRs adding such utils:
- #3548
- #3792
- #3875
- #3877

PR changing some functions that can be implemented using `Span` instead of `UnsafePointer`:
- #4169

PR adding a util to help `List` use the specialized `Span` methods:
- #3810

### What is your motivation for this change?

I wanted to get this out there as a standalone ticket to get some feedback and explain why I've been pushing for these PRs

### Implementation details

Many of the high performance implementations could be reduced to just a method and using `vectorize`. I'm no expert in SIMD, but I think there might be cases like `.reverse()` where e.g. `SIMD[DType.uint8, 8]` might be less than the vectorized loop width but should (AFAIK) still be faster using SIMD than doing it in the drain loop. This is something that might be worth exploring and even implementing a helper function that runs at compile time that returns the list of SIMD widths that are worth iterating over for the given datatype.

I would also like to extend these kind of optimizations in the future up to `List` and also cast trivial types to the datatype of the same width.

### Other methods we could add
- max
- min
- argmax
- argmin
- `Span[Scalar[dtype]].map[o_dtype: DType, func: fn [w: Int](SIMD[dtype, w]) -> SIMD[o_dtype, w]]() -> List[Scalar[o_dtype]]`

PR #4791, opened 2025-06-10
[stdlib] Add auto-vectorization infra for `List`

martinvuyk:
Add auto-vectorization infra for `List`. Part of #4758.

It would be best to measure the compilation time effect of adding this.

### DIFF
--- a/mojo/stdlib/stdlib/collections/list.mojo
+++ b/mojo/stdlib/stdlib/collections/list.mojo
@@ -244,13 +244,17 @@ struct List[T: Copyable & Movable](
         T: The type of elements stored in the list.
     """
 
+    # Aliases
+    alias _Span = Span[T, MutableAnyOrigin]
+    alias _AnyOriginUnsafePointerType = Self._Span.UnsafePointerType
+
     # Fields
-    var _data: UnsafePointer[T]
-    """The underlying storage for the list."""
+    var _data: Self._AnyOriginUnsafePointerType
+    """The underlying storage for the `List`."""
     var _len: Int
-    """The number of elements in the list."""
+    """The number of elements in the `List`."""
     var capacity: Int
-    """The amount of elements that can fit in the list without resizing it."""
+    """The amount of elements that can fit in the `List` without resizing it."""
 
     alias IteratorType[
         iterable_mut: Bool, //, iterable_origin: Origin[iterable_mut]
@@ -262,7 +266,7 @@ struct List[T: Copyable & Movable](
 
     fn __init__(out self):
         """Constructs an empty list."""
-        self._data = UnsafePointer[T]()
+        self._data = {}
         self._len = 0
         self.capacity = 0
 
@@ -273,9 +277,9 @@ struct List[T: Copyable & Movable](
             capacity: The requested capacity of the list.
         """
         if capacity:
-            self._data = UnsafePointer[T].alloc(capacity)
+            self._data = Self._AnyOriginUnsafePointerType.alloc(capacity)
         else:
-            self._data = UnsafePointer[T]()
+            self._data = {}
         self._len = 0
         self.capacity = capacity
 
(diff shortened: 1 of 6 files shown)


lsh:
Given some of the comments and aliases you have here, we may want to wait until we resolve the Upgrading Trivial proposal https://forum.modular.com/t/proposal-upgrading-trivial/1351/17

martinvuyk:
> Given some of the comments and aliases you have here, we may want to wait until we resolve the Upgrading Trivial proposal https://forum.modular.com/t/proposal-upgrading-trivial/1351/17

@lsh interesting cool to know that will make some parts of the code simpler. The only thing we would need to change in this pr is the handling of the different kinds of triviality. That means I just have to drop the new `_is_trivial` alias and the uses I made of it. The rest of this PR won't have much to do with the type of triviality but with what datatypes it can be transformed into.

I just dropped the `_is_trivial` alias and we can adjust all uses of `hint_is_trivial_type` once that proposal gets implemented

lsh:
Now that upgrading trivial is in, can you update this?

martinvuyk:
This is blocked until we have a `__copy_is_trivial` flag or something similar for `ExplicitlyCopyable`

JoeLoser:
Do we need to couple the auto-vectorization infra, or can we first start with just using the trivial destructor part, wait for us to fix copyinit trivial for `ExplicitlyCopyable`, and then proceed to leverage that and remove the `hint_trivial_type` parameter?

martinvuyk:
I just kept `hint_trivial_type` as a proxy for `__copy_is_trivial` for now. I can reduce the PR in scope more if you wish

martinvuyk:
@JoeLoser @lsh I've thought about this some more and reworked the compile time code to live in `Span` instead of `List`. I think it will be better because many collection types rely on `Span`, and also because maybe the compiler can optimize several instantiations of `Span[T, MutableAnyOrigin]` to only call the various SIMD estimation functions once.

I also think that taking most of the infra to `Span` makes more sense in delegating the responsibilities where the implementations are currently redundant. I implemented `__contains__` as an example here but there are many more cases. Implementing them generically with the optimizations in `Span` should also help apply them automatically for any collection that has contiguous memory.

I also think we might eventually be able to apply even more interesting optimizations like e.g. once we have a way to signal (compiler synthesized or through a trait alias) that the `__eq__` comparison for a type is trivial and its fields' are also trivial, we can just bitcast it to the corresponding uint dtype and use vectorized functions. We could likewise use `memcmp` when `>=`, `>`, `<=`, `<` are also trivial. This should give us a practically free speedup for any collection of trivial types :exploding_head: 

PS: I have no problem if this PR takes some time. I know these things take time to discuss and consider properly.

mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -234,21 +234,34 @@ struct List[T: ExplicitlyCopyable & Movable, hint_trivial_type: Bool = False](
...
+    alias _has_trivial_copy = hint_trivial_type  # Bool(T.__copy_is_trivial)
+    alias _has_trivial_move = Bool(T.__moveinit__is_trivial)
+    alias _is_fully_trivial = (
+        Self._has_trivial_move
+        and Self._has_trivial_copy
+        and Self._has_trivial_del
+    )
+
+    alias _Span = Span[T, MutableAnyOrigin]
+    alias _U = Self._Span.UnsafePointerType
```
lsh: I usually try not to block on naming but I'm pretty against single letter aliases like this

mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -234,21 +234,34 @@ struct List[T: ExplicitlyCopyable & Movable, hint_trivial_type: Bool = False](
...
+
     # Fields
-    var _data: UnsafePointer[T]
-    """The underlying storage for the list."""
+    var _data: Self._U
```
lsh: Furthermore I don't see how this is better, it just looks more confusing to me.

lsh (changes requested):

mojo/stdlib/stdlib/collections/list.mojo
martinvuyk: It is not about renaming just for brevity, it's because of all the parameter specifications. I'll make it more descriptive

PR #4789, opened 2025-06-10
[stdlib] Rename the `PythonObject` overload of `assert_equal` to `assert_equal_pyobj`

soraros:
This prevents values from being accidentally considered equal through implicit conversion to `PyObj`.

PR #4793, opened 2025-06-11
[stdlib] Refactor and optimize testing utils `Error` construction

martinvuyk:
Refactor and optimize testing utils `Error` construction. Partly motivated by  #4789, partly because this was on my TODO list.

This should improve the runtime performance of all tests by a significant amount.

I tried not having to pass around the `write_to` functions by building a wrapper type:
```mojo
struct _Wrapper[write_message: fn[W: Writer] (mut writer: W) capturing](Writable):
    fn write_to[W: Writer](self, mut writer: W):
        write_message(writer)
```
but the `Writable` trait won't be accepted because the `write_to` function would be capturing. I made all the involved functions private for now until that `Writer` feature (#5274) is implemented then going back to using the writable trait should be simply switching.

### DIFF
--- a/mojo/stdlib/std/builtin/_location.mojo
+++ b/mojo/stdlib/std/builtin/_location.mojo
@@ -1,0 +1,133 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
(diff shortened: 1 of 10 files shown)


martinvuyk:
ping @jackos

NathanSWard:
Error has recently been reworked to directly hold a `String`.
There's also some work on adding a new `Printable` trait (subsuming `Writable`, `Stringable`, `Representable`) after which we can rework `Error` to have an optimized implementation for working with the new trait!

Closing this for now - but feel free to reopen if you believe this is incorrect!

martinvuyk:
@NathanSWard This dealt with another issue which is long chains of write operations that can be solved by passing a capturing function around and writing only once to the final writer. You can see #5274 for more context and examples.

Could you please reopen this (I can't)? I'll refactor the internals in this PR (if necessary) when I have more time

martinvuyk:
@NathanSWard could you please reopen this? I can't. I will fix it for the recent changes once I have a bit more time (I'm coming back from a two week vacation so I'm a bit swamped at the moment). I could adapt it once we have unified closures a bit more polished; I could fix it after or before merging once that happens, whatever seems best to you. But this PR does solve an issue that isn't yet solved, and could serve as a use-case for adding a closure-based writing API (so I think that if you read the code you will find some of the examples of what I'm trying to solve here).

PR #4800, opened 2025-06-12
[stdlib] Fix integer formatting for types with a bitwidth greater than 64

soraros:
CC @jackos.

### DIFF
--- a/mojo/stdlib/test/builtin/test_format_int.mojo
+++ b/mojo/stdlib/test/builtin/test_format_int.mojo
@@ -26,27 +26,39 @@ fn test_format_int() raises:
     assert_equal(_format_int[DType.index](-123, 10), "-123")
     assert_equal(_format_int[DType.index](-999_999_999, 10), "-999999999")
 
-    #
-    # Max and min i64 values in base 10
-    #
-
+    # i64
     assert_equal(_format_int(Int64.MAX_FINITE, 10), "9223372036854775807")
-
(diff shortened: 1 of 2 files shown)


mojo/stdlib/test/builtin/test_format_int.mojo
```diff
@@ -26,27 +26,39 @@ fn test_format_int() raises:
...
+    assert_equal(_format_int(Int64.MAX_FINITE, 2), "1" * 63)
+    assert_equal(_format_int(Int64.MIN_FINITE, 2), "-1" + "0" * 63)
 
-    #
-    # Max and min i64 values in base 2
-    #
-
+    # i128
+    alias int128_max = Int128(UInt128(1) << 127) - 1
+    alias int128_min = Int128(UInt128(1) << 127)
```
soraros, line 37: We couldn't use `Int128.MAX` for `max_or_inf` because `max_or_inf` doesn't support it; and we can't easily add support due to #4049. One step at a time.

JoeLoser (commented):
LGTM, but I'll let @jackos also take a look.  Thanks for fixing this!

jackos (approved):
Cool thanks @soraros

Couple of things for the future, we should remove adding the null terminator in _format_int, it returns a span with the null terminator cut off anyway.

And yeah I have noticed that problem materializing an IntLiteral to UInt128 where the leftmost bit is 1, it thinks it's too large a value to fit, will fix this soon

jackos:
!sync

PR #4801, opened 2025-06-12
[stdlib] Improve os.remove error message (fixes #2342)

### LINKED ISSUE #2342: [stdlib] capture error message in os.remove
for now `os.remove` only raises a generic messages when it fails, ideally is should capture the error to have a more explicit message, the concerned code is here:
https://github.com/modularml/mojo/blob/20737dcb02ebf091c77b931e5e8b8abb2d4bb164/stdlib/src/os/os.mojo#L264-L271

hardikkgupta:
Fixes #2342

This patch captures errno after unlink failure and raises


hardikkgupta:

hardikkgupta:
recheck

mojo/stdlib/stdlib/os/os.mojo
```diff
@@ -269,45 +269,40 @@ fn abort[result: AnyType = NoneType._mlir_type](message: String) -> result:
...
+# --- errno / strerror bindings (keep near other C interop in this file) -------
 
-    Parameters:
-      PathLike: The a type conforming to the os.PathLike trait.
+@ctypes.extern var errno: c_int                 # POSIX errno
 
-    Args:
-      path: The path to the file.
+@ctypes.extern                                  # char *strerror(int)
+fn strerror(errnum: c_int) -> Pointer[c_char]
```
gryznar: Should we use flatcase here (`strerror` instead of `str_error` and `errnum` instead of `err_num`)?

hardikkgupta: I followed the flat-case convention for C interop symbols, so the current signature already uses `strerror` and `errnum`
Let me know if you’d prefer we adopt snake_case for Mojo-side names instead, but I think mirroring the original libc identifiers

hardikkgupta: makes it more clearer

hardikkgupta:
@gryznar What do you think about the PR? I am sure why tests are failing?

gryznar:
@hardikkgupta sorry, I don't feel qualified here. I had just small remark to the pr

mojo/stdlib/stdlib/os/os.mojo
```diff
@@ -269,45 +269,40 @@ fn abort[result: AnyType = NoneType._mlir_type](message: String) -> result:
...
-# remove/unlink
+# remove / unlink
 # ===----------------------------------------------------------------------=== #
+
+# --- errno / strerror bindings ----------------------------------------------
+@ctypes.extern
+var errno: c_int                     # POSIX errno
+
+@ctypes.extern                       # char *strerror(int)
+fn strerror(errnum: c_int) -> UnsafePointer[c_char]
```
laszlokindrat, line 280: This isn't valid mojo right now, and probably the reason why the tests are failing. Did you mean to make these a comment?

mojo/stdlib/stdlib/os/os.mojo
```diff
@@ -22,7 +22,7 @@ from os import listdir
 
 from collections import InlineArray, List
 from sys import external_call, is_gpu, os_is_linux, os_is_windows
-from sys.ffi import OpaquePointer, c_char
+from sys.ffi import ctypes, c_int, OpaquePointer, c_char
```
laszlokindrat, line 25: We don't have `ctypes` in `sys.ffi`. What is it that you're trying to do here?

mojo/stdlib/stdlib/os/os.mojo
```diff
@@ -269,45 +269,40 @@ fn abort[result: AnyType = NoneType._mlir_type](message: String) -> result:
...
-        # _ = external_call["perror", UnsafePointer[NoneType]](error_str.unsafe_ptr())
-        # _ = error_str
-        raise Error("Can not remove file: " + fspath)
-
+    var rc = external_call["unlink", Int32](fspath.unsafe_cstr_ptr())
+
+    if rc != 0:
+        # Capture ``errno`` *immediately*.
+        var code: Int32 = errno
+        var msg = String.from_c_str(strerror(code))
```
laszlokindrat, line 298: String doesn't have a `from_c_str` method. You probably need to use one of the unsafe constructor instead.

laszlokindrat (commented):
Thanks for the patch! I'm not sure I follow your code, so I left some comments and questions. Also, can you please add some tests?

mojo/stdlib/stdlib/os/os.mojo
```diff
@@ -269,45 +269,40 @@ fn abort[result: AnyType = NoneType._mlir_type](message: String) -> result:
...
+var errno: c_int                     # POSIX errno
+
+@ctypes.extern                       # char *strerror(int)
+fn strerror(errnum: c_int) -> UnsafePointer[c_char]
+# -----------------------------------------------------------------------------
+
 fn remove[PathLike: os.PathLike](path: PathLike) raises:
     """Removes the specified file.
 
-    If the path is a directory or it can not be deleted, an error is raised.
```
martinvuyk, line 277: This should not be deleted and should be in a `Raises` docstring section

mojo/stdlib/stdlib/os/os.mojo
```diff
@@ -269,45 +269,40 @@ fn abort[result: AnyType = NoneType._mlir_type](message: String) -> result:
...
+    var rc = external_call["unlink", Int32](fspath.unsafe_cstr_ptr())
+
+    if rc != 0:
+        # Capture ``errno`` *immediately*.
+        var code: Int32 = errno
+        var msg = String.from_c_str(strerror(code))
+        raise Error(
+            String("Unable to remove '{}' : {} (errno={})")
+                .format(fspath, msg, code)
+        )
```
martinvuyk, line 302: Nit: This should directly build using `Error`'s variadic constructor i.e. `Error("Unable to remove ", fspath, ....)`

mojo/stdlib/stdlib/os/os.mojo
```diff
@@ -269,45 +269,40 @@ fn abort[result: AnyType = NoneType._mlir_type](message: String) -> result:
...
-    Absolute and relative paths are allowed, relative paths are resolved from cwd.
-
-    Parameters:
-      PathLike: The a type conforming to the os.PathLike trait.
-
-    Args:
-      path: The path to the file.
-
-    """
+    """Alias for :pyfunc:`os.remove`."""
```
martinvuyk, line 305: Why are you removing documentation for a public function? And making it actively worse?

martinvuyk (changes requested):
This looks to me like AI generated slop that wasn't even checked for existing syntax, methods, or docstrings standards. Did you even try to build the stdlib locally?

To avoid being to harsh: please read the documentation and look at other functions in the stdlib as an example. There is no use of some of the functions and they don't exist in stdlib types. When using generative AI, you have to remember to check for every detail that might be hallucinated. The LSP should have shown you where most of the issues lie

hardikkgupta:
@martinvuyk Yes I agree. I was under the impression that my thought process could actually be good enough for the generative AI to actually tweak everything correctly. Closing this PR for now and will open a new one

PR #4804, opened 2025-06-12
[stdlib] [NFC] Make `__call_location[inline_count]` keyword only

martinvuyk:
Make `__call_location[inline_count]` keyword only. Split off from  #4793


laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

Issue #4249, opened 2025-04-01
[Feature Request] [proposal] Have the `Writable` trait require a function that returns an estimate of the buffer size needed.


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Have the `Writable` trait require a function that returns an estimate of the buffer size needed. The function could return 0 in the case of datastructures that don't know it ahead of time.

### What is your motivation for this change?

Currently many uses of the writer API look like this:
```mojo
fn __str__(self) -> String:
    var res = String(capacity=some_estimate) # most of the time it's even an empty constructor
    res.write(self)
    return res
```
One either has to know the exact size of each component or guess sizes. These calculations aren't of any use when the type's `write_to` function gets called without reserving it that way.

### Any other details?

This change will be useful for:
- Letting the type define its specific memory needs, without having to reserve the string size beforehand (which is not scalable for deeply nested datastructures)
- To avoid memory allocation overhead in loops
- And maybe it also opens the door to other optimizations

CC: @jackos

jackos:
I think this could be good idea to improve String.write performance even further after https://github.com/modular/modular/pull/4573,  but it might have a negative impact on SSO for short string writes e.g. each Int would need to reserve 20 bytes and so would quickly exceed the INLINE_CAPACITY 23 bytes.

Perhaps when we have default implementations on traits it could write out to _TotalWritableBytes to calculate size without writing out to a buffer e.g. for ints and floats, but then you can overide it with e.g. len(self.string_a) + len(self.string_b). This would also remove burden from making a type Writable. WDYT?

martinvuyk:
You're right about the possible hit of over-estimation, we could also just fall-back to running the stringifying algorithm for the type when the over-allocation isn't worth it, or have some rule-of-thumb criteria to use either implementation.

I wrote some code to estimate write length for FP types in [this PR for EmberJson](https://github.com/bgreni/EmberJson/pull/41/files#diff-917b58dc87329555eca8e1aa292596af8b1822a25f1220c91a48815ecc2b2f26R256). It is a bit over-estimating though for some numbers. It would probably be worth it polishing it/benchmarking it against directly running the stringify algorithm (writing to `_TotalWritableBytes`). We could start simple by using the existing stringify algorithm.

I think that for integers the value should be easier to estimate. Just as an idea:
```mojo
[...]

PR #4807, opened 2025-06-12
[stdlib] Optimize `Error(*args: *Writable)`

martinvuyk:
Optimize `Error(*args: *Writable)`. Split off from  #4793

### DIFF
--- a/mojo/stdlib/stdlib/builtin/error.mojo
+++ b/mojo/stdlib/stdlib/builtin/error.mojo
@@ -22,6 +22,7 @@ from sys import _libc, external_call, is_gpu
 from sys.ffi import c_char
 
 from memory import ArcPointer, memcpy
+from io.write import _WriteBufferStack, _TotalWritableBytes
 
 
 # ===-----------------------------------------------------------------------===#
@@ -92,6 +93,19 @@ struct StackTrace(ImplicitlyCopyable, Stringable):
 # ===-----------------------------------------------------------------------===#
 
 
+@fieldwise_init
+struct _ErrorWriter(Writer):
+    var data: List[Byte]
+
+    fn write_bytes(mut self, bytes: Span[Byte, _]):
+        self.data.extend(bytes)
+
+    fn write[*Ts: Writable](mut self, *args: *Ts):
+        @parameter
+        for i in range(args.__len__()):
+            args[i].write_to(self)
+
+
 @register_passable
 struct Error(
     Boolable,
@@ -149,61 +163,36 @@ struct Error(
         self.loaded_length = value.byte_length()
         self.stack_trace = StackTrace(depth=0)
 
-    @implicit
-    fn __init__(out self, src: String):
-        """Construct an Error object with a given string.
-
-        Args:
-            src: The error message.
-        """
-        var length = src.byte_length()
-        var dest = UnsafePointer[UInt8].alloc(length + 1)
-        memcpy(dest=dest, src=src.unsafe_ptr(), count=length)
-        dest[length] = 0
-        self.data = dest
-        self.loaded_length = -length
-        self.stack_trace = StackTrace(depth=0)
-
-    @implicit
-    fn __init__(out self, src: StringSlice):
-        """Construct an Error object with a given string ref.
-
-        Args:
-            src: The error message.
-        """
-        var length = src.byte_length()
-        var dest = UnsafePointer[UInt8].alloc(length + 1)
-        memcpy(dest=dest, src=src.unsafe_ptr(), count=length)
-        dest[length] = 0
-        self.data = dest
-        self.loaded_length = -length
-        self.stack_trace = StackTrace(depth=0)
-
     @no_inline
-    fn __init__[
-        *Ts: Writable
-    ](out self, *args: *Ts, sep: StaticString = "", end: StaticString = "",):
-        """
-        Construct an Error by concatenating a sequence of Writable arguments.
+    fn __init__[*Ts: Writable](out self, *args: *Ts):
+        """Construct an Error by concatenating a sequence of Writable arguments.
 
         Args:
             args: A sequence of Writable arguments.
-            sep: The separator used between elements.
-            end: The String to write after printing the elements.
 
         Parameters:
             Ts: The types of the arguments to format. Each type must be satisfy
                 `Writable`.
         """
-        var output = String()
-        var buffer = _WriteBufferStack(output)
 
         @parameter
-        for i in range(args.__len__()):
-            args[i].write_to(buffer)
+        fn _write(mut writer: Some[Writer]):
+            @parameter
+            for i in range(args.__len__()):
+                args[i].write_to(writer)
 
+        # Count the total length of bytes to allocate only once
+        var arg_bytes = _TotalWritableBytes()
+        _write(arg_bytes)
+        arg_bytes.size += 1  # nul terminator
+        var writer = _ErrorWriter(List[Byte](capacity=arg_bytes.size))
+        var buffer = _WriteBufferStack(writer)
+        _write(buffer)
         buffer.flush()
-        self = Error(output)
+        writer.data.append(0)  # nul terminator
+        self.loaded_length = -(len(writer.data) - 1)
+        self.data = writer.data.steal_data()
+        self.stack_trace = StackTrace(depth=0)
 
     fn __del__(deinit self):
         """Releases memory if allocated."""
--- a/max/kernels/src/_cublas/cublas.mojo
+++ b/max/kernels/src/_cublas/cublas.mojo
@@ -35,8 +35,8 @@ alias CUDA_CUBLAS_LIBRARY_PATHS = List[Path](
 )
 
 
-fn _on_error_msg() -> String:
-    return String(
+fn _on_error_msg() -> Error:
+    return Error(
         (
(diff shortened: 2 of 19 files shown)


martinvuyk:
ping @jackos

gryznar:
How this change looks like in benchmark? Could we measure it?

martinvuyk:
@gryznar I don't really want to measure it honestly. It should be similar to the improvement for `String.write()`.
@jackos most probably has to refactor some details due to compile times anyway.

mojo/stdlib/stdlib/builtin/error.mojo
```diff
@@ -77,59 +98,58 @@ struct Error(
         self.data = value.unsafe_ptr()
         self.loaded_length = value.byte_length()
 
+    @always_inline
     @implicit
     fn __init__(out self, src: String):
```
soraros: Why do we need this overload at all? I thought `StringSlice` already covered this.

martinvuyk: Hmm, in its current state it isn't necessary. But previously it used to receive an owned string and steal the buffer. I think that is not achievable anymore with the new String internals. I'll delete it

soraros: Actually, do we even need the string slice overload? (I'm on my phone so I didn't check the implementation in detail.)

martinvuyk:
```text
cannot implicitly convert 'String' value to 'Error'
            raise error
                  ^~~~~
```
Turns out the implicit inference doesn't quite work

soraros: Ah, right, they were there for implicit conversion. Do we want implicit conversion from sting though?

martinvuyk: Hm, I think actually not given it makes an alloc. The `StringSlice` overload as well. I'll remove it in another PR though. It probably is there to enable pythonic raises.

martinvuyk: Nevermind I'm going to do it in this PR

mojo/stdlib/stdlib/builtin/error.mojo
```diff
@@ -92,6 +92,23 @@ struct StackTrace(ImplicitlyCopyable, Stringable):
 # ===-----------------------------------------------------------------------===#
 
 
+@fieldwise_init
+struct _UnsafeWriteToErrorPtr(Writer):
+    var data: UnsafePointer[UInt8]
```
JoeLoser:
```suggestion
    var data: UnsafePointer[Byte]
```

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:

mojo/stdlib/stdlib/builtin/error.mojo
```diff
@@ -149,61 +166,36 @@ struct Error(
...
         for i in range(args.__len__()):
-            args[i].write_to(buffer)
+            args[i].write_to(arg_bytes)
+        var ptr = UnsafePointer[Byte].alloc(arg_bytes.size + 1)
+        var writer = _UnsafeWriteToErrorPtr(ptr, 0, arg_bytes.size)
+        var buffer = _WriteBufferStack(writer)
 
+        @parameter
+        for i in range(args.__len__()):
+            args[i].write_to(buffer)
```
ConnorGray: **Suggestion** I think the optimization to precompute the written bytes is valid, based on our past experience with optimizing things like `String.join()`.

However, I'm unsure that `_UnsafeWriteToErrorPtr` is correct in all cases. Consider a type whose `write_to()` function is nondeterministic. E.g. perhaps there is a type like `Now()` that always formats as the current time. The precomputed byte length from the first loop could differ from the actual number of written bytes in the second loop. We should ensure this code works without "clipping" the written bytes even in cases like that.

I suggest we simply use `List[Byte](capacity=arg_bytes.size + 1)` or a similar `String` here to accumulate the written bytes, and then use `.steal_data()` to take ownership of the pointer, and store that into the `Error` struct. In the common case, the capacity will exactly equal the written bytes. In the less common case, the `List`/`String` will simply re-allocate.

martinvuyk: Hm I didn't think about a nondeterministic case, good observation.

I do wonder however if that is worth considering for the sake of this PR's scope. I think we should open a specific issue and deal with that separately, maybe as a justification for #4249. Because currently `String`'s write methods also do the length estimation run before actually constructing the `String`. I think that rather than reallocating using `String`/`List` we should be looking into adapting the writable trait to give an estimate of the final buffer size.

NathanSWard: For this PR - using a `List`/`String` and then stealing the data is probably the solution I'd push for too. It should also simplify this implementation a little bit (no need for `_UnsafeWriteToErrorPtr`).

Regarding #4249 -> I totally agree with this. Similarly, `Iterator` now has a `bounds()` defaulted trait method added to and likely `Writeable` should have a similar one as well.

ConnorGray: > I do wonder however if that is worth considering for the sake of this PR's scope. I think we should open a specific issue and deal with that separately,

I'm not sure I agree. This PR is introducing the issue with this new helper type; I think it's probably easier and cleaner to simply fix the issue here, instead of introducing it just to fix it later.

I agree though that there are longer-term refactorings we could do to improve the situation, separate from this issue 🙂

Regardless of whether we "plumb through" writer estimations though, it will still be necessary to build this kind of logic on a _growable_ collection, since those will still be just estimations, not guaranteed.

martinvuyk: You guys were right, the implementation is actually much cleaner now :)

mojo/stdlib/stdlib/builtin/error.mojo
```diff
@@ -149,61 +162,36 @@ struct Error(
...
+            for i in range(args.__len__()):
+                args[i].write_to(writer)
+
+        # Count the total length of bytes to allocate only once
+        var arg_bytes = _TotalWritableBytes()
+        _write(arg_bytes)
+        arg_bytes.size += 1  # nul terminator
+        var writer = _ErrorWriter(List[Byte](capacity=arg_bytes.size))
+        var buffer = _WriteBufferStack(writer)
+        _write(buffer)
```
ConnorGray, line 190: **Question** Since we're writing to a list with pre-allocated capacity anyway, are we gaining anything by using this stack buffer? It might be just as efficient to simplify:

```suggestion
        var writer = _ErrorWriter(List[Byte](capacity=arg_bytes.size))
        _write(writer)
```

mojo/stdlib/stdlib/builtin/error.mojo
ConnorGray: Awesome, thank you Martin 🫶

mojo/stdlib/stdlib/builtin/error.mojo
martinvuyk: It made a big difference in benchmarks when I was doing this for `String.write`

ConnorGray:
!sync

JoeLoser:
Thanks for working together on this @martinvuyk and @ConnorGray.  Do you mind rebasing this so I can resync and land it, please?

Per internal review comment, the only API design quibble might be that we could still keep the `Error.__init__(var String)` constructor to avoid an extra allocation in a common case.  WDYT of that, Martin?

martinvuyk:
> Thanks for working together on this @martinvuyk and @ConnorGray. Do you mind rebasing this so I can resync and land it, please?

Yeah I'll do it in a bit, I've got some work stuff then I'll jump into this 

> Per internal review comment, the only API design quibble might be that we could still keep the `Error.__init__(var String)` constructor to avoid an extra allocation in a common case. WDYT of that, Martin?

What allocation is avoided? The previous implementation was implicitly allocating, that is why I removed those constructors

```mojo
    @implicit
    fn __init__(out self, src: String):
        var length = src.byte_length()
        var dest = UnsafePointer[UInt8].alloc(length + 1)
        memcpy(dest, src.unsafe_ptr(), length)
        dest[length] = 0
        self.data = dest
        self.loaded_length = -length
        self.stack_trace = StackTrace(depth=0)
```
I remember trying to make it non-allocating at some point when implementing this PR but I got several errors due to origin shenanigans. I could try again later today

martinvuyk:
@JoeLoser due to the new `String` internals, it is actively worse to use `.unsafe_cstr_ptr()` than directly building an `Error` because the reallocation that will happen anyway when missing a nul terminator. When using `.unsafe_ptr_mut()` (which `.unsafe_cstr_ptr()` uses internally)  it would also reallocate for all strings that have more than 1 reference. The only case where it wouldn't is when a uniquely referenced `String` with a nul terminator is passed in, which IMO is a very rare case given `String`'s write constructor does not add a nul terminator. But even then `Error`'s constructor has to copy the data anyway

The other option is making `Error` track the `String`'s origin, because just using its pointer causes several USAF bugs.

ConnorGray:
!sync

JoeLoser:
!sync

Issue #4794, opened 2025-06-11
[BUG] `float8` type have unclear spec


soraros:
### Bug description

Implementations of functions supporting these types sometimes have bugs. One prominent example is the `float8_e3m4` type. Its docstring says it doesn't support positive or negative infinity (`inf`/`-inf`), and that NaN is encoded as `00111111` and `11111111`. However, two online references disagree:

* **IntelLabs FP8 Emulation Toolkit**: no `inf`/`-inf`; NaN encoded as `S.111.1111`. [ref](https://github.com/IntelLabs/FP8-Emulation-Toolkit)
* **JAX-ML `ml_dtypes`**: supports `inf`/`-inf` per IEEE 754. [ref](https://github.com/jax-ml/ml_dtypes#float8_e3m4)

The `numerics` module also suffers from this unclear spec. Types like `f8e5m2fnuz` don't support `inf` (and treat certain bit patterns as NaN), yet the `inf` function still returns a value for them, and `isnan` ignores them entirely. There are plenty of other examples like this.

I'd like to get these fixed, but without a clear spec we can't know the intended behavior. Tagging @abduld, @jackos, @raiseirql as suggested by @JoeLoser.

### Steps to reproduce

One scary example:

```mojo
from utils.numerics import isnan, nan

fn main():
  n = nan[DType.float8_e4m3fnuz]()
  print(isnan(n))  # prints False
```

### System information

`Mojo 25.4.0.dev2025061005 (636c9749)`

PR #4814, opened 2025-06-13
[stdlib] Fix docstring for `DType.float8_*`

soraros:
Related to #4794. CC @jackos.

- [x] Add ref link for all `float8` types.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/dtype.mojo
+++ b/mojo/stdlib/stdlib/builtin/dtype.mojo
@@ -99,20 +99,25 @@ struct DType(
     alias float8_e3m4 = DType(
         __mlir_attr.`#kgen.dtype.constant<f8e3m4> : !kgen.dtype`
     )
-    """Represents an 8-bit e3m4 floating point format, encoded as `seeemmmm`:
+    """Represents an 8-bit `e3m4` floating point format, encoded as `seeemmmm`:
+
     - (s)ign: 1 bit
     - (e)xponent: 3 bits
     - (m)antissa: 4 bits
     - exponent bias: 3
-    - nan: 00111111, 11111111
-    - -0: 10000000
-    - fn: finite (no inf or -inf encodings)
+    - nan: {0,1}.111.1111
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/builtin/dtype.mojo
```diff
@@ -99,72 +99,78 @@ struct DType(
...
+    """Represents an 8-bit `e3m4` floating point format, encoded as `seeemmmm`:
+
     - (s)ign: 1 bit
     - (e)xponent: 3 bits
     - (m)antissa: 4 bits
     - exponent bias: 3
-    - nan: 00111111, 11111111
-    - -0: 10000000
-    - fn: finite (no inf or -inf encodings)
+    - `nan`: `S.111.1111`
```
jackos: Having the dots here is a good addition for separating the sign, exponent, and mantissa, but prefer not to have the \` marks as it makes the list formatting messier, and also prefer to specify the sign as `{0,1}` instead of `S`, as it's not immediately clear that `S` means 0 or 1

jackos (changes requested):
Cool thanks, the dots to separate sign, exponent, and mantissa are a good addition

mojo/stdlib/stdlib/builtin/dtype.mojo
soraros: I think the idea is `S` means `1` for positive. Do you think it's worth keeping?

jackos: Nah the S from the specs means it can be 1 or 0, that's what I mean in that it's not clear what it means

soraros: I think `S` indeed means `0` for positive and `1` for negative, notice the use for zero values and `inf`.

Screenshot 2025-06-14 at 19 32 14

jackos (approved):
Thanks!

jackos:
!sync

PR #4816, opened 2025-06-13
[stdlib] Make `SIMD.to_bits` return unsigned int

soraros:
- This change ensures that we use unsigned bitwise operations in SIMD, making the behavior consistent with other languages and libraries. It also fixes the cast operation on signed integers, which previously used `sext` and could lead to unexpected results and subtle bugs.
- Add support for `bool`.
- Add tests.


mojo/stdlib/stdlib/hashlib/_ahash.mojo
```diff
@@ -179,20 +179,12 @@ struct AHasher[key: U256](Defaultable, _Hasher):
...
+        alias rounds = max(1, new_data.dtype.sizeof() // 8)
 
         @parameter
         if rounds == 1:
             # vector values are not bigger than 8 bytes each
-            var u64: SIMD[DType.uint64, new_data.size]
-
-            @parameter
-            if new_data.dtype.is_floating_point():
-                u64 = new_data.to_bits().cast[DType.uint64]()
```
soraros, line 193: @mzaks This was subtly incorrect before, because `to_bits` returns a signed integer and `cast` performs a signed extension (`llvm.sext`), which explains the change in the hash value.

JoeLoser (approved):
Thanks for working on this!  I'll get this landed and then we can have @mzaks rebase https://github.com/modular/modular/pull/4863 since this fixes a subtle hashing bug.

JoeLoser:
!sync

PR #4824, opened 2025-06-13
fix compiler warning: Global variables

christoph-schlumpf:
fix compiler warning: Global variables are only partially implemented in Mojo


christoph-schlumpf:

PR #4825, opened 2025-06-13
[kernel] Fix `kl_div` test so for each input `a`, we have `a_i^i = 1`

soraros:
Motivation:

```suggestion
        a[i] = Scalar[dtype](1.0 / len)
        b[i] = Scalar[dtype](2.0 * (i + 1) / (len * (len + 1)))
```
NIT:
I would expect the inputs of KL Div to individually sum to 1.0, here's two which do, 
Since it's floating point math they won't really, but it's less confusing to me.

If it's a hassle to compute a new reference value, feel free to skip this comment

_Originally posted by @tboerstad in https://github.com/modular/modular/pull/4753#discussion_r2137956754_

### DIFF
--- a/max/kernels/test/internal_utils/test_measure.mojo
+++ b/max/kernels/test/internal_utils/test_measure.mojo
@@ -72,14 +72,14 @@ fn test_kl_div() raises:
     var a = InlineArray[Scalar[dtype], len](uninitialized=True)
     var b = InlineArray[Scalar[dtype], len](uninitialized=True)
     for i in range(len):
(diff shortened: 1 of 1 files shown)


tboerstad:
Thanks for the follow up 🙏 
Edit: Oops, didn't realize this was a draft PR, sorry

I'll hold up on our side until it's published

tboerstad:
!sync

soraros:
@tboerstad The OSS CI seems to be happy with this, so I think it's good to go.

PR #4827, opened 2025-06-13
[docs] Update mojo-from-python.mdx     Code line seems out of place.

ScottzCodez:
This code line seems out of place and it's not explained: x.attr = z

URL =https://docs.modular.com/mojo/manual/python/mojo-from-python#writing-python-in-mojo


ScottzCodez:
I can't figure out how to fix this error:
Check PR title / Check PR title format (pull_request).  Copilot suggested adding the text   [docs] <comment> 
which didn't help.   Maybe I'm too new / ignorant?

soraros:
The title should look like `[docs] ...`.

scottamain:
You’re right that it’s not explained but I think this is still relevant because it demonstrates that the Python object’s attributes are plainly accessible and mutable from Mojo.

PR #4830, opened 2025-06-13
[docs] Update types.mdx      typo -- 6 items,  not 4 items

ScottzCodez:
In the program output section:
"4 items in the set"    should be    "6 items in the set"
because the set is:
 var py_set = Python.evaluate('{2, 3, 2, 7, 11, 3}')

https://docs.modular.com/mojo/manual/python/types#mojo-wrapper-objects

### DIFF
--- a/mojo/docs/manual/python/types.mdx
+++ b/mojo/docs/manual/python/types.mdx
@@ -139,7 +139,7 @@ def main():
 ```
 
 ```output
-4 items in the set.
+6 items in the set.
 Is 7 in the set: True
 ```
 


soraros:
I don't think this is correct.

```python
  {2, 3, 2, 7, 11, 3}
=
  {2, 4, 7, 11}
```

It does have 4 items.

scottamain:
This surprised me at first also, but this is a Python set, which does not allow duplicate values. So the duplicate instances of 2 and 3 are eliminated, leaving us with 4 items in the set.

PR #4831, opened 2025-06-14
[docs] Fix vectorize sample code (fixes #4482)

### LINKED ISSUE #4482: [Docs]Sample code does not compile
### Where is the problem?

https://docs.modular.com/mojo/stdlib/algorithm/functional/vectorize/

### What can we do better?

Sample vectorization code does not compile.
APIs may have evolved - following lines in the sample code are problematic:

```mojo
p.store[width=simd_width](i, i) # UnsafePointer does not accept width param
```

```mojo
print(p.load[width=size]()) # size is not a power of 2
```

The sample code needs to be updated.

### Anything else?

_No response_

hardikkgupta:
`UnsafePointer.store/load` no longer accept keyword `width` and non-power-of-two widths
Changed two 'vectorize' doc-strings
Tackles #4482

### DIFF
--- a/mojo/stdlib/stdlib/algorithm/functional.mojo
+++ b/mojo/stdlib/stdlib/algorithm/functional.mojo
@@ -102,7 +102,7 @@ fn vectorize[
 
     ```mojo
     from algorithm.functional import vectorize
-    from memory import UnsafePointer
+    from memory import UnsafePointer, SIMD, simdwidthof
 
     # The amount of elements to loop through
     alias size = 10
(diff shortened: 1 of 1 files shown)


soraros:
Hi @hardikkgupta, thank you for tackling this. I don't think it's the correct fix though. You only need to add [a single line](https://github.com/modular/modular/issues/4482#issuecomment-2972217169).

hardikkgupta:
Oh! let me check and get back @soraros

hardikkgupta:
I thought that `p.store[width=width]` and `p.load[width=…]` were removed a few versions ago; stable expects the positional form `p.store[width](…)`, `p.load[width](…)`
But yes Nightly will succeed in this case, I suppose

PR #4832, opened 2025-06-14
[docs] add simdwidthof import in vectorize sample (fixes #4482)

### LINKED ISSUE #4482: [Docs]Sample code does not compile

hardikkgupta:
Adds `from sys import simdwidthof` to the import block in both `vectorize`
doc-string examples

Tackles #4482


JoeLoser (approved):
LGTM, thank you for the fix!

JoeLoser:
!sync

PR #4840, opened 2025-06-15
[stdlib] Use `List[UInt8]` as data type for `b[16|64][en|de]code` functions

richardkiss:
The `b64decode` and `b16decode` functions should really return a `bytes`-like structure rather than a `str`-like structure. Similarly, the `b64encode` and `b16encode` functions should accept a `bytes`-like structure (in this case, `Span[UInt8]`). This PR fixes that.

### DIFF
--- a/mojo/stdlib/std/base64/_b64encode.mojo
+++ b/mojo/stdlib/std/base64/_b64encode.mojo
@@ -203,16 +203,29 @@ def load_incomplete_simd[
 
 
 @no_inline
-def _b64encode(input_bytes: Span[mut=False, Byte, _], mut result: String):
+def _b64encode(
+    input_bytes: Span[mut=False, Byte, _],
+    output_bytes: Span[mut=True, Byte, _],
+) -> Int:
+    """Writes base64-encoded `input_bytes` into `output_bytes`.
+
+    The output buffer must be at least `4 * ceildiv(len(input_bytes), 3)`
+    bytes. Returns the number of bytes written.
+    """
+    debug_assert(
+        len(output_bytes) >= 4 * ceildiv(len(input_bytes), 3),
+        "output buffer too small for base64 encoding: need ",
+        4 * ceildiv(len(input_bytes), 3),
+        ", got ",
+        len(output_bytes),
+    )
     comptime simd_width = sys.simd_byte_width()
     comptime input_simd_width = simd_width * 3 // 4
     comptime equal_vector = SIMD[DType.uint8, simd_width](ord("="))
 
-    # 4 character bytes for each 3 bytes (or less) block
-    result.resize(unsafe_uninit_length=4 * ceildiv(len(input_bytes), 3))
     var input_bytes_len = len(input_bytes)
     var input_index = 0
-    var res_ptr = result.unsafe_ptr_mut()
+    var res_ptr = output_bytes.unsafe_ptr()
     var res_offset = 0
 
     # Main loop
@@ -266,7 +279,7 @@ def _b64encode(input_bytes: Span[mut=False, Byte, _], mut result: String):
         res_offset += nb_of_elements_to_store
         input_index += input_simd_width
 
-    result.resize(res_offset)
+    return res_offset
 
 
 # Utility functions
--- a/mojo/stdlib/std/base64/base64.mojo
+++ b/mojo/stdlib/std/base64/base64.mojo
@@ -20,6 +20,7 @@ from std.base64 import b64encode
 """
 
 
+from std.math import ceildiv
 from std.memory import Span
 
 from ._b64encode import _b64encode
@@ -79,18 +80,40 @@ def _ascii_to_value[validate: Bool = False](char: Byte) raises -> Byte:
 
 
 @always_inline
-def b64encode(input_bytes: Span[mut=False, Byte, _], mut result: String):
-    """Performs base64 encoding on the input string.
+def unsafe_b64encode(
+    input_bytes: Span[mut=False, Byte, _],
+    output_bytes: Span[mut=True, Byte, _],
+) -> Int:
+    """Performs base64 encoding from input bytes to output bytes.
+
+    The caller must ensure the output buffer is at least
+    `4 * ceildiv(len(input_bytes), 3)` bytes; no bounds check is performed.
 
     Args:
-        input_bytes: The input string buffer.
-        result: The string in which to store the values.
+        input_bytes: The input bytes to encode.
+        output_bytes: The output buffer to write encoded bytes into.
 
-    Notes:
-        This method reserves the necessary capacity. `result` can be a 0
-        capacity string.
+    Returns:
+        The number of bytes written to output_bytes.
     """
-    _b64encode(input_bytes, result)
+    return _b64encode(input_bytes, output_bytes)
+
+
+@always_inline
+def b64encode(input_bytes: Span[mut=False, Byte, _], mut result: String):
+    """Performs base64 encoding, writing into a reusable `String` buffer.
+
+    `result` is resized to fit the encoded output; existing contents are
+    overwritten. Capacity is reused when possible, so calling this in a
+    loop with the same `result` avoids per-call allocation.
+
+    Args:
+        input_bytes: The input bytes to encode.
+        result: The string buffer to write into.
+    """
+    var output_len = 4 * ceildiv(len(input_bytes), 3)
+    result.resize(unsafe_uninit_length=output_len)
+    _ = unsafe_b64encode(input_bytes, result.unsafe_as_bytes_mut())
 
 
 @always_inline
@@ -108,10 +131,10 @@ def b64encode(input_string: StringSlice[mut=False, _]) -> String:
 
 @always_inline
 def b64encode(input_bytes: Span[mut=False, Byte, _]) -> String:
-    """Performs base64 encoding on the input string.
+    """Performs base64 encoding on the input bytes.
 
     Args:
-        input_bytes: The input string buffer.
+        input_bytes: The input bytes to encode.
 
     Returns:
         The ASCII base64 encoded string.
@@ -126,30 +149,45 @@ def b64encode(input_bytes: Span[mut=False, Byte, _]) -> String:
 # ===-----------------------------------------------------------------------===#
 
 
-def b64decode[
+def unsafe_b64decode[
     *, validate: Bool = False
 ](
-    input_bytes: Span[mut=False, Byte, _],
+    str: StringSlice[mut=False, _],
     mut output_bytes: Span[mut=True, Byte, _],
 ) raises -> Int:
-    """Performs base64 decoding from input bytes to output bytes.
+    """Performs base64 decoding from a string to output bytes.
 
-    The output buffer must be at least (input length * 3) // 4 bytes.
+    The caller must ensure the output buffer is at least
+    `(len(str) * 3) // 4` bytes; no bounds check is performed.
 
     Parameters:
         validate: If true, the function will validate the input and that the
             output buffer is large enough.
 
     Args:
-        input_bytes: The input base64 encoded bytes.
+        str: A base64 encoded string.
         output_bytes: The output buffer to write decoded bytes into.
 
     Returns:
         The number of bytes written to output_bytes.
+
+    Raises:
+        If `validate` is true and the input length is not a multiple of 4,
+        the output buffer is too small, or the input contains non-base64
+        characters.
     """
     comptime `=` = Byte(ord("="))
-    var n = len(input_bytes)
+    var input_bytes = str.as_bytes()
+    var n = str.byte_length()
     var write_pos = 0
+
+    debug_assert(
+        len(output_bytes) >= n * 3 // 4,
+        "output buffer too small for base64 decoding: need ",
+        n * 3 // 4,
+        ", got ",
+        len(output_bytes),
+    )
 
     comptime if validate:
         if n % 4 != 0:
@@ -190,9 +228,37 @@ def b64decode[
     return write_pos
 
 
+@always_inline
 def b64decode[
     *, validate: Bool = False
-](str: StringSlice[mut=False, _]) raises -> List[UInt8]:
+](str: StringSlice[mut=False, _], mut result: List[Byte]) raises:
+    """Performs base64 decoding into a reusable `List[Byte]` buffer.
+
+    `result` is resized to fit the decoded output; existing contents are
+    overwritten. Capacity is reused when possible, so calling this in a
+    loop with the same `result` avoids per-call allocation.
+
+    Parameters:
+        validate: If true, the function will validate the input string.
+
+    Args:
+        str: A base64 encoded string.
+        result: The byte buffer to write into.
+
+    Raises:
+        If `validate` is true and the input is not valid base64.
+    """
+    var output_len = str.byte_length() * 3 // 4
+    result.resize(unsafe_uninit_length=output_len)
+    var output_span = Span(result)
+    var written = unsafe_b64decode[validate=validate](str, output_span)
+    result.resize(unsafe_uninit_length=written)
+
+
+@always_inline
+def b64decode[
+    *, validate: Bool = False
+](str: StringSlice[mut=False, _]) raises -> List[Byte]:
     """Performs base64 decoding on a string, returning decoded bytes.
 
     Parameters:
@@ -202,15 +268,13 @@ def b64decode[
         str: A base64 encoded string.
 
     Returns:
-        The decoded bytes as a List[UInt8].
+        The decoded bytes as a List[Byte].
+
+    Raises:
+        If `validate` is true and the input is not valid base64.
     """
-    var data = str.as_bytes()
-    var n = str.byte_length()
-    var result = List[UInt8](capacity=n * 3 // 4)
-    result.resize(n * 3 // 4, 0)
-    var output_span = Span(result)
-    var written = b64decode[validate=validate](data, output_span)
-    result.resize(written, 0)
+    var result = List[Byte]()
+    b64decode[validate=validate](str, result)
     return result^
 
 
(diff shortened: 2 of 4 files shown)


martinvuyk (changes requested):
I'm sorry but I don't agree with your base assumption, that this needs to return and receive bytes. It is specifically working to and from UTF-8 so it should be wrapped in a type that shows that (`StringSlice` and `String`). There is also currently no performant way to go from an owned `List` to a `String` without copying the whole thing, thus this PR also removes the existing performance.

I would rather this remains using `StringSlice` also because of another reason:  #3988. I would like encodings like base64 to actually be implemented to interop with other encodings like e.g. UTF-16, UTF-32, ASCII, ISO 8859-1, etc.

richardkiss:
> I'm sorry but I don't agree with your base assumption, that this needs to return and receive bytes. It is specifically working to and from UTF-8 so it should be wrapped in a type that shows that (`StringSlice` and `String`). There is also currently no performant way to go from an owned `List` to a `String` without copying the whole thing, thus this PR also removes the existing performance.
> 
> I would rather this remains using `StringSlice` also because of another reason: #3988. I would like encodings like base64 to actually be implemented to interop with other encodings like e.g. UTF-16, UTF-32, ASCII, ISO 8859-1, etc.

Hi @martinvuyk and thanks for your review! This is my first non-trivial work on this so I'm still getting acquainted with how things work (and obviously this PR is not ready since it doesn't even pass tests).

What I do think as being non-controversial is that `bXXencode` functions should accept binary data. That's their whole point! My understand of `StringSlice` (and this may be wrong) is that its closest python analog is `str`, ie. it represents Unicode code points (or possibly other codings). This is different from binary data, which can be an arbitrary sequence of bytes! Even sequences that do not represent any valid unicode (or any other encoding!). So something else, like `Span[UInt8]` (which I think of as `bytes` or `bytearray`, or rust's slice `&[u8]`) is required so that you can actually pass in binary blobs that cannot be represented as a `StringSlice`. Of course, the things that come OUT of `bXXencode` are string-like. Am I misunderstanding something here?

richardkiss:
Allowing `bXXencode` to accept `Span[UInt8]` still allows it to operate on `StringSlice`, since we can just go `.as_bytes()`. But we can't always go the other way!!

I can see your point with `.decode` that we may want to allow a direct write of blob data to `String`-like entity to prevent having to do a `memcpy`. However, we can't safely do that to a `String`-like entity that guarantees valid encodings because `.decode` can produces *any* binary output, including invalid encodings! So maybe we can continue to support "unsafe" api that decodes directly to a string-like thing. But IMO the base api should be one that decodes to a binary-like thing. If there is (unsafe) api to turn a `Span[UInt8]` into a `StringSpan` in a 0-copy way, that could be the final step (and it would have uses beyond this application, in any place where we know through some other guarantee that we're not going to get a garbage byte string).

Again, please let me know if I'm misunderstanding something.

martinvuyk:
Well I think you're right, they should work on binary data that might not be valid UTF-8 as well.

But I think we should approach this differently. Rather than changing everything to bytes, what if we (in several separate PRs):
- create an overload for `b64encode(data: Span[mut=False, Byte]) -> List[Byte]` (b16 as well) using the existing string impl
- create an overload for `b64decode(data: Span[mut=False, Byte]) -> List[Byte]` (b16 as well) using the existing string impl
- generalize the existing implementations to reuse code but keep optimum perf for building a `String`

richardkiss:
I can definitely be sensitive to your use case, which seems to be using b64/b16 for using on UTF-8 encoded strings. I don't think `StringSpan` can be the base case though because at some point `StringSlice` is supposed to have the guarantee that it represents UTF-8 code points (I think?). I read your linked proposal in #3988 which look great, and you have thought about this a lot more than I have. I'm pretty sure we can build the `StringSlice` `encode` implementation on the `Span` implementation without a single excess CPU cycle, and I can rewrite the `decode` implementation either overloaded or generic so `Span` and `StringSlice` has the same implementation with the same speed. I'll rework this patch to ensure no old api is removed or slowed since it seems to be important to you (which means it's likely important to others).

richardkiss:

martinvuyk:
> I don't think StringSpan can be the base case though because at some point StringSlice is supposed to have the guarantee that it represents UTF-8 code points (I think?)

This is correct. We do need an overload for bytes.

>  I'm pretty sure we can build the StringSlice encode implementation on the Span implementation without a single excess CPU cycle, I can rewrite the `decode` implementation either overloaded or generic so `Span` and `StringSlice` has the same implementation with the same speed

I have some ideas on how to make the underlying implementations generic. I think it is best if you constraint this PR to only adding some overloads reusing the existing `String`/`StringSlice` implementations. It is faster for the Mojo team to review, and overall easier to catch mistakes/improvement opportunities.

richardkiss:
> > I don't think StringSpan can be the base case though because at some point StringSlice is supposed to have the guarantee that it represents UTF-8 code points (I think?)
> 
> This is correct. We do need an overload for bytes.
> 
> > I'm pretty sure we can build the StringSlice encode implementation on the Span implementation without a single excess CPU cycle, I can rewrite the `decode` implementation either overloaded or generic so `Span` and `StringSlice` has the same implementation with the same speed
> 
> I have some ideas on how to make the underlying implementations generic. I think it is best if you constraint this PR to only adding some overloads reusing the existing `String`/`StringSlice` implementations. It is faster for the Mojo team to review, and overall easier to catch mistakes/improvement opportunities.

@martinvuyk I tried a couple things to make things generic over both `List[UInt8]` and `String` but it seems like it'll have to wait until you can pass in capturing functions as either parameters or generics.

There is one more issue though: we have to choose one of `String` and `List[UInt8]` as the return value of `b64decode` (and `16`) as we can't parameterize on return type only. Because `List[UInt8]` is more fundamental than `String`, I think we have to use that and maybe create a new thing like `b64decode_to_string` for `String`.

martinvuyk:
I would advice you to concentrate on just building an overload using the existing implementations first. Then, in another PR, we can add a generic version to avoid the back and forth copying.

My idea for the  generic version (for the next PR, not this one) is more or less:
```mojo
fn _b64encode[T: Variant[String, List[Byte]](input_bytes: Span[mut=False, Byte], mut result: T): ...
```

mojo/stdlib/std/base64/_b64encode.mojo
```diff
@@ -203,16 +203,29 @@ def load_incomplete_simd[
...
     comptime simd_width = sys.simd_byte_width()
     comptime input_simd_width = simd_width * 3 // 4
     comptime equal_vector = SIMD[DType.uint8, simd_width](ord("="))
 
-    # 4 character bytes for each 3 bytes (or less) block
-    result.resize(unsafe_uninit_length=4 * ceildiv(len(input_bytes), 3))
     var input_bytes_len = len(input_bytes)
     var input_index = 0
-    var res_ptr = result.unsafe_ptr_mut()
+    var res_ptr = output_bytes.unsafe_ptr()
```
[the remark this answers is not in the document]
richardkiss: Not a real issue. In Mojo, `UnsafePointer` is inherently unsafe — `.store()` writes through the raw address regardless of the `mut=` annotation on the enclosing `Span`. The mutability annotation is a compile-time access control on the `Span` itself, not a C++ const-pointer constraint on the underlying pointer. The code compiles and all tests pass.

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -79,18 +80,43 @@ def _ascii_to_value[validate: Bool = False](char: Byte) raises -> Byte:
...
+    overwritten. Capacity is reused when possible, so calling this in a
+    loop with the same `result` avoids per-call allocation.
+
+    Args:
+        input_bytes: The input bytes to encode.
+        result: The string buffer to write into.
+    """
+    var output_len = 4 * ceildiv(len(input_bytes), 3)
+    result.resize(unsafe_uninit_length=output_len)
+    var output_span = Span[Byte, origin_of(result)](
```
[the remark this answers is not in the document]
richardkiss: Not a real issue. The `Span` is constructed with `ptr=result.unsafe_ptr_mut()`, so Mojo infers the span's mutability from the pointer type — no explicit `mut=True` annotation is needed. The code compiles and all tests pass.

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -178,6 +284,58 @@ def b64decode[
...
+    Args:
+        input_bytes: The input bytes to encode.
+        result: The string buffer to write into.
+    """
+    var output_len = 2 * len(input_bytes)
+    result.resize(unsafe_uninit_length=output_len)
+    var output_span = Span[Byte, origin_of(result)](
+        ptr=result.unsafe_ptr_mut(), length=output_len
+    )
+    _ = b16encode(input_bytes, output_span)
```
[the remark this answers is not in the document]
richardkiss: Not a real issue on either point. (1) Same as the other span comment — mutability is inferred from `unsafe_ptr_mut()`, no explicit annotation needed. (2) For b16, `output_len = 2 * len(input_bytes)` is always exact and the encoder returns exactly `length * 2`, so there is nothing to resize. Discarding the return value here is intentional — contrast with b64encode where padding means the written count can vary. The code compiles and all tests pass.

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -79,18 +80,43 @@ def _ascii_to_value[validate: Bool = False](char: Byte) raises -> Byte:
...
+    Args:
+        input_bytes: The input bytes to encode.
+        result: The string buffer to write into.
+    """
+    var output_len = 4 * ceildiv(len(input_bytes), 3)
+    result.resize(unsafe_uninit_length=output_len)
+    var output_span = Span[Byte, origin_of(result)](
+        ptr=result.unsafe_ptr_mut(), length=output_len
+    )
+    var written = b64encode(input_bytes, output_span)
```
martinvuyk:
```suggestion
    var written = b64encode(input_bytes, result.unsafe_as_bytes_mut())
```

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -79,18 +80,43 @@ def _ascii_to_value[validate: Bool = False](char: Byte) raises -> Byte:
...
+    `result` is resized to fit the encoded output; existing contents are
+    overwritten. Capacity is reused when possible, so calling this in a
+    loop with the same `result` avoids per-call allocation.
+
+    Args:
+        input_bytes: The input bytes to encode.
+        result: The string buffer to write into.
+    """
+    var output_len = 4 * ceildiv(len(input_bytes), 3)
+    result.resize(unsafe_uninit_length=output_len)
```
martinvuyk, line 115: resize sets a fill byte (default 0) to the memory which we don't need here
also, since we're allowing the mutation of an incoming string instead of creating a new one we need to reserve/resize adding the current byte length of the string.
```suggestion
    result.reserve(result.byte_length() + output_len)
```

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -128,48 +154,128 @@ def b64encode(input_bytes: Span[mut=False, Byte, _]) -> String:
...
+
+    Args:
+        str: A base64 encoded string.
+        result: The byte buffer to write into.
+
+    Raises:
+        If `validate` is true and the input is not valid base64.
+    """
+    var output_len = str.byte_length() * 3 // 4
+    result.resize(output_len, 0)
```
martinvuyk:
```suggestion
    result.reserve(len(result) + output_len)
```

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -128,48 +154,128 @@ def b64encode(input_bytes: Span[mut=False, Byte, _]) -> String:
...
+    var output_span = Span(result)
+    var written = b64decode[validate=validate](str, output_span)
+    result.resize(written, 0)
```
martinvuyk: resize only uses the fill byte when the size is larger than the current size
```suggestion
    result.resize(written)
```

mojo/stdlib/std/base64/base64.mojo
martinvuyk:
```suggestion
    _ = b16encode(input_bytes, result.unsafe_as_bytes_mut())
```

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -209,24 +368,25 @@ def b16encode(str: StringSlice[mut=False, _]) -> String:
...
     """
-
     comptime `A` = Byte(ord("A"))
     comptime `a` = Byte(ord("a"))
     comptime `Z` = Byte(ord("Z"))
     comptime `z` = Byte(ord("z"))
     comptime `0` = Byte(ord("0"))
     comptime `9` = Byte(ord("9"))
 
-    # TODO: Measure perf against lookup table approach
```
martinvuyk, line 229: why did you delete this?

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -209,24 +368,25 @@ def b16encode(str: StringSlice[mut=False, _]) -> String:
 # ===-----------------------------------------------------------------------===#
 
 
-def b16decode(str: StringSlice[mut=False, _]) -> String:
-    """Performs base16 decoding on the input string.
+def b16decode(
+    str: StringSlice[mut=False, _],
+    mut output_bytes: Span[mut=True, Byte, _],
+):
```
martinvuyk: I think we should rename this function as unsafe since we assume the size is reserved beforehand

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -178,6 +284,58 @@ def b64decode[
 # ===-----------------------------------------------------------------------===#
 
 
+def b16encode(
+    input_bytes: Span[mut=False, Byte, _],
+    output_bytes: Span[mut=True, Byte, _],
+) -> Int:
```
martinvuyk: same here, we should rename this to `unsafe_b16encode`

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -128,48 +154,128 @@ def b64encode(input_bytes: Span[mut=False, Byte, _]) -> String:
 
 def b64decode[
```
martinvuyk: same here, we should rename it to unsafe_...

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -79,18 +80,43 @@ def _ascii_to_value[validate: Bool = False](char: Byte) raises -> Byte:
 
 
 @always_inline
-def b64encode(input_bytes: Span[mut=False, Byte, _], mut result: String):
-    """Performs base64 encoding on the input string.
+def b64encode(
+    input_bytes: Span[mut=False, Byte, _],
+    output_bytes: Span[mut=True, Byte, _],
+) -> Int:
```
martinvuyk: and here as well rename it to unsafe since we assume the size is correctly reserved

martinvuyk (changes requested):
Hi, this looks much better. My only nit is that I think we should rename some of these functions prepending them with `unsafe` since they expect the buffers to be of a certain size and otherwise they read/write OOB. We could also just make them private

richardkiss (commented):
~~All your review comments addressed - see inline replies. Requesting re-review.~~

Sorry, overzealous agent did this automatically. Not ready yet.

mojo/stdlib/std/base64/base64.mojo
richardkiss: Done — switched to `unsafe_b64encode(input_bytes, result.unsafe_as_bytes_mut())`.

mojo/stdlib/std/base64/base64.mojo
richardkiss: This is also using `unsafe_uninit_length` now.

mojo/stdlib/std/base64/base64.mojo
richardkiss: Done — used `result.resize(unsafe_uninit_length=written)` to trim without a fill write.

mojo/stdlib/std/base64/base64.mojo
richardkiss: Done — switched to `unsafe_b16encode(input_bytes, result.unsafe_as_bytes_mut())`.

mojo/stdlib/std/base64/base64.mojo
richardkiss: Done — renamed to `unsafe_b64encode`.

mojo/stdlib/std/base64/base64.mojo
richardkiss: Done — renamed to `unsafe_b16encode`.

mojo/stdlib/std/base64/base64.mojo
richardkiss: Done — renamed to `unsafe_b64decode`.

mojo/stdlib/std/base64/base64.mojo
richardkiss: Done — renamed to `unsafe_b16decode`.

mojo/stdlib/std/base64/base64.mojo
richardkiss: To be clear, `resize` with `unsafe_uninit_length` avoids the unnecessary `memset`.

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -126,50 +148,131 @@ def b64encode(input_bytes: Span[mut=False, Byte, _]) -> String:
...
 
     Raises:
-        If the operation fails.
+        If `validate` is true and the input length is not a multiple of 4,
+        the output buffer is too small, or the input contains non-base64
+        characters.
     """
     comptime `=` = Byte(ord("="))
-    var data = str.as_bytes()
+    var input_bytes = str.as_bytes()
```
richardkiss, line 180: Claude changed the name of this, and although it's gratuitous, it probably makes the code a tad easier to comprehend since `data` is a bit ambiguous.

mojo/stdlib/std/base64/base64.mojo
richardkiss: It's back now.

richardkiss:
Okay, I think it's ready now.

mojo/stdlib/std/base64/base64.mojo
martinvuyk: true I forgot about that second signature, and reserve doesn't set the byte length so getting the span out of it wouldn't work

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -79,18 +80,40 @@ def _ascii_to_value[validate: Bool = False](char: Byte) raises -> Byte:
...
+    overwritten. Capacity is reused when possible, so calling this in a
+    loop with the same `result` avoids per-call allocation.
+
+    Args:
+        input_bytes: The input bytes to encode.
+        result: The string buffer to write into.
+    """
+    var output_len = 4 * ceildiv(len(input_bytes), 3)
+    result.resize(unsafe_uninit_length=output_len)
+    _ = unsafe_b64encode(input_bytes, result.unsafe_as_bytes_mut())
```
martinvuyk, line 116: you should resize the string to the final actually written bytes here, the byte reservation above is just an estimate

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -178,6 +283,57 @@ def b64decode[
...
+    overwritten. Capacity is reused when possible, so calling this in a
+    loop with the same `result` avoids per-call allocation.
+
+    Args:
+        input_bytes: The input bytes to encode.
+        result: The string buffer to write into.
+    """
+    var output_len = 2 * len(input_bytes)
+    result.resize(unsafe_uninit_length=output_len)
+    _ = unsafe_b16encode(input_bytes, result.unsafe_as_bytes_mut())
```
martinvuyk, line 333: same here, need to resize the string to what was actually written

martinvuyk (commented):
Other than the two missing string resizes at the end of the encodings, LGTM

mojo/stdlib/std/base64/base64.mojo
richardkiss: For encoding, the estimate is always exactly right, as b64 encoding always pads up to multiple of four bytes, using `=` to indicate premature EOF.

mojo/stdlib/std/base64/base64.mojo
richardkiss: For b16, it's even easier -- encoded is always an even number of bytes, and always twice as long as the raw data.

mojo/stdlib/std/base64/base64.mojo
martinvuyk: could you at least add an assert to ensure that logic is correct?

PR #4850, opened 2025-06-17
[docs] Update roadmap.md -- mistake in code ?

ScottzCodez:
I think you meant this instead:
for var i in list:
    i += 1  # i is a copy; changing i doesn't update the list

for ref i in list:
    i += 1  # i is a mutable reference; changing i updates the list 

https://docs.modular.com/mojo/roadmap#scoping-and-mutability-of-statement-variables


scottamain (approved):
Good catch!

scottamain:
!sync

PR #4812, opened 2025-06-12
[stdlib] Apply trivial cleanups to `numerics.mojo` to prepare for the `float8` fixes

soraros:
Related to #4794. CC @jackos. Please review by commits.

PR #4852, opened 2025-06-17
[stdlib] Fix int formatting to not go out of bounds, add MIN and MAX for 128 and 256 sized integers

sstadick:
Fixes issue reported here: https://forum.modular.com/t/segmentation-fault-with-dict/1656/9 and identified by @soraros here: https://forum.modular.com/t/uint256-returns-wrong-value/1643/6

When creating a buffer to write the string, it was previously hardcoded to 64+1. This right-sizes it to `dtype.bitwidth() + 1`. Additionally, this adds support for MIN and MAX on 128/256 integers and tests for format_int on 128/256.

There's another todo here, which is that converting an `IntLiteral` to a `SIMD[...]` looks like it goes through a `si128`. So IntLiterals outside the range of si128 can't be converted easily. I've worked around that in this MR, but a better solution is probably warranted. 

I'd be open to fixing that if someone can point me in the right direction since it would require something larger than a 256, or more complex logic.

edit - the ubuntu test runner still failed when I accidentally still had the small dict example from the forum post in the tests. So this doesn't completely fix the all issues there.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/format_int.mojo
+++ b/mojo/stdlib/stdlib/builtin/format_int.mojo
@@ -331,8 +331,7 @@ fn _try_write_int[
     # Create a buffer to store the formatted value
 
     # Stack allocate enough bytes to store any formatted 64-bit integer
-    # TODO: use a dynamic size when #2194 is resolved
-    alias CAPACITY: Int = 64 + 1  # +1 for storing NUL terminator.
+    alias CAPACITY: Int = dtype.bitwidth() + 1  # +1 for storing NUL terminator.
 
     var buf = InlineArray[UInt8, CAPACITY](uninitialized=True)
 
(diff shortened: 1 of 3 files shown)


soraros:
Haha, I already made #4800. There's also #4049 where I have some ongoing work for the literal issue.

sstadick:
> Haha, I already made #4800. There's also #4049 where I have some ongoing work for the literal issue.

Dang! I'm not sure how I missed that when searching to see if you had a PR up already! Well I'm happy to close all of mine, or just part of it in favor of either of those two.

soraros:
I think the `MIN`/`MAX` changes are good, though I have some ongoing refactoring in `numerics.mojo` (#4812) and is slowly working through fixing the literal conversion and implement the constants properly. So if you are ok with it, I think we can close this for the moment.

PR #4858, opened 2025-06-18
[max] [stdlib] Fix consistent `dtype: DType` usage

martinvuyk:
Fix consistent `dtype: DType` usage. Part of #4215

### DIFF
--- a/max/kernels/benchmarks/gpu/bench_allreduce.mojo
+++ b/max/kernels/benchmarks/gpu/bench_allreduce.mojo
@@ -60,15 +60,15 @@ fn _human_memory(size: Int) -> String:
 
 # TODO: convert 'ngpus' to runtime variable
 fn bench_reduce[
-    type: DType, rank: Int, ngpus: Int, max_num_blocks: Int
+    dtype: DType, rank: Int, ngpus: Int, max_num_blocks: Int
 ](mut m: Bench, list_of_ctx: List[DeviceContext], num_bytes: Int) raises:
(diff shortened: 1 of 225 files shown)


martinvuyk:
@JoeLoser could we merge this ASAP? I don't really want to deal with the conflicts

JoeLoser:
!sync

JoeLoser (approved):
LGTM, thanks for working on this @martinvuyk!  I've synced this in and am working through some missed spots to get internal CI passing so we can land this.

In a follow up, it would be good to change `NDBuffer` itself to remove the `alias type = dtype` intermediate state as there will be a lot of fallout from that.

martinvuyk:
Yeah I figured we'd better leave that for another time. Sadly we don't yet have the tools to just use `F2` in an IDE to change all the occurrences.

JoeLoser:
> Yeah I figured we'd better leave that for another time. Sadly we don't yet have the tools to just use `F2` in an IDE to change all the occurrences.

I can sympathize with that completely, especially since the days of NDBuffer are numbered with work from @lsh in moving use over to LayoutTensor.

It would be great for the LSP to support rename all occurrences. There's an old internal issue I filed for that a long time ago.

JoeLoser:
FYI I've fixed the missed cases to get everything to build, but this is causing some issues with the graph compiler in one of the lowering passes. I'm out of office today and tomorrow, but I'll ping the graph compiler team to take a look next week.

JoeLoser:
@martinvuyk this just landed internally! 🎉

PR #4863, opened 2025-06-19
[stdlib] Migrates the hash function to use hasher based hashing

mzaks:
This PR makes a big step towards hasher based hashing. As the next step we would need to parametric the Dict data structure with a Hasher type. Currently we default to Fnv1a as it is simple, relatively fast, provides an ok-ish hash quality and works at compile time.


mojo/stdlib/benchmarks/hashlib/bench_hash.mojo
```diff
@@ -14,23 +14,16 @@
...
+from hashlib._ahash import AHasher
+from hashlib._fnv1a import Fnv1a
+from hashlib._djbx33a import DJBX33A
 
 from benchmark import Bench, BenchConfig, Bencher, BenchId, keep
 from bit import byte_swap, rotate_bits_left
 
 # Source: https://www.101languages.net/arabic/most-common-arabic-words/
-alias words_ar = """
+alias words_ar: String = """
```
JoeLoser, line 26: **Question (non-blocking)** Is this and the below specific type annotations strictly required?  Not a biggie either way since they are `StringLiteral` before.

JoeLoser (approved):
Awesome work, @mzaks - thanks for continuing in this area!  I'm ready to sync this in and land it.  Do you mind also adding a changelog entry for this?  Specifically, this will require users updating their `__hash__` functions to use the hasher bashed hashing/dataflow hashing.

No longer does it suffice to just implement `fn __hash__(self) -> UInt:` for your type like user code will have today, right?

JoeLoser:
!sync

mojo/docs/changelog.md
```diff
@@ -38,6 +38,10 @@ what we publish.
   declared.
 
 ### Standard library changes
+- The `Hashable` trait has been updated to use a new data flow / Hasher-based strategy.
```
soraros: Nit: I think we should either not capitalise "hasher" or wrap it in backticks.

mojo/stdlib/benchmarks/hashlib/bench_hash.mojo
mzaks: they were mixed before (some were String some were StringLiteral), so I harmonised it and went with String, but AFAIK it should be ok as `StringLiteral` as well.

mojo/docs/changelog.md
mzaks: yeah that's a good point.

JoeLoser:
!sync

JoeLoser (commented):
Can you please also update the `hash` call site [here](https://github.com/modular/modular/blob/76c388201048989e1ede2fa4556b4ead29ecf5f9/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo#L94) for the new API?  In the next nightly, we move these Python-Mojo bindings tests off of lit and allow them to run with Bazel externally if that's helpful to wait on.

mzaks:
> Can you please also update the `hash` call site [here](https://github.com/modular/modular/blob/76c388201048989e1ede2fa4556b4ead29ecf5f9/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo#L94) for the new API? In the next nightly, we move these Python-Mojo bindings tests off of lit and allow them to run with Bazel externally if that's helpful to wait on.

I think I fixed the issue, sadly I was not able to run the tests on my machine, but generally as far as I can tell there was only the issue of String not having the `__hash__()` method any more.

JoeLoser:
!sync

JoeLoser:
> > Can you please also update the `hash` call site [here](https://github.com/modular/modular/blob/76c388201048989e1ede2fa4556b4ead29ecf5f9/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo#L94) for the new API? In the next nightly, we move these Python-Mojo bindings tests off of lit and allow them to run with Bazel externally if that's helpful to wait on.
> 
> I think I fixed the issue, sadly I was not able to run the tests on my machine, but generally as far as I can tell there was only the issue of String not having the `__hash__()` method any more.

No worries, this was close.  There's just one remaining error it looks like:

```mojo
open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:99:55: error: invalid call to 'PyLong_FromSsize_t': method argument #0 cannot be converted from 'SIMD[uint64, 1]' to 'Int'
        var curr_hash_obj = cpython.PyLong_FromSsize_t(curr_hash)
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:1:1: note: function declared here
^
open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:100:35: error: invalid call to 'PyList_SetItem': method argument #2 cannot be converted from '!lit.typecheckerror' to 'PyObjectPtr'
        _ = cpython.PyList_SetItem(result_py_list, block_idx, curr_hash_obj)
            ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:1:1: note: function declared here
```

which is two things (signedness) and SIMD vs Scalar thing.

JoeLoser:
!sync

mzaks:
> > > Can you please also update the `hash` call site [here](https://github.com/modular/modular/blob/76c388201048989e1ede2fa4556b4ead29ecf5f9/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo#L94) for the new API? In the next nightly, we move these Python-Mojo bindings tests off of lit and allow them to run with Bazel externally if that's helpful to wait on.
> > 
> > 
> > I think I fixed the issue, sadly I was not able to run the tests on my machine, but generally as far as I can tell there was only the issue of String not having the `__hash__()` method any more.
> 
> No worries, this was close. There's just one remaining error it looks like:
> 
> ```mojo
> open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:99:55: error: invalid call to 'PyLong_FromSsize_t': method argument #0 cannot be converted from 'SIMD[uint64, 1]' to 'Int'
>         var curr_hash_obj = cpython.PyLong_FromSsize_t(curr_hash)
>                             ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
> open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:1:1: note: function declared here
> # ===----------------------------------------------------------------------=== #
> ^
> open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:100:35: error: invalid call to 'PyList_SetItem': method argument #2 cannot be converted from '!lit.typecheckerror' to 'PyObjectPtr'
>         _ = cpython.PyList_SetItem(result_py_list, block_idx, curr_hash_obj)
>             ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> open-source/max/mojo/integration-test/python-extension-modules/block-hasher/mojo_module.mojo:1:1: note: function declared here
> ```
> 
> which is two things (signedness) and SIMD vs Scalar thing.

Got it, from what I see and tried out a simple cast to `Int` should fix it. Pushed the fix. Fingers crossed :)

mojo/stdlib/test/collections/test_dict.mojo
```diff
@@ -405,8 +406,8 @@ struct DummyKey(KeyElement):
     fn __init__(out self, *, other: Self):
         self = other
 
-    fn __hash__(self) -> UInt:
-        return self.value
+    fn __hash__[H: Hasher](self, mut hasher: H):
+        return hasher.update(self.value)
```
soraros, line 410:
```suggestion
        hasher.update(self.value)
```

PR #4867, opened 2025-06-20
[stdlib] [proposal] Add `UnsafePointer[type, ...].alloc[count]()` to replace `stack_allocation[count, ...]()`

martinvuyk:
Add `UnsafePointer[type, ...].alloc[count]()` to replace `stack_allocation[count, ...]()`.

Motivation: this feels to better follow the nascent "mojo style" if I can even say that. It also makes more sense to add a factory method that returns a type to that type itself.

### DIFF
--- a/mojo/stdlib/stdlib/memory/memory.mojo
+++ b/mojo/stdlib/stdlib/memory/memory.mojo
@@ -398,6 +398,7 @@ fn stack_allocation[
     ]()
 
 
+# TODO: make this private
 @always_inline
 fn stack_allocation[
     count: Int,
@@ -406,7 +407,15 @@ fn stack_allocation[
     name: Optional[StaticString] = None,
     alignment: Int = alignof[type]() if is_gpu() else 1,
     address_space: AddressSpace = AddressSpace.GENERIC,
-]() -> UnsafePointer[type, address_space=address_space]:
+](
+    out res: UnsafePointer[
+        type,
+        address_space=address_space,
+        alignment=alignment,
+        mut=True,
+        origin = MutableOrigin.empty,
+    ]
+):
     """Allocates data buffer space on the stack given a data type and number of
     elements.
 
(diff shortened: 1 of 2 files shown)


soraros:
I thought most of the uses will be replaced by `InlineArray` ([ref](https://github.com/modular/modular/commit/fcfa9158a713c653c038dfd1509c3c46e003bbf7)). I'm not against this change though.

martinvuyk:
Hm true but it doesn't have the same capabilities :man_shrugging:. I think we should have both

lsh:
> Hm true but it doesn't have the same capabilities

Can you clarify what you mean by this?

martinvuyk:
> > Hm true but it doesn't have the same capabilities
> 
> Can you clarify what you mean by this?

@lsh Inline array is just a `!pop.array<size.value, Self.ElementType>` and there is no control over other aspects like `stack_allocation` gives:
```mojo
fn stack_allocation[
    count: Int,
    type: AnyType,
    /,
    name: Optional[StaticString] = None,
    alignment: Int = alignof[type]() if is_gpu() else 1,
    address_space: AddressSpace = AddressSpace.GENERIC,
]() -> UnsafePointer[type, address_space=address_space]: # this is wrong, I'm fixing it in this PR
    """Allocates data buffer space on the stack given a data type and number of
    elements.

    Parameters:
        count: Number of elements to allocate memory for.
        type: The data type of each element.
        name: The name of the global variable (only honored in certain cases).
        alignment: Address alignment of the allocated data.
        address_space: The address space of the pointer.

    Returns:
        A data pointer of the given type pointing to the allocated space.
    """
    ...
```

lsh:
Yes, to @soraros point we are working to add support to specify that address space / alignment of variables. Ideally `stack_allocation` will be relegated to hyper specific use cases.

lsh:
Just to add to this comment, once we can allocate an `InlineArray` with a custom address space and alignment, being able to grep for `stack_allocation` will be useful for removing its use as much as possible.

martinvuyk:
> Just to add to this comment, once we can allocate an `InlineArray` with a custom address space and alignment, being able to grep for `stack_allocation` will be useful for removing its use as much as possible.

ok, so I'll just close this

PR #4875, opened 2025-06-21
[stdlib] Make `_VariadicListIter` an `IteratorTrait`, and use them in element based `@parameter for ... in` loops

soraros:


lsh:
!sync

PR #4876, opened 2025-06-21
[stdlib] Remove the `SIMD.element_type` alias

soraros:
The `SIMD` type parameter has been renamed to `dtype`, so the alias is no longer necessary.

Part of #4215.


JoeLoser (approved):

JoeLoser:
!sync

PR #4877, opened 2025-06-21
[stdlib] Add `StaticSized` trait and `len` builtin for it

martinvuyk:
Add `StaticSized` trait and `len` builtin for it


soraros:
And we should be able to derive `Sized` from `StaticSized` when we have default implementation.

martinvuyk:
blocked by #4878

PR #4880, opened 2025-06-21
[stdlib] Trivial cleanups in `dtype.mojo`

soraros:
- Move the explicit MLIR code into `DType._match`
- Hide a few internal methods from docgen
- Use `@parameter for` to simplify `dispatch_integral`

### DIFF
--- a/mojo/stdlib/stdlib/builtin/dtype.mojo
+++ b/mojo/stdlib/stdlib/builtin/dtype.mojo
@@ -373,24 +373,28 @@ struct DType(
         """
         return self.value
 
+    @doc_private
     @staticmethod
-    fn _from_ui8(ui8: __mlir_type.ui8) -> DType:
-        return __mlir_op.`pop.dtype.from_ui8`(ui8)
-
-    @staticmethod
-    fn _from_ui8(ui8: __mlir_type.`!pop.scalar<ui8>`) -> DType:
-        return DType._from_ui8(
+    @always_inline("nodebug")
+    fn _from_ui8(ui8: UInt8._mlir_type) -> DType:
+        return __mlir_op.`pop.dtype.from_ui8`(
             __mlir_op.`pop.cast_to_builtin`[_type = __mlir_type.ui8](ui8)
         )
 
+    @doc_private
     @always_inline("nodebug")
-    fn _as_i8(
-        self,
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/builtin/dtype.mojo
```diff
@@ -530,14 +513,7 @@ struct DType(
...
         """
-        if self.is_integral():
-            return False
-        return Bool(
-            __mlir_op.`pop.cmp`[pred = __mlir_attr.`#pop<cmp_pred ne>`](
-                __mlir_op.`pop.simd.and`(self._as_i8(), _mIsFloat.value),
-                UInt8(0).value,
-            )
-        )
+        return self._match(_mIsFloat)
```
soraros, line 516: We're very confident this is correct, since I added systematic tests.

soraros:
@JoeLoser Could you please take a look at this?

mojo/stdlib/test/builtin/test_dtype.mojo
```diff
@@ -17,115 +17,132 @@ from sys import sizeof
...
 
-fn test_key_element() raises:
-    var set = Set[DType]()
-    set.add(DType.bool)
-    set.add(DType.int64)
+fn test_is_xxx() raises:
+    fn _is_category[
+        test: fn (DType) -> Bool,
+        true_dtypes: List[DType],
+        all_dtypes: List[DType] = all_dtypes,
```
JoeLoser: **Suggestion** Since this `all_dtypes` parameter is not specified below, should we just replace the `@parameter for` to use `all_dtypes` rather than the local parameter one?

JoeLoser (approved):
LGTM, thanks for the nice cleanup as always!

JoeLoser:
!sync

mojo/stdlib/test/builtin/test_dtype.mojo
soraros: Updated. Please resync.

JoeLoser: LGTM, I'll resync once `!sync` is working again (there were some internal infra changes that broke it earlier today).  CC @patrickdoc

JoeLoser:
!sync

PR #4881, opened 2025-06-21
[Docs] remove unused import in docs

zsiegel92:
The docs say "we currently need some more boilerplate code to make it work (but this will improve soon)" but I think it can improve *now* (`os` is unused)


JoeLoser (approved):
Thanks for the contribution, looks good to me!

JoeLoser:
!sync

PR #4890, opened 2025-06-24
[stdlib][NFC] Apply trivial cleanups in `_cpython.mojo` to prepare for refactoring

soraros:
- Add reference links to docstrings and comments
- Move magic-number constants closer to their usage
- Make other minor improvements

### DIFF
--- a/mojo/stdlib/stdlib/python/_cpython.mojo
+++ b/mojo/stdlib/stdlib/python/_cpython.mojo
@@ -39,43 +39,32 @@ from python.bindings import (
     Typed_newfunc,
 )
 
+alias Py_ssize_t = c_ssize_t
+alias Py_hash_t = Py_ssize_t
+
 # ===-----------------------------------------------------------------------===#
 # Raw Bindings
 # ===-----------------------------------------------------------------------===#
 
-# https://github.com/python/cpython/blob/d45225bd66a8123e4a30314c627f2586293ba532/Include/compile.h#L7
+# ref: https://github.com/python/cpython/blob/main/Include/compile.h
 alias Py_single_input = 256
 alias Py_file_input = 257
 alias Py_eval_input = 258
 alias Py_func_type_input = 345
 
-alias Py_tp_dealloc = 52
-alias Py_tp_init = 60
-alias Py_tp_methods = 64
-alias Py_tp_new = 65
-alias Py_tp_repr = 66
-
+# 0 when Stackless Python is disabled
+# ref: https://github.com/python/cpython/blob/main/Include/object.h
 alias Py_TPFLAGS_DEFAULT = 0
 
-alias Py_ssize_t = c_ssize_t
-alias Py_hash_t = Py_ssize_t
 
 # TODO(MOCO-1138):
 #   This should be a C ABI function pointer, not a Mojo ABI function.
+# ref: https://docs.python.org/3/c-api/structures.html#c.PyCFunction
 alias PyCFunction = fn (PyObjectPtr, PyObjectPtr) -> PyObjectPtr
-"""[Reference](https://docs.python.org/3/c-api/structures.html#c.PyCFunction).
-"""
 
+# Flag passed to newmethodobject
+# ref: https://github.com/python/cpython/blob/main/Include/methodobject.h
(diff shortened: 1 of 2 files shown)


martinvuyk (commented):
Nit-pick:
- IMO references should be part of the notes section not free floating text. We could create a "new header" like `References:`, but I'm strongly opposed to the amount of free floating (non-categorized) text that we already have in the stdlib
- Please use markdown syntax so that the link is rendered as a hyperlink
- Please use the python docstrings for the variables where documentation or references are added so that people can view them from the IDE when hovering without having to jump to definition

soraros:
Thanks for the review comments @martinvuyk!

Regarding "free floating" text, I see it as the docstring "body". We could file a feature request for a `References:` header, or better yet, standard markdown footnote support (do we have that already?). In the meantime I'm faking a footnote section with `---`, and I'm happy to migrate once a better option exists.

On markdown links: I went back and forth here. `_cpython` is internal, its docs are for maintainers, so we need a balance between IDE-rendered popups and plain-text readability. A generic label like `Reference` gives no clue where it leads, and crafting informative link names for the Python docs is tricky. I'd rather let the URL itself serve as documentation, keeping things a bit "raw" since this module isn't public.

On comments vs. docstrings for `typedef`s: for something like

```mojo
# typedef void (*destructor)(PyObject*)
```

a simple `#` comment is actually more eye-friendly in plain-text form than

```mojo
"""`typedef void (*destructor)(PyObject*)`"""
```

That said, I've still used triple-quoted docstrings for anything remotely public. I share your intent, just pointing out that not every rule maps perfectly to this internal module. I'd like to get these comments in now; the module will be refactored soon, and we can standardize formatting once things settle.

martinvuyk:
> Regarding "free floating" text, I see it as the docstring "body". We could file a feature request for a `References:` header, or better yet, standard markdown footnote support (do we have that already?). In the meantime I'm faking a footnote section with `---`, and I'm happy to migrate once a better option exists.

Yeah I opened #3235 a while ago because I saw this kind of issue coming. IMO we should try to create and keep to "peudo-standardized" section headers. They will evolve with time and according to needs we might not predict, so I don't think we'll be able to define every one in advance.

> On markdown links: I went back and forth here. `_cpython` is internal, its docs are for maintainers, so we need a balance between IDE-rendered popups and plain-text readability. A generic label like `Reference` gives no clue where it leads, and crafting informative link names for the Python docs is tricky. I'd rather let the URL itself serve as documentation, keeping things a bit "raw" since this module isn't public.

If you find reference too short you still can write something longer e.g. `[reference to python docs for this API](link)`. I personally just find a raw link very ugly and it occupies a bunch of space on the rendered popup.

> On comments vs. docstrings for `typedef`s: for something like
> 
> ```mojo
> # typedef void (*destructor)(PyObject*)
> ```

I think especially these should be in triple quotes. I agree they aren't as readable as `#` comments, but I personally find it very useful to be able to hover over cryptic variables and get a sense of what they are supposed to mean.

> That said, I've still used triple-quoted docstrings for anything remotely public. I share your intent, just pointing out that not every rule maps perfectly to this internal module. I'd like to get these comments in now; the module will be refactored soon, and we can standardize formatting once things settle.

Sounds good :+1:

soraros:
As a point of reference, the current setup looks neat/aligned in both plaintext and rendered format.

Screenshot 2025-06-24 at 20 34 18

arthurevans:
To be clear on the "free-floating text" question: our style is explicitly based on the Google Python docstrings style, which is in turn based on PEP257. That calls for longer descriptions to be in the main body of the docstring, separated from the summary by a blank line. We have no intention of changing this.

arthurevans (approved):
For consistency I'd prefer to remove the `---`. I tested this on a different doc and it shows up as a horizontal rule in our doc output, which looks odd since we don't have any other horizontal rules.

Since this is an internal module and liable to be reformatted when the module is refactored, I'm OK with merging this as-is, but we don't want this style leaking into the public docstrings.

soraros:
@arthurevans Thanks for the review! I end up removing all the `---`, and hopefully it's more consistent with the reset of the codebase.

PR #4897, opened 2025-06-25
[docs] Correct module for NDBuffer (#4861)

### LINKED ISSUE #4861: [Docs] Incorrect module for NDBuffer
### Where is the problem?

https://docs.modular.com/mojo/stdlib/buffer/buffer/

### What can we do better?

```
You can import these APIs from the memory package. For example:

from buffer import NDBuffer
```

which of these is the `memory` package?

### Anything else?

_No response_

hardikkgupta:
Edited `memory` -> `buffer`
Fixes #4861


abduld:
!sync

PR #4898, opened 2025-06-25
[stdlib] Perform trivial cleanups in `bool.mojo`

soraros:
- Use `select` in additional places
- Simplify the implementation of comparison operators

### DIFF
--- a/mojo/stdlib/stdlib/builtin/bool.mojo
+++ b/mojo/stdlib/stdlib/builtin/bool.mojo
@@ -25,7 +25,7 @@ from python import (
     ConvertibleFromPython,
 )
 
-from utils._select import _select_register_value
+from utils._select import _select_register_value as select
 from utils._visualizers import lldb_formatter_wrapping_type
 
 # ===----------------------------------------------------------------------=== #
@@ -287,7 +287,7 @@ struct Bool(
         Returns:
             1 if the Bool is True, 0 otherwise.
         """
-        return Int(1) if self else Int(0)
+        return select[Int](self, 1, 0)
 
     @always_inline("builtin")
     fn __as_int__(self) -> Int:
@@ -315,7 +315,7 @@ struct Bool(
         Returns:
             1.0 if True else 0.0 otherwise.
         """
-        return _select_register_value(self, Float64(1.0), Float64(0.0))
+        return select[Float64](self, 1, 0)
 
     @always_inline("builtin")
     fn __eq__(self, rhs: Bool) -> Bool:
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/builtin/bool.mojo
```diff
@@ -330,7 +330,7 @@ struct Bool(
         Returns:
             True if the two values match and False otherwise.
         """
-        return not self != rhs
+        return ~(self != rhs)
```
soraros, line 333: Why don't we generate a `icmp eq i1` here? Isn't that more friendly to pattern matching based passes?

mojo/stdlib/stdlib/builtin/bool.mojo
```diff
@@ -528,7 +528,7 @@ struct Bool(
         Returns:
             0 for False and -1 for True.
         """
-        return Int(-1) if self else Int(0)
+        return select[Int](self, -1, 0)
```
soraros, line 531: Do we even want to support `-Bool(value)` at all? Python allows this since `bool` is also an `int`; however, we already deviate from Python behaviour, because we have `~True = False`, whereas in Python `~True = -2`.

JoeLoser: I'd suggest we remove that behavior since Mojo has strong typing.  I remember me and you were talking about this last year sometime and it made one particular thing easier, but it is not a great thing for Mojo to keep from Python IMO.  We've moved on from the "superset of Python" and needing to copy some of its not-so-great things.  This is one of those things we should deviate from IMO.  Any interest in opening a PR?

mojo/stdlib/stdlib/builtin/bool.mojo
JoeLoser: Intuitively I'd expect an `icmp eq i1` as well - can you file a QoI compiler issue please to investigate this and @npanchen may be interested?

JoeLoser (approved):

JoeLoser:
!sync

mojo/stdlib/stdlib/builtin/bool.mojo
soraros: Indeed, let me remove this in a subsequent PR. Maybe this deserves a changelog entry? Just to make sure I understand you correctly, you are saying that we an remove `Bool.__neg__`, right?

mojo/stdlib/stdlib/builtin/bool.mojo
soraros: Opened #4907 for this.

mojo/stdlib/stdlib/builtin/bool.mojo
lattner: +1, thank you for removing this!

PR #4902, opened 2025-06-25
[Graph] Add the graph APIs

modularbot:
Filtered commits since previous sync


max/graph/dtype_promotion.py
```diff
@@ -0,0 +1,187 @@
+# ===----------------------------------------------------------------------=== #
+#
+# This file is Modular Inc proprietary.
+#
+# ===----------------------------------------------------------------------=== #
+"""Handles DType promotion for TensorValues.
+
+DType promotion in the graph api is used to decide the DType of the output of an operations with multiple inputs.
```
martinvuyk, line 8:
```suggestion
DType promotion in the graph api is used to decide the DType of the output of an operation with multiple inputs.
```

bethebunny (approved):

PR #4913, opened 2025-06-26
[stdlib] Generalise `from_bytes` and `as_bytes` so it works on any SIMD vector not only scalars

mzaks:
The behaviour before worked ok for scalar and silently trimmed a vector to first element, which IMHO is an unexpected behaviour. Now all values of a SIMD value are included.


mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -1961,32 +1961,35 @@ struct SIMD[dtype: DType, size: Int](
...
             The integer value.
         """
-        var ptr = bytes.unsafe_ptr().bitcast[Scalar[dtype]]()
+        var ptr = bytes.unsafe_ptr().bitcast[SIMD[dtype, size]]()
         var value = ptr[]
 
         @parameter
         if is_big_endian() != big_endian:
+
             return byte_swap(value)
```
soraros: I suspect that this is slightly wrong now.

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -1961,32 +1961,35 @@ struct SIMD[dtype: DType, size: Int](
 
     @staticmethod
     fn from_bytes[
-        big_endian: Bool = is_big_endian()
-    ](bytes: InlineArray[Byte, dtype.sizeof()]) -> Scalar[dtype]:
-        """Converts a byte array to an scalar integer.
+        big_endian: Bool = is_big_endian(),
+        size: Int = 1,
+    ](bytes: InlineArray[Byte, dtype.sizeof() * size]) -> SIMD[dtype, size]:
```
soraros: Maybe we could use `sizeof[Self]` this time?

mojo/stdlib/stdlib/builtin/simd.mojo
mzaks: The formatting was wrong but the logic is fine, I added a unit test now to confirm the endianness is supported.

mojo/stdlib/stdlib/builtin/simd.mojo
mzaks: Oh, that's a good point! It also simplifies the API! just pushed it.

mojo/stdlib/test/builtin/test_simd.mojo
```diff
@@ -1971,6 +1972,81 @@ def test_from_bytes_as_bytes():
                 x,
             )
 
+    var v8_u8 = SIMD[DType.uint8, 8](1, 2, 3, 4, 5, 6, 7, 8)
+    assert_equal(v8_u8, SIMD[DType.uint8, 8].from_bytes(v8_u8.as_bytes()))
+
+    var v8_u16 = SIMD[DType.uint16, 8](1, 2, 3, 4, 5, 6, 7, 8)
+    var expected_v8_u16_be_bytes = [
```
soraros, line 1982: Can we wrap `fmt: off` and `fmt: on` around them so they fit in a single line.

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -1974,7 +1974,7 @@ struct SIMD[dtype: DType, size: Int](
         Returns:
             The integer value.
         """
-        var ptr = bytes.unsafe_ptr().bitcast[Scalar[dtype]]()
+        var ptr = bytes.unsafe_ptr().bitcast[SIMD[dtype, size]]()
```
soraros:
```suggestion
        var ptr = bytes.unsafe_ptr().bitcast[Self]()
```

soraros:
I wonder what was the CI failure on ubuntu all about.

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -1961,9 +1961,9 @@ struct SIMD[dtype: DType, size: Int](
 
     @staticmethod
     fn from_bytes[
-        big_endian: Bool = is_big_endian()
-    ](bytes: InlineArray[Byte, dtype.sizeof()]) -> Scalar[dtype]:
-        """Converts a byte array to an scalar integer.
+        big_endian: Bool = is_big_endian(),
```
soraros, line 1960:
```suggestion
        *, big_endian: Bool = is_big_endian(),
```

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -1985,8 +1985,8 @@ struct SIMD[dtype: DType, size: Int](
 
     fn as_bytes[
         big_endian: Bool = is_big_endian()
```
soraros:
```suggestion
        *, big_endian: Bool = is_big_endian(),
```

mzaks:
> I wonder what was the CI failure on ubuntu all about.

Seems like Bazel Ubuntu test does not like long unit tests. Splitting the function into two did the trick.

soraros:
It would be nice to be able to download the log file from a failed CI run. WDYT, @JoeLoser?

soraros (approved):

soraros:
> Seems like Bazel Ubuntu test does not like long unit tests. Splitting the function into two did the trick.

I hope it's not because we triggered some kind of UB.

laszlokindrat (approved):
Nice catch, thanks for fixing it!

laszlokindrat:
!sync

PR #4915, opened 2025-06-26
[stdlib] [kernels] Adapt argsort cpu implementation and take it to stdlib `sort.mojo`

### LINKED ISSUE #2979: [Feature Request] [stdlib] Add `argsort` to the builtin `sort` module
### Review Mojo's priorities

### What is your request?

Add argsort similar to [numpy's](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html)

sorting inplace is not an option if you need to sort multiple arrays keeping the same permutations

### Any other details?

I don't know what would be needed in terms of `Span`'s capabilities. Because that permutation should be able to affect other items.
```mojo
var my_list_distances = List[UInt64]()
var my_structures = List[MyStruct]()
...
var permutation = argsort(my_list_distances)
var ordered_structs = my_structures[permutation]
```
Or for now we can just return a list of indexes

martinvuyk:
Adapt argsort cpu implementation and take it to stdlib `sort.mojo`. Closes #2979. IMO this should live in the stdlib because it is a useful API and the implementation is straightforward.

This PR makes the sort functions more generic over alignment and fixes immutable span inference for such cases.

This PR also adds a private convenience method to `LayoutTensor` to return a `Span`.

### DIFF
--- a/mojo/stdlib/stdlib/builtin/sort.mojo
+++ b/mojo/stdlib/stdlib/builtin/sort.mojo
@@ -15,7 +15,7 @@
 These are Mojo built-ins, so you don't need to import them.
 """
 
-from math import ceil
+from math import ceil, iota
 from sys import bitwidthof
 
 from bit import count_leading_zeros
@@ -41,7 +41,7 @@ fn _insertion_sort[
     T: Copyable & Movable,
     origin: MutableOrigin, //,
     cmp_fn: fn (_SortWrapper[T], _SortWrapper[T]) capturing [_] -> Bool,
-](span: Span[T, origin]):
+](span: Span[T, origin, address_space = AddressSpace.GENERIC, **_]):
     """Sort the array[start:end] slice"""
     var array = span.unsafe_ptr().origin_cast[origin=MutableAnyOrigin]()
     var size = len(span)
@@ -66,7 +66,7 @@ fn _quicksort_partition_right[
     T: Copyable & Movable,
     origin: MutableOrigin, //,
     cmp_fn: fn (_SortWrapper[T], _SortWrapper[T]) capturing [_] -> Bool,
-](span: Span[T, origin]) -> Int:
+](span: Span[T, origin, address_space = AddressSpace.GENERIC, **_]) -> Int:
     var array = span.unsafe_ptr().origin_cast[origin=MutableAnyOrigin]()
     var size = len(span)
 
@@ -95,7 +95,7 @@ fn _quicksort_partition_left[
     T: Copyable & Movable,
     origin: MutableOrigin, //,
     cmp_fn: fn (_SortWrapper[T], _SortWrapper[T]) capturing [_] -> Bool,
-](span: Span[T, origin]) -> Int:
+](span: Span[T, origin, address_space = AddressSpace.GENERIC, **_]) -> Int:
     var array = span.unsafe_ptr().origin_cast[origin=MutableAnyOrigin]()
     var size = len(span)
 
@@ -121,7 +121,7 @@ fn _heap_sort_fix_down[
     T: Copyable & Movable,
     origin: MutableOrigin, //,
     cmp_fn: fn (_SortWrapper[T], _SortWrapper[T]) capturing [_] -> Bool,
-](span: Span[T, origin], idx: Int):
+](span: Span[T, origin, address_space = AddressSpace.GENERIC, **_], idx: Int):
     var array = span.unsafe_ptr().origin_cast[origin=MutableAnyOrigin]()
     var size = len(span)
     var i = idx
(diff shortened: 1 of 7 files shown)


mojo/stdlib/stdlib/builtin/sort.mojo
```diff
@@ -696,3 +702,70 @@ fn _small_sort[
...
+fn argsort[
+    values_dtype: DType, //,
+    *,
+    ascending: Bool = True,
+    indices_dtype: DType = DType.index,
+](values: Span[mut=False, Scalar[values_dtype], **_]) -> List[
+    Scalar[indices_dtype]
+]:
+    """Sorts the index buffer according to the biggest/smallest values that they
+    correspond to.
```
laszlokindrat: Can you please update this to reflect that this overload returns the indices in a new list?

mojo/stdlib/stdlib/builtin/sort.mojo
```diff
@@ -696,3 +702,70 @@ fn _small_sort[
...
+
+    Parameters:
+        values_dtype: The dtype of the value buffer.
+        ascending: Sort direction (True for ascending, False for descending).
+        indices_dtype: The dtype of the indices buffer.
+
+    Args:
```
laszlokindrat, line 764: Might have to add a `Returns:` section?

mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -534,6 +534,8 @@ struct List[T: Copyable & Movable, hint_trivial_type: Bool = False](
         self._unsafe_next_uninit_ptr().init_pointee_move(value^)
         self._len += 1
 
+    # TODO: this should be an overload of extend, not append. And we need to
+    # give users better control over the allocation reserve calculation.
```
laszlokindrat: suggestion: Could you just fix this in a separate patch instead of leaving this TODO here? Or, if it's a bit more work, could you please open a ticket and reference it here?

mojo/stdlib/stdlib/prelude/__init__.mojo
```diff
@@ -109,7 +109,7 @@ from builtin.simd import (
     UInt128,
     UInt256,
 )
-from builtin.sort import partition, sort
+from builtin.sort import partition, sort, argsort
```
laszlokindrat: I'm a bit surprised that these are in the prelude, they don't seem commonly used at all. @JoeLoser how do you feel about removing these from here?

laszlokindrat (approved):
LGTM, just a few small things, thank you! Could you please also rebase on the latest nightly so I can take a final look and bring it in?

mojo/stdlib/stdlib/prelude/__init__.mojo
martinvuyk: I think they are here because sort is a builtin in Python. I think it would be confusing for users to know to import `from builtin.sort import sort`. Partition and argsort could probably be left out. I'll remove argsort.

laszlokindrat: `sorted` is a builtin in python, but it has a slightly different name, and a different signature. I think it's okay to deviate here, since we are not aiming to be a drop in replacement.

laszlokindrat:
!sync

laszlokindrat:
@martinvuyk be advised, this is causing an internal ASAN job to crash. We're investigating, but it might take a while.

martinvuyk:
@laszlokindrat if you can't find the issue causing the ASAN crash we could also try to split this PR into smaller pieces. I already created #4917 a while ago. We could split off the part where I make the sort functions generic as well.

laszlokindrat:
@martinvuyk Good callout, I checked the status of https://github.com/modular/modular/pull/4917, and its running into the same issue internally. If you want, you can open a new PR with the difference of these two, and see if we can move that forward while we wait for a fix.

martinvuyk:
Only problem is that the argsort signature is generic:
```mojo
fn _argsort_cpu[
    *,
    ascending: Bool = True,
](
    indices: LayoutTensor[mut=True, address_space = AddressSpace.GENERIC, **_],
    input: LayoutTensor,
) raises:
```
That is why I had to build out the other parts. So this is blocked on that issue.

Though technically the current implementation "casts" the span to a Span with specific alignment:
```mojo
sort[cmp_fn](
    Span[
        Scalar[indices.dtype],
        indices.origin,
    ](ptr=indices.ptr, length=indices.size())
)
```
so maybe we could still not make the argsort one generic. But to me these cases seem more like bugs than anything else. Because the underlying `UnsafePointer` are being silently casted to alignments and or address spaces that they actually aren't

Issue #4918, opened 2025-06-27
[Docs] Document usage of MAX graph compilation cache with containers


orpheezt:
### Where is the problem?

DockerHub Images: https://hub.docker.com/u/modular & Some MAX docs

### What can we do better?

Document how to enable and persist the MAX graph compilation cache when using containers. This reduces compilation time and significantly improves startup latency—especially on serverless and ephemeral platforms where cold-start performance is critical.

orpheezt:
Add `-v ~/.cache/max_cache:/opt/venv/share/max/.max_cache` to Dockerhub image description and MAX documentation entries.

judyheflin:
Hey @orpheezt, thanks so much for calling this out and putting together a fix! We’re actually working on a different caching approach internally, so we’re going to hold off on merging this for now, as we'll soon have a slightly different workflow recommendation.

PR #4919, opened 2025-06-27
[DOCS] Document usage of MAX graph compilation cache with containers

orpheezt:
Related to #4918


orpheezt:

scottamain:
Thank you for point this out! As @judyheflin said on [the issue](https://github.com/modular/modular/issues/4918#issuecomment-3054194392), this behavior will change soon, so we’ll keep that ticket open but won’t be accepting this change. We’ll update the docs as appropriate to resolve that issue soon.

PR #4922, opened 2025-06-27
[stdlib] Adds hasher parameter to Dict

mzaks:
This PR concludes the transformation of the hashing strategy to data flow based hashing. With this PR users are able to provide a parameter to the Dict in order to customise the way a hash value is computed for the keys.

Important note! As at this point in time the AHasher implementation does not work at compile time and the Dict parameter defaults to AHasher, the users would need to provide a different hasher (e.g. Fnv1a) if they chose to have a comp time Dict, Set, or other Dict based data structure.

### DIFF
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -39,6 +39,12 @@ what we publish.
 
 ### Standard library changes
 
+- The `Dict` now has an `H` parameter which allows users to provider a
+  custom `Hasher` type.
+  - `default_hasher` (AHasher) and `default_comp_time_hasher` (Fnv1a)
+    are now provided
+  - The `H` parameter of `Dict` defaults to `default_hasher`
+
 - The `Hashable` trait has been updated to use a new data flow strategy.
   - Users are now required to implement the method
     `fn __hash__[H: Hasher](self, mut hasher: H):`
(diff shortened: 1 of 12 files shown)


JoeLoser:
!gpu

JoeLoser (approved):
This is great, thanks for working on this @mzaks!  Can you please file a GitHub issue with repro for why AHasher doesn't work at compile time - what are the limitations/issues?  Fixing this would be great (CC: @dgurchenkov) so users don't have to explicitly specify the hasher type just because they want a compile time Dict.  This is in the spirit of making things "just work" and not needing to vary code between compile time and runtime in the initialization, declaration, etc.

That said, I want to land this and we can iterate on fixing the above issue and improving this for users.

JoeLoser:
!sync

mzaks:
@JoeLoser and @dgurchenkov this is the issue for compile time crash while using `AHasher` at compile time. #4924

JoeLoser:
@mzaks FYI this requires a bit of work in fixing things up internally. I'll be able to do that on Monday and land this. Feel free to build things on top of this PR. All of the changes I need to do are internal only and I don't expect anything to change from the standard library side of things.

PR #4932, opened 2025-06-29
[stdlib] Add casting for non-floating-point dtypes in log functions

martinvuyk:
Add casting for non-floating-point dtypes in log functions.

These functions should work fine for integral dtypes as well. Let the user decide which datatype to input and let them deal with casting issues.

### DIFF
--- a/mojo/stdlib/stdlib/math/math.mojo
+++ b/mojo/stdlib/stdlib/math/math.mojo
@@ -749,14 +749,32 @@ fn frexp[
 # ===----------------------------------------------------------------------=== #
 
 
+fn _fp_type[bitwidth: Int]() -> DType:
+    @parameter
+    if bitwidth <= 16:
+        return DType.float16
+    elif bitwidth == 32:
+        return DType.float32
+    else:
+        return DType.float64
+
+
 @always_inline
 fn _log_base[
-    dtype: DType, width: Int, //, base: Int
-](x: SIMD[dtype, width]) -> SIMD[dtype, width]:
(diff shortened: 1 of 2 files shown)


soraros:
I don't think this is the right approach. If we were to do it, I'd much rather leave `_log_base`, which is clearly a polynomial-based numerical method that only makes sense for certain floating-point types, untouched and put the dispatch logic in `log`.

martinvuyk:
I don't see the difference in casting the input value in `_log_base`

mojo/stdlib/stdlib/math/math.mojo
```diff
@@ -749,14 +749,32 @@ fn frexp[
...
 
 
+fn _fp_type[bitwidth: Int]() -> DType:
+    @parameter
+    if bitwidth <= 16:
+        return DType.float16
+    elif bitwidth == 32:
+        return DType.float32
+    else:
+        return DType.float64
```
laszlokindrat, line 759: I'm a bit hesitant about this: if the input is a 32 bit int, this implies that the computation will happen in float32, which guarantees that we lose precision. Moreover, we now have 128 and 256 bit integers that will fallback to float64 calculation.

Based on the (admittedly minimal) tests, it seems that our log implementation was never intended to work on integers. So maybe we should just enforce that, and push the conversion on the user?

martinvuyk:
Yeah I think I'm going to just close this one. My actual need was log where I knew that the number was a power of two.

PR #4935, opened 2025-06-30
[stdlib] Remove the `Copyable` and `Moveable` bounds from `Intable`

### LINKED ISSUE #4934: [Feature Request] Intable should not inherit Copyable or Movable
### Review Mojo's priorities

### What is your request?

`Intable` currently inherits from `Copyable` and `Moveable` and I don't really see any reason why it should. Interestingly, `IntableRaising` doesn't inherit from them so it really seems like `Intable` shouldn't either.

I have a type I want to be `Intable` but not `Copyable`.

### Any other details?

_No response_

soraros:
Fix #4934.


JoeLoser (approved):
Great clean up, thank you!  This will make @abduld happy as there was an old internal issue for this.

JoeLoser:
!sync

PR #4938, opened 2025-06-30
[CI] Disable tests that require network

keith:
The tests that we have that require network access are normally pulling
from huggingface. Internally we have some workarounds to try and make
that more stable, but since we're using ephemeral github runners
externally it's likely going to be too flaky to make it worth running.
These tests still run before running internally, and they can still be
run locally.

Signed-off-by: Keith Smiley <keithbsmiley@gmail.com>


keith:
!sync

Ahajha (approved):

PR #4949, opened 2025-07-02
[stdlib] [MAX] Change all uses of `raise "str"` or `raise String()` to use `Error`

martinvuyk:
Change all uses of `raise "str"` or `raise String()` to use `Error`. Error has to copy the data again otherwise.


JoeLoser (approved):
Nice, thank you! 👍

JoeLoser:
!sync

PR #4958, opened 2025-07-03
[Kbench] Added `init-kbench`

parallelistix:
Running the following line should install and enable `kbench` in your env:
```
source max/kernels/benchmarks/autotune/init-kbench.sh
```

Example:
```
cd $KERNEL_BENCHMARKS_ROOT
kbench -v bench_vec_add.yaml
kbench -v bench_matmul.yaml
```


parallelistix:
!sync

PR #4962, opened 2025-07-04
[stdlib] Fix unnecessary copy when moving elements, optimize copy

### LINKED ISSUE #4954: [BUG] Possible missing optimisation in `InlineArray.__moveinit__`
### Bug description

Currently, `InlineArray.__moveinit__` is implemented as follows, which appears harmless:

https://github.com/modular/modular/blob/b32b261bf32bac66cf7dcc5d6fe648d29d40c4bc/mojo/stdlib/stdlib/collections/inline_array.mojo#L339-L341

However, this version performs an extra copy. We can eliminate it by using:

```mojo
struct InlineArray[...]:
  fn __moveinit__(out self, owned other: Self):
    ...

    for idx in range(size):
      var ptr = self.unsafe_ptr() + idx
      var other_ptr = other.unsafe_ptr() + idx
      other_ptr.move_pointee_into(ptr)
```

I'm not sure whether this is the intended behaviour or a missing optimisation. If it's the former, we might need to audit the stdlib and change a few more places. CC @lattner.

Credit to @xtxtxtxtxtxtx for discovering the issue and proposing the fix.

### Steps to reproduce

```mojo
@fieldwise_init
struct S(Copyable, Movable):
  fn __moveinit__(out self, owned other: Self):
    print("moveinit")

  fn __copyinit__(out self, other: Self):
    print("copyinit")

fn main():
  a = InlineArray[S, 1](S())
  print("---")
  b = a^
```

Prints:

```text
copyinit
---
copyinit
moveinit
```

With the fix, prints:

```text
copyinit
---
moveinit
```

### System information

`Mojo 25.5.0.dev2025070207 (5e6c051c)`.

samufi:
This fixes #4954.

Furthermore, copy now directly emplaces the result into the output slot, which saves a move operation.

### DIFF
--- a/mojo/stdlib/stdlib/collections/inline_array.mojo
+++ b/mojo/stdlib/stdlib/collections/inline_array.mojo
@@ -290,7 +290,7 @@ struct InlineArray[
         # Do not destroy the elements when their backing storage goes away.
         __disable_del storage
 
-    fn copy(self) -> Self:
+    fn copy(self, out copy: Self):
         """Creates a deep copy of the array.
 
         Returns:
@@ -304,13 +304,14 @@ struct InlineArray[
         ```
         """
 
-        var copy = Self(uninitialized=True)
+        copy = Self(uninitialized=True)
 
+        @parameter
         for idx in range(size):
             var ptr = copy.unsafe_ptr() + idx
-            ptr.init_pointee_copy(self[idx])
+            ptr.init_pointee_copy(self.unsafe_get(idx))
 
-        return copy^
+        return copy
 
     fn __copyinit__(out self, other: Self):
         """Copy constructs the array from another array.
@@ -336,9 +337,10 @@ struct InlineArray[
 
         __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))
 
+        @parameter
         for idx in range(size):
-            var ptr = self.unsafe_ptr() + idx
-            ptr.init_pointee_move(other[idx])
+            var other_ptr = other.unsafe_ptr() + idx
+            other_ptr.move_pointee_into(self.unsafe_ptr() + idx)
 
     fn __del__(owned self):
         """Deallocates the array and destroys its elements.
--- a/mojo/stdlib/stdlib/collections/inline_array.mojo
+++ b/mojo/stdlib/stdlib/collections/inline_array.mojo
@@ -311,8 +311,6 @@ struct InlineArray[
             var ptr = copy.unsafe_ptr() + idx
             ptr.init_pointee_copy(self.unsafe_get(idx))
 
-        return copy
-
     fn __copyinit__(out self, other: Self):
         """Copy constructs the array from another array.
 
--- a/mojo/stdlib/stdlib/collections/inline_array.mojo
+++ b/mojo/stdlib/stdlib/collections/inline_array.mojo
@@ -306,7 +306,6 @@ struct InlineArray[
 
         copy = Self(uninitialized=True)
 
-        @parameter
         for idx in range(size):
             var ptr = copy.unsafe_ptr() + idx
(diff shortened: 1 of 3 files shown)


mojo/stdlib/stdlib/collections/inline_array.mojo
```diff
@@ -304,13 +304,12 @@ struct InlineArray[
         ```
         """
 
-        var copy = Self(uninitialized=True)
+        copy = Self(uninitialized=True)
 
+        @parameter
```
soraros: Can we leave this out for the moment? I think we might need manual unrolling like in the batched `__init__`.

soraros:
Could you sign-off your commits following the instructions here: https://github.com/modular/modular/pull/4962/checks?check_run_id=45366010093

mojo/stdlib/stdlib/collections/inline_array.mojo
samufi: If we are serious about this, we should remove it wherever it is not batched. It makes little sense to do it this way here and that way there, because performance and resource consumption will have the worst from both worlds then.

samufi: I have done now anyway.

samufi: (not touching code that is not subject of this PR)

soraros: > If we are serious about this, we should remove it wherever it is not batched. It makes little sense to do it this way here and that way there, because performance and resource consumption will have the worst from both worlds then.

I'm not entirely convinced that's true, though we probably *should* audit the use of `@parameter for`. The concern (also in `__init__`) was primarily compilation time, not runtime performance; as far as I understand, the Mojo compiler is still smart enough to unroll small loops.

samufi: My argument would be as follows: If `@parameter for` causes problematic compilation times, we will suffer from that anyway, since `@parameter for` (without batching) also occurs in `__del__` and `__contains__`. Though I admit that the latter may be used somewhat rarely, `__del__` will be used on every `InineArray` that calls destructors (which should be all `InlineArrays` with non-trivial elements. If this is the case, we are already suffering from bad compilation times, and the additional `@parameter for` will only cause a change by a constant multiplier. At the same time, if `@parameter for` is relevant for the runtime, copy and move without `@parameter for` could cause a bottleneck.

I agree, though, that we should use the batched operations everywhere. Isn't there some abstraction of that, (`vectorize`?) which we could use?

JoeLoser (approved):
Thanks for fixing this!  @laszlokindrat do you mind also taking a look at this for another set of eyes, please?

mojo/stdlib/test/collections/test_inline_array.mojo
```diff
@@ -293,3 +328,4 @@ def main():
     test_inline_array_runs_destructors()
     test_unsafe_ptr()
     test_sizeof_array[Int, capacity=32]()
+    test_move()
```
laszlokindrat, line 331: We seriously didn't have this test enabled? 😱

laszlokindrat (approved):
Great, thank you for fixing this!

laszlokindrat:
@samufi Can you please sign off on the commits so I can bring them in? Also, please rebase on the latest nightly, just in case.

mojo/stdlib/test/collections/test_inline_array.mojo
samufi: Tbh, I forgot it on my last PR, when I introduced the test. So I am simply cleaning up after me...

samufi:
> @samufi Can you please sign off on the commits so I can bring them in? Also, please rebase on the latest nightly, just in case.

Done.

laszlokindrat:
!sync

PR #4966, opened 2025-07-04
[stdlib][cpython-api] ABI hardening for CPython GIL/thread methods

soraros:


soraros:
CC @laszlokindrat.

laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

PR #4680, opened 2025-05-23
[Refactoring][stdlib] Replace `_UInt128` with `UInt128`

kasmith11:
Addresses #4569 by removing _UInt128 struct and references in `stdlib/builtin/_format_float.mojo` by adjusting the following functions to be inline with the UInt64 versions:

- `_umul128`
- `_umul192_lower128`
- `_umul192_upper128`
- `_compute_mul_parity`

PR #4971, opened 2025-07-05
[stdlib] Switch Dragonbox float writer to UInt128 helpers

### LINKED ISSUE #4569: [Refactoring][stdlib] Replace `_UInt128` with `UInt128` in `stdlib/builtin/_format_float.mojo`
### Review Mojo's priorities

### What is your request?

Replace `_UInt128` type in `stdlib/builtin/_format_float.mojo` with the new `UInt128` type from `stdlib/builtin/simd.mojo`.

Simplifying/consolidating on existing types.

### Any other details?

_No response_

hardikkgupta:
Replaces the custom _UInt128 struct with the built-in UInt128 type and updates
all helpers accordingly.
Fixes #4569


hardikkgupta:
Hey @soraros, any feedback?

soraros:
FYI, @kasmith11 already opened #4680 for this.

Regarding the code itself, just try to fix the error as the compiler points out. The general approach looks good to me.

hardikkgupta:
Yes, his work was great and helpful! It got little hard for me to follow his code. I am looking into this

hardikkgupta:
I think errors are fixed, and it can be merged now

mojo/stdlib/stdlib/builtin/_format_float.mojo
```diff
@@ -81,6 +69,14 @@ struct FP[dtype: DType, CarrierDType: DType = FPUtils[dtype].uint_type]:
...
     alias small_divisor = pow(10, Self.kappa)
 
+fn _uint64s_to_uint128(hi: UInt64, lo: UInt64) -> UInt128:
+    return (UInt128(hi) << 64) | UInt128(lo)
+
+fn _uint128_high(x: UInt128) -> UInt64:
+    return (x >> 64).cast[DType.uint64]()
+
+fn _uint128_low(x: UInt128) -> UInt64:
+    return (x & UInt128(0xFFFF_FFFF_FFFF_FFFF)).cast[DType.uint64]()
```
soraros: `cast` lowers into `trunc` already, no need to apply a mask first.

hardikkgupta: Yes, missed this one

mojo/stdlib/stdlib/builtin/_format_float.mojo
```diff
@@ -584,12 +581,8 @@ fn _compute_mul_parity[
...
         )
         return _MulParity(
-            ((r.high >> (64 - beta)) & 1) != 0,
-            (
-                ((r.high << beta) & UInt64(0xFFFFFFFFFFFFFFFF))
-                | (r.low >> (64 - beta))
-            )
-            == 0,
+            ((_uint128_high(r) >> (64 - beta)) & 1) != 0,
+            (_uint128_low(r) >> (32 - beta)) == 0
```
soraros: Are you sure this is equivalent to the original code?

mojo/stdlib/stdlib/builtin/_format_float.mojo
```diff
@@ -601,12 +594,12 @@ fn _compute_mul_parity[
         )
         return _MulParity(
             ((r >> (64 - beta)) & 1) != 0,
-            (UInt32(0xFFFFFFFF).cast[DType.uint64]() & (r >> (32 - beta))) == 0,
+            (r >> (32 - beta)) == 0
```
soraros: Same question as above. I don't think this is equivalent to the old code. Please provide a proof or a reference.

soraros:
- Please format the file by running `mojo format`.
- Please use your real name, not an alias, in the sign-off line.

hardikkgupta:
okay, will get back to you

soraros:
The PR title and description are also inaccurate. I don't think you've added support for `u128`; you've simply replaced the `_UInt128` struct with the native `UInt128` type.

hardikkgupta:
I changed/modified all the above (signoff, code, PR title/description) to match it correctly. Locally, the tests are passing but are failing here. I guess it is something to do with _masks‐out the upper 32 bits_ `UInt32(0xFFFFFFFF).cast[DType.uint64]()` because when I removed it earlier, they compiled it correctly.
Any insights?

mojo/stdlib/stdlib/builtin/_format_float.mojo
```diff
@@ -628,13 +634,11 @@ fn _check_divisibility_and_divide_by_pow10[
 
 
 @always_inline
-fn _truncate[
-    dtype: DType, S: Int, //, TruncateType: DType
-](u: SIMD[dtype, S]) -> SIMD[dtype, S]:
+fn _truncate[dtype: DType](u: Scalar[dtype]) -> Scalar[dtype]:
```
soraros: This change is unnecessary (and incorrect). This function is now a no-op.

hardikkgupta: reverted back

hardikkgupta: So now I changed a little
1. For `uint64` SIMD lanes, we cast down to `uint32` then back up (intended truncate)
2. For all other, we simply cast to the same dtype, satisifying original intent
For going through the compiler checks
It works I guess to pass through the checks!

mojo/stdlib/stdlib/builtin/_format_float.mojo
soraros: I think you reverted the reversion in the latest commit.

mojo/stdlib/stdlib/builtin/_format_float.mojo
soraros: Though I think for the scope of this PR, you should probably try to keep the change set minimal. We could always refactor later.

hardikkgupta: Yes, but if I can keep the `_truncate` function same, the checks are not passing
Although I can revert to the original, but the checks will be a problem
Let me know what you think abut this

soraros: I think you could, or perhaps should, leave `_truncate` in its original form.

laszlokindrat:
@hardikkgupta Thanks for the patch! I'm closing this in favor of https://github.com/modular/modular/pull/5452 that recently landed. We're looking forward to your other contributions, though!

PR #4978, opened 2025-07-07
[stdlib][cpython-api] ABI hardening for CPython very high level methods

soraros:


laszlokindrat (approved):
Nice, thank you!

laszlokindrat:
!sync

PR #4987, opened 2025-07-08
[stdlib] Simplify `SIMD.__init__(PythonObject)`

soraros:
`PyFloat_AsDouble` calls `__float__` when necessary, so there's no need to reproduce that logic in Mojo.


soraros:
CC @laszlokindrat.

laszlokindrat (approved):
Much cleaner, thank you!

laszlokindrat:
!sync

PR #4991, opened 2025-07-09
[stdlib][cpython-api] ABI hardening for CPython object protocol methods

soraros:


laszlokindrat (approved):
Keep `em coming, thank you!

laszlokindrat:
!sync

PR #4992, opened 2025-07-09
[stdlib] Introduce `PyObjectPtr.bitcast` and use it to simplify code

soraros:
- Rename `PyObjectPtr.unsized_obj_ptr` to `_unsized_obj_ptr` to mark it as a private field.
- Add the `PyObjectPtr.bitcast` method for downcasting.
- Add a `PyObjectPtr(upcast_from)` constructor that accepts an `UnsafePointer[T]` for upcasting.

### DIFF
--- a/mojo/stdlib/stdlib/python/python_object.mojo
+++ b/mojo/stdlib/stdlib/python/python_object.mojo
@@ -260,17 +260,12 @@ struct PythonObject(
             If no Python type object has been registered for `T` by a
             `PythonTypeBuilder`.
         """
-
         # NOTE:
         #   We can't use PythonTypeBuilder.bind[T]() because that constructs a
         #   _new_ PyTypeObject. We want to reference the existing _singleton_
         #   PyTypeObject that represents a given Mojo type.
         var type_obj = lookup_py_type_object[T]()
-
-        var type_obj_ptr = type_obj._obj_ptr.unsized_obj_ptr.bitcast[
-            PyTypeObject
-        ]()
-
+        var type_obj_ptr = type_obj._obj_ptr.bitcast[PyTypeObject]()
         return _unsafe_alloc_init(type_obj_ptr, alloc^)
 
     # TODO(MSTDL-715):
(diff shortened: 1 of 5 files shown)


laszlokindrat (approved):
Nice, thank you!

laszlokindrat:
!sync

mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -1461,16 +1456,14 @@ struct PythonObject(
             If `T` has not been bound to a Python type object.
         """
         var cpython = Python().cpython()
-        var type = PyObjectPtr(
-            cpython.Py_TYPE(self._obj_ptr).bitcast[PyObject]()
-        )
 
-        var expected_type_obj = lookup_py_type_object[T]()
+        var type = PyObjectPtr(upcast_from=cpython.Py_TYPE(self._obj_ptr))
+        var expected_type = lookup_py_type_object[T]()._obj_ptr
```
ConnorGray, line 1461:
```mojo
        var expected_type = lookup_py_type_object[T]()._obj_ptr
```

**Question** Does this accidentally  introduce a use-after-free bug? `lookup_py_type_object()` returns a `PythonObject`, but then we access the raw underlying pointer in `_obj_ptr` without keeping the `PythonObject` alive.

mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -1491,15 +1484,9 @@ struct PythonObject(
...
-        var obj_ptr = self._unchecked_downcast_object_ptr[PyMojoObject[T]]()
-
+        ref obj = self._obj_ptr.bitcast[PyMojoObject[T]]()[]
         # TODO(MSTDL-950): Should use something like `addr_of!`
-        return UnsafePointer[T](to=obj_ptr[].mojo_value)
-
-    @always_inline
-    fn _unchecked_downcast_object_ptr[T: AnyType](var self) -> UnsafePointer[T]:
-        """Assume that this Python object contains a wrapped Mojo value."""
-        return self._obj_ptr.unsized_obj_ptr.bitcast[T]()
```
ConnorGray, line 1502: **Suggestion** What do you think about keeping this function? I think that adding `PyObjectPtr.bitcast()` is valuable to simplify the low-level code, but I think having a way to downcast the higher-level `PythonObject` type is valuable as well.

I think ideally, most places would downcast a `PythonObject`, so they get the RAII benefits of keeping ownership managed by `PythonObject`, without having to dip into `PyObjectPtr`.

Right now it may be the case that most uses of downcasting are in our internal bindings code, but we should remember that user downcasting of `PythonObject` (for example to a numpy.array / `PyArrayObject`) is something we want to make ergonomic as well.

ConnorGray (commented):
Hey @soraros, thanks for all the great cleanup work you've been doing on our bindings recently! 🙂

The changes here are no exception, great work—I particularly love using `ref` in more places to avoid those `[]` derefs.

I did have one note about removing the `PythonObject._unchecked_downcast_object_ptr()` method (see comments in code), which I think would be beneficial to retain. However I'm open to hearing what you're plans here are as well 🙂

mojo/stdlib/stdlib/python/python_object.mojo
soraros: I don't really feel the *need* to *remove* this method, it's removed simply because it was mechanically possible. What you said makes sense and let's keep `_unchecked_downcast_object_ptr`.

mojo/stdlib/stdlib/python/python_object.mojo
soraros: You have a point, but I don't think this code is "more wrong" than the original.

soraros: Actually, I think it's fine. We return a copy of the `PyObj` from the global registry, which shares the same `_obj_ptr` as the one stored in (and kept alive by) the `MOJO_PYTHON_TYPE_OBJECTS` dict. Here, we temporarily increase and then decrease the refcount on the CPython side.

mojo/stdlib/stdlib/python/python_object.mojo
laszlokindrat: I actually do want to remove this since its implementation would now just become `return self._obj_ptr.bitcast[T]()`, which is both shorter, and more expressive than this shallow wrapper.

PR #4994, opened 2025-07-10
[stdlib][cpython-api] ABI hardening for CPython call protocol methods

soraros:


laszlokindrat (approved):
Thank you!

laszlokindrat:
Depends on https://github.com/modular/modular/pull/4991. !sync

Issue #1731, opened 2024-02-05
[Feature Request] Enhanced Compile-Time and Runtime Integration for Function Optimization in Mojo


YichengDWu:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### Introduction

The Mojo programming language, with its capacity to define compile-time parameters and runtime arguments, presents a unique opportunity for optimization. However, its current implementation requires explicit differentiation between compile-time and runtime contexts, limiting the potential for automatic optimization. This feature request seeks to address this limitation by proposing an enhancement for more intelligent compile-time constant detection and optimization in mixed contexts.

### Current Challenge

The existing Mojo language constructs allow for distinct handling of compile-time and runtime scenarios, as illustrated below:

```mojo
fn mul[l:Int, r:Int]() -> Int:
    @parameter
    if l==0 or r==0:
        return 0
    else:
        return l*r

fn mul(l: Int, r: Int) -> Int:
    return l*r
```

While this design supports compile-time optimizations, it introduces inefficiencies when dealing with mixed compile-time and runtime value scenarios. Consider the function overloads for `mul`:

```mojo
fn mul(lhs:Tuple[Int, Int], rhs: Tuple[Int, Int]) -> Tuple[Int, Int]:
    # Current implementation lacks optimization for compile-time known values within tuples
    return (mul(lhs.get[0, Int](), rhs.get[0](), mul(lhs.get[1,Int](), rhs.get[1, Int]()))
```

This implementation demonstrates a significant limitation: the inability to optimize based on compile-time constants when those constants are part of composite types like tuples. The distinction between compile-time and runtime optimizations becomes blurred, especially when handling mixed scenarios, leading to less efficient execution.

### Problem Illustration

In practical terms, the language's current model does not support a unified method for optimizing functions that might receive both compile-time constants and runtime variables. This limitation is evident when attempting to optimize operations on tuples that mix compile-time constants and runtime values:

```mojo
alias l = 4
alias r = 5
let a = 6  # runtime variable
let b = 7  # runtime variable

let lhs = (l, a) # mix of compile-time constant 'l' and runtime variable 'a'
let rhs = (r, b) # mix of compile-time constant 'r' and runtime variable 'b'

let result = mul(lhs, rhs) # runtime call without compile-time optimization for 'l' and 'r'
```

[...]

PR #4997, opened 2025-07-10
[stdlib][Proposal] Arguments can be parameters too without duplicating the function signature

### LINKED ISSUE #4887: [Feature Request] Opportunistic Lifting to Parameter Time
### Review Mojo's priorities

### What is your request?

A mechanism which allows developers to take advantage of things which happen to be compile-time known, allowing them to be "lifted" to parameter time. 

ex:

```mojo
fn broadcast[dtype: DType](mut a: Matrix[dtype], value: Scalar[dtype]):
    alias comptime_value: Optional[Scalar[dtype]] = __get_comptime_if_known(value)

    @parameter 
    if comptime_value and comptime_value.value() == 0:
        memset_zero(a.unsafe_pointer(), a.num_elements())
        return
    else:
        for i in range(matrix.dims[0]):
            for j in range(matrix.dims[0])
                matrix[i, j] = value
```

Being able to "lift" arguments to compile time if they are compile time known is useful for APIs where users may have the information at compile time or it may be at runtime. Currently, the only option is to duplicate the function, which causes a combinatorial explosion as more and more arguments can be used for optimizations. For instance, doing this just for matrix dimensions results in 4 functions. There are solutions like what is used in `LayoutTensor`, however, those are verbose and require passing around type wrappers. 

### Any other details?

I don't have strong feelings on the syntax, so whatever gives this functionality and is most convenient to implement in the compiler is fine.

This should be able to handle the user juggling an arbitrary number of potentially compile-time arguments.

gabrieldemarmiesse:
Closes https://github.com/modular/modular/issues/4887

Proposal can be found [here](https://github.com/gabrieldemarmiesse/mojo/blob/parameter_or_arg/mojo/proposals/parameter-or-arg.md).

The implementation is provided in this pull request with unit tests. We can change names as usual if the maintainers don't like them.

I have chosen to not add the docstrings in this PR to improve readability and agility. I can add them just before merging if the proposal is approved.


owenhilyard:
I don't like the dependency on `T: Copyable & Movable`, any thoughts on returning a ref instead and have other functions for "by value" retrieval? 

This is good to have for things that could be either runtime or compile-time configured, such as the width of a btree node (assuming we get unsized types), and could be used to collapse `List`, `InlineVector` (variable number of active members but all inline allocations), and `InlineArray` into a single thing. 

The reason I asked about this in #4887 is because I think that, while it is possible to do it this way, there are substantial ergonomics problems. My vision was more for an "enhanced constant folding" interface where you could take constant folded items and go do arbitrary computations with them at compile time to enable better runtime performance. This does work, but extensive use of it would mean that the stdlib would be littered with this construct. #1731 has a similar desire. 

Ideally, I would like the ability to transition large swaths of the stdlib over to a "maybe-parameter" solution (All indexing, any numeric function which takes multiple arguments, etc). If we do that with this, then we end up with a big ergonomics problem, which is why I was going for the "manual constant folding" angle. I don't think we have a good way around this for struct parameters/constructor args which need to hang around for a while, but for function arguments we can probably improve the situation with some compiler work.

gabrieldemarmiesse:
>I don't like the dependency on T: Copyable & Movable, any thoughts on returning a ref instead and have other functions for "by value" retrieval?

Totally agree with you there. I wonder if that was going to complexify the proposal with lifetimes and if (that I'm not sure) there is no performance penalty by using a `ref` when working with register_passable types. Any insight on that could help. We can maybe restrict the struct to register_passable types at first and start from there?

> If we do that with this, then we end up with a big ergonomics problem

Are you talking about the verbose syntax? Do you see things that can't be improved with some compiler work/tooling work?

I think we all want to go in the same general direction. We'll likely start by clarifying the mental model, and then build on this. I'm proposing a mental model which I believe is simple enough. How much we want to invest in ergonomics and when is another question.

christoph-schlumpf:
Here is an alternative take on the proposal based on the same Idea.
The difference is that the parameter and argument side are still kept apart but are mutual exclusive. This might be more flexible and avoids having to provide a parameter on the argument side.

```mojo
# Proposal
flexible_memcpy(dst, src, Parameter[64]())

flexible_memcpy(dst, src, get_buffer_size())

# Alternative
flexible_memcpy[64](dst, src)

flexible_memcpy(dst, src, get_buffer_size())
```

Here the code for this alternative:

```mojo
struct ArgOrParam[T: Copyable & Movable, //, param: Optional[T] = {}](
    Copyable, Movable
):
    var _value: InlineArray[T, 0 if param else 1]

    @implicit
    fn __init__(out self, owned value: T):
        constrained[not param, "param has value"]()
        self._value = __type_of(self._value)(value)

    fn __init__(out self):
        constrained[Bool(param), "param is missing"]()
        self._value = __type_of(self._value)(uninitialized=True)

    @always_inline
    fn value(self) -> T:
        @parameter
        if param:
            return param.value()
        else:
            return self._value[0]


fn some_func[x_param: Optional[String]](x: ArgOrParam[x_param] = {}):
    @parameter
    if x.param:
        print("Is comptime!! Value is:")
        alias a = x.param.value()
        # I can use the alias here
        print(a)
    else:
        print("Is runtime!! Value is:")
        print(x.value())

    print(
        "I don't care if it's comptime or runtime, here is the value:",
        x.value(),
    )


fn main():
    print("----")
    some_func["COMPTIME"]()
    print("----")
    some_func("RUNTIME")
    print("----")
    # some_func["COMPTIME"]("RUNTIME")  # This would not compile
```

This also allows for default values:
```mojo
fn some_func[
    x_param: Optional[String] = "DEFAULT"
](x: ArgOrParam[x_param] = {}):

...

fn main():
    some_func["COMPTIME"]()
    print("----")
    some_func("RUNTIME")
    print("----")
    some_func()
```

gryznar:
> ## Proposal
> flexible_memcpy(dst, src, Parameter\[64\]()) > flexible_memcpy(dst, src, getDynamicValue()) 
>
> ## Alternative 
> flexible_memcpy[64](dst, src) 
> flexible_memcpy(dst, src,  getDynamicValue())

Despite improper usage of camelCase (should be snake_case), I like this alternative much more. This usage looks more natural to me than wrapping compime value with `Parameter`

christoph-schlumpf:
> Despite improper usage of camelCase …

Fixed. Java/Kotlin and TypeScript is my daily business 😉

Brian-M-J:
Is this orthogonal to maybe adding [multi-stage programming](https://en.wikipedia.org/wiki/Multi-stage_programming) in the future or do we have to hash out the possibility now?

christoph-schlumpf:
@gabrieldemarmiesse:  After looking again at your proposal I saw that your proposal supports both syntax versions too. So both implementations are similar in that regard.

@owenhilyard : I think it is only possible to return a `ref` for runtime values. Maybe `parameters` do not have an origin that can be referenced?

Here an example that uses both syntax versions and  `ref`  for runtime values. I introduced an alias `Param[T]` which avoids the parentheses compared to the proposed struct `Parameter[T]()`.

```mojo
struct MaybeParam[T: Copyable & Movable, //, param: Optional[T] = {}](
    Copyable, Movable
):
    var _value: InlineArray[T, 0 if param else 1]

    @implicit
    fn __init__(out self, owned value: T):
        constrained[not param, "param has value"]()
        self._value = __type_of(self._value)(value)

    fn __init__(out self):
        constrained[Bool(param), "param is missing"]()
        self._value = __type_of(self._value)(uninitialized=True)

    @always_inline
    fn runtime_value(self) -> ref [self._value[0]] T:
        constrained[not param, "param has value"]()
        return self._value[0]

    @always_inline
    fn value(self) -> T:
        @parameter
        if param:
            alias value = self.param.value()
            return value
        else:
            return self._value[0]


alias Param[T: Copyable & Movable, //, O: Optional[T]] = MaybeParam[O]()


fn volume[
    length_param: Optional[Int],
    width_param: Optional[Int],
    height_param: Optional[Int],
](
    length: MaybeParam[length_param] = {},
    width: MaybeParam[width_param] = {},
    height: MaybeParam[height_param] = {},
) -> Int:
    """One function instead of 8 permutations."""

    @parameter
    if length.param and width.param and height.param:
        alias volume = length.param.value() * width.param.value() * height.param.value()
        print("Comptime volume:", volume)
        return volume

    @parameter
    if length.param and width.param:
        alias area = length.param.value() * width.param.value()
        print("Comptime length*width area:", area)
        var volume = area * height.runtime_value()
        print("Runtime volume:", volume)
        return volume

    @parameter
    if length.param and height.param:
        alias area = length.param.value() * height.param.value()
        print("Comptime length*height area:", area)
        var volume = area * width.runtime_value()
        print("Runtime volume:", volume)
        return volume

    @parameter
    if width.param and height.param:
        alias area = width.param.value() * height.param.value()
        print("Comptime width*height area:", area)
        var volume = area * length.runtime_value()
        print("Runtime volume:", volume)
        return volume

    @parameter
    if not length.param and not width.param and not height.param:
        var volume = (
            length.runtime_value()
            * width.runtime_value()
            * height.runtime_value()
        )
        print("Runtime volume:", volume)
        return volume

    print("Only one parameter is provided, using its value directly.")
    var volume = length.value() * width.value() * height.value()
    print("Runtime volume:", volume)
    return volume


fn main():
    print("----")
    var v1 = volume[10, 5, 2]()
    print("----")
    var v2 = volume(10, 5, 2)
    print("----")
    var v3 = volume[10]({}, 5, 2)
    print("----")
    var v4 = volume[{}, 5, 2](10)
    print("----")
    var v5 = volume[10, 5]({}, {}, 2)
    print("----")
    # Alternative syntax with Param
    var v6 = volume(Param[10], 5, Param[2])
    print("----")
    print("Total volume:", v1 + v2 + v3 + v4 + v5 + v6)
```

owenhilyard:
> > I don't like the dependency on T: Copyable & Movable, any thoughts on returning a ref instead and have other functions for "by value" retrieval?
> 
> Totally agree with you there. I wonder if that was going to complexify the proposal with lifetimes and if (that I'm not sure) there is no performance penalty by using a `ref` when working with register_passable types. Any insight on that could help. We can maybe restrict the struct to register_passable types at first and start from there?

I think that we could have `Movable`, which would require an owned `self` and effectively destroy the "container", and `Copyable`, where you get handed copies. Both are fine individually. `ref` is good because it works for all types, and iirc it will pass by value for immutable refs and will do pass by value for mutable ones as well if the compiler can optimize it. 

> > If we do that with this, then we end up with a big ergonomics problem
> 
> Are you talking about the verbose syntax? Do you see things that can't be improved with some compiler work/tooling work?
> 
> I think we all want to go in the same general direction. We'll likely start by clarifying the mental model, and then build on this. I'm proposing a mental model which I believe is simple enough. How much we want to invest in ergonomics and when is another question.

My hope was to jump directly to the compiler supported model, but I can see your point about wanting this to live outside the compiler for a bit.

owenhilyard:
> Is this orthogonal to maybe adding [multi-stage programming](https://en.wikipedia.org/wiki/Multi-stage_programming) in the future or do we have to hash out the possibility now?

I think we leave that bag of worms for the reflection API discussion.

owenhilyard:
> @owenhilyard : I think it is only possible to return a `ref` for runtime values. Maybe `parameters` do not have an origin that can be referenced?

That is a bit of a footgun, and something we might want to fix. It should be possible to pull an appropriate origin out of thin air when aliases are materialized.

lattner:
This proposal makes a lot of sense conceptually to me - I think it would be a valuable addition to the standard library.  Maybe a name like `Specializable[T]` would be more natural.  There are other refinements that seem possible as well. @JoeLoser have you taken a look at this?

christoph-schlumpf:
@owenhilyard : I updated the implementation to always return a `ref` value reference, removing the footgun you mentioned.

@lattner : I tried renaming the struct to `Specializable`. But not sure if the name is natural because I'm not a native English speaker.

@gabrieldemarmiesse: I hope this is useful for the proposal. Let me know if I can support you in any other way. 

```mojo
struct Specializable[T: Copyable & Movable, //, param: Optional[T] = {}](
    Copyable, Movable
):
    """A struct that can either hold a runtime value or a comptime parameter."""

    var _value: InlineArray[T, 0 if param else 1]

    @implicit
    fn __init__(out self, owned value: T):
        constrained[not param, "param has value"]()
        self._value = __type_of(self._value)(value)

    fn __init__(out self):
        constrained[Bool(param), "param is missing"]()
        self._value = __type_of(self._value)(uninitialized=True)

    @always_inline
    fn value(self) -> ref [self._value, StaticConstantOrigin] T:
        @parameter
        if param:
            alias ptr = UnsafePointer(to=param.value()).origin_cast[
                False, StaticConstantOrigin
            ]()
            return ptr[]
        else:
            return self._value[0]


alias Parameter[T: Copyable & Movable, //, O: Optional[T]] = Specializable[O]()


fn volume[
    length_param: Optional[Int],
    width_param: Optional[Int],
    height_param: Optional[Int],
](
    length: Specializable[length_param] = {},
    width: Specializable[width_param] = {},
    height: Specializable[height_param] = {},
) -> Int:
    """One function instead of 8 permutations."""

    @parameter
    if length.param and width.param and height.param:
        alias volume = length.param.value() * width.param.value() * height.param.value()
        print("Comptime volume:", volume)
        return volume

    @parameter
    if length.param and width.param:
        alias area = length.param.value() * width.param.value()
        print("Comptime length*width area:", area)
        var volume = area * height.value()
        print("Runtime volume:", volume)
        return volume

    @parameter
    if length.param and height.param:
        alias area = length.param.value() * height.param.value()
        print("Comptime length*height area:", area)
        var volume = area * width.value()
        print("Runtime volume:", volume)
        return volume

    @parameter
    if width.param and height.param:
        alias area = width.param.value() * height.param.value()
        print("Comptime width*height area:", area)
        var volume = area * length.value()
        print("Runtime volume:", volume)
        return volume

    print("Less than 2 parameters provided. Calculating at runtime.")
    var volume = length.value() * width.value() * height.value()
    print("Runtime volume:", volume)
    return volume


fn main():
    print("----")
    var v1 = volume[10, 5, 2]()
    print("----")
    var v2 = volume(10, 5, 2)
    print("----")
    var v3 = volume[10]({}, 5, 2)
    print("----")
    var v4 = volume[{}, 5, 2](10)
    print("----")
    var v5 = volume[10, 5]({}, {}, 2)
    print("----")
    # Alternative syntax with Parameter
    var v6 = volume(Parameter[10], 5, Parameter[2])
    print("----")
    print("Total volume:", v1 + v2 + v3 + v4 + v5 + v6)
```

I think to remove the requirement on `Copyable & Movable` types, `Optional`, `InlineArray` and `Variant` would have to be rewritten too.

gabrieldemarmiesse:
@christoph-schlumpf let's wait until we have a proper review from the stdlib team, we'll update the document then. Thanks for the help!

owenhilyard:
My only concern with making this a library feature instead of a language one is the potential ergonomics impact. While this is a very valuable feature for performance, since it lets us keep as much in the parameter domain as possible, I have some concerns about usability if we were to go replace everything that can be replaced with this. That would mean every function argument in the stdlib which isn't necessarily only runtime known (ex: file descriptors) would use this. For example, using this to generalize indexing in `LayoutTensor` so that if the layout is known enough to determine the offset (ex: one dimension may be unknown but if it may still be possible to determine the offset into the tensor's backing memory) then that happens at compile time without relying on constant folding. The downside is that this requires every index into `LayoutTensor` to construct a `Specializable`, which isn't that bad but may start to get annoying if you're doing a lot of compile-time known indexing. 

Given that nobody else seems to have these concerns, let's move ahead with the stdlib team review and we can hold discussion about making this a language feature until we see if it is actually a large enough usability issue to warrant dedicated support in the language. 

Given that this does look like a very weird construct without explanation, I think we should also discuss how to describe it. I want to call it "manual constant folding", but that might require too much compiler background of the average Mojo developer. 

For the stdlib team, I think some discussion on whether using this more general mechanism to implement the runtime vs static layout handing in `LayoutTensor` is warranted. Right now there seem to be a lot of footguns in trying to use runtime layout tensors which this could alleviate, but that's also going to cause a lot of kernel code churn. If this isn't a solution that works for `LayoutTensor` and friends, then I think we should discuss why that is because `LayoutTensor` is a good example of something that should benefit greatly from this because it has a lot of parts which can be runtime or comptime.  I'd also like to put whether it's valuable to use this to unify `List`, `InlineArray` and the old `InlineVector` into a single `List` construct with a flag to disable the allocator if the capacity is fixed at compile time, since  you might still want to put the allocation on the heap if it's a large enough allocation.

laszlokindrat:
@gabrieldemarmiesse Thanks for the proposal and the implementation! This looks like a really cool and powerful feature to have. We discussed it with the team and decided that we want to wait a little bit before committing to a feature like this. We see a few use cases that could benefit from this, but we also have a couple of concerns that we are not ready to address yet:
- This is a non-trivial feature that needs really good explanation and guidelines on when to use, and when to stick to just arguments or parameters.
- If this proliferates, the compile time implications can be severe. (This goes to my previous point too; we need to understand when it's worth doing this vs just forcing compile time or runtime.)
- We want to have discussions with the compiler team, in case we can do this with improved syntax (and maybe performance).

We propose the following: can you turn this into a [community package](https://github.com/modular/modular-community) so that people can start using this in their projects? We are really curious how this could be used in the wild. We would also like to circle back to this later, maybe in 2025Q4.

gabrieldemarmiesse:
As requested, I made a community package. The instructions to install are here: https://github.com/gabrieldemarmiesse/mojo-parameter-or-arg#installation

Happy to take PRs about it, notably to remove the `Copyable & Movable`, and show that there is no perf impact when doing so.

kartva:
@gabrieldemarmiesse do you think you should close out the PR now for the time being? It looks like everyone has decided on a path forward, which doesn't involve merging this PR soon.

gabrieldemarmiesse:
Sure thing, let's clean this up :)

lattner:
Thank you all for the help closing off issues!

PR #5000, opened 2025-07-10
[stdlib] [proposal] Make logarithm function generic over base, etc.

martinvuyk:
Make logarithm function generic over base, etc.

My main goal is to enable most optimizations by having the user only decide over the particular "sacrifices" they are willing to make for more speed/precision.

CC: @soraros @laszlokindrat WDYT?

### DIFF
--- a/mojo/stdlib/stdlib/bit/bit.mojo
+++ b/mojo/stdlib/stdlib/bit/bit.mojo
@@ -19,6 +19,7 @@ from bit import count_leading_zeros
 ```
 """
 
+from bit._mask import is_negative
 from sys import llvm_intrinsic, sizeof
 from sys.info import bitwidthof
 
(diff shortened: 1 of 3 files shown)


mojo/stdlib/stdlib/bit/bit.mojo
```diff
@@ -354,15 +355,38 @@ fn log2_floor(val: Int) -> Int:
 
     Returns:
         The floor of the base-2 logarithm of the input value, which is equal to
-        the position of the highest set bit. Returns -1 if val is 0.
+        the position of the highest set bit. Returns -1 if val is 0 or negative.
     """
-    if val == 32:
-        return 5
-    if val == 64:
-        return 6
```
laszlokindrat: This microoptimization was added explicitly to speed up these cases, so can you please add it back? You can probably also add it to the SIMD overload, but for the Int overload it is needed for sure.

mojo/stdlib/stdlib/math/math.mojo
```diff
@@ -809,36 +809,67 @@ fn _log_base[
...
+        return log2_floor(x).cast[out_dtype]()
+    elif dtype.is_integral():
+        alias log2base = log2(base)
+        return (log2_floor(x).cast[calc_dtype]() / log2base).cast[out_dtype]()
+    else:
+        # TODO: benchmark. supposedly due to IEEE floats being stored in log10
+        # in binary they are faster to compute for that logarithm. I found no
+        # reliable sources
+        alias log10base = log10(base).cast[calc_dtype]()
+        return (log(x.cast[calc_dtype]()) / log10base).cast[out_dtype]()
```
laszlokindrat: I have concerns about precision here. Wouldn't it be better to approximate directly? @abduld can you provide some insight?

mojo/stdlib/stdlib/bit/bit.mojo
martinvuyk: I'll add them back for `Int`. But I would like to actually benchmark this to make sure it is worth it

mojo/stdlib/stdlib/math/math.mojo
martinvuyk: Sidenote: `log()` is wrong here, I always get confused because in my mind `log` is base 10. I fixed it now.

What other approximation methods are there other than base change for logarithms? I'm not familiar with this space, I just wanted a generic base algorithm.

PR #5001, opened 2025-07-10
[stdlib] Add `log2_floor` for `SIMD` and optimize

martinvuyk:
Add `log2_floor` for `SIMD` and optimize. Split off from  #5000

### DIFF
--- a/mojo/stdlib/stdlib/bit/bit.mojo
+++ b/mojo/stdlib/stdlib/bit/bit.mojo
@@ -19,6 +19,7 @@ from bit import count_leading_zeros
 ```
 """
 
+from bit._mask import is_negative
 from sys import llvm_intrinsic
 from sys.info import bitwidthof
 
@@ -63,9 +64,14 @@ fn count_leading_zeros[
         leading zeros at position `i` of the input value.
     """
     constrained[dtype.is_integral(), "must be integral"]()
-    return llvm_intrinsic["llvm.ctlz", __type_of(val), has_side_effect=False](
-        val, False
+
+    # HACK(#5003): remove this workaround
+    alias d = dtype if dtype is not DType.index else (
+        DType.int32 if dtype.sizeof() == 4 else DType.int64
     )
+    return llvm_intrinsic["llvm.ctlz", SIMD[d, width], has_side_effect=False](
+        val.cast[d](), False
+    ).cast[dtype]()
 
 
 # ===-----------------------------------------------------------------------===#
@@ -354,14 +360,13 @@ fn log2_floor(val: Int) -> Int:
 
     Returns:
         The floor of the base-2 logarithm of the input value, which is equal to
-        the position of the highest set bit. Returns -1 if val is 0.
+        the position of the highest set bit. Returns -1 if val is 0 or negative.
     """
-    alias bitwidth = bitwidthof[Int]()
-    return select(val <= 1, 0, bitwidth - count_leading_zeros(val) - 1)
+    return Int(log2_floor(Scalar[DType.index](val)))
 
 
 @always_inline
-fn log2_floor(val: Scalar) -> __type_of(val):
+fn log2_floor(val: UInt) -> UInt:
     """Returns the floor of the base-2 logarithm of an integer value.
 
     Args:
@@ -369,13 +374,38 @@ fn log2_floor(val: Scalar) -> __type_of(val):
 
     Returns:
         The floor of the base-2 logarithm of the input value, which is equal to
-        the position of the highest set bit. Returns -1 if val is 0.
+        the position of the highest set bit. Returns UInt.MAX if val is 0.
     """
-    constrained[val.dtype.is_integral(), "the input dtype must be integral"]()
-    alias bitwidth = bitwidthof[val.dtype]()
-    return select(
-        val <= 1, __type_of(val)(0), bitwidth - count_leading_zeros(val) - 1
-    )
+    return bitwidthof[UInt]() - count_leading_zeros(val) - 1
+
+
+@always_inline
+fn log2_floor[
+    dtype: DType, width: Int, //
+](val: SIMD[dtype, width]) -> SIMD[dtype, width]:
+    """Returns the floor of the base-2 logarithm of an integer value.
+
+    Parameters:
+        dtype: The `dtype` of the input SIMD vector.
+        width: The width of the input and output SIMD vector.
+
+    Args:
+        val: The input value.
+
+    Returns:
+        The floor of the base-2 logarithm of the input value, which is equal to
+        the position of the highest set bit. Returns -1 if val is 0 or negative.
+    """
+    constrained[dtype.is_integral(), "dtype must be integral"]()
+
+    alias bitwidth = bitwidthof[dtype]()
+    var res = bitwidth - count_leading_zeros(val) - 1
+
+    @parameter
+    if dtype.is_signed():
+        return res | is_negative(val)
+    else:
+        return res
 
 
 # ===-----------------------------------------------------------------------===#
--- a/mojo/stdlib/test/bit/test_bit.mojo
+++ b/mojo/stdlib/test/bit/test_bit.mojo
@@ -487,7 +487,6 @@ fn _log2_ceil(n: Scalar) -> __type_of(n):
 
 
 def test_log2_floor():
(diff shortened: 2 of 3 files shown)


mojo/stdlib/stdlib/bit/bit.mojo
```diff
@@ -354,15 +355,38 @@ fn log2_floor(val: Int) -> Int:
...
+
+    Args:
+        val: The input value.
+
+    Returns:
+        The floor of the base-2 logarithm of the input value, which is equal to
+        the position of the highest set bit. Returns -1 if val is 0 or negative.
+    """
+    constrained[dtype.is_integral(), "dtype must be integral"]()
+
```
soraros, line 400: Maybe add a `debug_assert(val > 0)`.

martinvuyk: True, nice catch

mojo/stdlib/stdlib/bit/bit.mojo
```diff
@@ -354,15 +355,39 @@ fn log2_floor(val: Int) -> Int:
...
+    """
+    constrained[dtype.is_integral(), "dtype must be integral"]()
+    debug_assert(all(val > 0), "values should be bigger than 0")
+
+    alias bitwidth = bitwidthof[dtype]()
+    var res = bitwidth - count_leading_zeros(val) - 1
+
+    @parameter
+    if dtype.is_signed():
+        return res | is_negative(val)
```
soraros, line 406: I think `(val > 0).select(res, -1)` is a bit clearer (which doesn't rely implicitly on the fact that all-ones is -1).

martinvuyk: Yeah but I don't think it will have the same performance: Both are running `log2_floor(Int8(-1))`
`return (bitwidth - count_leading_zeros(val) - 1) | (val >> (bitwidth - 1))`
```asm
"example::log2_floor[::DType,::Int](::SIMD[$0, $1]),dtype=si8,width=1":
        movzbl  %dil, %eax
        lzcntl  %eax, %ecx
        addl    $-24, %ecx
        movb    $7, %dl
        subb    %cl, %dl
        sarb    $7, %al
        orb     %dl, %al
        retq
```
`(val > 0).select((bitwidth - count_leading_zeros(val) - 1), -1)`
```asm
"example::log2_floor[::DType,::Int](::SIMD[$0, $1]),dtype=si8,width=1":
        movzbl  %dil, %eax
        lzcntl  %eax, %ecx
        addl    $-24, %ecx
        movb    $7, %dl
        subb    %cl, %dl
        testb   %al, %al
        movzbl  %dl, %ecx
        movl    $255, %eax
        cmovgl  %ecx, %eax
        retq
```

soraros: I suspect they will have comparable performance, though it doesn't hurt to build the optimisation into the library. Please write a comment explaining that `is_neg` returns all-ones, which is `-1` in two's complement.

mojo/stdlib/stdlib/bit/bit.mojo
```diff
@@ -354,15 +355,39 @@ fn log2_floor(val: Int) -> Int:
...
+
+    Args:
+        val: The input value.
+
+    Returns:
+        The floor of the base-2 logarithm of the input value, which is equal to
+        the position of the highest set bit. Returns -1 if val is 0 or negative.
+    """
+    constrained[dtype.is_integral(), "dtype must be integral"]()
+    debug_assert(all(val > 0), "values should be bigger than 0")
```
soraros:
```suggestion
    debug_assert(all(val > 0), "input value should be positive")
```

soraros: Actually, I'm not so sure anymore. If we guarantee `-1` for `val <= 0`, maybe the `debug_assert` can be left out.

martinvuyk: True given we guarantee that. Sidenote: I don't think bigger than 0 means positive. 0+ and 0- exist in limit theorems, also in floats

mojo/stdlib/stdlib/bit/bit.mojo
martinvuyk: It is part of is_negative docstrings explicitly, IMO that's what IDEs are for

mojo/stdlib/stdlib/bit/bit.mojo
soraros: Haha. a) Aren't we in integer domain here? b) 0+/- are not numbers. c) positive is usually defined as "greater than 0" d) logarithm is usually said to be defined for "positive real something something" e) I think you are think of `sign` not positivity for signed zeros in floating point math.

martinvuyk: Hm, I guess it is an old debate whether zero is a natural number or not. And in the strictest sense it is a neutral position as the origin coordinate so neither positive nor negative :man_shrugging: . But I've seen way too many times people colloquially talk about positive numbers while still thinking about zero, so I would also prefer to keep being a bit more exact by using numbers instead of mathematical nuances in documentation.

mojo/stdlib/stdlib/bit/bit.mojo
```diff
@@ -63,9 +64,14 @@ fn count_leading_zeros[
         leading zeros at position `i` of the input value.
     """
     constrained[dtype.is_integral(), "must be integral"]()
-    return llvm_intrinsic["llvm.ctlz", __type_of(val), has_side_effect=False](
-        val, False
+
+    # HACK(#5003): remove this workaround
+    alias d = dtype if not (is_compile_time() and dtype is DType.index) else (
```
laszlokindrat: `is_compile_time()` will always evaluate to `True` here, because you are calling it in a parameter context.

mojo/stdlib/stdlib/bit/bit.mojo
```diff
@@ -354,15 +360,38 @@ fn log2_floor(val: Int) -> Int:
 
     Returns:
         The floor of the base-2 logarithm of the input value, which is equal to
-        the position of the highest set bit. Returns -1 if val is 0.
+        the position of the highest set bit. Returns -1 if val is 0 or negative.
     """
-    if val == 32:
-        return 5
-    if val == 64:
-        return 6
```
laszlokindrat: As I mentioned in the other PR, this microoptimization needs to be retained.

mojo/stdlib/test/bit/test_bit.mojo
```diff
@@ -476,19 +476,23 @@ fn _log2_floor(n: Int) -> Int:
...
             log2_floor(i),
             _log2_floor(i),
             msg=String("mismatching value for the input value of ", i),
         )
+    # test failures
+    assert_equal(log2_floor(0), -1)
+    assert_equal(log2_floor(-1), -1)
+    assert_equal(log2_floor(-2), -1)
+    assert_equal(log2_floor(-3), -1)
+    assert_equal(log2_floor(-999), -1)
```
laszlokindrat: Seems like this only tests `Int`? Can you add tests for SIMD types?

mojo/stdlib/stdlib/bit/bit.mojo
martinvuyk: True, fixed

laszlokindrat (approved):
LGTM, thank you!

laszlokindrat:
!sync

laszlokindrat:
@martinvuyk I'm seeing some test failures internally, can you please rebase this and check if anything needs fixing?

martinvuyk:
@laszlokindrat I think there is something wrong with the bazel build system, it might not be using the current stdlib in the local development branch for us OS contributors. I'm adding print statements in the implementation but the tests still seem to execute the old version that made the code return 0 when the value is `<= 1`, the statements aren't printing anything and I can find literally no way this code is that wrong (and these tests were passing with the previous build system).

martinvuyk:
@laszlokindrat my bad, there was an overload for scalar types that I hadn't seen

laszlokindrat:
@martinvuyk Could you please rebase on the latest so I can try to bring it in again? Thanks!

laszlokindrat:
!sync

Issue #3520, opened 2024-09-21
[Feature Request] [stdlib] A Foreign Function Interfacing exclusive package (`ffi`)


martinvuyk:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Reorganize and polish `sys/ffi.mojo` and take it out to be it's own package with it's own set of ease of use capabilities for using Mojo as the glue between languages.

For all major languages for which support is judged as worth being added, adding basic type conversions etc. In the case of C, also adding the basic POSIX functions (libc).

An excerpt from an [example implementation](https://github.com/martinvuyk/forge-tools/blob/main/src/forge_tools/ffi/c.mojo):
```mojo
# Adapted from https://github.com/crisadamo/mojo-Libc which doesn't currently
# (2024-07-22) have a licence, so I'll assume MIT licence.
# Huge thanks for the work done.


struct C:
    """C types. This assumes that the platform is 32 or 64 bit, and char is
    always 8 bit (POSIX standard).
    """

    alias char = Int8
    """C type: `char`. The signedness of `char` is platform specific. Most
    systems, including x86 GNU/Linux and Windows, use `signed char`, but those
    based on PowerPC and ARM processors typically use `unsigned char`."""
    alias s_char = Int8
    """C type: `signed char`."""
    alias u_char = UInt8
    """C type: `unsigned char`."""
    alias short = Int16
    """C type: `short`."""
    alias u_short = UInt16
    """C type: `unsigned short`."""
    alias int = Int32
    """C type: `int`."""
    alias u_int = UInt32
    """C type: `unsigned int`."""
    alias long = Int64
    """C type: `long`."""
    alias u_long = UInt64
    """C type: `unsigned long`."""
    alias long_long = Int64
    """C type: `long long`."""
    alias u_long_long = UInt64
    """C type: `unsigned long long`."""
    alias float = Float32
    """C type: `float`."""
    alias double = Float64
    """C type: `double`."""
    alias void = Int8
    """C type: `void`."""

...

fn socket(domain: C.int, type: C.int, protocol: C.int) -> C.int:
    """Libc POSIX `socket` function.

    Args:
        domain: Address Family see AF_ alises.
        type: Socket Type see SOCK_ alises.
        protocol: Protocol see IPPROTO_ alises.

    Returns:
        A pointer to a socket.

    Notes:
[...]

PR #5025, opened 2025-07-15
[stdlib] Add `get_errno`, `set_errno`, and refactor module structure

### LINKED ISSUE #3921: [Feature Request] Errno Getter
### Review Mojo's priorities

### What is your request?

A getter for errno, likely in the form `fn raw_errno() -> c_int` as a runtime builtin.

Many, many, **many** POSIX functions use errno to return error codes, and checking it is necessary for proper error handling. We can't give coherent errors to the user without this. As far as I am aware, there is no way to access a C global variable from Mojo, so it needs to be a runtime builtin. 

### Any other details?

The reason I'm asking for `raw_errno` is so we can wrap the function with something that returns an enum for an easier to use "errno" function. A general version of this solution would be a way to use C ABI globals directly by symbol name, such as some kind of "extern" variable modifier.

martinvuyk:
Add `get_errno`, `set_errno`, and refactor module structure. Closes #3921. Part of #3520 and #3766

CC: @JoeLoser @owenhilyard @soraros


mojo/stdlib/stdlib/sys/ffi/c/constants.mojo
```diff
@@ -0,0 +1,1401 @@
...
+alias _mac_or_linux = os_is_macos() or os_is_linux()
+alias _windows = os_is_windows()
+alias _mac = os_is_macos()
+alias _linux = os_is_linux()
+
+# ===----------------------------------------------------------------------=== #
+# Error constants (errno.h)
+# ===----------------------------------------------------------------------=== #
+
+alias SUCCESS = 0
```
owenhilyard, line 32: Can we package this up in an enum?

mojo/stdlib/stdlib/sys/ffi/c/constants.mojo
```diff
@@ -0,0 +1,1401 @@
...
+alias EHWPOISON = 133
+"""Memory page has hardware error."""
+
+
+# ===----------------------------------------------------------------------=== #
+# Networking constants
+# ===----------------------------------------------------------------------=== #
+
+# Address Family Constants
+alias AF_UNSPEC = 0
```
owenhilyard, line 305: Same as error values, please put it in an enum.

mojo/stdlib/stdlib/sys/ffi/c/constants.mojo
```diff
@@ -0,0 +1,1401 @@
...
+alias AF_VSOCK = 40
+"""Constant: AF_VSOCK."""
+alias AF_KCM = 41
+"""Constant: AF_KCM."""
+alias AF_QIPCRTR = 42
+"""Constant: AF_QIPCRTR."""
+alias AF_MAX = 43
+"""Constant: AF_MAX."""
+
+alias PF_UNSPEC = AF_UNSPEC
```
owenhilyard, line 394: Enums for everything else in here too please!

mojo/stdlib/stdlib/sys/ffi/c/constants.mojo
```diff
@@ -0,0 +1,1401 @@
...
+"""Xerox NS IDP."""
+alias IPPROTO_RDP = 27
+""""Reliable datagram" protocol."""
+alias IPPROTO_ISO_TP4 = 29
+"""ISO Transport Protocol class 4 [RFC905]."""
+alias IPPROTO_DCCP = 33
+"""Datagram Congestion Control Prot. [RFC4340]."""
+alias IPPROTO_XTP = 36
+"""Xpress Transfer Protocol."""
+alias IPPROTO_DDP = 37
```
owenhilyard, line 540: I know Windows doesn't support all of these protocols, so I think that separating these out by OS might make more sense.

owenhilyard (commented):
I think we may want to take a step back and look at segmenting some of this a bit more strongly by OS.

mojo/stdlib/stdlib/sys/ffi/c/types.mojo
```diff
@@ -0,0 +1,105 @@
...
+
+    @parameter
+    if is_64bit() and os_is_windows():
+        return DType.uint32  # LLP64
+    elif is_64bit():
+        return DType.uint64  # LP64
+    elif is_32bit():
+        return DType.uint32  # ILP32
+    else:
+        constrained[False, "size of C `long long` is unknown on this target"]()
```
owenhilyard, line 104: This should be `unsigned long`, `long long` and `unsigned long long` are different types.

owenhilyard:
As I said in the review, we'll need to be a bit more careful handling OS differences. Even for this PR I'm not quite sure that musl guarantees that errno is in the same place as it is in glibc, and I think we will want musl compatibility in the future to enable statically linked binaries. Some stuff can be unified for POSIX, but especially windows vs unix is going to have some major differences.

martinvuyk:
> I think we may want to take a step back and look at segmenting some of this a bit more strongly by OS.

@owenhilyard thanks for the feedback, changed it to draft to enable more discussion and to be clear that I don't pretend this to be merged immediately. I expect a lot of discussion on some of the things I've done here.

The enums are a good idea, they should also enable the type itself to require certain operating systems which is cool and convenient.

> I'm not quite sure that musl guarantees that errno is in the same place as it is in glibc, and I think we will want musl compatibility in the future to enable statically linked binaries.

Yes these things are a pain. But at least we'll have a way to test whether they work without being surprised.

martinvuyk:
@jackos any interest in this PR? given 2d8d2e065c02794bd1c39c6ebdd6166a6483c1ac . And FYI, the error messages you added aren't written exactly like what strerror returns. Also, I dissagree on hardcoding the constant values instead of adding a constant that is determined at compile time (where we can add better specific OS quirks later on)

mojo/stdlib/stdlib/sys/ffi/c/types.mojo
```diff
@@ -0,0 +1,105 @@
...
+from os import abort
+from utils import StaticTuple
+from memory import memcpy, UnsafePointer
+
+# ===----------------------------------------------------------------------=== #
+# Base Types
+# ===----------------------------------------------------------------------=== #
+
+
+struct C:
```
jackos, line 26: Can we keep this separate from this errno PR? Needs discussion on if this:

```
from sys.ffi.c import C

UnsafePointer[C.c_char]
```
Is better than just:
```
from sys.ffi.c import c_char

UnsafePointer[c_char]
```

mojo/stdlib/stdlib/sys/ffi/c/constants.mojo
jackos: Agree with @owenhilyard here, just writing a design doc, and will share how I think these libc enums should be modelled so you can comment

jackos:
@martinvuyk very interested in this yeah thanks, need to get the structure of how we interact with libc correct

jackos:
Hi @martinvuyk [updated errno handling with an enum here](https://github.com/modular/modular/blob/0a2c1586d482cde14b505b851a5a04bfe3b4ebbd/mojo/stdlib/stdlib/sys/_libc_errno.mojo).

Added a new [sys.info.platform_map](https://github.com/modular/modular/blob/0a2c1586d482cde14b505b851a5a04bfe3b4ebbd/mojo/stdlib/stdlib/sys/info.mojo#L377-L414) to make it easier to create cross-platform consts. Curious to get your thoughts if you have any feedback on structuring things this way for libc.

martinvuyk:
@jackos that looks good... yeah it is much better to add these constants to enums.

I would like it though if we could reorder the whole module instead of throwing everything into either `ffi.mojo`, `_libc.mojo`, or `_libc_errno.mojo`. So if you guys are ok with it, I can leave out all of the constants from this PR, then after you merge the ErrNo enum PR I can fix all conflicts to add the windows support and tests. I could also do the reordering of the ffi stuff into a single module in this PR or split it off into another, whatever suits you

Ideally in the future I would also like to bring the bindings in the python directory to a `ffi/python` module as well.

Issue #4685, opened 2025-05-24
[Feature Request] Fold `FPUtils` into `DType` / make them standalone functions in `dtype.mojo`


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

- `DType.sign_length[dtype]() -> Scalar[DType]`
- `DType.mantissa_length[dtype]() -> Scalar[DType]`
- `DType.exponent_length[dtype]() -> Scalar[DType]`
- `DType.exponent_bias[dtype]() -> Scalar[DType]`
- `DType.neg_zero[dtype]() -> Optional[Scalar[DType]]`
- `DType.nan[dtype]() -> List[Scalar[DType]]`
- `DType.inf[dtype]() -> Optional[Scalar[DType]]`
- `DType.neg_inf[dtype]() -> Optional[Scalar[DType]]`

### What is your motivation for this change?

We need to open the doors to true generic floating point programming.

soraros:
A large part of these are covered by the `FPUtils`. It's again, more of a API design problem.

martinvuyk:
I just found `FPUtils` by searching for NaN definitions like 3 minutes ago lol. Yeah I think this should go elsewhere. I don't know whether to close the issue or leave it as a proposal to move them to `dtype.mojo` or integrate the methods into `DType`

ConnorGray:
Hi Martin 👋 🙂. Apologies for the delay giving you feedback here. The team had a chance to discuss this proposal recently, and we think the suggestion to put this set of methods directly on `DType` makes sense. We'd be willing to accept a straightforward PR that performs this code movement 🙂

PR #5031, opened 2025-07-15
[stdlib] Move some functions from `FPUtils` to `DType`

martinvuyk:
Move some functions from `FPUtils` to `DType`. Part of #4685


ConnorGray:
!sync

Issue #4914, opened 2025-06-26
[Feature Request] Add a `@deprecated` decorator to `List` and `Dict` 's `__copyinit__()`


gabrieldemarmiesse:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

So the issue of collections being implicily deeply copyable when doing `other_list = old_list` or even `return local_list` is in a limbo at the moment. Work was done to move away from collections being implicitly copyable, but it seems this was delayed/cancelled. Probably due to compiler issues.

It's not clear if the solution is to retry removing the `Copyable` trait, waiting for true conditional conformance or making those copy-on-write, but the status quo is an issue as more and more Mojo users are writing libraries and triggering implicit deep copies all the time (even in the stdlib).

### What is your motivation for this change?

Since we don't know exactly where to go, I propose that we deprecate those methods with an helpful warning message explaining how to migrate. I understand that for some situations it might be impossible to avoid a `__copyinit__()` call, notably because we need conditional conformance, but it's better to have a warning than having a performance drop that's hard to find when reading the code.

Does the stdlib team have any opinions on the matter?

ConnorGray:
Hi Gabriel 👋 🙂. The team got a chance to discuss this yesterday, and we'd like to move forward with this proposal 👍 

You're correct that fully moving away from implicit `Copyable` bounds has been blocked on compiler issues unfortunately for a while. But using `@deprecated` seems like a nice incremental hill-climb step we could take with low risk. It won't catch every implicit copy (e.g. in generic code), but should warn users away from the most common cases.

If you or another community member following this discussion would like to prepare a PR, we'd be happy to incorporate it. I'd just ask that such a PR:

* [ ] Updates the changelog showing an example of code that would need to be updated. (i.e. `var list2 = list` => `var list2 = list.copy()`)

* [ ] The deprecation message informs the user that `list.copy()` / `dict.copy()` are the alternative.

* [ ] Let's start with only applying this to `List` and `Dict` and seeing how that goes. For the time being, I think we would not want to apply this to `String`, for example. (Cheap to copy types like `StringSlice` and `Span` should also remain (quietly!) implicitly copyable 🙂)

christoph-schlumpf:
@gabrieldemarmiesse: Maybe close this issue as it is not needed anymore.

gabrieldemarmiesse:
Thanks closing!

PR #5033, opened 2025-07-16
[stdlib] Use a warning on `Set.__copyinit__()`

gabrieldemarmiesse:
Fixes part of 
* https://github.com/modular/modular/issues/4914

I know `Set` was not in the initial issue, but `Set` uses `Dict`, which uses `List`. So we want to handle the downstream dependencies first. Furthermore, it has very little usage, so it's a good way to see the effect of the decorator


mojo/stdlib/stdlib/collections/set.mojo
```diff
@@ -78,6 +85,19 @@ struct Set[T: KeyElement, H: Hasher = default_hasher](
         for e in elements:
             self.add(e)
 
+    @doc_private
+    fn __init__(out self, owned data: Dict[T, NoneType, H]):
```
soraros:
```suggestion
    fn __init__(out self, var data: __type_of(self._data)):
```

gabrieldemarmiesse: Thanks!

mojo/stdlib/stdlib/collections/set.mojo
```diff
@@ -86,6 +106,17 @@ struct Set[T: KeyElement, H: Hasher = default_hasher](
         """
         self._data = other._data.copy()
 
+    fn copy(self) -> Self:
+        """Create an explicit copy of the set.
+
+        Prefer using this function over the copy constructor, as the copy
```
gryznar, line 115: This sentence seems to be redundant with depracation on `__copyinit__()`. Saying explicitly in its docstring that it is the default one is too verbose IMO

gabrieldemarmiesse: It's fine to be overly verbose in documentation IMO

JoeLoser (approved):

JoeLoser:
LGTM, thanks Gabriel!

JoeLoser:
!sync

ConnorGray:
Just wanted to give a brief update that this is blocked on a compiler bug. TL;DR: This simple repro demonstrates a spurious warning that is breaking our internal CI for this PR:

```mojo
trait Fooable:
    fn foo(self):
        pass


struct Foo(Fooable):
    @deprecated("foo() is deprecated.")
    fn foo(self):
        print("Foo.foo()")


fn main():
    pass
```

Issues warning:

```output
trait-deprecated-conformance.mojo:6:8: warning: foo() is deprecated.
struct Foo(Fooable):
       ^~~
trait-deprecated-conformance.mojo:8:8: note: 'foo' declared here
    fn foo(self):
       ^
```

If we remove the explicit `struct Set(Copyable)` conformance the warning goes away, but we're obviously not wanting to remove the conformance entirely, just make concrete calls to `Set.__copyinit__()` a warning.

I'll merge this PR once we can fix that internally. Thanks for your patience as we work through this 😌

gabrieldemarmiesse:
Closing as not necessary anymore thanks to the awesome work of @ConnorGray !

PR #5035, opened 2025-07-16
[stdlib] Clean up `PythonObject.__del__` a bit

soraros:
- Removed a branch since `Py_DecRef` is a no-op on `Null`.


JoeLoser (approved):

JoeLoser:
!sync

PR #5039, opened 2025-07-17
[stdlib] Clean up `CPython.destroy`

soraros:


mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -1340,13 +1336,13 @@ struct CPython(Copyable, Defaultable, Movable):
         pass
 
     @staticmethod
```
soraros: Why is this a `@staticmethod`?

laszlokindrat: Either some legacy of old conventions or an oversight. Can you please fix it and update the callsites?

laszlokindrat (approved):
LGTM! There is a typo in the PR title, can you please fix it?

laszlokindrat:
!sync

PR #5040, opened 2025-07-17
[docs] Change parameter type from `StringLiteral` to `String`

LeeLee26:
Cannot implicitly convert 'StringLiteral["woof"]' value to 'StringLiteral[value]' in doc https://docs.modular.com/mojo/manual/parameters/#optional-parameters-and-keyword-parameters


gryznar:
Title IMO should be corrected. This pr is about fixing this error, rephrasing it in title does not make too much sense ;)

soraros:
Little tip: you could use backquote `` ` `` instead of `"` to make it look a little nicer.

KenJones-Modular:
!sync

PR #5041, opened 2025-07-17
[stdlib] Clean up `_PyIter`

soraros:


mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -72,88 +72,42 @@ struct _PyIter(Defaultable, Sized):
...
-    @implicit
     fn __init__(out self, iter: PythonObject):
         """Initialize an iterator.
 
         Args:
             iter: A Python iterator instance.
         """
-        var cpython = Python().cpython()
+        var cpy = Python().cpython()
         self.iterator = iter
```
soraros, line 89: @laszlokindrat This could be a place to maybe put a [`PyIter_Check`](https://docs.python.org/3/c-api/iter.html#c.PyIter_Check) in `debug_assert`.

laszlokindrat (approved):
Thanks for the patch! I'm good with this, but can you please verify that we have some tests that cover this (and if not, add some)?

mojo/stdlib/stdlib/python/python_object.mojo
laszlokindrat: I don't feel strongly about it, but feel free to add it if you want to!

soraros: Right. I'm mainly giving you an example, since we discussed this yesterday. I think it's fine without a check.

soraros:
@laszlokindrat Yea, we have `test_iter` in `test_python_object.mojo`. I think it's good to go.

laszlokindrat:
!sync

PR #5044, opened 2025-07-18
[stdlib] Remove `Bool.__neg__`

soraros:
Mojo is a language with strong typing, there's no point mimic Python oddity here.

> I'd suggest we remove that behavior since Mojo has strong typing.  I remember me and you were talking about this last year sometime and it made one particular thing easier, but it is not a great thing for Mojo to keep from Python IMO.  We've moved on from the "superset of Python" and needing to copy some of its not-so-great things.  This is one of those things we should deviate from IMO.  Any interest in opening a PR?

_Originally posted by @JoeLoser in https://github.com/modular/modular/pull/4898#discussion_r2168080582_

### DIFF
--- a/mojo/stdlib/stdlib/builtin/bool.mojo
+++ b/mojo/stdlib/stdlib/builtin/bool.mojo
@@ -521,15 +521,6 @@ struct Bool(
         """
         return lhs ^ self
 
-    @always_inline("builtin")
-    fn __neg__(self) -> Int:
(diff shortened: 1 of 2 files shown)


laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

lattner:
Thank you for removing this!

PR #5045, opened 2025-07-18
[stdlib] Clean up docstring of `PythonObject.__init__(from_[owned|borrowed]_ptr)`

soraros:


mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -195,47 +195,36 @@ struct PythonObject(
...
         Args:
-            from_owned_ptr: The `PyObjectPtr` to take ownership of.
+            from_owned_ptr: An owned pointer to a Python object.
+
+        References:
+        - https://docs.python.org/3/glossary.html#term-strong-reference
         """
         self._obj_ptr = from_owned_ptr
 
     fn __init__(out self, *, from_borrowed_ptr: PyObjectPtr):
```
soraros, line 210: Maybe we could rename them to `from_[borrowed|owned]`. WDYT, @laszlokindrat?

mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -195,47 +195,36 @@ struct PythonObject(
         """Initialize this object from an owned reference-counted Python object
         pointer.
 
-        Ownership of the reference will be assumed by `PythonObject`.
+        This function should be used to construct a `PythonObject` from the
```
laszlokindrat: Just to make it clear that there might be other cases when you use this.
```suggestion
        For example, this function should be used to construct a `PythonObject` from the
```

mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -195,47 +195,36 @@ struct PythonObject(
...
-
-        # SAFETY:
-        #   We were passed a Python 'read-only reference', so for it to be
-        #   safe to store this reference, we must increment the reference
-        #   count to convert this to a 'strong reference'.
-        cpython.Py_IncRef(from_borrowed_ptr)
-
-        self = PythonObject(from_owned_ptr=from_borrowed_ptr)
+        var cpy = Python().cpython()
+        # increment the refcount to please RAII
```
laszlokindrat: Let's not call this RAII :) I think the original comment was fine, do you have something against it?

mojo/stdlib/stdlib/python/python_object.mojo
laszlokindrat: Sure, the type makes it clear that this is a pointers. Feel free to do the rename in this patch.

laszlokindrat (approved):
LGTM % few nits. Thanks!

mojo/stdlib/stdlib/python/python_object.mojo
soraros: Not really. Though, we use the same refcount-increasing trick elsewhere (the `PyObj` has a dtor that calls `DecRef` later), so I don't think this case is any different. Will put the comment back.

mojo/stdlib/stdlib/python/python_object.mojo
soraros: It will touch `_cpython.mojo`, I will wait until that part settle down.

laszlokindrat:
!sync

PR #5048, opened 2025-07-18
[stdlib][NFC] Move code in `_cpython.mojo` to reduce the diff size for subsequent patches

soraros:


soraros:
The diff algorithm isn't smart enough to recognise that I'm only moving code around, though it's fairly trivial to see that from the individual commits.

laszlokindrat (approved):
Thanks, this should make subsequent reviews easier!

laszlokindrat:
!sync

PR #5053, opened 2025-07-20
[stdlib][cpython-api] ABI hardening for CPython memory/reflection methods

soraros:
- Removed `PyEval_GetBuiltins` since it's not used.


soraros:
Dependes on #5048.

laszlokindrat (approved):
🎉 Thank you!

laszlokindrat:
!sync

PR #5060, opened 2025-07-20
[stdlib] Precondition Assisted Execution Proposal 

akneni:
This is a more formal proposal of the idea that was proposed in [this discussion](https://forum.modular.com/t/precondition-based-optimization-library-design-proposal/1568) on the modular forum.

Note, this is the same as #4741 (I had to create a fresh PR due to some technical issues).

### DIFF
--- a/mojo/proposals/precondition-assisted-execution.md
+++ b/mojo/proposals/precondition-assisted-execution.md
@@ -1,0 +1,188 @@
+## Proposal: Implement Precondition Assisted Execution Pattern for mojo standard library
+
+**Author:** Anish Kanthamneni (akneni)
+**Date:** June 3, 2025
+**Status:** Proposed
+
+---
+
+### Abstract
+
+This proposal suggests introducing a pattern of "unsafe assumption parameters"
+for functions in the Mojo standard library and for general use. These
+compile-time parameters would allow developers to assert specific
+preconditions, enabling functions to bypass certain checks or utilize more
+specialized, higher-performance code paths. This approach aims to provide
+significant performance gains in critical scenarios, reduce API fragmentation
+compared to numerous distinct `unsafe_` prefixed functions, and potentially
+improve code maintainability.
+
+---
+
+### 1. Motivation for this Proposal
+
+Mojo standard library functions prioritize safety and generality, which is
+essential. However, this can lead to performance overhead in situations where
+developers *know* specific preconditions about their data or environment are
+met. For instance, a string might be known to be ASCII, or a list may be
+known to have a valid length/capacity.
+
+Currently, one way to offer optimized paths is to create separate `unsafe_`
+prefixed functions (e.g., `List.unsafe_get()`). While clear, this approach
+can lead to significant API fragmentation if applied broadly, potentially
+doubling the number of relevant functions. It can also lead to boilerplate or
+near-duplicate code across the safe and unsafe variants.
+
+We need a mechanism that:
+1.  Allows opt-in to unsafe, specialized behavior for performance.
+2.  Clearly signals the associated risks and responsibilities to the developer.
+3.  Minimizes API surface bloat.
+
+---
+
+### 2. Proposed Solution: Unsafe Assumption Parameters
+
+I propose the use of parameters in function signatures to specify assumptions.
+These parameters would be prefixed with `unsafe_` to clearly indicate that
+enabling them by overriding the safe default value shifts responsibility to the
+caller for upholding the asserted precondition.
+
+**General Form:**
+
+```mojo
+fn some_function[
+    unsafe_assume_X: Bool = False, # Default to safe behavior
+](...) -> OutputType:
+    @parameter
+    if unsafe_assume_X:
+        # Path optimized assuming X is true
+        # ...
+    else:
+        # Default safe and general path
+        # ...
+```
+
+The `unsafe_` prefix serves as a contract: if the developer sets
+`unsafe_assume_X`, they guarantee that condition X holds. If the assumption is
+violated, the behavior is undefined. The default values ensure that calls
+without these explicit parameters retain current safe behavior.
+
+---
+
+### 3. Benefits
+
+- **Significant Performance Gains**: Allows bypassing checks (e.g., UTF-8
+    validation) or using hardware-specific optimizations (e.g., aligned SIMD
+    instructions) that are not possible in general-purpose code.
+- **Reduced API Fragmentation**: Instead of `foo()`, `unsafe_foo_ascii()`,
+    and `unsafe_foo_assume_capacity()`, we can have a single `foo()` with
+    `unsafe_assume_ascii` and `unsafe_assume_capacity` parameters. This keeps
+    the API surface cleaner and more focused.
+- **Improved Code Maintainability**:
+    - Consolidates related logic within a single function definition, making it
+        easier to understand the relationship between the general case and
+        optimized variants.
+    - Reduces boilerplate that might arise from creating many separate, largely
+        similar `unsafe_` functions. Common setup, teardown, or non-optimized
+        parts of the logic can be shared, with only the assumption-specific
+        sections varying based on compile-time parameters.
+
+---
+
+### 4. Concrete Examples
+- *Please note that these are merely examples, and that this pattern could be
+  implemented in many other places as well*.
+
+#### Example 1: String Processing - `unsafe_assume_ascii`
+
+Many text processing tasks, such as parsing structured data like JSON or
+specific log formats, may operate on inputs known to be ASCII.
+
+- **Proposed Parameterization**:
+
+    ```mojo
+    fn parse_json_ish_text[
+        unsafe_assume_ascii: Bool = False
+        # ... other parameters ...
+    ](text: String) -> ParsedData:
+        @parameter
+        if unsafe_assume_ascii:
+            # Optimized path for ASCII:
+            # - Skip UTF-8 validation/decoding steps.
+            # - Use direct byte comparisons/manipulations.
+            # - Potentially use SIMD for certain ASCII operations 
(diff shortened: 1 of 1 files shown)


gryznar:
@JoeLoser sorry for ping, but this proposal stays some time without answer from stdlib team. It seems to be straightforward and quite obvious improvement

ConnorGray:
Hi @akneni, thank you for taking the time to write up this proposal 🙂. I apologize for the delay in communication from our team; we've only recently had a chance to discuss your proposal.

The pattern you suggest is interesting. However, we had some questions regarding how broadly applicable this proposed transformation is. If the proposal was expanded with additional concrete examples, that may help us understand and evaluate it better.

We also were unsure if a tradeoff of this transformation might be an unintended harm to readability. The examples given in the proposal were all roughly of the form:

```mojo
fn some_operation[assumption: Bool = False](mut self, item: T):
    @parameter
    if assumption:
        # ...
    else:
        # ...
```

Implying that the two alternative bodies of the function `# ...` were effectively disjoint. In that situation, could it actually be be more readable (both to a caller and the author) to continue giving a single name to each alternative (the safe and unsafe logic), e.g. `take()` vs `unsafe_take()`? (Using dedicated functions for unsafe vs safe variants of operations also helps keep doc strings for the safe variant clean.)

One general design principle we've learned while developing the Mojo standard library is that often there are things that _could_ be modeled with either a parameter, or as separate, named functions—but that it's better to exercise restraint, and avoid complexity due to "over parameterizing". Unsafety is a possible example of that, but there are many others.

Understandably though, using parameters is tempting because Mojo makes it so easy—that is definitely a strength of Mojo 🙂 

We haven't articulated this principle formally—and its fundamentally subjective; reasonable people can disagree—but in our view, "unsafety" is potentially less a "knob" that can be tuned, and more a fundamental characteristic of an operation, and hence may justify being distinguished by separately named functions.

Regarding unsafety, another reason our instinct is that it doesn't fit cleanly as a parameter is that its rarely a "global" property, passed down through layers of parameterized types or functions (as, say, `simd_width` is). Instead, opting in to unsafe operations is often a local choice made as an optimization after benchmarking has been performed. That implies code that is parameterized on the _caller_
 side would be rare.

For example, is code like this common in Mojo today?

```mojo
@parameter
if some_condition:
    optional.unsafe_take()
else:
    optional.take()
```

If it is common, then implementing a parameter like this proposal suggests could simplify that logic:

```mojo
optional.take[unsafe_assume_present=some_condition]()
```

However, our instinct is that code calling e.g. `unsafe_take()` _conditionally_ is rare.

But, we'd love to see the results of an audit of our stdlib or other open source Mojo code out there in the world—perhaps this pattern is common and implementing this proposal will help simplify real-world code like that shown above 🙂

Regarding the concern about code duplication that is potentially avoided by adopting this proposal, I'm uncertain how common such duplication is in our code base. In practice, we try to follow a pattern like:

```mojo
fn foo(self):
     # ... Check preconditions ...

     self.unsafe_foo()

fn unsafe_foo(self):
    # ... do unsafe operations ...
```

which means that duplication is already avoided through normal reuse of code via named functions.

<details>

<summary>That being said, as I'm writing this, I realize that it's good practice to also include debug_asserts within `unsafe_foo()` like functions, so you actually often write something like this:</summary>

```mojo
fn foo(self):
     # ... Check preconditions ...

     self.unsafe_foo()

fn unsafe_foo(self):
    @parameter
    if debug_assertions_enabled:
        # ... Check preconditions ...

    # ... do unsafe operations ...
```

so the `# ... Check preconditions ...` logic is duplicated. That implies a potential simplification is:

```mojo
fn foo[unsafe: Bool](self):
     @parameter
     if unsafe or debug_assertions_enabled:
         # ... Check preconditions ...

     # ... do unsafe operations ...
```

That's interesting, but I think in practice could run into the problem that the safe function wants to `raise` (and hence needs to be marked `raises`), while the unsafe function _should not_ be marked `raises`, making it difficult to be parametric over this behavior.

</details>

Thank you again for writing this proposal @akneni.  Please take the potential issues I've described above as initial feedback, not as a rejection of your proposal🙂 It's a great thing that you're pushing us to clarify our design thinking around this part of Mojo. Mojo is still young, so "best practices" shouldn't simply be "whatever we tried first", but instead get poked and prodded and tested as you're pushing for here 🙂 That process is what will help us all be confident we've found the optimal position in the space of tradeoffs.

To be clear, we if folks want to experiment with applying this pattern in specific places in the standard library, we'll evaluate and potentially accept those on a _case by case_ basis. However, we're not quite sure yet that we want to accept this proposal as a general design principle that should be applied widely across the stdlib.

Please let me know if you have any questions for me or the team. And as always, please feel free to update the proposal and ping us again if you'd like us to give it another review 🙂

gryznar:
@ConnorGray how about sth like that?

Let's assume, `foo` has 2 safety knobs. To fully cover that in separate functions, we need to have:
- `foo()`
- `unsafe_foo_knob_a()`
- `unsafe_foo_knob_b()`
- `unsafe_foo_knob_a_b()`

And this is only for 2 unsafe cases! 
With this proposal we also have granularity over unsafeness (in usage we may want only to assume one unsafe variant, while treat others safely).

Currently we most often have 2 options: full safety or no safety at all.

akneni:
@ConnorGray
Thanks for taking the time to look over this and for the feedback! You bring up a lot of very good points and I think my proposal could stand to be refined. I'll dump some of my thoughts here and will update the PR once/if we come to a consensus.  

I do agree that with your example that `some_operation()` and `unsafe_assumpetion_some_operation()` is probably more readable that shoving an if statement in the body. In this case, the precondition assisted execution (PAE) version may be easier for the user (as all the unsafe knobs can be seen in one place as opposed to searching for multiple different functions with names `unsafe_.+_some_operation()`), but may lead to some difficulty for library authors. I'd argue that this may still be desirable for 2 reasons. Firstly, I think making library authorship harder for easier use of the libraries is typically worth it. The second reason is because using PAE parameters allow us to elegantly pass assumptions up the call stack (this is the most important reason, more on this later). 

Note that the trade off between ease of library authorship and ease of library usage only exists for simpler functions with one or two PAE parameters. This pattern is beneficial to both in more complex scenarios where we have multiple unsafe knobs that aren't mutually exclusive (as @gryznar brought up). In terms of a more concrete example, let's look at the csv parsing function below. We can parse a csv more quickly if we know there are no escaped commas (or commas embedded in quoted strings), and the same is true if we already know the types of each column (so we don't need to call `.isdigit()` before converting to an int for example). 
Providing all possible combinations of these would require 4 functions (more generally, a function with $n$ PAE parameters would need to be split into $2^n$ functions, which does not scale). Using PAE, we can handle it with a single function. 
```mojo
fn parse_csv_row[
    unsafe_assume_valid_commas: Bool = False, # Set to true if you know this row has no escaped commas. 
    unsafe_assume_types: Optional[List[...]] = None,
](
    row: String
):
    var columns: List
    if unsafe_assume_valid_commas:
        columns = row.split(',')
    else:
        # parse the row as normal watching out for escaped commas

    if unsafe_assume_types is not None:
        # ...
    else:
        # ...
```
But wait! We can make this even better if strings also implement the PAE pattern, we can also expose those knobs to the user through this function and pass them up the call stack. 
```mojo
fn parse_csv_row[
    unsafe_assume_valid_commas: Bool = False, # Set to true if you know this row has no escaped commas. 
    unsafe_assume_types: Optional[List[...]] = None,
    unsafe_assume_ascii: Bool = False,
](
    row: String
):
    var columns: List
    if unsafe_assume_valid_commas:
        columns = row.split[unsafe_assume_ascii = unsafe_assume_ascii](',')
    # ...

```

If string doesn't implement the PAE pattern, then each time we call a string method, it would look like the following code block. Keep in mind, this assumes that strings only have a single PAE parameter; if we have a more complex type with more PAE parameters, exposing the knobs of the underlying types becomes infeasible. 
```mojo
if unsafe_assume_ascii:
    row.unsafe_assume_ascii_foo()
else:
    row.foo(',')

# as opposed to 
row.foo[unsafe_assume_ascii = unsafe_assume_ascii]()
```

In cases where the underlying types don't implement the PAE pattern, these knobs would just never be exposed. **TLDR, PAE makes library usage easier, but more importantly, unlocks a world of lower level control and raises the performance ceiling.**

ConnorGray:
Hey @akneni, thanks for your prompt, thoughtful response, and apologies again for my delay in getting back to you.

I appreciate you writing out that CSV example 🙂. Your arguments make sense regarding more complicated usage scenarios where multiple parameters are involved. I can see how parameters would help control complexity, compared to a combinatorial explosion of parameterless functions.

However, we don't currently have a CSV parser in the Mojo standard library. I'd love to consider some cases from "real world" standard library code. If we had such a case, then we could compare the tradeoffs in benefits to API ergonomics against downsides like code size and other dimensions like performance of a parameter vs a runtime argument 🙂

My current sense of this proposal is that it makes sense as a design decision in some cases, but perhaps not in others, based on case-specific tradeoffs. I wonder if the main thrust of this idea might make sense as a suggestion/recommended in the Mojo style guide? In some sense, it's more of an API design "rule of thumb" than a language proposal 🙂 If we had a concrete "success story" where we applied this pattern that we could reference in such a style guide section, that would help motivate it and make it clear in the minds of readers. 🙂 

Thanks again for pushing on this, it's awesome to see folks trying to define and refine new API design principles for Mojo 🔥

akneni:
@ConnorGray My turn to apologize for the delay. Thank you for the response! I'll try to update this proposal sometime this weekend

akneni:
Ok, I've looked though the standard library, and frankly, I've only found a few places where PAE would help, and even then wouldn't result in performance improvements that significant. I'll close this for now and maybe bring it up again when the standard library gets built out a little more.

Issue #4924, opened 2025-06-27
[BUG] AHasher does not work at compile time


mzaks:
### Actual behavior
Currently when the hash function is provided with `AHash` as `HasherType` and is execute at compile time:
```
from hashlib._ahash import AHasher 

def main():
    alias h = hash[HasherType=AHasher[SIMD[DType.uint64, 4](0)]](1)
    print(h)
```

the compiler crashes:

```
open-source/max/mojo/stdlib/stdlib/builtin/_startup.mojo:95:4: error: call expansion failed
open-source/max/mojo/stdlib/stdlib/builtin/_startup.mojo:65:4: note: function instantiation failed
<unknown>:0: note: failed to compile-time evaluate function call
[...]

PR #5061, opened 2025-07-21
[stdlib] Introduce default hasher type as a workaround

mzaks:
This should make developer experience better until #4924 is fixed.

Side note:
I wanted to implement the finish function as following:
```
   fn finish(var self) -> UInt64:
        @parameter
        if is_compile_time():
            return self._fnv1a^.finish() 
        else:
            return self._ahasher^.finish()
```

But the compiler did not let me, so I had to make the Hasher Copyable, which in my opinion seems unnecessary.


JoeLoser:
@weiweichen fixed https://github.com/modular/modular/issues/4924, so do we want to close this PR perhaps?

PR #5062, opened 2025-07-21
[example] [proposal] Add parallelized version of life example

christoph-schlumpf:
Add parallelized version of life example to show off how easy it is to speed up performance with Mojo.

Example benchmark on my Apple MacBook Air (M2):

```
1000 evolutions of gridv1.Grid elapsed time:  4251.161 ms
1000 evolutions of gridv2.Grid elapsed time:  2588.249 ms
1000 evolutions of gridv3.Grid elapsed time:  773.583 ms
```


KenJones-Modular:
!sync

PR #5064, opened 2025-07-21
[stdlib] Add `ComplexSIMD[dtype] * Scalar[dtype]`

martinvuyk:
Add `ComplexSIMD[dtype] * Scalar[dtype]`. This is much more efficient than building `ComplexSIMD(scalar_value)` and doing a full complex multiplication.


abduld:
!sync

PR #5068, opened 2025-07-23
[kernels][layout] Update layout_tensor copy functions for 2D/3D threadblocks

cudawarped:
Currently several copy routines inside `layout_tensor` silently fail when passed a thread block of more than one dimension. e.g. Given an NxN `src`/`dst` layout:

 1. Setting the `thread_layout` to be a single row (`Layout.row_major(1, N)`) of the `src`/`dst` layout results in the threads from rows 2,..,N redundantly participating in the copy resulting in N times as much memory being transfered.
 2. Setting the `thread_layout` to be all the rows and cols (`Layout.row_major(N,N)`) only copies the first row because distribute uses `thread_idx.x`.

I am assuming that this function should work with 2D/3D threadblocks because there is currently no checks to prevent it?
If so should this fix be applied to all other copy function in layout tensor?


cudawarped:

abduld:
Hi @cudawarped, thanks for the contribution!

I noticed that this patch introduces additional thread computation for code that doesn't use 3D blocks, which could impact performance in some critical operations. Could you explore whether it's possible to preserve the current FLOP count for the general case while still achieving the flexibility you're aiming for in this PR?

cudawarped:
@abduld Agreed would it be acceptable to include an extra parameter such as `block_dim_count: Int = 1`?

If so shoudl I change the scope of this PR to apply this change to all copy routines inside `layout_tensor` which are silently failing for 2D/3D blocks?

cudawarped:
@abduld Updated to:
1. Include `block_dim_count` parameter
2. Apply changes to all copy routines
3. Include tests

JoeLoser:
!sync

PR #5072, opened 2025-07-23
[stdlib] Fix a refcount bug in the Python interop integration test

soraros:


mojo/integration-test/python-extension-modules/basic-raw/mojo_module.mojo
```diff
@@ -43,19 +43,24 @@ fn PyInit_mojo_module() -> PythonObject:
...
-    return PythonObject(cpython.PyObject_Length(args))._obj_ptr
+    Return value: New reference.
+    """
+    var cpy = Python().cpython()
+    return cpy.PyLong_FromSsize_t(cpy.PyObject_Length(args))
 
 
 @export
 fn mojo_count_args_with_kwargs(
     py_self: PyObjectPtr, args: PyObjectPtr, kwargs: PyObjectPtr
```
soraros, line 55: Actually, we could also just give `kwargs` an empty default.

kartva: If you mean:
- giving `kwargs` an empty default in `mojo_count_args_with_kwargs`
- then making `mojo_count_args` forward arguments to `mojo_count_args_with_kwargs`

Then yes, that's a very sensible thing to do. It seems to be a minor enough change that you can put it in this PR if you'd like.

kartva (commented):
Thanks a lot for the PR! You're right; the temporary PythonObject when constructed will have its refcount decremented at end of scope, which means that the pointer may become invalid immediately.

kartva:
!sync

PR #5073, opened 2025-07-23
[stdlib] Rename `PythonObject` ctors argument keyword from `from_[owned|borrowed]_ptr` to `from_[owned|borrowed]`

soraros:
- The type is already clear from the signature.


JoeLoser (approved):

JoeLoser:
!sync

PR #5074, opened 2025-07-23
[stdlib] Remove `__len__` from iter structs to match the `Iterator` trait

soraros:
- `__len__` was there as a step gap.
- Add conformance to `Iterator` when possible.

### DIFF
--- a/mojo/stdlib/stdlib/collections/deque.mojo
+++ b/mojo/stdlib/stdlib/collections/deque.mojo
@@ -985,7 +985,7 @@ struct _DequeIter[
     ElementType: Copyable & Movable,
     deque_lifetime: Origin[deque_mutability],
     forward: Bool = True,
-](Copyable, Movable):
+](Copyable, Iterator, Movable):
     """Iterator for Deque.
 
     Parameters:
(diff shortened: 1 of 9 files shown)


mojo/stdlib/stdlib/collections/deque.mojo
```diff
@@ -996,13 +996,22 @@ struct _DequeIter[
     """
 
     alias deque_type = Deque[ElementType]
```
JoeLoser, line 998: **Suggestion** Drive-by, but it's weird to have this be a public alias as it's only used in one place below in the `src` member declaration.  We can fix it later and separately.

mojo/stdlib/stdlib/collections/dict.mojo
```diff
@@ -69,6 +69,8 @@ struct _DictEntryIter[
         forward: The iteration direction. `False` is backwards.
     """
 
+    alias Element = DictEntry[K, V, H]
+
     var index: Int
     var seen: Int
     var src: Pointer[Dict[K, V, H], dict_origin]
```
JoeLoser, line 76:
```suggestion
    var src: Pointer[Element, dict_origin]
```
now if you'd like.  Can probably replace/use it in other places too. Non-blocking.

mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -56,26 +56,22 @@ struct _SpanIter[
...
-    fn __has_next__(self) -> Bool:
-        return self.__len__() > 0
-
-    @always_inline
-    fn __len__(self) -> Int:
+    fn __next_ref__(mut self) -> ref [origin, address_space] T:
         @parameter
         if forward:
-            return len(self.src) - self.index
+            self.index += 1
```
JoeLoser, line 70: **Quibble** You may save an instruction if you store a local to `self.index` to avoid the `-1` below.

```mojo
var current = self.index
self.index += 1
return self.src[current]
```

JoeLoser (approved):
Nice!  I left a few non-blocking suggestions, feel free to tackle in a follow-up PR. Going to land this.

JoeLoser:
!sync

soraros:
@JoeLoser Thanks for the review. I will address the comments in a followup PR.

PR #5080, opened 2025-07-23
[stdlib] Fix refcount bug in `PythonObject._dict`

soraros:


mojo/stdlib/stdlib/python/python.mojo
```diff
@@ -370,26 +370,25 @@ struct Python(Defaultable):
...
 
-            var val_obj = entry.value.to_python_object()
-            var result = cpython.PyDict_SetItem(
-                dict_obj_ptr, key_ptr, val_obj._obj_ptr
-            )
-            if result == -1:
-                raise cpython.get_error()
+            var val = entry.value.to_python_object()
+            var errno = cpy.PyDict_SetItem(dict, key, val._obj_ptr)
+            cpy.Py_DecRef(key)
```
soraros, line 387: This is the main line of concern.

JoeLoser (commented):
High-level question: do we have good testing coverage of our ref-counting in the python module in general?  CC: @laszlokindrat as well

I saw we didn't modify/add any tests in this PR to expose the bug.  How can we move forward safely with that in mind? 😄

soraros:
IMO, the lower-level CPython API is like `UnsafePointer`: it's inherently unsafe, and careful eyeballing is probably unavoidable. That said, more test coverage is almost always beneficial, let me write a test that can catch this case.

mojo/stdlib/stdlib/python/python.mojo
```diff
@@ -370,26 +370,25 @@ struct Python(Defaultable):
...
-        var cpython = Python().cpython()
-        var dict_obj_ptr = cpython.PyDict_New()
-        if not dict_obj_ptr:
-            raise Error("internal error: PyDict_New failed")
+        """Construct a Python dictionary from keyword arguments.
+
+        Return value: New reference.
+        """
+        var cpy = Python().cpython()
+        var dict = cpy.PyDict_New()
```
laszlokindrat: nit: can you please use a different name for this, just to be defensive?

mojo/stdlib/stdlib/python/python.mojo
laszlokindrat: Nice, catch! Docs are ambiguous here:
> This function does not steal a reference to val.

Fine, but what about the key? Do you know where this is documented?

soraros: I think things are by default to not stealing, some of the `SetItem` API do steal the pointer to `val` so the Python people added a line. This line here is probably one of those "double negation" note.

Anyways, they did put a `NewRef` in the implementation.

https://github.com/python/cpython/blob/ec7fad79d24e79961b86e17177a32b32bb340fe5/Objects/dictobject.c#L2639-L2650

soraros: FYI, there's also a `PyDict_SetItemString` (not part of our CPython struct yet) which is more suitable for this use case.

laszlokindrat (approved):
Nice, thanks!

laszlokindrat:
!sync

PR #5083, opened 2025-07-24
[stdlib] Migrate OS/Arch methods to `CompilationTarget`

soraros:


mojo/stdlib/stdlib/sys/info.mojo
```diff
@@ -106,6 +106,28 @@ struct CompilationTarget[value: _TargetType = _current_target()]:
             `> : i1`,
         ]
 
+    @always_inline("nodebug")
+    @staticmethod
+    fn _arch() -> StaticString:
```
JoeLoser, line 111: **Question** Nothing to do for this PR, but I idly wonder whether we ought to make these public in the future.

We could also factor out the `target_get_field` stuff into a private helper function in this struct if you think it's useful. I see we use it in a few places now with just the value and type parts changing.

mojo/stdlib/stdlib/sys/info.mojo
```diff
@@ -173,58 +193,7 @@ struct CompilationTarget[value: _TargetType = _current_target()]:
...
-                    `#kgen.param.expr<target_get_field,`,
-                    Self.value,
-                    `, "arch" : !kgen.string`,
-                    `> : !kgen.string`,
-                ],
-                `, "apple-m4" : !kgen.string`,
-                `> : i1`,
-            ]
-        )
+        return Self._has_feature["neon"]() or Self.is_apple_silicon()
```
JoeLoser, line 201: **Praise** So much cleaner 👍

JoeLoser (approved):
Awesome clean up, thanks for pushing on this!  Do you mind also updating the changelog note I started a few weeks back at https://github.com/modular/modular/blob/766643d2472446555d617c0c7183d4e6d4f120dc/mojo/docs/changelog.md#-removed?  Thanks!

mojo/stdlib/stdlib/sys/info.mojo
soraros: Yes to both, I'm planning to do both.

JoeLoser:
!sync

JoeLoser:
!sync

JoeLoser:
!sync

PR #5084, opened 2025-07-24
[stdlib] Finalise first pass of `_cpython.mojo` refactor

soraros:


kartva (commented):
Always nice to see negative lines of code!

**Suggestion** I recommend passing the `cpy: CPython` parameter by reference, making it `ref cpy: CPython` instead. I think there are plans to pass `CPython` around by reference in the future, since it's a relatively "heavy" struct to be copying.

This isn't a very important note since we can modify this in a separate PR later on, but this is a small enough change that I think just doing it here is helpful.

laszlokindrat (approved):
Very nice, thanks for all these patches and putting up with the less-than-ideal process for bringing this in! I'm +1 on @kartva's suggestion; can you please change the tests to pass by `ref` (other places can be updated separately as we go).

Update: I discussed the `ref` thing with @soraros offline, and he convinced me that based on how the `CPython` struct should be used in the common case, `read` is the right convention.

laszlokindrat:
!sync

ConnorGray (approved):

PR #5098, opened 2025-07-26
[CI] Improve formatting and update readability

Amet13:
- Standardize YAML indentation and formatting
- Update GitHub Actions versions from pinned hash to human-readable options
- Improve readability with better line breaks
- Use working-directory instead of cd commands
- Rename cla.yaml to cla.yml for consistency
- Update obsoleted link to GitHub Actions documentation
- Update shellchek 'variable quoting' errors (SC2086)


JoeLoser:
Gentle ping @patrickdoc @Ahajha @keith

.github/workflows/check_pr_title.yml
```diff
@@ -32,7 +34,7 @@ jobs:
     timeout-minutes: 5
 
     steps:
-      - uses: morrisoncole/pr-lint-action@51f3cfabaf5d46f94e54524214e45685f0401b2a
+      - uses: morrisoncole/pr-lint-action@v1.7.1
```
Ahajha: Pinning to specific commits is intentional and good practice for security, as tags can change but commits cannot.

Ahajha (changes requested):

.github/workflows/check_pr_title.yml
Amet13: Hey @Ahajha, thanks for the review.

I agree that it’s more secure to use the hash instead. I’ll revert these changes.

However, am I understanding correctly that for “official” actions, you pin major release versions ([example](https://github.com/modular/modular/blob/main/.github/workflows/build_and_test.yml#L51)), while for “external” actions, you pin the hash ([example](https://github.com/modular/modular/blob/main/.github/workflows/check_pr_title.yml#L35))?

I’m trying to find consistency across version usage. For example, despite using the “official” action in ([example](https://github.com/modular/modular/blob/main/.github/workflows/cla.yaml#L32)), it still uses the hash, and vice versa ([another example](https://github.com/modular/modular/blob/main/.github/workflows/deploy_docs.yml#L38C15-L38C36)).

Ahajha: Internally, we pin everything, we should do the same here as well.

.github/workflows/build_and_test.yml
```diff
@@ -65,6 +66,13 @@ jobs:
...
 
-          ./bazelw test --config=ci --config=public-cache "${args[@]}" --build_tag_filters=-skip-external-ci-${{ matrix.os }},-requires-network --test_tag_filters=-skip-external-ci-${{ matrix.os }},-requires-network -- //...
+          # Set up filter arguments
+          BUILD_TAG_FILTERS="-skip-external-ci-${{ matrix.os }},-requires-network"
+          TEST_TAG_FILTERS="-skip-external-ci-${{ matrix.os }},-requires-network"
+
+          ./bazelw test --config=ci --config=public-cache "${args[@]}" \
+            --build_tag_filters="$BUILD_TAG_FILTERS" \
+            --test_tag_filters="$TEST_TAG_FILTERS" \
+            -- //...
```
Ahajha, line 76: @keith should we just have these variables be the same? or is there a world where they might differ at some point?

.github/workflows/deploy_docs.yml
```diff
@@ -35,30 +35,26 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v3
-        with:
-          version: "latest"
+        uses: astral-sh/setup-uv@caf0cab7a618c569241d31dcd442f54681755d39
```
Amet13, line 38: Example: https://github.com/modular/modular/actions/runs/16379402218/job/46287225766#step:1:33

.github/workflows/build_and_test.yml
keith: probably the same, i don't think this part of the change is necessary but if you want to do it yea i'd use 1 var

PR #5075, opened 2025-07-23
[stdlib] add gpu support for nvidia gtx 1050

cyrillzadra:
I'm working on getting support for my GeForce GTX 1050. Here's the current state:

**Bazel test summary:**
Executed 707 tests:

571 passed
46 failed to build
19 failed locally
71 were skipped

Some tests initially failed due to timeouts but passed when rerun individually.
However, I'm still unsure what exactly is required to get the remaining failed tests working. I'm currently stuck and would appreciate any guidance or suggestions.

Is full support even possible with this older NVIDIA card, or are some features simply not compatible?

More details can be found below.

### nvidia-smi
```
cyrill@NB-TIEZAD-DELL:~/mojo-playground$ nvidia-smi
Wed Jul 23 09:17:45 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.163.01             Driver Version: 536.23         CUDA Version: 12.2     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce GTX 1050 ...    On  |   00000000:01:00.0 Off |                  N/A |
| N/A   66C    P0             N/A / ERR!  |     487MiB /   4096MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
```

### gpu-info.mojo
```
cyrill@NB-TIEZAD-DELL:~/modular/mojo$ pixi run mojo run stdlib/tools/gpu_info.mojo
Info(
name="NVIDIA GeForce GTX 1050 Ti with Max-Q Design",
vendor=Vendor.NVIDIA_GPU,
api="cuda",
arch_name="pascal",
compile_options="nvptx-short-ptr=true",
compute=6.1,
version="sm_61",
sm_count=6,
warp_size=32,
threads_per_sm=-1,
threads_per_warp=32,
warps_per_multiprocessor=64,
threads_per_multiprocessor=2048,
thread_blocks_per_multiprocessor=32,
shared_memory_per_multiprocessor=98304,
register_file_size=65536,
register_allocation_unit_size=256,
allocation_granularity="warp",
max_registers_per_thread=255,
max_registers_per_block=65536,
[...]

PR #5107, opened 2025-07-28
[max] nvml changes to initilization and driverversion.patch level handling (wsl2)

cyrillzadra:
After spending some more time investigating the test failures in https://github.com/modular/modular/pull/5075, I was able to identify the root cause: wrong libnvidia-ml.so library files were being used.

Initially, the test was failing with:

`Unhandled exception caught during execution: NVML_UNINITIALIZED`

After improving the NVML initialization error handling, the actual underlying issue became clear:

`ABORT: CUDA NVML library initialization failed: NVML_DRIVER_NOT_LOADED
`
This indicated that the NVML library couldn't be initialized because the driver wasn't properly loaded.

Once the correct library was in place and properly initialized, the test was able to proceed and retrieve the driver version as expected. However, I noticed that on WSL2 the driver version reported by `nvidia-smi` does not include the patch-level version. 

For reference, here's the output of nvidia-smi on my WSL2 setup:

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.163.01             Driver Version: 555.85         CUDA Version: 12.5     |
|-----------------------------------------+------------------------+----------------------+
```

### DIFF
--- a/max/kernels/src/nvml/nvml.mojo
+++ b/max/kernels/src/nvml/nvml.mojo
@@ -53,10 +53,12 @@ alias CUDA_NVML_LIBRARY = _Global[
 fn _init_dylib() -> _OwnedDLHandle:
     try:
         var dylib = _try_find_dylib(_get_nvml_library_paths())
-        _ = dylib._handle.get_function[fn () -> Result]("nvmlInit_v2")()
+        _check_error(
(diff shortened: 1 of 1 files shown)


JoeLoser (approved):
Nice find, thanks for fixing this!

JoeLoser:
!sync

PR #5119, opened 2025-07-30
[stdlib] Add convenience pixi testing and format config

martinvuyk:
Add convenience pixi testing and format config

### DIFF
--- a/mojo/stdlib/scripts/run-tests.sh
+++ b/mojo/stdlib/scripts/run-tests.sh
@@ -17,4 +17,20 @@ set -euo pipefail
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 REPO_ROOT="${SCRIPT_DIR}"/../../..
 
-exec "$REPO_ROOT"/bazelw test //mojo/stdlib/test/...
+
+FILTER=""
+if [[ $# -gt 0 ]]; then
+  FILTER="${1#./}" # remove leading relative file path if it has one
+  if [[ -f ${FILTER} ]]; then # specific file
+    FILTER="//mojo/$(dirname $FILTER):$(basename $FILTER)"
+    FILTER=$("$REPO_ROOT"/bazelw query $FILTER)
+  else
+    # specific test directory
+    FILTER="${FILTER%/}" # remove trailing / if it has one
+    FILTER="//mojo/${FILTER}/..."
+  fi
+else
+  FILTER="//mojo/stdlib/test/..."
+fi
+
+exec "$REPO_ROOT"/bazelw test ${FILTER}
--- a/mojo/pixi.toml
+++ b/mojo/pixi.toml
@@ -6,6 +6,7 @@ platforms = ["linux-64", "linux-aarch64", "osx-arm64"]
 
 [tasks]
 tests = "./stdlib/scripts/run-tests.sh"
+format = "mojo format ./stdlib"
 
 [dependencies]
 python = ">=3.9,<3.14"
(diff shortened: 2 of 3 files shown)


mojo/stdlib/scripts/run-tests.sh
```diff
@@ -17,4 +17,15 @@ set -euo pipefail
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 REPO_ROOT="${SCRIPT_DIR}"/../../..
 
-exec "$REPO_ROOT"/bazelw test //mojo/stdlib/test/...
+
+FILTER="stdlib/test"
```
JoeLoser: **Question** Do we want to maintain this script even at all now that we have https://github.com/modular/modular/commit/e591f3ef610d14ac98dc86f685b2db8818181171? People just can use `bazelw` directly and specify their subdirectory of interest.

Ahajha: Yea this is mainly meant as a courtesy option for transitioning to the bazel build. Personally I'd rather people just use bazel directly.

martinvuyk: I prefer using the pixi task. Faster to type as well.

ehsanmok: +1 as pixi task. A lot less users have used bazel and don't want to necessarily learn yet another thing. Just bc something is used internally doesn't have to leak into the user-space so it's great to have it hidden as much as possible.

keith: no pref on this script or not since it's tiny, but @ehsanmok note that most things in the repo are bazel only, this script only affects the stdlib

ehsanmok: Sure, would be great to IMO to wrap everything else as pixi tasks for outside users.

keith: I think there's probably a bit too many potential workflows for that to be very understandable

Ahajha: +1, the second someone says "I want to run a subset of the stdlib tests" we have to come back and update this workflow to pipe things through. I think a quick start guide to Bazel, and maybe some convenience aliases like we have internally (like bb and br) would help. All this is doing currently is just adding another layer of indirection.

ehsanmok: I understand there're nuances. My point is to lower the barrier to entry pixi tasks are perfect. Sure, more advanced stuff leaks bazel requirements.

martinvuyk: >  "I want to run a subset of the stdlib tests" 

That is the motivation for this. Now somebody can run a subdirectory. If we want to get down to the layer of running single files again, then yeah it could be too much of a hassle for you guys (although I would very much like having it).

JoeLoser: Each test file should have its own test target you can use `bazel query` to find and run directly FYI.

keith (approved):
i think we should land and keep this script afloat while it's easy like this

martinvuyk:
> Each test file should have its own test target you can use bazel query to find and run directly FYI.

And if I find a way to plug this into the script i.e. if the arg is a file execute that command, would you guys be okay with integrating it in the future to this script?

JoeLoser:
!sync

martinvuyk:
> Each test file should have its own test target you can use `bazel query` to find and run directly FYI.

Thanks @JoeLoser, just added the ability to test specific files as well :D

JoeLoser:
> > Each test file should have its own test target you can use `bazel query` to find and run directly FYI.
> 
> Thanks @JoeLoser, just added the ability to test specific files as well :D

Can you open a new PR for that functionality?  The subdir-only part landed already internally.  Thanks!

PR #5122, opened 2025-07-30
[stdlib] Add origins to `test_utils` to make them a bit safer

martinvuyk:
Add origins to `test_utils` to make them a bit safer. I came across some segfaults in the `InlineArray` tests in #4793 and #4807 due to how they currently don't carry an origin.

### DIFF
--- a/mojo/stdlib/test/collections/test_list.mojo
+++ b/mojo/stdlib/test/collections/test_list.mojo
@@ -866,10 +866,11 @@ def test_indexing():
 def test_list_dtor():
     var dtor_count = 0
 
-    var l = List[DelCounter]()
+    var ptr = UnsafePointer[mut=False](to=dtor_count)
+    var l = List[DelCounter[ptr.origin]]()
     assert_equal(dtor_count, 0)
 
-    l.append(DelCounter(UnsafePointer(to=dtor_count)))
+    l.append(DelCounter(ptr))
     assert_equal(dtor_count, 0)
 
     l^.__del__()
@@ -880,8 +881,9 @@ def test_list_dtor():
 def test_destructor_trivial_elements():
     var dtor_count = 0
 
-    var l = List[DelCounter, hint_trivial_type=True]()
-    l.append(DelCounter(UnsafePointer(to=dtor_count)))
+    var ptr = UnsafePointer[mut=False](to=dtor_count)
+    var l = List[DelCounter[ptr.origin], hint_trivial_type=True]()
+    l.append(DelCounter(ptr))
 
     l^.__del__()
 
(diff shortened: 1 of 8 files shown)


mojo/stdlib/test/collections/test_list.mojo
```diff
@@ -866,10 +866,11 @@ def test_indexing():
 def test_list_dtor():
     var dtor_count = 0
 
-    var l = List[DelCounter]()
+    var ptr = UnsafePointer(to=dtor_count).origin_cast[mut=False]()
```
soraros: I think this should be fine:


```suggestion
    var ptr = UnsafePointer[mut=False](to=dtor_count)
```

soraros: Actually, why do we take `UnsafePtr[mut=False]`? I don't think they are that much safer, and we end up having to `origin_cast`, which is kinda messy.

martinvuyk: >     var ptr = UnsafePointer[mut=False](to=dtor_count)

that is nice



> Actually, why do we take `UnsafePtr[mut=False]`? I don't think they are that much safer, and we end up having to `origin_cast`, which is kinda messy.

Yeah I don't love it but otherwise we get exclusivity issues. It is only a bit safer in the sense that the underlying referenced counter can't be destructed before the util. I got that error when changing how errors are built inside `assert_` testing things.

martinvuyk:
```mojo
def test_eq():
    var local = 1
    var p1 = UnsafePointer[mut=False](to=local)
    var p2 = p1
    assert_equal(p1, p2)

    var other_local = 2
    var p3 = UnsafePointer[mut=False](to=other_local)
    assert_not_equal(p1, p3)

    var p4 = UnsafePointer[mut=False](to=local)
    assert_equal(p1, p4)
```
fails with
```text
mojo/stdlib/test/memory/test_unsafepointer.mojo:153:17: note: call expansion failed
    assert_equal(p1, p4)
                ^
mojo/stdlib/stdlib/testing/testing.mojo:103:4: note: function instantiation failed
mojo/stdlib/stdlib/testing/testing.mojo:127:12: note: could not simplify operator get_alignof(?, #kgen.target<triple = "arm64-apple-darwin24.5.0", arch = "generic", features = "+ete,+fp-armv8,+neon", data_layout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-n32:64-S128-Fn32", relocation_model = "pic", simd_bit_width = 128, index_bit_width = 64, accelerator_arch = "NONE">)
```
so I'll revert to using origin_cast

soraros: Sad. Could you maybe file an issue for that? I think it should work.

martinvuyk: hold on, this works fine:
```mojo
fn main():
    var data: Int = 1
    var ptr = UnsafePointer[mut=False](to=data)
    print(Int(ptr))
```
I think it might be a lowering issue

ConnorGray:
Thanks for this Martin! 🙂

!sync

ConnorGray:
!sync

PR #5128, opened 2025-07-31
[stdlib] Remove FileCheck in test_elementwise.mojo

mmicu:


mojo/stdlib/test/algorithm/test_elementwise.mojo
```diff
@@ -19,52 +19,71 @@ from buffer import NDBuffer
...
+
+        elementwise[func, simd_width=1, use_blocking_impl=is_blocking](
+            rebind[IndexList[outer_rank]](out_buffer.get_shape()),
+        )
+
+        ok = True
+        for i2 in range(min(numelems, 64)):
+            if out_buffer.data.offset(i2).load() != 2 * (i2 + 1):
+                ok = False
+                break
```
laszlokindrat: Instead of returning a boolean and asserting it below, can you just use `assert_equal` here directly?

mojo/stdlib/test/algorithm/test_elementwise.mojo
```diff
@@ -78,94 +97,31 @@ def test_elementwise_implicit_runtime():
 
     elementwise[func, simd_width=1](20)
 
+    ok = True
     for i in range(len(vector)):
         if Int(vector[i]) != 42:
-            print("ERROR: Expecting the result to be 42")
-            return
+            ok = False
+            break
```
laszlokindrat: Same as above; can you use an assert directly?

laszlokindrat (approved):
Thanks for the patch! I have a few requests to make this neater, but LGTM otherwise.

mojo/stdlib/test/algorithm/test_elementwise.mojo
mmicu: Here I am not returning a boolean as before. I just assert that `ok` is `True` after the loop.

laszlokindrat: I meant that this entire loop should look like this:
```mojo
    for i in range(len(vector)):
        assert_equal(Int(vector[i]), 42)
```
And similarly for [the loop above](https://github.com/modular/modular/pull/5128/files#r2252271809).

mojo/stdlib/test/algorithm/test_elementwise.mojo
laszlokindrat: To simplify:
```mojo
    for i2 in range(min(numelems, 64)):
            assert_equal(out_buffer.data.offset(i2).load(), 2 * (i2 + 1))
```

mojo/stdlib/test/algorithm/test_elementwise.mojo
mmicu: Yeah, you are right. I was too focused on maintaining the previous structure and I missed this. Thanks for the suggestions!

JoeLoser:
@mmicu do you mind fixing up the conflicts, please?  Then @laszlokindrat can sync this in and land it. Thanks for the contribution!

mmicu:
> @mmicu do you mind fixing up the conflicts, please? Then @laszlokindrat can sync this in and land it. Thanks for the contribution!

@JoeLoser, done. There was also a double call to `test_elementwise[32, 4, False](DimList(4, 2, 2, 2))`, but one was supposed to be done with `is_blocking=True`. I fixed that too.

laszlokindrat:
!sync

PR #5132, opened 2025-08-01
[stdlib] Cleanup List.__getitem__()

christoph-schlumpf:
Refator implementation of `List.__getitem__()` to match `Span.__getitem__()` and other collection types.

Reuse of `from collections._index_normalization import normalize_index`

Relates to: https://github.com/modular/modular/issues/2948


gryznar:
Maybe this was done previously in such way for performance? What's the impact (benchmarks)?

kartva:
!sync

kartva (approved):
Thanks for cleaning this up! Nice to see `normalize_index` spread more here, and it's good to see the `assert_always=False` (For context: I introduced, then fixed, a perf-regression in `Span.__getitem__` because of forgetting to apply that to the debug statements inside `normalize_index`)

christoph-schlumpf:
Do you know if there is a proposal available (or in the making) on the topic of „bounds checking and negativ indices“? I saw the topic in the Q3 roadmap and I think this will be very useful for stdlib contributors.

kartva:
@christoph-schlumpf
I personally don't know whether that is the case; I recommend asking on the forum or a different communication channel.

PR #5141, opened 2025-08-02
[stdlib] Add `ComplexSIMD.conj()` and make `ComplexSIMD` `EqualityComparable`

martinvuyk:
Add `ComplexSIMD.conj()` and make `ComplexSIMD` `EqualityComparable`

### DIFF
--- a/mojo/stdlib/stdlib/complex/complex.mojo
+++ b/mojo/stdlib/stdlib/complex/complex.mojo
@@ -33,7 +33,9 @@ alias ComplexFloat64 = ComplexSIMD[DType.float64, 1]
 
 
 @register_passable("trivial")
-struct ComplexSIMD[dtype: DType, size: Int](Stringable, Writable, _Expable):
+struct ComplexSIMD[dtype: DType, size: Int](
+    EqualityComparable, Stringable, Writable, _Expable
+):
     """Represents a complex SIMD value.
 
     The class provides basic methods for manipulating complex values.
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/complex/complex.mojo
```diff
@@ -253,6 +253,15 @@ struct ComplexSIMD[dtype: DType, size: Int](Stringable, Writable, _Expable):
         """
         return ComplexSIMD(-self.re, -self.im)
 
+    @always_inline
+    fn __invert__(self) -> Self:
+        """Return the conjugate of self.
+
+        Returns:
+            The conjugate of self.
```
laszlokindrat: nit: for improved clarity
```suggestion
        """Return the complex conjugate of self.

        Returns:
            The complex conjugate of self.
```

laszlokindrat:
Thanks for the patch! This is an interesting idea, but I'm not sure we really want this; it breaks symmetry with SIMD, even though I agree that conjugation is pretty common when working with complex numbers. So I would recommend that you introduce just an ordinary method for now. Once you do that, could you please also add some tests?

martinvuyk:
I wanted to tell you that conjugation is the multiplicative inverse of a complex number [by definition](https://en.wikipedia.org/wiki/Multiplicative_inverse#Complex_numbers). But I forgot about the normalization that has to happen for that to be true (it is true for normalized complex vectors).

But overall the symbol `~` is used with complex numbers to mean conjugate all the time. [GNU extension for C](https://gcc.gnu.org/onlinedocs/gcc/Complex.html) also uses the `~` operator for conjugate. The `~` operator is also called the complement operator not just the invert, and it makes more sense to say that `~Complex` is the complement of a complex number.

laszlokindrat:
In `SIMD`, the `__invert__` operator corresponds to bitwise inversion (currently restricted to boolean elements), which means that if you wrote generic code like this:
```
trait Invertible:
    fn __mul__(self, other: Self) -> Self: ...

    fn __invert__(self, other: Self) -> Self: ...
    
fn foo[T: Invertible](x: T) -> T:
    return x * ~x
```
it would work very differently for complex and "real" SIMD. Since real numbers are a subset of complex numbers, I would expect generic code that performs calculation with complex numbers to also work with real numbers. With your proposed definition of `ComplexSIMD.__invert__` this would not be the case, hence my reluctance to accept it. Also note that if we wanted to force this consistency, `SIMD.__invert__` would become identity.

And a few more issues:
- Your proposed change would make `ComplexSIMD[DType.bool].__invert__` ambiguous; does it perform boolean inversion or not?
- Neither Python nor numpy allows `~` for complex conjugation. Both of these raise `TypeError`:
```python
~complex(-1.0, 1.0)
~np.array([1+2j])
```

So my request remains the same: can you please change this to be an explicit `conj` method (or something similar), and also add tests? If and when this becomes overwhelming syntactic burden, we can decide about whether the sugar is worth it.

martinvuyk:
@laszlokindrat makes sense, didn't think much about generic code

laszlokindrat (approved):
Thanks! The complex module (and its tests) could really use some love, but we currently don't have cycles for this, so I appreciate you looking into this!

laszlokindrat:
@martinvuyk CI is currently failing, I suspect because `ComplexSIMD` doesn't implement `EqualityComparable`, or something like that. Can you please look into this (maybe in another patch)?

martinvuyk:
@laszlokindrat I just added the functions for eq and ne since this PR is so small anyway :man_shrugging:

laszlokindrat:
!sync

PR #5144, opened 2025-08-03
Fix get-started.mdx

BPerlakiH:
the run display function should be fixed in the tutorial

### DIFF
--- a/mojo/docs/manual/get-started.mdx
+++ b/mojo/docs/manual/get-started.mdx
@@ -1109,7 +1109,7 @@ Then we'll update `main()` to create a random initial `Grid` and pass it to
 ```mojo title="life.mojo"
 from gridv1 import Grid
 
-def run_display(var grid: Grid) -> None:
+def run_display(mut grid: Grid) -> None:
     while True:
         print(String(grid))
         print()


christoph-schlumpf:
Hi @BPerlakiH ,

This change would make the „get started“ documentation different from the implementation in https://github.com/modular/modular/blob/main/examples/mojo/life/lifev1.mojo (and lifev2, lifev3).

So these implementations should be changed accordingly in this PR.

I‘m not sure if `mut` is the better argument convention than `var` in this case. It does not change the performance of the implementation. By using `var` the `start` grid gets moved into `run_ display` which is correct IMO.

BPerlakiH:
Thank you for the response. I've just started to learn mojo, so I wasn't sure about it, just followed the tutorial step by step.

It does indeed compiles and runs fine from the command-line, with "var".
Originally I thought it was a typo, since VSCode couldn't handle it:

Screenshot 2025-08-04 at 10 41 39

christoph-schlumpf:
Hi @BPerlakiH ,

Yes, this argument convention recently changed from `owned` to `var`. `owned` is deprecated now.

Please ensure to use the nightly Mojo VSCode plugin.

Please close this PR.

scottamain:
Thanks for fielding this @christoph-schlumpf. I agree the docs look correct and you’re probably using an older version, @BPerlakiH. So I’m closing this.

Sadly, we currently don’t publish docs for older versions, only the latest "stable" version (which is 25.5 as of today) and the nightly version. So if you plan on sticking to the stable version, I suggest you select the "v25.5" docs from the drop-down at the top of [the docs site](https://docs.modular.com/stable/mojo/manual/get-started).

As of today, there’s no difference between the two but they will start to diverge again until we release 25.6.

PR #5152, opened 2025-08-06
[stdlib] Add windows support for errno and improve tests

martinvuyk:
Add windows support for errno and improve tests.

CC: @jackos


JoeLoser:
Thanks for the contribution, @martinvuyk! We appreciate the effort you've put into adding Windows support for errno and improving the tests.

However, I need to share some important context about our current platform support strategy: **Mojo and the Mojo standard library have no immediate plans to support Windows at this time.**

A few key points about our approach:

1. **No Windows CI infrastructure**: We don't currently have any internal Windows CI systems, and we won't be investing in Windows CI until we make a formal commitment to supporting the platform.
2. **Maintenance overhead concerns**: Given that Windows is not a platform we officially support, the Mojo Standard Library team prefers to avoid taking on additional code maintenance burden for Windows-specific implementations.
3. **Avoiding confusion:** To prevent any misunderstanding about our current platform support, we're actually planning to go a step further and remove existing Windows references from our stdlib codebase. This will help avoid confusion about what platforms we currently support.

Rest assured that any Windows-related code will remain in the Git history, so when the time comes to revisit Windows support in the future, we'll be able to reference and potentially revive those contributions.

We understand this may be disappointing, but we want to be transparent about our current priorities and resource allocation. Thanks again for your interest in contributing to Mojo!

Given this, I'm going to go ahead and close this PR.

PR #5156, opened 2025-08-07
[stdlib] Add tests for errno in linux and macos

martinvuyk:
Add tests for errno in linux and macos


mojo/stdlib/stdlib/sys/_libc_errno.mojo
```diff
@@ -443,22 +450,48 @@ struct ErrNo(Copyable, EqualityComparable, Movable, Stringable, Writable):
...
+        Args:
+            other: The `ErrNo` value to compare with.
+
+        Returns:
+            True if the error codes are not equal, False otherwise.
+        """
+        return self.value != other.value
+
+    @always_inline
+    fn __is__(self, other: Self) -> Bool:
```
JoeLoser: **Question** Do we want to define `__is__` and `__isnot__` for `ErrNo`?  I'd suggest splitting that out into a separate patch to consider as I don't think any of the existing tests exercise this logic.  Conceptually, I think of `ErrNo` wrapping a value, and so it doesn't make sense to define` is` and `is_not` for them: people should continue to use `==` and `!=` for value-based comparisons.

mojo/stdlib/stdlib/sys/_libc_errno.mojo
```diff
@@ -443,22 +450,48 @@ struct ErrNo(Copyable, EqualityComparable, Movable, Stringable, Writable):
         """
         return String.write(self)
 
+    @always_inline
```
JoeLoser, line 452: **Question** Any particular reason to mark this and `__ne__` with `@always_inline` now? Ditto for `__str__` above and friends.

JoeLoser (commented):
Overall, PR looks pretty good - thanks Martin!  Do you mind rebasing to undo the changes to `mojo/pixi.lock` now that the issue you ran into should be fixed?  I like the test coverage you added! 👏

mojo/stdlib/stdlib/sys/_libc_errno.mojo
martinvuyk: Since they are just one-liners I didn't see the point of making a function call instead of inlining them. I can revert if you wish :man_shrugging:

JoeLoser: Feel free to keep the `__eq__` as `@always_inline`, but please remove it from `__str__`.  From a quick grep in the stdlib, we only have 8 uses of `@always_inline` for `str` functions and `139` or so are explicitly marked `@no_inline` (which is what I would recommend here).

martinvuyk: Done. I thought that `@no_inline` were for places where variadic unrolling was too expensive at compile time, like the `String.write()` implementation not where that implementation is called

mojo/stdlib/stdlib/sys/_libc_errno.mojo
```diff
@@ -99,317 +99,317 @@ struct ErrNo(Copyable, EqualityComparable, Movable, Stringable, Writable):
...
+    alias ENFILE         = Self(23)
     """File table overflow"""
-    alias EMFILE         : Self = 24
+    alias EMFILE         = Self(24)
     """Too many open files"""
-    alias ENOTTY         : Self = 25
+    alias ENOTTY         = Self(25)
     """Not a typewriter"""
-    alias ETXTBSY        : Self = 26
+    alias ETXTBSY        = Self(pm["ETXTBSY",          linux=26, macos=26]())
```
JoeLoser: **Question** Any particular reason to specify `linux=26, macos=26` since they have the same values? I suggest leaving it off in favor of `Self(26)` for consistency.

martinvuyk: It is a leftover from the windows setup, the number for it is 139 and not 26. Do you still want me to simplify it?

JoeLoser: Ah, I see.  Yes, let's go ahead and simplify it now please.  I'm happy to land this after that is done 😄 

Thank you!

martinvuyk: done

JoeLoser (approved):

JoeLoser:
!sync

jackos:
thanks @martinvuyk awesome

PR #5161, opened 2025-08-10
[mojo][example] Cleanup `__del__` in life example

christoph-schlumpf:
The current implementation of `__del__` in `gridv2.mojo` and `gridv3.mojo` is misleading.
`Int8` is a "trivial" `SIMD` struct and therefore no pointees are needed to be destroyed.


KenJones-Modular:
!sync

Issue #5148, opened 2025-08-06
[BUG] Mojo InlineArray move constructor introduced in v25.5 causes horrible compilation times when array size is large


1ahmadbassam:
### Actual behavior
Compiling programs with large InlineArrays results in horrible compilation times in the latest version of Modular/Mojo.

### Expected behavior
The program should compile relatively quickly as with previous releases (25.4 and earlier).

### Steps to reproduce

```mojo
def main():
    var a = InlineArray[UInt8, 512000](0)
```
This code does not compile in a timely manner (left it for over 13 minutes on my machine).
In addition, the following warning is generated:
```
open-source/max/mojo/stdlib/stdlib/collections/inline_array.mojo:286:9: warning: parameter for unrolling loop more than 65536 times may cause long compilation time and large code size. (use '--loop-unrolling-warn-threshold' to increase the threshold or set to `0` to disable this warning)
```
This is a clear indication where the problem is. I have tried removing the `@parameter` declaration on that particular line in the move constructor and recompiling the standard library, which restores the efficient compilation times. However, as this is more of a workaround, the optimal solution would be to batch the move constructor as the base constructor is batched as well.

### System information

```
System
------------
       Pixi version: 0.48.2
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.12.38=0
                   : __glibc=2.41=0
                   : __cuda=12.4=0
                   : __archspec=1=zen3
          Cache dir: /home/aelbizri/.cache/rattler/cache
       Auth storage: /home/aelbizri/.rattler/credentials.json
   Config locations: No config files found

Global
------------
            Bin dir: /home/aelbizri/.pixi/bin
    Environment dir: /home/aelbizri/.pixi/envs
       Manifest dir: /home/aelbizri/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: mojo-serial
            Version: 0.1.0
      Manifest file: /home/aelbizri/Documents/pixeltrack-standalone/src/mojo-serial/pixi.toml
       Last updated: 06-08-2025 10:17:44

Environments
------------
        Environment: default
           Features: default
           Channels: https://conda.modular.com/max-nightly, conda-forge
   Dependency count: 1
       Dependencies: mojo
   Target platforms: linux-64
    Prefix location: /home/aelbizri/Documents/pixeltrack-standalone/src/mojo-serial/.pixi/envs/default
```
```
Package        Version  Build    Size      Kind   Source
[...]

PR #5165, opened 2025-08-12
[stdlib] Batch the unrolling of InlineArray's move constructor

ehsan:
This is a WIP fix for #5148.

As far as I can tell this is using the same approach that the copy constructor uses in order to batch the unrolling of the move constructor, but for some reason with this PR the stdlib takes an excessive amount of time to build, which makes me think that I have somehow introduced another variant of the original issue -- except that I can't figure out why for the life of me :/
image

I'd appreciate any help figuring out what could be causing this, thanks!

### DIFF
--- a/mojo/stdlib/stdlib/collections/inline_array.mojo
+++ b/mojo/stdlib/stdlib/collections/inline_array.mojo
@@ -257,11 +257,9 @@ struct InlineArray[
         self = Self(storage=elems^)
 
     @always_inline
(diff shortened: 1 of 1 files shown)


ehsan:

kartva:
I can replicate the compiler hang from this change. Have you tried to fiddle around and see what particular line may be causing the issue / tried to create a minimal reproducible example?

This is possibly a compiler issue, so a minimal reproducible example would be invaluable in properly reporting this.

PR #5173, opened 2025-08-13
[max] nn: respect strict flag in Module.load_state_dict (fix PP staged loads)

dl-alexandre:
## Summary
- Fix `Module.load_state_dict` to respect `strict` for missing/unexpected keys.
- Previously raised on missing weights even when `strict=False`, breaking Pipeline Parallel staged loading.
- Now only raises when `strict=True`; otherwise tolerates missing/unexpected keys.

## Rationale
- Aligns with common frameworks (e.g., PyTorch) and unblocks PP.
- Verified with Llama3 PP where `rope_freqs.weight` or `lm_head.weight` may be absent per stage.

## Details
- Collect `missing_keys`; only raise if `strict=True`.
- Keep unexpected keys validation gated by `strict=True`.
- Updated docstring.

## Repro enabled
MAX_SERVE_OFFLINE_INFERENCE=1 \
CUDA_VISIBLE_DEVICES=0,1 \
max generate \
  --model-path meta-llama/Llama-3.2-1B-Instruct \
  --devices=gpu:0,1 \
  --pipeline-parallel-degree 2 \
  --max-length 8192 \
  --device-memory-utilization 0.6 \
  --max-batch-size 1 \
  --max-new-tokens 10 \
  --prompt "hi"

## Notes
- No API break for strict=True; strict=False becomes lenient as intended.
- Follow-up: add unit test exercising both strict modes.

### DIFF
--- a/max/nn/layer/layer.py
+++ b/max/nn/layer/layer.py
@@ -305,14 +305,16 @@ class Module(Layer, ABC):
         """
         loaded_keys = set()
         missing_keys = set()
+        
         for name, layer in recursive_named_layers(self):
             weight_prefix = f"{name}." if name else ""
             for weight_name, weight in layer.layer_weights.items():
-                # Skip the shared weights, since their values are loaded with
-                # the original layers.
+                # Skip shared weights, as they are loaded with the original layers.
                 if weight_name in layer._shared_weights:
                     continue
+                
                 full_weight_name = f"{weight_prefix}{weight_name}"
+                
                 if (data := state_dict.get(full_weight_name)) is not None:
                     loaded_keys.add(full_weight_name)
                     if isinstance(data, WeightData):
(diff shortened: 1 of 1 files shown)


dl-alexandre:

max/nn/layer/layer.py
```diff
@@ -330,23 +334,34 @@ def load_state_dict(
...
-                        msg += f" Did you mean '{possible_match[0]}'?"
-                    raise ValueError(msg)
+                    # Missing key: respect `strict` flag.
+                    if strict:
+                        msg = f"Could not find weight '{full_weight_name}'. "
+                        if possible_match := difflib.get_close_matches(
+                            full_weight_name, state_dict.keys(), n=1
+                        ):
+                            msg += f" Did you mean '{possible_match[0]}'?"
+                        raise ValueError(msg)
```
k-w-w: Since an error is raised below, do we need to raise an error here?

k-w-w (commented):
Thanks for the PR! Agreed, missing weights should be caught only if strict=True.

tboerstad:
Thanks for this, makes a lot of sense. 

Not for this PR, but we should probably return the `missing_keys` and `unused_keys`. 
This way the caller can use `strict=False` and still handle/fix things on their end if needed

k-w-w (approved):
LGTM, thank you!

k-w-w:
!sync

Issue #2776, opened 2024-05-21
[Feature Request] Improve ergonomics for writing functions generic across integral scalars.


bgreni:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Come up with a way to allow users to write a single function that accepts `Int`, `IntLiteral` and scalar `SIMD` types. I assume there is sound rational for `Int` not simply being defined as `alias Int = Scalar[DType.index]`, so maybe this would require the introduction of another trait?

### What is your motivation for this change?

I've noticed a bit of pain point when attempting to write functions that are meant to accept any possible integral scalar type (`Int` and integral `SIMD` scalars). You can easily write a function to accept `SIMD` scalars like such.
```mojo
fn foo[type: DType](s: Scalar[type]):
    constrained[type.is_integral(), "Expected an integral"]()
    print(s)
```
However, even though you can easily convert an `Int` to a `Scalar[DType.index]`, trying to call this function with an `Int` will fail since it can't implicitly parameterize `type`. So in order to maintain a clean interface one must write a second overload that accepts `Int`.
```mojo
fn foo(i: Int):
    foo[DType.index](i)
```
I imagine this behaviour could become burdensome for numerical library writers.

martinvuyk:
what about doing
```mojo
fn foo[type: DType = DType.int64](s: Scalar[type]):
    constrained[type.is_integral(), "Expected an integral"]()
    print(s)


fn main():
    foo(UInt64(123))
    foo(Int32(123))
    foo(UInt16(123))
    foo(Int8(123))
    foo(int(123))
```

bgreni:
> what about doing
> 
> ```mojo
> fn foo[type: DType = DType.int64](s: Scalar[type]):
>     constrained[type.is_integral(), "Expected an integral"]()
>     print(s)
> 
> 
> fn main():
>     foo(UInt64(123))
>     foo(Int32(123))
>     foo(UInt16(123))
>     foo(Int8(123))
>     foo(int(123))
> ```

That works well, but it results in accepting `Bool` implicitly as well which might be undesirable? I'm not sure that the opinion of the team is on that sort of behaviour honestly.

martinvuyk:
> results in accepting Bool implicitly

oof, might be a problem. But i think that would rather be on the developers side (?)

bgreni:
[...]

PR #5177, opened 2025-08-13
[stdlib][proposal] Add arithmetic trait proposal

bgreni:
Related to #2776

This proposal isn't 100% fleshed out at the moment, but I am putting it up here anyways to get some feedback, and hopefully stimulate some further discussion on the topic.

CC: @laszlokindrat @owenhilyard @martinvuyk

### DIFF
--- a/mojo/proposals/arithmetic_trait.md
+++ b/mojo/proposals/arithmetic_trait.md
@@ -1,0 +1,273 @@
+# Abstracting Over Arithmetic Types
+
+## Motivation
+
+As Mojo becomes more widely used across scientific computing domains, giving
+library writers the ability to effectively abstract over arithmetic types will
+be crucial to its success in those areas. As Mojo intends to give users
+low-level control over exotic hardware, organizing such an abstraction quickly
+becomes non-trivial as we must consider operations where the output type is
+different than the inputs. Including dependent type operations where the
+output type becomes a function of the input types.
+
+In this proposal I have taken some liberties assuming language features that
+do not exist, nor are they planned to be implemented in the near future. The
+exact syntax isn't particularly important, but I have added them simply as a way
+of representing the sort of solution that I would find to be elegant.
+
+Currently if library writers want to define functions that operate on all the
+common numeric types, they must provide individual overloads for
+each of `SIMD`, `Int`, `UInt`, and any other types they may want to support.
+
+```mojo
+fn sum(*args: Int) -> Int:
+    ...
+
+fn sum(*args: UInt) -> UInt:
+    ...
+
+fn sum(*args: SIMD) -> __type_of(args):
+    ...
+```
+
+## The Arithmetic trait
+
+The solution is to build a heirachy of traits that are each responsible for
+defining one particular operation. The most trivial form would look like the
+following:
+
+```mojo
+trait Add:
+
+    fn __add__(self, other: Self) -> Self:
+        ...
+
+trait Sub:
+
+    fn __sub__(self, other: Self) -> Self:
+        ...
+
+trait Arithmetic(Add, Sub):
+    ...
+```
+
+Since in-place operations like `__iadd__` are impossible to implement for
+dependent types due to the result always being different than `Self`.
+They must be kept separate from the non-mutating operations that can return a
+new type.
+
+```mojo
+
+trait IAdd:
+
+    fn __iadd__(self, other: Self):
+        ...
+
+trait IArithmetic(IAdd):
+    ...
+```
+
+We can handle certain special cases using associated aliases. Such as when an
+operation produces a concrete result that doesn't match the input types,
+a good example being `Int.__truediv__` which returns a `Float64`.
+It also might be nice to consider that associated aliases could have default
+values to save a lot of boilerplate in the common case of simply returning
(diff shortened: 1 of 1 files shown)


owenhilyard:
Here's a functional Rust equivalent of the C++ code with some quality of life tacked on. 

```rust
trait Addable<R> {
    type Output;

    fn add(self: &Self, rhs: &R) -> Self::Output;
}

trait AddableLOutput<R> {
   fn add(self: &Self, rhs: &R) -> Self;
}

impl<L, R> Addable<R> for L where L: AddableLOutput<R> {
    type Output = L;

    fn add(self: &L, rhs: &R) -> L {
        return AddableLOutput::add(self, rhs);
    }
}

trait AddableSelf {
    fn add(self: &Self, rhs: &Self) -> Self;
}

impl<T> AddableLOutput<T> for T where T: AddableSelf {
   fn add(self: &T, rhs: &T) -> T {
       return AddableSelf::add(self, rhs);
   }
}
```

owenhilyard:
I think that, given the existence of `IntLiteral`, we don't have good options except to declare the base `Add` trait as one which needs 3 types, `Lhs`, `Rhs` and `Output`. lhs will be `Self` in all cases, and beyond that the specializations I put into the Rust version should make it more reasonable for more normal uses (we can bikeshed the names later). 

I think that we'll need to do similar things for most operations, so I agree with this proposal.

laszlokindrat:
@bgreni Thanks for writing this up! You bring up very interesting points, and I would encourage y'all to continue iterating on this. There are a couple general points of feedback I would like you to consider:
- I'm really looking for more detail on a trait hierarchy, and how existing (e.g. literals, SIMD, complex) and future (e.g. BigInt) types might fit into it. Examples here would be very helpful, as well as looking at some concrete corner cases (e.g. how would these traits model adding a `SIMD` to an `Int`).
- You are quickly jumping to the conclusion that we need parametric traits, but I would like a more detailed description of why associated parametric aliases wouldn't work (@martinvuyk [provided some here](https://github.com/modular/modular/issues/2776#issuecomment-3163529278) but it needs a cleaner explanation). You also had the idea of overloaded aliases, which might help with this.

If we conclude that we cannot properly model numeric types without parametric traits, that can be evidence that we need parametric traits, but it will also likely mean that we will have to wait with accepting/implementing this proposal.

One more related thing I want you to be aware of: we are looking at replacing `Int`/`UInt`/`Bool` with just `SIMD` aliases. I don't know yet if this is possible, and if so, at what cost. But if we succeed, it would mean that our core numeric types would be only `SIMD` and the literals.

bgreni:
> * I'm really looking for more detail on a trait hierarchy, and how existing (e.g. literals, SIMD, complex) and future (e.g. BigInt) types might fit into it. Examples here would be very helpful, as well as looking at some concrete corner cases (e.g. how would these traits model adding a `SIMD` to an `Int`).

Good point, I will add more concrete examples of what an implementation would look like.

> * You are quickly jumping to the conclusion that we need parametric traits, but I would like a more detailed description of why associated parametric aliases wouldn't work (@martinvuyk [provided some here](https://github.com/modular/modular/issues/2776#issuecomment-3163529278) but it needs a cleaner explanation). You also had the idea of overloaded aliases, which might help with this.

I do briefly touch on the fact that parametric traits allow us to easily express that a type has multiple overloads, as well as express things like a function arg that is addable with a particular type. It's possible this falls more into a  "nice to have" though, so I'll do some more thinking about whether parametric traits would actually be a hard blocker here.

bgreni:
> One more related thing I want you to be aware of: we are looking at replacing `Int`/`UInt`/`Bool` with just `SIMD` aliases. I don't know yet if this is possible, and if so, at what cost. But if we succeed, it would mean that our core numeric types would be only `SIMD` and the literals.

@laszlokindrat I see some progress is being made on this (fbb341dfccd1046df1d7a51980b830dd4c2800ad). I actually believe this strengthens my proposal because while we should still eventually have a highly generic solution like the one I've outlined in this document. Having the power to cover 90% of use cases being being generic over `SIMD` types removes much of the urgency I was feeling to potentially implement any half measures in the mean time while we wait for certain language features.

laszlokindrat:
> @laszlokindrat I see some progress is being made on this ([fbb341d](https://github.com/modular/modular/commit/fbb341dfccd1046df1d7a51980b830dd4c2800ad)). I actually believe this strengthens my proposal because while we should still eventually have a highly generic solution like the one I've outlined in this document. Having the power to cover 90% of use cases being being generic over `SIMD` types removes much of the urgency I was feeling to potentially implement any half measures in the mean time while we wait for certain language features.

@bgreni, yeah, I agree, we should revisit once this unification is complete.

josiahls:
Bumping this. I have structs that I would like to parameterize on Float and custom FixedPointInt structs. I don't think this is possible without DType implementing some basic numeric traits.

bgreni:
We have achieved int unification with SIMD, and with Mojo 1.0 looming perhaps we could pickup this discussion again in a few months since we'll be able to take the parametric trait component more seriously.

@JoeLoser soft ping as I realize Laszlo is still assigned to this though he doesn't work at Modular anymore

PR #5182, opened 2025-08-16
[stdlib] Update `pixi.lock` file

martinvuyk:
Update `pixi.lock` file. Every time I do formatting or something pixi keeps deleting these lines from the lockfile. Not sure if this should be merged or not. I imagine it has something to do with the change of splitting max and mojo binaries


Ahajha:
Huh. Let me find the root cause of this.

martinvuyk:
@Ahajha I think it's because the pixi.toml file doesn't have max anymore
```toml
[dependencies]
python = ">=3.9,<3.14"
lit = "*"
mojo = "*"
```

Ahajha:
Okay I've figured out the root cause. You're correct this is from max/mojo split, I thought I had preemptively fixed it but this caught a small bug, it was always pinning `max` since that was the previous "lowest" package, it should have been pinning `mojo-compiler`. The newest nightly just went out, but this should be fixed in the next one. If you want to rebase this PR we can merge it, if it's bothering you. Thanks for finding this!

martinvuyk:
@Ahajha so the next nightly will make this PR redundant right? if so I can just close this

Ahajha:
Yes, correct. There should be a few other lock files that will get fixed similar to this one as well.

Issue #5179, opened 2025-08-14
[Feature Request] Support `UnsafePointer[T].__sub__(UnsafePointer[T])`


JoeLoser:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

I wanted to calculate the distance between two (related) pointers. In C terms, (p-q) for "void* p, *q". The following does not compile in Mojo, and the error message is generic and not helpful.

```
fn main():
  var y = UnsafePointer[Int32].alloc(1)
  var q = y-y
  print(q)
```

```
mojo align2.mojo
/Users/denis/work/v8/align2.mojo:4:12: error: invalid call to '__sub__': failed to infer parameter 'I'
  var q = y-y
          ~^~
/Users/denis/work/v8/align2.mojo:4:13: note: failed to infer parameter #5, argument type 'UnsafePointer[SIMD[int32, 1], origin={}]' does not conform to trait 'Indexer'
  var q = y-y
            ^
Included from /Users/denis/work/v8/open-source/max/mojo/stdlib/stdlib/memory/__init__.mojo:1:
Included from /Users/denis/work/v8/open-source/max/mojo/stdlib/stdlib/memory/__init__.mojo:22:
/Users/denis/work/v8/open-source/max/mojo/stdlib/stdlib/memory/unsafe_pointer.mojo:268:8: note: function declared here
    fn __sub__[I: Indexer, //](self, offset: I) -> Self:
```

### What is your motivation for this change?

Mojo is a low level language. Should it support pointer arithmetics, computing distance between two pointers in one address space? I think yes 😄 

### Any other details?

We should return the number of *elements* not bytes here from this methods.  I'd suggest having it be a member method on `UnsafePointer`.  Here is one implementation as a *free function* for reference

```
# Distance in _elements_ (not bytes) between two pointers
fn pointer_distance[T: AnyType](ptr1: UnsafePointer[T], ptr2: UnsafePointer[T]) -> Int:
    return abs(Int(ptr2) - Int(ptr1)) // sys.sizeof[T]()
```

This would be something like https://doc.rust-lang.org/std/primitive.pointer.html#method.offset_from.

PaulOberg1:
Hi @JoeLoser , I'm interested in contributing to this. Could I be assigned please? :)

JoeLoser:
[...]

PR #5185, opened 2025-08-17
[stdlib] Implement UnsafePointer.__sub__ for signed pointer subtraction

PaulOberg1:
This PR introduces a new signed subtraction method `__ssub__` for UnsafePointer, allowing
subtraction of pointers using both signed and unsigned integer types.  

Additionally, it adds comprehensive tests to validate pointer subtraction behavior:  
- `test_pointer_subtraction`: basic subtraction using standard integer types.  

These changes improve the safety and correctness of pointer arithmetic and ensure
consistent behavior across integer types.

Addresses issue: [#5179 ](https://github.com/modular/modular/issues/5179#issuecomment-3190547480)


PaulOberg1:

laszlokindrat:
@PaulOberg1 Thank you for the patch, this looks promising! Seems like there is a genuine compiler error caught by one of the CI jobs; can you please investigate and make sure it's all green before we review in detail?

mojo/stdlib/test/memory/test_unsafepointer.mojo
```diff
@@ -330,6 +330,50 @@ def test_offset():
     ptr2.free()
 
 
+# Test UnsafePointer subtraction with standard integer offsets
```
laszlokindrat: Can you please drop these comments in favor of inline comments within the test function?

mojo/stdlib/test/memory/test_unsafepointer.mojo
```diff
@@ -330,6 +330,50 @@ def test_offset():
...
+def test_pointer_subtraction():
+    # -----------------------------------------
+    # Basic pointer subtraction
+    # -----------------------------------------
```
laszlokindrat: Can you please remove these visual separators (here and elsewhere in this file).
```suggestion
    # Basic pointer subtraction
```

mojo/stdlib/test/memory/test_unsafepointer.mojo
```diff
@@ -330,6 +330,50 @@ def test_offset():
...
+    # Subtract two pointers to get element difference
+    assert_equal(end_ptr.__sub__(start_ptr), 5)
+    assert_equal(start_ptr.__sub__(end_ptr), -5)
+    assert_equal(end_ptr.__sub__(end_ptr), 0)
+
+    ptr.free()
+
+
+# Test UnsafePointer subtraction using SIMD and multiple integer types
+def test_pointer_subtraction_simd():
```
laszlokindrat: IMO it's okay to collapse these two into a single test case now, and just have a comment in the function body.

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
```diff
@@ -366,6 +366,24 @@ struct UnsafePointer[
...
+        Parameters:
+            T: The type of element the pointer points to.
+
+        Args:
+            other: Another pointer of the same type.
+
+        Returns:
+            The number of elements (of type T) between self and other.
+        """
+        var element_size = sizeof[T]()
```
NathanSWard:
```suggestion
        alias element_size = sizeof[T]()
```
This can be evaluated at compile time so you could use `alias` here

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
```diff
@@ -366,6 +366,24 @@ struct UnsafePointer[
...
+        var byte_diff = self - other
+        return Int(byte_diff / element_size)
```
NathanSWard: Something to note: by using `/` the `Int.__truediv__(Int)` function is called which converts both values to floats and then returns a float which we manually cast back to an `Int`. I believe you can avoid the int -> conversions completely by using the [index.divs](https://mlir.llvm.org/docs/Dialects/IndexOps/#indexdivs-indexdivsop) operation.

Looking at the llvm ir output [here](https://godbolt.org/z/Kh7cfvPTr) we can avoid the `fp <-> si` conversions.

mojo/stdlib/test/memory/test_unsafepointer.mojo
PaulOberg1: Done 👍

mojo/stdlib/test/memory/test_unsafepointer.mojo
PaulOberg1: Done 👍

mojo/stdlib/test/memory/test_unsafepointer.mojo
PaulOberg1: Done 👍

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
PaulOberg1: Done 👍

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
PaulOberg1: Thanks for the insight, Nathan! From what I’ve read, Mojo lowers through MLIR before LLVM, and MLIR already exposes the divs (signed) and divu (unsigned) ops for integer division. Would it make sense here to call that directly instead of using /?

Something along the lines of:
```mojo

var element_size = sizeof[T]()
var byte_diff = self.address - other.address
return intrinsics.divs(byte_diff, element_size)
```
That way we’d avoid the intermediate float conversion in the generated LLVM IR, and stay closer to the “raw” integer semantics. Curious if you think that’s overkill compared to keeping / for readability.

PaulOberg1:
Thanks for the feedback @laszlokindrat  and @NathanSWard !
I’ve made several iterations on this and resolved most of the issues, but I’m still hitting some CI/compiler errors I can’t reproduce locally. I’d love a bit of guidance on the best way to resolve these last hurdles so the PR can get green. I’m happy to do the work - just want to make sure I’m heading in the right direction :)

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
NathanSWard: Correct, we can directly use the
```mojo
__mlir_op.`index.divs`(element_size.value, byte_diff.value)
```

@lattner also mentions the LLVM `sdiv exact` instruction however, I'm unsure how we are able to convert the `__mlir_type.index` to a type that plays nicely with the llvm instruction. (I was looking at the `arith.index_cast` mlir dialect but I don't think we have access that that yet).

Ahajha:
Hi @PaulOberg1, please rebase your branch, things should be fixed now.

laszlokindrat:
@PaulOberg1 this looks good to me, but CI is still failing; can you please rebase and investigate?

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
```diff
@@ -366,6 +366,25 @@ struct UnsafePointer[
         """
         return self + (-1 * index(offset))
 
+    @always_inline("nodebug")
+    fn __sub__[T: AnyType](self: UnsafePointer[T], other: UnsafePointer[T]) -> Int:
```
laszlokindrat, line 370: What's the reason for restricting the self type here?

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
```diff
@@ -366,6 +366,25 @@ struct UnsafePointer[
...
+            T: The type of element the pointer points to.
+
+        Args:
+            other: Another pointer of the same type.
+
+        Returns:
+            The number of elements (of type T) between self and other.
+        """
+        alias element_size = sizeof[T]()
+        var byte_diff: Int = cast(Int, self) - cast(Int, other)
```
laszlokindrat: There isn't a `cast` free function in this module, which is why CI is failing. Also note that we already have a `__sub__` overload that works for `Indexer` instances; can we delegate to that (or the other way around)?

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
PaulOberg1: Good point, will remove that

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
PaulOberg1: Hmm, not sure how that would work though? I switched this to use builtin.ptrtoint since that seems to be the right way to handle pointer-to-int conversion here.

PaulOberg1: Also, there is an issue with my local environment so I am having to commit changes in order to test them, hence the delays.

PaulOberg1:
Hey @laszlokindrat @NathanSWard I am still having issues with my local development environment and am only able to test code by committing it. If possible, could someone identify what is causing the build errors so we can save time and prevent the PR being cluttered by loads of commits?

Ahajha:
@PaulOberg1 What issues are you running into? Error logs would be helpful, of any kind.

PaulOberg1:
@Ahajha Here is the error log:
image
I also attempted various other strategies, some of which I have shown below, none of these worked:
image
image
image
image
image
image
image
image

Ahajha:
That seems to be the same error as in CI. But not sure how to fix, that's more of a mojo issue than a build system one.

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
soraros: There's `Int._positive_div` which wraps `index.divs`.

mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
soraros: We may also want to generalize over the parameters of `other` here (I hate heterogeneous ops/comparisons, but that's the only way to make these useful). CC @NathanSWard.

soraros:
This should work

```mojo
fn main():
  p = UnsafePointer[Int]()
  q = p + 1

  var m = Int(p)
  var n = Int(q)
  print((n - m)._positive_div(4))
```

PR #4267, opened 2025-04-03
[stdlib] Add `repr` support for `List[RepresentableCollectionElement].

rd4com:
Hello,
this adds an overload to `repr` to support it.

It is a workaround that has to be replaced one day,
but seem like something widely used by users !

```mojo
 print(repr(List[UInt8](0, 1)))
 # [SIMD[DType.uint8, 1](0), SIMD[DType.uint8, 1](1)]
```
Also work with `List[DType]` for metaprogramming:
```mojo
def main():
    x = List(DType.int8, DType.int16)
    print(repr(x))
    print(DType.int8 in x)
```

PR #5189, opened 2025-08-18
[stdlib] Add `repr` support for `List`, `Deque`, `Dict`, `LinkedList`, `Optional`, `Set`.

rd4com:
Hello,
this adds an overload to `repr` to support it. (redo of PR 4267, ping to @laszlokindrat )

It is a workaround that has to be replaced one day,
but seem like something widely used by users !

```mojo
 print(repr(List[UInt8](0, 1)))
 # [SIMD[DType.uint8, 1](0), SIMD[DType.uint8, 1](1)]
```
Also work with `List[DType]` for metaprogramming:
```mojo
def main():
    x = List(DType.int8, DType.int16)
    print(repr(x))
    print(DType.int8 in x)
```


rd4com:

JoeLoser (approved):
LGTM, thanks!  Should we also add future convenient wrapper `repr` function support for `Dict`, `Set`, and friends until we have better conditional conformances?  This may future-proof some "why does this work on `List` but not <xyz> common core collection" questions.

rd4com:
Sure, that would make sense (ping to @JoeLoser) ,
would you prefer separate PR's ?

JoeLoser:
> Sure, that would make sense (ping to @JoeLoser) , would you prefer separate PR's ?

You can keep them in this PR if you'd like as long as you update the PR title.  Don't forget to add a changelog entry like you did in https://github.com/modular/modular/pull/4267/files#diff-e08d18780192c5dabac085c1c92471c74079075158239ff056b015f32f6b0094R119! 😄

rd4com:
Great :+1: , have i forgot anything ? (ping to @JoeLoser)
Few observations:
- Importing `Set`, `LinkedList`,.. in `repr`. (Does it make theses builtin ?) 
- Would not work when nested (ex: `List[List[Int]]`)

But it works so that is cool!

mojo/stdlib/test/collections/test_optional.mojo
```diff
@@ -186,6 +186,13 @@ def test_optional_unwrap():
         _ = a[]
 
 
+def test_optional_repr_wrap():
+    var o = Optional(10)
+    assert_equal(repr(o), "Optional(10)")
+    o = None
+    assert_equal(repr(o), "Optional(None)")
```
laszlokindrat, line 193: Unrelated to this PR, but this is a bit confusing. `Optional(10)` should probably just print `10`, since that's the actual value (and similarly, the second test should yield the same as `repr(None)`. The current implementation is kinda weird, since it prints `Optional`, but not the underlying type.

laszlokindrat:
> Few observations:
> 
> * Importing `Set`, `LinkedList`,.. in `repr`. (Does it make theses builtin ?)

No, the implicit imports are controlled by `stdlib/prelude/__init__.mojo`.

> * Would not work when nested (ex: `List[List[Int]]`)

Yep, and that's okay. We are working on conditional conformance, which should address this properly.

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

rd4com:
It's been a while since my last pr,
thanks to both you for reviving me :bow: !
I finally have bazel setup and can run tests :+1: .

mojo/stdlib/test/collections/test_optional.mojo
rd4com: Yep, `List` for example do give the type :+1:

laszlokindrat: If you want, seems like fixing this could be a good followup :)

rd4com: Sure, let's go about fixing this (followup)! 
Here is an overview of what types do on `repr`:

Collections type do not always print their own type, some hints at it ("[]","{}"):
-Set: "{a,b,c}"
-List: "[a,b,c]"
-Dict: "{a:b,c:d}"
-Deque: "Deque(a,b,c)"
-LinkedList: "LinkedList(a,b,c)"
-Optional: "Optional(a)"

SIMD include it's DT:  
- repr -> "SIMD\[DType.float64, 1\]\(1.0\)"  

Optional, deque, LinkedList, List,.. don't include T: 
- repr -> "[0], {0}, {a:b}"

PR #5192, opened 2025-08-19
[bazel] Add linker driver to modular_py_binary

keith:
This is required in some cases where the python test runs mojo


keith:
fixed by https://github.com/modular/modular/commit/0daffe5e746451651bccca2d8cd896895e94cbdf

PR #5208, opened 2025-08-23
[stdlib] Add tests for `T.__*__is_trivial`.

rd4com:
An type is given as an parameters and the struct is instantiated. 
The struct value is then interacted with in any ways.

As interactions goes, the struct register the corresponding events for:
- `__init__`
- `__moveinit__`
- `__copyinit__`
- `__del__`

The event registered depends of `T`.
For example, when `T.__moveinit__is_trivial`,
The event registered on `__moveinit__` is `event_move | event_trivial`.

We can then compare the sequence of events to the expected one.

### DIFF
--- a/mojo/stdlib/test/builtin/test_value.mojo
+++ b/mojo/stdlib/test/builtin/test_value.mojo
@@ -13,17 +13,16 @@
 
 
 from testing import assert_false, assert_true, assert_equal
-from memory import memcpy
 
 # ===-----------------------------------------------------------------------===#
 # Triviality Struct
 # ===-----------------------------------------------------------------------===#
 
-alias event_trivial = 1
-alias event_init = 2
-alias event_del = 4
-alias event_copy = 8
-alias event_move = 16
+alias EVENT_TRIVIAL = 0b1  # 1
+alias EVENT_INIT = 0b10  # 2
+alias EVENT_DEL = 0b100  # 4
+alias EVENT_COPY = 0b1000  # 8
+alias EVENT_MOVE = 0b10000  # 16
 
 
 struct ConditionalTriviality[T: Movable & Copyable](Copyable, Movable):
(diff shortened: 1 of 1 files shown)


mojo/stdlib/test/builtin/test_value.mojo
```diff
@@ -0,0 +1,110 @@
...
+
+# ===-----------------------------------------------------------------------===#
+# Triviality Struct
+# ===-----------------------------------------------------------------------===#
+
+alias event_trivial = 1
+alias event_init = 2
+alias event_del = 4
+alias event_copy = 8
+alias event_move = 16
```
martinvuyk: My nit-pick here would be that IMO using `0b0001_0000`, `0b0000_1000` or `0b10000` and `0b1000` will be clearer since these are bit flags

mojo/stdlib/test/builtin/test_value.mojo
```diff
@@ -0,0 +1,110 @@
...
+
+
+struct ConditionalTriviality[T: Movable & Copyable](Copyable, Movable):
+    var events: UnsafePointer[List[Int]]
```
martinvuyk: Since we are starting from scratch with this type, I think it would be best to make it use origins already (see  #5122 where I'm trying to patch existing testing utils)

martinvuyk (commented):
Hi, this looks interesting. How would we use it to test collections though? Since its destructor, move, and copy are non-trivial it should record every event happening which is cool, but does that mean this replaces all types in `test_utils` ?

This would also make collections that use different code-paths for trivial and non-trivial ops always go the non-trivial route due to how the type is made, so I'm not sure how this helps test the trivial paths

mojo/stdlib/test/builtin/test_value.mojo
laszlokindrat: +1, this would show the intent more clearly.

mojo/stdlib/test/builtin/test_value.mojo
```diff
@@ -0,0 +1,110 @@
...
+
+
+from testing import assert_false, assert_true, assert_equal
+from memory import memcpy
+
+# ===-----------------------------------------------------------------------===#
+# Triviality Struct
+# ===-----------------------------------------------------------------------===#
+
+alias event_trivial = 1
```
laszlokindrat: Can you please change these to follow the naming convention for global aliases? E.g.
```suggestion
alias EVENT_TRIVIAL = 1
```

mojo/stdlib/test/builtin/test_value.mojo
PeimingLiu: +1

laszlokindrat (approved):
LGTM, thank you! Can you please address the nits, and I'll bring it in.

rd4com:
Thanks @martinvuyk, @laszlokindrat, @PeimingLiu & others ,
i'll work on this pr today, 
so that the suggestions are integrated into the PR today :+1:

mojo/stdlib/test/builtin/test_value.mojo
rd4com: Thanks, reviews integrated :+1:

mojo/stdlib/test/builtin/test_value.mojo
rd4com: Thanks, review integrated :+1: 
(i'll make sure to learn the naming convention)

mojo/stdlib/test/builtin/test_value.mojo
rd4com: Thanks, the review is integrated :+1:

rd4com:
> Hi, this looks interesting. How would we use it to test collections though? Since its destructor, move, and copy are non-trivial it should record every event happening which is cool, but does that mean this replaces all types in `test_utils` ?
> 
> This would also make collections that use different code-paths for trivial and non-trivial ops always go the non-trivial route due to how the type is made, so I'm not sure how this helps test the trivial paths

Hello, 
no it does not replace all the `test_utils` types,
just an test for the trivial aliases :+1:

rd4com:
Reviews integrated @laszlokindrat :+1:

laszlokindrat:
!sync

Issue #5003, opened 2025-07-11
[BUG] `DType.index` is getting lowered to `DType.int64` too early in the comptime pipeline


martinvuyk:
### Bug description

```mojo
from sys.info import bitwidthof
from bit import count_leading_zeros


fn is_negative[dtype: DType, //](value: SIMD[dtype, _]) -> __type_of(value):
    return value >> (bitwidthof[dtype]() - 1)


fn main():
    alias value = Scalar[DType.index](-1)
    print(is_negative(value))  # works fine
    alias a = is_negative(value)  # crashes
    print(a)
    alias b = count_leading_zeros(value)  # crashes
    print(b)
```
```text
open-source/max/mojo/stdlib/stdlib/builtin/simd.mojo:1101:35: note: failed to fold operation pop.shr(#pop<simd -1> : !pop.scalar<index>, #pop<simd 63> : !pop.scalar<index>)
```
```text
note: failed to interpret operation pop.call_llvm_intrinsic{fastmathFlags: #pop<fmf none>, hasSideEffects: false, intrin: "llvm.ctlz" : !kgen.string}(#kgen.pack<-1, 0> : !kgen.pack<[[typevalue<#kgen.instref<@"stdlib::builtin::simd::SIMD,dtype=index,size=1">>, scalar<index>], [typevalue<#kgen.instref<@"stdlib::builtin::bool::Bool">>, i1]]>)
open-source/max/mojo/stdlib/stdlib/bit/bit.mojo:66:78: note: call expansion failed
open-source/max/mojo/stdlib/stdlib/sys/intrinsics.mojo:91:10: note: LLVM intrinsic operand has unknown value: #pop<simd -1> : !pop.scalar<index>
```

### Steps to reproduce

...

### System information

`Mojo 25.5.0.dev2025070905`

soraros:
CC @npanchen.

npanchen:
thanks. I'll take it.
I always want to go and fix all folders/canonicalizers w.r.t. index/dtype.index type, but that's quite a lot of tedious work.
Basically, in order to do this right, we have to query index's size from data layout, which is not yet done in many ops.

martinvuyk:
@npanchen  #5238 is still failing :(
```
[...]

PR #5211, opened 2025-08-23
[stdlib] Add generic min max finite for integers

martinvuyk:
Add generic min max finite for integers.

CC: @soraros

### DIFF
--- a/mojo/stdlib/stdlib/utils/numerics.mojo
+++ b/mojo/stdlib/stdlib/utils/numerics.mojo
@@ -702,26 +702,12 @@ fn max_finite[dtype: DType]() -> Scalar[dtype]:
     """
 
     @parameter
-    if dtype is DType.int8:
-        return 127
-    elif dtype is DType.uint8:
-        return 255
-    elif dtype is DType.int16:
-        return 32767
-    elif dtype is DType.uint16:
-        return 65535
-    elif dtype is DType.int32 or (
-        dtype is DType.index and bitwidthof[DType.index]() == 32
-    ):
-        return 2147483647
-    elif dtype is DType.uint32:
-        return 4294967295
-    elif dtype is DType.int64 or (
-        dtype is DType.index and bitwidthof[DType.index]() == 64
-    ):
-        return 9223372036854775807
-    elif dtype is DType.uint64:
-        return 18446744073709551615
+    if dtype.is_unsigned():
+        return ~Scalar[dtype](0)
+    elif dtype.is_integral():
+        return Scalar[dtype](
+            ~Scalar[_unsigned_integral_type_of[dtype]()](0) >> 1
+        )
     elif dtype is DType.float8_e4m3fn:
         return 448
     elif dtype is DType.float8_e4m3fnuz:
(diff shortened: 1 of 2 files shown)


soraros:
LGTM. Maybe we should add a fixme in test saying that we don't yet cover larger int types.

martinvuyk:
@soraros could you double check the implementation and test numbers? I changed some things and am trying to add tests for bigger integers. I'm running into this error when running `pixi run tests ./stdlib/test/utils/` locally:
```
mojo/stdlib/stdlib/builtin/simd.mojo:641:10: error: integer value 115792089237316195423570985008687907853269984665640564039457584007913129639935 requires 257 bits to store, but the destination bit width is only 128 bits wide
...
mojo/stdlib/stdlib/builtin/simd.mojo:641:10: error: integer value 57896044618658097711785492504343953926634992332820282019728792003956564819967 requires 256 bits to store, but the destination bit width is only 128 bits wide
...
mojo/stdlib/stdlib/builtin/simd.mojo:641:10: error: integer value 340282366920938463463374607431768211455 requires 129 bits to store, but the destination bit width is only 128 bits wide
...
mojo/stdlib/stdlib/builtin/simd.mojo:641:10: error: integer value -57896044618658097711785492504343953926634992332820282019728792003956564819968 requires 256 bits to store, but the destination bit width is only 128 bits wide
```
and in the CI it seems like another version of the compiler bug in #5003

mojo/stdlib/stdlib/utils/numerics.mojo
```diff
@@ -702,26 +702,12 @@ fn max_finite[dtype: DType]() -> Scalar[dtype]:
...
-    ):
-        return 9223372036854775807
-    elif dtype is DType.uint64:
-        return 18446744073709551615
+    if dtype.is_unsigned():
+        return ~Scalar[dtype](0)
+    elif dtype.is_integral():
+        return Scalar[dtype](
+            ~Scalar[_unsigned_integral_type_of[dtype]()](0) >> 1
+        )
```
martinvuyk, line 710: What do you think of this iteration of the implementation doing this assuming that unsigned shift right fills with zero? Do you know of any system that might not do this?

martinvuyk (commented):
@soraros I just redid the numbers using straight hexadecimal and I opened bug report #5214 because now I'm confident it is one

soraros:
Sorry, I'm on my phone these days so I didn't take the time to refer you to the int literal issue (that was the reason I didn't fix these functions sometime ago in the first place). Commented in your new issue as well.

mojo/stdlib/test/utils/test_numerics.mojo
```diff
@@ -170,6 +170,36 @@ def test_max_finite():
...
+    assert_equal(max_finite[DType.uint8](), 255)
+    assert_equal(max_finite[DType.int16](), 32767)
+    assert_equal(max_finite[DType.uint16](), 65535)
+    assert_equal(max_finite[DType.int32](), 2147483647)
+    assert_equal(max_finite[DType.uint32](), 4294967295)
+    assert_equal(max_finite[DType.int64](), 9223372036854775807)
+    assert_equal(max_finite[DType.uint64](), 18446744073709551615)
+    # FIXME(#5214): uncomment once it is closed
+    # assert_equal(
+    #     max_finite[DType.int128](), 0x7FFF_FFFF_FFFF_FFFF_FFFF_FFFF_FFFF_FFFF
```
lsh, line 183: Nit: I'm not sure I find the hex representation strictly better here. At least, better enough that these comments and the FIXME is worth it.

lsh:
!sync

mojo/stdlib/test/utils/test_numerics.mojo
martinvuyk: I find it easier to confirm that the number is correct by counting set hex bits given the number has to be base 2. The decimal version is also much longer

Issue #5166, opened 2025-08-12
[Feature Request] Rename `InstanceOf` to express trait implementation


gryznar:
### What is your request?

As title

### What is your motivation for this change?

`InstanceOf` introduced via [commit](https://github.com/modular/modular/commit/f2537ead3d03faaff4085727af9c1ebe08f84208) is used in context of traits which cannot be instantiated. In such case naming which empasises implementation should be used.

As of comments, we need 2 names: for trait implementation statically and dynamically.

@laszlokindrat

nmsmith:
Personally, I don't mind using "instance of" to describe an instance of a struct that has a trait.

In everyday English, it makes sense. My cat is both an instance of a cat, an instance of an animal, an instance of a feedable, etc. Instance just means an occurrence/example of something.

nmsmith:
That said, it's worth considering the context in which this `InstanceOf` feature will exist. Looking at this broader context, I can see some other potential issues with the name.

## The broader context

Swift and Rust offer _two_ closely related features, each with a distinct syntax:
1. Declaring a variable to have a **statically known** type conforming to a trait bound. (`some X` in Swift, `impl X` in Rust.)
2. Declaring a variable to have a **dynamically dispatched** type conforming to a trait bound. (`any X` in Swift, `dyn X` in Rust.)

I'll assume readers are similar with these features. If not, check out the Swift and Rust docs.

Mojo will likely need its own versions of the above two features, so we should try to find names that are self-explanatory.

Unfortunately, `InstanceOf` doesn't really clarify whether the type's trait methods are dispatched statically or dynamically. (It's the former.)

For readers who enjoy a bit of bikeshedding, that sets the stage. Let the bikeshedding commence?

## Syntax examples

To set the scene, here are some potential syntax for the equivalent of Swift's `some` + `any`:
- `fn foo(x: Static[Stringable])`  and `fn bar(x: Dynamic[Stringable])`
- `fn foo(x: Fixed[Stringable])`  and `fn bar(x: Dynamic[Stringable])`
- `fn foo(x: Known[Stringable])`  and `fn bar(x: Dynamic[Stringable])`
- `fn foo(x: Param[Stringable])`  and `fn bar(x: Runtime[Stringable])`
- `fn foo(x: Inline[Stringable])`  and `fn bar(x: Boxed[Stringable])`

...and so on.

[...]

PR #5213, opened 2025-08-24
[stdlib][doc] Add `Some[]` utility to style guide

christoph-schlumpf:
- Add a section in the `style-guide.md` to advice on when to use the `Some[]` utility
- As an example `write_to` is refactored accordingly in the stdlib

Relates to discussions in https://github.com/modular/modular/issues/5166


ConnorGray:
!sync

mojo/stdlib/stdlib/io/write.mojo
```diff
@@ -148,13 +148,10 @@ trait Writable:
     ```
     """
 
-    fn write_to[W: Writer](self, mut writer: W):
```
laszlokindrat, line 151: @ConnorGray Are you okay with this change?

PR #5215, opened 2025-08-24
[stdlib] Fix big IntLiteral casting

### LINKED ISSUE #4049: [BUG] Cant instantiate variable of UInt256 with more than 128 bit value

martinvuyk:
Fix big IntLiteral casting. Closes #4049. Closes  #4212

CC: @soraros

### DIFF
--- a/mojo/stdlib/stdlib/builtin/dtype.mojo
+++ b/mojo/stdlib/stdlib/builtin/dtype.mojo
@@ -22,10 +22,10 @@ from sys import CompilationTarget, bitwidthof, sizeof
 from sys.intrinsics import _type_is_eq
 
 
-alias _mIsSigned = UInt8(1)
-alias _mIsInteger = UInt8(1 << 7)
-alias _mIsNotInteger = UInt8(~(1 << 7))
-alias _mIsFloat = UInt8(1 << 6)
+alias _mIsSigned = UInt8._unsafe_from_raw_intliteral(1)
+alias _mIsInteger = UInt8._unsafe_from_raw_intliteral(1 << 7)
+alias _mIsNotInteger = UInt8._unsafe_from_raw_intliteral(~(1 << 7))
+alias _mIsFloat = UInt8._unsafe_from_raw_intliteral(1 << 6)
 
 
 @register_passable("trivial")
--- a/mojo/stdlib/stdlib/builtin/simd.mojo
+++ b/mojo/stdlib/stdlib/builtin/simd.mojo
@@ -43,7 +43,7 @@ domain-specific libraries for machine learning and scientific computing.
 import math
 from collections import InlineArray
 from hashlib.hasher import Hasher
-from math import Ceilable, CeilDivable, Floorable, Truncable
+from math import Ceilable, CeilDivable, Floorable, Truncable, ceildiv
 from math.math import _call_ptx_intrinsic
 from sys import (
     CompilationTarget,
@@ -71,6 +71,7 @@ from builtin.math import Powable
 from documentation import doc_private
 from memory import bitcast, memcpy
 from python import ConvertibleToPython, PythonObject, Python
+from os import abort
 
 from utils import IndexList, StaticTuple
 from utils._visualizers import lldb_formatter_wrapping_type
@@ -621,7 +622,7 @@ struct SIMD[dtype: DType, size: Int](
             self = Scalar[dtype]()
             constrained[False, "unsupported dtype"]()
 
-    @always_inline("nodebug")
+    @no_inline
     @implicit
     fn __init__(out self, value: IntLiteral, /):
         """Initializes the SIMD vector with an integer.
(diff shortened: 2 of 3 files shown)


mojo/stdlib/stdlib/builtin/dtype.mojo
```diff
@@ -22,10 +22,10 @@ from sys import CompilationTarget, bitwidthof, sizeof
...
 
 
-alias _mIsSigned = UInt8(1)
-alias _mIsInteger = UInt8(1 << 7)
-alias _mIsNotInteger = UInt8(~(1 << 7))
-alias _mIsFloat = UInt8(1 << 6)
+alias _mIsSigned = UInt8._unsafe_from_raw_intliteral(1)
+alias _mIsInteger = UInt8._unsafe_from_raw_intliteral(1 << 7)
+alias _mIsNotInteger = UInt8._unsafe_from_raw_intliteral(~(1 << 7))
+alias _mIsFloat = UInt8._unsafe_from_raw_intliteral(1 << 6)
```
laszlokindrat, line 28: Why is this change needed?

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -621,7 +622,7 @@ struct SIMD[dtype: DType, size: Int](
             self = Scalar[dtype]()
             constrained[False, "unsupported dtype"]()
 
-    @always_inline("nodebug")
+    @no_inline
```
laszlokindrat, line 625: Are you doing this to ensure this function always gets fully evaluated at compile time (since IntLiteral doesn't have a runtime representation)? If so, I request that this gets documented in the docstring, or better yet, the function refactored into a helper that's called in a parameter context, and only the result materialized.

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -633,19 +634,77 @@ struct SIMD[dtype: DType, size: Int](
...
+            v = Scalar[dtype](0)
+            alias ratios = [1, 2, 4, 8]
+
+            @parameter
+            for r in ratios:
+                if r < ceildiv(value, 128):
+                    if r == 8:
+                        abort(
+                            "the current implementation can handle up to"
+                            " integers of 1024 bitwidth"
```
laszlokindrat: How can we possibly support something of length 1024 when our largest datatype is 256 bit?

laszlokindrat:
Thanks @martinvuyk! Before we do a detailed review, I would like to request that you provide a very detail description of the algorithm.

mojo/stdlib/stdlib/builtin/simd.mojo
martinvuyk: I don't know what other exotic datatypes we might support in the future. I could also just hard-code it for 256 :man_shrugging:

mojo/stdlib/stdlib/builtin/simd.mojo
martinvuyk: I'm just trying to avoid the timeout that I'm running into in CI and locally. Not sure if this helps by trying to avoid unrolling the inner loops everywhere

mojo/stdlib/stdlib/builtin/dtype.mojo
martinvuyk: Otherwise there is a recursive use of the constructor since `DType` uses these flags to know whether a dtype is integral etc.

martinvuyk:
@laszlokindrat I will once it works, the idea is basically just building uint from the raw bit value and bit shifting it into the result. I might just end up doing something else, but I opened the PR as a draft to get some help or comments in case I'm missing something since this is a bit too close to the compiler for my comfort

mojo/stdlib/stdlib/builtin/dtype.mojo
```diff
@@ -22,10 +22,10 @@ from sys import CompilationTarget, bitwidthof, sizeof
 from sys.intrinsics import _type_is_eq
 
 
-alias _mIsSigned = UInt8(1)
-alias _mIsInteger = UInt8(1 << 7)
-alias _mIsNotInteger = UInt8(~(1 << 7))
-alias _mIsFloat = UInt8(1 << 6)
+alias _mIsSigned = UInt8._unsafe_from_raw_intliteral(1)
```
gryznar, line 25: @laszlokindrat shall these camel case aliases be fixed?

soraros: https://github.com/modular/modular/pull/4212#discussion_r2024286213

Issue #5196, opened 2025-08-21
[Feature Request] Atomic fences


NathanSWard:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

It would be nice along with the current suite of the `Atomic` type, to have the ability to create atomic fences (see [Rust's here](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html) and [C++'s here](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence.html) for more info). 

It would look something along the lines of:
```mojo
# in atomic.mojo

fn fence[*, ordering: Consistency = Consistency.SEQUENTIAL]():
    ...

# or maybe since the function wouldn't take any arguments it could be an `alias`
alias fence[*, ordering: Consistency = Consistency.SEQUENTIAL] = ...
```

### What is your motivation for this change?

With mojo being a performance forward language, the ability to use atomic fences in various lock-free algorithms and data structures seems like a must have. 

One example is in `ArcPointer`'s implementation, we could better optimize the ref counting with fences.

JoeLoser:
Fixed with https://github.com/modular/modular/pull/5216 which landed internally.

PR #5216, opened 2025-08-25
[stdlib] Add atomic `fence`

NathanSWard:
This adds an `os.atomic.fence` function (synonymous with C++'s [atomic_thread_fence](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence.html)).

A few notes:
- I did not add the equivalent of [atomic_signal_fence](https://en.cppreference.com/w/cpp/atomic/atomic_signal_fence.html) but I can easily add that for completeness if so desired. Not sure on a good name for this however, since in Rust it's called `compiler_fence` and in LLVM talk it is an atomic fence with `scope ="singlethread"`.
- I could add quite a bit to the documentation for `fence`, however the other atomic function's documentation is also quite lack-luster and it would probably be beneficial to have a single overhaul of all atomic docs.
- For tests, there are not currently any "real" atomic tests via fuzzing or actually spawning threading to check if the atomic behave as they're supposed to so I didn't add any for `fence` - but lmk if that is something I should do.

Linked issue: #5196

### DIFF
--- a/mojo/stdlib/stdlib/os/atomic.mojo
+++ b/mojo/stdlib/stdlib/os/atomic.mojo
@@ -157,7 +157,7 @@ struct Consistency:
 
 @always_inline("nodebug")
 fn fence[
-    *, ordering: Consistency = Consistency.SEQUENTIAL, scope: StaticString = ""
+    ordering: Consistency = Consistency.SEQUENTIAL, *, scope: StaticString = ""
 ]():
     """Creates an atomic fence.
 
--- a/mojo/stdlib/stdlib/os/atomic.mojo
+++ b/mojo/stdlib/stdlib/os/atomic.mojo
@@ -151,6 +151,37 @@ struct Consistency:
 
 
 # ===-----------------------------------------------------------------------===#
+# fence
+# ===-----------------------------------------------------------------------===#
+
+
+@always_inline("nodebug")
+fn fence[
+    *, ordering: Consistency = Consistency.SEQUENTIAL, scope: StaticString = ""
+]():
+    """Creates an atomic fence.
+
+    Parameters:
+        ordering: The memory ordering for the fence.
+        scope: The memory synchronization scope.
+
+    Fences create synchronization between themselves and atomic operations or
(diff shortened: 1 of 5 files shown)


NathanSWard:

NathanSWard:

mojo/stdlib/stdlib/os/atomic.mojo
```diff
@@ -150,6 +150,37 @@ struct Consistency:
...
 
 
+# ===-----------------------------------------------------------------------===#
+# fence
+# ===-----------------------------------------------------------------------===#
+
+
+@always_inline("nodebug")
+fn fence[
+    *, ordering: Consistency = Consistency.SEQUENTIAL, scope: StaticString = ""
```
laszlokindrat: A matter of taste perhaps, but ordering can probably be positional so the calls are less verbose (e.g. `fence[Consistency.RELEASE, scope="workgroup"]()`.:
```suggestion
    ordering: Consistency = Consistency.SEQUENTIAL, *, scope: StaticString = ""
```

laszlokindrat (approved):
LGTM, thank you!

mojo/stdlib/stdlib/os/atomic.mojo
NathanSWard: great idea, done 👍

laszlokindrat:
!sync

laszlokindrat:
@NathanSWard Thanks again for the patch! Feel free to work on either of the things you suggested in the PR description as followups; documentation would probably be the most straightforward, but fuzz testing this would break some new ground, so I would also love to see that.

JoeLoser:
@NathanSWard since this merged internally already, do you mind opening a new PR to add a [changelog entry](https://github.com/modular/modular/blob/2f7379752aa3ee8f2e2ebf20f1ad837bb9350add/mojo/docs/changelog.md#standard-library-changes) so other users know about this?  Thanks!

NathanSWard:
@JoeLoser - done 👍 [add to changelog](https://github.com/modular/modular/pull/5234)

PR #5222, opened 2025-08-26
[stdlib] feat: Add `parts()` method to `Path` struct.

josiahls:
Adds the `parts` method. Allows for cross platform (windows vs unix) path splitting.

### DIFF
--- a/mojo/stdlib/test/pathlib/test_pathlib.mojo
+++ b/mojo/stdlib/test/pathlib/test_pathlib.mojo
@@ -203,6 +203,25 @@ def test_name():
     assert_equal("file", Path("file").name())
 
 
+def test_parts():
+    var path_to_file = Path("/path/to/file")
+    assert_equal(path_to_file.parts(), path_to_file.path.split("/"))
+
+    var rel_path = Path("path/to/file")
+    assert_equal(rel_path.parts(), rel_path.path.split("/"))
+
+    var path_no_slash = Path("path")
+    assert_equal(path_no_slash.parts(), path_no_slash.path.split("/"))
+
+    var path_with_tail_slash = Path("path/")
+    assert_equal(
+        path_with_tail_slash.parts(), path_with_tail_slash.path.split("/")
+    )
+
+    var root_path = Path("/")
+    assert_equal(root_path.parts(), root_path.path.split("/"))
+
+
 def main():
     test_cwd()
     test_path()
@@ -216,3 +235,4 @@ def main():
     test_home()
     test_stat()
     test_name()
+    test_parts()
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -131,6 +131,21 @@ language across multiple phases.
 
 ### Standard library changes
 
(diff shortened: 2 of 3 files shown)


josiahls:
@lsh I'm realizing there is a `from ..os import sep` alias already present.

I'm thinking of switching to use that instead of `DIR_SEPARATOR`

Ideally, it looks like: `alias DIR_SEPARATOR = "\\" if CompilationTarget.is_windows() else "/"` should just be deprecated in favor of os.sep.

There seems to be a decent amount of inconsistency where the "separator" should be pulled from when I look in `os.path` and `pathlib`. I'm guessing `os.sep` would probably be the correct alias to use for path ops in general.

mojo/stdlib/test/pathlib/test_pathlib.mojo
```diff
@@ -177,6 +177,43 @@ def test_stat():
     )
 
 
+def test_parts():
+    if CompilationTarget.is_windows():
```
lsh: Let's remove this branch for now since we don't have a runnable target to verify the test against

lsh (changes requested):

mojo/stdlib/test/pathlib/test_pathlib.mojo
josiahls: done

lsh:
!sync

mojo/stdlib/test/pathlib/test_pathlib.mojo
josiahls: I'll resolve this since you ack'd

gryznar:
> I'm realizing there is a `from ..os import sep` alias already present.
> 
> I'm thinking of switching to use that instead of `DIR_SEPARATOR`

@josiahls But why?`DIR_SEPARATOR` is clear while `sep` is very ambiguous. Also screaming case looks much proper here than just snake_case (compile time constant). I don't see value in copying Python 1:1 if it worsens API. I'd suggest rather to replace `sep` with `DIR_SEPARATOR`.

lsh:
Can you rebase this? :)

josiahls:
> Can you rebase this? :)

It looks rebased to me? The previous commit in the branch is 2f737975.

When I compare:
https://github.com/josiahls/modular/commits/josiahls/feat/add-parts-to-pathlib/
and 
https://github.com/modular/modular/commits/main/

My branch appears rebased correctly.

Usually if its out of date, github shows a button to rebase. Not seeing it on my end.

lsh:
Ah the nightly must not be out yet. Your other PR causes a conflict with this

josiahls:
@lsh 
I had another pr that got merged yesterday that also modifies pathlib so thats not surprising. Once the nightly is available I'll rebase / fix the conflict. I already know what its going to complain about.

lsh:
Seems like the nightly dropped :)

josiahls:
@lsh Should be good to go

lsh:
!sync

PR #5223, opened 2025-08-26
[stdlib] Fix `Atomic.compare_exchange` and add tests.

NathanSWard:
This fixes some current bugs and UB in the previously named `Atomic.compare_exchange_weak`.

1. Updates name to `compare_exchange`: The previous name indicated that this function was the `weak` variant of compare exchange which is was *not*. The compiled llvm instructions show the use of `cmpxchg` without the `weak`. We should add the weak variant in the future, but first we need to have the `pop.atomic.cmpxchg` MLIR operation support the `weak` attribute.
2. Correctly update the `expected` argument for floating-point atomics: Before, if an `Atomic`'s `DType` was a floating point, the inout parameter `expected` would not be correctly updated if the `cmpxchg` failed. 
3. Fix UB when writing the atomic's loaded value into the `expected` argument: Previously when the `cmpxchg` operation was called, if the success flag was `False`, the expected inout parameter would be written to by non-atomically reading the atomic pointer via `expected = value_addr[]`. This is UB as we are not allowed to read/write to the atomic pointer without an atomic operation. Instead, now we extract the loaded value from the result of the `cmpxchg` operation and store that in `expected`.

### DIFF
--- a/mojo/stdlib/stdlib/os/atomic.mojo
+++ b/mojo/stdlib/stdlib/os/atomic.mojo
@@ -393,8 +393,71 @@ struct Atomic[dtype: DType, *, scope: StaticString = ""]:
         """
         _ = self.fetch_sub(rhs)
 
-    @always_inline
-    fn compare_exchange_weak[
+    @staticmethod
+    @always_inline("nodebug")
+    fn compare_exchange[
+        *,
+        failure_ordering: Consistency = Consistency.SEQUENTIAL,
+        success_ordering: Consistency = Consistency.SEQUENTIAL,
+    ](
+        ptr: UnsafePointer[Scalar[dtype], mut=True, **_],
+        mut expected: Scalar[dtype],
+        desired: Scalar[dtype],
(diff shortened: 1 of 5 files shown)


JoeLoser:
Hey @NathanSWard, thanks for working on this!  I'll take a closer look tomorrow at the PR.

With regard to (1) — adding a `weak` attribute to `pop.atomic.cmpxchg`, can you file a GitHub feature request please?  Then, I can route it to the compiler team and we can refer to it in the code then as a TODO so future contributors don't wonder "was this an oversight?" or something else of the like.

NathanSWard:
@JoeLoser - Issue created [here](https://github.com/modular/modular/issues/5227).
(As well as [this issue](https://github.com/modular/modular/issues/5228) so we can correctly implement `Atomic.load/store`)

laszlokindrat (approved):
Cool, thanks!

laszlokindrat:
!sync

PR #5233, opened 2025-08-27
[stdlib] Move `StringSlice.from_utf8` factory method to be a constructor

martinvuyk:
Move `StringSlice.from_utf8` factory method to be a constructor


soraros:
Finally! If this works, we could also move `SIMD.from_bits`.

JoeLoser (approved):
Nice clean up, thank you! Please add a changelog entry though so other users know about this breaking change and can adjust their code.

After that, I'm happy to sync this in and land it!

JoeLoser:
!sync

Issue #4637, opened 2025-05-19
[BUG] `mblack` cannot format files using initialiser lists with keyword arguments


soraros:
### Bug description

As title.

### Steps to reproduce

Code:

```mojo
fn main():
  var l: List[Int] = {length=3, fill=1}
```

Command:

```bash
mojo format bug.mojo
```

Error:

```txt
error: cannot format ./main.mojo: Cannot parse: 13:28:   var l: List[Int] = {length=3, fill=1}

Oh no! 💥 💔 💥
1 file failed to reformat.
```

### System information

```txt
Package     Version               Build    Size       Kind   Source
max         25.4.0.dev2025051905  release  9.2 KiB    conda  max
max-core    25.4.0.dev2025051905  release  211.1 MiB  conda  max-core
max-python  25.4.0.dev2025051905  release  11.1 MiB   conda  max-python
```

martinvuyk:
not reproducible in `Mojo 25.6.0.dev2025082605` 🎉

soraros:
Closing as fixed then. Thanks heaps, @AmaranthineCodices!

PR #5235, opened 2025-08-28
[stdlib] Make `SIMD.from_bits` a constructor

martinvuyk:
Make `SIMD.from_bits` a constructor

### DIFF
--- a/max/kernels/src/nn/mha_amd.mojo
+++ b/max/kernels/src/nn/mha_amd.mojo
@@ -142,7 +142,7 @@ fn convert_f32_to_bf16[dtype: DType](x: SIMD, out res: SIMD[dtype, x.size]):
 
     @parameter
     if use_truncation:
-        res = __type_of(res).from_bits((x.to_bits() >> 16).cast[DType.uint16]())
+        res = __type_of(res)(from_bits=(x.to_bits() >> 16).cast[DType.uint16]())
     else:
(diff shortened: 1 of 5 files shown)


JoeLoser (approved):
LGTM, thanks! Please add a changelog entry though so other users know about this change.

JoeLoser (approved):

JoeLoser:
!sync

max/kernels/src/nn/mha_amd.mojo
```diff
@@ -142,7 +142,7 @@ fn convert_f32_to_bf16[dtype: DType](x: SIMD, out res: SIMD[dtype, x.size]):
 
     @parameter
     if use_truncation:
-        res = __type_of(res).from_bits((x.to_bits() >> 16).cast[DType.uint16]())
+        res = __type_of(res)(from_bits=(x.to_bits() >> 16).cast[DType.uint16]())
```
soraros, line 145: We could use `{from_bits}` if automatic formatting for them works these days.

martinvuyk: NICE! Though I think I'm going to open a new PR and change all uses I can find that are fixable like this

soraros: Sounds good. Be aware of #4637 though.

PR #5237, opened 2025-08-28
[stdlib] Add `assert_equal` overload for `SIMD` again

martinvuyk:
Add `assert_equal` overload for `SIMD` again. This should not have been deleted, because without it doing simple comparisons to `IntLiteral` fail (due to Mojo inferring both `Int` and the `SIMD` `DType`) and the tests become hugely verbose by having to use constructors.


laszlokindrat:
Thanks for the patch! That overload was removed intentionally, because it allowed implicit conversion routes that could lead to false negative tests. We are currently improving the architecture of the basic numeric types, so let's not reintroduce these complications in the meanwhile.

PR #5242, opened 2025-08-29
[stdlib] Refactor the `Roundable` trait

### LINKED ISSUE #5200: [Feature Request] [proposal] Rename `align_up` to `next_multiple_of` and `align_down` to `prev_multiple_of`
### Review Mojo's priorities

### What is your request?

Similarly to how we renamed `bit_ceil` and `bit_floor` for clarity into `next_power_of_two` and `prev_power_of_two`, I would like to rename `align_up` to `next_multiple_of` and `align_down` to `prev_multiple_of`.

When developing a function to round up to the nearest multiple of a compile time known value in  #4416, I realized that is what `align_up` does (though only at runtime currently). I think the name is not very descriptive except for people who already know what that kind of function does (not everybody deals with memory alignment on a day-to-day basis).

### Any other details?

CC: @laszlokindrat @JoeLoser

### LINKED ISSUE #5224: [Feature Request] [proposal] [stdlib] Flesh out the `Roundable` trait
### Review Mojo's priorities

### What is your request?

# Flesh out the `Roundable` trait

Our current trait looks like this:
```mojo
trait Roundable:
    fn __round__(self) -> Self:
        """Get a rounded value for the type.

        Returns:
            The rounded value.
        """
        ...

    fn __round__(self, ndigits: Int) -> Self:
        """Get a rounded value for the type.

        Args:
            ndigits: Number of digits after the decimal point.

        Returns:
            The rounded value.
        """
        ...
```
This is limited due to not taking rounding modes into account, and adding the
runtime cost of calculating the amount of digits to round after the comma.

After talking with @laszlokindrat and @soraros, I think we can do something like
this:
```mojo
@fieldwise_init
struct RoundMode(Copyable, Movable):
    alias Up = RoundMode(0)
    """Round towards positive infinity."""
    alias Down = RoundMode(1)
    """Round towards negative infinity."""
    alias ToEven = RoundMode(2)
    """Round towards the even number."""
    alias ToZero = RoundMode(3)
    """Round towards zero."""
    alias HalfUp = RoundMode(4)
    """When in the middle, round towards positive infinity."""
    alias HalfDown = RoundMode(5)
    """When in the middle, round towards negative infinity."""
    alias HalfToEven = RoundMode(6)
    """When in the middle, round towards the even number."""
    alias HalfToZero = RoundMode(7)
    """When in the middle, round towards zero."""

    var _value: Int

trait RoundableToMultiple:
    fn __round__[
        mode: RoundMode, to_multiple_of: Int, num_digits: Int
    ](self) -> Self:
        ...

# as a builtin
fn round[
    mode: RoundMode = RoundMode.HalfToEven,
    *,
    to_multiple_of: Int = 1,
    num_digits: Int = 0,
](v: InstanceOf[RoundableToMultiple]) -> __type_of(v):
    return number.__round__[mode, to_multiple_of, num_digits]()
```

This doesn't mean that each implementation must implement each and every one
of the possible combinations, but it opens the door to better customization and
optimizations given they are known at compile time.

This started with  #4416, and evolved in #5200

The optimizations and ergonomics that become possible because of this design are pretty powerful IMO:

```mojo
struct Int:
    fn __round__[
        mode: RoundMode, to_multiple_of: Int, num_digits: Int
    ](self) -> Self:
        alias magnitude = 10 ** -(num_digits) if num_digits < 0 else 1
        alias base = to_multiple_of * magnitude

        @parameter
        if base == 0:
            return 0
        elif base == 1:
            return self

        @parameter
        if RoundMode.Up:
            @parameter
            if base.is_power_of_two():
                alias offset = base - 1
                return (self + offset) & ~offset
            else:
                ceildiv(self, base) * base
        elif RoundMode.Down:
            @parameter
            if base.is_power_of_two():
                alias offset = base - 1
                return self & ~offset
            else:
                (self // base) * base

        var rem = self % base
        var criteria: Bool

        @parameter
        if RoundMode.HalfUp:
            criteria = self < base // 2
        elif RoundMode.HalfDown:
            criteria = self <= base // 2
        elif RoundMode.ToEven:
            criteria = (self - rem) % 2 == 0
        elif RoundMode.HalfToEven:
            criteria = self < base // 2 or (
                self == base // 2 and (self - rem) % 2 == 0
            )
        elif RoundMode.ToZero:
            criteria = self >= 0
        elif RoundMode.HalfToZero:
            criteria = self < base // 2 or (self == base // 2 and self >= 0)
        else:
            constrained[False, "Rounding mode not implemented"]()
            criteria = False
        return self - rem if criteria else self + rem
```

### Any other details?

There still is the matter of whether or not to add the dynamic version with
num_digits as an argument to the same trait or to create a new one. Since I
can't really think of a type that can be rounded with num_digits at compile time
but not at runtime I see no issue with having both in the same trait. As for
the signedness of num_digits meaning lhs or rhs of the decimal separator, I also
see no issue when explained properly in the docstrings.

martinvuyk:
Refactor the `Roundable` trait. Closes #5224, closes #5200, closes  #4416. Closes MOCO-333

This adds an optimized version that handles most of the rounding modes that are used. It will also be a nice replacement for functions that are very used like `align_down` by being a compile time version (that is very optimized for powers of two) e.g. 
```mojo
var vectorized_end = align_down(length, bool_mask_width)
```
can become (we could also use an alias that binds the parametrized round builtin)
```mojo
var vectorized_end = round[RoundMode.Down, to_multiple_of=bool_mask_width](length)
```
which is the equivalent of doing: 
```mojo
alias offset = bool_mask_width - 1
var vectorized_end = length & ~offset
```

CC: @laszlokindrat

### DIFF
--- a/mojo/stdlib/stdlib/builtin/math.mojo
+++ b/mojo/stdlib/stdlib/builtin/math.mojo
@@ -379,32 +379,123 @@ fn pow(base: SIMD, exp: Int) -> __type_of(base):
 # ===----------------------------------------------------------------------=== #
 
 
+@fieldwise_init
+struct RoundMode(Copyable, EqualityComparable, Movable):
+    """The `RoundMode` to use for the operation."""
+
+    alias Up = RoundMode(0)
+    """Round towards positive infinity."""
+    alias Down = RoundMode(1)
+    """Round towards negative infinity."""
+    alias ToEven = RoundMode(2)
+    """Round towards the even number."""
+    alias ToZero = RoundMode(3)
+    """Round towards zero."""
+    alias HalfUp = RoundMode(4)
+    """When in the middle, round towards positive infinity."""
+    alias HalfDown = RoundMode(5)
+    """When in the middle, round towards negative infinity."""
+    alias HalfToEven = RoundMode(6)
+    """When in the middle, round towards the even number."""
+    alias HalfToZero = RoundMode(7)
+    """When in the middle, round towards zero."""
+
+    var _value: UInt
+
+    @always_inline("nodebug")
+    fn __eq__(self, rhs: Self) -> Bool:
+        """Compare this value to the RHS value using EQ comparison.
+
+        Args:
+            rhs: The other value to compare against.
+
+        Returns:
+            True if this value is equal to the RHS value.
+        """
+        return self._value == rhs._value
+
+    @always_inline("nodebug")
+    fn __ne__(self, rhs: Self) -> Bool:
+        """Compare this value to the RHS value using NE comparison.
+
+        Args:
+            rhs: The other value to compare against.
+
+        Returns:
+            True if this value is not equal to the RHS value.
+        """
+        return self._value != rhs._value
+
+    @always_inline("nodebug")
+    fn __is__(self, rhs: Self) -> Bool:
+        """Compare this value to the RHS value using EQ comparison.
+
+        Args:
+            rhs: The other value to compare against.
+
+        Returns:
+            True if this value is equal to the RHS value.
+        """
+        return self == rhs
+
+    @always_inline("nodebug")
+    fn __isnot__(self, rhs: Self) -> Bool:
+        """Compare this value to the RHS value using NE comparison.
+
+        Args:
+            rhs: The other value to compare against.
+
+        Returns:
+            True if this value is not equal to the RHS value.
+        """
+        return self != rhs
+
+
 trait Roundable:
-    """
-    The `Roundable` trait describes a type that defines a rounding operation.
+    """The `Roundable` trait describes a type that defines a rounding operation.
 
     Types that conform to `Roundable` will work with the builtin `round`
     function. The round operation always returns the same type as the input.
 
-    For example:
+    Examples:
     ```mojo
     @fieldwise_init
     struct Complex(Roundable):
+        alias _RoundMultipleType = Float64
+        alias _RoundMultipleDefault = Float64(1)
+        alias _RoundModeDefault = RoundMode.HalfToEven
+
         var re: Float64
         var im: Float64
 
-        fn __round__(self) -> Self:
-            return Self(round(self.re), round(self.im))
+        fn __round__[
+            mode: RoundMode, to_multiple_of: Self._RoundMultipleType
+        ](self) -> Self:
+            return Self(
+                round[mode, to_multiple_of=to_multiple_of](self.re),
+                round[mode, to_multiple_of=to_multiple_of](self.im),
+            )
 
         fn __round__(self, ndigits: Int) -> Self:
             return Self(round(self.re, ndigits), round(self.im, ndigits))
     ```
     """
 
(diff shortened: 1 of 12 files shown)


mojo/stdlib/stdlib/builtin/math.mojo
```diff
@@ -401,11 +473,14 @@ trait Roundable:
     ```
     """
 
-    # TODO(MOCO-333): Reconsider the signature when we have parametric traits or
-    # associated types.
-    fn __round__(self) -> Self:
+    fn __round__[mode: RoundMode, to_multiple_of: UInt](self) -> Self:
```
soraros: Should this have type uint?

martinvuyk: I don't see a reason to support negative multipliers (If you have one in mind I'm interested in hearing about it though). I'm actually not sure if the bitwise trick for powers of two works for negative numbers for example, and it adds a lot more test area that I honestly don't want to deal with. I also think that if someone is dealing with signed multipliers they can add a sign at the end after rounding :man_shrugging:

soraros: I'm more talking about 0.1 etc.

martinvuyk: Ah, that is fixed by using round on a multiple of the base and ndigits e.g.
```mojo
# someone wants a multiple of 0.5 -> multiple_of = 5, num_digits = 1
offset = 10**num_digits
return round[mode, to_multiple_of=5](self * offset) / offset
```

But maybe we could receive a float, though it would complicate the implementation to extract the base and magnitude.
something like this would run at compile time
```mojo
value = to_multiple_of
base = 0
magnitude = 0
while value - value % 10 != 0:
  base = (base * 10) + value % 10
  value = (value - value % 10) * 10
  magnitude += 1
```
I'm not sure it's worth it

soraros: I'd expect the type of `to_multiple_of` to an associated type, e.g. float for floating point and unsigned for integers.

martinvuyk: I don't know how to express that in the case of `SIMD`, because using `Scalar[dtype]` as the associated type would mean accepting signed multiples. It could be solved by constraining that the value must be positive, but it would reduce the usable multiples for signed integers (which is arguably safer than the current implementation that doesn't check for overflow :thinking: ). WDYT?

laszlokindrat: I agree with @soraros here; asking for a multiple of a float should be possible and easy. For example, we might want to round a dollar amount to the nearest quarter. Restricting to positive numbers seems reasonable to me.

martinvuyk: Sadly I had to hard-code `Float64` due to #5276. But the scaffolding is there to make this generic once that is fixed :)

PR #5244, opened 2025-08-29
[stdlib] Improve `String` and `ArcPointer` ref-counting

NathanSWard:
This improves the atomic operations for both `ArcPointer` and `String`'s reference counting.
We can use explicit memory orderings in the operations along with a `fence` to improve both the performance and to more clearly state the intent of the atomic ops.

If you're curious, you can read more about the standard ref-count pointer type in the [boost docs](https://www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html#boost_atomic.usage_examples.example_reference_counters.implementation).


laszlokindrat (approved):
Cool, thanks! I would love to see some way to test/benchmark this, but it seems good enough to not block on it for now.

laszlokindrat:
!sync

PR #5247, opened 2025-08-29
[stdlib][mojo] Introduce `__delitem__` dunder function

christoph-schlumpf:
Introduces the `__delitem__` dunder function.

- Trait `DeletableItem`.
- Added a `delete()` convenience function to wrap `__delitem__()`.
This can be removed in the future if/when Mojo supports the `del list[idx]`, `del list[::]` and `del dict[key]` syntax.
- Implementation on `List` and `Dict`.
- Used `Indexer` for `__delitem__()` and `pop()` to align with `__getitem__()` in `List`.
- Performance optimization of slice deletion provided by @NathanSWard.
- tests added.

Examples:
```mojo
var l = [10, 20, 30]
l.delete(1)  # [10, 30]
l.delete(-1)  # [10]
```
```mojo
var l = [0, 1, 2, 3, 4, 5]
l.delete(Slice(1, 5))  # [0, 5]
```
```mojo
var l = [0, 1, 2, 3, 4, 5]
l.delete(Slice(None, None, 2))  # [1, 3, 5]
```
```mojo
var dict = {"mojo": "lang", "max": "engine"}
dict.delete("mojo")  # {"max": "engine"}
```

If approved I intend to add this functionality to all relevant collection structs.

Python equivalent: [`__delitem__`](https://docs.python.org/3/reference/datamodel.html#object.__delitem__)


mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -613,6 +619,49 @@ struct List[T: ExplicitlyCopyable & Movable, hint_trivial_type: Bool = False](
...
+            idx: The index of the item to delete.
+        """
+        var normalized_idx = normalize_index["List", assert_always=False](
+            idx, UInt(len(self))
+        )
+
+        (self._data + normalized_idx).destroy_pointee()
+        for j in range(normalized_idx + 1, self._len):
+            (self._data + j).move_pointee_into(self._data + j - 1)
+        self._len -= 1
```
NathanSWard: There is already the `List.pop` function, why not directly call that here instead of needing to re-rewrite the function?
```suggestion
    fn __delitem__(mut self, idx: Some[Indexer]) -> None:
        """Delete an item from the list at the given index.

        Args:
            idx: The index of the item to delete.
        """
        _ = self.pop(idx)
```
Or conversely have `pop` call `__delitem__`

mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -613,6 +619,49 @@ struct List[T: ExplicitlyCopyable & Movable, hint_trivial_type: Bool = False](
...
+            for j in range(end, self._len):
+                (self._data + j).move_pointee_into(self._data + j - r_len)
+            self._len -= r_len
+        else:
+            # non-contiguous range
+            for i in reversed(r):
+                (self._data + i).destroy_pointee()
+                for j in range(i + 1, self._len):
+                    (self._data + j).move_pointee_into(self._data + j - 1)
+                self._len -= 1
```
NathanSWard: I don't believe this is the most efficient implementation of this - as time complexity could potentially be `O(n^2)` if we're deleting every other element (`Slice(0, len(list), 2)`

Also the number of `__moveinit__` this has to call is quite large as you iterate though the range backwards and delete the items in reverse order moving the entire end of the list back 1 position every iteration.
For example, in your test with list `[0, 1, 2, 3, 4, 5].__delitems__(Slice(0, 5, 2))`, `__moveinit__` is called 6 times for this list.

There should be an `O(n)` algorithm for this...

mojo/stdlib/stdlib/collections/list.mojo
christoph-schlumpf: I did not reuse `List.pop` as it would be less performant because `pop` takes the item and returns it. `__delitem__` just destroys the item (no move).

Conversely `pop` can’t call `__delitem__` because `__delitem__` does not return the item by design.

But as there is some common functionality I will try to move that in a separate function to be reused by `pop` and `__ delitem__`.

NathanSWard: I'm not sure that it would be less performant necessarily (w/out benchmarks showing it).
Since if we immediately discard the value `_ = self.pop(idx)` hopefully the compiler would see that the move of the value from `pop` is discarded and therefore never actually performs the move.

mojo/stdlib/stdlib/collections/list.mojo
christoph-schlumpf: Agree. But I do not know an algorithm that is `O(n)` for this. I would be ready to implement a more performant algorithm. Not sure if this is a blocker for this PR or can be improved in a follow up PR.

mojo/stdlib/stdlib/collections/list.mojo
christoph-schlumpf: You're right. I did some benchmarks and there is no performance advantage of `_ = list.pop(i)` vs. `list__delitem__(i)`

Therefor `pop` is now reused inside `__delitem__`

Thank you!

mojo/stdlib/stdlib/collections/list.mojo
christoph-schlumpf: OK, I've refactored it to be `O[n]`.

Please have a look.

mojo/stdlib/stdlib/builtin/value.mojo
```diff
@@ -220,3 +220,24 @@ trait Iterator(Movable):
 
     fn __next__(mut self) -> Element:
         ...
+
+
+trait DeletableItem:
+    """The `DeletableItem` trait describes a type that can delete items from a indexed collection.
+    """
+
+    fn __delitem__(mut self, i: Int) -> None:
```
martinvuyk: This could probably be a `Some[Indexer]` from the start

martinvuyk (commented):
~~I get what you're getting at. Though I'm not sure why this needs to be a dunder method, AFAIK they are meant to never be used directly and have some more ergonomic way to be called. So this could simply be a method `.del_item()` or something.~~

~~Also a nit-pick: Since this is a new method, I would personally prefer if we snake_case it. Even if the `__getitem__` and `__setitem__` dunders still follow python names (which I'm not sure will remain the case once extensions land)~~

I just followed the link, I didn't know that existed in Python. 

Still without parametric traits we can't yet abstract what the key type is

mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -613,6 +619,57 @@ struct List[T: ExplicitlyCopyable & Movable, hint_trivial_type: Bool = False](
...
+                    slice_range_idx += 1
+                else:
+                    if write_idx != read_idx:
+                        (self._data + read_idx).move_pointee_into(
+                            self._data + write_idx
+                        )
+                    write_idx += 1
+            for i in range(write_idx, self._len):
+                (self._data + i).destroy_pointee()
+            self._len = write_idx
```
NathanSWard: This is `O(n)` - nice!
But this implementation is a bit bugged.
Particularly the last destroy loop
```mojo
for i in range(write_idx, self._len):
    (self._data + i).destroy_pointee()
```
I don't believe is necessary and in fact
- It can cause a double delete. 
- Will delete valid elements at the end of the list

also after running some benchmarks - this is a bit slow - and it seems to be due to the nested branches in the main loop which is likely throwing off the branch predictor. Ideally you should try and avoid large branches in the "hot loop" so to speak.

NathanSWard:
@christoph-schlumpf Here are some benchmark results:
```
| (algo/O(..)/len/slice)
| name                                | met (ms)               | iters   |
| ----------------------------------- | ---------------------- | ------- |
| chris/O( n )/100/[0:len:2]          | 0.000192727            | 1000000 |
| chris/O(n^2)/100/[0:len:2]          | 0.000440085            | 1000000 |
| natew/O( n )/100/[0:len:2]          | 0.00010956300000000001 | 1000000 |
| chris/O( n )/100/[0:len:10]         | 0.000195464            | 1000000 |
| chris/O(n^2)/100/[0:len:10]         | 0.00021408             | 1000000 |
| natew/O( n )/100/[0:len:10]         | 0.000102471            | 1000000 |
| chris/O( n )/10000/[0:len:2]        | 0.016150833610096314   | 75275   |
| chris/O(n^2)/10000/[0:len:2]        | 3.221991869918699      | 369     |
| natew/O( n )/10000/[0:len:2]        | 0.007649668481139481   | 156703  |
| chris/O( n )/10000/[0:len:10]       | 0.016474948286154575   | 72031   |
| chris/O(n^2)/10000/[0:len:10]       | 1.165139               | 1000    |
| natew/O( n )/10000/[0:len:10]       | 0.006726480264073433   | 179647  |
| chris/O( n )/10000/[0:len:100]      | 0.017174474681364106   | 69672   |
| chris/O(n^2)/10000/[0:len:100]      | 0.1333799375139478     | 8962    |
| natew/O( n )/10000/[0:len:100]      | 0.006456661194868029   | 185192  |
| chris/O( n )/10000/[0:len:5000]     | 0.016731841551601623   | 71771   |
| chris/O(n^2)/10000/[0:len:5000]     | 0.007324632256570003   | 164041  |
| natew/O( n )/10000/[0:len:5000]     | 0.005996933390029766   | 199895  |
| chris/O( n )/100000/[0:len:2]       | 0.16765892658509457    | 7192    |
| chris/O(n^2)/100000/[0:len:2]       | 327.5309090909091      | 11      |
| natew/O( n )/100000/[0:len:2]       | 0.0767550105654095     | 15617   |
| chris/O( n )/100000/[0:len:10]      | 0.16818688570626228    | 7122    |
| chris/O(n^2)/100000/[0:len:10]      | 118.64881818181817     | 11      |
| natew/O( n )/100000/[0:len:10]      | 0.06737542220220671    | 17764   |
| chris/O( n )/100000/[0:len:100]     | 0.17533191364370687    | 6809    |
| chris/O(n^2)/100000/[0:len:100]     | 13.101186813186814     | 91      |
| natew/O( n )/100000/[0:len:100]     | 0.06510600802515996    | 18442   |
| chris/O( n )/100000/[0:len:5000]    | 0.17097246397488944    | 7009    |
| chris/O(n^2)/100000/[0:len:5000]    | 0.31030556988138214    | 3878    |
| natew/O( n )/100000/[0:len:5000]    | 0.06037591791570265    | 19882   |
| chris/O( n )/1000000/[0:len:2]      | 1.7083097982708935     | 694     |
| natew/O( n )/1000000/[0:len:2]      | 0.7682967948717948     | 1560    |
| chris/O( n )/1000000/[0:len:10]     | 1.7309941520467835     | 684     |
| natew/O( n )/1000000/[0:len:10]     | 0.6674856175972927     | 1773    |
| chris/O( n )/1000000/[0:len:100]    | 1.7705                 | 668     |
| natew/O( n )/1000000/[0:len:100]    | 0.6520157608695653     | 1840    |
| chris/O( n )/1000000/[0:len:5000]   | 1.7436231884057969     | 690     |
| natew/O( n )/1000000/[0:len:5000]   | 0.6022678756476684     | 1930    |
| chris/O( n )/1000000/[0:len:100000] | 1.7419312865497076     | 684     |
| natew/O( n )/1000000/[0:len:100000] | 0.6000270135067534     | 1999    |
```

This compares your original quadratic version, linear version, and then a linear one I wrote.

I think we're able to improve this a bit by only iterating through the slice it's self (instead of through the entire list) and then "backfill" the list by taking into account the number of elements we have deleted up to that point. We can also do some nice stuff with the `__del__is_trivial` to avoid calling `destroy_pointee()` in the contiguous branch :)

Here is a godbolt link to the implementation: https://godbolt.org/z/6vo3c4feM

mojo/stdlib/stdlib/builtin/value.mojo
christoph-schlumpf: > I just followed the link, I didn't know that existed in Python. Still without parametric traits we can't yet abstract what the key type is

I think the dunder method can be replaced with the same „magic“ as `__getitem__` by the compiler team in the future:
```mojo
del list[i]
del list[1:10:2]
del dict[key]
```
This is how python calls `__delitem__`

I woul keep the name as is to be consistent with `__getitem__` and `__setitem__` for now.

With the new default implementation on traits feature I could provide a `del()` function that just calls `__delitem__` until we get the syntactic sugar:
```mojo
list.del(i)
list.del(Slice(1,10,2))
dict.del(key)
```

christoph-schlumpf: @martinvuyk
I added a `delete()` convenience function. Additionally I changed the code to use `Some[Indexer].

```mojo
list.delete(i)
list.delete(Slice(1,10,2))
dict.delete(key)
```

mojo/stdlib/stdlib/collections/list.mojo
christoph-schlumpf: Thank you for the advice. I looked at your code in the benchmark and incorporated it in this PR.
I'll resolve this conversation.

christoph-schlumpf:
> Here is a godbolt link to the implementation: https://godbolt.org/z/6vo3c4feM

Thank you! I integrated your implementation in the PR.

mojo/stdlib/stdlib/builtin/value.mojo
christoph-schlumpf: @martinvuyk 
As of today Mojo allows to use hard keywords as function names. So I decided to rename `delete()` to `del()` which brings the implementation even closer to the Python equivalent.

```mojo
list.del(i) # del list[i]
list.del(Slice(1,10,2)) # del list[1:10:2]
dict.del(key) # del dict[key]
```

christoph-schlumpf: Unfortunately I have to revert to `delete()` because `mojo format` breaks on the PR pipeline. Bug report: https://github.com/modular/modular/issues/5268

NathanSWard:
@christoph-schlumpf - Apologies on the delay - I'm jumping back into looking at this PR. 

I need to check with/ the team to see how we feel on adding a new `DeletableItem` trait and if we ever actually want to support the `del` syntax or not. 

In the meantime - I think we should consider having `__delitem__(slice)` be a normal member function on `List` and then we can consider special syntax support in the future. And for the PR we should remove `DeletableItem` for now and have that be a followup PR in the future IMO. 

We could just call it `pop_slice(slice: Slice)` perhaps - I'm open to other options too.

martinvuyk:
> We could just call it pop_slice(slice: Slice) perhaps - I'm open to other options too.

My nit-pick: It could just be `pop(slice: Slice)` since we have nice overload mechanics :)

christoph-schlumpf:
Here my proposal on how to proceed:

I will split this PR into three parts:

#### 1. Introduce `fn pop(slice: Slice) -> Self:`
I think an overload of `pop(i: Int)` makes sense as @martinvuyk proposed.

#### 2. Introduce `delete(idx)` and `delete(slice)`
Without a trait and without a `__delitem__()` dunder function. I suspect `delete(slice)` to have better performance than `pop(slice)` because no new List has to be built and returned (@NathanSWard: correct me if I‘m wrong).

#### 3.  Add a feature request for `__delitem__`
Write a feature request `issue` and post it on the Modular forum to be be discussed and to document the decision on it and its `del list[idx]` and `del list[x:y:z]` syntax. If the Modular team already takes a decision in  the meantime this might be obsolete.

If you agree I will close this PR and start with the first part.

NathanSWard:
@christoph-schlumpf - Sounds like a good start to me.
I would say doing #1 sounds great to me, for #2&3 I would file issues (instead of making PRs) for those that we can have discussion on there. 
I want to make sure to don't end up doing too much extra work that may not get used depending on what the team decides. 
Thanks again for this!

christoph-schlumpf:
> for #2&3 I would file issues

OK, I‘ll open issues. But I think being able to explicitly delete item(s) from a collection is a profound feature that any programming language should support with an expressive syntax. We will see. And thank you for your time spent reviewing this PR.

christoph-schlumpf:
HI @NathanSWard ,

I've opened a new PR for part 1: https://github.com/modular/modular/pull/5308

I will close this PR now.

PR #5240, opened 2025-08-28
[stdlib] Deprecate functions with flatcase naming and add snake_case counterparts in `sys.info`

gryznar:
Closes https://github.com/modular/modular/issues/457

PR #5248, opened 2025-08-29
[stdlib] Fix changelog for renaming  in `sys.info`

gryznar:
I've messed up changelog in https://github.com/modular/modular/pull/5240 (duplicated names from overloads). This patch fixes this. Now names are also sorted

cc @laszlokindrat


laszlokindrat (approved):

laszlokindrat:
!sync

PR #5249, opened 2025-08-29
[stdlib] Add tests for struct inheriting triviality (T.__*__is_trivial).

rd4com:
An type with an parameter T is instantiated.
It sets its own triviality based on T.

For example,
if T is Float64,
Self.__copyinit__is_trivial is set to T.__copyinit__is_trivial.

Then the type goes trough an triviality tester,
to determine if it works as expected.

---

Possible auto-optimizations related to inherited triviality:
- `InlineArray[32, InlineArray[32 Int]]` -> Fully trivial
- `Variant[Int64, Bool]` -> Fully trivial
- `List[Optional[Int]]` -> No hint_trivial_type

### DIFF
--- a/mojo/stdlib/test/builtin/test_value.mojo
+++ b/mojo/stdlib/test/builtin/test_value.mojo
@@ -25,9 +25,9 @@ alias EVENT_COPY = 0b1000  # 8
 alias EVENT_MOVE = 0b10000  # 16
 
 
-struct ConditionalTriviality[O: MutableOrigin, //, T: Movable & Copyable](
-    Copyable, Movable
-):
+struct ConditionalTriviality[
+    O: MutableOrigin, //, T: Movable & ExplicitlyCopyable
+](ExplicitlyCopyable, Movable):
     var events: Pointer[List[Int], O]
 
     fn add_event(mut self, event: Int):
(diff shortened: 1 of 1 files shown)


mojo/stdlib/test/builtin/test_value.mojo
```diff
@@ -63,6 +63,12 @@ struct ConditionalTriviality[O: MutableOrigin, //, T: Movable & Copyable](
             self.add_event(EVENT_MOVE)
 
 
+struct StructInheritTriviality[T: Movable & Copyable](Movable & Copyable):
```
PeimingLiu: LGTM in general, but maybe use `ExplicitlyCopyable` here? Not sure whether the mojo nightly has catch up the change (if not, maybe better to wait for a few days?). 

Not a blocker, also depends on @ConnorGray's plan on updating `Copyable`.

rd4com: Thanks @PeimingLiu , the nightly just got released with the patch!
Here is an commit to integrate the review.

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

PR #5252, opened 2025-08-30
[stdlib] Replace uses of `String.__add__` with `write`

martinvuyk:
Replace uses of `String.__add__` with `write`

### DIFF
--- a/mojo/stdlib/stdlib/algorithm/functional.mojo
+++ b/mojo/stdlib/stdlib/algorithm/functional.mojo
@@ -1370,10 +1370,9 @@ fn elementwise[
         )
 
     # Intern the kind string as a static string so we don't allocate.
-    alias kind = get_static_string[
-        "elementwise",
-        ("(" + _trace_description + ")" if _trace_description else ""),
-    ]()
+    alias d = _trace_description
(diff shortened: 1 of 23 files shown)


lsh:
!sync

lsh:
!sync

lsh:
This fails on GPU but I haven't quite isolated why yet. I'll hopefully take a closer look this week

lsh:
Alternatively you can undo the changes in the `gpu` folder and I can land this without that probably

martinvuyk:
@lsh I'll remove the changes in the gpu directory :+1:

lsh:
!sync

soraros:
Sorry for being late to the party. I thought `__add__` calls `write` underneath?

martinvuyk:
> Sorry for being late to the party. I thought `__add__` calls `write` underneath?

@soraros several concatenated calls to add are repeatedly allocating a new buffer for each operation, whereas write calculates the full size and reserves only once. `iadd` is the same as using `String.write()` for only one item at a time (if the full size was previously reserved), whenever you use several it is best to just wrap it all in a single constructor call to malloc only once

PR #5255, opened 2025-08-30
[stdlib] Add UTF-16 decoding

martinvuyk:
Add UTF-16 decoding. Split off from  #3239

### DIFF
--- a/Mojo/stdlib/std/collections/string/_utf16.mojo
+++ b/Mojo/stdlib/std/collections/string/_utf16.mojo
@@ -1,0 +1,108 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+"""Implement UTF-16 utils."""
+
+from std.bit.mask import splat
+from std.sys.info import bit_width_of
+from std.sys.intrinsics import likely, unlikely
+from std.collections.string._unicode import BIGGEST_UNICODE_CODEPOINT
+
+
+def _decode_utf16[
+    dtype: DType,
+    //,
+    *,
+    strict: Bool = True,
+    replace: Codepoint = Codepoint.ord("�"),
+](*, from_utf16: Span[mut=False, Scalar[dtype], ...]) raises -> String:
+    comptime assert dtype.is_integral(), "The dtype must be integral"
+    comptime assert bit_width_of[dtype]() > 8, (
+        "Decoding UTF-16 from a buffer of bytes is ambiguous "
+        "bitcast it to UInt16 or use the codepoints decoder if they are "
+        "independent codepoints"
+    )
+    comptime replace_str = String(replace)
+    comptime replace_length = UInt(replace_str.byte_length())
+
+    # TODO: this could use reduce_sum with a function that calculates the utf8
+    # length from the utf16 codepoints. Performance needs to be measured
+    var utf16_len = UInt(len(from_utf16))
+    var result = String(capacity_bytes=2 * Int(utf16_len))
+    var utf16_ptr = from_utf16.unsafe_ptr()
+    var utf16_idx = UInt(0)
+    var utf8_offset = UInt(0)
+
+    while utf16_idx < utf16_len:
+        comptime cont_bytes = 0b1000_0000
+        var c_og = utf16_ptr[unsafe_offset=utf16_idx]
+        var c = c_og.cast[DType.uint32]()
+        var ptr = result.unsafe_ptr_mut().unsafe_offset(Int(utf8_offset))
+        var num_bytes: UInt
+        # high surrogate marker 0xD800 == 0b11_0110 << 10
+        var is_high_surrogate = (c >> 10).eq(0b11_0110)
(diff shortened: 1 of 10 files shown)


martinvuyk:
@NathanSWard given https://github.com/modular/modular/commit/0fac848368f0e82e58cc4207670ed99cac12f6bb I think you might be interested in the trick I used here to have similar constructors to Python's API (plus being able to specify the replacement character seems nice to me)

Mojo/stdlib/std/collections/string/_utf16.mojo
```diff
@@ -0,0 +1,108 @@
...
+        "bitcast it to UInt16 or use the codepoints decoder if they are "
+        "independent codepoints"
+    )
+    comptime replace_str = String(replace)
+    comptime replace_length = UInt(replace_str.byte_length())
+
+    # TODO: this could use reduce_sum with a function that calculates the utf8
+    # length from the utf16 codepoints. Performance needs to be measured
+    var utf16_len = UInt(len(from_utf16))
+    var result = String(capacity_bytes=2 * Int(utf16_len))
```
alisterburt, line 41: the assumption of two output bytes per UTF-16 unit is incorrect
`U+0800` to `U+FFFF` (CJK, Hangul, Devanagari, Thai…) are delivered as 3 bytes per unit so this can overflow

```suggestion
    var result = String(capacity_bytes=3 * Int(utf16_len))
```

This wasn't caught in the tests because of the stack vs. heap allocation, only short sequences were tested which go to the fixed size stack allocated buffer and don't overflow

Mojo/stdlib/std/collections/string/_utf16.mojo
```diff
@@ -0,0 +1,108 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+"""Implement UTF-16 utils."""
+
+from std.bit.mask import splat
+from std.sys.info import bit_width_of
+from std.sys.intrinsics import likely, unlikely
+from std.collections.string._unicode import BIGGEST_UNICODE_CODEPOINT
```
alisterburt, line 19: unused import

Mojo/stdlib/std/collections/string/_utf16.mojo
```diff
@@ -0,0 +1,108 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+"""Implement UTF-16 utils."""
+
+from std.bit.mask import splat
```
alisterburt, line 16: unused import

Mojo/stdlib/std/collections/string/_utf16.mojo
```diff
@@ -0,0 +1,108 @@
...
+            # get lower 10 bits
+            comptime low_10b = UInt32(0b0011_1111_1111)
+            var has_more = utf16_idx < utf16_len - 1
+            var c2 = utf16_ptr[unsafe_offset=utf16_idx + UInt(has_more)].cast[
+                DType.uint32
+            ]()
+            # low surrogate marker 0xDC00 == 0b0011_0111 << 10
+            var is_low_surrogate = (c2 >> 10).eq(0b0011_0111)
+            is_full_surrogate = is_high_surrogate & is_low_surrogate
+            var any_pair = is_high_surrogate | is_low_surrogate
```
alisterburt, line 85: this isn't quite right
- `[4E00, DC00]` c gets flagged as invalid because c2 is a low surrogate
- `[DC00, 0041]` low surrogate followed by non-low surrogate -> DC00 flows unchecked into `Codepoint(unsafe_unchecked_codepoint=)`. Errors but only with asserts on

something like below should fix:
```mojo
var next_is_low_surrogate = (c2 >> 10).eq(0b0011_0111)
is_full_surrogate = is_high_surrogate & next_is_low_surrogate
var is_surrogate = is_high_surrogate | (c >> 10).eq(0b0011_0111)

if unlikely(Bool(is_surrogate & ~is_full_surrogate)):
```

your low surrogate tests only pass because they are one element buffers, c2 == c for the comparison so it doesn't hit the failure - weird coincidence

Mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -404,6 +405,77 @@ struct String(
...
+            An exception is raised if the provided buffer byte values do not
+            form valid UTF-16 encoded codepoints.
+        """
+        var result = _decode_utf16[strict=True](from_utf16=from_utf16)
+        # NOTE: the decoder should've already raised/replaced when invalid data
+        # was encountered
+        debug_assert(
+            _is_valid_utf8(result.as_bytes()),
+            "Buffer is not valid UTF-16.",
+        )
```
alisterburt, line 436:
```suggestion
        debug_assert(
            _is_valid_utf8(result.as_bytes()),
            "UTF-16 decoder produced invalid UTF-8.",
        )
```

also this is duplicated across both constructors, consider adding to the decoder itself?

Mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -404,6 +405,77 @@ struct String(
...
+            var result = _decode_utf16[strict=False, replace=replace](
+                from_utf16=from_utf16
+            )
+            # NOTE: the decoder should've already raised/replaced when invalid
+            # data was encountered
+            debug_assert(
+                _is_valid_utf8(result.as_bytes()),
+                "Buffer is not valid UTF-16.",
+            )
+            self = result^
```
alisterburt, line 475: same point as above

alisterburt (commented):
hey @martinvuyk - thanks for parting this out, rebased for you

nits:
- no spec reference in file https://www.rfc-editor.org/rfc/rfc2781.txt
- 2025 copyright -> 2026 copyright
- assert_true/false imports unused

alisterburt (changes requested):

PR #5256, opened 2025-08-30
[Stdlib] Add missing comparison operators for StringSlice with String

msaelices:
## Summary
- Implement missing comparison operators (`<`, `<=`, `>`, `>=`) for StringSlice compared with String
- Add comprehensive test coverage for all comparison operations including edge cases
- Enables natural comparison syntax like `string_slice < "abc"`

## Examples

It's useful for comparison with string or string literals in functions that receives string slices coming from String to avoid allocations:

```mojo
fn is_digit(StringSlice s) -> Int:
    return s >= "0" and s <= "9"
```


mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -968,6 +968,55 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
+    fn __le__(self, rhs: String) -> Bool:
+        """Define whether this String slice is less than or equal to the RHS.
+
+        Args:
+            rhs: The other String to compare against.
+
+        Returns:
+            True if this String slice is less than or equal to the RHS String.
+        """
+        return not (rhs.as_string_slice() < self)
```
laszlokindrat: Here and below: instead of implementing these in terms of `<`, could we just implement in terms of the corresponding implementation that takes StringSlice?
```suggestion
        return self <= rhs.as_string_slice()
```

msaelices: Sure. It's not as easy as it seems as there were a couple of missing comparisons operators to implement. 

Fixed here: https://github.com/msaelices/mojo/commit/e85ccd76b466e82c29b98fa27ac24cca11a3ef33

msaelices: @laszlokindrat could you PTAL?

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

PR #5257, opened 2025-08-31
[max] Refactor Value.from_mlir to use a self-registering factory

YichengDWu:
This PR refactors the `Value.from_mlir` class method in `max/graph/value.py`. The previous implementation relied on a growing `if/isinstance` chain, which has been replaced with a robust, self-registering factory pattern.

###  Changes Made

   - The Value base class now leverages `__init_subclass__` to act as a self-registering factory.
   - The factory automatically registers new Value subclasses by introspecting the generic alias in their base class definition (e.g., `Value[mo.TensorType]`) to infer the associated MLIR type.
   - This change completely eliminates the conditional block for creating Value instances from MLIR types.

### Benefits of this Change

  This refactoring improves the software design in several key areas:

   - Decoupling: The `Value` base class is now fully decoupled from its concrete subclasses. It no longer needs explicit knowledge of them.
   - Open/Closed Principle: The system is now open for extension but closed for modification. New `Value` types can be added without any changes to the base class, significantly improving maintainability and extensibility.

### DIFF
--- a/max/graph/value.py
+++ b/max/graph/value.py
@@ -22,7 +22,8 @@ from typing import (
     Protocol,
     TypeVar,
     Union,
-    cast,
+    get_args,
+    get_origin,
     runtime_checkable,
 )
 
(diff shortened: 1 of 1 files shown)


YichengDWu:

JoeLoser:
@bethebunny do you mind taking a look at this, please? Thanks!

bethebunny:
Hi @YichengDWu, thank you so much for the contribution! Impressive use of Python, I like this pattern in general :) That said

> Open/Closed Principle: The system is now open for extension but closed for modification. New Value types can be added without any changes to the base class, significantly improving maintainability and extensibility.

We have no plans to extend the system here, by contrast we want to simplify the type system and make opaque types and tensor values use the same mechanism.

Your pattern is a great application of the principle of "Don't Repeat Yourself", but in this case I would rather defer to "Keep it Simple" / "Yain't Gonna Need It": we won't add more if/else clauses in the future, and with so few cases I prefer the explicitness if the if/else checks.

bethebunny:
I'm going to close this PR, but feel free to continue discussion or re-open if you think there's a better explicit pattern.

PR #5258, opened 2025-08-31
[stdlib] Add unicode codepoints decoding (UTF-32 as well)

martinvuyk:
Add unicode codepoints decoding (UTF-32 as well). Split off from #3239

### DIFF
--- a/Mojo/stdlib/std/collections/string/string.mojo
+++ b/Mojo/stdlib/std/collections/string/string.mojo
@@ -26,6 +26,7 @@ from std.collections.string.string_span import (
     _to_string_list,
     _unsafe_strlen,
 )
+from std.collections.string._unicode import _decode_codepoints
 from std.builtin.builtin_slice import ContiguousSlice
 from std.hashlib.hasher import Hasher
 from std.reflection import call_location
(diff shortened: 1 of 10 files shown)


alisterburt (commented):
Hi @martinvuyk ! I rebased onto main updating a few things with respect to API changes.

few things noticed in AI assisted review:
- when `errors="replace"` and a buffer is longer than one codepoint the output is corrupted e.g. `[0x41, 0xD800, 0x42 ` should give `41 EF BF BD 42` but gives `EF BF BD 00 42`. CI didn't catch as only tested single codepoint buffers
- `dtype.is_integral()` allows negative signed ints which lead to a `String` with invalid UTF-8 when asserts are off
- for the above, how about a where clause?
- UTF-32 arrives as bytes and impl. has no endianness/byte order handling
- `BIGGEST_UNICODE_CODEPOINT` should have a leading underscore to match conventions in file
- test_codepoint_parsing is a 41s test on my machine -> sample + boundary cases would be better
- `range` is non-inclusive so `UInt16.MAX` never gets tested
- `result.unsafe_ptr_mut()` is called once per codepoint in the loop, can it be hoisted?
- weird layering with codepoint importing from `_unicode.mojo` which uses `Codepoint`... only compiles because `_unicode.mojo` picks `Codepoint` up from prelude. Could be fixed by moving `BIGGEST_UNICODE_CODEPOINT` into `_unicode.mojo`

alisterburt (changes requested):

Mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -404,6 +405,149 @@ struct String(
...
+            from_codepoints=from_codepoints
+        )
+        # NOTE: the decoder should've already raised/replaced when invalid data
+        # was encountered
+        debug_assert(
+            _is_valid_utf8(result.as_bytes()),
+            "Buffer is not valid Unicode.",
+        )
+
+        self = result^
```
alisterburt, line 445: please replace message with "decoder produced invalid UTF-8"

consider moving into decoder to avoid duplication

Mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -404,6 +405,149 @@ struct String(
...
+            )
+            # NOTE: the decoder should've already raised/replaced when invalid
+            # data was encountered
+            debug_assert(
+                _is_valid_utf8(result.as_bytes()),
+                "Buffer is not valid Unicode.",
+            )
+            self = result^
+        except:
+            abort("This should never happen")
```
alisterburt, line 486: duplication and wrong error message

PR #5261, opened 2025-09-01
[stdlib] Improve performance in hashing tests and benchmark setup

martinvuyk:
Improve performance in hashing tests and benchmark setup

### DIFF
--- a/mojo/stdlib/benchmarks/hashlib/bench_hash.mojo
+++ b/mojo/stdlib/benchmarks/hashlib/bench_hash.mojo
@@ -18,7 +18,7 @@ from std.hashlib._fnv1a import Fnv1a
 from std.benchmark import Bench, BenchConfig, Bencher, BenchId, keep
 
 # Source: https://www.101languages.net/arabic/most-common-arabic-words/
-comptime words_ar: String = """
+comptime words_ar = """
 لا, من, هذا, أن, في, أنا, على, ما, هل,
  يا, و, لقد, ذلك, ماذا, أنت, هنا, لم, إلى, نعم, كان, هو, ان, هذه, هناك, عن, فى, كل, ليس, فقط, كنت, الآن, يجب, انا,
  لك, مع, شيء, لكن, لن, الذي, حسنا, كيف, سوف, هيا, نحن, إنه, ـ, أجل, لماذا, إذا, عندما, انه, كذلك, لي, الى, بعد, انت,
@@ -84,7 +84,7 @@ comptime words_ar: String = """
 """
 
 # Source: https://github.com/tkaitchuck/ahash/blob/7d5c661a74b12d5bc5448b0b83fdb429190db1a3/tests/map_tests.rs#L9
-comptime words_en: String = """
+comptime words_en = """
     a, ability, able, about, above, accept, according, account, across, act, action,
     activity, actually, add, address, administration, admit, adult, affect, after,
     again, against, age, agency, agent, ago, agree, agreement, ahead, air, all,
@@ -184,7 +184,7 @@ comptime words_en: String = """
     yet, you, young, your, yourself"""
 
 # Source: https://www.101languages.net/hebrew/most-common-hebrew-words/
-comptime words_he: String = """
+comptime words_he = """
 לא , את , אני , זה , אתה ,
  מה , הוא , לי, על, כן, לך, של, יש , בסדר , אבל , כל , שלי , טוב , עם, היא, אם, רוצה,
  שלך, היה, אנחנו, הם, אותך, יודע, אז, רק, אותו, יכול, אותי, יותר, הזה, אל, כאן, או,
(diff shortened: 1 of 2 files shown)


martinvuyk:
@mzaks I changed the `gen_word_pairs` function to split words by `words.split(",")` instead of `words.split(", ")` that was incorrectly splitting in the case of a word ending with `,` and then a newline character. But now the fill factor test fails. Should I just revert that change, should I delete some very short words from the tests, or what should I do?

mzaks:
> @mzaks I changed the `gen_word_pairs` function to split words by `words.split(",")` instead of `words.split(", ")` that was incorrectly splitting in the case of a word ending with `,` and then a newline character. But now the fill factor test fails. Should I just revert that change, should I delete some very short words from the tests, or what should I do?

Hi @martinvuyk,
Yeah I see what is happening. The `test_fill_factor` for DJBX33a is extremely flaky, which shows the issues with the hash function itself. It is non deterministic, as it gives different values on macOS and Linux and now as you fixed the splitting of words it returns different results again. Generally the simplest thing to do is either comment out or delete the test completely. It does not provide any value it just shows how terrible the function is and that nobody should use it. IMHO we should actually delete the DJBX33a hasher all together, I kept it for backwards compatibility reasons, if folks relied on it, but actually as it is non deterministic nobody could rely on it anyways, so I think it is time to remove it all together. In your Pr I would suggest just to comment out the failing test and in the next PR we can delete the Hasher.

martinvuyk:
@mzaks wow I didn't know that hasher is non-deterministic, that sounds bad. I'll comment out the tests and point to a github issue I'll create referencing your comment :+1:

JoeLoser:
> > @mzaks I changed the `gen_word_pairs` function to split words by `words.split(",")` instead of `words.split(", ")` that was incorrectly splitting in the case of a word ending with `,` and then a newline character. But now the fill factor test fails. Should I just revert that change, should I delete some very short words from the tests, or what should I do?
> 
> Hi @martinvuyk, Yeah I see what is happening. The `test_fill_factor` for DJBX33a is extremely flaky, which shows the issues with the hash function itself. It is non deterministic, as it gives different values on macOS and Linux and now as you fixed the splitting of words it returns different results again. Generally the simplest thing to do is either comment out or delete the test completely. It does not provide any value it just shows how terrible the function is and that nobody should use it. IMHO we should actually delete the DJBX33a hasher all together, I kept it for backwards compatibility reasons, if folks relied on it, but actually as it is non deterministic nobody could rely on it anyways, so I think it is time to remove it all together. In your Pr I would suggest just to comment out the failing test and in the next PR we can delete the Hasher.

Thanks for some color on this matter.  Please feel free to open a PR to remove DJBX33a hasher completely now that we have a better default.  I don't think we're relying on it internally at all now given the improved default hasher.

JoeLoser:
> > > @mzaks I changed the `gen_word_pairs` function to split words by `words.split(",")` instead of `words.split(", ")` that was incorrectly splitting in the case of a word ending with `,` and then a newline character. But now the fill factor test fails. Should I just revert that change, should I delete some very short words from the tests, or what should I do?
> > 
> > 
> > Hi @martinvuyk, Yeah I see what is happening. The `test_fill_factor` for DJBX33a is extremely flaky, which shows the issues with the hash function itself. It is non deterministic, as it gives different values on macOS and Linux and now as you fixed the splitting of words it returns different results again. Generally the simplest thing to do is either comment out or delete the test completely. It does not provide any value it just shows how terrible the function is and that nobody should use it. IMHO we should actually delete the DJBX33a hasher all together, I kept it for backwards compatibility reasons, if folks relied on it, but actually as it is non deterministic nobody could rely on it anyways, so I think it is time to remove it all together. In your Pr I would suggest just to comment out the failing test and in the next PR we can delete the Hasher.
> 
> Thanks for some color on this matter. Please feel free to open a PR to remove DJBX33a hasher completely now that we have a better default. I don't think we're relying on it internally at all now given the improved default hasher.

https://github.com/modular/modular/pull/5275 just landed internally, so feel free to rebase with the next nightly FYI.

mojo/stdlib/test/hashlib/test_djbx33a.mojo
```diff
@@ -83,47 +83,49 @@ def test_trailing_zeros():
...
-    assert_fill_factor["PL", DJBX33A](words, len(words) // 2, 0.485)
-    assert_fill_factor["PL", DJBX33A](words, len(words) // 4, 0.971)
-    assert_fill_factor["PL", DJBX33A](words, len(words) // 14, 0.5)
-
-    words = gen_word_pairs[words_ru]()
-    assert_fill_factor["RU", DJBX33A](words, len(words), 0.607)
-    assert_fill_factor["RU", DJBX33A](words, len(words) // 2, 0.038)
-    assert_fill_factor["RU", DJBX33A](words, len(words) // 4, 0.066)
-    assert_fill_factor["RU", DJBX33A](words, len(words) // 14, 1.0)
+    # NOTE: This was commented out due to them failing, see issue #5270
```
JoeLoser: #5270 is fixed now, so you can rebase in the next nightly 👍

JoeLoser (commented):
For my own understanding, is there any compile time perf differences in annotating these long word lists as `String` vs. the new type (`StringLiteral` since they are marked with `alias` and so won't materialize to a `String`)?

martinvuyk:
> For my own understanding, is there any compile time perf differences in annotating these long word lists as `String` vs. the new type (`StringLiteral` since they are marked with `alias` and so won't materialize to a `String`)?

It shouldn't make any performance difference on the materialization side since `String` just initializes with the given `StringLiteral` ptr and length. It is one less function call, but it is probably negligible

PR #5264, opened 2025-09-01
[stdlib] Make `File.read() -> String` read unicode codepoints and not potentially invalid UTF-8

### LINKED ISSUE #5262: [Feature Request] [proposal] `File.read() -> String` should read codepoints not bytes
### Review Mojo's priorities

### What is your request?

When doing PR  #5253 I realized we are directly reading bytes without validating that they are valid UTF-8. We can validate the read and raise an error like I am doing in that PR, but I would like us to reimplement the behavior.

I think this method should read actual unicode codepoints and not bytes. There is also [precedent](https://docs.python.org/3/tutorial/inputoutput.html#methods-of-file-objects) from Python which does the same.

We'll have to reimplement the read logic at some point anyway because it currently creates a `List` (when using `read_bytes`) and then it copies the data again into a `String`:
```mojo
var list = self.read_bytes(size)
return String(bytes=list)
```

### Any other details?

_No response_

martinvuyk:
Make `File.read() -> String` read unicode codepoints and not potentially invalid UTF-8. Closes #5262


martinvuyk:
Hi @JoeLoser sorry to ping you, but I don't want to go through the work of adapting this to the new io functions unless you tell me this will be merged at some point

PR #5265, opened 2025-09-01
[stdlib] Add support for 1080Ti

gabrieldemarmiesse:
There is still one test failing here: https://github.com/modular/modular/blob/main/mojo/stdlib/test/asyncrt/cuda.mojo#L107

because it assumes sm80+ but I'll handle it in a separate PR.


BradLarson (approved):
Thanks for adding this! There's a lower bound on DeviceContext support we may need to adjust, but I can do that internally. I may also follow on with support for `sm_60` to also enable the P100, which some people have requested.

BradLarson:
!sync

Issue #5266, opened 2025-09-01
[Feature Request] Allow the pre-commit to be run at the root of the repo without activating envs


gabrieldemarmiesse:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Currently to make the pre-commit work, we have to activate the pixi env in the mojo or max directory, this is because we need to call mojo format.

I would like to just be able to do 
```
pixi x pre-commit install
```
and then  run `git add && git commit` and it would just work, no other prerequisites.

There are multiple ways we can do this. First, we could call the formatting command in bazel to avoid depending on mojo, that would be nice. And also should we assume that python3 is always available on the dev system? This is because some steps of the pre-commit rely on python. I often rely on pixi or uv to get a working python interpreter, so maybe we shouldn't assume that and call `pixi run python ...` in some form.

### What is your motivation for this change?

I'd like to setup the pre-commit without having to worry if I activated a virtual environment or not, and also which environment to use since there are multiple pixi.toml in this repo. We should also update the contribution instructions since they advise `pip install pre-commit && pre-commit install` which assumes that a virtualenv is activated and pip is installed. `pixi x pre-commit install` should work all the time from any setup as long as pixi is installed.

gabrieldemarmiesse:
@Ahajha since you asked me to open this issue

Ahajha:
Thanks! There's a lot of minor (and a few major) issues with the build system in its current state. Hoping to chip away at those ~soonish, but in the meantime happy to review a PR if you want to take a stab at it. I think standardizing on `./bazelw run format` is ideal IMO.

I believe rules_python in its current state requires a system python in order to bootstrap itself, but there was a mention internally that there might be a flag we can flip on to eliminate that dependency. Then the bazel config wouldn't need anything on the system.

gabrieldemarmiesse:
While working on this I noticed that 
```
pixi run --manifest-path ./mojo mojo doc --diagnose-missing-doc-strings --validate-doc-strings -o /dev/null ./mojo/stdlib/stdlib
```
gives many errors, is this supposed to give no errors? We get things like:
```
Included from /projects/open_source/modular/./mojo/stdlib/stdlib/gpu/host/__init__.mojo:1:
[...]

PR #5272, opened 2025-09-03
[stdlib] Make the pre-commit work without any dependency

gabrieldemarmiesse:
Fix part for https://github.com/modular/modular/issues/5266

It allows users to just do `pixi x pre-commit install` and `pixi x pre-commit run --all-files` without any other requirements than pixi.

The only tool that was removed from the pre-commit is the docstring check because it wasn't working correctly and catching the errors. See https://github.com/modular/modular/issues/5266#issuecomment-3248626071 . We'll see if we want to re-enable it later but it would require fixing the 50+ errors/warnings we get when running the docs command.


gabrieldemarmiesse:
@Ahajha I think you're the right person for this.

gryznar:
What is the "x" stands for? I cannot see it in Pixi docs at first glance. Shortcut for "execute"? If yes, we should probably suggest full command name

Ahajha:
Running `pixi x --help` locally tells me it's short for "exec"

gryznar:
@Ahajha thanks for checking :) IMO still better is to reffer to full command name rather than to shortcuts in official docs

.pre-commit-config.yaml
```diff
@@ -14,29 +14,13 @@
...
         language: system
-        files: '^(mojo|examples/mojo).*\.(mojo|🔥|py)$'
+        files: '\.(mojo|🔥|py|md)$'
         stages: [pre-commit]
-      - id: check-docstrings
-        name: check-docstrings
-        entry: python3 ./mojo/stdlib/scripts/check-docstrings.py
-        language: system
         pass_filenames: false
-        stages: [pre-commit]
```
Ahajha, line 28: Seems the docstrings checker isn't in the bazel linter. I'm curious how there's a difference in behavior.

.github/workflows/test_pre_commit.yml
```diff
@@ -42,10 +42,5 @@ jobs:
...
           echo "$HOME/.pixi/bin" >> $GITHUB_PATH
 
-      - name: Install pre-commit
-        run: |
-          pip install pre-commit
-          pre-commit install
-
       - name: Run pre-commit
-        run: pixi run --manifest-path mojo pre-commit run --all-files
+        run: pixi x pre-commit run --all-files
```
Ahajha, line 46: Aside from the differences between the two, I think we should have this file just run `./bazelw run lint` (which is `./bazelw run format` but it doesn't change anything, would just be noise in CI). Probably fine to leave as-is until those are fully in sync.

Ahajha (commented):
Overall LGTM, let's just sort out the docstrings issue.

.pre-commit-config.yaml
gabrieldemarmiesse: It seems mojo/stdlib/scripts/check-docstrings.py wasn't working as intended. 
```
pixi run --manifest-path ./mojo mojo doc --diagnose-missing-doc-strings --validate-doc-strings -o /dev/null ./mojo/stdlib/stdlib
```
Gives the correct output and errors. But there are too many of them to be fixed in this PR

.github/workflows/test_pre_commit.yml
gabrieldemarmiesse: The pre-commit additionaly formats bazel files, I don't think it's something that bazelw format does currently

.pre-commit-config.yaml
Ahajha: Interesting. Let me try to mimic that behavior in bazel, I just don't want this to get lost.

.github/workflows/test_pre_commit.yml
Ahajha: Internally we're more unified on `./bazelw run lint/format`, the version in this repo is a partial recreation of that, so yea that just hasn't been ported over yet.

.pre-commit-config.yaml
Ahajha: Okay, actually after some internal discussion I think we can just remove it like you have here.

Ahajha: It's covered in a different way in bazel tests

.github/workflows/test_pre_commit.yml
Ahajha: I'm adding buildifier support here: https://github.com/modular/modular/pull/5273

Ahajha: Okay I just landed that change internally. Once that lands here tomorrow, we could change this file to just use `./bazelw run lint`. Thoughts? I figure at that point the `pixi` indirection is a bit unnecessary.

Ahajha: As of the last nightly, `./bazelw run format` also runs buildifier. So I think we can have this file just run that, and remove the separate buildifier step.

gabrieldemarmiesse: All good. I chose to keep the pre-commit run --all-files as it make sure the .pre-commit-config.yaml is valid in addition to also run the bazel format command.

Ahajha: Seems fair enough.

Ahajha:
!sync

Ahajha:
Copybara didn't close this (I did something wrong on my end), closing as completed!

PR #5282, opened 2025-09-05
[stdlib][proposal] Allow `Scalar[DType.bool].__index__()`

martinvuyk:
Allow `Scalar[DType.bool].__index__()`. This is useful for constructing things like `UInt`. A secondary effect this has that might or not be desired is that now one can do:
```mojo
var options = [1, 2]
var some_condition = Scalar[DType.bool](True)
print(options[some_condition]) # 2
```

### DIFF
--- a/mojo/stdlib/stdlib/builtin/simd.mojo
+++ b/mojo/stdlib/stdlib/builtin/simd.mojo
@@ -1859,7 +1859,8 @@ struct SIMD[dtype: DType, size: Int](
             The corresponding __mlir_type.index value.
         """
         constrained[
-            dtype.is_integral(), "cannot index using a floating point type"
+            not dtype.is_floating_point(),
+            "cannot index using a floating point type",
         ]()
         # NOTE: using Int(self) here would cause an infinite recursion.
         return self.__int__()._mlir_value


NathanSWard:
I feel like this is probably not very desirable, especially since it now allows the following (since `BoolLiteral` got removed):
```mojo
var l = [1, 2, 3, 4, 5]
var n = l[True]
```

Being able to index with `True` and `False` should (IMO) not be possible.

mojo/stdlib/stdlib/builtin/simd.mojo
```diff
@@ -1859,7 +1859,8 @@ struct SIMD[dtype: DType, size: Int](
             The corresponding __mlir_type.index value.
         """
         constrained[
-            dtype.is_integral(), "cannot index using a floating point type"
+            not dtype.is_floating_point(),
+            "cannot index using a floating point type",
```
NathanSWard, line 1863: I feel like this is probably not very desirable, especially since it now allows the following (since `BoolLiteral` got removed):
```mojo
var l = [1, 2, 3, 4, 5]
var n = l[True]
```

Being able to index with `True` and `False` should (IMO) not be possible.

martinvuyk: Yeah it is a bit weird, though the alternative is adding 2 constructors for `Scalar[DType.bool]` and `Bool` respectively to `UInt`.

> since `BoolLiteral` got removed

On another topic, I think it might be useful to have it back. Using dependent types to define function overloads that are colored differently (raising, async, etc.) might be useful later on

laszlokindrat:
Is there an immediate need for this change, i.e. some bigger objective that this unblocks? I'm currently trying to remove `UInt` in favor of a `SIMD` alias, and unless there is a strong motivation, I don't want to risk introducing more subtle ways the core numeric types can convert between each other.

martinvuyk:
Yeah I think I'll just drop this, it was just an idea

PR #5283, opened 2025-09-05
[stdlib] Rename `DType.bit_width()` and replace uses of `sys.info.bit_width_of`

martinvuyk:
Rename `DType.bit_width()` and replace uses of `sys.info.bit_width_of`

### DIFF
--- a/max/kernels/src/Mogg/MOGGKernelAPI/MOGGKernelAPI.mojo
+++ b/max/kernels/src/Mogg/MOGGKernelAPI/MOGGKernelAPI.mojo
@@ -33,7 +33,7 @@ from math import (
     tanh,
 )
 from random import randn, seed
-from sys import bit_width_of, external_call, llvm_intrinsic
+from sys import external_call, llvm_intrinsic
 from sys.info import simd_width_of, size_of
 from sys.intrinsics import _type_is_eq
 
--- a/max/kernels/src/layout/tensor_core_async.mojo
+++ b/max/kernels/src/layout/tensor_core_async.mojo
@@ -270,7 +270,7 @@ fn select_k_atom[
         `Layout` - A core matrix layout optimized for tensor core operations.
     """
     alias a = _select_k_atom_bits[swizzle_mode]()
-    return upcast(a, dtype.bitwidth())
+    return upcast(a, dtype.bit_width())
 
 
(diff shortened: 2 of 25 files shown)


laszlokindrat:
Thanks @martinvuyk! I understand the renaming, but what is the motivation behind rweplacing uses of `sys.info.bit_width_of`? I'm not aware of plans to deprecate it.

martinvuyk:
> Thanks @martinvuyk! I understand the renaming, but what is the motivation behind rweplacing uses of `sys.info.bit_width_of`? I'm not aware of plans to deprecate it.

Because it's more convenient (and IMO more readable) to use the dtype method directly

laszlokindrat:
> > Thanks @martinvuyk! I understand the renaming, but what is the motivation behind rweplacing uses of `sys.info.bit_width_of`? I'm not aware of plans to deprecate it.
> 
> Because it's more convenient (and IMO more readable) to use the dtype method directly

But your change also replaces uses like `bit_width_of[Int]()`, which has a slightly different API for this. Given that we are contemplating moving `Int` to be an alias to a `Scalar` type, this could actually introduce more churn, so I would like to request that we hold off with this. I would be happy to land the renaming, though :)

martinvuyk:
> > > Thanks @martinvuyk! I understand the renaming, but what is the motivation behind rweplacing uses of `sys.info.bit_width_of`? I'm not aware of plans to deprecate it.
> > 
> > 
> > Because it's more convenient (and IMO more readable) to use the dtype method directly
> 
> But your change also replaces uses like `bit_width_of[Int]()`, which has a slightly different API for this. Given that we are contemplating moving `Int` to be an alias to a `Scalar` type, this could actually introduce more churn, so I would like to request that we hold off with this. I would be happy to land the renaming, though :)

I could also replace all instances of `bit_width_of[Int]()` with `DType.index.bit_width()`. Which do you prefer?

laszlokindrat:
Let's just keep `bit_width_of` to minimize to blast radius. Thanks!

laszlokindrat:
I actually warmed up to the idea of doing just `.bit_width()`. Can you please rebase on the latest so I can do a final review and bring it in? Thanks!

laszlokindrat (approved):

laszlokindrat:
!sync

PR #5284, opened 2025-09-06
[stdlib] `Variant` and `Optional` auto-optimizes on type trivialities.

rd4com:
Hello,
this is early bottom-up work for auto-optimizing.

The PR is status: draft,
because i'm not sure if this is the way to go, is it?

What we have here is quite encouraging.
Mojo's meta-programing is powerful,
we can auto-optimize the types based on their parameters. 
(Even with *Ts)

So for example if T is trivial on an Optional,
the optional can become Trivial too for that dunder.

This is bottom-up, because Optional uses Variant[_NoneType, T]. 
The Optional itself ask the Variant for its trivialities. 
And the Variant get it from all T in Ts.

So if we implement this for InlineArray too,
InlineArray[Optional[Int]] would get its trivialities like this: 
_NoneType&Int -> Variant[_NoneType, Int] -> Optional -> InlineArray.

Any ideas and thoughts?

InlineArray trivialities could work in an nested way too: 
InlineArray[InlineArray[Int, 64], 64]

But not sure if it is not "too soon" for theses features, 
should we wait for using T.__*__is_trivial ?

### DIFF
--- a/mojo/stdlib/stdlib/collections/optional.mojo
+++ b/mojo/stdlib/stdlib/collections/optional.mojo
@@ -40,6 +40,10 @@ from utils import Variant
 # TODO(27780): NoneType can't currently conform to traits
 @fieldwise_init
 struct _NoneType(ImplicitlyCopyable, Movable):
+    alias __copyinit__is_trivial = True
+    alias __moveinit__is_trivial = True
+    alias __del__is_trivial = True
+
     fn __copyinit__(out self, other: Self):
         pass
 
@@ -86,6 +90,9 @@ struct Optional[T: Copyable & Movable](
     alias _type = Variant[_NoneType, T]
     var _value: Self._type
 
+    alias __copyinit__is_trivial = Self._type.__copyinit__is_trivial
+    alias __moveinit__is_trivial = Self._type.__moveinit__is_trivial
+    alias __del__is_trivial = Self._type.__del__is_trivial
     # ===-------------------------------------------------------------------===#
     # Life cycle methods
     # ===-------------------------------------------------------------------===#
--- a/mojo/stdlib/stdlib/collections/optional.mojo
+++ b/mojo/stdlib/stdlib/collections/optional.mojo
@@ -40,12 +40,7 @@ from utils import Variant
 # TODO(27780): NoneType can't currently conform to traits
 @fieldwise_init
 struct _NoneType(ImplicitlyCopyable, Movable):
-    alias __copyinit__is_trivial = True
-    alias __moveinit__is_trivial = True
-    alias __del__is_trivial = True
-
-    fn __copyinit__(out self, other: Self):
-        pass
+    ...
 
 
 # ===-----------------------------------------------------------------------===#
@@ -90,9 +85,6 @@ struct Optional[T: Copyable & Movable](
     alias _type = Variant[_NoneType, T]
     var _value: Self._type
 
-    alias __copyinit__is_trivial = Self._type.__copyinit__is_trivial
-    alias __moveinit__is_trivial = Self._type.__moveinit__is_trivial
-    alias __del__is_trivial = Self._type.__del__is_trivial
     # ===-------------------------------------------------------------------===#
     # Life cycle methods
     # ===-------------------------------------------------------------------===#
--- a/mojo/stdlib/stdlib/utils/variant.mojo
+++ b/mojo/stdlib/stdlib/utils/variant.mojo
@@ -128,6 +128,11 @@ struct Variant[*Ts: Copyable & Movable](ImplicitlyCopyable, Movable):
     ]
     var _impl: Self._mlir_type
 
+    # Trivialities from Ts
+    alias __copyinit__is_trivial = Self._all_t_trivial["__copyinit__"]()
+    alias __moveinit__is_trivial = Self._all_t_trivial["__moveinit__"]()
+    alias __del__is_trivial = Self._all_t_trivial["__del__"]()
+
     # ===-------------------------------------------------------------------===#
     # Life cycle methods
     # ===-------------------------------------------------------------------===#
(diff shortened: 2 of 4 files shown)


mojo/stdlib/test/builtin/test_value.mojo
```diff
@@ -69,6 +70,30 @@ struct StructInheritTriviality[T: Movable & Copyable](Movable & Copyable):
     alias __del__is_trivial = T.__del__is_trivial
 
 
+struct SpecificTrivialities[
+    trivialities: Int = 0,
+](Copyable & Movable):
+    alias __moveinit__is_trivial = Bool(trivialities & EVENT_MOVE)
+    alias __copyinit__is_trivial = Bool(trivialities & EVENT_COPY)
+    alias __del__is_trivial = Bool(trivialities & EVENT_DEL)
+    var events: UnsafePointer[List[Int]]
```
martinvuyk, line 79: Please put an origin in this struct for the events buffer

martinvuyk (commented):
Nice! I think this will allow us to finally get rid of `OptionalReg`

rd4com:
Ok, this can help for [upgrading-trivial](https://github.com/modular/modular/blob/main/mojo/proposals/upgrading-trivial.md) (proposal by C.L to improve ergonomics and register_passable opti),
The proposal suggest the inlinearray optis too in the same `Self.__*__is_trivial=` implemented in the pr.
There is an roadmap entry for phase 1! 
So maybe this overlap some WIP

mojo/stdlib/stdlib/collections/optional.mojo
```diff
@@ -40,6 +40,10 @@ from utils import Variant
 # TODO(27780): NoneType can't currently conform to traits
 @fieldwise_init
 struct _NoneType(ImplicitlyCopyable, Movable):
+    alias __copyinit__is_trivial = True
+    alias __moveinit__is_trivial = True
+    alias __del__is_trivial = True
```
NathanSWard: For `NoneType` these aliases are already synthesized by the compiler so we shouldn't need to manually add them here.

mojo/stdlib/stdlib/collections/optional.mojo
```diff
@@ -86,6 +90,9 @@ struct Optional[T: Copyable & Movable](
     alias _type = Variant[_NoneType, T]
     var _value: Self._type
 
+    alias __copyinit__is_trivial = Self._type.__copyinit__is_trivial
+    alias __moveinit__is_trivial = Self._type.__moveinit__is_trivial
+    alias __del__is_trivial = Self._type.__del__is_trivial
```
NathanSWard: So this is an odd one and from what I can see it looks like a compiler bug. 
Ideally a user should never have a manually add the `__x __is_trivial` aliases for a struct, however, the synthesized aliases are not correctly dependent on the struct's member variables. Let me see if we can't get this fixed :)

mojo/stdlib/stdlib/utils/variant.mojo
```diff
@@ -408,6 +417,26 @@ struct Variant[*Ts: Copyable & Movable](ImplicitlyCopyable, Movable):
...
+    @staticmethod
+    fn _all_t_trivial[_dunder: String]() -> Bool:
+        var is_trivial = True
+
+        @parameter
+        for i in range(len(VariadicList(Ts))):
+
+            @parameter
+            if _dunder == "__del__":
+                is_trivial &= Bool(Ts[i].__del__is_trivial)
```
NathanSWard: IMO - its a bit easier to ready without the bit logic, and instead we should just return here
```suggestion
                if not(Bool(Ts[ i ].__del__is_trivial)):
                    return False
```

mojo/stdlib/test/builtin/test_value.mojo
```diff
@@ -128,6 +153,113 @@ def test_type_inherit_non_triviality[T: Movable & Copyable]():
     )
 
 
+# ===-----------------------------------------------------------------------===#
+# Implementations tests
+# ===-----------------------------------------------------------------------===#
+
+
+def test_variant_specific_trivialities[trivialities: Int = 0]():
```
NathanSWard, line 161: These tests should live in the according container's test file
- `Varaint` -> `test_variant.mojo`
- `Optional` -> `test_optional.mojo`

Also the tests themselves are quite large. Ideally in these files, each `test_` function should be a unit-style test meaning the scope of them is small. These should get broken down into a couple smaller unit tests if possible 😄

NathanSWard (commented):
Thanks for this! :) 
It seems this may have surfaces a few compiler bugs too - I'll see if the compiler team can't get these addressed!

rd4com:
Thanks @NathanSWard and @martinvuyk ,
Just waiting for next nightly to start working on the reviews :+1:

mojo/stdlib/stdlib/collections/optional.mojo
rd4com: Thanks,
the struct have an `__copyinit__`, 
so i was expecting mojo to synthesize `Self.__copyinit__is_trivial = False` ?

NathanSWard: Oh odd I was thinking of the stdlib `NoneType` - I'm not sure of what the purpose of this type is. 
If we can just remove the `__copyinit__` function then the compiler should do everything correctly 👍

mojo/stdlib/stdlib/collections/optional.mojo
NathanSWard: @rd4com - this was fixed in the compiler (should be on next nightly release).
Now you can rely on the compiler generated aliases in `Optional` to be the correct value

rd4com: Thanks @NathanSWard,
`Mojo 0.25.6.0.dev2025090805 (1f84bce6)` is not yet
(waiting for next one :+1: )

mojo/stdlib/stdlib/collections/optional.mojo
rd4com: Review implemented @NathanSWard :+1: 
The Optional seem to get the correct triviality !

mojo/stdlib/stdlib/collections/optional.mojo
rd4com: Yep, review implemented @NathanSWard :+1:,
the Optional seem to get the correct triviality from Variant !

mojo/stdlib/stdlib/utils/variant.mojo
rd4com: Thanks, review implemented @NathanSWard :+1:

rd4com:
@NathanSWard , CI tests failed on last commits

JoeLoser:
@NathanSWard is this one still blocked on compiler bugs?  If not, I think we could pursue it now, especially with recent improvements in variadics for @lsh's work.

rd4com:
@JoeLoser , @NathanSWard ,
Yep, Nate got this going well with:
https://github.com/modular/modular/commit/392b0bcb2f106dcb12db1e6decfdb60cc55639a5
Great, internally it will land much faster and thoroughly refined !!
Feel free to scavage anything useful from the PR,
i'd love to side contribute additional tests if needed (another PR?) :+1: 

The change to auto-trivial would affect optional, variant, .and much of stdlib,
it comes with enormous benefit for users, 
which is both exciting and important :+1: 

Recommendations:
- Just have to make sure the transition comes with very trourough testing (never enough),
even it means removing few redudants tests after if needed.
- Check if the whole system works together all types as whole *at once* (implemented on all types)
  (`InlineArray[8, InlineArray[Optional[Variant[..]]]]`)
  
With this being done between two stable release, 
any possible bug if any should/could appear during nightly :+1: 

Happy new year :D

NathanSWard:
Clsoing this PR as these have been fixed over the past couple months. Ty for the PR though!

PR #5285, opened 2025-09-06
[kernels] Change function parameters `type: DType` -> `dtype`

martinvuyk:
Change function parameters `type: DType` -> `dtype`. Rename `MHAConfig.type` to `dtype` as well.

Also some small changes from `size_of[type]() * 8` -> `dtype.bitwidth()` or `size_of[type]()` -> `dtype.size_of()`, and other similar instances. Because those lines were being edited anyway due to the parameter name change

### DIFF
--- a/max/kernels/src/layout/tensor_core_async.mojo
+++ b/max/kernels/src/layout/tensor_core_async.mojo
@@ -253,7 +253,7 @@ fn _select_k_atom_bits[
 
 
 fn select_k_atom[
-    type: DType,
+    dtype: DType,
     swizzle_mode: TensorMapSwizzle,
 ]() -> Layout:
     """Creates a core matrix layout for tensor core operations.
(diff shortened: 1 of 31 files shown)


max/kernels/src/layout/tensor_core_async.mojo
```diff
@@ -263,25 +263,25 @@ fn select_k_atom[
...
-        type: Element data type of the tensor.
+        dtype: Element data type of the tensor.
         swizzle_mode: Memory access pattern swizzling mode.
 
     Returns:
         `Layout` - A core matrix layout optimized for tensor core operations.
     """
     alias a = _select_k_atom_bits[swizzle_mode]()
-    return upcast(a, size_of[type]() * 8)
+    return upcast(a, dtype.bitwidth())
```
NathanSWard, line 273: Ideally with the PR -> since it's goal is it simple rename `type` -> `dtype`, non-naming changes like this should probably be left to a separate PR since this change isn't noted in the PR change log.
There a few other instances of this too in the PR

NathanSWard:
Great clean up - I love when naming conventions are consistent across a codebase 😄

max/kernels/src/layout/tensor_core_async.mojo
martinvuyk: Good point, I added mention about those changes to the PR description, is that enough?

NathanSWard (approved):

NathanSWard:
!sync

PR #5289, opened 2025-09-08
[stdlib] Add `take_items` to Dict to allow for draining existing dicts via move ops

josiahls:
Adds `take_items` and `_TakeDictEntryIter` to Dict. This allows for writing clean APIs that take owned dics, and convert them into another data structure without copying the entries. 

Some questions for reviewers:

- Should we rename this to instead be `drain_items` and `_DrainDictEntryIter`
- Can we just repurpose existing iterators?

Some additional notes:
- ~~There is a lot of private ops in _TakeDictEntryIter. Ideally, I would want a no-raise version of pop (that doesn't branch at all. Possibly just have it as a `_pop_optional` or something like that.~~ I added `_unsafe_take_entry` which is a subset of code in `pop` however avoids the "raises". 
- ~~Possibly~~ changed _TakeDictEntryIter.__has_next__ to just do len(self.src[]) > 0

```
fn main():
    d:Dict[String, String] = {'a': 'b', 'c': 'd', 'e': 'f'}

    for item in d.take_items():
        print(item.key, item.value)

    print('Length: ', len(d))
    for item in d.take_items():
        print(item.key, item.value)
```
Output:
```
a b
c d
e f
Length:  0
```

### DIFF
--- a/mojo/stdlib/stdlib/collections/dict.mojo
+++ b/mojo/stdlib/stdlib/collections/dict.mojo
@@ -108,6 +108,55 @@ struct _DictEntryIter[
 
 
 @fieldwise_init
+struct _TakeDictEntryIter[
+    K: KeyElement, V: Copyable & Movable, H: Hasher, origin: Origin[True]
+](Iterator, Movable):
+    """Iterator over mutable DictEntry references that moves entries out of the dictionary.
+
+    Parameters:
+        K: The key type of the elements in the dictionary.
+        V: The value type of the elements in the dictionary.
+        H: The type of the hasher in the dictionary.
+        origin: The mutable origin of the Dict
+    """
+
+    alias Element = DictEntry[K, V, H]
+
+    var index: Int
+    var src: Pointer[Dict[K, V, H], origin]
+
+    fn __init__(out self, index: Int, ref [origin]dict: Dict[K, V, H]):
+        self.index = index
+        self.src = Pointer(to=dict)
+
+    @always_inline
+    fn __iter__(var self) -> Self:
+        return self^
+
+    @always_inline
+    fn __has_next__(self) -> Bool:
+        return len(self.src[]) > 0
+
+    @always_inline
+    fn __next__(
+        mut self,
+    ) -> Self.Element:
+        while True:
+            ref opt_entry_ref = self.src[]._entries[self.index]
+
+            self.index += 1
+
+            if opt_entry_ref:
+                ref key = opt_entry_ref.value().key
+                var hash = hash[HasherType=H](key)
+                var found, slot, index = self.src[]._find_index(hash, key)
+
+                if found:
+                    # TODO: Should we call `self.src[]._compact()` if len(self.src[]) is 0?
+                    return self.src[]._unsafe_take_entry(slot, index)
+
+
+@fieldwise_init
 struct _DictKeyIter[
     mut: Bool, //,
     K: KeyElement,
@@ -963,12 +1012,7 @@ struct Dict[K: KeyElement, V: Copyable & Movable, H: Hasher = default_hasher](
         var hash = hash[HasherType=H](key)
         var found, slot, index = self._find_index(hash, key)
         if found:
-            self._set_index(slot, Self.REMOVED)
-            ref entry = self._entries[index]
-            debug_assert(entry.__bool__(), "entry in index must be full")
-            var entry_value = entry.unsafe_take()
-            entry = None
-            self._len -= 1
+            var entry_value = self._unsafe_take_entry(slot, index)
             return entry_value^.reap_value()
         raise Error("KeyError")
 
@@ -992,6 +1036,7 @@ struct Dict[K: KeyElement, V: Copyable & Movable, H: Hasher = default_hasher](
         var val = Optional[V](None)
 
         for item in reversed(self.items()):
+            # TODO: Is there a way to do via move instead of copy?
             key = Optional(item.key.copy())
             val = Optional(item.value.copy())
             break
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/collections/dict.mojo
```diff
@@ -107,6 +107,64 @@ struct _DictEntryIter[
                 return opt_entry_ref.value()
 
 
+@fieldwise_init
+struct _TakeDictEntryIter[
+    K: KeyElement,
+    V: Copyable & Movable,
+    H: Hasher,
+    origin: Origin[True],
+    forward: Bool = True,
```
NathanSWard: `forward` isn't really needed here. IMO for now we don't need to add the complexity of being able to backwards "drain" from a `Dict`.

mojo/stdlib/stdlib/collections/dict.mojo
```diff
@@ -107,6 +107,64 @@ struct _DictEntryIter[
...
+](ImplicitlyCopyable, Movable):
```
NathanSWard:
```suggestion
](ImplicitlyCopyable, Iterator, Movable):
```
This type should implement the new `Iterator` trait if possible.

NathanSWard:
```suggestion
](Movable):
```

This should probably not be copyable IMO. Copying a draining iterator feel like a code smell and could potentially result in UB if the implementation assumed it had exclusive ownership over the collection.

mojo/stdlib/test/collections/test_dict.mojo
```diff
@@ -249,6 +249,27 @@ def test_iter_items():
...
+        sum += entry.value
+
+    assert_equal(keys, "ab")
+    assert_equal(sum, 3)
+    assert_equal(len(dict), 0)
+
+    var iterated_again = False
+    for entry in dict.take_items():
+        iterated_again = True
+    assert_false(iterated_again)
```
NathanSWard:
```suggestion
    assert_false(dict.take_items().__has_next__())
```

This could be simplified

mojo/stdlib/stdlib/collections/dict.mojo
```diff
@@ -961,6 +1019,7 @@ struct Dict[K: KeyElement, V: Copyable & Movable, H: Hasher = default_hasher](
         var hash = hash[HasherType=H](key)
         var found, slot, index = self._find_index(hash, key)
         if found:
+            # TODO: Replace with _unsafe_take_entry
```
NathanSWard: I would just replace this with `_unsafe_take_entry` to make this DRY in this PR. Otherwise we have duplicated code

mojo/stdlib/stdlib/collections/dict.mojo
josiahls: I had to change the iter to this:
```
    fn __iter__(var self) -> Self:
        return self^
```
is that expected then? I agree, there shouldn't be a copy for this. 

I tried:
```
fn __iter__(ref self) -> ref[self] Self:
    return self
```
But id get `is not implicitly copyable because it does not conform to 'ImplicitlyCopyable'` again.

Alternatively, Add Copyable to this tierator and return `self.copy()`. I think the copy would be "cheap" anyways though weird to do since this usually immediately follows a initialization.

mojo/stdlib/stdlib/collections/dict.mojo
```diff
@@ -107,6 +107,64 @@ struct _DictEntryIter[
...
+                self.index -= 1
+
+            if opt_entry_ref:
+                ref key = opt_entry_ref.value().key
+                var hash = hash[HasherType=H](key)
+                var found, slot, index = self.src[]._find_index(hash, key)
+
+                if found:
+                    # TODO: Should we call `self.src[]._compact()` if len(self.src[]) is 0?
+                    return self.src[]._unsafe_take_entry(slot, index)
```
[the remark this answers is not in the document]
josiahls: @NathanSWard This would be an issue in the original _DictEntryIter also. I can add a `if self.index > len(_entries)` to the new iter, or add it to both. Not sure how much "new" changes I want to make in this mr.

mojo/stdlib/stdlib/collections/dict.mojo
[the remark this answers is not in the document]
josiahls: done

josiahls: done

mojo/stdlib/test/collections/test_dict.mojo
josiahls: done

mojo/stdlib/stdlib/collections/dict.mojo
josiahls: done

mojo/stdlib/stdlib/collections/dict.mojo
josiahls: Agreed, done

mojo/stdlib/stdlib/collections/dict.mojo
NathanSWard: You shouldn't need the `__iter__` function on this type :)
`__iter__` is only needed for `Iterable` types whereas this type is an _iterator_ return from an _iterable_

josiahls: Removed!

Not recommending doing it in this pr, but `_DictKeyIter` and `_DictValueIter` are in a similar boat, and `_DictEntryIter` probably should remove its `__iter__`, and implement Iterator. I can make a separate pr to clean this up if that makes sense to you.

josiahls: I take it back, im getting:
```
mojo/stdlib/test/collections/test_dict.mojo:259:33: error: '_TakeDictEntryIter[String, Int, AHasher[0], dict]' does not implement the '__iter__' method
    for entry in dict.take_items():
```
Iterator is probably the wrong trait for these. These should be themselves Iterables. I re-added the __iter__ to this iterator. 

Alternatively:
I'm wondering if there is something that needs to be done to Dict's different iterator structs in general. There probably should instead be a single _InnerDictIter Iterable wrapper that accepts different iterators (item, key, value, take item).

One issue I see, we have parametric mutability for item, key, value, which is for the most part ok. But take item **must** be mutable. I think that it should still be possible to use a common inner iterable wrapper, but the mutability for take item can't be parametric, otherwise the call to `_unsafe_take_entry` is going to panic.

NathanSWard: Ahh my bad - I didn't fully grok the different iterator traits :)

NathanSWard (approved):
lgtm! 
There were some recent changes to `Iterator` (or was it `Iterable`) but this shouldn't conflict with it (or if it does I'll make some small adjustments to fix it up).

Thanks for this!

NathanSWard:
!sync

Issue #5300, opened 2025-09-11
[Feature Request] Add new MAX ops `bitwise_and`, `bitwise_or` and `bitwise_xor`


gabrieldemarmiesse:
### What is your request?

torch functions:
* https://docs.pytorch.org/docs/stable/generated/torch.bitwise_and.html
* https://docs.pytorch.org/docs/stable/generated/torch.bitwise_or.html
* https://docs.pytorch.org/docs/stable/generated/torch.bitwise_xor.html

Note that we could provide ops with those names but also if we want to follow torch's api, we'll want to implement it in `TensorValue`'s `__and__`, `__or__` and `__xor__`. The behavior in pytorch for those is to use the bitwise version:

```
In [3]: torch.tensor([1, 2], dtype=torch.uint8) |  torch.tensor([2, 4], dtype=torch.uint8)
Out[3]: tensor([3, 6], dtype=torch.uint8)
```

### What is your motivation for this change?

I need it to implement the max backend for torch, and surprisingly it's used in gemma3 (or at least it's what torch aotautograd decomposes some ops into).

### Any other details?

I could code them in torch-max-backend directly in Mojo, but then I end up with this bug: https://github.com/modular/modular/issues/5299

ferasboulala:
Looking at other implementations, there are two types of operations:
- Bitwise: Straightforward bitwise operations
- Logical: Operations that assume 0 is False and the rest is True
They work on all data types.

MAX currently does not support the logical kind of operations and it has a constrain on the data type of the bitwise kind (must be bool).

I propose that we remove the data type constraint on the existing operations. This should give us full bitwise support. Then, we can add logical operations later.

PR #5301, opened 2025-09-11
[MAX] Add bitwise_and() op

gabrieldemarmiesse:
Currently Max does not expose bitwise operations. Those are available in torch through operators like `&` or `|`. We are adding those now, starting with `bitwise_and()` in this PR. 

Address part of https://github.com/modular/modular/issues/5300


abduld:
Thanks @gabrieldemarmiesse . Can you add a commit message along with a test

gabrieldemarmiesse:
@abduld the issue we have is that I cannot run the unit tests from the public repo and the public CI cannot run them either. So since it's not possible for me to run any of the code I write, I'll let someone in Modular do this.

See https://github.com/modular/modular/issues/5303

abduld:
!sync

abduld:
> @abduld the issue we have is that I cannot run the unit tests from the public repo and the public CI cannot run them either. So since it's not possible for me to run any of the code I write, I'll let someone in Modular do this.
> 
> See #5303

makes sense. Please though add a commit message :)

gabrieldemarmiesse:
@abduld thanks! I edited the main message, you can use it as commit message.

ehsanmok:
@gabrieldemarmiesse Thank you for working on this! We're currently planning to make the rest of the testing modules available so please stay tuned.

lattner:
nice!

PR #5308, opened 2025-09-14
[stdlib][mojo] Add a `pop(slice)` method override to the `List` struct.

christoph-schlumpf:
Add a `pop(slice)` method override to the `List` struct.

This complements `__getitem__(slice)`.

Example: Pop every second item
```mojo
fn main():
    var list = [1, 2, 3, 4, 5, 6]
    var popped = list.pop({None, None, 2})
    print(popped.__str__()) # [1, 3, 5]
    print(list.__str__()) # [2, 4, 6]
```


mojo/stdlib/test/collections/test_list.mojo
```diff
@@ -185,6 +185,147 @@ def test_list_pop():
...
+    assert_equal(list[1], 3)
+    assert_equal(len(list), 2)
+
+    list = [1, 2, 3]
+    popped = list.pop({None, None, -1})
+    assert_equal(popped[0], 3)
+    assert_equal(popped[1], 2)
+    assert_equal(popped[2], 1)
+    assert_equal(len(popped), 3)
+    assert_equal(len(list), 0)
```
NathanSWard: For these tests - would you be able to simplify them a bit? Specifically you can rewrite:

```suggestion
    assert_equal(popped, [0, 1, 2])
    assert_equal(len(list), 0)
```

It just condenses the test a bit and makes it more readable imo!

Thank you again for this contribution!

mojo/stdlib/std/collections/list.mojo
```diff
@@ -837,6 +837,54 @@ struct List[T: Copyable & Movable, hint_trivial_type: Bool = False](
...
 
+    fn pop(mut self, slice: Slice) -> Self:
+        """Pops a slice of items from the list.
+
+        Args:
+            slice: The slice of items to pop.
+
+        Returns:
+            A list of popped items.
+        """
```
NathanSWard, line 956: Curious, is the reason this is returning a new `List` so this can be parallel with the `pop(index)` function? Is it useful to return the new list or should it be omitted - since creating the new list (specifically a new allocation) seems a bit wasteful unless the purpose of this function is to get the elements from the list and not just remove them.

christoph-schlumpf: Exactly. IMO the semantics of `pop` is to return the removed item(s). This aligns well with `pop(idx)`, `__getitem__(idx)` and `__getitem__(slice)`.

I think there are use cases for both options (returning the removed items or just deleting them). The `stdlib` should provide both.

My intention is to create an additional `Feature Request / PR ` that provides a `delete(slice)` method.
But if you like I can add `delete(slice)` to this PR so that both options are available in this PR.

mojo/stdlib/test/collections/test_list.mojo
christoph-schlumpf: OK, I will update the PR accordingly.

Thank you for the review.

christoph-schlumpf: I cleaned up the tests accordingly.

christoph-schlumpf:
Hi @NathanSWard 
Just let me know if I can do something concerning this PR.

PR #5314, opened 2025-09-16
[example][mojo] Fix matrix_multiplication.mojo example (naive case)

w32zhong:
1. The writeback operation needs to be guarded as well. 
2. The naive matmul is mapping (x, y) to MxN, not NxM.


w32zhong:
I have read the CLA Document and I hereby sign the CLA.

JoeLoser:
@BradLarson do you mind taking a look at this, please? Thanks!

BradLarson:
Thank you for taking a look at these, and I apologize for how long it has taken to review this. In the meantime, we've reworked some of these examples and I don't believe that this modification is necessary with the code in its current state. The naive and coalescing matrix multiplication in their current versions return correct results.

PR #5316, opened 2025-09-16
[stdlib] Fix origin mutability of pointers in `memory.mojo` and `qmatmul_k.mojo` and add `NDBuffer[mut=False]` implicit ctor

martinvuyk:
Part of https://github.com/modular/modular/pull/4396.

This also adds an implicit constructor from NDBuffer[mut=_] -> NDBuffer[mut=False] to avoid having to fix ambiguous function signatures all up the call-stack

 #4398 got closed because I accidentally pushed before doing the commit.

CC: @lsh @JoeLoser

### DIFF
--- a/max/kernels/src/nn/mla.mojo
+++ b/max/kernels/src/nn/mla.mojo
@@ -1434,21 +1434,9 @@ fn flare_mla_prefill_dispatch[
 
     alias smem_use = (q_smem + k_smem + v_smem) * config.dtype.size_of()
 
-    var softmax_info_ptr = (
-        softmax_info.value().data if softmax_info else UnsafePointer[
-            Scalar[softmax_type]
-        ]()
-    )
-    var prev_output_ptr = (
-        prev_output.value().data if prev_output else UnsafePointer[
-            Scalar[output_type]
-        ]()
-    )
-    var prev_softmax_info_ptr = (
-        prev_softmax_info.value().data if prev_softmax_info else UnsafePointer[
-            Scalar[softmax_type]
-        ]()
-    )
+    var softmax_info_ptr = softmax_info.or_else({}).data
+    var prev_output_ptr = prev_output.or_else({}).data
+    var prev_softmax_info_ptr = prev_softmax_info.or_else({}).data
 
     alias kernel = mla_prefill[
         config.dtype,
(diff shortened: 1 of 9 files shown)


NathanSWard:
This is super helpful @martinvuyk :) 
Greatly appreciate you taking the time to do this!

Is the current approach to fix all usages of `UnsafePointer` across the codebase and then having a final PR cleaning up the constructors in `unsafe_pointer.mojo`?

max/kernels/src/linalg/matmul_vnni.mojo
```diff
@@ -110,7 +110,11 @@ struct Inner_matmul_vnni[saturated_vnni: Bool](InnerMatmulKernel, Movable):
             # exclusivity checking, but it is safe in the sense that
             # `a` will be guaranteed to remain alive because
             # it is an argument to the function.
-        ) else a_base_ptr.origin_cast[True, MutableAnyOrigin]()
+        ) else a_base_ptr.origin_cast[
+            True, MutableAnyOrigin
+        ]().static_alignment_cast[
+            a_local.alignment
+        ]()
```
NathanSWard: `alignment` very recently got removed from `UnsafePointer` :)

martinvuyk: Yeah I'm still working through all remainders of that change

martinvuyk:
> Is the current approach to fix all usages of UnsafePointer across the codebase and then having a final PR cleaning up the constructors in unsafe_pointer.mojo?

Yeah at least that is what I was aiming for, because otherwise it would be a huge PR since there are many places where `UnsafePointer` get wrongfully casted implicitly

max/kernels/src/nn/mla.mojo
```diff
@@ -1434,21 +1434,15 @@ fn flare_mla_prefill_dispatch[
...
-            Scalar[output_type]
-        ]()
-    )
-    var prev_softmax_info_ptr = (
-        prev_softmax_info.value().data if prev_softmax_info else UnsafePointer[
-            Scalar[softmax_type]
-        ]()
-    )
+    var softmax_info_ptr = softmax_info.or_else(
+        __type_of(softmax_info.value())()
```
soraros: Would this work?
```suggestion
        {}
```

martinvuyk: Let's find out

martinvuyk: Nice, it worked

NathanSWard:
Is #5323 a partial replacement for this PR?

martinvuyk:
> Is https://github.com/modular/modular/pull/5323 a partial replacement for this PR?

more or less yeah, let me put this one as draft as well until I manage to reduce the scope, otherwise I'll just go with this one

martinvuyk:
Superseded by #5323

PR #5318, opened 2025-09-16
[stdlib] Chase after `UnsafePointer` origin bugs take 2

martinvuyk:
Chase after `UnsafePointer` origin bugs take 2. Rebase of  #4396. Part of #4386

### DIFF
--- a/mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
+++ b/mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
@@ -210,16 +210,79 @@ struct UnsafePointer[
     @always_inline("builtin")
     @implicit
     fn __init__(
-        out self, other: UnsafePointer[type, address_space=address_space, **_]
+        other: UnsafePointer[*_, **_],
+        out self: UnsafePointer[
+            type = other.type,
+            address_space = other.address_space,
+            mut=False,
+            origin = ImmutableOrigin.cast_from[other.origin],
+        ],
     ):
-        """Exclusivity parameter cast a pointer.
+        """Implicitly cast a pointer to an immutable one.
 
         Args:
             other: Pointer to cast.
         """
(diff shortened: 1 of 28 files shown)


mojo/stdlib/stdlib/memory/unsafe_pointer.mojo
```diff
@@ -210,16 +210,93 @@ struct UnsafePointer[
...
+        """
+        self.address = __mlir_op.`pop.pointer.bitcast`[
+            _type = __type_of(self)._mlir_type
+        ](other.address)
+
+    @always_inline("builtin")
+    @implicit
+    fn __init__(
+        out self: UnsafePointer[NoneType], other: UnsafePointer[NoneType, **_]
+    ):
```
NathanSWard: FWIW: I don't think we should special case for `OpaquePointer` here. Implicitly casting from `void const*` to `void*` is also just as dangerous. Even C++ prevent implicit casting for void*.
https://godbolt.org/z/fe4bKT9Y6

martinvuyk: ok so `other: UnsafePointer[NoneType, mut=True, **_]` then ?

NathanSWard: I think you should be able to just remove this overload entirely? Since `NoneType` isn't necessarily special and will fall into one of the other overloads.

JoeLoser:
As of the other day, ~all of the stdlib is now good with respect to origins for https://github.com/modular/modular/issues/4386. Nate is now continuing up in the stack in the kernel library right now for improving the mut/immut origin issues.  As such, closing this PR.

Thanks for the initial push on this, Martin - very helpful!  It took a while to fix all of the stdlib, but much better to do this sort of thing sooner rather than later.

Issue #4884, opened 2025-06-22
Decide on default parameters vs auto-parametrization in function inputs.


gabrieldemarmiesse:
### Bug description

Today I learned a quirk in the Mojo type system, it might even be a bug, I don't know, what do you think? 

All default values for struct parameters are applied only if at least one parameter has been specified. Otherwise it's assumed to be unbounded (same as _). Here is a surprising example: 

### Steps to reproduce

```mojo
struct Something[T: Copyable & Movable, size: Int=8]:
    var arr: InlineArray[T, size]

    fn __init__(out self, value: T):
        self.arr = InlineArray[T, size](fill=value)

fn print_len_any_type_and_size(a: Something):
    print(len(a.arr))

fn print_len_only_size_8[T: Copyable & Movable](a: Something[T]):
    print(len(a.arr))

fn main():
    a = Something[Int, 10](42)
    print_len_any_type_and_size(a)
    # argument #0 cannot be converted from 'Something[Int, 10]' to 'Something[Int]
    # print_len_only_size_8(a) 

    b = Something[Int, 8](42)
    print_len_any_type_and_size(b)
    print_len_only_size_8(b)
```

We see two very different behaviors if one of the parameters has been supplied or not (specifying the type influences the size). Intuitively, I'd say that when writing `a: Something[T]` then it means my function works for any `size`, like when writing `a: Something`, but it's not what happens at all. 

Is this considered a bug? 

### System information

Mojo 25.5.0.dev2025061905 (11446594)

lsh:
This is not exactly a bug, but something we have discussed internally. `Something` without `[T]` means "autoparameterize every parameter," while `Something[T]` means that all parameters with defaults will be concretized to those defaults. If you wanted it to autoparameterize the rest of those parameters, you would write `Something[T, *_, **_]`.

lsh:
cc @zyx-billy since we talked about this recently

soraros:
Related #4490.

martinvuyk:
This behavior is part of the cause for https://github.com/modular/modular/issues/4386 going unnoticed for so long. Many of the headaches I've been having trying to fix `UnsafePointer` origin bugs (#4396) are mainly due to the default parameters being set for signatures like `UnsafePointer[type]` and  having to go and chase after them to change to `UnsafePointer[type, mut=True, *_, **_]`. There are also some issues for `NDBuffer` (#4398) and `LayoutTensor` (#4701) I've been having.

[...]

PR #5319, opened 2025-09-16
[stdlib] Fix `ffi` and `_libc` `UnsafePointer` parameters

martinvuyk:
Fix `ffi` and `_libc` `UnsafePointer` parameters. Split off from  #5318

CC: @NathanSWard

### DIFF
--- a/mojo/stdlib/stdlib/sys/_libc.mojo
+++ b/mojo/stdlib/stdlib/sys/_libc.mojo
@@ -27,7 +27,7 @@ from sys.ffi import c_char, c_int, c_size_t, get_errno
 
 
 @always_inline
-fn free(ptr: UnsafePointer[NoneType, **_]):
+fn free(ptr: UnsafePointer[NoneType, mut=True, **_]):
     external_call["free", NoneType](ptr)
 
 
@@ -44,7 +44,9 @@ alias FILE_ptr = OpaquePointer
 
 
 @always_inline
-fn fdopen(fd: c_int, mode: UnsafePointer[c_char, **_]) -> FILE_ptr:
+fn fdopen(
+    fd: c_int, mode: UnsafePointer[c_char, mut=False, origin=_]
+) -> FILE_ptr:
     alias name = "_fdopen" if CompilationTarget.is_windows() else "fdopen"
 
     return external_call[name, FILE_ptr](fd, mode)
@@ -62,8 +64,8 @@ fn fflush(stream: FILE_ptr) -> c_int:
 
 @always_inline
 fn popen(
-    command: UnsafePointer[c_char],
-    type: UnsafePointer[c_char],
+    command: UnsafePointer[c_char, mut=False, origin=_],
+    type: UnsafePointer[c_char, mut=False, origin=_],
 ) -> FILE_ptr:
     return external_call["popen", FILE_ptr](command, type)
 
@@ -75,7 +77,10 @@ fn pclose(stream: FILE_ptr) -> c_int:
 
 @always_inline
 fn setvbuf(
-    stream: FILE_ptr, buffer: UnsafePointer[c_char], mode: c_int, size: c_size_t
+    stream: FILE_ptr,
+    buffer: UnsafePointer[c_char, mut=True, origin=_],
+    mode: c_int,
+    size: c_size_t,
 ) -> c_int:
     return external_call["setvbuf", c_int](stream, buffer)
 
@@ -107,7 +112,8 @@ fn dup(oldfd: c_int) -> c_int:
 
 @always_inline
 fn execvp(
-    file: UnsafePointer[c_char], argv: UnsafePointer[UnsafePointer[c_char]]
+    file: UnsafePointer[c_char, mut=False, origin=_],
+    argv: UnsafePointer[UnsafePointer[c_char, mut=False, origin=_]],
 ) -> c_int:
     """[`execvp`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/exec.html)
     — execute a file.
@@ -144,7 +150,7 @@ fn kill(pid: c_int, sig: c_int) -> c_int:
 
 
 @always_inline
-fn pipe(fildes: UnsafePointer[c_int]) -> c_int:
+fn pipe(fildes: UnsafePointer[c_int, mut=True, origin=_]) -> c_int:
     """[`pipe()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/pipe.html) — create an interprocess channel.
     """
     return external_call["pipe", c_int](fildes)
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/sys/_libc.mojo
```diff
@@ -75,7 +75,10 @@ fn pclose(stream: FILE_ptr) -> c_int:
 
 @always_inline
 fn setvbuf(
-    stream: FILE_ptr, buffer: UnsafePointer[c_char], mode: c_int, size: c_size_t
+    stream: FILE_ptr,
+    buffer: UnsafePointer[c_char, mut=True, **_],
```
NathanSWard: Totally agree with these set of changes 🥳 
From seeing all of this, it definitly feels like all functions _taking_ an `UnsafePointer` should use `**_` and not rely on the defaulted parameters.

martinvuyk: Yeah most of them do. There is an ongoing debate in #4884 on whether to keep the current defaulting behavior

soraros: > Totally agree with these set of changes 🥳 From seeing all of this, it definitly feels like all functions _taking_ an `UnsafePointer` should use `**_` and not rely on the defaulted parameters.

That said, does it make sense for `setvbuf` to take a pointer with, e.g., `address_space = SHARED`? We chose to make these signatures generic in these fixes, but I wonder whether that's really semantically correct.

martinvuyk: Good point, there are cases where address space should be defaulted.

lemme summon @owenhilyard re: `setvbuf`, do you think we'll ever get a stream that runs on a non-default address space using a POSIX API?

soraros: Yea. My point is not specific to `setvbuf`, though. It would be good for @NathanSWard to audit all your past fixes and fix up (this is what I meant by "technical decision have to be made" [here](https://github.com/modular/modular/issues/4386#issuecomment-2819754030)) as appropriate. We could resolve the typing issue first in this PR.

NathanSWard: That's a good point @soraros .
Since `**_` will also make the `address_space` unbound. 
@martinvuyk - we might need to be pedantic here and say:
```
fn foo[T: AnyType](ptr: UnsafePointer[T, mut=<True/False>, origin=_]):
    ...
```
so only the `origin` is unbound and not the address space.

martinvuyk: The main motivation behind using `**_` previously was because of not using the default `alignment` parameter that now isn't an issue anymore :thinking:.



> @martinvuyk - we might need to be pedantic here and say:
> 
> ```
> fn foo[T: AnyType](ptr: UnsafePointer[T, mut=<True/False>, origin=_]):
>     ...
> ```
> 
> so only the `origin` is unbound and not the address space.

I think this is a good approach, default the address space and only generalize where it blocks any other work. I'll have to go back and fix a bunch of stuff

NathanSWard (approved):

mojo/stdlib/stdlib/sys/_libc.mojo
```diff
@@ -144,7 +148,7 @@ fn kill(pid: c_int, sig: c_int) -> c_int:
 
 
 @always_inline
-fn pipe(fildes: UnsafePointer[c_int]) -> c_int:
+fn pipe(fildes: UnsafePointer[c_int, mut=False, origin=_]) -> c_int:
```
NathanSWard:
```suggestion
fn pipe(fildes: UnsafePointer[c_int, mut=True, origin=_]) -> c_int:
```

`pipe` requires a mutable pointer [docs](https://linux.die.net/man/2/pipe)

mojo/stdlib/stdlib/sys/_libc.mojo
```diff
@@ -44,7 +44,7 @@ alias FILE_ptr = OpaquePointer
 
 
 @always_inline
-fn fdopen(fd: c_int, mode: UnsafePointer[c_char, **_]) -> FILE_ptr:
+fn fdopen(fd: c_int, mode: UnsafePointer[c_char, origin=_]) -> FILE_ptr:
```
NathanSWard:
```suggestion
fn fdopen(fd: c_int, mode: UnsafePointer[c_char, mut=False, origin=_]) -> FILE_ptr:
```
this can be immutable :)

mojo/stdlib/stdlib/sys/_libc.mojo
```diff
@@ -214,15 +220,13 @@ fn dlsym[
...
-]:
+    name: UnsafePointer[c_char, mut=False, origin=_],
+) -> UnsafePointer[result_type]:
     return external_call["dlsym", UnsafePointer[result_type]](handle, name)
 
 
 fn realpath(
-    path: UnsafePointer[c_char],
+    path: UnsafePointer[c_char, mut=False, origin=_],
     resolved_path: UnsafePointer[c_char] = UnsafePointer[c_char](),
```
NathanSWard:
```suggestion
    resolved_path: UnsafePointer[c_char, mut=True] = UnsafePointer[c_char](),
```
So we can be explicitly about mutability :)

NathanSWard (changes requested):

NathanSWard (approved):

mojo/stdlib/stdlib/sys/_libc.mojo
```diff
@@ -27,7 +27,7 @@ from sys.ffi import c_char, c_int, c_size_t, get_errno
 
 
 @always_inline
-fn free(ptr: UnsafePointer[NoneType, **_]):
+fn free(ptr: UnsafePointer[NoneType, mut=True, **_]):
```
soraros, line 30: Does this work on address space other than generic?

martinvuyk: Idk, I guess it would depend on the hardware? i.e. if we define SRAM as always being the stack and stack pointer not being freeable, and address spaces that aren't generic as always being stack pointers, then yes this should be constrained to generic. But I really don't know.

On another note: Yesterday I tried using memcpy from a stack pointer (a comptime `alias`) to a GPU shared memory address (I had to cast the address space to generic), but it failed and I'm not sure whether its because of using the stack pointer (very probable as I've had other issues with it in GPU).

NathanSWard: FWIW: it was previously unbounded before so the behavior hasn't changed for `free`. Also `UnsafePointer` is getting a bit of a rework over then next couple weeks so we'll likely have to update this function again and we can consider if this should work for all address spaces.

NathanSWard:
!sync

PR #5320, opened 2025-09-16
[stdlib] Improve performance of `_str_iota`

martinvuyk:
Improve performance of `_str_iota`

### DIFF
--- a/mojo/stdlib/stdlib/gpu/mma.mojo
+++ b/mojo/stdlib/stdlib/gpu/mma.mojo
@@ -23,6 +23,8 @@ from sys.info import (
     _cdna_4_or_newer,
     CompilationTarget,
 )
+from io.write import _WriteBufferStack, _TotalWritableBytes
+
 
 from gpu.host._nvidia_cuda import TensorMapSwizzle
 from gpu.mma_operand_descriptor import MMAOperandDescriptor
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/gpu/mma.mojo
```diff
@@ -1780,18 +1786,32 @@ fn wgmma_async[
 
 @always_inline("nodebug")
 fn _str_iota[
-    count: Int, *, prefix: String = String(), sep: String = ", "
+    count: UInt,
+    *,
+    O1: ImmutableOrigin = StaticConstantOrigin,
+    O2: ImmutableOrigin = StaticConstantOrigin,
```
laszlokindrat, line 1792: Unrelated to your PR: I have run into similar situations. Would it make sense to allow default values for infer-only parameters?

laszlokindrat:
@martinvuyk Thanks for the patch! Do you have any measurements of what kind of speedup this delivers?

mojo/stdlib/stdlib/gpu/mma.mojo
martinvuyk: I'm not sure. Seems weird when taking into account that you can't specify a defaulted parameter before a non-defaulted one

martinvuyk:
> @martinvuyk Thanks for the patch! Do you have any measurements of what kind of speedup this delivers?

Nope, but it is obviously faster. Similar to the speedups I achieved for `String.write()`, it's using the same principles

mojo/stdlib/stdlib/gpu/mma.mojo
soraros: > Unrelated to your PR: I have run into similar situations. Would it make sense to allow default values for infer-only parameters?

*I* think the answer is yes. I don't see why we can't have default solution to a unification problem.

> I'm not sure. Seems weird when taking into account that you can't specify a defaulted parameter before a non-defaulted one

Infer-only parameters are filled in *after* all the rest, so the order is already kinda inverted.

laszlokindrat (approved):

laszlokindrat:
!sync

mojo/stdlib/stdlib/gpu/mma.mojo
lattner: Egads, none of this should be required.  You should be able to just write:

```
n _str_iota[
    O1: ImmutableOrigin,
    O2: ImmutableOrigin,
    /,
    count: Int,
    *,
    prefix: StringSlice[O1] = StaticString(),
    sep: StringSlice[O2] = StaticString(", "),
]() -> String
```

We should infer O1/O2 from the argument ... which could be the default arg if not specified.  Right now, I don't think inference is wired into default args correctly

lattner: Also, why are you changing count to UInt?  Please use Int as the generalized integer type

PR #5323, opened 2025-09-17
[stdlib] Fix origins in `memory.mojo`

martinvuyk:
Fix origins in `memory.mojo`. Supersedes  #5316, I managed to reduce the scope of the PR

### DIFF
--- a/max/kernels/src/quantization/qmatmul_k.mojo
+++ b/max/kernels/src/quantization/qmatmul_k.mojo
@@ -1149,7 +1149,11 @@ fn _matmul_Q4_K_columns[
 
     # Fast path for M=1 that avoids materializing the unpacked weights.
     if M == 1:
-        var b_q_bits_ptr = b_tile_ptr[].q_bits.bits.unsafe_ptr()
+        var b_q_bits_ptr = (
+            b_tile_ptr[]
+            .q_bits.bits.unsafe_ptr()
+            .origin_cast[True, MutableAnyOrigin]()
+        )
 
         @parameter
         fn matmul_group_packed(
@@ -1185,7 +1189,7 @@ fn _matmul_Q4_K_columns[
     @__copy_capture(b_tile_ptr, b_q_scales_and_mins_buf, b_q_bits)
     @always_inline
     fn process_rows[tile_m: Int](m: Int):
-        var b_q_bits_ptr = b_q_bits
+        var b_q_bits_ptr = b_q_bits.origin_cast[True, MutableAnyOrigin]()
 
         @parameter
         fn matmul_group_unpacked(
@@ -1391,7 +1395,11 @@ fn _matmul_Q6_K_columns[
 
     # Fast path for M=1 that avoids materializing the unpacked weights.
     if M == 1:
-        var b_q_bits_ptr = b_tile_ptr[].q_bits.bits.unsafe_ptr()
+        var b_q_bits_ptr = (
+            b_tile_ptr[]
+            .q_bits.bits.unsafe_ptr()
+            .origin_cast[True, MutableAnyOrigin]()
+        )
 
         @parameter
         fn matmul_group_packed(
@@ -1419,7 +1427,7 @@ fn _matmul_Q6_K_columns[
     @__copy_capture(b_tile_ptr, b_q_bits)
     @always_inline
     fn process_rows[tile_m: Int](m: Int):
-        var b_q_bits_ptr = b_q_bits
+        var b_q_bits_ptr = b_q_bits.origin_cast[True, MutableAnyOrigin]()
 
         @parameter
         fn matmul_group_unpacked(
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/memory/memory.mojo
```diff
@@ -242,16 +242,17 @@ fn _memcpy_impl(
 
 @always_inline
 fn memcpy[
-    T: AnyType
+    T: AnyType, address_space: AddressSpace
 ](
-    dest: UnsafePointer[T, address_space = AddressSpace.GENERIC, **_],
-    src: UnsafePointer[T, address_space = AddressSpace.GENERIC, **_],
+    dest: UnsafePointer[T, address_space=address_space, **_],
+    src: UnsafePointer[T, address_space=address_space, **_],
```
NathanSWard: Curious - why are we allowing other address spaces other than `AddressSpace.GENERIC`? 
Also `src` should be able to be marked as `mut = False` since we only need to write to `dst`.

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -758,7 +758,7 @@ fn _matmul_group_unpacked[
     group_size: Int,
 ](
     a_q_bits_ptr: UnsafePointer[Int8],
-    mut b_q_bits_ptr: UnsafePointer[UInt8],
+    var b_q_bits_ptr: UnsafePointer[UInt8],
```
NathanSWard: There still seem to be some left over changes from `qmatmul_k.mojo`

NathanSWard (commented):
@martinvuyk - if you want I can sync in the branch (even it its not 100% yet) and then internally push some changes if you want?
Since ideally I want you to get credit for the large portion of the work!

max/kernels/src/quantization/qmatmul_k.mojo
NathanSWard: (Ah, I just saw this was a draft PR) - feel free to ignore my comments if you're not ready to have this reviewed yet!

mojo/stdlib/stdlib/memory/memory.mojo
martinvuyk: >  Also src should be able to be marked as mut = False since we only need to write to dst.

True

> Curious - why are we allowing other address spaces other than AddressSpace.GENERIC?

Actually my doubt was on why are we limiting to only generic address spaces. wouldn't we also want to copy a tensor from one piece of mem to another in e.g. a block's shared memory?

NathanSWard: Totally a valid point -> just for limiting the scope of what this PR is changing, keeping the semantics of the adrress the same is probably best (until we have a use case for other address spaces)

NathanSWard: I can internally fix this up - but this looks great for now! Thank you for the work :)

NathanSWard (approved):

mojo/stdlib/stdlib/memory/memory.mojo
martinvuyk: Sounds good, I think I might open up another one in the future maybe even removing the constraint of them having the same address space. But I'm not sure whether that makes any sense, so yeah it's worth splitting off

NathanSWard:
!sync

mojo/stdlib/stdlib/memory/memory.mojo
```diff
@@ -506,7 +486,7 @@ fn _malloc[
 
 
 @always_inline
-fn _free(ptr: UnsafePointer[_, address_space = AddressSpace.GENERIC, *_, **_]):
+fn _free(ptr: UnsafePointer[_, mut=True, **_]):
```
ConnorGray, line 489: **Question** Why / do we need to remove the `AddressSpace.GENERIC` requirement here?

martinvuyk: Because I don't understand why it was there in the first place given the body of the function also works on GPUs
```mojo
    @parameter
    if is_gpu():
        libc.free(ptr.bitcast[NoneType]())
    else:
        __mlir_op.`pop.aligned_free`(ptr.address)
```
and the libc definition is:
```mojo
fn free(ptr: UnsafePointer[NoneType, **_]):
    external_call["free", NoneType](ptr)
```

I might be wrong and GPUs only allow libc free in generic address spaces though :man_shrugging:

martinvuyk: I can also just revert the change to reduce the scope of the PR, whatever works best for you guys

soraros: FTR, `libc.free` only works on GA before #4405.

martinvuyk: Yep, might have been an oversight :man_shrugging:

PR #5324, opened 2025-09-17
[stdlib] Improve the performance of the `PythonObject` dict constructor

martinvuyk:
Improve the performance of the `PythonObject` dict constructor

### DIFF
--- a/mojo/stdlib/stdlib/python/python_object.mojo
+++ b/mojo/stdlib/stdlib/python/python_object.mojo
@@ -389,15 +389,14 @@ struct PythonObject(
         if not dict_obj_ptr:
             raise Error("internal error: PyDict_New failed")
 
-        for i in range(len(keys)):
-            var key_obj = keys[i].copy().to_python_object()
-            var val_obj = values[i].copy().to_python_object()
+        for key, val in zip(keys, values):
             var result = cpython.PyDict_SetItem(
-                dict_obj_ptr, key_obj._obj_ptr, val_obj._obj_ptr
+                dict_obj_ptr, key._obj_ptr, val._obj_ptr
             )
             if result != 0:
+                # destroy the incorrectly built dict to avoid a memory leak
+                _ = PythonObject(from_owned=dict_obj_ptr)
                 raise Error("internal error: PyDict_SetItem failed")
-
         return PythonObject(from_owned=dict_obj_ptr)
 
     fn __copyinit__(out self, existing: Self):


soraros:
~~Can we use `zip` these days?~~

Actually, do we need this at all? `PythonObject`s are quite cheap to copy (just a pointer). That said, the `copy().to_python_object()` call is a weird/unnecessary.

mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -389,15 +389,20 @@ struct PythonObject(
...
-                dict_obj_ptr, key_obj._obj_ptr, val_obj._obj_ptr
+                dict_obj_ptr, keys[i]._obj_ptr, values[i]._obj_ptr
             )
             if result != 0:
+                # destroy the incorrectly built dict to avoid a memory leak
+                _ = PythonObject(from_owned=dict_obj_ptr)
                 raise Error("internal error: PyDict_SetItem failed")
-
+            keys[i]._obj_ptr = {}
+            values[i]._obj_ptr = {}
```
soraros: I don't think this is correct. The inner object pointer don't have their ref count decreased correctly this way.

martinvuyk: I might be wrong, but my intention is that this essentially moves the python object from the list to the dict so that when the dict goes out of scope it frees the `PythonObject`. The idea is to not trigger the refcount at all and steal ownership of the underlying pointer. But if I don't zero out the obj_ptr of the keys and values then when they go out of scope then their destructor wants to trigger the refcount

martinvuyk:
> ~Can we use `zip` these days?~

This would be a quite interesting example of a use-case for a consuming iterator actually 

> Actually, do we need this at all? `PythonObject`s are quite cheap to copy (just a pointer). That said, the `copy().to_python_object()` call is a weird/unnecessary.

My intention is to not have to increase or decrease the refcounts when the `PythonObject` are already newly allocated just for this constructor call since the list is a `var`

soraros:
The problem is twofold:

1. The `copy` is indeed redundant. There is no reason why we cannot write:

```mojo
        ...
        for key, val in zip(keys, values):
            var result = cpython.PyDict_SetItem(
                dict_obj_ptr, key._obj_ptr, val._obj_ptr
            )
        ...
````

2. We must call `Py_DECREF`, but using consuming iterator would give the wrong semantics.

We want something roughly equivalent to this translated C code:

```c
for (i...) {
  if (PyDict_SetItem(dict, keys[i], vals[i]) < 0) {
    return NULL;
  }
}
```

Note:

- 1) `PyDict_SetItem` increases the reference count of each key and value.
- 2) We can only `Py_DECREF` after each successful `PyDict_SetItem` call; otherwise, the objects might be freed prematurely.

Because of (i), we should not null out the internal pointers, otherwise every object in the lists will leak.
Because of (ii), we should run the destructors (which call `Py_DECREF`) only after the dictionary construction is complete, so moving out of the list would be the wrong choice.

mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -389,15 +389,20 @@ struct PythonObject(
...
         for i in range(len(keys)):
-            var key_obj = keys[i].copy().to_python_object()
-            var val_obj = values[i].copy().to_python_object()
             var result = cpython.PyDict_SetItem(
-                dict_obj_ptr, key_obj._obj_ptr, val_obj._obj_ptr
+                dict_obj_ptr, keys[i]._obj_ptr, values[i]._obj_ptr
             )
             if result != 0:
+                # destroy the incorrectly built dict to avoid a memory leak
+                _ = PythonObject(from_owned=dict_obj_ptr)
```
soraros, line 398: And for the record, this line is good. However, I'm not sure what the policy is regarding internal errors; perhaps we don't even need to handle the `result != 0` case.

martinvuyk: Why would we ignore an error? I think this should raise

martinvuyk:
> * `PyDict_SetItem` increases the reference count of each key and value.

Ah so that is what `This function *does not* steal a reference to value` means. Got it

> * We can only `Py_DECREF` after each successful `PyDict_SetItem` call; otherwise, the objects might be freed prematurely.

Does the new implementation I just pushed make sense ?

mojo/stdlib/stdlib/python/python_object.mojo
soraros: It's kinda a grey area, like how people often ignore the errno of `printf`. Internal error mostly indicates a programming error on our side. I'm not advocating for either side, just IDK.

martinvuyk: I would rather raise the error than silently fail

soraros:
> Does the new implementation I just pushed make sense ?

Yes! Looks great.

mojo/stdlib/stdlib/python/python_object.mojo
soraros: Noted. I will tag you should we discuss the CPython internal error policy on GH.

soraros:
> > * `PyDict_SetItem` increases the reference count of each key and value.
> 
> Ah so that is what `This function *does not* steal a reference to value` means. Got it

Haha, yeah. CPython doc (I simply reproduce the original doc in our docstring) can be a bit obscure at time.

soraros:
CC @laszlokindrat.

soraros:
I forgot about this PR and included a similar fix as part of #5420. Sorry for stepping on your toes, @martinvuyk, and thanks again for the patch.

laszlokindrat:
@martinvuyk Thanks for the patch and sorry for the miscoordination!

PR #5328, opened 2025-09-17
[stdlib] Add `take_items` to `List` with examples.

josiahls:
This allows for moving out the items from the list. For example:
```mojo
    for item in l.take_items():
        print(item)
```

Note, this does not work with zip, you need to do:

```
    var l = List[Optional[UnCopyable]](Optional(UnCopyable(x=1)), Optional(UnCopyable(x=2)), Optional(UnCopyable(x=3)))
    var l2 = List[Optional[Int]](Optional(10), Optional(20), Optional(30))
    var it = zip(l.take_items(), l2.take_items())
    for item in it:
        var v = item[0].take()^
        print('This is v:', v)
        var v2 = item[1].take()^
        print('This is v2:', v2)
```

I had to do a similar thing to dict take_items via adding a reap to dict entry https://github.com/modular/modular/pull/5327

My question is, should "take_items" type iterators always wrap the return elements in Optional? That would clean up the UX for all of these, and should plugin nicer with other iter tools like zip. 

Is there a smarter way to resolve this?

### DIFF
--- a/mojo/stdlib/stdlib/collections/list.mojo
+++ b/mojo/stdlib/stdlib/collections/list.mojo
@@ -93,6 +93,43 @@ struct _ListIter[
         return (iter_len, {iter_len})
 
 
+@fieldwise_init
+struct _TakeListIter[T: Copyable & Movable, origin: Origin[True]](
+    ImplicitlyCopyable, Iterable, Iterator, Movable
+):
+    """Iterator for List that pops the first element per iteration.
+
+    Parameters:
+        T: The type of the elements in the list.
+        origin: The origin of the List
+    """
+
+    alias Element = T  # FIXME(MOCO-2068): shouldn't be needed.
+
+    alias IteratorType[
+        iterable_mut: Bool, //, iterable_origin: Origin[iterable_mut]
+    ]: Iterator = Self
+
+    var src: Pointer[List[Self.Element], origin]
+
+    @always_inline
+    fn __iter__(ref self) -> Self.IteratorType[__origin_of(self)]:
+        return self.copy()
+
+    @always_inline
+    fn __has_next__(self) -> Bool:
+        return len(self.src[]) > 0
+
+    @always_inline
+    fn __next__(mut self) -> Self.Element:
+        return self.src[].pop(0)
+
+    @always_inline
+    fn bounds(self) -> Tuple[Int, Optional[Int]]:
+        var iter_len: Int = len(self.src[])
+        return (iter_len, {iter_len})
+
+
 struct List[T: Copyable & Movable](
     Boolable, Copyable, Defaultable, Iterable, Movable, Sized
 ):
@@ -842,6 +879,34 @@ struct List[T: Copyable & Movable](
         self._len -= 1
 
         return ret_val^
+
+    fn take_items(mut self) -> _TakeListIter[T, __origin_of(self)]:
+        """Returns a mutating iterator that pops items out of the list.
+
+        Examples:
+        ```mojo
+        var l = List[String]("a", "b", "c")
+        for item in l.take_items():
+            # Unlike the default iterator, we
+            # can take ownership of each item.
+            var v = item^
+            print(v)
+        print('List length:', len(l))
+        ```
+
+        This will print:
+
+        ```
+        a
+        b
+        c
+        List length: 0
+        ```
+
+        Returns:
+            A mutating iterator that pops items out of the list.
+        """
+        return _TakeListIter(Pointer(to=self))
 
     fn reserve(mut self, new_capacity: Int):
         """Reserves the requested capacity.
--- a/mojo/stdlib/test/collections/test_list.mojo
+++ b/mojo/stdlib/test/collections/test_list.mojo
@@ -1006,6 +1006,15 @@ def test_list_repr_wrap():
     )
 
 
+def test_list_take_items():
+    var l = List[String]("a", "b", "c")
+    var l2 = l.copy()
+    for item in l.take_items():
+        var v = item^
+        assert_true(v in l2)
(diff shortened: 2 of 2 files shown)


mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -93,6 +93,43 @@ struct _ListIter[
...
+    fn __iter__(ref self) -> Self.IteratorType[__origin_of(self)]:
+        return self.copy()
+
+    @always_inline
+    fn __has_next__(self) -> Bool:
+        return len(self.src[]) > 0
+
+    @always_inline
+    fn __next__(mut self) -> Self.Element:
+        return self.src[].pop(0)
```
bgreni, line 125: I think we'll want a new method on `List` to support this. Doing a front pop on each item could be quite expensive for large collections as it involves moving each item back one spot in the array each time.

https://github.com/modular/modular/blob/main/mojo/stdlib/stdlib/collections/list.mojo#L839-L842

josiahls: That's a good point, I'll try out a method that doesn't require moving everything back a slot.

josiahls: Ok main concern is if we don't move update the list and move all the elements left, then when should it move them left? On iterator destruction? What happens if we don't completely iterate through the elements?

I'm looking at https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain and https://doc.rust-lang.org/std/vec/struct.Drain.html for guidance. 

The way rust does this:
- User can specify a range. If not, borrows the whole vec.
- If the iterator is not fully iterated through the remaining elements, it will drop the remainder. 

> [drain](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain) If the iterator is dropped before being fully consumed, it drops the remaining removed elements
I'm a little fuzzy on when rust decides "the iterator is droppped". Like on iterator destruction?

I'm wondering if this should be a proposal, because I'm noticing multiple moving parts here, plus their Drain struct is very fleshed out.

bgreni: I think the Rust behaviour is the correct way forward

josiahls: I'm writing a proposal today that will try to get a lay of the land with current mutable collections + how this drain behavior will work with iter utils like map and zip + are there any compiler feature that would make this easier.

NathanSWard: Hey @josiahls :)
We just had a discussion about this and are thinking of moving the semantics of `take_item` to behave similarly to rust's `drain` iterator. 

> [drain](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain) If the iterator is dropped before being fully consumed, it drops the remaining removed elements
I'm a little fuzzy on when rust decides "the iterator is droppped". Like on iterator destruction?
Correct, when the iterator is destroyed. (Rust has a `drop()` function which is very similar/synonymous to Mojo's `__del__` function).

Specifically for the `List.take_items` it should likely take a `Slice` argument so a user can specify what they want to take. 
```mojo
var list [1, 2, 3, 4, 5]
for elem in list.take_items(Slice(0, 3)):
    print(elem) # <-- prints 1, 2, 3

assert_equal(len(list), 2)
```

Concerning the "consumes elements when the iterator is dropped, I think we've come to the conclusion that yes this should happen.
Take the following test example:
```mojo
def test_dict_take_items():
    var d: Dict[String, Int] = {"a": 1, "b": 2, "c": 3}
    var iter = d.take_items() # <-- The iterator is "destroyed" here consuming the rest of the elements

    assert_equal(len(d), 0)
```

josiahls: @NathanSWard Sounds good to me!

Is there anything we should do to handle this case? Or is this acceptable that if the user wants to delay the drain's destruction, they accept there will be weird side affects? 
```mojo
def test_dict_take_items():
    var d: Dict[String, Int] = {"a": 1, "b": 2, "c": 3}
    var iter = d.take_items() 

    assert_equal(len(d), 0) # <-- will raise
    _ = iter # User's fault for trying to delay the iterator's destruction here. 
```
Or even stranger:
```mojo
def test_list_take_items():
    var ls: List = ["a", "b", "c"]
    var iter = ls.take_items() 
    for element in iter:
        # Only removed the first item, and avoid moving the others left since that can be slow.
        break 
    
    # The items haven't been moved to the left yet, and the iterator has not be destroyed.
    # I think this may try to pop the location that "a" was moved out of.
    _ = d.pop(0) 

    assert_equal(len(d), 0) # <-- will raise
    _ = iter # User's fault for trying to delay the iterator's destruction here. 
```
If you're onboard with, "The user accepts they are doing something weird in delaying the iterators destruction" then I'll continue working on this.

An additional question for you @NathanSWard is how this plays into `zip` and `map`. I found that I need to wrap the elements in `Optional` so I can `take` the elements out of the resulting tuple. If mojo supports tuple-move unpacking, I don't think we need to wrap the elements in Optional then.

NathanSWard: Apologies for the late reply - I think we may want to hold off on this specific PR until we have a better grasp on what we want the other "take_items" semantics to be on all collections. 
(It'll likely be rust-ish but there are some easy to run into foot-guns due to the lack of exclusive mutability which we need to consider)

josiahls: @NathanSWard Makes sense, I did add a comment taking about a possible roadmap for this https://github.com/modular/modular/pull/5338#discussion_r2365735272 in the proposal

PR #5331, opened 2025-09-18
[stdlib][proposal] Introduce Optional Field Type Proposal

mzaks:


martinvuyk:
Question: Could we just call this `DependentOptional` or something ? It seems to me that this would be useful for a bunch of other cases besides defining fields in a struct

mzaks:
> DependentOptional

Yeah I am unattached to the name. Suggestions are welcome

christoph-schlumpf:
Cool! I was working on a similar proposal based on `InlineArray`. My naming was `StaticNonable`. Reason: It is either statically `None` or can hold a runtime value.

mojo/proposals/optional-field-type.md
```diff
@@ -0,0 +1,260 @@
...
+struct OptionalField[
+    active: Bool, ElementType: Copyable & Movable
+](Copyable, Movable):
+    var field: InlineArray[ElementType, 1 if active else 0]
+
+    fn __init__(out self):
+        constrained[
+            not active, "Constructor only available with no active field"
+        ]()
+        self.field = InlineArray[ElementType, 1 if active else 0](
```
christoph-schlumpf, line 36: I think you can simplify this based on the constraint:
`self.field = InlineArray[ElementType, 0]…`

The same can be applied in other places where you have a constraint on `active` or a `@parameter if active`.

martinvuyk: If we're nit-picking implementation I would say we could add an `alias _Array = InlineArray[ElementType, Int(active)]` and use `Self._Array` everywhere

owenhilyard:
Strong +1 from me. This is the kind of thing we need in order to be able to unify a runtime-sized array (~owning Span), a fixed-capacity but dynamic-length list, a List type which always stores on the heap (better for arrays of large lists), and List with a small size optimization.

Brian-M-J:
I'll just repeat what I said in the Discord for future reference:

Given that a design goal of Mojo is to eventually have user-controlled optimizations, another way to achieve the same effect is through dead field elimination. The same problems with any metaprogramming apply here (like being able to reason about what the code looks like after it is modified).

Example: https://scholar.google.com/scholar?cluster=5016535405140761761&hl=en&as_sdt=2005&sciodt=0,5

josiahls:
Currently working on ffi, and I'm dealing with structs that have fields that are turned on/off via macros. Made a version of the above OptionalField but where the element is trivially passable.
```
@register_passable("trivial")
struct TrivialOptionalField[
    active: Bool, ElementType: AnyTrivialRegType
](Copyable, Movable):
    comptime OptionalElementType = StaticTuple[
        Self.ElementType, 1 if Self.active else 0
    ]
    var field: Self.OptionalElementType
    fn __init__(out self):
        constrained[
            not Self.active, "Constructor only available with no active field"
        ]()
        self.field = Self.OptionalElementType()
    
    fn __init__(out self, var value: Self.ElementType):
        constrained[Self.active, "Constructor only available with active field"]()
        self.field = Self.OptionalElementType(value)

    fn __init__(out self, var value: Optional[Self.ElementType]):
        @parameter
        if Self.active:
            self.field = Self.OptionalElementType(value.take())
        else:
            self.field = Self.OptionalElementType()

    @always_inline
    fn __getitem__(ref self) -> Self.ElementType:
        constrained[
            Self.active, "Field is not active, you should not access it."
        ]()
        return self.field[0]
```

PR #5332, opened 2025-09-18
[stdlib] Replace uses of `raise String(...)`

martinvuyk:
Replace uses of `raise String(...)`. Split off from #4807

This is very inefficient because `Error`'s constructor copies the whole data again.

### DIFF
--- a/max/kernels/src/_cublas/cublas.mojo
+++ b/max/kernels/src/_cublas/cublas.mojo
@@ -35,8 +35,8 @@ alias CUDA_CUBLAS_LIBRARY_PATHS = List[Path](
 )
 
 
-fn _on_error_msg() -> String:
-    return String(
+fn _on_error_msg() -> Error:
+    return Error(
         (
             "Cannot find the cuBLAS libraries. Please make sure that "
             "the CUDA toolkit is installed and that the library path is "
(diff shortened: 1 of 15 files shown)


JoeLoser (approved):

max/kernels/src/_cublas/cublas.mojo
```diff
@@ -35,8 +35,8 @@ alias CUDA_CUBLAS_LIBRARY_PATHS = List[Path](
 )
 
 
-fn _on_error_msg() -> String:
-    return String(
+fn _on_error_msg() -> Error:
```
soraros, line 38: This wouldn't be on error *message* any more, right?

martinvuyk: Huh? It does give back an error message, I don't see the issue

soraros: It gives back a `Error`, no? Not a major issue though.

martinvuyk: Well yeah but all errors kind of are error messages (?). Anyway it's a private function, I don't think it matters that much

JoeLoser:
Given that https://github.com/modular/modular/pull/4807 just landed, can we close this now?

Issue #4790, opened 2025-06-10
[BUG] Incorrect exclusivity error when printing a StringSlice


soraros:
### Bug description

This should work but fails with exclusivity error:

```
fn main():
  print(String("test").removeprefix("")) 
```

martinvuyk:
AFAIK `StringSlice` cannot be implicitly converted to `String` to avoid invisible allocations

soraros:
That makes sense. Maybe the conversion should go the other way around. In any case, this was caught in #4789.

martinvuyk:
Ok but that is correct behavior. There should be an overload for `StringSlice` and no implicit allocations should take place

soraros:
This also doesn't work:
```mojo
fn main():
  print(String("test").removeprefix(""))  # error: argument of '__init__' call allows writing a memory location previously readable through another aliased argument
```

soraros:
Actually, it's more of a problem with `StringSlice` not being `Eq`, since we do have a `EqualityComparable & Stringable` overload for `debug_assert`.

martinvuyk:
> This also doesn't work:
> 
> fn main():
>   print(String("test").removeprefix(""))  # error: argument of '__init__' call allows writing a memory location previously readable through another aliased argument

Yeah I hate that error, have gotten it in a lot of places where it makes no sense. It's more of an error for anonymous origins (you even opened #4742 )

lattner:
I agree the first issue isn't an issue.  I agree the print origin thing is really bad, it looks related to variadic pack handling.

lattner:
Incidentally, while VariadicPack can/should be fixed, there is a deeper problem here.  The issue is that StringSlice allows itself to be unconditionally "writable" with a parametric mutability.  You can try make it force an immutable origin for writability - this would cause an implicit conversion from mutable slice to immutable slice, which will probably make lots of things better.

martinvuyk:
@lattner there is still an issue with the origin, even when using `.get_immutable()`. I opened #4846 to track it

PR #5334, opened 2025-09-18
[stdlib] [NFC] Remove workaround for #4790

martinvuyk:
Remove workaround for #4790

### DIFF
--- a/mojo/stdlib/test/collections/string/test_string.mojo
+++ b/mojo/stdlib/test/collections/string/test_string.mojo
@@ -1038,29 +1038,27 @@ def test_endswith():
     assert_false(str.endswith("llo", 2, 3))
 
 
-# TODO: Remove the explicit conversion to `String` once #4790 is resolved.
 def test_removeprefix():
-    assert_equal(String(String("hello world").removeprefix("")), "hello world")
-    assert_equal(String(String("hello world").removeprefix("hello")), " world")
+    assert_equal(StaticString("hello world").removeprefix(""), "hello world")
+    assert_equal(StaticString("hello world").removeprefix("hello"), " world")
     assert_equal(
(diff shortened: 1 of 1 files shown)


soraros:
Since this is test for string, can we simply use the literals?

martinvuyk:
> Since this is test for string, can we simply use the literals?

`StringLiteral` doesn't have `removeprefix` or `removesuffix` anymore

JoeLoser (approved):
LGTM, thanks!

JoeLoser:
!sync

soraros:
> > Since this is test for string, can we simply use the literals?
> 
> `StringLiteral` doesn't have `removeprefix` or `removesuffix` anymore

I thought they materialise into `String` automatically.

martinvuyk:
> > > Since this is test for string, can we simply use the literals?
> > 
> > 
> > `StringLiteral` doesn't have `removeprefix` or `removesuffix` anymore
> 
> I thought they materialise into `String` automatically.

If you do:
```mojo
var a = "1"
a. # gives you String's API
```
but with any materializable type like `IntLiteral` and `StringLiteral`
```mojo
"1". # gives you StringLiteral API
1. # gives you IntLiteral API
```

soraros:
Ah, right. Makes sense, `NumLiteral` also works this way. It would be nice to have the doc explaining this to pin the behaviour down though, shall we write a feature request?

martinvuyk:
> Ah, right. Makes sense, `NumLiteral` also works this way. It would be nice to have the doc explaining this to pin the behaviour down though, shall we write a feature request?

Idk to me it's quite obvious since you haven't materialized it yet at runtime. It's the same behavior if you declare them as an `alias`

PR #5335, opened 2025-09-18
[stdlib] Change `assert_equal` and `assert_not_equal` to use `Writable`

martinvuyk:
Change `assert_equal` and `assert_not_equal` to use `Writable`. Split off from #4793

### DIFF
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -672,6 +672,9 @@ what we publish.
 - `StringableRaising` has been deprecated and its usages in the stdlib have
   been removed.
 
+- `assert_equal` and `assert_not_equal` now take in `Writable` arguments instead
+  of `Stringable` ones.
+
 ### Tooling changes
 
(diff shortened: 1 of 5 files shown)


soraros:
I think the new policy is to make `__str__`/`write_to` always `@no_inline`? @JoeLoser.

JoeLoser:
@NathanSWard curious to get your thoughts on this as you're working in this area on Writable.

NathanSWard:
> @NathanSWard curious to get your thoughts on this as you're working in this area on Writable.

Oh totally, this is definitely a +1 for me.
Any of the functions we have in the stdlib that are bound on either `T: Stringable` or `T: Representable` I'm going to eventually add overloads of `T: Writable` for to eventually deprecate `Stringable & Representable` entirely :)

NathanSWard:
Closing this PR as we actually have overload for these functions that work with `Writable` arguments :)

PR #5338, opened 2025-09-18
[stdlib] [proposal] Introduce proposal for draining collections

josiahls:
[Introduce proposal for drain collection take_items support for basic collections.](https://github.com/modular/modular/pull/5338/commits/8b8c7b16b5212ce5af13289611aa7057395dab9f)

Note, I'm coming from python + c++ background. My comments on rust, I would like someone with past experience to check that that is accurate.

### DIFF
--- a/mojo/proposals/drain-collection-support.md
+++ b/mojo/proposals/drain-collection-support.md
@@ -1,0 +1,455 @@
+# Draining Collections in Mojo
+
+Josiah Laivins, September 18, 2025
+
+Status: Proposal
+
+This doc describes the design and behavior of draining collections in Mojo. This
+proposes the `take_items` as a standard way of moving elements efficiently out
+of their respective collections.  It walks through concerns related to the drain
+iterator vs the source collection, and weird behavior that can come out of
+interacting with collections with active drain iterators. Additionally it
+explains a workaround for tuple unpacking using `Optional`. It also goes through
+ compatibility with iter tools like `zip`.
+
+## Motivation
+
+Mojo has changed default copy behavior from
+[ImplicitlyCopyable to ExplicitlyCopyable][1]. This has surfaced more clearly
+places in the stdlib where we want to be moving values, not copying them.
+
+One major part of the stdlib where copies occur is when we work with
+collections. Moving elements out of collections involves clever, but unstable
+workarounds to avoid copying elements as we move them out of the collection.
+
+## Background
+
+The example below shows that prior to making ExplicitlyCopyable the default,
+we were copying values when taking ownership of items in a collection.
+
+Given struct `Foo` that is (Explicitly) `Copyable` and `Movable`. Also
+given 2 example functions `take_ownership`, one that takes a single `Foo`, and
+one that takes a `String` and a `Foo` (as a key value pair).
+
+```mojo
+from collections import Set
+from hashlib import Hasher
+# Foo is used as an example so we can print the copies.
+@fieldwise_init
+struct Foo(Copyable & Movable & Hashable & EqualityComparable):
+    var x: String
+
+    fn __hash__[H: Hasher](self, mut hasher: H):
+        hasher.update(self.x)
+
+    fn __eq__(self, other: Self) -> Bool:
+        return self.x == other.x
+
+    fn __copyinit__(out self, other: Self):
+        self.x = other.x
+        print('Foo.',self.x,' copied',sep='')
+
+fn take_ownership(var item: Foo):
+    print('Taking ownership of item:', item.x)
+
+fn take_ownership(var key: String, var item: Foo):
+    print('Taking ownership of: (key) ',key,' item:', item.x)
+```
+
+The straight forward way to interface with functions that take ownership
+of elements in collections involves copying the elements.
+
+```mojo
+
+fn test_copyable_a_iteration():
+    print('Testing list')
+    var l = List[Foo](Foo(x="a"), Foo(x="b"), Foo(x="c"))
+    for ref item in l:
+        take_ownership(item.copy()) # Must be copied, not moved (via ^)
+
+    print('Testing set')
+    var set: Set[Foo] = {Foo(x="a"), Foo(x="b"), Foo(x="c")}
+    for ref item in set:
+        take_ownership(item.copy()) # Must be copied, not moved (via ^)
+
+    print('Testing dict')
+    var dict: Dict[String, Foo] = {"a": Foo(x="a"), "b": Foo(x="b"), "c": 
+                                   Foo(x="c")}
+    for ref item in dict.items():
+        # Must be copied, not moved (via ^)
+        take_ownership(item.key.copy(), item.value.copy()) 
+
+```
+
+Collections `List` and `Set` and `Dict` all have `pop`/`popitem` methods that
+can move popped values (as of this writing `Set.pop` does a copy still).
+
+How would we do this currently? We may try a while loop:
+
+```mojo
+fn test_while_loop_moving_items_out_of_collection():
+    print('Testing list')
+    var l = List[Foo](Foo(x="a"), Foo(x="b"), Foo(x="c"))
+    while l:
+        # List.pop will do a full collection move-left for every call to pop(0)
+        take_ownership(l.pop(0))
+
+    print('Testing set')
+    var set: Set[Foo] = {Foo(x="a"), Foo(x="b"), Foo(x="c")}
+    while set:
+        # Set.pop calls require raising context. Hashable items get copied
+        # since it copies the reference in the underlying `_DictKeyIter` 
+        try:
+            take_ownership(set.pop())
+        except:
+            print('Set is empty')
+
+    print('Testing dict')
+    var dict: Dict[String, Foo] = {"a": Foo(x="a"), "b": Foo(x="b"), 
+                                   "c": Foo(x="c")}
+    print('beginning iteration')
+    while dict:
+        try:
+            # Dict.popitem calls require raising context. 
+            # Executes a move, however must be in LIFO order.
+            var item = dict.popitem()
+            # DictEntry does not support move-based unpacking, so we need to 
+            # copy the key, then take the value.
+            take_ownership(item.key.copy(), item^.reap_value())
+        except:
+            print('Dict is empty')
+```
+
+This is made more complicated by combining with iter tools like `zip`:
+
+```mojo
+fn test_zip_from_2_lists():
+    var keys = List[String]("a", "b", "c")
+    var values = List[Foo](Foo(x="d"), Foo(x="e"), Foo(x="f"))
+    var it = zip(keys^, values^)
+    for item in it:
+        # Copies happen during zip iteration, then additionally
+        # must happen when passing to take_ownership. We are copying twice per 
+        # iter!
+        take_ownership(item[0].copy(), item[1].copy())
+```
+
+This is a common pattern, given a list of keys and a list of values, we want to
+zip them together into a map-like object (Dict, Tree, etc.).
+
+### How do other languages handle this?
+
+Rust
+
+> TODO: Would like input from people with Rust heavy backgrounds to clarify this
+> is how Rust drain actually works, particularly "when do the source collections
+> get cleared"?
+
+Rust has a struct called
+[`vec/Drain`][2] and
+[`hashmap/Drain`][3]
+that is used by collections like `Vec` and `HashMap` to drain the collection.
+
+[Vec.drain](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain)
+
+```rust
+let mut v = vec![1, 2, 3];
+let u: Vec<_> = v.drain(1..).collect();
+assert_eq!(v, &[1]);
+assert_eq!(u, &[2, 3]);
+
+// A full range clears the vector, like `clear()` does
+v.drain(..);
+assert_eq!(v, &[]);
+```
+
+[HashMap.drain](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.drain)
+
+```rust
+use std::collections::HashMap;
+
+let mut a = HashMap::new();
+a.insert(1, "a");
+a.insert(2, "b");
+
+for (k, v) in a.drain().take(1) {
+    assert!(k == 1 || k == 2);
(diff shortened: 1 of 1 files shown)


mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+
+```mojo
+fn test_zip_from_2_lists():
+    var keys = List[String]("a", "b", "c")
+    var values = List[Foo](Foo(x="d"), Foo(x="e"), Foo(x="f"))
+    var it = zip(keys^, values^)
+    for item in it:
+        # Copies happen during zip iteration, then additionally
+        # must happen when passing to take_ownership. We are copying twice per 
+        # iter!
```
NathanSWard, line 133: Yep! Though a lot of this is due to the lack of language support we have to easily express ownership in a generic manner via origins. So ideally when the language supports it, we will be able to easily remove the unnecessary copies :D

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+For sequential collections like Vec:
+
+- `drain(..)` clears the full collection either on full iteration or when the
+iterator is dropped (destroyed).
+- `drain(start..end)` drains a range of elements either on full iteration or when
+the iterator is dropped (destroyed).
+
+For hashable collections like HashMap:
```
NathanSWard, line 190: Yep! Or for non-sequential/unordered collections where taking a specific "range" doesn't really make sense - their drain function won't support a range (synonymous to Mojo's slice)

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+
+- `drain().take(n)` drains the first `n` elements. *Clears the entire collection
+on destruction.*
```
NathanSWard: `drain().take(n)` will drain the entire collection, the `take(n)` iterator adaptor will only have the `Drain` iterator type return its first `n` elements. `take` is orthogonal to how `Drain` behaves.

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+
+### User Experience Cases
+
+In all cases, `ref` must be used to get mutable access to the item. `for` loops
+by default iterate over immutable references (even if the source collection is
+mutable).
+
+In all cases the call to `take_items` takes mutable reference of the underlying
+collection, and clears the source collection immediately on destruction of the
+iterator. Cases such as `take_items(Slice(..))` takes mutable reference of the
```
NathanSWard, line 231: FWIW: I wonder if taking a `range` is better? Since the only nice part about `slice` is the ability to use the `[:]` syntax with it. But the Mojo parser only supports the `[:]` syntax in combination with square brackets.

NathanSWard: So you'll be able to write `list.drain(range(1, 10))` which is much cleaner to read. Also we probably should not support backwards ranges? That feels like it will drive the complexity of the implementation to be too large.

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+```mojo
+collection = List("a", "b", "c") or Dict("a": 1, "b": 2, "c": 3) or \ 
+Set(1, 2, 3)
+take_iter = collection.take_items()
+assert(len(collection) == 3) 
+
+for ref item in take_iter:
+    # Do something with item or break early
+    # Note: What happens if we do `collection.pop(0)` here? Is this something 
+    # to worry about?
```
NathanSWard, line 251: Yes, this is _huge_ concern. And IMO it's a blocker for this proposal until we can figure out a solution on how to have the iterator take "exclusive mutability" to the entire collection. 

Without this, writing `collection.pop(0)` while the iterator is still alive would be UB in some cases depending on the implementation.

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+in tuples such as in `zip`. The examples below, wrap the elements in `Optional`
+so we can `take` the elements out. This allows the elements to be moved out of
+tuples without deinitializing the tuple overall. As part of this proposal a
+method such as `DictEntry.reap` will return a
+`Tuple[Optional[Key], Optional[Value]]` that allows each element to be
+independently moved out of the tuple without deinitializing the tuple overall.
+
+- Ideally above would not be needed if structs (with a unpackable trait) and
+tuples supported move unpacking. Reference issue
+[#5330](https://github.com/modular/modular/issues/5330).
```
NathanSWard, line 268: Move unpacking is a thing we're hopping to get soon!!

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+
+The basic case is a complete drain of each respective collection.
+
+```mojo
+def basic_user_experience_cases():
+    print('Testing list')
+    var l = List[Foo](Foo(x="a"), Foo(x="b"), Foo(x="c"))
+    for ref item in l.take_items():
```
NathanSWard: I don't believe you want `ref` here. Since `take_items` returns "owned" values that is it draining from the collection.


```suggestion
    for item in l.take_items():
```

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+    for ref item in dict.take_items():
+        # Dict take_items returns the DictEntry struct, which does not 
+        # automatically unpack itself.
+        # Ideally mojo would support `__unpack__` like in #5330
+        # However we can define method that returns a 
+        # Tuple[Optional[Key], Optional[Value]]
+        # using optional allows for moving the key and value out of the item 
+        # without deinitializing the item overall.
+        var key_value = item.reap()
+        take_ownership(key_value[0].take(), key_value[1].take())
```
NathanSWard, line 301: Ideally we could write
`var (key, value) = item.reap()`

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+    }
+    for ref item in dict.take_items():
+      break # Iterator clears the remaining elements on destruction
+    assert_equal(len(dict), 0)
+```
+
+## Questions
+
+**Do we think Mojo's Trait API is flexible enough to make Drain/TakeIterator a
+trait?**
```
NathanSWard, line 357: I don't think there is much value in making Drain/TakeIterator a trait. 
The only traits you should need to work with are `Iterator` and `Iterable`. `take_items` is just a way to get a specific kind of iterator you can then pass to generic functions.
```
fn foo(iter: Some[Iterable]):
    ...

var list = [1, 2, 3]
foo(list.take_items(range(1, 2)))
```

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+
+**Should we allow the user to explicitly keep the remaining elements?**
```
NathanSWard, line 359: I think this is definitly something to consider, but not a must have for a first iteration. I would only push for it if there are enough uses cases for it 😄

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+```mojo
+    print('Testing list')
+    var l = List[Foo](Foo(x="a"), Foo(x="b"), Foo(x="c"))
+    # Not sure what the iterator should do for keep_rest in this case.
+    # When do we shift the remaining elements left?
+    drain_iter = l.take_items()
+    for ref item in drain_iter:
+        break 
+    drain_iter.keep_rest()
+    assert_equal(len(l), 2)
```
NathanSWard: `keep_rest` should be a consuming function.
```mojo
fn keep_rest(deinit self):
    ...
```
That way when its called you're forced to release the iterator and never touch it again
```mojo
drain_iter^.keep_rest()
```

mojo/proposals/drain-collection-support.md
```diff
@@ -0,0 +1,452 @@
...
+Note, these edge cases can be partially avoided if the call to `take_items`
+instantly takes ownership of the underlying collection before any iteration is
+expected to occur.
+
+```mojo
+def test_dict_take_items():
+    var d: Dict[String, Int] = {"a": 1, "b": 2, "c": 3}
+    var iter = d.take_items() 
+
+    assert_equal(len(d), 0) # <-- will raise
```
NathanSWard: Ideally this should be a compile error

mojo/proposals/drain-collection-support.md
josiahls: This makes sense to me. Supporting forward for now would keep things simple.

mojo/proposals/drain-collection-support.md
josiahls: My main question though, should we even have a `reap` method? Should the take iterator just return a `Tuple[Optional[K],Optional[V]]` instead of a dict entry? I guess returning a dict entry should at least be consistent, I'm just not a fan of the extra step here. Ofcourse, this could just be solved later with a `__unpack__ / Unpackable` trait at some point so that DictEntry could automatically unpack if that syntax is used.

mojo/proposals/drain-collection-support.md
josiahls: I've tested zip with a take_items method call from list, it looks we don't need to make any changes to the zip functions itself to avoid copies. This would require the `take_items` to wrap the elements in `Optional` though to move (`take`) them back out of the resulting tuples cleanly.

mojo/proposals/drain-collection-support.md
josiahls: Yeah I agree, for now I personally don't have a need for it.

mojo/proposals/drain-collection-support.md
josiahls: Right, whether mojo supports xor mut, or we do `d^.take_items()` this would be a compiler error. Updating the comment here.

mojo/proposals/drain-collection-support.md
josiahls: Right, bad wording on my part, if i understand correctly `take` returns n elements, then the drain will drop the remaining elements on destruction, which id probably expect to happen when take's iteration loop completes.

mojo/proposals/drain-collection-support.md
josiahls: Makes sense, this would be dependent on xor mut presumably since the elements remain in `List`. I think the more immediate implementation of `l^.take_items()` will make `keep_rest ` not practical.

Assuming you mean above, but just for completeness: 
If you mean `keep_rest` would just avoid draining the remaining elements in drain, I don't see the utility in that since the remaining elements are kept until the iterators destruction anyways. My understanding with the rust drain implementation is that the purpose of `keep_rest` is for the benefit of the source collection e.g. when the drain is destroyed, don't clear the remaining elements in the source collection.

mojo/proposals/drain-collection-support.md
josiahls: I'm thinking we should break this into 2 phases:

Phase 1:
- Before xor mut
- Before tuple move unpacking
```mojo
ls = ["a","b","c"] 

drain = ls^.take_items() # Consume the collection
for el in drain:
     # Compiler error since we are moving / deiniting the list in take_items.
     # Avoids accidental UB.
    ls.pop(0)
```
In a typical loop we receive elements wrapped in `Optional`
```mojo
for el in drain:
    # Elements are wrapped in Optional to allow for compat with zip / tuple making functions
    consumer(el.take())
```
With zip:
```mojo
for pair in zip(drain_1, drain_2):
    # Elements are wrapped in Optional to allow for compat with zip / tuple making functions
    consumer(pair[0].take(), pair[1].take())
```
We could alternatively, require a more explicit optional wrap, that way the transtion to phase 2
is less disruptive.
```mojo
# or map(drain_1,Optional), I haven't tested whether this works right now.
for pair in zip(map(drain_1,wrap_optional), map(drain_2,wrap_optional)):
    # Elements are wrapped in Optional, but instead we just do this compositionally
    consumer(pair[0].take(), pair[1].take())
```

Phase 2.a:
- xor mut support
```mojo
ls = ["a","b","c"] 

drain = ls.take_items() # `fn take_items(xor self) -> _TakeItemIter(self)`
for el in drain:
     # Compiler error since drain has exclusive access to ls's memory
    ls.pop(0)
    break
# (assuming no compiler error above)
# We can support keep_rest since `ls` has not been deinit, and its elements
# have simply just been exclusively borrowed.
drain.keep_rest()
assert_equal(ls.pop(0), "b")
```

Phase 2.b:
- tuple / Unpackable move unpacking
```mojo
ls = ["a","b","c"] 

drain = ls.take_items() # `fn take_items(xor self) -> _TakeItemIter(self)`
for el in drain:
     # Don't wrap in optional since moving elements out of tuples would be trivial
    consume(el^)
```
With zip:
```mojo
for left,right in zip(drain_1, drain_2):
     # Don't wrap in optional since moving elements out of tuples would be trivial
    consumer(left^,right^)
```

josiahls: Phase 1:
Basic idea is that "take_items / iter" just severs any association with the source collection after the `take_items` call.

The alternative to:
```mojo
ls = ["a","b","c"] 

drain = ls^.take_items() # Consume the collection
for el in drain:
     # Compiler error since we are moving / deiniting the list in take_items.
     # Avoids accidental UB.
    ls.pop(0)
```
Would be:
```mojo
ls = ["a","b","c"] 

drain = ls.take_items() # Call `ls.steal_data` here, don't deinit list
for el in drain:
    # Would compile but the list would be clearly empty here
    ls.pop(0)
```
and for slices
```mojo
# Moves that range out of the list at this call and update the len / pointers / lshifts
# Move elements at indicies 1,2 at the call site here, don't deinit list.
drain = ls.take_items(range(1,3)) 
for el in drain:
    ls.pop(0) # would be "a"
```

christoph-schlumpf:
Hi @josiahls,
does the newly introduced `IterableOwned` solve this proposal or part of it?
https://github.com/modular/modular/blob/e874bfcf0cbf80719bde8a308cc3b18a9f41ec5e/mojo/stdlib/std/iter/__init__.mojo#L90

PR #5342, opened 2025-09-19
[stdlib] [NFC] Add `debug_assert`s to new `Span` unsafe methods

martinvuyk:
Add `debug_assert`s to new `Span` unsafe methods

### DIFF
--- a/mojo/stdlib/stdlib/builtin/debug_assert.mojo
+++ b/mojo/stdlib/stdlib/builtin/debug_assert.mojo
@@ -66,7 +66,7 @@ fn debug_assert[
     assert_mode: StaticString = "none",
     *Ts: Writable,
     cpu_only: Bool = False,
-](*messages: *Ts):
+](*messages: *Ts, location: Optional[_SourceLocation] = None):
     """Asserts that the condition is true at run time.
 
     If the condition is false, the assertion displays the given message and
@@ -147,6 +147,7 @@ fn debug_assert[
     Args:
         messages: A set of [`Writable`](/mojo/stdlib/utils/write/Writable/)
             arguments to convert to a `String` message.
+        location: The location of the error (defaults to `__call_location`).
     """
 
     @parameter
@@ -162,7 +163,9 @@ fn debug_assert[
 
         message.nul_terminate()
 
-        _debug_assert_msg(message.data, message.pos, __call_location())
+        _debug_assert_msg(
+            message.data, message.pos, location.or_else(__call_location())
+        )
 
 
 @always_inline
@@ -171,7 +174,7 @@ fn debug_assert[
     *Ts: Writable,
     cpu_only: Bool = False,
     _use_compiler_assume: Bool = False,
-](cond: Bool, *messages: *Ts):
+](cond: Bool, *messages: *Ts, location: Optional[_SourceLocation] = None):
     """Asserts that the condition is true at run time.
 
     If the condition is false, the assertion displays the given message and
@@ -255,6 +258,7 @@ fn debug_assert[
         cond: The bool value to assert.
         messages: A set of [`Writable`](/mojo/stdlib/utils/write/Writable/)
             arguments to convert to a `String` message.
+        location: The location of the error (defaults to `__call_location`).
     """
 
     @parameter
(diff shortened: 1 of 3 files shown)


mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -402,19 +402,20 @@ struct Span[
...
         Args:
-            index: The index of the element to get.
+            idx: The index of the element to get.
 
-        ### Safety
-            * This function does not do bounds checking and assumes the provided
-            index is `< len(self)`. Not upholding this contract will result in
-            undefined behavior.
-            * This function does not support wraparound for negative indices.
+        Safety:
```
NathanSWard, line 413: The route we're going for safety docs is the `###` (header 3) direction.
It's not unified across the codebase yet, but were going to be using `### Safety` (I'll need to double check w/ the team to make sure I'm not making this up lol)

mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -535,10 +536,13 @@ struct Span[
             a: The first element's index.
             b: The second element's index.
 
-        ## Safety:
-            * Both `a` and `b` must be `< len(self)`.
-            * `a` cannot be equal to `b`.
+        Safety:
```
NathanSWard, line 549:
```suggestion
        ### Safety
```

mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -747,8 +751,9 @@ struct Span[
             offset: The starting offset of the subspan (self._data + offset).
             length: The length of the new subspan.
 
-        ### Safety
+        Safety:
             This function does not do bounds checking and assumes the current
             span contains the specified subspan.
         """
+        debug_assert(0 <= offset + length < len(self), "Index out of bounds.")
```
NathanSWard: This is a bit hard to read IMO as to what exactly it's checking.
Since we need to check
1. `offset` is within bounds (since offset could technically be negative) 
2. `offset + length` is within bounds
(That may be what this expression is doing but its a bit hard to read)

```suggestion
        debug_assert(0 <= offset < len(self), "subspan out of bounds.")
        debug_assert(0 <= (offset + length) < len(self), "subspan out of bounds.")
```

mojo/stdlib/stdlib/memory/span.mojo
martinvuyk: For me this is part of an older request and discussion (see #3235). There were also some separate discussions in specific PRs separate from that request. I remember @ConnorGray and @JoeLoser also being involved. IMO we should allow custom section headers to be formatted as markdown to avoid having to create these inconsistent "standards"

mojo/stdlib/stdlib/memory/span.mojo
martinvuyk: Yeah the second assert was missing and the error message is better, thanks

mojo/stdlib/stdlib/memory/span.mojo
NathanSWard: Sounds good. Though in the meantime any new doc-strings we have should be consistent - but what is "consistent" not sure. I'll ask the team and then we can append this onto the current style guide doc.

mojo/stdlib/stdlib/builtin/sort.mojo
```diff
@@ -375,7 +376,8 @@ fn _partition[
     var right = size - 2
 
     var pivot_index = size - 1
-    span.unsafe_swap_elements(pivot, pivot_index)
+    if pivot != pivot_index:
+        span.unsafe_swap_elements(pivot, pivot_index)
```
NathanSWard, line 380: Look at that - already catching bugs lol

NathanSWard (approved):

mojo/stdlib/stdlib/builtin/debug_assert.mojo
```diff
@@ -66,7 +66,7 @@ fn debug_assert[
     assert_mode: StaticString = "none",
     *Ts: Writable,
     cpu_only: Bool = False,
-](*messages: *Ts):
+](*messages: *Ts, location: Optional[_SourceLocation] = None):
```
NathanSWard, line 69: What's the purpose of adding this?
The function is `@always_inline` so any callsites will use their source location. 
Were you seeing different behavior by chance?

martinvuyk: When the `debug_assert` is called from within a struct's function it doesn't indicate where that function was called from. So in the case of wanting to catch sort's bugs I needed to have the specific location where they were called. This functionality was already present in `debug_assert` but it was (I assume accidentally) removed some time back during some big refactor by Jack

NathanSWard:
!sync

PR #5345, opened 2025-09-21
[stdlib] Use TestSuite in collections, compile, complex, hashlib and io modules

bgreni:
Migrate to using the TestSuite pattern in the remainder of the collections tests, as well as the compile, complex, hashlib and io module tests.


mojo/stdlib/test/collections/test_list_getitem_invalid_index_int.mojo
```diff
@@ -23,5 +25,8 @@ fn test_fail_list_index():
     print("is never reached")
 
 
-fn main():
-    test_fail_list_index()
+def main():
```
NathanSWard: This file is a bit special as it actually tests that the resulting executable asserts and exits early. We should probably have some sort of way to test that a function "fails" instead of succeeds. For now though, files like these probably shouldn't use `TestSuite`.

mojo/stdlib/test/collections/test_list_getitem_invalid_index_uint.mojo
```diff
@@ -23,5 +25,9 @@ fn test_fail_list_index_uint():
     print("is never reached")
 
 
-fn main():
-    test_fail_list_index_uint()
+def main():
```
NathanSWard: Ditto not using test suite for this file.

NathanSWard:
Thank you for the quick work on this! 😄  Looks good other than the two mentioned files which should probably not use `TestSuite` for now.

mojo/stdlib/test/collections/test_list_getitem_invalid_index_int.mojo
bgreni: Ah right I was wonder about that

JoeLoser:
!sync

JoeLoser:
!sync

PR #5352, opened 2025-09-23
[stdlib] [kernels] Update NVIDIA Blackwell values and add reference

martinvuyk:
Update NVIDIA Blackwell values and add reference.

If I'm reading these wrong, or there is a reason why some values were so different, please correct me.

### DIFF
--- a/mojo/stdlib/stdlib/gpu/host/info.mojo
+++ b/mojo/stdlib/stdlib/gpu/host/info.mojo
@@ -582,6 +582,18 @@ alias H100 = GPUInfo(
     max_thread_block_size=1024,
 )
 
+# ===-----------------------------------------------------------------------===#
+# Blackwell constants
+# ===-----------------------------------------------------------------------===#
+
+# https://docs.nvidia.com/cuda/blackwell-tuning-guide/index.html#occupancy
+alias _blackwell_warp_size = 32
+alias _blackwell_shared_mem_10_0 = 228 * _KB
+alias _blackwell_shared_mem_12_0 = 100 * _KB
+alias _blackwell_max_registers_per_block = 64 * 1024
+alias _blackwell_max_threads_per_sm_10_0 = 64 * _blackwell_warp_size
+alias _blackwell_max_threads_per_sm_12_0 = 48 * _blackwell_warp_size
+alias _blackwell_max_thread_blocks_per_sm = 32
 
 # ===-----------------------------------------------------------------------===#
 # B100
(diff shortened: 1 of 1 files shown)


abduld:
!sync

mojo/stdlib/stdlib/gpu/host/info.mojo
```diff
@@ -582,6 +582,18 @@ alias H100 = GPUInfo(
...
+# ===-----------------------------------------------------------------------===#
+
+# https://docs.nvidia.com/cuda/blackwell-tuning-guide/index.html#occupancy
+alias _blackwell_warp_size = 32
+alias _blackwell_shared_mem_10 = 227 * _KB
+alias _blackwell_shared_mem_12 = 99 * _KB
+alias _blackwell_max_registers_per_block = 64 * _KB
+alias _blackwell_max_threads_per_sm_10 = 64 * _blackwell_warp_size
+alias _blackwell_max_threads_per_sm_12 = 48 * _blackwell_warp_size
+alias _blackwell_max_thread_blocks_per_sm = 32
```
hengjiew, line 596: Thanks for PR! A few comments:
* I see you're accommodating the 1KB L1 reservation. We subtract it in kernel launch so the values here should still be 228 and 100 KB.
* 64*1024 is the max register count, not in KB as memory notion. The original 65536 should be fine.
* Nit: can you please use 100/120 for 10/12? Thanks.

martinvuyk: > * I see you're accommodating the 1KB L1 reservation. We subtract it in kernel launch so the values here should still be 228 and 100 KB.

I've added back the 1KB, but I wonder whether we can always assume that it's only 1KB for different vendors

> * 64*1024 is the max register count, not in KB as memory notion. The original 65536 should be fine.

Yeah I missed that. But now that I actually read `64K 32-bit registers` I'm wondering whether this struct field shouldn't actually also have a register size field. Do all GPUs have/will have 32-bit registers?

> * Nit: can you please use 100/120 for 10/12? Thanks.

Imma go with `10_0` because it's supposed to be `10.0` and `12.0`

hengjiew: > I wonder whether we can always assume that it's only 1KB for different vendors

We can't. It should arch/vendor specific.

> Do all GPUs have/will have 32-bit registers?

It has been 4B so far for Nvidia GPUs.

> Imma go with 10_0 because it's supposed to be 10.0 and 12.0

👍🏼

hengjiew (approved):
Thanks! LGTM!

abduld:
!sync

PR #5360, opened 2025-09-24
[stdlib] [NFC] Nit-pick docstrings and make magic numbers clearer in `gpu/host/info.mojo`

martinvuyk:
Nit-pick docstrings and make magic numbers clearer in `gpu/host/info.mojo`


abduld:
!sync

PR #5369, opened 2025-09-26
[stdlib] Add bit_not function for Int type

sergiogallegos:
- Add missing bit_not function for Int type in bit module
- Add comprehensive tests for the new function
- Follows the same pattern as other bit manipulation functions
- Completes the bit_not API by providing both Int and SIMD variants

### DIFF
--- a/mojo/stdlib/stdlib/bit/bit.mojo
+++ b/mojo/stdlib/stdlib/bit/bit.mojo
@@ -274,6 +274,19 @@ fn pop_count[
 
 
 @always_inline("nodebug")
+fn bit_not(val: Int) -> Int:
+    """Performs a bitwise NOT operation on an integer value.
+
+    Args:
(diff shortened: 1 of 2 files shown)


gryznar:
You can surround `bit_not` and `Int` in title with backticks for better representation

sergiogallegos:

soraros:
I also don't think we need this. https://github.com/modular/modular/pull/4210#pullrequestreview-2730545254

NathanSWard:
@sergiogallegos Ty for this PR!

So currently `Int` and `UInt` are going under a re-write which will make them both aliases of `SIMD`. What this means is that `bit_not` will automatically work for `Int` since there is already an implementation of it for `SIMD` types. We'll probably want to hold off on this PR for those larger sets of changes - but we greatly appreciate the PR none the less!

PR #5374, opened 2025-09-28
[stdlib] Add variadic hypot function for multiple arguments

sergiogallegos:
- Add hypot(*values: Float64) -> Float64 for variadic arguments
- Add hypot(l: List[Float64, *_], /) -> Float64 for list arguments
- Add comprehensive tests covering all edge cases
- Update imports in test file
- Matches Python's math.hypot(*coordinates) behavior
- Fixes TODO: implement for variadic inputs as Python

Examples:
- hypot(3.0, 4.0, 5.0) returns ~7.071
- hypot(3.0) returns 3.0
- hypot() returns 0.0

### DIFF
--- a/mojo/stdlib/stdlib/math/math.mojo
+++ b/mojo/stdlib/stdlib/math/math.mojo
@@ -2017,6 +2017,92 @@ fn hypot[
     return _simd_apply[_float32_dispatch, dtype, width](arg0, arg1)
 
 
+fn hypot(l: List[Float64, *_], /) -> Float64:
+    """Computes the hypot of a list of floating-point values.
+
+    This function computes the Euclidean norm (square root of sum of squares)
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/math/math.mojo
```diff
@@ -2017,6 +2017,92 @@ fn hypot[
     return _simd_apply[_float32_dispatch, dtype, width](arg0, arg1)
 
 
+fn hypot(l: List[Float64, *_], /) -> Float64:
```
abduld, line 2020: can you make this operate on an Iteratable instead of jsut a list

soraros: Also, we recently removed the reg-type hint parameter from `List`, one could simply use `List[Float64]` these days.

Issue #5111, opened 2025-07-29
[Feature Request] Being able to run `sin` and/or `cos` at compile time + implement fallback for GPUs


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

```mojo
from math import pi, sin

fn main():
    alias s_45 = sin(pi/4)
    alias s_60 = sin(pi/3)
    alias s_30 = sin(pi/6)
    print(s_45, s_30, s_60)
```

### What is your motivation for this change?

Pre-computing expensive things like sin and cos for lookup tables is needed. Also, I seemingly can't run cosine in my RTX 3060 at runtime for `Float64`.

### Any other details?

We could implement a polynomial approximation function for compile time evaluation and when the accelerator doesn't have full support for `math.h` (e.g. Cuda seems to only support `Float32` with `cosf`) as a fallback

bgreni:
I'm a bit concerned about the possibility of a function potentially returning slightly different results depending on whether it was compile, or runtime computed. I think even the fallback itself would need to be platform-aware which could end up being a significant amount of work to implement and maintain.

JoeLoser:
I've added this to the backlog, but isn't an active priority for the team.  Feel free to open a pull request and we can consider it though if anyone is interested.  Note RTX 3060 is Tier 2 support: [https://docs.modular.com/max/faq/#tier-2-confirmed-compatibility](https://docs.modular.com/max/faq/#tier-2-confirmed-compatibility) meaning we do not have any CI tests for this platform, so it's on a "best supported" basis by the community/those with that hardware right now.

martinvuyk:
I agree with @bgreni I think I would prefer the approximation function to be a separate one and not introduce different behavior. For now I've just implemented my own taylor series expansion one. But I do wish we could run precise sin and cos in in CPUs at compile time

PR #5378, opened 2025-09-30
[kernels] Add fft implementation

martinvuyk:
# An evolution of the Fast fourier transform for the Modular hackathon.

This is the improved and generalized version.

## Sequential intra-block on an RTX 5090

The performance beats cufft for sequential fft executions.
cufft takes ~36 miliseconds running 10k 128-point fft. These are the results
for my algorithm (the lists are the radixes used):

```terminal
bench_intra_block_radix_n_rfft[[64, 2], 128], 34.393
bench_intra_block_radix_n_rfft[[32, 4], 128], 9.551
bench_intra_block_radix_n_rfft[[16, 8], 128], 4.793
bench_intra_block_radix_n_rfft[[16, 4, 2], 128], 4.698
bench_intra_block_radix_n_rfft[[8, 8, 2], 128], 4.618
bench_intra_block_radix_n_rfft[[8, 4, 4], 128], 2.563
bench_intra_block_radix_n_rfft[[8, 4, 2, 2], 128], 2.899
bench_intra_block_radix_n_rfft[[8, 2, 2, 2, 2], 128], 2.657
bench_intra_block_radix_n_rfft[[4, 4, 4, 2], 128], 2.525
bench_intra_block_radix_n_rfft[[4, 4, 2, 2, 2], 128], 2.649
bench_intra_block_radix_n_rfft[[4, 2, 2, 2, 2, 2], 128], 2.356
bench_intra_block_radix_n_rfft[[2], 128], 1.976
```

So the performance improvement depending on the radix used is ~ 85 - 94 %
(7 - 18 x) in sequential executions. Plus the benefit of no dynamic
allocation or planning step which wasn't included in the benchmark.

## CPU implementation on an Intel i5-12600KF

These are several implementations with different default behavior. Some use
multi-threading on by default depending on the platform's linalg libs.

```terminal
lib   100_000x128  1_000_000x128
numpy     44.663    431.364
scipy     35.113    345.956
pyfftw    31.148    805.163
```

Bellow are the results for different radix sizes, but for the sake of brevity
if we choose the best multi-threaded combination we get:

```terminal
bench_cpu_radix_n_rfft[[16, 8], 100_000, 128, workers=n], 28.802
bench_cpu_radix_n_rfft[[16, 8], 1_000_000, 128, workers=n], 306.103
```

Which means a ~ 10 % improvement for long sequences compared to the fastest
implementation of the 3 (scipy). This is highly dependent on the chosen radix
factors.

This is all ignoring the cost of interop between Python and the C/C++
implementations. I tried setting up fftw locally but it was a headache, and
this is just to give an estimate overview of the benefits of this
implementation.

## Conclusions

We can setup a simple function to estimate the best bases using the learnings
of the benchmarking. In the case of GPUs, the more threads the better. In the
case of CPUs, the closer `length // base` to the amount of logical
cores the better.

It is worth noting that these benchmarks highlight the strengths of the other
implementations. This implementation runs a generic Cooley-Tukey algorithm, and
thus leaves the other implementations in the dust when the sequence length is
not a power of the common prime factors (usually 2, 3, 5, and 7), because it
lets the user choose other prime factors when the default ones don't have all
those needed by the given sequence length instead of falling back to a slower
algorithm.

The only limitation is that a radix factor bigger than 100 leads to compilation
issues, it might be solvable by adding one of those slower fallback algorithms;
but I would rather see users using padding or other common ways to adjust the
sequence to the more efficient versions. Other fallback algorithms can be
implemented independently from this one and shouldn't be called fft given they
aren't running the Cooley-Tukey algorithm.

### Other results

Results for the Mojo fft implementation when using a single thread on CPU:

```terminal
bench_cpu_radix_n_rfft[[64, 2], 100_000, 128, workers=1], 434.76
bench_cpu_radix_n_rfft[[64, 2], 1_000_000, 128, workers=1], 4261.257
bench_cpu_radix_n_rfft[[32, 4], 100_000, 128, workers=1], 134.495
bench_cpu_radix_n_rfft[[32, 4], 1_000_000, 128, workers=1], 1301.183
bench_cpu_radix_n_rfft[[16, 8], 100_000, 128, workers=1], 115.614
bench_cpu_radix_n_rfft[[16, 8], 1_000_000, 128, workers=1], 1148.454
bench_cpu_radix_n_rfft[[16, 4, 2], 100_000, 128, workers=1], 88.025
bench_cpu_radix_n_rfft[[16, 4, 2], 1_000_000, 128, workers=1], 805.408
bench_cpu_radix_n_rfft[[8, 8, 2], 100_000, 128, workers=1], 110.555
bench_cpu_radix_n_rfft[[8, 8, 2], 1_000_000, 128, workers=1], 1082.624
bench_cpu_radix_n_rfft[[8, 4, 4], 100_000, 128, workers=1], 68.18
bench_cpu_radix_n_rfft[[8, 4, 4], 1_000_000, 128, workers=1], 686.936
bench_cpu_radix_n_rfft[[8, 4, 2, 2], 100_000, 128, workers=1], 82.91
bench_cpu_radix_n_rfft[[8, 4, 2, 2], 1_000_000, 128, workers=1], 808.935
bench_cpu_radix_n_rfft[[8, 2, 2, 2, 2], 100_000, 128, workers=1], 97.149
bench_cpu_radix_n_rfft[[8, 2, 2, 2, 2], 1_000_000, 128, workers=1], 956.262
bench_cpu_radix_n_rfft[[4, 4, 4, 2], 100_000, 128, workers=1], 89.543
bench_cpu_radix_n_rfft[[4, 4, 4, 2], 1_000_000, 128, workers=1], 843.657
bench_cpu_radix_n_rfft[[4, 4, 2, 2, 2], 100_000, 128, workers=1], 95.985
bench_cpu_radix_n_rfft[[4, 4, 2, 2, 2], 1_000_000, 128, workers=1], 1004.276
bench_cpu_radix_n_rfft[[4, 2, 2, 2, 2, 2], 100_000, 128, workers=1], 121.523
bench_cpu_radix_n_rfft[[4, 2, 2, 2, 2, 2], 1_000_000, 128, workers=1], 1127.333
bench_cpu_radix_n_rfft[[2], 100_000, 128, workers=1], 145.242
bench_cpu_radix_n_rfft[[2], 1_000_000, 128, workers=1], 1336.504
```

And these are the results when using multiple threads on CPU:

```terminal
bench_cpu_radix_n_rfft[[64, 2], 100_000, 128, workers=n], 101.729
bench_cpu_radix_n_rfft[[64, 2], 1_000_000, 128, workers=n], 992.453
bench_cpu_radix_n_rfft[[32, 4], 100_000, 128, workers=n], 41.277
bench_cpu_radix_n_rfft[[32, 4], 1_000_000, 128, workers=n], 356.703
bench_cpu_radix_n_rfft[[16, 8], 100_000, 128, workers=n], 28.802
bench_cpu_radix_n_rfft[[16, 8], 1_000_000, 128, workers=n], 306.103
bench_cpu_radix_n_rfft[[16, 4, 2], 100_000, 128, workers=n], 47.942
bench_cpu_radix_n_rfft[[16, 4, 2], 1_000_000, 128, workers=n], 412.996
bench_cpu_radix_n_rfft[[8, 8, 2], 100_000, 128, workers=n], 43.685
bench_cpu_radix_n_rfft[[8, 8, 2], 1_000_000, 128, workers=n], 430.616
bench_cpu_radix_n_rfft[[8, 4, 4], 100_000, 128, workers=n], 41.552
bench_cpu_radix_n_rfft[[8, 4, 4], 1_000_000, 128, workers=n], 399.767
bench_cpu_radix_n_rfft[[8, 4, 2, 2], 100_000, 128, workers=n], 52.064
bench_cpu_radix_n_rfft[[8, 4, 2, 2], 1_000_000, 128, workers=n], 507.537
bench_cpu_radix_n_rfft[[8, 2, 2, 2, 2], 100_000, 128, workers=n], 76.505
bench_cpu_radix_n_rfft[[8, 2, 2, 2, 2], 1_000_000, 128, workers=n], 646.568
bench_cpu_radix_n_rfft[[4, 4, 4, 2], 100_000, 128, workers=n], 63.228
bench_cpu_radix_n_rfft[[4, 4, 4, 2], 1_000_000, 128, workers=n], 571.997
bench_cpu_radix_n_rfft[[4, 4, 2, 2, 2], 100_000, 128, workers=n], 72.065
bench_cpu_radix_n_rfft[[4, 4, 2, 2, 2], 1_000_000, 128, workers=n], 699.373
bench_cpu_radix_n_rfft[[4, 2, 2, 2, 2, 2], 100_000, 128, workers=n], 76.774
bench_cpu_radix_n_rfft[[4, 2, 2, 2, 2, 2], 1_000_000, 128, workers=n], 771.757
bench_cpu_radix_n_rfft[[2], 100_000, 128, workers=n], 96.144
bench_cpu_radix_n_rfft[[2], 1_000_000, 128, workers=n], 939.783
```


max/kernels/src/nn/fft/utils.mojo
```diff
@@ -0,0 +1,333 @@
...
+    @parameter
+    for i in range(len(ordered_bases)):
+        alias base = ordered_bases[i]
+        base_offset //= base
+        reversed_idx += (current_val % base) * base_offset
+        current_val //= base
+    return reversed_idx
+
+
+# FIXME: mojo limitation. cos and sin can't run at compile time
```
abduld: can you file an issue and assign it to me. While mathematically sin/cos can be done via taylor approximation, computer numerics is a bit different and we should do this in a different manner.

max/kernels/src/nn/fft/fft.mojo
```diff
@@ -0,0 +1,969 @@
...
+        alias min_radix_for_block = length // common_thread_block_size
+
+        @parameter
+        if length // max_radix_number <= common_thread_block_size:
+            var radixes = range(min_radix_for_block, max_radix_number)
+            # TODO: replace this with inline for generators once they properly
+            # preallocate the sequence length (`[i for i in range(...)]`)
+            bases = List[UInt](capacity=len(radixes))
+            for i in range(len(radixes)):
+                bases.append(i)
```
abduld, line 81: the syntax


```suggestion
            bases = [UInt(i) for i in range(len(radixes))]
```

does not work?

max/kernels/benchmarks/nn/fft/bench_fft.mojo
```diff
@@ -0,0 +1,265 @@
...
+    b.iter[call_fn]()
+
+    _ = out_tensor
+    _ = x_tensor
+
+
+def main():
+    seed()
+    var m = Bench(BenchConfig(num_repetitions=1))
+    alias bases_list: List[List[UInt]] = [
```
abduld: can you lift those configs to a bench_fft.yaml file

max/kernels/src/nn/fft/fft.mojo
martinvuyk: AFAIK list comprehensions don't previously reserve the buffer size, so this makes compile times slower. I could be wrong (I remember somewhat vaguely reading about this somewhere), since this is a compiler black-box for me :man_shrugging:

max/kernels/src/nn/fft/utils.mojo
martinvuyk: I filed #5111 a couple of months back

max/kernels/benchmarks/nn/fft/bench_fft.mojo
martinvuyk: How would/should I do that? I'm not familiar with your internal tooling :sweat_smile:

abduld: take a look at https://github.com/modular/modular/blob/main/max/kernels/benchmarks/autotune/README.md

max/kernels/src/nn/fft/utils.mojo
abduld: should be fixed now

tjk213 (commented):
this is awesome!! 🚀 🚀

max/kernels/benchmarks/nn/fft/bench_fft.mojo
martinvuyk: Will what I pushed now be enough for running kbench? Maybe I'll have more time in the weekend to test it out

max/kernels/src/nn/fft/utils.mojo
JoeLoser: It just landed in the nightly that just got released, feel free to rebase and update this PR.  Thanks, Martin and Abdul!

max/kernels/src/nn/fft/fft.mojo
JoeLoser: List comprehensions don't have a size hint API to do the pre-allocation.  It's something we've talked about in the past, but is not implemented currently. The way the code is currently structured is more performant.

max/kernels/benchmarks/nn/fft/bench_fft.mojo
abduld: yep, that's fine

max/kernels/src/nn/fft/utils.mojo
```diff
@@ -0,0 +1,265 @@
...
+    elif length < UInt(UInt16.MAX):
+        return DType.uint16
+    elif length < UInt(UInt32.MAX):
+        return DType.uint32
+    elif UInt64(length) < UInt64.MAX:
+        return DType.uint64
+    elif UInt128(length) < UInt128.MAX:
+        return DType.uint128
+    else:
+        return DType.uint256
```
abduld, line 34: this can be implemented via 

```
return DType._uint_type_of_width[log2floor(length)]()
```

I think

max/kernels/src/nn/fft/utils.mojo
```diff
@@ -0,0 +1,265 @@
...
+fn _get_twiddle_factors[
+    length: UInt, dtype: DType, inverse: Bool = False
+](out res: InlineArray[ComplexSIMD[dtype, 1], length - 1]):
+    """Get the twiddle factors for the length.
+
+    Examples:
+        for a signal with 8 datapoints:
+        the result is: [W_1_8, W_2_8, W_3_8, W_4_8, W_5_8, W_6_8, W_7_8]
+    """
+    alias C = ComplexSIMD[dtype, 1]
```
abduld: can you use ComplexScalar and avoid using C here

max/kernels/src/nn/fft/utils.mojo
```diff
@@ -0,0 +1,265 @@
...
+    res = __type_of(res)(uninitialized=True)
+    alias N = length
+    for n in range(1, N):
+        # exp((-j * 2 * pi * n) / N)
+        var factor = 2 * n / N
```
abduld:
```suggestion
        var factor = 2 * n // N
```

max/kernels/src/nn/fft/fft.mojo
```diff
@@ -0,0 +1,969 @@
...
+    x_out: LayoutTensor[mut=True, out_dtype, x_out_layout],
+):
+    """A generic Cooley-Tukey algorithm. It has most of the generalizable radix
+    optimizations."""
+    constrained[length >= base, "length must be >= base"]()
+    alias Sc = Scalar[_get_dtype[length]()]
+    alias offset = Sc(processed)
+    var n = Sc(local_i) % offset + (Sc(local_i) // offset) * (offset * Sc(base))
+
+    alias Co = ComplexSIMD[out_dtype, 1]
```
abduld: Can you use ComplexScalar throughout the code

max/kernels/src/nn/fft/fft.mojo
```diff
@@ -0,0 +1,969 @@
...
+    local_i: UInt,
+    twiddle_factors: LayoutTensor[
+        out_dtype, twf_layout, twf_origin, address_space=twf_address_space
+    ],
+    x_out: LayoutTensor[mut=True, out_dtype, x_out_layout],
+):
+    """A generic Cooley-Tukey algorithm. It has most of the generalizable radix
+    optimizations."""
+    constrained[length >= base, "length must be >= base"]()
+    alias Sc = Scalar[_get_dtype[length]()]
```
abduld: would be great to drop these aliases to make the code more readable

max/kernels/src/nn/fft/utils.mojo
martinvuyk: This is the factor of angle in: `[0, 2*pi)`, so it has to be a floating point division

max/kernels/src/nn/fft/fft.mojo
martinvuyk: I didn't know that alias existed, will do

max/kernels/src/nn/fft/fft.mojo
martinvuyk: It's to use smaller datatypes for faster calculations and lower register use whenever possible. I'm not very sure how big of an impact this has, but I wanted to make sure not to use 32 bits for calculations that fit e.g. in 8 bits

max/kernels/src/nn/fft/utils.mojo
martinvuyk: We could go with
```mojo
fn _get_dtype[length: UInt]() -> DType:
    alias bit_width = Int(next_power_of_two(log2_ceil(length)))
    return _uint_type_of_width[bit_width << Int(bit_width <= 4)]()
```
But I find it way too hard to understand at a glance and it is also much less efficient than just doing a couple of comparisons.

We could actually introduce this helper into the dtype module itself in the future :man_shrugging:

owenhilyard: Should "smallest (u)int that can store this value" be something dtype exposes more generally? I think that might come up quite a bit.

abduld:
!sync

max/kernels/benchmarks/nn/fft/bench_fft.mojo
```diff
@@ -0,0 +1,371 @@
...
+    _ = out_tensor
+    _ = x_tensor
+
+
+def main():
+    seed()
+    var m = Bench(BenchConfig(num_repetitions=1))
+    alias name = env_get_string["name", "cpu_multi_threaded"]()
+    alias dtype = env_get_dtype["dtype", DType.float64]()
+    alias bases_int = env_get_shape["shape", "[16, 8]"]()
```
PeimingLiu:
```suggestion
    alias bases_int = env_get_shape["shape", "16x8"]()
```

this is the right format for shape to be parsed by `env_get_shape`, it should solve the compilation issue you are having ATM.

martinvuyk: Thanks!

abduld:
!sync

abduld:
@martinvuyk  still working on it

martinvuyk:
> @martinvuyk still working on it

@abduld don't worry I imagine the amount of stuff that needs adjusting behind the scenes. Let me know if there is anything I can do to help :)

lattner:
This is really great work Martin!

max/kernels/src/nn/fft/fft.mojo
```diff
@@ -0,0 +1,969 @@
...
+# ===----------------------------------------------------------------------=== #
+
+from algorithm import parallelize
+from complex import ComplexScalar
+from gpu import thread_idx, block_idx, block_dim, barrier
+from gpu.cluster import cluster_arrive_relaxed, cluster_wait
+from gpu.host import DeviceContext
+from gpu.host.info import is_cpu, Vendor
+from layout import Layout, LayoutTensor
+from layout.tensor_builder import LayoutTensorBuild as tb
```
JoeLoser: FYI I removed this type a few weeks back, you'll want to use `LayoutTensor` and its parameters directly now.

max/kernels/src/nn/fft/fft.mojo
```diff
@@ -0,0 +1,969 @@
...
+    alias grid_dim = (num_blocks, batches)
+    alias gpu_info = ctx.default_device_info
+    alias is_sm_90_or_newer = (
+        gpu_info.vendor == Vendor.NVIDIA_GPU and gpu_info.compute >= 9.0
+    )
+
+    @parameter
+    if is_sm_90_or_newer and num_blocks * batches <= max_cluster_size:
+        # TODO: this should use distributed shared memory for twfs
+        ctx.enqueue_function[
```
JoeLoser: **Suggestion** Here and elsewhere: can you please move to `enqueue_function_checked`?  We're almost done migrating the remaining stragglers off of `enqueue_function` so they get their arguments type-checked.

martinvuyk: done :)

max/kernels/src/nn/fft/fft.mojo
martinvuyk: done :)

martinvuyk:
Quick update, I'll leave this as draft until I'm done. I have a functional generic n-dimensional fft.

On the CPU side the implementation is currently faster than Numpy but slower than FFTW. I'll work on optimizing it later on.

I've finally got more thorough benchmarking setup and my implementation is slower than cufft for batched executions but I've found the biggest bottlenecks. I'll have to rework the whole architecture to have a similar setup to cufft by having a separate plan and an execution stage. I'll also look into having something like a circular shared memory buffer so that I can juxtapose the copying with the algorithm's execution.

So overall this will take me a couple more months of tinkering in my free time.

PR #5381, opened 2025-10-01
[stdlib] Remove unnecessary allocs in unicode casing

### LINKED ISSUE #5372: [Feature Request] [stdlib] `_binary_search_index()` shouldn't require allocation
### Review Mojo's priorities

### What is your request?

This should be in `Span`
```mojo
    fn _binary_search_index[
        dtype: DType, //,
    ](self: List[Scalar[dtype], **_], needle: Scalar[dtype]) -> Optional[UInt]:
    ...
```

Maybe also take the opportunity to see if SIMD can help speed it up.

### Any other details?

Flagged in [this forum post](https://forum.modular.com/t/how-do-you-put-bytes-into-rodata/2311)

martinvuyk:
Remove unnecessary allocs in unicode casing.

### DIFF
--- a/mojo/stdlib/std/collections/string/_unicode.mojo
+++ b/mojo/stdlib/std/collections/string/_unicode.mojo
@@ -11,6 +11,7 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 
+from std.builtin.globals import global_constant
 from std.collections.string._unicode_lookups import (
     has_uppercase_mapping,
     has_lowercase_mapping,
@@ -22,47 +23,28 @@ from std.collections.string._unicode_lookups import (
     uppercase_mapping3,
 )
 
-from std.memory import Span
 
-
-def _uppercase_mapping_index(rune: Codepoint) -> Int:
-    """Return index for upper case mapping or -1 if no mapping is given."""
-    return _to_index[has_uppercase_mapping](rune)
-
-
-def _uppercase_mapping2_index(rune: Codepoint) -> Int:
-    """Return index for upper case mapping converting the rune to 2 runes, or -1 if no mapping is given.
-    """
-    return _to_index[has_uppercase_mapping2](rune)
-
-
-def _uppercase_mapping3_index(rune: Codepoint) -> Int:
-    """Return index for upper case mapping converting the rune to 3 runes, or -1 if no mapping is given.
-    """
-    return _to_index[has_uppercase_mapping3](rune)
-
-
-def _lowercase_mapping_index(rune: Codepoint) -> Int:
-    """Return index for lower case mapping or -1 if no mapping is given."""
-    return _to_index[has_lowercase_mapping](rune)
+@always_inline
+fn _maybe_get_index[
+    lookup: InlineArray[UInt32, ...]
+](rune: Codepoint) -> Optional[UInt]:
+    if __is_run_in_comptime_interpreter:
+        return Span(materialize[lookup]())._binary_search_index(rune.to_u32())
+    return Span(global_constant[lookup]())._binary_search_index(rune.to_u32())
 
 
 @always_inline
-def _to_index[lookup: List[UInt32, ...]](rune: Codepoint) -> Int:
-    """Find index of rune in lookup with binary search.
-    Returns -1 if not found."""
+fn _get_mapping[
+    inner_size: Int,
+    outer_size: Int,
+    //,
+    mapping: InlineArray[SIMD[DType.uint32, inner_size], outer_size],
+](index: UInt) -> SIMD[DType.uint32, inner_size]:
+    if __is_run_in_comptime_interpreter:
+        return materialize[mapping]().unsafe_get(index)
+    return global_constant[mapping]().unsafe_get(index)
 
-    var result = Span(materialize[lookup]())._binary_search_index(rune.to_u32())
 
-    if result:
-        return Int(result.unsafe_value())
-    else:
-        return -1
-
-
-# TODO:
-#   Refactor this to return a Span[Codepoint, StaticConstantOrigin], so that the
-#   return `UInt` count and fixed-size `InlineArray` are not necessary.
 def _get_uppercase_mapping(
     char: Codepoint,
 ) -> Optional[Tuple[UInt, InlineArray[Codepoint, 3]]]:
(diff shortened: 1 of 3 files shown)


mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -776,3 +777,32 @@ struct Span[
...
+        Returns:
+            Returns None if `needle` is not present, or if `self` was not
+            sorted.
+
+        Notes:
+            This function will return an unspecified index if `self` is not
+            sorted in ascending order.
+        """
+
+        var ptr = self.unsafe_ptr()
```
NathanSWard: No need to use `unsafe_ptr` here. You can directly just call `self.unsafe_get(..)` in the loop

mojo/stdlib/test/memory/test_span.mojo
```diff
@@ -350,6 +352,41 @@ def test_unsafe_subspan():
...
 
 
+def test_binary_search():
+    def _test[dtype: DType]():
+        alias max_val = Int(Scalar[dtype].MAX)
+        var data = List[Scalar[dtype]](unsafe_uninit_length=max_val + 1)
+        iota(data)
+
+        # make sure we aren't reading an empty pointer
+        var view = Span(ptr=data.unsafe_ptr(), length=0)
```
NathanSWard: These could all be changed to use `__getitem__(slice)` -> `var view = Span(data)[:0]`.
So we don't need to interface w/ an UnsafePointer.

mojo/stdlib/test/memory/test_span.mojo
```diff
@@ -350,6 +352,41 @@ def test_unsafe_subspan():
     assert_equal(List(subspan2), [1, 2, 3])
 
 
+def test_binary_search():
+    def _test[dtype: DType]():
+        alias max_val = Int(Scalar[dtype].MAX)
+        var data = List[Scalar[dtype]](unsafe_uninit_length=max_val + 1)
```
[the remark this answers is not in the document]
NathanSWard: +1 on this. 
Not sure of the overhead of have a 2^32 element array.

mojo/stdlib/stdlib/collections/string/_unicode_lookups.mojo
```diff
@@ -5859,18 +5859,18 @@ alias has_uppercase_mapping3 = List[UInt32](
...
-    SIMD[DType.uint32, 4](0x03A5, 0x0313, 0x0342, 0),  #  ὖ -> Υ̓͂
-    SIMD[DType.uint32, 4](0x0399, 0x0308, 0x0300, 0),  #  ῒ -> Ϊ̀
-    SIMD[DType.uint32, 4](0x0399, 0x0308, 0x0301, 0),  #  ΐ -> Ϊ́
-    SIMD[DType.uint32, 4](0x0399, 0x0308, 0x0342, 0),  #  ῗ -> Ϊ͂
-    SIMD[DType.uint32, 4](0x03A5, 0x0308, 0x0300, 0),  #  ῢ -> Ϋ̀
-    SIMD[DType.uint32, 4](0x03A5, 0x0308, 0x0301, 0),  #  ΰ -> Ϋ́
-    SIMD[DType.uint32, 4](0x03A5, 0x0308, 0x0342, 0),  #  ῧ -> Ϋ͂
-    SIMD[DType.uint32, 4](0x0046, 0x0046, 0x0049, 0),  #  ﬃ -> FFI
-    SIMD[DType.uint32, 4](0x0046, 0x0046, 0x004C, 0),  #  ﬄ -> FFL
+alias uppercase_mapping3 = InlineArray[SIMD[DType.uint32, 4], 13](
```
NathanSWard: <3
Love the usage of `InlineArray` over `List`

mojo/stdlib/stdlib/collections/string/_unicode.mojo
```diff
@@ -210,28 +191,28 @@ fn to_uppercase(s: StringSlice[mut=False]) -> String:
...
+        input_offset += size
 
     return result^
 
 
-@always_inline
+@always_inline("nodebug")
 fn _estimate_needed_size(byte_len: Int) -> Int:
-    return 3 * (byte_len >> 1) + 1
+    return 3 * (byte_len // 2) + 1
```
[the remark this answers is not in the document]
NathanSWard: I also prefer the use of `//` over bitshifting. 
we can ignore copilot on this one haha

mojo/stdlib/stdlib/collections/string/_unicode.mojo
```diff
@@ -11,44 +11,44 @@
...
+    uppercase_mapping3,
+)
+from sys import is_compile_time
+
+
+fn _global_cst[T: AnyType, //, value: T]() -> ref [StaticConstantOrigin] T:
+    var ptr = __mlir_op.`pop.global_constant`[
+        value=value, _type = UnsafePointer[T]._mlir_type
+    ]()
+    return UnsafePointer[T](ptr)[]
```
NathanSWard: Hmmm, I'm not super sure on if we should be using `pop.global_constant`. This seems like a good use for it but I'm not sure how battle tested this is.
@JoeLoser - would likely know best

mojo/stdlib/std/collections/string/_unicode.mojo
```diff
@@ -64,45 +64,34 @@ fn _get_uppercase_mapping(
     """
     var array = InlineArray[Codepoint, 3](fill=Codepoint(0))
 
-    var index1 = _uppercase_mapping_index(char)
-    if index1 != -1:
-        var rune = materialize[uppercase_mapping]()[index1]
+    if index := _maybe_get_index[has_uppercase_mapping](char):
+        var rune = _get_mapping[uppercase_mapping](index.unsafe_value())
```
NathanSWard, line 59: Seeing this pattern - we should definitely add a `map` function to `Optional` so we clean up logic like this.

mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -776,3 +777,32 @@ struct Span[
...
+        dtype: DType, //,
+    ](self: Span[Scalar[dtype], **_], needle: Scalar[dtype]) -> Optional[UInt]:
+        """Finds the index of `needle` with binary search.
+
+        Args:
+            needle: The value to binary search for.
+
+        Returns:
+            Returns None if `needle` is not present, or if `self` was not
+            sorted.
```
NathanSWard:
```suggestion
            Returns None if `needle` is not present.
```

mojo/stdlib/stdlib/collections/string/_unicode.mojo
martinvuyk: Chris suggested it in this [forum post](https://forum.modular.com/t/how-do-you-put-bytes-into-rodata/2311/6). I just went with it because I found no other way to do this.

mojo/stdlib/test/memory/test_span.mojo
martinvuyk: True, will do

mojo/stdlib/std/collections/string/_unicode.mojo
martinvuyk: That is a very interesting idea

mojo/stdlib/test/memory/test_span.mojo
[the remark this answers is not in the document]
martinvuyk: I wanted to test overflow specifically, but I can just remove the 32 bit tests and go with the smaller ones

mojo/stdlib/stdlib/collections/string/_unicode.mojo
JoeLoser: Yeah, this is fine for now as Chris suggested. It may make sense to give this a public name in the function as the interim solution in the library, but we can do that later. Let's collect some uses and play with it some.

JoeLoser:
!sync

lsh:
These changes are exploding internally with different memory access violations. Can you make a version of this PR that just adds the tests and moves the function without doing any of the optimizations?

martinvuyk:
ping @lsh

PR #5385, opened 2025-10-03
[stdlib][benchmark] Force benchmark to respect max_iters

cudawarped:
The current benchmarking logic does not respect `max_iters`.  I've tried to fix this but I am not 100% sure my interpretation of the benchmarking logic is correct.  I have implemented the fix using the following interpretation:

1. Benchmarking should continue until `min_time_ns` has elapsed and either `max_time_ns` or `max_iters` is hit.
2. If `max_batch_size`  isn't specified then we attempt to choose a size which will last for at least `min_time_ns` if the other constraints from 1 (`max_iter` and `max_time_ns`) allow it.
3. If `max_batch_size` is specified use that on every iteration.
4. A batch is significant if:
   i.  `max_batch_size` is specified and it contains `max_batch_size` measurements. Currently there is an extra condition that the runtime should be greater than or equal to the min time for a batch but the condition currently evaluates `min_time_sec` and as there is no `min_batch_time_sec` argument I have removed the condition.
   ii.  Part of it is within the last 10% of batches.

If my understanding is correct should the `max_batch_size` argument be replaced with `batch_size` because it is fixed and not an upper bound on the size of the batch?

### DIFF
--- a/mojo/stdlib/stdlib/benchmark/bencher.mojo
+++ b/mojo/stdlib/stdlib/benchmark/bencher.mojo
@@ -271,9 +271,9 @@ struct BenchConfig(Copyable, Movable):
     var out_file: Optional[Path]
     """Output file to write results to."""
     var min_runtime_secs: Float64
-    """Upper bound on benchmarking time in secs."""
-    var max_runtime_secs: Float64
     """Lower bound on benchmarking time in secs."""
+    var max_runtime_secs: Float64
+    """Upper bound on benchmarking time in secs."""
     var min_warmuptime_secs: Float64
     """Lower bound on warmup time in secs."""
     var num_warmup_iters: Int
--- a/mojo/stdlib/stdlib/benchmark/benchmark.mojo
+++ b/mojo/stdlib/stdlib/benchmark/benchmark.mojo
@@ -405,8 +405,8 @@ fn run[
 
     Args:
         max_iters: Max number of iterations to run (default `1_000_000_000`).
-        min_runtime_secs: Upper bound on benchmarking time in secs (default `2`).
-        max_runtime_secs: Lower bound on benchmarking time in secs (default `60`).
+        min_runtime_secs: Lower bound on benchmarking time in secs (default `2`).
+        max_runtime_secs: Upper bound on benchmarking time in secs (default `60`).
         max_batch_size: The maximum number of iterations to perform per time
             measurement.
 
@@ -454,8 +454,8 @@ fn run[
 
     Args:
         max_iters: Max number of iterations to run (default `1_000_000_000`).
-        min_runtime_secs: Upper bound on benchmarking time in secs (default `2`).
-        max_runtime_secs: Lower bound on benchmarking time in secs (default `60`).
+        min_runtime_secs: Lower bound on benchmarking time in secs (default `2`).
+        max_runtime_secs: Upper bound on benchmarking time in secs (default `60`).
         max_batch_size: The maximum number of iterations to perform per time
             measurement.
 
@@ -491,8 +491,8 @@ fn run[
 
     Args:
         max_iters: Max number of iterations to run (default `1_000_000_000`).
-        min_runtime_secs: Upper bound on benchmarking time in secs (default `2`).
-        max_runtime_secs: Lower bound on benchmarking time in secs (default `60`).
+        min_runtime_secs: Lower bound on benchmarking time in secs (default `2`).
+        max_runtime_secs: Upper bound on benchmarking time in secs (default `60`).
         max_batch_size: The maximum number of iterations to perform per time
             measurement.
 
@@ -540,8 +540,8 @@ fn run[
 
     Args:
         max_iters: Max number of iterations to run (default `1_000_000_000`).
-        min_runtime_secs: Upper bound on benchmarking time in secs (default `2`).
-        max_runtime_secs: Lower bound on benchmarking time in secs (default `60`).
+        min_runtime_secs: Lower bound on benchmarking time in secs (default `2`).
+        max_runtime_secs: Upper bound on benchmarking time in secs (default `60`).
         max_batch_size: The maximum number of iterations to perform per time
             measurement.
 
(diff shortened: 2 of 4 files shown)


mojo/stdlib/stdlib/benchmark/benchmark.mojo
```diff
@@ -621,6 +624,13 @@ fn _run_impl(opts: _RunOptions) raises -> Report:
             n = max(n, prev_iters + 1)
             # The batch size should not be larger than 1.0e9.
             n = min(n, 1.0e9)
+            # Process at least one batch. i.e. Ensure n does not exceed opts.max_iters on the first iteration
+            if(total_iters == 0):
+                n = min(n, opts.max_iters)
+
+        # Respect hard limit of opts.max_iters if min_time_ns has elapsed
+        if time_elapsed > min_time_ns and (total_iters + Int(n)) > opts.max_iters:
```
parallelistix: Shouldn't this be `time_elapsed >= min_time_ns`?

mojo/stdlib/stdlib/benchmark/benchmark.mojo
```diff
@@ -640,17 +650,15 @@ fn _is_significant_measurement(
     idx: Int, batch: Batch, num_batches: Int, opts: _RunOptions
 ) -> Bool:
     # The measurement number of iteration is the same as the requested
-    # maxBatchSize and the measurement duration exceeded the requested min
-    # runtime.
+    # maxBatchSize
```
parallelistix: Please complete the comment.

mojo/stdlib/stdlib/benchmark/benchmark.mojo
```diff
@@ -640,17 +650,15 @@ fn _is_significant_measurement(
...
         opts.max_batch_size
         and batch.iterations >= opts.max_batch_size
-        and Float64(batch.duration) >= opts.min_runtime_secs
     ):
         return True
 
-    # This measurement occurred in the last 10% of the run.
-    if Float64(idx + 1) >= 0.9 * num_batches:
+    # All or part of this measurement occurred in the last 10% of batches.
+    if Float64(idx + 1) > 0.9 * num_batches:
```
parallelistix: Would rather keep this as `>=` instead of `>`.

mojo/stdlib/stdlib/benchmark/bencher.mojo
```diff
@@ -271,9 +271,9 @@ struct BenchConfig(Copyable, Movable):
     var out_file: Optional[Path]
     """Output file to write results to."""
     var min_runtime_secs: Float64
-    """Upper bound on benchmarking time in secs."""
-    var max_runtime_secs: Float64
     """Lower bound on benchmarking time in secs."""
+    var max_runtime_secs: Float64
+    """Upper bound on benchmarking time in secs."""
```
parallelistix, line 276: Thanks for fixing "Lower/Upper" typos.

mojo/stdlib/stdlib/benchmark/benchmark.mojo
cudawarped: Do you want to keep the `Float64(batch.duration) >= opts.min_runtime_secs` condition? i.e. only consider batches to be valid if their duration exceeds the minimum runtime for the entire benchmark?

mojo/stdlib/stdlib/benchmark/benchmark.mojo
cudawarped: @parallelistix Changed this back but it means that if there are 10 batches then both batch index 8 and 9 will be valid which feels wrong?

mojo/stdlib/stdlib/benchmark/benchmark.mojo
parallelistix: The code is fine, just adding more details to the comment would be great.

cudawarped: Is it possible for `run` not to execute exactly `n` runs? i.e. Is `batch.iterations != opts.max_batch_size` possible?

mojo/stdlib/stdlib/benchmark/benchmark.mojo
parallelistix: That's right, this will further limit the range of measurements, however, we want to change that in near future to avoid making multiple changes at once.

mojo/stdlib/stdlib/benchmark/benchmark.mojo
parallelistix: I think that is an acceptable case, doesn't have to run exactly `n` trials.

cudawarped: The variable is called `max_batch_size` but this logic allows batches to be bigger should it be changed from
```
    if (
        opts.max_batch_size
        and batch.iterations >= opts.max_batch_size
    ):
        return True
```

to

```
    if (
        opts.max_batch_size
        and batch.iterations == opts.max_batch_size
    ):
        return True
```

Additionaly should `max_batch_size` argument be called `batch_size` because it is fixed and not an upper bound on the size of the batch?  If `opts.timing_fn(Int(n))` results in a different number of iterations with the above change only measurements containing  `max_batch_size` will be significant.

mojo/stdlib/stdlib/benchmark/benchmark.mojo
cudawarped: Note added.

abduld:
!sync

mojo/stdlib/stdlib/benchmark/benchmark.mojo
```diff
@@ -600,18 +600,21 @@ fn _run_impl(opts: _RunOptions) raises -> Report:
...
         if opts.max_batch_size == 0:
             # We now count the next batchSize. A user might run the benchmark
             # with no warmup phase, so we need to make sure the divisor is not
             # zero.
-            # Compute the next batch size.
             if prev_dur > 0:
-                n = 1.2 * min_time_ns * prev_iters / Float64(prev_dur)
+                # Propose batch size which lasts at least min_time_ns or opts.max_iters
+                n = opts.max_iters
+                if(min_time_ns > 0):
```
abduld: drop the parens

mojo/stdlib/stdlib/benchmark/benchmark.mojo
```diff
@@ -621,6 +624,13 @@ fn _run_impl(opts: _RunOptions) raises -> Report:
             n = max(n, prev_iters + 1)
             # The batch size should not be larger than 1.0e9.
             n = min(n, 1.0e9)
+            # Process at least one batch. i.e. Ensure n does not exceed opts.max_iters on the first iteration
+            if(total_iters == 0):
```
abduld: drop parens

abduld:
!sync

PR #5386, opened 2025-10-03
[stdlib] Optimize lowercasing in `atof` by a lot

martinvuyk:
Optimize lowercasing in `atof` by a lot.

Improvement:

|name | markdown percentage | magnitude|
|-- | -- | --|
|atof/canada | 0.874 | 7.94|
|atof/mesh | 0.790 | 4.75|

Before:

|name | met (ms) | iters | throughput (GElems/s) | DataMovement (GB/s)|
|-- | -- | -- | -- | --|
|----------- | ----------------- | ----- | --------------------- | --------------------|
|atof/canada | 246.068 | 11 | 0.000452 | 0.008240 |
|atof/mesh | 69.738 | 17 | 0.001047 | 0.008059 |

After:

|name | met (ms) | iters | throughput (GElems/s) | DataMovement (GB/s)|
|-- | -- | -- | -- | --|
|----------- | ------------------ | ----- | --------------------- | -------------------|
|atof/canada | 30.971 | 38 | 0.003588 | 0.065469|
|atof/mesh | 14.678 | 82 | 0.004975 | 0.038293|

CC: @gabrieldemarmiesse this is just helping your implementation shine :)

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/_parsing_numbers/parsing_floats.mojo
+++ b/mojo/stdlib/stdlib/collections/string/_parsing_numbers/parsing_floats.mojo
@@ -284,6 +284,48 @@ fn lemire_algorithm(var w: UInt64, var q: Int64) -> Float64:
     return create_float64(m, p)
 
 
+alias _ascii_lower: Byte = ord("A") ^ ord("a")
+
+
+@always_inline
+fn _is_nan(stripped: StringSlice) -> Bool:
+    alias `n` = Byte(ord("n"))
+    alias `a` = Byte(ord("a"))
+    var ptr = stripped.unsafe_ptr()
+    return stripped.byte_length() == 3 and (
+        (ptr[0] | _ascii_lower == `n`)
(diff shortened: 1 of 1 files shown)


abduld:
!sync

soraros:
@martinvuyk Can we introduce a `StringSlice.is_[StringLiteral]()` that implements this unrolling programmatically?

martinvuyk:
> @martinvuyk Can we introduce a `StringSlice.is_[StringLiteral]()` that implements this unrolling programmatically?

That is an interesting idea. I could do it as a helper function related specifically to ascii (something like a `is_any_cased_ascii[c: StringSlice](v: StringSlice)` free floating function). So it would fall under the kind of optimizations I want to add after #3859.

But maybe it could also be a method on `StringSlice` and since the comparison value is a parameter I could add a branch that does ascii and another that uses unicode casing :thinking:. Then the implementations themselves remain in each module nice and orderly.

Thanks for the idea, I think I'll go and do it now lol

PR #5389, opened 2025-10-05
[stdlib] Add performant `String.is_like()` to replace `lhs.lower() == rhs.lower()` patterns

martinvuyk:
Add performant `String.is_like()` to replace `lhs.lower() == rhs.lower()` patterns.

CC: @soraros

### DIFF
--- a/max/kernels/src/internal_utils/_utils.mojo
+++ b/max/kernels/src/internal_utils/_utils.mojo
@@ -647,14 +647,13 @@ struct Mode(ImplicitlyCopyable, Movable, Stringable):
     alias SEP = "+"
 
     fn __init__(out self, handle: String = "run+benchmark+verify") raises:
-        var handle_lower = handle.lower().split(Self.SEP)
         self = Self.NONE
-        for h in handle_lower:
-            if String(Self.RUN.handle) == h:
+        for h in handle.split(Self.SEP):
+            if h.is_like[value = Self.RUN.handle]():
                 self.append(Self.RUN)
-            elif String(Self.BENCHMARK.handle) == h:
+            elif h.is_like[value = Self.BENCHMARK.handle]():
                 self.append(Self.BENCHMARK)
-            elif String(Self.VERIFY.handle) == h:
+            elif h.is_like[value = Self.VERIFY.handle]():
                 self.append(Self.VERIFY)
 
     fn append(mut self, other: Self):
--- a/mojo/stdlib/stdlib/collections/string/_parsing_numbers/parsing_floats.mojo
+++ b/mojo/stdlib/stdlib/collections/string/_parsing_numbers/parsing_floats.mojo
@@ -284,48 +284,6 @@ fn lemire_algorithm(var w: UInt64, var q: Int64) -> Float64:
     return create_float64(m, p)
 
 
-alias _ascii_lower: Byte = ord("A") ^ ord("a")
-
-
-@always_inline
-fn _is_nan(stripped: StringSlice) -> Bool:
-    alias `n` = Byte(ord("n"))
-    alias `a` = Byte(ord("a"))
-    var ptr = stripped.unsafe_ptr()
-    return stripped.byte_length() == 3 and (
-        (ptr[0] | _ascii_lower == `n`)
-        and (ptr[1] | _ascii_lower == `a`)
-        and (ptr[2] | _ascii_lower == `n`)
-    )
-
-
-@always_inline
-fn _is_inf(stripped: StringSlice) -> Bool:
-    alias `i` = Byte(ord("i"))
-    alias `n` = Byte(ord("n"))
-    alias `f` = Byte(ord("f"))
-    alias `t` = Byte(ord("t"))
-    alias `y` = Byte(ord("y"))
-    var ptr = stripped.unsafe_ptr()
-    var in_start = (ptr[0] | _ascii_lower == `i`) and (
-        ptr[1] | _ascii_lower == `n`
-    )
-    # f was removed previously
-    var is_in = stripped.byte_length() == 2 and in_start
-    return in_start and (
-        is_in
-        or (
-            stripped.byte_length() == 8
-            and (ptr[2] | _ascii_lower == `f`)
-            and (ptr[3] | _ascii_lower == `i`)
-            and (ptr[4] | _ascii_lower == `n`)
-            and (ptr[5] | _ascii_lower == `i`)
-            and (ptr[6] | _ascii_lower == `t`)
-            and (ptr[7] | _ascii_lower == `y`)
-        )
-    )
-
-
 fn _atof(x: StringSlice) raises -> Float64:
     """Parses the given string as a floating point and returns that value.
 
@@ -347,9 +305,10 @@ fn _atof(x: StringSlice) raises -> Float64:
     sign_and_stripped = get_sign(stripped)
     sign = sign_and_stripped[0]
     stripped = sign_and_stripped[1]
-    if _is_nan(stripped):
+    if stripped.is_like[value="nan"]():
         return FloatLiteral.nan
-    elif _is_inf(stripped):
+    # f was removed previously so it could be just "in"
+    elif stripped.is_like[value="infinity"]() or stripped.is_like[value="in"]():
         return FloatLiteral.infinity * sign
     try:
         w_and_q = _get_w_and_q_from_float_string(stripped)
--- a/mojo/stdlib/stdlib/collections/string/codepoint.mojo
+++ b/mojo/stdlib/stdlib/collections/string/codepoint.mojo
@@ -45,6 +45,7 @@ from sys.intrinsics import likely
 
 from bit import count_leading_zeros
 from bit._mask import splat
+from collections.string._unicode import _get_lowercase_mapping
 
 
 @always_inline
@@ -589,3 +590,29 @@ struct Codepoint(
         # Count how many of the minimums this codepoint exceeds, which is equal
         # to the number of bytes needed to encode it.
         return UInt(sizes.le(self.to_u32()).cast[DType.uint8]().reduce_add())
+
+    fn lower(self) -> Codepoint:
+        """Get the lowercase version of the codepoint.
+
+        Returns:
+            The lowercase version of the codepoint.
+        """
+        alias `a` = Byte(ord("a"))
+        alias `A` = Byte(ord("A"))
+        alias lower_ascii_latin1 = `A` ^ `a`
+
+        var size = self.utf8_byte_length()
+        if likely(size == 1):  # ASCII fast path
+            alias `z` = Byte(ord("z"))
+            var b = Byte(self.to_u32())
+            var low = b | lower_ascii_latin1
+            return Codepoint(low if `a` <= low <= `z` else b)
+        elif likely(size == 2):  # latin-1 fast path
+            alias `à` = Byte(ord("à"))
+            alias `þ` = Byte(ord("þ"))
+            alias `÷` = Byte(ord("÷"))
+            var b = Byte(self.to_u32())
+            var low = b | lower_ascii_latin1
+            return Codepoint(low if `à` <= low <= `þ` and low != `÷` else b)
+        else:
+            return _get_lowercase_mapping(self).or_else(self)
--- a/mojo/stdlib/stdlib/collections/string/string.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string.mojo
@@ -1872,6 +1872,41 @@ struct String(
         self._capacity_or_data = new_capacity
         self._set_ref_counted()
 
+    @always_inline
+    fn is_ascii(self) -> Bool:
+        """Return whether the values in the string are all ASCII.
+
+        Returns:
+            Whether the values in the string are all ASCII.
+        """
+        return self.as_string_slice().is_ascii()
+
+    @always_inline
+    fn is_like[value: StringSlice](self) -> Bool:
+        """Whether the string is like the given value. This is the equivalent
+        of doing `self.lower() == value.lower()`, but optimized.
+
+        Parameters:
+            value: The given value to compare against.
+
+        Returns:
+            Whether the string is like the given value.
+        """
+        return self.as_string_slice().is_like[value=value]()
+
+    @always_inline
+    fn is_like(self, value: StringSlice) -> Bool:
+        """Whether the string is like the given value. This is the equivalent
+        of doing `self.lower() == value.lower()`, but optimized.
+
+        Args:
+            value: The given value to compare against.
+
+        Returns:
+            Whether the string is like the given value.
+        """
+        return self.as_string_slice().is_like(value)
+
 
 # ===----------------------------------------------------------------------=== #
 # ord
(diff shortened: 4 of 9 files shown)


mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2360,6 +2363,86 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
+                alias char = value.as_bytes().unsafe_get(i)
+
+                @parameter
+                if `a` <= (char | ascii_lower) <= `z`:
+                    alias comp = char | ascii_lower
+                    result &= (data.unsafe_get(i) | ascii_lower) == comp
+                else:
+                    result &= data.unsafe_get(i) == char
+            return result
+        else:
```
soraros, line 2416: Can we use the dynamic `is_like` for this branch?

max/kernels/src/internal_utils/_utils.mojo
```diff
@@ -647,14 +647,13 @@ struct Mode(ImplicitlyCopyable, Movable, Stringable):
     alias SEP = "+"
 
     fn __init__(out self, handle: String = "run+benchmark+verify") raises:
-        var handle_lower = handle.lower().split(Self.SEP)
         self = Self.NONE
-        for h in handle_lower:
-            if String(Self.RUN.handle) == h:
+        for h in handle.split(Self.SEP):
+            if h.is_like[value = Self.RUN.handle]():
```
soraros, line 652: I think we can drop the keyword argument here.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2360,6 +2363,86 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
+        @parameter
+        fn _is_ascii[w: Int](v: SIMD[DType.uint8, w]) -> SIMD[DType.bool, w]:
+            return v.lt(0b1000_0000)
+
+        # FIXME(#5341): this should be map_reduce[_is_ascii, bool_and_fn]()
+        return self.as_bytes().count[func=_is_ascii]() == UInt(
+            self.byte_length()
+        )
+
+    fn is_like[value: StringSlice](self) -> Bool:
```
soraros: I'm not sure about the method name (is it too cute?), though I don't have better ideas.

soraros:
The "lowercase-before-comparison" semantics of `is_like` are a bit curious. I wonder whether we could introduce an `ascii_lower` method (assuming the input is ASCII) and write something like `s.ascii_lower().is_[some_literal]()`, making the design more orthogonal. I'd love to hear what others think.

YkdWaWEzVmphR1Z1:
Shouldn't this do proper case folding? https://www.unicode.org/Public/UCD/latest/ucd/CaseFolding.txt

max/kernels/src/internal_utils/_utils.mojo
martinvuyk: We can't due to the inferred origin parameter. It's the same issue as using `SIMD.shuffle[mask: IndexList]`

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: Why? we can lower once at compile time, and then reference the compile time string several times. This avoids having to pay the cost of having to fetch the lowercase mapping every time the function is called

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I'm open to ideas, it's the only thing that occurred to me that signifies "somewhat similar in any casing combination"

martinvuyk:
> Shouldn't this do proper case folding? https://www.unicode.org/Public/UCD/latest/ucd/CaseFolding.txt

Thanks for the link @bGVia3VjaGVu dang I was suspecting we were missing something when looking at the unicode casing code. We are currently uppercasing using the correct mappings from codepoints into the uppercase grapheme, but when lowercasing we are looking for the direct mapping per codepoint and not folding graphemes into those codepoints that map to more than one when uppercased.

I think this means we need a new mapping. I'll open an issue.

martinvuyk:
> The "lowercase-before-comparison" semantics of `is_like` are a bit curious. I wonder whether we could introduce an `ascii_lower` method (assuming the input is ASCII) and write something like `s.ascii_lower().is_[some_literal]()`, making the design more orthogonal. I'd love to hear what others think.

I think that would be doable but only if it returns an iterator, because it's otherwise allocating (we can't change the semantics and have `lower` mutate in-place).

YkdWaWEzVmphR1Z1:
This is python's versions: https://docs.python.org/3/library/stdtypes.html#str.casefold.
Iterator, compare directly, a new string.
I guess perhaps everything.

martinvuyk:
> This is python's versions: https://docs.python.org/3/library/stdtypes.html#str.casefold. Iterator, compare directly, a new string. I guess perhaps everything.

~~I'm going to go for using `.lower().upper()`, that should map every one of these weird cases to their respective values.~~

Nevermind, I'll stick with saying the function is the equivalent of `self.lower() == value.lower()`. Dealing with all the edge cases is something the user can do by adding the call themselves before sending it to this function, it is too expensive to set as the default.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I ended up using the dynamic version by iterating over the codepoints instead of allocating

YkdWaWEzVmphR1Z1:
> Lowercase_Mapping(C) is used for most characters, but there are instances in which the folding must be based on Uppercase_Mapping(C), instead.

https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G53253
I don't know if lower case instead of upper case will you here.

martinvuyk:
> > Lowercase_Mapping(C) is used for most characters, but there are instances in which the folding must be based on Uppercase_Mapping(C), instead.
> 
> https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G53253 I don't know if lower case instead of upper case will you here.

I'm going to go for offering lowercase equality and not case-insensitive equality. Users that deal with edge-case languages should do `.lower().upper()` on the values to remove the edge cases. Uppercasing is too expensive to set as the default for this function.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -308,7 +308,7 @@ struct CodepointSliceIter[
 
 
 struct CodepointsIter[mut: Bool, //, origin: Origin[mut]](
-    ImplicitlyCopyable, Movable, Sized
+    ImplicitlyCopyable, Iterable, Iterator, Movable, Sized
```
NathanSWard, line 311: Could you split this into a separate PR?
Conforming `CodepointsIter` to `Iterable` is a bit out of scope of this PR even if it was a drive-by fix :)

mojo/stdlib/stdlib/collections/string/string.mojo
```diff
@@ -1872,6 +1872,41 @@ struct String(
...
+    fn is_ascii(self) -> Bool:
+        """Return whether the values in the string are all ASCII.
+
+        Returns:
+            Whether the values in the string are all ASCII.
+        """
+        return self.as_string_slice().is_ascii()
+
+    @always_inline
+    fn is_like[value: StringSlice](self) -> Bool:
```
NathanSWard, line 1885: This definitely seems like a super useful function, but the name `is_like` is not very descriptive as to what this is doing. Something like `eq_ignore_case` (feel free to to bike shed) better captures what this function does and is clearer when read.

mojo/stdlib/stdlib/collections/string/codepoint.mojo
```diff
@@ -589,3 +590,29 @@ struct Codepoint(
         # Count how many of the minimums this codepoint exceeds, which is equal
         # to the number of bytes needed to encode it.
         return UInt(sizes.le(self.to_u32()).cast[DType.uint8]().reduce_add())
+
+    fn lower(self) -> Codepoint:
```
NathanSWard, line 594: I see this is a new function added as well. For changes like this it would be nice if these were broken into smaller PRs to more easily review.
Also adding some tests to bind this function would be great too 🙏

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2360,6 +2363,95 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
         """
         return String(elems, sep=self)
 
+    fn is_ascii(self) -> Bool:
```
NathanSWard, line 2366: Assuming this is a new function too - tests binding the behavior would also be appreciated :)

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2360,6 +2363,95 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
+            for i in range(length):
+                alias char = value.as_bytes().unsafe_get(i)
+                alias lowercase = char | ascii_lower
+
+                @parameter
+                if `a` <= lowercase <= `z`:
+                    result &= (data.unsafe_get(i) | ascii_lower) == lowercase
+                else:
+                    result &= data.unsafe_get(i) == char
+            return result
```
NathanSWard, line 2415: Not sure of the perf characteristics, but if this could use `zip` that would be greatly preferred for readability.

mojo/stdlib/stdlib/collections/string/_parsing_numbers/parsing_floats.mojo
```diff
@@ -347,9 +305,10 @@ fn _atof(x: StringSlice) raises -> Float64:
     sign_and_stripped = get_sign(stripped)
     sign = sign_and_stripped[0]
     stripped = sign_and_stripped[1]
-    if _is_nan(stripped):
+    if stripped.is_like[value="nan"]():
```
NathanSWard, line 308: This is a great cleanup 👏

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -2360,6 +2363,95 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut]](
...
+        Parameters:
+            value: The given value to compare against.
+
+        Returns:
+            Whether the string is like the given value.
+        """
+
+        @parameter
+        if value.is_ascii():
+            alias `A` = Byte(ord("A"))
```
NathanSWard, line 2395: It may be nice to refactor this out into a separate function `byte_eq_ascii` (name TBD) that checks if two bytes are ascii equal. It would help improve the readability of this function a bit :)

mojo/stdlib/stdlib/collections/string/string.mojo
YkdWaWEzVmphR1Z1: Given that the equality trait is called `EqualityComparable`, I don't think eq_ignore_case is an appropriate name.  I think the method should be called something like `is_equal_case_folded`, because this is what the functions does or should be doing.

YkdWaWEzVmphR1Z1:
> I'm going to go for offering lowercase equality and not case-insensitive equality. Users that deal with edge-case languages should do `.lower().upper()` on the values to remove the edge cases. Uppercasing is too expensive to set as the default for this function.

Lowercase equality is not something you should be doing. Case folding combines upper, lower and title cases. 
#5392 is an issue you have created, because you refuse to use the right tables. 
I guess you have to use List for now to add new tables. You can also optimize the table to use lowercase mapping where possible.

mojo/stdlib/stdlib/collections/string/string.mojo
martinvuyk: The thing is I don't really want to incur the penalty of doing case folded equality. I think something like `is_eq_lowercase` might be appropriate. WDYT?

mojo/stdlib/stdlib/collections/string/codepoint.mojo
martinvuyk: I kept this PR as draft because I haven't yet decided how to spin these changes off. For this one specifically I was waiting on landing  #5381 and #5395 and then refactoring what can be brought into codepoint and what should remain in the lowercasing/uppercasing functions

mojo/stdlib/stdlib/collections/string/_parsing_numbers/parsing_floats.mojo
martinvuyk: This was the start of the idea of creating this function actually. @soraros suggested it :)

martinvuyk:
> > I'm going to go for offering lowercase equality and not case-insensitive equality. Users that deal with edge-case languages should do `.lower().upper()` on the values to remove the edge cases. Uppercasing is too expensive to set as the default for this function.
> 
> Lowercase equality is not something you should be doing. Case folding combines upper, lower and title cases. #5392 is an issue you have created, because you refuse to use the right tables. I guess you have to use List for now to add new tables. You can also optimize the table to use lowercase mapping where possible.

I don't want this to do case folded equality. You can add another performant version for that if you'd like. IMO it is not necessary for most use-cases and has a big perf hit so it should not be the default.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I was actually thinking of vectorizing this with #5341 later on :)

mojo/stdlib/stdlib/collections/string/string.mojo
soraros: What about something like `equivalent[preprocess: ... = LOWER](...)`? Or for fellow mathheads, `equivalent[up_to]()`.

martinvuyk: How about something like `fn eq_ignore_case[case_fold: Bool = False](...): ...` ? So that we can provide the case folded functionality but don't use it as default. We can first implement the logic without case folding to make it simpler at the beginning.

> something like `equivalent[preprocess: ... = LOWER](...)`?

I think we can just reduce it to a boolean for now (use lowercasing or do full case folding). Unless we find a reason to create an API like `apply`/`map`/`map_reduce` but for strings/codepoints. But I do think this is an interesting pattern for other use-cases, I'll keep it in mind and keep an eye out in case I see more places where it might apply

soraros: Let's just pick a long and ugly name (so it's easy to search and replace in the future) and move forward.

mojo/stdlib/stdlib/collections/string/codepoint.mojo
NathanSWard: All good - apologies for jumping the gun since I just realized this is a draft PR. appreciate the response :)

YkdWaWEzVmphR1Z1:
There's a trick that simplifies the process of creating Unicode tables so you don't have to analyze 20 different files.
There's this gigantic [xml file ](https://www.unicode.org/Public/UCD/latest/ucdxml/ucd.nounihan.grouped.zip) with over 1 million codepoints. Thus parsing [the casing properties](https://www.unicode.org/reports/tr42/#case_properties_0) is simply a matter of specifying uc, lc, cf, tc.

YkdWaWEzVmphR1Z1:
> IMO it is not necessary for most use-cases and has a big perf hit so it should not be the default.

Why do you want to ignore regex, file systems, python's casefold. It's not a perf hit, because it's simply another table. Normalization is not preserved, but this is a with all casing forms.

PR #5395, opened 2025-10-05
[stdlib] Greatly optimize unicode casing

martinvuyk:
Greatly optimize unicode casing by adding fast paths for ASCII and Latin-1 encodings.

## Benchmark results

CPU: Intel® Core™ i7-7700HQ

|bench | old value (ms) | new value (ms) | markdown perc | magnitude|
|-- | -- | -- | -- | --|
|bench_string_lower[10] | 72.42 | 15.50 | 0.79 | 4.67|
|bench_string_upper[10] | 121.40 | 22.11 | 0.82 | 5.49|
|bench_string_lower[30] | 74.95 | 15.39 | 0.79 | 4.87|
|bench_string_upper[30] | 125.36 | 21.35 | 0.83 | 5.87|
|bench_string_lower[50] | 77.73 | 14.76 | 0.81 | 5.27|
|bench_string_upper[50] | 132.22 | 20.05 | 0.85 | 6.59|
|bench_string_lower[100] | 86.46 | 14.48 | 0.83 | 5.97|
|bench_string_upper[100] | 119.76 | 19.96 | 0.83 | 6.00|
|bench_string_lower[1000] | 69.18 | 14.03 | 0.80 | 4.93|
|bench_string_upper[1000] | 125.47 | 19.48 | 0.84 | 6.44|
|bench_string_lower[10000] | 67.78 | 13.90 | 0.79 | 4.88|
|bench_string_upper[10000] | 125.44 | 19.44 | 0.85 | 6.45|
|bench_string_lower[100000] | 67.40 | 14.68 | 0.78 | 4.59|
|bench_string_upper[100000] | 125.52 | 20.99 | 0.83 | 5.98|
|bench_string_lower[1000000] | 67.39 | 15.11 | 0.78 | 4.46|
|bench_string_upper[1000000] | 125.58 | 20.58 | 0.84 | 6.10|

### DIFF
--- a/mojo/stdlib/std/collections/string/_unicode.mojo
+++ b/mojo/stdlib/std/collections/string/_unicode.mojo
@@ -23,6 +23,7 @@ from std.collections.string._unicode_lookups import (
 )
 
 from std.memory import Span
+from std.sys.intrinsics import likely
 
 
 def _uppercase_mapping_index(rune: Codepoint) -> Int:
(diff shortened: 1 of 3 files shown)


mojo/stdlib/std/collections/string/_unicode.mojo
```diff
@@ -175,21 +176,41 @@ fn to_lowercase(s: StringSlice[mut=False]) -> String:
...
         )
-        var lowercase_char_opt = _get_lowercase_mapping(rune_and_size[0])
-        if lowercase_char_opt is None:
-            result.write_bytes(
-                data[input_offset : input_offset + rune_and_size[1]]
-            )
+        ref rune = rune_and_size[0]
+        ref size = rune_and_size[1]
+
+        if likely(size == 1):  # ASCII fast path
```
NathanSWard, line 199: Is there a perf difference when using `likely` rather than when just using the branch as is?
Not sure if it's worth adding here unless there is a measurable difference. 
In C++ the `[[unlikely]] / [[likely]]` attributes were added as well but using them can sometimes lead to weird perf implications so better to just have this verified here :)

martinvuyk: It seems like the second one for latin-1 was detrimental. After removing it the average speedup over the baseline (no intrinsics) was 2% (when using the intrinsic for ascii). Thanks for pointing it out

barcharcraz:
This is obviously an improvement, but I think we should optimize the whole algorithm as well. `_unicode_lookups` appears to be a very simple translation of the UCD case folding data, and we could probably get better performance (and better code size) by preprocessing that data to something more compact.

A starting point would be figuring out the offsets for each character, then merging ranges with the same offsets and storing just the range start and length. Obviously this is more "work" to look up but it's a lot less thrashing. If you start the search by looking at the first range then you end up doing basically the same thing as this optimization except you _might_ not get the bitwise version (which probably won't matter, arithmetic has the same TP as bitwise ops and there aren't any dependency chains on the output).

You'd probably want to first check at the beginning of your map (in case of ascii) then start searching from wherever the last processed character was.

martinvuyk:
> This is obviously an improvement, but I think we should optimize the whole algorithm as well. `_unicode_lookups` appears to be a very simple translation of the UCD case folding data, and we could probably get better performance (and better code size) by preprocessing that data to something more compact.
> 
> A starting point would be figuring out the offsets for each character, then merging ranges with the same offsets and storing just the range start and length. Obviously this is more "work" to look up but it's a lot less thrashing. If you start the search by looking at the first range then you end up doing basically the same thing as this optimization except you _might_ not get the bitwise version (which probably won't matter, arithmetic has the same TP as bitwise ops and there aren't any dependency chains on the output).
> 
> You'd probably want to first check at the beginning of your map (in case of ascii) then start searching from wherever the last processed character was.

I agree we can rework this whole thing more optimally. I was actually thinking of maybe using the utf-8 values and build some indexing with them somehow instead of paying the cost of transforming stuff to codepoints. Your idea of piling everything into a single map sounds viable as well

But this PR is just about optimizing this implementation as-is. Having a way to access a global table without materializing it at runtime (#5381) would also be useful for the next optimized refactoring. And I think we should leave the fast-path for ascii and latin-1, I'm not so sure a lookup table is going to be faster than that. So IMO this is something we should work on later on, because it will affect mostly languages that have 3 or more utf8 byte sequences which shouldn't be a priority at this stage IMO

martinvuyk:
ping @barcharcraz

PR #5396, opened 2025-10-05
Tweak git clone command

fadend:
As is, the command gives me the following error:

```
% git clone git@github.com:modular/modular
Cloning into 'modular'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

The replacement works (and I think matches pretty common usage?).

Thank you for the tutorial!


fadend:

abduld:
!sync

PR #5398, opened 2025-10-06
[stdlib] [proposal] Make `splitlines` return an iterator

martinvuyk:
Make `splitlines` return an iterator. Successor to  #3858

I think functions like this one and `split` should return iterators that are indexable and sliceable. Most of the use cases for them iterate over the result of calling the function anyway. So IMO we should nudge users towards allocating only when necessary.

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_string.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_string.mojo
@@ -13,7 +13,7 @@
 
 from collections import Optional
 from collections.string._utf8 import _is_valid_utf8
-from collections.string.string_slice import _split
+from collections.string.string_slice import _split, _to_string_slice_list
 from os import abort
 from pathlib import _dir_of_current_file
 from random import seed
@@ -160,8 +160,8 @@ fn bench_string_splitlines[
     @always_inline
     @parameter
     fn call_fn() raises:
-        for _ in range(1_000_000 // length):
-            var res = items.splitlines()
+        for _ in range(10**6 // length):
+            var res = _to_string_slice_list(items.splitlines())
             keep(res.unsafe_ptr())
 
     b.iter[call_fn]()
--- a/mojo/stdlib/stdlib/collections/string/string_slice.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string_slice.mojo
@@ -80,6 +80,169 @@ alias StaticString = StringSlice[StaticConstantOrigin]
 """An immutable static string slice."""
 
 
+@fieldwise_init
+struct _RangedIteratorWrapper[
+    is_mutable: Bool, //,
+    origin: Origin[is_mutable],
+    keep_ends: Bool,
+    forward: Bool = True,
+](ImplicitlyCopyable, Iterable, Iterator, Movable, Writable):
+    alias IteratorType[
+        iterable_mut: Bool, //, iterable_origin: Origin[iterable_mut]
+    ]: Iterator = Self
+    alias Element = StringSlice[origin]
+
+    var _inner_iter: _SplitlinesIter[
+        origin=origin, keep_ends=keep_ends, forward=forward
+    ]
+    var _how: Slice
+    var _idx: Int
+
+    fn __iter__(ref self) -> Self.IteratorType[__origin_of(self)]:
+        return self.copy()
+
+    @always_inline
+    fn __has_next__(self) -> Bool:
+        return (
+            self._inner_iter.__has_next__()
+            and self._idx < self._how.end.or_else(Int.MAX)
+        )
+
+    @always_inline
+    fn _next(mut self) -> StringSlice[origin]:
+        if self.__has_next__():
+            self._idx += 1
+            return self._inner_iter.__next__()
+        else:
+            return {}
+
+    fn __next__(mut self) -> StringSlice[origin]:
+        for _ in range(self._how.start.or_else(0)):
+            _ = self._next()
+        self._how.start = None
+        var value = self._next()
+        for _ in range(self._how.step.or_else(1) - 1):
+            _ = self._next()
+        return value
+
+    fn __getitem__(var self, idx: Some[Indexer]) -> StringSlice[origin]:
+        debug_assert(index(idx) >= 0, "no negative indexing.")
+        var i = 0
+        for elem in self:
+            if i < index(idx):
+                i += 1
+                continue
+            return elem
+        return {}
+
+    fn write_to(self, mut writer: Some[Writer]):
+        for elem in self.copy():
+            writer.write(elem)
+
+    fn __str__(self) -> String:
+        return String(self)
+
+
+@fieldwise_init
+struct _SplitlinesIter[
+    is_mutable: Bool, //,
+    origin: Origin[is_mutable],
+    *,
+    keep_ends: Bool = False,
+    forward: Bool = True,
+](ImplicitlyCopyable, Iterable, Iterator, Movable, Writable):
+    """Iterator for `StringSlice` over unicode line breaks.
+
+    Parameters:
+        is_mutable: Whether the slice is mutable.
+        origin: The origin of the underlying string data.
+        keep_ends: Whether to keep the line breaks in the resulting slices.
+        forward: The iteration direction. `False` is backwards.
+    """
+
+    alias IteratorType[
+        iterable_mut: Bool, //, iterable_origin: Origin[iterable_mut]
+    ]: Iterator = Self
+    alias Element = StringSlice[origin]
+    var _slice: StringSlice[origin]
+
+    fn __iter__(ref self) -> Self.IteratorType[__origin_of(self)]:
+        return self.copy()
+
+    fn __next__(mut self) -> StringSlice[origin]:
+        alias `\r` = UInt8(ord("\r"))
+        alias `\n` = UInt8(ord("\n"))
+
+        # highly performance sensitive code, benchmark before touching
+        @parameter
+        if forward:
+            var ptr = self._slice.unsafe_ptr()
+            var length = UInt(self._slice.byte_length())
+            var line_end = UInt(0)
+            var is_new_line = False
+            var b0 = Byte(0)
+            var char_len = UInt(0)
+
+            while not is_new_line and line_end < length:
+                b0 = ptr[line_end]
+                char_len = _utf8_first_byte_sequence_length(b0)
+                debug_assert(
+                    line_end + char_len <= length,
+                    "corrupted sequence causing unsafe memory access",
+                )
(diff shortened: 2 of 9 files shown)


mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+    fn __next__(mut self) -> StringSlice[origin]:
+        alias `\r` = UInt8(ord("\r"))
+        alias `\n` = UInt8(ord("\n"))
+
+        # highly performance sensitive code, benchmark before touching
+        @parameter
+        if forward:
+            var ptr = self._slice.unsafe_ptr()
+            var length = UInt(self._slice.byte_length())
+            var line_end = UInt(0)
```
NathanSWard, line 181: I would prefer this to be just an `Int` - no specific benefit of the unsigned-ness and the ops that come with it. Generally default to `Int` for indexing - especially for collections

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+    var _slice: StringSlice[origin]
+
+    fn __iter__(ref self) -> Self.IteratorType[__origin_of(self)]:
+        return self.copy()
+
+    fn __next__(mut self) -> StringSlice[origin]:
+        alias `\r` = UInt8(ord("\r"))
+        alias `\n` = UInt8(ord("\n"))
+
+        # highly performance sensitive code, benchmark before touching
```
NathanSWard, line 176: Do we have benchmarks for this currently? It would be nice to have a basis for what to compare this to if it's ever altered in the future.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+
+            while not is_new_line and line_end < length:
+                b0 = ptr[line_end]
+                char_len = _utf8_first_byte_sequence_length(b0)
+                debug_assert(
+                    line_end + char_len <= length,
+                    "corrupted sequence causing unsafe memory access",
+                )
+                # percentage-wise a newline is uncommon compared to a normal byte
+                is_new_line = unlikely(
```
NathanSWard, line 194: I'm generally against the use of `likely` & `unlikely` unless there is an actual measurable difference in benchmarks. Using these can some times lead to unexpected codgen or interfere with the branch predictor that may end up hurting perf instead of helping.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+                    _is_newline_char_utf8(ptr, line_end, b0, char_len)
+                )
+                line_end += char_len
+
+            var str_len = line_end
+            var is_r = unlikely(b0 == `\r`)
+            var may_be_r_n = is_r and likely(line_end < length)
```
NathanSWard, line 201: typo 
```suggestion
            var maybe_r_n = is_r and likely(line_end < length)
```

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+
+    fn __getitem__(self, idx: Some[Indexer]) -> StringSlice[origin]:
+        debug_assert(index(idx) >= 0, "no negative indexing.")
+        var i = 0
+        for elem in self.copy():
+            if i < index(idx):
+                i += 1
+                continue
+            return elem
+        return {}
```
NathanSWard: I believe that this could be more useful as a general iterator adaptor instead of a special case for just this iterator type. 
Rust has `nth` [here](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth) which would be useful. 

For now I would prefer to hold off on this functionally and maybe add a `TODO` until iterators have a clearer direction.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+
+    fn _to_string_list(self, out res: List[String]):
+        res = {capacity = 32}  # guessing
+        for elem in self.copy():
+            res.append(String(elem))
+
+    fn _to_string_slice_list(self, out res: List[StringSlice[origin]]):
+        res = {capacity = 32}  # guessing
+        for elem in self.copy():
+            res.append(elem)
```
NathanSWard: I see these functions are only used in test code. 
If that's the case it would be better to have them just be free functions in the test files so we don't bloat `_SplitlineIter` type.

mojo/stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -160,8 +160,8 @@ fn bench_string_splitlines[
     @always_inline
     @parameter
     fn call_fn() raises:
-        for _ in range(1_000_000 // length):
-            var res = items.splitlines()
+        for _ in range(10**6 // length):
```
NathanSWard, line 163: IMO `1_000_000` is more readable as "one million"

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+        for _ in range(self._how.start.or_else(0)):
+            _ = self._next()
+        for _ in range(self._how.step.or_else(1) - 1):
+            _ = self._next()
+        if self._idx < self._how.end.or_else(Int.MAX):
+            return self._next()
+        else:
+            return {}
+
+    fn __getitem__(self, idx: Some[Indexer]) -> StringSlice[origin]:
```
NathanSWard: Ditto my comment below about a more generic iterator "get item" adaptor.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
```diff
@@ -80,6 +80,184 @@ alias StaticString = StringSlice[StaticConstantOrigin]
...
+
+    fn _to_string_list(self, out res: List[String]):
+        res = {capacity = 32}  # guessing
+        for elem in self.copy():
+            res.append(String(elem))
+
+    fn _to_string_slice_list(self, out res: List[StringSlice[origin]]):
+        res = {capacity = 32}  # guessing
+        for elem in self.copy():
+            res.append(elem)
```
NathanSWard: Ditto my comment below about these only being used in test code

NathanSWard:
Love to see the use of iterators and lazy evaluation over eager/allocation interface! 🥳

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I micro-optimized the hell out of this one, so yeah there are benchmarks. I ran them to compare this iterator approach in https://github.com/modular/modular/pull/3858.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I want to use unsigned math, and these values are compared against using that. Also, many collections in the stdlib use `normalize_index` which does negative indexing. I don't want to alter the implementation itself more than necessary either here or in https://github.com/modular/modular/pull/3858

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I benchmarked this thoroughly, especially the `is_new_line` one makes a difference (newlines are unlikely ~`< 5%` in a sequence of bytes, so reordering the machine instructions is useful). I don't want to alter the implementation in this PR or the one introducing this iterator any more than necessary.

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I was thinking more along the lines of "this may be rn"

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I thought about generalizing this as well, but I went for implementing it for this iterator specifically because they aren't fleshed out yet. This is needed because many people do `str.splitlines()[index]` to get a specific line. I would like this same indexing and slicing for the `.split()` iterator, which is used much more often like `.split()[:3]` `.split()[0]`

mojo/stdlib/stdlib/collections/string/string_slice.mojo
martinvuyk: I wanted to provide them as adapters for people that are using the splitlines function outside the stdlib. Adapters that aren't promised to be kept around once the iterator design is more fleshed out. I can lower them to free floating functions if you wish though.

mojo/stdlib/benchmarks/collections/bench_string.mojo
martinvuyk: for me it's more about how many zeros the length removes from the 6 ones the value has e.g. `10**6 // 10**4 = 10**2 = 100` is much faster than counting the number of zeros :man_shrugging:

mojo/stdlib/stdlib/collections/string/string_slice.mojo
NathanSWard: That would be ideal for this PR if that's okay :)

mojo/stdlib/stdlib/collections/string/string_slice.mojo
NathanSWard: Ahhh gotcha, that makes sense.
I guess in the future we have have the `__getitem__` method just internally call whatever API we come up with.

NathanSWard (approved):

NathanSWard:
!sync

mojo/stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -160,8 +160,8 @@ fn bench_string_splitlines[
...
+            var res = _to_string_slice_list(items.splitlines())
```
martinvuyk, line 164: I'm not sure benchmarking this by building and appending to a list is quite "fair" to the real-world use-case which is just iterating over it without allocating. Any opinions?

PR #5399, opened 2025-10-07
[Kernels] Migrate `bicubic` from `NDBuffer` to `LayoutTensor`

msaelices:
- Migrate bicubic interpolation kernels from `NDBuffer` to `LayoutTensor`
- Update `ResizeBicubic` kernel to use `LayoutTensor` instead of `NDBuffer`
- Migrate `bicubic` interpolation tests from `NDBuffer` to `LayoutTensor`


abduld:
!sync

PR #5403, opened 2025-10-08
[Kernels] Conv Transpose tests refactor to use `TestSuite`

msaelices:
Migrate the tests in `max/kernels/test/nn/test_conv_transposed.mojo` to use `TestSuite`


JoeLoser:
Thanks for starting to migrate the kernel tests over to `TestSuite`: this is awesome to see.  This will make @NathanSWard and @laszlokindrat very happy as well!

FYI that I completed the migration of the stdlib over to completely using `TestSuite` today (will go out in next nightly).  We also will share some updates/goodies soon on improvements we're cooking up to `TestSuite` and the future of testing in Mojo - stay tuned! 🚀

lsh:
FYI this test is about to have conflicts coming down the pipeline, so you'll probably need to rebase tomorrow or the next day.

lsh:
!sync

max/kernels/test/nn/test_conv_transposed.mojo
```diff
@@ -523,8 +539,25 @@ def main():
         1,  # num_groups
     )
 
+
+def main():
+    var suite = TestSuite()
+
     # Test conv_transpose_shape function
-    test_conv_transpose_shape_basic()
+    suite.test[test_conv_transpose_shape_basic]()
```
JoeLoser: **Question** Instead of enumerating all of the test functions out, can we now use `TestSuite.discover_tests[__functions_in_module()]().run()` instead like in the other stdlib tests?

msaelices: done: https://github.com/msaelices/modular/commit/9546b11d5fc128d26e76f93e579d6ab6f3df007b

msaelices: @JoeLoser it seems the compiler does not like the parameters in the inner functions called: https://github.com/modular/modular/actions/runs/18687185756/job/53282932159?pr=5403

JoeLoser: @laszlokindrat IIUC, this is similar to the internal bug you filed the other day, right?

@msaelices please feel free to revert and just register the test functions manually into the TestSuite as a workaround. Sorry for the churn, thanks for trying it out!

msaelices: done: https://github.com/msaelices/modular/commit/0a001f433b476644c9b73871ce9ad58f10c798be

JoeLoser (approved):

JoeLoser:
!sync

PR #5405, opened 2025-10-08
[stdlib] Add safety asserts for string indexing and slicing

### LINKED ISSUE #5402: [BUG] Slicing UTF-8 strings can return invalid results due to byte based indexing
### Actual behavior
Slicing a UTF-8 String and StringSlice both return an unexpected response.

### Expected behavior
First codepoint should be returned.

### Steps to reproduce

```mojo
fn main():
    var ok: String = "-"
    print(ok.as_string_slice()[0])

    var not_ok: String = "─"
    print(not_ok.as_string_slice()[0])
```

### System information

- Provide the system information by running `pixi info`.

```
System
------------
       Pixi version: 0.47.0
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=15.6.1=0
                   : __archspec=1=m2
```

- Provide version information for MAX (includes Mojo) by pasting the output of `pixi list max`. `0.25.6.0`

martinvuyk:
Add safety asserts for string indexing and slicing. Closes #5402.

We should actually fix this and make it slice by unicode codepoints. But for now at least we can provide some asserts

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/string.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string.mojo
@@ -83,6 +83,10 @@ from collections.string.string_slice import (
     _to_string_list,
     _unsafe_strlen,
 )
+from collections.string._utf8 import (
+    _is_utf8_continuation_byte,
+    _utf8_first_byte_sequence_length,
+)
 from hashlib.hasher import Hasher
 from io.write import STACK_BUFFER_BYTES, _TotalWritableBytes, _WriteBufferStack
 from os import PathLike, abort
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
Thanks for adding some safety guard rails in the meantime!  What do you think about adding various API document notes on these APIs to explicitly mention how only ASCII is supported right now?  This can minimize confusion for users reading the docs and/or running with asserts off (the default right now).

Over time, like you said, I think we should improve all of these APIs to do indexing and such by codepoints (CC: @laszlokindrat @ConnorGray from internal discussions).

JoeLoser:
!sync

martinvuyk:
> What do you think about adding various API document notes on these APIs to explicitly mention how only ASCII is supported right now? This can minimize confusion for users reading the docs and/or running with asserts off (the default right now).

I would rather just make the breaking change to unicode codepoints as soon as possible. For those who want byte indexing and want to go into perf/unsafe mode, there is always `str_slice.as_bytes()`, and grapheme indexing should IMO remain our last priority and can be implemented later on (but not as default due to heavy perf hit).

Also another thing that makes me want to do this ASAP is ecosystem growth. changing the indexing and length and find scheme for string will be a huge breaking change for a lot of people.

PR #5410, opened 2025-10-09
[stdlib] Clean up `_cpython.mojo` since we dropped Python 3.9

soraros:
- Remove dummy functions and simplify version checks


soraros:
@laszlokindrat I'm not sure whether we should keep the compatibility patch in `Py_Is`.

laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

PR #5417, opened 2025-10-11
[stdlib] Add default implementation for ImplicitlyBoolable trait

bgreni:
Take advantage of default trait impls for `ImplicitlyBoolable`

Upon review, this unfortunately doesn't appear to be applicable to any of the stdlib types that currently use this trait as they all also include some sort of `@always_inline` annotation.

Also moved `FloatLiteral.__bool__` to use `@always_inline("builtin")` as that appears to have been missed in the past

### DIFF
--- a/mojo/stdlib/stdlib/builtin/bool.mojo
+++ b/mojo/stdlib/stdlib/builtin/bool.mojo
@@ -87,13 +87,14 @@ trait ImplicitlyBoolable(Boolable):
     ```
     """
 
+    @always_inline
     fn __as_bool__(self) -> Bool:
         """Get the boolean representation of the value.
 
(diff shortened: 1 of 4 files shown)


NathanSWard:
I believe you can decorate the defaulted function with `@always_inline`.

bgreni:
> I believe you can decorate the defaulted function with `@always_inline`.

Yeah I wasn't sure if that was something we wanted.

soraros:
It's a one-line default implementation, I'd say `@always_inline` is exactly what we want.

laszlokindrat (approved):
LGTM, thanks!

laszlokindrat:
!sync

PR #5418, opened 2025-10-11
[Graph] Add `acos()` Max operation

msaelices:
This adds the arccosine (inverse cosine) operation to MAX graph ops. The implementation uses a compositional approach with existing graph operations, following the Remez polynomial approximation from math.mojo.


max/graph/ops/acos.py
```diff
@@ -0,0 +1,126 @@
+# ===----------------------------------------------------------------------=== #
```
abduld, line 1: this is not super efficient and duplicates the mojo implementation. It would be better to just introduce a custom kernel which is a wrapper around foreach and math.acos

msaelices: I'm not sure if I did correctly but it's done in these commits:
https://github.com/modular/modular/pull/5418/commits/624035ab58f637d9582b53cbe190558d9adc2ffd
https://github.com/modular/modular/pull/5418/commits/ed2517706b099d0d6f7adf42b26cc78521755c0d

abduld: Thanks, this is almost how to do it. Can you move the acos function to the elementwise file

max/graph/ops/acos.py
```diff
@@ -0,0 +1,57 @@
...
+        - the same shape as the input
+
+    Raises:
+        Error: If the symbol doesn't represent a tensor value.
+        Error: If the input is not a floating-point dtype.
+    """
+    from . import dtype_promotion  # Avoid circular import
+
+    x = dtype_promotion._restrict_to_strong_dtypes(x)
+    return Graph.current._add_op(mo.mo_acos, x._mlir_value.type, x)[0].tensor
```
abduld: there is no mo op called acos, so you want to add this as a custom op via `ops.custom`. It should look something like

```
return ops.custom(
     name="mo.acos",
     device = device,
     values = [x],
     out_types = 
                TensorType(
                    dtype=x.dtype,
                    shape=x.tensor.shape,
                    device=DeviceRef.from_device(device),
                )
 )[0].tensor
```

abduld:
!sync

abduld:
Sorry for the long delay. This is finally merged

PR #5425, opened 2025-10-11
[stdlib] Add `is_in` function to `Variant` class.

josiahls:
Used to check if a variant instance also is in another variant instance / variadic list. Useful for operating on types that are in a super set variant and then filtering subsets.

For example:
```mojo
        from utils import Variant

        def main():
            var x = Variant[Int, Float64](1)
            alias OnlyFloats = Variant[Float64, Float32]
            # Note the unpacked *VariantName.Ts syntax. This is required to
            # see if x also belongs to OnlyFloats.
            print(x.is_in[*OnlyFloats.Ts]())
````

### DIFF
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -64,6 +64,17 @@ what we publish.
 
 ### Standard library changes
 
+- Added `is_in` function to `Variant` class.
+
+```mojo
+from utils import Variant
+
+def main():
+    var x = Variant[Int, Float64](1)
+    alias OnlyFloats = Variant[Float64, Float32]
+    print(x.is_in[*OnlyFloats.Ts]())
+```
+
 - Added `unsafe_get`, `unsafe_swap_elements` and `unsafe_subspan` to `Span`.
 
 - The deprecated `DType.index` is now removed in favor of the `DType.int`.
--- a/mojo/stdlib/stdlib/utils/variant.mojo
+++ b/mojo/stdlib/stdlib/utils/variant.mojo
@@ -427,3 +427,30 @@ struct Variant[*Ts: Copyable & Movable](ImplicitlyCopyable, Movable):
         For example, the `Variant[Int, Bool]` permits `Int` and `Bool`.
         """
         return Self._check[T]() != Self._sentinel
+
+    fn is_in[*V: Copyable & Movable](self) -> Bool:
+        """Check if a variant instance is also in another variant instance.
+
+        Parameters:
+            V: The variant to check.
+
+        Returns:
+            True if the variant contains the requested type.
+
+        Example:
+        ```mojo
+        from utils import Variant
+
+        def main():
+            var x = Variant[Int, Float64](1)
+            alias OnlyFloats = Variant[Float64, Float32]
+            print(x.is_in[*OnlyFloats.Ts]())
+        ```
+        """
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by josiahls** [comment]
Additional practical context:

https://github.com/josiahls/c_binder_mojo/blob/7ee932f4c29985844c3b277008d57e538c0a2f45/c_binder_mojo/ast/declarations/function_decl_node.mojo#L278-L283

I have a super set variant: AstNodes. When iterating through a list of them, I want to know which ones are CommentNodes and handle those differently.

**Comment by NathanSWard** [comment]
Hmm interesting function, I've never thought about this before 👀 
FWIW the name `is_in` is likely not the best choice here as it doesn't clearly state the intent of the function.
Something like `active_varaint_type_is_in_pack` or something (although more verbose, is also clearer as what it's checking). Since `is_in` could refer to a multitude of implementations.

I'm trying to be a bit conservative with this API since ideally in the future we'd have variants (product types) built into the langue so it's difficult to see this function working there.

Curious what's the case where this is better/clearer over use using `variant.isa[T] or variant.isa[U], etc...`?

**Comment by josiahls** [comment]
> Hmm interesting function, I've never thought about this before 👀 FWIW the name `is_in` is likely not the best choice here as it doesn't clearly state the intent of the function. Something like `active_varaint_type_is_in_pack` or something (although more verbose, is also clearer as what it's checking). Since `is_in` could refer to a multitude of implementations.
> 
> I'm trying to be a bit conservative with this API since ideally in the future we'd have variants (product types) built into the langue so it's difficult to see this function working there.
> 
> Curious what's the case where this is better/clearer over use using `variant.isa[T] or variant.isa[U], etc...`?

This is the code I'm trying to replace with `is_in`. Its weird doing a loop within a loop currently.
```mojo
        for child in self.children():
            if child.is_in[*AstAttributeVariant.Ts]():
                continue
            # vs: 
            var is_in = False
            @parameter
            for j in range(len(AstAttributeVariant.Ts)):
                alias T = AstAttributeVariant.Ts[j]
                if child.isa[T]():
                    is_in = True
                    break
            if is_in:
                continue
```
You're right that its a little weird. Ideally I wanted just `variant_instance.is_in[OtherVariant]()`. Alternative names being `is_type_of` or `is_contained_by`. 

I basically have a variant `AllNodes` and I want to see if its also a variant of `SomeNodes`. I guess the alternative is maybe using traits somehow, but I'm not sure how that would work when iterating collections.

I could alternatively have a `get_subset_type()` for the AllNodes and then just filter via `if child.get_subset_type() == AttrType:`. Maybe this is my python side coming out, trying to shove OOP into mojo too early.

Alternatively, alternatively, I could just wait for struct extensions lol.

josiahls:
@NathanSWard I'm fine closing this, though would like to know what direction modular plans to go in with regard to comparing variants with each other. I had issues doing direct variant comparisons because working with the variadic packs was awkward (maybe I don't understand how to make them play well with parameterized methods?)

josiahls:
@NathanSWard 
I'm closing this. I see https://github.com/modular/modular/commit/c0ca3840662f2a3b21d855eb99ba251014376b56 and https://github.com/modular/modular/commit/df4f517fbebbf81877ac7d8c53df8d72db090eb5 that might solve this

PR #5428, opened 2025-10-12
[stdlib][tests] Remove deprecated `__iter__` call in `String` test.

christoph-schlumpf:
Removes the following warning in test_string: "Use `str.codepoints()` or `str.codepoint_slices()` instead."


NathanSWard:
!sync

NathanSWard (approved):
thanks for this!

PR #5430, opened 2025-10-12
[stdlib] Make `CodepointsIter` conform to `Iterable` and `Iterator`

martinvuyk:
Make `CodepointsIter` conform to `Iterable` and `Iterator`


abduld:
!sync

PR #5431, opened 2025-10-12
[stdlib] Micro-optimize `_utf8_byte_type()`

martinvuyk:
I had previously changed this from

```mojo
return count_leading_zeros(~b)
```
to
```mojo
return (
	b.ge(0b1000_0000).cast[DType.uint8]()
	+ b.ge(0b1100_0000).cast[DType.uint8]()
	+ b.ge(0b1110_0000).cast[DType.uint8]()
	+ b.ge(0b1111_0000).cast[DType.uint8]()
)
```
Because I thought I found a bug for some sequences that have length 4. When the last byte can be e.g. `0b1111_1000` then the `~b` logic would make it count 5 leading zeros.

But the maximum unicode codepoint is `0x10FFFF` whose first utf-8 byte is `1111_0100` so we can safely go back to the previous implementation and fix the tests.

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/_utf8.mojo
+++ b/mojo/stdlib/stdlib/collections/string/_utf8.mojo
@@ -173,7 +173,7 @@ fn _is_valid_utf8_comptime(span: Span[mut=False, Byte, **_]) -> Bool:
         if byte_type == 0:
             offset += 1
             continue
-        elif byte_type == 1:
+        elif byte_type == 1 or byte_type > 4:
             return False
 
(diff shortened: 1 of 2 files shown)


abduld:
!sync

martinvuyk:
@abduld give me a minute, it seems that I was testing it wrong. The maximum valid unicode codepoint is `0x10FFFF` whose first utf-8 byte is `1111_0100` so the tests are actually wrong and the previous implementation was fine. I'll fix this PR up in a bit

PR #5434, opened 2025-10-13
[stdlib] Fix `_utf8_byte_type()` test and utf8 validation logic

martinvuyk:
I had previously changed this from
```mojo
return count_leading_zeros(~b)
```
to
```mojo
return count_leading_zeros(~b | 0b0000_1111)
```

Because I thought I found a bug for some sequences that have length 4. When the last byte can be e.g. `0b1111_1000` then the ~b logic would make it count 5 leading zeros.

But the maximum unicode codepoint is `0x10FFFF` whose first utf-8 byte is `0b1111_0100` so we can safely go back to the previous implementation and fix the tests and utf8 validation logic.

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/_utf8.mojo
+++ b/mojo/stdlib/stdlib/collections/string/_utf8.mojo
@@ -24,6 +24,10 @@ from memory import Span
 # Validate UTF-8
 # ===-----------------------------------------------------------------------===#
 
+alias BIGGEST_UTF8_FIRST_BYTE = Byte(0b1111_0100)
+"""Since the biggest unicode codepoint is 0x10FFFF then the biggest
+first byte that a utf-8 sequence can have is 0b1111_0100 (0xF4).
+"""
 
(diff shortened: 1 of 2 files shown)


laszlokindrat (approved):
Nice catch, thanks! Out of curiosity, how did you run into this?

laszlokindrat:
!sync

martinvuyk:
> Nice catch, thanks! Out of curiosity, how did you run into this?

When doing the PR for compile time `_validate_utf8`. I thought I'd caught a bug in the `_utf8_byte_type` function but it turned out the input had an invalid first byte that was making invalid input pass because I wasn't checking for it (since I assumed `_utf8_byte_type` always returned a max of 4). But when out of curiosity as to why no bug had cropped up due to the previous implementation (that didn't deal with the "bug"), I searched up the biggest codepoint and serialized it to utf-8 to look at the biggest first byte. And when I saw it couldn't be bigger than that then I finally understood why the utf8 tests were considered valid by the compile time version and not the runtime one.

PR #5444, opened 2025-10-14
[stdlib] Minor cosmetic and consistency improvements in `python_object.mojo` for better readability

soraros:
- Follow the convention that `PyObjectPtr` variables use the `_ptr` suffix, while `PythonObject` variables do not. This improves clarity when both types appear together.
- Rename the `Python().cpython()` instance to `cpy` (from `cpython`) to reduce visual noise and line length.
- Use `unsafe_get_error` in places where the presence of an error is guaranteed by CPython.
- Name the result of CPython API calls that return error codes `errno`, consistently distinguishing them from APIs that return `PyObject*`, where `NULL` indicates an error.

### DIFF
--- a/mojo/stdlib/stdlib/python/python_object.mojo
+++ b/mojo/stdlib/stdlib/python/python_object.mojo
@@ -300,7 +300,7 @@ struct PythonObject(
         ref cpy = Python().cpython()
         var unicode = cpy.PyUnicode_DecodeUTF8(string)
         if not unicode:
-            raise cpy.get_error()
+            raise cpy.unsafe_get_error()
         self = Self(from_owned=unicode)
 
     @implicit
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/python/python_object.mojo
```diff
@@ -1429,33 +1395,33 @@ fn _unsafe_alloc[
     Raises:
         If the Python object allocation fails.
     """
-    ref cpython = Python().cpython()
-    var obj_py_ptr = cpython.PyType_GenericAlloc(type_obj_ptr, 0)
-    if not obj_py_ptr:
+    ref cpy = Python().cpython()
+    var obj_ptr = cpy.PyType_GenericAlloc(type_obj_ptr, 0)
+    if not obj_ptr:
         raise Error("Allocation of Python object failed.")
```
soraros, line 1408: Maybe we should `raise cpy.get_error()` here?

laszlokindrat: I think this is fine for now. Maybe at some point we'll overhaul all the error propagation from cpython, but not yet.

laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

PR #5445, opened 2025-10-14
[stdlib] Parallelize `TestSuite.run()`

martinvuyk:
Parallelize `TestSuite.run()`

### DIFF
--- a/mojo/stdlib/stdlib/testing/suite.mojo
+++ b/mojo/stdlib/stdlib/testing/suite.mojo
@@ -11,8 +11,10 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 
+from algorithm import parallelize
 from math import ceil, floor
 from os import sep
+from runtime.asyncrt import parallelism_level
 from time import perf_counter_ns
 from utils._ansi import Color, Text
 
@@ -191,8 +193,11 @@ struct TestSuite(Movable):
         var replacement = String("\n", _Indent("", level=Self._ErrorIndent))
         return e.__str__().replace("\n", replacement)
 
-    def run(deinit self):
+    fn run(deinit self, *, parallelized: Bool = False) raises:
         """Runs the test suite and prints the results to the console.
+
+        Args:
+            parallelized: Whether to parallelize the run.
 
         Raises:
             An error if a test in the test suite fails.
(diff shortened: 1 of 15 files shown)


mojo/stdlib/stdlib/testing/suite.mojo
```diff
@@ -210,32 +216,33 @@ struct TestSuite(Movable):
...
+                runs.unsafe_set(i, e)
+            durations.unsafe_set(i, perf_counter_ns() - start)
+
+        parallelize[_run_test](n_tests, workers.or_else(parallelism_level()))
+
+        for test, error, duration in zip(self.tests, runs, durations):
+            var name = Text[Color.CYAN](test.name)
+            var nsec = _format_nsec(duration)
+            if not error:
+                runtime += duration
```
NathanSWard: `runtime += duration` could get refactored out of the `if/else` to avoid some duplication.

mojo/stdlib/stdlib/testing/suite.mojo
```diff
@@ -191,9 +193,13 @@ struct TestSuite(Movable):
         var replacement = String("\n", _Indent("", level=Self._ErrorIndent))
         return e.__str__().replace("\n", replacement)
 
-    def run(deinit self):
+    fn run(deinit self, workers: Optional[Int] = None) raises:
```
NathanSWard: Out-of-scope of this PR, but we should provide a way for individual tests to be run synchronously too.
There are cases for this where a test may use a global or rely on some shared state other tests may use meaning those should be run synchronously.

NathanSWard: There may also need to be a bit of design work here.
In the `stdlib` `bazel` (iirc) will already run multiple test files in parallel - meaning if each of the test files themselves are parallelized this may end up potentially end up adding overhead. We should probably benchmark/tests this to ensure tests don't take longer to run overall.
(None the less, this is still can be useful)

mojo/stdlib/stdlib/testing/suite.mojo
```diff
@@ -210,32 +216,33 @@ struct TestSuite(Movable):
...
-                    _Indent(passed, level=2),
-                    "[",
-                    _format_nsec(duration),
-                    "]",
-                    name,
-                )
+                self.tests[i].test_fn()
             except e:
+                runs.unsafe_set(i, e)
+            durations.unsafe_set(i, perf_counter_ns() - start)
```
NathanSWard, line 228: Since this is test code, we may as well just use the safer alternative of `__getiem__` since the overhead of a bounds check is nothing in comparison to the test suite as a whole.
Also prevents future code changes from silently introducing UB if for some reason `i` is out of bounds.

NathanSWard:
This seems to break some `FILE-CHECK` tests :( 
Ideally we should move off of `file-check` and use the regular `assert_equal/true/false`.
In the meantime, we could override the defaulted "parallelize" behavior for those tests and have them run synchronously

mojo/stdlib/stdlib/testing/suite.mojo
martinvuyk: I think we can just let people separate those out into another file and specify the amount of workers for the test or make the whole test suite run sequentially

martinvuyk: Yeah we should benchmark it. Worst case scenario we can make the default be 1 worker and add a boolean argument that sets the amount of workers to the runtime's parallelism level

NathanSWard: Maybe - but forcing another entire test file doesn't seem super great from a usability/simplicity aspect.
My thought is a `@sync_test` decorator or `suite.sync[file..]` would be nice. 

This is a bit out-of-scope though so we can revisit this later.

martinvuyk: Using a decorator might scale better when combined with the function discovery mechanism

martinvuyk:
@NathanSWard I went with making using 1 worker the default. There was a single function in the `test_string` suite that I had to take over to the python tests because it is not thread safe; so yeah we will need a way to mark how the tests should be run at a function level by using something like a decorator

laszlokindrat:
@martinvuyk I'm apprehensive about this change for the reasons @NathanSWard mentioned: bazel already takes care of parallel execution of these tests, and with remote execution and caching mixed in, there are diminishing returns on these optimizations (or they can become outright slowdowns). What is the motivation for this change? Do they actually speed up your local test runs? Do you have an external project that could benefit from this?

martinvuyk:
List's tests are 20 microseconds slower, and String's around 500 microseconds slower when using the parallelized version (when removing the PythonObject test, that test alone takes 49 ms). So yeah, I'm going to close this

PR #5447, opened 2025-10-14
[stdlib] Fold redundant overload of `get_linkage_name` into a default argument

soraros:


laszlokindrat (approved):
Nice catch, thank you!

laszlokindrat:
!sync

PR #5448, opened 2025-10-14
[stdlib] Minor cleanups to `compile_info`

soraros:


mojo/stdlib/stdlib/compile/compile.mojo
```diff
@@ -63,8 +63,8 @@ struct _Info:
 
     var asm: __mlir_type.`!kgen.string`
     var module_name: __mlir_type.`!kgen.string`
-    var num_captures: __mlir_type.index
-    var capture_sizes: __mlir_type.`!kgen.pointer<ui64>`
+    var num_captures: Int
```
soraros: This might not be desirable, since we are moving towards `Scalar[int]`, and the correspondence between `Int` and `mlir.index` no longer exists.

laszlokindrat: Yeah, let's not ask for trouble.

laszlokindrat (approved):
Thanks!

laszlokindrat:
!sync

PR #5450, opened 2025-10-14
[stdlib] Minor cosmetic and consistency improvements in `python.mojo`…

soraros:
… for better readability

- Follow the convention that `PyObjectPtr` variables use the `_ptr` suffix, while `PythonObject` variables do not. This improves clarity when both types appear together.
- Rename the `Python().cpython()` instance to `cpy` (from `cpython`) to reduce visual noise and line length.
- Use `unsafe_get_error` in places where the presence of an error is guaranteed by CPython.
- Name the result of CPython API calls that return error codes `errno`, consistently distinguishing them from APIs that return `PyObject*`, where `NULL` indicates an error.

### DIFF
--- a/mojo/stdlib/stdlib/python/python.mojo
+++ b/mojo/stdlib/stdlib/python/python.mojo
@@ -33,18 +33,14 @@ from ._cpython import (
 )
 from .python_object import PythonObject
 
-alias _PYTHON_GLOBAL = _Global["Python", _init_python_global]
-
-
-fn _init_python_global() -> _PythonGlobal:
-    return _PythonGlobal()
+alias _PYTHON_GLOBAL = _Global["Python", _PythonGlobal.__init__]
 
 
 struct _PythonGlobal(Defaultable, Movable):
     var cpython: CPython
 
     fn __init__(out self):
-        self.cpython = CPython()
+        self.cpython = {}
 
     fn __del__(deinit self):
         self.cpython.destroy()
@@ -57,13 +53,13 @@ fn _get_python_interface() raises -> Pointer[CPython, StaticConstantOrigin]:
     this global variable. Once it is initialized, it may not be mutated.
     """
 
-    var ptr = _PYTHON_GLOBAL.get_or_create_indexed_ptr(_Global._python_idx)
-    var ptr2 = (
-        UnsafePointer(to=ptr[].cpython)
+    var python = _PYTHON_GLOBAL.get_or_create_indexed_ptr(_Global._python_idx)
+    var cpython_instance = (
+        UnsafePointer(to=python[].cpython)
         .as_immutable()
         .unsafe_origin_cast[StaticConstantOrigin]()
     )
-    return Pointer(to=ptr2[])
+    return Pointer(to=cpython_instance[])
 
 
 struct Python(Defaultable, ImplicitlyCopyable):
@@ -114,9 +110,9 @@ struct Python(Defaultable, ImplicitlyCopyable):
             `True` if the code executed successfully or `False` if the code
             raised an exception.
         """
-        ref cpython = self.cpython()
+        ref cpy = Self().cpython()
         # return 0 if the code executed successfully, -1 if it raised an exception.
-        return cpython.PyRun_SimpleString(code^) == 0
+        return cpy.PyRun_SimpleString(code^) == 0
 
     @staticmethod
     fn evaluate(
@@ -196,8 +192,7 @@ struct Python(Defaultable, ImplicitlyCopyable):
             dir_path: The path to a Python module you want to import.
         """
         var sys = Python.import_module("sys")
-        var directory: PythonObject = dir_path
-        _ = sys.path.append(directory)
+        _ = sys.path.append(PythonObject(dir_path))
 
     # ===-------------------------------------------------------------------===#
     # PythonObject "Module" Operations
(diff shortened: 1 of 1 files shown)


laszlokindrat:
I really like the PR description in this patch! Could you please add these style guidelines to the `python.mojo` and maybe even the `_cpython.mojo` files? It could be just some comments after the module docstring.

mojo/stdlib/stdlib/python/python.mojo
```diff
@@ -114,9 +110,9 @@ struct Python(Defaultable, ImplicitlyCopyable):
             `True` if the code executed successfully or `False` if the code
             raised an exception.
         """
-        ref cpython = self.cpython()
+        ref cpy = Self().cpython()
```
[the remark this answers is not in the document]
laszlokindrat: I'm also curious; @soraros why this change?

mojo/stdlib/stdlib/python/python.mojo
```diff
@@ -230,10 +225,10 @@ struct Python(Defaultable, ImplicitlyCopyable):
         ref cpy = Self().cpython()
         cpy.check_init_error()
 
-        var module_ptr = cpy.PyImport_ImportModule(module^)
-        if not module_ptr:
+        var mod_ptr = cpy.PyImport_ImportModule(module^)
```
laszlokindrat, line 228: Q: Why shorten this? `module_ptr` seemed okay. (I'm fine with either, just want to understand)

mojo/stdlib/stdlib/python/python.mojo
```diff
@@ -357,22 +343,19 @@ struct Python(Defaultable, ImplicitlyCopyable):
 
         Return value: New reference.
         """
-        ref cpy = Python().cpython()
-        var dict_obj = cpy.PyDict_New()
-
+        ref cpy = Self().cpython()
+        var dict_ptr = cpy.PyDict_New()
```
laszlokindrat, line 347: I noticed this in previous patches, so wanted to ask you: you don't check `dict_ptr` here, even though `PyDict_New` can return null on failure. I think this is okay in most cases because the user of `dict_ptr` in the loop will fail and we do check for that. The subtle problem occurs if we don't enter the loop (i.e. `kwargs` is empty). So this means this function in theory can return null, which is probably not what we want, especially since we are already raising and checking after every entry. WDYT?

mojo/stdlib/stdlib/python/python.mojo
```diff
@@ -454,14 +430,12 @@ struct Python(Defaultable, ImplicitlyCopyable):
...
-        var obj_ptr = cpython.PyList_New(len(values))
-
+        ref cpy = Self().cpython()
+        var list_ptr = cpy.PyList_New(len(values))
         for i in range(len(values)):
             var obj = values[i].copy().to_python_object()
-            _ = cpython.PyList_SetItem(obj_ptr, i, obj.steal_data())
-
-        return PythonObject(from_owned=obj_ptr)
+            _ = cpy.PyList_SetItem(list_ptr, i, obj.steal_data())
```
laszlokindrat, line 437: As a followup from my other comment: you're not checking the error code here, so what if `PyList_New` fails for some reason?

laszlokindrat (approved):
LGTM, I just have a few questions and quibbles. Thank you!

mojo/stdlib/stdlib/python/python.mojo
soraros: I think I used `mod` else where. Will revert, I think `module` looks fine.

mojo/stdlib/stdlib/python/python.mojo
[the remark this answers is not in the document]
soraros: It was a mechanical change that I didn't give much thought to. Now that I think about it, why is `eval` a regular method while all the others are `@staticmethod`?

mojo/stdlib/stdlib/python/python.mojo
soraros: `PyDict_New` can only fail when the system runs out of memory, in which case there isn't much we can do. The same applies to `Py[Tuple|List]_New`. They do have other failure modes, but those indicate a bug in our implementation (e.g. a variadic pack producing a negative `len`). This is ultimately a policy question: we could throw on all of them just to be safe (Martin made similar points [here](https://github.com/modular/modular/pull/5451#issuecomment-3403877838)). What do you think?

mojo/stdlib/stdlib/python/python.mojo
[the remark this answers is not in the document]
laszlokindrat: Probably accidental complexity :) Feel free to fix it up in a followup!

mojo/stdlib/stdlib/python/python.mojo
laszlokindrat: > `PyDict_New` can only fail when the system runs out of memory

My understanding is that it can fail if python fails to allocate memory; can that only happen on OOM?

> This is ultimately a policy question: we could throw on all of them just to be safe

Since all these methods already raise, and the actual cost of the check is pretty small, I'm inclined to do the check. At the same time, I don't see a way to actually test these, and I'm also not sure how the user would recover from these situations.

laszlokindrat:
!sync

PR #5451, opened 2025-10-14
[stdlib] Avoid mem leaks in `PythonObject` constructors and make them consistent

martinvuyk:
Avoid mem leaks in `PythonObject` constructors and make them consistent

### DIFF
--- a/mojo/stdlib/stdlib/python/python.mojo
+++ b/mojo/stdlib/stdlib/python/python.mojo
@@ -260,8 +260,7 @@ struct Python(Defaultable, ImplicitlyCopyable):
 
     @staticmethod
     fn add_functions(
-        module: PythonObject,
-        var functions: List[PyMethodDef],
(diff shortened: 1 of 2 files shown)


soraros:
FYI, I’m planning to do this (decref on the error path) in one go after #5444 and #5450.

We deliberately ignore the return value of `Py[List|Tuple]_SetItem`, which can only fail if the index is out of bounds or the interpreter runs out of memory. The former is effectively guaranteed not to happen given our implementation (otherwise it's a bug in the stdlib), and there's little we can do about the latter, so it's reasonable to ignore it. The same reasoning applies to `PySet_New`, etc.

martinvuyk:
Hi @soraros thanks for the heads up. I'll close this then.

> or the interpreter runs out of memory

IMO we should totally raise even then. Otherwise we would give no way to recover from it. The end-user can decide whether they care about malloc failures or not. I would also like us to have the option to raise parametrically for all allocations so... this is more or less the same

PR #5454, opened 2025-10-14
[stdlib] Move `PythonObject` test from `test_string` to `test_python_object`

martinvuyk:
Not sure if there is some kind of warmup for the interpreter that might make this execute faster in the same file as the other python object tests, but it certainly slows string's tests considerably (48ms of the total that is less than 49 ms)


laszlokindrat (approved):
Thanks! Yeah, this is something I have wanted to do for some time. Struct extensions will help control the direction of dependency in the stdlib, and the tests need to follow that structure eventually (regardless of the fact that moving python tests together is probably the right choice for perf as well).

laszlokindrat:
!sync

PR #5456, opened 2025-10-15
[stldib] Rename `Span.apply[cond]` to `when`

martinvuyk:
This is a bit more expressive and readable at a glance

### DIFF
--- a/mojo/stdlib/stdlib/memory/span.mojo
+++ b/mojo/stdlib/stdlib/memory/span.mojo
@@ -688,7 +688,7 @@ struct Span[
         O: MutableOrigin, //,
         func: fn[w: Int] (SIMD[dtype, w]) capturing -> SIMD[dtype, w],
         *,
-        cond: fn[w: Int] (SIMD[dtype, w]) capturing -> SIMD[DType.bool, w],
+        when: fn[w: Int] (SIMD[dtype, w]) capturing -> SIMD[DType.bool, w],
     ](self: Span[Scalar[dtype], O]):
         """Apply the function to the `Span` inplace where the condition is
         `True`.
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -688,7 +688,7 @@ struct Span[
         O: MutableOrigin, //,
         func: fn[w: Int] (SIMD[dtype, w]) capturing -> SIMD[dtype, w],
         *,
-        cond: fn[w: Int] (SIMD[dtype, w]) capturing -> SIMD[DType.bool, w],
```
soraros, line 691: I think naming this function `cond` is quite common in numerical code. I don't have a very strong opinion though.

mojo/stdlib/test/memory/test_span.mojo
```diff
@@ -293,7 +293,7 @@ def test_apply():
         return x * 2
 
     @parameter
-    fn _where[D: DType, w: Int](x: SIMD[D, w]) -> SIMD[DType.bool, w]:
+    fn _is_even[D: DType, w: Int](x: SIMD[D, w]) -> SIMD[DType.bool, w]:
```
soraros, line 296: I *think* the more "pythonic" name would be simply even? Again, I'm fine with either.

martinvuyk: It's just a testing function :man_shrugging:

mojo/stdlib/stdlib/memory/span.mojo
```diff
@@ -714,12 +714,12 @@ struct Span[
...
                     var p_curr = ptr + processed
                     var vec = p_curr.load[width=w]()
-                    p_curr.store(cond(vec).select(func(vec), vec))
+                    p_curr.store(when(vec).select(func(vec), vec))
                     processed += w
 
         for i in range(length - processed):
             var vec = ptr[processed + i]
-            if cond(vec):
+            if when(vec):
```
laszlokindrat, line 722: The way this reads is awkward :( The `if` followed by `when` is strange, and then the variable named `vec` adds to the confusion. TBH I prefer `cond` over this.

martinvuyk: ok, I'll just close the PR

PR #5458, opened 2025-10-16
[stdlib] Simplify List iterable ctor and enable type inference

martinvuyk:
Simplify List iterable ctor and enable type inference

### DIFF
--- a/mojo/stdlib/stdlib/collections/list.mojo
+++ b/mojo/stdlib/stdlib/collections/list.mojo
@@ -329,20 +329,19 @@ struct List[T: Copyable & Movable](
         for value in span:
             self.append(value.copy())
 
-    fn __init__[
-        IterableType: Iterable
-    ](out self, iterable: IterableType) where _type_is_eq_parse_time[
-        T, IterableType.IteratorType[__origin_of(iterable)].Element
-    ]():
+    fn __init__(
+        var iterable: Some[Iterable],
+        out self: List[__type_of(iterable).IteratorType[__origin_of()].Element],
+    ):
         """Constructs a list from an iterable of values.
 
         Args:
             iterable: The iterable of values to populate the list with.
         """
         var lower, _ = iter(iterable).bounds()
-        self = Self(capacity=lower)
-        for value in iterable:
-            self.append(rebind[T](value).copy())
+        self = {capacity = lower}
+        for var value in iterable:
+            self.append(rebind_var[__type_of(self).T](value^))
 
     @always_inline
     fn __init__(out self, *, unsafe_uninit_length: Int):
--- a/mojo/stdlib/test/collections/test_list.mojo
+++ b/mojo/stdlib/test/collections/test_list.mojo
@@ -899,7 +899,7 @@ def test_list_init_span():
 def test_list_init_iter():
     var l = [String("a"), "bb", "cc", "def"]
     var it = iter(l)
-    var l2 = List[String](it)
+    var l2 = List(it)
     assert_equal(len(l), len(l2))
     assert_equal(l[0], l2[0])
     assert_equal(l[1], l2[1])


mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -329,20 +329,19 @@ struct List[T: Copyable & Movable](
...
         Args:
             iterable: The iterable of values to populate the list with.
         """
         var lower, _ = iter(iterable).bounds()
-        self = Self(capacity=lower)
-        for value in iterable:
-            self.append(rebind[T](value).copy())
+        self = {capacity = lower}
+        for var value in iterable:
+            self.append(rebind_var[__type_of(self).T](value^))
```
soraros, line 344: Is this beneficial? We don't have a consuming iteration, and it's unavoidable to copy once. Does the old spelling somehow cause it to copy twice?

martinvuyk: I think this will be easier to transition into a consuming iterator later on. Something like:
```mojo
fn __init__(deinit iterator, out self):
  for var elem in iterator: ...
```

mojo/stdlib/stdlib/collections/list.mojo
```diff
@@ -329,20 +329,19 @@ struct List[T: Copyable & Movable](
...
 
-    fn __init__[
-        IterableType: Iterable
-    ](out self, iterable: IterableType) where _type_is_eq_parse_time[
-        T, IterableType.IteratorType[__origin_of(iterable)].Element
-    ]():
+    fn __init__(
+        var iterable: Some[Iterable],
+        out self: List[__type_of(iterable).IteratorType[__origin_of()].Element],
+    ):
```
NathanSWard, line 335: I would personally prefer keeping `IterableType` in the parameter list since then you don't need to have `__type_of`.

```suggestion
    fn __init__[IterableType: Iterable](
        var iterable: IterableType,
        out self: List[IterableType.IteratorType[__origin_of()].Element],
    ):
```

Also I'm not a huge fan of exposing the `__origin_of()` in the signature of the function but it's unfortunate since we cannot write `__origin_of(iterable)` since it could be a register-passable type 😢 .

I wonder if `Iterable` should also include an `Element` alias so we could just write
`IterableType.Element`.
Since we're inspired by Rust, Rust's [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) uses this pattern and we could leverage it here.

soraros: I still think we should add `Element` to the `Iterable` trait and put a constraint `Element == IteratorType.Element` there (just like Rust). Hopefully `where` can be extended to put there soon.

martinvuyk: > I would personally prefer keeping IterableType in the parameter list since then you don't need to have __type_of.

Yeah I don't quite like it either for readability. But I also am a bit lazy and don't want to document a whole parameter just for that.

> Also I'm not a huge fan of exposing the __origin_of() in the signature of the function but it's unfortunate since we cannot write __origin_of(iterable) since it could be a register-passable type 😢 .

Not just that, but also the compiler seems to think there are exclusivity issues when trying to append to the list due to the elements having the "same origin" as the iterator

> I still think we should add Element to the Iterable trait and put a constraint Element == IteratorType.Element there (just like Rust). Hopefully where can be extended to put there soon.

That would be useful. But is it even possible? can we have two traits overlapping on the same alias?

NathanSWard: My thought would be:

```mojo
trait Iterable:
    alias Element
    
    fn __iter__(..): ...
    
trait Iterator(Iterable):
    fn __has_next__(self) -> Bool: ...
    fn __next__(self) -> Self.Element: ... # (Comes from Iterable)
```

martinvuyk: Ah ok so you mean taking the alias "up a level" not defining a new one, gotcha

lsh: I don't think `__origin_of()` is correct here. This fails type equality `Pointer[Int, __origin_of(a)] == Pointer[Int, __origin_of(b)]` if `a != b`. The origin in that parameter refers to the memory location, so I think this could lead to subtle bugs

martinvuyk: The origin is irrelevant here since we need the type of the element not the iterator return. I can put the `__origin_of(iterable)` but then get exclusivity errors when doing `self.append(value)` because the type system thinks the list's type is related with the iterable's origin which it actually isn't.

For correctness' sake I think we should do the change @NathanSWard showed of taking the element up into the iterable trait. But until then we can keep this ugly workaround as a placeholder

lsh: Consider:

```mojo
struct PointerIterator[origin: ImmutableOrigin](Iterator):
    alias Element = Pointer[Int, origin]
```

My point is that Element itself can look like this and I want to make sure that would pass a type equality test

martinvuyk: One of the motivations behind #5463 (due to finding the bug #5460 that shows origins are ignored after parse time) which indirectly blocks #5459

NathanSWard:
@martinvuyk - I'm planning on closing out this PR for now since we don't have a great final decision on how to write these constructors. Notably we don't want to use `origin_of()` and I believe the current constructor is _okay_ for now.
Iterators are likely getting a rework soon too.

martinvuyk:
Yeah makes sense to wait for a better defined iterator design :+1:

PR #5480, opened 2025-10-19
[stdlib] Add PyBool_Check and PyLong_Check* support

winding-lines:
When writing a Mojo module to interface with Python it is useful to be able to check the types of the input arguments.

This PR adds some missing APIs to check the type of PythonObject. The initial implementation handles boolean and long.

### DIFF
--- a/mojo/stdlib/stdlib/python/_cpython.mojo
+++ b/mojo/stdlib/stdlib/python/_cpython.mojo
@@ -32,6 +32,7 @@ from sys.ffi import (
     c_size_t,
     c_ssize_t,
     c_uint,
+    c_ulong,
 )
 
 from utils import Variant
@@ -52,6 +53,17 @@ alias Py_func_type_input: c_int = 345
 # 0 when Stackless Python is disabled
 # ref: https://github.com/python/cpython/blob/main/Include/object.h
 alias Py_TPFLAGS_DEFAULT = 0
+
+# These flags are used to determine if a type is a subclass.
+# ref: https://github.com/python/cpython/blob/main/Include/object.h
+alias Py_TPFLAGS_LONG_SUBCLASS = (1 << 24)
+alias Py_TPFLAGS_LIST_SUBCLASS = (1 << 25)
+alias Py_TPFLAGS_TUPLE_SUBCLASS = (1 << 26)
+alias Py_TPFLAGS_BYTES_SUBCLASS = (1 << 27)
+alias Py_TPFLAGS_UNICODE_SUBCLASS = (1 << 28)
+alias Py_TPFLAGS_DICT_SUBCLASS = (1 << 29)
+alias Py_TPFLAGS_BASE_EXC_SUBCLASS = (1 << 30)
+alias Py_TPFLAGS_TYPE_SUBCLASS = (1 << 31)
 
 
 # TODO(MOCO-1138):
@@ -971,6 +983,11 @@ alias PyType_FromSpec = ExternalFunction[
     # PyObject *PyType_FromSpec(PyType_Spec *spec)
     fn (UnsafePointer[PyType_Spec]) -> PyObjectPtr,
 ]
+alias PyType_GetFlags = ExternalFunction[
+    "PyType_GetFlags",
+    # unsigned long PyType_GetFlags(PyTypeObject *type)
+    fn (PyTypeObjectPtr) -> c_ulong,
+]
 
 # Integer Objects
 alias PyLong_FromSsize_t = ExternalFunction[
@@ -1338,12 +1355,14 @@ struct CPython(Defaultable, Movable):
     var _PyType_GenericAlloc: PyType_GenericAlloc.type
     var _PyType_GetName: PyType_GetName.type
     var _PyType_FromSpec: PyType_FromSpec.type
+    var _PyType_GetFlags: PyType_GetFlags.type
     # The None Object
     var _Py_None: PyObjectPtr
     # Integer Objects
     var _PyLong_FromSsize_t: PyLong_FromSsize_t.type
     var _PyLong_FromSize_t: PyLong_FromSize_t.type
     var _PyLong_AsSsize_t: PyLong_AsSsize_t.type
+    var _PyLong_Type: PyTypeObjectPtr
     # Boolean Objects
     var _PyBool_FromLong: PyBool_FromLong.type
     # Floating-Point Objects
@@ -1501,6 +1520,7 @@ struct CPython(Defaultable, Movable):
         else:
             self._PyType_GetName = _PyType_GetName_dummy
         self._PyType_FromSpec = PyType_FromSpec.load(self.lib)
+        self._PyType_GetFlags = PyType_GetFlags.load(self.lib)
         # The None Object
         if self.version.minor >= 13:
             # Py_GetConstantBorrowed is part of the Stable ABI since version 3.13
@@ -1521,6 +1541,10 @@ struct CPython(Defaultable, Movable):
         self._PyLong_FromSsize_t = PyLong_FromSsize_t.load(self.lib)
         self._PyLong_FromSize_t = PyLong_FromSize_t.load(self.lib)
         self._PyLong_AsSsize_t = PyLong_AsSsize_t.load(self.lib)
+        self._PyLong_Type = PyTypeObjectPtr(
+            # PyTypeObject PyLong_Type
+            self.lib.get_symbol[PyTypeObject]("PyLong_Type")
+        )
         # Boolean Objects
         self._PyBool_FromLong = PyBool_FromLong.load(self.lib)
         # Floating-Point Objects
@@ -2237,6 +2261,19 @@ struct CPython(Defaultable, Movable):
         """
         return self._PyType_FromSpec(spec)
 
+    fn PyType_GetFlags(
+        self,
+        type: PyTypeObjectPtr,
+    ) -> c_ulong:
+        """Return the feature flags for the `type` pointer.
+
+        Return value: Feature flags.
+
+        References:
+        - https://docs.python.org/3/c-api/type.html#c.PyType_GetFlags
+        """
+        return self._PyType_GetFlags(type)
+
     # ===-------------------------------------------------------------------===#
     # The None Object
     # ref: https://docs.python.org/3/c-api/none.html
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -53,6 +54,17 @@ alias Py_func_type_input: c_int = 345
 # ref: https://github.com/python/cpython/blob/main/Include/object.h
 alias Py_TPFLAGS_DEFAULT = 0
 
+# These flags are used to determine if a type is a subclass.
+# ref: https://github.com/python/cpython/blob/main/Include/object.h
+alias Py_TPFLAGS_LONG_SUBCLASS = (1 << 24)
```
soraros: Please wrap these in constructors or add type annotations.

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -1338,14 +1355,17 @@ struct CPython(Defaultable, Movable):
...
     var _PyType_GetName: PyType_GetName.type
     var _PyType_FromSpec: PyType_FromSpec.type
+    var _PyType_GetFlags: PyType_GetFlags.type
     # The None Object
     var _Py_None: PyObjectPtr
     # Integer Objects
     var _PyLong_FromSsize_t: PyLong_FromSsize_t.type
     var _PyLong_FromSize_t: PyLong_FromSize_t.type
     var _PyLong_AsSsize_t: PyLong_AsSsize_t.type
+    var _PyLong_Type: PyTypeObjectPtr
```
soraros: Please move this before the other functions (of the same section), since we order everything (fields, type def, methods) strictly following Python's documentation pages. Do the same for the other fields you added. It might feel a bit rigid, but I'd like to keep everything very organised and delegate the "boring" or potentially bikeshedding-prone details to Python.

I should probably put these in the style guide, too. What do you think, @laszlokindrat?

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -2237,6 +2266,19 @@ struct CPython(Defaultable, Movable):
...
+    fn PyType_GetFlags(
+        self,
+        type: PyTypeObjectPtr,
+    ) -> c_ulong:
+        """Return the feature flags for the `type` pointer.
+
+        Return value: Feature flags.
+
+        References:
+        - https://docs.python.org/3/c-api/type.html#c.PyType_GetFlags
```
soraros: We reproduce the Python documentation verbatim in these functions; please see the other docstrings as examples; same elsewhere.

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -2290,6 +2332,28 @@ struct CPython(Defaultable, Movable):
         """
         return self._PyLong_AsSsize_t(pylong)
 
+    fn PyLong_Check(self, obj: PyObjectPtr) -> Bool:
```
soraros: We follow the C api exactly, and this function should return `c_int`.

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -2730,3 +2804,9 @@ struct CPython(Defaultable, Movable):
 
         # TODO(MSTDL-950): Should use something like `addr_of!`
         return obj._unsized_obj_ptr[].object_type
+
+    fn PyType_HasFeature(self, ptr: PyTypeObjectPtr, feature: c_ulong) -> Bool:
```
soraros: Can we add a field to the struct instead of reproducing the implementation here? Anyways, please include reference links and fix the docstring.

soraros (commented):
Thanks for the PR @winding-lines! Sorry for the delay as I was sick for a few days.

I think the changes are good; whether we should include them in the stdlib is something the team has to decide, though. The current policy is that we only include what the stdlib itself uses, and `_cpython` is a private module. I'm not sure how complete our `extension` is right now, but maybe we could put these in an extension to the `CPython` struct.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Added, good call.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Heh, great idea. Reshuffled.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Copied the python documentation here and the other places.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Oops, changed this and a few other places.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: I am not sure I fully understand, I've added a method to the struct. Let me know if you want something different 🤗

winding-lines: Ready for another pass at your convenience.

mojo/stdlib/stdlib/python/_cpython.mojo
laszlokindrat: @soraros +1 on putting these in the guide as well, thank you!

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -2053,6 +2170,23 @@ struct CPython(Defaultable, Movable):
...
+        Note: this is a static inline function in the Python C API.
+        https://github.com/python/cpython/blob/3dab11f888fda34c02734e4468d1acd4c36927fe/Include/object.h#L431
+
+        References:
+        - https://docs.python.org/3/c-api/object.html#c.PyObject_TypeCheck
+        """
+        var type_ptr = self.Py_TYPE(obj)
+        if (type_ptr == type) or self._PyType_IsSubtype(type_ptr, type):
+            return 1
+        return 0
```
laszlokindrat: Can you please just add a `_PyObject_TypeCheck` field and delegate to cpython as in the other methods?

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -1335,18 +1358,98 @@ struct CPython(Defaultable, Movable):
...
+    var _PyType_IsSubtype: PyType_IsSubtype.type
     var _PyType_GenericAlloc: PyType_GenericAlloc.type
     var _PyType_GetName: PyType_GetName.type
     var _PyType_FromSpec: PyType_FromSpec.type
     # The None Object
     var _Py_None: PyObjectPtr
     # Integer Objects
+    var _PyLong_Type: PyTypeObjectPtr
+
+    fn _PyLong_Check(self, obj: PyObjectPtr) -> c_int:
```
soraros: Sorry if I wasn't being clear before. You could/should just inline these implementations, as long as you leave the comment "This is a macro in the Python C API".

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -1335,18 +1358,98 @@ struct CPython(Defaultable, Movable):
     var _PyIter_Next: PyIter_Next.type
     # Concrete Objects Layer
     # Type Objects
+    var _PyType_GetFlags: PyType_GetFlags.type
+
+    fn _PyType_HasFeature(
```
soraros: If we can load this from the library, please just add a field instead of reproducing the implementation in Mojo.

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -2198,6 +2332,40 @@ struct CPython(Defaultable, Movable):
...
     # ===-------------------------------------------------------------------===#
 
+    fn PyType_GetFlags(
+        self,
+        type: PyTypeObjectPtr,
+    ) -> c_ulong:
+        """Return the feature flags for the `type` pointer.
+
+        Return value: Feature flags.
+
```
soraros:
```suggestion
        """Return the `tp_flags` member of type.

```

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -2255,6 +2423,26 @@ struct CPython(Defaultable, Movable):
     # ref: https://docs.python.org/3/c-api/long.html
     # ===-------------------------------------------------------------------===#
 
+    fn PyLong_Check(self, obj: PyObjectPtr) -> c_int:
+        """Return true if its argument is a PyLongObject or a subtype of PyLongObject. This function always succeeds.
```
soraros: Please break the line at 80 char, same else where.

mojo/stdlib/stdlib/python/_cpython.mojo
soraros: I meant you could implement this like `PyType_GetFlags`.

soraros:
@winding-lines It's looking pretty good modulo a few nits. We are getting there!

mojo/stdlib/stdlib/python/_cpython.mojo
```diff
@@ -1335,18 +1358,98 @@ struct CPython(Defaultable, Movable):
...
+    fn _PyType_HasFeature(
+        self, ptr: PyTypeObjectPtr, feature: c_ulong
+    ) -> c_int:
+        """Return non-zero if the type object ptr sets the feature feature. Type features are denoted by single bit flags.
+
+        References:
+        - https://docs.python.org/3.13/c-api/type.html#c.PyType_HasFeature
+        """
+        # Note this is another static helper function in the C API.
+        if (self._PyType_GetFlags(ptr) & feature) != 0:
```
soraros: You could write simply `return c_int(expr)` or `return 1 if expr else 0`.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: I think this is done, let me know if I am missing something.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Unfortunately this is an inline static function defined in a header so it has no representation in the compiled library.
https://github.com/python/cpython/blob/88ad41fa3434444a8f316310bffde50e004156ce/Include/object.h#L802

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Applied.

mojo/stdlib/stdlib/python/_cpython.mojo
soraros: I see. Let's leave a comment like the one we did for macros, so we set a good precedent for the future. Thanks!

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Lines are now wrapping up before 80 chars.

mojo/stdlib/stdlib/python/_cpython.mojo
winding-lines: Changed here and in other places, thanks for the advice.

winding-lines:
Ready for another review @soraros and @laszlokindrat 🙇

mojo/stdlib/stdlib/python/_cpython.mojo
soraros: @winding-lines I think you missed this one.

laszlokindrat (approved):
Thank you!

laszlokindrat:
!sync

winding-lines:
Thank you @soraros and @laszlokindrat :-)

PR #5487, opened 2025-10-21
[stdlib] Add Apple GPU SIMD-group shuffle implementation for gpu.warp

### LINKED ISSUE #5469: [Feature Request] Support `shuffle_*` on Apple GPU
### Review Mojo's priorities

### What is your request?

`shuffle_down`, `shuffle_idx`, `shuffle_up` and `shuffle_xor` stdlib's functions are not supported yet on Apple GPU in `stdlib/gpu/warp.mojo`

```html
<html><head></head><body>
CUDA Function | Metal Equivalent | Description
-- | -- | --
__shfl_sync(mask, value, src_lane) | simd_shuffle(value, src_lane) | Get value from specific lane
__shfl_up_sync(mask, value, delta) | simd_shuffle_up(value, delta) | Get value from lane current - delta
__shfl_down_sync(mask, value, delta) | simd_shuffle_down(value, delta) | Get value from lane current + delta
__shfl_xor_sync(mask, value, mask) | simd_shuffle_xor(value, mask) | Get value from lane current ^ mask

</body></html>
```

Increase broader support for Apple Metal GPU

### Any other details?

_No response_

johnny19436:
Closes #5469

### DIFF
--- a/bazel/mojo.MODULE.bazel
+++ b/bazel/mojo.MODULE.bazel
@@ -1,6 +1,6 @@
 """Define the Mojo toolchain"""
 
-PACKAGE_VERSION = "25.7.0.dev2025110105"
+PACKAGE_VERSION = "25.7.0.dev2025110213"
 
 BASE_URL = "https://dl.modular.com/public/nightly/python"
 
--- a/examples/custom-graph-module/pixi.lock
+++ b/examples/custom-graph-module/pixi.lock
@@ -63,7 +63,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/gflags-2.2.2-h5888daf_1005.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/gguf-0.17.1-pyhc364b38_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/glog-0.7.1-hbabe93e_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/googleapis-common-protos-1.70.0-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/googleapis-common-protos-1.71.0-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/grpcio-1.73.1-py313h2b3948d_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/h11-0.16.0-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/h2-4.3.0-pyhcf101f3_0.conda
@@ -98,7 +98,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libcblas-3.9.0-38_h0358290_openblas.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libcrc32c-1.1.2-h9c3ff4c_0.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libcurl-8.16.0-h4e3cde8_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libdeflate-1.24-h86f0d12_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libdeflate-1.25-h17f619e_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libedit-3.1.20250104-pl5321h7949ede_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libev-4.33-hd590300_2.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libevent-2.1.12-hf998b51_1.conda
@@ -134,7 +134,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libstdcxx-15.2.0-h8f9b012_7.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libstdcxx-ng-15.2.0-h4852527_7.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libthrift-0.22.0-h454ac66_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libtiff-4.7.1-h8261f1e_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libtiff-4.7.1-h9d88235_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libutf8proc-2.11.0-hb04c3b8_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libuuid-2.41.2-he9a06e4_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libuv-1.51.0-hb03c661_1.conda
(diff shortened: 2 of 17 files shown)


johnny19436:

JoeLoser:
Thanks for working on this, @johnny19436!  If you remove the `test_shuffle` file from the extra constraints list [here](https://github.com/modular/modular/blob/53e05e781812f63b8011b03a9dcdc60bbffd7d77/max/kernels/test/gpu/basics/BUILD.bazel#L76), it should try running locally on your Apple Metal GPU setup.  Does the test pass?  I can also sync this PR internally and we can run it on internal macOS CI as well if you run into issues.

johnny19436:
The test_shuffle.mojo.test runs locally but fails to create a compute pipeline due to Metal language version mismatch (3.2 vs OS-supported 3.1). I’m currently on macOS 14.5, so the Metal runtime doesn’t support 3.2 yet.
Is it necessary to upgrade my machine for the version support?

JoeLoser:
> The test_shuffle.mojo.test runs locally but fails to create a compute pipeline due to Metal language version mismatch (3.2 vs OS-supported 3.1). I’m currently on macOS 14.5, so the Metal runtime doesn’t support 3.2 yet. Is it necessary to upgrade my machine for the version support?

Let's loop in @npanchen, @dgurchenkov, and @lsh here to best comment on this specifically. Thanks for trying! One of them will comment today when they're online.

mojo/stdlib/stdlib/gpu/warp.mojo
```diff
@@ -163,6 +164,74 @@ fn _shuffle_amd_helper[
...
+        dtype.is_half_float() or simd_width == 1,
+        "Unsupported simd_width",
+    ]()
+
+    var arg = UInt16(offset)  # AIR intrinsics use 16-bit offsets
+
+    @parameter
+    if dtype is DType.float32:
+        alias name = (
+            "llvm.air.simd_shuffle" + ("" if op == "indexed" else "_" + op) + ".f32"
```
npanchen: I assume you've used release version for testing.
Nightly compiler doesn't require to mangle type into AIR intrinsic name as it's done by the compiler, so you should be able to simplify this code into:
```
@parameter
if dtype in (DType.int64, DType.uint64):
  # emulation here
else:
  alias name = "llvm.air.simd_shuffle" + ("" if op == "indexed" else "_" + op)
  return llvm_intrinsic[name, Scalar[dtype]](val, arg)
```
This is done quite hackie so if you experience any issues with it let me know

mojo/stdlib/stdlib/gpu/warp.mojo
```diff
@@ -163,6 +164,74 @@ fn _shuffle_amd_helper[
...
+
+        var half1_n = rebind[SIMD[DType.uint32, simd_width]](half1)
+        var half2_n = rebind[SIMD[DType.uint32, simd_width]](half2)
+
+        var s1 = _shuffle_apple_helper[op, DType.uint32, simd_width](mask, half1_n, offset)
+        var s2 = _shuffle_apple_helper[op, DType.uint32, simd_width](mask, half2_n, offset)
+
+        var merged = s1.interleave(s2)
+        return bitcast[dtype, simd_width](merged)
+    elif dtype.is_half_float():
```
npanchen: Do we really need this ? From MLS spec half and bfloat types should be supported by `simd_shuffle`.

mojo/stdlib/stdlib/gpu/warp.mojo
```diff
@@ -163,6 +164,74 @@ fn _shuffle_amd_helper[
...
+](
+    mask: UInt,                           # Ignored, for API parity
+    val: SIMD[dtype, simd_width],
+    offset: UInt32
+) -> SIMD[dtype, simd_width]:
+    """
+    Matches the NVIDIA/AMD signature: (mask, val, offset)
+    but ignores `mask` since Metal intrinsics don’t support it.
+
+    Uses llvm.air.simd_shuffle[_down/_up/_xor].[f32|i32] intrinsics.
```
npanchen: IMO: that will be great to provide mapping between Metal's functions and generated AIR intrinsic. That is
```
simd_shuffle -> llvm.air.simd_shuffle
```

npanchen (commented):
Thank you for working on this. There are too many tests that are failing due to missed implementation of these functionalities, so when it will be merged in some of them can be enabled in regular runs.

It will be still great if you can enable https://github.com/modular/modular/blob/main/max/kernels/test/gpu/basics/BUILD.bazel#L76 test too

npanchen:
> The test_shuffle.mojo.test runs locally but fails to create a compute pipeline due to Metal language version mismatch (3.2 vs OS-supported 3.1). I’m currently on macOS 14.5, so the Metal runtime doesn’t support 3.2 yet. Is it necessary to upgrade my machine for the version support?

If you see an error
```
...using language version 3.2 which is incompatible with this OS.
```
then yes, update seems essential as we're not testing on old osx. Which xcode do you use ? It is the second time I hear that compilation succeeded, but execution failed (because metal does JIT). I really want to reproduce it and try to move the error from runtime to compile time.
But if you see
```
Failed to create compute pipeline state: Compiler encountered an internal error
```
then it's way more complex and requires to understand which LLV instruction/type/metadata/attribute/etc Metal compiler doesn't like

johnny19436:
> Thank you for working on this. There are too many tests that are failing due to missed implementation of these functionalities, so when it will be merged in some of them can be enabled in regular runs.
> 
> It will be still great if you can enable https://github.com/modular/modular/blob/main/max/kernels/test/gpu/basics/BUILD.bazel#L76 test too

Sure, I will upgrade and enable https://github.com/modular/modular/blob/main/max/kernels/test/gpu/basics/BUILD.bazel#L76 test.

johnny19436:
> > The test_shuffle.mojo.test runs locally but fails to create a compute pipeline due to Metal language version mismatch (3.2 vs OS-supported 3.1). I’m currently on macOS 14.5, so the Metal runtime doesn’t support 3.2 yet. Is it necessary to upgrade my machine for the version support?
> 
> If you see an error
> 
> ```
> ...using language version 3.2 which is incompatible with this OS.
> ```
> 
> then yes, update seems essential as we're not testing on old osx. Which xcode do you use ? It is the second time I hear that compilation succeeded, but execution failed (because metal does JIT). I really want to reproduce it and try to move the error from runtime to compile time. But if you see
> 
> ```
> Failed to create compute pipeline state: Compiler encountered an internal error
> ```
> 
> then it's way more complex and requires to understand which LLV instruction/type/metadata/attribute/etc Metal compiler doesn't like

I get the error message : 
```
using language version 3.2 which is incompatible with this OS.
```

Environment: 
```
Apple metal version 32023.404 (metalfe-32023.404)
Target: air64-apple-darwin23.5.0
Xcode 16.2
Build version 16C5032a
macOS 14.5
```

Hope this can help with the problem.

mojo/stdlib/stdlib/gpu/warp.mojo
johnny19436: Thanks for pointing this out — you’re right that according to the Metal Shading Language spec (Table 6.14), simd_shuffle is defined for any integer or floating-point T (except bool, void, size_t, ptrdiff_t).
That means in theory both half and bfloat16 should be supported.
However, in practice the current AIR backend doesn’t yet provide air.simd_shuffle.bf16 or air.simd_shuffle.f16 intrinsics. When lowering bf16 kernels, the runtime hits:
```
Failed to create compute pipeline state:
error: BFloat not handled at: air.simd_shuffle.bf16.i16
```
Does that mean MSL declaration exists, but the underlying AIR instruction isn’t implemented for bf16 ?

npanchen: Interesting. It is more complicated than I though. Consider next metal kernel (ignore its correctness):
```
#include <metal_stdlib>
using namespace metal;

#define TYPE half

kernel void simple_simd_shuffle(device TYPE* input [[buffer(0)]]) {
    TYPE my_value = input[0];
    TYPE lane0_value = simd_shuffle(my_value, 0);
    input[0] = lane0_value;
}
```
For this case Metal does generate `tail call fast half @air.simd_shuffle.f16(half %3, i16 0)`. The fact you see `.i16` is my issue in mangler and it has to be removed.

But `bfloat` seems to be unsupported:
```
include/metal/metal_simdgroup:189:14: note: candidate template ignored: requirement '__is_valid_simdgroup_type<bfloat, void>::value' was not satisfied [with T = bfloat]
```
Xcode 26 seems to have the same problem, so not sure if intrinsic of this type is not supported or it's just a header problem.

npanchen: >  The fact you see .i16 is my issue in mangler and it has to be removed.

The way you can overcome this is by manually mangling the type into the function name. However, you will need to use `pop.external_call` instead of `llvm_intrinsic`. In this case your original approach with `if dtype is DType....` will stay, but you will need to manually test for all supported dtypes.

johnny19436:
Closes #5469

mojo/stdlib/stdlib/gpu/warp.mojo
johnny19436: Thanks for the clarification — that makes sense. For now, I’ll stick with the current llvm_intrinsic-based approach and keep the explicit if dtype is DType... handling. Since air.simd_shuffle.f16 is handled correctly and bfloat16 is still unsupported at the AIR level, using direct intrinsics keeps the implementation consistent with the existing NVIDIA and AMD shuffle helpers.

mojo/stdlib/stdlib/gpu/warp.mojo
```diff
@@ -163,6 +164,81 @@ fn _shuffle_amd_helper[
...
+
+        @parameter
+        if simd_width == 1:
+            var pair = SIMD[dtype, 2](val._refine[new_size=1]())
+            var pair_i32 = bitcast[DType.int32, 1](pair)
+            var y_i32 = _shuffle_apple_helper[op, DType.int32, 1](
+                mask, pair_i32, offset
+            )
+            return bitcast[dtype, 2](y_i32)[0]
+        elif (simd_width % 2) == 0:
```
npanchen: even though it makes sense for me to check `simd_width` in general, but current `SIMD` is limited to [power-of-2 VL](https://github.com/modular/modular/blob/main/mojo/stdlib/stdlib/builtin/simd.mojo#L213)
so it seems not possible to hit `else` below.

maybe @JoeLoser can comment here

npanchen (commented):
overall, lgtm. Thanks again for working on this

mojo/stdlib/stdlib/gpu/warp.mojo
JoeLoser: +1 to simplifying things since SIMD is restricted to power of two sizes

JoeLoser (approved):
Thanks for working on this! Happy to sync this in once you rebase to fix the conflict. I think some of my recent bigger GPU module refactorings caused this, sorry about that.

JoeLoser:
Do you mind also adding a new changelog entry [here](https://github.com/modular/modular/blob/805c68b31b0863fe251d9686a23ee0ae325cd5ac/mojo/docs/changelog.md#library-changes-25-7-library-changes) please so others know about your added Apple GPU support? Thanks!

JoeLoser:
Hey @johnny19436, did you mean to close this PR or did it get auto-closed to a rebase gone awry perhaps?

Edit: I see a new PR opened at https://github.com/modular/modular/pull/5537 👍

PR #5491, opened 2025-10-21
[stdlib] Add x y z setters to dim

josiahls:
Adds setters to the DIm struct similar to what we can do with the `dim3` struct from cuda.

An alternative is via function overload: `fn [...]x(self, value: T)`. Let me know if there is a preference.

### DIFF
--- a/mojo/stdlib/stdlib/gpu/host/dim.mojo
+++ b/mojo/stdlib/stdlib/gpu/host/dim.mojo
@@ -193,3 +193,36 @@ struct Dim(Stringable, Writable):
             The value of the x dimension.
         """
         return self[0]
+
+    fn set_z[T: Indexer](mut self, z: T):
+        """Sets the z dimension.
+
+        Parameters:
+            T: An indexer type infered from the argument.
+
+        Args:
+            z: The value for the z dimension.
+        """
+        self._value[2] = index(z)
+
+    fn set_y[T: Indexer](mut self, y: T):
+        """Sets the y dimension.
+
+        Parameters:
+            T: An indexer type infered from the argument.
+
+        Args:
+            y: The value for the y dimension.
+        """
+        self._value[1] = index(y)
+
+    fn set_x[T: Indexer](mut self, x: T):
+        """Sets the x dimension.
+
+        Parameters:
+            T: An indexer type infered from the argument.
+
+        Args:
+            x: The value for the x dimension.
+        """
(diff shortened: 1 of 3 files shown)


lsh:
!sync

lsh:
Ah this fails internally because implicit UInt and Int conversions have been removed. You'll probably have to rebase in a day or two.

lsh:
Or I can fix internally

josiahls:
@lsh I'm afk today, could fix tomorrow, up to you, I'm guessing you could beat me to it

josiahls:
@lsh Resyned with develop. Fixed missing parameter docs. Looks like im getting out of space errorsa.

soraros:
Can we tweak `fn [x|y|z]` to return a ref?

josiahls:
@soraros 
> Can we tweak `fn [x|y|z]` to return a ref?

~~These are setters though, why would they return anything? Unless is your idea that you can chain them then?~~
*edit: Thought for 5 more seconds lol, this makes sense to me. If @lsh is on board with it, I can update my pr to just update the original getters to return a mutable ref instead.

mojo/stdlib/stdlib/gpu/host/dim.mojo
```diff
@@ -193,3 +193,36 @@ struct Dim(Stringable, Writable):
             The value of the x dimension.
         """
         return self[0]
+
+    fn set_z[T: Indexer](mut self, z: T):
```
abduld, line 197: what do you think about using setattr and getattr here (similar to what is done for ThreadIdx struct)?

josiahls: Up to @lsh I'm happy to change this to either just have the original getters just return mutable references or leverage `set/getattr`. 

My thought process was:
- Keep the setter separate so there is a immut api (getter) and mut api (setter).
- setattr and getattr I didn't get the impression was a acceptable use of this (gives python hack vibes). I remember asking about this missing docs, and modular folks felt that there should be friction in using these functions in general with preference toward explicit API. Additionally doesn't allow static type hinting.
    - Additional argument against this, I think setattr/getattr really only should be considered if it isn't possible to statically define these e.g. only in dynamic field instances. 
    - Additionally if we use set/getattr it will be a breaking change.

Might be nice if modular want's to make a programming guideline on what is good/bad practice for getters/setters tbh.

PR #5511, opened 2025-10-25
[stdlib] Replace unnecessary call to `Int.__init__(mlir_value)` with vanilla `Int` ctor

soraros:


soraros (commented):
I'm just trying out Codex.

**Summary**
This PR replaces uses of `Int.__init__(mlir_value=...)` with the ordinary `Int(...)` constructor (and then casts to `UInt`), and simplifies a few local types/expressions in `max/kernels/src/layout/layout_tensor.mojo`. The change trims MLIR‑level plumbing from the stdlib surface and makes the code easier to reason about.

**What changed**
- `UInt(mlir_value=Int(x)._mlir_value)` → `UInt(Int(x))`.
- A couple of `var` temporaries converted to `alias` where values are compile‑time/loop‑invariant.
- Removed redundant `: UInt` annotations where type inference is unambiguous.
- Normalised construction of `fragments_stride_i` from runtime layout with an eager `UInt(...)` so later uses don’t need per‑use casting.

**Correctness & semantics**
- The new construction keeps the same numeric path: `x -> Int(x) -> UInt(...)`. Assuming the source values are non‑negative (which they should be for shapes/strides), semantics are preserved and we avoid depending on `_mlir_value`.
- Using `alias` for loop‑invariant values is a small win for readability and likely for IR as well.
- Arithmetic like `(thread_id // stride_i) % shape_i` remains in unsigned space after your change, which is desirable if `thread_id` is already `UInt`.

**Style/nits**
- The file mixes `alias` and `var` for `fragments_stride_i` (e.g., lines around the second hunk). Consider consistently using `alias` whenever the value is invariant inside the loop.
- Where you removed `: UInt` type annotations (e.g., `thread_coord_i`), inference should be fine given the RHS operands are `UInt`. If there’s any chance `thread_id` is not `UInt` in some specialisations, keeping the explicit annotation may prevent accidental `Int` inference.
- It may be worth adding debug‑mode assertions that `thread_projected_{shape,stride}[i] >= 0` to make the signed→unsigned cast intent explicit.

**Follow‑ups**
- A quick repo‑wide sweep for `mlir_value=` usage in stdlib would be nice to finish the clean‑up.

Overall this looks good to me. LGTM with the minor consistency nit.

laszlokindrat (approved):
good catch, thanks!

laszlokindrat:
!sync

PR #5514, opened 2025-10-27
[stdlib] Enable Python module-level Mojo imports

miili:
Adds module-level imports to `MojoImporter`.

This enables shipping Python packages with mojo code, where the package only needs a runtime dependency on the Mojo package.

```python
from awesome_package.extensions.mojo import factorial

factorial(123)
```

The compiler and hashing logic are the same. The importer will create a `__mojocache__` directory in the module path. I.e. `<package_path>/autumn_package/extensions/mojo/__mojocache__/factorial.ae43a431b521.so`.

* Improved clarity, renamed helper functions and removed implicit comments.
* Replaced `assert` with meaningful `ImportError` raises.
* Implemented `invalidate_caches`, removing compiled Mojo extension.
  https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches
* Compiled files: Removed implicit `hash-` prefix from filename
* Docs: Added documentation of module-level Python imports

### DIFF
--- a/mojo/docs/manual/python/mojo-from-python.mdx
+++ b/mojo/docs/manual/python/mojo-from-python.mdx
@@ -19,13 +19,12 @@ work in progress. See below for [known limitations](#known-limitations).
 
 :::
 
-## Import a Mojo module in Python
+## Import a simple Mojo module in Python
 
-To illustrate what calling Mojo from Python looks like, we'll start with a
-simple example, and then dig into the details of how it works and what is
-possible today.
+We'll start with a simple example, what calling Mojo from Python looks like
+and then dig into the details of how it works and what is possible today.
 
-Consider a project with the following structure:
+Consider the following directory structure of a simple project:
 
 ```text
 project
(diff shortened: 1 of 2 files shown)


miili:

ConnorGray:
!sync

ConnorGray:
Hi @miili thank you so much for making this all around great contribution, especially from a first-time contributor! 🔥

This is a nice improvement to the UX of Python package authors, enabling them to better encapsulate their Mojo extensions. I've started the process of syncing this in to our internal sources. I may make a few minor changes as I integrate this, but all around this looks really good, and thank you again 🙂

ConnorGray:
!sync

YichengDWu:
What's blocking this?

miili:
We should also encode the compiling Mojo version in the filename hash.

miili:
@ConnorGray anything I can do to speed this up?

miili:
@Ahajha @ConnorGray , a bit surprised that this PR did not make it in. The handling of path replaces is a bit hacky imo. This PR leverages the Python Importer API.

https://github.com/modular/modular/commit/63daf53f1b44748a85734fe31aa133c15c9b38d3

Ahajha:
@miili Sorry about that, this got lost in the mix of things. I did glance at the PR when you wrote it (though it's not normally my neck of the woods) and I think this is a good direction. Looking at the imported PR internally, there were a couple issues that could be sorted out to make this go along quicker:
1. There were failing tests, I don't think these were caught at the time here because most of the `max` tests weren't working, but that should be fixed now.
2. I think this PR is trying to do too much, with a functional change like this it's best to make the commit small so it's easy to review the exact change. The docs can be separate, and I would leave any parts of the code that aren't necessary for the functionality change alone. I appreciate the desire to clean things up, but it's a separate concern :)

PR #5529, opened 2025-10-30
[stdlib] Add origin to `RecordingStrategy`

martinvuyk:
Add origin to `RecordingStrategy`.

CC: @NathanSWard just some small detail to avoid any surprises for tests that don't use the list after the last use of the `RecordingStrategy` :)


NathanSWard:
Thank you :)

NathanSWard:
!sync

soraros:
Silly question, why don't we use `Pointer` instead?

Maybe even
```mojo
    fn __init__(out self, ref[origin] list: List[Int]):
        self.list = Pointer(to=list)
```

PR #5531, opened 2025-10-31
[docs] Add note about 1:1 mapping of GPUs to target_arch.

josiahls:
Currently for someone working through this documentation they will be confused when they try to add their GPU if its cuda capability is already mapped to another GPU.


JoeLoser (approved):
LGTM, thanks!

JoeLoser:
!sync

PR #5535, opened 2025-11-01
[stdlib][proposal] Include UnsafePointer V1-V2 comparison in proposal

msaelices:
Show the signatures side by side in the proposal to emphasize the differences.


NathanSWard:
!sync

NathanSWard:
@msaelices, thanks for this :)

PR #5540, opened 2025-11-05
[stdlib] Add `Tuple` logic dunder methods

### LINKED ISSUE #2645: [Feature Request] [stdlib] Add logic dunder methods for Tuple
### Review Mojo's priorities

### What is your request?

With an iterator:
Implementing an iterator and comparing per item is an option.

With pointers:
I still don't quite understand how to "safely" use the unsafe pointer. I think this implementation example would have a problem when say, comparing a Tuple[UInt64, UInt64] to a Tuple[UInt8, UInt8], but I'm not sure since I barely undestand the memcmp implementation.

Could do something like:
```mojo
    fn _comp(self, other: Self) -> Optional[Int]:
        # doesn't have an __eq__ 
        # if self._mlir_type != other._mlir_type:
        if len(self) != len(other):
            return None
        return memcmp(
            UnsafePointer(self.storage).address,
            UnsafePointer(other.storage).address,
            len(self),
        )

    fn __eq__(self, other: Self) -> Bool:
        var result = self._comp(other)
        return result.value()[] == 0 if result else False

    fn __ne__(self, other: Self) -> Bool:
        var result = self._comp(other)
        return result.value()[] != 0 if result else False

    fn __gt__(self, other: Self) -> Bool:
        var result = self._comp(other)
        return result.value()[] == 1 if result else False

    fn __ge__(self, other: Self) -> Bool:
        var result = self._comp(other)
        return result.value()[] != -1 if result else False

    fn __lt__(self, other: Self) -> Bool:
        var result = self._comp(other)
        return result.value()[] == -1 if result else False

    fn __le__(self, other: Self) -> Bool:
        var result = self._comp(other)
        return result.value()[] != 1 if result else False
```

Some nice and intuitive syntax can come out of this
```mojo
var result: Tuple[Int, UInt16, UInt8] = some_func()
if result == (0, 0, 0):
  return "some thing"
return "some other thing"
```

### Any other details?

_No response_

martinvuyk:
Add `Tuple` logic dunder methods. Closes #2645

### DIFF
--- a/mojo/stdlib/stdlib/builtin/tuple.mojo
+++ b/mojo/stdlib/stdlib/builtin/tuple.mojo
@@ -198,7 +198,7 @@ struct Tuple[*element_types: Copyable & Movable](
         """
 
         @parameter
-        for i in range(len(VariadicList(element_types))):
+        for i in range(Self.__len__()):
 
             @parameter
             if _type_is_eq[element_types[i], T]():
@@ -208,13 +208,11 @@ struct Tuple[*element_types: Copyable & Movable](
         return False
 
     @always_inline("nodebug")
-    fn __init__[
-        *elt_types: Copyable & Movable & Defaultable
-    ](out self: Tuple[*elt_types]):
+    fn __init__[*U: Copyable & Movable & Defaultable](out self: Tuple[*U]):
         """Construct a tuple with default-initialized elements.
 
         Parameters:
-            elt_types: The types of the elements contained in the Tuple.
+            U: The types of the elements contained in the Tuple.
         """
 
         # Mark 'self.storage' as being initialized so we can work on it.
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/builtin/tuple.mojo
```diff
@@ -225,3 +225,129 @@ struct Tuple[*element_types: Copyable & Movable](
...
+
+        Args:
+            other: The other tuple to compare with.
+
+        Returns:
+            True if all `self[i] == other[i]`, False otherwise.
+        """
+
+        @parameter
+        for i in range(len(VariadicList(element_types))):
```
lsh: All the `VariadicList(element_types)` need to be `type_of(self).__len__()` instead

lsh (changes requested):

lsh:
!sync

abduld:
I did not see this PR and wanted a similar feature so I've implemented it. Sorry about that.

smart people think a like I guess :)

martinvuyk:
My logic was actually different than the default Python one (I was doing elementwise + reduce_and here), good thing this wasn't merged :sweat_smile:

PR #5542, opened 2025-11-05
test branch for gemma3 for conditional generation

RossCampbellDev:
** DRAFT PR **

this is being made so that Modular engineers can test it against their hardware, to investigate an issue with CUDA

the error occurring upon running `max generate...` being:  `ABORT: dlsym failed:<home>/modular/max/python/max/pipelines/architectures/.pixi/envs/default/bin/python3.13: undefined symbol: cublasCreate_v2`


ross:
> ❌ @ross Campbell

The action that commented above needs some tweaks around @ mentions as it appears to be sourcing them from somewhere that doesn't have correct info.

PR #5544, opened 2025-11-06
[stdlib] Improve `MojoImporter` concurrency safety with file locking

YichengDWu:
#### Motivation

The existing `MojoImporter` implementation did not account for concurrent access from multiple processes. This created several critical race conditions:
1.  **Thundering Herd:** Multiple processes could simultaneously detect a missing cache file and all trigger redundant, resource-intensive compilations.
2.  **Corrupt Cache Reads:** A process could attempt to load a `.so` file while another process was still in the middle of writing it, leading to `ImportError` or segmentation faults.
3.  **Filesystem Inconsistencies:** The implementation was not robust against transient delays common in network or containerized filesystems.

This PR refactors the importer to address these issues systematically, making it suitable for production use in parallel environments.

#### Solution Implemented

A multi-layered defense strategy has been implemented to ensure correctness and robustness:

1.  **Mutual Exclusion via `filelock`:** A per-module lock file (`.lock`) is used to serialize compilation attempts. This enforces a "one writer at a time" policy. A double-checked locking pattern is used for performance, avoiding lock acquisition when the cache is already valid.

2.  **Atomic Cache Updates via `os.replace`:** The "compile-to-temp-and-rename" pattern has been adopted.
    *   Compilation output is directed to a unique temporary file (named with PID and UUID).
    *   Upon successful compilation, `os.replace()` is used to atomically move the temporary file to its final cache path. This guarantees that any process looking for the cache file will see either the old version or the fully complete new version, never a partial file.

3.  **Resilience and Graceful Degradation:**
    *   **Lock Timeouts:** If a process fails to acquire a lock within the timeout, it doesn't immediately fail. It waits briefly and re-checks, allowing it to "piggyback" on a slow compilation completed by another process.
    *   **Filesystem Latency:** A small retry loop with a delay has been added before the final module loading step. This mitigates race conditions caused by metadata propagation delays on filesystems like NFS.
    *   **Tolerant Cleanup:** Deletion of stale cache files is now wrapped in a `try...except` block to prevent failures if an old library is still loaded or locked by another process.

### DIFF
--- a/mojo/python/mojo/importer.py
+++ b/mojo/python/mojo/importer.py
@@ -16,9 +16,13 @@ import logging
 import os
 import subprocess
 import sys
+import time
 from collections.abc import Sequence
 from importlib.util import spec_from_file_location
 from pathlib import Path
+from uuid import uuid4
+
+from filelock import FileLock, Timeout
 
 from .paths import MojoCompilationError, MojoModulePath, find_mojo_module_in_dir
 from .run import subprocess_run_mojo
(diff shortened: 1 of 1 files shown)


owenhilyard:
I like the atomic file swap option to avoid corruption, since that should be fairly reliable. 

However, I'm not a huge fan of the locking since that means that lock breaking will be necessary in some circumstances, such as crashes, and it makes it harder to do parallel cross-compiles in CI and might lead to brittle/stalled CI if you have crashing compiles. In this case, I don't think the reduction in compute usage is worth the extra complexity for automated build systems.

YichengDWu:
> I like the atomic file swap option to avoid corruption, since that should be fairly reliable.
> 
> However, I'm not a huge fan of the locking since that means that lock breaking will be necessary in some circumstances, such as crashes, and it makes it harder to do parallel cross-compiles in CI and might lead to brittle/stalled CI if you have crashing compiles. In this case, I don't think the reduction in compute usage is worth the extra complexity for automated build systems.

That's a very strong point, thank you for the detailed feedback. I completely agree that the risk of stale locks and added complexity outweighs the benefit of avoiding re-compiles, especially in automated CI environments. The robustness of the build system should indeed be the top priority.

I'll continue in another PR to remove the file locking mechanism entirely, while keeping the atomic file swap logic to prevent cache corruption. This approach will accept the trade-off of potential redundant compilations in favor of a simpler and more resilient implementation.

PR #5553, opened 2025-11-08
[SDK] Add entrypoint mode to notebook magic for Python object returns

msaelices:
- Enables returning any Python object (SVG, DataFrames, etc.) from Mojo cells
- Uses mojo.importer for dynamic compilation and Python integration
- Maintains backward compatibility
- Provides better visualization support in Jupyter notebooks

Example of cell:

```mojo
%%mojo draw
from python import Python, PythonObject

fn draw() raises -> PythonObject:
    Digraph = Python.import_module("graphviz").Digraph

    g = Digraph()
    g.node('A')
    g.node('B')
    g.node('C')
    
    return g
```

It's useful for e.g. explaining Mojo code using graphs, as you can see here: https://github.com/msaelices/mojo-unnet/blob/main/notebooks/grad.ipynb


msaelices:
I'm not sure this is the ideal approach, so for now I only did the minimum required to make it work. My main goal was simply to be able to plot SVGs directly in notebooks.

With some guidance, I could implement it properly, in the right way.

msaelices:
Sorry, the following PR is doing the same thing but has a non wanted changed file there. I cannot close (permissions) it but I don't want to confuse people. Please close the following PR: https://github.com/modular/modular/pull/5552

JoeLoser:
> Sorry, the following PR is doing the same thing but has a non wanted changed file there. I cannot close (permissions) it but I don't want to confuse people. Please close the following PR: #5552

Oops, sorry about that: I thought closing it was an accident at first so I had reopened it. Thanks for clarifying: I just closed #5552.

bethebunny:
My general sense is

1) this is very cool! love the demo, great work :D
2) this is probably too specialized for the main mojo hooks

This would be really great for you to publish as a community package. I think to make sense in the base image it would want to be really general, for instance allowing a `%%mojo` cell to behave like a normal python cell, and if the result of the final expression is a PythonObject then displaying it with `display`.

msaelices:
> My general sense is
> 
> 1. this is very cool! love the demo, great work :D
> 2. this is probably too specialized for the main mojo hooks
> 
> This would be really great for you to publish as a community package. I think to make sense in the base image it would want to be really general, for instance allowing a `%%mojo` cell to behave like a normal python cell, and if the result of the final expression is a PythonObject then displaying it with `display`.

Sure, I wanted to do it general but I was not sure if I should get rid of the `subprocess` approach. 

The issue I find is how we can know if a chunk of text (the code in the cell) is going to return an object or not?

If we have this example:
```
%%mojo
print("hello world")
```
We cannot embed this code inside this "template":
```
fn entrypoint() -> PythonObject:
    print("hello world")  # code injected from the cell
```
The above code is not going to compile, as it expect a return sentence.

Any ideas on how I can get this to be general?

ehsanmok:
This is nice but agree it's too specialized and would be suitable for a library not general mojo extension.

JoeLoser:
Given the feedback from @bethebunny and @ehsanmok, I'm closing to close this PR to help clean up the pull request review queue.

PR #5562, opened 2025-11-10
[stdlib] Add support for the mapping protocol to Python interop

winding-lines:
Currently if you define a `__getitem__` method and export the struct to Python you can call the 
given method but you cannot use []. That's because the methods of the Mapping protocol need 
to be plumbed through special `tp_slots` in the type definition.

This PR proposes an approach for how this could be done. I am not familiar with current and future
directions of the Python Interop design. There are many of these type slots here:
https://github.com/python/cpython/blob/main/Include/typeslots.h. The design should accommodate
all those.

Feedback appreciated.

### DIFF
--- a/mojo/integration-test/python-extension-modules/basic-columnar/mojo_module.mojo
+++ b/mojo/integration-test/python-extension-modules/basic-columnar/mojo_module.mojo
@@ -13,13 +13,14 @@
 
 from os import abort
 
-from memory import UnsafePointer, alloc
-from utils import IndexList
+from memory import UnsafePointer
+from utils import IndexList, Variant
 from python import Python, PythonObject
-from python._cpython import PyObjectPtr, Py_LT, Py_GT
+from python._cpython import RichCompareOps
 from python.bindings import (
     PythonModuleBuilder,
     PyTypeObjectSlot,
+    NotImplementedError,
 )
 
 comptime Coord1DColumn = List[Float64]
@@ -56,6 +57,9 @@ struct DataFrame(Defaultable, Movable, Representable):
     var pos_x: Coord1DColumn
     var pos_y: Coord1DColumn
 
+    #: Track the number of method calls for debugging purposes.
+    var call_counts: Dict[String, Int]
+
     # The bounding box area that contains all points.
     var _bounding_box_area: Float64
 
@@ -64,6 +68,7 @@ struct DataFrame(Defaultable, Movable, Representable):
         self.pos_x = []
         self.pos_y = []
         self._bounding_box_area = 0
+        self.call_counts = {}
 
     fn __init__(
         out self,
@@ -73,6 +78,7 @@ struct DataFrame(Defaultable, Movable, Representable):
         self._bounding_box_area = _compute_bounding_box_area(x, y)
         self.pos_x = x^
         self.pos_y = y^
+        self.call_counts = {}
 
     @staticmethod
     fn _get_self_ptr(
@@ -90,6 +96,15 @@ struct DataFrame(Defaultable, Movable, Representable):
                     e,
                 )
             )
+
+    @staticmethod
+    fn get_call_count(
+        py_self: PythonObject, name: PythonObject
+    ) raises -> PythonObject:
+        """Return the number of times the method was called, just for debugging purposes.
+        """
+        var self_ptr = Self._get_self_ptr(py_self)
+        return self_ptr[].call_counts.get(String(py=name), 0)
 
     @staticmethod
     fn with_columns(
@@ -125,20 +140,33 @@ struct DataFrame(Defaultable, Movable, Representable):
         """For __getitem__ we will return a row of the DataFrame."""
         var self_ptr = Self._get_self_ptr(py_self)
         var index_mojo = Int(py=index)
+        var length = len(self_ptr[].pos_x)
+        if index_mojo < 0 or index_mojo >= length:
+            raise Error("index out of range")
         return Python().tuple(
             self_ptr[].pos_x[index_mojo], self_ptr[].pos_y[index_mojo]
         )
 
     @staticmethod
     fn py__setitem__(
-        py_self: PythonObject, index: PythonObject, value: PythonObject
+        py_self: PythonObject,
+        index: PythonObject,
+        value: Variant[PythonObject, Int],
     ) raises -> None:
         """For __setitem__ we set the x and y values at the given index."""
         var self_ptr = Self._get_self_ptr(py_self)
         var index_mojo = Int(py=index)
-        # Expect value to be a tuple of (x, y)
-        self_ptr[].pos_x[index_mojo] = Float64(py=value[0])
-        self_ptr[].pos_y[index_mojo] = Float64(py=value[1])
+        var length = len(self_ptr[].pos_x)
+        if index_mojo < 0 or index_mojo >= length:
+            raise Error("index out of range")
+        if value.isa[PythonObject]():
+            # Expect value to be a tuple of (x, y) to be saved at index.
+            self_ptr[].pos_x[index_mojo] = Float64(py=value[PythonObject][0])
+            self_ptr[].pos_y[index_mojo] = Float64(py=value[PythonObject][1])
+        else:
+            # Delete the index.
+            _ = self_ptr[].pos_x.pop(index_mojo)
+            _ = self_ptr[].pos_y.pop(index_mojo)
 
     fn __repr__(self) -> String:
         return String("DataFrame( length=", len(self.pos_x), ")")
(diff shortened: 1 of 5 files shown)


winding-lines:
I changed the direction based on the feedback at the December Meetup @ConnorGray. Let me know what are the next changes :)

soraros:
@winding-lines Sorry for the very long delay. I will take a look this Saturday.

soraros:
Can/should we put this feature into an extension? I know that we currently split `CPython` into different file, but we could start with this one. WDYT? @laszlokindrat @ConnorGray

winding-lines:
There are many more typeslots to connect, over 20, so anything that can scale is great.

JoeLoser:
Thanks for this contribution! Adding mapping protocol support is a valuable addition to the Python interop story.

I've left some inline comments on specific areas. The main items to address:

1. Several docstrings incorrectly say "needs to be tp_richcompare" for non-richcompare slots
2. The `_mp_ass_subscript_wrapper` doesn't handle the NULL value case for `__delitem__`
3. The example's `rich_compare` returns `False` for unsupported ops instead of `NotImplemented`

Happy to discuss any of these!

mojo/stdlib/std/python/bindings.mojo
```diff
@@ -902,6 +1142,113 @@ struct PythonTypeBuilder(Copyable):
...
+                length of the container.
+            slot: The Python type slot to define, needs to be tp_richcompare.
+
+        Returns:
+            A reference to self for easier chaining.
+
+        References:
+        - https://docs.python.org/3/c-api/typeobj.html#c.PyMappingMethods.mp_length .
+        """
+
```
JoeLoser, line 1236:
```suggestion
            slot: The Python type slot to define, must be `mp_length`.
```

mojo/stdlib/std/python/bindings.mojo
```diff
@@ -902,6 +1142,113 @@ struct PythonTypeBuilder(Copyable):
...
+                will be None.
+            slot: The Python type slot to define, needs to be tp_richcompare.
+
+        Returns:
+            A reference to self for easier chaining.
+
+        References:
+        - https://docs.python.org/3/c-api/typeobj.html#c.PyMappingMethods.mp_ass_subscript .
+        """
+
```
JoeLoser, line 1267:
```suggestion
            slot: The Python type slot to define, must be `mp_setitem`.
```

mojo/stdlib/std/python/bindings.mojo
```diff
@@ -902,6 +1142,113 @@ struct PythonTypeBuilder(Copyable):
...
+        Returns:
+            A reference to self for easier chaining.
+
+        References:
+        - https://docs.python.org/3/c-api/typeobj.html#c.PyMappingMethods.mp_subscript .
+        """
+
+        self._insert_slot(
+            PyType_Slot.mp_subscript(_mp_subscript_wrapper[method])
```
JoeLoser, line 1296:
```suggestion
            slot: The Python type slot to define, must be `mp_getitem`.
```

mojo/stdlib/std/python/bindings.mojo
```diff
@@ -224,6 +225,245 @@ fn _tp_repr_wrapper[T: Representable](py_self: PyObjectPtr) -> PyObjectPtr:
...
+    When value is NULL, this indicates a deletion (__delitem__). Otherwise, it's an
+    assignment (__setitem__).
+
+    Parameters:
+        method: The user's method that takes self, key, and value.
+
+    Args:
+        py_self: Pointer to the Python object (the container).
+        key: Pointer to the Python object used as the subscript key.
+        value: Pointer to the value to set, or NULL for deletion.
```
JoeLoser, line 437: The docstring mentions that `value` being NULL indicates deletion (`__delitem__`), but this code unconditionally creates a `PythonObject` from the borrowed pointer. If `value` is NULL (when Python calls `del obj[key]`), this could cause issues.

Consider either:
1. Adding a check for `value.is_null()` and handling deletion separately
2. Documenting that `__delitem__` is not supported by this wrapper
3. Providing a separate wrapper or signature for deletion support

mojo/integration-test/python-extension-modules/basic-columnar/mojo_module.mojo
```diff
@@ -0,0 +1,184 @@
...
+        var self_df = Self._get_self_ptr(self_ptr)
+        var other_df = Self._get_self_ptr(other)
+        if op == Py_LT:
+            return self_df[]._bounding_box_area < other_df[]._bounding_box_area
+        elif op == Py_GT:
+            return self_df[]._bounding_box_area > other_df[]._bounding_box_area
+        # For other comparisons, return False
+        return False
+
+
```
JoeLoser, line 198: Returning `False` for unsupported comparison operations is incorrect behavior. Python's comparison protocol expects `NotImplemented` to be returned so the interpreter can try the reflected operation on the other operand.

For example, `a == b` where `a.__eq__(b)` returns `NotImplemented` will then try `b.__eq__(a)`.

Since users will likely use this as a template, consider either:
1. Implementing all comparison ops
2. Returning `NotImplemented` (may need support in the wrapper)
3. Adding a comment explaining this is a limitation of the example

mojo/integration-test/python-extension-modules/basic-columnar/mojo_module.mojo
```diff
@@ -0,0 +1,184 @@
...
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+from os import abort
+
+from memory import UnsafePointer, alloc
+from utils import IndexList
+from python import Python, PythonObject
+from python._cpython import PyObjectPtr, Py_LT, Py_GT
+from python.bindings import (
```
JoeLoser: Nit: `PyObjectPtr` is imported but not used in this file.

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -407,11 +429,15 @@ struct PyType_Spec:
 
 
 # https://github.com/python/cpython/blob/main/Include/typeslots.h
+comptime Py_mp_setitem = 3
```
JoeLoser: These constants are also defined as magic numbers in `PyTypeObjectSlot` in `bindings.mojo`. Consider having `PyTypeObjectSlot` reference these constants to avoid duplication:

```mojo
comptime mp_length = PyTypeObjectSlot(Py_mp_length)
```

soraros: And maybe we should type them

mojo/stdlib/std/python/bindings.mojo
```diff
@@ -902,6 +1142,113 @@ struct PythonTypeBuilder(Copyable):
             _py_function_wrapper[method, is_method=True](), static_method=False
         ](method_name, docstring)
 
+    fn def_method[
+        method: fn (PythonObject, PythonObject, Int) raises -> Bool,
+        slot: PyTypeObjectSlot,
+    ](mut self: Self,) -> ref [self] Self where slot.is_tp_richcompare():
```
soraros: Actually, I don't quite understand this API design. Why do don't we simply call the method `def_richcompare` and remove the type tag? Or if we want to keep the tag, we could make the "payload" a parameter, so the `where` clause is not necessary.

That said, `where` clause with a enum is a really interesting pattern.

mojo/stdlib/std/python/bindings.mojo
winding-lines: done

mojo/stdlib/std/python/bindings.mojo
winding-lines: done

mojo/stdlib/std/python/bindings.mojo
winding-lines: done

mojo/stdlib/std/python/bindings.mojo
winding-lines: I used a Variant to support __delitem__

mojo/integration-test/python-extension-modules/basic-columnar/mojo_module.mojo
winding-lines: Fixed to support NotImplemented and use in the example. According to the docs exceptions are very cheap in Mojo so I went down that route. Happy to implement a different design if you prefer.

mojo/stdlib/std/python/bindings.mojo
winding-lines: Agree, there are 2 design patterns for the API. Do we want one method name with type tags or do we want multiple and more specific method names. 

Like you said, the fact that the where clause worked was pretty cool and it's simpler to  have a small number of functions, so in this PR I went for that approach. Easy to change if the team prefers the second approach.

winding-lines:
@JoeLoser and @soraros  many thanks for the reviews 🙇 

I believe I've addressed the feedback and ready for another pass when you have some time.

mojo/integration-test/python-extension-modules/basic-columnar/mojo_module.mojo
winding-lines: Removed, thanks!

soraros:
@winding-lines I asked Joe offline, and it seems the extensions aren't ready to be used yet—so scratch that.

winding-lines:
Note: I am taking a different approach to this problem in [pontoneer](https://github.com/winding-lines/pontoneer)

winding-lines:
This is replaced by https://github.com/modular/modular/pull/6453

PR #5588, opened 2025-11-19
[max] implement Gemma3ForConditionalGeneration model

RossCampbellDev:
This PR adds the Gemma3ForConditionalGeneration model to the available MAX architectures, based on the technical report for multi modal Gemma3 and pre-existing implementations.
It is prepared for multi-gpu support, and tested against ChartQA

### DIFF
--- a/max/python/max/pipelines/architectures/gemma3multimodal/arch.py
+++ b/max/python/max/pipelines/architectures/gemma3multimodal/arch.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 
+import os
 
 from max.graph.weights import WeightsFormat
 from max.interfaces import PipelineTask
@@ -20,37 +21,70 @@ from max.pipelines.lib import (
     SupportedArchitecture,
     SupportedEncoding,
     TextAndVisionTokenizer,
+    TextTokenizer,
 )
 
-from .model import Gemma3_MultiModalModel
-
-gemma3_multimodal_arch = SupportedArchitecture(
-    name="Gemma3ForConditionalGeneration",
-    example_repo_ids=[
-        # it = Instruction tuned (recommended).
-        # pt = Pre-trained.
-        "google/gemma-3-12b-it",
-        "google/gemma-3-12b-pt",
-        "google/gemma-3-4b-it",
-        "google/gemma-3-4b-pt",
-        "google/gemma-3-12b-it",
-        "google/gemma-3-12b-pt",
-        "google/gemma-3-27b-it",
-        "google/gemma-3-27b-pt",
-    ],
-    default_encoding=SupportedEncoding.bfloat16,
-    supported_encodings={
-        SupportedEncoding.bfloat16: [KVCacheStrategy.PAGED],
-        SupportedEncoding.float8_e4m3fn: [KVCacheStrategy.PAGED],
-    },
-    pipeline_model=Gemma3_MultiModalModel,
-    task=PipelineTask.TEXT_GENERATION,
-    tokenizer=TextAndVisionTokenizer,
-    default_weights_format=WeightsFormat.safetensors,
-    multi_gpu_supported=True,
-    rope_type=RopeType.normal,
-    required_arguments={
-        "enable_prefix_caching": False,
-        "enable_chunked_prefill": False,
-    },
+ENABLE_NEW_IMPL = os.environ.get("MAX_ENABLE_GEMMA3_VISION", "0").lower() in (
+    "1",
+    "true",
 )
+
+if ENABLE_NEW_IMPL:
+    from .model import Gemma3_MultiModalModel
+else:
+    from . import weight_adapters_legacy as weight_adapters
+    from .model_legacy import Gemma3_MultiModalModelLegacy
+
+
+example_repo_ids = [
+    # it = Instruction tuned (recommended).
+    # pt = Pre-trained.
+    "google/gemma-3-12b-it",
+    "google/gemma-3-12b-pt",
+    "google/gemma-3-4b-it",
+    "google/gemma-3-4b-pt",
+    "google/gemma-3-12b-it",
+    "google/gemma-3-12b-pt",
+    "google/gemma-3-27b-it",
+    "google/gemma-3-27b-pt",
+]
+
+if ENABLE_NEW_IMPL:
+    gemma3_multimodal_arch = SupportedArchitecture(
+        name="Gemma3ForConditionalGeneration",
+        example_repo_ids=example_repo_ids,
+        default_encoding=SupportedEncoding.bfloat16,
+        supported_encodings={
+            SupportedEncoding.bfloat16: [KVCacheStrategy.PAGED],
+            SupportedEncoding.float8_e4m3fn: [KVCacheStrategy.PAGED],
+        },
+        pipeline_model=Gemma3_MultiModalModel,
+        task=PipelineTask.TEXT_GENERATION,
+        tokenizer=TextAndVisionTokenizer,
+        default_weights_format=WeightsFormat.safetensors,
+        multi_gpu_supported=True,
+        rope_type=RopeType.normal,
+        required_arguments={
+            "enable_prefix_caching": False,
+            "enable_chunked_prefill": False,
+        },
+    )
+else:
+    gemma3_multimodal_arch = SupportedArchitecture(
+        name="Gemma3ForConditionalGeneration",
+        example_repo_ids=example_repo_ids,
+        default_encoding=SupportedEncoding.bfloat16,
+        supported_encodings={
+            SupportedEncoding.bfloat16: [KVCacheStrategy.PAGED],
+            SupportedEncoding.float8_e4m3fn: [KVCacheStrategy.PAGED],
+        },
+        pipeline_model=Gemma3_MultiModalModelLegacy,
+        task=PipelineTask.TEXT_GENERATION,
+        tokenizer=TextTokenizer,
+        default_weights_format=WeightsFormat.safetensors,
+        multi_gpu_supported=True,
+        rope_type=RopeType.normal,
+        weight_adapters={
+            WeightsFormat.safetensors: weight_adapters.convert_safetensor_state_dict,
+        },
+    )
--- a/max/python/max/pipelines/architectures/gemma3multimodal/arch.py
+++ b/max/python/max/pipelines/architectures/gemma3multimodal/arch.py
@@ -19,10 +19,9 @@ from max.pipelines.lib import (
     RopeType,
     SupportedArchitecture,
     SupportedEncoding,
-    TextTokenizer,
+    TextAndVisionTokenizer,
 )
 
-from . import weight_adapters
 from .model import Gemma3_MultiModalModel
 
 gemma3_multimodal_arch = SupportedArchitecture(
@@ -46,11 +45,12 @@ gemma3_multimodal_arch = SupportedArchitecture(
     },
     pipeline_model=Gemma3_MultiModalModel,
     task=PipelineTask.TEXT_GENERATION,
-    tokenizer=TextTokenizer,
+    tokenizer=TextAndVisionTokenizer,
     default_weights_format=WeightsFormat.safetensors,
     multi_gpu_supported=True,
     rope_type=RopeType.normal,
-    weight_adapters={
-        WeightsFormat.safetensors: weight_adapters.convert_safetensor_state_dict,
+    required_arguments={
+        "enable_prefix_caching": False,
+        "enable_chunked_prefill": False,
     },
 )
--- a/max/python/max/pipelines/architectures/gemma3multimodal/model.py
+++ b/max/python/max/pipelines/architectures/gemma3multimodal/model.py
@@ -270,6 +270,8 @@ class Gemma3_MultiModalModel(PipelineModel[TextAndVisionContext], KVCacheMixin):
 
         self.vision_model, self.language_model = self.load_model(session)
 
+        self.vision_stacker = _VisionStacker()
+
     @classmethod
     def calculate_max_seq_len(
         cls, pipeline_config: PipelineConfig, huggingface_config: AutoConfig
@@ -716,7 +718,7 @@ class Gemma3_MultiModalModel(PipelineModel[TextAndVisionContext], KVCacheMixin):
         if not images:
             return None
 
-        final_images = _VisionStacker().stack(images)
+        final_images = self.vision_stacker.stack(images)
 
         tensor = cast_dlpack_to(
             final_images, DType.float32, DType.bfloat16, self.devices[0]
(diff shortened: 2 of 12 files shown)


ross:
> ❌ @ross Campbell

Please fix whatever is spamming @ mentions here.

RossCampbellDev:

tboerstad:
!sync

tboerstad:
!gpu

max/python/max/pipelines/architectures/gemma3multimodal/arch.py
```diff
@@ -19,10 +19,10 @@
     RopeType,
     SupportedArchitecture,
     SupportedEncoding,
-    TextTokenizer,
+    TextAndVisionTokenizer,
 )
+from weight_adapters import convert_safetensor_state_dict
```
k-w-w: I think this needs to be `from .weight_adapters`

max/python/max/pipelines/architectures/gemma3multimodal/vision_model/projection.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+
+        self.patches_per_image = int(
+            config.vision_config.image_size // config.vision_config.patch_size
+        )
+
+        self.tokens_per_side = int(config.mm_tokens_per_image**0.5)
+        self.kernel_size = self.patches_per_image // self.tokens_per_side
+
+    def __call__(self, vision_outputs: Tensor) -> TensorValue:
```
k-w-w: Both inputs and outputs should be TensorValue (they are all symbolic)

max/python/max/pipelines/architectures/gemma3multimodal/vision_model/projection.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+
+        pooled_vision_outputs = ops.permute(pooled_vision_outputs, [0, 3, 1, 2])
+        pooled_vision_outputs = pooled_vision_outputs.flatten(2)
+        pooled_vision_outputs = pooled_vision_outputs.transpose(1, 2)
+
+        normed_vision_outputs = self.mm_soft_emb_norm(pooled_vision_outputs)
+
+        projected_vision_outputs = matmul(
+            normed_vision_outputs, self.mm_input_projection_weight
```
k-w-w: `normed_vision_outputs @ self.mm_input_projection_weight`

max/python/max/pipelines/architectures/gemma3multimodal/vision_model/projection.py
```diff
@@ -0,0 +1,223 @@
...
+            device=self.device,
+            has_bias=True,
+        )
+
+    def __call__(self, x: TensorValue):
+        """Expands hidden states to intermediate size, applies GELU activation,
+        then projects back to hidden size."""
+        x = self.fc1(x)
+        x = ops.gelu(
+            x, getattr(self.config.vision_config, "hidden_act", "none")
```
k-w-w: Do we need to use getattr here? the vision config used here is well typed (unlike HuggingFace's autoconfig object)

max/python/max/pipelines/architectures/gemma3multimodal/model_config.py
```diff
@@ -0,0 +1,306 @@
...
+
+    patch_size: int
+    """The size (resolution) of each patch.  default 14"""
+
+    attention_bias: bool = True
+
+    attention_dropout: float = 0.0
+    """The dropout ratio for the attention probabilities"""
+
+    model_type: str = "Gemma3"
```
k-w-w: I think this can be removed

max/python/max/pipelines/architectures/gemma3multimodal/model_config.py
```diff
@@ -0,0 +1,306 @@
...
+    attn_logit_softcapping: float = None
+    """Scaling factor when applying tanh softcapping on the attention scores."""
+
+    float8_config: Float8Config | None = None
+    """Float8 quantization configuration."""
+
+    head_dim: int = 256
+    """The attention head dimension."""
+
+    model_type: str = "Gemma3ForConditionalGeneration"
```
k-w-w: This too can be removed

max/python/max/pipelines/architectures/gemma3multimodal/model.py
```diff
@@ -98,66 +284,46 @@ def __init__(
...
-        return super().get_num_layers(huggingface_config.text_config)
+    @staticmethod
+    def get_num_layers(huggingface_config: AutoConfig) -> int:
+        """Gets the number of hidden layers from the HuggingFace configuration."""
+        return Gemma3ForConditionalGenerationConfig.get_num_layers(
+            huggingface_config
+        )
 
-    @classmethod
+    @staticmethod
```
k-w-w: Making it a staticmethod makes sense, but for now we're stuck with classmethod because of the parent class (I think mypy would complain here)

k-w-w (commented):
Some drive by comments, great work!

tboerstad:
!sync
!gpu

max/python/max/pipelines/architectures/gemma3multimodal/vision_model/projection.py
RossCampbellDev: for some reason i can't resolve this conversation - but the change has been made anyway

max/python/max/pipelines/architectures/gemma3multimodal/vision_model/projection.py
RossCampbellDev: again, can't resolve but the change was made

tboerstad:
!sync

tboerstad:
!gpu

max/python/max/pipelines/architectures/gemma3multimodal/model.py
```diff
@@ -98,14 +284,22 @@ def __init__(
             return_logits,
         )
 
+        # # Initialize emtpy signal buffers
```
tboerstad:
```suggestion
        # Initialize empty signal buffers
```

max/python/max/pipelines/architectures/gemma3multimodal/model.py
```diff
@@ -14,36 +14,230 @@
...
+        self.image_token_indices = image_token_indices
+
+    @property
+    def has_vision_inputs(self) -> bool:
+        """Check if this input contains vision data."""
+        return self.pixel_values is not None
 
 
-class Gemma3_MultiModalModel(Gemma3Model):
+# ⚠️ some parts are finished, some are possibly not (prepare vision inputs for example)
```
tboerstad:
```suggestion
```

max/python/max/pipelines/architectures/gemma3multimodal/model.py
```diff
@@ -171,22 +338,514 @@ def estimate_kv_cache_size(
...
+            for img in context.next_images:
+                images.append(img.pixel_values)
+
+        if not images:
+            return None
+
+        final_images = _VisionStacker().stack(images)
+
+        tensor = Tensor.from_numpy(final_images)
+        tensor = _cast_to_dtype(
```
tboerstad: This is slow, since it creates an inference session under the hood. 
Can we do similar to InternVL? 

So in the tokenizer do:
`:rom max.pipelines.lib import float32_to_bfloat16_as_uint16`
 Store pixel_values as uint16 (bfloat16 bit representation)

 In model.py, reinterpret bytes:
  ```
tensor = Tensor.from_numpy(final_images)
tensor = tensor.view(DType.bfloat16, tensor.shape)
```

If that doesn't pan out, there's:
`from max.graph.tensor_utils import cast_dlpack_to` which does the same, but caches the inference session

Comment (tboerstad) [reply max/python/max/pipelines/architectures/gemma3multimodal/model.py]: Sorry, on second thought let's just reuse `cast_dlpack_to`. 
This way there's minimal change to the code, `_cast_to_dtype` can be removed, and no risk of issues.

File: max/python/max/pipelines/architectures/gemma3multimodal/weight_adapters.py [hunk 11]
```diff
@@ -14,39 +14,63 @@
...
+    "vision_tower.vision_model.": "",
+    "multi_modal_": "",
+}
 
 
 def convert_safetensor_state_dict(
     state_dict: dict[str, Weights],
-    huggingface_config: AutoConfig,
-    **unused_kwargs,
+) -> tuple[dict[str, WeightData], dict[str, WeightData]]:
```
tboerstad: Can we change the return value to `-> dict[str, WeightData]`?
To be consistent with the other implementations

And rather use a prefix for the vision weights, ala:
```
GEMMA3_MULTIMODAL_SAFETENSOR_MAP: dict[str, str] = {
+    "vision_tower.vision_model.": "vision_tower.",
+    "multi_modal_": "vision_tower.", 
 }
```

tboerstad:
Some comments from me, otherwise looks good 🙏

max/python/max/pipelines/architectures/gemma3multimodal/weight_adapters.py
RossCampbellDev: might be a bit redundant - originally i had mimicked InternVL and had called the two conversion functions directly in `model.py#load_model()` to get my language+vision weights, which is fairly clean.

i then added the combined function in weight adapters that returns the tuple of dicts, which moves the function calls into weight adapters and allowed us to use `self.adapter()` in `model.py`

if i add a `vision_tower.` prefix to the vision weights, i need to then remove it again later otherwise I run into problems with missing weights and unexpected weights due to the names no longer matching.

unless there is some way to tell the various layers to expect a prefix string on their weights, it seems like the cleanest route is either calling the 2 conversion functions in `load_model()`, or doing what I currently have in this PR?

most others seem to use the former approach and ignore the `SupportedArchitecture`'s `adapter` entirely

k-w-w: The weight adapter is only used internally by the `PipelineModel` (and we should honestly remove it from `SupportedArchitecture`), so the return type can be as flexible as needed. There might be issues with mypy with the current implementation, since the adapter must be: `WeightsAdapter = Callable[..., dict[str, WeightData]]`

I would recommend changing the return a single `dict`, and then splitting the weights when loading into the separate vision and language models

max/python/max/pipelines/architectures/gemma3multimodal/model_config.py
```diff
@@ -0,0 +1,300 @@
...
+@dataclass
+class Gemma3VisionConfig:
+    """
+    The vision-specific config for Gemma3
+    More info at: https://huggingface.co/google/gemma-3-4b-it/blob/main/config.json
+    """
+
+    hidden_act: str | None
+    """The non-linear activation function (function or string) in the encoder and pooler. If string, `"gelu"`,
+    `"relu"`, `"selu"` and `"gelu_new"` `"gelu"` are supported."""
```
k-w-w: "silu", also here's the list of supported activations: https://github.com/modular/modular/blob/d2dadf1a723d7e341f84d34e93223517b23bfca3/max/python/max/nn/linear.py#L1005-L1012

max/python/max/pipelines/architectures/gemma3multimodal/model_config.py
```diff
@@ -0,0 +1,300 @@
...
+
+    hidden_size: int
+    """Dimensionality of the encoder layers and the pooler layer.  default 1152"""
```
k-w-w: Clean up the default values listed in the comments (since the defaults aren't actually set)

max/python/max/pipelines/architectures/gemma3multimodal/model_config.py
```diff
@@ -0,0 +1,300 @@
...
+
+    image_size: int
+    """The size (resolution) of each image.  default 896"""
+
+    intermediate_size: int
+    """maybe not required? default 4304"""
```
k-w-w: Looks like this was used, can you update this comment?

max/python/max/pipelines/architectures/gemma3multimodal/model_config.py
```diff
@@ -0,0 +1,300 @@
...
+    text_config: Gemma3Config
+    """The config object of the text backbone"""
+
+    vision_config: Gemma3VisionConfig
+    """Custom vision config or dict"""
+
+    attention_bias: bool = False
+    """Whether to use a bias in the query, key, value and output projection layers during self-attention."""
+
+    attn_logit_softcapping: float = None
```
k-w-w: It looks like this isn't used (and I don't think we support softcapping), perhaps this should be removed

k-w-w (commented):
A few docstring comments, but the rest LGTM! Will let @tboerstad do the final approval

tboerstad:
!sync

tboerstad:
!sync

tboerstad:
!gpu

max/python/max/pipelines/architectures/gemma3multimodal/arch.py
```diff
@@ -46,11 +45,8 @@
     },
     pipeline_model=Gemma3_MultiModalModel,
     task=PipelineTask.TEXT_GENERATION,
-    tokenizer=TextTokenizer,
+    tokenizer=TextAndVisionTokenizer,
     default_weights_format=WeightsFormat.safetensors,
     multi_gpu_supported=True,
     rope_type=RopeType.normal,
```
tboerstad:
```suggestion
    rope_type=RopeType.normal,
    required_arguments={
        "enable_prefix_caching": False,
        "enable_chunked_prefill": False,
    },
```

We do not support prefix_caching and chunked_prefill for VLMs at the moment. 
This fixes the issue with batch_size>1 failing

max/python/max/pipelines/architectures/gemma3multimodal/model.py
```diff
@@ -171,22 +314,517 @@ def estimate_kv_cache_size(
...
+        and available memory.
+
+        Args:
+            session: The MAX Engine inference session.
+            available_cache_memory: The amount of memory available for the KV cache in bytes.
+
+        Returns:
+            An initialized :obj:`KVCacheManager` instance.
+        """
+        return load_kv_manager(
```
tboerstad, line 782:
```suggestion
        if available_cache_memory is not None:
            available_cache_memory = int(available_cache_memory * 0.9)

        return load_kv_manager(
```
Could you add this workaround? Otherwise our memory estimation fails, and we OOM

tboerstad:
!gpu
!sync

tboerstad:
!sync

tboerstad:
!sync
!gpu

max/python/max/pipelines/architectures/gemma3multimodal/model.py
```diff
@@ -171,22 +315,525 @@ def estimate_kv_cache_size(
...
+        for context in context_batch:
+            for img in context.next_images:
+                images.append(img.pixel_values)
+
+        if not images:
+            return None
+
+        final_images = _VisionStacker().stack(images)
+
+        tensor = Tensor.from_numpy(final_images)
```
tboerstad:
```suggestion
```
This can be removed iiuc

tboerstad:
!sync

tboerstad:
!gpu

PR #5589, opened 2025-11-19
Test CI

keith:


keith:
https://github.com/bazelbuild/bazel/issues/27730

PR #5599, opened 2025-11-20
add ability to set compile time variables in python->mojo interop layer

alisterburt:
This PR adds a mechanism for setting compile time variables in Mojo from runtime values in Python to the Mojo-Python interop layer

Key changes:
- added `mojo.importer.set_comptime_variables(module_name: str, variables: dict[str, str])
- added hash of comptime variables to cached dynamic library filename
- added example usage

Motivation:
- enabling compile time optimizations in mojo based on information only available at runtime in Python

This API is intended as a starting point for discussion, open to making any required changes

I put this together with Jeff Martin

### DIFF
--- a/mojo/python/mojo/importer.py
+++ b/mojo/python/mojo/importer.py
@@ -16,9 +16,12 @@ import logging
 import os
 import subprocess
 import sys
+from collections import defaultdict
 from collections.abc import Sequence
 from importlib.util import spec_from_file_location
 from pathlib import Path
+from itertools import chain
+import threading
 
 from .paths import MojoCompilationError, MojoModulePath, find_mojo_module_in_dir
 from .run import subprocess_run_mojo
@@ -26,7 +29,7 @@ from .run import subprocess_run_mojo
 # ---------------------------------------
 # Helper Functions
 # ---------------------------------------
-
+_thread_local_storage = threading.local()
 
 def _calculate_mojo_source_hash(mojo_dir: Path) -> str:
     """Calculates a truncated SHA256 hash of all .mojo/.🔥 files in a directory."""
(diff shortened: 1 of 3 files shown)


alisterburt:

mojo/python/mojo/importer.py
```diff
@@ -59,14 +62,33 @@ def _calculate_mojo_source_hash(mojo_dir: Path) -> str:
...
+    if not comptime_variables:
+        return ""
+
+    hasher = hashlib.sha256()
+    # Sort items to ensure consistent hashing regardless of dict ordering
+    for key, value in sorted(comptime_variables.items()):
+        hasher.update(key.encode("utf-8"))
+        hasher.update(b"=")
+        hasher.update(value.encode("utf-8"))
+        hasher.update(b"\0")  # null separator between entries
```
cuchaz, line 76: Hashing correctly can get pretty tricky. I believe UTF-8 encoding allows zero bytes? So using a zero byte as a delimiter here could potentially lead to unwanted hash collisions. A better choice would be to use something like [pre-authentication encoding (PAE)](https://github.com/MatthiasValvekens/python-pae), which avoids the problem by adding explicit shape information to the encoding, rather than relying on delimiters.

cuchaz:
> I put this together with Jeff Martin

Nice, that's me!

It's good to see this make it into a PR. Hopefully someone will appreciate it. =)

mojo/python/mojo/importer.py
cuchaz: And I suppose the `=` delimiter could be problematic for the same reason. Again, PAE solves the issue.

alisterburt:
@cuchaz couldn't find your GitHub handle! I hope to see your Mojo version of this functionality make it into a PR too. 

Thanks for the feedback - will take a look into what you've mentioned when back at a computer :-)

PR #5609, opened 2025-11-26
[stdlib] Fix: Switch `TestReport.name` from `StaticString` to `String`

josiahls:
This is to support runtime test discovery.

Example runtime test discovery:
```mojo
from subprocess import run
from testing.suite import TestSuiteReport, TestReport, TestResult
from builtin._location import __call_location
from pathlib import Path

comptime TEST_DIR = Path("tests")

def test_file(file: Path) -> TestReport:
    var result = run("pixi run test " + String(file))
    if "Unhandled exception caught during execution" in result:
        return TestReport.failed(
            name=file.name(), duration_ns=0, error=Error(result)
        )
    return TestReport.passed(name=file.name(), duration_ns=0)

def walk_tests(path: Path, mut test_results: List[TestReport]):
    for f in path.listdir():
        file = path / f
        if file.is_file() and file.suffix() == ".mojo":
            var result = test_file(file)
            test_results.append(result^)
        elif file.is_dir():
            walk_tests(file, test_results)


def main():
    var test_results = List[TestReport]()
    walk_tests(TEST_DIR, test_results)
    var report = TestSuiteReport(
        reports=test_results^, location=__call_location[inline_count=0]()
    )
    print(report)
```
Compile time inter-file test discover also still works:
```mojo
from testing import assert_equal, TestSuite

def test_lightbug_http_backend():
    assert_equal(2, 1)

def main():
    TestSuite.discover_tests[__functions_in_module()]().run()
```
Example output:
Screenshot 2025-11-26 at 3 43 29 PM

### DIFF
--- a/mojo/stdlib/stdlib/testing/suite.mojo
+++ b/mojo/stdlib/stdlib/testing/suite.mojo
@@ -121,7 +121,7 @@ struct TestReport(Copyable, Movable, Writable):
 
     comptime _ErrorIndent = 3
 
-    var name: StaticString
+    var name: String
     """The name of the test."""
 
     var duration_ns: UInt
@@ -134,7 +134,7 @@ struct TestReport(Copyable, Movable, Writable):
     """The error associated with a failing test."""
 
     @staticmethod
-    fn passed(*, name: StaticString, duration_ns: UInt) -> Self:
+    fn passed(*, var name: String, duration_ns: UInt) -> Self:
         """Create a passing test report.
 
         Args:
@@ -145,15 +145,13 @@ struct TestReport(Copyable, Movable, Writable):
             A new passing test report.
         """
         return {
-            name = name,
+            name = name^,
             duration_ns = duration_ns,
             result = TestResult.PASS,
         }
 
     @staticmethod
-    fn failed(
-        *, name: StaticString, duration_ns: UInt, var error: Error
-    ) -> Self:
+    fn failed(*, var name: String, duration_ns: UInt, var error: Error) -> Self:
         """Create a failing test report.
 
         Args:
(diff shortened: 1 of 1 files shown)


mojo/stdlib/stdlib/testing/suite.mojo
```diff
@@ -151,9 +151,7 @@ struct TestReport(Copyable, Movable, Writable):
         }
```
NathanSWard: This function (also `failed` and `skipped`) should take a `var String` and then move the string to avoid unnecessary copies.

```mojo
return {
    name = name^,
    duration_ns = duration_ns,
    result = TestResult.PASS,
}
```

mojo/stdlib/stdlib/testing/suite.mojo
```diff
@@ -187,7 +185,7 @@ struct TestReport(Copyable, Movable, Writable):
     fn __init__(
         out self,
         *,
-        name: StaticString,
+        name: String,
```
NathanSWard: Ditto `var String` and then later moving the string 
```mojo
self.name = name^
```

josiahls: done

mojo/stdlib/stdlib/testing/suite.mojo
josiahls: done

mojo/stdlib/stdlib/testing/suite.mojo
```diff
@@ -397,7 +395,7 @@ struct TestSuite(Movable):
         """
         self.tests = List[_Test]()
         self.location = location.or_else(__call_location())
-        self.skip_list = Set[StaticString]()
+        self.skip_list = Set[String]()
```
soraros:
```suggestion
        self.skip_list = {}
```

josiahls: done

soraros (approved):

NathanSWard (approved):

NathanSWard:
!sync

PR #5622, opened 2025-11-29
[stdlib] Fix ```Dict.__str__``` to support custom hashers

sdgunaa:
#### Fix `Dict.__str__` for custom hashers

`Dict.__str__` was defined to take `Dict[T, U]`, which only matches dictionaries using the `default_hasher`. When a `custom_hasher` is used, the method becomes incompatible and fails to resolve.

Previously, `Dict.__str__` was defined as:

```mojo
@no_inline
fn __str__[
    T: KeyElement & Representable,
    U: Copyable & Representable,
](self: Dict[T, U]) -> String:
```

### Reproduction

```mojo
from collections import Dict
from hashlib import default_comp_time_hasher

def main():
  var d: Dict[Int, Int, default_comp_time_hasher] = {}
  d[1] = 10
  d.__str__()  # error: cannot convert Dict[Int, Int, default_comp_time_hasher] to Dict[Int, Int]
```

### Fix
Update `Dict.__str__` to use the hasher of the concrete `Dict` instance:
```mojo
@no_inline
fn __str__[
    T: KeyElement & Representable,
    U: Copyable & Representable,
](self: Dict[T, U, Self.H]) -> String:
```

This keeps the existing constraints on `T` and `U` and makes `__str__`
compatible with all `Dict` instantiations, regardless of the hasher type.


sdgunaa:

sdgunaa:
recheck

JoeLoser (approved):
Thanks for the fix, looks good to me! Do you mind rebasing to resolve the merge conflict so I can sync this change in and land it?

sdgunaa:
> Thanks for the fix, looks good to me! Do you mind rebasing to resolve the merge conflict so I can sync this change in and land it?

Heyy, Joe! While resolving the conflict I noticed that main has a newer Dict.__str__ + write_to implementation that already supports custom hashers.
Because of that, my original fix is no longer needed... still added the test case anyway...

JoeLoser:
> > Thanks for the fix, looks good to me! Do you mind rebasing to resolve the merge conflict so I can sync this change in and land it?
> 
> Heyy, Joe! While resolving the conflict I noticed that main has a newer Dict.**str** + write_to implementation that already supports custom hashers. Because of that, my original fix is no longer needed... still added the test case anyway...

No worries, that's good to hear. Let's sync this in for the additional test you added. Thanks!

JoeLoser:
!sync

PR #5619, opened 2025-11-28
[stdlib] Optimize `StringSlice.char_length` using SIMD `pack_bits` and `pop_count` (12x speedup)

YichengDWu:
This PR significantly optimizes `StringSlice.char_length` by maximizing SIMD register throughput.

The Optimization: The previous (or naive) SIMD approaches often require casting `int8` predicates to wider integer types (e.g., `int32`) to perform reductions, which reduces the effective SIMD width by 4x or 8x.

This implementation uses `pack_bits` to compress the comparison results (continuation byte detection) into a single integer mask, followed by `pop_count`. This approach:

1. Maintains 100% data density: Processes 64 bytes per instruction on AVX-512 (vs 16 bytes if casted to 32-bit ints).

2. Utilizes hardware acceleration: Maps efficiently to `vpmovmskb/popcnt` on x86 and equivalent instructions on ARM.

Benchmarks (1MB text): The new implementation removes the performance penalty for multi-byte characters seen in the current implementation.

Language | Before (ms) | After (ms) | Speedup
-- | -- | -- | --
EN (ASCII) | ~0.019 | ~0.020 | 1.0x
ZH (Chinese) | 0.241 | 0.020 | 12.0x 🚀
AR (Arabic) | 0.245 | 0.017 | 14.4x 🚀

Device: Mac mini M4

PR #5623, opened 2025-11-30
[stdlib] Optimize `vectorize` tail handling (up to 10x speedup on small/unaligned sizes)

YichengDWu:
This PR updates the `vectorize` function to handle tail elements (the remainder of `size % simd_width`) much more efficiently using the bitmask dispatching strategy.

### The Context

Currently, if you process a buffer that isn't perfectly aligned to the SIMD width, the `vectorize` function falls back to a scalar loop for the leftovers:
```mojo
for i in range(simd_end, size):
    closure[1](i)
```

This causes a massive performance cliff. For example, processing 31 items is currently ~4.5x slower than processing 32 items, simply because those 31 items are processed one by one.

### The Changes
This PR replaces the scalar loop with a Cascading Bitmask Dispatching strategy. Instead of looping 31 times for a remainder of 31, we now just perform a quick bitwise check and dispatch:

1. Check bit 4 -> handle 16 items.

2. Check bit 3 -> handle 8 items.

3. Check bit 2 -> handle 4 items.

4. ...and so on.

This results in straight-line, branch-predictor-friendly code.

For the `comptime size` overload, it's even better: the compiler performs dead code elimination on these checks, resulting in a fully static, branchless sequence of calls perfectly matched to the data size.

### Benchmarks
Ran this on Intel i5-13600K (AVX2). The difference on small, unaligned sizes is huge.

The most important comparison is vectorize_127 (worst case tail) vs vectorize_128 (best case aligned).

* Before: 127 items took 17.1µs (vs 0.73µs for 128). That's a ~23x penalty for missing alignment.

| name                   | met (ms)               | iters |
| ---------------------- | ---------------------- | ----- |
| bench_vectorize_31     | 5.01e-06               | 100   |
| bench_vectorize_32     | 1.4e-06                | 100   |
| bench_vectorize_127    | 1.7080000000000002e-05 | 100   |
| bench_vectorize_128    | 7.3e-07                | 100   |
| bench_vectorize_1023   | 9.56e-06               | 100   |
| bench_vectorize_1024   | 3.79e-06               | 100   |
| bench_vectorize_8191   | 3.308e-05              | 100   |
| bench_vectorize_8192   | 2.7499999999999998e-05 | 100   |
| bench_vectorize_32767  | 0.00010906             | 100   |
| bench_vectorize_32768  | 0.00010564             | 100   |
| bench_vectorize_131071 | 0.0016952500000000001  | 100   |
| bench_vectorize_131072 | 0.0016428400000000002  | 100   |

* After: 127 items take 1.6µs. The penalty is largely gone.

| name                   | met (ms)               | iters |
| ---------------------- | ---------------------- | ----- |
| bench_vectorize_31     | 1.41e-06               | 100   |
| bench_vectorize_32     | 4.6999999999999995e-07 | 100   |
| bench_vectorize_127    | 1.6000000000000001e-06 | 100   |
| bench_vectorize_128    | 6.3e-07                | 100   |
| bench_vectorize_1023   | 4.54e-06               | 100   |
| bench_vectorize_1024   | 3.7399999999999998e-06 | 100   |
| bench_vectorize_8191   | 2.845e-05              | 100   |
| bench_vectorize_8192   | 2.819e-05              | 100   |
| bench_vectorize_32767  | 0.00010848             | 100   |
| bench_vectorize_32768  | 0.00010801             | 100   |
| bench_vectorize_131071 | 0.00177422             | 100   |
| bench_vectorize_131072 | 0.0016857600000000001  | 100   |

### Relation to Forum Discussion

I analyzed the [Vectorize and remainder loops](https://forum.modular.com/t/vectorize-and-remainder-loops/665) discussion, where two main optimization paths were debated:

1.  Greedy Loop (Iterative):
    A nested loop approach (`for ... range((size - offset) // w)`). As noted by `@owenhilyard` in the thread, this introduces a deep data dependency on the `offset` variable, forcing serialized execution.

2.  Bitmask Dispatching (The approach in this PR):
    `@gphlipot` suggested checking "whether the $i$-th bit is set" to avoid deep dependencies.
    This PR implements exactly this strategy. By checking `(tail_len & width)`, we ensure that:

      * No Dependency Chains: All branch conditions depend only on the initial, immutable `tail_len`.
      * Instruction Level Parallelism: The CPU can resolve these checks independently.
      * Low Overhead: The benchmarks confirm that the cost of these straight-line branches is negligible compared to the latency of the naive scalar loop.

### Assembly Analysis
I've also prepared a [Godbolt link](https://godbolt.org/z/88E43oqoP) to verify the generated assembly.

The naive implementation (`_memset_impl`) falls back to a `memset` call for the tail.

```asm
.LBB0_3:
    ; ... setup overhead ...
    jmp     memset@PLT          ; Fallback to library call
```
The optimized implementation (`_memset_impl2`) generates fully inlined, straight-line AVX stores without loops or function calls.

```asm
.LBB1_3:
    testb   $16, %dl            ; Check bit 4
    jne     .LBB1_4
    testb   $8, %dl             ; Check bit 3
    jne     .LBB1_6
    ; ...
.LBB1_4:
    vmovdqu %xmm0, (%rdi,%rax)  ; Inline 16-byte store
    testb   $8, %dl             ; Straight-line continuation
    ; ...
```


YichengDWu:
Blocked by #5619

YichengDWu:
Although #5619 is resolved, it appears that many functions in the current codebase still implicitly assume that the vectorize tail handling falls back to a scalar loop. I personally believe we should completely abandon this programming model, as it is clearly a performance trap.

 If the consensus is to move away from the scalar fallback assumption, I will proceed with further updates based on that direction.

soraros:
We could also add a flag parameter, and do the migration later, for feels like a very high risk change. WDYT, @JoeLoser?

YichengDWu:
Thinking about this further, I realized that the bitmask dispatching strategy in this PR is still not the theoretical optimum for modern ISAs like AVX-512 (or RVV/SVE).

On AVX-512, handling the tail often only requires a single masked instruction (or `llvm.vp` equivalent), whereas the bitmask strategy still generates a cascade of checks and smaller vector ops.

I propose adding a new overload for `vectorize` that allows the closure to accept an additional `evl` (effective vector length) argument. This would allow users to handle the tail logic themselves (e.g., via masked load/stores) within the closure.

If this sounds good, I can add this overload in a new PR.

barcharcraz:
> Thinking about this further, I realized that the bitmask dispatching strategy in this PR is still not the theoretical optimum for modern ISAs like AVX-512 (or RVV/SVE).
> 
> On AVX-512, handling the tail often only requires a single masked instruction (or `llvm.vp` equivalent), whereas the bitmask strategy still generates a cascade of checks and smaller vector ops.
> 
> I propose adding a new overload for `vectorize` that allows the closure to accept an additional evl (effective vector length) argument. This would allow users to handle the tail logic themselves (e.g., via masked load/stores) within the closure.
> 
> If this sounds good, I can add this overload in a new PR.

I like giving the users the ability to handle the tail themselves, but fixing our handling still seems appropriate. I need to better understand why it's such a disruptive change, but if we really can't make such an improvement then something is wrong with the vectorize API generally.

YichengDWu:
> > Thinking about this further, I realized that the bitmask dispatching strategy in this PR is still not the theoretical optimum for modern ISAs like AVX-512 (or RVV/SVE).
> > On AVX-512, handling the tail often only requires a single masked instruction (or `llvm.vp` equivalent), whereas the bitmask strategy still generates a cascade of checks and smaller vector ops.
> > I propose adding a new overload for `vectorize` that allows the closure to accept an additional evl (effective vector length) argument. This would allow users to handle the tail logic themselves (e.g., via masked load/stores) within the closure.
> > If this sounds good, I can add this overload in a new PR.
> 
> I like giving the users the ability to handle the tail themselves, but fixing our handling still seems appropriate. I need to better understand why it's such a disruptive change, but if we really can't make such an improvement then something is wrong with the vectorize API generally.

The issue is that the current codebase relies on the implicit assumption that vectorize falls back to a simple scalar loop for the tail. A example is [here](https://github.com/modular/modular/blob/main/max/kernels/src/internal_utils/_measure.mojo#L174), which also demonstrates that users are already compelled to implement their own tail handling logic.

Note that on AVX-512 hardware, the bitmask dispatching strategy proposed in this PR is almost never what we want, considering native masked operations are superior. The existing vectorize interface strictly limits us to support this optimization because it doesn't expose a `mask` or `evl` to the closure.

I lean towards adding an overload that allows users to handle the tail manually (via `evl` or masks). I hope to see a gradual migration toward that pattern, as the example above proves users are already trying to do exactly that.

YichengDWu:
Followed by #5687

YichengDWu:
Followed in #5854

PR #5624, opened 2025-11-30
[stdlib] Use `memcmp` to dispatch to optimized implementation in `StringSlice`

YichengDWu:
This PR updates the comparison logic in `StringSlice` to use the public `memcmp` API instead of the internal `_memcmp_impl_unconstrained`.

Previously, `StringSlice` was directly calling `_memcmp_impl_unconstrained`. By switching to the standard `memcmp`, we allow the call to be properly dispatched to the optimized implementation (`_memcmp_impl_opt_unconstrained`).

This ensures that string slice comparisons leverage the best available memory comparison implementation provided by the standard library.

### DIFF
--- a/mojo/stdlib/stdlib/collections/string/string_slice.mojo
+++ b/mojo/stdlib/stdlib/collections/string/string_slice.mojo
@@ -78,7 +78,6 @@ from memory import (
     memcpy,
     pack_bits,
 )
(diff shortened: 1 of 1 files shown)


JoeLoser (approved):

JoeLoser:
!sync

PR #5636, opened 2025-12-05
[stdlib] Add `Result` type (take 2)

### LINKED ISSUE #2748: [Feature Request] [stdlib] [proposal] Add `ErrorReg`, `Result`, and `ResultReg` types

martinvuyk:
Closes #2748.

### DIFF
--- a/mojo/stdlib/stdlib/collections/result.mojo
+++ b/mojo/stdlib/stdlib/collections/result.mojo
@@ -1,0 +1,713 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Defines Result, a type modeling a value which may or may not be present.
+With an Error in the case of failure.
+Result values can be thought of as a type-safe nullable pattern.
+Your value can take on a value or `None`, and you need to check
+and explicitly extract the value to get it out.
+
+Examples:
+```mojo
+from collections.result import Result
+var a = Result(1)
+var b = Result[Int]()
+if a:
+    print(a.value())  # prints 1
+if b:  # bool(b) is False, so no print
+    print(b.value())
+var c = a.or_else(2)
+var d = b.or_else(2)
+print(c)  # prints 1
+print(d)  # prints 2
+```
+And if more information about the returned Error is wanted it is available.
+```mojo
+from collections.result import Result
+var a = Result(1)
+var b = Result[Int](err=Error("something went wrong"))
+var c = Result[Int](None, Error("error 1"))
+var d = Result[Int](err=Error("error 2"))
+if a:
+    print(a.err)  # prints ""
+if not b:
+    print(b.err) # prints "something went wrong"
+if c.err:
+    print("c had an error")
+# TODO: pattern matching
+if String(d.err_value()) == "error 1":
+    print("d had error 1")
+elif String(d.err_value()) == "error 2":
+    print("d had error 2")
+```
+A Result with an Error can also be retuned early:
+```mojo
+fn func_that_can_err[A: Copyable]() -> Result[A]:
+    ...
+
+fn return_early_if_err[T: Copyable, A: Copyable]() -> Result[T]:
+    var result: Result[A] = func_that_can_err[A]()
+    if not result:
+        # the internal err gets transferred to a Result[T]
+        return result
+        # its also possible to do:
+        # return None, Error("func_that_can_err failed")
+    var val = result.value()
+    var final_result: T
+    ...
+    return final_result
+```
+A Result can be unwrapped within a raising function:
+```mojo
+fn func_that_can_err[A: Copyable]() -> Result[A]:
+    ...
+
+fn return_early_if_err[T: Copyable, A: Copyable]() raises  -> T:
+    var val = func_that_can_err[A]()[]
+    var final_result: T
+    ...
+    return final_result
+```
+"""
+
+
+from os import abort
+
+from utils import Variant
+
+from builtin.constrained import _constrained_conforms_to
+from builtin.device_passable import DevicePassable
+from compile import get_type_name
+from memory import LegacyOpaquePointer as OpaquePointer
+
+from sys.intrinsics import _type_is_eq, _type_is_eq_parse_time
+
+# ===-----------------------------------------------------------------------===#
+# Result
+# ===-----------------------------------------------------------------------===#
+
+
+struct Result[
+    T: Copyable & Movable,
+    E: Copyable & Movable & Defaultable = Error,
+](
+    Boolable,
+    Defaultable,
+    ImplicitlyCopyable,
+    Iterable,
+    Iterator,
+    Movable,
+    Representable,
+    Stringable,
+    Writable,
+):
+    """A type modeling a value which may or may not be present.
+    With an Error in the case of failure.
+    Result values can be thought of as a type-safe nullable pattern.
+    Your value can take on a value or `None`, and you need to check
+    and explicitly extract the value to get it out.
+    Currently T is required to be a `Copyable` so we can implement
+    copy/move for Result and allow it to be used in collections itself.
+
+    Parameters:
+        T: The type of value stored in the `Result`.
+        E: The error type of the Result.
+
+    Examples:
+
+        ```mojo
+        from collections.result import Result
+        var a = Result(1)
+        var b = Result[Int]()
+        if a:
+            print(a.value())  # prints 1
+        if b:  # bool(b) is False, so no print
+            print(b.value())
+        var c = a.or_else(2)
+        var d = b.or_else(2)
+        print(c)  # prints 1
+        print(d)  # prints 2
+        ```
+        And if more information about the returned Error is wanted it is available.
+        ```mojo
+        from collections.result import Result
+        var a = Result(1)
+        var b = Result[Int](err=Error("something went wrong"))
+        var c = Result[Int](None, Error("error 1"))
+        var d = Result[Int](err=Error("error 2"))
+        if a:
+            print(a.err)  # prints ""
+        if not b:
+            print(b.err) # prints "something went wrong"
+        if c.err:
+            print("c had an error")
+        # TODO: pattern matching
+        if String(d.err_value()) == "error 1":
+            print("d had error 1")
+        elif String(d.err_value()) == "error 2":
+            print("d had error 2")
+        ```
+        A Result with an Error can also be retuned early:
+        ```mojo
+        fn func_that_can_err[A: Copyable]() -> Result[A]:
+            ...
+
+        fn return_early_if_err[T: Copyable, A: Copyable]() -> Result[T]:
+            var result: Result[A] = func_that_can_err[A]()
+            if not result:
+                # the internal err gets transferred to a Result[T]
+                return result
+                # its also possible to do:
+                # return None, Error("func_that_can_err failed")
+            var val = result.value()
+            var final_result: T
+            ...
+            return final_result
+        ```
+        A Result can be unwrapped within a raising function:
+        ```mojo
+        fn func_that_can_err[A: Copyable]() -> Result[A]:
+            ...
+
+        fn return_early_if_err[T: Copyable, A: Copyable]() raises  -> T:
+            var val = func_that_can_err[A]()[]
+            var final_result: T
+            ...
+            return final_result
+        ```
+    """
+
+    # Iterator aliases
+    comptime IteratorType[
+        iterable_mut: Bool, //, iterable_origin: Origin[iterable_mut]
+    ]: Iterator = Self
+    comptime Element = Self.T
+
+    comptime _container_type = InlineArray[Self.T, 1]
+    comptime _type = Variant[Self._container_type, Self.E]
+    var _value: Self._type
+
+    # ===-------------------------------------------------------------------===#
+    # Life cycle methods
+    # ===-------------------------------------------------------------------===#
+
+    fn __init__(out self):
+        """Construct a `Result` with an empty error."""
+        self = Self(err={})
+
+    @implicit
+    fn __init__(out self, var value: Self.T):
+        """Construct an `Result` containing a value.
+
+        Args:
+            value: The value to store in the `Result`.
+        """
+        self._value = {Self._container_type(value^)}
+
+    @implicit
+    fn __init__(out self, var err: Self.E):
+        """Construct a `Result` containing an error.
+
+        Args:
+            err: The error to store in the `Result`.
+        """
+        self._value = {err^}
+
+    # TODO(MSTDL-715):
+    #   This initializer should not be necessary, we should need
+    #   only the initializer from a `NoneType`.
+    @doc_private
+    @implicit
+    fn __init__(out self, value: NoneType._mlir_type):
+        """Construct a `Result` with an empty error.
+
+        Args:
+            value: Must be exactly `None`.
+        """
+        self = Self(err={})
+
+    @implicit
+    fn __init__(out self, value: NoneType):
+        """Construct a `Result` with an empty error.
+
+        Args:
+            value: Must be exactly `None`.
+        """
+        self = Self(err={})
+
+    @implicit
+    fn __init__(out self, var value: type_of(Tuple(None, Self.E()))):
+        """Construct a `Result` with an empty error.
+
+        Args:
+            value: Must be exactly `None`.
+        """
+        self = UnsafePointer(to=value[1]).unsafe_mut_cast[True]().take_pointee()
+
+    @implicit
+    fn __init__(
+        out self, var value: Result[_, Self.E]
+    ) where not _type_is_eq_parse_time[value.T, Self.T]():
+        """Construct a `Result` from the error in another `Result` type.
+
+        Args:
+            value: Must be exactly `None`.
+        """
+        self = Self(err=value.err_unsafe_take() if not value else {})
+
+    # ===-------------------------------------------------------------------===#
+    # Operator dunders
+    # ===-------------------------------------------------------------------===#
+
+    fn __is__(self, other: NoneType) -> Bool:
+        """Return `True` if the Result has no value.
+
+        Args:
+            other: The value to compare to (None).
+
+        Returns:
+            True if the Result has no value and False otherwise.
+
+        Notes:
+            It allows you to use the following syntax:
+            `if my_Result is None:`.
+        """
+        return not self.__bool__()
+
+    fn __isnot__(self, other: NoneType) -> Bool:
+        """Return `True` if the Result has a value.
+
+        Args:
+            other: The value to compare to (None).
+
+        Returns:
+            True if the Result has a value and False otherwise.
+
+        Notes:
+            It allows you to use the following syntax:
+            `if my_Result is not None:`.
+        """
+        return self.__bool__()
+
+    fn __eq__(self, rhs: NoneType) -> Bool:
+        """Return `True` if a value is not present.
+
+        Args:
+            rhs: The `None` value to compare to.
+
+        Returns:
+            `True` if a value is not present, `False` otherwise.
+        """
+        return self is None
+
+    fn __eq__[
+        _T: Equatable & Copyable & Movable
+    ](self: Result[_T], rhs: Result[_T]) -> Bool:
+        """Return `True` if this is the same as another `Result` value,
+        meaning both are absent, or both are present and have the same
+        underlying value.
+
+        Parameters:
+            _T: The type of the elements in the list. Must implement the
+                traits `Copyable`, `Movable` and `Equatable`.
+
+        Args:
+            rhs: The value to compare to.
+
+        Returns:
+            True if the values are the same.
+        """
+        if self:
+            if rhs:
+                return self.value() == rhs.value()
+            return False
+        return not rhs
+
+    fn __ne__(self, rhs: NoneType) -> Bool:
+        """Return `True` if a value is present.
+
+        Args:
+            rhs: The `None` value to compare to.
+
+        Returns:
+            `False` if a value is not present, `True` otherwise.
+        """
+        return self is not None
+
+    fn __ne__[
+        _T: Equatable & Copyable & Movable, //
+    ](self: Result[_T], rhs: Result[_T]) -> Bool:
+        """Return `False` if this is the same as another `Result` value,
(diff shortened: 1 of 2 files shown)


mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+from sys.intrinsics import _type_is_eq, _type_is_eq_parse_time
+
+# ===-----------------------------------------------------------------------===#
+# Result
+# ===-----------------------------------------------------------------------===#
+
+
+struct Result[
+    T: Copyable & Movable,
+    E: Copyable & Movable & Defaultable = Error,
```
NathanSWard, line 101: Defaulting to `Error` seems pretty reasonable, but requiring the error type to be `Defaultable` seems like an unneeded requirement and would make error types need to implement `Defaultable` which may not make sense for all error types.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+    comptime _type = Variant[Self._container_type, Self.E]
+    var _value: Self._type
+
+    # ===-------------------------------------------------------------------===#
+    # Life cycle methods
+    # ===-------------------------------------------------------------------===#
+
+    fn __init__(out self):
+        """Construct a `Result` with an empty error."""
+        self = Self(err={})
```
NathanSWard, line 205: Personally, I'm not a fan of this default constructor.
A `Result` should be either explicitly constructed with either a value or error - not implicitly with a defaulted error.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+    fn __init__(out self, value: NoneType._mlir_type):
+        """Construct a `Result` with an empty error.
+
+        Args:
+            value: Must be exactly `None`.
+        """
+        self = Self(err={})
+
+    @implicit
+    fn __init__(out self, value: NoneType):
```
NathanSWard, line 239: Ditto with these constructors. Being to to write `Result[MyType, MyError](None)` does not feel nice. It's a bit misleading too since `None` implies this result holds "nothing" not necessarily the "defaulted" error.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+    @implicit
+    fn __init__(
+        out self, var value: Result[_, Self.E]
+    ) where not _type_is_eq_parse_time[value.T, Self.T]():
+        """Construct a `Result` from the error in another `Result` type.
+
+        Args:
+            value: Must be exactly `None`.
+        """
+        self = Self(err=value.err_unsafe_take() if not value else {})
```
NathanSWard, line 265: For the first pass of `Result` we should try to keep the API minimal. We can add these extra constructors later if we think they're necessary but I don't believe they will be super useful (most of the time) and only in very niche cases.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+
+    # ===-------------------------------------------------------------------===#
+    # Operator dunders
+    # ===-------------------------------------------------------------------===#
+
+    fn __is__(self, other: NoneType) -> Bool:
```
NathanSWard, line 271: Writing `result is None` does not semantically make sense (imo). The result is either
1) Holding a result
2) Holding an error
the result is never in a `None` state.

a user (can and should imo) rust write `if result` or `if result.is_ok()` (or something like that)

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+        Returns:
+            False if the values are the same.
+        """
+        return not (self == rhs)
+
+    # ===-------------------------------------------------------------------===#
+    # Trait implementations
+    # ===-------------------------------------------------------------------===#
+
+    fn __iter__(ref self) -> Self.IteratorType[origin_of(self)]:
```
NathanSWard, line 369: We can probably delay conforming this to `Iterable` until we have actual conditional conformance. Otherwise this enforces `T` and `E` to both also be `Copyable` which is not a super nice requirement.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+            abort(
+                "`Result.take()` called trying to take a value from a"
+                "`Result` with an error. Consider"
+                " using `if Result:` to check whether the `Result` is empty"
+                " before calling `.take()`, or use `.or_else()` to provide"
+                " a default value."
+            )
+        return self.unsafe_take()
+
+    fn err_take(mut self) -> Self.E:
```
NathanSWard, line 601: Having this be name `take_err()` reads more cleanly than `err_take`. It should also provide parity by having a `take_` prefix.

Ditto with `unsafe_` versions.

IMO: 
`take()`
`take_err()`
`unsafe_take()`
`unsafe_take_err()`

NathanSWard (commented):
Took a first pass at some initial comments on this.
I'll need to take a second pass but I do think a `Result` type would be nice in the stdlib.

Specifically I would LOVE TO WRITE this:
```mojo
var res: Result[Int, Error] = try some_raising_function()
```

I'll bring this up with the stdlib team next week too. Since now we have typed errors we need to think how this can nicely play with that.

mojo/stdlib/stdlib/collections/result.mojo
martinvuyk: If it is not defaultable I don't know how we could implement `take()` for the inner value, because if we replace the inner array with an uninitialized array then the `__bool__` check will falsely return `True` given the type is not an error

NathanSWard: For this I would have `take` require a `deinit self`.
Something like:
```mojo
fn take(deinit self) -> Self.T: ...

fn take_err(deinit self) -> Self.E: ...
```

Not sure if it would actually ever be useful to "reuse" the result type.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+from memory import LegacyOpaquePointer as OpaquePointer
+
+from sys.intrinsics import _type_is_eq, _type_is_eq_parse_time
+
+# ===-----------------------------------------------------------------------===#
+# Result
+# ===-----------------------------------------------------------------------===#
+
+
+struct Result[
```
owenhilyard, line 99: Should it be a linear type to prod users to deal with it? We can add a `.ignore()` that gets rid of it if both the value and error variants are non-linear.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+    T: Copyable & Movable,
```
owenhilyard, line 100: Can we use the conditional conformance workaround to toss `Copyable` here? I can think of a lot of things which could go here which shouldn't be copied. Same with the error variant, a common result type in rust for `queue.try_push(value)` is `Result[None, type_of(value)]` where you get the value back on a full queue instead of blocking, which means that the error variant might need to be non-copyable as well.

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard: Can we get `abort` and `raises` variants of both versions of `take`?

`fn take(deinit self) raises Self.E -> Self.T` and `fn take_err(deinit self) raises Self.T -> Self.E` would end up in a similar place to Rust's `?` operator, and that provides an `out` for linear types.

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+            print(b.err) # prints "something went wrong"
+        if c.err:
+            print("c had an error")
+        # TODO: pattern matching
+        if String(d.err_value()) == "error 1":
+            print("d had error 1")
+        elif String(d.err_value()) == "error 2":
+            print("d had error 2")
+        ```
+        A Result with an Error can also be retuned early:
```
owenhilyard, line 159: s/retuned/returned

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard: +1

mojo/stdlib/stdlib/collections/result.mojo
```diff
@@ -0,0 +1,713 @@
...
+    # ===-------------------------------------------------------------------===#
+    # Life cycle methods
+    # ===-------------------------------------------------------------------===#
+
+    fn __init__(out self):
+        """Construct a `Result` with an empty error."""
+        self = Self(err={})
+
+    @implicit
+    fn __init__(out self, var value: Self.T):
```
owenhilyard, line 208: Can this still be safely implicit when `Self.T == Self.E` (thinking of `c_int, c_int` from FFI).

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard, line 265: I'm not a fan of an implicit ctor that can invoke UB. I think that happens when `value` is `Ok` should be well defined.

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard: Could we make it an owning iterator?

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard: I think that after raising and aborting versions are added, it might make sense to collapse some of these down into singular functions.

owenhilyard (changes requested):
I like the idea but I think we need to tighten up some semantics and not do either UB or abort by default.

owenhilyard:
> Took a first pass at some initial comments on this. I'll need to take a second pass but I do think a `Result` type would be nice in the stdlib.
> 
> Specifically I would LOVE TO WRITE this:
> 
> ```mojo
> var res: Result[Int, Error] = try some_raising_function()
> ```
> 
> I'll bring this up with the stdlib team next week too. Since now we have typed errors we need to think how this can nicely play with that.

Once concern I have with this syntax is how it would interact with a library like Rust's tracing that lets you add extra context info to errors as they travel up the stack.

ex:

```mojo
# mojo
var weights = (
try # I don't think it's instantly clear which error I'm choosing to capture here
    (try transformers_load(...)).context("during transformers lib load from disk").take_raising()
    .to(ctx)
).context("while moving to GPU").take_raising()
```

vs 

```rust
// rust
var weights = transformers_load(...)).context("during transformers lib load from disk")?
    .to(ctx)).context("while moving to GPU")?;
```

However, I can see how dealing with postfix syntax could be annoying given that Mojo is "raising first" so I can live with big piles of nested `try`. I prefer the more sequential style of processing errors to try-catch because IO-heavy code means most of your functions can fail and you may want to take very different actions based on which function returned a particular IO error. Having `Result` at all is a win here, but I also want to try to make code that handles all errors as easy to write as possible. Perhaps match expressions (not statements) would help fix that issue.

mojo/stdlib/stdlib/collections/result.mojo
martinvuyk: @owenhilyard I'm guessing you mean to use the case that a result's E type is `Never`, then the take would abort internally? I think that's a neat idea.

For context, we do have an empty `__getitem__` that should raise, but return a reference

mojo/stdlib/stdlib/collections/result.mojo
martinvuyk: Yeah I also thought of that case, that's why the internal structure uses an `InlineArray`. I think the compiler should figure out that using this in a generic context is ambiguous and require explicit use. If we remove this then we get more verbose code for returns which I'm not a huge fan of. Not sure if this is solvable at all, if not then yeah we'll have to make this explicit

lattner:
Thank you for working on this Martin!

mojo/stdlib/stdlib/collections/result.mojo
NathanSWard: Not sure it's worth making _this_ a linear type. The compiler will already emit a warning if you ignore a result.
```mojo
fn foo() -> Result[..]: ...

foo() # <- Emits compiler error on ignored result
```

Explicitly writing `_ = foo()` feels okay to me?

mojo/stdlib/stdlib/collections/result.mojo
NathanSWard: I think if the error and value type are the same this should still work. You can disambiguate by using the named constructors `Result(value=..)` and `Result(error=..)`.
And it also works in a generic context too 
See here: https://godbolt.org/z/sb3beGPv3

mojo/stdlib/stdlib/collections/result.mojo
NathanSWard: +1
Though this constructor probably shouldn't exist and I don't think it'll be able to once `Defaultable` is removed from the error type.

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard: @martinvuyk I provided an example in a further comment, but for some types of queues you might have a `fn try_push(var value: T) raises T:` which raises if the queue is full and otherwise returns `None`. If the error type is Never, that means it can raise, which means it becomes an infallible operation.

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard: `_ = foo()` is probably fine so long as we get a switch in the future to make that into an error. (ex: `#![deny(ignored_result)]`)

mojo/stdlib/stdlib/collections/result.mojo
owenhilyard: I guess that once we have sum types this should move over to that, at which point it will be explicit.

Issue #5209, opened 2025-08-23
[Feature Request] Change `repr` for collection types to consistently print their own type


rd4com:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](<https://docs.modular.com/mojo/roadmap.html#overall-priorities>) and I believe this request falls within the priorities.

### What is your request?

An followup for `repr` was asked by (ping to @laszlokindrat),<br>so creating this to hep define the task.

see: [https://github.com/modular/modular/pull/5189](<https://github.com/modular/modular/pull/5189>)

I'd be glad to go about this !

Here is an overview of what types do on repr:

Collections type do not always print their own type, some hints at it ("\[\]","{}"):<br>-Set: "{a,b,c}"<br>-List: "\[a,b,c\]"<br>-Dict: "{a:b,c:d}"<br>-Deque: "Deque(a,b,c)"<br>-LinkedList: "LinkedList(a,b,c)"<br>-Optional: "Optional(a)"

SIMD include it's DT:

* repr -> "SIMD\[DType.float64, 1\](1.0)"

Optional, deque, LinkedList, List,.. don't include T:

* repr -> "\[0\], {0}, {a:b}"

### What is your motivation for this change?

Asked by @laszlokindrat

### Any other details?

*No response*

christoph-schlumpf:
`List` is the default type for list literals `[]`. Should `list.repr()` therefore just return the list literal without the type name when it is a `List`?

akovacevicHTEC:
What would be desirable implementation for this? Do we want to print all collection types like this, e.g. :
-Set: "Set{a,b,c}"
-List: "List[a,b,c]"
-Dict: "Dict{a:b,c:d}"
-Deque: "Deque(a,b,c)"
-LinkedList: "LinkedList(a,b,c)"
-Optional: "Optional(a)"

martinvuyk:
I think we should lean into refactoring the stringification related traits into something more performant rather than keep trying to implement different methods for repr or str. IMO we should lean into the concept of a `FormatSpec` that is passed to every stringification function down the pipe. An example would be:

`Writer` would become:
```mojo
trait StringifySink:
  fn append(mut self, source: Some[StringifySource], *, spec: FormatSpec):
```
and `Writable`:
```mojo
trait StringifySource: 
  fn stringify(self, mut sink: Some[StringifySink], *, spec: FormatSpec):
```

I opened issue #4667 to talk about this a while back.

[...]

PR #5649, opened 2025-12-10
[Feature Request] Change repr for collection types to consistently print their own type

rajuptvs:
This is a draft pr to unify the behaviour  of repr as discussed in #5209 and also to include datatype. 

Its not perfect yet
Below is the print statement of how this looks

```bash
repr for dict:  {1: SIMD[DType.float64, 1](1.1), 2: SIMD[DType.float64, 1](2.2)} #hadn't made any changes  
repr for list:  List[Int][1, 2, 3]
repr for linked list:  LinkedList[SIMD[DType.float64, 1]](1.0, 2.5)
repr for set:  Set[Int]{1, 2, 3}
repr for deque:  Deque[Int](1, 2, 3)
```

Appreciate any feedback and pointers for further refinement if needed. 

Also one quick question would be, dict doesn't have a __repr__ method built in rather uses __str__, is there a need to fix that ?


rajuptvs:
@JoeLoser @laszlokindrat  was wondering if this falls in line with the expectation of the outline specified in the issue #5209

martinvuyk:
I think we should hold off on the implementation. I'll explain more in the issue you mentioned, sorry I didn't read it before

NathanSWard:
Since `Writable` is not replacing `Stringable` & `Representable`.

PR #5657, opened 2025-12-14
[max][layout] Add load_scalar method to LayoutTensor for scalar element access

### LINKED ISSUE #4847: [Feature Request] add a scalar `__getitem__` for `LayoutTensor`
### What is your request?

Add a `__getitem__` overload to `LayoutTensor` that returns `Scalar[DType]`.

Right now, `__getitem__` returns `LayoutTensor.element_type` which is defined as:

```python
struct LayoutTensor[...]:
    alias element_size = element_layout.size()
    alias element_type = SIMD[dtype, Self.element_size]
    
    @always_inline("nodebug")
    fn __getitem__(self, *dims: Int) -> Self.element_type:
        var strides = self.runtime_layout.stride.value
        var offset = Self._get_offset(strides, dims)
        return self._load_offset(offset)
```

Since `Self.element_size` is not guaranteed to be `1`, when working in a generic context Mojo is returning a generically sized `SIMD` value. This constantly surprises users and continues to cause misunderstandings.

### Any other details?

_No response_

KrxGu:
Fixes #4847

## Problem

Currently, `LayoutTensor.__getitem__` returns `Self.element_type` which is `SIMD[dtype, element_size]`. Since `element_size` isn't guaranteed to be 1 (e.g., in tiled/vectorized layouts), users working in generic contexts often get a SIMD vector when they just want a single scalar value. This leads to confusion and awkward workarounds like `tensor[i, j][0]`.

## Solution

This PR adds a `load_scalar` method that explicitly returns `Scalar[dtype]` for cases where you need a single value:

```mojo
# Before: have to extract lane 0 manually
var val = tensor[i, j][0]

# After: clear intent, returns Scalar directly  
var val = tensor.load_scalar(i, j)
```

## Changes

Added `_load_scalar_offset` private helper that reuses existing `_load_offset `and returns lane 0
Added `load_scalar[*Tys: Indexer]` for variadic indices like `tensor.load_scalar(i, j)`
Added load_scalar(RuntimeTuple) overload for coordinate-based access
Added tests covering static layouts, dynamic layouts, and RuntimeTuple coordinates

## Design note

I went with a named method rather than a `__getitem__` overload because having two methods with identical parameter signatures but different return types would cause ambiguity issues. The `load_scalar` name also makes the intent explicit at the call site, which seems cleaner for this use case.

### DIFF
--- a/max/kernels/test/layout/test_layout_tensor_load_scalar.mojo
+++ b/max/kernels/test/layout/test_layout_tensor_load_scalar.mojo
@@ -1,0 +1,211 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Tests for LayoutTensor.load_scalar() method.
+
+These tests verify that the load_scalar method correctly returns scalar values
+from tensors, addressing the issue where __getitem__ returns SIMD[dtype, element_size]
+which can be surprising in generic contexts when element_size > 1.
+"""
+
+from layout import Layout, LayoutTensor, RuntimeLayout, RuntimeTuple
+from layout._fillers import arange
+from layout.int_tuple import UNKNOWN_VALUE
+from memory import LegacyUnsafePointer as UnsafePointer
+from testing import assert_equal
+
+
+def main():
+    test_load_scalar_static_layout()
+    test_load_scalar_dynamic_layout()
+    test_load_scalar_with_runtime_tuple()
+    test_load_scalar_matches_getitem_lane0()
+    test_load_scalar_vectorized_element_size_gt_1()
+    print("All tests passed!")
+
+
+fn test_load_scalar_static_layout() raises:
+    """Test load_scalar with a static 2x3 row-major layout."""
+    print("== test_load_scalar_static_layout")
+
+    comptime layout = Layout.row_major(2, 3)
+    var storage = UnsafePointer[Float32].alloc(6)
+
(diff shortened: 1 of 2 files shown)


KrxGu:

KrxGu:
Hey @JoeLoser , this PR is ready for review!

I've added a `load_scalar` method to `LayoutTensor` that returns Scalar[dtype] for cases where users need a single value instead of the SIMD vector that `__getitem__` returns.

Went with a named method rather than a `__getitem__` overload to avoid signature ambiguity - happy to discuss if you'd prefer a different approach.

CI all green(except the expected CLA test), Good To Merge!!

Let me know if you have any feedback!

max/kernels/test/layout/test_layout_tensor_load_scalar.mojo
```diff
@@ -0,0 +1,211 @@
...
+    test_load_scalar_vectorized_element_size_gt_1()
+    print("All tests passed!")
+
+
+fn test_load_scalar_static_layout() raises:
+    """Test load_scalar with a static 2x3 row-major layout."""
+    print("== test_load_scalar_static_layout")
+
+    comptime layout = Layout.row_major(2, 3)
+    var storage = UnsafePointer[Float32].alloc(6)
```
martinvuyk: You can just use a `List` or an `InlineArray` as the backing storage, no need to use a pointer

max/kernels/test/layout/test_layout_tensor_load_scalar.mojo
```diff
@@ -0,0 +1,211 @@
...
+from testing import assert_equal
+
+
+def main():
+    test_load_scalar_static_layout()
+    test_load_scalar_dynamic_layout()
+    test_load_scalar_with_runtime_tuple()
+    test_load_scalar_matches_getitem_lane0()
+    test_load_scalar_vectorized_element_size_gt_1()
+    print("All tests passed!")
```
martinvuyk: please remove the print statements and just use `from testing import TestSuite` and auto import the tests, you can see other places in the stdlib where it's used

martinvuyk (commented):
Hi, thanks for contributing!

I'll ask a question I have regarding the actual need for this in the linked issue, but I'll leave a couple of comments of things that stand out to me to give you some feedback

KrxGu:
@martinvuyk i have made the required changes.
CI all green, Good To Merge!!
cc: @lsh

lsh:
!sync

PR #5659, opened 2025-12-14
[stdlib] Fix list and dict repr formatting

lamija-i:
## Summary
This PR fixes the string representation (`repr`) formatting for `list` and `dict` collections to ensure consistent and correct output.

## What changed
- Updated `repr` implementation for `list` to produce the expected formatted output
- Updated `repr` implementation for `dict` to ensure correct key–value representation
- Updated and added tests to cover the corrected `list` and `dict` `repr` behavior

## Why this change is needed
The previous `repr` formatting for `list` and `dict` could produce inconsistent or incorrect output, which made debugging and value inspection less reliable. This change aligns the behavior with expected and documented conventions.


NathanSWard:
Since `Writable` is not replacing `Stringable` & `Representable`.

PR #5660, opened 2025-12-15
Add Representable repr for LinkedList and Deque

nejraberberovic:
Implement inspect-style repr for collections while preserving existing writer-based output. 
Changes:

Add write_to_repr and Representable-based repr for Deque and LinkedList.
Keep write_to variants for Writable elements (no change to runtime algorithms).
Align builtin repr dispatch in mojo/stdlib/stdlib/builtin/repr.mojo.

Why:
Make collection repr output consistent and use repr(elem) (e.g., Deque('a','b'), LinkedList(1,2,3)) without introducing overload ambiguity or changing collection semantics.
To reproduce: checkout branch, run the collection tests above and inspect repr output for Deque/LinkedList.


NathanSWard:
Hi @nejraberberovic and thank you for the PR!
The stdlib is actually moving in a slightly different direction for `repr` implementations and that is via `downcast`ing the element types to write them instead of providing overloads of the `repr` function.
If you wanted, you could take a stab at implementing that for `LinkedList`- if you look at the current implementations of `__str__`, `__repr__` and `write_to` for `List` that should give you a good idea on how to start :)

NathanSWard:
Since `Writable` is now replacing `Stringable` & `Representable`.

PR #5661, opened 2025-12-15
[stdlib] Deprecate `Codepoint.ord()`

### LINKED ISSUE #5645: [Feature Request] [stdlib] [proposal] Deprecate `Codepoint.ord()` in favor of just a constructor `Codepoint(str)`
### Review Mojo's priorities

### What is your request?

As title. I think that making it a static method of that name was to mimic Python's freestanding `ord` function; and I believe that is not necessary anymore since we are leaning more towards improving Python APIs where we can. A constructor feels much more natural and simple.

Python chose the word `ordinal` because it means a position in a series. They use it in the datetime logic as well. But we aren't dealing with an `Int` per se but an instance of a `Codepoint` (which is arguably an ordinal unicode value, but nevermind), so IMO we should reflect that using a constructor

#5637 adds a constructor for `StringLiteral` and I think it is more intuitive to use a constructor like that than the `Codepoint.ord()` static method. 

### Any other details?

CC: @KrxGu I think you'll like this idea

martinvuyk:
Deprecate `Codepoint.ord()`. Closes #5645

### DIFF
--- a/mojo/stdlib/std/collections/string/codepoint.mojo
+++ b/mojo/stdlib/std/collections/string/codepoint.mojo
@@ -65,7 +65,7 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
     from std.testing import assert_true
 
     # Create a codepoint from a character
-    var c = Codepoint.ord('A')
+    var c = Codepoint('A')
 
     # Check properties
     assert_true(c.is_ascii())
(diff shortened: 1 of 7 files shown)


mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -138,6 +138,33 @@ struct Codepoint(
         """
         self._scalar_value = UInt32(Int(codepoint))
 
+    fn __init__(out self, string: StringSlice[mut=False]):
```
YkdWaWEzVmphR1Z1: I'm curious if the constructor should be raising to avoid confusion between the StringSlice and StringLiteral constructors.

martinvuyk: I don't think that's necessary. AFAIK this isn't the first case of having both `StringSlice` and `StringLiteral` constructors, and the compiler should figure out the shortest path (least amnt of conversions) to building a type

ConnorGray:
!sync

ConnorGray:
!sync

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -138,6 +137,33 @@ struct Codepoint(
...
+        var char, num_bytes = Codepoint.unsafe_decode_utf8_codepoint(
+            string.as_bytes()
+        )
+
+        debug_assert(
+            string.byte_length() == Int(num_bytes),
+            "input string must be one character",
+        )
+
+        self = char
```
ConnorGray, line 162: **Suggestion** Instead of duplicating the implementation, could we move the currently implementation from `Codepoint.ord()` here, and then make `Codepoint.ord()` call the new initializer?

That would avoid duplication, and preserve the existing safety check that `Codepoint.ord()` performs that the `StringSlice` is never empty (even when debug assertions are disabled).

PR #5677, opened 2025-12-16
[bazel] Fix wheel download URLs

Ahajha:


bazel/mojo.MODULE.bazel
```diff
@@ -2,11 +2,11 @@
 
 PACKAGE_VERSION = "26.1.0.dev2025121605"
 
-BASE_URL = "https://dl.modular.com/public/nightly/python"
+BASE_URL = "https://us-central1-python.pkg.dev/distribution-prod-0a7c7b3/nightly"
```
patrickdoc, line 5: Why doesn't the `dl.modular.com` url work? Does the actual index handle mapping the request to the underlying file?

Ahajha: Totally unsure. Maybe the redirect isn't mapping these files specifically? Or maybe it's redirecting incorrectly?

Relatedly, the URL has the package name as an extra jump, so for example the mojo compiler wheel URL looks something like `https://us-central1-python.pkg.dev/distribution-prod-0a7c7b3/nightly/mojo-compiler/mojo_compiler-0.26.1.0.dev2025121605-py3-none-macosx_13_0_arm64.whl`, note the `.../mojo-compiler/mojo_compiler...`

Ahajha: That's probably what broke this. This isn't a concern when using indexes.

Ahajha:
Closing since we reverted the redirect

PR #5685, opened 2025-12-18
[Kernels] Migrate CPU matmul kernels from LegacyUnsafePointer to UnsafePointer

### LINKED ISSUE #5670: [Feature Request] Migrate `kernels/src/linagl/matmul/cpu` off of `LegacyUnsafePointer`.
### Review Mojo's priorities

### What is your request?

See [parent issue](https://github.com/modular/modular/issues/5665)

The following kernel files should be migrated off of `LegacyUnsafePointer`. (complexity noted next to file)

```
linalg/matmul/cpu/apple_accelerate.mojo (Medium)
linalg/matmul/cpu/default.mojo (Medium)
linalg/matmul/cpu/i8mm.mojo (Medium)
linalg/matmul/cpu/impl.mojo (Low)
linalg/matmul/cpu/neon.mojo (Medium)
linalg/matmul/cpu/vnni.mojo (Medium)
```

`LegacyUnsafePointer` is going to be deprecated and removed.

### Any other details?

_No response_

KrxGu:
Fixes #5670

This PR migrates the CPU matmul kernel implementations to use the new `UnsafePointer` API, removing the deprecated `LegacyUnsafePointer` alias.

### Changes
Updated 6 files in `max/kernels/src/linalg/matmul/cpu/`:
- `impl.mojo` - Core matmul orchestration
- `apple_accelerate.mojo` - Apple Accelerate BLAS wrapper
- `default.mojo` - Default CPU microkernel
- `i8mm.mojo` - ARM i8mm instruction microkernel
- `neon.mojo` - ARM NEON microkernel
- `vnni.mojo` - x86 VNNI microkernel

### Implementation Notes
The new `UnsafePointer` API requires explicit mutability (`mut=True/False`) and origin tracking. Where needed for compatibility with unmigrated code (like `_Accumulator`) and FFI interfaces, we use `rebind[]` to convert between pointer types.

### Testing
- All matmul tests pass
- Code formatted with `bazelw run //bazel/lint:format`

### DIFF
--- a/max/kernels/src/linalg/matmul/cpu/apple_accelerate.mojo
+++ b/max/kernels/src/linalg/matmul/cpu/apple_accelerate.mojo
@@ -13,7 +13,6 @@
 
 from collections import OptionalReg
 from math import fma
-from memory import LegacyUnsafePointer as UnsafePointer
 from os import abort
 from sys import CompilationTarget, simd_width_of
 from sys.ffi import _get_dylib_function as _ffi_get_dylib_function
@@ -46,12 +45,12 @@ comptime cblas_gemm_type = fn (
     Int32,
     Int32,
     Float32,
-    UnsafePointer[Float32],
+    LegacyUnsafePointer[Float32],
     Int32,
-    UnsafePointer[Float32],
+    LegacyUnsafePointer[Float32],
     Int32,
     Float32,
-    UnsafePointer[Float32],
+    LegacyUnsafePointer[Float32],
     Int32,
 ) -> None
 
@@ -187,12 +186,12 @@ fn _cblas_f32[
         n,
         k,
         alpha,
-        rebind[UnsafePointer[Float32]](a_ptr),
+        rebind[LegacyUnsafePointer[Float32]](a_ptr),
         lda,
-        rebind[UnsafePointer[Float32]](b_ptr),
+        rebind[LegacyUnsafePointer[Float32]](b_ptr),
         ldb,
         beta,
-        rebind[UnsafePointer[Float32]](c_ptr),
+        rebind[LegacyUnsafePointer[Float32]](c_ptr),
         ldc,
     )
 
@@ -261,14 +260,16 @@ fn apple_gemv[
     var N = b.dim[0]() if transpose_b or b_packed else b.dim[1]()
 
     var transposed_b = NDBuffer[b.type, 2, MutAnyOrigin]()
-    var transposed_b_ptr = UnsafePointer[Scalar[b.type]]()
+    var transposed_b_ptr = LegacyUnsafePointer[Scalar[b.type]]()
 
     # If both b_packed and transpose_b are False, we need to transpose B at
     # runtime (which is suboptimal, but enables faster gemv below).
     @parameter
     if b_packed == False and not transpose_b:
         var transposed_b_shape = Index(b.dim[1](), b.dim[0]())
-        transposed_b_ptr = UnsafePointer[Scalar[b.type]].alloc(b.num_elements())
+        transposed_b_ptr = LegacyUnsafePointer[Scalar[b.type]].alloc(
+            b.num_elements()
+        )
         transposed_b = NDBuffer[b.type, 2](transposed_b_ptr, transposed_b_shape)
 
         pack_b_ndbuffer[
@@ -377,15 +378,15 @@ fn apple_matmul[
             ldc,
             alpha,
             beta,
-            rebind[UnsafePointer[Float32, address_space = c.address_space]](
-                c.data
-            ),
-            rebind[UnsafePointer[Float32, address_space = a.address_space]](
-                a.data
-            ),
-            rebind[UnsafePointer[Float32, address_space = b.address_space]](
-                b.data
-            ),
+            rebind[
+                LegacyUnsafePointer[Float32, address_space = c.address_space]
+            ](c.data),
+            rebind[
+                LegacyUnsafePointer[Float32, address_space = a.address_space]
+            ](a.data),
+            rebind[
+                LegacyUnsafePointer[Float32, address_space = b.address_space]
+            ](b.data),
         )
 
         @parameter
--- a/max/kernels/src/linalg/matmul/cpu/apple_accelerate.mojo
+++ b/max/kernels/src/linalg/matmul/cpu/apple_accelerate.mojo
@@ -13,6 +13,7 @@
 
 from collections import OptionalReg
 from math import fma
+from memory import alloc
 from os import abort
 from sys import CompilationTarget, simd_width_of
 from sys.ffi import _get_dylib_function as _ffi_get_dylib_function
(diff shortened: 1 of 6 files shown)


KrxGu:
@NathanSWard Hi! I've updated the CPU matmul kernels to use the new `UnsafePointer` API as discussed in #5670. 

The migration keeps `_Accumulator` using `LegacyUnsafePointer` for now and uses `rebind[]` at the interop points. All tests are passing.

CI all green. Good To Merge!!
Let me know if you'd prefer a different approach or if there's anything that needs adjustment. Happy to make changes!

abduld (approved):

max/kernels/src/linalg/matmul/cpu/i8mm.mojo
```diff
@@ -102,12 +102,16 @@ struct LoadStore_i8mm[
     @always_inline
     fn _store_c_tile(
         mut self,
-        c_ptr: UnsafePointer[Scalar[Self.dtype]],
+        c_ptr: UnsafePointer[mut=True, Scalar[Self.dtype], **_],
```
NathanSWard, line 105:
```suggestion
        c_ptr: UnsafePointer[mut=True, Scalar[Self.dtype]],
```

No need to add the `**_` syntax here as the previous pointer had it's address space bound to `GENERIC`.

max/kernels/src/linalg/matmul/cpu/i8mm.mojo
```diff
@@ -102,12 +102,16 @@ struct LoadStore_i8mm[
...
         c_stride: Int,
         tile_n_idx: Int,
         c_bound: IndexList[2],
     ):
-        var c_ptr_loc = c_ptr.offset(tile_n_idx)
+        # Convert to LegacyUnsafePointer for partial_simd_store compatibility
+        var c_ptr_legacy = rebind[LegacyUnsafePointer[Scalar[Self.dtype]]](
```
NathanSWard: **question**
Do we have to convert this to a `LegacyUnsafePointer`? Even though `partial_simd_store` takes a `LegacyUnsafePointer`, you _should_ be able to pass an `UnsafePointer` to it without any needed explicit conversions.

max/kernels/src/linalg/matmul/cpu/impl.mojo
```diff
@@ -388,7 +388,7 @@ struct TiledMatmul[
     #  need to remap every time K and kernel_cols changes.
     fn _view_buffer_as(
         self,
-        b_packed_ptr: UnsafePointer[Scalar[Self.b_type]],
+        b_packed_ptr: UnsafePointer[Scalar[Self.b_type], **_],
```
NathanSWard, line 391:
```suggestion
        b_packed_ptr: UnsafePointer[Scalar[Self.b_type]],
```
Ditto here, no need to unbind the `address_space` parameter :)

max/kernels/src/linalg/matmul/cpu/impl.mojo
```diff
@@ -595,9 +595,9 @@ fn _matmul_cpu_impl[
         comptime alignment = align_of[SIMD[c.type, simd_size]]()
         var kh = align_up(k, 8)
         var mh = align_up(m, 2)
-        var a_packed_ptr = UnsafePointer[Scalar[a.type]]()
+        var a_packed_ptr = LegacyUnsafePointer[Scalar[a.type]]()
         if use_i8mm:
-            a_packed_ptr = UnsafePointer[Scalar[a.type]].alloc(
+            a_packed_ptr = LegacyUnsafePointer[Scalar[a.type]].alloc(
                 mh * kh, alignment=alignment
             )
```
NathanSWard: Ideally these should also be updated to use the new `UnsafePointer` and `alloc` methods.
```mojo
var a_packed_ptr = UnsafePointer[Scalar[a.type], MutOrigin.external]()
if use_i8mm:
    a_packed_ptr = alloc[Scalar[a.type]](mh * kh, alignment=alignment)
```

max/kernels/src/linalg/matmul/cpu/neon.mojo
```diff
@@ -141,7 +141,7 @@ struct Inner_matmul_neon(InnerMatmulKernel, Movable):
                 acc.init(0)
             else:
                 acc.load(
-                    rebind[UnsafePointer[Scalar[c.dtype]]](c_ptr),
+                    rebind[LegacyUnsafePointer[Scalar[c.dtype]]](c_ptr),
```
NathanSWard, line 144: **questions**
Just double checking that the `rebind` is necessary here and we can't just pass `c_ptr` directly?

max/kernels/src/linalg/matmul/cpu/apple_accelerate.mojo
```diff
@@ -261,14 +260,16 @@ fn apple_gemv[
...
 
     # If both b_packed and transpose_b are False, we need to transpose B at
     # runtime (which is suboptimal, but enables faster gemv below).
     @parameter
     if b_packed == False and not transpose_b:
         var transposed_b_shape = Index(b.dim[1](), b.dim[0]())
-        transposed_b_ptr = UnsafePointer[Scalar[b.type]].alloc(b.num_elements())
+        transposed_b_ptr = LegacyUnsafePointer[Scalar[b.type]].alloc(
+            b.num_elements()
+        )
```
NathanSWard: Ditto using `UnsafePointer` and the free standing `alloc` function.

NathanSWard (changes requested):

max/kernels/src/linalg/matmul/cpu/neon.mojo
KrxGu: The rebind is necessary here because _Accumulator in accumulate.mojo still uses `from memory import LegacyUnsafePointer as UnsafePointer`. The load() and store() methods expect LegacyUnsafePointer, so we need to convert from the new UnsafePointer.

If _Accumulator gets migrated to the new API in the future, we can remove these rebind calls.

max/kernels/src/linalg/matmul/cpu/i8mm.mojo
KrxGu: Thanks for the review! I tried removing the **_ syntax, but ran into type inference issues. The functions receive pointers from LayoutTensor[mut=True, **_] which has a generic origin, and without **_ the compiler can't infer the origin parameter at call sites. 

I kept the **_ syntax for now - let me know if there's a better pattern I should use!

NathanSWard: Ahh intersting, thanks for pointing that out!

NathanSWard:
!sync

max/kernels/src/linalg/matmul/cpu/neon.mojo
soraros: I thought we have two-way implicit conversion between the two types?

max/kernels/src/linalg/matmul/cpu/i8mm.mojo
soraros: Is this fix, which relaxes the API as a side effect, really correct? Shouldn't we instead origin-cast the pointer obtained from the LayoutTensor?

PR #5687, opened 2025-12-19
[stdlib] Add `vectorize` overload with predicated tail handling

YichengDWu:
This PR adds a new overload of `vectorize` whose closure takes an additional `evl` (effective vector length) argument. Full SIMD iterations call the closure with `evl == simd_width`, while the remainder (`size % simd_width`) is handled by a single final call with `0 < evl < simd_width`

### Motivation

The existing `vectorize` handles the tail by running scalar iterations (`width=1`) for `size % simd_width`. This is safe and simple, but it prevents users from handling the tail via predication/masking and can add overhead (extra loop + multiple calls).

The new overload enables a common SIMD pattern:

- Keep a fixed SIMD width (`simd_width`) for the closure (compile-time specialization still applies).
- Pass the runtime lane count (`evl`) so the closure can implement masked/predicated loads/stores and avoid OOB.

This is especially useful when tail performance matters, or the closure naturally supports predication. In the future, we can migrate performance-critical kernels from the existing scalar-tail `vectorize` to this EVL-based overload to potentially gain better performance on AVX-512 capable machines while reducing tail overhead.

### New overload:

```mojo
fn vectorize[
    func: fn[width: Int](idx: Int, evl: Int) unified -> None,
    simd_width: Int,
    /,
    *,
    unroll_factor: Int = 1,
](size: Int, closure: func)
```

#### Semantics / Safety
- For full-width blocks: `closure[simd_width](idx, simd_width)`
- For the tail: `closure[simd_width](simd_end, remainder)`
- The closure must honor `evl` to keep memory accesses in-bounds.


abduld:
!sync

PR #5689, opened 2025-12-19
[stdlib] Make Deque conform to Writable, Stringable and Representable

msaelices:
This change adds explicit trait conformance to Deque for:
- Writable: via `write_to()` method
- Stringable: via `__str__()` method
- Representable: via `__repr__()` method

Also removed the obsolete `repr()` overload
for Deque in builtin/repr.mojo since Deque now conforms to Representable directly.
```
from collections import Deque

fn main() raises:
    # Test with Int deque
    var deque = Deque[Int](1, 2, 3, 4, 5)

    print("Deque string:", deque)  # Deque(1, 2, 3, 4, 5)
```


JoeLoser (approved):
LGTM, thanks!

JoeLoser:
!sync

PR #5694, opened 2025-12-20
[stdlib] Make LinkedList conform to Writable, Stringable and Representable

msaelices:
Modernizes `LinkedList` to use explicit trait conformance for `Writable`, `Stringable`, and `Representable` traits, following the same pattern as `Deque` and `InlineArray`.

- Updates struct declaration to include trait conformance
- Removes generic constraints from methods  
- Uses `_constrained_conforms_to` for compile-time checking
- Removes obsolete `repr` overload from `builtin/repr.mojo`


mojo/stdlib/std/collections/linked_list.mojo
```diff
@@ -791,33 +788,46 @@ struct LinkedList[
...
+        Constraints:
+            ElementType must conform to `Representable`.
 
         Args:
             writer: The writer to write the list to.
 
         Notes:
             Time Complexity: O(n) in len(self).
         """
+        _constrained_conforms_to[
```
JoeLoser: **Suggestion** This is called in both `write_to` and in `_write`. Since `write_to` delegates to `_write`, the check here in `write_to` is redundant and can be removed.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
Hi @msaelices, do you mind rebasing and resolving the merge conflicts? I've merged a few of your other recent PRs in this area and now there's some conflicts. Then I'm happy to sync it in and land it. Thanks!

msaelices:
> Hi @msaelices, do you mind rebasing and resolving the merge conflicts? I've merged a few of your other recent PRs in this area and now there's some conflicts. Then I'm happy to sync it in and land it. Thanks!

Sure. Done!

mojo/stdlib/std/collections/linked_list.mojo
msaelices: Good catch! Fixed: https://github.com/msaelices/modular/commit/4e3eea57a39274d8d0ea87bc354222eff196dfad

ConnorGray:
!sync

mojo/stdlib/std/collections/linked_list.mojo
```diff
@@ -826,6 +828,9 @@ struct LinkedList[
         for i in range(len(self)):
             if i:
                 writer.write(", ")
-            writer.write(curr[].value)
+            ref representable_element = trait_downcast[Representable](
+                curr[].value
+            )
+            writer.write(repr(representable_element))
```
ConnorGray, line 834: **Question** This seems to be changing `LinkedList.write_to` to call `Representable` on the elements, where previously it called `Writer` on the elements. Is that the direction we want to go? It would seem more natural to me if `Writer` "recursively" used elements' `Writer` implementations as well. That would also let us avoid allocating a `String` with the `repr(..)` call here.

I recognize there is a broader problem that we have two traits—`Stringable` and `Representable`—that return `String`, but only one trait—`Writer`—that does streaming writing, so we're shoe-horning two semantically different ways of formatting a type into one trait. However, I think `Writer` is probably semantically closer to `Stringable` (human-readable) than it is to `Representable`.

NathanSWard: > It would seem more natural to me if Writer "recursively" used elements' Writer implementations as well.

This is the direction we are eventually going (Once we remove `Representable` and `Stringable` entirely)! 

However, all other collections in the stdlib `List`, `Deque`, `Optional`, etc... all default to using the contained type's `Representable` trait (instead of `Writable`).
This PR at least makes that consistent, then we can change all of that in the future in one larger PR address all the `write_to` impls.

ConnorGray: I see, in that case then let's leave this as it is (calling `Representable`) and be "consistently inconsistent" for now, while we're in the middle of that broader refactoring that you mention 🙂

PR #5695, opened 2025-12-20
[stdlib] Make Set conform to Writable, Stringable and Representable

msaelices:
`Set` comforms to `Writable`, `Stringable`, and `Representable` traits.

- Updates struct declaration to include trait conformance
- Removes generic constraints from methods
- Uses `_constrained_conforms_to` for compile-time checking
- Removes obsolete repr overload from builtin/repr.mojo


JoeLoser:
!sync

PR #5702, opened 2025-12-22
[stdlib] Add `FastMathFlag` to `ComplexSIMD`

martinvuyk:
Add `FastMathFlag` to `ComplexSIMD`


barcharcraz:
What does passing through the flag even do today? I don't think it propagates to any expressions in the function arguments, does the fma pop instruction expand to a mult and and add if you don't pass the flag?

Maybe we could just remove the flag from the `fma` method and _always_ generate pop ops with it.

Maybe we could even remove it from the pop op in the first place

PR #5703, opened 2025-12-22
[stdlib] add time unit us (Microseconds) to benchmark Unit

jmikaelr:
Adds microseconds (us) as a supported time unit in std/benchmark’s Unit and updates the divisor mapping accordingly.

Motivation: benchmark results often fall between ns and ms. Supporting us makes reports easier to read without manual conversion.


jmikaelr:

abduld:
!sync

PR #5711, opened 2025-12-27
[stdlib] Generalize Index factory function to use variadic parameters

msaelices:
Replace specific 1/2/3/4/5-D Index factory functions with a single generic function using variadic type parameters `[*Ts: Intable]`. This works for any number of dimensions and eliminates ~160 lines of repetitive code.

Add tests for 1-D, 2-D, 4-D, 5-D, and 6-D indices (6-D drives the variadic implementation).


mojo/stdlib/std/utils/index.mojo
```diff
@@ -755,179 +755,28 @@ struct IndexList[size: Int, *, element_type: DType = DType.int64](
 # ===-----------------------------------------------------------------------===#
 @always_inline
 fn Index[
-    T0: Intable, //, *, dtype: DType = DType.int64
-](x: T0, out result: IndexList[1, element_type=dtype]):
-    """Constructs a 1-D Index from the given value.
+    *Ts: Intable, dtype: DType = DType.int64
+](*args: *Ts,) -> IndexList[args.__len__(), element_type=dtype]:
```
abduld:
```suggestion
](*args: *Ts) -> IndexList[args.__len__(), element_type=dtype]:
```

mojo/stdlib/std/utils/index.mojo
```diff
@@ -755,179 +755,28 @@ struct IndexList[size: Int, *, element_type: DType = DType.int64](
...
 
     Args:
-        x: The initial value.
+        args: The initial values (variable number of arguments).
 
     Returns:
         The constructed IndexList.
     """
-    return {Int(x)}
+    constrained[args.__len__() > 0, "At least one argument is required"]()
```
abduld: can this be a where statement? Also, does it make sense to have this incoded in the type something like


```suggestion

fn Index[
    T: Intable, *Ts: Intable, dtype: DType = DType.int64
](arg0 : T, *args: *Ts) -> IndexList[args.__len__() + 1, element_type=dtype]:
```

that way you do not need this constraint

msaelices: I tried using `where` but it did not work. Probably is not featured enough for `args.__len__()` expressions coming from variadics.

However, encoding the type using `arg0` to get rid of the constraint is actually a great idea.  Will do

msaelices: Done: https://github.com/msaelices/modular/commit/3977aa8f21f4301d1de285975369751688f66e0c

abduld:
!sync

PR #5714, opened 2025-12-28
[stdlib] Add `filter_types` to variadic

saviorand:
A few convenience functions to make it easier to work with variadics


saviorand:

soraros:
@lsh Idea: can we make them conditional methods on `Variadic` itself (so chaining works etc)?

mojo/stdlib/test/builtin/test_variadic.mojo
```diff
@@ -305,6 +305,55 @@ def test_map_types_to_types():
     assert_true(_type_is_eq[variadic[1], String]())
```
JoeLoser, line 307: **Suggestion** Can you add a test for filtering to an empty result (a predicate that matches nothing) to ensure that edge case is handled correctly.

JoeLoser (commented):
Thanks for this contribution, @saviorand! Filtering types is a useful ergonomic improvement for working with variadics.   Before merging, I think we should resolve @soraros's https://github.com/modular/modular/pull/5714#issuecomment-3716941897 about whether this should be a conditional method on Variadic itself. The current API:

```mojo
  comptime step2 = Variadic.filter_types[*step1, predicate=IsNotInt]
```

vs potentially:
```mojo
  comptime step2 = step1.filter[predicate=IsNotInt]
```

@lsh - could you weigh in on the preferred API direction?

Overall this is solid work - just want to nail down the API design before merging. Thanks again for the contribution!

lsh:
@soraros I think that would be really nice, but there are some tradeoffs to consider. I would expect an API kind of like

```mojo
VariadicIter[*elts].map[mapper].filter[pred].result
```
though this means that every use now needs to explicitly pull the elements out at the end. I think another important distinction is that this API would not be lazily evaluated. I'm still open to exploring that API though!

soraros:
@lsh Yea, and we can of course land and iterate.

saviorand:
@JoeLoser @soraros thanks for the feedback! I definitely like the idea.
@lsh I guess this affects other functions in this file then, not just the filter? E.g will `map_types_to_types` also have to become a `map` method like this etc., for everything to stay consistent?
To keep this PR small I'd definitely like to focus on `filter` only, but this would mean we'll have two styles of API in the file, at least for a short while

lsh:
I'd say we can land this and write the abstraction after. I have an idea I'll try out

lsh:
!sync

ConnorGray:
This is super cool, nice work!

lsh:
!sync

PR #5715, opened 2025-12-29
[stdlib] feat: Add unsafe_ptr to StaticTuple

josiahls:
Needed for some cases of interfacing with ffi libs such as ffmpeg.

### DIFF
--- a/mojo/stdlib/std/utils/static_tuple.mojo
+++ b/mojo/stdlib/std/utils/static_tuple.mojo
@@ -229,6 +229,43 @@ struct StaticTuple[element_type: __TypeOfAllTypes, size: Int](
         )
         return UnsafePointer[origin = origin_of(self)](ptr)[]
 
+    @always_inline
+    fn unsafe_ptr[
+        origin: Origin, address_space: AddressSpace, //
+    ](ref [origin, address_space]self) -> UnsafePointer[
+        Self.element_type,
+        origin,
+        address_space=address_space,
+    ]:
+        """Returns an unsafe pointer to the tuple's underlying memory.
+
+        The returned pointer is valid only while the tuple exists and has not
+        been moved. Accessing elements beyond `len(self) - 1` is undefined
+        behavior.
+
+        This is useful for FFI interop where C functions expect array pointers.
+
+        Parameters:
+            origin: The origin of the StaticTuple.
+            address_space: The address space of the StaticTuple.
+
+        Returns:
+            An `UnsafePointer` to the underlying memory, with origin tied to `self`.
+
+        Example:
+            ```mojo
+            var t = StaticTuple[Int, 3](1, 2, 3)
+            var ptr = t.unsafe_ptr()
+            # Pass ptr to a C function expecting int*
+            ```
+        """
+        return (
+            UnsafePointer(to=self._mlir_value)
+            .bitcast[Self.element_type]()
+            .unsafe_origin_cast[origin]()
+            .address_space_cast[address_space]()
+        )
+
     @always_inline("nodebug")
     fn _replace[idx: Int](self, val: Self.element_type) -> Self:
         """Replaces the value at the specified index.
--- a/mojo/stdlib/test/utils/test_static_tuple.mojo
+++ b/mojo/stdlib/test/utils/test_static_tuple.mojo
@@ -51,5 +51,36 @@ def test_setitem():
     assert_equal(t[0], 400)
 
 
+def test_get_unsafe_ptr():
+    var t = StaticTuple[Int, 3](1, 2, 3)
+    var ptr: UnsafePointer[Int, ImmutAnyOrigin] = t.unsafe_ptr()
+    assert_equal(ptr[0], 1)
+    assert_equal(ptr[1], 2)
+    assert_equal(ptr[2], 3)
+
+
+def test_unsafe_ptr_mutable():
+    """Test writing through the pointer."""
+    var t = StaticTuple[Int, 3](1, 2, 3)
+    var ptr = t.unsafe_ptr()
+    ptr[1] = 42
+    assert_equal(t[1], 42)
+
+
+def test_unsafe_ptr_single_element():
+    """Edge case: single element tuple."""
+    var t = StaticTuple[Int, 1](99)
+    var ptr = t.unsafe_ptr()
+    assert_equal(ptr[0], 99)
+
+
+def test_unsafe_ptr_different_types():
+    """Test with non-Int types."""
+    var t = StaticTuple[Float64, 2](1.5, 2.5)
+    var ptr = t.unsafe_ptr()
+    assert_equal(ptr[0], 1.5)
+    assert_equal(ptr[1], 2.5)
+
+
 def main():
     TestSuite.discover_tests[__functions_in_module()]().run()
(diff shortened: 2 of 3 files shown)


mojo/stdlib/std/utils/static_tuple.mojo
```diff
@@ -229,6 +229,21 @@ struct StaticTuple[element_type: AnyTrivialRegType, size: Int](
         )
         return UnsafePointer[origin = origin_of(self)](ptr)[]
 
+    @always_inline("nodebug")
+    fn unsafe_ptr(
```
JoeLoser: **Question**: Why not implement `unsafe_ptr` using the same `pop.array.gep` pattern with index 0? This would be more consistent:

```mojo
  fn unsafe_ptr(ref self) -> UnsafePointer[Self.element_type, origin_of(self)]:
      var ptr = __mlir_op.`pop.array.gep`(
          UnsafePointer(to=self._mlir_value).address, Int(0)._mlir_value
      )
      return UnsafePointer[origin = origin_of(self)](ptr)
```

Or alternatively, implement it in terms of `_unsafe_ref`:

```mojo
  fn unsafe_ptr(ref self) -> UnsafePointer[Self.element_type, origin_of(self)]:
      return UnsafePointer(to=self._unsafe_ref(0))
```

The bitcast approach works but relies on the assumption that the array's memory layout starts at the first element, while gep is semantically explicit about getting element 0.

mojo/stdlib/test/utils/test_static_tuple.mojo
JoeLoser, line 1: Suggested additional tests:

```mojo
  def test_unsafe_ptr_mutable():
      """Test writing through the pointer."""
      var t = StaticTuple[Int, 3](1, 2, 3)
      var ptr = t.unsafe_ptr()
      ptr[1] = 42
      assert_equal(t[1], 42)

  def test_unsafe_ptr_single_element():
      """Edge case: single element tuple."""
      var t = StaticTuple[Int, 1](99)
      var ptr = t.unsafe_ptr()
      assert_equal(ptr[0], 99)

  def test_unsafe_ptr_different_types():
      """Test with non-Int types."""
      var t = StaticTuple[Float64, 2](1.5, 2.5)
      var ptr = t.unsafe_ptr()
      assert_equal(ptr[0], 1.5)
      assert_equal(ptr[1], 2.5)
```

mojo/stdlib/std/utils/static_tuple.mojo
```diff
@@ -229,6 +229,21 @@ struct StaticTuple[element_type: AnyTrivialRegType, size: Int](
...
+        ref self,
+    ) -> UnsafePointer[Self.element_type, origin_of(self)]:
+        """Returns a pointer to the static tuple.
```
JoeLoser: **Suggestion**

```
"""Returns an unsafe pointer to the tuple's first element.

  The returned pointer is valid only while the tuple exists and has not
  been moved. Accessing elements beyond `len(self) - 1` is undefined
  behavior.

  This is useful for FFI interop where C functions expect array pointers.

  Returns:
      An `UnsafePointer` to the first element, with origin tied to `self`.

  Example:
      ```mojo
      var t = StaticTuple[Int, 3](1, 2, 3)
      var ptr = t.unsafe_ptr()
      # Pass ptr to a C function expecting int*
      ```
  """
```

JoeLoser (commented):
The feature is good, but the test should verify mutable access works (this is the primary FFI use case), and the docstring should warn about safety constraints. These are small fixes that would significantly improve the PR.  PTAL at my comments and ping me when they are addressed and we can land this!

mojo/stdlib/std/utils/static_tuple.mojo
josiahls: Quick tests,
```mojo
        # var ptr = __mlir_op.`pop.array.gep`(
        #     UnsafePointer(to=self._mlir_value).address, Int(0)._mlir_value
        # )
        # return UnsafePointer[origin = origin_of(self)](ptr)
        return (
            UnsafePointer(to=self._mlir_value)
            .bitcast[Self.element_type]()
            .unsafe_origin_cast[origin_of(self)]()
        )
```
Of the 3, bitcast passed all tests.

The other 2, `ptr[1] = 42` doesn't actually assign the value. 
```bash
      Error('At /home/mojo_user/ash_dynamics/third_party/modular/mojo/stdlib/test/utils/test_static_tuple.mojo:69:17: AssertionError: `left == right` comparison failed:
         left: 2
        right: 42')
```
Going to try to solve this in the next hour or 2 since `return UnsafePointer[origin = origin_of(self)](ptr)` would be preferable.

josiahls: @JoeLoser I guess I don't quite understand `pop.array.gep`. The only other usage of it is in `InlineArray`, but its only used in `unsafe_get`. `InlineArray` doesn't use it for `unsafe_ptr`, instead it does something similar to me:
```mojo
    @always_inline
    fn unsafe_ptr[
        origin: Origin, address_space: AddressSpace, //
    ](ref [origin, address_space]self) -> UnsafePointer[
        Self.ElementType,
        origin,
        address_space=address_space,
    ]:

        return (
            UnsafePointer(to=self._array)
            .bitcast[Self.ElementType]()
            .unsafe_origin_cast[origin]()
            .address_space_cast[address_space]()
        )

```

josiahls: Ok I give up. 
```mojo
@register_passable("trivial")
struct TestArray:
    var _mlir_value: __mlir_type[`!pop.array<3, index>`]

# ... minimal ops for this
```
Then, when I try assigning it, the change doesn't propagate
```mojo
fn test_gep_pointer_assignment():
    """Test assignment through a pointer created from gep at index 0, then use ptr[1]."""
    var arr = TestArray(1, 2, 3)
    print("\n--- Testing gep pointer assignment ---")
    print("Initial array:", arr[0], arr[1], arr[2])
    
    # Create pointer using gep at index 0 (starting point)
    var gep_result = __mlir_op.`pop.array.gep`(
        UnsafePointer(to=arr._mlir_value).address, Int(0)._mlir_value
    )
    
    # Create UnsafePointer using implicit constructor (same pattern as _unsafe_ref)
    var elem_ptr = UnsafePointer[origin = origin_of(arr)](gep_result)
    
    # Try to access through the pointer using pointer arithmetic
    var val1_ptr = elem_ptr + 1
    
    print("Pointer created at index 0, accessing ptr[1]")
    
    # Try to assign through the pointer at offset 1 (should modify arr[1])
    val1_ptr[] = Int(99)._mlir_value
    print("After assignment ptr[1] = 99")
    print("arr[0]:", arr[0], "arr[1]:", arr[1])
    print("Expected arr[1]: 99, Actual:", arr[1])
    
    if arr[1] == 99:
        print("✓ Assignment through gep pointer at offset 1 works!")
    else:
        print("✗ Assignment through gep pointer at offset 1 FAILED") # this prints

```

josiahls: If there are docs that talk about the usage of `pop.array.gep`, can you point me to them? I'm getting the feeling that an implicit copy is happening somewhere here either directly in the `pop.array.gep`  or somewhere else.

Issue #4937, opened 2025-06-30
[BUG] cosh does not work on NVIDIA L40S


Philipp413:
### Actual behavior

Calling cosh results in 
```
from gpu.host import DeviceContext,DeviceAttribute
^
/scratch/pkueppers/eloc.mojo:96:22: error: call expansion failed
    var result = cosh(5.0)
                     ^
open-source/max/mojo/stdlib/stdlib/math/math.mojo:1727:4: note: function instantiation failed
open-source/max/mojo/stdlib/stdlib/math/math.mojo:1743:30: note: call expansion failed
open-source/max/mojo/stdlib/stdlib/math/math.mojo:2580:4: note: function instantiation failed
open-source/max/mojo/stdlib/stdlib/math/math.mojo:2596:6: note: call expansion failed
open-source/max/mojo/stdlib/stdlib/builtin/constrained.mojo:58:6: note: constraint failed: the libm operation is not available on the CUDA target
mojo: error: failed to run the pass manager
```

Interestingly, the function `log` from the math module works. 

### Expected behavior

Return cosh(5)=74.2

### Steps to reproduce

At top of file, add `from math import cosh`

In GPU kernel, add `var result = cosh(5.0)`

### System information

Mojo 25.4.0

JoeLoser:
Thanks for filing the issue!  The docs team is clarifying that L40s are actually meant to be "Tier 2 support" which we do not run any internal CI on.  So this is supported on a best effort basis by others.  We will not actively prioritize fixing this, but I would happily welcome a PR from the community if someone is interested in making this work.

rajuptvs:
Hey,  I did a quick check and reproduced this on an L40s using:

`26.1.0.dev2025120907`

Running:

```bash
pixi run mojo test.mojo
```

produces:

```
/lustre/home/sairamaraju.penmatsa/mojo_exp/test.mojo:1:1: error: failed to run the pass manager for offload functions
from gpu.host import DeviceContext,DeviceAttribute
^
/lustre/home/sairamaraju.penmatsa/mojo_exp/test.mojo:11:4: error: function instantiation failed
fn print_threads():
   ^
/lustre/home/sairamaraju.penmatsa/mojo_exp/test.mojo:14:22: note: call expansion failed with parameter value(s): (..., "width": 1)
    var result = cosh(5.0)
                     ^
open-source/max/mojo/stdlib/stdlib/math/math.mojo:1990:4: note: function instantiation failed
[...]

PR #5719, opened 2025-12-31
[Stdlib][GPU] Add GPU support for cosh and sinh math functions

BlackWingedKing:
Addresses https://github.com/modular/modular/issues/4937#issuecomment-3634552867

This PR adds GPU support for the `cosh` and `sinh` functions in the math module.
Previously, calling these functions on GPU targets resulted in a compile error:
"the libm operation is not available on the CUDA target".

**Implementation:**
- Added GPU-specific implementations using the exponential identity:
  - `cosh(x) = (exp(x) + exp(-x)) * 0.5`
  - `sinh(x) = (exp(x) - exp(-x)) * 0.5`
- Uses `@parameter if is_gpu()` to dispatch to the identity-based implementation
  on GPU targets while preserving the existing libm-based implementation for CPU.

**Testing:**
- Added `cosh` and `sinh` to the GPU math function tests in `test_math.mojo`
- Added numerical accuracy tests for `float32`, `float16`, and `bfloat16` types
  in `test_math_accuracy.mojo`


BlackWingedKing:

abduld:
!sync

PR #5747, opened 2026-01-08
[stdlib] Fix complex division

### LINKED ISSUE #5742: [BUG] Incorrect imaginary part in Complex numbers division
### Actual behavior
when you use the complex number mojo implementation and try to do a division, the current implementation give a flipped sign. 
https://github.com/modular/modular/blob/a632e3a11f6efa3bf20c94b9e08d1010244f7dc5/mojo/stdlib/std/complex/complex.mojo#L277

**The current result:** 
> (1+i) / (1-i) =  0.0  +  -1.0 i

**Expected:** 
> 0.0 + 1.0i 

### Expected behavior

in principle the modular/mojo/stdlib/std/complex/complex.mojo implementation should be:

```
fn __truediv__(self, rhs: Self) -> Self:
    var denom = rhs.squared_norm()
    return Self(
        self.re.fma(rhs.re, self.im * rhs.im) / denom,
        self.im.fma(rhs.re, self.re * rhs.im) / denom,  // <<<< Fixed line
    )
```

Expected result: 
> "0 + 1i" or "I"

### Steps to reproduce

using : https://github.com/modular/modular/blob/a632e3a11f6efa3bf20c94b9e08d1010244f7dc5/mojo/stdlib/std/complex/complex.mojo#L277
 
run this:
 ```
from complex import ComplexFloat32

alias Complex = ComplexFloat32

fn main():
    # Test: (1+i) / (1-i) should equal I
    var num = Complex(1.0, 1.0)
    var den = Complex(1.0, -1.0)
    var result = num / den

    print("(1+i) / (1-i) = ", result.re, " + ", result.im, "I")
    print("Expected: 0 + 1i")

```

### System information

       Pixi version: 0.49.0
       Platform: osx-arm64

### PR DESCRIPTION
**Pull request by bgreni**
Fixes #5742

### DIFF
--- a/mojo/stdlib/std/complex/complex.mojo
+++ b/mojo/stdlib/std/complex/complex.mojo
@@ -274,7 +274,7 @@ struct ComplexSIMD[dtype: DType, size: Int](
         var denom = rhs.squared_norm()
         return Self(
             self.re.fma(rhs.re, self.im * rhs.im) / denom,
-            self.re.fma(rhs.im, -self.im * rhs.re) / denom,
+            self.im.fma(rhs.re, -(self.re * rhs.im)) / denom,
         )
 
     @always_inline
--- a/mojo/stdlib/test/complex/test_complex.mojo
+++ b/mojo/stdlib/test/complex/test_complex.mojo
@@ -69,7 +69,11 @@ def test_math():
 
     var u = x / y
     assert_almost_equal(u.re, 0.44)
-    assert_almost_equal(u.im, -0.08)
+    assert_almost_equal(u.im, 0.08)
+
+    var t = ComplexFloat32(1.0, 1.0) / ComplexFloat32(1.0, -1.0)
+    assert_almost_equal(t.re, 0)
+    assert_almost_equal(t.im, 1)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: mojo/stdlib/std/complex/complex.mojo [hunk 1]
```diff
@@ -274,7 +274,7 @@ struct ComplexSIMD[dtype: DType, size: Int](
         var denom = rhs.squared_norm()
         return Self(
             self.re.fma(rhs.re, self.im * rhs.im) / denom,
-            self.re.fma(rhs.im, -self.im * rhs.re) / denom,
+            self.im.fma(rhs.re, -(self.re * rhs.im)) / denom,
```
abduld, line 277: isn't this the same operation mathematically?

bgreni: Not quite, the correct formula is `(self.im*rhs.re)-(self.real*rhs.im)` and the original is doing `(self.re*rhs.im)-(self.im*rhs.re)` so they negate each other.

abduld:
!sync

PR #5749, opened 2026-01-08
[stdlib] Fix usage of string length when byte length is required

martinvuyk:
Fix usage of string length when byte length is required.

@NathanSWard in https://github.com/modular/modular/commit/a4ffad780fc19c23b19b03d966cc89f3b0ff3528 you introduced a lot of usage of `String.__len__` which I've been trying for the last year to remove everywhere except where unicode length is required to pave the way for the switch to codepoint indexing. There are also a lot of places where byte slicing processing was changed to string slicing which will also take a performance hit once the switch happens now, but it's in unimportant places like the path library and outside the scope of this PR.

@JoeLoser Could we please officialize that we will switch to codepoint indexing? I see way too many bugs like this every single time and I'm constantly fixing them. Or let's just straight up deprecate `String.__len__` and be done with the debate of what it should return in favor of explicit functions (currently `byte_length` and `char_length` in `StringSlice`), what do you guys think?

### DIFF
--- a/max/kernels/test/gpu/basics/test_amd_format.mojo
+++ b/max/kernels/test/gpu/basics/test_amd_format.mojo
@@ -28,7 +28,7 @@ struct Buffer[capacity: Int](Defaultable, Writer):
         self.pos = 0
 
     fn write_string(mut self, string: StringSlice):
-        len_bytes = len(string)
+        len_bytes = string.byte_length()
         # If empty then return
         if len_bytes == 0:
             return
--- a/mojo/stdlib/std/builtin/debug_assert.mojo
+++ b/mojo/stdlib/std/builtin/debug_assert.mojo
@@ -161,7 +161,9 @@ fn debug_assert[
         message.nul_terminate()
 
         var slice = message.as_string_slice()
-        _debug_assert_msg(slice.unsafe_ptr(), len(slice), __call_location())
+        _debug_assert_msg(
+            slice.unsafe_ptr(), slice.byte_length(), __call_location()
+        )
 
 
 @always_inline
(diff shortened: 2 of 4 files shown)


martinvuyk:
If we go the deprecation route I would also like to rename things like the `unsafe_uninit_length` ctor to `unsafe_uninit_byte_length`, `char_length` to `codepoint_length`, and so on

NathanSWard:
>  lot of usage of String.__len__ which I've been trying for the last year to remove everywhere except where unicode length is required to pave the way for the switch to codepoint indexing.

Yes this is true, however the current behavior of `String.__len__` returns the byte length of the string.
We actually have not officially decided on if/what `String.__len__` should return or if we should even have it. 
And ditto with codepoint indexing. @barcharcraz is working on a larger `String` redesign to make these sort of operations explicit. E.g. `string[byte=42]` and `string[char=42]` or something along those lines.

> byte slicing processing was changed to string slicing

The current implementation uses byte slicing (which can ofc cause invalid UTF-8 and needs to be fixed❗) but those changes are out of scope of that PR.
Once we finalize on a API for string indexing/slicing etc.. we can make sure to audit usages of them across the codebase and fix them accordingly 😄.

NathanSWard:
!sync

NathanSWard (approved):

NathanSWard:
!sync

NathanSWard:
Hmm the Copybara is not working (my guess is the merge conflict).
@martinvuyk are you able to rebase this PR so I can merge it in 😄? Thanks!

JoeLoser:
> Hmm the Copybara is not working (my guess is the merge conflict). @martinvuyk are you able to rebase this PR so I can merge it in 😄? Thanks!

Yep, if there's merge conflicts, the Copybara job internally will fail and you won't be able to sync the PR.

NathanSWard:
!sync

PR #5758, opened 2026-01-11
Remove MAX_THREADS_PER_BLOCK_METADATA annotations

w32zhong:
MAX_THREADS_PER_BLOCK_METADATA=256 causes many example kernels failed to compile on Nvidia platform because BM x BN > 256 (e.g.,  in coalescing_matrix_multiplication, BM x BN = 32 x 32)


dlangbe:
I think changing the hard-coded 256 (which is there for default performance behavior on both AMD/Nvidia) to `OPTIMIZED_NUM_THREADS` would be a better solution, as removing the metadata drastically effects performance.

w32zhong:
Sg. as a regular user, I don't know if metadata affects so much. But I am certain the old code does not complie.

thanks for looking into this

PR #5765, opened 2026-01-12
[sdtlib] Deprecate `String.__len__()`

martinvuyk:
Deprecate `String.__len__()`. Provide a new "builtin" `len(value: StringSlice) -> Int:` method with a deprecation warning. Remove the conformance to sized and the length methods from `String`, `StringSlice`, and `StringLiteral`


martinvuyk:
I think I've got everything in the stdlib.

@NathanSWard @barcharcraz I think I'll need you guys to pull this in and fix some internal MAX code. This is an error in the torch tests which aren't open sourced:
```
error: 'String' value has no attribute '__len__'
        MOGGPrimitives.mgp_assert(var_58, var_59.unsafe_ptr(), var_59.__len__())
                                                               ~~~~~~^
error: 'String' value has no attribute '__len__'
        MOGGPrimitives.mgp_assert(var_69, var_70.unsafe_ptr(), var_70.__len__())
                                                               ~~~~~~^
error: 'String' value has no attribute '__len__'
        MOGGPrimitives.mgp_assert(var_78, var_79.unsafe_ptr(), var_79.__len__())
                                                               ~~~~~~^
```

mojo/stdlib/std/builtin/string_literal.mojo
```diff
@@ -185,26 +184,14 @@ struct StringLiteral[value: __mlir_type.`!kgen.string`](
         """
         return PythonObject(self)
 
-    @always_inline("nodebug")
-    fn __len__(self) -> Int:
```
NathanSWard: I would recommend we mark this (and the other `StringSlice.__len__`, `String.__len__`) functions as `@deprecated` instead of just removing them for now. Since there is a chance that people are using them and this would make the transition easier 😄 
```
@deprecated("String.__len__ is being deprecated in favor of using `.byte_length()`")
```

martinvuyk: I tried but I couldn't because the conformance to `Sized` triggers the deprecation warning even though it isn't used in the stdlib. Not sure if it's because of the MAX code or just a "language bug". The workaround I found is adding a new "builtin" `fn len(StringSlice) -> Int` function that is marked as deprecated

barcharcraz (approved):
I'm fine with this. I don't really mind __len__ but given we decided to keep byte_length I don't mind removing it either. I doubt the fallout will be too bad in the rest of the codebase either.

Please rebase and add changelog entry; We'll fix the max/mogg stuff before merging.

martinvuyk:
@barcharcraz I already got everything in the public repo, only internal stuff remaining

barcharcraz:
!sync

barcharcraz:
!sync

barcharcraz:

barcharcraz:
!sync

barcharcraz:
I've merged this internally.

martinvuyk:
@barcharcraz I saw the commit was reverted, do we need to reopen this PR or is that something you guys handle internally?

PR #5766, opened 2026-01-12
[stdlib] [NFC] Use `_is_utf8_continuation_byte` instead of `_utf8_byte_type(b) == 1`

martinvuyk:
Use `_is_utf8_continuation_byte` instead of `_utf8_byte_type(b) == 1`.

This is slightly faster (micro-op level diff), and a bit easier to understand at a glance


ConnorGray:
!sync

PR #5768, opened 2026-01-13
[Kernels] LegacyUnsafePointer-to-UnsafePointer-migration

turakz:
## Summary
migrates scoped kernel files from `LegacyUnsafePointer` to `UnsafePointer`.

files migrated:
- `extensibility/tensor/managed_tensor_slice.mojo` (UPDATE: impacted by upstream changes, not migrated yet)
- `internal_utils/_measure.mojo`
- `internal_utils/_testing.mojo`
- `internal_utils/_utils.mojo`
- `kv_cache/types.mojo`

## Test Plan
- `./bazelw build //max/kernels/...` passes
- `./bazelw test //max/kernels/test/...` passes

### DIFF
--- a/max/kernels/src/internal_utils/_measure.mojo
+++ b/max/kernels/src/internal_utils/_measure.mojo
@@ -18,9 +18,6 @@ from sys import simd_width_of
 from algorithm import elementwise, mean, sum, vectorize
 from algorithm.functional import unswitch
 
-from memory import LegacyUnsafePointer
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
 from utils import IndexList
 
 # ===----------------------------------------------------------------------=== #
@@ -54,7 +51,7 @@ fn kl_div(
 fn kl_div[
     dtype: DType, //
 ](
-    output: UnsafePointer[Scalar[dtype]],
+    output: UnsafePointer[Scalar[dtype], MutAnyOrigin],
     x: type_of(output),
     y: type_of(output),
     len: Int,
@@ -77,7 +74,7 @@ fn kl_div[
 fn kl_div[
     dtype: DType, //, out_type: DType = DType.float64
 ](
-    x: UnsafePointer[Scalar[dtype]],
+    x: UnsafePointer[Scalar[dtype], ImmutAnyOrigin],
     y: type_of(x),
     len: Int,
 ) -> Scalar[
@@ -113,8 +110,8 @@ fn kl_div[
 fn correlation[
     dtype: DType, //, out_type: DType = dtype
 ](
-    u: UnsafePointer[Scalar[dtype]],
-    v: type_of(u),
+    u: UnsafePointer[Scalar[dtype], MutAnyOrigin],
+    v: UnsafePointer[Scalar[dtype], MutAnyOrigin],
     len: Int,
     *,
     w: OptionalReg[type_of(u)] = None,
@@ -137,7 +134,7 @@ fn correlation[
     var vmu = Scalar[out_type]()
     var w_val = type_of(u)()
     if w:
-        w_val = type_of(u).alloc(len)
+        w_val = alloc[Scalar[dtype]](len)
         _div(w_val, w.value(), _sum(w.value(), len), len)
     if centered:
         if w:
@@ -200,12 +197,10 @@ fn correlation[
 fn uncentered_unweighted_correlation[
     dtype: DType, //, out_type: DType = dtype
 ](
-    u: UnsafePointer[Scalar[dtype]],
+    u: UnsafePointer[Scalar[dtype], ImmutAnyOrigin],
     v: type_of(u),
     len: Int,
-) -> Scalar[
-    out_type
-]:
+) -> Scalar[out_type]:
     """Compute the uncentered and unweighted correlation
     distance between two 1-D arrays.
     Unlike `correlation` with arguments set, this does not raise.
@@ -236,7 +231,11 @@ fn uncentered_unweighted_correlation[
 fn cosine[
     dtype: DType,
     //,
-](u: UnsafePointer[Scalar[dtype]], v: type_of(u), len: Int,) -> Float64:
+](
+    u: UnsafePointer[Scalar[dtype], ImmutAnyOrigin],
+    v: type_of(u),
+    len: Int,
+) -> Float64:
     """Compute the Cosine distance between 1-D arrays.
 
     The Cosine distance between `u` and `v`, is defined as
@@ -259,7 +258,7 @@ fn relative_difference[
     dtype: DType,
     //,
 ](
-    output: LegacyUnsafePointer[mut=False, Scalar[dtype]],
+    output: UnsafePointer[Scalar[dtype], ImmutAnyOrigin],
     ref_out: type_of(output),
     len: Int,
 ) -> Float64:
@@ -289,7 +288,11 @@ fn relative_difference[
 
 fn _sqrt[
     dtype: DType, //
-](output: UnsafePointer[Scalar[dtype]], x: type_of(output), len: Int) raises:
+](
+    output: UnsafePointer[Scalar[dtype], MutAnyOrigin],
+    x: type_of(output),
+    len: Int,
+) raises:
     @parameter
     fn apply_fn[
         simd_width: Int, rank: Int, alignment: Int = 1
@@ -307,7 +310,7 @@ fn _sqrt[
 fn _mul[
     dtype: DType, //
 ](
-    output: UnsafePointer[Scalar[dtype]],
+    output: UnsafePointer[Scalar[dtype], MutAnyOrigin],
     x: type_of(output),
     y: type_of(output),
     len: Int,
@@ -330,7 +333,7 @@ fn _mul[
 fn _div[
     dtype: DType, //
 ](
-    output: UnsafePointer[Scalar[dtype]],
+    output: UnsafePointer[Scalar[dtype], MutAnyOrigin],
     x: type_of(output),
     c: Scalar[dtype],
     len: Int,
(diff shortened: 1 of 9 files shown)


turakz:
note: looks like my initial migration of `managed_tensor_slice.mojo` was impacted by some upstream changes. currently looking into it.

max/kernels/src/internal_utils/_measure.mojo
```diff
@@ -18,9 +18,7 @@ from sys import simd_width_of
 from algorithm import elementwise, mean, sum, vectorize
 from algorithm.functional import unswitch
 
-from memory import LegacyUnsafePointer
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
+from memory import UnsafePointer, alloc
```
NathanSWard: You can remove the explicit `UnsafePointer` and `alloc` imports as they're automatically imported in the Mojo prelude :) 
```suggestion
```

max/kernels/src/kv_cache/types.mojo
```diff
@@ -32,12 +32,11 @@ from layout.tma_async import (
...
     RaggedTMA3DTile,
 )
-from memory import LegacyUnsafePointer
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
-comptime OpaquePointer = LegacyUnsafePointer[
-    mut=True, NoneType, origin=MutAnyOrigin
-]
+from memory import (
+    LegacyOpaquePointer as OpaquePointer,
```
NathanSWard: Are you able to also update the usages of `LegacyOpaquePointer` in this PR as well? If you want to have that be a follow-up that's okay too, but ideally if we can knock them both out together, that would be great!

max/kernels/src/internal_utils/_measure.mojo
```diff
@@ -111,7 +107,7 @@ fn kl_div[
 fn correlation[
     dtype: DType, //, out_type: DType = dtype
 ](
-    u: UnsafePointer[Scalar[dtype]],
+    u: UnsafePointer[Scalar[dtype], MutExternalOrigin],
```
NathanSWard: You generally should never take a `MutExternalOrigin` as an argument to a function unless that function plans on free the associated memory.
This can simply use `MutExternalOrigin` or `mut=True` and it should suffice.

max/kernels/src/internal_utils/_testing.mojo
```diff
@@ -88,7 +87,7 @@ fn assert_almost_equal[
     dtype: DType,
     //,
 ](
-    x: UnsafePointer[Scalar[dtype]],
+    x: UnsafePointer[Scalar[dtype], ...],
     y: type_of(x),
```
NathanSWard: This is not quite correct. By saying `type_of(x)` everything (including all of the parameters) need to be equal, however, the origins of the pointers _can_ be different.
To mimic this behavior but correctly account for origins, you can do something like this:
```
    x: UnsafePointer[Scalar[dtype]],
    y: UnsafePointer[Scalar[dtype]],
```
Simply specifying the type of `y` instead of using `type_of` should work here and allow for differing origins.

One last thing, is by using the `...` syntax, you are saying this function can accept and pointer with any address space. However, the previous behavior of the function only worked on `AddressSpace.GENERIC`. (Hence why de don't want to use the `...` syntax here).

max/kernels/src/internal_utils/_testing.mojo
```diff
@@ -144,6 +143,37 @@ fn assert_almost_equal[
         )
 
 
+# LegacyUnsafePointer overload for backward compatibility
```
NathanSWard: Is this overload actually needed?
Ideally you can pass a `LegacyUnsafePointer` to a function that accepts a `UnsafePointer` and it should "Just work". If not can you please provide me an example of the error message you are getting in case we need to fix some of the constructors. Ideally we should be able to delete this overload and only use the one that accepts a normal `UnsafePointer`.

max/kernels/src/internal_utils/_testing.mojo
```diff
@@ -154,7 +184,7 @@ fn assert_equal[
     dtype: DType,
     //,
 ](
-    x: UnsafePointer[Scalar[dtype]],
+    x: UnsafePointer[Scalar[dtype], ...],
     y: type_of(x),
```
NathanSWard: Ditto on the usage of `type_of` here not being completely correct.

max/kernels/src/internal_utils/_testing.mojo
```diff
@@ -307,3 +337,27 @@ fn pytorch_like_tolerances_for[dtype: DType]() -> Tuple[Float64, Float64]:
         return (1e-7, 1e-7)
     else:
         return (0.0, 0.0)
+
+
+fn assert_with_measure[
```
NathanSWard: What is the purpose of this new function/overload? It seems to be just redefining the other `assert_with_measure` function.

max/kernels/src/kv_cache/types.mojo
```diff
@@ -511,13 +510,15 @@ struct ContinuousBatchingKVCache[
         start_tok_idx: Int,
         head_idx: Int,
         head_dim_idx: Int = 0,
-    ) -> UnsafePointer[Scalar[Self.dtype]]:
+    ) -> UnsafePointer[Scalar[Self.dtype], MutExternalOrigin]:
```
NathanSWard: Generally you should never return a `MutExternalOrigin` from a struct's method. This can lead to accidentally "early freeing" of types as external origins don't extend lifetimes. Ideally we would use `origin_of(self)` here, but for simplicity's sake we can just use `MutAnyOrigin` instead.
```suggestion
    ) -> UnsafePointer[Scalar[Self.dtype], MutAnyOrigin]:
```

max/kernels/src/kv_cache/types.mojo
```diff
@@ -818,7 +819,7 @@ struct PagedKVCache[
...
```
NathanSWard: Ditto on not returning an external origin but instead a `MutAnyOrigin`.

NathanSWard (changes requested):
Thank you for taking on migrating these files! I greatly appreciate it 😄 
I left a round of review comments, please let me know if you need me to expand on any of them or if you have other questions/concerns. Hopefully the review comments make sense!

max/kernels/src/internal_utils/_testing.mojo
turakz: removing the overload and building the kernels target results in

```
ERROR .../BUILD.bazel:66:14: //max/kernels/test/nn:test_fused_qk_rope_ragged.mojo.test compiling mojo object failed: (Exit 1): ... test_fused_qk_rope_ragged.mojo:209:17: error: invalid call to 'assert_almost_equal': value passed to 'y' cannot be converted from 'LegacyUnsafePointer[Scalar[dtype], origin=q_buffer]' to 'UnsafePointer[Scalar[dtype], q_out_buffer]'
               assert_almost_equal(
               ^~~~~~~~~~~~~~~~~~~~
```

i had also run into an ambiguous call with 2 candidates `LegacyUnsafePointer` and `UnsafePointer` earlier in the dev cycle which prompted me to (re)consider the conversion.

turakz:
> Thank you for taking on migrating these files! I greatly appreciate it 😄 I left a round of review comments, please let me know if you need me to expand on any of them or if you have other questions/concerns. Hopefully the review comments make sense!


yo! thank you very much for taking the time, stoked on the feedback and your willingness to help me improve my understanding, stoked to contribute!

max/kernels/src/kv_cache/types.mojo
turakz: yep, let me get the PR stable with the current feedback integrated (so at worse if you decide we should just merge, i can always follow up with another PR), and then i'll segue.

max/kernels/src/internal_utils/_testing.mojo
turakz: i had ran into an ambiguous call with 2 candidates, but it could also have been a red herring at the time. looking into this as i work on PR feedback

max/kernels/src/internal_utils/_testing.mojo
turakz: update: this appears to have been a red herring -- ive partially integrated PR feedback on my branch and removed this overload, and kernels build/tests pass

max/kernels/src/internal_utils/_testing.mojo
turakz: is origin inferred here from the caller? i was under the impression origin would be needed to be explicitly passed, but this builds.

max/kernels/src/internal_utils/_measure.mojo
turakz: just to make sure i understand, you mean use `MutAnyOrigin` or `mut=True`?

max/kernels/src/internal_utils/_testing.mojo
turakz: this has been removed.

turakz:
pr feedback integrated (looks like there's a typo in the lint script `bazel` vs `bazelw` -- happy to include a revision, lmk), kernels build/tests pass locally.

looking into `LegacyOpaquePointer` in the meantime.

turakz:
`LegacyOpaquePointer` imports removed, relevant areas in scope files where applicable already make use of `MutOpaquePointer`. kernels build/tests pass locally. looking into `managed_tensor_slice.mojo`

turakz:
migrated `managed_tensor_slice.mojo` away from `LegacyUnsafePointer` and `LegacyOpaquePointer`. build/tests pass locally.

max/kernels/src/internal_utils/_testing.mojo
NathanSWard: Correct. So when you _don't_ specify a non-defaulted parameter for some type (in this case the `origin` parameter for `UnsafePointer`) that parameter will become automatically "unbound" and the compiler will infer it.

In short the compiler will turn this
```mojo
fn foo(p: UnsafePointer[Int]): ...
```
into this
```mojo
fn foo[_mut: Bool, _origin: Origin[mut]](p: UnsafePointer[Int, _origin]): ...
```

NathanSWard (approved):
Apologies for the delay on my half. But thank you for doing this 😄

NathanSWard:
!sync

max/kernels/src/internal_utils/_measure.mojo
NathanSWard: Correct!

turakz:
> Apologies for the delay on my half. But thank you for doing this 😄

no worries! thanks for taking the time, appreciate it 👊

PR #5778, opened 2026-01-15
[stdlib] Add a safer `String[bytes=...]` getter

### LINKED ISSUE #5402: [BUG] Slicing UTF-8 strings can return invalid results due to byte based indexing

martinvuyk:
Add a safer `String[bytes=...]` getter. This should eventually replace the current getter and deprecate it (I don't think it should be deleted right away because a lot of people index strings using slices). Closes #5402. Closes  #5405

### DIFF
--- a/mojo/stdlib/std/collections/string/string_slice.mojo
+++ b/mojo/stdlib/std/collections/string/string_slice.mojo
@@ -902,6 +902,24 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
         """
         return Self(unsafe_from_utf8=self._slice[span])
 
+    @always_inline
+    fn __getitem__(self, *, bytes: ContiguousSlice) -> Self:
+        """Gets the sequence of characters at the specified positions.
+
+        Args:
+            bytes: A slice that specifies positions of the new substring.
+
+        Returns:
+            A new StringSlice containing the substring at the specified positions.
+        """
+        var sliced = Self(unsafe_from_utf8=self._slice[bytes])
+
+        if not sliced.is_codepoint_boundary(0):
+            abort("incorrectly sliced a multi-byte sequence at the start")
+        elif not sliced.is_codepoint_boundary(UInt(sliced.byte_length())):
+            abort("incorrectly sliced a multi-byte sequence at the end")
+        return sliced
+
     fn to_python_object(var self) raises -> PythonObject:
         """Convert this value to a PythonObject.
 
--- a/mojo/stdlib/std/collections/string/string.mojo
+++ b/mojo/stdlib/std/collections/string/string.mojo
@@ -913,15 +913,20 @@ struct String(
         Returns:
             A new string containing the string at the specified positions.
         """
-        var start: Int
-        var end: Int
-        # TODO(#933): implement this for unicode when we support llvm intrinsic evaluation at compile time
+        return self.as_string_slice()[bytes=span]
 
-        start, end = span.indices(self.byte_length())
-        return StringSlice(
-            ptr=self.unsafe_ptr() + start,
-            length=end - start,
-        )
+    fn __getitem__(
+        self, *, bytes: ContiguousSlice
+    ) -> StringSlice[origin_of(self)]:
+        """Gets the sequence of characters at the specified positions.
+
+        Args:
+            bytes: A slice that specifies positions of the new substring.
+
+        Returns:
+            A new string containing the string at the specified positions.
+        """
+        return self.as_string_slice()[bytes=bytes]
 
     fn __eq__(self, rhs: String) -> Bool:
         """Compares two Strings if they have the same values.


mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -902,6 +902,39 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+            return Byte(_utf8_first_byte_sequence_length(b0)) == amnt
+
+        debug_assert(
+            len(span) == 0 or not _is_utf8_continuation_byte(span[0]),
+            "incorrectly sliced a multi-byte sequence at the start",
+        )
+        debug_assert(
+            _last_bytes_are_ok(),
+            "incorrectly sliced a multi-byte sequence at the end",
+        )
```
NathanSWard: I would argue this should `abort` instead of simply `debug_assert` if the byte span is invalid UTF-8.

We could have a `__getitem__[*, unsafe_bytes: ContiguousSlice]` (or something like this) but by default we should be aborting/raising an error.

martinvuyk: I didn't think about that. I could even go with raising a generic type and aborting when it's a `Never`, WDYT? Or we could just go with abort and revisit this once we've got some better effect system like what owen proposed

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -902,6 +902,39 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+        return Self(unsafe_from_utf8=span)
```
ConnorGray: **Question** Could we simplify this implementation using `StringSlice.is_codepoint_boundary()`? Something like:

```mojo
    fn __getitem__(self, *, bytes: ContiguousSlice) -> Self:
        """..."""

        if not self.is_codepoint_boundary(bytes.start):
            abort("incorrectly sliced a multi-byte codepoint sequence at the start")

        if not self.is_codepoint_boundary(bytes.end):
            abort("incorrectly sliced a multi-byte codepoint sequence at the end")

        return Self(unsafe_from_utf8=self._slice[bytes])
```

martinvuyk: maybe for the start, but I'm not a fan of the overhead that introduces compared to using the helper that is used inside that function anyway.

For the end of the sequence, the last byte being a continuation byte doesn't tell you if the amount of continuation bytes is the correct one for the last codepoint

ConnorGray: > the overhead that introduces compared to using the helper that is used inside that function anyway.

Could you elaborate this, I'm not sure what overhead this is referring to? Both your current implementation and `is_codepoint_boundary()` check the length, and use `_is_utf8_continuation_byte()`.  

> the last byte being a continuation byte doesn't tell you if the amount of continuation bytes is the correct one for the last codepoint

`StringSlice` is guaranteed to be valid UTF-8. If `bytes.end` is a valid codepoint boundary, that means that `bytes.end` is either A) 1 byte passed the end of the string, or B) is a valid "start" byte.

Is it necessary that indexing into a `StringSlice` defensively validates that the last codepoint has the correct encoded length?

martinvuyk: > > the overhead that introduces compared to using the helper that is used inside that function anyway.
> 
> Could you elaborate this, I'm not sure what overhead this is referring to? Both your current implementation and `is_codepoint_boundary()` check the length, and use `_is_utf8_continuation_byte()`.

```mojo
fn is_codepoint_boundary(self) -> Bool:
    if index >= UInt(len(self)): # <--- unnecessary branch
        return index == UInt(len(self))

    var byte = self.as_bytes()[index] # <--- index normalization
    # If this is not a continuation byte, then it must be a start byte.
    return not _is_utf8_continuation_byte(byte)
```
vs
```mojo
not _is_utf8_continuation_byte(span[0])
```
we could arguably even use `.unsafe_get(0)` since we've checked `len(span) == 0 or ...` beforehand


> > the last byte being a continuation byte doesn't tell you if the amount of continuation bytes is the correct one for the last codepoint
> 
> `StringSlice` is guaranteed to be valid UTF-8. If `bytes.end` is a valid codepoint boundary, that means that `bytes.end` is either A) 1 byte passed the end of the string, or B) is a valid "start" byte.
> 
> Is it necessary that indexing into a `StringSlice` defensively validates that the last codepoint has the correct encoded length?

As an example:


```mojo
comptime EVERY_CODEPOINT_LENGTH_STR = StringSlice("߷കൈ🔄!")
"""A string that contains at least one of 1-, 2-, 3-, and 4-byte UTF-8
sequences.

Visualized as:

┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃                      ߷കൈ🔄!                    ┃
┣━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━┫
┃   ߷  ┃     ക     ┃     ൈ    ┃       🔄      ┃! ┃
┣━━━━━━━╋━━━━━━━━━━━╋━━━━━━━━━━━╋━━━━━━━━━━━━━━━╋━━┫
┃ 2039  ┃   3349    ┃   3400    ┃    128260     ┃33┃
┣━━━┳━━━╋━━━┳━━━┳━━━╋━━━┳━━━┳━━━╋━━━┳━━━┳━━━┳━━━╋━━┫
┃223┃183┃224┃180┃149┃224┃181┃136┃240┃159┃148┃132┃33┃
┗━━━┻━━━┻━━━┻━━━┻━━━┻━━━┻━━━┻━━━┻━━━┻━━━┻━━━┻━━━┻━━┛
  0   1   2   3   4   5   6   7   8   9  10  11  12

For further visualization and analysis involving this sequence, see:
<https://connorgray.com/ephemera/project-log#2025-01-13>.
"""
```
I could index this by doing `EVERY_CODEPOINT_LENGTH_STR[bytes=2:6]` and using the logic of `is_codepoint_boundary` it would be considered valid even though the byte 6 is the first of 2 continuation bytes that complete the sequence

ConnorGray: Ah, I see what you're getting at. I'm sure due to how this code is factored we're duplicating certain checks, and we should avoid that to the extent we can while still making the function safe and guarantee UTF-8 validity.

We can certainly optimize `is_codepoint_boundary()` to perform only one in-bounds comparison and use `unsafe_get()` as you suggest to avoid the index normalization.

I'm happy to accept any suggestions that improve performance, as long as they maintain predictable safety guarantees 🙂 

> I could index this by doing EVERY_CODEPOINT_LENGTH_STR[bytes=2:6] and using the logic of is_codepoint_boundary it would be considered valid even though the byte 6 is the first of 2 continuation bytes that complete the sequence

I'm not sure that's correct.

This program prints `False`; index 6 does not satisfy `is_codepoint_boundary()` because it is not a UTF-8 start byte:

```mojo
fn main():
    var s = StringSlice("߷കൈ🔄!")

    print(s.is_codepoint_boundary(6)) # False
```

martinvuyk: Sorry for some reason I completely forgot that slicing in end-exclusive during this discussion. But yeah checking that the first byte and last byte aren't continuation bytes should suffice. I'll change it, thanks for insisting it will look way simpler and nicer :)

barcharcraz:
I have an internal version of this PR too (we had the same idea). Sadly the linting ci is not going to pass because mblack can't parse subscripts with named slice parameters.

martinvuyk:
@barcharcraz sorry to step on your toes. I knew you made the byte one so I thought I could spare you the work since you guys have other stuff to do. And this is something I've been really trying to push forward for a while now. Do whatever you want with this PR

Issue #5779, opened 2026-01-15
[Feature Request] [stdlib] Add codepoint based String indexing and slicing


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

I'm creating this as a parent issue to keep track of the initiative. I'll tackle this myself.

The goal is to have a keyword-only string getter for both single-codepoint indexing and full codepoint based contiguous slicing. Stepped slicing is a possibility but I'll leave it for later.

### What is your motivation for this change?

Safety and ergonomics

### Any other details?

CC: @thatstoasty I think you'll like this :)

JoeLoser:
This is done now thanks to your work, Martin! See modular/modular#6552 and modular/modular#5780

PR #5780, opened 2026-01-15
[stdlib] Add codepoint indexing

martinvuyk:
Add codepoint indexing. Part of #5779


mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -902,6 +903,53 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+            codepoint, self.byte_length()
+        )
+        # NOTE: Edge case: when the element at idx 0 in a string is fetched and
+        # it's a multi-byte sequence, the code would assume it's an ascii
+        # sequence. Fetch 1 more byte (when not OOB) just to have a continuation
+        # byte and force the code to go into the clipped branch
+        var i0_multi_case = Int(c_idx == 0 and self.byte_length() > 1)
+        var span = self._slice[: c_idx + 1 + i0_multi_case]
+        var c_count = len(span) - _count_utf8_continuation_bytes(span)
+        if likely(c_count == len(span)):  # ASCII
```
NathanSWard, line 507: I'm still not a big fan of `likely` and `unlikely` usage unless you have measurable perf benchmarks that this actually produces a faster runtime or better codegen.

There are many cases where introducing `(un)likely` will actually cause a perf regression as it can trip up the llvm optimizer.

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -902,6 +903,53 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+            var b_idx = c_idx + 1
+            while _is_utf8_continuation_byte(self._slice[b_idx]):
+                b_idx += 1
+            for s in Self(
+                unsafe_from_utf8=self._slice[b_idx:]
+            ).codepoint_slices():
+                if c_count == c_idx:
+                    return s
+                c_count += 1
+            debug_assert(False, "codepoint index is out of bounds")
```
NathanSWard: Ideally we should `abort` here instead.
Also can you add the passed in `codepoint` index that was used as part of this error message?

mojo/stdlib/test/collections/string/test_string_slice.mojo
```diff
@@ -1074,5 +1074,18 @@ def test_merge():
     _ = cond(True, a, b)
 
 
+def test_codepoint_indexing():
```
NathanSWard: I see that the implementation also support index normalization. Are you able to add a few tests for this, e.g. `[codepoint=-1]` for example.

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -902,6 +903,53 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+            codepoint: The codepoint index.
+
+        Returns:
+            A `StringSlice` view containing the unicode codepoint at the
+            specified position.
+        """
+
+        var c_idx = normalize_index["StringSlice"](
+            codepoint, self.byte_length()
+        )
```
martinvuyk: If we want to support negative indexing this will have to use `char_length()` (curently renaming it to `count_codepoints()` in another PR), which potentially doubles the execution time for this.

If we won't support negative indexing, what would be the correct way to make it explicit? Should we just add `debug_assert(Int(codepoint) >= 0)` ?

If we go for supporting negative indexing, we could take advantage of the fact that when `self.count_codepoints() == self.byte_length()` -> the string is pure ASCII

NathanSWard:
Thank you for jumping right onto this and implementing it 😄 🥳

mojo/stdlib/std/collections/string/string_slice.mojo
martinvuyk: @NathanSWard we commented at the same time, we can bring the discussion about negative indexing here

mojo/stdlib/std/collections/string/string_slice.mojo
martinvuyk: In my experience it does a very good job of making the happy-path more performant when there are several branches that do very different things. I think that most cases where it does what you say it's because the branches could be optimized by some sort of "logic coalescing"/pipelining. In this case I avoid memory traversal entirely, so I think it will improve things. I'll add benchmarks later when I have some time, either in this PR or another one and I'll be able to check my intuition

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -902,6 +903,53 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+        # NOTE: Edge case: when the element at idx 0 in a string is fetched and
+        # it's a multi-byte sequence, the code would assume it's an ascii
+        # sequence. Fetch 1 more byte (when not OOB) just to have a continuation
+        # byte and force the code to go into the clipped branch
+        var i0_multi_case = Int(c_idx == 0 and self.byte_length() > 1)
+        var span = self._slice[: c_idx + 1 + i0_multi_case]
+        var c_count = len(span) - _count_utf8_continuation_bytes(span)
+        if likely(c_count == len(span)):  # ASCII
+            return Self(ptr=self._slice.unsafe_ptr() + c_idx, length=1)
+        elif c_count == c_idx + 1:  # clipped the multi-byte sequence
```
barcharcraz, line 509: can this happen if the sequence before the codepoint index contains a lead byte?

While there are currently some ways to form StringSlices with invalid UTF-8 going forward we want UTF-8 validity to be a precondition of string slices

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -902,6 +903,53 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+        var c_idx = normalize_index["StringSlice"](
+            codepoint, self.byte_length()
+        )
+        # NOTE: Edge case: when the element at idx 0 in a string is fetched and
+        # it's a multi-byte sequence, the code would assume it's an ascii
+        # sequence. Fetch 1 more byte (when not OOB) just to have a continuation
+        # byte and force the code to go into the clipped branch
+        var i0_multi_case = Int(c_idx == 0 and self.byte_length() > 1)
+        var span = self._slice[: c_idx + 1 + i0_multi_case]
+        var c_count = len(span) - _count_utf8_continuation_bytes(span)
```
barcharcraz, line 506: dose this do a scan of the string preceding `code point`? If so I'm not sure why we're bothering? If we need to scan the string anyway we may as well just write:
```mojo
var it = CodepointSliceIter(self)
for i in range(codepoint):
    try: it.__next__()
    except: debug_assert("oob")
return it.next().value()
```

barcharcraz (changes requested):
I don't see the point of optimizing this like you did. I think just doing the spiritual equivalent of:
```self.codepoints().nth(codepoint)```

could be just as good.

It seems like with this implementation we have to scan the whole string anyway.

Actually, I would rather add a method to seek the codepoint iterator anyway, since that doesn't encourage slow code.

The general plan for strings is to keep indexing in terms of bytes but panic if those bytes don't fall on codepoint boundaries.

mojo/stdlib/std/collections/string/string_slice.mojo
martinvuyk: Yes if you're initially slicing the span at an arbitrary byte offset. The tests I added expose this case

mojo/stdlib/std/collections/string/string_slice.mojo
martinvuyk: I'll reply in a more thorough comment to your points above

martinvuyk:
> I don't see the point of optimizing this like you did. I think just doing the spiritual equivalent of: `self.codepoints().nth(codepoint)`
> 
> could be just as good.

Could be is the key here, and I respectfully disagree. `_count_utf8_continuation_bytes()` is vectorized, while the iterator has a data dependency between each consecutive byte. The approach here is to traverse the preceding string to check if the bytes are ascii and if not take the (I think) slower path of iterating per element until the given codepoint is reached.

My original idea was to use the continuation byte count to slice the span again and count the continuation bytes there progressively until the given codepoint is reached. But for simplicity's sake of the initial implementation I went for using the iterator after the first slice and cont byte count. We can always iterate on this and benchmark the different approaches to make sure what the best one is.

> It seems like with this implementation we have to scan the whole string anyway.

Either implementation requires it. The iterator scans the first byte but skips unpredictable amounts of offsets in memory based on that, so it introduces a data dependency. The count bytes (or potentially faster checking if all bytes are ascii) are better approaches IMO. We can benchmark this later on.

> Actually, I would rather add a method to seek the codepoint iterator anyway, since that doesn't encourage slow code.

Let's wait until we have actual performance results to draw conclusions

> The general plan for strings is to keep indexing in terms of bytes but panic if those bytes don't fall on codepoint boundaries.

I think we could support both raising methods and an unsafe variant that panics (I'm  not particularly fond of panicking because it's unrecoverable), WDYT? We could also just panic an wait until we have a better configurable effect system to let the end-developer determine how to deal with these errors

barcharcraz:
> I think we could support both raising methods and an unsafe variant that panics (I'm not particularly fond of panicking because it's unrecoverable), WDYT? We could also just panic an wait until we have a better configurable effect system to let the end-developer determine how to deal with these errors

I strongly think we should panic. Being on a code point boundary is a precondition of an indexing function that produces a codepoint and the index will almost always be derived in a way that ensures that. Instead of handling the error users should ask first if the index is valid. We should also provide methods to get the closest codepoint to a byte index.

We don't really need an unsafe version because you can use `.as_bytes[...]` along with `StringSlice`'s `unsafe_from_utf8` constructor.

I did try and come up with an index type that ensures it's derived in a way that keeps things valid but I realized that that type is actually just `StringSlice` :).

barcharcraz:
>Could be is the key here, and I respectfully disagree. _count_utf8_continuation_bytes() is vectorized, while the iterator has a data dependency between each consecutive byte. The approach here is to traverse the preceding string to check if the bytes are ascii and if not take the (I think) slower path of iterating per element until the given codepoint is reached.

Right, I think it's reasonable to have a specialized path for advancing the iterator multiple characters and/or advancing without actually retrieving the current codepoint. However we should just _do that_ instead of having different code for figuring out if the start of the string is ascii vs actually advancing through the remaining codepoints

martinvuyk:
> However we should just _do that_ instead of having different code for figuring out if the start of the string is ascii vs actually advancing through the remaining codepoints

Most strings on the internet are ascii or latin-1 encoded, so I think we should optimize for those cases as the default. And checking if all bytes are ascii or roughly how many bytes to the right to keep reading I think will be faster. But we'll have to benchmark and see

barcharcraz:
But the check we need to do is *exactly the same* as what we need in order to advance over multiple characters to the correct location.<br><br>We could actually make advancing and returning copy the span, then iterate forward until the next lead (or the end of the stirng) then return the difference. This might eliminate the loop-carried data dependency too, since although such a dependency would exist between the values returned each iteration those values would be unused when advancing multiple characters, so after DCE the dependency might go away.

martinvuyk:
> But the check we need to do is _exactly the same_ as what we need in order to advance over multiple characters to the correct location.
> 
> We could actually make advancing and returning copy the span, then iterate forward until the next lead (or the end of the stirng) then return the difference. This might eliminate the loop-carried data dependency too, since although such a dependency would exist between the values returned each iteration those values would be unused when advancing multiple characters, so after DCE the dependency might go away.

The iterator requires checking what the `_utf8_byte_sequence_length()` is for the first byte and it skips that amount in memory and then reads the next first byte, and so on. Counting continuation bytes (or just seeing if `byte <= 0x7F` to check for ascii) will load a full vector and do a `reduce_add()` or `reduce_or()` (checking if ascii) in a vectorized manner. I don't think that doing unpredictable jumps in memory will be faster for big strings than a properly vectorized approach with a much higher throughput.

It could be the case that the vectorized approach is slower than iteration for very small jumps in memory (small strings, or strings with very few non-ascii characters) such that the latency of the vector ops lose to the pipelineability of the iterator, so the best approach might be a hybrid (e.g. a branch checking if `idx < simd_width_of[UInt8]()`).

But anyway, this is all theory until we can prove it with the string benchmarking we've got

NathanSWard:
I'd prefer, from a maintainability and reviewability standpoint for new PRs like this, starting with a simpler implementation `self.codepoints().nth(codepoint)`, even if it isn’t the most performant version upfront. A straightforward approach makes the intent clearer, reduces review complexity, and helps us get the functionality merged sooner. This is a really useful addition, and I'd hate to see it blocked on discussions around complexity or performance characteristics.

Once the basic version is in, we can always follow up with a separate PR focused on performance improvements. That also gives us a clean place to track optimizations explicitly (e.g. via a GitHub issue) and to back them with benchmarks, so we’re making data-driven improvements rather than relying on assumptions.

Thanks for all the work you’ve put into this, @martinvuyk 🙏! I would just love to prioritize landing the core functionality first and then iterating from there. 😄

mojo/stdlib/std/collections/string/string_slice.mojo
NathanSWard: See comment here: https://github.com/modular/modular/pull/5780#issuecomment-3761595806

In this case, w/out benchmarks I'd still push to remove `likely/unlikely`. I'm not the biggest fan of increasing complexity based on assumptions. If this does show a perf improvement w/ benchmarks great, we could add it! But until then removing it is the simpler solution.

I hope my perspective makes sense - apologies for the churn on this PR 🙏

martinvuyk:
@NathanSWard This is the simpler version without hyper-optimizing like I did `split` and `splitlines` which took months of iteration in my free time. I don't want the first version to be slow like I think using the iterator will be because some PRs take ages to merge. For example I've got an optimization PR #3860 which has been waiting for more than a year, or #5395 which is still in the pipe even though both are very targeted and proven improvements. And I don't really think 40 lines of code is all that bad ( #3528 was initially 1200+ LOC that I split off into several PRs). I will be iterating on this and making sure that the performance is the best that it can be, you can be sure of that.

mojo/stdlib/std/collections/string/string_slice.mojo
martinvuyk: Any thoughts on this @NathanSWard ? We could also revisit it later on and add a debug assert or an abort for negative indexing in the mean time

martinvuyk: I just did that, let's leave it at not supporting negative indexing for now

martinvuyk:
@NathanSWard @barcharcraz @JoeLoser any chance we might push this through the finish line before 1.0? Either this or a simpler alternative. I can find ways to optimize this for ASCII later. Let me know and I can simplify this PR if that helps, or open a new one adding both codepoint indexing and slicing. But I think that being able to index and slice codepoints and graphemes would be a very nice-to-have

barcharcraz:
> @NathanSWard @barcharcraz @JoeLoser any chance we might push this through the finish line before 1.0? Either this or a simpler alternative. I can find ways to optimize this for ASCII later. Let me know and I can simplify this PR if that helps, or open a new one adding both codepoint indexing and slicing. But I think that being able to index and slice codepoints and graphemes would be a very nice-to-have

we discussed and we'll merge it. Please resolve the conflicts and then I'll sync it (go ahead and @ me afterwards)

barcharcraz (approved):
I still don't love how complicated this is, but it's self-contained and has tests.

martinvuyk:
@barcharcraz I rebased. Tests pass locally, the failing CI seems to be due to a flaky max test

barcharcraz:
!sync

barcharcraz:
!sync

barcharcraz:
merged :)

thanks Martin

You should see the code in the public repo by tomorrow

Ahajha:
Sorry for the copybara mishap, this landed in https://github.com/modular/modular/commit/40466c8399d918bd82217bcb990bcc1f6c13d082!

PR #5782, opened 2026-01-16
[MAX] Add Qwen3 embedding architecture

sbrunk:
Add support for [Qwen3 Embedding](https://github.com/QwenLM/Qwen3-Embedding) models in the MAX pipeline. The embedding variant uses the same architecture as Qwen3, so I've tried to re-use as much as possible. There are a few minor changes required on the Qwen3 generative models, i.e. a config to return the hidden states.

I also added the verification pipeline for all variants, which needed some changes when creating the golden set with torch, due to last token pooling and normalization. I used `--find-tolerances` to fill in the smallest recommended thresholds. On my machine, running `verify_pipelines` is now passing for all sizes (0.6B/4B/8B):

```shell
./bazelw run //max/tests/integration/pipelines/python:verify_pipelines -- --pipeline "Qwen/Qwen3-Embedding-4B-bfloat16" --devices gpu
```

There was one blocker I ran into which is a conflict in the pipeline registry: Because both the generative as well as the embedding model are using `Qwen3ForCausalLM`, it's not possible to register both with the current implementation of the registry. I changed the pipeline registry to support secondary lookup of architectures by both name and task and do a Huggingface `pipeline_tag` lookup, preparing for cases where multiple architectures share the same name but serve different tasks (`max/python/max/pipelines/lib/registry.py`).

I saw this comment which hints towards work already in progress on that. The registry changes are  contained in one commit a9f3bf76c1c3321751dd957236aa1b9d7a6c1e8c so it should be easy to remove/replace them.
https://github.com/modular/modular/blob/3200058c3537769aff36b1e03191fec129debcf7/max/python/max/entrypoints/cli/serve/serve_api_and_model_worker.py#L49-L55

**Qwen3 Embedding Model Support:**

* Added new architecture `qwen3_embedding_arch` for Qwen3-Embedding models, supporting tasks like embeddings generation for `"Qwen/Qwen3-Embedding-0.6B"`, `"Qwen/Qwen3-Embedding-4B"`, and `"Qwen/Qwen3-Embedding-8B"`
* Implemented `Qwen3EmbeddingPipelineModel` to handle embedding generation, including last-token pooling and L2 normalization, reusing Qwen3 generative model logic where possible (`max/python/max/pipelines/architectures/qwen3_embedding/model.py`).


sbrunk:
I have read the CLA Document and I hereby sign the CLA.

max/python/max/pipelines/architectures/qwen3_embedding/pooling.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    # Compute sequence lengths from row offsets
+    # For each sequence i: length = row_offsets[i+1] - row_offsets[i]
+    # Get ending offsets for each sequence
+    end_offsets = input_row_offsets[1:]  # Remove first element
+
+    # Compute last token index for each sequence (end_offset - 1)
+    # Since end_offset is exclusive, the last token is at end_offset - 1
+    one = ops.constant(1, DType.uint32, device=hidden_states.device)
+    last_token_indices = ops.sub(end_offsets, one)
```
k-w-w: `end_offsets - one` (and other usages of `ops.sub`, `ops.add`, `ops.mul`, `ops.div` can be replaced with operators)

max/python/max/pipelines/architectures/qwen3_embedding/layers/pooling.py
```diff
@@ -0,0 +1,81 @@
...
+def normalize_embeddings(embeddings: TensorValue) -> TensorValue:
+    """Apply L2 normalization to embeddings.
+
+    Args:
+        embeddings: Embeddings to normalize [batch_size, hidden_size]
+
+    Returns:
+        Normalized embeddings [batch_size, hidden_size]
+    """
+    # Cast to float32 BEFORE normalization for better numerical precision
```
k-w-w, line 67: Thanks for the clear comments! Casting the wrong place has bitten us many times before

max/python/max/pipelines/architectures/qwen3_embedding/model.py
```diff
@@ -0,0 +1,253 @@
...
+        for compatibility with the pipeline interface. Unlike the generative model,
+        our graph outputs a single tensor (embeddings) rather than multiple outputs.
+        """
+        assert isinstance(model_inputs, Llama3Inputs)
+
+        # Get KV cache inputs
+        curr_kv_cache_inputs = model_inputs.kv_cache_inputs or ()
+
+        # Execute the model - unlike parent, we don't pass through signal_buffers
+        # because the Qwen3 graph doesn't include them as inputs
```
k-w-w: nit: "Qwen3 embedding graph"

max/python/max/pipelines/architectures/qwen3_embedding/model.py
```diff
@@ -0,0 +1,253 @@
...
+        as inputs, so we allocate temporary KV cache entries for this batch.
+        """
+        # First, use parent to prepare basic inputs (tokens, row_offsets, etc.)
+        # But we pass None for kv_cache_inputs initially
+        inputs = super().prepare_initial_token_inputs(
+            replica_batches, None, return_n_logits
+        )
+
+        # Now populate KV cache inputs from the manager
+        # For embeddings, we need to temporarily allocate KV cache entries
```
k-w-w: Might be worth going into a bit more details about why this class needs to do it while other PipelineModels don't need to manually call alloc on the KV Cache (because the embeddings pipeline doesn't automatically generate KVCache inputs)

max/python/max/pipelines/architectures/qwen3_embedding/model_config.py
```diff
@@ -0,0 +1,60 @@
...
+
+from max.dtype import DType
+from max.graph import DeviceRef
+from max.nn.kv_cache import KVCacheParams
+from max.pipelines.architectures.qwen3.model_config import Qwen3Config
+from max.pipelines.lib import KVCacheConfig, MAXModelConfigBase, PipelineConfig
+from transformers import AutoConfig
+
+
+class Qwen3EmbeddingConfig(MAXModelConfigBase):
```
k-w-w: Is this needed?

max/python/max/pipelines/lib/registry.py
```diff
@@ -279,6 +324,14 @@ def retrieve_architecture(
...
             if use_module_v3:
                 architecture_name += "_ModuleV3"
+
+            # If task is provided and we have a task-specific match, use it
+            if task is not None:
+                task_key = (architecture_name, task)
+                if task_key in self._architectures_by_task:
+                    return self._architectures_by_task[task_key]
+
+            # Fall back to name-only match
```
k-w-w, line 394: Will `max serve --model-path Qwen/Qwen3-4B` still choose the text generation pipeline?

k-w-w (commented):
Amazing work! You were very thorough with the model implementation, dealing with the registry conflict and updating logit verification.

max/python/max/pipelines/lib/registry.py
sbrunk: Ah good catch! The registry changes are broken. I should have tested them with serving properly instead of just running the verification pipelines.

lattner:
Super exciting Sören, great work!

max/python/max/pipelines/architectures/qwen3_embedding/pooling.py
sbrunk: done

max/python/max/pipelines/architectures/qwen3_embedding/model_config.py
sbrunk: That was a leftover from refactoring. Removed now.

max/python/max/pipelines/lib/registry.py
sbrunk: From my understanding the issue with the registry occurs in cases where we want do derive the task type just from the model id, as in `max serve` in contrast to `max generate/max encode` where we can derive it from the command.

I couldn't find anything just from the Huggingface model config so the approach is to let the registry look-up the huggingface `pipeline_tag` to distinguish the task in case it's not clear from just the architecture name as in the Qwen3 case. `max serve`should now pick up `TextGenerationPipeline` or `EmbeddingsPipeline` for Qwen3/Qwen3-embedding respectively.

I'm not too familiar with the serving layer yet so you probably want to look closely if this makes sense. The registry changes should be contained in https://github.com/modular/modular/pull/5782/commits/a9f3bf76c1c3321751dd957236aa1b9d7a6c1e8c

Here's a "trace" of it (LLM generated so take it with a grain of salt):
```
1. max serve --model Qwen/Qwen3-Embedding-0.6B
   └─> serve_api_and_model_worker()
       ├─> PipelineConfig created with model_path="Qwen/Qwen3-Embedding-0.6B"
       │
       ├─> PipelineConfig.__init__()
       │   └─> @model_validator(mode="after")
       │       └─> _postprocess_configs()
       │           └─> resolve()  [line ~520]
       │               │
       │               ├─> _validate_and_resolve_remaining_pipeline_config(self.model)  [line 721]
       │               │   │
       │               │   └─> PIPELINE_REGISTRY.retrieve_architecture(
       │               │           huggingface_repo=HuggingFaceRepo("Qwen/Qwen3-Embedding-0.6B"),
       │               │           use_module_v3=False,
       │               │           task=None  ❌ NO TASK PROVIDED!
       │               │       )  [config.py:836]
       │               │       │
       │               │       ├─> Gets HF config: architectures=["Qwen3ForCausalLM"]
       │               │       │
       │               │       ├─> Finds TWO architectures with name "Qwen3ForCausalLM":
       │               │       │   • (Qwen3ForCausalLM, TEXT_GENERATION)
       │               │       │   • (Qwen3ForCausalLM, EMBEDDINGS_GENERATION)
       │               │       │
       │               │       ├─> ⚠️  AMBIGUOUS! Which one to return?
       │               │       │
       │               │       ├─> 🔍 Fetches model_info from HF Hub
       │               │       │   └─> Gets pipeline_tag="feature-extraction"
       │               │       │
       │               │       ├─> 🎯 Infers task: "feature-extraction" → EMBEDDINGS_GENERATION
       │               │       │
       │               │       └─> ✓ Returns qwen3_embedding_arch (correct!)
       │               │
       │               └─> Uses arch to validate model config (encoding, rope, etc)
       │
       └─> serve_api_and_model_worker() [line 43-100]
           │
           ├─> ⚠️ WORKAROUND: Hard-coded check for embedding models [line 52-58]
           │   if model_path in ["Qwen/Qwen3-Embedding-0.6B", ...]:
           │       pipeline_task = EMBEDDINGS_GENERATION
           │
           └─> PIPELINE_REGISTRY.retrieve_factory(
                   pipeline_config,
                   task=pipeline_task,  ✓ TASK NOW PROVIDED
                   override_architecture=None
               )  [line 65]
               │
               └─> retrieve_architecture(
                       huggingface_repo=...,
                       task=EMBEDDINGS_GENERATION  ✓ TASK KNOWN
                   )
                   └─> Directly looks up (Qwen3ForCausalLM, EMBEDDINGS_GENERATION)
                   └─> ✓ Returns qwen3_embedding_arch

**Comment by sbrunk** [comment]
The KV cache usage is still buggy. I wonder if it makes sense to try to get rid of the KV cache here after all as it isn't used, but we'll probably need to either duplicate more parts of the model, or refactor the base model class(es) to make the KV cache optional.

**Comment by sbrunk** [comment]
I re-worked this a bit now to avoid the KV cache completely in e44fa0de64910127bd0dfb81e47684fa2c3345cf. Validation is still passing with similar results.

It's a bit more duplication as we now have a separate transformer and attention implementation. But I think it's worth to avoid coupling with the generative layers too much due to KV cache and now we can also optimize independently.

## Benchmarks
I also ran initial throughput benchmarks and the results look reasonable (in line with vLLM) on a RTX 4090:

```
./bazelw run //max/python/max/benchmark:benchmark_throughput -- \
                                                             --pipeline.model Qwen/Qwen3-embedding-0.6B \
                                                             --pipeline-task embeddings_generation \
                                                             --other.num-prompts 1000 \
                                                             --input-len 1024 \
                                                             --pipeline.max-batch-size 4 \
                                                             --output-len 0

Benchmark Result:
--------
dataset_basename: Synthetic-hi
elapsed_time_ms: 9017.640986014158
num_requests: 1000
total_num_input_tokens: 1024000
total_num_output_tokens: 0
total_num_tokens: 1024000
requests_per_second: 110.89374721736455
tokens_per_second: 113555.1971505813

```

```
./bazelw run //max/python/max/benchmark:benchmark_throughput -- \
                                                             --pipeline.model Qwen/Qwen3-embedding-4B \
                                                             --pipeline-task embeddings_generation \
                                                             --other.num-prompts 1000 \
                                                             --input-len 1024 \
                                                             --pipeline.max-batch-size 4 \
                                                             --output-len 0

Benchmark Result:
--------
dataset_basename: Synthetic-hi
elapsed_time_ms: 52783.520614990266
num_requests: 1000
total_num_input_tokens: 1024000
total_num_output_tokens: 0
total_num_tokens: 1024000
requests_per_second: 18.945306950897187
tokens_per_second: 19399.99431771872
```

max/python/max/pipelines/architectures/qwen3_embedding/model.py
sbrunk: Superseded by the implementation without KV cache

max/python/max/pipelines/lib/registry.py
k-w-w: Amazing, thanks for adding this fix!

k-w-w (approved):
LGTM!

k-w-w:
!sync

PR #5771, opened 2026-01-13
[MAX] Add visual generation pipeline variant support

tolgacangoz:
Introduce support for the image generation pipeline using diffusion models. This includes new data structures and an enumeration for pixel generation tasks, along with updates to documentation to reflect these capabilities. Fixes duplicate export entries in the codebase.

Thanks for your review, @KCaverly! This is the first PR.

PR #5784, opened 2026-01-16
[Flux1] add interfaces for image generation

kkimmk:
This PR adds a first-class image generation task to MAX’s pipeline interface layer and wires up the pipelines registry to construct and run diffusion-style pipelines.

- New public interfaces: `ImageGenerationInputs`, `ImageGenerationOutput`, `PipelineTask.IMAGE_GENERATION`
- New pipeline variant wrapper (`ImageGenerationPipeline`):
   - This class serves as the base class for image generation tasks in **model-agnostic** manner.
   - It corresponds to `TextGenerationPipeline` in the LLM workflow.
- Registry updates to enable pipeline registry and retrieval.


KCaverly (commented):
I think we should use this PR instead: https://github.com/modular/modular/pull/5771.

The main difference surrounds Context generation and its use in the ImageGenerationInputs object below. ImageGenerationInputs should be tensors for the model forward pass, and likely should not include single-request inputs like prompt: str.

kkimmk:
@KCaverly Sounds good. I'll update our implementations based on this interface.
For the Context perspective, as I left as a NOTE, Context was planned be introduced with the following PR we are currently working on to support serving. Since this initial PR focuses on offline generation, I excluded the concept of "request" or "scheduler" to make it minimal. Eventually, this interface looks similar to our current work for serving support, so we can easily use it.

Issue #5672, opened 2025-12-16
[Feature Request] Migrate `kernels/src/Mogg` off of `LegacyUnsafePointer`.


NathanSWard:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

See [parent issue](https://github.com/modular/modular/issues/5665)

The following kernel files should be migrated off of `LegacyUnsafePointer`. (complexity indicated next to file)

```
Mogg/MOGGKernelAPI/MOGGKernelAPI.mojo (Low)
Mogg/MOGGPrimitives/MOGGIntList.mojo (Medium)
Mogg/MOGGPrimitives/MOGGPrimitives.mojo (High)
```

### What is your motivation for this change?

`LegacyUnsafePointer` is going to be deprecated and removed.

turakz:
HI, i've addressed this as part of working on #5676.

see branch here: https://github.com/turakz/modular/tree/migrate-LegacyUnsafePointer-to-UnsafePointer

will open a PR once work is done for remaining modules in #5676:

```
internal_utils/_measure.mojo
internal_utils/_testing.mojo
internal_utils/_utils.mojo
kv_cache/types.mojo
```

turakz:
regarding

```
Mogg/MOGGKernelAPI/MOGGKernelAPI.mojo (Low)
Mogg/MOGGPrimitives/MOGGIntList.mojo (Medium)
```

do we want this to remain a separate PR or do we just want to include them in the current PR: #5768

turakz:
have cut a branch and started working on this: https://github.com/turakz/modular/tree/5672-migrate-LegacyUnsafePointer-to-UnsafePointer

turakz:
update: merged #5768 into the branch cut for this task (to bring in `MOGGKernelAPI.mojo` from upstream and `MOGGPrimitives.mojo` from pr)

active work related to this ticket completes migration of `MOGGIntList.mojo`
- kernels build locally
- kernels test pass locally

am waiting on nightly to hit upstream and then ill squash that branch and open a PR.

PR #5792, opened 2026-01-17
[Kernels] migrate-MOGGIntList-to-UnsafePointer

turakz:
## Summary

migrates `Mogg/MOGGPrimitives/MOGGIntList.mojo` to `UnsafePointer`.

- issue: https://github.com/modular/modular/issues/5672#issuecomment-3762301533

## Test Plan

- `./bazelw build //max/kernels/...`     passes
- `./bazelw test //max/kernels/test/...` passes

### DIFF
--- a/max/kernels/src/Mogg/MOGGPrimitives/MOGGIntList.mojo
+++ b/max/kernels/src/Mogg/MOGGPrimitives/MOGGIntList.mojo
@@ -14,9 +14,7 @@
 
 
 from buffer.dimlist import DimList, _make_tuple
-from memory import LegacyUnsafePointer, memset_zero
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
+from memory import memset_zero
 
 from utils.index import IndexList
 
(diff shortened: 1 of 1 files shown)


max/kernels/src/Mogg/MOGGPrimitives/MOGGIntList.mojo
```diff
@@ -259,5 +257,5 @@ struct IntList[static_values: DimList = DimList()](
     fn __del__(deinit self):
         @parameter
         if not Self.has_static_length():
-            if self.data != UnsafePointer[Int]():
+            if self.data != UnsafePointer[Int, MutExternalOrigin]():
```
JoeLoser: **Suggestion** I understand this is pre-existing, but we can simplify a bit as the null pointer check is not necessary. We can just always call `self.data.free()` within the body of the `@parameter if not Self.has_static_length()`.

Why:
1. UnsafePointer.free() delegates to LegacyUnsafePointer.free() which calls _free() (unsafe_pointer.mojo:1440)
2. _free() implementation (memory.mojo:519-524):
fn _free(ptr: UnsafePointer[mut=True, ...]):
    @parameter
    if is_gpu():
        libc.free(ptr.bitcast[NoneType]())
    else:
        __mlir_op.`pop.aligned_free`(ptr.address)
3. libc.free() (_libc.mojo:29-37) - This is the standard C free() function. Per the C standard (C11 §7.22.3.3), free(NULL) is explicitly defined as a no-op - safe to call with null.
4. pop.aligned_free - Looking at usage in the codebase (e.g., _closure.mojo:37), it's called directly in __del__ without any null check, suggesting it also handles null gracefully.

JoeLoser (approved):
LGTM but consider my one cleanup comment 👍

max/kernels/src/Mogg/MOGGPrimitives/MOGGIntList.mojo
turakz: ahh, that makes sense. given that, it appears some of the existing nullchecks are redundant.

i can revise, np. thank you for the feedback!

NathanSWard (approved):
thank you!

NathanSWard:
!sync

PR #5794, opened 2026-01-17
[stdlib] Use `@align(64)` for `BlockingSpinLock`

msaelices:
Cache-line align BlockingSpinLock to prevent false sharing when multiple locks are accessed concurrently from different threads in the same cache-line, which would slowdown the accesses per cache invalidation degradation.


NathanSWard:
See here: https://github.com/modular/modular/pull/5793#issuecomment-4042125862

PR #5798, opened 2026-01-18
[stdlib] Optimize the String repr implementation

martinvuyk:
Optimize the String repr implementation. I also updated the `_write_hex` util to use the safer writer API. And I also added a static method version of the `is_ascii_printable` `Codepoint` method given parsing a codepoint from UTF-8 is not free; all to then just do a byte comparison.

## Benchmark results

CPU: Intel® Core™ i7-7700HQ

|  | old value (ms) | new value (ms) | markdown perc | magnitude|
|-- | -- | -- | -- | --|
|bench_string_repr[10] | 12.53 | 9.26 | 0.26 | 1.35|
|bench_string_repr[30] | 14.05 | 10.67 | 0.24 | 1.32|
|bench_string_repr[50] | 14.64 | 9.92 | 0.32 | 1.48|
|bench_string_repr[100] | 14.20 | 9.22 | 0.35 | 1.54|
|bench_string_repr[1000] | 12.48 | 8.21 | 0.34 | 1.52|
|bench_string_repr[10000] | 11.97 | 7.84 | 0.35 | 1.53|
|bench_string_repr[100000] | 11.89 | 7.73 | 0.35 | 1.54|
|bench_string_repr[1000000] | 11.88 | 7.65 | 0.36 | 1.55|

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_string.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_string.mojo
@@ -376,6 +376,28 @@ fn bench_string_write[short: Bool](mut b: Bencher) raises:
 
 
 # ===-----------------------------------------------------------------------===#
+# Benchmark string repr
+# ===-----------------------------------------------------------------------===#
+@parameter
+fn bench_string_repr[
+    length: Int = 0, filename: StaticString = "UN_charter_EN"
+](mut b: Bencher):
+    var items = make_string[length](filename + ".txt")
+
+    @always_inline
+    @parameter
+    fn call_fn():
+        # this is to help with instability when measuring small strings
+        for _ in range(10**6 // length):
+            var s = String()
+            items.write_repr_to(s)
+            keep(s)
+
+    b.iter[call_fn]()
+    keep(Bool(items))
+
+
+# ===-----------------------------------------------------------------------===#
 # Benchmark Main
 # ===-----------------------------------------------------------------------===#
 def main():
@@ -471,6 +493,9 @@ def main():
             m.bench_function[bench_write_utf8[length, fname]](
                 BenchId(String("bench_write_utf8", suffix))
             )
+            m.bench_function[bench_string_repr[length, fname]](
+                BenchId(String("bench_string_repr", suffix))
+            )
 
     m.bench_function[bench_string_join[True]](
         BenchId(String("bench_string_join_short"))
@@ -479,4 +504,15 @@ def main():
         BenchId(String("bench_string_join_long"))
     )
 
-    print(m)
+    # NOTE: do not delete this. This is supposed to measure the average for
+    # different languages. You can use print(m) if you wish to see the
+    # per-language breakdown
+    results = Dict[String, Tuple[Float64, Int]]()
+    for info in m.info_vec:
+        n = info.name
+        time = info.result.mean("ms")
+        avg, amnt = results.get(n, (Float64(0), 0))
+        results[n] = ((avg * amnt + time) / (amnt + 1), amnt + 1)
+    print("")
+    for k_v in results.items():
+        print(k_v.key, k_v.value[0], sep=", ")
--- a/mojo/stdlib/std/collections/string/codepoint.mojo
+++ b/mojo/stdlib/std/collections/string/codepoint.mojo
@@ -378,15 +378,32 @@ struct Codepoint(
         comptime ord_z = UInt32(ord("z"))
         return ord_a <= self.to_u32() <= ord_z
 
+    @staticmethod
+    @always_inline
+    fn _is_ascii_printable(codepoint: Scalar) -> Bool:
+        """Determines whether the given character is a printable character.
+
+        Args:
+            codepoint: The codepoint to check.
+
+        Returns:
+            True if the character is a printable character, otherwise False.
+        """
+        __comptime_assert (
+            codepoint.dtype.is_integral()
+        ), "only integral codepoints exist"
+        comptime ` ` = type_of(codepoint)(ord(" "))
+        comptime `~` = type_of(codepoint)(ord("~"))
+        return ` ` <= codepoint <= `~`
+
+    @always_inline
     fn is_ascii_printable(self) -> Bool:
         """Determines whether the given character is a printable character.
 
         Returns:
             True if the character is a printable character, otherwise False.
         """
-        comptime ord_space = UInt32(ord(" "))
-        comptime ord_tilde = UInt32(ord("~"))
-        return ord_space <= self.to_u32() <= ord_tilde
+        return Self._is_ascii_printable(self.to_u32())
 
     @always_inline
     fn is_python_space(self) -> Bool:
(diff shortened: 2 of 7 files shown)


NathanSWard:
Ty for this!
Before merging though, I would request we add benchmarks for this so we can run and see the perf improvements.
We already have a `bench_string.mojo` file, you would just need to add a `bench_string_write_repr_to` so we can capture these improvements.

martinvuyk:
> Before merging though, I would request we add benchmarks for this so we can run and see the perf improvements. We already have a `bench_string.mojo` file, you would just need to add a `bench_string_write_repr_to` so we can capture these improvements.

Yeah I can add a benchmark, I was just lazy seeing if I could slip this without notice lol. I wrote most of the benchmarks in that file and the whole bench setup by the way :)

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -378,15 +378,32 @@ struct Codepoint(
         comptime ord_z = UInt32(ord("z"))
         return ord_a <= self.to_u32() <= ord_z
 
+    @staticmethod
+    @always_inline
+    fn is_ascii_printable(codepoint: Scalar) -> Bool:
```
soraros: Is it a good idea to expose this? If we do, we'd need to come up with semantics for out of range input values, which feels a bit adhoc to me. I'm not against that idea though. Maybe we should have `dtype.is_unsigned()` if we want this to be public API. Also, why don't we have the where clause on the type signature?

mojo/stdlib/std/format/_utils.mojo
```diff
@@ -209,28 +208,25 @@ fn _hex_digits_to_hex_chars(
...
-    _hex_digits_to_hex_chars(ptr, UInt16(ord("你")))
+    ptr.store(_hex_digits_to_hex_chars(UInt16(ord("你"))))
     assert_equal("4f60", S(ptr=ptr, length=4))
-    memset_zero(ptr, len(items))
-    _hex_digits_to_hex_chars(ptr, UInt8(ord("Ö")))
+    ptr.store(_hex_digits_to_hex_chars(UInt8(ord("Ö"))))
     assert_equal("d6", S(ptr=ptr, length=2))
     ```
     """
     comptime size = size_of[decimal.dtype]()
```
soraros, line 219: Maybe also add a `__comptime_assert` here since we are already touching this code.

mojo/stdlib/std/collections/string/codepoint.mojo
martinvuyk: > we'd need to come up with semantics for out of range input values, which feels a bit adhoc to me

I think we could just do as I did which is either have an upper bound constraint or parametrize on the dtype width, or just only accept `UInt8`

> Maybe we should have dtype.is_unsigned() if we want this to be public API

I don't know, an unsigned integer still can represent a valid codepoint. It would just be a matter of checking the bounds correctly

> Is it a good idea to expose this?

Overall you've made me doubt it now. We could totally make it private since we'll be only using this type of function for very performance sensitive cases in the stdlib

NathanSWard:
> Yeah I can add a benchmark, I was just lazy seeing if I could slip this without notice lol. I wrote most of the benchmarks in that file and the whole bench setup by the way :)

I'm just doing my best with 1.0 around the corner to make sure we do a better job at adding tests & benchmarks for appropriate PRs. Since I assume the OSS contributions will increase a bit with 1.0 so I'm trying to ideally set some "PR standards" haha :)
And greatly appreciate you putting the benchmark file there in the first place! We'll have to add similar stuff for other stdlib collections & algorithms here over the new couple months!

mojo/stdlib/std/collections/string/codepoint.mojo
NathanSWard: +1 on making this private as I'm not sure we want to expose it.
I can happily do that after I pull this into the internal repo or you can do it directly @martinvuyk - lmk which is easiest!

martinvuyk: done

NathanSWard (approved):

NathanSWard:
!sync

martinvuyk:
@NathanSWard just fixed an issue that I wasn't escaping single quotes. It is technically different from Python's behavior but I do agree its a better direction to go in. I added some comments about this in the `write_repr_to` docstrings

NathanSWard:
!sync

mojo/stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -375,6 +375,28 @@ fn bench_string_write[short: Bool](mut b: Bencher) raises:
...
+    var items = make_string[length](filename + ".txt")
+
+    @always_inline
+    @parameter
+    fn call_fn():
+        # this is to help with instability when measuring small strings
+        for _ in range(10**6 // length):
+            var s = String()
+            items.write_repr_to(s)
+            keep(s)
```
NathanSWard, line 394: One things about this benchmark is there is a chance that the `String` allocation could potentially cause invalid perf benchmarks since the cost of string allocation could be considerably greater than that of the `write_repr_to` implementation.

I went ahead and updated this internally by using a `NullWriter`
```
@fieldwise_init
struct NullWriter(ImplicitlyCopyable, Writer):
    fn write_string(mut self, string: StringSlice):
        keep(string)
```

So we don't pay the cost of allocation, but don't get the `write_string` call optimized out either :) 
And the good news is your changes are still a positive perf improvement 🥳

martinvuyk: That's a nice trick! Though I was thinking of using this as an example of what kind of perf improvements we could have once we develop an API for the writer to fetch the estimated byte size of the output so that it allocates only once. But yeah we could totally separate that out into another benchmark

mojo/stdlib/test/io/test_write.mojo
```diff
@@ -88,40 +88,38 @@ def test_hex_digits_to_hex_chars():
...
-    assert_equal(r"\U0001f525", String(S(ptr=ptr, length=10)))
-    memset_zero(ptr, len(items))
-    _write_hex[4](ptr, ord("你"))
-    assert_equal(r"\u4f60", String(S(ptr=ptr, length=6)))
-    memset_zero(ptr, len(items))
-    _write_hex[2](ptr, ord("Ö"))
-    assert_equal(r"\xd6", String(S(ptr=ptr, length=4)))
+    var s = String()
+    _write_hex[amnt_hex_bytes=8](s, Scalar[DType.int](ord("🔥")))
+    assert_equal(r"\U0001f525", s)
```
martinvuyk, line 116: @NathanSWard sorry I forgot to run the tests and since CI isn't running them it took me looking at this code to realize it was wrong. I was missing constructing a scalar and unescaping the sequences since I copied them from the docstrings

NathanSWard: No worries! I fixed this internally already :)

PR #5786, opened 2026-01-16
[Flux1] add model definitions for components of Flux1 pipeline

kkimmk:
This PR adds the model-side building blocks needed to run the Flux1 diffusion pipeline in MAX. It introduces MAX implementations for the FLUX transformer, its text encoders (CLIP + T5), and the VAE decoder.

- Adds `flux1_arch` as a `SupportedArchitecture` for `PipelineTask.IMAGE_GENERATION`, pointing to `FluxPipeline`. This serves as a registration point for all the components. So, other remaining components don't have `arch.py` defined.
- Adds model implementations for each component: clip, t5, flux, and autoencoder.
- Adds BaseModel interface for diffusion sub-models, to restrict the initialization flow. `PipelineModel` is implemented specifically for LLM models, so this simpler base class is introduced.

**NOTES**
Currently, V3 migration is implemented partially in this PR. I'll make updates to V3 migration as soon as possible.

PR #5811, opened 2026-01-21
[MAX] Add Clip implementation for Flux.1 pipeline

kkimmk:
### Overview
This PR adds the CLIP model components needed to support the Flux1 pipeline. It introduces a dedicated CLIP architecture package, a model interface layer to standardize integration, and supporting dtype/norm utilities so the CLIP stack can be wired consistently with the rest of MAX’s modules.

This PR is made by splitting Clip implementation from the original PR(#5786). 

### What’s included
- New CLIP architecture with core model implementation, common model wrapper, and configuration objects under `max/python/max/pipelines/architectures/clip/`.
- A new shared base model interface (`max/python/max/pipelines/lib/interfaces/max_model.py`) to align intialization and execution methods of models.
- Dtype extension helpers (`max/python/max/dtype/dtype_extension.py`) and exports to support new model components.
- Module V3 LayerNorm implementation and registry updates to enable CLIP’s normalization requirements.

### NOTES
This model implementation does NOT serve as a standalone executable model. It is expected to be executed within the Flux1 pipeline. (Will update the related PR link later)


pei0033:

pei0033:
The current PR's Config implementation, which uses both XXConfigBase and XXConfig simultaneously, is an outdated approach.
While it functions without issues, it needs to align with the config methods of other TextGeneration models in the future.

max/python/max/nn/norm/layer_norm.py
```diff
@@ -0,0 +1,118 @@
...
+            self.weight = Tensor.ones([dim])
+            self.bias = Tensor.zeros([dim]) if use_bias else None
+        else:
+            self.weight = None
+            self.bias = None
+
+    def __rich_repr__(self):
+        """Repr matching the Linear constructor."""
+        yield "dim", self.dim
+        yield "eps", self.eps, 1e-6
```
katelyncaldwell: Should this be 1e-5 to match the default value in the `init`?

katelyncaldwell (commented):
One minor comment, but nothing blocking. We do need to get CI passing before I merge, though. Do you understand the errors?

I will follow up about adding the `dtype` extension to `dtype` library itself. Are you planning to tackle the outdated XXConfigBase and XXConfig in a follow up?

max/python/max/nn/norm/layer_norm.py
pei0033: I fixed it!

pei0033:
> We do need to get CI passing before I merge, though. Do you understand the errors?

I'm not sure. Looking at the [test log](https://modular-public.buildbuddy.io/invocation/4c04b92a-c18d-41a8-8f5b-41ee8cf9dd77), tests unrelated to this PR appear to be failing because the worker crashed. I ran a few in my environment, and they succeeded.

> Are you planning to tackle the outdated XXConfigBase and XXConfig in a follow up?

Once the entire flux pipeline is completed, I plan to address this issue.

KCaverly:
!sync

PR #5812, opened 2026-01-21
[MAX] Add T5 implementation for Flux.1 pipeline

kkimmk:
### Overview
This PR adds the T5 model stack required for the Flux1 pipeline. It introduces a new T5 architecture package with the main model implementation, configuration, and weight adapter utilities.

This PR is made by splitting T5 implementation from the original PR(#5786).

### What’s included
- New T5 architecture package under `max/python/max/pipelines/architectures/t5/`.
- Weight adapter utilities for mapping HF checkpoints (weight_adapters.py).

### NOTES
This model implementation does NOT serve as a standalone executable model. It is expected to be executed within the Flux1 pipeline. (Will update the related PR link later)
Definition of the base abstract class `MaxModel` is defined in #5811


KCaverly:
!sync

PR #5813, opened 2026-01-21
[MAX] Add Flux1 implementation for Flux.1 pipeline

kkimmk:
### Overview
This PR introduces the core Flux1 model implementation required by the Flux1 pipeline. It adds the architecture package, model definition, configuration, and supporting layers (attention, embeddings, activations, normalizations) along with weight adapter utilities.

This PR is made by splitting Flux1 implementation from the original PR(https://github.com/modular/modular/pull/5786).

### What’s included
- New Flux1 architecture package under `max/python/max/pipelines/architectures/flux1/`.
- Supporting layers for attention, embeddings, activations, and normalization, used in Flux1 model.
- Weight adapter utilities for mapping HF checkpoints (weight_adapters.py).

### NOTES
This model implementation does NOT serve as a standalone executable model. It is expected to be executed within the Flux1 pipeline. (Will update the related PR link later)
Definition of the base abstract class MaxModel is defined in https://github.com/modular/modular/pull/5811

PR #5810, opened 2026-01-21
[MAX] Add AutoencoderKL VAE decoder implementation for Flux.1 pipeline

byungchul-sqzb:
## Summary
This PR implements the VAE decoder for Flux.1 pipeline, enabling the conversion of latent representations to images. The implementation follows the [diffusers library's AutoencoderKL](https://github.com/huggingface/diffusers/tree/main/src/diffusers/models/autoencoders) architecture and integrates with MAX's `module_v3` framework.

This PR splits the VAE decoder implementation from #5786 into a separate, focused PR to facilitate review and maintainability.

This PR consists of two commits:
- `[MAX] Add Conv2d and GroupNorm to module_v3` - Adds foundational layers required for VAE implementation
- `[Pipelines] Add AutoencoderKL VAE decoder implementation for Flux.1` - Implements the complete decoder pipeline

## Architecture

The implementation follows a modular structure designed for extensibility:
- `autoencoders/`
    - `autoencoder_kl.py` - AutoencoderKL model + MaxModel wrapper
    - `model.py` - BaseAutoencoderModel (shared logic)
    - `model_config.py` - Configuration classes
    - `vae.py` - Decoder implementation
    - `layers/`
        - `attention.py` - VAEAttention
        - `resnet.py` - ResnetBlock2D
        - `upsampling.py` - Upsample2D

## Extensibility

The modular architecture allows easy extension for future generative model tasks. For example, to add support for Flux.2, one would only need to:
- Create `autoencoder_kl_flux2.py` with a new `AutoencoderKLFlux2Model` class
- Inherit from `BaseAutoencoderModel` and reuse existing layers
- The shared components (`Decoder`, `ResnetBlock2D`, `VAEAttention`, etc.) can be reused without modification

This design pattern follows the diffusers library's approach and ensures maintainability as new VAE variants are added.

## Notes

- Only the decoder path is implemented (encode functionality not included) as Flux.1 T2I task only requires decoding. This structure will be gradually extended in future PRs to support additional functionality.
- Implementation is based on diffusers' `AutoencoderKL` architecture
- All code follows MAX framework conventions and uses `module_v3` APIs

PR #5814, opened 2026-01-21
[MAX] Add pipeline implementation for Flux.1 pipeline

kkimmk:
### Overview
This PR adds the core Flux1 diffusion pipeline implementation. It introduces the Flux1 pipeline module and architecture wiring, along with a shared diffusion pipeline interface to standardize how diffusion models integrate with MAX pipelines.

### What’s included
- New Flux1 pipeline implementation (`max/python/max/pipelines/architectures/flux1/pipeline_flux.py`).
- New diffusion pipeline interface (`max/python/max/pipelines/lib/interfaces/diffusion_pipeline.py`)

### NOTES
- The `DiffusionPipeline` abstract class will serve as a base interface for pipeline definition for each model. It's registered as the `pipeline_model` for pipeline registry.
- Current implementation requires each model to declare its components through a class variable. This may be better handled by registering through `SupportedArchitecture` on the `Registry` side; however, for now, it is implemented this way to minimize dependencies on the `Registry`.
- For the individual component implementations, please refer to #5810 #5811 #5812 #5813 .


lee-jingu:

KCaverly:
!sync

max/python/max/pipelines/architectures/flux1/pipeline_flux.py
```diff
@@ -0,0 +1,396 @@
...
+        noise_pred = self.transformer(
+            latents,
+            prompt_embeds,
+            pooled_prompt_embeds,
+            timestep,
+            latent_image_ids,
+            text_ids,
+            guidance,
+        )[0]
+        return noise_pred
```
tolgacangoz: Do you think this is a necessary abstraction?

max/python/max/pipelines/architectures/flux1/pipeline_flux.py
```diff
@@ -0,0 +1,396 @@
...
+                model_inputs.guidance_scale,
+                device=self.transformer.devices[0],
+                dtype=dtype,
+            )
+        else:
+            guidance = Tensor.zeros(
+                [latents.shape[0]],
+                device=self.transformer.devices[0],
+                dtype=dtype,
+            )
```
tolgacangoz, line 308: Since @KCaverly said making as much preprocessing as possible at the tokenizer stage, I already generated the `guidance` tensor there. WDYT?

max/python/max/pipelines/architectures/flux1/pipeline_flux.py
jglee-sqbits: I tend to define the method first and fill in the implementation afterward, but I agree this might be an unnecessary abstraction here.

max/python/max/pipelines/architectures/flux1/pipeline_flux.py
jglee-sqbits: I agree this part is a bit ambiguous as well. If you look at the code below:
https://github.com/modular/modular/blob/28217293e1918cf47cc7ce7d8eaf78e6b98c87c7/max/python/max/pipelines/lib/pixel_tokenizer.py#L519-L520

the pipeline receives a numpy value that doesn’t account for batching. My understanding is that batching is determined internally within the pipeline.

I assumed the split—doing CPU-bound work in PixelGenerationTokenizer and GPU-bound work in the pipeline—was mainly for efficiency. In that case, rather than looking at a NumPy value, expanding it to batch_size, and then transferring it from CPU → GPU (tensor), I thought it would be more efficient to determine batch_size in the pipeline and create the corresponding GPU tensor directly from the scalar value(`guidance_scale`).

That said, I think we could make this more explicit and potentially switch to an even more efficient approach later.

PR #5816, opened 2026-01-21
[MAX] Propose to add `PixelGenerationTokenizer` and `PixelContext` for image/video generation

tolgacangoz:
This PR proposes to add `PixelGenerationTokenizer` and `PixelContext` for image and video generation.


kkimmk:
Hi @tolgacangoz , I'd like to continue discussing the `diffusers_config`, you mentioned, in this thread. 

> but I'm happy to defer to your judgment there

Is it right that you are referring to using `MAXModelConfig` defined for each component, in the Flux1 model-component PRs? If it's right, I believe those configs are not accessible at this tokenizing stage where there's no context for the `PixelGenerationPipeline`. So, I think your approach to pull the configs from `pipeline_config` is more reasonable here!
Could you share a code snippet for saving those Diffusers configs into the `pipeline_config`? It would be helpful because it's related to the initialization process of `PixelGenerationPipeline` regarding the checkpoint downloading. Also, in the same context, I'd like to ask how the `subfolder` is being resolved when initializing the tokenizer.

tolgacangoz:
Since we are supposed to run `retrieve_timesteps()`, `prepare_latents()`, and other CPU-bound small calculations in the tokenizer, we should be able to utilize the `diffusers` model's configuration info in the tokenizer, thus I propose adding `diffusers` configs into the `pipeline_config`. I was thinking of saving those `diffusers` configs into the `pipeline_config` like that: https://github.com/modular/modular/pull/5604/changes#diff-e8ceba6af15259b1a2c58017e1878dab57236fb9691c517e6f3796d9d9060396

I thought just adding `subfolder="tokenizer"` [here](https://github.com/modular/modular/pull/5604/changes#diff-650c22b61d8b64d35300f9e63d85c5024dec9c2b4e623b4b80d43eaede4913d8R509) seems OK, because our tokenizer accepts it [normally](https://github.com/modular/modular/pull/5816/changes#diff-8dafd283fe1be08c9c5b19dfbe5bd5c46e9de327d853ef1353aff7d6c404ac9aR871), as others accept it in their `unused_kwargs`.

WDYT? Is there something I missed? Feel free to recommend if there might be better ways to do them.

kkimmk:
Thank you. I also think it's a good choice!

max/python/max/interfaces/pipeline_variants/pixel_generation.py
```diff
@@ -28,38 +28,42 @@
...
     """
-    guidance_scale: float = 7.5
+    guidance_scale: float = 3.5
     """
     Guidance scale for classifier-free guidance. Set to 1.0 to disable CFG.
     """
+    true_cfg_scale: float = 1.0
+    """
+    True classifier-free guidance is enabled when true_cfg_scale > 1.0 and negative_prompt is provided.
+    """
```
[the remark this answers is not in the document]
tolgacangoz: Hmm, you might be right. Should I open another PR for `PixelGenerationRequest`-related changes @kcaverly before working on this PR?

tolgacangoz:
- Adding configuration support for `diffusers`-based models is ~WIP at the moment~: https://github.com/modular/modular/pull/5849 I had worked on it in my first main PR, but I don't have the proper bandwidth for it to continue anymore at the moment.
- Scheduler/Sampler/Solver of diffusion models is a placeholder for now. It is going to be added in another PR: https://github.com/modular/modular/pull/5815

max/python/max/pipelines/lib/diffusion_schedulers/scheduling_flow_match_euler_discrete.py
```diff
@@ -0,0 +1,41 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+
+import numpy as np
+import numpy.typing as npt
+
+
+class FlowMatchEulerDiscreteScheduler:
```
KCaverly, line 19: What is this class doing? Is it just a stub, if so is it going to be updated?

max/python/max/pipelines/lib/diffusion_schedulers/scheduler_factory.py
```diff
@@ -0,0 +1,58 @@
...
+"""Scheduler factory for creating schedulers from diffusers configuration."""
+
+import importlib
+from typing import Any
+
+# Registry mapping scheduler class names to (module_path, class_name) tuples.
+# Module paths are relative to this package.
+_SCHEDULER_REGISTRY: dict[str, tuple[str, str]] = {
+    "FlowMatchEulerDiscreteScheduler": (
+        ".scheduling_flow_match_euler_discrete",
```
KCaverly, line 23: Using modular_path & class name, here feels quite brittle to future refactors. Can you just import the class and use it directly here instead?

max/python/max/pipelines/lib/diffusion_schedulers/scheduling_flow_match_euler_discrete.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, **kwargs) -> None:
+        self.config = type("Config", (), {"use_flow_sigmas": False})()
+        self.timesteps = np.array([], dtype=np.float32)
+        self.sigmas = np.array([], dtype=np.float32)
+        self.order = 1
+
+    def set_timesteps(
+        self,
+        num_inference_steps: int | None = None,
```
KCaverly, line 30: Do we need all this optionality here? It seems like a potential for bugs, to have a method that can accept all None values, and essentially silently no-op.

max/python/max/pipelines/lib/diffusion_schedulers/scheduling_flow_match_euler_discrete.py
KCaverly, line 1: If this is just a stub to illustrate how Scheduler Factory / _SCHEDULER_REGISTRY works. Can we move it to an example comment in the scheduler_factory.py, as opposed to introducing a stubbed class here?

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+    ]
+):
+    """Encapsulates creation of PixelContext and specific token encode/decode logic.
+
+    Args:
+        model_path: Path to the model/tokenizer
+        revision: Git revision/branch to use
+        max_length: Maximum sequence length
+        trust_remote_code: Whether to trust remote code from the model
+        enable_llama_whitespace_fix: Enable whitespace fix for Llama tokenizers
```
KCaverly, line 862: Please ensure these doc strings align with the arguments in the function below.

max/python/max/pipelines/core/context.py
```diff
@@ -569,6 +571,143 @@ def next_speech_tokens(
...
+
+    # Request identification
+    request_id: RequestID = field(default_factory=RequestID)
+
+    model_name: str = field(default="")
+
+    mask: npt.NDArray[np.bool_] | None = field(default=None)
+    """Mask for text encoder's attention."""
+
+    tokens_2: TokenBuffer | None = field(default=None)
```
KCaverly, line 614: Can you please rename this to secondary_tokens. To more clearly identify what it is used for.

max/python/max/pipelines/core/context.py
```diff
@@ -569,6 +571,143 @@ def next_speech_tokens(
...
+    """Secondary encoder tokens. None for single-encoder models."""
+
+    negative_tokens: TokenBuffer | None = field(default=None)
+    """Negative tokens for primary encoder."""
+
+    negative_tokens_2: TokenBuffer | None = field(default=None)
```
KCaverly, line 620: Can you please rename this to secondary_negative_tokens, to more clearly identify what it is used for.

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
         return _rgba_to_rgb(image)
     else:
         return image.convert(to_mode)
+
+
+class PixelGenerationTokenizer(
+    PipelineTokenizer[
+        PixelContext,
+        tuple[npt.NDArray[np.integer[Any]], npt.NDArray[np.bool_]],
```
KCaverly, line 851: Can you please enforce proper np.int64 type hints for tokens, throughout this class. We should have very limited use of Any.

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+        subfolder: str,
+        *,
+        subfolder_2: str | None = None,
+        revision: str | None = None,
+        max_length: int | None = None,
+        max_length_2: int | None = None,
+        trust_remote_code: bool = False,
+        chat_template: str | None = None,
+        context_validators: list[Callable[[PixelContext], None]] | None = None,
+        wrap_prompt_as_chat: bool = False,
```
KCaverly: Practically, is this used in Image Generation models? Since it is only wrapping the prompt & not the negative prompt, it wouldnt appear that both are expected to be used together.

While FLUX2.0 does not appear to use negative prompt at all, and does not expect messages as inputs, is this going to be immediately used?

Curious, if we can remove this, for now, to ignore messages, unless passed directly in via the Request.

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+        else:
+            scheduler.set_timesteps(
+                num_inference_steps, device=device, **kwargs
+            )
+            timesteps = scheduler.timesteps
+        return timesteps.astype(np.float32), num_inference_steps
+
+    async def _generate_tokens_ids(
+        self,
+        prompt: str,
```
KCaverly, line 1048: This type signature:
- Accepts the scenario in which prompt and messages are provided. I do not believe they both should be used at the same time. Please clarify exactly what the accepted behaviour here is, and error early, before this method is called.
- Accepts the scenario in which prompt is optional. If this is also not-supported, please raise an error early.

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+        npt.NDArray[np.integer[Any]] | None,
+    ]:
+        """Tokenize prompt(s) with encoder model(s).
+
+        Returns:
+            Tuple of (token_ids, attn_mask, token_ids_2, negative_token_ids, negative_token_ids_2).
+            token_ids_2 and negative_token_ids_2 are None if no secondary tokenizer is configured.
+        """
+        if messages is not None:
+            prompt = self.apply_chat_template(messages, chat_template_options)
```
KCaverly, line 1069: You likely have to convert the messages to [message.model_dump() for message in messages] here.

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+            )
+        negative_token_buffer_2 = None
+        if negative_token_ids_2 is not None:
+            negative_token_buffer_2 = TokenBuffer(
+                array=negative_token_ids_2.astype(np.int64, copy=False),
+            )
+
+        # 3. Resolve image dimensions
+        # Get defaults from diffusers_config
+        vae_config = self.diffusers_config.components["vae"].config_dict
```
KCaverly, line 1248: From a performance perspective, retrieving this repeatedly for each request is likely to be a problem. Maybe we can move the values that are Identical for each request, either to __init__, or (to keep things cleaner) to a cached property?

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+        mu = self._calculate_shift(
+            image_seq_len,
+            scheduler_config.get("base_image_seq_len", 256),
+            scheduler_config.get("max_image_seq_len", 4096),
+            scheduler_config.get("base_shift", 0.5),
+            scheduler_config.get("max_shift", 1.15),
+        )
+
+        # Create scheduler from config
+        scheduler_component = self.diffusers_config.components["scheduler"]
```
KCaverly, line 1280: For perf, can we bring this into the __init__? I imagine we can probably create on scheduler there, and reuse it?

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+        scheduler = SchedulerFactory.create(
+            scheduler_component.class_name, scheduler_config
+        )
+
+        sigmas: npt.NDArray[np.float32] | None = np.linspace(
```
KCaverly, line 1285: Where is sigmas None? This type hint indicates None, and then it is universally set, can we remove the optionality here?

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+            1.0,
+            1.0 / request.num_inference_steps,
+            request.num_inference_steps,
+            dtype=np.float32,
+        )
+        if (
```
KCaverly, line 1291: Can we cache this if inside a cached_property?

max/python/max/pipelines/lib/tokenizer.py
KCaverly: See the below, where we set sigmas=None. If we are setting sigmas to None, when use_flow_sigmas=True. Can we instantiate the linspace sigmas, only when needed, instead of unconditionally?

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+        ):
+            sigmas = None
+        timesteps, num_inference_steps = self._retrieve_timesteps(
+            scheduler,
+            request.num_inference_steps,
+            CPU(),
+            sigmas=sigmas,
+            mu=mu,
+        )
+        if request.model_name == "Tongyi-MAI/Z-Image-Turbo":
```
KCaverly, line 1303: For my understanding, why do we hardcode the model name in here. Is there anyway around this?

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+    def _prepare_latents(
+        self,
+        batch_size: int,
+        num_channels_latents: int,
+        height: int,
+        width: int,
+        seed: int | None,
+        vae_scale_factor: int,
+    ) -> tuple[npt.NDArray[np.float32], npt.NDArray[np.float32]]:
+        height = 2 * (int(height) // (vae_scale_factor * 2))
```
KCaverly, line 991: The values appear to already be calculated as latent_height, latent_width earlier in new_context_blocking. Can we combine these? Maybe, just pass in the existing variables, directly into _prepare_latents?

max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -839,3 +843,522 @@ def _convert_image_mode(image: Image.Image, to_mode: str):  # noqa: ANN202
...
+        latents, latent_image_ids = self._prepare_latents(
+            request.num_images_per_prompt,
+            num_channels_latents,
+            height,
+            width,
+            request.seed,
+            vae_scale_factor,
+        )
+
+        guidance: npt.NDArray[np.float32] | None = None
```
KCaverly, line 1320: This value, appears to be static given the transformer_config. Can we initialize this once in the __init__ and pass it to the PixelContent as needed?

KCaverly (commented):
Great work! We are getting there, just added a few comments.

max/python/max/interfaces/pipeline_variants/pixel_generation.py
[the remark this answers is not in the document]
KCaverly: For now, we can move forward with the existing request changes in this PR as is. We are likely going to rewrite/simplify the PixelGenerationRequest with the serve changes, so we can tackle cleanup then.

KCaverly:
!sync

max/python/max/pipelines/lib/tokenizer.py
KCaverly: Why we do this:
- Z-Image uses inverted timestep normalization compared to Flux

To manage this, I am updating this to use the Pipeline class name instead of the repo id, to allow for model variants.

max/python/max/pipelines/lib/diffusion_schedulers/scheduling_flow_match_euler_discrete.py
tolgacangoz: [It is just a placeholder.](https://github.com/modular/modular/pull/5816#:~:text=Scheduler/Sampler/Solver%20of%20diffusion%20models%20is%20a%20placeholder%20for%20now.%20It%20is%20going%20to%20be%20added%20in%20another%20PR%3A%20%5BMAX%5D%20Add%20scheduler%20and%20image%20processor%20implementation%20for%20Flux.1%20pipeline%C2%A0%235815)

max/python/max/pipelines/lib/diffusion_schedulers/scheduler_factory.py
tolgacangoz: I put that file there just as a placeholder/example and assumed that it would be beautified [later](https://github.com/modular/modular/pull/5815).

max/python/max/pipelines/lib/diffusion_schedulers/scheduling_flow_match_euler_discrete.py
tolgacangoz: IMO, this PR will add this scheduler properly: https://github.com/modular/modular/pull/5815

max/python/max/pipelines/lib/diffusion_schedulers/scheduling_flow_match_euler_discrete.py
tolgacangoz: Of course. IMO, let's move these discussions into this PR: https://github.com/modular/modular/pull/5815

max/python/max/pipelines/core/context.py
tolgacangoz: Hmm, actually, a coding agent named it like that first, but then I changed it; thus, it knew something :thinking:.

max/python/max/pipelines/core/context.py
tolgacangoz: https://github.com/modular/modular/pull/5816#discussion_r2736361546

max/python/max/pipelines/lib/tokenizer.py
tolgacangoz: [Flux.1 uses negative prompt(s).](https://github.com/huggingface/diffusers/blob/2c669e84808f40ac5b5026fa44e948cfbc325119/src/diffusers/pipelines/flux/pipeline_flux.py#L658-L659)

[We thought that](https://github.com/modular/modular/pull/5850#discussion_r2730374553) we should add `messages` and `chat_template_options` into `PixelGenerationRequest`, similar to `TextGenerationRequest`, since [Flux.2 has potentially customizable `system_message`s](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/flux2/system_messages.py), right @kkimmk?
I am now arranging the parsing of these two from the request, just like in `TextTokenizer`.

max/python/max/pipelines/lib/tokenizer.py
tolgacangoz: [Since `prompt` and `messages` are being included in the `PixelGenerationRequest`](https://github.com/modular/modular/pull/5816#discussion_r2736455960), I will add raising errors similar to `TextGenerationRequest`.

tolgacangoz:
Thanks for merging and improving internally!

max/python/max/pipelines/lib/tokenizer.py
tolgacangoz: AFAIS, people just use terminology differently. Some people use moving from `timestep=1` to `timestep=0` as the denoising direction, where the clean image is reached. Other people, such as P. Thiel, define this process as the opposite, from `0` to `1`. Since this comes from their training, we should put an `if` there.

Issue #5671, opened 2025-12-16
[Feature Request] Migrate `kernels/src/quantization` off of `LegacyUnsafePointer`.


NathanSWard:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

See [parent issue](https://github.com/modular/modular/issues/5665)

The following kernel files should be migrated off of `LegacyUnsafePointer`. (complexity noted next to file)

```
quantization/per_channel_grouped_4bit.mojo (Medium)
quantization/qmatmul_gpu.mojo (Medium)
quantization/qmatmul_k.mojo (High)
quantization/qmatmul.mojo (High)
```

Dependencies:
- `qmatmul_k.mojo` is imported by `nn/kv_cache_ragged.mojo` and `Mogg/`

### What is your motivation for this change?

`LegacyUnsafePointer` is going to be deprecated and removed.

KrxGu:
Hi @NathanSWard, @JoeLoser  and Mojo Libraries team,

I’d also be interested in taking on the quantization kernel migration in #5671 as a follow-up once my CPU matmul work for #5670 is in good shape.

From the issue, I understand the scope as:

- `quantization/per_channel_grouped_4bit.mojo` (Medium)
- `quantization/qmatmul_gpu.mojo` (Medium)
- `quantization/qmatmul_k.mojo` (High)
- `quantization/qmatmul.mojo` (High)

with `qmatmul_k.mojo` being used by `nn/kv_cache_ragged.mojo` and Mogg code paths.

A couple of clarifications before I dive in:

1. Would you prefer a **single PR** that migrates all four quantization files to the new `UnsafePointer` API, or should I split this into **two PRs** (for example: `per_channel_grouped_4bit` + `qmatmul_gpu` first, then `qmatmul` + `qmatmul_k` as a second step given their higher complexity and downstream dependencies)?
2. For `qmatmul_gpu.mojo` and the GPU-specific paths inside `qmatmul*_` files, what’s the expected **testing strategy** for external contributors without direct access to the exact GPU environment(GPU POOR 😭)? Is “build + run existing tests that touch these kernels” sufficient, relying on CI / internal harnesses for deeper coverage?
[...]

PR #5817, opened 2026-01-21
[kernels] Migrate qmatmul_k.mojo to UnsafePointer v2 API

KrxGu:
This PR completes the migration of `max/kernels/src/quantization/qmatmul_k.mojo` from `LegacyUnsafePointer` to the new `UnsafePointer` v2 API.

## Changes
- Replaced all `LegacyUnsafePointer` instances with `UnsafePointer[T, origin]`
- Used `MutAnyOrigin` for all mutable pointer parameters across tile and column functions
- Updated function type signatures to use `OptionalReg` instead of `Optional`
- Added `OptionalReg` import from collections

## Implementation Notes
This migration was particularly challenging due to several issues:

1. **Origin and address space inference**: Initial attempts using `...` syntax failed because it unbinds both origin and address space, causing type mismatches when multiple pointers with different origins are passed to the same function.

2. **Parameter passing**: The `//,` separator requires all subsequent parameters to be passed by name once any keyword argument is used, which caused several "parameter passed both as positional and keyword" errors.

3. **Address space handling**: Early attempts required explicit `address_space_cast` calls, but the final solution uses `MutAnyOrigin` which defaults to `GENERIC` address space, eliminating the need for casts.

4. **Pointer conversion**: Converting between pointer types required explicit construction using `UnsafePointer[T, MutAnyOrigin](unsafe_from_address=Int(ptr))` rather than simple casts.

5. **Merge conflicts**: Rebasing on main required resolving conflicts where upstream had changed `Optional` to `OptionalReg`.

The final approach uses `MutAnyOrigin` throughout, which provides a concrete origin that accepts any mutable pointer source while maintaining type safety.

## Testing
- All builds pass with no errors
- All quantization tests pass successfully

Addresses #5671

### DIFF
--- a/max/kernels/src/nn/kv_cache_ragged.mojo
+++ b/max/kernels/src/nn/kv_cache_ragged.mojo
@@ -2519,9 +2519,14 @@ fn _qmatmul_gguf_quantized_common[
             output,
         )
     elif quantization_encoding == "q4_k":
+
         @parameter
         if elementwise_lambda_fn:
-            matmul_Q4_K[elementwise_lambda_fn=OptionalReg(elementwise_lambda_fn.value())](
+            matmul_Q4_K[
+                elementwise_lambda_fn = OptionalReg(
+                    elementwise_lambda_fn.value()
+                )
+            ](
                 hidden_state,
                 weight,
                 output,
@@ -2533,9 +2538,14 @@ fn _qmatmul_gguf_quantized_common[
                 output,
             )
     elif quantization_encoding == "q6_k":
+
         @parameter
         if elementwise_lambda_fn:
-            matmul_Q6_K[elementwise_lambda_fn=OptionalReg(elementwise_lambda_fn.value())](
+            matmul_Q6_K[
+                elementwise_lambda_fn = OptionalReg(
+                    elementwise_lambda_fn.value()
+                )
+            ](
                 hidden_state,
                 weight,
                 output,
--- a/max/kernels/src/nn/kv_cache_ragged.mojo
+++ b/max/kernels/src/nn/kv_cache_ragged.mojo
@@ -2522,9 +2522,7 @@ fn _qmatmul_gguf_quantized_common[
 
         @parameter
         if elementwise_lambda_fn:
-            matmul_Q4_K[
-                elementwise_lambda_fn = elementwise_lambda_fn
-            ](
+            matmul_Q4_K[elementwise_lambda_fn=elementwise_lambda_fn](
                 hidden_state,
                 weight,
                 output,
@@ -2539,9 +2537,7 @@ fn _qmatmul_gguf_quantized_common[
 
         @parameter
         if elementwise_lambda_fn:
-            matmul_Q6_K[
-                elementwise_lambda_fn = elementwise_lambda_fn
-            ](
+            matmul_Q6_K[elementwise_lambda_fn=elementwise_lambda_fn](
                 hidden_state,
                 weight,
                 output,
--- a/max/kernels/src/nn/kv_cache_ragged.mojo
+++ b/max/kernels/src/nn/kv_cache_ragged.mojo
@@ -2519,17 +2519,33 @@ fn _qmatmul_gguf_quantized_common[
             output,
         )
     elif quantization_encoding == "q4_k":
-        matmul_Q4_K[elementwise_lambda_fn=elementwise_lambda_fn](
-            hidden_state,
-            weight,
-            output,
-        )
+        @parameter
+        if elementwise_lambda_fn:
+            matmul_Q4_K[elementwise_lambda_fn=OptionalReg(elementwise_lambda_fn.value())](
+                hidden_state,
+                weight,
+                output,
+            )
+        else:
+            matmul_Q4_K[elementwise_lambda_fn=None](
+                hidden_state,
+                weight,
+                output,
+            )
     elif quantization_encoding == "q6_k":
-        matmul_Q6_K[elementwise_lambda_fn=elementwise_lambda_fn](
-            hidden_state,
-            weight,
-            output,
-        )
+        @parameter
+        if elementwise_lambda_fn:
+            matmul_Q6_K[elementwise_lambda_fn=OptionalReg(elementwise_lambda_fn.value())](
+                hidden_state,
+                weight,
+                output,
+            )
+        else:
+            matmul_Q6_K[elementwise_lambda_fn=None](
+                hidden_state,
+                weight,
+                output,
+            )
     else:
         raise Error(
             "Unsupported quantization encoding: ", quantization_encoding
--- a/max/kernels/src/nn/kv_cache_ragged.mojo
+++ b/max/kernels/src/nn/kv_cache_ragged.mojo
@@ -2523,9 +2523,7 @@ fn _qmatmul_gguf_quantized_common[
         @parameter
         if elementwise_lambda_fn:
             matmul_Q4_K[
-                elementwise_lambda_fn = OptionalReg(
-                    elementwise_lambda_fn.value()
-                )
+                elementwise_lambda_fn = elementwise_lambda_fn
             ](
                 hidden_state,
                 weight,
@@ -2542,9 +2540,7 @@ fn _qmatmul_gguf_quantized_common[
         @parameter
         if elementwise_lambda_fn:
             matmul_Q6_K[
-                elementwise_lambda_fn = OptionalReg(
-                    elementwise_lambda_fn.value()
-                )
+                elementwise_lambda_fn = elementwise_lambda_fn
             ](
                 hidden_state,
                 weight,
--- a/max/kernels/src/quantization/qmatmul_k.mojo
+++ b/max/kernels/src/quantization/qmatmul_k.mojo
@@ -233,91 +233,6 @@ struct _packed_bit_array[bit_width: Int, block_m: Int, block_n: Int]:
         else:
             constrained[False, "unsupported bit width"]()
 
-    @staticmethod
-    @always_inline
-    fn unpack_from_ptr[
-        *, zero_point: UInt8 = 0
-    ](
-        self_ptr: UnsafePointer[Self, ...],
-        var dst_ptr: UnsafePointer[mut=True, UInt8],
-    ):
-        """Unpacks the local storage to the supplied external buffer using a pointer.
-        """
-        __comptime_assert (Self._packed_stride % Self._simd_width) == 0
-
-        var bits_ptr = self_ptr[].bits.unsafe_ptr()
-
-        @parameter
-        if Self.bit_width == 4:
-            __comptime_assert zero_point == 0, "zero point not implemented"
-            Self._unpack_int4_static(bits_ptr, dst_ptr)
-        elif Self.bit_width == 6:
-            Self._unpack_int6_static[zero_point](bits_ptr, dst_ptr)
-        else:
-            constrained[False, "unsupported bit width"]()
-
-    @staticmethod
-    @always_inline
-    fn _unpack_int4_static(
-        var bits_ptr: UnsafePointer[UInt8, ...],
-        var dst_ptr: UnsafePointer[mut=True, UInt8],
-    ):
-        __comptime_assert Self.bit_width == 4
-        __comptime_assert (Self.block_m % (2 * Self._tuple_width)) == 0
-
-        for _ in range(0, Self.block_m, 2 * Self._tuple_width):
-
-            @parameter
-            for col in range(Self._tile_n):
-                var packed_bits = bits_ptr.load[width = Self._simd_width]()
-                bits_ptr += Self._simd_width
-
-                @parameter
-                for i in range(2):
-                    var bytes = (packed_bits >> (i * 4)) & 15
-                    (dst_ptr + i * Self._packed_stride).store(bytes)
-
-                dst_ptr += Self._simd_width
-
-            dst_ptr += Self._packed_stride
-
-    @staticmethod
-    @always_inline
-    fn _unpack_int6_static[
-        zero_point: UInt8
-    ](
-        var bits_ptr: UnsafePointer[UInt8, ...],
-        var dst_ptr: UnsafePointer[mut=True, UInt8],
-    ):
-        __comptime_assert Self.bit_width == 6
-        __comptime_assert (Self.block_m % (4 * Self._tuple_width)) == 0
-
-        for _m in range(0, Self.block_m, 4 * Self._tuple_width):
-            var dst_col_ptr = dst_ptr
-
-            @parameter
-            for col in range(Self._tile_n):
-                var hi_bytes = SIMD[DType.uint8, size = Self._simd_width](0)
-
-                @parameter
-                for i in range(3):
-                    var packed_bits = bits_ptr.load[width = Self._simd_width]()
-                    bits_ptr += Self._simd_width
-
-                    (dst_col_ptr + i * Self._packed_stride).store(
-                        (packed_bits & 63) - zero_point,
-                    )
-
-                    hi_bytes |= (packed_bits >> 6) << (i * 2)
-
-                (dst_col_ptr + 3 * Self._packed_stride).store(
-                    hi_bytes - zero_point,
-                )
-
-                dst_col_ptr += Self._simd_width
-
-            dst_ptr += Self._packed_stride * 4
-
 
 struct _block_Q4_K_packed[block_n: Int = 1]:
     var base_scales: InlineArray[Float16, Self.block_n]
@@ -822,8 +737,8 @@ fn _matmul_group_unpacked[
     //,
     group_size: Int,
 ](
-    a_q_bits_ptr: UnsafePointer[Int8, ...],
-    mut b_q_bits_ptr: UnsafePointer[UInt8, ...],
+    a_q_bits_ptr: UnsafePointer[Int8, ImmutAnyOrigin],
+    mut b_q_bits_ptr: UnsafePointer[UInt8, MutAnyOrigin],
     mut c_int32_group: _Accumulator[DType.int32, tile_m, tile_n, simd_width],
 ):
     """Streaming matrix multiplication where the B matrix has been unpacked to
@@ -1090,11 +1005,17 @@ fn _matmul_Q4_K_tile[
     elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None,
 ](
     a_ptr: UnsafePointer[
-        _block_Q8_K_packed[_block_Q4_K.group_size], MutAnyOrigin
+        _block_Q8_K_packed[_block_Q4_K.group_size], ImmutAnyOrigin
     ],
-    b_ptr: UnsafePointer[_block_Q4_K_packed[], MutAnyOrigin],
-    b_q_scales_and_mins_buf: UnsafePointer[UInt8, MutAnyOrigin],
-    c_ptr: UnsafePointer[Float32, MutAnyOrigin],
+    b_ptr: UnsafePointer[
+        _block_Q4_K_packed[], origin=_, address_space = AddressSpace.GENERIC
+    ],
+    b_q_scales_and_mins_buf: UnsafePointer[
+        UInt8, origin=_, address_space = AddressSpace.GENERIC
+    ],
+    c_ptr: UnsafePointer[
+        Float32, origin=_, address_space = AddressSpace.GENERIC
+    ],
     N: Int,
     accumulate: Bool,
     m: Int,
@@ -1169,8 +1090,12 @@ fn _matmul_Q4_K_columns[
     var a_ptr: UnsafePointer[
         _block_Q8_K_packed[_block_Q4_K.group_size], MutAnyOrigin
     ],
-    b_ptr: UnsafePointer[_block_Q4_K_packed[], MutAnyOrigin],
-    var c_ptr: UnsafePointer[Float32, MutAnyOrigin],
+    b_ptr: UnsafePointer[
+        _block_Q4_K_packed[], origin=_, address_space = AddressSpace.GENERIC
+    ],
+    var c_ptr: UnsafePointer[
+        Float32, origin=_, address_space = AddressSpace.GENERIC
+    ],
     M: Int,
     N: Int,
     accumulate: Bool,
@@ -1233,7 +1158,7 @@ fn _matmul_Q4_K_columns[
     @__copy_capture(b_tile_ptr, b_q_scales_and_mins_buf, b_q_bits)
     @always_inline
     fn process_rows[tile_m: Int](m: Int):
-        var b_q_bits_ptr = b_q_bits
+        var b_q_bits_ptr = b_q_bits.unsafe_origin_cast[MutAnyOrigin]()
 
         @parameter
         fn matmul_group_unpacked(
@@ -1332,10 +1257,14 @@ fn _matmul_Q6_K_tile[
     elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None,
 ](
     a_ptr: UnsafePointer[
-        _block_Q8_K_packed[_block_Q6_K.group_size], MutAnyOrigin
+        _block_Q8_K_packed[_block_Q6_K.group_size], ImmutAnyOrigin
     ],
-    b_ptr: UnsafePointer[_block_Q6_K_packed[], MutAnyOrigin],
-    c_ptr: UnsafePointer[mut=True, Float32, MutAnyOrigin],
+    b_ptr: UnsafePointer[
+        _block_Q6_K_packed[], origin=_, address_space = AddressSpace.GENERIC
+    ],
+    c_ptr: UnsafePointer[
+        Float32, origin=_, address_space = AddressSpace.GENERIC
+    ],
     N: Int,
     accumulate: Bool,
     m: Int,
@@ -1419,8 +1348,12 @@ fn _matmul_Q6_K_columns[
     var a_ptr: UnsafePointer[
         _block_Q8_K_packed[_block_Q6_K.group_size], MutAnyOrigin
     ],
-    b_ptr: UnsafePointer[_block_Q6_K_packed[], MutAnyOrigin],
-    var c_ptr: UnsafePointer[Float32, MutAnyOrigin],
+    b_ptr: UnsafePointer[
+        _block_Q6_K_packed[], origin=_, address_space = AddressSpace.GENERIC
+    ],
+    var c_ptr: UnsafePointer[
+        Float32, origin=_, address_space = AddressSpace.GENERIC
+    ],
     M: Int,
     N: Int,
     accumulate: Bool,
@@ -1482,7 +1415,7 @@ fn _matmul_Q6_K_columns[
     @__copy_capture(b_tile_ptr, b_q_bits)
     @always_inline
     fn process_rows[tile_m: Int](m: Int):
-        var b_q_bits_ptr = b_q_bits
+        var b_q_bits_ptr = b_q_bits.unsafe_origin_cast[MutAnyOrigin]()
 
         @parameter
         fn matmul_group_unpacked(
@@ -1530,8 +1463,12 @@ fn _matmul_Qb_K[
         elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type],
     ](
         var a_ptr: UnsafePointer[_block_Q8_K_packed[group_size], MutAnyOrigin],
-        b_ptr: UnsafePointer[b_type, MutAnyOrigin],
-        var c_ptr: UnsafePointer[Float32, MutAnyOrigin],
+        b_ptr: UnsafePointer[
+            b_type, origin=_, address_space = AddressSpace.GENERIC
+        ],
+        var c_ptr: UnsafePointer[
+            Float32, origin=_, address_space = AddressSpace.GENERIC
+        ],
         M: Int,
         N: Int,
         accumulate: Bool,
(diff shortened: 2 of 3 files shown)


NathanSWard:
Ty for opening this draft! I've been OOO the end of last week so I'll take a look at this on Monday!

KrxGu:
> Ty for opening this draft! I've been OOO the end of last week so I'll take a look at this on Monday!

Gotcha, happy weekend.

KrxGu:
> Ty for opening this draft! I've been OOO the end of last week so I'll take a look at this on Monday!

Any update?

NathanSWard:
I'm going to sync in the PR and debug internally! We'll see what I can figure out 👀

NathanSWard:
!sync

NathanSWard:
Hi @KrxGu - so I was able to figure out the issue!

So the issue was the 3rd argument for 
```mojo
fn _pack_block_Q4_K[block_n: Int](
    var src_ptr: UnsafePointer[mut=False, _block_Q4_K],
    stride: Int,
    mut dst_ptr: UnsafePointer[mut=True, _block_Q4_K_packed[block_n]], # <--- this one
):
```

The problem is that `dst_ptr` is a `mut dst_ptr` which means it's a _mutable reference_ to an `UnsafePointer`. (almost like a pointer-to-a-pointer in C-code). 
When the function was being called, it was being passed a `LegacyUnsafePointer` (from the `b_packed` `LayoutTensor`).
However, you are not able to convert between types when the argument is a mutable reference.

The solution here is one of two options:
1. Simple remove the `mut` of `mut dst_ptr: UnsafePointer[..]` and just make it `dst_ptr: UnsafePointer[..]`. Granted you can _only_ do this if the `_pack_block_Q4_K` function never modifies the passed in `dst_ptr` argument (as in changed what it is pointing to).
2. Before passing the pointer to the function, you'll have to convert it to a `UnsafePointer` first. So something like:
```mojo
var dst_ptr = b_packed.ptr.bitcast[_block_Q4_K_packed[block_n]]().as_unsafe_pointer()
# ^ convert the `LegacyUnsafePointer` from the LayoutTensor and turn it into a regular UnsafePointer

    for _kb in range(k_blocks):
        var src_n_ptr = src_ptr

        for _n in range(0, N, block_n):
            _pack_block_Q4_K[block_n](src_n_ptr, k_blocks, dst_ptr) # <-- now you can pass it in here!
```

Please let me know if this makes sense!

KrxGu:
> Hi @KrxGu - so I was able to figure out the issue!
> 
> So the issue was the 3rd argument for
> 
> ```mojo
> fn _pack_block_Q4_K[block_n: Int](
>     var src_ptr: UnsafePointer[mut=False, _block_Q4_K],
>     stride: Int,
>     mut dst_ptr: UnsafePointer[mut=True, _block_Q4_K_packed[block_n]], # <--- this one
> ):
> ```
> 
> The problem is that `dst_ptr` is a `mut dst_ptr` which means it's a _mutable reference_ to an `UnsafePointer`. (almost like a pointer-to-a-pointer in C-code). When the function was being called, it was being passed a `LegacyUnsafePointer` (from the `b_packed` `LayoutTensor`). However, you are not able to convert between types when the argument is a mutable reference.
> 
> The solution here is one of two options:
> 
> 1. Simple remove the `mut` of `mut dst_ptr: UnsafePointer[..]` and just make it `dst_ptr: UnsafePointer[..]`. Granted you can _only_ do this if the `_pack_block_Q4_K` function never modifies the passed in `dst_ptr` argument (as in changed what it is pointing to).
> 2. Before passing the pointer to the function, you'll have to convert it to a `UnsafePointer` first. So something like:
> 
> ```mojo
> var dst_ptr = b_packed.ptr.bitcast[_block_Q4_K_packed[block_n]]().as_unsafe_pointer()
> # ^ convert the `LegacyUnsafePointer` from the LayoutTensor and turn it into a regular UnsafePointer
> 
>     for _kb in range(k_blocks):
>         var src_n_ptr = src_ptr
> 
>         for _n in range(0, N, block_n):
>             _pack_block_Q4_K[block_n](src_n_ptr, k_blocks, dst_ptr) # <-- now you can pass it in here!
> ```
> 
> Please let me know if this makes sense!

Thanks @NathanSWard ! That makes perfect sense - I was treating `mut dst_ptr` as "pointer to mutable data" when it actually means "mutable pointer reference." 

Looking at the code, `dst_ptr` is only modified in the caller loop (`dst_ptr += 1`), not inside `_pack_block_Q4_K` itself. So Option 1 (removing `mut` from the parameter) should work perfectly.

Should I also use `.as_unsafe_pointer()` at the call sites, or will removing `mut` from the parameter signature handle the LegacyUnsafePointer → UnsafePointer conversion automatically?

Also, should I apply this same pattern to `src_ptr`? Currently it's `var src_ptr` which allows `src_ptr += stride` inside the function.

NathanSWard:
> Should I also use .as_unsafe_pointer() at the call sites, or will removing mut from the parameter signature handle the LegacyUnsafePointer → UnsafePointer conversion automatically?

I don't think `. as_unsafe_pointer()` will be needed, so no need to add it!

> Also, should I apply this same pattern to src_ptr? Currently it's var src_ptr which allows src_ptr += stride inside the function.

I believe you need to keep the `var` there in order to do `src_ptr += stride`, but the conversions will work correctly when calling the function since `var src_ptr` just means the pointer is passed by value into the function (and not reference).

KrxGu:
> > Should I also use .as_unsafe_pointer() at the call sites, or will removing mut from the parameter signature handle the LegacyUnsafePointer → UnsafePointer conversion automatically?
> 
> I don't think `. as_unsafe_pointer()` will be needed, so no need to add it!
> 
> > Also, should I apply this same pattern to src_ptr? Currently it's var src_ptr which allows src_ptr += stride inside the function.
> 
> I believe you need to keep the `var` there in order to do `src_ptr += stride`, but the conversions will work correctly when calling the function since `var src_ptr` just means the pointer is passed by value into the function (and not reference).

Made progress with your mut fix (17→15 errors). Remaining issues:

1. Unpack on rvalue (lines 1161, 1199, 1451): b_tile_ptr[].q_bits.unpack(...) fails with "invalid use of mutating method on rvalue". Can't use var binding due to ImplicitlyCopyable. What's the pattern for calling mutating methods on fields behind UnsafePointer?

2. Function param inference (lines 1176, 1218, 1429, 1470): _matmul_Q*_K_tile[tile_m, tile_n, simd_width, func, lambda] fails with "failed to infer parameter 'tile_m'". Tried positional, keyword, mixed - all fail. Were these working before?

also is there any other communication channel where we can discuss things in a better way and get this PR mergable?

KrxGu:
@NathanSWard would love to get answers for above questions.

NathanSWard:
> Unpack on rvalue (lines 1161, 1199, 1451): b_tile_ptr[].q_bits.unpack(...) fails with "invalid use of mutating method on rvalue". Can't use var binding due to ImplicitlyCopyable. What's the pattern for calling mutating methods on fields behind UnsafePointer?

You should just be able to dereference the pointer and then call the function e.g. `ptr[].function()`. 

> Function param inference (lines 1176, 1218, 1429, 1470): _matmul_Q*_K_tile[tile_m, tile_n, simd_width, func, lambda] fails with "failed to infer parameter 'tile_m'". Tried positional, keyword, mixed - all fail. Were these working before?

I'm not sure here, I think they were failing before yes just they were overtaken by the other compiler errors. Consider taking a peek at the `matmut_group_fn` parameter and see if the types match up with it now? It might be an issue they incorrect types is my guess.

KrxGu:
Hi Nathan,

Made progress on qmatmul_k.mojo (down from 17 to 9 errors). Here's where I'm stuck:

**What I did:**
- Removed `mut` from `dst_ptr` in pack functions
- Added `...` origin to helper function pointer params
- Used `MutAnyOrigin` in function type signatures
- Used `unsafe_origin_cast[MutAnyOrigin]().address_space_cast[AddressSpace.GENERIC]()` for pointer casts

**Remaining errors:**

**1. "non-trivial value cannot be copied from non-default address space" (3)**
```mojo
b_tile_ptr[].q_scales_and_mins.unpack(b_q_scales_and_mins_buf)
b_tile_ptr[].q_bits.unpack(b_q_bits)
```
Dereferencing `b_tile_ptr` and calling a method on the field fails because it's in a non-GENERIC address space.

**2. "failed to infer parameter 'origin'" (5)**
```mojo
_matmul_Q4_K_tile[1, tile_n, simd_width, matmul_group_packed, elementwise_lambda_fn](...)
```
The `...` in runtime params like `a_ptr: UnsafePointer[..., ...]` creates an origin param that can't be inferred at call sites.

**3. "__getitem__" address space mismatch (1)**
```mojo
a_ptr[].group_sums[g * tile_m + row]
```
Same address space issue when indexing InlineArray through a non-GENERIC pointer.

**Questions:**
1. For the unpack errors - copy to local first, or is there a way to make methods work across address spaces?
2. For origin inference - should I use `MutAnyOrigin` in runtime params instead of `...`?

Thanks

NathanSWard:
> For the unpack errors - copy to local first, or is there a way to make methods work across address spaces?

The issue for this is that when you make the argument point "unbound" via `b_ptr: UnsafePointer[_block_Q4_K_packed[], ...],` (note the `...`) this also unbind the `AddressSpace` meaning this pointer can be of any address space. The best way to fix this is to _not_ unbind the address space to keeps it then as `AddressSpace.GENERIC`.
You should just be able to leave the argument as `b_ptr: UnsafePointer[_block_Q4_K_packed[]]` - is there a specific reason you need to be unbinding it? 🤔

> For origin inference - should I use MutAnyOrigin in runtime params instead of ...?

Generally don't do the `...` as this will unbind everything (including the `AddressSpace`) leading to other issues.
The ways to go about this are
- Adding `mut=True` if it needs to be mutable.
- Or if that doesn't work - because the origin must be know, using `MutAnyOrigin`/`ImmutAnyOrigin`.

In your case for `_matmul_Q4_K_tile` the definition should look like this:
```mojo
@always_inline
fn _matmul_Q4_K_tile[
    tile_m: Int,
    tile_n: Int,
    simd_width: Int,
    //,
    matmul_group_fn: fn(
        a_ptr: UnsafePointer[Int8, ImmutAnyOrigin], # <--- I believe this can be immutable
        mut c_int32: _Accumulator[DType.int32, tile_m, tile_n, simd_width],
    ) capturing[_] -> None,
    elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None,
](
    # The below arguments may need `mut=True` if they are required to be mutable.
    a_ptr: UnsafePointer[_block_Q8_K_packed[_block_Q4_K.group_size]],
    b_ptr: UnsafePointer[_block_Q4_K_packed[]],
    b_q_scales_and_mins_buf: UnsafePointer[UInt8],
    c_ptr: UnsafePointer[Float32],
    N: Int,
    accumulate: Bool,
    m: Int,
    n: Int,
    is_last_k_iter: Bool,
):
```
Notably for the `a_ptr` in the `matmul_group_fn` you _need_ to specify an origin since this pointer is a part of the closure signature. So this is one of the few cases where you need to specify `ImmutAnyOrigin`.

For the `a_ptr`, `b_ptr`, `b_q_scales_and_mins_buf`, and `c_ptr` arguments, you should not being using `...`, but simply either adding `mut=True` if the pointer needs to be mutable.
I would be surprised if you needed to explicitly say `Mut/ImmutAnyOrigin` for the arguments (unless this function is being used directly as a kernel, in which case you'll need to spell out the `AnyOrigin`s)

KrxGu:
@NathanSWard @JoeLoser 

Finally after 2.5weeks of efforts now this PR is ready for review.
Failure of Build and Test on large-oss-linux is expected and is due to an issue unrelated to this PR.
Good To Merge!!

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -244,6 +243,91 @@ struct _packed_bit_array[bit_width: Int, block_m: Int, block_n: Int]:
         else:
             constrained[False, "unsupported bit width"]()
 
+    @staticmethod
+    @always_inline
+    fn unpack_from_ptr[
```
NathanSWard: I see these 3 helper function are unused, are we able to remove them if so?

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -644,7 +724,7 @@ fn _matmul_group_stream_x86[
         ]
     ) capturing[_] -> None,
 ](
-    a_q_bits_ptr: UnsafePointer[Int8],
+    a_q_bits_ptr: UnsafePointer[Int8, ...],
```
NathanSWard, line 629: **question**: Do these need the unbound `...`? Or is there a compilation error when it's left fully bound?

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1303,11 +1389,13 @@ fn _matmul_Q6_K_tile[
...
-    elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None,
+    elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None,
 ](
-    a_ptr: UnsafePointer[_block_Q8_K_packed[_block_Q6_K.group_size]],
-    b_ptr: UnsafePointer[_block_Q6_K_packed[]],
-    c_ptr: UnsafePointer[Float32],
+    a_ptr: UnsafePointer[
+        _block_Q8_K_packed[_block_Q6_K.group_size], MutAnyOrigin
+    ],
+    b_ptr: UnsafePointer[_block_Q6_K_packed[], MutAnyOrigin],
```
NathanSWard: **questoin**: Do a & b need to be mutable? Or will `ImmutAnyOrigin` suffice?

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1531,11 +1641,18 @@ fn _matmul_Qb_K[
         var task_n_count = block_range[1] * grain_size
 
         var a_packed_ptr = a_packed_base_ptr
-        var b_packed_ptr = b.ptr.bitcast[b_type]()
+        var b_packed_ptr = UnsafePointer[b_type, MutAnyOrigin](
+            unsafe_from_address=Int(b.ptr.bitcast[b_type]())
+        )
```
NathanSWard: I believe you can simplify this a bit with the following:
```suggestion
        var b_packed_ptr = b.ptr.bitcast[b_type]().as_unsafe_pointer()
```
(Ditto with the `unsafe_from_address=` usage below.

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1580,7 +1697,7 @@ fn _matmul_Qb_K[
 
 
 fn matmul_Q4_K[
-    elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None
+    elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None
```
NathanSWard: Is this change from `Optional` to `OptionalReg` needed?

NathanSWard (changes requested):
Thanks for plugging away and figuring this out! Left a few more comments, but great work so far :)

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: Done.

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: Partially addressed. Updated _matmul_group_unpacked to use concrete origins (ImmutAnyOrigin for a_q_bits_ptr, MutAnyOrigin for b_q_bits_ptr).

However, I kept ... (with bound address space) for most other function parameters because:

The top-level tensors have ... origins that propagate through the call chain
Converting everywhere would require extensive unsafe_origin_cast calls throughout
Used origin=_, address_space=AddressSpace.GENERIC pattern to bind address space while keeping origin flexible - this was necessary to fix "non-trivial value cannot be copied from a non-default address space" errors with .unpack() calls
The pattern now is: concrete origins where practical (like _matmul_group_unpacked), bound address space with flexible origin elsewhere.

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: Applied where practical. Changed _matmul_group_unpacked's a_q_bits_ptr to ImmutAnyOrigin (it's read-only). Note that b_q_bits_ptr uses MutAnyOrigin because the pointer itself is incremented inside the function (line 782: b_q_bits_ptr += simd_width * 4).

For other functions, kept flexible origins with bound address space due to the propagation issue mentioned above.

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: Done.

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: Done.

KrxGu:
@NathanSWard have made the required changes.

The failing `test_module` tests with "Illegal instruction" errors are unrelated to this PR. 
GoodToMerge!!

max/kernels/src/nn/kv_cache_ragged.mojo
```diff
@@ -2516,17 +2516,35 @@ fn _qmatmul_gguf_quantized_common[
...
+                hidden_state,
+                weight,
+                output,
+            )
+        else:
+            matmul_Q4_K[elementwise_lambda_fn=None](
+                hidden_state,
+                weight,
+                output,
+            )
```
NathanSWard, line 2535: These branches hopefully should not be needed. You should just be able to pass the `elementwise_lambda_fn` closure through. Was there a compilation error with how the code was before?

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -825,9 +820,9 @@ fn _apply_base_scales[
 fn _apply_zero_point_correction[
     group_count: Int, tile_m: Int, tile_n: Int, simd_width: Int
 ](
-    a_group_sums_ptr: UnsafePointer[Int16],
-    b_q_mins_ptr: UnsafePointer[UInt8],
-    b_base_mins_ptr: UnsafePointer[Float16],
+    a_group_sums_ptr: UnsafePointer[Int16, ...],
```
NathanSWard, line 792: Again asking about if the `...` syntax here is necessary? (For all instances in this file that you added).
It ideally shouldn't be required, but are you getting compilation errors if its left in?

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1063,12 +1058,20 @@ fn _matmul_Q4_K_tile[
...
+    ],
+    b_ptr: UnsafePointer[
+        _block_Q4_K_packed[], origin=_, address_space = AddressSpace.GENERIC
+    ],
+    b_q_scales_and_mins_buf: UnsafePointer[
+        UInt8, origin=_, address_space = AddressSpace.GENERIC
+    ],
+    c_ptr: UnsafePointer[
+        Float32, origin=_, address_space = AddressSpace.GENERIC
+    ],
```
NathanSWard, line 1018: Was the previous syntax not working? E.g. hopefully this should just work
```mojo
c_ptr: UnsafePointer[Float32],
``
No need to do `origin=_` or `address_space=Generic`.

File: max/kernels/src/quantization/qmatmul_k.mojo [hunk 9]
```diff
@@ -1209,7 +1216,7 @@ fn _matmul_Q4_K_columns[
     @__copy_capture(b_tile_ptr, b_q_scales_and_mins_buf, b_q_bits)
     @always_inline
     fn process_rows[tile_m: Int](m: Int):
-        var b_q_bits_ptr = b_q_bits.as_any_origin().as_legacy_pointer()
+        var b_q_bits_ptr = b_q_bits.unsafe_origin_cast[MutAnyOrigin]()
```
NathanSWard, line 1161:
```suggestion
        var b_q_bits_ptr = b_q_bits.as_any_origin()
```

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1252,8 +1263,8 @@ fn _matmul_group_packed_Q6_K[
     *,
     zero_point: UInt8,
 ](
-    a_q_bits_ptr: UnsafePointer[Int8],
-    mut b_q_bits_ptr: UnsafePointer[UInt8],
+    a_q_bits_ptr: UnsafePointer[Int8, ...],
```
NathanSWard, line 1208: Ditto question about needing the `...` here.

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1303,11 +1314,17 @@ fn _matmul_Q6_K_tile[
             DType.int32, tile_m, tile_n, simd_width
         ],
     ) capturing[_] -> None,
-    elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None,
+    elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None,
 ](
-    a_ptr: UnsafePointer[_block_Q8_K_packed[_block_Q6_K.group_size]],
-    b_ptr: UnsafePointer[_block_Q6_K_packed[]],
-    c_ptr: UnsafePointer[Float32],
+    a_ptr: UnsafePointer[
```
NathanSWard, line 1259: Ditto about weather you can just write `UnsafePointer[_block_Q8_K_packed[_block_Q6_K.group_size]]` or if that doesn't not work?

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1303,11 +1314,17 @@ fn _matmul_Q6_K_tile[
             DType.int32, tile_m, tile_n, simd_width
         ],
     ) capturing[_] -> None,
-    elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None,
+    elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None,
```
NathanSWard, line 1257:
```suggestion
    elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None,
```

Why was this changed from `Optional` to `OptionalReg`? Unless its necessary can we please revert this change.

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1218,12 +1225,16 @@ fn _matmul_Q4_K_columns[
...
                 a_ptr, b_q_bits_ptr, c_int32_group
             )
 
         _matmul_Q4_K_tile[
-            matmul_group_unpacked, elementwise_lambda_fn=elementwise_lambda_fn
+            tile_m=tile_m,
+            tile_n=tile_n,
+            simd_width=simd_width,
+            matmul_group_fn=matmul_group_unpacked,
+            elementwise_lambda_fn=elementwise_lambda_fn,
```
NathanSWard, line 1179: I see these parameter callsties were changed. Is this necessary? Ideally it shouldn't be. The previous code should _hopefully_ work 🤔 
Was there an issue you ran into here?

max/kernels/src/quantization/qmatmul_k.mojo
```diff
@@ -1495,7 +1542,7 @@ fn _matmul_Qb_K[
     ) capturing -> None,
     *,
     interleave_group_sums: Bool = False,
-    elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None,
+    elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None,
```
NathanSWard, line 1480: Ditto: why was this changed to `OptionalReg`? Was that needed?

NathanSWard (changes requested):
Just reviewed again.
Apologies for the churn on this.
If you would feel more comfortable, you can simply rebase on the current `main` and then I can fixup the comments I posted internally so we could push this PR though faster. 

LMK if you want to address my comments, or lmk if you just want to rebase and then allow me to fix them internally.

max/kernels/src/nn/kv_cache_ragged.mojo
KrxGu: Yes, there was a compilation error. The issue is that `matmul_Q4_K` and `matmul_Q6_K` accept `Optional[elementwise_epilogue_type]` in their public API, but internally they need to convert to `OptionalReg` because the internal `_matmul_Qb_K` function uses `OptionalReg.`

When I tried to pass Optional directly to _matmul_Qb_K, I got:
```bash
error: cannot implicitly convert 'Optional[elementwise_epilogue_type]' value to 'OptionalReg[elementwise_epilogue_type]'
```
The @ `parameter` if/else pattern allows compile-time conversion from `Optional` to `OptionalReg` inside `matmul_Q4_K/matmul_Q6_K.`

However, if you have a better way to handle this conversion, I'm happy to implement it!

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: I initially tried using concrete origins (`ImmutAnyOrigin`, `MutAnyOrigin`) as you suggested, but ran into multiple issues:

1. Address space errors: Got "non-trivial value cannot be copied from a non-default address space" when calling `.unpack()` methods
2. Origin conversion constraints: `unsafe_origin_cast` requires matching mutability (`mut = Self.mut`), making it impossible to convert from `...` (unknown mut) to concrete origins

My approach was:

1. Use `origin=_, address_space=AddressSpace.GENERIC` to bind address space (fixes `.unpack()` errors) while keeping origin flexible
2. Keep `...` where the parameters come from top-level tensors that have `...` origins

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: Will give it a shot.

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: Great suggestion! I used `.unsafe_origin_cast[MutAnyOrigin]()` because I thought I needed to explicitly specify `MutAnyOrigin` since the pointer gets incremented. But if `.as_any_origin()` works (which should give `MutAnyOrigin` for mutable pointers automatically), that's much cleaner.

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: This was automatic formatting by `mblack` - I didn't manually change the call format. The linter required these formatting changes to pass. If there's a specific format you prefer, I can configure mblack differently or apply manual formatting.

max/kernels/src/quantization/qmatmul_k.mojo
KrxGu: This is related to Comment 1. The internal _matmul_Qb_K function has:
```bash
elementwise_lambda_fn: OptionalReg[elementwise_epilogue_type] = None
```
When I kept the public functions with Optional and tried to pass it through, I got type mismatch errors because Optional can't implicitly convert to OptionalReg.

KrxGu:
> Just reviewed again. Apologies for the churn on this. If you would feel more comfortable, you can simply rebase on the current `main` and then I can fixup the comments I posted internally so we could push this PR though faster.
> 
> LMK if you want to address my comments, or lmk if you just want to rebase and then allow me to fix them internally.

Hey @NathanSWard,

Thanks for the detailed review and the offer to handle this internally!

I think that might be the better path forward. A lot of my changes (the `origin=_, address_space=AddressSpace`.GENERIC stuff, `unsafe_origin_cast`, the `@ parameter` branches, etc.) were workarounds for compilation errors I hit, but they're probably not the cleanest solution.

I'm happy to rebase to current main so you can take it from here.

**One request though**: I've put a lot of time into this migration and would really appreciate understanding what went wrong with my approach. Could we do a quick call sometime after this is merged? I'd love to learn:

- Why the simpler syntax should work (when I kept hitting errors)
- The right way to handle Optional/OptionalReg conversions
- Best practices for the origin system

This would help me a ton for future contributions.
Let me know when you want me to rebase!

NathanSWard:
> This would help me a ton for future contributions.
Let me know when you want me to rebase!

I can do my best to try and plan something!
I'm a bit swamped with other work right now as well as a number of other OSS contributions I'm doing my best to juggle.
However, if you have the changes and rebase, I can get this merged in and we can try to set up some time so I can further explain questions you might have!

As you can tell, trying to explain stuff over Github can sometimes be difficult.

KrxGu:
> > This would help me a ton for future contributions.
> > Let me know when you want me to rebase!
> 
> I can do my best to try and plan something! I'm a bit swamped with other work right now as well as a number of other OSS contributions I'm doing my best to juggle. However, if you have the changes and rebase, I can get this merged in and we can try to set up some time so I can further explain questions you might have!
> 
> As you can tell, trying to explain stuff over Github can sometimes be difficult.

Hey @NathanSWard, I tried rebasing on latest main but hit some merge conflicts in qmatmul_k.mojo. It looks like there have been some updates to the pointer syntax in main (switching to UnsafePointer[mut=True, T] style and using __comptime_assert instead of comptime assert).

A few options:

1. I can manually resolve the conflicts and push
2. You could pull my current branch and handle the rebase/merge internally (might be easier since you know the latest changes)
3. I can squash my commits to make the rebase cleaner

What would work best for you? My branch is currently at fork/krxgu-quantization-high-unsafe-pointer with all the changes working and tests passing.

NathanSWard:
> What would work best for you? My branch is currently at fork/krxgu-quantization-high-unsafe-pointer with all the changes working and tests passing.

If you can manage to rebase/resolve the conflicts that wound be easiest! Since I cannot sync this into the mono repo if there are outstanding merge conflicts unfortunately. Hopefully it's not too hard, but once you get it rebased I'll sync it in ASAP to avoid any other issues.
Thanks!

KrxGu:
> > What would work best for you? My branch is currently at fork/krxgu-quantization-high-unsafe-pointer with all the changes working and tests passing.
> 
> If you can manage to rebase/resolve the conflicts that wound be easiest! Since I cannot sync this into the mono repo if there are outstanding merge conflicts unfortunately. Hopefully it's not too hard, but once you get it rebased I'll sync it in ASAP to avoid any other issues. Thanks!

I've rebased the branch onto latest main (83cd20634c) and resolved all merge conflicts. Also fixed the deprecated Int→UInt8 implicit conversion warnings that were causing Linux CI build failures.

The test_mlp.py failures appear to be unrelated infrastructure issues (same "Illegal instruction" errors seen in earlier test_module runs) - the quantization-specific tests pass successfully.

Good To Merge!!

NathanSWard:
Thank you @KrxGu - happy to finally land this one! And thank you for being patient with me, I greatly appreciate it!

NathanSWard:
!sync

Issue #5772, opened 2026-01-13
[Tracking] Linear Attention & State Space Model Support for MAX 


ulmentflam:
# Linear Attention & State Space Model Support for MAX

## Overview

This tracking issue monitors the implementation of linear attention and state space model (SSM) architecture support for MAX. The roadmap covers six model families unified by their sub-quadratic sequence processing: **Mamba (SSM)**, **Mamba2 (SSD)**, **Nemotron 3 (Hybrid Mamba2-Transformer-MoE)**, **Mamba3 (MIMO)**, **Gated Delta Networks**, and **Kimi Linear (KDA)**.

These architectures replace quadratic self-attention with linear-time recurrences, enabling efficient long-context inference with constant memory per token during decoding. MAX's kernel infrastructure is designed to maximize reuse across these families while exposing architecture-specific optimizations.

## Architecture Roadmap

- [x] Phase 1: Mamba (Selective SSM)
- [ ] Phase 2: Mamba2 (State Space Duality / SSD) **<-- Next**
- [ ] Phase 3: Nemotron 3 Nano & Super (Hybrid)
- [ ] Phase 4: Mamba3 (MIMO + Complex State)
- [ ] Phase 5: Gated Delta Networks
- [ ] Phase 6: Kimi Linear (KDA)

**Ordering rationale:**

- Mamba2 reuses most Mamba1 infrastructure (causal conv1d, RMSNorm) and adds chunk-based SSD, making it the lowest incremental effort
- Nemotron 3 is a hybrid architecture that interleaves Mamba2 layers with GQA attention and MoE. Once Mamba2 kernels land, bringup is primarily pipeline integration with existing MAX attention and MoE infrastructure
- Mamba3 extends Mamba2 with MIMO and complex state, building directly on Phase 2
- GDN introduces the delta rule with gating, a fundamentally different recurrence that shares chunk-parallel structure with earlier phases
- Kimi Linear extends GDN's gating from head-wise to channel-wise, building on Phase 5

---

## Phase 1: Mamba (Selective State Space Models)

Full Mamba1 architecture support with selective scan, causal conv1d, variable-length sequences, and fused kernels. Production-ready pipeline with prefill/step separation and slot-based SSM state caching.

### Kernels (Mojo)

- [x] `causal_conv1d` kernel + ops (#5820)
- [x] `selective_scan` kernel + ops (#5821)
- [x] Variable-length sequence kernels (#5822)
- [x] Fused kernels + normalization + ops (#5827)

[...]

PR #5820, opened 2026-01-22
[Kernels] Add causal conv1d kernel for Mamba SSM Architecture

ulmentflam:
## Summary
  - Introduces `state_space` kernel package with causal 1D convolution for Mamba architectures
  - Adds CPU and GPU kernel implementations with channel-first/last layouts and SiLU activation
  - Includes compiler-registered ops (`CausalConv1D`, `CausalConv1DUpdate`) for graph integration
  - Adds CPU and GPU kernel tests

  Part of #5772

  ## Test plan
  - [x] `./bazelw run //bazel/lint:lint` passes
  - [x] `./bazelw build //max/kernels/src/state_space/...` passes
  - [x] `./bazelw test //max/kernels/test/state_space/...` passes (2/2 CPU tests)
  - [x] `./bazelw test //max/kernels/test/gpu/state_space/...` passes (2/2 GPU tests)


ulmentflam:

max/kernels/src/state_space/__init__.mojo
```diff
@@ -0,0 +1,13 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2025, Qwerky AI Inc. All rights reserved.
```
BradLarson: Minor thing, but I think all header blocks need to match the same license name and style as is present in other files: https://github.com/modular/modular/blob/main/mojo/stdlib/std/benchmark/_progress.mojo#L1 or our linter will have a fit. It maintains a consistent licensing across the repository.

max/kernels/test/state_space/test_causal_conv1d_update.mojo
```diff
@@ -0,0 +1,377 @@
...
+    input_heap.free()
+    conv_state_heap.free()
+    conv_state_ref_heap.free()
+    weight_heap.free()
+    bias_heap.free()
+    result_fused_heap.free()
+    result_unfused_heap.free()
+
+
+def main():
```
BradLarson: One suggestion on the tests is that you might be able to use the TestSuite functionality: https://github.com/modular/modular/blob/main/max/kernels/test/layout/test_tile_tensor.mojo#L30 to do auto-discovery of specific tests, then group the below variants into those test cases. That way, you wouldn't need to have the print statement after each test grouping and could test each in isolation.

ulmentflam: Working on this now. This was a side effect of writing the tests in our custom module; I just ported them over directly.

BradLarson:
!sync

max/kernels/src/state_space/causal_conv1d.mojo
```diff
@@ -0,0 +1,3623 @@
...
+                    )
+                    conv_sum = conv_sum + Scalar[output_dtype](
+                        input_val * Scalar[x_dtype](weight_val)
+                    )
+
+            var out_offset: UInt32 = out_base + UInt32(l * out_l_stride)
+            var out_val: Scalar[output_dtype] = conv_sum
+            if silu_activation:
+                # SiLU: x * sigmoid(x) = x / (1 + exp(-x))
+                out_val = out_val / (1 + exp(-out_val))
```
BradLarson: As part of our unification of SIMD types, we've been adding requirements on `exp()` and a few other numerical functions to restrict them to floating-point types. When I pulled this in to our internal CI, this failed a constraint because `output_dtype` here is too open-ended. If you build this locally, you'll probably see something like:
```
error: invalid call to 'exp': lacking evidence to prove correctness
```
For the dtypes, in a free function like this you may be able to constrain this in the function signature using `where dtype.is_floating_point()`. For custom op bodies, you may need to attest to this using
```
__comptime_assert dtype.is_floating_point(), "dtype must be floating point"
```

ulmentflam: So, instead, I made the silu function require a floating-point dtype. Then I added a compiler parameter to branch silu if out_dtype is floating-point. This way, we maintain the same floating-point precision across the activation function, and if we run in Int quantization, we cast to float32 for now.

BradLarson:
!sync

BradLarson:
!sync

PR #5821, opened 2026-01-22
[Kernels] Add selective scan kernel for Mamba SSM Architecture

ulmentflam:
## Summary
  - Add selective_scan kernel (CPU and GPU implementations)
  - Add SelectiveScanFwd, SelectiveScanFwdMinimal, SelectiveScanUpdate op registrations
  - Add CPU and GPU unit tests

  Depends on #5820

  ## Test plan
  - [x] `./bazelw run //bazel/lint:lint` - Lint passes
  - [x] `./bazelw build //max/kernels/src/state_space:state_space` - Build passes
  - [x] `./bazelw test //max/kernels/test/state_space/...` - CPU tests pass (3/3)
  - [x] `./bazelw test //max/kernels/test/gpu/state_space/...` - GPU tests pass (3/3)


max/kernels/src/state_space/selective_scan.mojo
```diff
@@ -0,0 +1,3688 @@
...
+# ===----------------------------------------------------------------------=== #
+# Core selective scan kernel implementation
+# ===----------------------------------------------------------------------=== #
+# This file contains the core GPU and CPU kernel implementations for selective scan.
+#
+# The kernel functions `selective_scan_fwd_gpu` and `selective_scan_fwd_cpu` are
+# the actual kernels that perform the selective scan computation.
+#
+# For MAX framework integration, see MOGGKernelAPI.mojo.
+# ===----------------------------------------------------------------------=== #
```
BradLarson: Pretty minor, but the comment style at the top here is different than the one used for the causal_conv1d kernel. We could probably also drop the reference to the MOGGKernelAPI.mojo, since we're not incorporating these into that registry yet.

max/kernels/src/state_space/selective_scan.mojo
```diff
@@ -0,0 +1,3688 @@
...
+# Optimized sigmoid using fast approximation for large negative values
+fn sigmoid(val: Float32) -> Float32:
+    if val < -20.0:
+        return 0.0
+    var exp_neg = math.exp(-val)
+    return 1.0 / (1.0 + exp_neg)
+
+
+# Optimized SiLU (swish) activation
+fn silu(val: Float32) -> Float32:
```
BradLarson: Did you want to generalize to use the same `silu` as was used for `causal_conv1d`, or is this more of a special-case `fast_silu`?

max/kernels/src/state_space/selective_scan.mojo
```diff
@@ -0,0 +1,3688 @@
...
+    if val < -20.0:
+        return 0.0
+    var exp_neg = math.exp(-val)
+    return val / (1.0 + exp_neg)
+
+
+# GPU kernel function - must be at module scope for compile_function_checked
```
BradLarson: This seems like an extraneous placeholder comment.

max/kernels/src/state_space/selective_scan.mojo
```diff
@@ -0,0 +1,3688 @@
...
+    var curr_C_offset = UInt32(b) * C_b_stride + UInt32(group_id) * C_g_stride
+    var curr_z_offset = UInt32(b) * z_b_stride + UInt32(d) * z_d_stride
+    var curr_out_z_offset = (
+        UInt32(b) * out_z_b_stride + UInt32(d) * out_z_d_stride
+    )
+
+    # Bounds checking for tensor access
+    # Disabled for now - the kernel assumes valid tensor layouts from the framework
+    # TODO: Implement proper bounds checking based on actual tensor sizes
+    var u_total_size = UInt32.MAX
```
BradLarson: If these bounds checks aren't actually doing anything, do we want to remove their logic and conditionals for now?

max/kernels/src/state_space/selective_scan.mojo
ulmentflam: @BradLarson Yes! We can use the new generic one for causal_conv1d, though the reason this is separate is that the SSM ops must be in Float32. I'm always a big fan of resusability so I will eliminate some lines here.

max/kernels/src/state_space/selective_scan.mojo
```diff
@@ -0,0 +1,3688 @@
...
+                if b_offset >= B_total_size:
+                    return  # Invalid access - early exit
+                B_tile_15[i] = Scalar[kernel_dtype](B.ptr[b_offset]).cast[
+                    DType.float32
+                ]()
+
+        # Load C tiles - always use scalar loads to handle different layouts from slicing/reshaping
+        for i in range(TILE_SIZE):
+            var c_base = curr_C_offset + UInt32(i) * C_t_stride
+            if dstate > 0:
```
BradLarson: There's a lot of repeated code keying off of `dstate` here and below. I wonder if there's a way to make this more elegant to avoid the repetition. Similarly, if it could be lifted to a parameter, we could reduce the runtime overhead of the branching via `@parameter if`, but I don't know if that's feasible.

max/kernels/src/state_space/selective_scan.mojo
ulmentflam: I was thinking about this as I submitted this one. I'll take another pass through as I clean up.

max/kernels/src/state_space/selective_scan.mojo
ulmentflam: We can probably add the early termination logic to the causal_conv1d function. As you can tell, these functions were really completed at two different times, also starting with two different versions of Mojo.

max/kernels/src/state_space/selective_scan_ops.mojo
```diff
@@ -0,0 +1,727 @@
...
+                dt_lt,
+                A_lt,
+                B_lt,
+                C_lt,
+                D_lt,
+                z_lt,
+                dt_bias_lt,
+                UInt32(state_out_strides[0]),
+                UInt32(state_out_strides[1]),
+                UInt32(state_out_strides[2]),
```
BradLarson: Given that the strides are grouped in this way, I wonder if we can have typed arguments that take these in and cut down on the number of arguments going into these kernel functions.

BradLarson:
!sync

BradLarson:
!sync

BradLarson:
!sync

PR #5822, opened 2026-01-22
[Kernels] Add variable length causal conv1d and selective scan kernels for Mamba SSM Architecture

ulmentflam:
## Summary
  - Add varlen_causal_conv1d kernel for packed sequences
  - Add varlen_selective_scan kernel with cumulative sequence lengths
  - Add CPU and GPU unit tests

  Depends on #5821

  ## Test plan
  - [x] `./bazelw run //bazel/lint:lint` - Lint passes
  - [x] `./bazelw build //max/kernels/src/state_space:state_space` - Build passes
  - [x] `./bazelw test //max/kernels/test/state_space/...` - CPU tests pass (5/5)
  - [x] `./bazelw test //max/kernels/test/gpu/state_space/...` - GPU tests pass (5/5)

### DIFF
--- a/max/kernels/src/state_space/varlen_causal_conv1d.mojo
+++ b/max/kernels/src/state_space/varlen_causal_conv1d.mojo
@@ -1,0 +1,1007 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Causal Conv1D with variable length sequence support (vLLM interface).
+
+This module implements causal 1D convolution operations that support variable
+length sequences using cumulative sequence lengths (cu_seqlens), compatible
+with the vLLM inference interface.
+
+Key Functions:
+    - causal_conv1d_varlen_fwd: Forward pass for varlen sequences
+    - causal_conv1d_varlen_update: Update function for decode
+    - causal_conv1d_varlen_states: Extract states from varlen sequences
+
+vLLM Interface:
+    - x: (dim, cu_seq_len) for varlen - sequences concatenated left to right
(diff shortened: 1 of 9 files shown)


max/kernels/test/gpu/state_space/test_varlen_causal_conv1d.mojo
```diff
@@ -0,0 +1,1254 @@
...
+    conv_state_heap.free()
+    conv_state_cpu_heap.free()
+    conv_state_gpu_heap.free()
+    cache_seqlens_heap.free()
+    conv_state_indices_heap.free()
+    output_gpu_heap.free()
+    output_cpu_heap.free()
+
+
+def main():
```
BradLarson: It might be useful to update these tests to use the same TestSuite and discovery strategy as the other kernels that have gone in.

max/kernels/src/state_space/varlen_causal_conv1d.mojo
```diff
@@ -0,0 +1,1000 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+"""Causal Conv1D with variable length sequence support (vLLM interface).
```
BradLarson, line 13: Are the comments about the vLLM interface useful to denote the specific implementation, or were they brought across from the reference?

max/kernels/test/gpu/state_space/test_varlen_causal_conv1d.mojo
ulmentflam: I realized that as you were probably writing the comment!

max/kernels/src/state_space/varlen_causal_conv1d.mojo
ulmentflam: Yes, we wanted to keep the vLLM interface as the reference AND compatibility with the same interface. This way, in Python, we can reference a function that calls this graph op that has the exact same interface as the vLLM function that calls this graph op.

ulmentflam: I'm happy to drop the comment if you don't think it's relevant.

BradLarson: No, that sounds like good justification. Was just curious.

BradLarson:
!sync

BradLarson:
!sync

PR #5823, opened 2026-01-22
[Models] Add missing normalization to Qwen3 embeddings

sbrunk:
Small follow-up to #5782 to fix an issue with the Qwen 3 embeddings that I missed in the original PR. The crucial change is calling `norm` on the hidden states before returning them. The other changes are simplifications on the transformer layer as it is specialized for this case.

Applying the norm gets us **much** closer to PyTorch upstream which is reflected in the updated `cos_dist_threshold` values in the verification pipeline going down significantly from `0.24` to `0.00025`.

@k-w-w you might want to have a look as you also reviewed the original PR #5782.

### DIFF
--- a/max/python/max/pipelines/architectures/qwen3_embedding/layers/transformer.py
+++ b/max/python/max/pipelines/architectures/qwen3_embedding/layers/transformer.py
@@ -17,10 +17,9 @@ from __future__ import annotations
 from max.dtype import DType
 from max.graph import DeviceRef, TensorType, TensorValue, TensorValueLike, ops
 from max.nn.embedding import Embedding
-from max.nn.layer import Layer, LayerList, Module
+from max.nn.layer import LayerList, Module
 from max.nn.linear import Linear
 from max.nn.rotary_embedding import RotaryEmbedding
-from max.nn.transformer import ReturnHiddenStates
 
 
 class Qwen3EmbeddingTransformerBlock(Module):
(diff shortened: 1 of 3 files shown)


sbrunk:
Rebased against current main to make it work with the other fixes for model from e584a34d9af2d446142bd20c74f46dcb2a0c5184

It'd be great if the fix could make it into the 26.1 release, otherwise I'll be embarrassed for a while 😄

sbrunk:
I created another branch that just contains just the minimal change, a one line version of the fix: https://github.com/modular/modular/compare/main...sbrunk:modular:qwen3-embedding-fix-norm-minimal

sbrunk:
With 26.1 out, can I ask for a review on this fix, to get it into nightly?

sbrunk:
Rebased against main as `verify_pipelines` has changed.

Confirmed that the pipeline is still working and passing with the adjusted test command:
```bash
./bazelw run //max/tests/integration/accuracy:verify_pipelines -- --pipeline "Qwen/Qwen3-Embedding-0.6B-bfloat16" --devices gpu
```

sbrunk:
@k-w-w @tboerstad is there anything I could help with getting this fix into main?

I'm trying to keep the PR up-to-date, resolving merge conflicts that are popping up due to changes in `verify_pipelines` but if there's a way to get this merged it'd be much appreciated.

k-w-w (approved):
Sorry for the late review! Thank you so much for catching and fixing this, as well as the other code cleanup in this PR

k-w-w:
!sync

sbrunk:
Thanks @k-w-w :)

PR #5825, opened 2026-01-22
[Kernels] Migrate `kernels/src/nn` core ops off LegacyUnsafePointer (part 1)

### LINKED ISSUE #5675: [Feature Request] Migrate `kernels/src/nn` core ops off of `LegacyUnsafePointer`
### Review Mojo's priorities

### What is your request?

See [parent issue](https://github.com/modular/modular/issues/5665)

The following kernel files should be migrated off of `LegacyUnsafePointer`. (complexity noted next to file)

```
nn/broadcast.mojo (Medium)
nn/concat.mojo (High)
nn/conv_transpose.mojo
nn/conv.mojo (High)
nn/irfft.mojo (Medium)
nn/mla_graph.mojo (Medium)
nn/mla.mojo (High)
nn/moe.mojo (High)
nn/normalization.mojo (Medium)
nn/pad_gpu.mojo (High)
nn/pad.mojo (High)
nn/resize.mojo (Medium)
nn/softmax.mojo (Medium)
nn/topk_fi.mojo (Medium)
nn/topk.mojo (High)
nn/toppminp_gpu.mojo (High)
nn/toppminp.mojo (Medium)
```

`LegacyUnsafePointer` is being deprecated and removed.

### Any other details?

_No response_

brian-carroll:
Partially fix https://github.com/modular/modular/issues/5675

Already migrated on `main` [57f2f860](https://github.com/modular/modular/commit/57f2f860e509339dd8f4d805bba97c09c778036a):
- broadcast.mojo
- concat.mojo
- irfft.mojo
- normalization.mojo
- pad_gpu.mojo
- pad.mojo
- resize.mojo
- toppminp.mojo

Migrated in this PR:
- moe.mojo
- softmax.mojo
- topk_fi.mojo
- topk.mojo
- toppminp_gpu.mojo
- mla_graph.mojo
- mla.mojo

To migrate in [next PR](https://github.com/modular/modular/pull/5826):
- conv_transpose.mojo
- conv.mojo

### DIFF
--- a/max/kernels/src/nn/mla.mojo
+++ b/max/kernels/src/nn/mla.mojo
@@ -16,9 +16,6 @@ from collections import OptionalReg
 from math import ceildiv, recip
 from nn.mha_utils import DynamicInt
 from math.constants import log2e
-from memory import LegacyUnsafePointer
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
 from sys import (
     align_of,
     has_nvidia_gpu_accelerator,
@@ -277,7 +274,7 @@ fn flare_mla_decoding[
     var valid_length = LayoutTensor[
         DType.uint32, Layout.row_major(UNKNOWN_VALUE)
     ](
-        UnsafePointer[UInt32](),
+        UnsafePointer[UInt32, MutExternalOrigin](),
         RuntimeLayout[Layout.row_major(UNKNOWN_VALUE)].row_major(Index(0)),
     )
 
@@ -483,7 +480,9 @@ fn flare_mla_decoding_dispatch[
             decoding_warp_split_k=decoding_warp_split_k,
         ]
 
-        comptime nullptr = UnsafePointer[Scalar[accum_type]]()
+        comptime nullptr = UnsafePointer[
+            Scalar[accum_type], MutExternalOrigin
+        ]()
 
         var num_partitions_value: Int = 1
         var q_device = DeviceBuffer[q.dtype](ctx, q.ptr, q.size(), owning=False)
@@ -545,11 +544,11 @@ fn mla_decoding[
     _is_cache_length_accurate: Bool = False,
     decoding_warp_split_k: Bool = False,
 ](
-    q_ptr: UnsafePointer[Scalar[q_type]],
+    q_ptr: UnsafePointer[Scalar[q_type], MutAnyOrigin],
     k: k_t,
-    output_ptr: UnsafePointer[Scalar[output_type]],
-    exp_sum_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()]],
-    qk_max_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()]],
+    output_ptr: UnsafePointer[Scalar[output_type], MutAnyOrigin],
+    exp_sum_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()], MutAnyOrigin],
+    qk_max_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()], MutAnyOrigin],
     scale: Float32,
     batch_size: Int,
     num_partitions: Int,
@@ -709,11 +708,11 @@ fn mla_decoding_single_batch[
     use_score_mod: Bool = False,
     decoding_warp_split_k: Bool = False,
 ](
-    q_ptr: UnsafePointer[Scalar[q_type]],
+    q_ptr: UnsafePointer[Scalar[q_type], MutAnyOrigin],
     k: k_t,
-    output_ptr: UnsafePointer[Scalar[output_type]],
-    exp_sum_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()]],
-    qk_max_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()]],
+    output_ptr: UnsafePointer[Scalar[output_type], MutAnyOrigin],
+    exp_sum_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()], MutAnyOrigin],
+    qk_max_ptr: UnsafePointer[Scalar[get_accum_type[q_type]()], MutAnyOrigin],
     scale: Float32,
     num_keys: UInt,
     num_partitions: UInt,
@@ -1789,11 +1788,11 @@ fn mla_prefill[
     use_score_mod: Bool = False,
     _ndbuffer_mha_operand: Bool = False,
 ](
-    q_ptr: UnsafePointer[Scalar[q_type]],
+    q_ptr: UnsafePointer[Scalar[q_type], MutAnyOrigin],
     k: k_t,
     v: v_t,
     k_rope: k_rope_t,
-    output_ptr: UnsafePointer[Scalar[output_type]],
+    output_ptr: UnsafePointer[Scalar[output_type], MutAnyOrigin],
     scale: Float32,
     batch_size: Int,
     seq_len_arg: Int,
@@ -1918,11 +1917,11 @@ fn mla_prefill_single_batch[
     cache_depth: Int = 576,
     use_score_mod: Bool = False,
 ](
-    q_ptr: UnsafePointer[Scalar[q_type]],
+    q_ptr: UnsafePointer[Scalar[q_type], MutAnyOrigin],
     k: k_t,
     v: v_t,
     k_rope: k_rope_t,
-    output_ptr: UnsafePointer[Scalar[output_type]],
+    output_ptr: UnsafePointer[Scalar[output_type], MutAnyOrigin],
     scale: Float32,
     seq_len: Int,  # valid sequence length i.e. w/o padding.
     max_seq_len: Int,  # sequence length after padding.
--- a/max/kernels/src/nn/topk.mojo
+++ b/max/kernels/src/nn/topk.mojo
@@ -12,9 +12,7 @@
 # ===----------------------------------------------------------------------=== #
 
 from math import ceildiv, exp, iota
-from memory import LegacyUnsafePointer
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
+from memory import alloc
 from sys import align_of, simd_width_of, size_of, env_get_bool
 
 import gpu.primitives.warp as warp
@@ -144,10 +142,11 @@ fn top_k_shape_impl[
 
 @always_inline
 fn _adjust_top_p[
-    T: DType
+    T: DType,
+    address_space: AddressSpace = AddressSpace.GENERIC,
 ](
     top_p: Scalar[T],
-    values: UnsafePointer[Scalar[T], ...],
+    values: UnsafePointer[Scalar[T], MutAnyOrigin, address_space=address_space],
     k: Int,
     total_sum: Scalar[T],
 ) -> Scalar[T]:
@@ -450,7 +449,7 @@ fn fused_token_sampling_cpu[
     var out_vals_shape = input.runtime_layout.shape.value.canonicalize()
     out_vals_shape[input.rank - 1] = bound_max_k
     var out_vals = LayoutTensor[dtype, Layout.row_major[input.rank]()](
-        UnsafePointer[Scalar[dtype]].alloc(out_vals_shape.flattened_length()),
+        alloc[Scalar[dtype]](out_vals_shape.flattened_length()),
         RuntimeLayout[Layout.row_major[input.rank]()].row_major(out_vals_shape),
     )
 
@@ -593,7 +592,7 @@ fn _top_k_sampling[
     var out_idxs_tmp = LayoutTensor[
         DType.int64, Layout.row_major[internal_rank]()
     ](
-        UnsafePointer[Int64].alloc(out_vals.size()),
+        alloc[Int64](out_vals.size()),
         RuntimeLayout[Layout.row_major[internal_rank]()].row_major(
             internal_out_shape
         ),  # topk returns K as last dim
@@ -644,7 +643,7 @@ fn _top_k_sampling[
         # Calculate softmax normalization
         var max_val = internal_out_vals[batch, 0][0]
         var sum_exp = Scalar[dtype](0)
-        var exp_vals = UnsafePointer[Scalar[dtype]].alloc(k_val)
+        var exp_vals = alloc[Scalar[dtype]](k_val)
         var temp_val = temperature_val.cast[dtype]()
         for i in range(k_val):
             var val = internal_out_vals[batch, i][0]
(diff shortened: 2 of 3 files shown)


brian-carroll:

max/kernels/src/nn/conv_transpose.mojo
```diff
@@ -13,13 +13,6 @@
...
 from collections import OptionalReg
 from math import align_down, ceildiv
-from memory import LegacyUnsafePointer
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
-comptime OpaquePointer = LegacyUnsafePointer[
-    mut=True, NoneType, origin=MutAnyOrigin
-]
-
 from sys import align_of, simd_width_of
```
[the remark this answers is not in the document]
brian-carroll: It's part of the prelude
https://github.com/modular/modular/blob/main/mojo/stdlib/std/prelude/__init__.mojo#L167

max/kernels/src/nn/pad_gpu.mojo
```diff
@@ -10,9 +10,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
-from memory import LegacyUnsafePointer
-
-comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]
 from gpu import block_dim, block_idx, grid_dim, thread_idx
```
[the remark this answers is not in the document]
brian-carroll: It's part of the prelude
https://github.com/modular/modular/blob/main/mojo/stdlib/std/prelude/__init__.mojo#L178

max/kernels/src/nn/concat.mojo
```diff
@@ -159,7 +161,7 @@ fn _canonical_reshape_output[
     for i in range(1, len(inputs)):
         out_w += inputs[i].dim(axis)
     return _CanonicallyReshapedBuffer(
-        out_buf.ptr.bitcast[Int8](),
+        out_buf.ptr.bitcast[Int8]().unsafe_origin_cast[MutAnyOrigin](),
```
NathanSWard: When casting to an `AnyOrigin` you can use `as_any_origin()` instead of the unsafe casting :) 

```suggestion
        out_buf.ptr.bitcast[Int8]().as_any_origin(),
```

max/kernels/src/nn/concat.mojo
```diff
@@ -397,9 +401,13 @@ fn _concat_inner[
     for i in range(len(inputs)):
         var buffer_len = inputs[i].size()
         memcpy_or_fuse[output.rank, dtype, epilogue_fn](
-            output.ptr.bitcast[Int8](),
+            rebind[UnsafePointer[Int8, MutAnyOrigin]](
+                output.ptr.bitcast[Int8]()
+            ),
```
NathanSWard: For these, you don't need to use `rebind`. `rebind` generally should be avoided when possible as it basically just does an explicit cast to the destination type.
I would recommend updated this to just call the constructor of `UnsafePointer` directly.

```suggestion
            UnsafePointer[Int8, MutAnyOrigin](output.ptr.bitcast[Int8]()),
```

max/kernels/src/nn/irfft.mojo
```diff
@@ -46,14 +36,16 @@ from layout import LayoutTensor
 # This should eventually be moved to ffi.mojo with a more general global cache method
 # cache key is a string and cache value is a pointer.
 @always_inline
-fn global_cache_lookup(key: String) -> OpaquePointer:
-    return external_call["KGEN_CompilerRT_GetGlobalOrNull", OpaquePointer](
-        key.unsafe_ptr(), key.byte_length()
-    )
+fn global_cache_lookup(key: String) -> UnsafePointer[NoneType, MutAnyOrigin]:
+    return external_call[
+        "KGEN_CompilerRT_GetGlobalOrNull", UnsafePointer[NoneType, MutAnyOrigin]
```
NathanSWard: For `UnsafePointer[NoneType, origin]` you can use the `OpaquePointer` alias instead.

```suggestion
        "KGEN_CompilerRT_GetGlobalOrNull", OpaquePointer[MutAnyOrigin]
```

max/kernels/src/nn/mla.mojo
```diff
@@ -483,7 +480,9 @@ fn flare_mla_decoding_dispatch[
             decoding_warp_split_k=decoding_warp_split_k,
         ]
 
-        comptime nullptr = UnsafePointer[Scalar[accum_type]]()
+        comptime nullptr = UnsafePointer[
+            mut=True, Scalar[accum_type], MutExternalOrigin
+        ]()
```
NathanSWard, line 485:
```suggestion
        comptime nullptr = UnsafePointer[
            Scalar[accum_type], MutExternalOrigin
        ]()
```
If you specify an origin explicitly e.g. `MutExternalOrigin`, you shouldn't need to explicitly specify the `mut` parameter, is it will be inferred :)

max/kernels/src/nn/pad.mojo
```diff
@@ -175,7 +176,10 @@ fn pad_constant[
     @parameter
     fn pad_constant_wrapper(
         output: UnsafePointer[
-            Scalar[dtype], address_space = AddressSpace.GENERIC, ...
+            mut=True,
+            Scalar[dtype],
+            address_space = AddressSpace.GENERIC,
+            ...,
```
NathanSWard: These `...` shouldn't be needed anymore as `origin` will be automatically unbound by the compiler. We only need to specify these if we want to explicitly un-bind the `address_space` parameter as it has a defaulted valued.

```suggestion
```

max/kernels/src/nn/topk.mojo
```diff
@@ -1115,8 +1116,18 @@ fn _topk_stage2[
...
+        Scalar[T],
+        MutExternalOrigin,
+        address_space = AddressSpace.SHARED,
+    ]()
+    var s_id = UnsafePointer[
+        mut=True,
+        Int,
+        MutExternalOrigin,
+        address_space = AddressSpace.SHARED,
+    ]()
```
NathanSWard: To make this a bit more readable, I would recommend taking the `type_of(vals_sram)` as these are both assigned to offsets of that memory. It makes it a bit easier to read that these pointers will eventually point to offsets of that pointer.


```suggestion
    var s_val2 = type_of(vals_sram)()
    var s_id = type_of(vals_sram)()
```

max/kernels/src/nn/topk_fi.mojo
```diff
@@ -326,8 +321,12 @@ fn device_sampling_from_prob[
...
     prob_vec: SIMD[DType.float32, vec_size],
     aggregate: Float32,
-    sampled_id_sram: UnsafePointer[Int, address_space = AddressSpace.SHARED],
-    last_valid_id_sram: UnsafePointer[Int, address_space = AddressSpace.SHARED],
+    sampled_id_sram: UnsafePointer[
+        mut=True, Int, MutExternalOrigin, address_space = AddressSpace.SHARED
+    ],
+    last_valid_id_sram: UnsafePointer[
+        mut=True, Int, MutExternalOrigin, address_space = AddressSpace.SHARED
+    ],
```
NathanSWard, line 366: When taking an `UnsafePointer` as an argument, unless the function is taking ownership of that pointer (and expected to free it later on) we should avoid `MutExternalOrigin`. 
It would be easier to just specify these as `MutAnyOrigin` or `mut=True`.

```suggestion
    sampled_id_sram: UnsafePointer[
        Int, MutAnyOrigin, address_space = AddressSpace.SHARED
    ],
    last_valid_id_sram: UnsafePointer[
        Int, MutAnyOrigin, address_space = AddressSpace.SHARED
    ],
```

max/kernels/src/nn/toppminp_gpu.mojo
```diff
@@ -251,12 +248,18 @@ fn radix_sort_pairs_kernel[
     BLOCK_SIZE: Int = 256,  # found empirically
     NUM_BITS_PER_PASS: Int = 4,
 ](
-    input_keys_: UnsafePointer[Scalar[dtype]],  # modifies input
-    output_keys_: UnsafePointer[Scalar[dtype]],
-    input_key_ids_: UnsafePointer[Scalar[out_idx_type]],  # modifies input
-    output_key_ids_: UnsafePointer[Scalar[out_idx_type]],
+    input_keys_: UnsafePointer[
```
NathanSWard, line 252: Ditto here with not needing `mut=True` as it will be inferred from the explicit origin.

NathanSWard (changes requested):
Thank you for tackling these migrations, I greatly appreciate it!!
I left a round of feedback most in terms of usage of `rebind`, `mut=`, and some nuisances to origins.

After these get fixed up I'll be able to get this merged into the main repository. 😄

max/kernels/src/nn/concat.mojo
brian-carroll: Thanks, it looks like this became unnecessary after I started using UnsafePointer constructor as per your comment below.

max/kernels/src/nn/concat.mojo
brian-carroll: Thanks yeah it felt ugly but I was having trouble finding a way to get it to compile. The constructor seems obvious in hindsight, thanks!

brian-carroll:
Thanks for the feedback!
I hadn't expected you to review while still in draft, I was trying to get tests to pass. Although I might not have succeeded without your suggestions! I tried to apply all the suggestions to all files.
I can't run some of the tests locally on Mac, so if I still get failures on linux CI I might need to bisect by pushing a bunch of times. I rebased and broke it up into one commit per file to make that easier.

brian-carroll:
CI is failing with "illegal instruction" errors so I'm going to try bisecting it to find the first commit that fails.

brian-carroll:
First failing commit is the one where I migrate mla.mojo
Will try to debug

brian-carroll:
Through bisecting the git commits I found that the mla.mojo is causing the CI failures. There are callsites in mla_graph.mojo and kv_cache_ragged.mojo that probably need to be updated at the same time. However kv_cache_ragged.mojo is not currently listed in the same batch.
For now I'm thinking it might be best to leave mla.mojo out of this PR and get everything else merged. Then come back to it in a targeted follow-up PR.

brian-carroll:
I don't understand what's going on with the CI failures.
It is failing 7 tests that pass on a Lambda Cloud instance at the same git commit.
(ARM64 & H100)

```
ubuntu@192-222-57-1:~/modular$ git rev-parse HEAD
8f5df805000ea5776699f85d241c576ed5faccce
ubuntu@192-222-57-1:~/modular$ nvidia-smi
Sun Jan 25 23:34:47 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.195.03             Driver Version: 570.195.03     CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GH200 480GB             On  |   00000000:DD:00.0 Off |                    0 |
| N/A   32C    P0             92W /  700W |       1MiB /  97871MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
                                                                                         
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
ubuntu@192-222-57-1:~/modular$ ./bazelw test --cache_test_results=no \
  //max/tests/integration/API/python/nn/module_v3:rope/test_rope \
  //max/tests/integration/API/python/nn/module_v3:test_module \
  //max/tests/integration/pipelines/python/kv_cache/attention:attention_no_opaque_tests \
  //max/tests/integration/pipelines/python/nn:test_conv \
  //max/tests/integration/pipelines/python/nn:test_mlp \
  //max/tests/integration/pipelines/python/nn/norm:norm_tests \
  //max/tests/tests/nn:test_module
INFO: Invocation ID: 7596dc85-501d-4c16-949d-64b81fe05f97
INFO: Analyzed 7 targets (0 packages loaded, 62 targets and 18 aspects configured).
INFO: Found 7 test targets...
INFO: Elapsed time: 128.799s, Critical Path: 128.39s
INFO: 8 processes: 35 action cache hit, 1 internal, 7 linux-sandbox.
INFO: Build completed successfully, 8 total actions

Executed 7 out of 7 tests: 7 tests pass.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

brian-carroll:
I thought the MLA kernels were the problem so I took those commits out of the PR but it still doesn't pass even though it passes on Lambda Cloud as above.
The first 4 commits on their own do seem to pass CI but after that it fails.

NathanSWard:
@brian-carroll - are you able to rebase this on main? I can sync this into the internal repo and try to debug the issue you're having! :)

brian-carroll:
@NathanSWard done! ✅

NathanSWard:
!sync

NathanSWard:
@brian-carroll - after syncing this in, the PR passed all of our internal CI and tests 😄.
Was this expected or were you expecting it to also fail with the same issues you say on the OSS CI?

brian-carroll:
I'm not sure! I guess the question boils down to "is the bug real or not?"

The evidence we have is:

| | |
| ----------- | ------ |
| development | pass   |
| OSS CI      | fail   |
| Modular CI  | pass   |

So there are two possible scenarios:
1. The code is actually good. OSS CI is broken because it says it's bad.
2. The code is actually bad. Dev env and Modular CI are both broken because they say it's good.

OSS CI certainly seems like the odd one out. Maybe it has some special combination of OS, CPU and CPU that are required for the bug to show up? Or maybe there is a bug in that environment? I'm really not sure.

brian-carroll:
It turns out somebody else migrated a lot of these files already on main, so I had to rebase.
See updated PR description for details.

max/kernels/src/nn/topk.mojo
```diff
@@ -1217,7 +1216,11 @@ fn _topk_stage2[
         @parameter
         if sampling:
             # Storing the top-K logits in shmem for sampling
-            s_id = (idxs_sram + vals_smem_size).bitcast[Int]()
+            s_id = (
+                (idxs_sram + vals_smem_size).bitcast[Int]()
+                # comes from external_memory via vals_sram
+                .unsafe_origin_cast[MutExternalOrigin]()
```
brian-carroll: This unsafe_origin_cast might be the cause of the test failures.
idxs_sram is a bitcast of vals_sram which comes from a call to external_memory
From the names I would expect it to have origin MutExternalOrigin but it doesn't seem to.

brian-carroll: Actually I also notice that this line is adding `vals_smem_size` even though it was already added above at the definition. Could this be a bug or is it intended?
```mojo
    var idxs_sram = (vals_sram + vals_smem_size).bitcast[Int]()
```

NathanSWard:
!sync

PR #5827, opened 2026-01-22
[Kernels] Add RMSNorm fused residual and SSM combined kernels

ulmentflam:
## Summary

  - Adds `rms_norm_fused_residual` kernel with CPU and GPU implementations for Mamba blocks
  - Includes compiler-registered op (`RMSNormFusedResidual`) for graph integration
  - Adds SSM combined operation tests (ssd_combined, mamba_split_conv1d_scan_combined)

  The fused residual kernel performs RMSNorm on (input + residual) and outputs both the normalized result and
  pre-normalized sum, matching the pattern used in Mamba model blocks.

  Depends on #5821

  ## Test plan

  - [x] `./bazelw run //bazel/lint:lint` passes
  - [x] `./bazelw build //max/kernels/src/state_space/...` passes
  - [x] `./bazelw test //max/kernels/test/state_space/...` passes (5/5 CPU tests)
  - [x] `./bazelw test //max/kernels/test/gpu/state_space/...` passes (5/5 GPU tests)


BradLarson:
!sync

max/kernels/src/state_space/rms_norm_fused_residual.mojo
```diff
@@ -0,0 +1,699 @@
...
+    weight_offset: Scalar[dtype],
+    out_shape: IndexList[2],
+    dropout_p: Scalar[dtype] = Scalar[dtype](0.0),
+    seed: UInt64 = 0,
+):
+    """Core 2D implementation of RMSNorm with fused residual.
+
+    Uses simple (row, col) indexing to avoid compile-time evaluation issues.
+    """
+    __comptime_assert gamma.rank == 1, "gamma must have rank 1"
```
BradLarson: We recently deprecated `__comptime_assert` in favor of `comptime assert`, so these uses are failing on internal CI. They can all be safely migrated to `comptime assert`.

max/kernels/test/gpu/state_space/test_mamba_split_conv1d_scan_combined.mojo
```diff
@@ -0,0 +1,818 @@
...
+    var x_chunk_stride: UInt32 = 2 * dstate
+    var x_n_stride: UInt32 = 1
+    var out_z_b_stride: UInt32 = dim * seqlen
+    var out_z_d_stride: UInt32 = seqlen
+    var out_z_t_stride: UInt32 = 1
+    var dt_b_stride: UInt32 = nheads * seqlen
+    var dt_h_stride: UInt32 = seqlen
+    var dt_s_stride: UInt32 = 1
+    var A_stride: UInt32 = 1
+    var B_b_stride: UInt32 = ngroups * dstate * seqlen
```
BradLarson: I wonder if this long list of arguments can't be simplified. Many of these strides are the products of the same combinations of values. Could we pass these component values instead, and then calculate the resulting strides inside the kernel function, reusing similar products in the calculations? That could cut down a large number of lines here.

BradLarson:
!sync

max/kernels/test/state_space/test_ssd_combined.mojo
```diff
@@ -0,0 +1,329 @@
...
+    dim: Int,
+    seqlen: Int,
+    n_groups: Int,
+    rtol: Float64 = 0.01,
+) raises:
+    """Test SSD combined kernel against reference implementation."""
+    constrained[DSTATE <= MAX_DSTATE, "DSTATE exceeds kernel limit"]()
+    comptime dstate = DSTATE
+
+    var group_size = dim // n_groups
```
BradLarson, line 68: Minor thing, but it was pointed out internally that `group_size` is calculated here and other places and not used.

max/kernels/src/state_space/rms_norm_fused_residual.mojo
```diff
@@ -0,0 +1,699 @@
...
+        Scalar[dtype],
+        address_space = AddressSpace.SHARED,
+        alignment = align_of[SIMD[dtype, simd_width]](),
+        name="intermediate_shared_memory",
+    ]()
+    with PDL():
+        # First stage: apply dropout, add residual to input and store in shared memory
+        var tid = thread_idx.x
+        var row = block_idx.x
+        var idx = tid * UInt(simd_width)
```
BradLarson: It's been pointed out that this doesn't loop, where `_rms_norm_gpu_block_subkernel` does, so when `num_cols > block_dim * simd_width`, the subkernel's iterations beyond x=0 read uninitialized shared memory. This might not impact the Mamba model, but could be a potential problem with future models.

max/kernels/src/state_space/rms_norm_fused_residual.mojo
```diff
@@ -0,0 +1,699 @@
...
+
+        if idx < UInt(num_cols):
+            var input_val = input_fn[simd_width](Int(row), Int(idx))
+
+            # Apply dropout if enabled
+            var zero_scalar = Scalar[dtype](0.0)
+            if dropout_p > zero_scalar:
```
BradLarson: It was pointed out that the `dropout_p > 0` path here might not be tested in the current unit tests, due to this defaulting to 0.0.

max/kernels/test/state_space/test_ssd_combined.mojo
```diff
@@ -0,0 +1,329 @@
...
+    # Check a few sample outputs to verify kernel executed
+    var has_nonzero = False
+    var sample_size = min(10, batch * dim * seqlen)
+    for i in range(sample_size):
+        var val = Float32(output_h.ptr[i])
+        if abs(val) > 1e-8:
+            has_nonzero = True
+            break
+
+    if not has_nonzero:
```
BradLarson: This only checks for nonzero values, do we want a more robust numerical verification for the correctness of the calculations in this test?

BradLarson:
!sync

BradLarson:
!sync

max/kernels/src/state_space/rms_norm_fused_residual.mojo
```diff
@@ -0,0 +1,707 @@
...
+            seed,
+        )
+
+
+# ===----------------------------------------------------------------------=== #
+# Public API
+# ===----------------------------------------------------------------------=== #
+
+
+@register_internal("rms_norm_fused_residual")
```
BradLarson: Sorry, one last thing: `@register_internal` is used for marking MO ops, and we'll be using `rms_norm_fused_residual` as a custom op, so we probably want to remove this decorator.

ulmentflam: Darn, that was actually an artifact from when it was in normalization.mojo. I'll get it removed.

BradLarson:
!sync

BradLarson:
!sync

BradLarson:
!sync

PR #5829, opened 2026-01-23
[MAX] porting Gemma3MultiModal to V3 API

RossCampbellDev:
[DO NOT MERGE] this is a draft

for now, working with a single-GPU approach in a separate `gemma3multimodal_v3` module


tolgacangoz:
Thanks a bunch for working on this @RossCampbellDev!

I am currently porting the LTX-2 video-audio-gen pipeline. Its text encoder is `Gemma3ForConditionalGeneration`, and the V3 version of it is recommended. Could you also consider adding to accumulate the hidden states and output them optionally, similar to `transformers`:
```py
text_encoder_hidden_states = self.text_encoder(
    input_ids=text_input_ids, attention_mask=prompt_attention_mask, output_hidden_states=True
).hidden_states
```

RossCampbellDev:
> Thanks a bunch for working on this @RossCampbellDev!
> 
> I am currently porting the LTX-2 video-audio-gen pipeline. Its text encoder is `Gemma3ForConditionalGeneration`, and the V3 version of it is recommended. Could you also consider adding to accumulate the hidden states and output them optionally, similar to `transformers`:
> 
> ```python
> text_encoder_hidden_states = self.text_encoder(
>     input_ids=text_input_ids, attention_mask=prompt_attention_mask, output_hidden_states=True
> ).hidden_states
> ```

@tolgacangoz  hiya, I am working on an internal version of these changes [here](https://github.com/modularml/modular/pull/76890) (if you have access to that) and will be deleting this PR soon.  I'm only part way through so still some changes to come, but hopefully won't take a lot longer

I'm sorry, I am not following your comment regarding LTX-2 - it sounds like this would be changing the architecture implementation itself

Issue #5799, opened 2026-01-18
[Feature Request] Parametric traits


owenhilyard:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

The addition of parametric traits to Mojo, enabling traits to be parameterized with values to produce multiple implementations of a trait specialized on a type.

### What is your motivation for this change?

[...]

PR #5831, opened 2026-01-24
[Proposal] Allow users to trivially swap ADT implementations

akneni:
## Motivation

Today, swapping between collection implementations in Mojo often requires refactoring due to small but pervasive API differences. This discourages experimentation and prematurely locks users into concrete data structures.

This proposal introduces a **minimal set of abstract data type (ADT) traits** that define *semantic intent* (maps, sets, sequences) rather than implementation details, allowing implementations to be swapped **without changing user code**.

Early discussion on the forum was positive:
[https://forum.modular.com/t/adts-with-swappable-implementations/2604](https://forum.modular.com/t/adts-with-swappable-implementations/2604)

---

## What this PR does

* Proposes base ADT traits: `Map`, `Set`, and `Sequence`
* Introduces semantic extensions such as `SortedMap`, `ContiguousSequence`, and others. 
* Documents design principles, scope, and intended usage

This PR is a **design proposal only** and consists of a single document.

---

## What this PR does not do

* Does not modify existing stdlib collections
* Does not define performance guarantees
* Does not introduce capability or allocator traits

---

## Why this matters

A shared ADT vocabulary in the standard library reduces fragmentation, lowers the cost of performance tuning, and makes third-party data structures easier to adopt—while keeping dispatch static and explicit.

### DIFF
--- a/mojo/proposals/adt-traits.md
+++ b/mojo/proposals/adt-traits.md
@@ -1,0 +1,269 @@
+# Swappable Abstract Data Types (ADTs) in the Mojo Standard Library
+
+**Status**: Draft
+
+Author: Anish Kanthamneni
+
+Date: January 2026
+
+---
+
+## Summary
+
+This proposal introduces a small set of **base abstract data type (ADT) traits** in the Mojo standard library. These traits define *semantic categories of data structures* with stable core operations, allowing users to **swap implementations without refactoring code**.
+
+The goal is not to expose every possible capability, but to establish a **common, minimal vocabulary** for the most widely used ADT classes (maps, sets, sequences, etc.).
+
+---
+
+## Scope
+
+This proposal is scoped to:
+
+- **Base ADT classes only**
+- **Static (compile-time) dispatch by default**
+
+It explicitly does **not** attempt to:
+
+- Define all possible collection operations
+- Replace concrete collection types
+- Specify performance characteristics
+- Introduce capability or extension traits
+
+Those may be layered on later.
+
+---
+
+## Motivation
+
+In practice, developers frequently experiment with different data structure implementations to improve performance or memory behavior:
+- hash-based vs tree-based maps (ex, rust's HashMap vs BTreeMap)
+- standard library structure vs a third part tuned implementation (ex, rust's HashMap vs hashbrown::HashMap)
+- typical data structure vs thread safe variants (ex rust's HashMap vs dashmap::DashMap)
+
+In Mojo today (as in Rust and many other languages), switching between implementations often requires **non-trivial refactoring** because APIs differ slightly in naming, return values, or iteration behavior.
+
+This discourages experimentation and leads users to prematurely lock in choices.
+
+The proposed solution is to define **ADT Traits** that capture *semantic intent* or *sets of supported operations* rather than implementation details.
+
+---
+
+## Design principles
+1. **Strict Decoupling of Functionality and Implementation**
+    - Each base ADT trait represents a distinct kind of data structure with different invariants and operations.
+    - Base ADT Traits are not differentiated by performance, only by behavior.
+
+2. **Explicit ordering guarantees**
+   - Ordering semantics are part of the ADT’s identity, not optional behavior.
+
+3. **Minimalist Hierarchical Base APIs**
+    - In order for library authors to create functions that take ADT traits rather than concrete types, the ADT traits need to have all the functionality they want. Otherwise, they start using concrete types for their function signatures which subtracts from the value of this proposal. For this reason, a pure minimalist API base API (where the base ADT trait only requires the smallest subset of operations all implementations can perform) doesn't work. 
+    - A better approach is to define a hierarchy of ADT traits. For example, a `BTreeMap` can implement `.range_scan()` while a `HashMap` cannot. Rather than grouping them into the same ADT trait and leaving out the range scan, we will make `HashMap` implement the `Map` ADT trait and make `BTreeMap` implement the `SortedMap` ADT trait. Since `SortedMap` is a strict superset of `Map`, we can create a default trait implementation for `Map` for all types that implement `SortedMap`. 
+    - Since we also don't want an explosion of ADT trait hierarchies, we can allow functions without an observable effect (like `.reserve_capacity()`) to be implemented as no ops. This allows us to use this function on sequence types without having `ReservableSequence`, `ShrinkableSequence`, etc. etc. 
+
+## Vocabulary
+- **ADT Trait:** A trait that defines all operations that a category of ADTs should be able to implement. For example, a map ADT trait may define get, set, and delete operations. 
+- **ADT Trait Extensions:** An ADT trait that is a strict superset of another underlying ADT trait and creates a default implementation of the underlying ADT trait for every structure that implements it. 
+- **Unobservable Functions:** These are functions that do not affect the logical state of the data structure. For example, calling `.reserve_capacity()` on a sequence may perform an heap allocation but does not change the logical state of our sequence, so its considered to be an un observable function for the purposes of ADT Traits. Unobservable functions should have a no op default trait implementation. 
+
+---
+
+## Proposed ADT Traits
+
+### ADT Traits
+
+#### `Map[K, V]`
+Unordered key–value association.
+
+**Semantics**
+- Each key maps to at most one value
+- No guarantees about iteration order
+
+**Base operations**
+- `__init__() -> Self`
+- `__get__(k: K) -> Optional[ref V]`
+- `__set__(k: K, v: V)`
+- `delete(k: K) -> Bool`         # Return True if it deleted a value
+- `iter() -> Iterator[(ref K, ref V)]` (order unspecified)
+- `len() -> UInt` 
+
+**Default Trait Implementations**
+- `__init__(a: (K, V), b: (K, V), ...) -> Self`
+- `get_or_insert(k: K, v: V) -> ref V`
+- `get_or_insert_with(k: K, f: fn() -> V) -> ref V`
+- `contains(k: K) -> Bool`
+- `is_empty() -> Bool`
+- `update(k, f: fn(ref mut V) -> None) -> Bool`   # Returns True if a value was updated
+- `map[NewVType](f: Fn(k: ref K, v: ref V) -> NewVType) -> Self[K, NewVType]`
+- `map_inplace(f: Fn(k: ref K, v: ref mut V))`
+- `filter(f: Fn(k: ref K, v: ref V) -> Bool) -> Self[K, V]`
+- `filter_inplace(f: Fn(k: ref K, v: ref V) -> Bool)`
+- `reserve_capacity(cap: UInt)`
+- `shrink_to_fit()`
+
+#### `Set[T]`
+
+Unordered membership collection.
+
+**Semantics**
+- Each element appears at most once
+- No iteration order guarantees
+
+**Base operations**
+- `__init__() -> Self`
+- `add(item: T) -> Bool`      # Return True if item already exists
+- `delete(item: T) -> Bool`   # Return True if it deleted a value
+- `contains(item: T) -> Bool`
+- `iter() -> Iterator[ref T]`
+
+**Default Trait Implementations**
+- `filter(f: Fn(f: ref T) -> Bool) -> Self[T]`
+- `filter_inplace(f: Fn(f: ref T) -> Bool)`
+- `reserve_capacity(cap: UInt)`
+- `shrink_to_fit()`
+
+#### `Sequence[T]`
+Index-based sequence.
+
+**Semantics**
+- Elements are stored in a linear order
+- Supports indexed insertion
+- Implemented by List, LinkedList, Deque
+
+**Base operations**
+- `__init__() -> Self`
+- `get_safe(idx: UInt) -> Optional[ref T]`
+- `set_safe(idx: UInt, item: T) -> Bool`  # True if the operation was a success
+- `insert(item: T, idx: UInt)`
+- `extent_at(slice: Slice[T], idx: Uint)`
+- `pop_at(idx: Uint) -> T`
+- `iter() -> Iterator[ref T]`
+- `len() -> Uint`
+
+**Default Trait Implementations**
+- `__init__(a: T, b: T, ...) -> Self`
+- `__get__(idx: UInt) -> ref T`
+- `__set__(idx: UInt, item: T)`
+- `push(item: T)`
+- `push_front(item: T)`
+- `pop() -> T`
+- `first() -> Optional[ref T]`
+- `last() -> Optional[ref T]`
+- `map[NewType](f: Fn(t: T) -> NewType) -> Self[NewType]`
+- `map_inplace(f: Fn(f: ref mut T))`
+- `filter(f: Fn(f: ref T) -> Bool) -> Self[T]`
+- `filter_inplace(f: Fn(f: ref T) -> Bool)`
+- `extend_front(slice: Slice[T])`
+- `extend(slice: Slice[T])`
+- `is_empty() -> Bool`
+- `reserve_capacity(cap: UInt)`
+- `shrink_to_fit()`
+
+## ADT Trait Extensions
+
+#### `InsertionOrderedMap[K, V]`
+Insertion-ordered key–value association.
+
+**Semantics**
+- Preserves insertion order during iteration
+- Matches Python-style dict behavior
+- Extends `Map[K, V]`
+
+**Additional operations**
+- No additional operations. But by implementing this, the library author guarantees that their `.iter()` function returns elements in the order they were inserted. 
+
+---
+
+#### `SortedMap[K, V]`
+Key-ordered key–value association.
+
+**Semantics**
+- Iteration order is sorted by key
+- Ordering defined by `K: Comparable`
+- Extends `Map[K, V]`
+
+**Additional operations**
+- `range_scan(start: K, end: K) -> Iterator[(ref K, ref V)]`
+- By implementing this, the library author guarantees that their `.iter()` function returns elements in sorted order as is defined by `K: Comparable`
+
+---
+
+#### `InsertionOrderedSet[T]`
+Insertion-ordered membership collection.
+
+**Semantics**
+- Preserves insertion order during iteration
+- Extends `Set[T]`
+
+**Additional operations**
+- No additional operations. But by implementing this, the library author guarantees that their `.iter()` function returns elements in the order they were inserted. 
+
+
+---
+
+#### `SortedSet[T]`
+Ordered membership collection.
+
+**Semantics**
+- Iteration order is sorted
+- Ordering defined by `T: Comparable`
+- Extends `Set[T]`
+
+**Additional operations**
+- `range_scan(start: T, end: T) -> Iterator[ref T]`
+- By implementing this, the library author guarantees that their `.iter()` function returns elements in sorted order. 
+
+---
+
+#### `ContiguousSequence[T]`
+Index-based sequence whose elements are laid out contiguously in memory. 
+
+**Semantics**
+- Extends `Sequence[T]`
+
+**Additional Operations**
+- `as_slice() -> Slice[T]`
+- `to_unsafe_pointer() -> UnsafePointer[T]`
+
+
+## Interaction with concrete implementations
+Concrete types in `std.collections` (or equivalent) are expected to:
+- Implement one or more base ADT traits
+- Clearly document any extra guarantees
+
+Users are encouraged to write APIs generic over the base ADT rather than a specific implementation:
+```mojo
+fn foo[M: Map[String, Int]](m: M):
+    """
+    This function needs a type that has basic get and set operations
+    """
+
+fn bar[M: SortedMap[String, Int] + Sync](m: M):
+    """
+    This function needs a type that has get, set, and range_scan operations. 
+    It also needs the type to be thread safe. 
+    """
+```
+
+---
+
+## Why this belongs in the standard library
+Without a standardized base each library defines its own pseudo-Map trait and APIs fragment quickly. More importantly, a proposal like this is only feasible in the standard library. 
+
+---
+
+## Conclusion
+This proposal establishes a **small, stable foundation** for collection abstractions in Mojo. It deliberately prioritizes **semantic clarity and swappability** over completeness. There is also some precedent for this: Zig did something very similar with allocators, and it was so successful for them they decided to do it again with IO!
+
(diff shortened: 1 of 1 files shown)


mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+### ADT Traits
+
+#### `Map[K, V]`
+Unordered key–value association.
+
+**Semantics**
+- Each key maps to at most one value
+- No guarantees about iteration order
+
+**Base operations**
```
owenhilyard: All operations should probably take `self` as a parameter or have `out self`. Ideally, these should be valid Mojo function signatures. 

Point of Discussion: Implementations with Interior Mutability (Thread Safe Collections and more)

How do we want to accommodate thread safe data structures which have interior mutability (ex: Rust's `DashMap` or a `java.util.concurrent.ConcurrentMap`)? They would likely be able to take self by immutable reference for mutating operations. Similar applies to other ADTs. I could see the base trait requiring that mutability and then having an `InteriorMutableMap` (name quite WIP) which adds `fn __setitem__(read self, var k: K, var v: V)` and similar. I'm open to ideas for how to make the APIs work transparently, perhaps with a flag or set of flags on `Map` to enable the `read self` variants of mutating operations. However, this mean lead to `fn foo[M: Map[K, V]](..., m: ref [Some[Origin[mut=M.has_interior_mutability]]] M, ...)`, which might mean attaching an associated type for "The type of origin used to pass this map into something that does mutable operations on it". 

ex:

```mojo
from builtin.rebind import downcast

trait Map(...):
    comptime K: AnyType
    comptime V: AnyType

    comptime has_interior_mutability: Bool # = False once we can do that, most implementations never need to change this. 

    comptime DataMutateOriginType = Origin[mut=Self.has_interior_mutability]

    ...

    fn __setitem__(mut self, var k: Self.K, var v: Self.V): 
        ...

    fn __setitem__(read self, var k: Self.K, var v: Self.V) where Self.has_interior_mutability:
        ...

fn get_foo[M: Map](m: M) -> Int where _type_is_eq[M.K, String]() and _type_is_eq[M.V, Int]():
    return rebind[Int](m[rebind[M.K](String("foo"))])

fn set_foo[M: Map, origin: M.DataMutateOriginType](ref [origin] m: M, var v: M.V) where _type_is_eq[M.K, String]() and _type_is_eq[M.V, Int]():
    m.__setitem__(rebind_var[downcast[Movable, M.K]](String("foo"))^, v^) # needs `M.K(Movable)` to subtype `M.K`, which it currently doesn't. 

# less generic code
fn set_foo_mutref[M: Map](mut m: M, var v: M.V) where _type_is_eq[M.K, String]() and _type_is_eq[M.V, Int]():
    m.__setitem__(rebind_var[downcast[Movable, M.K]](String("foo"))^, v^) # needs `M.K(Movable)` to subtype `M.K`, which it currently doesn't. 
```

Not exactly the nicest API I've ever worked with, but just using `mut` should be fine for code where the map can be safely mutably borrowed. It's only code that is intended to run in a context where maps with interior mutability might show up, which means it's mostly only library authors or people who actually need it who will need to deal with it frequently. 

The other point this raises is that, without parametric traits ( #5799 ), this API is a gigantic mess from a UX perspective. Using any kind of `Map` will require using what are nominally stdlib-internal helper functions, and even if that is fixed, `fn set_foo_mutref[M: Map](mut m: M, var v: M.V) where M.K == String and M.V == Int:` is still a far cry from `fn set_foo_mutref(mut m: Some[Map[String, Int]], var v: M.V):`. This is why I have used parametric traits throughout my review so far and I think it's a soft dependency for making this ergonomic.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- **ADT Trait Extensions:** An ADT trait that is a strict superset of another underlying ADT trait and creates a default implementation of the underlying ADT trait for every structure that implements it. 
+- **Unobservable Functions:** These are functions that do not affect the logical state of the data structure. For example, calling `.reserve_capacity()` on a sequence may perform an heap allocation but does not change the logical state of our sequence, so its considered to be an un observable function for the purposes of ADT Traits. Unobservable functions should have a no op default trait implementation. 
+
+---
+
+## Proposed ADT Traits
+
+### ADT Traits
+
+#### `Map[K, V]`
```
owenhilyard, line 121: Type bounds on `K` and `V`? I think `AnyType` is fine for the ADT trait since I think we should be fairly permissive in what these foundational traits require.

Implementations may add other bounds, such as `Comparable`, `Hashable`, and `Movable`. You could, in theory, design a hash table that only works via copying data, if you wanted to avoid mutation as a solution to some kind borrow checking pains or threading issues.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+Unordered key–value association.
+
+**Semantics**
+- Each key maps to at most one value
+- No guarantees about iteration order
+
+**Base operations**
+- `__init__() -> Self`
+- `__get__(k: K) -> Optional[ref V]`
+- `__set__(k: K, v: V)`
```
owenhilyard: Do we want to make accommodations for read-only maps? The most realistic place I could see that showing up is for config off of JSON or YAML files in a multi-threaded application, where strings may be borrowed from a single buffer and for performance reasons access can't be synchronized. In that case, breaking out set and delete into a separate sub-trait (I agree they should be part of a `Map` ADT trait) although slightly annoying due to the lack of a "add required traits" lsp action, may be beneficial.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+#### `Map[K, V]`
+Unordered key–value association.
+
+**Semantics**
+- Each key maps to at most one value
+- No guarantees about iteration order
+
+**Base operations**
+- `__init__() -> Self`
+- `__get__(k: K) -> Optional[ref V]`
```
owenhilyard: This should be `__getitem__` and `__setitem__`. 

`__setitem__` needs to take ownership of `v`.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `__set__(k: K, v: V)`
+- `delete(k: K) -> Bool`         # Return True if it deleted a value
+- `iter() -> Iterator[(ref K, ref V)]` (order unspecified)
```
owenhilyard: I'm not sure if a `Tuple` is the best choice here. Using `MapPair` or another unique type may be helpful in specializing iterators for map operations as opposed to needing to implement them for general tuples of the shape `(T1, T2)`, where `.keys()` may make more sense than `.first_values()`. 

Also, what about owning iteration?

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `len() -> UInt` 
+
+**Default Trait Implementations**
+- `__init__(a: (K, V), b: (K, V), ...) -> Self`
```
owenhilyard: Maybe a default constructor from `DictLiteral` (which should probably be renamed to `MapLiteral`)?

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `get_or_insert(k: K, v: V) -> ref V`
+- `get_or_insert_with(k: K, f: fn() -> V) -> ref V`
+- `contains(k: K) -> Bool`
+- `is_empty() -> Bool`
+- `update(k, f: fn(ref mut V) -> None) -> Bool`   # Returns True if a value was updated
```
owenhilyard: Do you mean `fn(mut V) -> None`?

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+---
+
+## Proposed ADT Traits
+
+### ADT Traits
+
+#### `Map[K, V]`
+Unordered key–value association.
+
+**Semantics**
```
owenhilyard: An associated boolean which specifies whether the map provides pointer stability for values would likely be quite useful.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### `Set[T]`
+
+Unordered membership collection.
+
+**Semantics**
+- Each element appears at most once
+- No iteration order guarantees
+
+**Base operations**
```
owenhilyard: I think that `union` and `intersection` should be part of the set ADT, even if we provide default implementations.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `filter(f: Fn(k: ref K, v: ref V) -> Bool) -> Self[K, V]`
+- `filter_inplace(f: Fn(k: ref K, v: ref V) -> Bool)`
+- `reserve_capacity(cap: UInt)`
+- `shrink_to_fit()`
+
+#### `Set[T]`
+
+Unordered membership collection.
+
+**Semantics**
```
owenhilyard: Point of discussion: Should the documentation for `Set` explicitly ban probabilistic sets like Bloom Filters? 

Probabilistic sets require special care and handling, and I don't think it's reasonable to expect everything that takes a `Set` to deal with that. Having to accommodate bloom filters, especially negative bloom filters, in this API would also require a lot of changes.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- Each element appears at most once
+- No iteration order guarantees
+
+**Base operations**
+- `__init__() -> Self`
+- `add(item: T) -> Bool`      # Return True if item already exists
+- `delete(item: T) -> Bool`   # Return True if it deleted a value
```
owenhilyard: Same question as before, do we break out a `ReadOnlySet` type and require that `Set` implementations implement that?

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+Index-based sequence.
+
+**Semantics**
+- Elements are stored in a linear order
+- Supports indexed insertion
+- Implemented by List, LinkedList, Deque
+
+**Base operations**
+- `__init__() -> Self`
+- `get_safe(idx: UInt) -> Optional[ref T]`
```
owenhilyard: Should be `__getitem__`. 

Point of Discussion: The return type of `__getitem__`

There are corners of the Rust community which regret that Rust's `__getitem__` equivalent does not return an `Option`, but there are performance and ergonomics concerns. This problem was one of the motivations for my effect system proposal ( #3946 ) , since it would allow for someone to set what `__getitem__` does in a region of code, while other behavior is accessible via `get_or_abort()`, `get_if_present()`, and `unsafe_get_unchecked()`. While technically possible in Mojo today, that proposal has some fairly substantial ergonomics issues without more compiler work (`**kwparams` at a minimum, ideal some of the implicit passing I mentioned there). Changing this would also require reworking `List` and all of the code that uses it, but I think that unless we find this to be a sizable error it should likely remain unchanged past 1.0, meaning that this is possibly the last time we can discuss the issue.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `set_safe(idx: UInt, item: T) -> Bool`  # True if the operation was a success
```
owenhilyard: As before, `ReadOnlySequence` as a separate trait?

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `extent_at(slice: Slice[T], idx: Uint)`
+- `pop_at(idx: Uint) -> T`
+- `iter() -> Iterator[ref T]`
+- `len() -> Uint`
+
+**Default Trait Implementations**
+- `__init__(a: T, b: T, ...) -> Self`
+- `__get__(idx: UInt) -> ref T`
+- `__set__(idx: UInt, item: T)`
+- `push(item: T)`
```
owenhilyard: `append` to match python

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `push_front(item: T)`
```
owenhilyard: `prepend`?

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- Elements are stored in a linear order
+- Supports indexed insertion
+- Implemented by List, LinkedList, Deque
+
+**Base operations**
+- `__init__() -> Self`
+- `get_safe(idx: UInt) -> Optional[ref T]`
+- `set_safe(idx: UInt, item: T) -> Bool`  # True if the operation was a success
+- `insert(item: T, idx: UInt)`
+- `extent_at(slice: Slice[T], idx: Uint)`
```
owenhilyard: Did you mean `extend_at`? In that case, this should take an owning Iterator or `Some[Sequence[T]]`, ideally both.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `push(item: T)`
+- `push_front(item: T)`
+- `pop() -> T`
+- `first() -> Optional[ref T]`
+- `last() -> Optional[ref T]`
+- `map[NewType](f: Fn(t: T) -> NewType) -> Self[NewType]`
+- `map_inplace(f: Fn(f: ref mut T))`
+- `filter(f: Fn(f: ref T) -> Bool) -> Self[T]`
+- `filter_inplace(f: Fn(f: ref T) -> Bool)`
+- `extend_front(slice: Slice[T])`
```
owenhilyard: `Iterator[T]` or `Some[Sequence[T]]` again.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `extend(slice: Slice[T])`
+- `is_empty() -> Bool`
+- `reserve_capacity(cap: UInt)`
+- `shrink_to_fit()`
+
+## ADT Trait Extensions
+
+#### `InsertionOrderedMap[K, V]`
```
owenhilyard, line 228: Should a trait below this which guarantees an implementation-defined stable iteration order exist? This means that any two versions of the type, which consist of the same set of key/value pairs, will iterate the pairs in the same order. It's a weaker guarantee but one I think is helpful for comparisons between maps and in a few other niche places.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+#### `SortedMap[K, V]`
+Key-ordered key–value association.
+
+**Semantics**
+- Iteration order is sorted by key
+- Ordering defined by `K: Comparable`
+- Extends `Map[K, V]`
+
+**Additional operations**
+- `range_scan(start: K, end: K) -> Iterator[(ref K, ref V)]`
```
owenhilyard: Should this be done via slicing? ex: `my_map["a":"e"]`

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+#### `InsertionOrderedMap[K, V]`
+Insertion-ordered key–value association.
+
+**Semantics**
+- Preserves insertion order during iteration
+- Matches Python-style dict behavior
+- Extends `Map[K, V]`
+
+**Additional operations**
+- No additional operations. But by implementing this, the library author guarantees that their `.iter()` function returns elements in the order they were inserted.
```
owenhilyard: Does adding the other guarantee that Python has, which is that `pop` removes the last inserted element, make sense here? Similar for other insertion ordered ADT extensions.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- Ordering defined by `T: Comparable`
+- Extends `Set[T]`
+
+**Additional operations**
+- `range_scan(start: T, end: T) -> Iterator[ref T]`
+- By implementing this, the library author guarantees that their `.iter()` function returns elements in sorted order. 
+
+---
+
+#### `ContiguousSequence[T]`
```
owenhilyard, line 299: Does having a trait below this which provides an `fn(ref self) -> Iterator[Span[T]]` make sense? A separate one for compile-time sized `fn(ref self) -> Iterator[Either[ref [self] InlineList[T, N], Span[T, Origin[self]]]` would be nice, and with parametric traits ( #5799 ) that `N` could be any value, which would allow `LayoutTensor` tile iterators to implement that trait nicely. And, of course, this is a great place to have `SIMDIterable`.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+Index-based sequence whose elements are laid out contiguously in memory. 
+
+**Semantics**
+- Extends `Sequence[T]`
+
+**Additional Operations**
+- `as_slice() -> Slice[T]`
```
owenhilyard: I'd like to see a conditionally conformed `fn simd_iter(ref self) -> SIMDIterator[dtype, width, Origin[self]]` here.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+
+
+## Interaction with concrete implementations
+Concrete types in `std.collections` (or equivalent) are expected to:
+- Implement one or more base ADT traits
+- Clearly document any extra guarantees
+
+Users are encouraged to write APIs generic over the base ADT rather than a specific implementation:
+```mojo
+fn foo[M: Map[String, Int]](m: M):
```
owenhilyard: To help clean this up, consider using `Some` (ex: `fn foo(m: Some[Map[String, Int]])`)

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+    """
+    This function needs a type that has basic get and set operations
+    """
+
+fn bar[M: SortedMap[String, Int] + Sync](m: M):
```
owenhilyard: `m: Some[SortedMap[String, Int] & Sync]`

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- `len() -> UInt` 
+
+**Default Trait Implementations**
+- `__init__(a: (K, V), b: (K, V), ...) -> Self`
+- `get_or_insert(k: K, v: V) -> ref V`
+- `get_or_insert_with(k: K, f: fn() -> V) -> ref V`
+- `contains(k: K) -> Bool`
+- `is_empty() -> Bool`
+- `update(k, f: fn(ref mut V) -> None) -> Bool`   # Returns True if a value was updated
+- `map[NewVType](f: Fn(k: ref K, v: ref V) -> NewVType) -> Self[K, NewVType]`
```
owenhilyard: This should probably be broken out. Specialized map types, especially for things like `String -> Int` or `Int -> T`, are reasonably common, and I don't think we should exclude them since I can easily see an interpreter or database wanting a specialized type here.

mojo/proposals/adt-traits.md
owenhilyard: Point of Discussion: `Entry` API

In Rust, some Maps and Sets have an `Entry` API, which allows provides the ability to check if something is present, and if it is get a reference to the value, but if it isn't it allows you to initialize the value. This allows for implementing "get or insert" in half the number of lookups, and for tree-based data structures implementing most of this API is very convenient to do as part of breaking apart tree traversal and operations on node data. Even for hash tables/sets, assuming a resize is not required, grabbing a reference to the slot for a value is trivial.

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+    """
+    This function needs a type that has get, set, and range_scan operations. 
+    It also needs the type to be thread safe. 
+    """
+```
+
+---
+
+## Why this belongs in the standard library
+Without a standardized base each library defines its own pseudo-Map trait and APIs fragment quickly. More importantly, a proposal like this is only feasible in the standard library.
```
owenhilyard: I agree. The standard library is a good place to place interfaces which have stood the test of time (in the case of many ADTs, their interfaces are decades old or rooted in centuries of math) and I think that we risk ecosystem fragmentation by not having these interfaces. However, I think we also need to be conservative in what is placed into these APIs, since I think they will become part of the glue that holds the Mojo ecosystem together and having mistakes in them will cost us dearly since they will be difficult to change later. 

As a result, until we have a better grasp on what iteration is going to look like, it may be best to avoid iterators in the API for now. Similarly, given that closures are getting a big rework right now, it may be best to avoid those as well until we have better tools to generically describe functions. For example, `filter*` and `map*` must take functions which are safe to call multiple times, which means you can't destroy any captures without replacing them. Additionally, I want to discuss `ReadOnly` versions of the ADTs for the reasons mentioned above and how ADTs with interior mutability will be integrated. 

This is a good start and provides a lot of places for discussion!

mojo/proposals/adt-traits.md
```diff
@@ -0,0 +1,269 @@
...
+- **Increases Productivity:** Whether you're a library author, or are just making a data structure for your own application, you can implement a few functions and the rest of the functions with default trait implementations get implemented for you. 
+
+
+## Next Steps
+If this proposal is implemented it could potentially open the door to some interesting features
+- Firstly, if mojo decides to include generic allocators like Zig, then this means we can choose to construct a map with a tuned implementation (one from std, a third party one, or our own) and with a specialized allocator (like an arena or fixed buffer allocator if our use case allows), and this hand-crafted type 
+will "just work" with any mojo function anywhere that deals with map types. 
+- This one is *very* speculative, but we could create a "PGO on steroids" feature into the compiler that not only chooses between low level compilation decisions, (should I inline this function or not) but also between different implementations of ADTs.
```
owenhilyard: This does seem like a very interesting language design option to explore, especially if Mojo grows a rich set of alternative ADT implementations.

owenhilyard (commented):
I have a few other discussions that don't really fit anywhere inline with the proposal.

owenhilyard:
Point of Discussion: SIMT Safety

One other discussion point to have is the idea of "SIMT Safe" operations. For instance, stores to and loads from a List or Span are safe on a GPU, subject to a lack of resizing. However, having all of the threads in a warp trying to rebalance a tree at the same time sounds like a bad idea. If each thread has a single collection, it's not an issue, but I think some sort of marker that an operation on a type has been checked to work correctly in a SIMT environment is reasonable.

barcharcraz:
I like the idea of defining a standard "container interface" and writing tests to ensure our containers conform. I'm less enthusiastic about providing _actual traits_ for them, because I think typically iterator traits or view types are better, and they usually have less type-system issues.

lsh:
Small naming nit: I don't particularly like the ADT acronym since it has a naming collision with algebraic data types.

PR #5842, opened 2026-01-25
[MAX] Propose to add pixel generation support to `max.serve`

tolgacangoz:
Here is the initial design for the serving layer.

I'm shifting my focus to other priorities for now, so I'd like to hand this over to the SqueezeBits team to finalize. Since you have the full context on the upstream dependencies, you are in the best position to get this merged and optimized quickly.

Please feel free to adopt this branch or cherry-pick the logic as you see fit!

I'll circle back to update the Z-Image and LTX-2 implementations to align with the updates once they land.

@kkimmk


kkimmk:
@jglee-sqbits

PR #5849, opened 2026-01-26
[MAX] Extend Max config system to support PixelGeneration and Diffusers models

kkimmk:
## Overview
This PR extends the existing model_config to allow resolving weight paths for Diffusers models.

## Changes
- Extends `HuggingFaceRepo` to resolve `weight_files` and `supported_encodings` for Diffusers repos.
- Extends `retrieve_architecture` to work with Diffusers models as well.
- Prevents `ValueError` raised when accessing pipeline_config or pipeline_config.model with Diffusers models.

### DIFF
--- a/max/python/max/pipelines/lib/config.py
+++ b/max/python/max/pipelines/lib/config.py
@@ -38,6 +38,7 @@ from pydantic import (
 from typing_extensions import Self
 
 from .config_enums import PipelineRole
+from .hf_utils import is_diffusion_pipeline
 from .kv_cache_config import KVCacheConfig
 from .lora_config import LoRAConfig
 from .memory_estimation import MemoryEstimator, to_human_readable_bytes
@@ -1028,6 +1029,11 @@ class PipelineConfig(ConfigFileModel):
         # Resolve final pipeline-specific changes to the config before doing
         # memory estimations.
         arch.pipeline_model.finalize_pipeline_config(self)
+
+        if is_diffusion_pipeline(model_config.huggingface_model_repo):
+            # Skip memory estimation for diffusion pipelines,
+            # since they don't use KV cache.
+            return
 
         MemoryEstimator.estimate_memory_footprint(
             self,
--- a/max/python/max/pipelines/lib/model_config.py
+++ b/max/python/max/pipelines/lib/model_config.py
@@ -452,16 +452,24 @@ class MAXModelConfig(MAXModelConfigBase):
 
     @computed_field  # type: ignore[prop-decorator]
     @property
-    def huggingface_config(self) -> AutoConfig:
+    def huggingface_config(self) -> AutoConfig | None:
         # Note: For multiprocessing, __getstate__ clears _huggingface_config
         # before pickling. Each worker process will reload the config fresh,
         # which properly handles trust_remote_code dynamic class loading.
         if self._huggingface_config is None:
-            self._huggingface_config = (
-                PIPELINE_REGISTRY.get_active_huggingface_config(
-                    huggingface_repo=self.huggingface_model_repo
+            try:
+                self._huggingface_config = (
+                    PIPELINE_REGISTRY.get_active_huggingface_config(
+                        huggingface_repo=self.huggingface_model_repo
+                    )
                 )
-            )
+            except Exception as e:
+                # Not a transformers-style model (e.g., diffusers model)
+                logger.debug(
+                    f"Could not load HuggingFace config for "
+                    f"{self.model_path}: {e}"
+                )
+                return None
         return self._huggingface_config
 
     @computed_field  # type: ignore[prop-decorator]
--- a/max/python/max/pipelines/lib/hf_utils.py
+++ b/max/python/max/pipelines/lib/hf_utils.py
@@ -28,7 +28,7 @@ import time
 from dataclasses import dataclass
 from functools import cached_property, lru_cache
 from pathlib import Path
-from typing import Any, cast
+from typing import Any, BinaryIO, cast
 
 import huggingface_hub
 from huggingface_hub import errors as hf_hub_errors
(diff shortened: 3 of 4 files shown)


kkimmk:
@tolgacangoz It'd be great if you could take a look at this config system PR! The main differences from the initial implementation in the z-image PR, that might require some modifications on your codebase, are:
- Implemented `DiffusersConfig` as a subclass of the `ConfigFileModel` Pydantic model.
- Updated PixelGeneration configs to avoid using `huggingface_config`. I felt it was more intuitive for Diffusers models to set `huggingface_config` to `None` and rely solely on `diffusers_config`. So, to access `huggingface_config` for text encoders, it would be `pipeline_config.model.diffusers_config.text_encoder_model_repo` rather than `pipeline_config.model.huggingface_config`.
- The type of `DiffusersComponentConfig.config_dict` is just `dict` instead of `HuggingFaceRepo | dict` to provide a unified access pattern.

max/python/max/pipelines/lib/config_enums.py
```diff
@@ -86,6 +86,71 @@ def uses_dispatch_service(self) -> bool:
...
+    """Generate an image from a text prompt."""
+
+    TEXT_TO_VIDEO = "text_to_video"
+    """Generate a video from a text prompt."""
+
+    # Image + text input (conditioning)
+    IMAGE_TO_IMAGE = "image_to_image"
+    """Transform an image using a text prompt; e.g., style transfer."""
+
+    IMAGE_TO_VIDEO = "image_to_video"
```
KCaverly: Are these all supported with the existing pipelines you are working at bringing in? If not, I would prefer we not over engineer early, and introduce all these different enums.

max/python/max/pipelines/lib/config.py
```diff
@@ -1486,3 +1500,92 @@ def from_flags(
...
+    """
+
+    # Generation mode - architectural choice set at server startup
+    generation_type: PixelGenerationType = Field(
+        default=PixelGenerationType.TEXT_TO_IMAGE,
+        description="The type of pixel generation to perform.",
+    )
+
+    # Scheduler override - server-wide setting
+    scheduler_type: str | None = Field(
```
KCaverly: What does the scheduler refer to in this context? How is this used?

max/python/max/pipelines/lib/config.py
```diff
@@ -1486,3 +1500,92 @@ def from_flags(
...
+    @staticmethod
+    def help() -> dict[str, str]:
+        """Return help text for PixelGenerationConfig-specific fields."""
+        return {
+            "generation_type": "The type of pixel generation: text_to_image, text_to_video, image_to_image, image_to_video, image_editing, video_to_video, inpainting, outpainting, or controlnet.",
+            "scheduler_type": "Optional scheduler override (e.g., 'ddim', 'euler'). Uses model default if not specified.",
+        }
+
+    @classmethod
+    def from_flags(
```
KCaverly: I don't understand the role of this method. If you are not introducing changes to the CLI in this stack, where is this used?

max/python/max/pipelines/lib/model_config.py
```diff
@@ -441,18 +447,73 @@ def huggingface_model_repo(self) -> HuggingFaceRepo:
...
+
+        Returns:
+            The AutoConfig for this model, or None if this is not a
+            transformers-style model (e.g., diffusers models).
+        """
         if self._huggingface_config is None:
-            self._huggingface_config = (
-                PIPELINE_REGISTRY.get_active_huggingface_config(
-                    huggingface_repo=self.huggingface_model_repo
+            try:
```
KCaverly, line 460: This is a positive cleanup, thanks!

max/python/max/pipelines/lib/config.py
```diff
@@ -1486,3 +1500,92 @@ def from_flags(
...
+        self.generation_type = generation_type
+        self.scheduler_type = scheduler_type
+
+    @property
+    def is_video(self) -> bool:
+        """Whether this config produces video output."""
+        return self.generation_type.outputs_video
+
+    @property
+    def requires_input_image(self) -> bool:
```
KCaverly: How are these validation methods expected to be used downstream? 

For LLMs we use Context Validators, which are defined on the Architecture. Would it be possible to pass a list of callables, there that would verify if the Context object as created by the Tokenizer is valid for this architecture.

KCaverly (commented):
I don't want to block on introducing these, but added a few comments.

As a general guideline, please refrain from introducing unnecessary or forward looking features, that we are not immediately addressing in the follow up PRs. It is much easier for us to introduce objects like ImageGenerationType etc, as we have a need, instead of up front. This clutters up the code base, and hides whats actively supported.

max/python/max/pipelines/lib/config_enums.py
kkimmk: I removed all the unused types.

max/python/max/pipelines/lib/config.py
kkimmk: It's currently not being used for offline inference. I'll remove it for now, and let @tolgacangoz configures it properly when he implements serving feature.

max/python/max/pipelines/lib/config.py
kkimmk: As with the previous comment, it is not being used in offline inference, and thereby removed it for now. It should also be applied properly with entrypoints for `max.serve`. cc. @tolgacangoz

max/python/max/pipelines/lib/config.py
kkimmk: Since most of the enum types are removed, these validations are no longer used, thereby removed. I'll tackle them again when those new tasks are introduced!

kkimmk:
FYI, @tolgacangoz 

> it would be pipeline_config.model.diffusers_config.text_encoder_model_repo rather than pipeline_config.model.huggingface_config.

I changed `text_encoder_model_repo` to `text_encoder_hf_config`!

KCaverly:
I believe this should be closed, For now MAXModelConfig.diffusers_config, is available on main, and will sync with this repo tonight. If we need to move this around, we can do so, but I don't think we should need to make any additional adjustments to external models to support diffusers models.

For initial testing, you can use the PipelineConfig object, just pass defer_resolve=True, during __init__.

kkimmk:
@KCaverly I didn't get the comment. Does it mean there's already diffusers_config available in main, requiring to close this PR? or this PR will get merged?

KCaverly:
!sync

PR #5850, opened 2026-01-26
[MAX] Extend Max registry system to support PixelGenerationPipeline

kkimmk:
## Overview
This PR extends the Registry system to support `PixelGenerationPipeline`, including updated registry logic.

## Changes
- Update registry handling to accommodate `PixelGenerationPipeline`
- Add comprehensive tests in max/tests/tests/pipelines/lib/test_registry_diffusers.py

## Usage Example
```
pipeline_config = PixelGenerationConfig(model_path=model_id_for_diffusers_model)
pipeline_task = PIPELINE_REGISTRY.retrieve_pipeline_task(pipeline_config)
tokenizer, model_factory = PIPELINE_REGISTRY.retrieve_factory(
    pipeline_config,
    task=pipeline_task,
)
pixel_generation_pipeline = model_factory()
```

### DIFF
--- a/max/python/max/pipelines/lib/registry.py
+++ b/max/python/max/pipelines/lib/registry.py
@@ -35,7 +35,7 @@ from max.interfaces import (
     TextGenerationRequest,
 )
 from max.nn.legacy.kv_cache import KVCacheStrategy
-from max.pipelines.core import (
+from max.pipelines.core import (  # type: ignore[attr-defined] #TODO: can be removed once #5816 is merged.
     PixelContext,
     TextAndVisionContext,
     TextContext,
@@ -56,7 +56,10 @@ from .config_enums import RopeType, SupportedEncoding
 from .embeddings_pipeline import EmbeddingsPipeline
 from .hf_utils import HuggingFaceRepo
 from .interfaces import ArchConfig, PipelineModel
-from .pipeline_variants import PixelGenerationPipeline, TextGenerationPipeline
+from .pipeline_variants import (  # type: ignore[attr-defined] #TODO: can be removed once #5835 is merged.
+    PixelGenerationPipeline,
+    TextGenerationPipeline,
+)
 from .pipeline_variants.overlap_text_generation import (
     OverlapTextGenerationPipeline,
 )
@@ -373,7 +376,7 @@ class PipelineRegistry:
         """
         # Check for diffusers-style model first (via pipeline_config)
         if pipeline_config is not None:
-            diffusers_config = pipeline_config.model.diffusers_config
+            diffusers_config = pipeline_config.model.diffusers_config  # type: ignore[attr-defined] #TODO: can be removed once #5849 is merged.
             if diffusers_config is not None:
                 # Use pipeline_class from model_index.json for architecture lookup
                 arch_name = diffusers_config.pipeline_class
@@ -551,7 +554,7 @@ class PipelineRegistry:
                 f"No architecture found for {pipeline_config.model.model_path}"
             )
 
-        is_diffusers = pipeline_config.model.diffusers_config is not None
+        is_diffusers = pipeline_config.model.diffusers_config is not None  # type: ignore[attr-defined] #TODO: can be removed once #5849 is merged.
 
         # Calculate Max Length (skip for diffusers models)
         huggingface_config = (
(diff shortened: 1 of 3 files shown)


max/python/max/pipelines/lib/registry.py
```diff
@@ -504,21 +543,59 @@ def retrieve_tokenizer(
...
+                model_path=pipeline_config.model.model_path,
+                pipeline_config=pipeline_config,
+                subfolder="tokenizer",
+                subfolder_2=subfolder_2,
+                revision=pipeline_config.model.huggingface_model_revision,
+                max_length=max_length,
+                trust_remote_code=pipeline_config.model.trust_remote_code,
+                chat_template=pipeline_config.retrieve_chat_template(),
+                context_validators=arch.context_validators,
+            )
```
kkimmk: @tolgacangoz Could you check how should I resolve `wrap_prompt_as_chat` and `default_chat_template_options` for PixelGenerationTokenizer initialization?

tolgacangoz: I imagined it like that for those who need them:

```py
z_image_module_v3_arch = SupportedArchitecture(
	...
     tokenizer=partial(
         PixelGenerationTokenizer,
         wrap_prompt_as_chat=True,
         default_chat_template_options={
             "add_generation_prompt": True,
             "enable_thinking": True,
         },
     ),
	...
)
```

kkimmk: I wonder if we can handle the prompt wrapping from the caller side, similarly to how the `TextTokenizer` assumes. And how about `default_chat_template_options` being parsed through the request, like the `TextTokenizer`?

If that would be possible, this part could remain as it is, not parsing the `wrap_prompt_as_chat`, and `default_chat_template_options`. What do you think?

tolgacangoz: In my first PR, I did exactly what you said. But then, I realized that `"enable_thinking": True` is only used for Z-Image and `"add_generation_prompt": True` for just several pipelines, and `role": ..., "content": ...` style text processing too. I thought that just hardcoding these into the pipelines' `arch.py` files would be enough. Since they seem hardcoded, users don't even need to input them in their requests.

But now I see: Flux.2 has several potentially configurable system messages, so users would want to give them themselves. Thus, I am returning to `TextTokenizer`'s style.

kkimmk: Makes sense. Then, I guess we might need to consider that hardcoded path for those specific models, if they require a fixed setting.
As you said, let's return to the TextTokenizer's style by default. And could you review extending it to allow that hard-coded config when you work on z-image?

tolgacangoz: Yes, I will add it to Z-Image.
Could you convert pixel generation-related parts into `TextTokenizer`'s style? I don't have time atm.

kkimmk: For this part, there's no need to change anything for that. I've just removed the TODO comment. Did you perhaps mean the Tokenizer PR?

tolgacangoz: Yes, the tokenizer and its request, etc.

kkimmk: Okay, I'll append a commit on that PR.

kkimmk: [0001-feat-remove-option-inside-and-add-to.patch](https://github.com/user-attachments/files/24878665/0001-feat-remove-option-inside-and-add-to.patch)
Since I can not directly push to your repo, I'm just sharing the patch. The change is small, so I’m not sure if this will be helpful😅

tolgacangoz: My PR seems imported internally, I guess it would be better to open a PR as if it were merged?

kkimmk: Upon a chat with Kyle, it seems we can still append a commit to a PR in imported-internally. Could you make the commit from your side? If you are not available at the moment, I can make a separate PR later, when that PR is explicitly merged.

tolgacangoz: OK, doing it now.

tolgacangoz:
Did you mean "And could you review extending it to allow that hard-coded config when you work on z-image?" that emphasizing how its inputs are supposed to be given by the users like that:
```console
curl -X POST http://localhost:8000/v1/images/generations \
  -H "Content-Type: application/json" \
  -d '{
    "messages": [{
                  "role": "user",
                  "content": "A cute baby sea otter floating in crystal clear water"
                  }],
    "chat_template_options": [{
                  "add_generation_prompt": True,
                  "enable_thinking": True
                  }],
    "n": 1,
    "size": "1024x1024",
    "guidance_scale": 0.0,
    "num_inference_steps": 9
  }' -o result.json
```

If we "allow that hard-coded config when using z-image", then why did we remove the two options `wrap_prompt_as_chat` and `default_chat_template_options`?

PR #5853, opened 2026-01-26
[stdlib] Add `Writable.estimate_bytes_to_write()` with a default implementation

martinvuyk:
Part of #4249. This will enable every `Writable` type to implement a faster size estimation function in cases where calling `self.write_to(writer)` is expensive. It will also allow other `Writer` types like string to previously reserve the necessary buffer size, enabling equivalent optimizations to what was done by using the `_TotalWritableBytes` internal tool.

This should eventually be extended to enable one to pass the format specification and calculate the necessary buffer size according to it as well.

This also adds integer write size estimation size for `SIMD` as a proof of the improvements this enables

## Benchmark results

CPU: Intel® Core™ i7-7700HQ

|name | old value (ms) | new value (ms) | markdown perc | magnitude|
|-- | -- | -- | -- | -- |
|bench_string_write_int_short | 17.0325 | 17.2421 | -0.0123 | 0.9878|
|bench_string_write_int_long | 37.1422 | 0.0881 | 0.9976 | 421.5114|

### DIFF
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -40,6 +40,7 @@ components, including tensor operations, linear algebra routines, and
 domain-specific libraries for machine learning and scientific computing.
 """
 
+from bit import next_power_of_two
 import math
 from collections import InlineArray
 from hashlib.hasher import Hasher
@@ -73,7 +74,7 @@ from bit import bit_width, byte_swap, pop_count
 from builtin._format_float import _write_float
 from builtin.device_passable import DevicePassable
 from builtin.format_int import _write_int
-from math import DivModable, Powable
+from math import DivModable, Powable, log10, ceil
 from documentation import doc_private
 from memory import bitcast, memcpy, pack_bits
 from python import ConvertibleToPython, Python, PythonObject
@@ -89,6 +90,7 @@ from utils.numerics import max_or_inf as _max_or_inf
 from utils.numerics import min_finite as _min_finite
 from utils.numerics import min_or_neg_inf as _min_or_neg_inf
 from utils.numerics import nan as _nan
+from format._utils import _TotalWritableBytes
 
 from .dtype import (
     _integral_type_of,
@@ -2271,6 +2273,119 @@ struct SIMD[dtype: DType, size: Int](
             _write_scalar(writer, element)
         writer.write_string(")")
 
+    fn estimate_bytes_to_write(self) -> Int:
+        """Estimate the total bytes to write the type into a writer.
+
+        Returns:
+            The estimated bytes needed to write the type into a writer.
+        """
+
+        @parameter
+        if not Self.dtype.is_integral():
+            var total_bytes = _TotalWritableBytes()
+            self.write_to(total_bytes)
+            return total_bytes.size
+        else:
+            comptime extra = (2 * Self.size) if Self.size > 1 else 0
+            comptime is_unsigned = Self.dtype.is_unsigned()
+            comptime MAX = Scalar[Self.dtype].MAX
+            comptime f_MAX = MAX.cast[DType.float64]()
+            comptime log10_f_MAX = Int(ceil(log10(f_MAX)))
+            comptime max_digits_size[offset: Int] = next_power_of_two(
+                log10_f_MAX - offset
+            )
+
+            fn _get_max_digits[
+                offset: Int = 0
+            ](out res: SIMD[Self.dtype, max_digits_size[offset]]):
+                res = {MAX}
+
+                var idx = 0
+
+                for i in range(offset, offset + type_of(res).size):
+                    if log10_f_MAX >= i + 1:
+                        res[idx] = Scalar[Self.dtype](10) ** i - 1
+                        idx += 1
+
+            @parameter
+            fn _full_digits[offset: Int = 0]() -> Int:
+                var abs_v: Self
+
+                @parameter
+                if is_unsigned:
+                    abs_v = self
+                else:
+                    abs_v = abs(self + self.eq(Self.MIN).cast[Self.dtype]())
+
+                var count = offset + extra
+
+                @parameter
+                if bit_width_of[Self.dtype]() <= 64:
+                    comptime max_digits = _get_max_digits[offset]()
+
+                    @parameter
+                    for i in range(Self.size):
+                        count += max_digits.lt(abs_v[i]).reduce_bit_count()
+                else:  # 128 and 256 bit integers would take 1KiB+
+                    var max_digit = Self(10) ** offset
+
+                    @parameter
+                    for i in range(offset, max_digits_size[0]):
+
+                        @parameter
+                        if log10_f_MAX < i + 1:
+                            break
+                        count += max_digit.le(abs_v).reduce_bit_count()
+                        max_digit *= 10
+
+                return count + self.le(0).reduce_bit_count()
+
+            @parameter
+            if bit_width_of[Self.dtype]() <= 16 or Self.size > 1:
+                return _full_digits()
+            else:
+                # NOTE: We can take advantage of 4 facts:
+                # 1. We are human and our measurement systems don't tend towards
+                # numbers bigger than 10_000
+                # 2. They can be casted to [u]int16 and use a size of 4 which
+                # only requires a machine simd width of 64 bits
+                # 3. We can lower latecy by letting the branch predictor do its
+                # job, instead of brute-forcing all possible digits
+                # 4. After this comparison 32bit integers need only 8 digit
+                # comparisons instead of 16 and 64bit need only 16 instead of 32
+                var comp: Bool
+
+                @parameter
+                if is_unsigned:
+                    comp = self < 10_000
+                else:
+                    comp = Self(-10_000) < self < Self(10_000)
+                if not comp:
+                    return _full_digits[4]()
+                else:
+                    comptime dt = DType.uint16 if is_unsigned else DType.int16
+                    comptime max_digits = SIMD[dt, 4](0, 10, 100, 1000)
+                    var vec = self.cast[dt]()
+                    var abs_v: type_of(vec)
+
+                    @parameter
+                    if is_unsigned:
+                        abs_v = vec
+                    else:
+                        abs_v = abs(vec)
+
+                    var count = 0
+
+                    @parameter
+                    for i in range(Self.size):
+                        count += max_digits.le(abs_v[i]).reduce_bit_count()
+
+                    @parameter
+                    if is_unsigned:
+                        return count
+                    else:
+                        return count + vec.lt(0).reduce_bit_count()
+
     @always_inline
     fn to_bits[
         _dtype: DType = _uint_type_of_width[bit_width_of[Self.dtype]()]()
@@ -3051,17 +3166,14 @@ struct SIMD[dtype: DType, size: Int](
         ), "Expected either integral or bool type"
 
         @parameter
-        if Self.dtype == DType.bool:
-
-            @parameter
-            if Self.size == 1:
-                return Int(self)
+        if Self.dtype == DType.bool and Self.size == 1:
+            return Int(self)
+        elif Self.dtype == DType.bool:
+            if is_compile_time():  # FIXME(#5855): remove this workaround
+                return self.cast[DType.uint8]().reduce_bit_count()
             else:
-                var packed_mask = pack_bits(
-                    rebind[SIMD[DType.bool, Self.size]](self)
-                )
-                var count = pop_count(packed_mask)
-                return Int(count)
+                var s = self._refine[DType.bool, Self.size]()
+                return Int(pop_count(pack_bits(s)))
         else:
             return Int(pop_count(self).reduce_add())
 
--- a/mojo/stdlib/std/builtin/variadics.mojo
+++ b/mojo/stdlib/std/builtin/variadics.mojo
@@ -1040,6 +1040,42 @@ struct VariadicPack[
                 self[i].write_to(writer)
         writer.write_string(end)
 
+    fn _estimate_bytes_to_write[
+        O1: ImmutOrigin = StaticConstantOrigin,
+        O2: ImmutOrigin = StaticConstantOrigin,
+        O3: ImmutOrigin = StaticConstantOrigin,
+    ](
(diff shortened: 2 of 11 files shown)


mojo/stdlib/std/format/__init__.mojo
```diff
@@ -251,6 +251,22 @@ trait Writable:
 
         _reflection_write_to[f=call_write_repr_to](self, writer)
 
+    @always_inline("nodebug")
+    fn estimate_bytes_to_write(self) -> UInt:
```
NathanSWard:
```suggestion
    fn estimate_bytes_to_write(self) -> Int:
```
We're standardizing on `Int` being the canonical return type in the stdlib :)

mojo/stdlib/std/format/__init__.mojo
```diff
@@ -251,6 +251,22 @@ trait Writable:
 
         _reflection_write_to[f=call_write_repr_to](self, writer)
 
+    @always_inline("nodebug")
```
NathanSWard:
```suggestion
    @always_inline
```
I definitely believe we can remove the `"nodebug"` - not too sure what benefit we get by hiding the debug info for this function.

Also not entirely sure of the benefit of `always_inline` here either, it's probably okay but I wonder if there's actually a benefit. Especially since this may greatly increase binary size if the `write_to` of `self` is massive and inlined as well.

mojo/stdlib/std/format/_utils.mojo
```diff
@@ -392,6 +392,7 @@ struct _TotalWritableBytes(Writer):
             for i in range(1, length):
                 self.write(sep, values[i])
 
+    @always_inline("nodebug")
```
NathanSWard:
```suggestion
    @always_inline
```

mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -645,23 +645,22 @@ struct String(
             args: Sequence of arguments to write to this Writer.
         """
         comptime length = args.__len__()
-        var total_bytes = _TotalWritableBytes()
-        total_bytes.size += self.byte_length()
+        var total_bytes = UInt(0)
```
NathanSWard:
```suggestion
        var total_bytes = self.byte_length()
```

The previous implementation had the initial byte_length of the string. Should we keep this?

mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -645,23 +645,22 @@ struct String(
...
             @parameter
             for i in range(length):
                 args[i].write_to(writer)
 
-        if total_bytes.size <= Self.INLINE_CAPACITY:
+        if Int(total_bytes) <= Self.INLINE_CAPACITY:
             _write(self)
         else:
-            self.reserve(total_bytes.size)
+            self.reserve(Int(total_bytes))
```
NathanSWard: One potential issue I see with this current design is what if the `estimate_bytes_to_write` ends up overshooting the total number of bytes needed. So we are left with an excess heap allocation.

It might be worth adding in the docs of `estimate_bytes_to_write` that this should be either an exact or conservative estimate.

mojo/stdlib/std/format/__init__.mojo
```diff
@@ -251,6 +251,22 @@ trait Writable:
...
+            The estimated bytes needed to write the type into a writer.
+
+        Notes:
+            By default this function runs the type's `write_to` function. If
+            doing that is not cheap for a given type, consider writing a faster
+            size estimation function instead.
+        """
+        var total_bytes = _TotalWritableBytes()
+        self.write_to(total_bytes)
+        return UInt(total_bytes.size)
```
NathanSWard: Can we add a unit test in `test/format` that binds this default behavior?

NathanSWard (changes requested):

mojo/stdlib/std/format/__init__.mojo
martinvuyk: I think that this should all inline into just adding integers, otherwise we have massive call chains for very trivial ops. And having this in debug will just add noise for a single add operation and each call to estimate size. The whole idea about this is to simplify a lot of the function calls just down to arithmetic

mojo/stdlib/std/collections/string/string.mojo
martinvuyk: Yeah I missed this while copy pasting code, will fix

mojo/stdlib/std/format/__init__.mojo
NathanSWard: It's fine to keep it always_inline - we just need to be very conscious about these are we are way to liberal with it across the codebase generally.

I'd still request you remove `"nodebug"` please.

martinvuyk:
@NathanSWard added code for `SIMD` integers and benchmarks to prove the improvement this new setup enables. I've updated the PR description with the results 421x improvement (for sequential writing) :tada: :rocket:

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -2282,6 +2284,49 @@ struct SIMD[dtype: DType, size: Int](
             _write_scalar(writer, element)
         writer.write_string(")")
 
+    fn estimate_bytes_to_write(self) -> Int:
```
NathanSWard, line 2276: I would recommend pushing off the `SIMD.estimate_bytes_to_write` to a separate PR - or if you wanted to add it in this PR, we now need to add tests for this function as well :)

mojo/stdlib/std/math/math.mojo
```diff
@@ -2223,10 +2223,10 @@ fn log10[dtype: DType, width: Int, //](x: SIMD[dtype, width]) -> type_of(x):
     __comptime_assert (
         dtype.is_floating_point()
     ), "input type must be floating point"
+    comptime `log10(2)` = 0.301029995663981195213738894724493027
 
     @parameter
     if is_nvidia_gpu():
-        comptime log10_2 = 0.301029995663981195213738894724493027
```
NathanSWard: This looks like a drive-by change, but I'm not a fan of using the back-tick variable name here, IMO it a bit messier than before. `log10_2` was a okay name.

soraros: You could add a inline `llvm_intrinsic["llvm.log10", ...](...)`: https://github.com/modular/modular/issues/5856#issuecomment-3802789632

mojo/stdlib/std/builtin/simd.mojo
martinvuyk: I wanted to have some proof of the improvements this allows. If you want I can split it off, the implementation got a bit more complicated than I originally expected. I did add tests that cover pretty much every edge case now.

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -2284,6 +2287,120 @@ struct SIMD[dtype: DType, size: Int](
...
+        """Estimate the total bytes to write the type into a writer.
+
+        Returns:
+            The estimated bytes needed to write the type into a writer.
+        """
+
+        @parameter
+        if Self.dtype.is_integral():
```
soraros: Nit: could you swap the two branches, so the simpler case (`not is_integral()`) is handled first? I think that'd make the code a bit more readable.

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -2284,6 +2287,120 @@ struct SIMD[dtype: DType, size: Int](
...
+                # 4. After this comparison 32bit integers need only 8 digit
+                # comparisons instead of 16 and 64bit need only 16 instead of 32
+                var comp: Bool
+
+                @parameter
+                if is_unsigned:
+                    comp = self < 10_000
+                else:
+                    comp = Self(-10_000) < self < Self(10_000)
+                if comp:
```
soraros: Nit: could you change this part to the following structure:
```mojo
if not comp:
  return ...
longer_body
return ...
```

NathanSWard:
Awesome, this is looking good! I'll get to this first thing next week and get it synced in :)

NathanSWard:
Hey @martinvuyk - if you can rebase, I'll get this merged in asap.
I can already think of places to apply it in the updated formatting code :)

mojo/stdlib/std/builtin/variadics.mojo
```diff
@@ -1040,6 +1040,42 @@ struct VariadicPack[
                 self[i].write_to(writer)
         writer.write_string(end)
 
+    fn _estimate_bytes_to_write[
```
martinvuyk, line 1043: @NathanSWard I just added this as a fly-by change to cleanup string's use of it. I can split it off if you wish

martinvuyk: The amount of repetition when doing this for collections is tickling at my brain for a shared implementation. I don't know if we'll ever have something like `ComptimeIterable` or `ComptimeIndexable` + the equivalent runtime things and implement this and write_to generically for iterable collections

NathanSWard: We ideally want to be generic over `Iterable`, however our current iterator traits don't let you return references in generic contexts, meaning if we had a `write_iterable_to`, it would invoke a copy of every element unfortunately. 

It'll first take some reworking of `Iterable` before we can do this generically unfortunately.

NathanSWard: We hopefully wouldn't need to do "comptime iterable" as the `Iterable` trait should allows ways to do this.
Things like defining an optional static size, or something like Rust's `ExactSizeIterator` are possible routes.

NathanSWard:
!sync

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -2271,6 +2273,119 @@ struct SIMD[dtype: DType, size: Int](
             _write_scalar(writer, element)
         writer.write_string(")")
 
+    fn estimate_bytes_to_write(self) -> Int:
```
NathanSWard, line 2276: @martinvuyk - do you have a reference/link to the algorithm used here? E.g. if it's a paper or if it's implemented in another language?

NathanSWard (approved):

mojo/stdlib/std/builtin/simd.mojo
martinvuyk: Reference: my braincells lol

I like inventing the stuff I use, it's what I like about contributing to Mojo; I get to try and come up with something no one did before and try to beat every other implementation

NathanSWard: Cool, that works too. Just curious to if there is a reference implementation, as it's easier to debug/maintain especially if we end up finding bugs in the future.

NathanSWard:
Hey @martinvuyk!
Talked with the stdlib team. We're in a phase where we want to be judicious and about what new APIs we add before 1.0, since every API added now is one we'll have to review as a potential candidate to stabilize. Given that this feature is not found in many other standard libraries, and there are available workaround like manual pre-allocation, I suggest we back-burner this until later when we can give it a more careful consideration?

A few questions/concerns about this:
1. With the current API, if an implementor of this function _under_ estimates the size, we have the issues where the backing string would need to grow (reallocate) once the values are written to, which slightly defeats the purpose of this utility.
- I could see a use for `get_exact_size` or something like that which uses the `_WritableBytes` helper type under the hood if the user _needs_ the exact number of bytes for preallocation (where either under or over allocating would not be acceptable).
2. Does this utility have an application outside of pre-allocating for `String`? 
- For example, if we write directly to an underlying buffer (File handle) we don't need the check the size before hand.
- The `Writable` trait is really only useful when used for debugging, I could see some use in a more generic (de)serialization library, however I'm unsure if this optimization will really be necessary in generic debug/printing code.

martinvuyk:
> Talked with the stdlib team. We're in a phase where we want to be judicious and about what new APIs we add before 1.0, since every API added now is one we'll have to review as a potential candidate to stabilize

@NathanSWard I completely understand and agree, this should not be stabilized yet. FYI I also don't want the separation of `write_to` from `write_to_repr` to be stabilized yet, `Writable` is not only about debugging and that separation is already giving issues for nested structures (so even debugging is kind of harder/uglier). 

> Given that this feature is not found in many other standard libraries, and there are available workaround like manual pre-allocation

I'm honestly not surprised given the only language where I've liked dealing with strings is Python, and my goal is to make Mojo's even better.

> I suggest we back-burner this until later when we can give it a more careful consideration?

Yeah I think it's gotten to a point where I should write out a proposal to explain a bit more of my motivations and vision for the whole `Writable` and `Writer` APIs. I'll be writing it in my free time this week and open a PR shortly.

> 1. With the current API, if an implementor of this function _under_ estimates the size, we have the issues where the backing string would need to grow (reallocate) once the values are written to, which slightly defeats the purpose of this utility.
> 
> 
>     * I could see a use for `get_exact_size` or something like that which uses the `_WritableBytes` helper type under the hood if the user _needs_ the exact number of bytes for preallocation (where either under or over allocating would not be acceptable).

I think we could adapt this by having the allowance/preference of over or under-estimation be determined by an argument. It does deserve a bit more consideration, and we could start by making it an exact requirement and thus the fallback to `_TotalWritableBytes` and later on relax it to include e.g. float estimation (which is over-estimating in the algorithm I've managed to setup) which is another case where transforming it to a string is very expensive

> 2. Does this utility have an application outside of pre-allocating for `String`?

Anything that has to do with manipulating strings, and potentially some other logic like IO could make use of the knowledge of how much data is to be transmitted in total.
 
>     * For example, if we write directly to an underlying buffer (File handle) we don't need the check the size before hand.

If you want to assume the file system has enough space for storing the string sure, but if you're working on constrained environments where e.g. writing out an array of numbers takes more memory than what is available either in the file system or in RAM, then we need to have this kind of setup.

>     * The `Writable` trait is really only useful when used for debugging, I could see some use in a more generic (de)serialization library, however I'm unsure if this optimization will really be necessary in generic debug/printing code.

I will show more use-cases in the proposal I'll write, but suffice it to say that debugging is just the tip of the iceberg in what I want to achieve with `Writable` and `Writer`. If we want to follow Python's lead (which I strongly want to since I like their string abilities), we need to think more in the direction of template strings. I want our trait family to be used for everything from json, yaml, to html and beyond. I think we have a huge opportunity here to setup something that is easily extensible by external libraries in such a way that we could provide native-feeling templating infrastructure

PR #5863, opened 2026-01-28
[stdlib] add `downcast` parametric alias to stdlib prelude

bgreni:
I have found the `downcast` alias to be extremely useful during my adventures with the new reflection functionality for calling trait static methods amongst other things. As more people dig into it I believe it would helpful to import it into the prelude to avoid having to explicitly import it all the time.

### DIFF
--- a/mojo/stdlib/std/builtin/device_passable.mojo
+++ b/mojo/stdlib/std/builtin/device_passable.mojo
@@ -12,7 +12,6 @@
 # ===----------------------------------------------------------------------=== #
 
 from sys.intrinsics import _type_is_eq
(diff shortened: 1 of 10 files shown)


JoeLoser (commented):
Hey Brian, I chatted with the team last night, and we would prefer not to put this in the prelude right now. Some rationale:
- Conditional conformances in the compiler is going to land _very soon_ which will reduce the ned for most people to need these functions for `downcast`

That said, I'm going to close this for now - let's see how conditional conformances bake for a bit in the stdlib in the coming month or two and then check on things.

Issue #5873, opened 2026-01-29
[BUG] [mojo-lang] Function `get_type_name` is returning inconsistent values


martinvuyk:
### Bug description

This should just print unqualified type name and the `write_to` value (not repr), or at least let us control whether or not to use repr. Results are too incosistent.<br>e.g. `_test[Int(0)]()` correctly produces `Int, 0`

```mojo
from reflection import get_type_name
from testing import assert_equal


struct TestParamWriting[param: Some[AnyType]]():
    ...


def main():
    def _test[T: Writable, //, value: T]():
        comptime name = get_type_name[TestParamWriting[value]]()
        var params = name[name.find("[") + 1 : name.rfind("]")]
        var v = materialize[value]()
        assert_equal(String(get_type_name[T](), ", ", v), params)

    _test[UInt8(0)]()
    _test[InlineArray[UInt8, 2](3, 4)]()
    _test[List[Int]({3, 4})]()
```

output for `UInt8(0)`

```
AssertionError: `left == right` comparison failed:
   left: SIMD[DType.uint8, 1], 0
  right: SIMD[DType.uint8, 1], 0 : SIMD[DType.uint8, 1]
```

output for `InlineArray[UInt8, 2](3, 4)`

```
AssertionError: `left == right` comparison failed:
   left: std.collections.inline_array.InlineArray[SIMD[DType.uint8, 1], 2], InlineArray[SIMD[DType.uint8, 1], 2](SIMD[DType.uint8, 1](3), SIMD[DType.uint8, 1](4))
  right: std.collections.inline_array.InlineArray[SIMD[DType.uint8, 1], 2], {<unprintable>}
```

output for `List[Int]({3, 4})`

```
AssertionError: `left == right` comparison failed:
   left: List[Int], [3, 4]
  right: List[Int], {<unprintable>, 2, 2}
```

This probably supersedes issue [MSTDL-2236](https://linear.app/modularml/issue/MSTDL-2236/bug-function-get-type-name-returns-wrong-data-for-string-parameters)

### Steps to reproduce

...

### System information

`Mojo 0.26.2.0.dev2026012806`

Issue #5872, opened 2026-01-29
[BUG] [mojo-lang] Weird lowering error when trying to use reflection: `'get_type_name' requires a concrete type`


martinvuyk:
### Bug description

```mojo
from reflection import get_type_name
from testing import assert_equal


@fieldwise_init
struct TestDefaultWritable[T: Writable, //, param: T](Writable):
    var x: Int
    var y: Int


def main():
    def _test[type: Writable, //, value: type]():
        comptime T = TestDefaultWritable[value]
        comptime TStr = "TestDefaultWritable["
        var v = materialize[value]()
        assert_equal(String(T(1, 2)), String(TStr, v, "](1, 2)"))
        assert_equal(
            repr(T(1, 2)), String(TStr, repr(v), "](x=Int(1), y=Int(2))")
        )

    _test[InlineArray[Int, 3](1, 2, 3)]()
```
outputs
```
            repr(T(1, 2)), String(TStr, repr(v), "](x=Int(1), y=Int(2))")
                                            ^
oss/modular/mojo/stdlib/std/builtin/repr.mojo:83:4: note: function instantiation failed
oss/modular/mojo/stdlib/std/builtin/repr.mojo:98:24: note: call expansion failed with parameter value(s): (..., "size": 3, ...)
oss/modular/mojo/stdlib/std/collections/inline_array.mojo:63:8: note: function instantiation failed
oss/modular/mojo/stdlib/std/collections/inline_array.mojo:63:8: note: call expansion failed with parameter value(s): (..., ...)
oss/modular/mojo/stdlib/std/format/__init__.mojo:224:8: note: function instantiation failed
oss/modular/mojo/stdlib/std/format/__init__.mojo:257:51: note: call expansion failed with parameter value(s): (..., ..., ...)
oss/modular/mojo/stdlib/std/format/__init__.mojo:261:4: note: function instantiation failed
oss/modular/mojo/stdlib/std/format/__init__.mojo:274:5: note: call expansion failed with parameter value(s): (...)
oss/modular/mojo/stdlib/std/reflection/type_info.mojo:120:4: note: function instantiation failed
oss/modular/mojo/stdlib/std/reflection/type_info.mojo:120:4: note: call expansion failed with parameter value(s): (..., ..., "qualified_builtins": false)
oss/modular/mojo/stdlib/std/reflection/type_info.mojo:89:4: note: function instantiation failed
[...]

Issue #5871, opened 2026-01-29
[BUG] Function `get_type_name` returns wrong data for String parameters


martinvuyk:
### Bug description

```mojo
from testing import assert_equal


@fieldwise_init
struct TestDefaultWritable[param1: Bool, param2: String](Writable):
    var x: Int
    var y: Int


fn main() raises:
    comptime T = TestDefaultWritable[True, "\0'"]
    assert_equal(String(T(1, 2)), "TestDefaultWritable[True, \0'](x=1, y=2)")
```
output
```
 AssertionError: `left == right` comparison failed:
   left: TestDefaultWritable[True, {"'", 2, 2305843009213693952}](x=1, y=2)
  right: TestDefaultWritable[True, '](x=1, y=2)
```

### Steps to reproduce

...

### System information

`Mojo 0.26.2.0.dev2026012806`

martinvuyk:
Actually it seems like it's an issue of what `get_type_name` returns for strings:
```mojo
from reflection import get_type_name


struct TestDefaultWritable[param1: Bool, param2: String](Writable):
    ...


fn main() raises:
    comptime T = TestDefaultWritable[True, "hi"]
    print(get_type_name[T]())
```
outputs
```
asd.TestDefaultWritable[True, {"hi", 2, 2305843009213693952}]
```

JoeLoser:
> Actually it seems like it's an issue of what `get_type_name` returns for strings:
> 
> from reflection import get_type_name
> 
> 
> struct TestDefaultWritable[param1: Bool, param2: String](Writable):
>     ...
> 
> 
> fn main() raises:
>     comptime T = TestDefaultWritable[True, "hi"]
>     print(get_type_name[T]())
> outputs
> 
> ```
> asd.TestDefaultWritable[True, {"hi", 2, 2305843009213693952}]
> ```

The issue is that `get_type_name` returns the internal String struct representation (with all its fields like data, length, capacity) instead of just the string value when a struct has a String parameter.

Issue #5870, opened 2026-01-29
[Feature Request] Add `write_repr_to` implementations for stdlib types.


NathanSWard:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

## Background

The Mojo stdlib is transitioning from the `Stringable` and `Representable` traits to the unified `Writable` trait. As part of this transition, `Writable` now includes a `write_repr_to()` method that replaces the functionality of `Representable`.

## Problem

While `write_repr_to()` has a default implementation using reflection, this default is not suitable for most stdlib types. We need explicit implementations that provide:
- Clear, readable debug output
- Complete information about the type (including all parameters)
- Consistent formatting across the stdlib

**Note**: Some stdlib types may not yet implement `Writable` at all. For these types, we need to add the full trait implementation including both `write_to()` and `write_repr_to()` methods.

## Requirements

When implementing `write_repr_to()` for stdlib types:

1. **Print all useful information**: Include all type parameters and relevant internal state
2. **Recurse for containers**: Call `write_repr_to()` on element types for container types
3. **Use consistent formatting**: Leverage the helper utilities in `format/_utils.mojo` to ensure uniform style across the stdlib
4. **Tests**: Add appropriate unit tests to validate the output of each writing function.

For types that don't yet implement `Writable`:
- Add `Writable` conformance
- Implement `write_to()` for string representation (equivalent to the old `Stringable`)
- Implement `write_repr_to()` for debug representation (equivalent to the old `Representable`)

## Reference Implementations

The following structs already have correct `write_to`/`write_repr_to()` implementations that can serve as examples:
- `List`
- `Set` 
- `UnsafePointer`
- `Pointer`
- `StringSlice`

## Task

Audit the stdlib and implement `write_repr_to()` for all remaining types that either:
- Lack `Writable` conformance entirely
- Implement `Writable` but lack an explicit `write_repr_to()` (relying on the reflection default), or
- Have an incorrect/incomplete `write_repr_to()` implementation

Each implementation should follow the patterns established in the reference implementations above.

---

### High Priority - Core Types Need BOTH Methods

| Type | Module | File | Done |
|------|--------|------|-------|
| `FloatLiteral` | builtin | `builtin/floatable.mojo` | |
| `IntLiteral` | builtin | `builtin/int_literal.mojo` | |
[...]

PR #5874, opened 2026-01-29
[stdlib] Improve default `write_to` function implementations

martinvuyk:
Added tests for previously undiscovered bugs: #5871, #5872, #5873.

This should already make several types only need to use the default implementation for #5870

Changes in behavior:
- Slightly change the behavior of the default `write_to` implementation making it not print the name of the fields
- Make the default skip mlir fields
- SIMD now prints e.g. `SIMD[uint8, 2]([1, 2])`
- InlineArray now prints as a list `[1, 2]` and using repr `InlineArray[Int, 2]([1, 2])`

What is missing:
- A proper way to iterate over parameters and getting their write_repr_to/write_to behavior depending on which one was triggered
- A way to have the Writable interface trickle down the information of whether to use write_repr_to or write_to for nested structures, such that we can use it in all colection types in a generic way and provide a default implementation that does this. It would allow types to implement `write_to` only once

### DIFF
--- a/mojo/stdlib/std/format/__init__.mojo
+++ b/mojo/stdlib/std/format/__init__.mojo
@@ -34,7 +34,7 @@ struct Point(Writable):
     var y: Float64
 
 var p = Point(1.5, 2.7)
-print(p)  # Point(x=1.5, y=2.7)
+print(p)  # Point(1.5, 2.7)
 print(repr(p)) # Point(x=Float64(1.5), y=Float64(2.7))
 ```
 
@@ -80,6 +80,7 @@ from reflection import (
     struct_field_types,
     struct_field_count,
     get_type_name,
+    is_struct_type,
 )
 from reflection.type_info import _unqualified_type_name
 
@@ -167,7 +168,7 @@ trait Writable:
         var y: Int
 
     var p = Point(1, 3)
-    print(p)       # Point(x=1, y=3)
+    print(p)       # Point(1, 3)
     print(repr(p)) # Point(x=Int(1), y=Int(3))
     ```
 
@@ -191,6 +192,11 @@ trait Writable:
     ```
     """
 
+    comptime _type_name = _unqualified_type_name[Self]()
+    comptime _write_self_not_fields: Bool = False
+    """Whether to call `self.write_to(writer)` when writing the representation
+    instead of printing all fields' values."""
+
     fn write_to(self, mut writer: Some[Writer]):
         """Write this value's text representation to a writer.
 
@@ -212,14 +218,7 @@ trait Writable:
             writer.write("(", self.x, ", ", self.y, ")")
         ```
         """
-
-        @always_inline
-        fn call_write_to[
-            FieldType: Writable
-        ](field: FieldType, mut writer: type_of(writer)):
-            field.write_to(writer)
-
-        _reflection_write_to[f=call_write_to](self, writer)
+        _reflection_write_to(self, writer)
 
     fn write_repr_to(self, mut writer: Some[Writer]):
         """Write this value's debug representation to a writer.
@@ -247,46 +246,52 @@ trait Writable:
             of the repr. Any single quote inside a string should be escaped
             (`\\'`).
         """
-
-        @always_inline
-        fn call_write_repr_to[
-            FieldType: Writable
-        ](field: FieldType, mut writer: type_of(writer)):
-            field.write_repr_to(writer)
-
-        _reflection_write_to[f=call_write_repr_to](self, writer)
+        _reflection_write_to[write_repr=True](self, writer)
 
 
 @always_inline
 fn _reflection_write_to[
-    T: Writable,
-    W: Writer,
-    //,
-    f: fn[FieldType: Writable](field: FieldType, mut writer: W),
-](this: T, mut writer: W,):
+    T: Writable, //, *, write_repr: Bool = False
+](this: T, mut writer: Some[Writer]):
     comptime names = struct_field_names[T]()
     comptime types = struct_field_types[T]()
-    comptime type_name = _unqualified_type_name[T]()
-    writer.write_string(type_name)
+    writer.write_string(T._type_name)
     writer.write_string("(")
 
     @parameter
-    for i in range(names.size):
-        comptime FieldType = types[i]
-        _constrained_field_conforms_to[
-            conforms_to(FieldType, Writable),
-            Parent=T,
-            FieldIndex=i,
-            ParentConformsTo="Writable",
-        ]()
+    if T._write_self_not_fields and write_repr:
+        this.write_to(writer)
+    else:
 
         @parameter
-        if i > 0:
-            writer.write_string(", ")
-        writer.write_string(materialize[names[i]]())
-        writer.write_string("=")
+        for i in range(names.size):
+            comptime FieldType = types[i]
 
-        ref field = trait_downcast[Writable](__struct_field_ref(i, this))
-        f(field, writer)
+            @parameter
+            if not is_struct_type[FieldType]():  # skip mlir types
+                continue
+            _constrained_field_conforms_to[
+                conforms_to(FieldType, Writable),
+                Parent=T,
+                FieldIndex=i,
+                ParentConformsTo="Writable",
+            ]()
+
+            @parameter
+            if i > 0:
+                writer.write_string(", ")
+
+            @parameter
+            if write_repr and names.size > 1:
+                writer.write_string(materialize[names[i]]())
+                writer.write_string("=")
+
+            ref field = trait_downcast[Writable](__struct_field_ref(i, this))
+
+            @parameter
+            if write_repr:
+                field.write_repr_to(writer)
+            else:
+                field.write_to(writer)
 
     writer.write_string(")")
--- a/mojo/stdlib/std/builtin/bool.mojo
+++ b/mojo/stdlib/std/builtin/bool.mojo
@@ -107,6 +107,7 @@ struct Bool(
     comptime __del__is_trivial: Bool = True
     comptime __moveinit__is_trivial: Bool = True
     comptime __copyinit__is_trivial: Bool = True
+    comptime _write_self_not_fields = True
 
     # ===-------------------------------------------------------------------===#
     # Life cycle methods
@@ -215,18 +216,6 @@ struct Bool(
         """
 
         writer.write("True" if self else "False")
-
-    @no_inline
-    fn write_repr_to(self, mut writer: Some[Writer]):
-        """Writes the repr of this boolean to a writer.
-
-        The repr of a boolean is the same as its string representation:
-        `True` or `False`.
-
-        Args:
-            writer: The object to write to.
-        """
-        self.write_to(writer)
 
     fn __repr__(self) -> String:
         """Get the bool as a string.
--- a/mojo/stdlib/std/builtin/int.mojo
+++ b/mojo/stdlib/std/builtin/int.mojo
@@ -204,6 +204,8 @@ struct Int(
     # Aliases
     # ===-------------------------------------------------------------------===#
 
+    comptime _write_self_not_fields = True
+
     comptime BITWIDTH: Int = bit_width_of[DType.int]()
     """The bit width of the integer type."""
 
@@ -1003,14 +1005,6 @@ struct Int(
         """
 
         writer.write(Int64(self))
-
-    fn write_repr_to(self, mut writer: Some[Writer]):
-        """Write the string representation of the Int".
-
-        Args:
-            writer: The value to write to.
-        """
-        writer.write("Int(", self, ")")
 
     fn write_padded[W: Writer](self, mut writer: W, width: Int):
         """Write the int right-aligned to a set padding.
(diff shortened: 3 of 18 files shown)


mojo/stdlib/std/format/__init__.mojo
```diff
@@ -247,31 +240,33 @@ trait Writable:
...
-](this: T, mut writer: W,):
+    T: Writable, //, *, write_repr: Bool = False
+](this: T, mut writer: Some[Writer]):
     comptime names = struct_field_names[T]()
     comptime types = struct_field_types[T]()
     comptime type_name = _unqualified_type_name[T]()
     writer.write_string(type_name)
     writer.write_string("(")
 
+    fn _first_non_private() -> Int:
```
NathanSWard: I'm going to push back against not printing "private" fields.
I understand why this would be desirable for the `write_to` method, however Mojo doesn't have an official stance of what "private" fields are (since they don't exist). We just have the current practice of the `_` prefix denoting a private field.

I think we should wait until we officially have these and what their defined syntax before implementing this. :)

If someone really doesn't want their fields with `_` printed, they can simply add a custom `write_to` method.

mojo/stdlib/std/format/__init__.mojo
```diff
@@ -281,12 +276,20 @@ fn _reflection_write_to[
         ]()
 
         @parameter
-        if i > 0:
+        if i > _first_non_private():
             writer.write_string(", ")
-        writer.write_string(materialize[names[i]]())
```
NathanSWard, line 286: I'm also not totally sure on not printing the name of the fields in the default `write_to` method. I'm personally in favor of having the name as it can help differentiate between fields of the same type. Though I can also understand why if someone is trying to debug, they should probably just use `repr`.

mojo/stdlib/std/reflection/type_info.mojo
```diff
@@ -114,18 +114,26 @@ fn get_type_name[
     return StaticString(res)
 
 
-# TODO: This currently does not strip the module name from the inner type name.
-# For example, Generic[Foo] should return "Generic[Foo]" but currently returns
-# "Generic[module_name.Foo]".
-fn _unqualified_type_name[type: AnyType]() -> StaticString:
+fn _unqualified_type_name[type: AnyType]() -> String:
```
NathanSWard: Having this return a `String` now means any invocation of this will likely allocate. We ideally want to not allocate by returning a `StaticString` or by providing a writer to this function.

NathanSWard: I'm not sure the implementation of this is fully correct.
For examples cases like `module.Foo[other.Bar]`, and nested generics `Tuple[Tuple[Int, Bool], String]` I don't think work correctly with implementation.

If we do update this, can you please add some unit tests to verify the logic? 
I have a feeling this implementation may be a bit complex, so we it'll likely be nice to pull this out into it's own PR so we can ensure the edge cases are covered :)

NathanSWard (changes requested):

NathanSWard:
Ahh apologies if I jumped on gun on reviewing this, just realized it was a Draft...

mojo/stdlib/std/format/__init__.mojo
martinvuyk: A lot of our stdlib types have mlir values, so this would be way more tricky to implement for them by default. The point of having a default that is overwritable is for people to implement their own version if they're not happy with it. I would argue having the private fields not printed is an acceptable default that applies in more cases than not

mojo/stdlib/std/format/__init__.mojo
martinvuyk: > Though I can also understand why if someone is trying to debug, they should probably just use repr.

Yep, I was thinking more along those lines and reducing visual noise for the default `write_to` version

mojo/stdlib/std/reflection/type_info.mojo
martinvuyk: What is complicated is adding the logic for removing the fully qualified name from the inner parameters without using `String`, since every for loop iteration would create a new `StaticString`. I can try to do it again, I changed this previously because I found a bug for `Int` that I was trying to work around by using `DevicePassable.get_type_name()` which returns a `String`.

Having this function take in a writer would be very similar to what I'm trying to achieve by adding an interface for the Writer trait to be able to use functions and not just structs (#4793). It could also be solved by allowing a Writer to use a capturing function (#5274)

martinvuyk: Yeah I think you're right, we could leave this for another time. I think the implementation I did here would be able to deal only with 1 layer of parameter nesting. It will probable need to be recursive which always requires extensive testing :)

martinvuyk:
> Ahh apologies if I jumped on gun on reviewing this, just realized it was a Draft...

Don't worry I did start it off as a review-ready PR, but it seems I have more bugs to isolate besides the ones I found. It could also be that now that I'll remove the `_unqualified_type_name` changes it all just works :) :crossed_fingers: (update: it worked :D)

mojo/stdlib/std/format/__init__.mojo
NathanSWard: The workaround for dealing with MLIR types should be for us to use `is_struct_type` inside the reflection implementation for `Writable` and then either erroring out or printing `<unprintable>` or something like that.

Until mojo has real "public/private" fields, we should just print it all.
Especially if in the future the syntax is something like this:
```
struct Foo:
    private var x: Int
```

Relying on the stringified name of the field is not a good long term solution IMO.

martinvuyk: Ok how about this? We only print fields (including private named ones) if the type doesn't include a flag `_write_self_not_fields` set to true, and those types just call `this.write_to(writer)` instead of writing the fields (when calling repr)? This would help us unify all types like `UnsafePointer` and `Int` in one stroke

NathanSWard: > _write_self_not_fields

I still think this is over engineering a bit and bloating the `Writable` trait.
I generally think it is _okay_ for us to have to write custom `write_(repr_)to` implementations for stdlib types, especially those the use raw `__mlir_type`s.
If something has a `__mlir_type` then we should be expected to write a `write_to` method.

I'm not sure exactly how this would unify `UnsafePointer` & `Int` as they both have different MLIR types under the hood and need to be written differently: `UnsafePointer`: the address, `Int`: the numeric value.

I appreciate all of the ideas none the less - it's good to brainstorm 😄. Though I'd like to make small incremental changes, so for this PR I'm open to merging the "not printing field names" in `write_to`, and that would be a good first start.

martinvuyk: > I'm not sure exactly how this would unify UnsafePointer & Int as they both have different MLIR types under the hood and need to be written differently: UnsafePointer: the address, Int: the numeric value.

I just did it, you can see the changes in the PR. This unifies most types except for collection types that need to print the repr of their inner elements

>  If something has a __mlir_type then we should be expected to write a write_to method.

That's an alternative but having the flag would allow other custom types to take advantage of their `write_to` having some different representation than their field values

> I still think this is over engineering a bit and bloating the Writable trait.
> I generally think it is okay for us to have to write custom write_(repr_)to implementations for stdlib types, especially those the use raw __mlir_types.

I incorporated the changes and it allowed me to delete a lot of boilerplate and several different types' implementations since they followed the same pattern. Most of the code I've added is a workaround for an issue when printing parameter names, and editing tests for the new expected returns

setsource:
@martinvuyk, I’ve been following the progress here, spot on work! Your work on standardizing the infrastructure with `_write_self_not_fields` is exactly the architectural direction we are requesting for the `stdlib`. However, to fully close out the "Prettier" requirement in **#5838**, I see we have one last gap: **Alias Resolution.**

Currently, code changes print `SIMD[DType.uindex, 1]` for scalars. To satisfy the original request while respecting @NathanSWard’s preference for keeping the `Writable` trait lean, I make a humble proposal of dynamically resolving the `_type_name` within the `SIMD` struct itself. @martinvuyk please consider reviewing my suggestions.

---

- Proposed Refinement for `mojo/stdlib/std/builtin/simd.mojo`

By moving from a static string to a `comptime` helper, we give the reflection engine the "Pretty" name it needs without adding any complexity to the `Writable` trait logic.

**Current in PR:**

```mojo
comptime _type_name = String("SIMD[", Self.dtype, ", ", Self.size, "]")
comptime _write_self_not_fields = True

```

**Proposed Change:**

```mojo
# Inside struct SIMD:

@always_inline
@parameter
fn _get_type_name() -> String:
    @parameter
    if Self.size == 1:
        # Resolve to standard Mojo aliases for visual clarity
        if Self.dtype == DType.int64: return "Int64"
        if Self.dtype == DType.int32: return "Int32"
        if Self.dtype == DType.uint64: return "UInt64"
        if Self.dtype == DType.uint32: return "UInt32"
        if Self.dtype == DType.float64: return "Float64"
        if Self.dtype == DType.index: return "Int"
        if Self.dtype == DType.uindex: return "UInt"
        if Self.dtype == DType.bool: return "Bool"
        return String("Scalar[") + str(Self.dtype) + "]"
    
    return String("SIMD[") + str(Self.dtype) + ", " + str(Self.size) + "]"

comptime _type_name = Self._get_type_name()
comptime _write_self_not_fields = True

```

##### Why these suggestions:

* **Zero Over-Engineering:** As per @NathanSWard suggestions, we aren't changing the trait API or adding runtime overhead. Instead now we are simply providing a smarter `comptime` constant.
* **Satisfies #5838:** That `List[UInt]` will finally print as `[UInt(1), UInt(2)]` instead of the verbose details SIMD internal representation (prettier..)
* **Boilerplate Reduction:** We keep @martinvuyk's excellent work + logic for pruning the `stdlib` of redundant `write_repr_to` implementations.

what do you both think? If we're aligned on this, I hope this will unblock the merge and finally give us those cleaner debug logs we're after. Thanks.

martinvuyk:
@setsource thanks for the proposal, I was actually thinking on doing something way simpler:
```mojo
comptime _upper_idx = 1 + Int(Self.dtype.is_unsigned())
comptime _type_name = String("SIMD[", Self.dtype, ", ", Self.size, "]") if Self.size > 1 else (
    String(Self.dtype)[:Self._upper_idx].upper() + String(Self.dtype)[Self._upper_idx:]
)
```
I didn't implement this in this PR to reduce the scope and try to fix the issue in a separate one, but I could introduce it here if @NathanSWard approves :man_shrugging:

NathanSWard:
@martinvuyk - I don't believe the path of using a `comptime _write_self_not_fields: Bool` and the `compete _type_name` flags is the best approach here. The general idea is that the default reflection-based `write_to` implementation should be a "best guess" at how a type would want to format itself. If for some reason, the type doesn't like the default output it should provide a manual implementation.

This begins to bloat the surface of `Writable` which we should keep simple and make it work for "most cases".
A lot of the stdlib types are not these "most cases" - e.g. custom collections that abstract over pointers, or types that abstract over raw `MLIR` types. These are the exact types that _should_ provide a custom `write_(repr_)to` implementation.

The idea is we should be providing nice "format" utilities. This is was I was trying to achieve with the `FormatStruct/Named` etc... helper types that allow you to compose and build a nice formatter without having to worry about the nuances of adding appropriate whitespace or adding an `=` in the output.

martinvuyk:
@NathanSWard the `_type_name` was actually added as a workaround because the builtin `get_type_name` has issues (https://github.com/modular/modular/issues/5871, https://github.com/modular/modular/issues/5872, https://github.com/modular/modular/issues/5873), it shouldn't be needed to be overwritten for most types except those like SIMD which would like to change their representation based on some parameters.

This is intended to replace a lot of repetitive boilerplate, I don't see it as increasing the surface area by much (It's why I put the `_type_name` as a private alias). I'm just making the default implementation more flexible. I haven't looked into removing recently added custom implementations, but my guess is this will replace a lot more code than what I replaced the last time I rebased on top of the tree.

NathanSWard:
That all makes sense! Though I would instead push for fixing what is currently broken e.g. `_unqualified_type_name` rather than coming up with a temporary work around which is ideally not even needed if `_unqualified_type_name` is correctly fixed.

martinvuyk:
> That all makes sense! Though I would instead push for fixing what is currently broken e.g. `_unqualified_type_name` rather than coming up with a temporary work around which is ideally not even needed if `_unqualified_type_name` is correctly fixed.

@NathanSWard The thing is we will need for types like SIMD to be able to specify their own naming based on parameters if they so wish. I would like to eventually make `_type_name` public, but that decision can wait until post 1.0 . This is intended as a workaround/maybe future feature. I have use-cases for custom type naming (without having to write a bunch of boilerplate) easily come to mind for other community projects, I think this feature will be nice.

FWIW I tried patching just `_unqualified_type_name` by having `AnyType` have an alias with the `get_type_name` result but everything exploded, I also tried other approaches but it won't work because the `get_type_name` function returns are just a mess. Also, I've reported these bugs a month ago and I don't think they'll be quickly fixed. And the concern of stringifying a type name does seem to me like the responsibility of the `Writable` trait.

PR #5877, opened 2026-01-30
[stdlib] Take `write_sequence_to` into `VariadicPack` and use it

martinvuyk:
Take `write_sequence_to` into `VariadicPack` and use it in several call-sites improving readability :)

Also rename open and close delimiters to start and end to keep nomenclature consistent accross the codebase.

CC: @NathanSWard

### DIFF
--- a/mojo/stdlib/std/builtin/variadics.mojo
+++ b/mojo/stdlib/std/builtin/variadics.mojo
@@ -992,6 +992,46 @@ struct VariadicPack[
         """
         return __mlir_op.`kgen.pack.load`(self.get_as_kgen_pack())
 
+    fn write_to[
+        O1: ImmutOrigin = StaticConstantOrigin,
+        O2: ImmutOrigin = StaticConstantOrigin,
+        O3: ImmutOrigin = StaticConstantOrigin,
+    ](
+        self: VariadicPack[_, Writable, ...],
+        mut writer: Some[Writer],
+        start: StringSlice[O1] = rebind[StringSlice[O1]](StaticString("")),
+        end: StringSlice[O2] = rebind[StringSlice[O2]](StaticString("")),
+        sep: StringSlice[O3] = rebind[StringSlice[O3]](StaticString(", ")),
+    ):
+        """Writes a sequence of writable values from a pack to a writer with
+        delimiters.
+
+        This function formats a variadic pack of writable values as a delimited
+        sequence, writing each element separated by the specified separator and
+        enclosed by start and end delimiters.
+
+        Parameters:
+            O1: The origin of the open `StringSlice`.
+            O2: The origin of the close `StringSlice`.
+            O3: The origin of the separator `StringSlice`.
+
+        Args:
+            writer: The writer to write to.
+            start: The starting delimiter.
+            end: The ending delimiter.
+            sep: The separator between items.
+        """
+        writer.write_string(start)
+
+        @parameter
+        for i in range(self.__len__()):
+
+            @parameter
+            if i != 0:
+                writer.write_string(sep)
+            self[i].write_to(writer)
+        writer.write_string(end)
+
 
 # ===-----------------------------------------------------------------------===#
 # VariadicReduce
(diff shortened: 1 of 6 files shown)


mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -478,37 +478,15 @@ struct String(
...
-        for i in range(length):
-            args[i].write_to(total_bytes)
-
-            @parameter
-            if i < length - 1:
-                sep.write_to(total_bytes)
-        end.write_to(total_bytes)
-
-        if total_bytes.size == 0:
-            return String()
```
martinvuyk, line 492: I decided to drop this silly edge-case. I don't know what I was thinking, if somebody really wants to call `String.write()` on an empty string with empty separators and ending, then it's their problem

mojo/stdlib/std/builtin/variadics.mojo
```diff
@@ -992,6 +992,46 @@ struct VariadicPack[
...
+        O1: ImmutOrigin = StaticConstantOrigin,
+        O2: ImmutOrigin = StaticConstantOrigin,
+        O3: ImmutOrigin = StaticConstantOrigin,
+    ](
+        self: VariadicPack[_, Writable, ...],
+        mut writer: Some[Writer],
+        start: StringSlice[O1] = rebind[StringSlice[O1]](StaticString("")),
+        end: StringSlice[O2] = rebind[StringSlice[O2]](StaticString("")),
+        sep: StringSlice[O3] = rebind[StringSlice[O3]](StaticString(", ")),
+    ):
```
NathanSWard, line 1005: Quick question, I assume this syntax is used so we can pass `StringSlice` with different origins to this instead of only requiring `StaticConstantOrigin`?

Do we have a use case where we need a non-static start/end/sep? If so then this looks good to me, but if we don't IMO we should leave it as `StaticString` to make this a bit simpler :)

mojo/stdlib/std/builtin/variadics.mojo
```diff
@@ -992,6 +992,46 @@ struct VariadicPack[
         """
         return __mlir_op.`kgen.pack.load`(self.get_as_kgen_pack())
 
+    fn write_to[
```
NathanSWard, line 995: We may want to keep this a private function for now since it may end up conflicting with the `Writable.write_to` method.
Or if we do keep it public, having a better name to indicate this write this as a comma separate delimited sequence.

NathanSWard (commented):
Really nice clean up :)
Left a few comments, but others than those this looks good to me!

Can we also add unit tests for `VariadcPack.write_to` (or whatever the method's name ultimately is) so we can bind the behavior?

mojo/stdlib/std/builtin/variadics.mojo
martinvuyk: Yeah this is a trick I found last year and tried to incorporate it everywhere (#4414) but the PR has remained in the backlog for some reason. There are several cases in the stdlib itself, but this is very useful for external developers as well.

mojo/stdlib/std/builtin/variadics.mojo
martinvuyk: We have similar divergence to what is strictly the `Writer` interface in `String`. I think we could probably use defaulted trait implementations for many of those

> it may end up conflicting with the Writable.write_to method.

We would first need conditional trait conformance for `VariadicPack` to be able to be writable.

I also think this is another argument in favor of having a `FormatSpec` argument for the `Writable.write_to` interface, we could shove a lot of this complexity into it and unify several types' implementations

martinvuyk:
> Can we also add unit tests for VariadcPack.write_to (or whatever the method's name ultimately is) so we can bind the behavior?

`/stdlib/std/tests/format/test_utils.mojo` has a lot of tests for the overload of `write_sequence_to` that just uses `args.write_to(...)`

mojo/stdlib/std/builtin/variadics.mojo
NathanSWard: Yes, agreed. Custom format specs will be a thing for sure :)

NathanSWard:
Thanks for this :)

NathanSWard:
!sync

PR #5804, opened 2026-01-19
[stdlib] Optimize `String.count()` with single-pass SIMD algorithm

msaelices:
Replace the repeated `find()` calls in `String.count()` with a dedicated `_memmem_count` function that counts all non-overlapping occurrences in a single pass. The `_memmem_count` logic is based on the existing `_memmem_impl` one.

**Before:** each `find()` creates a new slice and restarts search
**After:** single traversal with SIMD-optimized scanning

## Changes

- Add `_memmem_count()` function with SIMD first/last character matching
- Simplify `count()` to delegate to the new function
- Add comprehensive benchmarks for various count scenarios

## Benchmark Results

| Scenario | 1K | 10K | 100K | 1M |
|----------|-----|------|-------|-----|
| Single char (`"a"`) | 1.0x | **2.3x** | 1.4x | 1.0x |
| Short substr (`"the"`) | **2.6x** | **2.3x** | 1.3x | 1.1x |
| High frequency (`"ab"`) | **2.3x** | **2.2x** | 1.3x | **1.3x** |

*Speedup vs main branch. Bold = ≥1.25x improvement.*

### Speedup Comparison

| Benchmark | 1K bytes | 10K bytes | 100K bytes | 1M bytes |
|-----------|----------|-----------|------------|----------|
| count_single_char | 1.03x | **2.26x** | 1.36x | 1.03x |
| count_short_substr | **2.58x** | **2.34x** | 1.33x | 1.13x |
| count_long_substr | 1.70x | **2.07x** | 1.05x | 0.89x |
| count_not_found | 1.52x | 1.45x | 0.85x | 0.71x |
| count_high_freq | **2.26x** | **2.17x** | 1.30x | **1.29x** |

*Values > 1.0x indicate the optimized version is faster. Bold indicates ≥1.25x speedup.*

### Conclusion

- **Best case:** 2.6x speedup for short substrings in small-medium strings
- **Typical case:** 1.3-2.3x speedup for strings with multiple matches
- **Worst case:** ~0.7x for not-found patterns in very large strings (rare use case)

PR #5882, opened 2026-01-31
[Stdlib] Use single-pass algorithm in StringSlice.replace()

msaelices:
## Summary
- Replace the two-pass approach (count then find) with a single-pass algorithm that finds and replaces in one traversal
- Return early without allocating when no match is found
- Add `bench_string_replace_no_match` and `bench_string_replace_multi_char` benchmarks

## Benchmarks

Here are the results with 5 repetitions (using median of 5 runs for stability):
```
  ┌──────────────────────────────┬───────────┬────────────────┬─────────┐
  │          Benchmark           │ main (ms) │ optimized (ms) │ Speedup │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_single_char[1000]    │ 0.00584   │ 0.00388        │ 1.51x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_multi_char[1000]     │ 0.00146   │ 0.00093        │ 1.57x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_no_match[1000]       │ 0.00027   │ 0.00019        │ 1.42x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_grow[1000]           │ 0.00534   │ 0.00393        │ 1.36x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_shrink[1000]         │ 0.00132   │ 0.00093        │ 1.42x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_single_char[10000]   │ 0.05216   │ 0.03991        │ 1.31x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_multi_char[10000]    │ 0.01465   │ 0.00641        │ 2.29x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_no_match[10000]      │ 0.00112   │ 0.00073        │ 1.53x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_grow[10000]          │ 0.04402   │ 0.01839        │ 2.39x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_shrink[10000]        │ 0.01344   │ 0.00626        │ 2.15x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_single_char[100000]  │ 0.36175   │ 0.20307        │ 1.78x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_multi_char[100000]   │ 0.12577   │ 0.07384        │ 1.70x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_no_match[100000]     │ 0.00851   │ 0.00819        │ 1.04x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_grow[100000]         │ 0.34158   │ 0.19862        │ 1.72x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_shrink[100000]       │ 0.14677   │ 0.06727        │ 2.18x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_single_char[1000000] │ 3.85886   │ 2.98504        │ 1.29x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_multi_char[1000000]  │ 1.20595   │ 0.82984        │ 1.45x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_no_match[1000000]    │ 0.10425   │ 0.11750        │ 0.89x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_grow[1000000]        │ 3.28581   │ 2.24118        │ 1.47x   │
  ├──────────────────────────────┼───────────┼────────────────┼─────────┤
  │ replace_shrink[1000000]      │ 1.25597   │ 0.81103        │ 1.55x   │
  └──────────────────────────────┴───────────┴────────────────┴─────────┘
```

The single-pass algorithm is faster across nearly all cases.


mojo/stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -219,6 +219,40 @@ fn bench_string_replace[
...
+        var res = black_box(items).replace(
+            black_box("ZZZZZZ"), black_box("YYYYYY")
+        )
+        keep(res)
+
+    b.iter(call_fn)
+
+
+@parameter
+fn bench_string_replace_multi_char[
```
NathanSWard, line 240: Thanks for the benchmarks! Curious can you attach results you saw before & after the change with these benchmarks?

msaelices: Will do. BTW this is a continuation of this work: https://github.com/modular/modular/pull/5804

martinvuyk:
@NathanSWard @msaelices I've had  #3860 be in the backlog for 2 years now...

msaelices:
> @NathanSWard @msaelices I've had #3860 be in the backlog for 2 years now...

@martinvuyk, OMG. totally my bad. I had completely forgotten about it. Do you think both optimizations could be complementary?

mojo/stdlib/benchmarks/collections/bench_string.mojo
msaelices: done

martinvuyk:
> > @NathanSWard @msaelices I've had #3860 be in the backlog for 2 years now...
> 
> @martinvuyk, OMG. totally my bad. I had completely forgotten about it. Do you think both optimizations could be complementary?

I think your implementation will be interesting to test out after merging the one I did, because I'm getting 2-4x improvements but I'm also changing the interleave function which was suboptimal. There is pobably a point where estimating the required size and potentially having reallocations is worse the bigger the string and the cost of counting the occurences gets amortized. But we would need to compare only that change in both implementations equally, so I would suggest we wait until the other one is merged.

We could also add an optimization that I hadn't thought of, if the new string is smaller than or equal in length to the old string then we wouldn't need to count at all.

As a bit of a nit-pick with your benchmarks, I would try to just reuse the existing replace test and pass different width old and new strings to test growing vs shringking

PR #5883, opened 2026-01-31
[stdlib] Revamp the `_FormatSpec` logic for a new direction

### LINKED ISSUE #5881: [Feature Request] [proposal] New and simpler `FormatSpec` for the `Writable` & `Writer` interfaces
### Review Mojo's priorities

### What is your request?

We currently have some part of Python's format strings implemented:
- struct `_FormatSpec` has several flags (none are currently implemented and they aren't parsed from a fmt string):
  - `fill: UInt8`: If a valid align value is specified, it can be preceded by a fill character that can be any character and defaults to a space if omitted
  - `align: UInt8`: Various alignment options
  - `sign: UInt8`: Defines the treatment of signs for numeric types
  - `coerce_z: Bool`: coerces negative zero floating-point values to positive zero
  - `alternate_form: Bool`: The alternate form is defined differently for different types. For integers, when binary, octal, or hexadecimal output is used, this option adds the respective prefix
  - `width: UInt8`: Defines the minimum total field width
  - `grouping_option: UInt8`: The ',' option signals the use of a comma for a thousands separator. For
    integer presentation types 'b', 'o', 'x', and 'X', underscores will be inserted every 4 digits
  - `precision: UInt8`: how many digits should be displayed after the decimal point
  - `type: UInt8`: which integer, floating point, and date-like types' representation format to use
- struct `_FormatCurlyEntry`:
  - supports 2 conversion flags "s" (normal stringify) or "r" (repr)

### Proposed alternative

Python's format spec, while amazing and flexible, is practically a mini-language unto itself ([their own docs call it so](https://docs.python.org/3/library/string.html#format-specification-mini-language)). Python has a single-type type system, so it can be very dynamic. I propose we rework the format spec to better suit Mojo's capabilities, and also simplify it for the new age.

We can make `FormatSpec` also mean the conversion flag in the specification i.e. Python's
`replacement_field: "{" [field_name] ["!" conversion] [":" format_spec] "}"`
becomes: `replacement_field: "{" [field_name] [MojoFormatSpec] "}"`

Conversion flags have been up until now only "s" (default) and "r" (repr), python also has "a" (ascii) which we haven't implemented yet. I think we should allow other types to implement their own custom conversion flags, this would enable things like for example built-in feeling json serialization: `"{!j}".format(SomeStruct())` (or even also pretty printing) to be built outside the stdlib but use it without an issue; we would of course need to add a check that the conversion flags are supported by each given type.

Let's have a new `FormatSpec` struct with the fields:
- `conversion: UInt8`: The conversion flag to use, defaults to the ascii byte representing "s" i.e. `Byte(ord("s"))`
- `fill: Codepoint`: If a valid align value is specified, it can be preceded by a fill codepoint that can be any codepoint and defaults to a space if omitted
- `align: AlignmentOptions`: The same alignment options as in python, backend by a `UInt8` enum. When parsing it is assumed to fit in it and represent the raw value. We can have e.g. `"<"` be represented by `Byte(ord("<"))`
- `grapheme_width: Int`: Defines the minimum total field width. But this one takes into account what the printed width looks like by using the grapheme width as the specified value
- `numeric_options: Optional[NumericOptions]`: All of the Integer and Floating point specifications

struct `NumericOptions` would have fields:
- `sign: SignOptions`: Defines the treatment of signs for numeric types.
- `float_options: Optional[FloatOptions]`
- `thousand_separator: UInt8`: Thousand separator ascii character
- `numeric_form: Variant[IntForm, FloatForm]`: The alternative form to print the integer as. Defaults to decimal for integers, a more complex set of criteria for floating point

struct `FloatOptions` would have fields:
- `coerce_z: Bool`: The 'z' option coerces negative zero floating-point values to positive zero after rounding to the format precision
- `precision: UInt8`: How many digits should be displayed after the decimal point, defaults to max.

`FloatForm` and `IntForm` would have all of the same alternatives as in Python, but represented as enums of `UInt8` with the corresponding ascii character value as the string representation of each form, e.g. binary `"b"` be represented by `Byte(ord("b"))`.

It is practically impossible to implement nested structures that should print the write_repr_to or the write_to versions of their elements generically with our current setup.

### Any other details?

I would happily tackle this

We can discuss later on exactly how to incorporate this into the `Writer` and `Writable` traits. I'm leaning towards passing the format spec as a parameter to avoid the runtime branching and have the format spec fail at compile time if it isn't correctly configured for a given type

CC: @NathanSWard

martinvuyk:
Revamp the `_FormatSpec` logic for a new direction. Closes #5881

This is all in preparation for the switch to having each type define its format spec type, with a default one for SIMD (numeric), and the other one for all other `Writable` types by default.

The new Mojo format specifications deviate very little from Python. Where they do deviate is to make the format specification parsing less convoluted. This can be revisited in the future.

CC: @NathanSWard

### DIFF
--- a/mojo/stdlib/std/collections/string/format.mojo
+++ b/mojo/stdlib/std/collections/string/format.mojo
@@ -39,7 +39,8 @@ indicate whether the argument should be formatted using `repr()` or `String()`,
 respectively:
 
 ```mojo
-var s = "{!r}".format(myComplicatedObject)
+%# var some_object = String()
+var s = "{!r}".format(some_object)
 ```
 
 Note that the following features from Python's `str.format()` are
@@ -252,7 +253,9 @@ struct _FormatUtils:
 # And going a step further it might even be worth it adding custom format
 # specification start character, and custom format specs themselves (by defining
 # a trait that all format specifications conform to)
-struct _FormatCurlyEntry[origin: ImmutOrigin](ImplicitlyCopyable):
+struct _FormatCurlyEntry[
+    origin: ImmutOrigin, spec_type: _FormatSpecType = _DefaultFormatSpec
+](ImplicitlyCopyable):
     """The struct that handles string formatting by curly braces entries.
     This is internal for the types: `StringSlice` compatible types.
     """
@@ -261,14 +264,8 @@ struct _FormatCurlyEntry[origin: ImmutOrigin](ImplicitlyCopyable):
     """The index of an opening brace around a substitution field."""
     var last_curly: Int
     """The index of a closing brace around a substitution field."""
-    # TODO: ord("a") conversion flag not supported yet
-    var conversion_flag: UInt8
-    """The type of conversion for the entry: {ord("s"), ord("r")}."""
-    # TODO: ord("a") conversion flag not supported yet
-    comptime supported_conversion_flags = SIMD[DType.uint8, 2](
-        UInt8(ord("s")), UInt8(ord("r"))
-    )
-    """Currently supported conversion flags: `__str__` and `__repr__`."""
+    var format_spec: Self.spec_type
+    """The format specifier."""
     comptime _FieldVariantType = Variant[
         StringSlice[Self.origin], Int, NoneType, Bool
     ]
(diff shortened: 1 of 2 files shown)


NathanSWard:
Just as a heads up I have a change coming in the next nightly that will likely conflict with this. It removed the _FormatSpec type & parsing (similar to this PR) since I'm trying to get us to a point where we have can f/t-strings that aren't too slow.

Issue #5666, opened 2025-12-16
[Feature Request] Migrate `kernels/src/layout` off of `LegacyUnsafePointer`.


NathanSWard:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](<https://docs.modular.com/mojo/roadmap.html#overall-priorities>) and I believe this request falls within the priorities.

### What is your request?

See [parent issue](<https://github.com/modular/modular/issues/5665>)

The following kernel files should be migrated off of `LegacyUnsafePointer`. (complexity noted next to file)

```
layout/_mixed_tuple.mojo (Low)
layout/_utils.mojo (Low)
layout/layout_tensor.mojo (Very High)
layout/tma_async.mojo (High)
```

Notes<br>`layout_tensor.mojo`: This is likely the most complex and most involved change as the entire kernel library uses `LayoutTensor`, this is likely best saved for last (or at least toward the end).

Ramshankar07:
Hi @NathanSWard , Can you assign this to me?? and Can I start with Low priority then High ( tma_async )?

NathanSWard:
Hi @Ramshankar07 thanks for wanting to pick this up!
I personally would save this sub-issue to the end (or at least `LayoutTensor` as this is arguably the most difficult of them to address).

Or if you want, you can take on 
```
layout/_mixed_tuple.mojo (Low)
layout/_utils.mojo (Low)
layout/tma_async.mojo (High)
```
(minus `layout_tensor.mojo` to start)

Ramshankar07:
That make sense, I would like to start with 
`
layout/_mixed_tuple.mojo (Low),
layout/_utils.mojo (Low),
layout/tma_async.mojo (High)
` in a single PR. Can you assign it to me?

Ramshankar07:
layout/_mixed_tuple.mojo file is missing

PR #5884, opened 2026-02-01
[Kernels] Migrated utils and tma_async from LegacyUnsafePointer (#5666)

Ramshankar07:
## Summary
Migrates `layout/tma_async.mojo` and `layout/_utils.mojo` from `LegacyUnsafePointer` to the new `UnsafePointer`. for issue #5666 

## Changes

**Imports**
- **tma_async.mojo:** `from memory import MutOpaquePointer, UnsafePointer`.
- **_utils.mojo:** `from memory import UnsafePointer` (uses new type in `hash()`).

UnsafePointer usage
- All pointer types now specify an **origin** (`MutAnyOrigin` or `ImmutAnyOrigin`) and, where needed.
- `UnsafePointer(to=...)` requires a reference; every method that took the address of `self.descriptor` now takes `ref self` so the constructor gets a ref.
- Descriptor pointers are created with `ImmutAnyOrigin` so they work with `ref self`; casts to `MutAnyOrigin` are added only where the API requires it.

TensorMapDescriptorArray._get_descriptor_ptr
- Builds the pointer with `ImmutAnyOrigin(to=self.descriptor)` then `.bitcast[NoneType]().unsafe_mut_cast[True]().unsafe_origin_cast[MutAnyOrigin]()` so it can take `ref self` and still return the mutable descriptor pointer when called on rvalues (e.g. `prefetch_descriptor(self)`).

- Replaced `UnsafePointer(to=val)` with `UnsafePointer[type_of(val), ImmutAnyOrigin](to=val)` so the pointer has a defined origin.

### Testing
- `./bazelw build //max/kernels/src/layout:layout` passes.


Ramshankar07:

NathanSWard:
@Ramshankar07 - I see this is still a draft. Is this ready for review yet?

Ramshankar07:
yes

max/kernels/src/layout/tma_async.mojo
```diff
@@ -443,28 +444,57 @@ struct SharedMemBarrier(TrivialRegisterType):
 
     @always_inline
     fn unsafe_ptr[
-        origin: Origin
+        origin: ImmutOrigin
```
NathanSWard: **question**: Is this change needed for this PR? Ideally we should have the origin of the returned pointer be equal to the origin of this type (including the matching of mutabilities).
Did you by change hit a compiler error?

max/kernels/src/layout/tma_async.mojo
```diff
@@ -443,28 +444,57 @@ struct SharedMemBarrier(TrivialRegisterType):
...
             An unsafe pointer to the barrier's memory location in shared memory,
             properly typed and aligned for barrier operations.
         """
-        return UnsafePointer(to=self.mbar).unsafe_origin_cast[origin]()
+        return UnsafePointer[
+            Int64, ImmutAnyOrigin, address_space = AddressSpace.SHARED
+        ](to=self.mbar).unsafe_origin_cast[origin]()
+
+    @always_inline
+    fn unsafe_ptr[
```
NathanSWard: Ahh I see, you shouldn't need this overload if you leave the other one generic over the Origin's mutability `origin: Origin`. That should let you delete this overload and only use the previous one :)

max/kernels/src/layout/tma_async.mojo
```diff
@@ -734,22 +764,24 @@ struct TMATensorTile[
...
         """
         Prefetches the TMA descriptor into cache to reduce latency.
 
         This method helps hide memory access latency by prefetching the descriptor
         before it's needed for actual data transfers.
         """
-        var desc_ptr = UnsafePointer(to=self.descriptor).bitcast[NoneType]()
+        var desc_ptr = UnsafePointer[TMADescriptor, ImmutAnyOrigin](
+            to=self.descriptor
+        ).bitcast[NoneType]()
```
NathanSWard: For all of these cases in this file, you shouldn't need to to specify the specific origin here. As the `to=...` syntax will actually infer the origin for you!
Also ditto with the type, it can also be inferred!

```suggestion
        var desc_ptr = UnsafePointer(to=self.descriptor).bitcast[NoneType]()
```

max/kernels/src/layout/tma_async.mojo
```diff
@@ -734,22 +764,24 @@ struct TMATensorTile[
...
         prefetch_tma_descriptor(desc_ptr)
 
     @always_inline
     fn async_copy[
         cta_group: Int = 1,
         eviction_policy: CacheEviction = CacheEviction.EVICT_NORMAL,
     ](
-        self,
+        ref self,
```
NathanSWard: Are these `ref` argument conventions needed? I don't believe they should be a they do also change the semantics of this function to work on immutable or mutable references to self. Were as just `self` (without the `ref`) requires the value to be _immutable_.

max/kernels/src/layout/tma_async.mojo
```diff
@@ -2130,9 +2180,9 @@ struct TMATensorTile[
         *,
         rank: Int,
     ](
-        self,
+        ref self,
         smem_tma_descriptor_ptr: UnsafePointer[
-            TMADescriptor, address_space = AddressSpace.SHARED, ...
+            TMADescriptor, MutAnyOrigin, address_space = AddressSpace.SHARED
```
NathanSWard: This is a slight preference, but if possible when you have an `UnsafePointer` as an argument to a function, instead of manually specifying the origin, you can use `mut=True` (if and only if the pointer actually needs to be mutable).

```suggestion
            mut=True, TMADescriptor, address_space = AddressSpace.SHARED
```

NathanSWard (changes requested):
Thanks for this!
I left an initial round a feedback. Please let me know if you have any questions I can help answer, but hopefully these make sense.

max/kernels/src/layout/tma_async.mojo
Ramshankar07: I wanted the returned pointer’s mutability to exactly mirror the barrier’s origin and I wasn’t sure the more generic origin: Origin signature would type‑check. I'll revert it to origin: Origin and delete the extra one.

max/kernels/src/layout/tma_async.mojo
Ramshankar07: I already got some compiler errors around UnsafePointer(to=...) (because of self not being ref, etc.), so it felt safer to specify the arguments. I'll remove it :}

max/kernels/src/layout/tma_async.mojo
Ramshankar07: reverting back to self, except at some place where we are reference using ptr. Like _get_descriptor_ptr(ref self ) and TensorMapDescriptorArray.prefetch_descriptor(ref self) which calls get descriptor ptr

NathanSWard:
Thank you for taking a stab at this 😄! We ended up finishing up this migration internally, so I'm going to close this PR.

Ramshankar07:
Cool!

PR #5891, opened 2026-02-01
[MAX] Add Flux.2 DiT models for Flux.2 Pipeline

byungchul-sqzb:
## Overview

This PR adds the Flux2 DiT (Diffusion Transformer) model stack required for the Flux2 pipeline. It introduces a new Flux2 architecture package with the main transformer model implementation, attention mechanisms, and supporting layers.

## What's included

- **New Flux2 architecture package** under `max/python/max/pipelines/architectures/flux2/`:
  - `Flux2Transformer2DModel`: Main DiT model with dual-stream and single-stream transformer blocks
  - `Flux2Model`: ComponentModel wrapper for pipeline integration
  - `Flux2Config`: Configuration class for Flux2 model parameters
  - **Layers**:
    - `Flux2Attention`: Multi-head attention with image/text conditioning
    - `Flux2FeedForward`: Feed-forward with SwiGLU activation
    - `TimestepEmbedding`: Diffusion timestep conditioning
    - `AdaLayerNormContinuous`: Adaptive layer normalization
    - Positional embeddings and normalization utilities

## NOTES

- This model implementation does NOT serve as a standalone executable model. It is expected to be executed within the Flux2 pipeline.
- Definition of the base abstract class ComponentModel is defined in the pipeline interfaces module.


max/python/max/pipelines/architectures/flux2/model.py
```diff
@@ -0,0 +1,149 @@
...
+                *input_types, weights=self._state_dict
+            )
+            # compile returns a callable, but we need to store it as Model
+            # The actual Model is created when the callable is invoked
+            self._compiled_model = compiled  # type: ignore[assignment]
+            self._compiled_shapes = current_shapes
+
+        if self._compiled_model is None:
+            raise RuntimeError("Model compilation failed")
+        return self._compiled_model
```
kkimmk: Is this method essential? When should we handle compilation explicitly?

max/python/max/pipelines/architectures/flux2/model.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def __call__(
+        self,
+        hidden_states: Any,
+        encoder_hidden_states: Any,
+        timestep: Any,
+        img_ids: Any,
+        txt_ids: Any,
+        guidance: Any,
+    ) -> Any:
```
kkimmk: Can we specify exact types rather than `Any` here?

max/python/max/pipelines/architectures/flux2/model.py
```diff
@@ -0,0 +1,149 @@
...
+            enc_shape = encoder_hidden_states.shape
+            text_seq_len = (
+                enc_shape[1].dim
+                if hasattr(enc_shape[1], "dim")
+                else enc_shape[1]
+            )
+        else:
+            raise ValueError(
+                "encoder_hidden_states must have a shape attribute"
+            )
```
kkimmk: I guess we currently accept `max.tensor.Tensor` only. Could you check if it's true in Flux2 pipeline and if the logic for `driver.Buffer` can be removed?

max/python/max/pipelines/architectures/flux2/flux2.py
```diff
@@ -0,0 +1,707 @@
...
+        )
+
+        return (
+            hidden_states_type,
+            encoder_hidden_states_type,
+            timestep_type,
+            img_ids_type,
+            txt_ids_type,
+            guidance_type,
+        )
```
kkimmk: I guess it would be enough to use the symbolic shapes, `input_types`, and this method can be safely removed. If it's confirmed that it really affects the runtime performance, I believe we'd better analyze it separately and report it.

max/python/max/pipelines/architectures/flux2/layers/normalizations.py
```diff
@@ -0,0 +1,195 @@
...
+            else Tensor.zeros(
+                self.normalized_shape, dtype=x.dtype, device=x.device
+            )
+        )
+        return F.layer_norm(
+            x,
+            gamma=gamma,
+            beta=bias,
+            epsilon=self.eps,
+        )
```
kkimmk: Would it be possible to remove these custom Norms and use the existing `nn.norm` instead? It was possible with Flux1.

max/python/max/pipelines/architectures/flux2/layers/flux2_attention.py
```diff
@@ -0,0 +1,533 @@
...
+        heads: int = 8,
+        dim_head: int = 64,
+        dropout: float = 0.0,
+        bias: bool = False,
+        added_kv_proj_dim: int | None = None,
+        added_proj_bias: bool | None = True,
+        out_bias: bool = True,
+        eps: float = 1e-5,
+        out_dim: int | None = None,
+        elementwise_affine: bool = True,
```
kkimmk: If these [custom Norms can be removed](elementwise_affine), `elementwise_affine` should also be removable.

max/python/max/pipelines/architectures/flux2/model.py
byungchul-sqzb: I've removed it.

max/python/max/pipelines/architectures/flux2/model.py
byungchul-sqzb: Done — the parameters are now typed as Tensor.

max/python/max/pipelines/architectures/flux2/model.py
byungchul-sqzb: Yes, that logic can be removed. It’s no longer needed anyway, since we removed _ensure_compiled and no longer do any shape extraction in __call__.

max/python/max/pipelines/architectures/flux2/flux2.py
byungchul-sqzb: Done. I’ve removed it.

max/python/max/pipelines/architectures/flux2/layers/normalizations.py
byungchul-sqzb: Done. I replaced custom WeightedRMSNorm and WeightedLayerNorm with max.nn.norm.RMSNorm and max.nn.norm.LayerNorm.

max/python/max/pipelines/architectures/flux2/layers/flux2_attention.py
byungchul-sqzb: Done.

katelyncaldwell (approved):
Thanks!

katelyncaldwell:
!sync

katelyncaldwell:
Merged!

PR #5895, opened 2026-02-02
[Docs] Update variables.mdx with regard to scope of explicit and implicit

DerekVeit:
The current descriptions of the distinction between implicitly declared and explicitly declared variables is using "lexical scope" to refer to block-level scope.  But it is lexical scoping in both cases, with the difference being in the granularity, block-level or function-level.

I think a larger edit to better organize these descriptions might be helpful too.


DerekVeit:

fairlydurable:
!sync

DerekVeit:
recheck

PR #5899, opened 2026-02-03
[stdlib][optional] add explicit write_repr_to and tests

lamija-i:
This PR adds an explicit write_repr_to implementation for Optional in the stdlib, replacing the default reflection-based behavior.

The new implementation produces a clear and consistent debug representation and delegates to write_to for the contained value when present. This aligns Optional with other stdlib types that already define custom write_repr_to behavior.

Unit tests are included to cover both simple and nested Optional cases.

Related issue:
https://github.com/modular/modular/issues/5870

### DIFF
--- a/mojo/stdlib/std/collections/optional.mojo
+++ b/mojo/stdlib/std/collections/optional.mojo
@@ -467,11 +467,10 @@ struct Optional[T: Movable](
             writer.write(trait_downcast[Representable](self.value()).__repr__())
         else:
             writer.write("None")
-
     
-        fn write_repr_to(self: Self, mut writer: Some[Writer]):
-        FormatStruct(writer, "Optional").fields(self)
-        
+    fn write_repr_to(self: Self, mut writer: Some[Writer]):
+    FormatStruct(writer, "Optional").fields(self)
+
     # ===-------------------------------------------------------------------===#
     # Methods
     # ===-------------------------------------------------------------------===#
--- a/mojo/stdlib/std/collections/optional.mojo
+++ b/mojo/stdlib/std/collections/optional.mojo
@@ -467,6 +467,13 @@ struct Optional[T: Movable](
         else:
             writer.write("None")
 
+    fn write_repr_to[
+        U: Representable & Copyable & Movable, //
+    ](self: Optional[U], mut writer: Some[Writer]):
+        writer.write("Optional(")
+        self.write_to(writer)
+        writer.write(")")
+
     # ===-------------------------------------------------------------------===#
     # Methods
     # ===-------------------------------------------------------------------===#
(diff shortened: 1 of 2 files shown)


NathanSWard:
@lamija-i Could you please sign the CLA as commented here: https://github.com/modular/modular/pull/5899#issuecomment-3841114905
I can then get this merged into the repo :)

mojo/stdlib/std/collections/optional.mojo
```diff
@@ -467,6 +467,13 @@ struct Optional[T: Movable](
         else:
             writer.write("None")
 
+    fn write_repr_to[
+        U: Representable & Copyable & Movable, //
+    ](self: Optional[U], mut writer: Some[Writer]):
+        writer.write("Optional(")
+        self.write_to(writer)
+        writer.write(")")
```
NathanSWard: We shouldn't need to constrain `U` to `Representable`. If you look at the implementation of `write_to` we can model that by using `_constrained_conforms_to`.

Also, to keep this formatting consistent across the stdlib, ideally we should use the formatting helpers in `format/_utils.mojo`
With that, the implementation would look as such:
```suggestion
    fn write_repr_to(self: Self, mut writer: Some[Writer]):
        FormatStruct(writer, "Optional").fields(self)
```

mojo/stdlib/test/collections/test_optional.mojo
```diff
@@ -159,6 +159,12 @@ def test_optional_str_repr():
     assert_equal(Optional[Int](None).__str__(), "None")
     assert_equal(Optional[Int](None).__repr__(), "Optional(None)")
 
+def test_optional_write_repr_to():
+    var o1 = Optional[Int](None)
+    var o2 = Optional(3)
+
+    assert_equal(repr(o1), "Optional(None)")
+    assert_equal(repr(o2), "Optional(3)")
```
NathanSWard: Unfortunately, this doesn't actually call the `write_repr_to` method, it actually delegates to `Representable`.
To get this to work, can you directly call `write_repr_to` instead and then check the output?

NathanSWard (changes requested):

lamija-i:

lamija-i:
Thanks for the feedback!  
I’ve signed the CLA and pushed the updated changes addressing your comments.

mojo/stdlib/std/collections/optional.mojo
```diff
@@ -466,6 +467,9 @@ struct Optional[T: Movable](
             writer.write(trait_downcast[Representable](self.value()).__repr__())
         else:
             writer.write("None")
+    
+    fn write_repr_to(self: Self, mut writer: Some[Writer]):
+    FormatStruct(writer, "Optional").fields(self)
```
NathanSWard, line 472: Odd that the formatter didn't add a tab here, but I can just fix this internally!
Thanks for the PR and I'll get this merged in.

NathanSWard (approved):

NathanSWard:
!sync

PR #5904, opened 2026-02-04
[MAX] Fix diffusers config to properly load components configs and to consider local path.

kkimmk:
This PR fixes two issues that are not properly handled in the current diffusers config implementation.

- loads configs of components having a name different from `config.json`. e.g., `scheduler_config.json`.
- loads configs properly when pre-downloaded local path is given.


KCaverly:
!sync

PR #5913, opened 2026-02-07
[Stdlib] Add `write_repr_to` to Counter

msaelices:
## Summary
- Add `write_repr_to` to `Counter` following the same pattern used by `List`, `Deque`, and other collections
- Refactor `write_to` to use a shared `_write_self_to` helper parameterized on key/value formatting functions.
- Update `__repr__` to delegate to `write_repr_to`.


mojo/stdlib/std/collections/counter.mojo
```diff
@@ -256,44 +259,71 @@ struct Counter[V: KeyElement, H: Hasher = default_hasher](
...
+            if index >= len(items):
+                raise StopIteration()
+            ref item = items[index]
+            write_key(item._value, w)
+            w.write_string(": ")
+            write_value(item._count, w)
+            index += 1
+
+        fmt.write_sequence_to[ElementFn=iterate](writer, start="{", end="}")
+        _ = items^
```
NathanSWard, line 288: Super clean implementation and nice use of the formatting utilities :) 

One small suggestion, when we write the `item._count` I'd suggest _always_ using the `write_to` method of `Int` instead of needing the repr. Writing the `repr` of the count isn't super useful, as we're only really concerned about writing the repr of the key.
This would change the output of repr to look as such
```
"Counter[Int]({Int(3): 3, Int(2): 2, Int(1): 1})"
``` 


```suggestion
    fn _write_self_to[
        write_key: fn(Self.V, mut Some[Writer]),
    ](self, mut writer: Some[Writer]):
        fmt.constrained_conforms_to_writable[Self.V, Parent=Self]()

        var items = self.most_common(UInt(len(self)))
        var index = 0

        @parameter
        fn iterate(mut w: Some[Writer]) raises StopIteration:
            if index >= len(items):
                raise StopIteration()
            ref item = items[index]
            write_key(item._value, w)
            w.write_string(": ")
            item._count.write_to(w)
            index += 1

        fmt.write_sequence_to[ElementFn=iterate](writer, start="{", end="}")
        _ = items^
```

NathanSWard (commented):
After the small suggestion, this looks great and I'll merge it in 😄

mojo/stdlib/std/collections/counter.mojo
msaelices: Great suggestion. Done: https://github.com/msaelices/modular/commit/1a559969200d5a889d4bbc98c05020c663444677

NathanSWard (approved):
thank you! :)

NathanSWard:
!sync

NathanSWard:
Thank you for this PR 😄 , however we ended up addressing this internally!

PR #5918, opened 2026-02-07
[Stdlib] Add SIMD-optimized List.__contains__() for scalar types

msaelices:
## Summary
- Delegate `List.__contains__` for scalar types to `Span.__contains__` instead of duplicating SIMD logic
- Fix `Span.__contains__` SIMD width cascade to include widths 4 and 2, enabling SIMD for 64-bit types on AVX2
- Add benchmarks for Int, Float64, and UInt8 across sizes 10-100k (50 repetitions)

## Benchmark results (main vs this branch)

Machine: x86_64, Intel i7-12700H (AVX2, 256-bit SIMD), 50 repetitions

| Benchmark | Size | main (ms) | branch (ms) | Speedup |
|---|---|---|---|---|
| **Float64** | 10 | 0.000924 | 0.000404 | **2.3x** |
| **Float64** | 50 | 0.003033 | 0.001032 | **2.9x** |
| **Float64** | 100 | 0.005506 | 0.000785 | **7.0x** |
| **Float64** | 1,000 | 0.034644 | 0.010380 | **3.3x** |
| **Float64** | 10,000 | 0.355334 | 0.073229 | **4.9x** |
| **Float64** | 100,000 | 3.227314 | 1.003895 | **3.2x** |
| **UInt8** | 50 | 0.008404 | 0.001952 | **4.3x** |
| **UInt8** | 100 | 0.011563 | 0.000948 | **12.2x** |
| **UInt8** | 1,000 | 0.032597 | 0.001294 | **25.2x** |
| **UInt8** | 10,000 | 0.369522 | 0.012411 | **29.8x** |
| **UInt8** | 100,000 | 2.731753 | 0.098097 | **27.8x** |
| **Int hit** | 100 | 0.006290 | 0.003364 | **1.9x** |
| **Int miss** | 100 | 0.015014 | 0.006682 | **2.2x** |
| **Int hit** | 100,000 | 3.515771 | 3.039424 | **1.2x** |


mojo/stdlib/test/collections/test_list.mojo
```diff
@@ -885,6 +885,40 @@ def test_list_contains():
...
+
+    # Test with UInt8.
+    var bytes = List[UInt8]()
+    for i in range(256):
+        bytes.append(UInt8(i))
+
+    assert_true(UInt8(0) in bytes)
+    assert_true(UInt8(127) in bytes)
+    assert_true(UInt8(255) in bytes)
+
```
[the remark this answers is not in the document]
msaelices: Done: https://github.com/msaelices/modular/commit/a614847b103b5c995f078c3f6fbe7ccd7b21596b

mojo/stdlib/benchmarks/collections/bench_list_contains.mojo
```diff
@@ -0,0 +1,156 @@
...
+    for i in range(size):
+        l.append(Float64(i) * 1.5)
+    return l^
+
+
+fn make_list_uint8[size: Int]() -> List[UInt8]:
+    var l = List[UInt8](capacity=size)
+    for i in range(size):
+        l.append(UInt8(i % 256))
+    return l^
```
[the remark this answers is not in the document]
msaelices: Fixed: https://github.com/msaelices/modular/commit/0686f8eaa54c379ad74c6f97a0e4c7bbd13e4cc3

mojo/stdlib/benchmarks/collections/bench_list_contains.mojo
```diff
@@ -0,0 +1,156 @@
...
+
+# ===-----------------------------------------------------------------------===#
+# Benchmark List contains (UInt8)
+# ===-----------------------------------------------------------------------===#
+@parameter
+fn bench_list_contains_uint8[size: Int](mut b: Bencher) raises:
+    """Search for a UInt8 element present in the middle of the list."""
+    var items = make_list_uint8[size]()
+    var target = UInt8((size // 2) % 256)
+
```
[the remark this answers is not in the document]
msaelices: Fixed: https://github.com/msaelices/modular/commit/0686f8eaa54c379ad74c6f97a0e4c7bbd13e4cc3

mojo/stdlib/std/collections/list.mojo
```diff
@@ -537,6 +537,60 @@ struct List[T: Copyable](
...
+            True if the value is contained in the list, False otherwise.
+
+        Examples:
+
+        ```mojo
+        var x = [1, 2, 3]
+        print("x contains 3" if 3 in x else "x does not contain 3")
+        ```
+        """
+        var count = len(self)
```
NathanSWard: **question**: We already currently have a [SIMD search algorithm](https://github.com/modular/modular/blob/a968ccc177c8d1b2caca8b4f4dd62cecb75cd726/mojo/stdlib/std/collections/string/string_slice.mojo#L2690) for finding a scalar in a span of other scalars `_memchr` and `_memchr_impl` (which are both used inside of `_memmem_impl` and by extension in `StringSlice.find`).

Are you able to simply reuse the current `_memchr` algorithm or is this one measurably better?

NathanSWard: If yours is better, I'd recommend updating `_memchr` to use it instead so we can also get the perf benefit inside of `StringSlice.find` too! 😁

mojo/stdlib/std/collections/list.mojo
```diff
@@ -537,6 +537,60 @@ struct List[T: Copyable](
                 return True
         return False
 
+    fn __contains__[
+        dtype: DType, //
+    ](self: List[Scalar[dtype], ...], value: Scalar[dtype]) -> Bool:
```
NathanSWard: **suggestion**: I'd recommend moving this implementation to `Span.__contains__` and then delegating `List` to call that method. That way we can get this benefit in both cases.

msaelices: Done! I've replaced the SIMD implementation in `List.__contains__` with a simple delegation to `Span.__contains__`, which already has a SIMD-optimized implementation on `main`. The overload now just does `return value in Span(self)`.

mojo/stdlib/std/collections/list.mojo
msaelices: Good question! I looked into reusing `_memchr`/`_memchr_impl` but they serve a different purpose, as they return a **pointer** to the found element (find semantics), which requires `pack_bits` + `count_trailing_zeros` to compute the exact position. `Span.__contains__` only needs a **bool** (existence check), so it can use a simpler `value in simd_vector` check without the bit-position computation.

That said, since `Span.__contains__` already has a SIMD implementation on `main`, I've now simplified this PR to just delegate `List.__contains__` to `Span.__contains__`, eliminating the code duplication entirely. Unifying the `_memchr` and `Span.__contains__` approaches could be a good follow-up.

msaelices: Since `Span.__contains__` already has a multi-width SIMD implementation, the delegation approach addresses this, `List.__contains__` now benefits from whatever optimizations `Span.__contains__` has, and any future improvements to `Span.__contains__` will automatically apply to `List` as well.

mojo/stdlib/std/collections/list.mojo
msaelices: Just realized that, once we start using the Span SIMD-optimized logic, the benchmarks are much worse, being the non-SIMD approach faster in the Int and Float64 cases

I will dig into the Span logic to see why

msaelices: Fixed the issue for Span, so now both are way faster in Int and Float64 cases (64-bit scalars): https://github.com/msaelices/modular/commit/515e098930c6c9d460b83b871caaf77e8bf8a0e2

mojo/stdlib/benchmarks/collections/bench_list_contains.mojo
```diff
@@ -0,0 +1,160 @@
...
+fn bench_list_contains_hit[size: Int](mut b: Bencher) raises:
+    """Search for an element present in the middle of the list."""
+    var items = make_list[size]()
+    var target = size // 2
+
+    @always_inline
+    @parameter
+    fn call_fn() raises:
+        for _ in range(100):
+            var res = target in items
```
NathanSWard: I'm not totally sure we need these benchmarks as this PR is not doing any performance improvements, it's merely delegating the call to `Span.__contains__`.
If we do keep the benchmarks updating them to `Span` feels like a better approach.
However, if you just want to remove this benchmark, I'd be okay with that too :)

mojo/stdlib/benchmarks/collections/bench_list_contains.mojo
```diff
@@ -0,0 +1,160 @@
...
+            keep(res)
+
+    b.iter[call_fn]()
+    keep(Bool(items))
+    keep(target)
```
NathanSWard: A few style suggestion on benchmarks if we do keep them (IMO it'll be easier to just remove them), 

1. Let's use `black_box` in the testing function
2. Prefer the overload of `.iter` that works with unified closures - this allows you to remove the items/target at the end of this function as they were only needed because `@parameter` closures emit a false-positive warning about "unused variables".
3. No need to mark the function `raises` unless it does actually raise an error.


```suggestion
    fn call_fn() unified {read}:
        for _ in range(100):
            var res = black_box(target) in items
            keep(res)

    b.iter(call_fn)
```

NathanSWard (changes requested):

mojo/stdlib/benchmarks/collections/bench_list_contains.mojo
msaelices: Removed: https://github.com/msaelices/modular/commit/d717ab0ea9bedfba66232c9aabca5ccd8a891996

msaelices:
Superseded: upstream's `Span.__contains__` now uses the same SIMD width-stepping approach (comptime `widths` table). Closing as already implemented upstream.

Issue #4144, opened 2025-03-13
[Feature Request] [stdlib] Compile time `VariadicList.__len__()`


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

As title.

### What is your motivation for this change?

#4142. There is a recursive instantiation if one tries to use the variadic `SIMD` constructor for a `StringSlice` constructor (`unsafe_from_utf8`) and it is then run at compile time, because the `SIMD` constructor has a `debug_assert` instead of a `constrained` due to the limitation of `VariadicList.__len__()` being runtime only.

### Any other details?

```mojo
struct VariadicList[type: AnyTrivialRegType](Sized):
    alias _mlir_type = __mlir_type[`!kgen.variadic<`, type, `>`]
    var value: Self._mlir_type

    fn __len__(self) -> Int:
        return __mlir_op.`pop.variadic.size`(self.value)
```
Since the type itself stores the value as a variable, maybe we can go down a route similar to f5f06ed61ad1d00c7bda904b9938e7cd5fab1be3 and change it to parameter space.

JoeLoser:
I agree we should do this. @lattner it may be a good thing for others to pick up on top of your dependent types work.

lattner:
Unfortunately, we can't do this.  Homogenous variadics are (intentionally) not instantiated with the different list  sizes.  This means that in:

```
fn example(*a: Int):
```

that 'a' doesn't have a comptime known list length.  You can use variadic packs to do this though, because the type list has a comptime known length. This is why VariadicPack can have this member:

```
   @staticmethod
   fn __len__() -> Int:
```

martinvuyk:
@lattner Can we somehow bypass that limitation? Because I think there are several types of APIs that would need the ability to fail at compile time if the length doesn't have certain characteristics, or choose different code-paths for operating on the data.

One example is the one I mentioned above:
```mojo
struct SIMD:
    ...
    fn __init__(out self, *elems: Scalar[type]):
        _simd_construction_checks[type, size]()

        # TODO: Make this a compile-time check when possible.
        debug_assert(
            size == len(elems),
            (
                "mismatch in the number of elements in the SIMD variadic"
                " constructor"
            ),
        )

        __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))

        @parameter
        for i in range(size):
[...]

PR #5919, opened 2026-02-08
[Stdlib] Use SIMD-accelerated brace scanning in format string parsing

msaelices:
## Summary

- Generalize `_memchr` in `string_slice.mojo` into a single variadic function that accepts N needle characters via a `VariadicPack`, using SIMD to search for all of them simultaneously in a single pass
- Inner loops use `@parameter for` with `rebind[Scalar[dtype]]` for compile-time unrolling
- Add `_find_next_brace()` helper in `format.mojo` that uses `_memchr` to locate `{`/`}` characters
- Replace byte-by-byte `for` loop in `compile_entries_runtime()` with `while` loop that jumps directly to the next brace
- Add `bench_format.mojo` benchmark and SIMD boundary tests

## Benchmark results (`main` vs this PR)

Median of 5 repetitions, 1000 format calls per iteration:

| Benchmark | main (ms) | PR (ms) | Speedup |
|---|---|---|---|
| `literal[4B x 4]` | 0.46 | 0.39 | **1.2x** |
| `literal[64B x 4]` | 0.99 | 0.55 | **1.8x** |
| `literal[512B x 4]` | 4.46 | 0.67 | **6.7x** |
| `literal[4096B x 4]` | 28.59 | 2.20 | **13.0x** |
| `runtime_short` | 0.59 | 0.24 | **2.5x** |


mojo/stdlib/std/collections/string/format.mojo
```diff
@@ -341,32 +393,35 @@ struct _FormatUtils:
...
         var fmt_len = format.byte_length()
         var total_estimated_entry_byte_width = 0
 
-        for i in range(fmt_len):
-            if skip_next:
-                skip_next = False
-                continue
+        var i = 0
+        while True:
+            var next = _find_next_brace(fmt_ptr, i, fmt_len)
```
NathanSWard, line 377: *question*: This seems like a more general useful facility than just finding braces for this specific context. Something along the lines of `StringSlice.find_one_of(codepoints...)` or something like that.

mojo/stdlib/std/collections/string/format.mojo
```diff
@@ -70,12 +70,64 @@ methods.
...
+from sys import is_compile_time, simd_width_of
 from utils import Variant
 
+# ===-----------------------------------------------------------------------===#
+# SIMD Helpers
+# ===-----------------------------------------------------------------------===#
+
+
+@always_inline
+fn _find_next_brace(
```
NathanSWard: **comment**: This is a nice optimization for sure, however a few things to consider:
1. For format-strings and when we eventually have `f/t-strings` this code will be run at compile time meaning the vectorized approach isn't saving us much as there is no runtime performance benefits and it may end up slowing down compile times.
2. If you do add benchmarks for this, we need to really only be testing for small strings, where I'm not sure how much perf improvement we'll get. Since most format strings relatively small e.g. `"Hello, {}! I am {}".format(...)`. So I'd be interested to see these cases as long format sting aren't really a very common thing 👀

msaelices: 1. I did not know it. I guess we can parametrize the code to differentiate regular Strings vs f-ones, right?
2. Added benchmarks.

mojo/stdlib/std/collections/string/format.mojo
msaelices: Ups, Just read. Actually, I found we can use the `_memchr` function and I did it here: https://github.com/modular/modular/pull/5919/changes/2270a30315f82800bdafada1d1e9be8dbc8c4ec1

But, it is slower than before, so I've created a new `_memchr2` to find any of the two chars passed: https://github.com/msaelices/modular/commit/724458c3f92541529e6e01ffd20eeb848d4c9fa4

It could be generalized to support n-char searches, I agree, but I’m not sure whether that would be within the scope of this PR.

msaelices: I've added a TODO line for generalizing the `_memchr` and `_memchr2` functions: https://github.com/modular/modular/commit/7990e347cf95169979ba8ba9d8f1f4a27ffcd65b

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -2734,6 +2734,78 @@ fn _memchr_impl[
...
+
+
+@always_inline
+fn _memchr2[
+    dtype: DType, //
+](
+    source: Span[mut=False, Scalar[dtype], ...],
+    char1: Scalar[dtype],
+    char2: Scalar[dtype],
+) -> source.UnsafePointerType:
```
martinvuyk: I think this is already generalizable nowadays. It should be just receiving a variadic pack and iterating over each element and `or`ing  a result variable

NathanSWard: +1 to this

msaelices: Sure. Done: https://github.com/modular/modular/pull/5919/changes/4582c8e44c0935dbfa47beb474577f143005eef4

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -2690,35 +2690,51 @@ fn _unsafe_strlen(
...
+    """
     if is_compile_time() or len(source) < simd_width_of[Scalar[dtype]]():
         var ptr = source.unsafe_ptr()
 
         for i in range(len(source)):
-            if ptr[i] == char:
-                return ptr + i
+            for j in range(len(chars)):
+                if ptr[i] == chars[j]:
+                    return ptr + i
```
martinvuyk: IMO you should unroll the loop at compile time here. Similarly in other places

We might leave this more complicated version for another time, up to you:

I'm not 100% sure how well the compiler might optimize this but it's best if you give it a shove in the right direction. Not sure if this optimization is faster or not either, just eyeballing it

```suggestion
            var data = SIMD[dtype, next_power_of_to(len(chars))](chars[0])

            @parameter
            for j in range(1, len(chars)):
                data[j] = chars[j]
            
            if data.eq(ptr[i]).reduce_or():
                return ptr + i
```

msaelices: I think this does not work because `@parameter for` requires compile-time bounds, and `VariadicList` length is not known at compile-time in Mojo

martinvuyk: > I think this does not work because `@parameter for` requires compile-time bounds, and `VariadicList` length is not known at compile-time in Mojo

Yeah that limitation has irked me for a while now (see #4144). It's why I oginally meant for this to use a `VariadicPack` but while I know of a way to do it currently, it will be ugly. We can polish this later on (i.e. just leave the runtime loop for now).

This actually gave me an idea of how we could allow some nicer version of it, I'll go play around and hopefully figure it out

msaelices: I've checked that now the speed-up is even better than before :rocket: 

Will try the `VariadicPack` approach

NathanSWard: If variadic pack is too verbose and since variadic list doesn't have a comptime size, you could use InlineArray as an alternative too.

msaelices: it works like a charm: https://github.com/msaelices/modular/commit/c4a379605fea22587bd7d11861e1efc3842a241d

msaelices: The benchmarks are similar or slightly better, specially in the `runtime_short` bench

martinvuyk: Nice! FYI I opened #5935 to see if we ever get a nicer way to implement this without the rebind workaround. If this were a public function we'd have to comptime assert that the type of each of the `VariadicPack` elements is `Scalar[dtype]`, IMO we can leave it as is since this is private

msaelices: @NathanSWard @martinvuyk Does this PR look good to go?

PR #5923, opened 2026-02-09
stdlib: add explicit write_repr_to for Deque and LinkedList

lamija-i:
### Summary

This PR adds explicit `write_repr_to` implementations for `Deque` and `LinkedList`,
routing `__repr__` through the `Writable` trait instead of relying on reflection.

### Details

- Implements `write_repr_to` for both containers with recursive `Writable` formatting
- Updates `__repr__` to produce debug-style output distinct from user-facing `str`
- Adjusts tests to explicitly validate the separation between `str` and `repr`
- Keeps existing `write_to` / `__str__` behavior unchanged

### Motivation

This aligns `Deque` and `LinkedList` with existing stdlib container semantics
and addresses the reflection-related concerns discussed in issue #5870


NathanSWard:
Hi @lamija-i thank you for the PR!
Unfortunately, both `Deque` and `LinkedList` already have a `write_repr_to` implementation, so those are already crossed off the list. Please though take a stab at any of the other types that don't have the method implementation :)

PR #5927, opened 2026-02-10
[MAX] Hotfix diffusers config to properly load

jglee-sqbits:
# Overview
Update component config discovery to correctly handle repos that contain both `config.json` and `generation_config.json`.

# What's included
- Detect component folders and only load valid targets (`config.json` or `{key}_config.json`) to avoid mis-picking when multiple config-like files exist.

    - For `tokenizer`, `tokenizer_2`, and `scheduler`, load configs explicitly from `tokenizer_config.json` / `scheduler_config.json`.


KCaverly (approved):
!sync

katelyncaldwell:
Merged! Should be in next nightly

PR #5929, opened 2026-02-10
[stdlib][dict]: align Dict string and repr formatting with fmt utilities

lamija-i:
### Context

The `Dict` type already implements `Writable`, `Stringable`, and `Representable`,
but its string and repr formatting was implemented manually and was not aligned
with the standard `format._utils` helpers used across other stdlib collections
(e.g. `Deque`, `List`).

In particular:
- `__str__` and `__repr__` shared logic implicitly
- there was no explicit `write_repr_to` implementation following the standard
  formatting conventions

This PR brings `Dict` in line with the rest of the stdlib collections.

---

### What this PR does

- Adds an explicit `write_repr_to` implementation for `Dict`
- Uses `format._utils` (`fmt`) for structured formatting, matching `Deque`
- Keeps `__str__` and `__repr__` thin wrappers around `write_to` /
  `write_repr_to`
- Does **not** change behavior, semantics, or public API of `Dict`
- Does **not** modify or add tests (existing tests already cover repr behavior)

---

### Testing

- All existing `collections` tests pass
- No test changes were required, as current tests already validate
  `repr(dict)` and string output

addresses the reflection-related concerns discussed in issue #5870


mojo/stdlib/std/collections/dict.mojo
```diff
@@ -1008,6 +1007,29 @@ struct Dict[
             i += 1
         writer.write("}")
 
+       
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
```
NathanSWard: The format we are following for `write_repr_to` includes printing the typename and parameters of the struct. e.g. `Dict[Int, String]({1: "value", ...})`

You can take a look at some of the other collections like `List`, `LinkedList`, `Deque` for examples on how to do this :)

mojo/stdlib/std/collections/dict.mojo
NathanSWard: We should also ideally update the `write_to` method too, so `write_repr_to` can share code.

NathanSWard (changes requested):

lamija-i:
Updated Dict.write_repr_to to follow the FormatStruct pattern used by other stdlib collections (e.g. List).

- __repr__ now includes the type name and type parameters.
- write_to and write_repr_to share a common _write_dict_body implementation to avoid duplication.
- Preserved Python-like __str__ behavior (String keys use repr formatting).
- Updated tests to reflect the new __repr__ format.

All stdlib tests pass locally.

lamija-i:
Resolved merge conflicts with main.
Added explicit write_repr_to implementations for DictKeyError and EmptyDictError to avoid reflection fallback.
Updated tests accordingly.
All stdlib tests pass locally.

NathanSWard (approved):
Thanks! There's a few extraneous imports but I'll just remove those internally!

NathanSWard:
!sync

PR #5937, opened 2026-02-11
[proposal] Fix and clarify a small typo in Division Semantics proposal

### LINKED ISSUE #5924: [proposal] Division Semantics proposal has incorrect floating-point example
### Where is the problem?

https://github.com/modular/modular/blob/8caeea3afc7acd7f275c9de60253a8e55023ae35/mojo/proposals/truediv.md?plain=1#L61-L63

### What can we do better?

The floating-point example for `7.0 / 3.0` is incorrect, as both before and after this proposal is applied, the answer will always be `2.33333...`. It looks like a copy-paste oversight from integer examples.

I assume the intent was to provide a floordiv example with a positive result, i.e.

```mojo
7.0 // 3.0    # = 2.0
```

... but to avoid ambiguity perhaps it would be good to have an example of regular division preceding the two floordiv examples, so the complete example would be:

```mojo
# Float __floordiv__ follows the same semantics, but still returns Self
7.0 / 3.0     # = 2.3333... (regular floating-point division)
7.0 // 3.0    # = 2.0 (floor toward -infinity)
7.0 // -3.0   # = -3.0 (floor toward -infinity)
```

I have a patch ready to go if submitting here is OK, but creating this issue in case the change needs to be made first in an internal Modular repo.

@laszlokindrat FYI :-)

### Anything else?

_No response_

JosephWakeling:
The floating point example given is incorrect: regardless of whether the proposal gets applied or not, 7.0 / 3.0 will always be 2.3333...

This patch tweaks the example for complete clarity of intent, with cases for regular division, floordiv with a positive result, and floordiv with a negative result.

Fixes https://github.com/modular/modular/issues/5924

### DIFF
--- a/mojo/proposals/truediv.md
+++ b/mojo/proposals/truediv.md
@@ -59,8 +59,9 @@ Python's float-returning semantics.
 7 // -3  # = -3 (floor toward -infinity)
 
 # Float __floordiv__ follows the same semantics, but still returns Self
-7.0 / 3.0     # = 2.0
-7.0 // -3.0   # = -3.0 (floor toward -infinity)
+7.0 / 3.0     # = 2.333... (regular floating-point division)
+7.0 // 3.0    # = 2.0      (floor toward -infinity)
+7.0 // -3.0   # = -3.0     (floor toward -infinity)
 ```
 
 ### Behavior Summary


mojo/proposals/truediv.md
```diff
@@ -59,7 +59,8 @@ Python's float-returning semantics.
 7 // -3  # = -3 (floor toward -infinity)
 
 # Float __floordiv__ follows the same semantics, but still returns Self
-7.0 / 3.0     # = 2.0
+7.0 / 3.0     # = 2.3333... (regular floating-point division)
```
soraros: Nit: please remove one of the "3"s in the fractional part to match the rest of the document, and consider aligning the left parentheses.

JosephWakeling:
<details><summary>Updated to address feedback from @soraros</summary>

```diff
diff --git a/mojo/proposals/truediv.md b/mojo/proposals/truediv.md
index 776375c2c4..93454ddc37 100644
--- a/mojo/proposals/truediv.md
+++ b/mojo/proposals/truediv.md
@@ -59,9 +59,9 @@ Python's float-returning semantics.
 7 // -3  # = -3 (floor toward -infinity)
 
 # Float __floordiv__ follows the same semantics, but still returns Self
-7.0 / 3.0     # = 2.3333... (regular floating-point division)
-7.0 // 3.0    # = 2.0 (floor toward -infinity)
-7.0 // -3.0   # = -3.0 (floor toward -infinity)
+7.0 / 3.0     # = 2.333... (regular floating-point division)
+7.0 // 3.0    # = 2.0      (floor toward -infinity)
+7.0 // -3.0   # = -3.0     (floor toward -infinity)
 ```
 
 </details>

laszlokindrat (approved):

laszlokindrat:
Thank you!

laszlokindrat:
!sync

PR #5940, opened 2026-02-12
[stdlib] Replace uses of `constrained[False, ...]()` with `comptime assert "" != "", ...`

martinvuyk:
Replace uses of `constrained[False, ...]()` with `comptime assert "" != "", ...`. This is a workaround until we can have expressions that always evaluate to false at some sooner stage in the compilation pipeline

### DIFF
--- a/max/kernels/src/layout/_coord.mojo
+++ b/max/kernels/src/layout/_coord.mojo
@@ -122,7 +122,7 @@ struct ComptimeInt[val: Int](CoordLike, TrivialRegisterPassable):
 
     @always_inline("nodebug")
     fn tuple(var self) -> Coord[*Self.VariadicType]:
-        constrained[False, "ComptimeInt is not a tuple type"]()
+        comptime assert "" != "", "ComptimeInt is not a tuple type"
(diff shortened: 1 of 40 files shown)


max/kernels/src/layout/_coord.mojo
```diff
@@ -122,7 +122,7 @@ struct ComptimeInt[val: Int](CoordLike, TrivialRegisterPassable):
 
     @always_inline("nodebug")
     fn tuple(var self) -> Coord[*Self.VariadicType]:
-        constrained[False, "ComptimeInt is not a tuple type"]()
+        comptime assert "" != "", "ComptimeInt is not a tuple type"
```
NathanSWard, line 125: I'm not a big fan of this change as this is a temporary hack until we can get `comptime assert False` to work correctly.
I think a better solution would be to open an issue about this so the compiler team can triage and perhaps figure out a better solution and then we can go from there! :)

NathanSWard (changes requested):

martinvuyk:
Putting this as draft until the fix for #5945 is merged

PR #5947, opened 2026-02-13
[Stdlib] Use memmove in List.pop() for trivially movable types

msaelices:
For types with trivial move semantics, use `memmove` to shift trailing elements in `List.pop()` instead of an element-by-element loop. This follows the same pattern already used in `List._realloc()` and `List.extend()`.

`memmove` (not `memcpy`) is required here because the source and destination regions overlap when shifting elements left.

## Benchmarks (median, 5 repetitions)

| Benchmark | main (ms) | optimized (ms) | speedup |
|---|---|---|---|
| `pop_front[100]` | 0.0070 | 0.0046 | **1.52x** |
| `pop_front[1000]` | 0.284 | 0.220 | **1.29x** |
| `pop_front[10000]` | 33.9 | 12.5 | **2.72x** |
| `pop_middle[100]` | 0.0033 | 0.0032 | ~same |
| `pop_middle[1000]` | 0.316 | 0.157 | **2.01x** |
| `pop_middle[10000]` | 14.4 | 5.13 | **2.80x** |

Non-trivially movable types (e.g. `String`) are unaffected and continue using the element-by-element loop.


mojo/stdlib/benchmarks/collections/bench_list.mojo
```diff
@@ -0,0 +1,126 @@
...
+# ===-----------------------------------------------------------------------===#
+@parameter
+fn bench_list_pop_last[size: Int](mut b: Bencher):
+    """Benchmark popping the last element repeatedly."""
+
+    @always_inline
+    fn call_fn() unified {}:
+        var l = make_int_list[size]()
+        for _ in range(size):
+            keep(black_box(l).pop())
```
NathanSWard: Creation of the list should not be in the benchmark function, otherwise your benchmarks will measure how long the list takes to create which is likely significantly more expensive than the pop operation (as it incurs an allocation which takes an un-determinstic amount of time). 

We should probably have a way to pass an argument to the `call_fn` that is not measure in the benchmark output.
For now we can probably just remove these benchmarks since the implementation is so small no need to block on the adding the benchmarks. :)

NathanSWard (changes requested):

mojo/stdlib/benchmarks/collections/bench_list.mojo
msaelices: Done. I force-pushed the changes, so there is no trace of the benchmarks.

msaelices: @NathanSWard Additionally, I addressed the benchmark issue by introducing an `iter_with_setup()` function. You can review it here: https://github.com/modular/modular/pull/5953/

mojo/stdlib/std/collections/list.mojo
```diff
@@ -983,8 +983,18 @@ struct List[T: Copyable](
...
+        @parameter
+        if Self.T.__moveinit__is_trivial:
+            memcpy(
+                dest=self._data + normalized_idx,
+                src=self._data + normalized_idx + 1,
+                count=trailing,
+            )
+        else:
+            for j in range(normalized_idx + 1, self._len):
+                (self._data + j - 1).init_pointee_move_from(self._data + j)
```
NathanSWard, line 997: Upon looking at the implementation of our `memcpy`, I'm not actually sure this is a safe thing we can do. Normally `memcpy` requires the regions of memory **do not** overlap, meaning what you want is actually more of a `memmove`. this usage could spike UB or invalidate the memory as it doesn't account for overlapping regions.

msaelices: Great catch! I actually had the `memmove` implementation here: https://github.com/modular/modular/pull/5949. 

I will bring this implementation here.

msaelices: Done

NathanSWard:
Closing this PR as this was address in the most recently nightly with `uninit_copy_n`! Appreciate the work @msaelices and for driving a lot of this effort forward!

PR #5949, opened 2026-02-15
[Stdlib] Add memory fast paths to List for trivially copyable/movable types

msaelices:
## Summary
- `__init__(Span)`: use `memcpy` instead of element-by-element `append` for trivially copyable types
- `insert()`: use `memmove` to shift trailing elements in bulk instead of swap loop for trivially movable types
- `__getitem__(slice)`: use `memcpy` for contiguous (step=1) slices with trivially copyable types
- Added `memmove` to the `memory` module (alongside existing `memcpy`)
- Added tests covering both trivial and non-trivial code paths for all three methods
- Added `bench_list.mojo` benchmarks

## Benchmark results (main vs this PR)

**`__init__(Span)` — memcpy fast path**
| Size | main | optimized | Speedup |
|---|---|---|---|
| 100 | 1.04ms | 0.20ms | **5.2x** |
| 1,000 | 1.88ms | 0.65ms | **2.9x** |
| 10,000 | 8.25ms | 4.26ms | **1.9x** |
| 100,000 | 55.0ms | 49.3ms | **1.1x** |

**`insert()` front — memmove fast path (Int)**
| Size | main | optimized | Speedup |
|---|---|---|---|
| 1,000 | 3.11ms | 1.02ms | **3.1x** |
| 10,000 | 21.8ms | 6.74ms | **3.2x** |

**`insert()` middle — memmove fast path (Int)**
| Size | main | optimized | Speedup |
|---|---|---|---|
| 1,000 | 1.64ms | 0.87ms | **1.9x** |
| 10,000 | 11.8ms | 5.72ms | **2.1x** |


mojo/stdlib/std/memory/memory.mojo
```diff
@@ -283,6 +283,39 @@ fn memcpy[
         _memcpy_impl(dest.bitcast[Byte](), src.bitcast[Byte](), n)
 
 
+# ===-----------------------------------------------------------------------===#
+# memmove
+# ===-----------------------------------------------------------------------===#
+
+
+@always_inline
+fn memmove[
```
JoeLoser, line 292: FYI @NathanSWard as you filed an issue internally about this the other day

NathanSWard:
Thank you for this work @msaelices !
Internally - which you'll see in the next nightly - we added some generic functions for this `uninit_copy_n`, `uninit_move_n` and `destroy_n`, so we can avoid writing the `@paratmer if trivial` branch everywhere and hopefully make this a bit easier to use across multiple collections.
The nightly change also modifies `List` to use these so it'll end up clashing with this PR.
I'm going to close this PR since the internal changes will subsume this, but appreciate you work.
I'll get your `Deque` changes merged in though :)

PR #5952, opened 2026-02-16
[Stdlib] Add Deque benchmarks and tests for copy, extend, and append

msaelices:
## Summary

Adds benchmarks and tests for `Deque`'s `copy()`, `extend()` and `append()`,
covering both a trivial element type (`Int`) and a non-trivial one (`String`)
across sizes from 100 to 100,000.

## Background

This PR started as an attempt to replace `Deque`'s manual element-by-element
loops with `uninit_copy_n`/`uninit_move_n`, which lower to `memcpy` for
trivial types. The benchmarks added here are what measured that change, and
they showed it was not a clear win: `extend` regressed at large sizes
(100,000 elements went from 116 ms to 147 ms), so the `Deque` changes were
reverted and only the measurement infrastructure remains.

No functional change to `Deque` itself: this PR is benchmarks and tests only.
Keeping them in tree means the next attempt at this optimization starts with
the numbers already available.


NathanSWard:
Hey @msaelices !
We are exposing some new function `uninit_move_n`, `unint_copy_n` and `destroy_n` in the stdlib which we should be using instead of the manually if/else trivial check and direct calls to memcpy/memmove. 
Once the next nightly lands, can you rebase and use these new functions? thank you 😄

msaelices:
Done — rebased from main and replaced all manual `@parameter if`/`memcpy` patterns with `uninit_copy_n` and `uninit_move_n`. The new APIs handle the trivial type optimization internally, so the code is simpler now.

mojo/stdlib/benchmarks/collections/bench_deque.mojo
```diff
@@ -0,0 +1,148 @@
...
+
+
+# ===-----------------------------------------------------------------------===#
+# Benchmark Main
+# ===-----------------------------------------------------------------------===#
+def main():
+    var m = Bench(BenchConfig(num_repetitions=5))
+    comptime sizes = (100, 1_000, 10_000, 100_000)
+
+    @parameter
```
JoeLoser:
```suggestion
    comptime
```
here and elsewhere now 😄

msaelices:
@JoeLoser I'm investigating some regressions happening after using `unit_*_n functions`, specially in the `append` benchs.

msaelices:
> @JoeLoser I'm investigating some regressions happening after using `unit_*_n functions`, specially in the `append` benchs.

Investigation results:

  - `uninit_move_n` (and direct `memcpy`) in `_realloc` caused a ~3.8x regression in append benchmarks, even in opt mode

What was changed here: https://github.com/modular/modular/pull/5952/changes/34a060d6d78be9f9cc1cbf444ee4ba7a7f643838

  - Reverted _realloc to use original init_pointee_move_from loops
  - Kept uninit_copy_n/uninit_move_n optimizations in __copyinit__, extend, and _prepare_for_new_elements (where they show big wins)

PR description updated with opt-mode benchmarks showing up to 5.6x speedup for copy_int and no append regression

Added a comment in case we revisit this in the future and mistakenly try to "optimize" this path again: https://github.com/modular/modular/pull/5952/changes/4d4cab0cd3ab4439d45b355aa814de7dcbaba83d

mojo/stdlib/benchmarks/collections/bench_deque.mojo
```diff
@@ -0,0 +1,148 @@
...
+fn bench_deque_extend_string[size: Int](mut b: Bencher) raises:
+    var lst = List[String]()
+    for i in range(size):
+        lst.append(String("item_") + String(i))
+
+    @always_inline
+    @parameter
+    fn call_fn():
+        var q = Deque[String]()
+        q.extend(black_box(lst).copy())
```
NathanSWard, line 82: This is also benchmarking copy of list which is non trivial. These copies should not be done in the benchmark function. Let me merge in your "iter_with_starup" and we can use that here.

mojo/stdlib/benchmarks/collections/bench_deque.mojo
```diff
@@ -0,0 +1,148 @@
...
+fn bench_deque_copy_int[size: Int](mut b: Bencher) raises:
+    var q = Deque[Int]()
+    for i in range(size):
+        q.append(i)
+
+    @always_inline
+    @parameter
+    fn call_fn():
+        var p = black_box(q).copy()
+        keep(p[size - 1])
```
NathanSWard:
```suggestion
        keep(p)
```
This should be sufficient if you're intention is to prevent the optimizer from removing `p`

PR #5953, opened 2026-02-16
[Stdlib] Add Bencher.iter_with_setup for benchmarking destructive operations

msaelices:
## Summary

- Adds `Bencher.iter_with_setup` (non-raising and raising variants) that runs a setup closure before each iteration, passing the produced state to the timed benchmark closure. Only the benchmark is timed.
- This solves the problem of benchmarking destructive operations (e.g. `List.pop()`) where fresh state is needed per iteration but setup cost (allocations) should not be measured.
- Adds `bench_list.mojo` with `List.pop()` benchmarks (end, front, middle) using the new API.


mojo/stdlib/benchmarks/collections/bench_list.mojo
```diff
@@ -0,0 +1,110 @@
+# ===----------------------------------------------------------------------=== #
```
NathanSWard: Can we remove these `bench_list` benchmarks in this PR to keep this PR minimal?
Ideally this PR should only be adding the `iter_with_setup` functions to the `Bencher` struct and any tests for those functions :)

mojo/stdlib/std/benchmark/bencher.mojo
```diff
@@ -1359,6 +1359,52 @@ struct Bencher(RegisterPassable):
...
+        for _ in range(self.num_iters):
+            var state = setup_fn()
+            var start = time.perf_counter_ns()
+            bench_fn(state)
+            var stop = time.perf_counter_ns()
+            self.elapsed += Int(stop - start)
+
+    fn iter_with_setup[
+        T: Movable & ImplicitlyDestructible,
+        setup_fn: fn () raises capturing [_] -> T,
```
NathanSWard: IMO we don't need to provide an overload that works with raising `setup_fn`. Setup should never raise an error.

mojo/stdlib/std/benchmark/bencher.mojo
```diff
@@ -1359,6 +1359,52 @@ struct Bencher(RegisterPassable):
             var stop = time.perf_counter_ns()
             self.elapsed += Int(stop - start)
 
+    fn iter_with_setup[
```
NathanSWard, line 1362: Are you able to unify the implementations of `iter_with_setup` and the normal `iter`?
Ideally `iter` should call `iter_with_setup` and simply provide an "empty" setup function.
something like:
```
fn iter[IterFn: fn() unified](mut self, f: IterFn):
        """Returns the total elapsed time by running a target closure a
        particular number of times.

        Parameters:
            IterFn: Type of the closure to benchmark.

        Args:
            f: The closure to benchmark.
        """

        @parameter 
        fn empty_setup(): pass
        self.iter_with_setup[empty_setup](f)
```

NathanSWard (changes requested):
This is great thanks! Left some suggestions for how we can improve this, but this will be a really nice addition to the `Bencher` struct :)

mojo/stdlib/benchmarks/collections/bench_list.mojo
msaelices: Done — removed bench_list.mojo.

mojo/stdlib/std/benchmark/bencher.mojo
msaelices: Done — removed the raising overload.

mojo/stdlib/std/benchmark/bencher.mojo
msaelices: I looked into this but they can't be trivially unified, as `iter` takes a runtime `unified` closure while `iter_with_setup` uses compile-time `capturing[_]` parameters, so the signatures are incompatible AFAIK.

Also, `iter` uses a single timing measurement around all iterations (more accurate for fast ops) while `iter_with_setup` necessarily accumulates per-iteration timings to exclude setup. Happy to revisit if there's a pattern I'm missing.

NathanSWard:
!sync

PR #5956, opened 2026-02-17
[Stdlib] explicit Writable repr for Range

nejraberberovic:
Small changes:
range.mojo
Add Writable to _SequentialRange traits.
Implement fn write_repr_to(self, mut writer: Some[Writer]) using FormatStruct(writer, "range").fields(self.start, self.end).
Add import from format._utils import FormatStruct.
test_range_repr.mojo
New focused test file with two assertions: repr(range(2,5)) == "range(2, 5)" and repr(range(3,3)) == "range(3, 3)".
BUILD.bazel
Add minimal mojo_test target test_range_repr.test so Bazel can run the new test directly.

Related issue: #5870

### DIFF
--- a/mojo/stdlib/std/builtin/range.mojo
+++ b/mojo/stdlib/std/builtin/range.mojo
@@ -23,6 +23,7 @@ from sys.intrinsics import unlikely
 from python import PythonObject
 
 from utils._select import _select_register_value as select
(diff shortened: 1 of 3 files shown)


nejraberberovic:
This change is uneccessary.

PR #5957, opened 2026-02-17
[Stdlib] explicit Writable repr for Bool

nejraberberovic:
Add an explicit, non-reflective `write_repr_to` for `Bool` that emits
\"True\" or \"False\" to preserve existing repr() output. Adds a focused
test asserting exact repr output for both cases.

Related issue: #5870


NathanSWard:
Thanks for this PR, however I don't believe this change is needed. `Bool` already has a custom `write_repr_to` function which simply delegates to it's `write_to` function :)

PR #5958, opened 2026-02-17
[Stdlib] explicit Writable repr for NoneType

nejraberberovic:
Add an explicit, non-reflective debug representation for built-in NoneType by implementing Writable.write_repr_to.
Add a focused test asserting repr(None) == "None".
Keep scope minimal: only NoneType + its focused test.
What changed

none.mojo
Add write_repr_to(self, mut writer: Some[Writer]) that writes the literal "None".
No changes to existing write_to() or runtime behavior.
test_none_repr.mojo
Add focused test test_none_repr_simple asserting exact repr output.
Removed/avoided a container-based assertion that required additional Writable constraints.

Related issue: #5870


mojo/stdlib/test/builtin/test_none_repr.mojo
```diff
@@ -0,0 +1,20 @@
+# ===----------------------------------------------------------------------=== #
+# Focused tests for None repr
+# ===----------------------------------------------------------------------=== #
+
+from testing import assert_equal, TestSuite
+
+
+def test_none_repr_simple():
```
NathanSWard: Thank you for adding a test!
However, we can actually just move this test directly into `test_none.mojo` instead of creating its own separate test file :)

NathanSWard (changes requested):

mojo/stdlib/std/builtin/none.mojo
```diff
@@ -72,3 +72,11 @@ struct NoneType(
...
         """
         writer.write("None")
+
+    @always_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        """Explicit non-reflective repr for NoneType.
+
+        Writes the literal `None` to preserve existing repr() behavior.
+        """
+        writer.write("None")
```
NathanSWard:
```suggestion
        writer.write_string("None")
```

If we're writing a string literal, we can just use `writer.write_string` instead, it's only marginally faster compile times, but worth it IMO!

mojo/stdlib/test/builtin/test_none_repr.mojo
nejraberberovic: Moved the test into test_none.mojo.

NathanSWard (approved):
thank you!

NathanSWard:
!sync

PR #5959, opened 2026-02-17
[Kernel] CausalConv1DUpdate op: functional-graph input + GPU enqueue fix

ulmentflam:
Summary:

- Conv state as input: Op now takes conv_state_in and writes updated conv state to an output buffer (copy then kernel). No in-place mutation so it works with the functional graph.
- shape(): Returns (output_shape, updated_conv_state_shape) and takes conv_state_in.
- GPU enqueue fix: Removed bias_stride from enqueue_function so the 22-argument list matches causal_conv1d_update_gpu and graph compilation succeeds.
- Why: Lets Mamba (and other pipelines) use the custom kernel via F.custom("causal_conv1d_update", ...) and unblocks the model-side PR.


BradLarson:
!sync

BradLarson:
!sync

max/kernels/src/state_space/causal_conv1d_ops.mojo
BradLarson, line 478: In the CPU path here, `CS_IN` appears to be never copied to `CS`, where it is copied on the GPU path. Is CS then uninitialized? Maybe something like this is needed before the kernel call:
```
memcpy(CS.ptr, CS_IN.ptr, total_state_elements)
```

BradLarson: `bias_stride` might no longer be used, if I'm reading things right, so you might be able to remove this.

BradLarson:
!sync

PR #5960, opened 2026-02-17
[Stdlib] Fix LinkedList.reverse() not updating prev pointers

msaelices:
## Summary

`reverse()` only swapped `next` pointers, leaving `prev` pointers stale. This caused `__reversed__()` to produce wrong results after a `reverse()`.


JoeLoser (approved):

JoeLoser:
!sync

PR #5962, opened 2026-02-18
[Stdlib] Fix FileDescriptor.write_bytes() silently dropping data on partial writes

msaelices:
## Summary

Loop until all bytes are written, matching `FileHandle.write_bytes()` behavior. Previously a single `write()` call with only a `debug_assert` meant partial writes (pipes, sockets, large writes) silently lost data in release builds.


lsh:
Thanks, this seems like an improvement! Can you please add a test case where you maybe write more bytes than `PIPE_BUF ` to a pipe with `O_NONBLOCK` fd?

msaelices:
> Thanks, this seems like an improvement! Can you please add a test case where you maybe write more bytes than `PIPE_BUF ` to a pipe with `O_NONBLOCK` fd?

https://github.com/modular/modular/pull/5962/changes/f9a9d9493e2b6add2b64f97a96cd9ff2e47fd1dc

PR #5963, opened 2026-02-18
[Stdlib] Replace O(n^2) pop() loop in Set.clear() with Dict.clear()

msaelices:
## Summary

`Set.clear()` called `pop()` in a loop, which restarts iteration from the beginning on each call, resulting in O(n^2) behavior. Delegate to the underlying `Dict.clear()` which is O(n).


JoeLoser (approved):
Nice cleanup, thank you!

JoeLoser:
!sync

PR #5968, opened 2026-02-18
[Stdlib][Error]: add explicit write_repr_to implementation for Error

lamija-i:
Adds an explicit write_repr_to implementation for Error to remove reliance on the default reflection-based Writable fallback.

Preserves existing repr(Error(...)) behavior (Error('message')) to maintain backward compatibility.

All stdlib tests pass locally.

Part of issue: #5870


NathanSWard (approved):
Thanks!
Internally I'm going to add some unit tests for this, but it looks great, thanks!

NathanSWard:
!sync

PR #5969, opened 2026-02-18
[Stdlib][FloatLiteral]: implement Writable for FloatLiteral and add tests

lamija-i:
Background

As part of the ongoing migration from Stringable/Representable to the unified Writable trait, stdlib types must provide explicit write_to() and write_repr_to() implementations.

While Writable provides a default reflection-based write_repr_to(), this is not suitable for stdlib types, which require consistent and controlled formatting.

Changes

- Added Writable conformance to FloatLiteral
- Implemented explicit write_to()
- Implemented explicit write_repr_to()
- Preserved existing formatting semantics by delegating to String(self)
- Added unit tests to validate Writable behavior

Rationale

FloatLiteral previously relied only on __str__ and did not provide an explicit Writable implementation.

The new implementation:
- Ensures compatibility with the Writable migration
- Avoids reflection-based default formatting
- Keeps output consistent with Float64 formatting
- Adds explicit test coverage for string and repr behavior

Tests

All builtin tests pass locally:
bazel test builtin:all

Part of issue: #5870

### DIFF
--- a/mojo/stdlib/std/builtin/float_literal.mojo
+++ b/mojo/stdlib/std/builtin/float_literal.mojo
@@ -28,6 +28,7 @@ struct FloatLiteral[value: __mlir_type.`!pop.float_literal`](
     Intable,
     Stringable,
     TrivialRegisterPassable,
+    Writable,
 ):
     """Mojo floating point literal type.
 
@@ -168,6 +169,16 @@ struct FloatLiteral[value: __mlir_type.`!pop.float_literal`](
             The Float value.
         """
         return Float64(self)
+    
+    @no_inline
+    fn write_to(self, mut writer: Some[Writer]):
+        """Writes the FloatLiteral in string form."""
+        writer.write(String(self))
+
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        """Writes the FloatLiteral in repr form."""
+        writer.write(String(self))
 
     # ===------------------------------------------------------------------===#
     # Unary Operators
--- a/mojo/stdlib/std/builtin/float_literal.mojo
+++ b/mojo/stdlib/std/builtin/float_literal.mojo
@@ -172,13 +172,21 @@ struct FloatLiteral[value: __mlir_type.`!pop.float_literal`](
     
     @no_inline
     fn write_to(self, mut writer: Some[Writer]):
-        """Writes the FloatLiteral in string form."""
-        writer.write(String(self))
+        """Writes the FloatLiteral in string form.
+
+        Args:
+            writer: The Writer to write the value to.
+        """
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/builtin/float_literal.mojo
```diff
@@ -168,6 +169,16 @@ struct FloatLiteral[value: __mlir_type.`!pop.float_literal`](
...
+    
+    @no_inline
+    fn write_to(self, mut writer: Some[Writer]):
+        """Writes the FloatLiteral in string form."""
+        writer.write(String(self))
+
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        """Writes the FloatLiteral in repr form."""
+        writer.write(String(self))
```
NathanSWard: This current implementation will create a temporary string which leads to an unnecessary allocation.
Instead for these can we just use the `write_to` and `write_repr_to` methods from Float?

```suggestion
    @no_inline
    fn write_to(self, mut writer: Some[Writer]):
        """Writes the FloatLiteral in string form."""
        Float64(self).write_to(writer)

    @no_inline
    fn write_repr_to(self, mut writer: Some[Writer]):
        """Writes the FloatLiteral in repr form."""
        Float64(self).write_repr_to(writer)
```

mojo/stdlib/std/builtin/float_literal.mojo
```diff
@@ -168,6 +169,16 @@ struct FloatLiteral[value: __mlir_type.`!pop.float_literal`](
...
         return Float64(self)
+    
+    @no_inline
+    fn write_to(self, mut writer: Some[Writer]):
+        """Writes the FloatLiteral in string form."""
+        writer.write(String(self))
+
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        """Writes the FloatLiteral in repr form."""
```
NathanSWard: For doc comments - in order to pass our linter, they need to document the arguments.

```suggestion
        """Writes the FloatLiteral in repr form.
        
        Args:
            writer: The value to write to.
        """
```

NathanSWard (changes requested):

lamija-i:
Updated to delegate to Float64.write_to / write_repr_to and added required doc comments. Thanks!

NathanSWard (approved):
Thanks! In the future, can you also make sure to run the formatter on the code? 
You can see in the [Run pre-commit / lint (pull_request)](https://github.com/modular/modular/actions/runs/22204294579/job/64224610271?pr=5969) - how this failed the formatter. 
No worries for the PR as I'll fix it internally :)

NathanSWard:
!sync

PR #5970, opened 2026-02-18
[Stdlib][complex]: add canonical write_repr_to for ComplexSIMD

lamija-i:
Adds explicit write_repr_to implementation for ComplexSIMD as part of the stdlib Writable migration.

Uses FormatStruct(...).fields(self) to ensure canonical stdlib formatting
and avoids relying on the default reflection-based implementation.

Adds unit test verifying repr output.

All complex and builtin tests pass locally.
Part of issue: #5870

### DIFF
--- a/mojo/stdlib/std/complex/complex.mojo
+++ b/mojo/stdlib/std/complex/complex.mojo
@@ -22,6 +22,7 @@ from complex import ComplexSIMD
 import math
 from math.math import _Expable
 from sys import llvm_intrinsic
+from format._utils import FormatStruct
 
 comptime ComplexScalar = ComplexSIMD[size=1]
 """Represents a scalar complex value."""
@@ -164,6 +165,10 @@ struct ComplexSIMD[dtype: DType, size: Int](
         @parameter
         if Self.size > 1:
             writer.write("]")
+    
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        FormatStruct(writer, "ComplexSIMD").fields(self)
 
     @always_inline
     fn __abs__(self) -> SIMD[Self.dtype, Self.size]:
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/complex/complex.mojo
```diff
@@ -164,6 +165,10 @@ struct ComplexSIMD[dtype: DType, size: Int](
         @parameter
         if Self.size > 1:
             writer.write("]")
+    
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
```
NathanSWard, line 170: Can you add appropriate doc comments for this function please?

mojo/stdlib/test/complex/test_complex.mojo
```diff
@@ -125,5 +126,10 @@ def test_exp():
     assert_almost_equal(c.im, 0)
 
 
+def test_complex_repr():
+    var c = ComplexFloat32(1, 2)
+    assert_equal(repr(c), "ComplexSIMD(1.0 + 2.0i)")
```
NathanSWard: For the `write_repr_to` tests, I have a helper test function in `test_utils/` called `check_write_to`. Can we use that to validate these functions? 

```suggestion
from test_utils import check_write_to

def test_complex_repr():
    var c = ComplexFloat32(1, 2)
    check_write_to(c, expected="ComplexSIMD(1.0 + 2.0i)", is_repr=True)
```

NathanSWard (changes requested):

lamija-i:
Added doc comments to write_repr_to and updated tests to use check_write_to as suggested. Thanks!

NathanSWard (approved):
thanks!

NathanSWard:
!sync

Issue #5726, opened 2026-01-02
[BUG] Segfault when creating mojo extension module with Python 3.14 free-threaded


SeeRich:
### Bug description

I'm interested in using mojo to speed up CPU tasks in conjunction with Python 3.14 no-gil version. However, when I try to import a simple mojo extension, I get the following segmentation fault:

```
Program received signal SIGSEGV, Segmentation fault.
__strlen_asimd () at ../sysdeps/aarch64/multiarch/strlen_asimd.S:195
195     ../sysdeps/aarch64/multiarch/strlen_asimd.S: No such file or directory.
(gdb) bt
#0  __strlen_asimd () at ../sysdeps/aarch64/multiarch/strlen_asimd.S:195
#1  0x0000aaaaaae05c4c in PyUnicode_FromString (u=0xffffffffffffffff <error: Cannot access memory at address 0xffffffffffffffff>) at Objects/unicodeobject.c:2171
#2  0x0000aaaaaaefd6b0 in PyModule_New (name=0xffffffffffffffff <error: Cannot access memory at address 0xffffffffffffffff>) at Objects/moduleobject.c:147
#3  _PyModule_CreateInitialized (module=0x5005ffe02000, module_api_version=<optimized out>) at Objects/moduleobject.c:235
#4  0x0000fffff7731534 in PyInit_mojo_module () from /home/edge/Development/mojo-sandbox/mojo_module.so
#5  0x0000aaaaaafc4b18 in _PyImport_RunModInitFunc (p0=0x0, info=0xffffffffd9d0, p_res=0xffffffffd930) at ./Python/importdl.c:428
#6  0x0000aaaaaafc3130 in import_run_extension (tstate=0xaaaaabd8f170 <_PyRuntime+346736>, p0=0xfffff7731400 <PyInit_mojo_module>, info=0xffffffffd9d0, spec=0x26ba87c0fb0, modules=<optimized out>) at Python/import.c:2054
#7  0x0000aaaaaafc3e5c in _imp_create_dynamic_impl (spec=0x26ba87c0fb0, module=<optimized out>, file=<optimized out>) at Python/import.c:4756
#8  _imp_create_dynamic (module=<optimized out>, args=<optimized out>, nargs=<optimized out>) at Python/clinic/import.c.h:488
[...]

PR #5971, opened 2026-02-18
[Stdlib] Initial support for building a free-threaded python extension.

SeeRich:
Just starting to experiment with Mojo and because it allows me to use GPU + SIMD in a cross platform way, I'm very interested in using it as a high performance python extension module. However, it currently doesn't support the free-threaded implementation of the cpython interperter (officially supported in python 3.14). 

This PR attempts to add that support. I have tested this against python 3.14 (free-threaded and regular) builds.

The current implementation of python module extensions in mojo's stdlib segfaults when trying to run with python 3.14 free-threaded. This is because the PyObject struct in cpython has a different size (it's larger due to tracking per object thread information) than the GIL version. So when the module is loaded, it gets invalid data for the module name and crashes trying to calculate strlen(...) of the module name.

The following mojo + python code demonstrates a simplified use case:

#### **`mojo_module.mojo`**
```mojo
import std.math as math
from std.python import PythonObject
from std.python.bindings import PythonModuleBuilder
from std.os import abort
from std.time import perf_counter_ns
from std.builtin.simd import SIMD

# Build with: `mojo build mojo_module.mojo --emit shared-lib -o mojo_module.so`

@export
def PyInit_mojo_module() abi("C") -> PythonObject:
    try:
        var m = PythonModuleBuilder("mojo_module")
        m.def_function[cpu_spin]("cpu_spin", docstring="Spin CPU for given seconds")
        return m.finalize()
    except e:
        abort(String("error creating Python Mojo module:", e))

def cpu_spin(py_secs: PythonObject) raises -> PythonObject:
    var secs = UInt(py=py_secs)
    var ns = Int(secs * 1_000_000_000)
    # Vector size 16 (e.g., 4 floats at once)
    var v1 = SIMD[DType.float32, 4](1.0)
    var v2 = SIMD[DType.float32, 4](2.0)

    var start_ns = perf_counter_ns()
    var sum = SIMD[DType.float64, 1](0.0)
    while perf_counter_ns() - start_ns < ns:
        sum += (v1 + v2).reduce_add().cast[DType.float64]()  # Vector addition to keep CPU busy
    print(sum)
    return sum
```

#### **`main.py`**
```python
import asyncio
import os
import sys
import sysconfig
import time

# Add PWD to sys.path so we can import mojo_module.so
sys.path.append(os.path.dirname(os.path.abspath(__file__)))
import mojo_module  # type: ignore

CPU_SPIN_SECS = 3


def cpu_spin_worker(secs: int) -> int:
    return mojo_module.cpu_spin(secs)


async def async_main() -> None:
    workers = os.cpu_count() or 1

    print(
        f"Launching {workers} asyncio.to_thread tasks: "
        f"cpu_spin({CPU_SPIN_SECS}) seconds per task"
    )
    start = time.perf_counter()
    results = await asyncio.gather(
        *[asyncio.to_thread(cpu_spin_worker, CPU_SPIN_SECS) for _ in range(workers)]
    )
    elapsed = time.perf_counter() - start
    print(f"Finished {workers} tasks in {elapsed:.3f}s ")
    print(f"Final sum: {sum(results)}")


def main() -> None:
    print(f"Python version: {sys.version}")
    print(f"Debug build: {sysconfig.get_config_var('Py_DEBUG') == 1}")
    print(f"GIL enabled: {sys._is_gil_enabled()}")
    asyncio.run(async_main())


if __name__ == "__main__":
    main()
```

### DIFF
--- a/Mojo/stdlib/std/python/_cpython.mojo
+++ b/Mojo/stdlib/std/python/_cpython.mojo
@@ -552,8 +552,10 @@ struct PyObject(
     """All object types are extensions of this type. This is a type which
     contains the information Python needs to treat a pointer to an object as an
     object. In a normal “release” build, it contains only the object's reference
-    count and a pointer to the corresponding type object. Nothing is actually
-    declared to be a PyObject, but every pointer to a Python object can be cast
+    count and a pointer to the corresponding type object. In the free-threaded
+    build (i.e. Py_GIL_DISABLED), it contains additional fields, see
+    _PyObject_FreeThreaded for more details. Nothing is actually declared to be a
+    PyObject, but every pointer to a Python object can be cast
     to a PyObject.
 
     References:
@@ -592,6 +594,86 @@ struct PyObject(
         """
         fmt.FormatStruct(writer, "PyObject").fields(
             fmt.Named("object_ref_count", self.object_ref_count),
+            fmt.Named("object_type", self.object_type),
+        )
+
+
+@fieldwise_init
+struct _PyMutex(Defaultable, ImplicitlyCopyable):
+    """CPython `PyMutex` from `Include/cpython/lock.h`."""
+
+    # https://github.com/python/cpython/blob/629da5c914b4407e01c1dc06cbcbd8dce825fef3/Include/cpython/pylock.h#L33
+
+    var bits: UInt8
+
+    def __init__(out self):
+        self.bits = 0
+
+
+@fieldwise_init
+struct _PyObject_FreeThreaded(
+    Defaultable,
+    ImplicitlyCopyable,
+    Writable,
+):
+    """CPython free-threaded `_object` layout (`Py_GIL_DISABLED`). Same docs for `PyObject` apply,
+    but with additional fields used by the free-threaded interpreter.
+    """
+
+    # https://github.com/python/cpython/blob/629da5c914b4407e01c1dc06cbcbd8dce825fef3/Include/object.h#L156
+
+    var object_tid: c_size_t
+    var object_flags: UInt16
+    var object_mutex: _PyMutex
+    var object_gc_bits: UInt8
+    var object_ref_local: UInt32
+    var object_ref_shared: Py_ssize_t
+    var object_type: PyTypeObjectPtr
+
+    def __init__(out self):
+        self.object_tid = 0
+        self.object_flags = 0
+        self.object_mutex = {}
+        self.object_gc_bits = 0
+        self.object_ref_local = 0
+        self.object_ref_shared = 0
+        self.object_type = {}
+
+    # ===-------------------------------------------------------------------===#
+    # Methods
+    # ===-------------------------------------------------------------------===#
+
+    def write_to(self, mut writer: Some[Writer]):
+        """Formats to the provided Writer.
+
+        Args:
+            writer: The object to write to.
+        """
+
+        writer.write("_PyObject_FreeThreaded(")
+        writer.write("object_tid=", self.object_tid, ",")
+        writer.write("object_flags=", self.object_flags, ",")
+        writer.write("object_mutex=", self.object_mutex.bits, ",")
+        writer.write("object_gc_bits=", self.object_gc_bits, ",")
+        writer.write("object_ref_local=", self.object_ref_local, ",")
+        writer.write("object_ref_shared=", self.object_ref_shared, ",")
+        writer.write("object_type=", self.object_type)
+        writer.write(")")
+
+    @no_inline
+    def write_repr_to(self, mut writer: Some[Writer]):
+        """Writes the repr of this `_PyObject_FreeThreaded` to a writer.
+
+        Args:
+            writer: The object to write to.
+        """
+        fmt.FormatStruct(writer, "_PyObject_FreeThreaded").fields(
+            fmt.Named("object_tid", self.object_tid),
+            fmt.Named("object_flags", self.object_flags),
+            fmt.Named("object_mutex", self.object_mutex.bits),
+            fmt.Named("object_gc_bits", self.object_gc_bits),
+            fmt.Named("object_ref_local", self.object_ref_local),
+            fmt.Named("object_ref_shared", self.object_ref_shared),
             fmt.Named("object_type", self.object_type),
         )
 
@@ -677,15 +759,13 @@ struct PyModuleDef_Slot:
     var value: OpaquePointer[MutUntrackedOrigin]
 
 
-struct PyModuleDef(Movable, Writable):
-    """The Python module definition structs that holds all of the information
-    needed to create a module.
+struct _PyModuleDef_Body(Movable):
+    """The trailing CPython `PyModuleDef` fields after `m_base`.
 
-    References:
-    - https://docs.python.org/3/c-api/module.html#c.PyModuleDef
+    Shared between the GIL-build `PyModuleDef` and the free-threaded
+    `_PyModuleDef_FreeThreaded` so the two layouts share this functionality after
+    their (differently sized) `PyModuleDef_Base`.
     """
-
-    var base: PyModuleDef_Base
 
     var name: Optional[CStringSlice[ImmStaticOrigin]]
     """Name for the new module."""
@@ -727,7 +807,6 @@ struct PyModuleDef(Movable, Writable):
     or `NULL` if not needed."""
 
     def __init__(out self, name: StaticString):
-        self.base = {}
         self.name = name.as_c_string_slice()
         self.docstring = {}
         # setting `size` to -1 means that the module does not support sub-interpreters
@@ -737,6 +816,25 @@ struct PyModuleDef(Movable, Writable):
         self.traverse_fn = _null_fn_ptr[Self._traverse_fn_type]()
         self.clear_fn = _null_fn_ptr[Self._clear_fn_type]()
         self.free_fn = _null_fn_ptr[Self._free_fn_type]()
+
+
+struct PyModuleDef(Movable, Writable):
+    """The Python module definition structs that holds all of the information
+    needed to create a module.
+
+    References:
+    - https://docs.python.org/3/c-api/module.html#c.PyModuleDef
+    """
+
+    var base: PyModuleDef_Base
+    """The initial `PyModuleDef_Base` segment (GIL-enabled build `PyObject` head)."""
+
+    var body: _PyModuleDef_Body
+    """The trailing CPython `PyModuleDef` fields after `m_base`."""
+
+    def __init__(out self, name: StaticString):
+        self.base = {}
+        self.body = _PyModuleDef_Body(name)
 
     # ===-------------------------------------------------------------------===#
     # Trait implementations
@@ -755,11 +853,11 @@ struct PyModuleDef(Movable, Writable):
 
         writer.write("PyModuleDef(")
         writer.write("base=", self.base, ",")
-        writer.write("name=", self.name, ",")
-        writer.write("docstring=", self.docstring, ",")
-        writer.write("size=", self.size, ",")
-        writer.write("methods=", self.methods, ",")
-        writer.write("slots=", self.slots, ",")
+        writer.write("name=", self.body.name, ",")
+        writer.write("docstring=", self.body.docstring, ",")
+        writer.write("size=", self.body.size, ",")
+        writer.write("methods=", self.body.methods, ",")
+        writer.write("slots=", self.body.slots, ",")
         writer.write("traverse_fn=<unprintable>", ",")
         writer.write("clear_fn=<unprintable>", ",")
         writer.write("free_fn=<unprintable>")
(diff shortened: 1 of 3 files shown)


SeeRich:

SeeRich:
Closes #5726

SeeRich:
Not really sure what to do for the build/test failures. Looks like main branch is currently failing and the failing tests are unreleated to the changes in this PR?

SeeRich:
@ConnorGray, anything I can do to help move this forward?

I did have one question. Is is possible to call a C macro from mojo? For example, it would be nice to call `PyModuleDef_HEAD_INIT` for initializing the module base instead of have to redefine it (e.g. `PyModuleDef_Base` struct)?

SeeRich:
@ConnorGray any chance this gets reviewed?

SeeRich:
@JoeLoser Hi! Is there any iterest in supporting python interop with python 3.14 free-threaded? If not, I will just take down the PR.

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -1478,6 +1602,11 @@ struct CPython(Defaultable, Movable):
         else:
             self.version = PythonVersion(0, 0, 0)
 
+        # Present only when CPython is built with `Py_GIL_DISABLED`.
+        self.is_free_threaded = self.lib.check_symbol(
+            "PyUnstable_Module_SetGIL"
+        )
```
JoeLoser: **Observation (blocking)**: I don't think this detection is correct.

`PyUnstable_Module_SetGIL` was added in CPython 3.13 and is present in **all** 3.13+ builds, not just `Py_GIL_DISABLED` ones (see https://docs.python.org/3/c-api/module.html#c.PyUnstable_Module_SetGIL — "Added in version 3.13", no `Py_GIL_DISABLED` gating). On a stock GIL-enabled CPython 3.13/3.14 this will set `is_free_threaded = True`, you'll allocate a `_PyModuleDef_FreeThreaded` against a CPython that expects the GIL `PyModuleDef_Base` layout, and you'll reproduce the exact segfault this PR aims to fix.

I think we have a few options here:
- Look up a symbol that is genuinely conditional on `Py_GIL_DISABLED` (verify against current CPython source).
- After `Py_Initialize`, call `sys._is_gil_enabled()` (3.13+) once and cache the result.

Please also add a test that asserts this returns `False` against a stock 3.13 GIL build.

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -2816,19 +2954,31 @@ struct CPython(Defaultable, Movable):
...
+            module_def_ptr.init_pointee_move(_PyModuleDef_FreeThreaded(name))
+            var module = self._PyModule_Create2_FreeThreaded(
+                module_def_ptr, module_api_version
+            )
+            if module:
+                if (
+                    self._PyUnstable_Module_SetGIL(module, Py_MOD_GIL_NOT_USED)
+                    == -1
+                ):
+                    return {}
```
JoeLoser, line 3810: **Observation (blocking)**: module leaked on `PyUnstable_Module_SetGIL` failure.

`module` is a strong reference (new reference from `PyModule_Create2`). If `PyUnstable_Module_SetGIL` returns -1 we return null without decref'ing it. I think this should be

```mojo
if module:
    if self._PyUnstable_Module_SetGIL(module, Py_MOD_GIL_NOT_USED) == -1:
        self.Py_DecRef(module)
        return {}
```

(`module_def_ptr` is also never freed in either branch, but that matches the existing pybind11 pattern — leave that as-is.)

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -543,6 +543,63 @@ struct PyObject(
...
+    # https://github.com/python/cpython/blob/323c59a5e348347be2ce2b7ea55fcb30bf68b2d3/Include/cpython/lock.h#L29
+
+    var bits: UInt8
+
+    def __init__(out self):
+        self.bits = 0
+
+
+@fieldwise_init
+struct PyObject_FreeThreaded(
```
JoeLoser: Naming inconsistency across the new types:

| Type | Convention |
|---|---|
| `_PyMutex` | leading underscore |
| `PyObject_FreeThreaded` | no underscore |
| `PyModuleDef_Base_FreeThreaded` | no underscore |
| `_PyModuleDef_FreeThreaded` | leading underscore |
| `_PyModuleDef_Body` | leading underscore |

Suggestion: prefix everything that's an internal layout mirror without a stable Python C API name with `_`. So `_PyObject_FreeThreaded`, `_PyModuleDef_Base_FreeThreaded`, `_PyModuleDef_FreeThreaded`. That signals these aren't user-facing.

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -1134,6 +1239,13 @@ comptime PyModule_Create2 = ExternalFunction[
     # PyObject *PyModule_Create2(PyModuleDef *def, int module_api_version)
     def(_CPointer[PyModuleDef, MutAnyOrigin], c_int) thin -> PyObjectPtr,
 ]
+comptime PyModule_Create2_FreeThreaded = ExternalFunction[
```
JoeLoser, line 1492: This binds the same C symbol (`"PyModule_Create2"`) as the existing `PyModule_Create2` above, just with a different Mojo pointer type for the moduledef arg. Works, but it's surprising — a reader sees "FreeThreaded" and expects a distinct C symbol.

Options:
- Bind once with `OpaquePointer`/`UnsafePointer[NoneType]` for the def arg and cast at the two call sites.
- Keep both and add a one-line comment explaining that this is one C function with two compatible Mojo signatures.

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -1202,6 +1314,14 @@ def _PyType_GetName_dummy(type: PyTypeObjectPtr) -> PyObjectPtr:
     abort("PyType_GetName is not available in this Python version")
 
 
+def _PyUnstable_Module_SetGIL_dummy(
+    module: PyObjectPtr, gil: OpaquePointer[MutAnyOrigin]
+) -> c_int:
+    _ = module
+    _ = gil
+    return 0
```
JoeLoser: Returning `0` silently lies that `PyUnstable_Module_SetGIL` succeeded if a future caller ever forgets to gate on `is_free_threaded`. Prefer `abort("PyUnstable_Module_SetGIL is not available in this Python version")`, matching the pattern of `_PyType_GetName_dummy` just above (lines 1310-1314).

mojo/stdlib/std/python/_cpython.mojo
SeeRich: Hi @JoeLoser sorry it took me a bit to get back to this. Thank you for the review! Any chance this can get in before 1.0.0 release?

For this specific issue, while I do agree that the documentation says that, to me, the code reads differently.

Python 3.13 branch: https://github.com/python/cpython/blob/d804d2817b58c8a637d34dac53114f1a11f50cc5/Include/moduleobject.h#L103

Python main branch: https://github.com/python/cpython/blob/41cc78a7a47fe584e2c2899737fec877d61ae331/Include/moduleobject.h#L92

I'm open to suggestions, but this was the best gate I could find.

mojo/stdlib/std/python/_cpython.mojo
SeeRich: Thanks, done!

mojo/stdlib/std/python/_cpython.mojo
SeeRich: Thanks, done!

mojo/stdlib/std/python/_cpython.mojo
SeeRich: Thanks, I added a comment to clarify!

mojo/stdlib/std/python/_cpython.mojo
SeeRich: Thanks, done!

mojo/stdlib/std/python/_cpython.mojo
SeeRich: Also, I added a test but I was having a hard time querying wether the interperter was compiled with free-threading support from python so I used the build flags.

SeeRich:
@JoeLoser any chance you could review this?

Mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -1796,6 +1970,16 @@ struct CPython(Defaultable, Movable):
         self._PyGILState_Ensure = PyGILState_Ensure.load(self.lib.borrow())
         self._PyGILState_Release = PyGILState_Release.load(self.lib.borrow())
         self._PyGILState_Check = PyGILState_Check.load(self.lib.borrow())
+        # `PyUnstable_Module_SetGIL` only exists in free-threaded builds; load
+        # the real symbol when present, otherwise an aborting stub.
+        if self.lib.check_symbol("PyUnstable_Module_SetGIL"):
```
JoeLoser: **Retracting my earlier blocking comment on this — you were right.**

I checked the CPython source rather than just the docs. `PyUnstable_Module_SetGIL` is declared under `#if !defined(Py_LIMITED_API) && defined(Py_GIL_DISABLED)` (`Include/moduleobject.h:103-105`) *and* defined under `#ifdef Py_GIL_DISABLED` (`Objects/moduleobject.c:449-460`). The symbol genuinely is absent from a stock 3.13/3.14 GIL build, so `check_symbol` discriminates correctly. The "Added in version 3.13" note in the docs misled me.

Going further: my suggested alternative, `sys._is_gil_enabled()`, would have been *less* correct. That reports runtime GIL state — a free-threaded build launched with `PYTHON_GIL=1` still has the 32-byte `PyObject`, so it would select the wrong layout. This is a build property, not a runtime one, and the symbol probe is the right instrument.

One request: please add a short comment here citing the `Objects/moduleobject.c` gating specifically, not just the header. The existing comment points at `moduleobject.h`, and a future reader checking only the docs will land where I did and re-litigate this.

Mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -596,6 +598,86 @@ struct PyObject(
...
+    # https://github.com/python/cpython/blob/629da5c914b4407e01c1dc06cbcbd8dce825fef3/Include/cpython/pylock.h#L33
+
+    var bits: UInt8
+
+    def __init__(out self):
+        self.bits = 0
+
+
+@fieldwise_init
+struct _PyObject_FreeThreaded(
```
JoeLoser, line 614: I verified this layout against the CPython 3.14 headers: `Include/object.h:152-163` and the offsets work out to 8/2/1/1/4/8/8 = 32 bytes.

That said, nothing in the PR checks that: it works today but does depend on Mojo laying out nested structs flag with C-compatible padding, including the new `_PyModuleDef_Body` nesting level that changes `PyModuleDef`'s composition. 

Do you mind adding a few asserts to guard ourselves here?

```mojo
comptime assert size_of[_PyObject_FreeThreaded]() == 32
comptime assert size_of[PyModuleDef]() == size_of[PyModuleDef_Base]() + size_of[_PyModuleDef_Body]()
comptime assert size_of[_PyModuleDef_FreeThreaded]() == size_of[_PyModuleDef_Base_FreeThreaded]() + size_of[_PyModuleDef_Body]()
comptime assert size_of[_PyModuleDef_Base_FreeThreaded]() == size_of[PyModuleDef_Base]() + 16
```

Given that the free-threaded path is otherwise untested in CI right now today, these asserts are the only regression protection this change would have.

Mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -596,6 +598,86 @@ struct PyObject(
...
+    # ===-------------------------------------------------------------------===#
+
+    def write_to(self, mut writer: Some[Writer]):
+        """Formats to the provided Writer.
+
+        Args:
+            writer: The object to write to.
+        """
+
+        writer.write("_PyObject_FreeThreaded(")
```
JoeLoser, line 653: **Suggestion** Can you also make `_PyModuleDef_FreeThreaded` conform to `Writable` so it mirrors `PyModuleDef`?

Mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -677,16 +759,14 @@ struct PyModuleDef_Slot:
     var value: OpaquePointer[MutUntrackedOrigin]
 
 
-struct PyModuleDef(Movable, Writable):
-    """The Python module definition structs that holds all of the information
-    needed to create a module.
+struct _PyModuleDef_Body(Movable):
```
JoeLoser, line 762: The `_PyModuleDef_Body` extraction is a nice way to avoid duplicating nine fields, and I'd suggest we kepe it.

The thing I'd like to see is the size assert noted on `_PyObject_FreeThreaded`: this introduces a nesting level that didn't exist before, and if Mojo ever inserted padding between `base` and `body` the result would be a silent ABI break with the same symptom as the bug being fixed here — a garbage `m_name` pointer.

Mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -3635,6 +3818,34 @@ struct CPython(Defaultable, Movable):
...
+
+        # On a free-threaded build (`Py_GIL_DISABLED`) the embedded `PyObject`
+        # head is larger, so `PyModuleDef` has a different layout. Allocate the
+        # matching struct and declare that the module supports running without
+        # the GIL.
+        if self.is_free_threaded:
+            var module_def_ptr = alloc(
+                Layout[_PyModuleDef_FreeThreaded].single()
+            ).unsafe_leak()
+            module_def_ptr.unsafe_write(_PyModuleDef_FreeThreaded(name))
```
JoeLoser, line 3830: **Blocking: this fixes module creation but leaves the same layout assumption broken one level down, where we'd get corrupted data instead of segfaulting.** 

If I understand correctly, any module that uses `PythonTypeBuilder.bind[T]()` is now memory-unsafe on a free-threaded build:

- `bindings.mojo:156` — `PyMojoObject[T].ob_base: PyObject`, the 16-byte GIL head.
- `bindings.mojo:617` — `basicsize=size_of[PyMojoObject[T]]()`, so `tp_basicsize` is 16 bytes short of what the interpreter needs.
- `python_object.mojo:1576` — `PyType_GenericAlloc` allocates that short size, and CPython writes its 32-byte header into it. Not great.
- `python_object.mojo:1601` — `obj_ptr.bitcast[PyMojoObject[T]]()` then writes `mojo_value` at offset 16, landing on `ob_ref_shared` / `ob_type`.

In short, under-allocation combined with clobbering the type pointer on first write. There's two acceptable resolutions to this problem:

1. Extend the same treatment to `PyMojoObject` — a free-threaded head variant and a matching `basicsize`.
2. Have `PythonTypeBuilder.finalize()` raise on free-threaded builds, so the unsupported part of the API fails loudly instead of corrupting memory.

Either is fine for an "initial support" PR, but the PR body should say which, because right now "initial support" reads as "functions work, types are untested" rather than "types corrupt memory".

Mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -3635,6 +3818,34 @@ struct CPython(Defaultable, Movable):
...
+
+            var module = self._PyModule_Create2_FreeThreaded(
+                module_def_ptr, module_api_version
+            )
+            if module:
+                # `Py_MOD_GIL_NOT_USED` is `(void *)1` in CPython.
+                var gil_not_used = OpaquePointer[MutUntrackedOrigin](
```
JoeLoser, line 3837: **Blocking: this declares every Mojo extension module thread-safe on the author's behalf.**

`Py_MOD_GIL_NOT_USED` suppresses CPython's automatic GIL re-enable for the whole module. But the Mojo binding runtime has unsynchronized shared state today:

- `bindings.mojo:56` — `MOJO_PYTHON_TYPE_OBJECTS` is a plain `Dict` behind `_Global` with no locking.
- `bindings.mojo:115` — it's read on every `lookup_py_type_object[T]()`, which is on the path for constructing any Mojo-wrapped type.

Two threads constructing Mojo-wrapped types concurrently is a straightforward data race. `_get_global` first-init has the same exposure.

I'd suggest making this opt-in — e.g. a `gil` parameter on `PythonModuleBuilder` defaulting to `Py_MOD_GIL_USED`. That still fixes the segfault, since the layout correction is what actually matters here, and the default becomes "correct, but the GIL comes back" instead of "fast, but racy". Authors who have audited their module opt in explicitly.

That also seems like the right conservative default for a PR titled "Initial support" 😄

Mojo/stdlib/test/python/test_python_cpython.mojo
```diff
@@ -579,5 +579,18 @@ def test_pymoduledef_write_repr_to() raises:
     assert_true("size=" in s)
 
 
+def test_is_free_threaded_detection() raises:
+    # Is the CPython instance free-threaded
+    var python = Python()
+    ref cpython = python.cpython()
+    var is_free_threaded = cpython.is_free_threaded
```
JoeLoser, line 586: **Quibble** `is_free_threaded` is assigned and never read — the assertion below uses `cpython.is_free_threaded` directly. Drop the local.

Mojo/stdlib/test/python/test_python_cpython.mojo
```diff
@@ -579,5 +579,18 @@ def test_pymoduledef_write_repr_to() raises:
...
+
+    # Was the interpreter compiled with free-threading support
+    var sys = Python.import_module("sys")
+    var expected_free_threaded = "t" in String(sys.abiflags)
```
JoeLoser, line 590: **Thought** This may be better to fix on the build/Bazel side.

What do you think, @Ahajha?

JoeLoser:
> @JoeLoser any chance you could review this?

Apologies for the long re-review. I left another round of comments, a few of them blocking. Once those are addressed, I'm happy to sync this in and land it. Thanks! Don't forget to add a Mojo changelog entry as well 😄

PR #5977, opened 2026-02-19
[Stdlib] Use len() check + single issubset() in Set.__gt__/__lt__

msaelices:
## Summary

Previously `__gt__`/`__lt__` did two full traversals: `issuperset`/`issubset` + `__ne__`. Now an O(1) `len()` check ensures strictness, followed by a single `issubset()` traversal.


JoeLoser (approved):

JoeLoser:
!sync

PR #5978, opened 2026-02-19
[Stdlib] Use Dict.popitem() in Set.pop() to avoid redundant rehash

msaelices:
## Summary

Previously `Set.pop()` iterated to find the first element, copied it, then called `remove()` which rehashed to find it again. Now delegates to `Dict.popitem()` which removes directly from the slot. Order changes from FIFO to LIFO, matching Python's unordered `set.pop()` semantics.

### DIFF
--- a/mojo/stdlib/std/collections/set.mojo
+++ b/mojo/stdlib/std/collections/set.mojo
@@ -418,9 +418,8 @@ struct Set[T: KeyElement, H: Hasher = default_hasher](
     fn pop(mut self) raises -> Self.T:
         """Remove any one item from the set, and return it.
 
-        As an implementation detail this will remove the first item
-        according to insertion order. This is practically useful
-        for breadth-first search implementations.
+        As an implementation detail this will remove the last item
+        according to insertion order.
 
         Returns:
             The element which was removed from the set.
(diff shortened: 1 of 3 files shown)


mojo/stdlib/std/collections/set.mojo
```diff
@@ -419,22 +419,19 @@ struct Set[T: KeyElement, H: Hasher = default_hasher](
...
 
         Returns:
             The element which was removed from the set.
 
         Raises:
             If the set is empty.
         """
-        if not self:
+        try:
+            return self._data.popitem().key.copy()
```
JoeLoser, line 431:
```suggestion
            return self._data.popitem().key^
```

JoeLoser (approved):
Please rebase to fix the conflict and then I can sync this in. Thanks!

mojo/stdlib/std/collections/set.mojo
msaelices: Tried both `.key^` on the temporary and binding to a variable first, both fail with: "field destroyed out of the middle of a value, preventing the overall value from being destroyed". The compiler doesn't allow partial moves out of `DictEntry` (which has both `key` and `value`), so `.copy()` is needed here. Rebased onto main.

JoeLoser:
!sync

PR #5979, opened 2026-02-19
[StdLib] Add recursive flatten method with optional depth

TristanLabrousse:
_**This PR adds a `flatten()` method for Lists with an optional depth parameter.**_

- If `n` is 0, the list is returned unchanged.
- If `n` is 1, it flattens one level.
- If `n` > 1, it flattens up to `n` levels recursively.
- If `n` is None, the method performs a full recursive flatten.

**Examples:**

[[1, [2, 3]], 4].flatten[1]() -> [1, [2, 3], 4]
[[1, [2, 3]], 4].flatten[2]() -> [1, 2, 3, 4]
[[1, [2, 3]], 4].flatten[None]() -> [1, 2, 3, 4]


TristanLabrousse:

mojo/stdlib/std/collections/list.mojo
```diff
@@ -1462,6 +1462,42 @@ struct List[T: Copyable](
         var length = self._len
         return self.unsafe_ptr() + length
 
+fn flatten[depth: Optional[AnyType] = None](self) -> List:
```
JoeLoser:
```suggestion
fn flatten[depth: Optional[Int] = None](self) -> List:
```
since `AnyType` is a bit loose and we know it's going to be something with well-defined semantics of supporting `== 0` as we use it below.

mojo/stdlib/std/collections/list.mojo
```diff
@@ -1462,6 +1462,42 @@ struct List[T: Copyable](
...
+        A flattened list with elements from nested lists up to the specified depth.
+    """
+    # if depth is 0 no changes will be done to the list, so we can return early
+    if depth is not None and depth == 0:
+        return self
+
+    var result = List()
+
+    for item in self:
+        if item is List:
```
JoeLoser, line 1481: **Question** Does this syntax work as you'd expect? I don't think this is properly supported in Mojo at the moment.

TristanLabrousse: Did I need to first write something like : var list = self?

mojo/stdlib/std/collections/list.mojo
NathanSWard: Hey @TristanLabrousse, thank you for this PR!
My one design-focused question would be, I think this would be better off as an `Iterator` adaptor (like map, filter, etc...) instead of being solely implemented on `List`. This is a generally useful utility and applies to all other collections types too (not only `List`, but also `Deque`, `InlineArray`, etc...).

Would you consider reworking this into an iterator adaptor instead?
That may be a bit more complicated than this approach, but it would be more generically useful.

Let me know if this makes sense and I can help guide you on how to implement such a thing.
I would consider looking at `take_while`, or `repeat` in the `itertools.mojo` file as a start from there.

My idea for how this would look would be as such:
```
struct _FlattenIterator[I: InnerIteratorType]:
    ...

@always_inline
fn flatten[
    origin: ImmutOrigin,
    IterableType: Iterable, //,
](ref[origin] iterable: IterableType) -> _FlattenIterator[...]:
    ...
```

NathanSWard (changes requested):

mojo/stdlib/std/collections/list.mojo
TristanLabrousse: Thank you for your idea, I'll try ti do it. Did I need to do another PR or stay on this one?

NathanSWard: Up to you! You can reword this one, or close this and open new one.

For some background - the best reference material would be Rust's flatten iterator adaptor: https://doc.rust-lang.org/beta/std/iter/struct.Flatten.html

NathanSWard:
Thanks for initially working on this @TristanLabrousse
I'm going to go ahead and close this PR though since its now stale - and we are trying to move any major decision making over to Github issues instead.
Please re-open a Gtihub issues with the feature request and we can continue discussion there, thanks!

PR #5980, opened 2026-02-19
[Stdlib] Use discard() instead of try/except remove() in Set.difference_update()

msaelices:
## Summary

Replaces the try/except per-element pattern with a direct `discard()` call, which silently ignores missing elements without exception overhead.


JoeLoser:
`discard()` itself uses try/except internally. The change is a code quality improvement, but the "avoiding exception overhead" framing in the changelog is not quite right — the exception overhead still exists, it's just encapsulated in `discard()`.  That said, is it worth optimizing `discard()` itself?

PR #5982, opened 2026-02-19
[Stdlib] Use memset/memcpy in _justify() instead of repeated concatenation

msaelices:
## Summary

Previously `_justify()` appended the fill character one at a time in a loop, causing repeated reallocations. Now uses `memset` for padding and `memcpy` for content in a single pre-allocated buffer.


mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -2477,13 +2478,19 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+        var ptr = result.unsafe_ptr_mut(capacity=width)
 
-        while result.byte_length() < width:
-            result += fillchar
+        # Left padding
+        memset(ptr, fill_byte, start)
+        # Content
+        memcpy(ptr + start, self.unsafe_ptr(), len(self))
+        # Right padding
+        var end_padding = width - start - len(self)
```
JoeLoser: **Question** What happens if `end_padding` is negative? Does it make sense to add `debug_assert(end_padding >= 0)` and `debug_assert(start >= 0)`?.

In any case, I'd suggest we audit and add some edge case tests such as empty string, exact width, and ensure we have some tests with UTF-8 for this path.

Maybe something like this in the end

```mojo
var end_padding = width - start - len(self)

debug_assert(start >= 0, "start padding cannot be negative")
debug_assert(end_padding >= 0, "end padding cannot be negative")

memset(ptr, fill_byte, start)
memcpy(ptr + start, self.unsafe_ptr(), len(self))
memset(ptr + start + len(self), fill_byte, end_padding)
```

PR #5984, opened 2026-02-19
[Stdlib] Return self.path directly in Path.__fspath__()

msaelices:
## Summary

Previously used `String(self)` which went through the `Writable` trait, creating an unnecessary intermediate copy. Now returns `self.path` directly.

### DIFF
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -328,6 +328,9 @@ what we publish.
 
 ### Library changes
 
+- `Path.__fspath__()` now returns `self.path` directly instead of creating an
+  unnecessary copy via `String(self)`.
+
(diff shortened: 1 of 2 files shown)


mojo/docs/changelog.md
```diff
@@ -328,6 +328,9 @@ what we publish.
 
 ### Library changes
 
+- `Path.__fspath__()` now returns `self.path` directly instead of creating an
```
JoeLoser: **Suggestion** Same comment as in in https://github.com/modular/modular/pull/5985#discussion_r2833288405 - no need to changelog this detail.

JoeLoser (approved):

JoeLoser:
!sync

PR #5986, opened 2026-02-19
[Stdlib] Optimize `Span.__eq__` with `memcmp` for byte and integer types

msaelices:
## Summary
- Adds a compile-time `memcmp` fast path for `Span[Byte].__eq__()` instead of element-by-element comparison
- Uses `@parameter if _type_is_eq[_T, Byte]()` guard so non-Byte spans retain the existing generic path


mojo/stdlib/std/memory/span.mojo
```diff
@@ -547,6 +548,17 @@ struct Span[
...
+        @parameter
+        if _type_is_eq[_T, Byte]():
+            return (
+                memcmp(
+                    self.unsafe_ptr().bitcast[Byte](),
+                    rhs.unsafe_ptr().bitcast[Byte](),
+                    len(self),
+                )
+                == 0
+            )
```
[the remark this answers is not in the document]
soraros: We can do this for all trivial reg types, right?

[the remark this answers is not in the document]
msaelices: Good point! Extended to all integer scalar types (Int8, UInt16, Int16, UInt32, Int32, UInt64, Int64, Bool) via a `_is_byte_eq_comparable` helper. Float types are excluded due to IEEE 754 semantics (NaN != NaN, +0.0 == -0.0).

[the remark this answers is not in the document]
soraros: I was thinking maybe we could use `conforms_to(type_of(...), TrivialRegisterPassable)`.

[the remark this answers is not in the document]
msaelices: Great suggestion. Done: https://github.com/modular/modular/pull/5986/changes/c6201c049bbc1a54407942eb0b7dcb77d81ab968

mojo/stdlib/std/memory/span.mojo
```diff
@@ -555,6 +555,16 @@ struct Span[
         # same pointer and length, so equal
         if self.unsafe_ptr() == rhs.unsafe_ptr():
             return True
+        # Fast path: use memcmp for TrivialRegisterPassable types.
+        comptime if conforms_to(_T, TrivialRegisterPassable):
+            return (
```
JoeLoser, line 746: `TrivialRegisterPassable` includes float types (`Float32`, `Float64`, `SIMD[DType.floatXX, N]`), which breaks equality semantics due to IEEE 754:

| Expression | Element-by-element | `memcmp` | Correct |
|---|---|---|---|
| `Span([0.0]) == Span([-0.0])` | True (`-0.0 == 0.0`) | **False** (different bits) | True |
| `Span([NaN]) == Span([NaN])` | False (`NaN != NaN`) | **True** (same bits) | False |

You identified this yourself earlier in the thread:

> Float types are excluded due to IEEE 754 semantics (NaN != NaN, +0.0 == -0.0).

The subsequent switch to `conforms_to(_T, TrivialRegisterPassable)` reintroduced the problem. I'd suggest going back to your original approach of restricting to integer/byte types, or using a narrower trait that excludes floats.

mojo/docs/changelog.md
```diff
@@ -205,6 +205,9 @@ what we publish.
 
 ### Library changes
```
JoeLoser: **Quibble**: The changelog says `TrivialRegisterPassable` but the PR title says "for Byte type." These should match once the final scope is decided. If you narrow the guard back to integer types, update this accordingly.

mojo/stdlib/std/memory/span.mojo
soraros: Maybe we could add a `TriviallyEquitable` and add a `memcmp` based default implementation for them (outside of this PR, that is).

msaelices: Thanks for the catch. Fixed: https://github.com/msaelices/modular/commit/6ed6f8274a3a2c4215bbe120eb81ed599dc24bb9

Added some of the edge cases you mentioned. 

TIL: `IEEE 754 semantics (NaN != NaN, +0.0 == -0.0)`

mojo/docs/changelog.md
msaelices: Fixed.

PR #5993, opened 2026-02-20
[Stdlib] Use uninit_copy_n for trivial types in Span.copy_from

msaelices:
## Summary
- Use `uninit_copy_n` (memcpy) for trivial types in `Span.copy_from`, keeping the original single-pass assignment loop for non-trivial types.

## Benchmarks (opt mode, 20 reps, `iter_with_setup`)

| Benchmark | Size | main (ms) | optimized (ms) | Speedup |
|---|---|---|---|---|
| copy_from Int | 1,000 | 0.938 | 0.254 | **3.7x** |
| copy_from Int | 10,000 | 5.568 | 2.075 | **2.7x** |
| copy_from String | all | neutral | neutral | ~1.0x |

Benchmark uses `Bencher.iter_with_setup` from PR #5953 to isolate setup cost from the timed `copy_from` call. The benchmark code is included below since it depends on that PR being merged first.

<details>
<summary>Benchmark code (bench_span_copy_from.mojo)</summary>

```mojo
from memory import Span, UnsafePointer

from benchmark import Bench, BenchConfig, Bencher, BenchId, black_box, keep


@fieldwise_init
struct IntState(Movable, ImplicitlyDestructible):
    var src: List[Int]
    var dst: List[Int]

    fn __moveinit__(out self, deinit take: Self):
        self.src = take.src^
        self.dst = take.dst^


@fieldwise_init
struct StringState(Movable, ImplicitlyDestructible):
    var src: List[String]
    var dst: List[String]

    fn __moveinit__(out self, deinit take: Self):
        self.src = take.src^
        self.dst = take.dst^


@parameter
fn bench_span_copy_from_int[size: Int](mut b: Bencher) raises:
    @parameter
    fn setup() -> IntState:
        var src = List[Int]()
        var dst = List[Int]()
        for i in range(size):
            src.append(i)
            dst.append(0)
        return IntState(src^, dst^)

    @parameter
    fn bench(mut state: IntState):
        var dst_span = Span(state.dst)
        var src_span = Span(black_box(state.src))
        dst_span.copy_from(src_span)
        keep(state.dst[size - 1])

    b.iter_with_setup[IntState, setup, bench]()


@parameter
fn bench_span_copy_from_string[size: Int](mut b: Bencher) raises:
    @parameter
    fn setup() -> StringState:
        var src = List[String]()
        var dst = List[String]()
        for i in range(size):
            src.append(String("item_") + String(i))
            dst.append(String("____") + String(i))
        return StringState(src^, dst^)

    @parameter
    fn bench(mut state: StringState):
        var dst_span = Span(state.dst)
        var src_span = Span(black_box(state.src))
        dst_span.copy_from(src_span)
        keep(len(state.dst[size - 1]))

    b.iter_with_setup[StringState, setup, bench]()


def main():
    var m = Bench(BenchConfig(num_repetitions=20))
    comptime sizes = (100, 1_000, 10_000, 100_000)

    comptime
    for i in range(len(sizes)):
        comptime size = sizes[i]
        m.bench_function[bench_span_copy_from_int[size]](
            BenchId(String("bench_span_copy_from_int[", size, "]"))
        )
        m.bench_function[bench_span_copy_from_string[size]](
            BenchId(String("bench_span_copy_from_string[", size, "]"))
        )

    results = Dict[String, Tuple[Float64, Int]]()
    for info in m.info_vec:
        n = info.name
        time = info.result.mean("ms")
        avg, amnt = results.get(n, (Float64(0), 0))
        results[n] = (
            (avg * Float64(amnt) + time) / Float64((amnt + 1)),
            amnt + 1,
        )
    print("")
    for k_v in results.items():
        print(k_v.key, k_v.value[0], sep=",")
```

</details>


mojo/stdlib/std/memory/span.mojo
```diff
@@ -504,8 +504,13 @@ struct Span[
             len(self) == len(other),
             "Spans must be of equal length",
         )
-        for i in range(len(self)):
-            self[i] = other[i].copy()
+        var count = len(self)
+        var dest = self.unsafe_ptr()
+        var src = other.unsafe_ptr()
+        destroy_n(dest, count=count)
+        uninit_copy_n[overlapping=False](
```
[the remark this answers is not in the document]
msaelices: This is outdated. 

Also, the concern about overlapping spans is also a non-issue for our current code: the original `self[i] = other[i].copy()` loop also doesn't handle overlapping spans correctly. `copy_from` was never designed for overlapping spans.

mojo/stdlib/std/memory/span.mojo
```diff
@@ -504,8 +504,19 @@ struct Span[
...
+        # passes over memory with worse cache locality.
+        comptime if _T.__copyinit__is_trivial:
+            uninit_copy_n[overlapping=False](
+                dest=self.unsafe_ptr(),
+                src=other.unsafe_ptr(),
+                count=len(self),
+            )
+        else:
+            for i in range(len(self)):
+                self[i] = other[i].copy()
```
NathanSWard, line 519: Nice! we could probably generalize this with a `copy_n` algorithm or something that assumes _both_ memory regions are initialized. This is good for now though so I'll get it merge in :)

NathanSWard (approved):

NathanSWard:
!sync

mojo/stdlib/std/memory/span.mojo
ConnorGray: **Question** Are we certain this is correct in the case of a type that is trivial to copy, but not trivial to destroy?

I'll admit such a type is certainly rare, perhaps even pathological; but Mojo affords us the ability to model such a type, so we should try to handle it correctly if it comes up. I think all we need to do it modify the check:

```diff
-        comptime if _T.__copyinit__is_trivial:
+        comptime if _T.__copyinit__is_trivial and _T.__del__is_trivial:
```

msaelices: Sure, but the PR is already merged. Do you want me to open a new one?

NathanSWard: I'll fix this internally.

PR #5999, opened 2026-02-21
[Stdlib] SIMD byte-range check for is_ascii_digit and is_ascii_printable

msaelices:
Add `_all_bytes_in_ascii_range[lo, hi]` that uses the unsigned-subtraction trick—`(byte - lo) > (hi - lo)`—to collapse the two-sided range check into one comparison, then vectorises it with SIMD for strings longer than one register width.

`StringSlice.is_ascii_digit` and `is_ascii_printable` now delegate to this helper instead of iterating codepoints.

Benchmarks on UN charter data (5-language average, ms/call):

| Size | `is_ascii_printable` scalar | SIMD | Speedup |
|------|---------------------------|------|---------|
| 100 | 0.150 | 0.0146 | 10.3x |
| 1000 | 0.0181 | 0.00163 | 11.1x |
| 10000 | 0.00194 | 0.000220 | 8.8x |
| 100000 | 0.000244 | 0.0000244 | 10.0x |
| 1M | 0.0000163 | 0.0000032 | 5.1x |

`is_ascii_digit` shows smaller gains on text corpora (which are mostly non-digits, so both paths exit early), but benefits on all-digit inputs.


mojo/stdlib/std/collections/string/string_span.mojo
```diff
@@ -2680,6 +2678,49 @@ fn _unsafe_strlen(
     return offset
 
 
+@always_inline
+fn _all_bytes_in_ascii_range[lo: Byte, hi: Byte](
```
NathanSWard, line 2631: Can you please add a `comptime assert` in this function to ensure that `lo < hi`?
That way this function can ensure the `[lo, hi]` byte range _is_ a valid byte range?

PR #6002, opened 2026-02-22
[Stdlib] Add Stringable and Writable to os.ProcessStatus

msaelices:
Adds `Stringable` and `Writable` trait implementations to `ProcessStatus` in `os/process.mojo`, enabling `print(status)` and `String(status)` to work.

The string output is:
- `ProcessStatus(exit_code: 0)` when the process exited normally
- `ProcessStatus(term_signal: 15)` when terminated by a signal
- `ProcessStatus(running)` when still running

Tests added in `test/os/test_process.mojo` covering all three cases.

### DIFF
--- a/mojo/stdlib/std/builtin/builtin_slice.mojo
+++ b/mojo/stdlib/std/builtin/builtin_slice.mojo
@@ -205,7 +205,7 @@ struct Slice(
         return (start.value(), end.value(), step)
 
 
-struct StridedSlice(ImplicitlyCopyable):
+struct StridedSlice(ImplicitlyCopyable, Stringable, Writable):
     """Represents a slice expression that has a stride.
 
     This type is used to support different behavior for strided vs unstrided
@@ -241,6 +241,28 @@ struct StridedSlice(ImplicitlyCopyable):
 
         self._inner = Slice(start, end, stride)
 
+    # ===-------------------------------------------------------------------===#
+    # Trait implementations
+    # ===-------------------------------------------------------------------===#
+
+    @no_inline
+    fn __str__(self) -> String:
+        """Gets the string representation of the strided slice.
+
+        Returns:
+            The string representation of the strided slice.
+        """
+        return String.write(self)
+
+    @no_inline
+    fn write_to(self, mut writer: Some[Writer]):
+        """Write StridedSlice string representation to a `Writer`.
+
+        Args:
+            writer: The object to write to.
+        """
+        writer.write(self._inner)
+
     fn indices(self, length: Int) -> Tuple[Int, Int, Int]:
         """Returns a tuple of 3 integers representing start, end, and step
         of the slice if applied to a container of given length.
@@ -254,7 +276,7 @@ struct StridedSlice(ImplicitlyCopyable):
         return self._inner.indices(length)
 
 
-struct ContiguousSlice(ImplicitlyCopyable):
+struct ContiguousSlice(ImplicitlyCopyable, Stringable, Writable):
     """Represents a slice expression without a stride.
 
     This type is used to support different behavior for strided vs unstrided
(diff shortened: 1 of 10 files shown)


mojo/stdlib/std/builtin/builtin_slice.mojo
```diff
@@ -205,7 +205,7 @@ struct Slice(
         return (start.value(), end.value(), step)
 
 
-struct StridedSlice(ImplicitlyCopyable):
+struct StridedSlice(ImplicitlyCopyable, Stringable, Writable):
```
NathanSWard, line 208:
```suggestion
struct StridedSlice(ImplicitlyCopyable, Writable):
```

`Stringable` is soon going to be deprecated, so we are only going to support `Writable`.
Can you please remove the `Stringable` and `__str__` method?

Ditto with all of the other types in the PR.

mojo/stdlib/std/builtin/builtin_slice.mojo
```diff
@@ -285,6 +307,40 @@ struct ContiguousSlice(ImplicitlyCopyable):
...
+            if opt:
+                writer.write(repr(opt.value()))
+            else:
+                writer.write(repr(None))
+
+        writer.write("slice(")
+        write_optional(self.start)
+        writer.write(", ")
+        write_optional(self.end)
+        writer.write(", None)")
```
NathanSWard, line 342: Please take a look at how the other types in the stdlib are implement the `write_to` and `write_repr_to` methods. We are using the `format._utils` helper types so we can ensure consistent formatting across all of our types.
E.g. this should be something like:
```
FormatStruct(writer, "ContiguousSlice").fields(self.start, self.end)
```

mojo/stdlib/std/gpu/memory/memory.mojo
```diff
@@ -330,13 +366,23 @@ struct Fill(Equatable, TrivialRegisterPassable):
...
+            writer: The writer to output the fill pattern to.
+        """
         if self == Self.NONE:
-            return "none"
-        if self == Self.ZERO:
-            return "zero"
-        if self == Self.NAN:
-            return "nan"
-        return "unknown fill"
+            writer.write("none")
```
NathanSWard, line 379: When you are writing a string literal such as `"none"` in this case, ideally suing `writer.write_string("none")` is a bit better as it avoid the compile time overhead of the variadic pack parameters of `writer.write`.
```suggestion
            writer.write_string("none")
```

mojo/stdlib/std/os/process.mojo
```diff
@@ -87,6 +87,56 @@ struct ProcessStatus(Copyable, ImplicitlyCopyable, Movable):
...
+            writer.write("ProcessStatus(running)")
+
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        """Formats the repr of this `ProcessStatus` to the provided Writer.
+
+        Args:
+            writer: The object to write to.
+        """
+        writer.write("ProcessStatus(")
```
NathanSWard: Please use the `format._utils` helper structs/methods so we can have consistent formatting across all stdlib types.

mojo/stdlib/std/sys/intrinsics.mojo
```diff
@@ -314,9 +357,48 @@ struct PrefetchRW(TrivialRegisterPassable):
...
+            writer.write("WRITE")
+
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        """Writes the representation of the prefetch read-write option to a writer.
+
+        Args:
+            writer: The writer to write to.
+        """
+        self.write_to(writer)
```
NathanSWard, line 379: the `repr` methods should be used to provide maximal information with the intention of helping a user debug their code.
We generally always want to include the type-name of the struct in the `write_repr_to`.

```suggestion
        FormatStruct(writer, "PrefetchRW").fields(self)
```

mojo/stdlib/test/os/test_process.mojo
```diff
@@ -81,9 +81,55 @@ def test_process_run_missing():
         _ = Process.run(missing_executable_file, List[String]())
 
 
+def test_processstatus_str():
+    # Common exit codes.
+    assert_equal(
+        String(ProcessStatus(exit_code=0)), "ProcessStatus(exit_code: 0)"
+    )
```
NathanSWard: When writing tests for the `write_to` and `write_repr_to` methods, please use the `check_write_to` function instead.
Currently `String(..)` and `repr(...)` route though the `Stringable` and `Representable` traits, however these are going to be removed in the near future.

NathanSWard (commented):
Thank you for starting to work on these! 😄 

I left some comments about how the stdlib implements these.
1. Don't implement `Stringable` or `Writable` for types, these traits are going to be completely replaced with `Writable`.
2. Try to use the `format._utils` helper methods/types so we can have consistent formatting across the stdlib.
3. Use the `check_write_to` helper function in test code for testing `write_to` and `write_repr_to` methods

NathanSWard (changes requested):

NathanSWard:
@msaelices - Can you also please update the title of the PR to reflect what you're doing here?
The title only indicates that you're adding support to `os.ProcessStatus` but there are quite a lot of types in this PR that are getting `Writable` implement for them as well.

msaelices:
Closing in favour of a clean branch — was accidentally created on main branch with unrelated commits.

msaelices:
> @msaelices - Can you also please update the title of the PR to reflect what you're doing here? The title only indicates that you're adding support to `os.ProcessStatus` but there are quite a lot of types in this PR that are getting `Writable` implement for them as well.

Sorry, I messed it up. I've created clean branches for each type.

PR #6004, opened 2026-02-22
[Stdlib] Add Stringable and Writable to benchmark.Report

msaelices:
Adds `Stringable` and `Writable` trait implementations to the `Report` struct in `benchmark.mojo`.

This allows using `print(report)` and `String(report)` directly instead of calling `report.as_string()` explicitly.


mojo/stdlib/std/benchmark/benchmark.mojo
```diff
@@ -210,7 +210,7 @@ struct Unit:
 # Report
 # ===-----------------------------------------------------------------------===#
 @fieldwise_init
-struct Report(Copyable, Defaultable):
+struct Report(Copyable, Defaultable, Representable, Stringable, Writable):
```
NathanSWard:
```suggestion
struct Report(Copyable, Defaultable, Writable):
```

Re the other PR, `Stringable` and `Representable` are being deprecated soon, so no need to implement these.

NathanSWard (changes requested):

NathanSWard:
Thank you for this PR 😄 , however we ended up addressing this internally!

PR #6009, opened 2026-02-22
[Stdlib] Add Writable to StridedSlice and ContiguousSlice

msaelices:
Add `Writable` trait to `StridedSlice` and `ContiguousSlice`, consistent with `Slice` which already implements it.

- `StridedSlice.write_to` delegates to the inner `Slice`
- `ContiguousSlice.write_to` formats as `slice(start, end, None)` (no step)
- `write_repr_to` delegates to `write_to` for both (repr == str, matching Python)

Tests use `check_write_to`.

### DIFF
--- a/mojo/stdlib/std/builtin/builtin_slice.mojo
+++ b/mojo/stdlib/std/builtin/builtin_slice.mojo
@@ -205,7 +205,7 @@ struct Slice(
         return (start.value(), end.value(), step)
 
 
-struct StridedSlice(ImplicitlyCopyable, Representable, Stringable, Writable):
+struct StridedSlice(ImplicitlyCopyable, Writable):
     """Represents a slice expression that has a stride.
 
     This type is used to support different behavior for strided vs unstrided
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/builtin/builtin_slice.mojo
```diff
@@ -285,6 +307,40 @@ struct ContiguousSlice(ImplicitlyCopyable):
...
+            if opt:
+                writer.write(repr(opt.value()))
+            else:
+                writer.write_string("None")
+
+        writer.write_string("slice(")
+        write_optional(self.start)
+        writer.write_string(", ")
+        write_optional(self.end)
+        writer.write_string(", None)")
```
NathanSWard, line 333: I'll just fix this up internally, but we don't want to use `repr` in the body of `write_to`. `repr` will allocate a string which defeats the purpose of a no-allocation `write_to` method. 
Also`, Optional implements `Writable`, so I think you can just directly inline the fields instead of routing through the `write_optional` helper function.

NathanSWard:
!sync

Issue #5894, opened 2026-02-02
[Feature Request] [stdlib] Integrate `FormatSpecType` into the `Writable` and `Writer` traits, and update the `"".format()` logic


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### Integrate `FormatSpecType` into the `Writable` trait
I propose we do it thusly:
```mojo
trait FormatSpecType(Defaultable):
    comptime amnt_conversion_flags: Int = 2
    comptime valid_conversion_flags: InlineArray[
        UInt8, Self.amnt_conversion_flags
    ] = ([Byte(ord("s")), Byte(ord("r"))])

    fn __init__(out self, fmt_str: StringSlice) raises:
        """Parses the format spec from a string that should comply with the
        spec.

        Args:
            fmt_str: The StringSlice with the format spec.
        """
        ...

trait Writable:
    comptime spec_type: FormatSpecType = _DefaultSpecType
    comptime spec_default: Self.spec_type = {}

    fn write_to(self, mut writer: Some[Writer], spec: Self.spec_type = Self.spec_default): ...
```
### Integrate `FormatSpecType` into the `Writer` trait
I would propose something like this:
#### Option 1: if we have the language features
```mojo
struct Formatter[W: Writable, origin: ImmutOrigin](Writable):
    var element: Pointer[Self.W, Self.origin]
    var spec: Self.W.spec_type 

    fn __init__(out self, ref [origin]element: Self.W, spec: Self.W.spec_type): ...
    fn write_to(...): ...

comptime _mapper[origin: ImmutOrigin, T: Writable] = Formatter[T, origin]

trait Writer:
    ...
    fn write[*Ts: Writable](
        self, *args: *Variadic.map_types_to_types[Ts.element_types, _mapper[Ts.origin]]
    ): ...
```
#### Option 2: simpler alternative until we have the language features
```mojo
trait Writer:
    ...
    fn write(self, element: Some[Writable], spec: element.spec_type): ...
```

### Update the `"".format()` logic
That logic should take into account that each curly format entry should depend on the given index's (or kwarg's) `FormatSpecType`

### What is your motivation for this change?

Better ergonomics

### Any other details?

I would like to tackle this

CC: @NathanSWard

PR #6012, opened 2026-02-23
[Stdlib] Add format spec support to String.format()

msaelices:
Add `[[fill]align][width]` format specifier support to `String.format()`, matching Python's format mini-language.

Supported syntax: `{:[fill]align[width]}` where fill is any char, align is `<`/`>`/`^`, and width is an integer.

```mojo
"{:>10}".format(42)      # "        42"
"{:<10}".format("hi")   # "hi        "
"{:*^10}".format("hi")  # "****hi****"
"{!r:>10}".format(42)   # "   Int(42)"
```

The implementation uses a two-pass allocation-free approach (`_CountingWriter`) to avoid heap allocation.

### DIFF
--- a/mojo/docs/changelog.md
+++ b/mojo/docs/changelog.md
@@ -357,6 +357,18 @@ what we publish.
 
 ### Library changes
 
+- `String.format()`, `StringSlice.format()`, and format string literals now
+  support format specifiers with the `[[fill]align][width]` mini-language,
+  matching Python's format spec syntax:
+
+  ```mojo
+  "{:>10}".format(42)     # "        42"  (right-align)
+  "{:<10}".format("hi")   # "hi        "  (left-align)
+  "{:^10}".format("hi")   # "    hi    "  (center)
+  "{:*^10}".format("hi")  # "****hi****"  (custom fill char)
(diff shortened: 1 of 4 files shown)


NathanSWard:
Hey @msaelices thanks for taking as stab at this 🙂!
However I would request this be a more formal proposal approach. Adding format specs to the language is a pretty big feature especially since we'd like to support extensible custom format specs. 
Also it may be worth investigating if mojo should have its own spec language. 

Appreciate this PR - but this will take some larger design work and discussion that is outside the scope of the PR.

NathanSWard:
Please take a look at #5881 : https://github.com/modular/modular/issues/5881#issue-3879367002
and #5894 : https://github.com/modular/modular/issues/5894#issue-3885783128 for more discussion points.

msaelices:
> Hey @msaelices thanks for taking as stab at this 🙂! However I would request this be a more formal proposal approach. Adding format specs to the language is a pretty big feature especially since we'd like to support extensible custom format specs. Also it may be worth investigating if mojo should have its own spec language.
> 
> Appreciate this PR - but this will take some larger design work and discussion that is outside the scope of the PR.

No problem! I was not aware of these discussions. Will close it.

PR #6013, opened 2026-02-23
[Flux2] Align prompt embedding prep with attention mask and padded sequence length

pei0033:
## Summary

In diffusers, `FLUX.1` prompt embedding is effectively handled with token padding only (no attention-mask-aware prompt embedding), while FLUX.2 uses attention mask semantics in prompt embedding.
This PR updates MAX FLUX.2 behavior to follow the FLUX.2-style mask-aware path (encode valid tokens + pad back to target prompt length)

## Changes

- Added mask to Flux2ModelInputs.
- Updated prompt-embedding compile signature
    - `_prepare_prompt_embeddings` now compiles with a CPU bool mask input (`[max_seq_len]`) plus selected hidden states.

## Testing

Run below command
```
./bazelw run //max/examples/diffusion:simple_offline_generation --    --model black-forest-labs/FLUX.2-dev   --prompt "A beautiful woman in a red dress walking down a street." --seed 42
```
<table>
  <tr>
      <td align="center">
      max_now
      <br/>max main branch
    </td>
    <td align="center">
      max_modify
      <br/>max current PR
    </td>
    <td align="center">
      diffusers
      <br/>diffusers with same latent
    </td>

  </tr>
</table>
CC. @byungchul-sqzb


max/python/max/pipelines/architectures/flux2/pipeline_flux2.py
```diff
@@ -144,12 +145,19 @@ def prepare_inputs(self, context: PixelContext) -> Flux2ModelInputs:  # type: ig
...
-                self.text_encoder.config.dtype,
-                shape=["seq_len", "hidden_dim"],
-                device=self.text_encoder.devices[0],
+                DType.bool,
+                shape=["max_seq_len"],
+                device=DeviceRef.CPU(),
             )
-            for _ in range(3)
         ]
+        for _ in range(3):
```
k-w-w: Can you add a comment of why "3" is used here?

max/python/max/pipelines/architectures/flux2/pipeline_flux2.py
```diff
@@ -396,14 +432,27 @@ def prepare_prompt_embeddings(
...
-    def _prepare_prompt_embeddings(self, *hidden_states: Tensor) -> Tensor:
-        # [B, L, S, D] -> [B, S, L, D] -> [B, S, L*D]
+    def _prepare_prompt_embeddings(
+        self,
+        mask: Tensor,  # [max_seq_len]
+        *hidden_states: Tensor,
+    ) -> Tensor:
+        """Stack selected hidden states and left-pad to the mask length."""
+        # [B, L, S, D] -> [B, S, L, D], then left-pad to mask length.
+        target_seq_len = mask.shape[0]
```
k-w-w: nit: since only mask.shape[0] is used, can this method just take the max_seq_len as an input?

max/python/max/pipelines/architectures/flux2/pipeline_flux2.py
```diff
@@ -366,14 +377,39 @@ def prepare_prompt_embeddings(
...
+        padded_seq_len = int(token_ids.shape[0])
+        if mask is not None:
+            if mask.shape[0] != padded_seq_len:
+                raise ValueError(
+                    "Prompt mask length must match token length. "
+                    f"Got mask={mask.shape[0]}, tokens={padded_seq_len}."
+                )
+            valid_tokens = token_ids[mask]
+        else:
+            mask = np.ones(padded_seq_len, dtype=np.bool_)
```
KCaverly: If you are always creating this mask here. Can we not make it mandatory in the method arguments, and ensure it is created upstream appropriately. That feels simpler than making it optional, and creating it here, anyways.

max/python/max/pipelines/architectures/flux2/pipeline_flux2.py
pei0033: You’re right. In this function, `mask` is only used to provide the dynamic `max_seq_len` via `mask.shape[0]`.
`_prepare_prompt_embeddings()` is `max_compile`’d and executed as a graph, and I couldn’t find a way to pass a dynamic shape directly as a graph input. So this is currently a workaround to carry `max_seq_len` through an input tensor shape.

If you have a better approach for passing dynamic shape information (without a full tensor input), I’d be happy to switch to it.

max/python/max/pipelines/architectures/flux2/pipeline_flux2.py
pei0033: The `3` came from the number of text-encoder hidden states used for prompt embeddings (layers `[10, 20, 30]`).
To make that explicit and avoid a magic number, I replaced `for _ in range(3):` with iteration over `self.prompt_embedding_hidden_states_layers`.
I also kept those layer indices as a class-level constant because they are architecture-level defaults used in both compile-time signature setup and runtime hidden-state selection.

max/python/max/pipelines/architectures/flux2/pipeline_flux2.py
pei0033: I checked the tokenizer flow, and `pixel_tokenizer.new_context()` always creates and passes an attention mask (`attn_mask`) into `PixelContext`.
Based on that, I made `mask` mandatory in `prepare_prompt_embeddings` and changed `Flux2ModelInputs.mask` to a non-optional type as well.

KCaverly (approved):
!sync

pei0033:
In commit 3dd49724b3, the prepare embedding section has been integrated into the text encoder and merged into a single graph.
Accordingly, I will open a PR to resolve this issue using a different approach.

PR #6017, opened 2026-02-23
[Stdlib] Vectorize Span.fill with SIMD stores for Scalar types

msaelices:
Add a specialized `Span.fill` overload for `Span[Scalar[dtype]]` that uses SIMD broadcast stores instead of a scalar element loop.

The value is splatted into SIMD vectors of decreasing width (256 → 4), storing multiple elements per instruction. Remaining elements are handled by a scalar tail loop. Follows the same `comptime widths` pattern used by `Span.apply()`.


msaelices:
Superseded: upstream now implements `Span.fill` with a SIMD fast path and its own coverage (`test_fill_bytes` / `test_fill_empty` / `test_fill_multibyte`). Closing as already implemented upstream.

PR #6020, opened 2026-02-24
[stdlib] [proposal] Add mmap module

damiankucharski:
Proposal for a new `std/mmap` package providing POSIX memory-mapped I/O through an RAII `MmapRegion` struct and `mmap_file()` convenience function.

Picks up the discussion from #1134 and #3218, addressing the stdlib team's review feedback (numeric bit-flag constants, no stringly typed APIs).

Assisted-by: AI (copy-editing)

## Checklist

      sequence of smaller PRs

### DIFF
--- a/mojo/proposals/mmap-support.md
+++ b/mojo/proposals/mmap-support.md
@@ -1,0 +1,142 @@
+# Add `mmap` Module
+
+**Status**: Proposed.
+
+This picks up the discussion from #1134 and the earlier proposal in #3218,
+incorporating the review feedback from the stdlib team.
+
+## Problem
+
+Mojo has no way to memory-map files or allocate anonymous mapped regions.
+Users who need this today must manually call `external_call["mmap", ...]`,
+check for `MAP_FAILED`, remember to `munmap` on every exit path, and handle
+the platform constant differences between Linux and macOS.
+
+```mojo
+# What users have to write today:
+var ptr = external_call["mmap", UnsafePointer[UInt8, MutExternalOrigin]](
+    UnsafePointer[UInt8, MutExternalOrigin](),
+    length, c_int(0x1), c_int(0x02), c_int(fd), Int64(0),
+)
+```
+
+## Prior Discussion
+
+The original feature request (#1134) and @KCaverly's proposal (#3218) were
+reviewed by the stdlib team. The feedback was:
+
+- Mode constants should be **numeric bit-flags**, not strings — they need to
+  be OR-able to match POSIX semantics.
+- Include `PROT_NONE`.
+- Avoid "stringly typed" APIs.
+
+This proposal addresses all three points. Plain `comptime Int` constants are
+used (matching the `O_CREAT` / `O_TRUNC` pattern in `io/file.mojo`) and raw
+`prot`/`flags` arguments are accepted rather than a mode enum.
+
+@martinvuyk's suggestion of folding mmap into `FileHandle` as
+`open[memory_mapped=True]()` was also considered. A separate module is
+cleaner: mmap supports anonymous mappings (no file at all), the lifetime of
+a mapping is independent of the fd, and the access pattern (`Span` over
+bytes) is fundamentally different from `FileHandle`'s read/write/seek API.
+
+## Proposal
+
+A new `std/mmap` package with an RAII `MmapRegion` struct and a `mmap_file()`
+convenience function.
+
+### `MmapRegion`
+
+A move-only struct that owns a mapped region and unmaps it on destruction,
+following the same pattern as `FileHandle`:
+
+```mojo
+struct MmapRegion(Movable, Sized):
+    var _ptr: UnsafePointer[UInt8, MutExternalOrigin]
+    var _length: Int
+
+    fn __init__(out self, *, length, prot, flags, fd, offset) raises
+    fn __del__(deinit self)           # munmap, errors silenced
+    fn close(mut self) raises         # munmap, errors raised
+    fn as_bytes(ref self) -> Span[UInt8, __origin_of(self)]
+    fn sync(self, flags: Int = MS_SYNC) raises   # msync
+    fn protect(self, prot: Int) raises            # mprotect
+    fn __len__(self) -> Int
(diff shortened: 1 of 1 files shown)


damiankucharski:

mojo/proposals/mmap-support.md
```diff
@@ -0,0 +1,142 @@
...
+A new `std/mmap` package with an RAII `MmapRegion` struct and a `mmap_file()`
+convenience function.
+
+### `MmapRegion`
+
+A move-only struct that owns a mapped region and unmaps it on destruction,
+following the same pattern as `FileHandle`:
+
+```mojo
+struct MmapRegion(Movable, Sized):
```
owenhilyard, line 62: I expect allocators to be one of the main users of this API, and allocators will often do their own size tracking, so I'd like to also expose the underlying `mmap` function wrapper.

mojo/proposals/mmap-support.md
```diff
@@ -0,0 +1,142 @@
...
+
+Stats the file for size, opens it, maps it, closes the fd, and returns the
+region.
+
+### Constants
+
+Standard POSIX constants, with `platform_map` where values differ:
+
+```mojo
+comptime PROT_NONE  = 0x0
```
owenhilyard: This is missing a lot of standard mmap flags, and also a lot of really useful Linux-specific ones like HUGETLB and MAP_LOCKED.

owenhilyard (changes requested):
This will likely run into the issue that the actual defines are technically not portable between OSes, so hardcoding values may result in headaches.

mojo/proposals/mmap-support.md
damiankucharski: Hey, thanks for pointing this out! Honestly when I was writing this proposal I was thinking about my specific high-level use case, but you're right, this should serve wider needs.

In that case, would you think adding thin  `@always_inline` wrappers for `mmap`, `munmap`, `msync`, `mprotect`, and `madvise` to `sys/_libc.mojo` (following the existing pattern of `close()`, `write()`, etc. there) would cover the low-level side? They'd return raw pointers and error codes so allocators can do their own lifetime and metadata management, partial unmapping, etc.
Then `MmapRegion` and `mmap_file()` would just be a convenience layer built on top for the common use-cases.
What do you think?

mojo/proposals/mmap-support.md
damiankucharski: Sure, honestly I skipped the other flags in proposal to not make it much longer but clearly the intent was not obvious from the content.
So yes, all of the flags including the Linux specific ones would need to be added. Regarding the portability between OSes I would follow how it's done in `std/sys/_libc_errno.mojo` with selective `platform_map` for the platform specific ones. I would hardcode the ones that are uniform between platforms, like `PROT_NONE`, akin to how it's done in `std/io/file.mojo`.

mojo/proposals/mmap-support.md
owenhilyard: Raw pointer versions should be there, but I think that also having a linear typed version of `MmapRegion` which requires you to provide the length to unmap while still making it clear it's not a normal pointer would be helpful. `madvise` and `mprotect` should probably be free floating functions regardless, since they have uses even in normal malloced memory.

damiankucharski: I'm not sure I fully understand the intent here. Do you mean a type that makes it clear at the type level that this is not a normal pointer, but still requires explicit management (caller provides the length to unmap, no automatic cleanup)? Could you sketch a concrete example of the API you're imagining and what the use case would be?

owenhilyard:
```mojo
@explicit_destroy
struct MMapRegionLinear:
    var ptr: UnsafePointer[UInt8, MutExternalOrigin]
    
    fn munmap(deinit self, length: UInt):
        ... #munmap impl
        
    # rest of struct as normal

struct MMapRegionDependent[length: UInt]:
    var ptr: UnsafePointer[UInt8, MutExternalOrigin]

    fn __init__(out self, *, prot, flags, fd, offset) raises
        # use Self.length for the length instead
        
    fn free_prefix[free_length: UInt](deinit self) -> MMapRegionDependent[Self.length - free_length]:
        comptime assert free_length < Self.length, "Invalid length"
        ... # munmap only part of the mapping, return a pointer to the new start
        
    fn __del__(deinit self):
        # munmap using Self.length
```

damiankucharski: I think I understand the intent, but I reviewed the rest of the stdlib looking for similar patterns and found no @explicit_destroy used on resource management types and no constructs that support partial deallocation. The common pattern is thin wrappers in sys/_libc.mojo plus a public RAII type.

That said, mmap is probably the only common system resource where partial deallocation is even viable. Yet Rust (memmap2), Go, and Python all handle it the same way - all-or-nothing RAII, partial unmap left to raw syscalls.

I can see why this could be useful, but it would introduce patterns not yet present in the stdlib, so I am not sure this is something I should be driving as a community contribution. I've updated the proposal to mention the flags and _libc.mojo wrappers, and I think the right level of abstraction here is a good question for the stdlib team.

owenhilyard: Linear types working properly with the rest of the stdlib is fairly new, which is why you won't find a lot of uses of them. I think there is a place for slightly less unsafe types like this.

Rust is the only one of those languages which can safely do partial deallocations, and that requires a certain way of doing type level programming that the stdlib and associated projects try to avoid because it makes really messy errors. Mojo has much better support for things like this, which means we get reasonable error messages.

If you don't want to introduce them, I can add them when I need them later.

damiankucharski: Please don't think I am against the idea; I would be happy to implement that as well. I can see how these would be useful, but what I meant is that this would make this proposal quite large, as it would introduce multiple new types and APIs simultaneously. This is why I mentioned that I would appreciate an opinion from the main developers here. Anyhow, I really appreciate your feedback here. I was quite surprised to get some so quickly, and it actually forced me to explore the codebase in more detail, which was quite useful :)

PR #6022, opened 2026-02-24
[Stdlib] Add explicit Writable repr for Codepoint

nejraberberovic:
This change adds a `write_repr_to` method to the `Codepoint` type, providing a clear and consistent string representation for debugging and logging purposes. A corresponding test was also added.

Related Issue: #5870


NathanSWard:
We ended up addressing internally! Ty for the PR :)

PR #6025, opened 2026-02-24
[Stdlib] Add zfill() to String and StringSlice

msaelices:
Add `zfill()` to `String` and `StringSlice`, matching Python's string API. Pads with zeros on the left, inserting zeros after a leading `+` or `-` sign character.

### DIFF
--- a/mojo/stdlib/std/collections/string/string_slice.mojo
+++ b/mojo/stdlib/std/collections/string/string_slice.mojo
@@ -2433,6 +2433,49 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
         """
         return self._justify(width - len(self) >> 1, width, fillchar)
 
+    fn zfill(self, width: Int) -> String:
+        """Pads the string on the left with zeros to fill a field of the given
+        width.
+
+        If the string starts with a sign character (`+` or `-`), the zeros are
+        inserted after the sign. If the string is already at least as wide as
+        `width`, it is returned unchanged.
+
+        Args:
+            width: The total width (in bytes) of the resulting string.
+
+        Returns:
+            A zero-padded string of (byte) length `width`, or the original
+            string if it is already at least as wide.
+
+        Examples:
+
+        ```mojo
+        print(StringSlice("42").zfill(5))   # "00042"
+        print(StringSlice("-42").zfill(5))  # "-0042"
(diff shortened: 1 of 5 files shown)


mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -2434,6 +2434,49 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+
+        Examples:
+
+        ```mojo
+        print(StringSlice("42").zfill(5))   # "00042"
+        print(StringSlice("-42").zfill(5))  # "-0042"
+        print(StringSlice("+42").zfill(5))  # "+0042"
+        ```
+        """
+        var length = len(self)
```
martinvuyk: I'm trying to deprecate the use of `len(str)`
```suggestion
        var length = self.byte_length()
```

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -2434,6 +2434,49 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+            return String(self)
+        var padding = width - length
+        var result = String(capacity=width)
+        if length > 0:
+            var first = self.unsafe_ptr()[]
+            if first == UInt8(ord("+")) or first == UInt8(ord("-")):
+                result += StringSlice(ptr=self.unsafe_ptr(), length=1)
+                for _ in range(padding):
+                    result += "0"
+                result += StringSlice(
```
martinvuyk, line 2470: This is assuming everything is an integer (which is not what Python does), do we want to go in that direction?
```python
>>> print("asd".zfill(3))
asd
>>> print("+asd".zfill(3))
+asd 
```

martinvuyk (commented):
Hi, sorry to be a bummer but I don't quite get why this should be in the stdlib. To me it seems like a very niche use-case, and I personally wouldn't like us to migrate every single Python API as-is just because. Especially given this is easily replaceable by just using format strings e.g. something like this
```python
def zfill(x: str, y: int):
  stripped = x.strip("+-")
  return f"{x[0] if x == stripped else ''}{stripped:0<f'{y}'}"
```

msaelices:
Closed as per @martinvuyk's comment

PR #6032, opened 2026-02-26
[pipelines] use the max_model_length from model config for tokenizers

frostming:
Signed-off-by: Frost Ming <me@frostming.com>

Quote:

> In the MAX pipeline, the tokenizer is limiting model_max_length to 131072 based on the value in tokenizer_config.json, even when a higher --max-length (e.g. 163840) is provided and the model supports a larger context.
> 
> This has been observed with DeepSeekV3 and will also affect DeepSeek-R1, since both have model_max_length set to 131072 in their tokenizer configs:
> 
> DeepSeekV3: https://huggingface.co/kathywu95/deepseek-v3-small-random-fp8/blob/main/tokenizer_config.json#L22
> 
> DeepSeek-R1: https://huggingface.co/deepseek-ai/DeepSeek-R1/blob/main/tokenizer_config.json#L22
> 
> The current behavior causes warnings/errors like:
> 
> Token indices sequence length is longer than the specified maximum sequence length for this model (405315 > 131072). Running this sequence through the model will result in indexing errors
> 
> The MAX pipeline tokenizer should obtain the correct model_max_length from the model/config instead of automatically using the value from tokenizer_config.json.
>

### DIFF
--- a/max/python/max/pipelines/lib/tokenizer.py
+++ b/max/python/max/pipelines/lib/tokenizer.py
@@ -227,6 +227,24 @@ async def run_with_default_executor(
     return await loop.run_in_executor(None, fn, *args, **kwargs)
 
 
+def _resolve_tokenizer_max_length(
+    max_length: int | None, pipeline_config: PipelineConfig | None
+) -> int | None:
+    """Resolve tokenizer max length with pipeline config fallback."""
+    if max_length is not None:
+        return max_length
+
+    if pipeline_config is None:
+        return None
+
+    model_config = getattr(pipeline_config, "model", None)
+    if model_config is None:
+        return None
+
+    model_max_length = getattr(model_config, "max_length", None)
+    return model_max_length if isinstance(model_max_length, int) else None
+
+
 class TextTokenizer(
     PipelineTokenizer[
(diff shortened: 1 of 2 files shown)


max/python/max/pipelines/lib/tokenizer.py
```diff
@@ -256,15 +274,16 @@ def __init__(
         **unused_kwargs,
     ) -> None:
         self.model_path = model_path
+        resolved_max_length = _resolve_tokenizer_max_length(
+            max_length, pipeline_config
```
wrongbad: Why not just `max_length = max_length or pipeline_config.model.max_length`?

`pipeline_config` and `pipeline_config.model` aren't allowed to be `None` with mypy/pydantic so the extra checking could be a little misleading.

wrongbad:
For the title, you could use `[pipelines]` or `[max serve]`. `[stdlib]` could be confused with the mojo stdlib.

max/python/max/pipelines/lib/tokenizer.py
frostming: Updated, sorry for the late reply

AustinDoolittle (approved):
LGTM

PR #6034, opened 2026-02-26
[Stdlib] Add List.remove(value) method

msaelices:
## Summary

- Add `List.remove(value)` which removes the first occurrence of a value, raising if not found. Matches Python's `list.remove()` behavior.


mojo/stdlib/std/collections/list.mojo
```diff
@@ -1440,6 +1440,32 @@ struct List[T: Movable, /](
                 count += 1
         return count
 
+    def remove[
+        _T: Equatable & Copyable & ImplicitlyDestructible, //
+    ](mut self: List[_T], value: _T) raises:
```
soraros, line 1445: We should use the `where` clause.

PR #6035, opened 2026-02-26
[Stdlib] Add List.sort() in-place sort method

msaelices:
## Summary

- Add `List.sort()` method for `Comparable` types (ascending order).
- Add `List.sort[cmp_fn]()` overload for custom comparison functions.
- Both delegate to the existing `sort()` free function.

### DIFF
--- a/mojo/stdlib/std/collections/list.mojo
+++ b/mojo/stdlib/std/collections/list.mojo
@@ -1380,6 +1380,47 @@ struct List[T: Copyable](
                 count += 1
         return count
 
+    fn sort[
+        _T: Copyable & Comparable, //
+    ](mut self: List[_T, ...]):
+        """Sorts the list in-place in ascending order.
+
+        Parameters:
+            _T: The type of the elements in the list. Must implement the
+                traits `Copyable` and `Comparable`.
+
+        Examples:
+
+        ```mojo
+        var numbers = [3, 1, 4, 1, 5]
+        numbers.sort()
+        print(numbers)  # [1, 1, 3, 4, 5]
+        ```
+        """
+        sort(self)
+
+    fn sort[
+        _T: Copyable,
+        //,
+        cmp_fn: fn(_T, _T) capturing[_] -> Bool,
+    ](mut self: List[_T, ...]):
+        """Sorts the list in-place using a custom comparison function.
+
+        Parameters:
+            _T: The type of the elements in the list.
+            cmp_fn: The comparison function that returns True if the first
+                argument should be ordered before the second.
+
+        Examples:
+
+        ```mojo
+        var numbers = [3, 1, 4, 1, 5]
+        numbers.sort[cmp_fn = lambda a, b: a > b]()  # descending
+        print(numbers)  # [5, 4, 3, 1, 1]
+        ```
+        """
+        sort[cmp_fn](self)
+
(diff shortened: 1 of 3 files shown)


mojo/stdlib/std/collections/list.mojo
```diff
@@ -1380,6 +1380,47 @@ struct List[T: Copyable](
                 count += 1
         return count
 
+    fn sort[
+        _T: Copyable & Comparable, //
```
owenhilyard, line 1384: Shouldn't this be `Movable`?

mojo/stdlib/std/collections/list.mojo
```diff
@@ -1380,6 +1380,47 @@ struct List[T: Copyable](
...
+        ```mojo
+        var numbers = [3, 1, 4, 1, 5]
+        numbers.sort()
+        print(numbers)  # [1, 1, 3, 4, 5]
+        ```
+        """
+        sort(self)
+
+    fn sort[
+        _T: Copyable,
```
owenhilyard, line 1403: Same as above, shouldn't this be `Movable`?

mojo/stdlib/std/collections/list.mojo
```diff
@@ -1380,6 +1380,47 @@ struct List[T: Copyable](
...
+        //,
+        cmp_fn: fn(_T, _T) capturing[_] -> Bool,
```
owenhilyard, line 1405: Do we want to make this parametrically raise?

mojo/stdlib/std/collections/list.mojo
NathanSWard: List requires `T: Copyable` so we're stuck with Copyable on this method until we support `T: Movable`.

mojo/stdlib/std/collections/list.mojo
```diff
@@ -1380,6 +1380,47 @@ struct List[T: Copyable](
...
+
+        Examples:
+
+        ```mojo
+        var numbers = [3, 1, 4, 1, 5]
+        numbers.sort()
+        print(numbers)  # [1, 1, 3, 4, 5]
+        ```
+        """
+        sort(self)
```
NathanSWard, line 1400: What's the benefit of this being a member function as apposed to just calling the `sort` free function?
This seems to just be introducing noise IMO.

We already have `sort(list)` available, so I'm not sure the benefits of having `list.sort()` as this just provide 2 ways to write the same thing.

NathanSWard (changes requested):

mojo/stdlib/std/collections/list.mojo
msaelices: The main benefit of a member function like `list.sort()` is ergonomics and discoverability.

From a usability perspective, users naturally look for operations on the type itself. If I have a list, it’s intuitive to check what methods it exposes.

Also, Python developers will try it just by musle memory.

I'm also ok to close this PR if consistency and DRY are higher priorities.

NathanSWard: I'd prefer to close this unless we have a tracking issue where there are multiple people supporting this extra functionally.

NathanSWard:
Closing this PR - as we have the free `sort` method which is usable. 

However, if there is more a desire to have a `.sort` method on `List`, we should open an issue to track that work. Thanks!

PR #6036, opened 2026-02-26
[Stdlib] Add as_span() method to List

msaelices:
## Summary
- Adds `as_span()` method to `List` for explicitly obtaining a `Span` view over the list's elements, matching the pattern used by `HostBuffer.as_span()`.


NathanSWard:
I don't believe this function is necessary. You can already construct a span from a list directly and we generally try to prefer constructors over method casting.
```
var list = [1, 2 , 3]
var span = Span(list)
```

msaelices:
As per @NathanSWard 's comment, close the PR

PR #6037, opened 2026-02-27
[Stdlib] Add fill() method to Array

msaelices:
## Summary

- Adds a `fill(value)` method to `Array` that assigns a value to every
  element, delegating to `Span.fill()`.

`Array` already supports `Array[Int, 5](fill=42)` at construction time, but
there is no way to refill an array you already have without writing the loop
by hand. This closes that gap, and reuses `Span.fill()` so the two paths stay
consistent.

```mojo
var arr: Array[Int, 3] = [1, 2, 3]
arr.fill(0)  # arr is now [0, 0, 0]
```

Note: this PR predates the `InlineArray` -> `Array` rename and has been
migrated to the current API (`Array`, `Deinitable`, and the `where
conforms_to(...)` form), with its tests moved to `test_array.mojo`.


mojo/stdlib/std/collections/inline_array.mojo
```diff
@@ -650,6 +650,27 @@ struct InlineArray[ElementType: Copyable, size: Int](
             .address_space_cast[address_space]()
         )
 
+    fn fill[
```
abduld: what woudl be the different between this and `InlinedArray[...](fill = xyz)` ?

soraros: I suppose it's because it is a mutating method on a pre-existing array. However, I also wonder whether we could simply write `var a = ...; a = {fill = val}` for this use case.

msaelices: It's because code ergonomics. `array.fill(x)` clearly communicates what the programmer intends: populate an existing array with the value `x`. With `InlinedArray(x)`, the meaning could is ambiguous.

PR #6038, opened 2026-02-27
[Kernels] Vectorize GPU concat elementwise loads/stores

prsabahrami:
## Summary

Vectorize GPU concat elementwise loads/stores from scalar (width=1) to SIMD width=4 when the concat axis is not the innermost dimension. In this case, the SIMD group spans the innermost (non-concat) dimension where data is contiguous in both input and output, making vectorized access safe.

Applied to both `_concat_gpu_elementwise` and `_fused_concat_gpu_elementwise`.

**Shape:** [1,8,1024,128] + [1,8,1,128], axis=2, float32

| Metric | Baseline | Optimized | Change |
|--------|----------|-----------|--------|
| Execution Time | 14.13 µs | 6.04 µs | **2.34x faster** |
| NCU Duration | 15,264 ns | 7,104 ns | -53.5% |
| Throughput | 594 GElems/s | 1,391 GElems/s | +134% |
| Executed Instructions | 7.6M | 2.0M | -73.7% |

## Changes

- `max/kernels/src/nn/concat.mojo` — 2 commits, +25/-20 lines

## Test Plan

- [x] Benchmarked with verified speedup via NCU profiling
- [x] `./bazelw test //max/kernels/test/gpu/nn:test_concat.mojo.test` — PASSED

### DIFF
--- a/max/kernels/src/nn/concat.mojo
+++ b/max/kernels/src/nn/concat.mojo
@@ -806,10 +806,12 @@ fn _concat_gpu_elementwise[
                     comptime func = epilogue_fn.value()
                     func[dtype, _rank, simd_width](
                         out_index,
-                        input.load[width=1](in_coord),
+                        input.load[width=simd_width](in_coord),
                     )
                 else:
-                    output.store(out_coord, input.load[width=1](in_coord))
+                    output.store[width=simd_width](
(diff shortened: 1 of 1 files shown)


abduld:
!sync

PR #6039, opened 2026-02-27
[Kernels] Improve GEMV split-K tile width and non-aligned K dispatch

prsabahrami:
## Summary

Two GEMV optimizations:

1. **Increase split-K tile_n from 2 to 4** — Each thread processes 4 output columns instead of 2 in the GEMV_SPLIT_K kernel, improving arithmetic throughput per thread.

2. **Use GEVM kernel for non-WARP_SIZE-aligned K** — When K is not a multiple of WARP_SIZE (32), the dispatch previously fell through to a slow generic path. Now the single GEVM_KERNEL path handles both aligned and non-aligned K by passing `check_k_bounds` at compile time based on `k % WARP_SIZE`. When K is aligned, the bounds check is eliminated entirely via `comptime if`.

| Workload | Baseline | Optimized | Speedup |
|----------|----------|-----------|---------|
| M=1 N=4096 K=4096 transpose | 7.59 µs | 7.43 µs | **1.02x** |
| M=1 N=4096 K=4095 default | 255.3 µs | 39.6 µs | **6.4x** |

The K=4095 case is a pathological edge case in the original code where non-aligned K caused a fallthrough to MATMUL_NAIVE.

## Changes

- `max/kernels/src/linalg/gemv.mojo` — 3 commits

## Test Plan

- [x] Benchmarked baseline vs optimized with verified speedups
- [x] `./bazelw test //max/kernels/test/gpu/linalg:test_gemv.mojo.test` — PASSED
- [x] No regressions across 13 additional shape/mode combinations


max/kernels/src/linalg/gemv.mojo
```diff
@@ -808,6 +812,34 @@ fn gemv_gpu[
     elif m == 1 and n % WARP_SIZE == 0 and k % WARP_SIZE == 0:
         kernel_func = GEMVAlgorithm.GEVM_KERNEL
 
+    elif m == 1 and n % WARP_SIZE == 0:
+        # K is not aligned to WARP_SIZE; use GEVM_KERNEL with bounds checking.
```
govindansriram: you could make this the only gevm path

```python
elif m == 1 and n % WARP_SIZE == 0:
```

and in `gemv_gpu_dispatch` check for ` k % WARP_SIZE == 0` and add `check_k_bounds=True` based on the that.

When you do that, you could also use parameterized aliases to reduce code duplication

here is a simple example

```rust

from gpu import global_idx
from gpu.host import DeviceContext
from memory import LegacyUnsafePointer

comptime UnsafePointer = LegacyUnsafePointer[mut=True, ...]

fn add_kernel[
    dtype: DType,
    check_bounds: Bool = False,
](
    result: UnsafePointer[Scalar[dtype]],
    a: UnsafePointer[Scalar[dtype]],
    b: UnsafePointer[Scalar[dtype]],
    n: Int,
):
    var tid = Int(global_idx.x)
    comptime if check_bounds:
        if tid >= n:
            return
    result[tid] = a[tid] + b[tid]


# Parameterized alias: pins dtype=float32, leaves check_bounds open
comptime f32_add_kernel[check_bounds: Bool] = add_kernel[
    DType.float32, check_bounds
]


fn main() raises:
    var ctx = DeviceContext()
    comptime N = 64

    var a_host = alloc[Float32](N)
    var b_host = alloc[Float32](N)
    var c_host = alloc[Float32](N)

    for i in range(N):
        a_host[i] = Float32(i)
        b_host[i] = Float32(i * 2)

    var a_dev = ctx.enqueue_create_buffer[DType.float32](N)
    var b_dev = ctx.enqueue_create_buffer[DType.float32](N)
    var c_dev = ctx.enqueue_create_buffer[DType.float32](N)

    ctx.enqueue_copy(a_dev, a_host)
    ctx.enqueue_copy(b_dev, b_host)

    comptime kernel = f32_add_kernel[True]

    ctx.enqueue_function_experimental[kernel](
        c_dev,
        a_dev,
        b_dev,
        N,
        grid_dim=1,
        block_dim=N,
    )

    ctx.enqueue_copy(c_host, c_dev)
    ctx.synchronize()

    for i in range(N):
        var expected = Float32(i * 3)
        var got = c_host[i]
        if got != expected:
            print("FAIL at", i, ": expected", expected, "got", got)
            return

    print("PASS")
```

govindansriram (changes requested):

max/kernels/src/linalg/gemv.mojo
prsabahrami: Thanks for the review! Implemented your suggestion — merged the two GEVM branches into a single `m == 1 and n % WARP_SIZE == 0` path. The dispatch in `gemv_gpu_dispatch` now checks `k % WARP_SIZE` at runtime and passes `check_k_bounds=True/False` to `gevm_kernel` accordingly.

Tried the parameterized alias approach but hit Mojo's `raises` constraint (nested parametric function can't call `ctx.enqueue_function` which may raise). Used an if/else with two comptime kernel instantiations instead — same compile-time specialization, just a bit more verbose.

Re-benchmarked: 6.4x confirmed for non-aligned K, no regressions across 13 shapes.

prsabahrami:
hey @govindansriram Just got the agent to implement your feedbacks. let me know what u think!

zbosons:
@govindansriram is this PR ready for approval?

govindansriram (approved):
Thanks

zbosons:
!sync

prsabahrami:

prsabahrami:
Superseded by #6206 which is a clean rebase onto current main with reviewer feedback incorporated.

PR #6042, opened 2026-02-27
[Kernels] SM-aware block scaling and register optimization for topk

prsabahrami:
## Summary

Four optimizations to the topk GPU kernel:

1. **Register-based local max** — Replaced shared memory writes with register variables in `_topk_stage1`'s inner loop, eliminating unnecessary shared memory traffic.

2. **DMA copy engine** — Moved the input-to-temp-buffer copy from inside the GPU kernel to a host-level `enqueue_copy`, offloading to the GPU's dedicated DMA copy engine which runs in parallel with compute.

3. **SM-aware block scaling** — Replaced the fixed cap of 8 `num_blocks_per_input` with a heuristic targeting 128 total blocks across all inputs, ensuring GPU SMs are saturated regardless of batch size. The old cap of 8 left 94% of the H100's 132 SMs idle for small batch sizes.

4. **Consistent defaults** — Applied the same SM-aware heuristic to `_topk_gpu`'s internal path.

| Configuration | Baseline | Optimized | Speedup |
|---------------|----------|-----------|---------|
| N=256K, K=1, B=1 | 46.4 µs | 6.9 µs | **6.7x** |
| N=256K, K=50, B=1 | 594 µs | 144 µs | **4.1x** |
| N=256K, K=50, B=8 | 599 µs | 308 µs | **1.95x** |
| N=256K, K=255, B=8 | 2954 µs | 1618 µs | **1.83x** |

## Changes

- `max/kernels/src/nn/topk.mojo` — 4 commits, +40/-39 lines

## Test Plan

- [x] Benchmarked with verified speedup via NCU profiling
- [x] `./bazelw test //max/kernels/test/gpu/nn:test_topk_gpu.mojo.test` — PASSED

### DIFF
--- a/max/kernels/src/nn/topk.mojo
+++ b/max/kernels/src/nn/topk.mojo
@@ -1289,10 +1289,17 @@ fn _topk_gpu[
     if batch_size == 0:
         return
 
-    # Define the number of blocks per grid
-    var num_blocks_per_input_: Int = ceildiv(
-        N, block_size
-    ) if not num_blocks_per_input else num_blocks_per_input.value()
+    # Define the number of blocks per grid.
+    # Target enough total blocks to saturate the GPU's SMs.
+    var num_blocks_per_input_: Int
+    if num_blocks_per_input:
+        num_blocks_per_input_ = num_blocks_per_input.value()
+    else:
+        comptime target_total_blocks = 128
+        num_blocks_per_input_ = min(
+            ceildiv(N, block_size),
+            max(ceildiv(target_total_blocks, batch_size), 1),
+        )
     # Calculate largest num bytes of shmem for each stage
     if block_size % WARP_SIZE != 0:
         # TODO: Need to pad in this case
--- a/max/kernels/src/nn/topk.mojo
+++ b/max/kernels/src/nn/topk.mojo
@@ -816,12 +816,13 @@ fn _topk_stage1_old[
     ]()
 
     with PDL():
-        # Pack the topk_vals and topk_idxs into shared memory
+        # Find local max per thread using registers, then store once in shmem
         var block_offset = block_lane * block_size
         var stride = block_size * UInt(num_blocks_per_input)
-        topk_sram[tid] = TopK_2[T, largest]()
+        var local_max = TopK_2[T, largest]()
         for i in range(tid + block_offset, num_elements, stride):
-            topk_sram[tid].insert(_in_buffer[i], i)
+            local_max.insert(_in_buffer[i], i)
+        topk_sram[tid] = local_max
         barrier()
         var k_batch = max_k
         if K:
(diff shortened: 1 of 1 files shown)


zbosons:
Sorry for the delay in reviewing this. This PR correctly identifies that at small batch sizes, the old heuristic (`min(ceildiv(N, block_size), 8)`) under-utilizes the GPU — at B=1, only 8 of 132 SMs are active on an H100.

However, the proposed fix of targeting a fixed 128 total blocks (`min(ceildiv(N, block_size), max(ceildiv(128, batch_size), 1))`) introduces a potential regression at large batch sizes, where each input gets fewer blocks than before. At B=128, each input gets just 1 block scanning all 128K elements alone, vs 8 blocks before.

| Batch size | Old (per input, total) | PR (per input, total) | Suggested (per input, total) |
|---|---|---|---|
| 1 | 8, 8 | 128, 128 | 128, 128 |
| 8 | 8, 64 | 16, 128 | 16, 128 |
| 16 | 8, 128 | 8, 128 | 8, 128 |
| 64 | 8, 512 | 2, 128 | 8, 512 |
| 128 | 8, 1024 | 1, 128 | 8, 1024 |
| 256 | 8, 2048 | 1, 256 | 8, 2048 |

The two approaches converge at B=16. Below that the PR is better; above that the old code is better — there's no reason to cap total blocks since the GPU can schedule thousands of blocks fine.

Suggested fix — combine both by flooring at 8 blocks per input:

```
min(ceildiv(N, block_size), max(ceildiv(128, batch_size), 8))
```

This gets the small-batch win without regressing large-batch performance.

max/kernels/src/nn/topk.mojo
```diff
@@ -1304,10 +1289,17 @@ fn _topk_gpu[
...
-        N, block_size
-    ) if not num_blocks_per_input else num_blocks_per_input.value()
+    # Define the number of blocks per grid.
+    # Target enough total blocks to saturate the GPU's SMs.
+    var num_blocks_per_input_: Int
+    if num_blocks_per_input:
+        num_blocks_per_input_ = num_blocks_per_input.value()
+    else:
+        comptime target_total_blocks = 128
+        num_blocks_per_input_ = min(
```
zbosons, line 1299: Consider changing `1` to `8` here so blocks per input never drops below the old default:
```
max(ceildiv(target_total_blocks, batch_size), 8)
```
At large batch sizes (B=64, 128), the current floor of 1 means each input gets far fewer blocks than before, potentially regressing performance. See my top-level comment for the full table.

max/kernels/src/nn/topk.mojo
```diff
@@ -1618,10 +1611,18 @@ fn topk_gpu[
...
+    # saturate the GPU's SMs. 128 is a good target for modern GPUs.
+    var num_blocks_per_input_: Int
+    if num_blocks_per_input:
+        num_blocks_per_input_ = num_blocks_per_input.value()
+    else:
+        comptime target_total_blocks = 128
+        num_blocks_per_input_ = min(
+            ceildiv(N, block_size_),
+            max(ceildiv(target_total_blocks, internal_bs), 1),
+        )
```
zbosons, line 1625: Same here — change `1` to `8`:
```
max(ceildiv(target_total_blocks, internal_bs), 8)
```

zbosons (approved):
LGTM — nice find on the small-batch under-utilization. One change needed: on lines 1299 and 1625, change the floor from `1` to `8` so blocks per input never drops below the old default at large batch sizes:

```
max(ceildiv(target_total_blocks, batch_size), 8)
max(ceildiv(target_total_blocks, internal_bs), 8)
```

See my earlier comment for the full table showing why this matters.

zbosons:
!sync

prsabahrami:

prsabahrami:
Superseded by #6205 which is a clean rebase onto current main. The reviewer feedback (floor at 8) has been incorporated.

PR #6044, opened 2026-02-27
[Stdlib] Use SIMD scanning in _unsafe_strlen for null-byte detection

msaelices:
Replace the byte-by-byte null scan in `_unsafe_strlen` with SIMD-width block scanning, mirroring the pattern already used in `_memchr_impl`.

Two paths:
- **Unbounded** (`max == UInt.MAX`): scan `bool_mask_width`-wide blocks in a `while True` loop until a block with a null byte is found. Valid because the caller guarantees null-termination.
- **Bounded** (`max < UInt.MAX`): scan complete SIMD blocks up to `align_down(max, width)`, then a scalar tail for remaining bytes.

Compile-time contexts keep the original scalar loop via `is_compile_time()`.


mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -2665,7 +2665,49 @@ fn _unsafe_strlen(
...
+) -> UInt:
+    """SIMD-accelerated helper for `_unsafe_strlen`.
+
+    Scans for a null byte using SIMD-width blocks, then a scalar tail for any
+    remaining bytes within `max`.
+    """
+    comptime bool_mask_width = simd_width_of[DType.bool]()
+    var zero = SIMD[DType.uint8, bool_mask_width](0)
+
+    if max == UInt.MAX:
```
owenhilyard: I think `Optional` makes more sense here than using a magic number. 

If we want to stick with the magic number, then this behavior should be documented in the docstring.

mojo/stdlib/std/collections/string/string_span.mojo
```diff
@@ -2665,7 +2665,49 @@ fn _unsafe_strlen(
...
+            var block = ptr.load[width=bool_mask_width](i)
+            var bool_mask = block.eq(zero)
+            var mask = pack_bits(bool_mask)
+            if mask:
+                return UInt(i) + UInt(count_trailing_zeros(mask))
+            i += bool_mask_width
+
+    # Bounded case: process only complete SIMD blocks that fit within max.
+    var vectorized_end = align_down(Int(max), bool_mask_width)
+    for i in range(0, vectorized_end, bool_mask_width):
```
owenhilyard, line 2682: Does it make sense to offer a `rep cmpsb`/scalar equivalent for icache sensitive areas?

mojo/stdlib/std/collections/string/string_span.mojo
```diff
@@ -2665,7 +2665,49 @@ fn _unsafe_strlen(
...
+
+fn _unsafe_strlen_impl(
+    ptr: UnsafePointer[mut=False, Byte], max: UInt
+) -> UInt:
+    """SIMD-accelerated helper for `_unsafe_strlen`.
+
+    Scans for a null byte using SIMD-width blocks, then a scalar tail for any
+    remaining bytes within `max`.
+    """
+    comptime bool_mask_width = simd_width_of[DType.bool]()
```
owenhilyard, line 2650: I think this should be a user-controlled parameter with a default value. We, sadly, still have to deal with early AVX512 impls where 512-bit ops are slower than 2 256-bit ones.

owenhilyard (changes requested):

mojo/stdlib/std/collections/string/string_slice.mojo
msaelices: Good point. I've added explicit documentation to the `max` parameter explaining that `UInt.MAX` is the sentinel for 'no bound' and how callers should use it. Changing the public signature to `Optional[UInt]` would touch all existing callers and feels like a separate refactor, happy to follow up in a separate PR if preferred.

mojo/stdlib/std/collections/string/string_span.mojo
msaelices: The `simd_width_of[DType.bool]()` constant intentionally mirrors `_memchr_impl`, which also hard-codes the same width. Adding a user-controllable width parameter here but not in `_memchr_impl` and the other SIMD search helpers would create an inconsistency. If we want to address early-AVX512 throttling across the board, that seems like a separate, stdlib-wide effort. Happy to open a follow-up issue to track it.

mojo/stdlib/std/collections/string/string_span.mojo
msaelices: The `is_compile_time()` guard in `_unsafe_strlen` already routes compile-time calls through the scalar loop. For runtime icache-sensitive code that truly needs a scalar path, callers can keep the original byte-by-byte loop pattern themselves. Adding an explicit scalar overload to this private helper seems out of scope here, but it could be considered as a follow-up if there's a concrete use case.

mojo/stdlib/std/collections/string/string_slice.mojo
owenhilyard: Having that as a separate PR is fine.

mojo/stdlib/std/collections/string/string_span.mojo
owenhilyard: A follow up issue is probably warranted.

msaelices: Will do once this is handled.

mojo/stdlib/std/collections/string/string_slice.mojo
msaelices: I've opened this PR, which depends on another branch, so for now it lives in my fork: https://github.com/msaelices/modular/pull/2

mojo/stdlib/std/collections/string/string_span.mojo
msaelices: I've added a `simd_width` compile-time parameter (defaulting to `simd_width_of[DType.bool]()`) to both `_unsafe_strlen` and `_unsafe_strlen_impl` so callers can pass a narrower width on early AVX-512 CPUs. This is part of the follow-up PR that also addresses the `Optional[UInt]` change: https://github.com/msaelices/modular/pull/2

mojo/stdlib/std/collections/string/string_span.mojo
```diff
@@ -2657,15 +2657,64 @@ fn _unsafe_strlen(
...
+    """SIMD-accelerated helper for `_unsafe_strlen`.
+
+    Scans for a null byte using SIMD-width blocks, then a scalar tail for any
+    remaining bytes within `max`. When `max == UInt.MAX` the scan is
+    unbounded; all other values cap the scan to at most `max` bytes.
+
+    The SIMD width matches `_memchr_impl` (`simd_width_of[DType.bool]()`),
+    keeping SIMD behaviour consistent across stdlib string search helpers.
+    """
+    comptime bool_mask_width = simd_width_of[DType.bool]()
```
barcharcraz, line 2650: Isn't this wrong if the pointer is not aligned to the simd register width? In that case we might load off the end of a page!

msaelices: Great catch! thanks! Fixed here: https://github.com/msaelices/modular/commit/04fdc7902acc9de79d26e2d3bfb37ee12bdf6ef4

mojo/stdlib/std/collections/string/string_span.mojo
owenhilyard: Fantastic

PR #6046, opened 2026-03-01
[Stdlib] Add String.__iadd__ overload for Writable types

msaelices:
Add a generic `__iadd__[T: Writable]` overload to `String` so that any `Writable` value can be appended with `+=`, including t-strings (`s += t"{x}"`), integers, and other formattable types.

The existing `StringSlice` overload is preserved as the fast path (direct `memcpy` via `_iadd`). Mojo's overload resolution prefers the concrete `StringSlice` overload for `StringSlice` and `String` arguments, so there is no ambiguity or performance regression.

### DIFF
--- a/mojo/stdlib/std/collections/string/string.mojo
+++ b/mojo/stdlib/std/collections/string/string.mojo
@@ -982,6 +982,20 @@ struct String(
         """
         self._iadd(other.as_bytes())
 
+    fn __iadd__[T: Writable](mut self, other: T):
+        """Appends any writable value to this string.
+
+        This overload allows appending any type that implements `Writable`,
+        such as t-strings, integers, or other formattable types.
+
+        Parameters:
+            T: The type of the value to append, which must implement `Writable`.
+
+        Args:
+            other: The writable value to append.
+        """
+        self.write(other)
+
     @deprecated("Use `str.codepoints()` or `str.codepoint_slices()` instead.")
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/collections/string/string.mojo
```diff
@@ -982,6 +982,20 @@ struct String(
         """
         self._iadd(other.as_bytes())
 
+    fn __iadd__[T: Writable](mut self, other: T):
```
NathanSWard, line 985: I'm not a big fan of this.

Seeing stuff like
```
var s = "hello"
s += 42 # you can add a number to a string?
print(s) "Hello42"
```
is very counter intuitive and makes code difficult to understand.

We should just keep the `.write` method as the way to append something writable.

NathanSWard (changes requested):

mojo/stdlib/std/collections/string/string.mojo
msaelices: I don't have a strong opinion, so happy to discuss it.

NathanSWard:
Hey @msaelices - this is a pretty big change to the API of string. For something like this, I'd prefer if we had an issue opened for it to then get further feedback. Feel free to open an issue if you want to continue this discussion, however, in the interim, I'm going to close the PR.

PR #6048, opened 2026-03-02
[Stdlib][GPU] Add multi-dimensional block support for block primitives

bertaveira:
## Summary
Add 2D and 3D thread block support to all block-level GPU operations: sum, max, min, broadcast, and prefix_sum. The existing 1D `block_size` API is preserved; new overloads use `block_dim_x`, `block_dim_y`, and `block_dim_z` parameters. Thread IDs are linearized in row-major order: `x + y * dim_x + z * dim_x * dim_y`.

Key changes:
- `_block_reduce` now takes `block_dim_x/y/z`.
- `_prefix_sum` extracted as a shared helper to avoid duplication between 1D and ND prefix_sum overloads.
- New public overloads for sum, max, min, broadcast, and prefix_sum with `block_dim_x/y/z` parameters to maintain backwards compatibility.
- Added GPU tests for all 1D, 2D, and 3D block operations.

Please let me know if there is anything I can do to improve this PR (it is my first one here). I wasn't sure if I had to make an issue first but it seemed a simple enough change that I could help by making a PR, and it was a feature I was personally in need of for my project.

Assisted-by: Claude

## Testing
Tests have been added for these operations including the existing 1D versions of it. 

## Checklist

- [x] PR is small and focused — consider splitting larger changes into a
      sequence of smaller PRs
- [x] I ran `./bazelw run format` to format my changes
- [x] I added or updated tests to cover my changes
- [x] If AI tools assisted with this contribution, I have included an
      `Assisted-by:` trailer in my commit message or this PR description
      (see [AI Tool Use Policy](../AI_TOOL_POLICY.md))

PR #6049, opened 2026-03-02
[Stdlib][GPU] Fix warp.max/warp.min CUDA_ERROR_INVALID_PTX on SM 120

bertaveira:
## Summary
The `redux.sync.{max,min}.NaN.f32` PTX instruction is only available on SM 100 family GPUs (sm_100a, sm_101a), not on SM 120 (RTX 5090). The `_has_redux_f32_support` guard was using `_is_sm_100x_or_newer()` which incorrectly includes SM 110/120. This caused CUDA_ERROR_INVALID_PTX when calling warp.max or warp.min with Float32 on any SM 120 device (such as an RTX 5090).

This commit:
- Restricts `_has_redux_f32_support` to `sm_100a` and `sm_101a` only, matching the LLVM NVPTX backend's `hasReduxSyncF32()` guard.
- Fixes inline assembly register constraints from `"=r,r,i"` (integer) to `"=f,f,i"` (float), which is correct for f32 operands.

SM 120 devices now fall back to the shuffle-based reduction path which works correctly on all architectures.

Assisted-by: Claude

## Testing

Tests were added in #6048 and these PRs were tested together on an RTX 5090.

## Checklist

      sequence of smaller PRs

### DIFF
--- a/mojo/stdlib/std/gpu/primitives/warp.mojo
+++ b/mojo/stdlib/std/gpu/primitives/warp.mojo
@@ -1117,7 +1117,11 @@ fn prefix_sum[
 
 @always_inline("nodebug")
 fn _has_redux_f32_support[dtype: DType, simd_width: Int]() -> Bool:
-    return _is_sm_100x_or_newer() and dtype == DType.float32 and simd_width == 1
+    return (
+        (is_nvidia_gpu["sm_100a"]() or is_nvidia_gpu["sm_101a"]())
+        and dtype == DType.float32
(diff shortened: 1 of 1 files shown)


bertaveira:
Given that I only have an rtx5090 to test I could not test the change in inline assembly register constraints. i can remove that from this PR if you think that is the best way forward for this PR. I would like some inout there perhaps as it looks like it was a bug but can't say for sure without testing

abduld:
!sync

PR #6052, opened 2026-03-02
[MAX][config] Auto-default DeepSeek EP/DP for multi-GPU pipeline config

larme:
## Summary

This PR adds DeepSeek-specific multi-GPU parallelism defaults during `PipelineConfig.resolve()`. It will assign both EP and DP to the number of GPUs.

Because currently we are only doing special treatment for DeepSeek models in `config.py`, I add helper functions directly. We may consider a rule-table based approach like `[(rule1_pred, rule1_resolve)...]` if more model arch need special treatments so each model arch can add their own parameter defaults to this rule-table.

## Testing

minimal tests:

```
./bazelw test --config=disable-mypy \
  //max/tests/integration/unorganized:test_config_pure \
  --test_arg=-k \
  --test_arg='test_apply_deepseek_defaults'
```

## Checklist

      sequence of smaller PRs

### DIFF
--- a/max/python/max/entrypoints/cli/config.py
+++ b/max/python/max/entrypoints/cli/config.py
@@ -189,8 +189,16 @@ def is_multiple(field_type: Any) -> bool:
     return get_origin(field_type) is list
 
 
+def get_public_field_name(field_name: str, field_info: Any) -> str:
+    alias = getattr(field_info, "alias", None)
+    if isinstance(alias, str):
+        return alias
+    return field_name
+
+
 def get_normalized_flag_name(dataclass_field: Any, field_type: Any) -> str:
-    normalized_name = dataclass_field.name.lower().replace("_", "-")
+    public_name = getattr(dataclass_field, "public_name", dataclass_field.name)
+    normalized_name = public_name.lower().replace("_", "-")
 
     if is_flag(field_type):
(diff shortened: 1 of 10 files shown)


larme:

larme:
currently tests failed at `test_kbench.py`, which is unrelated to this PR

larme:
The PR is now a bit more complex than I originally planned. I found that if I override `ep_size`/`data_parallel_degree` when they equal 1, then even when the user passes `--ep-size=1` explicitly, the resolving function will still override `ep_size`. While this would help DeepSeek model get the correct `ep_size`, I think implicitly overriding user-provided arguments is not a good idea. I'd rather pass the invalid value to downstream code and let it raise an error.

So I need to distinguish whether the user explicitly provided `ep_size=1` or Click filled `ep_size` with the default value of 1. Unfortunately this is tricky because at the Click parsing stage we don't know the model arch, and at the `PipelineConfig` stage we don't have access to Click's data. My current solution is to add Click info to `PipelineConfig` [0]. I'd like to hear your feedback about this approach.

[0]: https://github.com/modular/modular/pull/6052/changes#diff-e7cb4e5b8777f5d6ae970015afbebedf6ca95f6dfaf09f8faf7c92f8c62308d6R418

PR #6054, opened 2026-03-02
[Stdlib] Change _unsafe_strlen max parameter to Optional[UInt]

msaelices:
Replace the `UInt.MAX` magic number sentinel with `Optional[UInt] = None` for the `max` parameter of `_unsafe_strlen`. `None` means unbounded (scan until the null terminator); an explicit value caps the scan.

Follows up on review feedback in #6044.


msaelices:
Superseded — will re-open targeting the simd-unsafe-strlen base branch instead.

Issue #5707, opened 2025-12-24
[BUG]: MAX fails to build models on Blackwell (sm_120) because tcgen05 instruction not supported


rolson24:
### On which model did the bug occur?

Qwen/Qwen3-8B

### Bug description

The model fails to build because the fp8 and fp4 matmul kernels for Blackwell (sm_100a, sm_101a) use the tcgen05 instruction which doesn't exist on sm_120. If there was a fallback to a naive fp8 matmul kernel for sm_120 that would be ideal before fp8 kernels are written to support mxf4 instructions (used only in sm_120 gpus). Issue [MAXR-1808](https://linear.app/modularml/issue/MAXR-1808/bugcant-load-some-gemma-models-on-blackwell-gpu) appears to be the same issue.

### Steps to reproduce

`max serve --model Qwen/Qwen3-8B` running on RTX 5090 fails and produces this stack trace:

```
max/kernels/src/Mogg/MOGGKernelAPI:1:1: error: failed to run the pass manager for offload functions
max/kernels/src/linalg/matmul/gpu/sm100/matmul.mojo:1988:4: error: function instantiation failed
max/kernels/src/linalg/matmul/gpu/sm100/matmul.mojo:1988:4: note: call expansion failed with parameter value(s): (..., ..., ..., ..., ..., ..., ..., ..., ..., "transpose_b": true, ..., ..., ..., "register_based_epilogue": true, "pdl_level": 0, ...)
max/kernels/src/linalg/matmul/gpu/sm100/matmul.mojo:1988:4: note: function instantiation failed
max/kernels/src/linalg/matmul/gpu/sm100/matmul.mojo:2380:44: note: call expansion failed
open-source/max/mojo/stdlib/std/gpu/compute/tcgen05.mojo:84:31: note: call expansion failed
open-source/max/mojo/stdlib/std/builtin/constrained.mojo:56:6: note: constraint failed: The tcgen05 instructions are only applicable on nVidia Blackwell (sm_100a, sm_101a) hardware.
max/kernels/src/linalg/matmul/gpu/sm100/matmul.mojo:1988:4: error: function instantiation failed
max/kernels/src/linalg/matmul/gpu/sm100/matmul.mojo:1988:4: note: call expansion failed with parameter value(s): (..., ..., ..., ..., ..., ..., ..., ..., ..., "transpose_b": true, ..., ..., ..., "register_based_epilogue": true, "pdl_level": 0, ...)
[...]

PR #6059, opened 2026-03-04
[MAX] Fix SM100 matmul dispatch gating on SM120 GPUs

rolson24:
## Summary

Fixes issue #5707 by fixing gating for tcgen05 kernels (sm120 does not have tcgen05 instructions)
(Assisted-by Codex)

## Testing

`./bazelw run //max/python/max/entrypoints:pipelines -- serve --model-path Qwen/Qwen3-8B`
Model now responds to short queries at ~95 tok/s for short queries on RTX 5090.
Obviously this could be improved with better kernels, but still pretty decent for the naive matmul kernels.

## Checklist

      sequence of smaller PRs


rolson24:

rolson24:
Updated to allow all architectures newer than h100 except sm_120 to use tcgen05 path.

abduld:
@BradLarson do you mind taking a look

max/kernels/src/linalg/matmul/gpu/__init__.mojo
```diff
@@ -496,10 +496,15 @@ fn _matmul_gpu[
...
 
-    comptime if (
+    # Keep post-H100 NVIDIA GPUs on the SM100 structured path by default,
+    # but explicitly exclude consumer Blackwell sm_120 where tcgen05 is
+    # not available.
+    comptime use_sm100_structured_kernels = (
         has_nvidia_gpu_accelerator()
         and ctx.default_device_info.compute > H100.compute
-    ):
+        and "sm_120" not in _accelerator_arch()
```
abduld, line 505: can you use the existing `_is_sm_120x` in the sys.info module?

BradLarson: Another approach, which is what I used locally, is to use `and _has_blackwell_tcgen05()` in place of `and ctx.default_device_info.compute > H100.compute` to gate these, so that this works for `sm_110`, `sm_120`, and `sm_121`.

BradLarson:
Thanks for your work on this, and for putting up this PR. In parallel, I had been working on the same sort of gating for the `sm_100` matmuls. I needed that for a use case I was preparing, so I went ahead and merged my own version of this internally. I used the `_has_blackwell_tcgen05()` check I suggested above, which will also support `sm_110` and `sm_121` Blackwell hardware, in addition to `sm_120`.

This should hopefully be out in the next nightly. Again, sorry for stepping on your toes here, but thanks for taking the time to point this out.

BradLarson:
Actually, my bad, `_has_blackwell_tcgen05()` won't work for a check here and we're reverting my internal change. That is a device-side check, rather than a host-side one, so there's still an opportunity to get this right by using a proper host-side check for these conditions to make sure they are built only when targeting an accelerator that supports tcgen05 instructions.

rolson24:
No worries. I will take a look and see if I can come up with a good way to do it.

BradLarson:
Again, sorry, we may have fixed this robustly internally as a follow-on. If you can wait until the next nightly, that should have a fix in it that you can try. If that works for your RTX 5090, we might be good here.

BradLarson:
Revisiting this older PR, we should have this being properly checked in main since the above-referenced internal fixes, and have been running successfully on `sm_120` / `sm_121` since those fixes, so I'll close this. Thanks for pushing us in the right direction here.

PR #6069, opened 2026-03-05
[Refactor] Pixel Generation Model's config

pei0033:
## Summary

This PR aligns pixel-generation config handling with the LLM architecture-config pattern.

### What Changed
- Removed hardcoded pixel tokenizer lengths from `registry.py`.
- Switched pixel tokenizer setup to use `arch.config.initialize(...)`.
- Added/updated Flux architecture configs to explicitly define tokenizer lengths:
  - Flux1: `max_seq_len=77`, `secondary_max_seq_len=512`
  - Flux2: `max_seq_len=512`
- Kept Flux startup constraints (single-device checks) in architecture config initialization.
- Standardized component config construction to `initialize_from_config(...).`
- Updated all related call sites.

## Testing
Added `test_registry__retrieve_factory_pixel_uses_arch_config_max_length()` to verify pixel tokenizer `max_length` is sourced from `ArchConfig.get_max_seq_len()`.
```
./bazelw test //max/tests/integration/unorganized:test_registry 
```

## Checklist

      sequence of smaller PRs
---
`Assisted-by: Codex`


max/python/max/pipelines/architectures/flux1_modulev3/arch.py
```diff
@@ -11,17 +11,40 @@
...
 
-from .model_config import FluxConfig
 from .pipeline_flux import FluxPipeline
 
+
+@dataclass(kw_only=True)
+class FluxArchConfig(ArchConfig):
+    """Pipeline-level config for Flux1 (implements ArchConfig; no KV cache)."""
+
+    max_seq_len: int = 77
```
KCaverly, line 34: Are these values ever overridden?

max/python/max/pipelines/architectures/flux2_modulev3/arch.py
```diff
@@ -25,21 +25,24 @@
...
 
+_FLUX2_DEFAULT_MAX_SEQ_LEN = 512
+
 
 @dataclass(kw_only=True)
 class Flux2ArchConfig(ArchConfig):
     """Pipeline-level config for Flux2 (implements ArchConfig; no KV cache)."""
 
-    pipeline_config: PipelineConfig
+    max_seq_len: int = 512
```
KCaverly, line 34: This conflicts with the max_seq_len of 77 set above, which is the true constraint here?

KCaverly (approved):
Looks good! Just two small questions on consistency between the two config classes. Once thats resolved, ill bring this in.

max/python/max/pipelines/architectures/flux1_modulev3/arch.py
pei0033: This value is typically fixed per architecture and does not change during service.

max/python/max/pipelines/architectures/flux2_modulev3/arch.py
pei0033: FLUX.1 uses `max_seq_length` (for clip) set to 77 and `secondary_max_seq_length` (for T5) set to 512, while FLUX.2 uses `max_seq_length` (for Mistral3) set to 512.
These values differ depending on the architecture.

pei0033: FLUX.2-klien also uses the FLUX.2 ArcConfig. While the klien model employs Qwen3 as its text encoder, it uses a max length of 512, identical to FLUX2, so this should not pose a problem.

KCaverly:
!sync

Issue #5838, opened 2026-01-24
[Feature Request] [stdlib] Prettier `SIMD` scalar printing


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

After the change making `UInt` an alias for `Scalar[DType.uindex]` printing any collection of `UInt` gives e.g. `[SIMD[DType.uindex, 1](1), ...]`. I think we should make the printing of scalars use the aliases that we have, since IMO this introduces too much visual noise when debugging for no apparent benefit beyond making a parser for this a bit simpler. IMO it's a change that will be worth it.

### What is your motivation for this change?

...

### Any other details?

...

soraros:
Actually, what is preferable? Do we print the alias `UInt`? What happens when `size > 1`? What about other scalar types? Do we single out `Int` so it always prints the number and never the type alias? etc. etc. These are not particularly difficult questions, but we do have to answer and document them before any implementation attempt.

NathanSWard:
+1 to this.

I think we should have the `Scalar` SIMD types print their type alias name.
e.g.
```
SIMD[DType.uint32, 1] -> UInt32
SIMD[DType.int64, 1] -> Int64
SIMD[DType.uint, 1] -> UInt
```
etc...

nejraberberovic:
Hi, can I take this on?

martinvuyk:
I actually wanted to solve this using https://github.com/modular/modular/pull/5874/changes#diff-9bf9f81e4191b9f2ea900b79b5960e3b206cbf3f308f2d266ea91a05456b222aR528 by changing the string representation with the uppercased dtype initials (when size is 1). @NathanSWard is that PR still being considered? I think it will replace a lot of the write_to and write_repr_to custom implementations that follow the same pattern

soraros:
@JoeLoser This issue is marked as completed, yet we still have quite a few open PRs for it. Are they still relevant, or should we close them?

JoeLoser:
Let's take a look at the PRs to see if any are worth pursuing still after the next nightly goes out which contains a solution to this problem. I suspect we can close them, but worth verifying. Thanks for flagging this!  This was one of the issues just burned down in our rush to Mojo 1.0 🙂

PR #6071, opened 2026-03-05
[Feature Request] [stdlib] Prettier SIMD scalar printing

nejraberberovic:
SIMD.write_repr_to now checks comptime if Self.size == 1 and, in that case, prints the alias name derived from the dtype. A compile‑time helper (_scalar_alias_from_dtype) computes the alias by transforming the short dtype name returned by DType.write_to.
Prefix rules:
- uint… → UInt…
- int… → Int…
- otherwise, capitalize the first letter (so float32→Float32, float8_e4m3fn→Float8_e4m3fn, etc.).

For size > 1 the original SIMD[DType.foo, N](…) output is unchanged. No reflection or runtime hacks were introduced. The change is local to write_repr_to as requested.

Related issue: #5838 

Tests added/modified
test_simd.mojo
The scalar expectations were updated and new assertions inserted (including a uint32‑4 test). A new collection test ensures wrappers print nicely. All tests pass.

### DIFF
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -103,56 +103,29 @@ from .dtype import (
 
 
 # Compute a scalar alias name for `SIMD[dtype, 1]`.
-# Maps the DType short name to the usual alias (e.g. "uint32" -> "UInt32",
-# "int64" -> "Int64", otherwise Capitalize(first)).  Runs at comptime.
+# Maps the DType string name to a more readable form (e.g. "uint32" -> "UInt32", "float32" -> "Float32")
 @always_inline("nodebug")
 fn _scalar_alias_from_dtype(dtype: DType) -> String:
     var name = String()
     dtype.write_to(name)
 
-    # Prefer explicit aliases for common integral dtypes for readability.
-    if dtype == DType.uint8:
-        return String("UInt8")
-    elif dtype == DType.int8:
-        return String("Int8")
-    elif dtype == DType.uint16:
-        return String("UInt16")
-    elif dtype == DType.int16:
-        return String("Int16")
-    elif dtype == DType.uint32:
-        return String("UInt32")
-    elif dtype == DType.int32:
-        return String("Int32")
-    elif dtype == DType.uint64:
-        return String("UInt64")
-    elif dtype == DType.int64:
-        return String("Int64")
-    elif dtype == DType.uint128:
-        return String("UInt128")
-    elif dtype == DType.int128:
-        return String("Int128")
-    elif dtype == DType.uint256:
-        return String("UInt256")
-    elif dtype == DType.int256:
-        return String("Int256")
-    elif dtype == DType.uint:
+    if len(name) == 0:
+        return name
+
+    if name == "uint":
         return String("UInt")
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -101,6 +101,60 @@ from .dtype import (
...
+        var iter = s.codepoint_slices()
+        var first_slice = iter.peek_next().value()
+        var first_upper = first_slice.upper()
+        var total = s.byte_length()
+        var first_len = first_slice.byte_length()
+        var rest = StringSlice(
+            ptr=s.unsafe_ptr() + first_len, length=total - first_len
+        )
+        return String(first_upper, rest)
+    return String()
```
barcharcraz: I think we should either _only_ do the fallback path here or only do the "bit switch statement" and either error out or just print the raw dtype repr in the fallback. I lean toward the former.

If you like I can simply make this change when I merge

barcharcraz (approved):
Amazing timing! Once we transition the Int type to really be "Scalar[dtype.int]" this this will mean the value of "repr" for ints will remain as-is!

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -101,6 +101,60 @@ from .dtype import (
...
+    elif dtype == DType.int128:
+        return String("Int128")
+    elif dtype == DType.uint256:
+        return String("UInt256")
+    elif dtype == DType.int256:
+        return String("Int256")
+    elif dtype == DType.uint:
+        return String("UInt")
+    elif dtype == DType.int:
+        return String("Int")
```
abduld: can you write as 

```
if dtype.is_integral():
  return str(dtype).replace("uint", "UInt").replace("int", "Int")
```

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -101,6 +101,60 @@ from .dtype import (
...
 # ===----------------------------------------------------------------------=== #
 
+
+# Compute a scalar alias name for `SIMD[dtype, 1]`.
+# Maps the DType short name to the usual alias (e.g. "uint32" -> "UInt32",
+# "int64" -> "Int64", otherwise Capitalize(first)).  Runs at comptime.
+@always_inline("nodebug")
+fn _scalar_alias_from_dtype(dtype: DType) -> String:
+    var name = String()
+    dtype.write_to(name)
```
abduld, line 110: why compute this and then discard that in the switch statement below?

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -101,6 +101,60 @@ from .dtype import (
...
+    if len(s) > 0:
+        var iter = s.codepoint_slices()
+        var first_slice = iter.peek_next().value()
+        var first_upper = first_slice.upper()
+        var total = s.byte_length()
+        var first_len = first_slice.byte_length()
+        var rest = StringSlice(
+            ptr=s.unsafe_ptr() + first_len, length=total - first_len
+        )
+        return String(first_upper, rest)
```
abduld: this seams overkill. You know the results are ascii and not empty, so you can just convert the first byte using toUpper

PR #6074, opened 2026-03-05
[Stdlib] Fix Interval comparison operators to use lexicographic order

msaelices:
The four comparison operators on `Interval[T]` were inconsistent:
- `__lt__` and `__le__` compared by `start` only
- `__gt__` and `__ge__` compared by `end` only

This violated transitivity. For example, with `A=[1,10]`, `B=[2,5]`, `C=[3,8]`:
- `B < C` held (`start` 2 < 3)
- `B > C` also held (`end` 5 > 8 = False, wait, that's False, but the inconsistency meant sorting was undefined)

More concretely: `Interval(1,10) >= Interval(2,9)` returned `True` (because `end` 10 >= 9) while `Interval(1,10) < Interval(2,9)` returned `True` (because `start` 1 < 2), both held simultaneously for distinct intervals.

Fix all four operators to use lexicographic order (`start` first, `end` as tiebreaker), giving a proper total order consistent with how `IntervalTree` already sorts nodes.

### DIFF
--- a/mojo/stdlib/std/collections/interval.mojo
+++ b/mojo/stdlib/std/collections/interval.mojo
@@ -210,51 +210,69 @@ struct Interval[T: IntervalElement](
         """
         return self.start <= other.start and self.end >= other.end
 
-    fn __le__(self, other: Self) -> Bool:
-        """Returns whether this interval is less than or equal to another
-        interval.
+    fn __lt__(self, other: Self) -> Bool:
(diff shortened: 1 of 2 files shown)


msaelices:
Superseded by #6091 which fixes the same comparison semantics and also adds the `Comparable` trait, removing the redundant operator implementations.

PR #6075, opened 2026-03-05
[Stdlib] Use _simd_apply in cbrt for compile-time loop unrolling

msaelices:
The float32 path of `cbrt()` used a runtime `for` loop to apply the scalar `_cbrtf()` function element-by-element:

```mojo
for i in range(width):
    result[i] = _cbrtf(rebind[Float32](x[i]))
```

All similar functions in `math.mojo` (`hypot`, `atan2`, `fma`, etc.) use `_simd_apply` with a `@parameter` dispatch function, which uses `comptime for` internally and allows the compiler to unroll the loop at compile time.

Replace the runtime loop with `_simd_apply` and a local `@parameter` dispatch wrapper. The `_cbrtf` implementation is kept unchanged since it is a pure Mojo function compatible with GPU targets (unlike `_call_libm` which is CPU-only).

### DIFF
--- a/mojo/stdlib/std/math/math.mojo
+++ b/mojo/stdlib/std/math/math.mojo
@@ -2454,12 +2454,16 @@ def cbrt[
     elif dtype == DType.float64:
         return _call_libm["cbrt"](x)
 
-    var result = SIMD[DType.float32, width]()
+    @always_inline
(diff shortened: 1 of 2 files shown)


abduld:
!sync

Issue #5856, opened 2026-01-27
[BUG] [mojo-lang] Can't run `log10` at compile time


martinvuyk:
### Bug description

A lot of the math functions can't be run at compile time. When one wants precision and to pre-caclulate everything and store in a collection they are necessary

### Steps to reproduce

...

### System information

`Mojo 0.26.2.0.dev2026012605`

soraros:
A lot of them use the `libm` implementation. I suspect we could make many of them work by simply adding a comptime branch that uses the LLVM intrinsic.

```mojo

from std.math.math import _llvm_unary_fn

fn log10[dtype: DType, width: Int, //](x: SIMD[dtype, width]) -> type_of(x):
  return _llvm_unary_fn["llvm.log10"](x)

fn main():
  comptime res = log10(100.0)
  print(res)  # prints 2.0
```

soraros:
Actually, quite a few of functions will work with this fix. I think it's a good first issue? @JoeLoser

JoeLoser:
Yep, that seems reasonable. I'll add the appropriate labels 👍

dgurchenkov:
@JoeLoser two questions:

1. why don't we always use the llvm intrinsic? In other words, how the intrinsic is supposed to be lowered in Clang or other compilers?
2. If the right way is to use llvm intrinsic for compile-time compute only, then we have the `sys.is_compile_time` so this is just a change in log in math.mojo? That function already branches for various gpus, it just needs a comptime branch?

samufi:
This might be blocked by #6449. I have not tested all functions, but some are not implemented in the MLIR interpreter.

PR #6079, opened 2026-03-06
[stdlib][math] Enable compile-time evaluation of log10

lamija-i:
log10 currently falls back to libm on CPU, which cannot be evaluated during compile-time execution. As a result, expressions such as:

comptime log10(100.0)

fail because libm functions are not available in compile-time contexts.

This patch adds a compile-time branch using the LLVM intrinsic:

@parameter
if is_compile_time():
    return _llvm_unary_fn["llvm.log10"](x)

This allows log10 to be evaluated during compilation while preserving the existing runtime behavior.

referencing issue #5856


abduld:
!sync

PR #6082, opened 2026-03-06
[Stdlib] Fix tanh() float64 accuracy by using expm1-based formula

msaelices:
## Summary

- `tanh()` used a P6/Q3 Padé polynomial designed for float32 accuracy (~1e-8 relative error). For float64 inputs this produced only ~8 decimal digits instead of the expected ~15.
- For float64 on CPU, switch to the numerically stable identity `tanh(x) = -expm1(-2|x|) / (expm1(-2|x|) + 2)`. `expm1` avoids catastrophic cancellation for small arguments and matches the approach used by FreeBSD libm / Cephes for double-precision tanh.
- GPU float64 continues to use the existing polynomial (no hardware intrinsic available; libm is not accessible on GPU).
- Adds `test_tanh_libm_f64` which compares Mojo float64 tanh against libm with a 1e-14 tolerance (previously the error was ~1e-8).

### DIFF
--- a/mojo/stdlib/std/math/math.mojo
+++ b/mojo/stdlib/std/math/math.mojo
@@ -1141,6 +1141,18 @@ fn tanh[
                 instruction="tanh.approx.f32", constraints="=f,f"
             ](x)
 
+    comptime if dtype == DType.float64 and not is_gpu():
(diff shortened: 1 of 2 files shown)


abduld:
!sync

PR #6083, opened 2026-03-06
[Stdlib] Add @always_inline to Span-based reduction wrapper functions

msaelices:
## Summary

The single-argument overloads of `max`, `min`, `sum`, `product`, and `mean` in `algorithm/reduction.mojo` that accept `Span[Scalar[dtype], _]` are thin wrappers that immediately delegate to the multi-argument versions. Without `@always_inline`, the call is not elided, adding unnecessary overhead in tight loops.

Adds `@always_inline` to all five functions.

### DIFF
--- a/mojo/stdlib/std/algorithm/reduction.mojo
+++ b/mojo/stdlib/std/algorithm/reduction.mojo
@@ -998,6 +998,7 @@ fn _simd_max_elementwise[
     return _max(x, y.cast[acc_type]())
 
 
+@always_inline
 fn max[dtype: DType](src: Span[Scalar[dtype], _]) raises -> Scalar[dtype]:
     """Computes the max element in a buffer.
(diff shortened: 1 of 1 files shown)


NathanSWard:
> Without @always_inline, the call is not elided, adding unnecessary overhead in tight loops.

Is this true? there is a chance the function will still be inlined even without the `@always_inline` attribute - especially since the function is so small. Ideally for something like this I'd like to see a codegen comparison to prove that the functions weren't inlined before.

msaelices:
Closing this PR, as codegen comparison using `compile_info` confirms the compiler already inlines these wrapper functions without `@always_inline`. The generated assembly is identical with and without the decorator (same line counts, same call instructions). Thanks @NathanSWard for the suggestion to verify.

PR #6084, opened 2026-03-06
[Stdlib] Implement SIMD-width-based unroll factor heuristic in elementwise/reduction loops

msaelices:
## Summary

Five hot loops in `algorithm/functional.mojo` and `algorithm/reduction.mojo` had `unroll_factor = 8` hard-coded with TODO comments asking for a cost heuristic.

Replace with `max(1, min(8, simd_width // 4))`:
- Scales unrolling to the native SIMD width of the target.
- Caps at 8 to avoid excessive code-size on wide-SIMD targets (AVX-512, SVE).
- Clamps to at least 1 to satisfy the `unroll_factor > 0` invariant.

Affected sites: `_elementwise_impl_cpu_1d`, `_elementwise_impl_cpu_nd`, `_stencil_impl_cpu` (functional.mojo) and `map_reduce`, `reduce_boolean` (reduction.mojo).


soraros:
I can't comment on whether this is a good heuristics. I think you could use `clamp` though.

mojo/stdlib/std/algorithm/functional.mojo
```diff
@@ -1591,7 +1591,7 @@ fn _elementwise_impl_cpu_1d[
     """
     comptime assert rank == 1, "Specialization for 1D"
 
-    comptime unroll_factor = 8  # TODO: Comeup with a cost heuristic.
+    comptime unroll_factor = max(1, min(8, simd_width // 4))
```
abduld: can you use the clamp function for these

msaelices:
@abduld could you PTAL?

mojo/stdlib/std/algorithm/functional.mojo
msaelices: done

PR #6088, opened 2026-03-06
[Stdlib] Add variadic and Span overloads for math.hypot()

msaelices:
## Summary

- Add `hypot[dtype](*args)` variadic overload for n-dimensional Euclidean distance, matching Python's `math.hypot(*coordinates)`.
- Add `hypot[dtype](Span)` overload for computing distance from a span of values.
- Both use max-abs scaling to avoid overflow/underflow with large/small values.
- Remove the existing TODO comment requesting this feature.

---

Assisted-by: AI


mojo/stdlib/std/math/math.mojo
```diff
@@ -2487,6 +2486,126 @@ fn hypot[
...
+
+    if max_abs == 0:
+        return Scalar[dtype](0)
+
+    # SIMD pass 2: sum of scaled squares.
+    var inv_max = Scalar[dtype](1) / max_abs
+    var sum_vec = SIMD[dtype, w](0)
+    i = 0
+    while i + w <= n:
+        var v = (ptr + i).load[width=w]() * inv_max
```
abduld:
```suggestion
        var v = ptr.load[width=w](i) * inv_max
```

msaelices:
@abduld could you PTAL?

PR #6090, opened 2026-03-06
[Stdlib] Pre-allocate output list in Counter.elements()

msaelices:
## Summary

- Pre-allocate `List` capacity in `Counter.elements()` by computing the total positive count upfront, avoiding repeated reallocations for large counters.

---

Assisted-by: AI


JoeLoser (commented):
Thanks for the PR! The change looks correct — two comments before merging:

1. **Benchmarks**: This adds a second full pass over the dict to compute capacity. Can you share before/after numbers for a few counter sizes (e.g., 10, 1000, and 1M total elements)? For small counters the extra iteration could be a net regression, so it'd be good to see where the crossover is.

2. **Test for negative counts**: The `if count > 0` guard is new logic that isn't covered by the existing `test_elements`. Negative counts are possible via `subtract()`, so a test case like this would exercise that path:

```mojo
def test_elements_with_negative_counts() raises:
    var c = Counter[String]()
    c["a"] = 3
    c["b"] = 0
    c["c"] = -2

    var elements = c.elements()
    assert_equal(len(elements), 3)
```

PR #6091, opened 2026-03-06
[Stdlib] Add Comparable to Interval with lexicographic ordering

msaelices:
## Summary

- Add `Comparable` trait to `Interval`, replacing `Equatable` (which `Comparable` extends).
- Fix broken comparison semantics where `__lt__`/`__le__` compared by `start` while `__gt__`/`__ge__` compared by `end`, violating transitivity.
- Use lexicographic ordering (start first, end as tiebreaker) in `__lt__`, and let `Comparable` derive `__gt__`, `__le__`, `__ge__` consistently.
- Note: this also addresses item 1.2 (comparison semantics bug) from the improvement opportunities, superseding PR #6074.

### DIFF
--- a/mojo/stdlib/std/collections/interval.mojo
+++ b/mojo/stdlib/std/collections/interval.mojo
@@ -75,7 +75,7 @@ trait IntervalElement(Comparable, Copyable, Intable, Writable):
 
 struct Interval[T: IntervalElement](
     Boolable,
-    Equatable,
+    Comparable,
     ImplicitlyCopyable,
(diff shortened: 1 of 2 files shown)


NathanSWard:
@msaelices - do you mind if we hold off on `IntervalTree` fixes? I'm not even sure this is a type that belongs in the standard library - so I think there needs to be some discussion there before we spend energy improving its current API. It's a bit of a legacy type and never (hardly) used anywhere I believe.

PR #6092, opened 2026-03-06
[Stdlib] Add Sized and iteration support to IntervalTree

msaelices:
## Summary

- Add `__len__` to `IntervalTree` exposing the existing `_len` field
- Add `__iter__` with in-order traversal yielding `Interval[T]` in sorted order
- Fix pre-existing bug: `_del_helper` crashed on empty trees (missing null check)


NathanSWard:
See here: https://github.com/modular/modular/pull/6091#issuecomment-4013347714

PR #6099, opened 2026-03-06
[Stdlib] Add Equatable and Writable to benchmark Mode

msaelices:
## Summary

- Add `Equatable` and `Writable` to `Mode` trait list
- `write_to` outputs `"Benchmark"` or `"Test"`
- `write_repr_to` outputs `"Mode.Benchmark"` or `"Mode.Test"`
- `__ne__` is not needed — `Equatable` provides a default implementation

---

Assisted-by: AI

### DIFF
--- a/mojo/stdlib/std/benchmark/bencher.mojo
+++ b/mojo/stdlib/std/benchmark/bencher.mojo
@@ -459,16 +459,11 @@ struct BenchConfig(Copyable, Writable):
             writer: The writer to write to.
         """
 
-        @parameter
-        fn fields(mut w: Some[Writer]):
-            w.write_string("min_runtime_secs=")
-            w.write(self.min_runtime_secs)
-            w.write_string(", max_runtime_secs=")
-            w.write(self.max_runtime_secs)
-            w.write_string(", max_iters=")
-            w.write(self.max_iters)
-
-        fmt.FormatStruct(writer, "BenchConfig").fields[FieldsFn=fields]()
+        fmt.FormatStruct(writer, "BenchConfig").fields(
+            fmt.Named("min_runtime_secs", self.min_runtime_secs),
+            fmt.Named("max_runtime_secs", self.max_runtime_secs),
+            fmt.Named("max_iters", self.max_iters),
+        )
 
 
 @fieldwise_init
--- a/mojo/stdlib/std/benchmark/bencher.mojo
+++ b/mojo/stdlib/std/benchmark/bencher.mojo
@@ -515,7 +515,7 @@ struct BenchmarkInfo(Copyable):
 
 
 @fieldwise_init
-struct Mode(ImplicitlyCopyable):
+struct Mode(Equatable, ImplicitlyCopyable, Writable):
     """Defines a Benchmark Mode to distinguish between test runs and actual benchmarks.
     """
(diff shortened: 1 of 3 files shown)


abduld:
!sync

mojo/stdlib/std/benchmark/bencher.mojo
```diff
@@ -453,6 +453,36 @@ struct BenchConfig(Copyable):
 
         argparse()
```
JoeLoser, line 437: Nit: all existing `write_to`/`write_repr_to` in this file use `def`, but these new ones use `fn`. Please switch all four new methods to `def` for consistency.

Also — `write_to` and `write_repr_to` produce nearly identical output here (both `BenchConfig(min_runtime_secs=..., ...)`). Compare with `BenchMetric` where `write_to` → `"elements (GElems/s)"` (human-friendly) and `write_repr_to` → `BenchMetric(code=0, name="elements", ...)` (structural). If you want them identical, consider having `write_to` delegate to `write_repr_to` to avoid duplication.

mojo/stdlib/std/benchmark/bencher.mojo
```diff
@@ -453,6 +453,36 @@ struct BenchConfig(Copyable):
...
+        )
+
+    @no_inline
+    fn write_repr_to(self, mut writer: Some[Writer]):
+        """Writes the repr of this `BenchConfig` to a writer.
+
+        Args:
+            writer: The writer to write to.
+        """
+
```
JoeLoser, line 461: `write_repr_to` is typically the complete, unambiguous representation. This only includes 3 of 10 fields — `out_file`, `num_warmup_iters`, `max_batch_size`, `num_repetitions`, `flush_denormals`, `show_progress`, and `format` are all omitted. Consider including all fields in `write_repr_to`, even if `write_to` stays abbreviated.

mojo/stdlib/std/benchmark/bencher.mojo
```diff
@@ -558,6 +588,27 @@ struct Mode(ImplicitlyCopyable):
 
         return self.value == other.value
 
+    fn write_to(self, mut writer: Some[Writer]):
+        """Writes the mode name to a writer.
+
+        Args:
+            writer: The writer to write the `Mode` to.
+        """
+        if self.value == Self.Benchmark.value:
```
JoeLoser, line 579: Nit: other `write_to` implementations in this file use `writer.write(...)`. Using `writer.write("Benchmark")` / `writer.write("Test")` here would be more consistent.

JoeLoser (commented):
Thanks for the contribution! This needs a rebase to resolve merge conflicts with main.

One thing I can't comment on inline (it's outside the diff context): the existing `__eq__` docstring on `Mode` is misleading — it says "Check if its Benchmark mode or test mode" / returns "If its a test mode or benchmark mode." Since you're already touching this struct, could you fix it to match the pattern at `Format.__eq__`? e.g.:

```mojo
    def __eq__(self, other: Self) -> Bool:
        """Checks if two Mode objects are equal.

        Args:
            other: The `Mode` to compare with.

        Returns:
            True if the two `Mode` objects are equal, false otherwise.
        """
```

PR #6103, opened 2026-03-06
[Stdlib] Add LinkedList.index() method

msaelices:
Adds `index(value, start=0, stop=None)` to `LinkedList`, making it consistent with `List.index()`.

Traverses the list once up to `stop`, returning the index of the first match at or after `start`. Negative `start`/`stop` values are normalized from the end of the list. Raises `ValueError` if not found.


soraros:
I think it would be better not to provide this API at all, since this is a linked list, where index-based access is not really idiomatic.

msaelices:
As per @soraros 's comment. closing the PR

PR #6104, opened 2026-03-06
[Stdlib] Add assert_in() and assert_not_in() to testing module

msaelices:
Adds `assert_in(value, container)` and `assert_not_in(value, container)` for richer membership assertions.

`assert_true(x in collection)` only shows `False` on failure. The new helpers show the value and container:
```
AssertionError: `value in container` assertion failed
  value: 99
  container: [1, 2, 3]
```

### DIFF
--- a/mojo/stdlib/std/testing/testing.mojo
+++ b/mojo/stdlib/std/testing/testing.mojo
@@ -577,8 +577,10 @@ fn assert_in[
         raise _assert_error(
             String(
                 "`value in container` assertion failed",
-                "\n  value: ", value,
-                "\n  container: ", container,
+                "\n  value: ",
+                value,
+                "\n  container: ",
+                container,
                 ("\n  reason: " + msg) if msg else "",
             ),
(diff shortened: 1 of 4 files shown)


mojo/stdlib/std/testing/testing.mojo
```diff
@@ -540,6 +540,99 @@ fn _assert_cmp_error[
     return _assert_error(err, loc)
 
 
+@always_inline
+fn assert_in[
```
NathanSWard, line 544: I'm not a fan of these function as you can easily write `assert_true/false(value in collection, "message")` and achieve the exact same things.

Plus ideally these would be generic over a trait that encapsulates `__in__` so we don't have to hard-code for every collection type that implement `__in__`.

I think PR would be okay to close as I don't see the benefit of it over the already available syntax & method calls.

NathanSWard (changes requested):

msaelices:
Closed as per @NathanSWard 's request.

PR #6105, opened 2026-03-06
[Stdlib] Add Report.percentile() to benchmark module

msaelices:
Adds `percentile(p, unit)` to `benchmark.Report` for measuring tail latency.

## Summary

- Collects significant batch means, sorts them, and returns the p-th percentile via linear interpolation.
- Returns 0 for an empty report; clamps `p <= 0` to min and `p >= 100` to max.
- 6 new unit tests covering empty, single-batch, odd/even count, quartiles, clamping, and non-significant batch exclusion.

---

Assisted-by: AI


abduld:
!sync

PR #6106, opened 2026-03-06
[Stdlib] Fix Pipe FD leak when set_close_on_exec fails

msaelices:
If `_set_close_on_exec()` fails after `pipe()` succeeds, both raw FDs were leaked because `self.fd_in`/`fd_out` are not yet initialized, so `__del__` cannot close them. Close both FDs explicitly before raising.


JoeLoser (approved):

JoeLoser:
!sync

PR #6107, opened 2026-03-06
[Stdlib] Add SIMD.first_true() and SIMD.count_true() for bool masks

msaelices:
Adds two ergonomic methods to `SIMD[DType.bool, N]` that replace the verbose `pack_bits()` + `count_trailing_zeros()` pattern:

- `first_true()` — index of the first `True` lane, or `-1` if all are `False`
- `count_true()` — number of `True` lanes (bool-specific alias for `reduce_bit_count()`)

Uses `_memchr_impl` in `string_slice.mojo` as a concrete simplification example.

---

Assisted-by: AI


soraros:
I've always wanted something like this. @martinvuyk, I think you'd also find this interesting.

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -3009,6 +3009,56 @@ struct SIMD[dtype: DType, size: Int](
...
+        comptime assert (
+            Self.dtype == DType.bool
+        ), "first_true requires SIMD[DType.bool, N]"
+        comptime if Self.size == 1:
+            return 0 if self[0] else -1
+        else:
+            var mask = pack_bits(rebind[SIMD[DType.bool, Self.size]](self))
+            if not mask:
+                return -1
+            return Int(count_trailing_zeros(mask))
```
martinvuyk: We can use some tricks here `from bit._mask import is_negative`. I'm not sure if it's necessarily faster
```suggestion
        comptime if Self.size == 1:
            return Int(self[0]) - 1
        else:
            var mask = pack_bits(rebind[SIMD[DType.bool, Self.size]](self))
            return Int(count_trailing_zeros(mask)) | is_negative(mask - 1)
```

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -2701,11 +2701,9 @@ fn _memchr_impl[
...
     for i in range(0, vectorized_end, bool_mask_width):
         var bool_mask = haystack.load[width=bool_mask_width](i).eq(first_needle)
-        var mask = pack_bits(bool_mask)
-        if mask:
-            output = haystack + Int(
-                type_of(mask)(i) + count_trailing_zeros(mask)
-            )
+        var pos = bool_mask.first_true()
+        if pos >= 0:
+            output = haystack + i + pos
```
martinvuyk: I'm not 100% sure whether taking the count_trailing_zeros out into the main body of the loop is faster or slower, we need to benchmark this

mojo/stdlib/std/builtin/simd.mojo
martinvuyk: I think we have to have a branch according to endianness. We are assuming little-endian

martinvuyk (commented):
Hi, good idea :)

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -3009,6 +3009,56 @@ struct SIMD[dtype: DType, size: Int](
...
+
```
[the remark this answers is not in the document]
msaelices: Fixed: added @always_inline to count_true.

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -3009,6 +3009,56 @@ struct SIMD[dtype: DType, size: Int](
         else:
             return Int(pop_count(self).reduce_add())
 
+    fn first_true(self) -> Int:
```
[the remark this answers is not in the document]
msaelices: Fixed: added @always_inline to both first_true and count_true.

mojo/stdlib/std/builtin/simd.mojo
msaelices: Thanks for the suggestion. Since you noted the speed is uncertain I'll leave the branch-free optimization as a potential follow-up.

msaelices: Good catch. Fixed: added a comptime if is_big_endian() branch that uses count_leading_zeros on big-endian platforms.

samufi:
Maybe it would be a good idea to add the option to pass a default for non-existent indices explicitly. In some cases this could save the effort of checking for -1 explicitly. For example, I am thinking about a case where I want to get the last index where `arr < max_val`. Here, it would be more natural to default to `arr.size`.

PR #6108, opened 2026-03-06
[Stdlib] Add overflow assertion to IndexList.product()

msaelices:
Adds a runtime `assert` to `product()` in `utils/index.mojo` that detects integer overflow before multiplication, instead of silently wrapping. The check uses `Int.MAX // dim` to verify the running product fits.

Also adds tests for basic usage, zero dimensions, partial ranges, and the `end_idx` parameter.

---

Assisted-by: AI


soraros:
Since it's an index list, I wonder whether it's likely to actually exceeding `Int.MAX`.

msaelices:
> Since it's an index list, I wonder whether it's likely to actually exceeding `Int.MAX`.

Not sure if I get it. The threshold is `Int.MAX // dim`. BTW, I agree that it's not an extremely useful assertion, just to prevent overflow.

PR #6109, opened 2026-03-07
[Stdlib] Avoid String allocation in stat_result.write_to() for timestamps

msaelices:
`write_to()` was calling `String(self.st_atimespec)` for each of the four timestamp fields, allocating a heap string only to immediately write and discard it. `_CTimeSpec` already implements `Writable`, so the values can be passed directly to the writer.


NathanSWard (approved):
thanks!

NathanSWard:
!sync

PR #6110, opened 2026-03-07
[Stdlib] Avoid temporary Set allocation in Set.intersection_update()

msaelices:
`intersection_update()` previously implemented as `self.difference_update(self - other)`, which allocates a full temporary `Set` (a hash table) just to hold the keys to remove.

Replace with a two-pass approach that collects removals into a `List`, then calls `discard()` on each, eliminating the temporary hash table allocation while keeping iteration and mutation safely separated.

Also adds a `bench_set_intersection_update` benchmark using `iter_preproc` to reset the set between destructive iterations.

## Benchmarks

50% overlap between sets, median of 10 runs (time in ms, lower is better):

| Size | Before | After | Speedup |
|------|--------|-------|---------|
| 10 | 0.000297 | 0.000166 | 1.8x |
| 100 | 0.001378 | 0.000804 | 1.7x |
| 1000 | 0.014780 | 0.005545 | 2.7x |
| 10000 | 0.143700 | 0.065400 | 2.2x |

Consistent ~2x speedup across all sizes. The gain comes from eliminating the temporary `Set` hash table — the old code built a full hash table just to iterate and discard it, while the new code uses a plain `List` with much lower allocation and insertion overhead.

---

Assisted-by: AI


JoeLoser (approved):

JoeLoser:
!sync

PR #6112, opened 2026-03-07
[Kernels] Update CLAUDE.md import example to use TileTensor

msaelices:
## Summary

Update the canonical import example in `kernels/CLAUDE.md` to use `TileTensor` instead of the deprecated `LayoutTensor`.

`LayoutTensor` is being replaced by `TileTensor` across the kernels codebase. `CLAUDE.md` is used by contributors and AI coding assistants as the reference for how to write kernel code, so having the wrong type here causes new code to be written against the deprecated API.


abduld:
!sync

PR #6114, opened 2026-03-07
[stdlib] Add `Comparable` conformance to `Tuple`

msaelices:
`Tuple` already had lexicographic `__lt__`, `__le__`, `__gt__`, `__ge__`, `__eq__`, `__ne__` via parametric overloads, but the struct did not declare `Comparable`, making it unusable in generic algorithms that require it.

This PR adds conditional `Comparable` conformance using the same pattern as `Writable where AllWritable[*element_types]` introduced in upstream:

```mojo
struct Tuple[*element_types: Movable](
    Comparable where AllComparable[*element_types],
    ...
):
```

It also adds `AllComparable` to `reflection/traits.mojo` following the existing `AllWritable`/`AllEquatable` pattern.

## Why add new overloads alongside the existing ones?

The existing parametric overloads have the signature:

```mojo
fn __lt__[self_elt_types: ..., other_elt_types: ...](
    self: Tuple[*self_elt_types], other: Tuple[*other_elt_types]
) -> Bool
```

These are not equivalent to `Comparable` conformance. The `Comparable` trait requires:

```mojo
fn __lt__(self, other: Self) -> Bool
```

The two overloads serve different purposes and are **complementary, not redundant**:

| Overload | When used |
|---|---|
| Parametric | Direct comparison between tuples of **different lengths**: `(1,) < (1, 2)` |
| Conforming (`Self`) | Trait dispatch in generic code: `fn f[T: Comparable](a: T, b: T)` called with tuples |

The `Self`-typed overloads use `where AllComparable[*element_types]` — the compile-time conditional conformance enforces the element constraint without any runtime helper.

## What this enables

```mojo
fn min_val[T: Comparable & ImplicitlyCopyable](a: T, b: T) -> T:
    return a if a < b else b

min_val((1, 2), (1, 3))  # Now works — previously failed to compile
```

```mojo
fn clamp[T: Comparable](value: T, lo: T, hi: T) -> T:
    if value < lo: return lo
    if value > hi: return hi
    return value

clamp((3, 5), (1, 0), (5, 9))  # Now works
```

Tuples with non-`Comparable` elements are still valid — they simply don't conform to `Comparable` (conditional conformance).

---

Assisted-by: AI


mojo/stdlib/test/builtin/test_tuple.mojo
```diff
@@ -310,9 +310,20 @@ def test_tuple_assert_not_equal() raises:
...
+
+def test_tuple_comparable_trait_conformance() raises:
+    """Verify Tuple conforms to Comparable so it can be used as a Comparable."""
+
+    fn min_comparable[T: Comparable & ImplicitlyCopyable](a: T, b: T) -> T:
+        return a if a < b else b
+
+    assert_equal(min_comparable((1, 2), (1, 3)), (1, 2))
+    assert_equal(min_comparable((2, 0), (1, 9)), (1, 9))
+    assert_equal(min_comparable((1, 1), (1, 1)), (1, 1))
```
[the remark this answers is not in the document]
msaelices: Fixed, moved the test before main() in the latest commit.

mojo/stdlib/test/builtin/test_tuple.mojo
```diff
@@ -310,9 +310,20 @@ def test_tuple_assert_not_equal() raises:
 
 
 def test_tuple_assert_equal_failure_message() raises:
-    with assert_raises(contains="Tuple[Int, Int](Int(1), Int(2))"):
+    with assert_raises(contains="(1, 2)"):
         assert_equal((1, 2), (1, 3))
```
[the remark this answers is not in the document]
msaelices: The format change from repr to str is intentional, as once Tuple conforms to `Equatable & Writable`, the generic `assert_equal` overload wins, and `String(tuple)` produces the more readable `(1, 2)` format. The richer repr is still available via `write_repr_to`.

mojo/stdlib/std/builtin/tuple.mojo
```diff
@@ -244,6 +244,126 @@ struct Tuple[*element_types: Movable](ImplicitlyCopyable, Sized, Writable):
...
+            All element types must conform to `Comparable`.
+
+        Args:
+            other: The other tuple to compare against.
+
+        Returns:
+            True if this tuple is lexicographically >= the other, False otherwise.
+        """
+        return not self.__lt__(other)
+
```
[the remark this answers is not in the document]
msaelices: The `_compare` method requires `Variadic.TypesOfTrait[Movable & Comparable]` element types, but the conforming `__lt__(self, other: Self)` operates on `Tuple[*element_types: Movable]`, the static type does not carry the `Comparable` constraint. Delegating would require the compiler to narrow the type from the comptime assertion, which it does not do. I believe the `trait_downcast` approach is the correct pattern here (same as how `write_to` uses `trait_downcast` after `constrained_conforms_to_writable`). The constraint-check helpers are now extracted into `_constrained_elements_equatable`/`_comparable` (analogous to `constrained_conforms_to_writable`) in the latest commits.

mojo/stdlib/std/builtin/tuple.mojo
```diff
@@ -264,6 +266,106 @@ struct Tuple[*element_types: Movable](
...
             UnsafePointer(to=self[i]).init_pointee_move(elt_types[i]())
 
+    # ===-------------------------------------------------------------------===#
+    # Comparable / Equatable trait conformance (other: Self overloads)
+    # ===-------------------------------------------------------------------===#
+
+    @always_inline
+    def __eq__(
+        self, other: Self
+    ) -> Bool where AllComparable[*Self.element_types]:
```
JoeLoser: This `__eq__` duplicates the existing `__eq__(self, other: Self) where AllEquatable` that already exists below (at the old line 268, now ~line 370). Since `Comparable` extends `Equatable`, any type satisfying `AllComparable` also satisfies `AllEquatable`, so both overloads would be valid simultaneously for `Comparable` elements.

The `Comparable` trait inherits `__eq__`/`__ne__` from `Equatable`, and the existing `Equatable where AllEquatable` conformance already satisfies the equality requirements. These new `__eq__` and `__ne__` appear redundant and may cause overload ambiguity.

Same applies to the `__ne__` below (line 295).

If the compiler *requires* these to satisfy `Comparable` conformance specifically, that's worth noting in a comment explaining why.

mojo/stdlib/std/builtin/tuple.mojo
```diff
@@ -264,6 +266,106 @@ struct Tuple[*element_types: Movable](
...
+        """Compare this tuple to another tuple of the same type for inequality.
+
+        Args:
+            other: The other tuple to compare against.
+
+        Returns:
+            True if this tuple is not equal to the other tuple, False otherwise.
+        """
+        return not self.__eq__(other)
```
JoeLoser: **Suggestion** Can we say `return not self == other` for consistency?

mojo/stdlib/std/builtin/tuple.mojo
```diff
@@ -264,6 +266,106 @@ struct Tuple[*element_types: Movable](
...
+
+    @always_inline
+    def __lt__(
+        self, other: Self
+    ) -> Bool where AllComparable[*Self.element_types]:
```
JoeLoser, line 265: This re-implements the same lexicographic comparison logic that already exists in `_compare()` (around line ~526 in the new file). Could this delegate instead?

```mojo
def __lt__(self, other: Self) -> Bool where AllComparable[*Self.element_types]:
    return self._compare(other) < 0
```

mojo/stdlib/std/builtin/tuple.mojo
```diff
@@ -264,6 +266,106 @@ struct Tuple[*element_types: Movable](
...
+            if lhs < rhs:
+                return True
+            if rhs < lhs:
+                return False
+        return False  # equal
+
+    @always_inline
+    def __le__(
+        self, other: Self
+    ) -> Bool where AllComparable[*Self.element_types]:
```
JoeLoser, line 286: `Comparable` already provides default implementations for `__le__`, `__gt__`, and `__ge__` in terms of `__lt__`, and these implementations are identical to those defaults. Are they needed to satisfy the conditional conformance, or can they be dropped?

mojo/stdlib/test/builtin/test_tuple.mojo
```diff
@@ -354,5 +354,16 @@ def test_tuple_assert_equal_failure_message() raises:
...
 
 
+def test_tuple_comparable_trait_conformance() raises:
```
JoeLoser, line 377: **Suggestion** Can we audit existing test coverage and consider adding more tests if they don't exist? Some ideas include:
- Tuples with >2 elements or heterogeneous-but-comparable types (e.g. `Tuple[Int, String]`)
- Single-element tuple edge case
- A compile-time negative test: tuple with non-`Comparable` elements should NOT satisfy `Comparable`

JoeLoser (commented):
Generally looks good, I'd like to see my comments addressed before I approve and sync this in though.

msaelices:
Superseded: upstream added `Comparable` conformance to `Tuple` (`__lt__`/`__le__`/`__gt__`/`__ge__` gated on `AllComparable`). Closing as already implemented upstream.

PR #6116, opened 2026-03-07
[stdlib] Add `Hashable` conformance to `ArcPointer`

msaelices:
`ArcPointer` already implements `Identifiable` (pointer-identity semantics via `__is__`/`__isnot__`), but had no `__hash__()` or `__eq__()`. Without them, `ArcPointer` values cannot be stored in `Dict` or `Set`.

`Equatable` and `Hashable` are both based on the inner allocation address, consistent with identity: two `ArcPointer` instances pointing to the same object (`a is b`) are equal and hash identically.

```mojo
var p = ArcPointer(42)
var q = p          # same allocation
print(p == q)      # True
print(hash(p) == hash(q))  # True

var d = Dict[ArcPointer[Int], String]()
d[p] = "hello"
print(d[q])        # "hello"
```

---

Assisted-by: AI

### DIFF
--- a/mojo/stdlib/std/memory/arc_pointer.mojo
+++ b/mojo/stdlib/std/memory/arc_pointer.mojo
@@ -282,20 +282,7 @@ struct ArcPointer[T: Movable & ImplicitlyDestructible](
             True if the two `ArcPointer` instances point at the same object and
             False otherwise.
         """
-        return self._inner == rhs._inner
-
-    def __ne__(self, rhs: Self) -> Bool:
-        """Returns True if the two `ArcPointer` instances point at different
-        objects.
-
-        Args:
-            rhs: The other `ArcPointer`.
-
-        Returns:
-            True if the two `ArcPointer` instances point at different objects
-            and False otherwise.
-        """
-        return self._inner != rhs._inner
+        return self is rhs
 
     def __hash__[H: Hasher](self, mut hasher: H):
         """Hash this pointer by its address.
--- a/mojo/stdlib/std/memory/arc_pointer.mojo
+++ b/mojo/stdlib/std/memory/arc_pointer.mojo
@@ -70,6 +70,7 @@ struct _ArcPointerInner[T: Movable & ImplicitlyDestructible]:
 
 
 struct ArcPointer[T: Movable & ImplicitlyDestructible](
+    Equatable,
     Hashable,
     Identifiable,
     ImplicitlyCopyable,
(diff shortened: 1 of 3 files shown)


msaelices:
@JoeLoser FYI, since you're working on stdlib type conformances

soraros:
Are explicit function bodies required, or are the default implementations sufficient?

mojo/stdlib/std/memory/arc_pointer.mojo
```diff
@@ -265,6 +268,49 @@ struct ArcPointer[T: Movable & ImplicitlyDestructible](
...
+        Args:
+            rhs: The other `ArcPointer`.
+
+        Returns:
+            True if the two `ArcPointer` instances point at the same object and
+            False otherwise.
+        """
+        return self._inner == rhs._inner
+
+    def __ne__(self, rhs: Self) -> Bool:
```
JoeLoser: **Question** Does the default affordance synthesized by the `Eqautable` trait work such that we don't need to define it explicitly here? It would still compare the `_inner` fields.

JoeLoser (approved):
LGTM, but please look at the `__ne__` comment I made before I sync this in. Thanks!

mojo/stdlib/std/memory/arc_pointer.mojo
```diff
@@ -265,6 +268,49 @@ struct ArcPointer[T: Movable & ImplicitlyDestructible](
...
+        heap allocation, consistent with `__hash__` and `__is__`.
+
+        Args:
+            rhs: The other `ArcPointer`.
+
+        Returns:
+            True if the two `ArcPointer` instances point at the same object and
+            False otherwise.
+        """
+        return self._inner == rhs._inner
```
JoeLoser: **Question** Should we delegate to `is` so this is explicit, e.g. `return self is rhs`?

soraros: What's intended use of `is` these days? I thought we want it for all pointer like types, but `UnsafePointer` doesn't have a `__is__`.

msaelices:
@JoeLoser @soraros: `UnsafePointer` doesn't conform to `Equatable`, the default fieldwise synthesis would fail to compile, so an explicit `__eq__` is required. Updated it to `return self is rhs` to make the pointer-identity semantics explicit. Also removed the explicit `__ne__` since `Equatable` already provides a default `not self == other`.

JoeLoser:
!sync

JoeLoser:
!sync

PR #6120, opened 2026-03-07
[stdlib] Add `bit_mask()` to the `bit` module

msaelices:
The `bit` module provides `pop_count()`, `bit_width()`, `count_leading_zeros()`, etc. but had no function to generate a mask for a contiguous range of bits. This requires users to write `(1 << end) - (1 << start)` everywhere, with no bounds checking.

`bit_mask[dtype](start, end)` returns a `Scalar[dtype]` with bits in the half-open range `[start, end)` set to 1. It asserts that `start >= 0`, `start < end`, and `end <= bitwidth(dtype)`.

The edge case where `end == bitwidth(dtype)` is handled with `~Scalar[dtype](0) << start` to avoid left-shifting by the full bit width (which is undefined behavior).

**Where this is useful:**

- **Hardware/register fields**: isolate a known-width field at a known bit offset, e.g. `value & bit_mask[DType.uint32](5, 11)` to extract bits 5–10 of a register word.
- **Protocol parsing**: when a spec says "bits 12–15 are the opcode", `bit_mask[DType.uint16](12, 16)` is clearer and less error-prone than the magic constant `0xF000`.
- **Bit field manipulation**: set or clear a contiguous range without hardcoding a literal that must be re-derived when the field position changes.

Note: this only covers *contiguous* ranges. Non-contiguous patterns (e.g. `0x55` = 01010101) have no regular structure and are best expressed as literals or composed from bitwise operations.

---

Assisted-by: AI


martinvuyk:
I think this should be named something like `ones()` and be inside the `bit._mask` module, not in the bit one.

msaelices:
> I think this should be named something like `ones()` and be inside the `bit._mask` module, not in the bit one.

Thanks! done: https://github.com/msaelices/modular/commit/c79c54de9371be253921f032d3c64a022a4b062b

PR #6135, opened 2026-03-10
[MAX] Add a module v2 FLUX.2 architecture, pipeline, and example path

pei0033:
## Summary

This PR adds a new module v2 FLUX.2 architecture alongside the existing `flux2_modulev3` path.

The scope is intentionally focused on three areas:
- the FLUX.2 DiT implementation
- the FLUX.2 diffusion pipeline
- the offline example used to exercise the new path

The overall model behavior is kept aligned with the existing `module_v3` implementation where possible, while the code is migrated onto module v2 primitives.

### DiT

- Adds a new `max.pipelines.architectures.flux2` architecture and registers it as `Flux2Pipeline`.
- Introduces a module v2 FLUX.2 transformer implementation built on `max.nn` and `max.graph.ops` instead of `max.experimental.*`.
- Keeps the v2 implementation structurally close to the existing `flux2_modulev3` logic where possible:
  - Adds a local FLUX2-specific `LayerNorm` compatibility implementation so the v2 code can preserve the `module_v3` `elementwise_affine` / `use_bias` behavior.

### Pipeline

- Adds a new module v2 `Flux2Pipeline` implementation.
- The new pipeline runs on `Buffer` / graph-based execution rather than the `module_v3` eager `Tensor` path.
- Pipeline helper stages are compiled explicitly and invoked as graph callables.
- Updates the shared diffusion pipeline loader so `session` is passed only to components that require it, which is needed for graph-loaded v2 components.
- Adds `float32_array_to_buffer()` to materialize device buffers directly from host float32 data, including bf16 buffer creation without requiring an extra cast-only graph.

### Example

- Updates the offline diffusion example so FLUX.2 can run through the new module v2 path.
- Adds a new `--prefer-module-v3` flag to the example CLI.

## Validation

- `MODULAR_DEVICE_CONTEXT_SYNC_MODE=true ./bazelw run //max/examples/diffusion:simple_offline_generation -- --model black-forest-labs/FLUX.2-dev --prompt "A cat in a garden"`
- For comparison, the same example can be run with `--prefer-module-v3`.

## TODO
- [ ] AutoEncoder Migration
- [ ] TextEncoder Migration
- [ ] FLUX.2-Klein Migration

## Checklist

      sequence of smaller PRs
---
```
Assisted-by: Codex
```


katelyncaldwell (approved):

katelyncaldwell:
!sync

pei0033:
This landed in 6d5b1c1

PR #6136, opened 2026-03-10
[MAX] Add shift_terminal support for Qwen-Image

jglee-sqbits:
### Overview

This PR adds `shift_terminal` support required for Qwen-Image scheduler behavior in `FlowMatchEulerDiscreteScheduler`.

It introduces an optional terminal shifting path while preserving the existing scheduler behavior for current users when `shift_terminal` is unset.

### What’s included

* Optional `shift_terminal` support in `FlowMatchEulerDiscreteScheduler`.
* Preservation of the current scheduler behavior when `shift_terminal` is unset.
* Qwen-Image-compatible terminal shifting behavior without affecting other pipelines.

### NOTES

* This PR only adds the scheduler support needed by Qwen-Image.
* It does not introduce the Qwen-Image model or pipeline by itself.


ansley:
!sync

PR #6138, opened 2026-03-10
[MAX] Add Qwen2.5-VL encoder for Qwen-Image

jglee-sqbits:
### Overview

This PR adds the shared Qwen2.5-VL encoder required for the Qwen-Image pipelines.

It introduces a reviewable encoder unit on the module-v2 path with `Buffer`-based interfaces, while keeping edit-only multimodal helpers out of this branch.

### What’s included

* Shared Qwen2.5-VL encoder for the Qwen-Image pipelines.
* Module-v2 based encoder path with `Buffer`-based interfaces.
* Separation of edit-only multimodal helpers so the encoder remains reviewable on its own.

### NOTES

* This PR contains the shared encoder needed by the Qwen-Image family of pipelines.
* Edit-specific multimodal handling is introduced separately.


ansley:
!sync

PR #6132, opened 2026-03-10
[MAX] Add First-Block Cache Implementation for FLUX.2 Pipeline

byungchul-sqzb:
## Summary

This PR adds FBC support to the FLUX.2 pipeline.

At a high level, the cache decision is made from the residual of the first transformer block:
- compute the first-block residual for the current denoising step
- compare it against the previous step's first-block residual
- if the relative difference is below `rdt`, reuse the previous model output
- otherwise, continue through the normal FLUX.2 transformer path

This PR wires that behavior through the FLUX.2 stack:
- [`flux2.py`](/root/modular/max/python/max/pipelines/architectures/flux2_modulev3/flux2.py)
  adds step-cache inputs (`prev_residual`, `prev_output`, `cache_enabled`, `rdt`) and uses `F.cond(...)` to choose between cache reuse and the normal miss path
- [`pipeline_flux2.py`](/root/modular/max/python/max/pipelines/architectures/flux2_modulev3/pipeline_flux2.py)
  manages the runtime cache state across denoising steps and threads the cache tensors into the transformer
- [`model.py`](/root/modular/max/python/max/pipelines/architectures/flux2_modulev3/model.py)
  updates the transformer wrapper so the compiled model can accept the expanded input signature
- ~~[`functional.py`](/root/modular/max/python/max/experimental/functional.py)~~
~~exposes `F.cond(...)` for graph-level conditional branching~~ (added 1ce7b57a4cac057803a54bcaa023f114eb4f7b3d)
- [`simple_offline_generation.py`](/root/modular/max/examples/diffusion/simple_offline_generation.py)
  adds CLI support for `--step-cache` and `--rdt`

This PR also includes the FLUX.2-specific RoPE changes needed to make the step-cache graph compile correctly:
- [`flux2_attention.py`](/root/modular/max/python/max/pipelines/architectures/flux2_modulev3/layers/flux2_attention.py)
  externalizes `position_ids` construction instead of creating them inside the attention RoPE helper
- [`kernels.py`](/root/modular/max/python/max/nn/kernels.py)
  updates the `rope_ragged_with_position_ids()` path used by FLUX.2 so it is compatible with the step-cache graph structure

The important behavioral change is that FLUX.2 now has an explicit cache-hit / cache-miss path, with cache reuse driven by the relative change in the first-block residual rather than by a separate scheduler-level heuristic.

## Results
- Baseline (w/o --step-cache) 
output

```
Generating image for prompt: 'dog dancing near the sun'
[...]

PR #6150, opened 2026-03-12
[MAX] Fix FLUX2 Klein pipeline to always use the standard transformer path

pei0033:
## Summary

PR #6132 introduced step cache support for FLUX2. Although FLUX2 Klein uses the same transformer, that logic was not applied to the FLUX2 Klein pipeline.

FLUX2 Klein typically runs with a very small number of denoising steps by default (4), so first-block (FB) cache does not provide meaningful benefit. This PR makes that explicit by always disabling step cache for FLUX2 Klein and forcing the pipeline to use the standard transformer path.

## Testing

Previously, running:

```
./bazelw run //max/examples/diffusion:simple_offline_generation -- --model "black-forest-labs/FLUX.2-klein-4B" --prompt 'A beautiful woman in a red dress walking down a streetn' --num-inference-steps 4 --guidance-scale 1.0 --seed 42
```
could fail with:
```
ERROR: Flux2 transformer model is not ready yet. Call use_standard_model or use_step_cache_model first.
```

After this change, the same command runs successfully.

CC. @byungchul-sqzb 

## Checklist

      sequence of smaller PRs


pei0033:
It is no longer needed due to commit 0223052

PR #6163, opened 2026-03-14
[stdlib] Add barrier_count() GPU synchronization primitive

### LINKED ISSUE #6051: [Feature Request] Add __syncthreads_count() equivalent
### Review Mojo's priorities

### What is your request?

I noticed that there is no `__syncthreads_count()` equivalent.

This is far from my specialty but I think that there is some performance to gain by adding this operation where needed.

This operation is used in some applications liek gaussian splatting. One can reproduce this with 2 barriers and an atomic add but that wont be resolved into a single instruction in PTX `BAR.RED.POPC`.

### Any other details?

_No response_

mahendrarathore1742:
- Implement barrier_count(predicate) mapping to NVVM bar.red.popc
- Mirrors CUDA __syncthreads_count() for block-wide predicate counting
- NVIDIA GPU only; compile-time error on other targets
- Export via std.gpu.sync and std.gpu packages
- Add comprehensive GPU tests (half/zero/all predicates)
- Document barrier_count alongside barrier() in GPU manual

close #6051


abduld:
!sync

abduld:
Getting 

```
oss/modular/mojo/stdlib/test/gpu/sync/test_barrier_count.mojo:14:21: error: package 'gpu' does not contain 'DType'
from std.gpu import DType, thread_idx
                    ^
oss/modular/mojo/stdlib/test/gpu/sync/test_barrier_count.mojo:24:5: error: use of unknown declaration 'let'
    let predicate = thread_idx.x < UInt(active)
    ^~~
oss/modular/mojo/stdlib/test/gpu/sync/test_barrier_count.mojo:24:9: error: statements must start at the beginning of a line
    let predicate = thread_idx.x < UInt(active)
        ^
```

as errors

mahendrarathore1742:
Thanks for the report! After investigating, the failures were caused by two issues we introduced:

1. **`barrier_count` fix** — The NVVM `popc` result is an MLIR `i32`. We initially tried to wrap it with `Int32(...)`, but that doesn't match any constructor, causing the build/doc gen error. The correct fix is to use `rebind[Int32](...)` instead.

2. **`test_barrier_count.mojo` fix** — Two problems in the new test file:
   - `DType` was imported from `std.gpu`, but it isn't exported there — it's already available via the prelude, so the import can be dropped entirely.
   - `let predicate` was used, which the parser rejected. Fixed by changing it to `var predicate`.

Both fixes are now pushed. Let me know if anything else needs attention!

abduld:
!sync

abduld:
still getting errors. Looks like oss/modular/mojo/stdlib/test/gpu/sync/test_barrier_count.mojo is missing a main function. Please make sure the tests run and are correct and after confirmed I can sync again. Thanks

mahendrarathore1742:
Fixed main() using TestSuite.discover_tests, replaced comptime with alias, added has_nvidia_gpu_accelerator() skip guard. Verified locally — compiles cleanly and skips correctly on non-NVIDIA hardware. Ready for !sync.

Screenshot from 2026-03-16 23-34-50

mojo/stdlib/test/gpu/sync/test_barrier_count.mojo
```diff
@@ -0,0 +1,84 @@
...
+
+
+def test_barrier_count_extremes() raises:
+    if not has_nvidia_gpu_accelerator():
+        print("SKIP: test_barrier_count_extremes requires NVIDIA GPU")
+        return
+
+    alias BLOCK = WARP_SIZE * 2
+
+    with DeviceContext() as ctx:
```
abduld, line 57: please create DeviceContext once and not once per test

mojo/stdlib/test/gpu/sync/test_barrier_count.mojo
```diff
@@ -0,0 +1,84 @@
...
+    output[thread_idx.x] = barrier_count(predicate)
+
+
+def test_barrier_count_half() raises:
+    if not has_nvidia_gpu_accelerator():
+        print("SKIP: test_barrier_count_half requires NVIDIA GPU")
+        return
+
+    alias BLOCK = WARP_SIZE * 2
+    alias ACTIVE = BLOCK // 2
```
abduld, line 35: replace all these with comptime

.bazelversion
```diff
@@ -1,2 +1,3 @@
 buildbuddy-io/5.0.268
 9.0.1
+
```
abduld, line 3: drop these changes

PR #6167, opened 2026-03-14
[Stdlib] Optimize `Counter.most_common(n)` with heap-based partial sort for small n

msaelices:
When `n < total // 2`, use a min-heap of size `n` to find the top-n elements in O(total · log n) instead of sorting all elements in O(total · log total).

The full sort path is kept for `n >= total // 2` and for the `n is None` case.

Adds private helpers `_heap_sift_up`, `_heap_sift_down`, `_most_common_heap` and a test exercising the heap code path.

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_counter.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_counter.mojo
@@ -37,7 +37,7 @@ def bench_most_common_small_n[total: Int, n: Int](mut b: Bencher) raises:
 
     @always_inline
     def call_fn() unified {read}:
-        var result = black_box(c).most_common(black_box(n))
+        var result = black_box(c).most_common(UInt(black_box(n)))
         keep(result)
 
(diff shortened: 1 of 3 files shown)


msaelices:
Closing after benchmarking showed the heap optimization does not deliver the expected speedup.

Benchmarks across counter sizes (100, 1000, 10000 elements) consistently show the heap path is slower than the full sort path at every tested n value:

- total=1000, n=5: heap 8.75µs vs sort 4.9µs (1.8x slower)
- total=1000, n=10: heap 10.8µs vs sort 4.9µs (2.2x slower)
- total=10000, n=10: heap 110µs vs sort 65µs (1.7x slower)

The SIMD-optimized sort processes all elements in a single linear cache-friendly pass. The heap's sift operations involve pointer chasing and List allocation overhead that erases the theoretical O(total · log n) advantage. The optimization adds ~120 lines of complexity for no measurable benefit.

PR #6168, opened 2026-03-14
[Stdlib] Add partition() and rpartition() to String and StringSlice

msaelices:
Add Python-compatible `partition()` and `rpartition()` to `StringSlice` and `String`.

- `partition(sep)` — splits on the first occurrence; returns `(before, sep, after)` or `(self, "", "")` if not found.
- `rpartition(sep)` — splits on the last occurrence; returns `(before, sep, after)` or `("", "", self)` if not found.

`String` delegates to `StringSlice`. Tests added to both `test_string.mojo` and `test_string_slice.mojo`.

---

Tracking issue: #6176. Keeping as draft until consensus is reached there.

---

Assisted-by: AI


NathanSWard:
@msaelices - can we hold off on adding capabilities to existing types without proper reasoning or valid usage?
We don't just want to be copying Python for Python's sake. Often times there can be better mojo first-class alternatives to these sort of functions/utilities.
And often times we don't actually need/want to support everything that python does.

For new methods/APIs like this lets try to track these in a GitHub issue instead to see if there is valid need for them and then after we come to agreement we can review the implementation.

msaelices:
Opened issue #6176 to track the justification for this API as requested. Converted PR to draft

PR #6171, opened 2026-03-14
[Stdlib] Remove misleading TODO in `LinkedList.insert()` index normalization

msaelices:
`LinkedList.insert()` had a `# TODO: use normalize_index` comment, but `normalize_index` asserts bounds and cannot implement Python's `list.insert()` clamping semantics, where out-of-range negative indices silently clamp to 0 (head) rather than raising. Replacing the comment with one explaining the intentional difference.

---

Assisted-by: AI


JoeLoser (approved):

JoeLoser:
!sync

PR #6172, opened 2026-03-14
[Stdlib] Add `FileHandle.tell()` to query current file position

### LINKED ISSUE #6186: [Stdlib] Add FileHandle.tell() to query current file position
`FileHandle` already has `seek()` to move to a position, but no way to query the current position. Python's file objects have `.tell()` for this. The current workaround requires manually tracking the byte offset, which is error-prone.

**Use cases:**
- Save and restore a file position (save with \`tell()\`, restore with \`seek()\`)
- Track read progress without an external counter
- Implement parsers that need to backtrack

**Proposed API:**  `def tell(self) raises -> UInt64`

`seek()` already exists and returns the new position; `tell()` is a natural complement. Implementation is a single call to `seek(0, SEEK_CUR)`.

Draft PR: #6172

msaelices:
`FileHandle` had `seek()` but no `tell()` to query the current byte offset. Adds `tell()` as a thin wrapper around `lseek(fd, 0, SEEK_CUR)`, consistent with Python's `file.tell()`.

This fixes https://github.com/modular/modular/issues/6186

Assisted-by: AI


JoeLoser (approved):
LGTM but please add a changelog entry and update the PR description to say this fixes https://github.com/modular/modular/issues/6186. Thanks, @msaelices!

JoeLoser:
Did you explore a more descriptive name like `stream_position` or something? I assume the `tell()` name was chosen because that is what Python calls it?

msaelices:
> Did you explore a more descriptive name like `stream_position` or something? I assume the `tell()` name was chosen because that is what Python calls it?

Exactly. I come from Python, and IMO, if there's a strong argument for it, I would stick with Python's naming conventions.

Feel free to request me to change it for a better name

msaelices:
@JoeLoser rebased onto latest `main`; the changelog conflict is resolved and it's mergeable. Changelog entry is in and the description links #6186 as requested. Ready for `!sync` whenever you have a moment. Thanks!

PR #6173, opened 2026-03-14
[Stdlib] Add `math.prod()` for computing product of a span

msaelices:
Adds `math.prod()` as a Python-compatible function that computes the product of all elements in a `Span[Scalar[dtype]]`, with an optional `start` parameter (defaults to `1`).

Delegates to the existing `algorithm.product()` reduction, adding the `start` multiplier and the empty-span short-circuit.

### DIFF
--- a/mojo/stdlib/std/math/__init__.mojo
+++ b/mojo/stdlib/std/math/__init__.mojo
@@ -89,6 +89,7 @@ from .math import (
     min,
     modf,
     perm,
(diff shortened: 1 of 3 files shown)


msaelices:
Closing — algorithm.product() already exists and math.sum() is not in math either (it's in algorithm). Adding math.prod() would be a redundant wrapper with no clear precedent.

PR #6174, opened 2026-03-15
[stdlib] Add `Iterable` overload for `max`/`min`

bgreni:
## Summary

Add `Iterable` overload for the `max`/`min` functions. The implementation is a little messy due to needing to use
a placeholder origin since using the proper origin fails because `seq` _could_ be `RegisterPassable`.

## Testing

Added tests to stdlib test suite. This implementation raises on empty containers which I don't love, but I'm not sure how else to handle it and that does match the Python behaviour.

## Checklist

      sequence of smaller PRs

### DIFF
--- a/mojo/stdlib/std/math/math.mojo
+++ b/mojo/stdlib/std/math/math.mojo
@@ -3863,6 +3863,50 @@ def max[T: Copyable & Comparable](x: T, *ys: T) -> T:
     return res.copy()
 
 
+fn max[
+    origin: ImmutOrigin, T: Iterable, //
+](
+    ref[origin] seq: T,
+    out max_element: T.IteratorType[origin_of(seq)].Element,
+) raises StopIteration where conforms_to(
+    type_of(max_element), ImplicitlyDestructible & Comparable & Movable
+):
+    """
+    Gets the maximum value from a sequence of values.
+
+    Parameters:
+        origin: The origin of the sequence.
+        T: A type that is both copyable and comparable with greater than.
+
+    Args:
+        seq: The sequence to get the maximum value from.
+
+    Raises:
+        StopIteration: If the sequence is empty.
+
+    Returns:
+        The maximum value from the input sequence.
+    """
+    var it = iter(seq)
+
+    # Will raise on empty container
+    var m = trait_downcast_var[Comparable & Movable & ImplicitlyDestructible](
+        next(it)
+    )
+
+    while True:
+        try:
+            var n = trait_downcast_var[
+                Comparable & Movable & ImplicitlyDestructible
+            ](next(it))
+
+            if n > m:
+                m = n^
+        except:
+            break
+    max_element = m^
+
+
 # ===----------------------------------------------------------------------=== #
 # min
 # ===----------------------------------------------------------------------=== #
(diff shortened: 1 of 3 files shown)


mojo/stdlib/std/math/math.mojo
```diff
@@ -3863,6 +3863,44 @@ def max[T: Copyable & Comparable](x: T, *ys: T) -> T:
...
 
 
+fn max[
+    T: Iterable, //
+](
+    seq: T, out max_element: T.IteratorType[ImmutExternalOrigin].Element
+) raises StopIteration where conforms_to(
+    T.IteratorType[ImmutExternalOrigin].Element,
+    Comparable & Movable,
+):
```
NathanSWard, line 3873: In order to use the correct origin here, you'll need to manually specify `ref seq`.

```suggestion
fn max[
    T: Iterable, //
](
    ref seq: T, out max_element: T.IteratorType[origin_of(seq)].Element
) raises StopIteration where conforms_to(
    T.IteratorType[origin_of(seq)].Element,
    Comparable & Movable,
):
```

`ImmutExternalOrigin` is not correct (neither would the `AnyOrigin`) be as these could lead to exclusivity violations or early ASAP destruction when its unintended.

mojo/stdlib/std/math/math.mojo
```diff
@@ -3863,6 +3863,44 @@ def max[T: Copyable & Comparable](x: T, *ys: T) -> T:
     return res.copy()
 
 
+fn max[
+    T: Iterable, //
+](
+    seq: T, out max_element: T.IteratorType[ImmutExternalOrigin].Element
+) raises StopIteration where conforms_to(
```
NathanSWard, line 3871: Raising `StopIteration` here doesn't really make sense. A typed error should clearly convey what kind of error is produced and why. For example if I pass an empty `List`, seeing a `StopIteration` error raised isn't super telling of what happened especially if the is in the context of multiple other errors.

That all being said, I think returning an `Optional[T]` as the result would make more sense. :)

NathanSWard (changes requested):

NathanSWard:
This also feels like it may better belong in the `iter`/`itertools` module rather than `math`, but I'm not totally sure on that.

mojo/stdlib/std/math/math.mojo
bgreni: Seems like that doesn't quite work either.

```mojo
fn max[
    T: Iterable, //
](
    ref seq: T,
    out max_element: T.IteratorType[origin_of(seq)].Element,
) raises StopIteration where conforms_to(
    T.IteratorType[origin_of(seq)].Element,
    Comparable & Movable,
):
    var it = iter(seq)

    # Will raise on empty container
    max_element = trait_downcast_var[Comparable & Movable](next(it))

    while True:
        try:
            var n = trait_downcast_var[Comparable & Movable](next(it))
            """
            error: argument of call allows writing a memory location previously writable through another aliased argument
            if n > max_element:
               ~ ^ ~~~~~~~~~~~
            note: 'seq' memory accessed through reference embedded in value of type 'T.IteratorType[seq_is_mut, origin_of(seq), origin_of(seq)].Element(Comparable & Movable)'
            if n > max_element:
            """
            if n > max_element:
                max_element = n^
        except:
            break
```

Seems like this could be a gap in the usability of iterators since in this case i don't actually care about the origin of the container, I just need to know what type the iterator is going to return

mojo/stdlib/std/math/math.mojo
bgreni: I used `StopIteration` since that is what is thrown by `Iterator.__next__()`. Either option seems a bit annoying to only cover an edge case you _shouldn't_ do, but I suppose we'll run into this in lots of other places so it'll be fairly consistent.

mojo/stdlib/std/math/math.mojo
NathanSWard: If you make the origin `Immutable` it might work?
```
fn max[
    origin: ImmutOrigin,
    T: Iterable, //
](
    ref [origin] seq: T,
    out max_element: T.IteratorType[origin_of(seq)].Element,
)
```

Since then you won't get exclusivity violations.

All that being said - this is a very unfortunate part of the current iterator/origin design.
Can you create an issues to add `min/max` overloads that work w/ iterators and we can tackle it in the future once we cleanup iterators and ideally get origins working correctly?

bgreni: Yeah maybe this ended up being a bit more nasty to implement than I expected. I'll open a ticket and close this.

bgreni:
Closing because any implementation of this with the current features is quite gross, filed a ticket to track it's eventual implementation. #6179

PR #6183, opened 2026-03-16
[Serve] Validate target_endpoint for disaggregated inference requests

pperanich:
## Summary

Addresses the two TODO comments in `decode_scheduler.py`:

```
# TODO: Do not crash the scheduler if a request does not have a target endpoint.
#       Instead we should validate this in the frontend.
```

When running in disaggregated inference mode (`pipeline_role='decode_only'`),
requests missing a `target_endpoint` can crash the decode scheduler with an
unhandled `ValueError`. This PR adds validation at the two affected endpoints:

- **Frontend routes:** `/v1/chat/completions` and `/v1/completions` return
  HTTP 400 when `target_endpoint` is missing in decode-only mode, with a
  message explaining how to provide it.
- **Scheduler safety net:** `reserve_memory_and_send_to_prefill()` validates
  `target_endpoint` before allocating KV cache blocks. Invalid requests are
  dropped with a cancelled result, preventing both crashes and memory leaks.
- The two existing `raise ValueError` guards are replaced with
  `raise RuntimeError` since they should be unreachable after upstream
  validation.

## Testing

- `test_missing_target_endpoint_drops_request_gracefully` — verifies no
  crash, cancelled result sent back, no KV cache block leak, and a
  subsequent valid request still processes normally.

## Checklist

- [x] PR is small and focused

Assisted-by: AI

### DIFF
--- a/max/python/max/serve/router/openai_routes.py
+++ b/max/python/max/serve/router/openai_routes.py
@@ -1493,6 +1493,30 @@ def get_prompts_from_openai_request(
     raise Exception(f"unknown element type {type(prompt[0])}")
 
 
+def _get_and_verify_target_endpoint(
+    request: Request, completion_request: CreateCompletionRequest
+) -> str | None:
+    target_endpoint = _get_target_endpoint(
+        request, completion_request.target_endpoint
+    )
+    pipeline_config = get_app_pipeline_config(request.app)
+    if (
+        pipeline_config.runtime.pipeline_role == "decode_only"
+        and target_endpoint is None
+    ):
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                "This server is running in disaggregated inference"
+                " mode (pipeline_role='decode_only') and requires a"
+                " target_endpoint to route requests to a prefill"
+                " worker. Provide it via the 'X-Target-Endpoint'"
+                " header or the 'target_endpoint' request body field."
+            ),
+        )
+    return target_endpoint
+
+
 @router.post("/completions", response_model=None)
 async def openai_create_completion(
     request: Request,
@@ -1513,24 +1537,9 @@ async def openai_create_completion(
         )
         assert isinstance(pipeline, TokenGeneratorPipeline)
 
-        target_endpoint = _get_target_endpoint(
-            request, completion_request.target_endpoint
+        target_endpoint = _get_and_verify_target_endpoint(
+            request, completion_request
         )
-        pipeline_config = get_app_pipeline_config(request.app)
-        if (
-            pipeline_config.runtime.pipeline_role == "decode_only"
-            and target_endpoint is None
-        ):
-            raise HTTPException(
-                status_code=400,
-                detail=(
-                    "This server is running in disaggregated inference"
-                    " mode (pipeline_role='decode_only') and requires a"
-                    " target_endpoint to route requests to a prefill"
-                    " worker. Provide it via the 'X-Target-Endpoint'"
-                    " header or the 'target_endpoint' request body field."
-                ),
-            )
 
         logger.debug(
             "Path: %s, Request: %s%s, Model: %s",
--- a/max/python/max/serve/router/openai_routes.py
+++ b/max/python/max/serve/router/openai_routes.py
@@ -943,24 +943,9 @@ async def openai_create_chat_completion(
                 else 1
             )
 
-        target_endpoint = _get_target_endpoint(
+        target_endpoint = _get_and_verify_target_endpoint(
             request, completion_request.target_endpoint
         )
-        pipeline_config = get_app_pipeline_config(request.app)
-        if (
-            pipeline_config.runtime.pipeline_role == "decode_only"
-            and target_endpoint is None
-        ):
-            raise HTTPException(
-                status_code=400,
-                detail=(
-                    "This server is running in disaggregated inference"
-                    " mode (pipeline_role='decode_only') and requires a"
-                    " target_endpoint to route requests to a prefill"
-                    " worker. Provide it via the 'X-Target-Endpoint'"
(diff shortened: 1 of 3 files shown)


pperanich:

abduld:
fyi @AustinDoolittle / @wrongbad

AustinDoolittle:
@langong347 Could you take a look at this?

max/python/max/serve/router/openai_routes.py
```diff
@@ -1480,6 +1497,25 @@ async def openai_create_completion(
...
+                status_code=400,
+                detail=(
+                    "This server is running in disaggregated inference"
+                    " mode (pipeline_role='decode_only') and requires a"
+                    " target_endpoint to route requests to a prefill"
+                    " worker. Provide it via the 'X-Target-Endpoint'"
+                    " header or the 'target_endpoint' request body field."
+                ),
+            )
+
```
langong347: Consider encapsulating these into a reusable function. Could be just absorbing the check into something like `_get_and_verify_target_endpoint()`

langong347 (commented):
@pperanich Thank you for the PR! LG overall, please review my comment below.

max/python/max/serve/router/openai_routes.py
pperanich: Done!

max/python/max/serve/router/openai_routes.py
```diff
@@ -948,6 +948,25 @@ async def openai_create_chat_completion(
                 else 1
             )
 
+        target_endpoint = _get_target_endpoint(
```
langong347: Please replace this part too.

pperanich: target_endpoint is needed later in the TextGenerationRequest.

langong347: The reusable function also returns `target_endpoint`:

```python
target_endpoint = _get_and_verify_target_endpoint(
            request, completion_request
        )
```

pperanich:
Good catch, applied `_get_and_verify_target_endpoint` to the chat completion endpoint too. Generalized the helper to take the body `target_endpoint` value directly so both `CreateChatCompletionRequest` and `CreateCompletionRequest` work.

Let me know if there is anything else I can do to get this over the finish line.

langong347 (approved):

max/python/max/serve/router/openai_routes.py
```diff
@@ -1476,6 +1478,28 @@ def get_prompts_from_openai_request(
...
 
 
+def _get_and_verify_target_endpoint(
+    request: Request, body_target_endpoint: str | None
+) -> str | None:
+    target_endpoint = _get_target_endpoint(request, body_target_endpoint)
+    pipeline_config = get_app_pipeline_config(request.app)
+    if (
+        pipeline_config.runtime.pipeline_role == "decode_only"
+        and target_endpoint is None
```
ns-jjwathen: Presence-only check: `"target_endpoint": ""` in the body is not `None`, so it skips this 400 (note `_get_target_endpoint` treats an empty *header* as absent via truthiness but returns the body value as-is), then skips the scheduler safety net, and reaches `ZmqDealerSocket(endpoint="")` in `dispatcher_v2.py` — an uncaught `zmq.ZMQError` that kills the scheduler, the exact crash class this PR fixes.

```suggestion
        and not target_endpoint
```

Ideally also validate the value parses as a supported ZMQ endpoint (`tcp://host:port` / `ipc://...`) here so malformed strings get a 400 too.

ns-jjwathen (commented):
_Review withdrawn — posted here in error (automation pointed at the wrong PR list). Apologies for the noise._

max/python/max/serve/scheduler/decode_scheduler.py
```diff
@@ -238,6 +239,23 @@ def reserve_memory_and_send_to_prefill(self) -> None:
             req_id = context.request_id
             del self.pending_reqs[req_id]
 
+            # Validate target_endpoint before allocating resources.
+            # NOTE: The client receives a cancelled response with no error
+            # detail. Frontend route validation should catch this earlier
+            # with a proper 400 error. This guard is a safety net.
+            if context.target_endpoint is None:
```
ns-jjwathen: Same empty-string gap as the frontend guard: `""` is not `None`, so it sails past this safety net into `send_prefill_request` → `send_request_nowait(metadata, "")` → `ZmqDealerSocket(endpoint="")`, and the resulting `zmq.ZMQError` propagates uncaught through `run_iteration`.

```suggestion
            if not context.target_endpoint:
```

Also, after the (required) rebase, main now has `SchedulerResult.failed(error)` — please use it here instead of `cancelled()` so the client gets the actual reason rather than a silent drop (details in the main review comment).

max/tests/tests/serve/scheduler/test_di.py
```diff
@@ -1400,3 +1400,57 @@ def test_overlap_di_both_sides_minimal_output() -> None:
...
+    # Run the scheduler — should not crash
+    decode.run_iteration()
+
+    # The bad request should get a cancelled result
+    found_bad = False
+    while not q.response_queue.empty():
+        batch = q.response_queue.get()
+        if bad_req_id in batch:
+            found_bad = True
+            assert batch[bad_req_id].result is None  # cancelled
```
ns-jjwathen: This doesn't actually verify termination: per the `SchedulerResult` docstring, `result` "may be None for cancelled operations **or during intermediate states of streaming operations**". If the scheduler regressed to emitting a non-done placeholder, the client would hang forever and this test would stay green.

```suggestion
            assert batch[bad_req_id].is_done and batch[bad_req_id].result is None
```

PR #6184, opened 2026-03-16
[Serve] Export speculative decoding metrics to OTEL/Prometheus

pperanich:
## Summary

The speculative decoding pipeline computes acceptance stats and logs them to
console via `pretty_format()`, but never publishes them to the telemetry
system. This PR bridges that gap by adding three OTEL counters at `DETAILED`
metric level:

- `maxserve.speculative.draft_tokens_generated`
- `maxserve.speculative.draft_tokens_accepted`
- `maxserve.speculative.bonus_tokens_used` — this field already existed on
  `SpeculativeDecodingMetrics` but was never propagated to `BatchMetrics`.
  Adding it required touching the dataclass, `create()`, and the test
  constructor. It could be dropped deriving it from
  the other two counters is preferred.

`BatchMetrics.publish_metrics()` emits these counters only when
`draft_tokens_generated > 0`, avoiding noise when speculative decoding
is not active.

Additionally, `SpeculativeDecodingMetrics` is now reset after each batch in
`BatchMetrics.create()`, matching the existing `kv_cache.reset_metrics()`
pattern. Without this reset, `BatchMetrics.create()` reads cumulative
lifetime totals rather than per-batch deltas. This was already an issue
for `pretty_format()` console logs (which showed ever-growing totals
instead of per-batch stats), and would cause the new OTEL counters to
double-count since counters are additive.

**Note:** This changes the existing `pretty_format()` console log output
from cumulative lifetime stats to per-batch stats (e.g., "Draft Tokens:
15/30" instead of an ever-growing total). Every other metric in the log
line (batch size, throughput, KV cache stats) was already per-batch —
the speculative decoding fields were the only cumulative outlier.

## Testing

- `test_create_resets_speculative_metrics_between_batches` — regression
  test that calls `BatchMetrics.create()` twice with the same
  `SpeculativeDecodingMetrics` object. Verifies the second call returns
  only the new batch's values, not cumulative totals. Fails without the
  `reset()` fix (batch 2 reads 30 instead of 10).

## Checklist

- [x] PR is small and focused

Assisted-by: AI

### DIFF
--- a/max/python/max/serve/scheduler/utils.py
+++ b/max/python/max/serve/scheduler/utils.py
@@ -79,6 +79,7 @@ class BatchMetrics:
     draft_tokens_accepted: int
     avg_acceptance_length: float
     max_acceptance_length: int
+    bonus_tokens_used: int
 
     nixl_read_latency_avg_ms: float
     nixl_write_latency_avg_ms: float
@@ -196,6 +197,7 @@ class BatchMetrics:
         draft_tokens_accepted = 0
         avg_acceptance_length = 0.0
         max_acceptance_length = 0
+        bonus_tokens_used = 0
         if speculative_decoding_metrics is not None:
             draft_tokens_generated = (
                 speculative_decoding_metrics.draft_tokens_generated
@@ -209,6 +211,8 @@ class BatchMetrics:
             max_acceptance_length = (
                 speculative_decoding_metrics.num_speculative_tokens
             )
+            bonus_tokens_used = speculative_decoding_metrics.bonus_tokens_used
+            speculative_decoding_metrics.reset()
 
         return cls(
             batch_type=inputs.batch_type,
@@ -241,6 +245,7 @@ class BatchMetrics:
             draft_tokens_accepted=draft_tokens_accepted,
             avg_acceptance_length=avg_acceptance_length,
             max_acceptance_length=max_acceptance_length,
+            bonus_tokens_used=bonus_tokens_used,
             nixl_read_latency_avg_ms=nixl_read_latency_avg_ms,
             nixl_write_latency_avg_ms=nixl_write_latency_avg_ms,
             rpc_acquire_latency_avg_ms=rpc_acquire_latency_avg_ms,
@@ -339,6 +344,15 @@ class BatchMetrics:
             METRICS.dkv_rpc_acquire_latency(self.rpc_acquire_latency_avg_ms)
         if self.rpc_read_latency_avg_ms > 0:
             METRICS.dkv_rpc_read_latency(self.rpc_read_latency_avg_ms)
+
+        if self.draft_tokens_generated > 0:
+            METRICS.speculative_draft_tokens_accepted(
+                self.draft_tokens_accepted
+            )
+            METRICS.speculative_draft_tokens_generated(
+                self.draft_tokens_generated
+            )
+            METRICS.speculative_bonus_tokens_used(self.bonus_tokens_used)
 
 
 class SchedulerLogger:
--- a/max/python/max/serve/scheduler/utils.py
+++ b/max/python/max/serve/scheduler/utils.py
@@ -75,11 +75,15 @@ class BatchMetrics:
     disk_blocks_written: int
     disk_blocks_read: int
 
+    # Cumulative lifetime totals, used by pretty_format for console logs.
     draft_tokens_generated: int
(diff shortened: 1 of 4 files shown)


abduld:
fyi @AustinDoolittle / @wrongbad

atomicapple0 (commented):
This change makes sense to me. Thank you!

One request. Can we keep printing the culmulative acceptance rates instead of the per batch acceptance rates? We may move the kvcache cache hit rate to culmulative. Folks have been confused about how to interpret the per batch cache hit rate in cases like chunked prefill.

pperanich:
One thing worth noting: the cumulative behavior before this PR appears to have been a bug. 

SpeculativeDecodingMetrics was never reset between batches, so BatchMetrics.create() was reading ever-growing lifetime totals. This meant pretty_format() showed cumulative stats, but it also means the new OTEL counters would double-count: batch 1 emits 20 tokens, batch 2 emits 30 (20 + 10), batch 3 emits 45 (20 + 10 + 15), etc. Since OTEL counters are additive, that's incorrect. The reset() fix is necessary for correct telemetry.

The complication is that fixing the bug for OTEL also changed the console output from cumulative to per-batch, since both were reading from the same source. To get what you're describing, we'd need to decouple the two. E.g. compute the cumulative acceptance rate before the reset and pass it through to pretty_format() separately from the per-batch deltas used by OTEL.

Happy to add that in this PR or as a follow-up.

atomicapple0:
Yes, we want OTEL to publish per batch metrics but the console logs to display cumulative stats (possibly switching to rolling 60s or something fancier down the line). Perhaps track something like last batch stats in the metrics object?

Also heads up that there are some incoming changes from the internal repo that might conflict a bit with this PR.

pperanich:
Rebased on latest main and split the metrics as suggested. `SpeculativeDecodingMetrics` now keeps cumulative totals (so `pretty_format()` shows lifetime stats) and tracks a private snapshot via `publish_delta()` that returns per-batch increments for the additive OTEL counters. `BatchMetrics` carries both cumulative fields (console) and `_delta` fields (OTEL).

Also dropped `bonus_tokens_used` since main's simplified `SpeculativeDecodingMetrics` no longer has the field and it's derivable from the accepted/generated counters anyway.

max/python/max/serve/telemetry/metrics.py
```diff
@@ -170,6 +170,16 @@
         unit="ms",
         description="Audio output length in milliseconds",
     ),  # type: ignore
+    "maxserve.speculative.draft_tokens_accepted": _meter.create_counter(
```
ns-jjwathen: **Namespace conflict with main:** main already exports speculative-decoding metrics under `maxserve.spec_decode.*` (`avg_acceptance_length`, `acceptance_rate_per_position`), documented in `docs/max/serve/metrics.mdx`. Introducing `maxserve.speculative.*` gives dashboards two prefixes for the same subsystem — and metric names are effectively frozen once released. Please name these `maxserve.spec_decode.draft_tokens_accepted` / `maxserve.spec_decode.draft_tokens_generated`, with `spec_decode_draft_tokens_*` accessor methods to match the existing ones.

max/python/max/serve/scheduler/utils.py
```diff
@@ -209,6 +216,10 @@ def create(
             max_acceptance_length = (
                 speculative_decoding_metrics.num_speculative_tokens
             )
+            (
+                draft_tokens_generated_delta,
+                draft_tokens_accepted_delta,
+            ) = speculative_decoding_metrics.publish_delta()
```
ns-jjwathen: `create()` reads like a pure constructor but now mutates the shared `SpeculativeDecodingMetrics` (advances the `_published_*` snapshot) as a side effect. Any `create()` not followed by `publish_metrics()` on the same instance — an exception between the two calls, a future caller that only wants `pretty_format()`, or a test — silently and permanently drops those tokens from the OTEL counters. Today the sole production call site pairs them, but that contract is undocumented. If the delta split survives the rebase (see main-drift comment — it likely shouldn't), move the delta consumption into `publish_metrics()` so it happens at emission time, or at minimum rename to `take_delta()` and document the one-create-one-publish contract.

ns-jjwathen (commented):
_Review withdrawn — posted here in error (automation pointed at the wrong PR list). Apologies for the noise._

PR #6189, opened 2026-03-17
[tests] Enhance local testing setup and add model download utility

frostming:
- Updated CONTRIBUTING.md and development.md to clarify local test prerequisites and usage of a minimal test matrix.
- Introduced download_models.py script for downloading Hugging Face models into the local cache.
- Added tests for model path generation and download functionality.
- Updated BUILD.bazel files to include new test suites and dependencies.

Signed-off-by: Frost Ming <me@frostming.com>

## Checklist

- [ ] PR is small and focused — consider splitting larger changes into a
      sequence of smaller PRs


trojanezra-del:
Do we then have any LoRA adapters for this local testing? Or perhaps support on this Hugging Face transformers under the Max engine.

abduld:
nice, fyi @mdanatg

max/tests/integration/tools/download_models_for_testing.py
mdanatg, line 1: Suggest: how about we call it download_models_for_testing.py, so that it doesn't get confused with a more general tool?

max/docs/development.md
```diff
@@ -50,6 +50,34 @@ From the repo root, run this `bazelw` command to run all the MAX tests:
 If it's your first time, it starts by installing the Bazel version manager,
 [Bazelisk](https://github.com/bazelbuild/bazelisk), which then installs Bazel.
 
+### Know the local test prerequisites
+
+Not every MAX test has the same local requirements. Before running a broad
+target, check which of these constraints apply:
+
+- **`HF_TOKEN`**: Some tests exercise gated Hugging Face repos or fetch remote
```
mdanatg: Could you include something in the lines of "See https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hftoken for mode information about HF_TOKEN"

max/docs/development.md
```diff
@@ -66,6 +94,23 @@ To find all the test targets, you can run:
 ./bazelw query 'tests(//max/tests/...)'
 ```
 
+### Use a minimal test matrix
```
mdanatg: Suggest: Minimal test matrix (just so that we don't encourage users to run as few tests as possible - we want all tests to pass)

max/tests/tests/BUILD.bazel
```diff
@@ -108,3 +108,22 @@ modular_py_test(
         requirement("torch"),
     ],
 )
+
+test_suite(
+    name = "cpu_local_tests",
```
mdanatg, line 115: We should popularize this test suite, and encourage everyone at Modular to write tests that fit in here.

mdanatg (approved):
Thanks for driving these!

cc @scottamain @BradLarson @conorbronsdon @ansley in case the exact verbiage needs some wordsmithing

max/tests/integration/tools/download_models_for_testing.py
frostming: Updated the PR to address your feedback, what does it look now?

max/docs/development.md
```diff
@@ -50,6 +50,36 @@ From the repo root, run this `bazelw` command to run all the MAX tests:
 If it's your first time, it starts by installing the Bazel version manager,
 [Bazelisk](https://github.com/bazelbuild/bazelisk), which then installs Bazel.
 
+### Know the local test prerequisites
+
+Not every MAX test has the same local requirements. Before running a broad
```
AustinDoolittle, line 55: should we instead advocate that folks use `hf auth login`?

frostming: Updated, thanks

max/python/max/pipelines/lib/hf_utils.py
```diff
@@ -658,8 +658,9 @@ def is_diffusion_pipeline(repo: HuggingFaceRepo) -> bool:
 def generate_local_model_path(repo_id: str, revision: str) -> str:
     """Generate the local filesystem path where a Hugging Face model repo is cached.
 
-    This function uses Hugging Face's official snapshot_download with local_files_only=True
-    to resolve the local cache path for a model repository.
+    This function resolves the model from the local Hugging Face cache only.
+    Missing snapshots should be pre-downloaded explicitly so tests tagged as
```
conorbronsdon: Nit: the repo uses a `requires-network` Bazel tag, not a `network-free` tag — this phrasing could send contributors looking for a tag that doesn't exist. How about:

```suggestion
    Missing snapshots should be pre-downloaded explicitly so tests without the
    ``requires-network`` tag do not silently fetch remote artifacts at runtime.
```

max/python/max/pipelines/lib/hf_utils.py
```diff
@@ -670,16 +671,23 @@ def generate_local_model_path(repo_id: str, revision: str) -> str:
...
+        FileNotFoundError: If the model is not found in the local cache.
     """
     try:
         return huggingface_hub.snapshot_download(
             repo_id=repo_id,
             revision=revision,
             local_files_only=True,
         )
-    except huggingface_hub.utils.LocalEntryNotFoundError as e:
+    except huggingface_hub.utils.LocalEntryNotFoundError as local_error:
```
conorbronsdon: The tests import from `huggingface_hub.errors` (as `hf_hub_errors`), but the production code catches from `huggingface_hub.utils`. These are aliases today, but they could diverge in a future `huggingface-hub` release. Switching to the `errors` module here keeps test and production imports consistent:

```suggestion
    except huggingface_hub.errors.LocalEntryNotFoundError as local_error:
```

max/tests/integration/tools/download_models_for_testing.py
```diff
@@ -0,0 +1,124 @@
...
+            f"Downloading {request.repo_id} "
+            f"(revision: {revision_display}, "
+            f"source: {request.revision_source})..."
+        )
+        try:
+            snapshot_path = download_repo(
+                request,
+                force_download=force_download,
+            )
+        except Exception as error:
```
conorbronsdon, line 113: When a download fails, the error message drops the resolved revision, which makes it harder to tell whether the failure was against a lockfile pin, an explicit override, or the default branch. Including it would make debugging easier:

```suggestion
            revision_display = request.revision or "default branch"
            failures.append(
                f"{request.repo_id} (revision: {revision_display}): {error}"
            )
```

max/tests/tests/nn/BUILD.bazel
```diff
@@ -44,9 +50,18 @@ _LONG_TIMEOUT_TESTS = [
             "//max/python/max/graph",
             "//max/python/max/mlir",
             "//max/python/max/nn",
-            "//max/python/max/pipelines/lib",
             requirement("numpy"),
-        ],
+        ] + ([
+            "//max/python/max/pipelines/lib",
```
conorbronsdon: Minor: this conditional is easy to miss for someone adding a new test later. A short comment explaining the "why" would help future contributors:

```suggestion
        ] + (
            # pipelines/lib is only needed by the GPU sampling test; keep it
            # out of the default dep list so cpu_tests stays lightweight.
            ["//max/python/max/pipelines/lib"]
            if src == "test_sampling_gpu.py"
            else []
        ),
```

max/CONTRIBUTING.md
```diff
@@ -21,7 +21,14 @@ For technical details on developing for MAX and models, see the following
 document:
 
 - [Developing the MAX framework](/max/docs/development.md) covers building,
-  testing, and other information you’ll need to work in the MAX framework.
+  testing, including local test prerequisites such as `HF_TOKEN`, model
+  downloads, and GPU-only targets, and other information you’ll need to work
+  in the MAX framework.
```
conorbronsdon: This sentence is doing a lot of heavy lifting. The nested clauses make it hard to parse on first read. How about breaking it up:

```suggestion
  testing, and other information you'll need to work in the MAX framework.
  It also covers local test prerequisites (`HF_TOKEN`, model downloads,
  GPU-only targets) and a minimal test matrix for local iteration.
```

conorbronsdon (commented):
This is great work, @frostming. The minimal test matrix table alone is going to save contributors a lot of time, and `download_models_for_testing` solves an actual pain point for anyone setting up on a fresh machine. Docs are solid.

Left a few suggestions inline: consistency nits and one diagnostic improvement. Nothing blocking.

& sorry for the delayed review! Let's get this merged 💪

conorbronsdon:
The lint failure isn't a code issue — Bazel got a 502 from GitHub's CDN when downloading `bazel_clang_tidy`. I kicked off a re-run. Should go green once the transient error clears.

Also dismissed the Copilot reviews since my inline suggestions cover the same ground.

conorbronsdon (approved):

conorbronsdon:
The lint failure is buildifier formatting on `max/tests/tests/nn/BUILD.bazel` — my suggestion used multi-line formatting but buildifier wants the ternary on one line. Fix:

```bash
./bazelw run format
```

That should collapse lines 56-58 to:

```starlark
            ["//max/python/max/pipelines/lib"] if src == "test_sampling_gpu.py" else []
```

Sorry for the false start on that one!

Ahajha:
!sync

PR #6190, opened 2026-03-17
[stdlib] Make Dict reserve the requested minimum capacity

martinvuyk:
Make Dict reserve the requested minimum capacity

Previously, `Dict(capacity=N)` applied `next_power_of_two(N)` directly, but the usable capacity before resizing is only `7/8` of the internal capacity. So `Dict(capacity=1000)` gave an internal capacity of `1024` but could only hold `1024 * 7 / 8 = 896` entries — less than the requested `1000`. This PR compensates by computing `next_power_of_two(ceildiv(capacity * 8, 7))`, ensuring the usable capacity meets or exceeds the requested amount.

### DIFF
--- a/mojo/stdlib/std/collections/dict.mojo
+++ b/mojo/stdlib/std/collections/dict.mojo
@@ -41,6 +41,7 @@ from std.compile import get_type_name
 from std.hashlib import Hasher, default_comp_time_hasher, default_hasher
 import std.format._utils as fmt
 from std.sys.intrinsics import likely
+from std.math import ceildiv
 
 from std.bit import count_trailing_zeros, next_power_of_two
 from std.memory import alloc, bitcast, memcpy, memset, pack_bits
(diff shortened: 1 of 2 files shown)


JoeLoser (approved):
Interesting find, thanks for the fix! Can we please update the commit message to say something along the lines of 

"Previously, `Dict(capacity=N)` applied `next_power_of_two(N)` directly, but the usable capacity before resizing is only 7/8 of the internal capacity. So Dict(capacity=1000) gave an internal capacity of 1024 but could only hold 1024 * 7 / 8 = 896 entries — less than the requested 1000. This PR compensates by computing `next_power_of_two(ceildiv(capacity * 8, 7))`, ensuring the usable capacity meets or exceeds the requested amount." 

Should there be a test that directly validates the contract — i.e., insert exactly N items into Dict(capacity=N) and assert no resize occurred? The current tests check _reserved() (raw capacity) but don't directly verify the "no resize" guarantee.

martinvuyk:
> Should there be a test that directly validates the contract — i.e., insert exactly N items into Dict(capacity=N) and assert no resize occurred? The current tests check _reserved() (raw capacity) but don't directly verify the "no resize" guarantee.

@JoeLoser I think the existing check_reserved tests already cover this, the reserved capacity should be equal to what is required given a load factor. What I did wonder while I was doing this is that maybe some people would like to define the load factor they will be using as a parameter for `Dict` at some point, so these tests would all need to use the parameter to check and be changed to `expected <= reserved` instead of strict equality

JoeLoser:
!sync

PR #6191, opened 2026-03-17
[stdlib] Add `Dict.__getitem_param__` to hash keys at comptime

martinvuyk:
Add `Dict.__getitem_param__` to hash keys at comptime

### DIFF
--- a/mojo/stdlib/std/collections/dict.mojo
+++ b/mojo/stdlib/std/collections/dict.mojo
@@ -537,7 +537,7 @@ struct DictEntry[
             key: The key of the entry.
             value: The value of the entry.
         """
-        self.hash = hash[HasherType=Self.H](key)
+        self.hash = hash[Self.H](key)
         self.key = key^
         self.value = value^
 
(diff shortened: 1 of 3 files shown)


martinvuyk:
> Mojo now has a simple and explicit policy: if a type implements a `__getitem_param__` or `__getattr_param__` method and the indices are valid parameter expressions (e.g. doesn't refer to a dynamic value as a subexpr), the compiler will pick it (but will not support a `__setitem__` pair).

@lattner Thoughts on enabling a `__setitem_param__` ? Besides being able to lift some of the `setitem` computations into compile time, I don't know if you guys would be able to solve this kind of ambiguity:
```
mojo/stdlib/test/collections/test_dict.mojo:1090:10: error: cannot call function that may raise in a context that cannot raise
        d["ct"] = 42
         ^~~~~~
```
I added this getter to compute the hash at compile time
```mojo
    def __getitem_param__[
        key: Self.K
    ](ref self) raises DictKeyError[Self.K] -> ref[self] Self.V:
        comptime h = hash[Self.H](key)
        return self._find_ref(h, materialize[key]())

    def __getitem__(
        ref self, ref key: Self.K
    ) raises DictKeyError[Self.K] -> ref[self] Self.V:
        return self._find_ref(hash[Self.H](key), key)

    def __setitem__(mut self, var key: Self.K, var value: Self.V):
        self._insert(key^, value^)
```
It would need the equivalent setitem (that does't raise) for compile time, or we need a way to make the compiler understand it should choose the runtime setitem

PR #6194, opened 2026-03-17
[stdlib] Simplify `Hashable` and `Hasher` traits fn signatures

martinvuyk:
Simplify `Hashable` and `Hasher` traits fn signatures

### DIFF
--- a/mojo/stdlib/std/hashlib/hash.mojo
+++ b/mojo/stdlib/std/hashlib/hash.mojo
@@ -70,12 +70,9 @@ trait Hashable:
     fields implement both traits correctly.
     """
 
-    def __hash__[H: Hasher](self, mut hasher: H):
+    def __hash__(self, mut hasher: Some[Hasher]):
         """Accepts a hasher and contributes to the hash value
         by calling the update function of the hasher.
-
-        Parameters:
-            H: Any Hasher type.
 
         Args:
             hasher: The hasher instance to contribute to.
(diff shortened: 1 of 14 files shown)


mojo/stdlib/std/hashlib/hash.mojo
```diff
@@ -95,7 +92,7 @@ trait Hashable:
 
 
 def hash[
-    T: Hashable, HasherType: Hasher = default_hasher
+    T: Hashable, //, HasherType: Hasher = default_hasher
 ](hashable: T) -> UInt64:
```
NathanSWard, line 96: Could this use `Some[Hasable]` here?
```
def hash[HasherType: Hasher = default_hasher](hashable: Some[Hashable])
```

NathanSWard (approved):
great cleanup! 😄 
Left a small comment about one of the methods - but I can also fix this internally!

mojo/stdlib/std/hashlib/hash.mojo
martinvuyk: There is one test in dict that binds the overload to a single alias (`comptime _hash_with_hasher`) and can't do it without that T to be explicitly left as unbound. So until we either remove that hurdle or allow comptime binding to overloads, this has to stay like this :/

NathanSWard: Ahh makes sense, thanks for the explanation :)

NathanSWard:
!sync

NathanSWard (approved):

PR #6195, opened 2026-03-17
[stdlib] [proposal] String iterators > List allocations APIs

martinvuyk:
[Forum link for discussion](https://forum.modular.com/t/proposal-string-iterators-list-allocations-apis/2856)

### DIFF
--- a/mojo/proposals/string-iterators.md
+++ b/mojo/proposals/string-iterators.md
@@ -1,0 +1,70 @@
+# String iterators > List allocations APIs
+
+date: 2026-03-17, status: proposed
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


NathanSWard:
Hey @martinvuyk !
After taking with the team I think we're good to green light this approach. I went ahead and opened a larger tracking issue here: https://github.com/modular/modular/issues/6359 to encapsulate all of the iterator types returned from the string functions.

The one thing that I'd hold off on for the initial implementation is the `StringSliceIterator`. I could see this potentially being useful, but let's first implement all of the iterator themselves and see if they can't compose better via more generic iterator adaptors.

PR #6200, opened 2026-03-18
[stdlib] [proposal] Codepoint as UTF-8 readonly view

martinvuyk:
[forum discussion thread](https://forum.modular.com/t/stdlib-proposal-codepoint-as-utf-8-readonly-view-6200/2859)

### DIFF
--- a/mojo/proposals/codepoint-as-utf8-readonly-view.md
+++ b/mojo/proposals/codepoint-as-utf8-readonly-view.md
@@ -1,0 +1,36 @@
+# Codepoint as UTF-8 readonly view
+
+date: 2026-03-18, status: proposed
+
+## Motivation
+
+Currently the `Codepoint` type stores Unicode codepoints as their UTF-32
+representation. It is a niche type and not often used because the performance
+hit of parsing a unicode codepoint from a UTF-8 iterator is significant.
+Codepoint's API is also somewhat limited compared to `String`, `StringSlice`,
+and `StringLiteral`.
+
+## Concrete proposal
+
+While I would prefer to parametrize it based on encoding, we can go in a simpler
+direction first that still achieves the same purpose:
+
+1. Let's switch the struct's internals to be the raw UTF-8 value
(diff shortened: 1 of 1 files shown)


ConnorGray:
Hey @martinvuyk, really interesting idea 💡. The team had a chance to discuss this, and our takeaway was that it seems like a good idea. It's an elegant way to optimize these operations, and avoiding repeated UTF-8 decode/encode steps.

Our only uncertainty was how much of a performance win this would be in practice. It seems intuitive that there would be gains for things like iterating over codepoints (no decode steps), appending codepoints, even printing codepoints, since those all today would encode the `Codepoint` to a 1-4 byte sequence. However, always good to have numbers to validate a change like this in practice 🙂.

If this is something you're interested in working on, would love to see this move forward 🙂

martinvuyk:
Hi @ConnorGray, great to know you guys are considering this :tada: !!

> Our only uncertainty was how much of a performance win this would be in practice. It seems intuitive that there would be gains for things like iterating over codepoints (no decode steps), appending codepoints, even printing codepoints, since those all today would encode the `Codepoint` to a 1-4 byte sequence.

I agree with the sentiment, but I wasn't thinking much about `Codepoint` itself. To me it seems obvious this will be a win for everything to do with the struct itself in terms of performance. I was thinking more around what this will allow in terms of iteration patterns and some more encoding-native lookup tables/logic, etc. This will allow us to create a bunch of unicode-codepoint-based iterators (that can return different data than what the origin had).

The only part where I want to be cautious is around making it the default string iterator (vs using `CodepointSliceIter`), because I'm not sure what kind of impact copying the 1-4 UTF-8 bytes will have on iteration time vs just dereferencing the first byte and jumping forward like `CodepointSliceIter` does.

> However, always good to have numbers to validate a change like this in practice 🙂.

I can add a benchmark and compare the parsing of a unicode codepoint ordinal value vs the new utf8-native one in the PR where I do the switch if that helps.

PR #6203, opened 2026-03-19
[stdlib] Fix Float64 parsing accuracy for subnormal values and add regression …

### LINKED ISSUE #6196: [BUG] Float parsing accuracy issues
### Bug description

I wrote a bit of a [torture test](https://github.com/bgreni/EmberJson/blob/main/test_float_roundtrip.mojo) for my float parsing and string formatting implementations in EmberJson. It simply does a roundtrip of each side and verifies that the result has the same bit pattern as the input number. After I had ironed out all the edge case issues in my implementation, I figured I might as well apply it to the Mojo stdlib equivalents, and it appears there are a few correctness issues still hiding in there. Note that when I was testing the stdlib implementations I was only using Float64 even though the full test suite also covers Float32 and Float16.

### Steps to reproduce

```mojo
def main() raises():
    print(Float64("4.4501363245856945e-308")) # prints 2.225062466078493e-308
```

### System information

```
       Pixi version: 0.63.2
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=26.3.1=0

mojo: 0.26.3.0.dev2026031705 
```

mahendrarathore1742:
Fix Float64 parsing accuracy for subnormal values

- Fixed lemire_algorithm to properly handle subnormal (denormalized) float values
- Previously, Float64("4.4501363245856945e-308") was incorrectly parsed as 2.225062466078493e-308
- Added proper handling for edge cases where q < -342 or p <= -1086
- Added regression test to verify roundtrip accuracy for subnormal values
- Ensures bit-pattern preservation for all Float64 values within representable range

Fixes #6196

### DIFF
--- a/mojo/stdlib/std/collections/string/_parsing_numbers/parsing_floats.mojo
+++ b/mojo/stdlib/std/collections/string/_parsing_numbers/parsing_floats.mojo
@@ -216,12 +216,8 @@ def lemire_algorithm(var w: UInt64, var q: Int64) -> Float64:
     # This algorithm has 22 steps described
     # in https://arxiv.org/pdf/2101.11408 (algorithm 1)
     # Step 1
-    if w == 0:
+    if w == 0 or q < -342:
         return 0.0
-    # Allow parsing for subnormal values near lower Float64 limit
-    if q < -342:
-        # Use smallest subnormal value for Float64
-        return Float64(5e-324)
 
     # Step 2
     if q > 308:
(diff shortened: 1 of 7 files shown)


pixi.lock
JoeLoser, line 1: @Ahajha do we want these changes to the pixi files? I'd suggest removing the `pixi.toml` and `pixi.lock`, and therefore also changes to the `.gitignore` and `.gitattributes` please.

mojo/stdlib/std/collections/string/_parsing_numbers/parsing_floats.mojo
```diff
@@ -257,10 +257,11 @@ def lemire_algorithm(var w: UInt64, var q: Int64) -> Float64:
     # Subnormal case
     if p <= -1022:
         s = -1022 - p
-        m = m // (2 ** UInt64(s))
-        if m % 2 == 1:
-            m += 1
-        m >>= 1
+        m = m >> UInt64(s)
```
JoeLoser, line 260: **Question** Is this algorithmic fix "correct"? If I understand correctly, the new code is equivalent for unsigned integers. The test is failing in CI as well.

JoeLoser:
Thanks for the contribution. I went ahead and poked at this since it was easy to fix and there seemed to be multiple bugs. Please take a look at https://github.com/modular/modular/commit/0aeaaaf52ec2a4fbd523144f872886debc9dad69 and open a new PR if you see any other issues, please 🙏

PR #6208, opened 2026-03-19
[Kernels] Reduce signal SM overhead in EP dispatch by handling 2 experts per warp

prsabahrami:
PRAGMA-guided optimization of the Expert Parallel (EP) dispatch kernel signal overhead.

## Summary

Optimize the  in  so each signal warp handles **2 experts sequentially** instead of 1, halving the number of signal SMs needed for the same number of experts.

**Problem**: In the EP dispatch kernel, signal warps are used to notify receiving GPUs when their expert data is ready. With one expert per warp, a 256-expert H100 job consumed 8 full SMs just for signaling. Signal work is latency-light but was over-provisioned, leaving fewer SMs for actual communication.

**Fix**: Add  constant. Each signal warp now iterates over 2 experts sequentially. Since signal work runs concurrently with communication work and is not on the critical path, the additional loop overhead per warp is negligible.

**Effect**: Signal SM count halved (8→4 for 256 experts on H100), freeing 4 SMs (+3.2%) for communication. At H100-scale configs (256 experts, 8 GPUs), this translates to lower latency on the signal-limited EP dispatch path.

## Changes

- max/kernels/src/shmem/ep_comm.mojo

## Test Plan

- Builds successfully for , , and full  package
-  - PASSED
-  - PASSED

### DIFF
--- a/max/kernels/src/shmem/ep_comm.mojo
+++ b/max/kernels/src/shmem/ep_comm.mojo
@@ -1175,9 +1175,14 @@ struct EPDispatchKernel[
     comptime hid_dim = Self.token_fmt_type.hid_dim
     comptime msg_bytes = Self.token_fmt_type.msg_size()
 
-    # Aux SMs for dispatch_async kernel: one SM handles n_warps experts for
-    # monitoring.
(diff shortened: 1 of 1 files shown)


prsabahrami:

Issue #6198, opened 2026-03-18
[BUG] [GPU][SM120] sink-aware flash attention is incorrect on RTX 5090: test_flash_attention_sink_kernel returns 1.0 instead of ~0.3013


rolson24:
### Bug description

 On an SM120 consumer GPU (RTX 5090), the sink-aware flash-attention path is not numerically correct on the current main-style attention code.

  The clearest signal is the existing kernel test:

  - //max/kernels/test/gpu/nn:test_flash_attention.mojo.test
  - specifically test_flash_attention_sink_kernel

  This test constructs a case where:

  - QK logits are forced to exactly 0
  - V is all 1
  - each head has a positive sink logit
  - so the output should equal the probability mass assigned to real keys, not 1.0

  For head 0:

  - expected mass is about 0.3012996
  - observed value on this machine is 1.0

  That indicates sink normalization is being handled incorrectly on this GPU path. In practice, this was part of a larger GPT-OSS decode correctness problem on SM120-class hardware.

  This appears to be specific to the sink/ragged attention path on affected NVIDIA consumer Blackwell GPUs, not a general flash-attention failure. I have found that by falling back to the naive kernel it fixes this issue.

### Steps to reproduce

  Run:

  ./bazelw --batch test //max/kernels/test/gpu/nn:test_flash_attention.mojo.test --test_output=errors

  Observed failure includes:

  test_flash_attention_sink_kernel
  Unhandled exception caught during execution: At max/kernels/test/gpu/nn/test_flash_attention.mojo:889:32: AssertionError: 1.0 is not close to 0.3012996 with a diff of 0.6987004

  This is coming from the sink-aware kernel validation in:

  - max/kernels/test/gpu/nn/test_flash_attention.mojo:714

  Expected behavior:

  - sink-aware flash attention should return approximately the analytically expected probability mass for real keys
  - in this case, head 0 should be about 0.3012996, not 1.0

### System information

```shell
MAX: nightly
GPU: RTX 5090 (SM120)
CUDA: 13.0
```

rolson24:
This can either be fixed by falling back to the naive kernel or we can try to make an implementation of the correct flash attention sink kernel

PR #6209, opened 2026-03-19
[Kernels] Sm120 attention correctness

rolson24:
## Summary

Fix ragged and decode attention behavior on SM120-class GPUs, including sink-aware softmax handling and fused RoPE correctness in the live attention path.

This keeps the changes focused on kernel correctness and the associated GPU test gating.

This fixes issue #6198 

## Testing

Ran:
`./bazelw --batch test //max/kernels/test/gpu/nn:test_flash_attention.mojo.test --test_output=errors`

## Checklist

      sequence of smaller PRs
  Note: flash_attention[sink=True] receives a non-owning view of the sink weights (sinks_device.get_immutable()), while the actual owner is sinks_dev. Because the launch is async, the test must keep sinks_dev alive until the GPU work and the output readback are complete; otherwise the sink buffer can be reclaimed too early and the test becomes flaky/incorrect. The final ctx.synchronize() is still needed separately to ensure the D2H copy into out_ptr has completed before reading out_host. We could change the API of the flash attention kernel, but I didn't want to do that yet.

Assisted-by: Codex

### DIFF
--- a/max/kernels/src/nn/fused_qk_rope.mojo
+++ b/max/kernels/src/nn/fused_qk_rope.mojo
@@ -45,10 +45,20 @@ def rope_value[
     freq_dtype: DType,
     width: Int,
 ](val: SIMD[dtype, width], freq: SIMD[freq_dtype, width]) -> SIMD[dtype, width]:
-    x_re, x_im = val.cast[freq_dtype]().deinterleave()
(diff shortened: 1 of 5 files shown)


rolson24:
This is on the critic path to getting GPT-OSS working correctly on SM120. See #6216

PR #6207, opened 2026-03-19
[Kernels] Increase reduce_kernel default BLOCK_SIZE from 128 to 256

prsabahrami:
PRAGMA-guided optimization of the GPU reduction kernel default block size.

## Summary

Increase the default  from 128 to 256 for the GPU reduction kernel.

**Problem**: For latency-bound reduction workloads with few rows relative to SM count (common in LLM decode/generation), the default BLOCK_SIZE of 128 resulted in poor SM occupancy (~12% on H100 with 256 rows). With only 4 warps per block and limited block count, most of the GPU was idle.

**Fix**: Increasing BLOCK_SIZE to 256 (8 warps per block) doubles the active warps per SM, improving occupancy from ~12% to ~24% and saturating the streaming multiprocessors on common LLM shapes.

## Benchmark Results (H100, MAX nightly 26.3.0.dev2026031822)

| Shape | Axis | Dtype | Baseline | Optimized | Speedup |
|---|---|---|---|---|---|
| (1, 256, 4096) | 2 | float16 | 42.3 us | 26.4 us | **1.6x** |
| (8, 256, 4096) | 2 | float16 | 48.1 us | 31.5 us | **1.53x** |
| (1, 2048, 4096) | 2 | float16 | 91.2 us | 85.6 us | 1.07x |

The optimization specifically targets the latency-bound path (small M, large N along reduction axis). Large reduction workloads are unaffected.

## Changes

- mojo/stdlib/std/algorithm/backend/gpu/reduction.mojo: default BLOCK_SIZE 128 -> 256

## Test Plan

-  - PASSED
-  - PASSED

Co-Authored-By: modular-kernel-agent <modular@speedtrain.co>

PR #6215, opened 2026-03-19
[Kernels] Vectorize loads in reduce_kernel (128-bit for contiguous dim)

prsabahrami:
PRAGMA-guided optimization of reduce_kernel memory access pattern.

## Summary

When reducing the contiguous (last) dimension in row-major layout, use 128-bit (16-byte) vectorized loads instead of scalar simd_width loads. This matches the optimal GPU memory transaction size on NVIDIA H100.

For fp16: load_width increases from 2 to 8 elements per transaction.
For fp32: load_width increases from 1 to 4 elements per transaction.

## Changes

Adds a  parameter to  that defaults to  (preserving existing behavior for non-contiguous axes). When , loaded vectors are reduced inline before accumulation using a tree reduction.

In , automatically selects  when  (contiguous reduction).

## Performance

H100 benchmark (1x1024x3072 fp16 axis=2):
- Baseline (BLOCK_SIZE=128): 16.40 us, 63.8 GElems/s
- After BLOCK_SIZE=256 (PR #6207): 10.78 us (1.52x vs orig)
- After vectorized loads (this PR): 7.08 us (2.32x vs orig, 1.52x vs PR #6207)

Note: This PR builds on PR #6207 (BLOCK_SIZE change). The speedup here (1.52x over PR #6207) comes entirely from better memory coalescing - each thread now issues one 128-bit load per iteration instead of one 16-bit load.

## Test Plan

./bazelw test //mojo/stdlib/test/algorithm:test_reduce

### DIFF
--- a/mojo/stdlib/std/algorithm/backend/gpu/reduction.mojo
+++ b/mojo/stdlib/std/algorithm/backend/gpu/reduction.mojo
@@ -42,7 +42,7 @@ from std.utils import IndexList
 from std.utils.numerics import get_accum_type
 from std.utils.static_tuple import StaticTuple
 from std.sys import get_defined_int
-from std.sys.info import simd_width_of
+from std.sys.info import bit_width_of, simd_width_of
 
 
 @always_inline
@@ -274,6 +274,7 @@ def row_reduce[
     simd_width: Int,
     rank: Int,
     accum_type: DType = get_accum_type[dtype](),
+    load_width: Int = simd_width,
 ](
     mut row_coords: IndexList[rank],
     axis: Int,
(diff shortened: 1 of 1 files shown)


prsabahrami:

prsabahrami:
Pushed a fix for the load_width parameter placement (was in single-reduction overload, moved to multi-reduction overload). 

Note on the vectorized reduction: the reduce_kernel is always launched with packing_factor=1 (simd_width=1) from reduce_launch, so the reduction path SIMD[accum_type, 1](scalar) is correct for the actual use case. If this optimization is ever extended to simd_width > 1, the reduction logic would need updating.

PR #6224, opened 2026-03-19
[Kernels][GPU] Reduce MLA prefill BM to 32 for better SM utilization

prsabahrami:
[Kernels][GPU] Reduce MLA prefill BM to 32 for better SM utilization

BEGIN_PUBLIC
[Kernels][GPU] Reduce MLA prefill BM to 32 for better SM utilization

Reduce the MLA prefill query block size from 64 to 32 on NVIDIA GPUs.
This halves shared memory usage per block and doubles the number of
thread blocks for small sequence lengths, improving SM occupancy.
Combined with BK=64, this yields ~3% improvement in prefill latency.
END_PUBLIC


prsabahrami:

PR #6227, opened 2026-03-19
[Kernels][GPU] Fix shared memory bound calculation in MHA config

prsabahrami:
[Kernels][GPU] Fix shared memory bound calculation in MHA config

BEGIN_PUBLIC
[Kernels][GPU] Fix shared memory bound calculation in MHA config

The BN tile size calculation for SM90 FA3 used depth instead of
padded_depth when computing the shared memory upper bound. For
non-power-of-two depths like 72 (padded to 128), this underestimated
shared memory usage, allowing BN values that exceed available shared
memory when combined with smaller BM values.
END_PUBLIC

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/max/kernels/mha/profiling_config.yaml
+++ b/max/kernels/mha/profiling_config.yaml
@@ -1,0 +1,25 @@
+# Nsight Compute Profiling Configuration
+# Kernel: mha
+# Target: NVIDIA H100 (SM90)
+
+profiling:
+  tool: ncu-cli
+  sections:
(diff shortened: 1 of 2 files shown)


prsabahrami:

prsabahrami:
Closing per policy: draft PRs must not be opened without prior approval. Will re-open after getting explicit approval.

PR #6237, opened 2026-03-19
[Kernels][GPU] Reduce MLA decode BM to 16 on H100 for better SM utilization

prsabahrami:
On H100, the MLA decoding kernel was using BM=32, resulting in only num_heads/32 CTAs in the Y grid dimension. With BM=16, we double the number of CTAs, improving SM utilization and reducing latency for workloads with moderate head counts.

Benchmark Results (H100, bf16, 128 heads):
- ~9% improvement in decode latency on default MLA configuration

This change adds  to the preferred_BM condition in , so H100 uses BM=16 instead of BM=32.


prsabahrami:

prsabahrami:
Closing to fix unintended changes (sed replaced too many occurrences). Will reopen with corrected diff.

PR #6244, opened 2026-03-19
[Kernels] Increase broadcast block size from 256 to 1024

prsabahrami:
[Kernels] Increase broadcast block size from 256 to 1024

BEGIN_PUBLIC
[Kernels] Increase broadcast block size from 256 to 1024

Increase BLOCK_SIZE from 256 to 1024 for both the 1-stage and 2-stage
broadcast kernels. With 256 threads/block and a typical grid of ~216
blocks on 132 SMs, each SM runs at most 2 blocks (512 threads) which
is only ~25% occupancy. With 1024 threads/block, occupancy rises to
~80%, improving memory latency hiding for this bandwidth-bound kernel.

The broadcast barriers only use the first ngpus threads per block, so
larger blocks do not increase synchronization overhead.

Improves broadcast bandwidth by ~20%.
END_PUBLIC

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/max/kernels/broadcast/profiling_config.yaml
+++ b/max/kernels/broadcast/profiling_config.yaml
@@ -1,0 +1,25 @@
+# Nsight Compute Profiling Configuration
+# Kernel: broadcast
+# Target: NVIDIA H100 (SM90)
+
+profiling:
+  tool: ncu-cli
+  sections:
+    - SpeedOfLight
+    - Occupancy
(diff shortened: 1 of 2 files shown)


prsabahrami:

prsabahrami:
Closing per policy: draft PRs must not be opened without prior approval. Will re-open after getting explicit approval.

PR #6250, opened 2026-03-21
[Stdlib] Optimize `StringSlice[byte=]` subscript performance

msaelices:
## Summary
- Replace `debug_assert[assert_mode="safe"]` with `assert` in `__getitem__(byte=)` using a t-string message. The default `ASSERT_MODE` is `"safe"`, so `assert_mode="safe"` asserts were running unconditionally. Plain `assert` desugars to `debug_assert` with `assert_mode="none"`, which is compiled out in the default build mode.
- Add ASCII fast path to skip `_utf8_first_byte_sequence_length` when byte < 128
- Add `@always_inline` to `__getitem__(byte=)`

## Benchmark results (byte= subscript, main vs optimized)

System: Intel i7-12700H, 64GB RAM, Linux 6.17.0. Verified via `strings` that the safe assert is compiled in (main) vs compiled out (optimized). Config: 5 repetitions, 1000 max iterations.

16KB text:

| Benchmark | main (ms) | optimized (ms) | Speedup |
|---|---|---|---|
| `byte=` ASCII (EN) | 0.145 | 0.081 | **1.8x** |
| `byte=` UTF-8 (RU) | 0.053 | 0.072 | ~same |

1KB text:

| Benchmark | main (ms) | optimized (ms) | Speedup |
|---|---|---|---|
| `byte=` ASCII (EN) | 0.011 | 0.005 | **2.0x** |
| `byte=` UTF-8 (RU) | 0.008 | 0.006 | ~same |

Assisted-by: AI

### DIFF
--- a/mojo/stdlib/std/collections/string/string_slice.mojo
+++ b/mojo/stdlib/std/collections/string/string_slice.mojo
@@ -1132,7 +1132,10 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
         )
         var b = self._slice.unsafe_get(normalized_idx)
 
-        assert _is_utf8_start_byte(b), t"String slice index {normalized_idx} does not lie on a codepoint boundary."
+        assert _is_utf8_start_byte(b), (
+            t"String slice index {normalized_idx} does not lie on a codepoint"
+            t" boundary."
+        )
 
         # ASCII fast path: sequence length is always 1 for bytes < 128.
         var seq_len = 1 if b < 128 else _utf8_first_byte_sequence_length(b)
--- a/mojo/stdlib/std/collections/string/string_slice.mojo
+++ b/mojo/stdlib/std/collections/string/string_slice.mojo
@@ -1109,6 +1109,7 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
         """
         return self.codepoint_slices()
 
+    @always_inline
     def __getitem__[I: Indexer, //](self, *, byte: I) -> Self:
         """Gets a single byte at the specified byte index.
 
@@ -1129,19 +1130,19 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
         var normalized_idx = normalize_index["StringSlice"](
             byte, self.byte_length()
         )
-        # _utf8_first_byte_sequence_length also checks for this, but
-        # we want subscripting to check unconditionally.
-        debug_assert[assert_mode="safe"](
-            _is_utf8_start_byte(self._slice.unsafe_get(normalized_idx)),
+        var b = self._slice.unsafe_get(normalized_idx)
+
+        debug_assert(
+            _is_utf8_start_byte(b),
             "String slice index, ",
             normalized_idx,
             " does not lie on a codepoint boundary.",
         )
+
+        # ASCII fast path: sequence length is always 1 for bytes < 128.
+        var seq_len = 1 if b < 128 else _utf8_first_byte_sequence_length(b)
         return StringSlice(
-            ptr=self.unsafe_ptr() + normalized_idx,
-            length=_utf8_first_byte_sequence_length(
-                self._slice.unsafe_get(normalized_idx)
-            ),
+            ptr=self.unsafe_ptr() + normalized_idx, length=seq_len
         )
 
     def __contains__(self, substr: StringSlice) -> Bool:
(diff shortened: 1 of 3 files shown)


mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -1129,19 +1130,17 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
         var normalized_idx = normalize_index["StringSlice"](
             byte, self.byte_length()
         )
-        # _utf8_first_byte_sequence_length also checks for this, but
-        # we want subscripting to check unconditionally.
-        debug_assert[assert_mode="safe"](
```
NathanSWard, line 1134: This is intentionally a `assert_mode="safe"`. We want to be aggressive with asserting here if the byte index is not valid.
If a user really want to bypass these safety checks, they should just use `.as_bytes().unsafe_get(idx)` instead.

mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -1129,19 +1130,17 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
-            " does not lie on a codepoint boundary.",
+        var b = self._slice.unsafe_get(normalized_idx)
+
+        assert _is_utf8_start_byte(b), (
+            t"String slice index {normalized_idx} does not lie on a codepoint"
+            t" boundary."
         )
+
+        # ASCII fast path: sequence length is always 1 for bytes < 128.
+        var seq_len = 1 if b < 128 else _utf8_first_byte_sequence_length(b)
```
NathanSWard, line 1141: This doesn't seem to actually produce better codegen. The `_utf8_first_byte_sequence_length` is very optimal and when you look at the lowered codegen - the IR is essentially identical. The version w/out the branch that only uses `_utf8_first_byte_sequence_length` actually produces clearer codegen. And we actually got lucky here since this branch would actually introduce overhead, but LLVM is able to see through and remove this branch.

NathanSWard (changes requested):
@msaelices - appreciate you taking a look at this, but I don't believe this PR is what we want. Since we _do_ want the `debug_assert` to be firing in "safe" mode since this is a "safe" function.

The extra ASCII branch trick also doesn't seem to actually improve performance at all. The benchmarks are only comparing the lack of `assert` checks that the byte is on a valid byte index and not the logic of getting the string length.

mojo/stdlib/std/collections/string/string_slice.mojo
msaelices: You are right. I should have compared it before you need to do it. Sorry about it.

*  with fast path (return range 0..9):
```
  %2 = icmp sgt i8 %0, -65        ; _is_utf8_continuation_byte check
  %3 = icmp sgt i8 %0, -1         ; b < 128 (ASCII check)
  br i1 %3, label %8, label %4    ; branch on ASCII
  %5 = icmp samesign ult i8 %0, -11  ; b <= BIGGEST_UTF8_FIRST_BYTE
  br i1 %5, ...                    ; assert branch
  br i1 %2, ...                    ; continuation byte assert
  %9 = xor i8 %0, -1
  %10 = ctlz.i8(%9, false)        ; count_leading_zeros (is_zero_poison=false)
  %narrow = select i1 %3, i8 1, i8 %10  ; select 1 if ASCII, else ctlz result
  %11 = zext %narrow to i64
  ret i64 %11
```
*  without_fast_path (return range 0..8):
```
  %2 = icmp ult i8 %0, -11        ; b <= BIGGEST_UTF8_FIRST_BYTE
  br i1 %2, ...                    ; assert branch
  %4 = icmp sgt i8 %0, -65        ; continuation byte check
  br i1 %4, ...                    ; assert branch
  %7 = xor i8 %0, -1
  %8 = ctlz.i8(%7, true)          ; count_leading_zeros (is_zero_poison=true)
  %9 = icmp sgt i8 %0, -1         ; b < 128
  %10 = zext i1 %9 to i8          ; cast bool to i8 (1 or 0)
                                   ; or with ctlz result
  %12 = zext %11 to i64
  ret i64 %12
```
The fast path produces worse codegen. TIL the `is_zero_poison` effect on optimizing IR. Confirmed with `compile_info[..., emission_kind="llvm-opt"]`. 

Sorry about this. I should have checked the IR before submitting. Will make sure to verify IR before   submitting optimizations next time.  I'm actually collecting many of the pitfalls and patterns in a new `mojo-stdlib-contributing`, just in case you find it useful: https://github.com/modular/skills/pull/3/

mojo/stdlib/std/collections/string/string_slice.mojo
msaelices: Got it. The reason I submitted this was an unexpected performance regression in mojo-regex after upgrading to 0.26.2, where the compiler required switching to byte= subscript. I [fixed](https://github.com/msaelices/mojo-regex/commit/67710b4c4778f1171390c85269d9a5b3270ced59) it using raw pointers (equivalent to `.as_bytes().unsafe_get()`), but assumed the overhead was a stdlib performance bug rather than an intentional safety check.

The `byte=` subscript is ~12x slower than `.as_bytes()` access due to the UTF-8 boundary check. Ok, I get this is by design, but may surprise users migrating byte-scanning code. Could we add a note to the `__getitem__(byte=)` docstring pointing to `.as_bytes()` for performance-sensitive loops? or maybe it could be in the changelog.

NathanSWard: Adding it to the doc string for the method would be great!

msaelices: Thanks! done: https://github.com/modular/modular/pull/6251

Closing this PR...

PR #6251, opened 2026-03-22
[Stdlib] Document `StringSlice[byte=]` subscript performance in docstring

msaelices:
## Summary
Add a note to `StringSlice.__getitem__(byte=)` docstring explaining that the UTF-8 boundary check runs in release builds, and pointing users to `as_bytes()` for performance-sensitive byte-scanning loops.

Assisted-by: AI


NathanSWard:
!sync

PR #6256, opened 2026-03-23
[Pipelines] Add Mamba SSM architecture implementation

ulmentflam:
[Pipelines] Add Mamba SSM architecture implementation

## Summary
- Adds complete Mamba (Selective State Space Model) architecture for MAX, enabling inference with models like state-spaces/mamba-130m-hf
- Includes model definition, config, weight adapters, tokenizer, SSM cache, and functional ops (10 new files in max/python/max/pipelines/architectures/mamba/)
- Registers MambaForCausalLM in the pipeline registry
- Adds ArchConfigWithSSMCache protocol interface for SSM-based models
- Adds state_space Bazel alias needed by custom extensions
- Config follows upstream Dataclass + initialize() → initialize_from_config() + finalize() pattern (matching Llama3Config)
- Model uses dual compiled graphs (prefill + step) since SSM state must be carried forward incrementally 

## Dependencies

- Requires framework changes from the serving PR (mamba-serving-framework) for the model to actually run, but this PR is self-contained code that builds independently
- Requires smoke and pipeline tests that are also included separately 

## Checklist

      sequence of smaller PRs
- [x] I added or updated tests to cover my changes [In another PR as requested by @BradLarson]


max/python/max/pipelines/lib/interfaces/arch_config.py
```diff
@@ -84,6 +84,20 @@ def get_kv_params(self) -> KVCacheParamInterface:
         """KV cache parameters to use when running the model."""
 
 
+@runtime_checkable
+class ArchConfigWithSSMCache(ArchConfig, Protocol):
```
BradLarson: For now, could we move this into mamba/model_config.py? That way, it'll minimize the footprint of the change for this one model, and then we can generalize it out to other state-space models when we start working on those. That may also help handle the cyclic dependency and allow for typing on the output of `get_ssm_cache_params()`.

ulmentflam: Sounds good!

BradLarson:
!sync

BradLarson:
!sync

PR #6270, opened 2026-03-24
[stdlib] Graduate the `bit._mask` module into the public

martinvuyk:
Graduate the `bit._mask` module into the public.

This has proved that it is useful in several instances. I already have several use-cases outside the stdlib, so I'd rather this be made public.


JoeLoser (approved):

JoeLoser:
!sync

PR #6281, opened 2026-03-27
[Kernels] Fix argmax/argmin input shape validation

cuiweixie:
## Summary

Fixes the `_argn` dimension validation so the input tensor is compared against the output tensor per subaxis. Previously `output.dim` was used for both sides, so mismatched input shapes were not detected.

## Change

- `input_subaxis`: use `input.dim(subaxis)` instead of `output.dim(subaxis)`.

BEGIN_PUBLIC
[Kernels] Fix argmax/argmin input shape validation

The _argn dimension check used output.dim for both tensors when validating that input and output shapes match outside the reduced axis. Use input.dim for the input side so incorrect input shapes are rejected as intended.
END_PUBLIC

### DIFF
--- a/max/kernels/src/nn/argmaxmin.mojo
+++ b/max/kernels/src/nn/argmaxmin.mojo
@@ -55,7 +55,7 @@ def _argn[
 
     comptime for subaxis in range(rank):
         var output_subaxis = output.dim(subaxis)
-        var input_subaxis = output.dim(subaxis)
+        var input_subaxis = input.dim(subaxis)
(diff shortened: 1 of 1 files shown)


conorbronsdon (approved):
Good fix — `output.dim` was compared against itself instead of `input.dim`. Build failure is unrelated (pre-existing `multi_version_tests_3.14` parse error). A regression test for shape mismatch would be a nice follow-up but not blocking. Sign the CLA (see the failing GitHub Actions check) and this is ready to merge.

PR #6282, opened 2026-03-27
[MAX] Fix MinPSampler inline comment (min_p)

cuiweixie:
## Summary

Corrects a misleading comment in `MinPSampler.forward`: it referred to `top_p` while the parameter is `min_p`.

## Change

- Single-line comment fix in `max/python/max/nn/sampling/min_p.py`.

BEGIN_PUBLIC
[MAX] Fix MinPSampler inline comment (min_p)

The MinPSampler forward path documents min_p handling; the comment incorrectly referred to top_p. Align the comment with the actual parameter name.
END_PUBLIC

### DIFF
--- a/max/python/max/nn/sampling/min_p.py
+++ b/max/python/max/nn/sampling/min_p.py
@@ -47,7 +47,7 @@ class MinPSampler(Module):
         self, input: TensorValue, min_p: TensorValueLike = 0.0
     ) -> TensorValue:
         batch_size = input.shape[0]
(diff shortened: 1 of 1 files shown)


conorbronsdon (approved):
Comment fix looks right. There's one more `top_p` leftover in the error message on line 63 — same file, the ValueError string says `"top_p tensor shape ..."` but should be `"min_p tensor shape ..."`. Commit that fix too and once you've signed the CLA (the failing check in GitHub Actions), this is good to merge.

Issue #6177, opened 2026-03-15
[Stdlib] Add capitalize() and title() to String and StringSlice


msaelices:
## Motivation

`lower()` and `upper()` already exist on `String` and `StringSlice`. `capitalize()` and `title()` are the natural complements for display formatting — they are to mixed-case what `lower`/`upper` are to all-lowercase/all-uppercase.

The current workaround requires manual byte iteration:

```mojo
# capitalize() by hand
var bytes = List[UInt8](s.as_bytes())
if bytes and bytes[0] >= ord('a') and bytes[0] <= ord('z'):
    bytes[0] -= 32
for i in range(1, len(bytes)):
    if bytes[i] >= ord('A') and bytes[i] <= ord('Z'):
        bytes[i] += 32
```

This is verbose, error-prone, and duplicated everywhere it's needed.

## Consistency argument

The stdlib already accepts:
- `lower()` — all characters lowercased
- `upper()` — all characters uppercased

`capitalize()` (first char uppercased, rest lowercased) and `title()` (each word's first char uppercased) are direct extensions of the same design. If `lower()` and `upper()` belong in the stdlib, these do too.

## Real use cases

- Formatting user-supplied names for display: `name.capitalize()`
- Generating report headers or UI labels: `column_name.replace("_", " ").title()`
- Normalizing input before comparison

## Proposed API

```mojo
def capitalize(self) -> String
def title(self) -> String
```

On `StringSlice` first, delegating from `String`. Initial implementation ASCII-only (documented), with Unicode support as a follow-up.

## Related

PR #6169 has a working implementation ready for review once this is agreed upon.

martinvuyk:
This is the kind of thing that I would love for us to implement, but if I can weigh in on the approach itself I would ask that we deviate from Python and make it return an iterator. This should also work for all unicode codepoints IMO.

I would like for us to make these other functions also return iterators: lower, upper, replace . Ideally we would make all of them be able to interop to be concatenated calls. I'll open a proposal sometime this week if I have time diving deeper into the APIs that I'd like us to change into iterators and their interop.

I think this is something that can wait until after 1.0 (mostly due to the back-log and other priorities), but it is something that would be a great addition

lsh:
[...]

PR #6293, opened 2026-03-28
[stdlib] Adding decision tree based upper and lower functions

mzaks:
## Summary

This PR introduces a much faster `to_lower` and `to_upper` functions which was discussed in #6177.
This PR is purely additive, the functions are defined as private function mainly in order to be included in the benchmarks and showcase the performance difference.

This functions can be generated from Unicode data (Unicode-Data.txt and Special-Casing.txt) which is showcased in https://github.com/mzaks/mojo-unicode if we decide to switch to this approach I suggest to add the code generation scripts to the repo as well, so that when Unicode rules for upper and lower casing changes, we can easily regenerate this functions.

## Testing

Enhanced testing was doen in https://github.com/mzaks/mojo-unicode
In this PR we introduce a very simple unit test suite and the main purpose of this PR is the change to the benchmarks.

Here is an excerpt from the run I performed on MacBook Pro Apple M4 Max

```
bench_string_lower[10], 0.00075
bench_string_upper[10], 0.001372
bench_string_lower_2[10], 1.758102766798419e-05
bench_string_upper_2[10], 1.6000000000000003e-05

bench_string_lower[30], 0.0023360000000000004
bench_string_upper[30], 0.004200000000000001
bench_string_lower_2[30], 7.599999999999999e-05
bench_string_upper_2[30], 7.599999999999999e-05

bench_string_lower[50], 0.00415
bench_string_upper[50], 0.007580000000000001
bench_string_lower_2[50], 0.0001
bench_string_upper_2[50], 0.0001

bench_string_lower[100], 0.008974
bench_string_upper[100], 0.013700000000000004
bench_string_lower_2[100], 0.000166
bench_string_upper_2[100], 0.00016

bench_string_lower[1000], 0.07053799999999999
bench_string_upper[1000], 0.14207399999999998
bench_string_lower_2[1000], 0.001328
bench_string_upper_2[1000], 0.0013479999999999998

bench_string_lower[10000], 0.6840119999999998
bench_string_upper[10000], 1.3755399999999998
bench_string_lower_2[10000], 0.014704
bench_string_upper_2[10000], 0.01471

bench_string_lower[100000], 6.71796
bench_string_upper[100000], 14.037942000000001
bench_string_lower_2[100000], 0.153844
bench_string_upper_2[100000], 0.158394

bench_string_lower[1000000], 68.13025
bench_string_upper[1000000], 144.86831
bench_string_lower_2[1000000], 1.543504
bench_string_upper_2[1000000], 1.583038
```

As you can see `bench_string_lower_2` is about 40x faster and `bench_string_upper_2` is almost 100x faster

## Checklist

      sequence of smaller PRs

Assisted-by: Claude Code


mzaks:
I think it is helpful to give a brief explanation of how the `to_lower` and `to_upper` works and how they can be generated from Unicode-Data.txt and Special-Casing.txt.

This algorithm is purely computational (no lookup tables). We compute the difference in bytes between the source UTF8 representation and the destination UTF-8 representation. To get a better feeling for the solution please have a look at following CSV files:

https://github.com/mzaks/mojo-unicode/blob/main/to_upper.csv
https://github.com/mzaks/mojo-unicode/blob/main/to_lower2.csv

The source and destination columns represent the chars, the source bytes / destination bytes, represent a space separated byte array in decimal numbers. The diff bytes column represents the difference. Important to see that in some cases the transformation from source to destination is asymmetrical, meaning the amount of bytes of a UTF-8 representation in source and destination char varies. This phenomenon is accounted for. 

The CVS files are generated from Unicode-Data.txt and Special-Casing.txt through following scripts:

https://github.com/mzaks/mojo-unicode/blob/main/gen_to_lower_csv.py
https://github.com/mzaks/mojo-unicode/blob/main/gen_to_upper_csv.py

The `to_lower` and `to_upper` can be generated from the CSV files through:

https://github.com/mzaks/mojo-unicode/blob/main/gen_to_lower_tree_v3.py
https://github.com/mzaks/mojo-unicode/blob/main/gen_to_upper_tree.py

All the Python scripts above were generated with Claude Code.

The generated `to_upper` and `to_lower` Mojo code can be validated against the CSV files see:

https://github.com/mzaks/mojo-unicode/blob/main/test_to_lower_v4.mojo
https://github.com/mzaks/mojo-unicode/blob/main/test_to_upper.mojo

As mentioned in the PR comment I would suggest to include the `Unicode-Data.txt`, `Special-Casing.txt`, `gen_to_lower_csv.py`, `gen_to_upper_csv.py`, `gen_to_lower_tree.py`, `gen_to_upper_tree.py` and the tests in a separate PR if we decide to promote the `to_lower` and `to_upper` as default implementation for `String` and `StaticString`. It is also sensible to write a small documentation file for how to prompt Claude Code or other LLM to generate the scripts.

---

Additional thought:

As I mentioned above some of the source destination transformations are asymmetric, this is why the functions return a new string. We could however provide an in-place conversion which would ignore the asymmetric cases (leaving chars which would lead to an asymmetric conversion as is).  This conversion would work very well for many use case and would be  more efficient.

mojo/stdlib/std/collections/string/_to_lower.mojo
```diff
@@ -0,0 +1,590 @@
...
+            8 ASCII bytes per iteration before falling back to per-character
+            Unicode handling.
+
+    Args:
+        s: The input string to convert.
+
+    Returns:
+        A new `String` with all characters converted to lowercase.
+    """
+    comptime SIMD_WIDTH = 8
```
martinvuyk, line 529: Question, why 8? What if we did 4 and used those loaded bytes (`vec = (p + offset).load[SIMD_WIDTH]()`) to process the rest of the code? (loading scalars when the string length doesn't allow it into the `SIMD[DTtype.uint8, 4]`)

mzaks: This is a very good question and also worth additional exploration. The Rust codebase for ASCII fast path uses 16, as this is the lower SIMD width. I tried 16, 32, 64 first then I tried 8 and it worked better than the wider ones. Specifically as I decided to keep trying the ASCII fast path on every loop, which helps with tons of European languages where ASCII is dominant with exceptions. This is why in my tests DE and LT text is much faster then Python and Rust:
(see https://github.com/mzaks/mojo-unicode/blob/main/benchmark_comparison_x86_64.png)

I haven't tried 4 bytes at a time which could be reused later even if the the read 4 bytes are non ASCII. This is an interesting experiment to run.

mzaks: Ok just tested it on M4, the optimization of reusing 4 bytes later if it is not an ASCII is actually contra productive 🤷‍♂️

soraros: Can we please document these reasonings in the code itself?

mzaks: @soraros I added a comment, is this enough or should or do we need more context?

mzaks: Fun fact, I woke up today with a realization that the computation of the diff uses subtraction / addition, so that a `Diff` needs to be `SIMD[DType.int16, 4]` and we need to do casting after we do addition, where we could use xor bit operation for diff and then all the overflow considerations and casting can be removed. So I run an experiment to see if it will improve the performance and sadly at least based on my test, there is no difference.

soraros: Thanks @mzaks, I think it's enough (I'm fine as long as we document the fact that it's a *choice* made here which is inherently contingent).

mzaks: See:

benchmark_comparison_arm64

mzaks: you can see that lower_v4 and lower_v5 are same, and upper_v2, v3 and v4 are also quite same

PR #6300, opened 2026-03-29
[MAX] Add Wan VAE and refactor autoencoder module

jglee-sqbits:
## Summary

Add a Wan VAE (3D causal video autoencoder) and restructure the autoencoder module to separate Module V2 and V3 implementations.

### Wan VAE
- Implements the Wan 3D causal VAE with temporal caching for chunked encode/decode
- Encoder: processes video in temporal chunks (first frame + subsequent chunks) with cached convolution state to maintain temporal consistency
- Decoder: same chunked approach with 3 specialized graphs (post-quant conv, first frame, subsequent frames)
- Uses symbolic spatial dims for resolution flexibility
- Adds 3D convolution support via cuDNN (`conv.mojo`) with depth-tiled execution for large volumes

### Autoencoder restructuring
- Moves existing Module V3 (Flux) autoencoder files to `autoencoders_modulev3/`
- The `autoencoders/` directory now contains Module V2 graph-based implementations (Wan VAE, Qwen Image VAE)
- Updates `flux1_modulev3` and `flux2_modulev3` import paths accordingly

This follows the same pattern as #6278 which established the V2/V3 split.

## Dependencies

Should be merged **before** #6301 (transformer) and #6302 (pipeline-t2v), which import from `autoencoders`.

## Checklist

- [x] PR is small and focused
- [x] I ran `./bazelw run format` to format my changes

Assisted-by: Claude Code

PR #6301, opened 2026-03-29
[MAX] Add Wan transformer model with block-level compilation

jglee-sqbits:
## Summary

Add the Wan DiT (Diffusion Transformer) model with block-level compilation for memory-efficient inference.

## Description

- Implements the full Wan transformer architecture: patch embedding, RoPE 3D positional encoding, self-attention, cross-attention, and adaptive LayerNorm
- Uses **block-level compilation**: each of the 40 transformer blocks is compiled as a separate graph sharing the same compiled program, so only one block's activation workspace is live at a time
- Block graphs use **symbolic seq_len** for resolution flexibility (480p/720p without recompilation)
- Pre/post processing graphs use symbolic spatial dims
- Supports diffusers weight key remapping and QKV fusion
- Includes 3D RoPE computation matching the Wan frequency schedule

### Memory strategy
The block-level approach keeps peak VRAM low (~6.5 GB per block execution vs ~18.5 GB for a single monolithic graph). This is critical for running the 14B parameter model at 720p (seq_len=75,600) on a single GPU.

**Note:** `MODULAR_DEVICE_CONTEXT_MEMORY_MANAGER_CHUNK_PERCENT=100` is required at 720p to avoid memory manager fragmentation with symbolic dims.

## Dependencies

Should be merged **after** #6300 (VAE, for autoencoder restructuring).

## Checklist

- [x] PR is small and focused

Assisted-by: Claude Code

## Stack
Depends on #6300

### DIFF
--- a/max/python/max/pipelines/architectures/wan/wan_transformer.py
+++ b/max/python/max/pipelines/architectures/wan/wan_transformer.py
@@ -1,0 +1,894 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+from __future__ import annotations
+
+from math import prod
+
+from max.dtype import DType
+from max.graph import DeviceRef, TensorValue, Weight, ops
+from max.nn.attention.mask_config import MHAMaskVariant
+from max.nn.kernels import flash_attention_gpu
+from max.nn.layer import LayerList, Module
+from max.nn.linear import Linear
+
+from .embeddings import (
+    TimestepEmbedding,
+    Timesteps,
+    apply_rotary_emb,
+)
+from .model_config import WanConfigBase
+
+
+class WanConv3d(Module):
+    """3D conv for WAN patch embedding (NDHWC/QRSCF layout)."""
+
+    def __init__(
+        self,
+        kernel_size: tuple[int, int, int],
+        in_channels: int,
+        out_channels: int,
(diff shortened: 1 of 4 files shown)


ansley:
!sync

max/python/max/pipelines/architectures/wan/wan_transformer.py
```diff
@@ -0,0 +1,894 @@
...
+        self.bias = (
+            Weight("bias", dtype, [out_channels], device) if has_bias else None
+        )
+        self.stride = stride
+
+    def __call__(self, x: TensorValue) -> TensorValue:
+        return ops.conv3d(x, self.filter, stride=self.stride, bias=self.bias)
+
+
+class WanLayerNorm(Module):
```
KCaverly, line 60: Why do we need float32 here?

max/python/max/pipelines/architectures/wan/wan_transformer.py
```diff
@@ -0,0 +1,894 @@
...
+    may also hit resource limits for dim=5120.
+    """
+
+    def __init__(
+        self,
+        dim: int,
+        eps: float = 1e-6,
+        *,
+        dtype: DType = DType.bfloat16,
+        device: DeviceRef = DeviceRef.CPU(),
```
KCaverly, line 115: Having defaults that are incompatible, seems confusing to me.

cc @ansley is this a common pattern within Modules?

max/python/max/pipelines/architectures/wan/wan_transformer.py
jglee-sqbits: There is a slight difference if we do not cast to `float32` in LayerNorm. However, in places like T2V, that difference gets significantly amplified, so the final output itself can end up noticeably different.

jglee-sqbits: However, this cast itself slows things down, so it is planned to be removed in the `wan-animate` job. We do see this difference in T2V for now, but we plan to treat it as a known issue and continue with parity checks later, adding back only the elements that turn out to be truly necessary.
cc @kkimmk

max/python/max/pipelines/architectures/wan/wan_transformer.py
ansley: Yeah, I agree this is kind of weird. I can just fix it internally though

ansley:
!sync

PR #6299, opened 2026-03-29
[MAX] Add UMT5 text encoder for Wan diffusion

jglee-sqbits:
## Summary

Add a MAX-native UMT5 text encoder for Wan diffusion pipelines.

## Description

- Implements the UMT5 encoder architecture using `max.nn` (Module V2 graph API)
- Supports float32 → bfloat16 weight casting via `WeightData.astype()` for Wan checkpoints that store text encoder weights in float32
- Includes T5-style relative position bias and gated GeLU feed-forward
- Handles diffusers weight key remapping (e.g. `shared.weight` alias dedup)

UMT5 is the text encoder used by all Wan 2.1/2.2 models. It produces 4096-dim text embeddings consumed by the Wan transformer.

## Dependencies

None — can be merged independently.

## Checklist

- [x] PR is small and focused
- [x] I ran `./bazelw run format` to format my changes

Assisted-by: Claude Code

PR #6298, opened 2026-03-29
[MAX] Add UniPC multistep scheduler for Wan diffusion

jglee-sqbits:
## Summary

Add a numpy-only UniPC multistep scheduler for Wan diffusion pipelines.

## Description

- Implements the UniPC-BH2 algorithm with corrector and predictor steps
- Supports flow-matching sigma schedules (used by Wan 2.1/2.2)
- Provides `build_step_coefficients()` to precompute per-step coefficient matrices on the host, enabling on-device scheduler steps without Python-side numpy calls during denoising
- Registers `UniPCMultistepScheduler` in the diffusion scheduler factory

This is a numpy-only port of the diffusers `UniPCMultistepScheduler`, specialized for the Wan pipeline configuration.

## Dependencies

None — can be merged independently.

## Checklist

- [x] PR is small and focused
- [x] I ran `./bazelw run format` to format my changes

Assisted-by: Claude Code

PR #6302, opened 2026-03-29
[MAX] Add Wan T2V diffusion pipeline with MoE support

jglee-sqbits:
## Summary

Add the Wan text-to-video (T2V) diffusion pipeline with MoE (Mixture of Experts) dual-transformer support.

## Description

- Implements the full Wan T2V pipeline: text encoding → latent preparation → denoising loop → VAE decode
- Supports **Wan 2.2 MoE models** (A14B) with dual transformers: high-noise expert for early steps, low-noise expert for later steps, with configurable boundary timestep
- Supports **Wan 2.1 single-transformer models** (14B) with the same code path
- LoRA support with automatic download from HuggingFace (e.g. Lightning turbo LoRAs for 4-step generation)
- Classifier-free guidance with batched forward pass (positive + negative in one call)
- On-device UniPC scheduler steps via compiled graphs — no Python-side numpy during denoising
- Architecture registration for `Wan-AI/Wan2.2-T2V-A14B-Diffusers`, `Wan-AI/Wan2.1-T2V-14B-Diffusers`, etc.
- Adds `guidance_scale_2` field to `VideoProviderOptions` for MoE boundary guidance control
- Minimal upstream changes: only `_weight_paths` storage in `DiffusionPipeline.__init__` and Wan registration in `pixel_tokenizer.py` / `registry.py`

## Dependencies

Depends on all previous PRs: #6298 (scheduler), #6299 (UMT5), #6300 (VAE), #6301 (transformer).

## Checklist

- [x] PR is small and focused

Assisted-by: Claude Code


ansley:
!sync

KCaverly (commented):
Taking a look at this PR, can we break this out into smaller PRs.
To start with, I don't think this is a pattern we want to emulate for the PixelGenerationTokenizer.

We expose the `SupportedArchitecture.tokenizer` class specifically in arch.py files, so that model specific implementations can override default behaviour when necessary. This is common in other parts of our code base, the [Qwen Tokenizer](https://github.com/modular/modular/blob/main/max/python/max/pipelines/architectures/qwen2_5vl/tokenizer.py) is an example. I do not think we want to continue to add forking inside the PixelGenerationTokenizer as we introduce new image/video variants. Instead, we should be able to implement a `WanTokenizer` which is co-located wihin the wan archietcture code, that extends the `PixelGenerationTokenizer` as needed. `PixelContext` follows the same pattern, please extend the `PixelContext` and identify the `WanContext` on the architecture itself.

To start, can we pull the tokenizer changes out of this PR, and implement a `WanTokenizer` as discussed above. A good rule of thumb for this, is if this code is changing elements in `pipelines/core` or our code `lib` it is likely not the right pattern. While we are bringing up new modalities, we will likely have to make some changes to core. But these should be relatively minor, and architecture agnostic. If you find you are making model specific changes outside of the architectures directory, it is likely the wrong pattern.

If there are any further questions on this, please let me know.

jglee-sqbits:
@KCaverly
I think we can split this PR into smaller ones as you suggested, separating the tokenizer changes from the pipeline changes. I also agree that continuing to add model-specific logic whenever we introduce a new model does not seem like the right long-term pattern. Now that we have accumulated enough examples compared to before, I think we are in a better position to come up with a cleaner structure.

I’d like to spend a bit of time thinking through the structure properly, taking future model integrations into account, and then propose it together with the tokenizer PR.

Issue #4878, opened 2025-06-21
[Feature Request] Allow passing arguments when the function operates in parameter space


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

```mojo
trait StaticUIntSized:
    @staticmethod
    fn __len__() -> UInt:
        ...


fn len[T: StaticUIntSized](value: T) -> UInt:
    alias length = T.__len__()
    return length


@fieldwise_init
struct SomeTypeLikeVariadicOrTuple(StaticUIntSized):
    @staticmethod
    fn __len__() -> UInt:
        return 1


fn main():
    var a = SomeTypeLikeVariadicOrTuple()

    @parameter
    for i in range(len(a)):
        print(i)
```
error:
```text
 error: cannot use a dynamic value in call argument
    for i in range(len(a)):
                       ^
```

### What is your motivation for this change?

This blocks us from having nice things like `len(args)` for variadic packs and tuple in `@parameter for` loops (see  #4877).

### Any other details?

I proposed this in #3756 by requesting a `@uses_arguments_statically` decorator that should bypass this compiler blocking feature. But I don't know whether that is the best approach or not.

soraros:
CC @dgurchenkov.

soraros:
Another interesting (motivating?) example usage would be:

```mojo
alias type[T: AnyType, //, x: T] = __type_of(x)
```

It looks nicer than the intrinsic, and closer to Python's `type(val)`.

soraros:
If you restrict the scope to methods, this is kinda a Python-style `@classmethod`: it take `cls = __type_of(self)` as the first argument.

dgurchenkov:
Looking at this holistically, supporting passing runtime arguments into functions that operate in the parameter space, this is hard. The runtime argument does not exist during comptime, it cannot be given to the interpreter as a value.

What Martin really wants (which is a very reasonable thing!) is to be able to comptime-eval function calls where the input is not really used as a value. This would be trivial if the Mojo compiler was built on partial evaluation, but it is not — it eagerly constructs values and each value is concrete. 

[...]

PR #6309, opened 2026-03-30
[stdlib] Trivial cleanups to `Index`

soraros:
## Summary

`__len__` is a `@staticmethod` on `VariadicPack`, there's no need to use `type_of(...).__len__()`.


soraros:
I'd really help if we could have a decorator like `@callable_at_comptime`, so we get to write simply `len(args)` (useful for `Tuple` and a few other types as well). Related: #4878.

JoeLoser (approved):

JoeLoser:
!sync

JoeLoser:
!sync

PR #6313, opened 2026-03-30
[Stdlib] Add `UnsafePointer` pointer subtraction to compute element distance

### LINKED ISSUE #5179: [Feature Request] Support `UnsafePointer[T].__sub__(UnsafePointer[T])`

bro4all:
## Summary
- Add `UnsafePointer.__sub__(UnsafePointer) -> Int` so subtracting two pointers of the same type and address space returns a signed distance in elements
- Compute the result as `(Int(self) - Int(rhs)) // size_of[Self.type]()`
- Leave existing pointer-minus-index behavior unchanged

Closes #5179

## Test plan
- Added tests in `test_unsafe_pointer_v2.mojo` covering same-pointer subtraction, positive/negative related-pointer offsets, and existing pointer-minus-index arithmetic
- `./bazelw test //mojo/stdlib/test/memory/unsafe_pointer/...`
- `./bazelw run format`

## Notes
- `./bazelw build //mojo/stdlib/...` and `./bazelw test //mojo/stdlib/test/...` hit unrelated local Metal compiler failures outside this patch (`mojo/stdlib/test/algorithm/gpu/test_binary_apply.mojo` and `mojo/stdlib/test/asyncrt/test_function.mojo`)

### DIFF
--- a/mojo/stdlib/test/memory/unsafe_pointer/test_unsafe_pointer_v2.mojo
+++ b/mojo/stdlib/test/memory/unsafe_pointer/test_unsafe_pointer_v2.mojo
@@ -403,6 +403,9 @@ def test_offset() raises:
     var y = Int(4)
     assert_equal((ptr + x)[], 3)
     assert_equal((ptr + y)[], 4)
+    assert_equal(ptr - ptr, 0)
+    assert_equal((ptr + 4) - ptr, 4)
+    assert_equal(ptr - (ptr + 4), -4)
 
     var ptr2 = alloc[Int](5)
     var ptr3 = ptr2
--- a/mojo/stdlib/test/memory/unsafe_pointer/test_unsafe_pointer_v2.mojo
+++ b/mojo/stdlib/test/memory/unsafe_pointer/test_unsafe_pointer_v2.mojo
@@ -403,9 +403,6 @@ def test_offset() raises:
     var y = Int(4)
     assert_equal((ptr + x)[], 3)
     assert_equal((ptr + y)[], 4)
-    assert_equal(ptr - ptr, 0)
-    assert_equal((ptr + 4) - ptr, 4)
-    assert_equal(ptr - (ptr + 4), -4)
 
     var ptr2 = alloc[Int](5)
     var ptr3 = ptr2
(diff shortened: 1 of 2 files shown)


mojo/stdlib/test/memory/unsafe_pointer/test_unsafe_pointer_v2.mojo
```diff
@@ -403,6 +403,9 @@ def test_offset() raises:
     var y = Int(4)
     assert_equal((ptr + x)[], 3)
     assert_equal((ptr + y)[], 4)
+    assert_equal(ptr - ptr, 0)
```
NathanSWard: Would you mind moving these into their own unit test please?
Ideally each unit tests only tests a single thing.
```
def test_sub_same_type():
     # ... test code
```

NathanSWard (changes requested):
thank you for this!

Do you mind moving the tests to their own unit test function please?
Other than that, the implementation looks good!

PR #6319, opened 2026-03-31
[Kernel] Reuse eager custom-op models

bro4all:
## Summary
- cache compiled eager models for graphs with custom extensions, keyed by the eager session, stable MLIR, and resolved kernel-library paths
- route the eager compile path through that cache so repeated identical `F.custom(...)` calls reuse the compiled model instead of recompiling
- add a deterministic regression test that patches `session.load()` and asserts two identical eager custom-op calls only load once

## Validation
- `./bazelw run format`
- `CUDA_VISIBLE_DEVICES= HIP_VISIBLE_DEVICES= ./bazelw test //max/tests/integration/tensor:test_functional_custom --test_output=errors`
- `./bazelw test //max/tests/integration/tensor:test_functional_custom --test_filter=test_custom_extensions_cached_across_calls --test_output=errors`
- `./bazelw test //max/tests/integration/tensor:test_functional_custom --test_filter=test_custom_with_custom_extensions --test_output=errors`

### DIFF
--- a/max/python/max/experimental/realization_context.py
+++ b/max/python/max/experimental/realization_context.py
@@ -50,8 +50,8 @@ import os
 import threading
 import weakref
 from collections import OrderedDict
+from concurrent.futures import Future
 from contextvars import ContextVar
-from pathlib import Path
 from types import TracebackType
 from typing import Any, TypeVar
 
@@ -75,11 +75,18 @@ Ex = TypeVar("Ex", bound=BaseException)
 _SESSION: ContextVar[engine.api.InferenceSession] = ContextVar("_SESSION")
 _SEED: Tensor | None = None
 _EAGER_MODEL_CACHE_MAX_SIZE = 128
-_EAGER_MODEL_CACHE_LOCK = threading.Lock()
-_EAGER_MODEL_CACHE: OrderedDict[
-    tuple[int, str, tuple[str, ...]],
-    engine.Model,
-] = OrderedDict()
+_EAGER_MODEL_CACHES_LOCK = threading.Lock()
+
+
+class _SessionModelCache:
+    def __init__(self) -> None:
+        self.lock = threading.Lock()
+        self.models: OrderedDict[object, Future[engine.Model]] = OrderedDict()
+
+
+_EAGER_MODEL_CACHES: weakref.WeakKeyDictionary[
+    engine.api.InferenceSession, _SessionModelCache
+] = weakref.WeakKeyDictionary()
 
 # Environment variable to control interpreter usage.
 # Set to "0" or "false" to disable the interpreter (always compile).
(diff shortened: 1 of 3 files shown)


max/python/max/experimental/realization_context.py
```diff
@@ -160,6 +169,43 @@ def _session() -> engine.api.InferenceSession:
...
+    session: engine.api.InferenceSession, graph: Graph
+) -> tuple[int, str, tuple[str, ...]]:
+    module_asm = graph._module.operation.get_asm(
+        assume_verified=True,
+        enable_debug_info=False,
+        pretty_debug_info=False,
+        use_local_scope=True,
+    )
+    return (
+        id(session),
```
ehsanmok: 1. Generally the `id(session)` is fragile as cache key and if the session is recreated the cache will miss but won't clear stake entries referencing the old session.
2. `get_asm` serializes the entire MLIR on evey cache lookup which can be costly for large regions also has more critical issues if any non-semantic details of the ASM changes between calls.

max/python/max/experimental/realization_context.py
```diff
@@ -160,6 +169,43 @@ def _session() -> engine.api.InferenceSession:
...
+    if not graph.kernel_libraries_paths:
+        return session.load(graph)
+
+    key = _eager_model_cache_key(session, graph)
+    with _EAGER_MODEL_CACHE_LOCK:
+        if model := _EAGER_MODEL_CACHE.get(key):
+            _EAGER_MODEL_CACHE.move_to_end(key)
+            return model
+
+        model = session.load(graph)
```
ehsanmok: session inside the lock adds unnecessary overheard and is a major issue. I recommend following [experimental.torch](https://github.com/modular/modular/blob/aa6d915d620210d372ab73d9d2658a559180439c/max/python/max/experimental/torch/torch.py#L479) pattern instead.

max/tests/integration/tensor/test_functional_custom.py
```diff
@@ -123,29 +125,44 @@ def test_custom_with_string_path(kernel_verification_ops_path: Path) -> None:
 def test_custom_extensions_cached_across_calls(
     kernel_verification_ops_path: Path,
 ) -> None:
-    """Test that custom_extensions are cached and not reloaded on every call."""
-    x = Tensor.ones([64], dtype=DType.float32, device=CPU())
-    y = Tensor.ones([64], dtype=DType.float32, device=CPU())
+    """Test that identical eager custom op calls reuse compiled models."""
+    x = Tensor.ones([65], dtype=DType.float32, device=CPU())
```
ehsanmok: This doesn't look good. Looks like needed to change to avoid collision with other test which is better to explicitly clear caches in a fixture for example.

max/tests/integration/tensor/test_functional_custom.py
```diff
@@ -123,29 +125,44 @@ def test_custom_with_string_path(kernel_verification_ops_path: Path) -> None:
...
-    # Second call - should use cached extension
-    result2 = F.custom(
-        "my_add",
-        device=CPU(),
-        values=[x, y],
-        out_types=[x.type],
-        custom_extensions=kernel_verification_ops_path,
-    )
-    assert result2[0].real
+    assert load_count == 1
```
ehsanmok, line 173: There's no test eviction key correctness with different ops or thread safety.

ehsanmok (changes requested):
Thanks for putting up this PR! It's an important gap that we need to fix however the approach here doesn't work and has major issues as list below as well as having no cache invalidation and using global mutable states that can make testing and reasoning harder.

ehsanmok:
FYI due to higher priority we're addressing this issue internally. Thanks for your help and sorry we didn't communicate it earlier.

bro4all:
> FYI due to higher priority we're addressing this issue internally. Thanks for your help and sorry we didn't communicate it earlier.

It’s ok, glad you told me.
If you guys are hiring I’m interested. 
Best,

ehsanmok:
We are. Please check the career page  https://www.modular.com/company/careers#open-roles

bro4all:
> We are. Please check the career page https://www.modular.com/company/careers#open-roles

Thanks for sharing the link. I want to be straightforward: Modular's hiring process has serious problems. I've attached a screenshot of the most recent failure in your pipeline — this despite having worked directly with multiple engineers on your team.
If you or a sister team have a role where the hiring process reflects the caliber of work you're actually trying to do, I'm reachable at ohabra.com.

Screenshot 2026-04-03 at 5 32 40 PM

PR #6322, opened 2026-03-31
[bazel] Use single clang binary on CI

keith:
CI already doesn't share caches with local builds because it uses `-g0` and some other different flags. For local builds we use dynamic execution and potentially a different exec platform with remote execution, so we need multiple clang binaries available, on CI we can be sure only 1 will be used. This saves some upload/download of all these duplicate binaries that would never be called.

### DIFF
--- a/bazel/internal/cc-toolchain/tools/BUILD.bazel
+++ b/bazel/internal/cc-toolchain/tools/BUILD.bazel
@@ -17,7 +17,7 @@ cc_tool(
 )
 
 cc_tool(
-    name = "multi-platform-clang++".format(platform),
+    name = "multi-platform-clang++",
     src = ":multi-platform-clang++.sh",
     data = [
         "@clang-linux-aarch64//:bin/clang++",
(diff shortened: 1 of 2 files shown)


bazel/internal/cc-toolchain/tools/BUILD.bazel
```diff
@@ -17,7 +17,7 @@ cc_tool(
 )
 
 cc_tool(
-    name = "clang++".format(platform),
+    name = "multi-platform-clang++".format(platform),
```
keith:
```suggestion
    name = "multi-platform-clang++",
```

bazel/internal/cc-toolchain/tools/BUILD.bazel
```diff
@@ -28,7 +28,7 @@ cc_tool(
 )
 
 cc_tool(
-    name = "linker_driver".format(platform),
+    name = "multi-platform-linker_driver".format(platform),
```
keith:
```suggestion
    name = "multi-platform-linker_driver",
```

Ahajha:
thanks keith!

Ahajha:
!sync

PR #6326, opened 2026-04-01
[stdlib] Add take_items() method to List

### LINKED ISSUE #5312: [Feature Request] Add `drain`/`take_items` API to `List`
### Review Mojo's priorities

### What is your request?

Similar to `Dict.take_items` I propose adding a `drain()` or `take_items()` to `List`.

Ensure that `zip` is compatible with this (supports mutable iterators).

I would like a returned iterator that *move*s items out of the list as we iterate through it. Practical use case is situations like in EmberJson:
```mojo
    fn __init__(out self, var keys: List[String], var values: List[Value],__dict_literal__: (),):
        debug_assert(len(keys) == len(values))
        self._data = Self.Type()
        for i in range(len(keys)):
            # We do not want to be doing copies here since values is owned
            self._data.insert(keys[i], values[i].copy()) 
```
Ideally we would rewrite this as

```mojo
    fn __init__(out self, var keys: List[String], var values: List[Value]):
        # possibly not even need this if using zip handles this.
        debug_assert(len(keys) == len(values)) 
        self._data = Self.Type()
        # Might not need the "ref" keyword
        for ref key_value in zip(keys.drain(),values.drain()):
            self._data.insert(key_value^)
```
`dict.mojo` has a similar concern:
```
    @always_inline
    fn __init__(out self, var keys: List[K], var values: List[V],  __dict_literal__: (),):
        ... # assertions
        # TODO: Should transfer the key/value's from the list to avoid copying
        # the values.
        for i in range(len(keys)):
            self._insert(keys[i].copy(), values[i].copy())
```

This covers a larger concept of converting a List -> SomeOtherType.

- [EmberJson](https://github.com/bgreni/EmberJson/blob/main/emberjson/object.mojo#L41-L51)
- [Dict](https://github.com/modular/modular/blob/main/mojo/stdlib/stdlib/collections/dict.mojo#L656-L659)
- [PythonObject](https://github.com/modular/modular/blob/main/mojo/stdlib/stdlib/collections/dict.mojo#L656)

### Any other details?

I would like to make the PR for this. Possibly break into parts:
- Introduce this new function
- Update places where lists are should have their elements moved into other data structures instead of copying. (assumes that the list is owned)

Noting other PRs related to moving elements out of collections:
- https://github.com/modular/modular/pull/5308 pop(slice)
- https://github.com/modular/modular/pull/5289 Dict.take_items

dl-alexandre:
This PR implements #5312 by adding a `take_items()` method to `List` that returns an iterator draining the list by moving elements out.

## Motivation

When transferring elements from a `List` to another data structure (like a `Dict`), users currently need to copy elements or consume the entire list with owned iteration. This PR adds `take_items()` similar to `Dict.take_items()` to enable efficient transfer without copying.

### Core Implementation
- Added `_ListTakeIter` struct that takes a mutable reference to a List
- Iterator moves elements out using `take_pointee()` and updates `_len`
- Added `List.take_items()` method returning the draining iterator

### Tests
- `test_list_take_items()` - Basic functionality
- `test_list_take_items_empty()` - Empty list edge case
- `test_list_take_items_partial()` - Partial consumption behavior
- `test_list_take_items_bounds()` - Bounds reporting

## Usage Example

```mojo
var my_list = List[String]("a", "b", "c")

for element in my_list.take_items():
    print(element[])  # prints a, b, c

print(len(my_list))  # 0 - list is drained
```

## Comparison

**Owned iteration (consumes list):**
```mojo
for x in my_list^:
    pass
# my_list no longer accessible
```

**take_items() (drains but keeps list accessible):**
```mojo
for x in my_list.take_items():
    pass
# my_list still accessible but empty
```

## Checklist
- [x] Code follows style guidelines
- [x] Tests added and passing
- [x] Documentation added (docstrings with examples)
- [x] Signed-off commits
- [x] Fixes #5312

### DIFF
--- a/mojo/stdlib/std/collections/list.mojo
+++ b/mojo/stdlib/std/collections/list.mojo
@@ -206,8 +206,9 @@ struct _ListTakeIter[
 
         # Destroy remaining elements from current index to original length
         while self._index < self._original_len:
-            (self._list[].unsafe_ptr().bitcast[TDestructible]() + self._index)
-                .destroy_pointee()
+            (
+                self._list[].unsafe_ptr().bitcast[TDestructible]() + self._index
+            ).destroy_pointee()
             self._index += 1
 
     @always_inline
--- a/mojo/stdlib/std/collections/list.mojo
+++ b/mojo/stdlib/std/collections/list.mojo
@@ -147,6 +147,54 @@ struct _ListIterOwned[T: Copyable](IterableOwned, Iterator, Movable):
         return (iter_len, {iter_len})
 
 
+@fieldwise_init
+struct _ListTakeIter[
+    T: Copyable,
+    origin: MutOrigin,
+](Copyable, Iterable, Iterator):
+    """Iterator over mutable List element references that moves elements out of the list.
+
+    This iterator drains the list as it iterates, moving elements out using
+    `take_pointee()` and updating the list's length accordingly.
+
+    Parameters:
+        T: The type of the elements in the list.
+        origin: The mutable origin of the List.
+    """
+
+    comptime IteratorType[
+        iterable_mut: Bool, //, iterable_origin: Origin[mut=iterable_mut]
+    ]: Iterator = Self
+    comptime Element = Self.T
+
+    var _index: Int
+    var _list: Pointer[List[Self.T], Self.origin]
+
+    def __init__(out self, ref[Self.origin] list: List[Self.T]):
+        self._index = 0
+        self._list = Pointer(to=list)
+
+    @always_inline
+    def __iter__(ref self) -> Self.IteratorType[origin_of(self)]:
+        return self.copy()
+
+    @always_inline
+    def __next__(mut self) raises StopIteration -> Self.Element:
+        if self._index >= len(self._list[]):
+            raise StopIteration()
+
+        # Move the element out and update the list state
+        var element = (self._list[].unsafe_ptr() + self._index).take_pointee()
+        self._index += 1
+        self._list[]._len -= 1
+        return element
+
+    @always_inline
+    def bounds(self) -> Tuple[Int, Optional[Int]]:
+        var iter_len = len(self._list[]) - self._index
+        return (iter_len, {iter_len})
+
+
 struct List[T: Copyable](
     Boolable,
     Copyable,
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/collections/list.mojo
```diff
@@ -147,6 +147,54 @@ struct _ListIterOwned[T: Copyable](IterableOwned, Iterator, Movable):
         return (iter_len, {iter_len})
 
 
+@fieldwise_init
+struct _ListTakeIter[
```
NathanSWard, line 151: Hi @dl-alexandre thanks for taking a stab at this. Looking through the code there are a few memory safety issues that need to be addressed first

1. The current way `_ListTakeIter` is implement is not totally correct. The `__next__` method will move the elements out of the `List` from index `0 -> len(self) - 1`, which is correct! However, as it stands this will cause a _double destroy_ if the entire iterator it not fully conumsed since the `List`'s destructor gets ran. 
Example:
```mojo
var l = ["a", "b", "c"]
# len(list) == 3

var iter = l.take_items()
print(iter.__next__()) 
# ^ prints ("a")
# len(list) == 2

# Compiler inserts a call to `List.__del__`
# l.__del__() 
# ^ This is the bug.
# The list's destructor now gets ran iterating from `0 -> len(self) - 1` and calls all 
# of the destructors on the elements. 
# This means that `"a"` even though it was moved out of the list in the
# iter.__next__, the list # doesn't know this so it calls `"a".__del__`
# again leading to a double free!
```

The solution here is when creating the `_ListTakeIter` we need to immediately set the `List`'s length to `0`, and then we need to have a custom `__del__` method on `_ListTakeIter` that destroys the remaining elements that were not yet iterated over.

2. `_ListTakeIter` cannot conform to `Iterable` and cannot conform to `Copayble`. but instead needs to conform to `IterableOwned`. You cannot safely copy this iterator since you could also hit a double free again. For example
```mojo
var l = ["a", "b", "c"]
var iter1 = l.__iter__()
var iter2 = iter1.copy()

print(iter1.__next__()) # prints "a"
print(iter2.__next__()) 
# ^ Issue. this will also try to print "a" but that value is already moved from 
# so this will likely cause an invalid read.
```

Let me know if these make sense or if you would like more explanation! Thank you again for taking a stab at implementing this! 😄

NathanSWard (changes requested):

mojo/stdlib/std/collections/list.mojo
NathanSWard: @dl-alexandre - I saw you updated the PR with the fixes thanks!
In the future would you please respond/resolve the PR comments so that we can get notified on our end?

Otherwise I don't see all the notification that come though and this helps make sure there is a HITL 😄

JoeLoser: @dl-alexandre it looks like you updated things based on @NathanSWard's feedback and this looks good to me, thank you! In the future, do you mind just replying to Nate's comment so he knows when the PR is ready for re-review? Thank you!

NathanSWard:
!sync

PR #6332, opened 2026-04-01
[Stdlib] Enforce non-NULL PythonObjects

### LINKED ISSUE #5414: [stdlib] Enforce non-NULL `PythonObject`s
This will avoid branching on NULL safety checks in `PythonObject.__del__()` and other methods, and help avoid segfaults and similar bugs in our low-level binding code.

There are 2 known situations where we currently must deal with NULL `PyObject*` values:

1. As the return value from a `PyCFunction` that raises an exception

2. As the kwargs argument of a `PyCFunctionWithKeywords` or `initproc`, which is NULL when no keyword arguments were specified.

For (1), we can solve this by modeling the `def_py_function()` signature's ability for a Mojo function to raise a Python exception using:

```mojo
alias PyFunction = fn(PythonObject, PythonObject) -> Variant[PythonException, PythonObject]
```

For (2), we can solve this by modeling `PyFunctionWithKeywords` in one of two ways:

1. A signature: `alias PyFunctionWithKeywords = fn(PythonObject, PythonObject, Optional[PythonObject]) -> Variant[…]`

2. By "canonicalizing" a low-level NULL kwargs: `PyObjectPtr` into an empty `Python.dict()` `PythonObject`.

c-pozzi:
## Summary

- Change `PythonObject._obj_ptr` from nullable `PyObjectPtr` to `NonNullPyObjectPtr` (`NonNullUnsafePointer[PyObject, MutAnyOrigin]`), as suggested by @soraros in #5414
- NULL enforcement happens at the `from_owned`/`from_borrowed` boundary via `debug_assert`
- `steal_data()` swaps in an IncRef'd `Py_None` instead of setting NULL, so `__del__` always has a valid pointer
- CPython FFI functions continue to return nullable `PyObjectPtr` — no public API changes

## Design decisions

- **No new wrapper struct** — `NonNullPyObjectPtr` is a single-line `comptime` alias
- **`debug_assert` over `raises`** — the constructors are internal; NULL is always a caller bug, not a user-facing error
- **`Py_None` sentinel in `steal_data`** — zero-cost on CPython 3.12+ where None is immortalized
- **`Bool(kwargs)` replaces `kwargs._obj_ptr`** — existing truthiness checks on kwargs were relying on pointer nullability, which no longer applies

## Test plan

- [ ] Existing `test_python_object.mojo` tests pass
- [ ] New tests: `test_nonnull_obj_ptr`, `test_steal_data_leaves_valid_object`, `test_steal_data_roundtrip`
- [ ] Integration tests in `python-extension-modules/` pass

Fixes #5414

### DIFF
--- a/mojo/integration-test/python-extension-modules/def-method/mojo_module.mojo
+++ b/mojo/integration-test/python-extension-modules/def-method/mojo_module.mojo
@@ -231,7 +231,7 @@ struct Person(Defaultable, ImplicitlyCopyable, Writable):
         """
         var self_ptr = Self._get_self_ptr(py_self)
         var total = 0
-        if py_kwargs._obj_ptr:
+        if Bool(py_kwargs):
             for entry in py_kwargs.values():
                 total += Int(py=entry)
         self_ptr[].age += total
--- a/mojo/integration-test/python-extension-modules/feature-overview/mojo_module.mojo
+++ b/mojo/integration-test/python-extension-modules/feature-overview/mojo_module.mojo
@@ -71,26 +71,26 @@ def case_return_arg_tuple(
     return args
 
 
-def case_raise_empty_error() -> PythonObject:
+def case_raise_empty_error() raises -> PythonObject:
     ref cpython = Python().cpython()
 
     var error_type = cpython.get_error_global("PyExc_ValueError")
-
     cpython.PyErr_SetNone(error_type)
 
-    return PythonObject(from_owned=PyObjectPtr())
+    # Raise to unwind; the wrapper preserves the already-set ValueError.
+    raise Error()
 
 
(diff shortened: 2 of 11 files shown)


c-pozzi:
I have checked the CI failures and it seems that the same `Build and Test on large-oss-linux` job also fails on `main` (e.g. run b9fc9ac1) with the same set of tests (`test_shard_and_stack`, `test_vnni_intrinsics`, `test_mojobench`, python-interop tests, etc.). None of the failing tests appear related to the changes in this PR.

Happy to re-trigger if needed once the flakes are resolved.

c-pozzi:
@NathanSWard @JoeLoser

PR #6335, opened 2026-04-02
Refine SM100 softmax large-row paths

prsabahrami:
## Summary
This change keeps the SM100 large-row softmax path and the supporting benchmark harness updates required to validate that path reliably on the current branch.

## Benchmark Notes
- Fresh exact-branch rerun on the 2x B200 workspace completed all 13 generated benchmark items.
- Current exact-branch latencies:
  - `1x128256`: `15.05568 us`
  - `32x128256`: `13.43296 us`
  - `128x128256`: `29.48159 us`
  - `256x4096`: `4.20064 us`
  - `1024x4096`: `8.46527 us`
  - `fp32 32x128256`: `25.37695 us`
- The merge-base harness no longer fails at wrapper generation after a validation-only compatibility patch, but the key `128x128256` baseline still crashes at runtime on the 2x B200 workspace (`worker died`, exit code `-11`).
- That means there is still no trustworthy fresh merge-base-to-head percentage for the same exact large-row shape.
- Historical branch-note context on this path recorded roughly `+50.4%` on the exact `128x128256` route with effectively flat controls; that remains historical context, not the fresh headline.

## Verification
- Exact-branch `kbench` rerun on the 2x B200 workspace.
- Existing PR checks are green.

## Known Limits
- The exact merge-base softmax baseline can now be coerced into the current benchmark harness, but the large-row baseline still crashes before producing a usable row, so the current rerun evidence is exact branch-head latency rather than a fresh before/after delta.

### DIFF
--- a/max/kernels/benchmarks/autotune/kbench.py
+++ b/max/kernels/benchmarks/autotune/kbench.py
@@ -123,8 +123,9 @@ def run(
     timeout_secs: int | None = None,
     plot: str = "bars",
     use_shared_lib: bool = False,
+    reset_shared_lib_worker_per_binary_group: bool = False,
     cache_dir: Path | None = None,
-) -> None:
+) -> int:
     if yaml_path_list:
         # Load specs from a list of YAML files and join them in 'spec'.
         assert len(yaml_path_list), "There should be at least 1 YAML as input."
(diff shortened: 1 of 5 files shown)


govindansriram:
!sync

govindansriram:
This is an amazing PR, but I feel like it should be broken up. Maybe make the kbench segment one part, etc ...

prsabahrami:
will do it tomorrow!

PR #6338, opened 2026-04-02
Expand contiguous reduce specializations

prsabahrami:
## Summary
- expand the contiguous reduction specializations in `reduction.mojo`
- keep the host-side benchmark sizing fix in `bench_reduce.mojo`
- retain the live `3072` fixed-turn ILP2 and `4096` warp0-epilogue paths for contiguous axis-2 reductions

## Benchmark Notes
- historical branch-note `3072` path:
  - `1x1024x3072 axis=2`: `+27.53%` bf16 and `+32.53%` f16 versus the conservative control
  - even against the prior high-water controls, the same shape remains `+19.65%` bf16 and `+27.08%` f16
- historical branch-note `4096` tail path:
  - `1x256x4096 axis=2`: `241.07` bf16 / `237.48` f16 GElems/s
  - versus the kept R100 controls: `+28.58%` bf16 and `+23.55%` f16
- fresh exact-head B200 rerun on the current PR head produced:
  - `1x1024x3072 axis=2`: `376.77` bf16 / `365.46` f16 GElems/s
  - `1x256x4096 axis=2`: `161.52` bf16 / `158.92` f16 GElems/s
  - `32x1024x256x1024 axis=3`: `1178.02` bf16 / `1385.40` f16 GElems/s
- the current exact-head rerun still succeeds on all 23 benchmark items, but it does not reproduce every earlier control-relative percentage or throughput note exactly; those earlier percentages should be read as branch-note campaign context

### DIFF
--- a/max/kernels/benchmarks/gpu/bench_reduce.mojo
+++ b/max/kernels/benchmarks/gpu/bench_reduce.mojo
@@ -11,7 +11,13 @@
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
 
-from std.sys import align_of, get_defined_int, get_defined_string, simd_width_of
+from std.sys import (
+    align_of,
+    get_defined_bool,
+    get_defined_int,
+    get_defined_string,
+    simd_width_of,
+)
 from std.sys.info import _TargetType
 
 from std.algorithm.backend.gpu.reduction import reduce_launch
@@ -24,6 +30,7 @@ from std.benchmark import (
 )
 from layout import Layout, LayoutTensor, RuntimeLayout
 from std.gpu.host import DeviceContext, get_gpu_target
+from std.memory import bitcast
 from internal_utils import (
     CacheBustingBuffer,
     get_defined_shape,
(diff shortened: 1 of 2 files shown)


max/kernels/benchmarks/gpu/bench_reduce.mojo
```diff
@@ -65,7 +65,7 @@ def run_reduce[
     var expected_vals = alloc[Scalar[dtype]](out_size, alignment=align)
 
     var in_host = alloc[Scalar[dtype]](cb_in.alloc_size())
-    var res_host = alloc[Scalar[dtype]](out_size)
+    var res_host = alloc[Scalar[dtype]](in_size)
```
joeatodd, line 75: This doesn't seem right to me. The host side results array should be of `out_size`. Shouldn't instead `res_buffer` be set to size `out_size`?

abduld:
!sync

mojo/stdlib/std/algorithm/backend/gpu/reduction.mojo
```diff
@@ -998,9 +1113,35 @@ def reduce_launch[
...
+                        init,
+                        grid_dim=num_blocks,
+                        block_dim=BLOCK_SIZE,
+                        attributes=pdl_launch_attributes(pdl_level),
+                    )
         else:
+            comptime contig_simd = simd_width_of[dtype, get_gpu_target()]()
             comptime for ax in range(rank):
                 if axis == ax:
+                    comptime is_contig = (ax == rank - 1)
```
joeatodd, line 2682: Don't need to recompute `is_contig` here. Just use `reduce_contig_dim` and pull `reduce_simd` out of the for loop.

joeatodd (changes requested):
Hi @prsabahrami, thanks for this PR. Did you get a chance to look over my review comments, and those from Copilot?

prsabahrami:
Apologies for the delay here, will address these tomorrow!

PR #6340, opened 2026-04-02
[Stdlib] Use t-strings in HAL driver and plugin error messages

msaelices:
Replace string concatenation with t-string interpolation in HAL error messages to avoid intermediate heap allocations.

Assisted-by: AI


szbergeron:
Thanks for the catch!

szbergeron:
!sync

szbergeron (approved):

PR #6343, opened 2026-04-02
[bazel] Add support for interface shared libraries

keith:
This change enables bazel's support for producing interface libraries
for use when linking. This means it adds an extra output to shared
library linking that produces a slim version of the actual library
with just enough information for downstream links. This means you no
longer have to upload/download the massive libraries just to link
something that depends on them.


patrickdoc:
!sync

patrickdoc:
!sync

PR #6348, opened 2026-04-03
[stdlib][simd] Add test for simd construction checks

josiahls:
Note, currently @always_inline("builtin") basically strips the asserts out of the code.

## Summary

SIMD `_simd_construction_checks` is noop due to `@always_inline("builtin")`

When you run the test, you will notice that `_simd_construction_checks` isn't the thing that fails.
```
def main() raises:
    ^
/home/mojo_user/momanim/third_party/modular/mojo/stdlib/test/builtin/test_simd_compile_fail.mojo:18:49: note: call expansion failed with parameter value(s): (..., "size": 3)
    _simd_construction_checks2[DType.float32, 3]()
                                                ^
mojo/stdlib/std/builtin/simd.mojo:255:5: note: function instantiation failed
mojo/stdlib/std/builtin/simd.mojo:265:5: note: constraint failed: simd width must be power of 2
/home/mojo_user/momanim/.pixi/envs/stdlib-dev/bin/mojo: error: failed to run the pass manager
```

## Testing

I have a compile fail test that demos this. 

## Checklist

- [X] PR is small and focused — consider splitting larger changes into a
      sequence of smaller PRs

### DIFF
--- a/mojo/stdlib/test/builtin/BUILD.bazel
+++ b/mojo/stdlib/test/builtin/BUILD.bazel
@@ -37,6 +37,7 @@ _LIT_TESTS = [
     "test_format_compile_fail.mojo",
     "test_format_float_lut_optimization.mojo",
     "test_global_constant_compile_fail.mojo",
+    "test_simd_compile_fail.mojo",
     "test_issue_3908.mojo",
     "test_print_long_string.mojo",
     "test_print_stderr.mojo",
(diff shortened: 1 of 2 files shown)


josiahls:
cc @NathanSWard `@always_inline("builtin")` appears to strip any comptime assertions. `_simd_construction_checks` is noop as it stands. I can update the test to be smaller, just putting a demo copy of the function without the builtin line.

Basically this pr isn't going to be mergable until an API change or compiler fix.

PR #6351, opened 2026-04-05
[Stdlib] Fix llvm.masked.scatter CUDA_ERROR_ILLEGAL_ADDRESS in shared memory

### LINKED ISSUE #5366: [BUG] [stdlib] `llvm.masked.scatter` causes `CUDA_ERROR_ILLEGAL_ADDRESS` in shared memory
### Bug description

It works in global memory but it doesn't in shared memory. We need to either provide a fallback or constrain it to `AddressSpace.GENERIC`

### Steps to reproduce

```mojo
from gpu.host import DeviceContext
from layout import Layout, LayoutTensor
from layout.tensor_builder import LayoutTensorBuild as tb
from testing import assert_equal


fn kernel(tensor: LayoutTensor[mut=False, DType.float32, Layout.row_major(4)]):
    var shared_mem = tb[DType.float32]().row_major[4]().shared().alloc()
    shared_mem.ptr.scatter(
        SIMD[DType.uint, 4](3, 2, 1, 0), SIMD[DType.float32, 4](1, 2, 3, 4)
    )


fn main() raises:
    var data = List[Float32](18, 7, 29, 27)

    with DeviceContext() as ctx:
        var dev_buf = ctx.enqueue_create_buffer[DType.float32](4)
        ctx.enqueue_copy(dev_buf, data.unsafe_ptr())
        var tensor = LayoutTensor[
            mut=False, DType.float32, Layout.row_major(4)
        ](dev_buf.unsafe_ptr())
        ctx.enqueue_function[kernel](tensor, grid_dim=1, block_dim=1)
        ctx.synchronize()
        with dev_buf.map_to_host() as out_host:
            assert_equal(out_host[0], 4)
            assert_equal(out_host[1], 3)
            assert_equal(out_host[2], 2)
            assert_equal(out_host[3], 1)
```

hardikkgupta:
Fixes #5366 

## Summary

Add a GPU fallback for the scatter function in std/sys/intrinsics.mojo that performs element-wise stores

## Testing

- Validated that the fallback logic mirrors the existing gather GPU fallback pattern

### DIFF
--- a/mojo/stdlib/std/sys/intrinsics.mojo
+++ b/mojo/stdlib/std/sys/intrinsics.mojo
@@ -237,7 +237,6 @@ def scatter[
         return
 
     comptime if is_gpu():
-
         comptime for i in range(size):
             if mask[i]:
                 UnsafePointer[Scalar[dtype], MutExternalOrigin](
--- a/mojo/stdlib/std/sys/intrinsics.mojo
+++ b/mojo/stdlib/std/sys/intrinsics.mojo
@@ -236,14 +236,6 @@ def scatter[
             ptr.store(value[0])
         return
 
-    comptime if is_gpu():
-        comptime for i in range(size):
-            if mask[i]:
-                UnsafePointer[Scalar[dtype], MutExternalOrigin](
-                    unsafe_from_address=Int(base[i])
-                ).store(value[i])
-        return
-
     llvm_intrinsic["llvm.masked.scatter", NoneType](
         value,
         UnsafePointer(to=base).bitcast[
--- a/mojo/stdlib/std/sys/intrinsics.mojo
+++ b/mojo/stdlib/std/sys/intrinsics.mojo
@@ -235,6 +235,16 @@ def scatter[
             )
             ptr.store(value[0])
         return
+
+    comptime if is_gpu():
+
+        comptime for i in range(size):
+            if mask[i]:
+                UnsafePointer[Scalar[dtype], MutExternalOrigin](
+                    unsafe_from_address=Int(base[i])
+                ).store(value[i])
+        return
+
     llvm_intrinsic["llvm.masked.scatter", NoneType](
         value,
         UnsafePointer(to=base).bitcast[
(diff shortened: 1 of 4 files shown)


mojo/stdlib/std/sys/intrinsics.mojo
```diff
@@ -235,6 +235,15 @@ def scatter[
             )
             ptr.store(value[0])
         return
+
+    comptime if is_gpu():
```
martinvuyk: Shouldn't check that the address space is not `AddressSpace.GENERIC` here? I'm not sure whether `llvm.masked.scatter` just does the same scalar storing that we're doing here for GPU global memory. I also don't know how we should treat stack pointers, AFAIK they also use `AddressSpace.GENERIC` but they are technically pointing to GPU registers which I'm also not sure whether they support the instruction.

I think this should be accompanied by some gpu test cases, but I'm also unfamiliar with how we're handling those and configuring the test suite to only run when an accelerator is available. @JoeLoser how should we move forward with this?

JoeLoser: +1 to the concern here.

This proposed fix applies the fallback to all GPU scatter operations unconditionally. This bypasses the `llvm.masked.scatter` intrinsic for global memory too, where it works fine IIUC. Compare with the gather fallback earlier in the file, which is conditional:

```
  comptime if is_gpu() and invariant:  # gather — targeted
```

  vs. the proposed:
```
  comptime if is_gpu():  # scatter — unconditional
```

Also +1 for a test. You can see various bazel rules where we use `target_compatible_with = ["//:has_gpu"]`.

hardikkgupta: Makes sense, I can revise the approach:
Move the fallback to `UnsafePointer.scatter()` in  `unsafe_pointer.mojo`, where address_space is a known compile-time parameter.
The condition will become:                                                                   
 ```
comptime if is_gpu() and address_space != AddressSpace.GENERIC and address_space != AddressSpace.GLOBAL:
```                                         
So only shared/local/cluster memory scatter uses the scalar fallback. Global memory scatter still uses the optimized llvm.masked.scatter intrinsic
What do you think?

hardikkgupta: Also will add tests (`target_compatible_with = ["//:has_gpu"]`) covering global scatter, shared memory scatter, and masked shared memory scatter

hardikkgupta: Let me know if this seems fine

PR #6356, opened 2026-04-06
[stdlib] Add `SIMD` `VariadicPack` constructor to eventually replace the `VariadicList` one

martinvuyk:
Add `SIMD` `VariadicPack` constructor to eventually replace the `VariadicList` one

### DIFF
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -776,33 +776,78 @@ struct SIMD[dtype: DType, size: Int](
         )
 
     @always_inline("nodebug")
-    def __init__(
-        out self, *elems: Scalar[Self.dtype], __list_literal__: () = ()
-    ):
+    def __init__[
+        *Ts: type_of(Scalar[Self.dtype])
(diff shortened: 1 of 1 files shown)


lsh:
I'd like to see a more detailed motivation for this change.

martinvuyk:
@lsh I want to get rid of `VariadicList` because it doesn't have a compile time size. But either way I think I'll throw this PR in the bin once Chris finishes his `TypeList` work and we can do stuff like `*args: *SomeTypeList[SomeTrait]`

lsh:
I can see the value of that in some cases, but I'm not sure it's actually helping the types you're applying this to (such as `InlineArray` or `IndexList`).

martinvuyk:
Yeah but I think it's currently more of a language limitation, not a conceptual one. We should be able to have a list of types that an API receives be able to be bound to a single type but also be implicitly convertible. And the horrible overloads I had to build for inline array and such are a side-effect of issues with `VariadicPack` (https://github.com/modular/modular/issues/6357, https://github.com/modular/modular/issues/6381).

PR #6367, opened 2026-04-06
Optimize Gemma 3 decode rope and fused QK paths

prsabahrami:
## Summary
- add the current decode-focused rope, QK, and KV-cache fast paths used by the Gemma decode stack
- wire in the supporting Gemma integration profiling tests and benchmark harnesses
- keep the validated exact BF16/128 decode wins for the live q-only, qk, and k-only paths

## Benchmark Notes
- exact BF16/128 live q-only decode:
  - `1.16138153x` speedup
  - `6.92620000 us` saved
- exact BF16/128 live QK decode with correctness restored:
  - baseline `62.27719714 us`
  - fused `33.24680147 us`
  - `1.87317860x` speedup
  - `29.03039567 us` saved
- exact BF16/128 live K-only decode:
  - `1.23905479x` speedup
  - `9.10064846 us` saved
- the stronger composed QK result remains larger than the separable q-only and k-only wins, which matches the intended end-to-end decode path this branch is targeting

### DIFF
--- a/max/kernels/benchmarks/gpu/nn/bench_gemma3_k_norm_rope_boundary.mojo
+++ b/max/kernels/benchmarks/gpu/nn/bench_gemma3_k_norm_rope_boundary.mojo
@@ -1,0 +1,1054 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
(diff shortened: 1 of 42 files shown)


prsabahrami:
Superseded by the split follow-up PRs:\n- #6373 Optimize Gemma 3 rope and fused QK decode paths\n- #6372 Add Gemma 3 decode profiling coverage

Issue #6357, opened 2026-04-06
[BUG] [mojo-lang] Implicit conversion isn't being detected for `VariadicPack`s bound to a type


martinvuyk:
### Bug description

```mojo
from std.builtin.variadics import Variadic


struct SIMDWrapper[dtype: DType, size: Int]:
    var vec: SIMD[Self.dtype, Self.size]

    # VariadicList worked
    # def __init__(
    #     out self, *elems: Scalar[Self.dtype], __list_literal__: () = ()
    # ):
    #     __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))

    #     comptime for i in range(Self.size):
    #         self.vec[i] = elems[i]

    def __init__[
        *Ts: type_of(Scalar[Self.dtype])
    ](out self: SIMDWrapper[Self.dtype, Variadic.size_types[Ts]], *elems: *Ts):
        __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))

        comptime for i in range(Variadic.size_types[Ts]):
            self.vec[i] = elems[i]


def main():
    # This works
    print(SIMDWrapper[DType.uint8](UInt8(1)).vec)
    print(SIMDWrapper[DType.uint8](UInt8(1), UInt8(2)).vec)
    print(SIMDWrapper[DType.uint8](UInt8(1), UInt8(2), UInt8(3)).vec)
    # This fails with:
    # invalid initialization: could not convert element of 'elems' with type
    # 'Int' to expected type 'AnyStruct[UInt8]'
    print(SIMDWrapper[DType.uint8](1).vec)
```

### Steps to reproduce

...

### System information

`Mojo 0.26.3.0.dev2026040605`

soraros:
I think this is technically correct. When things gets lifted to the `AnyStruct[T]` singleton level, things gets a bit tricky. CC @zyx-billy.

martinvuyk:
To ilustrate other examples of what I'm forced to do to add this constructor for `InlineArray` in order to replace the `VariadicList` one
```mojo
    @always_inline
    def __init__[
        *Ts: type_of(Self.ElementType)
    ](out self, var *elems: *Ts, __list_literal__: ()) where (
        elems.__len__() == Self.size
        # FIXME(#6357): somehow I'm getting: error: rebind input type
        # '!kgen.pointer<index>' does not match result type '!kgen.pointer<scalar<ui8>>'
        # even though this constructor should only accept the same type
[...]

PR #6369, opened 2026-04-06
[stdlib] Use `VariadicPack` instead of `VariadicList` where possible

martinvuyk:
Use `VariadicPack` instead of `VariadicList` where possible. #6356 remains blocked in another PR, these here should work with no issues

### DIFF
--- a/max/kernels/src/layout/runtime_tuple.mojo
+++ b/max/kernels/src/layout/runtime_tuple.mojo
@@ -108,13 +108,22 @@ struct RuntimeTuple[
                 self.value[i] = UNKNOWN_VALUE
 
     @always_inline
-    def __init__(out self, *values: Int):
+    def __init__[*Ts: type_of(Int)](out self, *values: *Ts):
         """Initialize a `RuntimeTuple` with the provided values.
 
+        Parameters:
+            Ts: The integers.
+
         Args:
-            values: Variadic number of integer values to initialize the tuple with.
+            values: Variadic number of integer values to initialize the tuple
+                with.
         """
-        self.value = type_of(self.value)(*values)
+        comptime assert (
+            values.__len__() == Self.scalar_length
+        ), "mismatch in the number of elements"
+        self.value = {}
+        comptime for i in range(values.__len__()):
+            self.value[i] = values[i]
 
     @always_inline
     @implicit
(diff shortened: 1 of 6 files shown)


martinvuyk:
Blocked by #6381

lsh:
I'd like to see a more detailed motivation for this change.

mojo/stdlib/std/collections/inline_array.mojo
```diff
@@ -432,9 +433,20 @@ struct InlineArray[ElementType: Copyable, size: Int](
         )
 
     @always_inline
-    def __init__(out self, var *elems: Self.ElementType, __list_literal__: ()):
+    def __init__[
+        *Ts: type_of(Self.ElementType)
+    ](out self, var *elems: *Ts, __list_literal__: ()) where (
```
lsh, line 438: This is almost certainly not correct. An `InlineArray` takes a homogenous type.

max/kernels/src/layout/runtime_tuple.mojo
```diff
@@ -621,31 +630,30 @@ def shape_div[
...
-                    res.value[i + j] = res_i.value[j]
-            return res
+                    result.value[i + j] = res_i.value[j]
         else:
-            var res = RuntimeTuple[shape_div_int_tuple(a_t, b_t)]()
+            result = {}
             # FIXME: this used to be simpler
             # var vb = Int(to_int(b))
-            var vb = RuntimeTuple[IntTuple(UNKNOWN_VALUE)](Int(b))
+            comptime R = RuntimeTuple[IntTuple(UNKNOWN_VALUE)]
```
lsh, line 644: I'm strongly -1 on these single letter aliases.

mojo/stdlib/std/collections/inline_array.mojo
```diff
@@ -447,29 +459,101 @@ struct InlineArray[ElementType: Copyable, size: Int](
...
+            (self.unsafe_ptr() + idx).init_pointee_move(self.ElementType(elt))
+
+        elems^.consume_elements[init_elt]()
+
+    # TODO(#6357): remove once implicit conversion in VariadicPack is solved
+    @always_inline
+    def __init__[
+        *Ts: type_of(String)
+    ](
+        out self: InlineArray[StaticString, Self.size],
```
lsh, line 522: We're moving away from custom `Self` types in favor of where clauses, so this is not ideal.

mojo/stdlib/std/collections/inline_array.mojo
martinvuyk: it does, as an example `*Ts: type_of(Int)` enforces a single type

mojo/stdlib/std/collections/inline_array.mojo
martinvuyk: yeah I don't know what is going on with the compiler but it eagerly lowers `StringLiteral` into `String` so anything wanting to construct `InlineArray[StaticString, 1] = ["asd"]`  where the constructor expects the initializer type to equal the target type it would fail. We currently don't have a nice way for variadic packs to properly take implicit type conversion into account (#6357) and also carries type information incorrectly (https://github.com/modular/modular/issues/6381)

mojo/stdlib/std/collections/inline_array.mojo
lsh: If the goal is to try and optimize `InlineArray` using this information, I think the better solution is to have `__list_literal__` carry an actual size parameter, but I can be convinced otherwise.

martinvuyk: It's not about optimization but enforcing a compile-time known constraint at compile time

PR #6371, opened 2026-04-06
[kernels] Stabilize matmul benchmark cache-busting controls

prsabahrami:
## Summary
- make matmul benchmark cache busting configurable instead of always forcing it on
- thread shared benchmark config arguments into the matmul benchmark harness
- allow matched reruns with and without cache-busting for exact-shape investigations

## Benchmark Notes
- this change does not claim a matmul kernel speedup on its own
- it narrows the benchmark harness so exact-shape reruns can be compared under the same cache-busting configuration
- the intent is measurement stability and reproducibility, not a production kernel behavior change

## Verification
- reviewed against the current exact-shape matmul benchmark workflow on B200

### DIFF
--- a/max/kernels/benchmarks/gpu/linalg/bench_matmul.mojo
+++ b/max/kernels/benchmarks/gpu/linalg/bench_matmul.mojo
@@ -33,6 +33,7 @@ from internal_utils import (
     CacheBustingBuffer,
     arg_parse,
     pytorch_like_tolerances_for,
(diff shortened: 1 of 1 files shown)


yongmzhang:
Please add [kernels] in the title

yongmzhang (approved):

prsabahrami:
is this good to merge?

Issue #6359, opened 2026-04-06
[Feature Request] `String` iterator APIs


NathanSWard:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

This refers back to this proposal here: https://github.com/modular/modular/pull/6195

This is the larger tracking issue for implementing all of the desired iterators for `String(Slice)`'s methods.

### Any other details?

JoeLoser:
Fyi @martinvuyk

martinvuyk:
Idea dump, if we added `String.__add__` to this list of lazy iterators we would easily fold things like:
```mojo
var res = String()
for _ in range(5):
  res += "some_long_string" + "some_other_long_string"
```
Into 5 malloc calls instead of 10

PR #6384, opened 2026-04-07
[stdlib] Add `String + Iterable[Writable]` and `String + IterableOwned[Writable]`

martinvuyk:
Add `String + Iterable[Writable]` and `String + IterableOwned[Writable]`. Phase 0 of https://github.com/modular/modular/issues/6359

### DIFF
--- a/mojo/stdlib/std/collections/string/string.mojo
+++ b/mojo/stdlib/std/collections/string/string.mojo
@@ -856,6 +856,43 @@ struct String(
         memcpy(dest=result_ptr + lhs_len, src=rhs.unsafe_ptr(), count=rhs_len)
         return result^
 
+    @staticmethod
+    def _add[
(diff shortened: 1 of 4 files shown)


NathanSWard:
I'm a big -1 on this api. Being able to add a `String` + any `Writable` doesn't seem correct and bloats that API surface. It enables unintuitive things like
```
var s = "hello"
var n = 42
var s2 = s + n
print(s2) # "hello42"
```

there are better ways to do this with t-strings.
```
var s = "hello"
var n = 42
var s2 = String(t"{s}{n}")
```

martinvuyk:
@NathanSWard yeah I wasn't so sure myself about adding `Some[Writable]`, what about the `Iterable[Writable]` API ?

martinvuyk:
I'm going to close this off in favor of possibly making a reimplementation of this that merges iterators instead of returning a string

Issue #3572, opened 2024-09-30
[Feature Request] Make all examples in docstring compile


soraros:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

As title.

### What is your motivation for this change?

Otherwise, the LSP generates lots of errors.

### Any other details?

N/A

JoeLoser:
I think if you were to point `mojo test` at the stdlib source directory, this may help flag them all in one shot perhaps.  Is that right, @lbrown?

AmaranthineCodices:
I believe so. It's been a little bit since I tried this but from what I remember, you'll also need to build the stdlib and include it with `-I` - some experimentation may be required to avoid using the bundled stdlib.

kasmith11:
Have there been any updates to finding all docstring errors / issues using `mojo test`? @JoeLoser

JoeLoser:
> Have there been any updates to finding all docstring errors / issues using `mojo test`? [@JoeLoser](https://github.com/JoeLoser)

I'll defer to @rparolin on this, but I'm not aware of anything new being worked on here, although I think it would be nice still.

rparolin:
Thanks for flagging. We haven't been working on this recently due other priorities but I can raise the priority of this now.<br>cc: @AmaranthineCodices

PR #6392, opened 2026-04-09
[stdlib] Add docstring examples to testing module functions

isaacuselman:
## Summary
6 assertion functions in std.testing had no usage examples in their docstrings: assert_false, assert_equal, assert_not_equal, assert_almost_equal, assert_is, assert_is_not. No logic changes — docstrings only.
Partial fix for #3572.

## Testing

./bazelw test //mojo/stdlib/std:std.docs_test — passes, all docstring examples compile

## Checklist

- [x] PR is small and focused 
- [x] I ran ./bazelw run format
- [x]  If AI tools assisted with this contribution, I have included an Assisted-by: trailer in my commit message or this PR description

### DIFF
--- a/mojo/stdlib/std/testing/testing.mojo
+++ b/mojo/stdlib/std/testing/testing.mojo
@@ -96,6 +96,14 @@ def assert_false[
 
     Raises:
         An Error with the provided message if assert fails and `None` otherwise.
+
+    Example:
+    ```mojo
+    from testing import assert_false
+
+    assert_false(False)  # passes
+    assert_false(0)      # passes (0 is falsy)
+    ```
     """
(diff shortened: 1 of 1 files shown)


fairlydurable:
Thank you for the PR.

Suggestion: it would be great to use examples that show more meaningful or real-world use cases.

For example, this doesn’t give much context for how someone would use it:

```mojo
from testing import assert_false

assert_false(False)  # passes
assert_false(0)      # passes
```

It also leans on “falsy” behavior, which we probably don’t want to encourage in docs examples. A stronger sample would show assert_false() in a realistic test condition where the expected result is clear.

PR #6393, opened 2026-04-09
[stdlib] Add docstring examples to hashlib hash functions Body:

isaacuselman:
## Summary
2 hash function overloads in std.hashlib had no usage examples in their docstrings. No logic changes — docstrings only.
Partial fix for #3572.

## Testing

./bazelw test //mojo/stdlib/std:std.docs_test — passes, all docstring examples compile

## Checklist

- [x]  PR is small and focused
- [x]  I ran ./bazelw run format
- [x]  I added or updated tests to cover my changes
- [x]  If AI tools assisted with this contribution, I have included an Assisted-by: trailer in my commit message or this PR description

### DIFF
--- a/mojo/stdlib/std/hashlib/hash.mojo
+++ b/mojo/stdlib/std/hashlib/hash.mojo
@@ -112,6 +112,14 @@ def hash[
 
     Returns:
         A 64-bit integer hash based on the underlying implementation.
+
+    Example:
+    ```mojo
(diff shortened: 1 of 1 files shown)


fairlydurable:
Thanks for the PR.

Suggestion: these examples show how to call hash(), but they don’t illustrate why you’d use it. It would be great to include a more contextual example where the hash value plays a role.

PR #6399, opened 2026-04-09
[stdlib] Fix docstring examples imports and other details

martinvuyk:
Hide all imports that are for elements in the same module as the function/struct being documented.

Fix several non-working examples

### DIFF
--- a/max/kernels/test/gpu/layout/test_wgmma_with_static_tuple_output_fp8.mojo
+++ b/max/kernels/test/gpu/layout/test_wgmma_with_static_tuple_output_fp8.mojo
@@ -15,7 +15,7 @@ import linalg.matmul.vendor.blas as vendor_blas
 from std.gpu import barrier, warp_id, lane_id
 from std.gpu.host import DeviceContext
 
-# from testing import assert_almost_equal
+# from std.testing import assert_almost_equal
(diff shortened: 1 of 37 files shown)


fairlydurable:
Thank you for the PR — it’s looking good.

I noticed the examples are marked with docs exclusion for imports. We prefer to keep examples visible so developers can try them directly without needing to track them down elsewhere. Could you update those?

martinvuyk:
> I noticed the examples are marked with docs exclusion for imports. We prefer to keep examples visible so developers can try them directly without needing to track them down elsewhere. Could you update those?

@fairlydurable I commented out imports from the same module where the function is defined and imports from things like `assert_equal` etc. I can add them back if you wish, I just wanted to point that out

fairlydurable:
> I commented out imports from the same module where the function is defined and imports from things like assert_equal etc. I can add them back if you wish, I just wanted to point that out

@martinvuyk Please do add them back in.

People will see this in the module source code (where it is less important) but also in the generated docs pages. It is the latter where it really helps because the information they need (the import) is right next to the sample code, so they can copy/paste the import exactly.

Thank you for your thoughtful consideration on this work.

fairlydurable:
!sync

PR #6402, opened 2026-04-09
[stdlib] Add docstring example to swap function

isaacuselman:
## Summary
Add docstring example to `swap` showing basic value swapping. No logic changes — docstrings only.
Partial fix for #3572.
## Testing
./bazelw test //mojo/stdlib/std:std.docs_test — all tests pass
## Checklist
- [x] PR is small and focused
- [x] I ran ./bazelw run format
- [x] If AI tools assisted with this contribution, I have included an Assisted-by: trailer in my commit message or this PR description


fairlydurable:
Thank you for the PR. Suggestion: it would be great to use examples that show a more meaningful or real-world use case.

PR #6404, opened 2026-04-09
[stdlib] Add `std.iter.chain` function to concatenate iterators

martinvuyk:
Add `std.iter.chain` function to concatenate iterators

### DIFF
--- a/mojo/stdlib/std/iter/__init__.mojo
+++ b/mojo/stdlib/std/iter/__init__.mojo
@@ -54,6 +54,8 @@ for squared in map[square](values):
 """
 
 from std.builtin.constrained import _constrained_conforms_to
+from std.sys.intrinsics import _type_is_eq_parse_time
+
 
 from std.builtin.variadics import TypeList
 from std.reflection.traits import AllImplicitlyDestructible, AllCopyable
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/iter/__init__.mojo
```diff
@@ -1089,6 +1092,157 @@ def peekable(
...
+            if self._idx <= i:
+                try:
+                    return rebind_var[Self.Element](next(self._values[i]))
+                except:
+                    self._idx += 1
+        raise StopIteration()
+
+
+@fieldwise_init
+struct _ChainedIterableOwned[origin: MutOrigin, *Ts: IterableOwned & Iterator](
```
NathanSWard: I believe you should be able to simply remove the `_ChainedIterableRef` and `_ChainedIterableOwned` types and instead have this iterator hold the actual `Iterator` types instead of the `Iterable` types. 

Something like
```mojo
_ChainIterator[*Iterators: Iterator]:
    ...
```

NathanSWard: You'll have to them update the `chain` functions themselves to call `iter` on each of the passed in iterators. I think this should hopefully clean up this iterator nicely :)

NathanSWard: Lastly, we should also make sure to implement `def bounds()` for this iterator as well. You'll basically just have to sum the lower & upper bounds making to sure to not accidentally wrap on `Int.MAX` and if any of the upper bounds are `None`, this iterator should also return `None` for its upper bound.

NathanSWard (changes requested):

mojo/stdlib/std/iter/__init__.mojo
martinvuyk: I think you forgot to hit submit review on this one lol. Yeah I reworked the iterator logic based on the zip iterator work I did

NathanSWard: Ahh whoops yea my review was a bit stale haha. Sry!!

JoeLoser:
@copilot resolve the merge conflicts in this pull request

JoeLoser:
!sync

JoeLoser:
!sync

martinvuyk:
@JoeLoser sorry I missed some dangling imports, just fixed them

JoeLoser:
!sync

Issue #6381, opened 2026-04-07
[BUG] [mojo-lang] `VariadicPack` isn't carrying around type information correctly


martinvuyk:
### Bug description

There is something wrong in how the type information is being carried around that only surfaces in lower levels in the compilation pipeline. The implicit conversions aren't being respected either ([MOCO-3698](https://linear.app/modularml/issue/MOCO-3698/bug-mojo-lang-implicit-conversion-isnt-being-detected-for))

```mojo
struct InlineArrayWrapper[T: Copyable, size: Int]:
    var arr: InlineArray[Self.T, Self.size]

    def __init__[
        *Ts: type_of(Self.T)
    ](out self, var *elems: *Ts, __list_literal__: () = ()):
        comptime assert (
            elems.__len__() == Self.size
        ), "mismatch in the number of elements"
        __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))

        @parameter
        def init_elt[idx: Int](var elt: elems.element_types[idx]):
            (self.arr.unsafe_ptr() + idx).init_pointee_move(
                rebind_var[self.T](elt^)
            )

        elems^.consume_elements[init_elt]()


def main():
    print(InlineArrayWrapper[Int32, 4](1, 2, 3, 4).arr)
```

output

```
note: call expansion failed
                rebind_var[self.T](elt^)
                                  ^
oss/modular/mojo/stdlib/std/builtin/rebind.mojo:105:25: note: error: rebind input type '!kgen.pointer<index>' does not match result type '!kgen.pointer<scalar<si32>>'
```

### Steps to reproduce

...

### System information

`Mojo 0.26.3.0.dev2026040605`

PR #6406, opened 2026-04-10
[stdlib] Simplify Zip iterator and enable unlimited args

martinvuyk:
Simplify Zip iterator and enable unlimited args.

We could actually make it yield `VariadicPack` but I kept the current Tuple approach. The internal storage could also be switched.

This also simplifies `_PyIter` and fixes two small memory leaks when creating dicts and sets with `PythonObject` constructors (if setting the values inside the dict/set failed and the functions raised without taking ownership of the dict/set reference)

### DIFF
--- a/mojo/docs/code/stdlib/pathlib/path.mojo
+++ b/mojo/docs/code/stdlib/pathlib/path.mojo
@@ -151,7 +151,9 @@ def test_parts() raises:
     from std.pathlib import Path
     from std.testing import assert_true
 
-    for p, q in zip(Path("a/path/foo.txt").parts(), ["a", "path", "foo.txt"]):
+    for p, q in zip(
+        Path("a/path/foo.txt").parts(), List(["a", "path", "foo.txt"])
+    ):
         assert_true(p == q)
 
 
--- a/mojo/stdlib/std/iter/__init__.mojo
+++ b/mojo/stdlib/std/iter/__init__.mojo
@@ -55,6 +55,10 @@ for squared in map[square](values):
 
 from std.builtin.constrained import _constrained_conforms_to
 
+from std.builtin.variadics import Variadic
+from std.reflection.traits import AllImplicitlyDestructible, AllCopyable
+
+
 # ===-----------------------------------------------------------------------===#
 # Iterable
 # ===-----------------------------------------------------------------------===#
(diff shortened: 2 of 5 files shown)


martinvuyk:
@JoeLoser  @lsh not sure which one of you is currently working more with iterators, so I'm pinging you both.

I don't know what else to do to get this to work, but the groundwork is laid. Something about the way I'm passing around the origins is failing and I'm getting a segfault when testing. I've tried finding the issue, but I think this is as far as I can take it with my limited understanding of our origin system

NathanSWard:
> Something about the way I'm passing around the origins is failing and I'm getting a segfault when testing.

I wasn't able to exactly locate the error, but I was able to get this to work by instead of holding a `VariadicPack` in the `_ZipIterator` to just hold a `Tuple` instead. I generally find you shouldn't hold a `VariadicPack` for owned values, but just instead holding a `Tuple` is simpler :)

NathanSWard (approved):
awesome!! :) Excited to see you got this working!

NathanSWard:
!sync

martinvuyk:
@NathanSWard I fixed a couple of breakages due to the latest nightly

JoeLoser:
!sync

JoeLoser:
Hey @martinvuyk, thanks for this refactor — the direction is right, but trying to land it internally surfaced two compiler bugs that this PR trips over. The issues seem to be compiler-side, not in your implementation.

Bug 1: pack inference fails on list literals through out params

Here's a minimal repro without `zip` at all.

```mojo
from std.iter import IterableOwned

struct Container[*Ts: IterableOwned]:
    pass

def dummy_call[*Ts: IterableOwned](
    var *iterables: *Ts,
    out res: Container[*Ts],
):
    __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(res))

def main():
    # error: could not infer type of parameter pack 'iterables'
    _ = dummy_call([1, 2, 3], [4, 5, 6])
```

Named variables with `^` transfer work fine — it's specifically list literals flowing into a pack through an out parameter that fails.

Bug 2: SIGABRT in `processParamForOp` in the Elaborator

Triggered when `comptime for` unpacks a tuple yielded by an iterator whose `Element = Tuple[*TypeList[...].map[...]()]`. This one required changed from your branch.

```
from std.iter import zip

def main():
    comptime lst0 = [1, 4]
    comptime lst1 = [1, 4]
    comptime for i, j in zip(lst0, lst1):  # SIGABRT
        pass
```
This hits

```
Assertion failed: (iterParamDecl.getType() == value.getType() &&
    "iterator value type should match the loop decl type")
```

Neither pattern above has equivalent public test coverage, which is why OSS CI looked green. I filed internal compiler bugs and linked to this PR so we don't lose track and I'll mark this PR with label `blocked` until they are fixed.

martinvuyk:
@JoeLoser yeah the first bug made me change a couple of call-sites to do `zip(List([...]), ...)`, the second one seems bad :/

Could this have something to do with #6381 and #6357 ?

JoeLoser:
> @JoeLoser yeah the first bug made me change a couple of call-sites to do `zip(List([...]), ...)`, the second one seems bad :/
> 
> Could this have something to do with #6381 and #6357 ?

I don't think they're related necessarily. For (1), I just implemented a fix in the compiler for this: this was just a bug in parameter inference. It is set to merge and should go out in the next nightly and then I can take a look at your branch to see what the errors look like and if there's any compiler bugs remaining. Feel free to help me out if you have time when the nightly comes out to remove your workarounds for (1) when the fix lands.

JoeLoser:
@martinvuyk I see how to fix (2), I'll put up a PR internally tonight for it with hopes the compiler team can review it tomorrow.

JoeLoser:
@martinvuyk there was a CI flake so the compiler fix for (1) didn't make it into the nightly today. I landed the compiler fix for both (1) and (2) today and am updating your branch/PR internally and I'll run with it from here to get it over the finish line. Thank you!

martinvuyk:
@JoeLoser I was going to update this later on today, also to remove some of the `_PyIter` (and other python) changes that I split off into #6436. You can still do it if you want, whatever is best for you :man_shrugging:

JoeLoser:
> @JoeLoser I was going to update this later on today, also to remove some of the `_PyIter` (and other python) changes that I split off into #6436. You can still do it if you want, whatever is best for you 🤷‍♂️

@martinvuyk I changed the internal PR to just have the zip changes and I'll take a look at your other PR, thanks for splitting those off!

JoeLoser:
@martinvuyk I cleaned things up and got everything working for this branch with the workarounds removed since I fixed both compiler bugs. It should land internally soon. I also added some tests for the zip5/6 cases to show the generalization works.

lattner:
This is awesome @martinvuyk : `for a, b, c, d, e, f in zip(l1, l2, l3, l4, l5, l6):`!  Nice work!

Issue #6379, opened 2026-04-07
[Feature Request] [stdlib] Implement internal shrinking for property test failures


NathanSWard:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

When `PropTest.test` finds a failing input, it currently reports the raw generated value. This value is often unnecessarily large/complex, making it hard to understand the root cause. We need **shrinking**: automatically reducing the failing input to the smallest/simplest case that still fails.

We should implement **internal shrinking** (as pioneered by Python's Hypothesis library), where the random number stream feeding the strategies is recorded and then shrunk, rather than requiring each `Strategy` to implement its own shrinking logic.

### Why internal shrinking (not per-strategy shrinking)?

The traditional approach (Rust's proptest, Haskell's QuickCheck) requires each strategy to define a `shrink` method. This has significant downsides:

1. **Every new strategy needs a shrinker**: more work for strategy authors
2. **Composition breaks shrinking**: if you `map(f, strategy)`, you can't invert `f` to shrink through the map. Rust's proptest needs separate `prop_flat_map` vs `prop_ind_flat_map` vs `prop_ind_flat_map2` to handle different shrinking behaviors through composition
3. **Filter invalidation**: a per-value shrinker can produce values that no longer pass filter predicates, requiring retry loops

Internal shrinking avoids all of this. Since strategies are deterministic functions of the random stream, shrinking the stream automatically shrinks the output values while preserving all invariants.

### How it works

The existing `Rng._next()` method already has a TODO comment describing this design:

```mojo
# TODO: Add playback support.
def _next(mut self, max: UInt64 = UInt64.MAX) raises -> UInt64:
    """If in playback mode, returns the next value in the history, otherwise
    generates a random value and records it.

    All random values are built on top of a random `UInt64` so we can record
    and keep a history of all generated values. When a test failure is
    encountered, this history is used to "shrink" the generated input values
    from Strategies by reordering, reducing, and removing the individual
    `UInt64` values in the history.
    """
```

The implementation has three phases:

#### Phase 1: Recording

During test execution, `Rng._next()` records every `UInt64` it generates into a `List[UInt64]` (the "choice sequence").

#### Phase 2: Playback

When replaying a choice sequence, `_next()` returns values from the playback list instead of generating new random values. If the playback list is exhausted - it errors out and tired a different shrinking approach.

[...]

PR #6408, opened 2026-04-11
[stdlib] Begin reducer implementation for property testing

bgreni:
## Summary

Begin an implementation of test case reduction for use in property testing.

@NathanSWard This is very naive still, as I'm writing this as I learn about it, as I find that to be effective. I just figured I would open a draft so you can take a look and we can collaborate on a solid basis to work off of going forward.

Begins work for #6379

## Checklist

      sequence of smaller PRs

### DIFF
--- a/mojo/stdlib/std/builtin/error.mojo
+++ b/mojo/stdlib/std/builtin/error.mojo
@@ -148,6 +148,7 @@ struct StackTrace(Copyable, Movable, Writable):
 
 struct Error(
     Copyable,
+    Equatable,
     Writable,
 ):
     """This type represents an Error."""
@@ -258,6 +259,20 @@ struct Error(
         if self._stack_trace:
             return String(self._stack_trace.value())
         return None
+
+    def __eq__(self, other: Self) -> Bool:
+        """Checks equality of two Error objects.
+
+        Two Error objects are considered equal if their error messages are
+        equal. Stack traces are not considered for equality.
+
+        Args:
+            other: The other Error object to compare with.
+
+        Returns:
+            `True` if the error messages are equal, `False` otherwise.
+        """
+        return self._error == other._error
 
 
 @doc_hidden
--- a/mojo/stdlib/std/testing/prop/_shrinking/_passes.mojo
+++ b/mojo/stdlib/std/testing/prop/_shrinking/_passes.mojo
@@ -1,0 +1,172 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+from std.sys.info import simd_width_of
+
+from ._shrinker import ShrinkOracle, Stream, StreamChunk
+
+
+# TODO: add larger block sizes and handle mulitples of simd width
+def block_sizes() -> List[Int]:
+    var l = List[Int]()
+    var start = simd_width_of[DType.uint64]()
+
+    while start > 0:
+        l.append(start)
+        start >>= 1
+
+    return l^
+
+
+def delete_chunk(stream: Stream, start: Int, chunk_size: Int) -> Stream:
+    var cpy = Stream()
+    for j in range(start):
+        cpy.append(stream[j])
+    for j in range(start + chunk_size, len(stream)):
+        cpy.append(stream[j])
+    return cpy^
+
+
+def binary_search_value[
+    O: ShrinkOracle, //
+](mut oracle: O, i: Int, var hi: UInt64, mut improved: Bool):
+    """Binary search the smallest value at index `i` that is still interesting.
+    """
+    var lo = UInt64(0)
+    while lo < hi:
+        var mid = lo + (hi - lo) // 2
+        var cpy = oracle.best_stream().copy()
+        cpy[i] = mid
+
+        if oracle.try_stream(cpy):
+            improved = True
+            # best was updated by try_stream, use its value
+            hi = oracle.best_stream()[i]
+        else:
+            if mid == lo:
+                break
+            lo = mid
+
+
+def multi_element_pass[
+    O: ShrinkOracle,
+    //,
+    op: def[width: Int, //](StreamChunk[width]) thin -> StreamChunk[width],
+](mut oracle: O) -> Bool:
+    ref stream = oracle.best_stream()
+    comptime for size in block_sizes():
+        if size <= len(stream):
+            for i in range(0, len(stream) - size + 1, size):
+                var cpy = stream.copy()
+                var ptr = cpy.unsafe_ptr().unsafe_offset(i)
+                ptr.unsafe_store[width=size](0, op(ptr.unsafe_load[width=size]()))
+
+                if oracle.try_stream(cpy):
+                    return True
+
+    return False
+
+
+trait ShrinkOp(Movable):
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        ...
+
+
+@fieldwise_init
+struct DeleteChunkPass(Movable, ShrinkOp):
+    """Try to delete contiguous chunks from the stream, largest first."""
+
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        ref stream = oracle.best_stream()
+        var n = len(stream)
+        var chunk_size = n - 1
+
+        while chunk_size >= 1:
+            for i in range(n - chunk_size + 1):
+                var candidate = delete_chunk(stream, i, chunk_size)
+                if oracle.try_stream(candidate):
+                    return True
+
+                # Fallback: binary search the predecessor value with the
+                # chunk deleted. This handles "length + items" generation
+                # patterns where deleting items without adjusting the
+                # preceding length causes a playback overrun.
+                if i > 0 and stream[i - 1] > 0:
+                    var hi = stream[i - 1]
+                    var improved = False
+
+                    binary_search_value(oracle, i, hi, improved)
+
+                    if improved:
+                        return True
+
+            chunk_size -= 1
+
+        return False
+
+
+@fieldwise_init
+struct ZeroChunkPass(Movable, ShrinkOp):
+    """Try to zero out contiguous chunks of the stream."""
+
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        @always_inline
+        def zero[width: Int, //](s: StreamChunk[width]) -> StreamChunk[width]:
+            return 0
+
+        return multi_element_pass[zero](oracle)
+
+
+@fieldwise_init
+struct ReduceChunkPass(Movable, ShrinkOp):
+    """Binary-search each individual element down to its smallest interesting
+    value."""
+
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        var improved = False
+
+        # TODO: Use SIMD/chunked for this as well
+        var n = len(oracle.best_stream())
+        for i in range(n):
+            ref current = oracle.best_stream()
+            var hi = current[i]
+            if hi == 0:
+                continue
+
+            binary_search_value(oracle, i, hi, improved)
+
+        return improved
+
+
+@fieldwise_init
+struct SortChunkPass(Movable, ShrinkOp):
+    """TODO: Try and find a smaller ordering of the existing data."""
+
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        return False
+
+
+@fieldwise_init
+struct RedistributePass(Movable, ShrinkOp):
+    """TODO: Adjust adjacent values while keeping their sum constant."""
+
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        return False
+
+
+@fieldwise_init
+struct SwapChunkPass(Movable, ShrinkOp):
+    """TODO: Swap adjacent chunks."""
+
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        return False
--- a/mojo/stdlib/std/testing/prop/_shrinking/_passes.mojo
+++ b/mojo/stdlib/std/testing/prop/_shrinking/_passes.mojo
@@ -1,0 +1,174 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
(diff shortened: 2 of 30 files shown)


mojo/stdlib/std/builtin/error.mojo
```diff
@@ -258,6 +259,20 @@ struct Error(
             return String(self._stack_trace.value())
         return None
 
+    def __eq__(self, other: Self) -> Bool:
```
bgreni: Probably need a better way of figuring out "is this error the same as the one I'm looking for"

mojo/stdlib/test/testing/prop/test_runner.mojo
```diff
@@ -81,5 +83,43 @@ def test_prop_test_runner_using_same_seed_produces_deterministic_results() raise
...
+    def prop(values: List[Int64]) raises:
+        var total: Int64 = 0
+        for v in values:
+            total += v
+        assert_true(total < 100)
+
+    var config = PropTestConfig(seed=1235)
+
+    # TODO: This is generating a long list of ones for some reason
+    with assert_raises():
```
bgreni: Trying to reduce a list just creates a long list of 1s for some reason

bgreni: Looks like the was because Rng.rand_bool treated zero as always true so the zero-ing pass didn't work properly, seems to be fixed now. Still need to make it smarter to find a more optimal solution, but it's a start.

NathanSWard:
I'll take a look today/tomorrow and provide some early feedback!
If there's anything specific you want me to look at/focus on, let me know!

bgreni:
> I'll take a look today/tomorrow and provide some early feedback! If there's anything specific you want me to look at/focus on, let me know!

Thank! Nothing really in particular just some feedback about the general structure/approach. I'm not really attached to anything I have here so far.

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
```diff
@@ -0,0 +1,182 @@
...
+from std.testing.prop._errors import PLAYBACK_EXHAUSTED
+from std.collections import Set
+from std.sys.info import simd_width_of
+from std.os import abort
+
+comptime _Stream = List[UInt64]
+comptime _StreamChunk[width: Int] = SIMD[DType.uint64, width]
+
+
+def __block_sizes() -> List[Int]:
```
NathanSWard: small nit, no need for the "double underscore" or even "single underscore" for any of the methods/types in the module. Since this is a private `_shrinker.mojo` module, you can keep everything normally named and make it bit easier on the eyes.

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
```diff
@@ -0,0 +1,182 @@
...
+                    var cpy = stream.copy()
+                    var ptr = cpy.unsafe_ptr() + i
+                    ptr.store[size](0, op(ptr.load[size]()))
+
+                    if self.try_stream(cpy):
+                        return True
+
+        return False
+
+    def reduce_chunk_pass(mut self, stream: _Stream) -> Bool:
```
NathanSWard: Can we pull all of these "shrinking operations" into separate structs/methods that aren't tied to this larger `_Shrinker` so we can more easily unit test them?
Something like:
```
trait ShrinkOp:
    def shrink(chunk: _StreamChunk[_]) -> type_of(chunk):
        ...
```
(it doesn't have to be behind a trait but could just be a free standing method too)

NathanSWard (commented):
This is a great first pass. 😄 
Great job on the `block_sizes()` as this is an important optimization in shrinking algorithms since often times shrinking large chunks (e.g. 4 or 8 blocks) will produce a smaller failing input faster than incrementally doing 1 block at a time.

So overlooking the current shrinking ops we have:
- zeroing
- reduce
- delete

From what I've seen about shrinking strategies I'd recommend the following

**Replace `reduce_chunk_pass` with binary search minimization**

The current `halve` and `reduce_10` ops are a bit ad-hoc. I've seen algorithm which binary-searches between `0` and the current value for each element - much more systematic and converges in O(log n) test runs instead of many linear steps. The refactor would be: for each index `i`, binary search `[0, stream[i]]` by writing candidate values into a copy of the stream and calling `try_stream`.

**Add `sort_chunk_pass`**

I've seen passes that sorts a chunk in ascending order and tests the result. This helps find simpler orderings that still fail (important for collection-valued inputs). This would just be a new pass alongside `zero_chunk_pass` et al.

Notably, we want to skip the 1-sized block retruend from `__block_sizes()` since sorting a 1-sized chunk doesn't mean much 🙃 

**Add `redistribute_pass`**

This pass keeps the sum of two adjacent choices constant while moving weight toward the left (smaller index). This is powerful when two stream values are correlated (e.g. a length field and a value field). It first tries swapping if they're out of order, then binary-searches the redistribution.

**Add `swap_chunk_pass`**

This pass swaps a chunk with its right neighbor. Cheap to implement and helps reorder correlated values.

**`delete_chunk_pass` is missing the dec-previous optimization**

The delete pass could also try decrementing the value *before* the deleted chunk. This handles the common "length + items" generation pattern (deleting an item without adjusting the preceding length would cause an overrun). Worth adding the same fallback logic.

---

If you didn't want to implement these all at once (which would be better as it makes this PR easier to review) you could just stub them out for now and we can implement them in follow-ups!

bgreni:
> This is a great first pass. 😄 Great job on the `block_sizes()` as this is an important optimization in shrinking algorithms since often times shrinking large chunks (e.g. 4 or 8 blocks) will produce a smaller failing input faster than incrementally doing 1 block at a time.
> 
> So overlooking the current shrinking ops we have:
> 
> * zeroing
> * reduce
> * delete
> 
> From what I've seen about shrinking strategies I'd recommend the following
> 
> **Replace `reduce_chunk_pass` with binary search minimization**
> 
> The current `halve` and `reduce_10` ops are a bit ad-hoc. I've seen algorithm which binary-searches between `0` and the current value for each element - much more systematic and converges in O(log n) test runs instead of many linear steps. The refactor would be: for each index `i`, binary search `[0, stream[i]]` by writing candidate values into a copy of the stream and calling `try_stream`.
> 
> **Add `sort_chunk_pass`**
> 
> I've seen passes that sorts a chunk in ascending order and tests the result. This helps find simpler orderings that still fail (important for collection-valued inputs). This would just be a new pass alongside `zero_chunk_pass` et al.
> 
> Notably, we want to skip the 1-sized block retruend from `__block_sizes()` since sorting a 1-sized chunk doesn't mean much 🙃
> 
> **Add `redistribute_pass`**
> 
> This pass keeps the sum of two adjacent choices constant while moving weight toward the left (smaller index). This is powerful when two stream values are correlated (e.g. a length field and a value field). It first tries swapping if they're out of order, then binary-searches the redistribution.
> 
> **Add `swap_chunk_pass`**
> 
> This pass swaps a chunk with its right neighbor. Cheap to implement and helps reorder correlated values.
> 
> **`delete_chunk_pass` is missing the dec-previous optimization**
> 
> The delete pass could also try decrementing the value _before_ the deleted chunk. This handles the common "length + items" generation pattern (deleting an item without adjusting the preceding length would cause an overrun). Worth adding the same fallback logic.
> 
> If you didn't want to implement these all at once (which would be better as it makes this PR easier to review) you could just stub them out for now and we can implement them in follow-ups!

Thanks this is really great stuff. I'll implement the immediately applicable stuff and stub out the new passes with `TODOs` to add in follow up PRs

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
bgreni: I approached it like that originally, but since I found it easiest to encapsulate things into `try_stream` I would have to inject that callback back into the shrink operations anyways? Do you have an idea for a better approach?

mojo/stdlib/std/testing/prop/random.mojo
```diff
@@ -128,8 +153,38 @@ struct Rng(Movable):
             var uint64 = self._next(diff)
             return Scalar[dtype](uint64) + min
         elif dtype.is_floating_point():
-            var f = self._xoshiro_float()
-            var result = Float64(min) * (1.0 - f) + Float64(max) * f
+            # Shrink-monotonic decoding: smaller UInt64 must produce a float
+            # with smaller (or equal) |result|, with u=0 producing 0.0. The
+            # shrinker reduces the raw stream bits, so the encoding is what
+            # makes shrinking converge to small magnitudes rather than to the
```
bgreni, line 183: Added a slightly more complicated test case and realized floats don't reduce properly using the `_xoshiro_float` based approach, so Claude whipped this up for me. It works better, but perhaps this will need a closer look at some point as I'm not sure if there is a better way of doing this.

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
bgreni: nvm I have something that seems nice enough for now

bgreni:
@NathanSWard I'm just mostly tinkering with the implementation at this point, but I think it should be ready for a proper review at this point.

NathanSWard:
@bgreni - Awesome thanks for pinging me. I'll take a look on Monday :)

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
```diff
@@ -0,0 +1,132 @@
...
+            return False
+        except e:
+            var used = Stream(stream[: rng.playback_index])
+            if e == self.expected_error and stream_is_better(used, self.best):
+                self.best = used^
+                return True
+            return False
+
+    def best_stream(self) -> Stream:
+        return self.best.copy()
```
NathanSWard: I see this will likely introduce a lot of copying of potentially large lists. Could we instead update the trait/type to return a `ref [self] Stream`?

mojo/stdlib/std/testing/prop/_shrinking/_passes.mojo
```diff
@@ -0,0 +1,172 @@
...
+
+        return improved
+
+
+@fieldwise_init
+struct SortChunkPass(Movable, ShrinkOp):
+    """TODO: Try and find a smaller ordering of the existing data."""
+
+    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
+        return False
```
NathanSWard, line 156: I like this trait a lot :)

NathanSWard, line 156:
```suggestion
@fieldwise_init
struct SortChunkPass(Movable, ShrinkOp):
    """TODO: Try and find a smaller ordering of the existing data."""

    def run[O: ShrinkOracle, //](mut self, mut oracle: O) -> Bool:
        ...
```
Let's not provide a default implementation of `run` instead enforce the implementor of this trait to defined one.

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
```diff
@@ -0,0 +1,132 @@
...
+        var delete_pass = DeleteChunkPass()
+        var zero_pass = ZeroChunkPass()
+        var reduce_pass = ReduceChunkPass()
+        while reductions < self.max_reductions:
+            runs += 1
+
+            if not (
+                delete_pass.run(self)
+                or zero_pass.run(self)
+                or reduce_pass.run(self)
```
NathanSWard, line 99: *question*: is there a specific reason for the order of these passes or are they arbitrary?
IIRC some property testing frameworks do them in a specific order to try and minimize shrinking time.

mojo/stdlib/std/testing/prop/random.mojo
```diff
@@ -121,15 +147,68 @@ struct Rng(Movable):
         comptime if dtype == DType.bool:
             return rebind[Scalar[dtype]](Scalar[DType.bool](self.rand_bool()))
         elif dtype.is_integral():
-            var offset = UInt64(0) - UInt64(Scalar[dtype].MIN)
-            var a = UInt64(min) + offset
-            var b = UInt64(max) + offset
-            var diff = a - b if a > b else b - a
-            var uint64 = self._next(diff)
-            return Scalar[dtype](uint64) + min
+            comptime bits = bit_width_of[dtype]()
```
NathanSWard, line 150: Can you expand a bit on why the rng implementation for integral d-types had to change here?

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
bgreni: I've ordered them with a general heuristic of "what _shrinks_ the most if successful?". I'd like to add some benchmarking that actually counts the number of reduction attempts to make a more data-informed decision on the ordering in the future though.

mojo/stdlib/std/testing/prop/random.mojo
bgreni: I noticed that since we were only pulling a single `u64` even for longer 128/256 bit integers we wouldn't be testing the full range of their values. We can revert this if you want since it's not directly related to the rest of the PR

mojo/stdlib/std/testing/prop/_shrinking/_passes.mojo
bgreni: I'm not sure what you mean, `ShrinkOp` already doesn't have a default impl

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
bgreni: A bit odd that this gives me a compiler error

```mojo
def best_stream(self) -> ref[self] Stream: # error: cannot return 'self's origin, because it might expand to a RegisterPassable type
        ...
```

This seems ok but could lead to accidental mutations right?

```mojo
def best_stream(ref self) -> ref[self] Stream:
        ...
```

mojo/stdlib/std/testing/prop/_shrinking/_passes.mojo
NathanSWard: Ohh apologies, I thought this was the trait... 🙃

mojo/stdlib/std/testing/prop/_shrinking/_shrinker.mojo
NathanSWard: You can do
```mojo
def best_stream[o: ImmutOrigin, //](ref [o] self) -> ref [o] Stream:
        ...
```

bgreni: I guess that sort of works, but then you need to do an origin cast

```mojo
def best_stream[o: ImmutOrigin, //](ref [o] self) -> ref [o] Stream:
        return UnsafePointer(to=self.best).unsafe_origin_cast[o]()[]
```

NathanSWard:
@bgreni - I'm taking another look/pass over this now. :)

mojo/stdlib/std/builtin/error.mojo
NathanSWard: What is the use of this equality for error? I don't believe we want to add this to the `Error` type as if someone wants this behavior they should just manually comparing the stringificatino of two errors.

bgreni: Its for the criteria of "a shrink is only valid if it produces the _same_ error as the original test case". Comparing the stringified versions is probably fine like you said though

PR #6414, opened 2026-04-13
[stdlib] Add default indexing for all iterators

martinvuyk:
Add default indexing for all iterators. Phase 0 of https://github.com/modular/modular/issues/6359

### DIFF
--- a/mojo/stdlib/std/iter/__init__.mojo
+++ b/mojo/stdlib/std/iter/__init__.mojo
@@ -54,6 +54,8 @@ for squared in map[square](values):
 """
 
 from std.builtin.constrained import _constrained_conforms_to
+from std.builtin.builtin_slice import ContiguousSlice, StridedSlice
+from std.os import abort
 
 # ===-----------------------------------------------------------------------===#
 # Iterable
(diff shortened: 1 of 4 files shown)


mojo/stdlib/std/iter/__init__.mojo
```diff
@@ -175,6 +253,72 @@ trait Iterator(ImplicitlyDestructible, Movable):
         """
         return (0, None)
 
+    def __getitem__(
```
NathanSWard: I think this may need more discussion/design. While I do think it's important to be able to get the "nth" element from an iterator, using the `__getitem__` (`[n]`) syntax I don't believe is the best choice. Since usually `[n]` syntax implies a fast lookup. Also since this requires us to take the iterator by `var` (to ensure consecutive calls to `[n]` produce the same value) this requires a copy of iterators which may not necessarily be cheap. 

Instead I would prefer a method like `nth(index: Int) -> Optional[Element]` (name to be bikeshed). This is more explicit and doesn't give a false sense of similarity to getitem.

mojo/stdlib/std/iter/__init__.mojo
```diff
@@ -175,6 +253,72 @@ trait Iterator(ImplicitlyDestructible, Movable):
...
+        var j = 0
+        while True:
+            var val = trait_downcast_var[ImplicitlyDestructible & Movable](
+                next(self)
+            )
+            if i == j:
+                return val^
+            j += 1
+
+    def __getitem__(
```
NathanSWard, line 294: I'm also not totally sold on the need to have special casing for slices. Since you can Have this exact same behavior with a combination of iterator adapters - though the adaptors are more verbose.

NathanSWard: All that said I'm not totally against this, I just want to be careful with people confusing collection slicing w: iterator slicing (though perhaps I'm just being too caution lol) 😅

mojo/stdlib/std/iter/__init__.mojo
NathanSWard: I may be being a bit too caution and allowing indexing into iterators may be nice (I would prefer it return an `Optional`)

Thoughts @lsh @ConnorGray ?

martinvuyk: Iterators that aren't trivial to copy would require the user to do either `iter^[2]` or `iter.copy()[2]`, so the cost would be obvious IMO. Also still all of these would raise `StopIteration` when OOB so it's also not easy to abuse by accident

mojo/stdlib/std/iter/__init__.mojo
martinvuyk: Making iterator slicing feel natural is exactly my goal with this :) . I want people to just use any return of any one of the string iterators (or any others) and make it feel as if they were dealing with lists or views over the data. I also quite like the composability of this in general with any set of generic iterator operations, making them all just raise `StopIteration` to signify any OOB or stop signal instead of doing a bunch of optional checking everywhere

mojo/stdlib/std/iter/__init__.mojo
martinvuyk: I ended up following the lead of our `range` iterators and just aborted on OOB access for `def __getitem__(var self, idx: Some[Indexer])`. I also updated the `range` iterators to receive `var self` for indexing. We should still add some faster functions for iterators like that which don't actually need to iterate to yield slices of themselves, these are just the default implementations

NathanSWard:
Hey @martinvuyk ! We talked this through as a team as we'd be open to the single element get of an iterator, however, we are not too sure about the slicing functionality.

There's a bit of an issue we have with iterators where they are still a WIP and unfortunately likely will not hit stability for 1.0 (unless we get a huge rework to the origin system). Since we have no way to express an `Iterator` that returns references. Currently this is hacked into the compiler and because of this we ideally want to reduce churn on iterators.

That being said, we'd be open to having some sort of `nth(var self, idx: Int) -> Optional[Self.Element]` method like you have here. However we're not sold on the indexing syntax via `__getitem__`.

If you could break this apart into the two parts or even just move the iterator slicing to its own GH issue first so we can have design discussion there since these are both pretty bit API features for `Iterator` and we need to have a discussion about them first.

martinvuyk:
@NathanSWard what would you guys think about hiding all of this inside a trait `IndexableIterator` that could be public or private, but we treat as separate and experimental for now?

Without this basis I won't be able to easily provide a path for people to interact with the string APIs when they get switched to returning iterators and the places where they previously sliced the `List` returns will break.

Also, I wasn't quite intending on marking any of this stable before or on 1.0 . I would like to make the switch as soon as possible to avoid much breakage in the future, but I know this won't be fast

NathanSWard:
I'm going to close this PR as we now have `Iterator.nth` which is synonymous with iterator indexing.
Please re-open to file a new issues if `Iterator.nth` doesn't convert all of these cases :)

PR #6420, opened 2026-04-16
[NFC] Fix docstrings in std/collections

martinvuyk:
Fix docstrings in std/collections

### DIFF
--- a/mojo/stdlib/std/collections/__init__.mojo
+++ b/mojo/stdlib/std/collections/__init__.mojo
@@ -10,7 +10,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===----------------------------------------------------------------------=== #
-"""Core data structures: List, Dict, Set, Optional, plus specialized collections.
+"""Core data structures: List, Dict, Set, Optional, plus specialized
+collections.
 
 The `collections` package provides fundamental data structures for organizing
 and managing data in Mojo programs. It includes general-purpose collections
--- a/mojo/stdlib/std/collections/_conditional.mojo
+++ b/mojo/stdlib/std/collections/_conditional.mojo
@@ -15,7 +15,9 @@
 from std.builtin.device_passable import DevicePassable
 from std.utils.type_functions import ConditionalType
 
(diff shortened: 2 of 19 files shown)


fairlydurable:
Thanks for the PR. I appreciate the effort you put into this.

Could you break this into a few [smaller, more focused](https://github.com/modular/modular/blob/870957719ce8d513441cdb082c967b997288634f/mojo/CONTRIBUTING.md#about-pull-request-sizes) PRs? That will make it easier to review and land.

Also, summary strings should follow [PEP-257](https://peps.python.org/pep-0257): single line, start with a capital letter, and end with a period. This will likely need a bit of hand-editing.

Thanks again for the contribution. This is helpful.

martinvuyk:
> Could you break this into a few [smaller, more focused](https://github.com/modular/modular/blob/870957719ce8d513441cdb082c967b997288634f/mojo/CONTRIBUTING.md#about-pull-request-sizes) PRs? That will make it easier to review and land.

I've done so in other instances and I ended up flooding the repo with very few line fixes, and it was more noise than necessary.

> Also, summary strings should follow [PEP-257](https://peps.python.org/pep-0257): single line, start with a capital letter, and end with a period. This will likely need a bit of hand-editing.

Yeah I don't know why you mention that. If it's because of the line-breaks, then the summary is allowed to have a line-break (or several) and a dot at the end, our docstring linter tool allows it. We have a lot of code that follows this pattern.

while the pep says `it is important that it [summary] fits on one line` not even python follows this as there are some summaries which require line-breaks in between the text (just due to the complexity of what the functions do). And we require it especially since we use an 80 character limit.

> Thanks again for the contribution. This is helpful.

np, I've been trying to keep docstrings consistent since this repo was open sourced :)

---

We actually still have a lot of places where we are even allowing 2-space indentation in docstrings and our linter isn't catching those inconsistencies. I also keep finding instances of `# Example` and `Note:` instead of `Examples:` and `Notes:`, and a lot of other things. I've been actually asking for some better enforcing of other parts of our docstring standard for some time now in #3235

PR #6427, opened 2026-04-19
[External] [stdlib] Add support for NVIDIA RTX 2070 SUPER

ThomasMader:
I ran the following two commands to test the changes successfully.

```
./bazelw build //mojo/stdlib/std

./bazelw test //mojo/stdlib/test/gpu/...
```


JoeLoser:
LGTM, @BradLarson do you mind taking a peak as well?

ThomasMader:
I think this PR is not needed anymore.
I updated my environment to the latest state and I am now at: Mojo 1.0.0b2.dev2026060306 (e15be343)

With that running gpu_check.mojo detects my GPU successfully and running the mojo-gpu-puzzles tests also succeeds for all of the tests which are not skipped due to insufficient compute capability.

I don't know why that works now because as far as I know it was always needed to make the changes I did in those files in order to get it working.
I also checked the current commit of the repository (7aa053560034c8c5b4f9acb0a5b450e79d2f7c18) if I see changes to those files which might explain why it is not necessary anymore but didn't find anything.

I guess it's a misunderstanding on my side.

BradLarson:
The Bazel changes would still be necessary if you were building the Mojo standard library, that needs to match against the text name of your GPU, but the existing `sm_75` registration should have supported your GPU from the beginning at the Mojo level. I think we may have had too-strict checking in the separate GPU detection script in the Mojo GPU puzzles repo, and we reworked that to be a little more robust a few months ago.

I think our one hesitation about pulling this in was whether the new 2070 entries were necessary in gpu.host with the existing 2060 ones, plus the commenting-out of the older 2060 entries.

ThomasMader:
> The Bazel changes would still be necessary if you were building the Mojo standard library, that needs to match against the text name of your GPU, but the existing sm_75 registration should have supported your GPU from the beginning at the Mojo level. I think we may have had too-strict checking in the separate GPU detection script in the Mojo GPU puzzles repo, and we reworked that to be a little more robust a few months ago.

Thanks for the clarification.
I reviewed the changes in the [Mojo GPU puzzles repo](https://github.com/modular/mojo-gpu-puzzles/commit/e780dacbaf48b3f14cb1b5730d903a412af20e9f#diff-75eb454d720ef6378d3acbf8571ad45aeee3d8fbb401a21794bbbdda2f128831), and I have a much better understanding of the current detection logic now.

> I think our one hesitation about pulling this in was whether the new 2070 entries were necessary in gpu.host with the existing 2060 ones, plus the commenting-out of the older 2060 entries.

I initially shared that concern because removing the older 2060 entries felt risky. However, seeing that the same pattern was applied for the GTX1060 vs. GTX1080Ti (where also only the sm_count differed), I thought it might be ok to do that.

For me it doesn't matter if this is not merged soon because currently I have anyway other things to look into.
In the worst case I can define the target myself via the command line parameter as far as I understood in the meantime.

Looking forward to see how the refactoring to support multiple GPUs per target_arch is going to look like in the future when it is available.

Is the very long term idea for those GPU specifications to be integrated by the vendors themselves in their own MLIR dialects or something?

PR #6436, opened 2026-04-21
[stdlib] Simplify and modernize `_PyIter`, and fix small mem-leaks

martinvuyk:
Simplify and modernize `_PyIter`, and fix small mem-leaks.

This also makes the use of `PythonVersion` nicer by adding default implicit constructors from tuples and comparison operations. It also fixes many call-sites that would fail if Python ever realeased a 4.0 version to use the 3 major version for the comparisons.

### DIFF
--- a/mojo/stdlib/std/python/_cpython.mojo
+++ b/mojo/stdlib/std/python/_cpython.mojo
@@ -240,7 +240,7 @@ struct PyObjectPtr(
 
 
 @fieldwise_init
-struct PythonVersion(ImplicitlyCopyable, RegisterPassable):
+struct PythonVersion(Comparable, ImplicitlyCopyable, RegisterPassable):
     """Represents a Python version with major, minor, and patch numbers."""
 
     var major: Int
@@ -249,6 +249,28 @@ struct PythonVersion(ImplicitlyCopyable, RegisterPassable):
     """The minor version number."""
     var patch: Int
     """The patch version number."""
+
+    @implicit
+    def __init__(out self, version: Tuple[Int, Int, Int]):
+        """Construct a version with a Tuple.
+
+        Args:
+            version: The version values.
+        """
+        self.major = version[0]
+        self.minor = version[1]
+        self.patch = version[2]
+
+    @implicit
+    def __init__(out self, version: Tuple[Int, Int]):
+        """Construct a version with a Tuple.
+
+        Args:
+            version: The version values.
+        """
+        self.major = version[0]
+        self.minor = version[1]
+        self.patch = 0
 
     def __init__(out self, version: StringSlice):
         """Initialize a PythonVersion object from a version string.
@@ -276,6 +298,22 @@ struct PythonVersion(ImplicitlyCopyable, RegisterPassable):
                 start = next_idx + 1
             next_idx += 1
         self = PythonVersion(components[0], components[1], components[2])
+
+    @always_inline
+    def __lt__(self, other: Self) -> Bool:
+        """Compare self to a full version.
+
+        Args:
+            other: The other version.
+
+        Returns:
+            The result.
+        """
+        return (self.major, self.minor, self.patch) < (
+            other.major,
+            other.minor,
+            other.patch,
+        )
 
 
 def _py_get_version(lib: _DLHandle) -> StaticString:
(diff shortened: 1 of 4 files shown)


mojo/stdlib/std/python/python_object.mojo
```diff
@@ -75,15 +75,26 @@ struct _PyIter(ImplicitlyCopyable, Iterable, Iterator):
...
             points to.
         """
-        if not self.next_item:
-            raise StopIteration()
         ref cpy = Python().cpython()
-        var curr_item = self.next_item
-        self.next_item = cpy.PyIter_Next(self.iterator._obj_ptr)
-        return PythonObject(from_owned=curr_item)
+        if cpy.version >= (3, 14):
+            var elem = PythonObject()
```
JoeLoser: This leaks one `Py_None` strong reference per `__next__` call. `PythonObject()` → `Self(None)` → `Self(from_borrowed=cpy.Py_None())`, which `Py_IncRef`s `Py_None`. The next line then writes a new pointer directly into `elem._obj_ptr` through `UnsafePointer(to=...)`, clobbering the field with no `Py_DecRef`.

To fix this, I think this should work:

```mojo
var ptr = PyObjectPtr()  # null
var err = cpy.PyIter_NextItem(
    self.iterator._obj_ptr, UnsafePointer(to=ptr)
)
if err < 1:
    raise StopIteration()
return PythonObject(from_owned=ptr)
```

The existing test exercises `PyIter_NextItem` against a raw `PyObjectPtr`, so it doesn't see this leak. A Mojo-side test that iterates a finite iterable to exhaustion via `_PyIter` and asserts `sys.getrefcount(None)` is unchanged would bind this behavior.

mojo/stdlib/std/python/python_object.mojo
```diff
@@ -75,15 +75,26 @@ struct _PyIter(ImplicitlyCopyable, Iterable, Iterator):
...
+            var err = cpy.PyIter_NextItem(
+                self.iterator._obj_ptr, UnsafePointer(to=elem._obj_ptr)
+            )
+            # FIXME: a return of -1 is an error, we should deal with it
+            # but we currently don't have a way to propagate it from here and
+            # still conform to the Iterator trait.
+            if err < 1:
```
JoeLoser: I think there's a few problems here:

1. `err < 1` collapses `0` (no more items) and `-1` (error) into `StopIteration`. On `-1`, CPython has set the error indicator; we swallow the exception and leave the indicator set, so the *next* CPython call will see a spurious pending exception and behave incorrectly. At minimum:

   ```mojo
   if err == -1:
       cpy.PyErr_Clear()  # don't leak the indicator into the next FFI call
       raise StopIteration()
   if err == 0:
       raise StopIteration()
   ```

2. The FIXME says "we don't have a way to propagate", but the constraint is the `raises StopIteration` clause on `__next__`. Worth widening that contract so real errors can propagate, or at least filing a ticket. The legacy `PyIter_Next` path has the same latent bug (NULL is ambiguous), but it can't be fixed there without an extra `PyErr_Occurred` check — which would be cheap to add for parity.

Style quibble: `err != 1` reads nicer than `err < 1`.

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -250,6 +250,28 @@ struct PythonVersion(ImplicitlyCopyable, RegisterPassable):
     var patch: Int
     """The patch version number."""
 
+    @implicit
```
JoeLoser: **Thought** Marking both the 2- and 3-tuple constructors `@implicit` is a broad implicit conversion from a somewhat generic type. The call sites read nicely (`self.version >= (3, 12)`), but anywhere a `PythonVersion` is expected, any `Tuple[Int, Int]` will silently convert.

Consider either dropping `@implicit` here or limiting `@implicit` to the 3-tuple variant only — the 2-tuple form defaulting `patch = 0` is the more surprising one to apply implicitly.

mojo/stdlib/std/python/python.mojo
```diff
@@ -380,7 +380,7 @@ struct Python(Defaultable, ImplicitlyCopyable):
     def dict[
         V: ConvertibleToPython & Copyable = PythonObject
     ](**kwargs: V) raises -> PythonObject:
-        """Construct an Python dictionary from keyword arguments.
+        """Construct a Python dictionary from keyword arguments.
```
JoeLoser, line 380: **Quibble** while you're cleaning up typos: `python.mojo` still has `"an Python"` at lines 406, 466, and 492. Quick to clean up the rest.

JoeLoser (changes requested):

mojo/stdlib/std/python/python_object.mojo
martinvuyk: I opened https://github.com/modular/modular/issues/6579, I'll add it to the comment for reference and update the code to explicitly clear the error

mojo/stdlib/std/python/_cpython.mojo
martinvuyk: I get your concern, I'll just change it to a constructor with a default patch version of zero

mojo/stdlib/std/python/python_object.mojo
martinvuyk: nice catch, I didn't realize that `PythonObject()` was incrementing a ref to `None`. Though the example you gave is also repeating a leak that I've seen in several places: the raising branch will leak if we don't build the object that takes ownership of the ref

PR #6442, opened 2026-04-22
[stdlib] Add comptime assert to tuple getitem

martinvuyk:
Add comptime assert to tuple getitem. Caught by @gabrieldemarmiesse in a discord convo

### DIFF
--- a/mojo/stdlib/std/builtin/tuple.mojo
+++ b/mojo/stdlib/std/builtin/tuple.mojo
@@ -218,6 +218,10 @@ struct Tuple[*element_types: Movable](
         Returns:
             A reference to the specified element.
         """
+        comptime assert (
+            idx >= 0
+        ), "negative indexing is not supported, use e.g. `x[len(x) - 1]`"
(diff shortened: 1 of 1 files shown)


NathanSWard (approved):
Thank you for catching this!
I'll add this internally but I believe you can use a t-string in the assert message to provide better error diagnostics :)

NathanSWard:
!sync

martinvuyk:
@NathanSWard could you sync this again? I tried using the Indexer trait but it won't work, so I went back to using Int

>  I'll add this internally but I believe you can use a t-string in the assert message to provide better error diagnostics :)

Yeah I agree, I think we should revisit all other `__getitem_param__` like `InlineArray`'s to give them some clearer messages (I just copied these from there)

NathanSWard:
!sync

rd4com:
Hi, 👍 
Not sure if it could be baked into `TypeList` struct,
so that all things that uses it get the comp time assert ?
(And not sure if desirable to do it in TypeList too)

(ping to @NathanSWard)

PR #6444, opened 2026-04-22
[bazel] Refactor wheel repositories to split Mojo and MAX versions

Ahajha:
They're related currently, but preparing for a split.


Ahajha:
!sync

PR #6453, opened 2026-04-24
[Stdlib] Add CPython protocol builders for Mojo-defined Python types

winding-lines:
BEGIN_PUBLIC
[Stdlib] Add CPython protocol builders for Mojo-defined Python types

Adds a family of protocol builders under `std.python` that let Mojo structs exposed as Python types implement the standard CPython type protocols without writing CPython slot wrappers by hand.

New builders, each constructed from a `PythonTypeBuilder`:

  - `BufferProtocolBuilder` — `bf_getbuffer` / `bf_releasebuffer`, enabling zero-copy `memoryview(obj)` / `numpy.frombuffer(obj)`.
  - `MappingProtocolBuilder` — `mp_length`, `mp_subscript`, `mp_ass_subscript` (assignment and deletion).
  - `NumberProtocolBuilder` — full `PyNumberMethods` surface: unary (`__abs__`, `__neg__`, `__pos__`, `__invert__`, `__int__`, `__float__`, `__index__`), inquiry (`__bool__`), binary (arithmetic, bitwise, in-place), and ternary (`__pow__`, `__ipow__`).
  - `SequenceProtocolBuilder` — `sq_length`, `sq_item`, `sq_ass_item`, `sq_contains`, `sq_concat`, `sq_repeat`.
  - `TypeProtocolBuilder` — `tp_richcompare` for the six comparison operators.

Each `def_*` method accepts the corresponding Python dunder signature and overloads on receiver kind (pointer, value, mut) and on raising vs non-raising methods. Binary and ternary number methods may raise `NotImplementedError` to return `Py_NotImplemented`, triggering Python's reflected-operation fallback. Number methods may also return any `ConvertibleToPython` type rather than `PythonObject` directly.

Supporting additions:

  - `std.python.adapters` — internal C ABI wrappers, `_PySlotIndex` constants, and lift/conv helpers shared by all builders.
  - `std.python.utils` — `RichCompareOps` constants and a `NotImplementedError` sentinel.
  - `std.python.builders` — re-exports the four `*ProtocolBuilder` types as a convenience entry point.

Adds a `columnar` example under `mojo/examples/python-interop` demonstrating buffer-protocol interop, and integration tests covering each protocol.
END_PUBLIC

## Summary

Extend Python extension capabilities provided by the standard library. This PR adds support for:

 - mapping protocol — obj[key], len(obj), obj[key] = val
 - number protocol — arithmetic operators, abs(), bool(), etc.
 - sequence protocol — indexed access, in operator, concatenation, repetition
 - rich comparison — ==, !=, <, <=, >, >=

Assisted-by: Claude Opus 4.7 (Anthropic) <noreply@anthropic.com>

## Testing

Adds a `columnar` example under `mojo/examples/python-interop` demonstrating buffer-protocol interop, and integration tests covering each protocol.

Additionally the standalone library that this code comes from (https://pontoneer.dev) is used by https://github.com/kszucs/marrow

## Checklist

- [ ] PR is small and focused — consider splitting larger changes into a
      sequence of smaller PRs

### DIFF
--- a/mojo/examples/python-interop/BUILD.bazel
+++ b/mojo/examples/python-interop/BUILD.bazel
@@ -46,6 +46,21 @@ modular_py_binary(
     ],
 )
 
+modular_py_binary(
+    name = "columnar",
+    srcs = ["columnar.py"],
+    data = ["columnar_mojo.mojo"],
+    imports = ["."],
+    target_compatible_with = select({
+        "//:asan": ["@platforms//:incompatible"],
+        "//:ubsan": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    }),
+    deps = [
+        "@mojo//:python",
+    ],
+)
+
 modular_run_binary_test(
     name = "hello_test",
     binary = "hello",
@@ -60,4 +75,9 @@ modular_run_binary_test(
     name = "mandelbrot_test",
     binary = "mandelbrot",
     tags = ["gpu"],
+)
+
+modular_run_binary_test(
+    name = "columnar_test",
+    binary = "columnar",
 )
--- a/mojo/examples/python-interop/BUILD.bazel
+++ b/mojo/examples/python-interop/BUILD.bazel
@@ -46,6 +46,22 @@ modular_py_binary(
     ],
 )
 
+modular_py_binary(
+    name = "columnar",
+    srcs = ["columnar.py"],
+    data = ["columnar_mojo.mojo"],
+    imports = ["."],
+    target_compatible_with = select({
+        "//:asan": ["@platforms//:incompatible"],
+        "//:tsan": ["@platforms//:incompatible"],
+        "//:ubsan": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    }),
+    deps = [
+        "@mojo//:python",
+    ],
+)
+
 modular_run_binary_test(
     name = "hello_test",
     binary = "hello",
@@ -60,4 +76,9 @@ modular_run_binary_test(
     name = "mandelbrot_test",
     binary = "mandelbrot",
     tags = ["gpu"],
+)
+
+modular_run_binary_test(
+    name = "columnar_test",
+    binary = "columnar",
 )
--- a/mojo/stdlib/std/python/adapters.mojo
+++ b/mojo/stdlib/std/python/adapters.mojo
@@ -1,0 +1,1325 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+# ===----------------------------------------------------------------------=== #
+# CPython slot adapter functions introduced in:
+# https://github.com/modular/modular/pull/5562
+#
+# These adapt user-friendly Mojo function signatures to the low-level C ABI
+# expected by each CPython type slot.  They are passed to
+# PontoneerTypeBuilder.def_method as template parameters.
+# ===----------------------------------------------------------------------=== #
+
+from std.ffi import c_int, c_long
+from std.memory import OpaquePointer, UnsafePointer
+from std.os import abort
+from std.python import Python, PythonObject
+from std.python._cpython import PyObject, PyObjectPtr, Py_ssize_t, PyType_Slot
+from std.python.bindings import PythonTypeBuilder
+from std.python.conversions import ConvertibleToPython
+from std.utils import Variant
+
+from .utils import NotImplementedError
+
+
+@always_inline
+def _unwrap_self[
+    T: ImplicitlyDestructible
+](py_self: PyObjectPtr) -> UnsafePointer[T, MutAnyOrigin]:
+    """Downcast a raw PyObjectPtr to a typed Mojo pointer, aborting on failure.
+    """
+    try:
+        return PythonObject(from_borrowed=py_self).downcast_value_ptr[T]()
+    except e:
+        abort(
+            String("Python method receiver did not have the expected type: ", e)
+        )
+
+
+def _mp_length_wrapper[
+    self_type: ImplicitlyDestructible,
+    method: def(UnsafePointer[self_type, MutAnyOrigin]) thin raises -> Int,
+](py_self: PyObjectPtr) abi("C") -> Py_ssize_t:
+    """CPython `lenfunc` adapter for the `mp_length` slot (__len__).
+
+    Parameters:
+        self_type: The Mojo struct type whose instances back the Python object.
+        method: User function `def(self: UnsafePointer[self_type, MutAnyOrigin]) raises -> Int`.
+
+    Returns:
+        Length as `Py_ssize_t`, or -1 with an exception set on error.
+    """
+    ref cpython = Python().cpython()
+    try:
+        var result = method(_unwrap_self[self_type](py_self))
+        return Py_ssize_t(result)
+    except e:
+        var error_type = cpython.get_error_global("PyExc_Exception")
+        var msg = String(e)
+        cpython.PyErr_SetString(
+            error_type, msg.as_c_string_slice().unsafe_ptr()
+        )
+        return Py_ssize_t(-1)
+
+
+def _mp_subscript_wrapper[
+    self_type: ImplicitlyDestructible,
+    method: def(
+        UnsafePointer[self_type, MutAnyOrigin], PythonObject
+    ) thin raises -> PythonObject,
+](py_self: PyObjectPtr, key: PyObjectPtr) abi("C") -> PyObjectPtr:
+    """CPython `binaryfunc` adapter for the `mp_subscript` slot (__getitem__).
+
+    Parameters:
+        self_type: The Mojo struct type whose instances back the Python object.
+        method: User function `def(self: UnsafePointer[self_type, MutAnyOrigin], key: PythonObject) raises -> PythonObject`.
+
+    Returns:
+        New reference to the result, or null with an exception set on error.
+    """
+    ref cpython = Python().cpython()
+    try:
+        var result = method(
+            _unwrap_self[self_type](py_self),
+            PythonObject(from_borrowed=key),
+        )
+        return result.steal_data()
+    except e:
+        var error_type = cpython.get_error_global("PyExc_Exception")
+        var msg = String(e)
+        cpython.PyErr_SetString(
+            error_type, msg.as_c_string_slice().unsafe_ptr()
+        )
+        return PyObjectPtr()
+
+
+def _mp_ass_subscript_wrapper[
+    self_type: ImplicitlyDestructible,
+    method: def(
+        UnsafePointer[self_type, MutAnyOrigin],
+        PythonObject,
+        Variant[PythonObject, Int],
+    ) thin raises -> None,
+](py_self: PyObjectPtr, key: PyObjectPtr, value: PyObjectPtr) abi("C") -> c_int:
+    """CPython `objobjargproc` adapter for the `mp_ass_subscript` slot.
+
+    When `value` is NULL the operation is a deletion (__delitem__); the `method`
+    receives `Variant[PythonObject, Int](Int(0))` as the third argument.
+    Otherwise the operation is an assignment (__setitem__) and `method` receives
+    `Variant[PythonObject, Int](value_object)`.
+
+    Parameters:
+        self_type: The Mojo struct type whose instances back the Python object.
+        method: User function with signature
+            `def(self, key, value: Variant[PythonObject, Int]) raises -> None`.
+
+    Returns:
+        0 on success, -1 with an exception set on error.
+    """
+    comptime PassedValue = Variant[PythonObject, Int]
+    ref cpython = Python().cpython()
+    try:
+        var passed_value = PassedValue(
+            PythonObject(from_borrowed=value)
+        ) if value else PassedValue(Int(0))
+        method(
+            _unwrap_self[self_type](py_self),
+            PythonObject(from_borrowed=key),
+            passed_value,
+        )
+        return c_int(0)
+    except e:
+        var error_type = cpython.get_error_global("PyExc_Exception")
+        var msg = String(e)
+        cpython.PyErr_SetString(
+            error_type, msg.as_c_string_slice().unsafe_ptr()
+        )
+        return c_int(-1)
+
+
+def _unaryfunc_wrapper[
+    self_type: ImplicitlyDestructible,
+    method: def(
+        UnsafePointer[self_type, MutAnyOrigin]
+    ) thin raises -> PythonObject,
+](py_self: PyObjectPtr) abi("C") -> PyObjectPtr:
+    """CPython `unaryfunc` adapter for unary nb_ slots (__neg__, __abs__, etc.).
+
+    Parameters:
+        self_type: The Mojo struct type whose instances back the Python object.
+        method: User function `def(self: UnsafePointer[self_type, MutAnyOrigin]) raises -> PythonObject`.
+
+    Returns:
+        New reference to the result, or null with an exception set on error.
+    """
+    ref cpython = Python().cpython()
+    try:
+        var result = method(_unwrap_self[self_type](py_self))
+        return result.steal_data()
+    except e:
+        var error_type = cpython.get_error_global("PyExc_Exception")
+        var msg = String(e)
+        cpython.PyErr_SetString(
+            error_type, msg.as_c_string_slice().unsafe_ptr()
+        )
+        return PyObjectPtr()
+
+
+def _binaryfunc_wrapper[
+    self_type: ImplicitlyDestructible,
+    method: def(
+        UnsafePointer[self_type, MutAnyOrigin], PythonObject
+    ) thin raises -> PythonObject,
+](lhs: PyObjectPtr, rhs: PyObjectPtr) abi("C") -> PyObjectPtr:
+    """CPython `binaryfunc` adapter for binary nb_ slots (__add__, __mul__, etc.).
+
+    If `method` raises `NotImplementedError` (by name), the wrapper returns
+    `Py_NotImplemented`, signalling Python to try the reflected operation.
+
+    Parameters:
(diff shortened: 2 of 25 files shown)


JoeLoser:
Hi @winding-lines - thanks for opening this PR! It's been in my queue as well as @ConnorGray's for a bit: I have started reviewing this and will leave a more detailed review this afternoon. Appreciate your patience - I'm excited by this work and what it can help unlock!

mojo/stdlib/std/python/adapters.mojo
```diff
@@ -0,0 +1,1325 @@
...
+    T: ImplicitlyDestructible
+](py_self: PyObjectPtr) -> UnsafePointer[T, MutAnyOrigin]:
+    """Downcast a raw PyObjectPtr to a typed Mojo pointer, aborting on failure.
+    """
+    try:
+        return PythonObject(from_borrowed=py_self).downcast_value_ptr[T]()
+    except e:
+        abort(
+            String("Python method receiver did not have the expected type: ", e)
+        )
```
JoeLoser: `_unwrap_self` aborts the process on a type mismatch. Every protocol slot funnels through here, so a wrong-type object reaching an unbound method kills the whole interpreter.

Pythonic behavior is to set `PyExc_TypeError` and return the slot's C error indicator (`NULL` for `*func`, `-1` for `lenfunc`/`inquiry`/`getbufferproc`/`objobjproc`, etc.). Either return a `Variant`/`Result`-style outcome to each wrapper, or have `_unwrap_self` itself raise and let the existing `try`/`except` in each wrapper handle it.

Side benefit: makes negative tests for the protocol surface actually testable — right now any test that misroutes a receiver type aborts the test binary.

mojo/stdlib/std/python/adapters.mojo
```diff
@@ -0,0 +1,1325 @@
...
+            PythonObject(from_borrowed=rhs),
+        )
+        return result.steal_data()
+    except e:
+        var msg = String(e)
+        if NotImplementedError.name == msg:
+            var not_implemented = cpython.lib.call[
+                "Py_GetConstantBorrowed", PyObjectPtr
+            ](4)
+            return cpython.Py_NewRef(not_implemented)
```
JoeLoser: I see two issues co-located on these lines.

**1. `Py_GetConstantBorrowed` is Python 3.13+ only.** `_cpython.mojo:1547-1569` already documents this and version-branches `Py_None`. The CI matrix runs 3.10, 3.11, 3.12, 3.13, 3.14 (`bazel/common.MODULE.bazel:117-123`). On 3.10–3.12 the first binary handler that raises `NotImplementedError` will fail with a missing-symbol error. Suggest routing `Py_NotImplemented` through `CPython.__init__` the same way `Py_None` is, so the version branch lives in one place; this call site then becomes `cpython.Py_NotImplemented` and the same fix covers lines 253 and 424.

**2. Dispatch by message string is fragile.** `if NotImplementedError.name == msg` matches any error whose message text equals the literal `"NotImplementedError"`. The sentinel also collides by name with Python's built-in `NotImplementedError`, which is semantically different (a real exception vs. a singleton fallback). Suggest renaming the sentinel (e.g. `PyNotImplementedSignal`) and dispatching via a side-channel — for example a thread-local flag set by the sentinel's constructor and cleared in the wrapper — rather than a string compare. Same comment applies at lines 251 and 421.

mojo/stdlib/std/python/adapters.mojo
```diff
@@ -0,0 +1,1325 @@
...
+            PythonObject(from_borrowed=mod),
+        )
+        return result.steal_data()
+    except e:
+        var msg = String(e)
+        if NotImplementedError.name == msg:
+            var not_implemented = cpython.lib.call[
+                "Py_GetConstantBorrowed", PyObjectPtr
+            ](4)
+            return cpython.Py_NewRef(not_implemented)
```
JoeLoser: Same 3.13+ `Py_GetConstantBorrowed` portability issue as line 208 (ternary path). Same string-match dispatch fragility. Please fix together — all three call sites (here, 208, 424) should consume one shared `cpython.Py_NotImplemented` accessor.

mojo/stdlib/std/python/adapters.mojo
```diff
@@ -0,0 +1,1325 @@
...
+        return cpython.PyBool_FromLong(c_long(Int(result)))
+    except e:
+        # Mojo lacks multiple except branches; dispatch on the error name.
+        var msg = String(e)
+        if NotImplementedError.name == msg:
+            # Py_CONSTANT_NOT_IMPLEMENTED = 4 (CPython 3.13+ stable ABI)
+            var not_implemented = cpython.lib.call[
+                "Py_GetConstantBorrowed", PyObjectPtr
+            ](4)
+            return cpython.Py_NewRef(not_implemented)
```
JoeLoser: Same 3.13+ `Py_GetConstantBorrowed` portability issue as lines 208/253 (richcompare path). The `# Py_CONSTANT_NOT_IMPLEMENTED = 4 (CPython 3.13+ stable ABI)` comment acknowledges this but the call site doesn't guard for it.

While fixing, please add a test that triggers this branch under Python 3.10 and 3.12 — today nothing in `test_type_protocol.py` or `test_number.py` exercises the `NotImplementedError → Py_NotImplemented` path on the pre-3.13 matrix entries, so the bug would land silently.

mojo/stdlib/std/python/adapters.mojo
```diff
@@ -0,0 +1,1325 @@
...
+    comptime tp_doc = Int32(56)
+    comptime tp_getattr = Int32(57)
+    comptime tp_getattro = Int32(58)
+    comptime tp_hash = Int32(59)
+    comptime tp_init = Int32(60)
+    comptime tp_is_gc = Int32(61)
+    comptime tp_iter = Int32(62)
+    comptime tp_iternext = Int32(63)
+    comptime tp_methods = Int32(64)
+    comptime tp_new = Int32(65)
```
JoeLoser: `_PySlotIndex` re-numbers slots that `_cpython.mojo:419-423` already defines as named constants (`Py_tp_dealloc = 52`, `Py_tp_init = 60`, `Py_tp_methods = 64`, `Py_tp_new = 65`, `Py_tp_repr = 66`) with typed `PyType_Slot.tp_X(func)` named constructors. Two parallel sources of truth for ABI numbers is a bug magnet.

Either extend the `_cpython.mojo` set with the additional `mp_*` / `nb_*` / `sq_*` / `bf_*` entries this PR needs, or — if the intent is to *replace* the `Py_tp_*` constants — do that conversion in this PR or call it out as an explicit follow-up. Today `buffer.mojo:34-35` forks a *third* copy of the same numbers, see comment there.

mojo/stdlib/std/python/buffer.mojo
```diff
@@ -0,0 +1,356 @@
...
+from std.python import Python, PythonObject
+from std.python._cpython import PyObjectPtr, Py_ssize_t, PyType_Slot
+from std.python.bindings import PythonTypeBuilder
+
+from .adapters import _unwrap_self
+
+
+# Slot indices for the buffer protocol (from CPython Include/typeslots.h).
+comptime _BF_GETBUFFER = Int32(1)
+comptime _BF_RELEASEBUFFER = Int32(2)
```
JoeLoser: `_BF_GETBUFFER`/`_BF_RELEASEBUFFER` are a third copy of ABI slot numbers that already exist as `_PySlotIndex.bf_getbuffer` / `bf_releasebuffer` in `adapters.mojo:448-449` (which themselves duplicate CPython's `typeslots.h`). This file already imports from `.adapters` — please drop these two `comptime` constants and use `_PySlotIndex.bf_getbuffer` directly at the `PyType_Slot(...)` call sites on lines 263 and 278.

mojo/stdlib/std/python/buffer.mojo
```diff
@@ -0,0 +1,356 @@
...
+
+# PyBUF_ flag constants (from CPython Include/cpython/object.h).
+comptime _PyBUF_WRITABLE = Int32(0x0001)
+comptime _PyBUF_FORMAT = Int32(0x0004)
+comptime _PyBUF_ND = Int32(0x0008)
+comptime _PyBUF_STRIDES = Int32(0x0018)  # 0x0010 | PyBUF_ND
```
JoeLoser: `_PyBUF_ND` and `_PyBUF_STRIDES` are defined but never read in this file. The strides path on line 209 unconditionally provides strides regardless of the flag (correct per the docstring on lines 206–209, since we always populate strides for a 1D buffer), so these constants are pure dead code. Delete them, or wire them up if a future commit intends to make strides request-gated.

mojo/stdlib/std/python/buffer.mojo
```diff
@@ -0,0 +1,356 @@
...
+        # + format string (fmt_len bytes) + null terminator (1 byte).
+        # On 64-bit platforms Int = 8 bytes, so the two Int fields occupy 16 bytes.
+        var fmt_bytes = info.format.as_bytes()
+        var fmt_len = len(fmt_bytes)
+        var alloc_size = 16 + fmt_len + 1  # 2 * sizeof(Int64) + format + NUL
+
+        # List.steal_data() gives us an owned UnsafePointer we can free later.
+        var store = List[UInt8](capacity=alloc_size)
+        store.resize(alloc_size, 0)
+        var alloc = store.steal_data()
```
JoeLoser: One `List[UInt8]` allocation + `resize(alloc_size, 0)` + `steal_data()` per `memoryview(obj)` / `numpy.frombuffer(obj)` call, just to hold one `shape[0]`, one `strides[0]`, and the format string. For 1D zero-copy hot paths (e.g. calling into numpy in a tight loop) this churns the allocator unnecessarily.

Two options worth considering:
1. Move shape/strides/format storage *into* the host Mojo    struct (or alongside it) so `_bf_getbuffer_wrapper`    just hands back pointers — no allocation per call.
2. At minimum, leave a `# TODO` here so future profiling    work has a breadcrumb.

Same applies to `_bf_releasebuffer_impl` on line 232 since it has to free the matching block.

mojo/stdlib/std/python/builders.mojo
```diff
@@ -0,0 +1,17 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+
+from .mapping import MappingProtocolBuilder
+from .number import NumberProtocolBuilder
+from .sequence import SequenceProtocolBuilder
+from .type_protocol import TypeProtocolBuilder
```
JoeLoser, line 24: Re-exports `MappingProtocolBuilder`, `NumberProtocolBuilder`, `SequenceProtocolBuilder`, `TypeProtocolBuilder` — but not `BufferProtocolBuilder`. `test/python/buffer_mojo_module.mojo` has to reach into `std.python.buffer` directly as a result. Either include all five or document why buffer is intentionally separate.

mojo/stdlib/std/python/utils.mojo
```diff
@@ -0,0 +1,66 @@
...
+            if op == RichCompareOps.Py_EQ:
+                ...
+            raise NotImplementedError()
+        ```
+    """
+
+    comptime name: String = "NotImplementedError"
+    """Well-known name used by `_richcompare_wrapper` for dispatch."""
+
+    def write_to(self, mut writer: Some[Writer]):
```
JoeLoser: Naming this sentinel `NotImplementedError` collides with Python's built-in `NotImplementedError` exception. The Mojo source `raise NotImplementedError()` and Python source `except NotImplementedError:` look identical at the syntactic level but refer to different concepts: this is a CPython protocol fallback *sentinel* (returns `Py_NotImplemented`), whereas Python's built-in is a genuine exception users may legitimately raise.

Suggest renaming to `PyNotImplementedSignal` (or just `NotImplemented`, which mirrors Python's singleton), and expanding the docstring to explicitly call out that this is the sentinel for `Py_NotImplemented` dispatch — not a user-facing exception type. Pair this with replacing the string-match dispatch in `adapters.mojo:206/251/421` (see comment there).

mojo/examples/python-interop/BUILD.bazel
```diff
@@ -46,6 +46,21 @@ modular_py_binary(
...
+modular_py_binary(
+    name = "columnar",
+    srcs = ["columnar.py"],
+    data = ["columnar_mojo.mojo"],
+    imports = ["."],
+    target_compatible_with = select({
+        "//:asan": ["@platforms//:incompatible"],
+        "//:ubsan": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    }),
```
JoeLoser, line 59: Inconsistent sanitizer exclusion. The matching `mojo_shared_library` and `modular_py_test` targets in `mojo/stdlib/test/python/BUILD.bazel:82-86, 101-106` exclude **asan, tsan, ubsan** for the same code path; this example excludes only **asan and ubsan**. tsan is missing here — likely an oversight. Please match the test-side config.

mojo/stdlib/test/python/BUILD.bazel
```diff
@@ -56,3 +64,46 @@ _NUMPY_TESTS = [
...
+        tags = ["no-pydeps"] if name == "buffer" else [],
+        target_compatible_with = select({
+            "//:asan": ["@platforms//:incompatible"],
+            "//:tsan": ["@platforms//:incompatible"],
+            "//:ubsan": ["@platforms//:incompatible"],
+            "//conditions:default": [],
+        }),
+    )
+    for name in _PROTOCOL_TESTS
+]
```
JoeLoser, line 109: Three test-coverage gaps worth closing as part of this PR rather than as follow-ups:

1. **Pre-3.13 `Py_NotImplemented` path.** Nothing in    `test_number.py` / `test_type_protocol.py` exercises a    raised-`NotImplementedError → Py_NotImplemented`    handoff on Python 3.10–3.12. That is exactly the    matrix where `Py_GetConstantBorrowed` will fail to    resolve (see `adapters.mojo:208`). Add at least one    such test per protocol that uses the sentinel.
2. **Buffer writable-rejection branch.**    `buffer.mojo:161-167` returns -1 with `PyExc_BufferError`    when `PyBUF_WRITABLE` is requested on a read-only    buffer. No test exercises this path — it would be    straightforward to add via `memoryview(obj, writable=True)` (or the lower-level `__buffer__` protocol).
3. **Type-mismatch handling.** No negative tests around    `_unwrap_self`'s wrong-type path because the current    implementation aborts; replacing the abort with a    `TypeError` (see comment on `adapters.mojo:36`) makes    this testable.

JoeLoser (commented):
@winding-lines thanks for your hard work on extending these capabilities and building these out! I think they will be super useful.

Once we align on the feedback at large, I'd suggest we split this up into 5 separate PRs - one for each builder essentially. @ConnorGray mentioned you and him had already chatted about this. This will make each PR much more bite-sized and easier to review, especially for ABI and other issues.

I would also recommend having a short design doc for each of the builders to help give context and understand the proposed changes for people that may not be as close to the Python-Mojo interop so far (such as @NathanSWard). It would be good to help understand the use cases each one unlocks and how it fits into the broader Python-Mojo interop story for the stdlib.

mojo/stdlib/std/python/adapters.mojo
winding-lines: Great feedback, removed the try/except and see how to handle the error in the caller.

mojo/stdlib/std/python/adapters.mojo
winding-lines: Agree, I enhanced the error integration. In some cases specific exceptions are expected, for example KeyError.

mojo/stdlib/std/python/adapters.mojo
winding-lines: Changed all of them, thanks for the feedback.

mojo/examples/python-interop/BUILD.bazel
winding-lines: Done, thanks!

mojo/stdlib/test/python/BUILD.bazel
winding-lines: Add tests for the above, the `_unwrap_self` exception path is dictated by the required behavior on the `NotImplemented` path.

winding-lines:
> @winding-lines thanks for your hard work on extending these capabilities and building these out! I think they will be super useful.
> 
> Once we align on the feedback at large, I'd suggest we split this up into 5 separate PRs - one for each builder essentially. @ConnorGray mentioned you and him had already chatted about this. This will make each PR much more bite-sized and easier to review, especially for ABI and other issues.
> 
> I would also recommend having a short design doc for each of the builders to help give context and understand the proposed changes for people that may not be as close to the Python-Mojo interop so far (such as @NathanSWard). It would be good to help understand the use cases each one unlocks and how it fits into the broader Python-Mojo interop story for the stdlib.

Added a design doc and I've addressed the feedback. I am ready to split this into smaller PRs, let me know if you want to verify that all the feedback is properly addressed before I do that.

mojo/stdlib/std/python/utils.mojo
winding-lines: There was a design gap here. The Type Slot functions need to be able to control their exceptions. Also the exception story has change in Mojo from end of last year when I first looked at this. I redid all of this code, let me know what you think 🙇

mojo/stdlib/std/python/builders.mojo
winding-lines: This is a miss, thanks for the review.

winding-lines:
I actually need to address some more feedback, so this is not ready for re-review.

mojo/stdlib/std/python/adapters.mojo
winding-lines: I've replaced the free constants in _cpython.mojo with this struct since the struct is more general.

mojo/stdlib/std/python/buffer.mojo
winding-lines: Done, sorry about this.

mojo/stdlib/std/python/buffer.mojo
winding-lines: Good eye! Deleted for now.

mojo/stdlib/std/python/buffer.mojo
winding-lines: One way or another we need to return pointers to some memory that outlives the current call frame.  I changed the code to use data inside the BufferInfo, but now BufferInfo must be in the heap. The python API requires us to return pointers to dimensions, shape, format. 

We can push this responsibility on the user and they can likely optimize further. But then the API becomes more difficult. It could well be that very high performance libraries would want that.

Let me know what you think....

winding-lines:
@JoeLoser this is ready for another high level review before being split in smaller parts of reviewed individually.

Many thanks for your guidance!

PR #6455, opened 2026-04-24
[stdlib] Refactor `Codepoint` to use UTF-8 internally

martinvuyk:
# Refactor `Codepoint` to use UTF-8 internally

Implements #6200

One caveat of going in this direction is that this now makes this pattern slower:
```mojo
for cp in str.codepoints():
  some_utf_32_func(cp.to_u32())
  ..
```
The replacement for the same performance as previously would be
```mojo
var str = StringSlice(og_str)
while str.byte_length() > 0:
  var cp, byte_length = Codepoint.unsafe_decode_utf8_codepoint(str)
  str = StringSlice(ptr=str.unsafe_ptr() + byte_length, length=str.byte_length() - byte_length)
  some_utf_32_func(cp)
  ...
```
Unless we create a `UTF32Iterator` that returns the raw `UInt32` that wraps this functionality.

IMO this is a tradeoff that is worth it and we should provide another iterator that explicitly corresponds to UTF-32 codepoints.

This pattern of iterating over codepoints and then building a `UInt32` is only used in the stdlib for looking up unicode tables in the upper and lower casing code. The code itself will be simple to fix in the future, and I also expect #6293 to be merged at some point and replace these functions anyway.

We could also go in the direction of making this `Codepoint` be `UTF8Codepoint` and renaming the current one `UTF32Codepoint`, or parametrizing `Codepoint` based on the encoding and using the same raw storage.

## Benchmark results

CPU: Intel® Core™ i7-7700HQ

### Interpretation

- `str.codepoints()`: For Iteration in English text the improvements are from 5-10 %, and Chinese ~45% on average. This is a net gain
- `ord(str) -> Int`: The implementation is essentially the same, but due to some adjustments there is a noticeable improvement in English of around 20%
- `chr(Int) -> String`: English text has an improvement of around 20%, and Chinese around 50% (2x faster)

### English

| name | old (ms) | new (ms) | markdown perc. |
-- | -- | -- | --
bench_string_codepoints[10] | 4.06 | 4.28 | -0.05
bench_string_codepoints[30] | 4.96 | 4.64 | 0.06
bench_string_codepoints[50] | 5.11 | 4.76 | 0.07
bench_string_codepoints[100] | 5.55 | 4.86 | 0.12
bench_string_codepoints[1000] | 5.45 | 4.98 | 0.08
bench_string_codepoints[10000] | 5.48 | 4.88 | 0.11
bench_string_codepoints[100000] | 5.59 | 4.90 | 0.12
bench_string_codepoints[1000000] | 5.43 | 4.81 | 0.11
  |   |   |  
bench_string_ord[10] | 1.27 | 0.88 | 0.31
bench_string_ord[30] | 1.08 | 0.77 | 0.28
bench_string_ord[50] | 1.06 | 0.85 | 0.20
bench_string_ord[100] | 1.06 | 0.86 | 0.19
bench_string_ord[1000] | 1.08 | 0.86 | 0.21
bench_string_ord[10000] | 1.19 | 0.92 | 0.23
bench_string_ord[100000] | 1.31 | 1.12 | 0.15
bench_string_ord[1000000] | 4.26 | 3.85 | 0.10
  |   |   |  
bench_string_chr[10] | 14.34 | 11.77 | 0.18
bench_string_chr[30] | 15.48 | 12.66 | 0.18
bench_string_chr[50] | 15.65 | 12.73 | 0.19
bench_string_chr[100] | 15.74 | 12.97 | 0.18
bench_string_chr[1000] | 16.42 | 13.07 | 0.20
bench_string_chr[10000] | 17.55 | 12.95 | 0.26
bench_string_chr[100000] | 17.10 | 13.11 | 0.23
bench_string_chr[1000000] | 16.75 | 13.44 | 0.20

### Chinese

| name | old (ms) | new (ms) | markdown perc. |
-- | -- | -- | --
bench_string_codepoints[10] | 3.22 | 2.33 | 0.28
bench_string_codepoints[30] | 3.15 | 2.26 | 0.28
bench_string_codepoints[50] | 3.28 | 2.36 | 0.28
bench_string_codepoints[100] | 3.36 | 2.36 | 0.30
bench_string_codepoints[1000] | 3.33 | 2.73 | 0.18
bench_string_codepoints[10000] | 3.46 | 2.42 | 0.30
bench_string_codepoints[100000] | 3.33 | 2.48 | 0.26
bench_string_codepoints[1000000] | 3.44 | 2.61 | 0.24
  |   |   |  
bench_string_ord[10] | 3.69 | 3.53 | 0.04
bench_string_ord[30] | 3.63 | 3.63 | 0.00
bench_string_ord[50] | 4.02 | 3.48 | 0.13
bench_string_ord[100] | 4.03 | 3.66 | 0.09
bench_string_ord[1000] | 4.22 | 3.64 | 0.14
bench_string_ord[10000] | 4.23 | 3.95 | 0.07
bench_string_ord[100000] | 4.56 | 4.05 | 0.11
bench_string_ord[1000000] | 7.27 | 6.89 | 0.05
  |   |   |  
bench_string_chr[10] | 20.17 | 10.44 | 0.48
bench_string_chr[30] | 20.93 | 10.68 | 0.49
bench_string_chr[50] | 21.10 | 10.89 | 0.48
bench_string_chr[100] | 20.61 | 11.11 | 0.46
bench_string_chr[1000] | 22.75 | 10.84 | 0.52
bench_string_chr[10000] | 21.28 | 10.99 | 0.48
bench_string_chr[100000] | 23.03 | 11.20 | 0.51
bench_string_chr[1000000] | 22.43 | 10.74 | 0.52

### DIFF
--- a/mojo/stdlib/benchmarks/collections/bench_string.mojo
+++ b/mojo/stdlib/benchmarks/collections/bench_string.mojo
@@ -25,9 +25,7 @@ from std.benchmark import Bench, BenchConfig, Bencher, BenchId, black_box, keep
 # ===-----------------------------------------------------------------------===#
 # Benchmark Data
 # ===-----------------------------------------------------------------------===#
-def make_string[
-    length: Int = 0
-](filename: String = "UN_charter_EN.txt") -> String:
+def make_string[length: Int](filename: String = "UN_charter_EN.txt") -> String:
     """Make a `String` made of items in the `./data` directory.
 
     Parameters:
@@ -48,9 +46,9 @@ def make_string[
             while length > len(items):
                 items.append(items[i])
                 i = i + 1 if i < len(items) - 1 else 0
-            return String(unsafe_from_utf8=items)
+            return String(from_utf8_lossy=items)
         else:
-            return String(unsafe_from_utf8=f.read_bytes())
+            return String(from_utf8_lossy=f.read_bytes())
     except e:
         print(e, file=stderr)
     abort(String())
@@ -75,7 +73,7 @@ def bench_string_init(mut b: Bencher) raises:
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_count[
-    length: Int = 0,
+    length: Int,
     filename: StaticString = "UN_charter_EN",
     sequence: StaticString = "a",
 ](mut b: Bencher) raises:
@@ -94,7 +92,7 @@ def bench_string_count[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_split[
-    length: Int = 0,
+    length: Int,
     filename: StaticString = "UN_charter_EN",
     sequence: Optional[StaticString] = None,
 ](mut b: Bencher) raises:
@@ -149,7 +147,7 @@ def bench_string_join[short: Bool](mut b: Bencher) raises:
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_splitlines[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = StringSlice(make_string[length](filename + ".txt"))
 
@@ -167,7 +165,7 @@ def bench_string_splitlines[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_lower[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = make_string[length](filename + ".txt")
 
@@ -184,7 +182,7 @@ def bench_string_lower[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_upper[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = make_string[length](filename + ".txt")
 
@@ -201,7 +199,7 @@ def bench_string_upper[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_replace[
-    length: Int = 0,
+    length: Int,
     filename: StaticString = "UN_charter_EN",
     old: StaticString = "a",
     new: StaticString = "A",
@@ -221,7 +219,7 @@ def bench_string_replace[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_count_codepoints[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = make_string[length](filename + ".txt")
 
@@ -238,7 +236,7 @@ def bench_string_count_codepoints[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_find_single[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = make_string[length](filename + ".txt")
 
@@ -259,7 +257,7 @@ def bench_string_find_single[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_find_multiple[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = make_string[length](filename + ".txt")
     var sequence = "ZZZZ"  # something that probably won't be there
@@ -279,7 +277,7 @@ def bench_string_find_multiple[
 # ===-----------------------------------------------------------------------===#
 @parameter
 def bench_string_is_valid_utf8[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = make_string[length](filename + ".html")
 
@@ -292,29 +290,27 @@ def bench_string_is_valid_utf8[
 
 
 # ===-----------------------------------------------------------------------===#
-# Benchmark write_utf8
+# Benchmark codepoint_write_to
 # ===-----------------------------------------------------------------------===#
 @parameter
-def bench_write_utf8[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+def bench_codepoint_write_to[
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher) raises:
     var items = make_string[length](filename + ".txt")
     var codepoints_iter = items.codepoints()
     # appending to a list to avoid paying the overhead of codepoint parsing
-    var codepoints = List[Codepoint](capacity=len(codepoints_iter))
+    var codepoints = List[UInt32](capacity=len(codepoints_iter))
     for c in codepoints_iter:
-        codepoints.append(c)
+        codepoints.append(c.to_u32())
 
     @always_inline
     def call_fn() unified {read}:
-        var data = InlineArray[Byte, 4](uninitialized=True)
         # this is to help with instability when measuring small strings
         for _ in range(10**6 // length):
             for i in range(len(codepoints)):
-                var res = black_box(codepoints.unsafe_get(i)).unsafe_write_utf8(
-                    black_box(data).unsafe_ptr()
-                )
-                keep(res)
+                var writer = NullWriter()
+                black_box(codepoints.unsafe_get(i)).write_to(writer)
+                keep(writer)
 
     b.iter(call_fn)
 
@@ -371,7 +367,7 @@ struct NullWriter(ImplicitlyCopyable, Writer):
 
 @parameter
 def bench_string_repr[
-    length: Int = 0, filename: StaticString = "UN_charter_EN"
+    length: Int, filename: StaticString = "UN_charter_EN"
 ](mut b: Bencher):
     var items = make_string[length](filename + ".txt")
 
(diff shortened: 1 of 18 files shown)


mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -158,124 +286,25 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
...
-            True if this character's value is less than the other codepoint value;
-            False otherwise.
+            True if this character's value is less than the other codepoint
+            value; False otherwise.
         """
-        return self.to_u32() < other.to_u32()
+        return (
+            UnsafePointer(to=self._values).bitcast[UInt32]()[]
+            < UnsafePointer(to=other._values).bitcast[UInt32]()[]
+        )
```
JoeLoser: Is this correct in all cases? It looks a little fishy, especially on little-endian platforms.

`bitcast[UInt32]` makes `_values[0]` the *least* significant byte. UTF-8 byte-wise lexicographic order requires the *first* byte to be the *most* significant. So this comparison disagrees with codepoint order whenever two codepoints differ outside the first byte.

Concrete counter-example:
- `"Я"` U+042F → UTF-8 `[0xD0, 0xAF, 0, 0]` → LE-u32 = `0x0000AFD0` (44944)
- `"ا"` U+0627 → UTF-8 `[0xD8, 0xA7, 0, 0]` → LE-u32 = `0x0000A7D8` (42968)

Codepoint order says `Я (1071) < ا (1575)`. This `__lt__` returns `Я > ا`. That also breaks `Comparable`'s antisymmetry vs. `__eq__`.

Cleanest fix is `return self.to_u32() < other.to_u32()`. If you want to keep something fast, byte-swap before compare (or do the lex compare across the SIMD vector).

No test caught this because there's no test for `__lt__` across multibyte codepoints — please add one (Latin/Cyrillic/Arabic/CJK/emoji).

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -97,17 +99,64 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
     scalar values.
     """
 
-    var _scalar_value: UInt32
-    """The Unicode scalar value represented by this type."""
+    var _values: SIMD[DType.uint8, 4]
```
JoeLoser, line 103: Please document the load-bearing invariant on this field directly: **bytes at indices `>= byte_length()` must be zero**. `to_u32`, `__lt__`, and the `StringSlice(Codepoint)` bitcast all rely on it, but it's only called out in a comment inside `to_u32`. Any future constructor / mutator on `Codepoint` needs to honor it.

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -97,17 +99,64 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
...
 
     # ===-------------------------------------------------------------------===#
     # Life cycle methods
     # ===-------------------------------------------------------------------===#
 
     @always_inline
-    def __init__(out self, *, unsafe_unchecked_codepoint: UInt32):
-        """Construct a `Codepoint` from a code point value without checking that it
-        falls in the valid range.
+    def __init__(out self, single_char: StringLiteral):
```
JoeLoser, line 114: Should this be `@implicit`? If it were, all the test churn from `Codepoint.ord("x")` → `Codepoint("x")` could just disappear at the call sites that already pass a literal into a `Codepoint`-typed parameter. If you intentionally avoided `@implicit` (e.g., to dodge surprising overload selection on `String + StringLiteral`), please add a comment saying so.

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -97,17 +99,64 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
...
+        if Self.is_ascii(b0):
+            return
+
+        var num_bytes = Self.byte_length(first_utf8_byte=b0)
+        if unsafe_parse_first.byte_length() >= 4:
+            var bool_vec = SIMD[DType.uint8, 4](1, 2, 3, 4).le(UInt8(num_bytes))
+            self._values = ptr.load[4]() & splat[DType.uint8](bool_vec)
+        else:
+            for i in range(1, num_bytes):
+                self._values[i] = ptr[i]
```
JoeLoser, line 153: This `else` branch trusts `num_bytes` derived from `b0` without bounding it by `unsafe_parse_first.byte_length()`. If a caller passes `"\xC3"` (length 1, but `b0` claims a 2-byte sequence), this reads OOB.

The function name has `unsafe_` and the doc says "non-empty", but "non-empty" is weaker than the precondition the body actually requires. Please tighten the docstring's safety section to require the slice to begin with a complete, valid UTF-8 sequence. Or, if this should be safe-against-truncation, clamp the loop with `min(num_bytes, byte_length)`.

mojo/stdlib/std/collections/string/codepoint.mojo
JoeLoser: (Already raised the LE bug above — separately:) please add a short Safety/Notes block to the docstring documenting that this is a UTF-8 byte-order comparison, not a UInt32 comparison, once the bug is fixed.

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -312,12 +338,25 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
         Args:
             writer: The object to write to.
         """
-        fmt.FormatStruct(writer, "Codepoint").fields(self._scalar_value)
+        fmt.FormatStruct(writer, "Codepoint").fields(self._values)
```
JoeLoser: The new repr leaks the internal storage representation:

```
Codepoint(97)            →  Codepoint([97, 0, 0, 0])
Codepoint(0)             →  Codepoint([0, 0, 0, 0])
Codepoint(0x1F642)       →  Codepoint([240, 159, 153, 130])
```

The old repr (numeric scalar value) was much friendlier and matches what `__int__` returns. Suggest: `fmt.FormatStruct(writer, "Codepoint").fields(self.to_u32())`.

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -478,117 +514,164 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
...
+                mask >> (num_bytes - is_ascii), mask, mask, mask
+            )
+            return (
+                (self._values & b_type_mask).cast[DType.uint32]() << shift_mask
+            ).reduce_or()
+
+    @staticmethod
+    def unsafe_decode_utf8_codepoint[
+        optimize_ascii: Bool = True
+    ](string: StringSlice[mut=False, _]) -> Tuple[UInt32, Int]:
```
JoeLoser: Return type changed from `(Codepoint, Int)` to `(UInt32, Int)`. This walks back the type-safety the previous API provided, and forces every existing caller that wanted a `Codepoint` to wrap manually.

Given we already have all 4 bytes in registers when constructing the result, returning `(Codepoint, Int)` should be ~free. Suggest keeping the type-safe return; if you need a UInt32-returning fast path, expose it as a separate function (e.g. `unsafe_decode_utf8_to_u32`).

Also the docstring Notes line says `Codepoint(str).to_u32()` — there's no `Codepoint(StringSlice)` ctor; presumably this means `Codepoint(unsafe_parse_first=str).to_u32()`.

mojo/stdlib/std/collections/string/codepoint.mojo
```diff
@@ -478,117 +514,164 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
...
+        return result, Int(num_bytes)
+
+    # ===-------------------------------------------------------------------===#
+    # other methods
+    # ===-------------------------------------------------------------------===#
+
+    @staticmethod
     @always_inline
-    def utf8_byte_length(self) -> Int:
+    def byte_length(*, first_utf8_byte: UInt8) -> Int:
```
JoeLoser: There are now three `byte_length` overloads on `Codepoint` (instance, `first_utf8_byte=`, `codepoint:`) and the static one is a one-line wrapper around `_utf8_first_byte_sequence_length`.

Also: the rename from `utf8_byte_length()` → `byte_length()` makes the meaning more ambiguous on a Codepoint ("size of the in-memory representation?" vs. "size of the UTF-8 encoding?"). They happen to coincide today but the old name was clearer IMO.

mojo/stdlib/benchmarks/collections/bench_string.mojo
```diff
@@ -386,6 +382,75 @@ def bench_string_repr[
...
+
+    var codepoints_iter = items.codepoints()
+    # appending to a list to avoid paying the overhead of codepoint parsing
+    var codepoints = List[UInt32](capacity=len(codepoints_iter))
+    for c in codepoints_iter:
+        codepoints.append(c.to_u32())
+
+    @always_inline
+    def call_fn() unified {read}:
+        for codepoint in codepoints:
```
JoeLoser, line 447: Same `codepoint` value is reused across the inner loop, allowing CSE/hoisting to no-op the body. Wrap the value with `black_box` (similar to `bench_string_codepoints`) to keep the optimizer honest.

mojo/stdlib/std/bit/mask.mojo
```diff
@@ -58,14 +58,14 @@ def is_negative[dtype: DType, //](value: SIMD[dtype, _]) -> type_of(value):
 
 @always_inline
 def splat[
-    size: SIMDSize, //, dtype: DType
+    dtype: DType, size: SIMDSize = 1
 ](value: SIMD[DType.bool, size]) -> SIMD[dtype, size]:
```
JoeLoser, line 62: Unrelated breaking signature change buried in a Codepoint refactor. The PR only updates two call sites (here and `codepoint.mojo`), but `splat` is in `std.bit.mask` so anything outside this file that calls it as `splat[size, dtype]` will silently break (parameter order swap + new default). Please split this into a standalone NFC PR with a sweep of all call sites.

mojo/stdlib/test/collections/string/test_codepoint.mojo
```diff
@@ -24,7 +24,7 @@ from std.testing import TestSuite
 def test_char_validity() raises:
```
JoeLoser, line 26: Test coverage gaps to address before merge:

- No test exercises `Codepoint.__lt__` / `__eq__` across multibyte codepoints — would have caught the LE bitcast bug.
- No test for `unsafe_parse_first` on truncated/invalid UTF-8 (intentionally unsafe, but at least one debug-assert negative test would document the contract).
- No test for the `optimize_ascii=False` parameter on `Codepoint(unsafe_unchecked_codepoint=…)` or `to_u32`.

mojo/stdlib/std/collections/string/codepoint.mojo
martinvuyk: ooh yeah thanks for catching this, I was just thinking in big-endian terms. I'll add some more tests

mojo/stdlib/std/collections/string/codepoint.mojo
martinvuyk: I can make it implicit, I went for the conservative choice since we've been avoiding implicit things in many places

mojo/stdlib/std/collections/string/codepoint.mojo
martinvuyk: > If a caller passes "\xC3" (length 1, but b0 claims a 2-byte sequence), this reads OOB.
>  
> The function name has unsafe_ and the doc says "non-empty", but "non-empty" is weaker than the precondition the body actually requires. Please tighten the docstring's safety section to require the slice to begin with a complete, valid UTF-8 sequence. Or, if this should be safe-against-truncation, clamp the loop with min(num_bytes, byte_length).

Should we expect a `StringSlice` to have invalid UTF-8? My understanding is that that is a safe assumption in the whole stdlib. I did change the previous APIs from using raw bytes to using `StringSlice` exactly for that reason

mojo/stdlib/std/collections/string/codepoint.mojo
martinvuyk: > if you need a UInt32-returning fast path, expose it as a separate function (e.g. `unsafe_decode_utf8_to_u32`).

Yeah I'm going to rename this `unsafe_decode_utf8_to_utf32` because the whole purpose is to bypass constructing a `Codepoint` and then doing `.to_u32()`. This is the cheap fast path for things like `ord(StringSlice) -> Int` to remain as fast as they were (I verified this with benchmarks)

> presumably this means Codepoint(unsafe_parse_first=str).to_u32().

yep, I'll fix it

mojo/stdlib/std/collections/string/codepoint.mojo
martinvuyk: I can go back without an issue. I was thinking more along the lines of that maybe in the future we will want to parametrize `Codepoint` on the encoding that it has internally, and also in being more aligned with `String` et al. Yeah I think I'll go back to the old name to be clearer

mojo/stdlib/std/bit/mask.mojo
martinvuyk: Ok I'll remove this change. This was recently changed from how I previously left it which is exactly like this. AFAIK there are still no other uses of scalar simd vectors with this function in the stdlib and is to just avoid having to do `splat(Scalar[DType.bool](a < b))` and just do `splat(a < b)`. The size parameter is inferred anyway

mojo/stdlib/test/collections/string/test_codepoint.mojo
martinvuyk: Got it, will add them

> No test for unsafe_parse_first on truncated/invalid UTF-8 (intentionally unsafe, but at least one debug-assert negative test would document the contract).

I'm going to skip this one for now until you can read this because AFAIK we should expect `StringSlice` to have valid UTF-8

mojo/stdlib/benchmarks/collections/bench_string.mojo
martinvuyk: I actually didn't understand (or used) `black_box` until you put it that way, I updated the benchmarks and the benchmark results above. The chr improvements in english were more modest than previously but the chinese ones jumped from 20 to 40 %. It also seems that the new ord is faster than I thought for some reason

barcharcraz:
The benchmark results are weird, it looks like we're including "something else" in the benchmark that should not be there

mojo/stdlib/std/collections/string/codepoint.mojo
barcharcraz: we should be able to assume valid UTF-8 as a precondition. We should still defend against invalid data if it's not too much trouble, this extra length check could actually have significant cost though.

Do we do the spiritual equivalent of this check in the old UTF-32 version?

martinvuyk:
> The benchmark results are weird, it looks like we're including "something else" in the benchmark that should not be there

@barcharcraz what is that "something else"? This PR affects all three aspects: codepoint iteration, chr, and ord.

If you mean that in terms of the results having a long execution time, they are running `10**6 // length` times each operation, which is to help with the stability of results for small strings

PR #6476, opened 2026-04-30
[Kernels][Benchmarking] bump vendor libraries versions for benchmark

larme:
## Summary

Upgrade the following vendor libraries versions:

- FlashInfer 0.5.2 -> 0.6.7.post3
- FlashAttention 2.8.3 -> 4.0.0b4

I'm not bumping FlashAttention to the absolute latest (4.0.0b11) because later version requires cutedsl 4.4.2, which will require a huge changes for our lockfile.

## Checklist

      sequence of smaller PRs

Assisted-by: Claude

### DIFF
--- a/bazel/common.MODULE.bazel
+++ b/bazel/common.MODULE.bazel
@@ -455,20 +455,6 @@ http_file(
     urls = ["https://modular-bazel-artifacts-public.s3.amazonaws.com/artifacts/deep_gemm/sm100/91a898b8704b91726646e98403f6028b1e601603b2d34f8e4530370072e474cf/deep_gemm-2.2.0_38f8ef7-cp312-cp312-linux_x86_64.whl"],
 )
 
-http_file(
-    name = "flashinfer_sm100_wheel",
-    downloaded_file_path = "flashinfer_python-0.6.7.post3-py3-none-any.whl",
-    sha256 = "9d3f1aa0313cf9e5cf99f7560b8e003c57876088c8abfe7a3d330cccd4873052",
-    urls = ["https://files.pythonhosted.org/packages/01/6b/4117cd7cbeff07818ae7c6b8bf5a6d1ee3eed29356672b731b55af3d4453/flashinfer_python-0.6.7.post3-py3-none-any.whl"],
-)
-
-http_file(
-    name = "flash_attn_sm100_wheel",
-    downloaded_file_path = "flash_attn_4-4.0.0b4-py3-none-any.whl",
-    sha256 = "daaec7e76e9aed0bbc717228b4b8461387a99fa71629fb34a8dd00fc56e291e4",
-    urls = ["https://files.pythonhosted.org/packages/73/8e/b27eed4f2dcb9e5a89719e2233be36e86faa64fd1bead802265c7a7b046c/flash_attn_4-4.0.0b4-py3-none-any.whl"],
-)
-
 # https://github.com/modularml/breathe
 http_archive(
     name = "breathe",
--- a/bazel/common.MODULE.bazel
+++ b/bazel/common.MODULE.bazel
@@ -457,16 +457,16 @@ http_file(
 
 http_file(
     name = "flashinfer_sm100_wheel",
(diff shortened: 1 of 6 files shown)


chriselrod (approved):

bazel/common.MODULE.bazel
```diff
@@ -457,16 +457,16 @@ http_file(
 
 http_file(
     name = "flashinfer_sm100_wheel",
-    downloaded_file_path = "flashinfer_python-0.5.2-py3-none-any.whl",
-    sha256 = "0afdb8f536a3a101ac7f653d66a5f4f63991f820352daf64950dab89f9806c1c",
-    urls = ["https://modular-bazel-artifacts-public.s3.amazonaws.com/artifacts/flashinfer/sm100/0afdb8f536a3a101ac7f653d66a5f4f63991f820352daf64950dab89f9806c1c/flashinfer_python-0.5.2-py3-none-any.whl"],
+    downloaded_file_path = "flashinfer_python-0.6.7.post3-py3-none-any.whl",
+    sha256 = "9d3f1aa0313cf9e5cf99f7560b8e003c57876088c8abfe7a3d330cccd4873052",
+    urls = ["https://files.pythonhosted.org/packages/01/6b/4117cd7cbeff07818ae7c6b8bf5a6d1ee3eed29356672b731b55af3d4453/flashinfer_python-0.6.7.post3-py3-none-any.whl"],
```
Ahajha: If this means we don't have to prebuild these ourselves, probably better to modify the version in `bazel/pip/requirements/pyproject.toml` and then run `./bazelw run //bazel/pip/requirements:update_requirements`. Would probably be able to clean up these manual `http_file`s and the manual `pycross_lock_file`s.

But this is a reasonable incremental improvement.

larme: Hi @Ahajha , I have pushed changes according to the review. However this change seems to make flashinfer do some compiling works (calling cicc) at the start of bazel benchmarks, so I need to add a --timeout-secs 0 to wait for the compiling finished. So a benchmark may looks like:

```
./bazelw run //max/kernels/benchmarks/autotune:kbench -- \
  max/kernels/benchmarks/misc/comparison/bench_mla_decode_sm100_deepseek_bf16.yaml \
  --target-accelerator nvidia:sm_100 \
  --timeout-secs 0 \
  --output-dir benchmark_results/decode_bf16_full \
  -o decode_bf16_full.csv
```

Do you think we should reverse the newest change?

larme: I think the reason is now we are using `requirement("flashinfer-python")`, which introducing the dependency of `cuda-tile`. The old way of manual wheel pinning avoid this problem

Ahajha: So sorry, I never got notified of the reply. Looking now.

Ahajha: Okay you have some interesting points. Few points:
1. Calling out to `cicc` sounds... dangerous in a Bazel env. I assume it would get that from one of the wheels? If so maybe fine.
2. How long does this compilation take? If it's ~seconds maybe okay, too long and maybe we should go back to the precompiled wheel.
3. Doing it this way (via the pyproject.toml) _is_ nicer in terms of deps management, but not if it comes with runtime drawbacks.

If you could answer for me 1 and 2 we can make a decision. And I'll keep the tab open this time so I see your reply :sweat_smile:

larme: @Ahajha 

1. if we do `requirement("flashinfer-python")`, we will involve `cicc`. I think the reason is when installing `flashinfer-python` on a B200 system, some additional compiling is involved. Let me check this part in details and report back later.
2. It will takes quite some minutes to finish compiling.

Ahajha: Okay, ~minutes is not great, I think maybe we should go back to the prebuilt wheel then. We could also still use the pyproject file by specifying `[tool.uv.sources]` for flashinfer-python and pointing to the prebuilt one. I _think_ that would get us the best of both worlds, wdyt?

PR #6487, opened 2026-05-03
[stdlib] Add iterator for homogeneous tuples

martinvuyk:
Add iterator for homogeneous tuples

### DIFF
--- a/mojo/stdlib/std/builtin/tuple.mojo
+++ b/mojo/stdlib/std/builtin/tuple.mojo
@@ -32,15 +32,87 @@ from std.reflection.traits import (
     AllRegisterPassable,
     AllWritable,
 )
-from std.sys.intrinsics import _type_is_eq
+from std.sys.intrinsics import _type_is_eq, _type_is_eq_parse_time
 
 from std.reflection.type_info import _unqualified_type_name
-
 from std.utils._visualizers import lldb_formatter_wrapping_type
 
 # ===-----------------------------------------------------------------------===#
 # Tuple
 # ===-----------------------------------------------------------------------===#
+
+comptime _all_same_type_condition[
+    T0: Movable, T: Movable
+] = _type_is_eq_parse_time[T, T0]()
+
+comptime _all_same_type[*Ts: Movable]: Bool = Ts.all_satisfies[
+    _all_same_type_condition[Ts[0], _]
+]()
+
+
+struct _TupleIterOwned[*element_types: Movable](
+    IterableOwned, Iterator, Movable
+):
+    """An owning iterator for a homogeneous `Tuple`.
+
+    Parameters:
+        element_types: The elements types.
+    """
+
+    comptime Element = Self.element_types[0]
+    comptime IteratorOwnedType = Self
+
+    var _tuple: Tuple[*Self.element_types]
+    var _index: Int
+
+    def __init__(
+        var tup: Tuple, out self: _TupleIterOwned[*tup.element_types]
+    ) where _all_same_type[*tup.element_types]:
+        """Consume an tuple and create an iterator over its elements.
+
+        Args:
+            tup: The tuple to consume.
+        """
+        self._tuple = tup^
+        self._index = 0
+
+    @always_inline
+    def __del__(deinit self):
+        # Move fields out of self so we can manage their lifetimes.
+        var idx = self._index
+        var tup = self._tuple^
+
+        _constrained_conforms_to[
+            conforms_to(Self.Element, ImplicitlyDestructible),
+            Parent=Self,
+            Element=Self.Element,
+            ParentConformsTo="ImplicitlyDestructible",
+        ]()
+        # Destroy the remaining elements that have not yet been
+        # iterated over.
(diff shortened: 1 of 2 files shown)


martinvuyk:
It seems like I found a compiler bug?

jorel666:
Hi @modular! Hey @modular, your AI/ML Developer work caught my eye.

ARKHAN is a fractal AI system with 59 specialized chips: autonomous scraping, crypto monitoring, content generation, income automation. API access $29/mo.

Store: https://fractalaixipetotec.online
Card payment: https://xipetotec.lemonsqueezy.com/checkout/buy/d5943e3e-b661-4011-a995-826847b87156

Reply STOP to opt out.

Pago directo USDT TRC20: TWJv7sD1nTWeHFsaMDBuJEEAc62VoxPmvm
Ref: pay_4542b0903a59

*Automated outreach — reply NO to stop.*

barcharcraz:
I don't think we want this. It's sorta a new subtype of tuple, and a homogeneous tuple is essentially just an InlineArray. Right now it's somewhat hard to do this conversion but we should fix that instead of adding this iterator. If there's useful code that uses this then maybe we could reconsider.

Also, we don't want to guarantee that tuples are laid out sequentially in memory like this, so using UnsafePointer in this way is not acceptable.

szbergeron:
@martinvuyk I'm curious if you have concrete motivators for this that could drive future design.

martinvuyk:
@barcharcraz could you reopen the PR? You closed it off a bit prematurely without any discussion IMO.

I could rework this to use `_InlineArrayIterOwned` instead. I'm not sure we'll ever get the capability to go from `for i in (...):` to use an implicit conversion to `InlineArray` to then use its iterator.

> Also, we don't want to guarantee that tuples are laid out sequentially in memory like this, so using UnsafePointer in this way is not acceptable.

That is a problem. That means we'll never be able to index tuples dynamically. I was thinking of adding another iterator later for inhomogeneous tuples where the returned elements are `Variant[*Self.element_types]`, but without any way to dynamically index tuples that remains impossible. Unless we are given some way to be able to index it dynamically this severely limits tuple's usefulness at runtime, and hampers a lot of python code from being used as-is

szbergeron:
I'm curious what cases we would want to index tuples dynamically rather than using an InlineArray to start with.

martinvuyk:
> @martinvuyk I'm curious if you have concrete motivators for this that could drive future design.

@szbergeron 

I changed some tests in a project from doing
```mojo
# deeply nested for loops
for i in range(2):
  some_thing = Bool(i)
```
to
```mojo
for some_thing in [True, False]: ...
```
and noticed a significant increase in testing runtime.

So motivations:
1. AFAIK the compiler materializes that list literal as a `List` and not an `InlineArray`.
2. It is very common to iterate over tuples for such short cases in Python


> I'm curious what cases we would want to index tuples dynamically rather than using an InlineArray to start with.

```mojo
def main():
    print(StringSlice("hi") in ("hi", "bye")) # prints False
```
because the type `StringSlice` isn't a type `String`, which is how `Tuple.__contains__` is implemented. So until we have parametric traits and can do `Equatable[T]` or whatever, we can't really simplify it.

But if we were able to iterate over it, then the types would be lowered to `String` and the compiler would realize that it is equatable to a `StringSlice` on it's own without any fancy parametrized trait

szbergeron:
How much could we solve the above by making list literals consistently fast/faster?

I've used the pattern you mention in Python, but this is mainly because tuples are much cheaper in python than lists. It may make sense to try to turn this into a comptime iteration across any tuple if you want (barely) heterogenous values in it (such as a mix of string, slice, and literal), but the truly homogenous case with _dynamic_ indexing seems more solvable with cheap inline array (which is effectively already a dynamically indexable, homogenous tuple)

martinvuyk:
hm now that I think about it I might be wrong on iteration fixing the contains issue. I'm not sure we can yet use an anytype there. But it would allow us to do
```mojo
StrignSlice("hi") in iter(("hi", "bye"))
```

szbergeron:
@barcharcraz / @NathanSWard may be more familiar with the trait conformance fiddlyness you mention above fwiw

martinvuyk:
> How much could we solve the above by making list literals consistently fast/faster?
> 
> I've used the pattern you mention in Python, but this is mainly because tuples are much cheaper in python than lists. It may make sense to try to turn this into a comptime iteration across any tuple if you want (barely) heterogenous values in it (such as a mix of string, slice, and literal), but the truly homogenous case with _dynamic_ indexing seems more solvable with cheap inline array (which is effectively already a dynamically indexable, homogenous tuple)

yeah if we get a cheap way to iterate over list literals I think that would solve the issue I was trying to fix here.

I do still want to enable iteration on homogeneous tuples though. But yeah we can pipe them through an `InlineArray` constructor and the use its iterator

martinvuyk:
Ok so if I had a wishlist it would be that doing `for x in [...]: ...` and `if x in [...]: ...` be cheaper, probably by materializing into `InlineArray` by default when used inline

PR #6491, opened 2026-05-04
Make `BenchId` `Movable`, `Copyable`, `Equatable` and `Hashable`

### LINKED ISSUE #6490: [Feature Request] Make `BenchId` `Copyable` or at least `Movable`
### Review Mojo's priorities

### What is your request?

Make `BenchId` in `stdlib.benchmark` implement the `Copyable` and/or `Movable` trait.

`BenchId` is a trivial struct (contains a `String` and an `Optional[String]`) which is used while implementing benchmark frameworks to reference a specific benchmark function execution. Therefore it is common to be stored in `List` or `Tuple`. But this doesn't currently work, because `Tuple` requires `Movable` and `List` even needs `Copyable`. As this seems like a trivial change enabling more flexibility and future work on benchmarking frameworks I'd suggest implementing this now.

### Any other details?

_No response_

moseschmiedel:
## Summary

`BenchId` now implements the `Movable`, `Copyable`, `Equatable` and `Hashable` trait through default implementations.
Closes #6490 

## Testing

Ran the benchmark tests via `./bazelw test ////mojo/stdlib/benchmarks/...`

## Checklist

      sequence of smaller PRs


moseschmiedel:

mojo/stdlib/std/benchmark/bencher.mojo
```diff
@@ -437,7 +437,7 @@ struct BenchConfig(Copyable):
 
 
 @fieldwise_init
-struct BenchId:
+struct BenchId(Copyable, ImplicitlyCopyable, Movable):
```
JoeLoser: **Question** Do we need the `ImplicitlyCopyable` or is the explicit copying just fine?

Do you mind adding a test case for this? Something like constructing a `List[BenchId]` and show copying them works fine.

JoeLoser (commented):
Hi there! 👋 

Thanks for the contribution. I left a few comments to check out before we merge this.

moseschmiedel:
While adding the test case I also thought about adding `Equatable` and `Hashable`. It would allow us to use the `BenchId` as a `Dict` `KeyElement` which for an identifier would make sense in my opinion. What do you think @JoeLoser ?

NathanSWard:
@moseschmiedel 

> While adding the test case I also thought about adding Equatable and Hashable. It would allow us to use the BenchId as a Dict KeyElement which for an identifier would make sense in my opinion. What do you think @JoeLoser ?

This would also be totally acceptable! Let me know if you want to do that as a followup or if you want to add it to this PR.

moseschmiedel:
@NathanSWard 
> This would also be totally acceptable! Let me know if you want to do that as a followup or if you want to add it to this PR.

I would add it to this PR.

NathanSWard (approved):
thanks!

NathanSWard:
!sync

Issue #6489, opened 2026-05-04
[Feature Request] [stdlib] [proposal] Use `_WriteBufferStack` in the default `write_to` implementation


martinvuyk:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

As title

### What is your motivation for this change?

Most types have functions that repeatedly flush data all the way to the heap, when we could just accumulate it and send it once.

### Any other details?

I'm not sure what kind of impact this would have on stack sizes or other places which is why I'm opening this as a proposal instead of directly implementing it

PR #6494, opened 2026-05-04
[stdlib] Optimize `SIMD.write_to`

martinvuyk:
The highest priority for this function should be to have low latency.

I spent a long time coming up with a very fancy vectorized approach where it processed by power_of_ten chunks, but it turned out a good old scalar while loop was faster.

The approach uses a lookup table for the first 100 digit pairs to be able to extract 2 digits at a time, this is the sweet spot because the next power of two (to be able to use vec loads) for the amount of digits would be 10000 which requires ~40 KiB of memory for the LUT.

## Benchmark results

CPU: Intel® Core™ i7-7700HQ

name | old value (ms) | new value (ms) | markdown perc. | magnitude
-- | -- | -- | -- | --
bench_writer_int_42 | 5.37 | 3.42 | 0.36 | 1.57
bench_writer_int_2**64 | 2.63 | 2.27 | 0.14 | 1.16
bench_writer_simd | 40.42 | 23.21 | 0.43 | 1.74
bench_writer_simd_2**16 | 60.29 | 37.56 | 0.38 | 1.61

### DIFF
--- a/mojo/stdlib/benchmarks/utils/bench_formatter.mojo
+++ b/mojo/stdlib/benchmarks/utils/bench_formatter.mojo
@@ -13,11 +13,13 @@
 
 from std.sys import simd_width_of
 
-from std.benchmark import Bench, BenchConfig, Bencher, BenchId
+from std.benchmark import Bench, BenchConfig, Bencher, BenchId, black_box, keep
 
-# ===-----------------------------------------------------------------------===#
-# Benchmark Data
-# ===-----------------------------------------------------------------------===#
+
+@fieldwise_init
+struct NullWriter(ImplicitlyCopyable, Writer):
+    def write_string(mut self, string: StringSlice):
+        keep(string)
 
 
 # ===-----------------------------------------------------------------------===#
@@ -28,9 +30,10 @@ def bench_writer_int[n: Int](mut b: Bencher) raises:
     @always_inline
     @parameter
     def call_fn():
-        var s1 = String()
-        s1.write(n)
-        _ = s1^
+        for _ in range(10**6):
+            var writer = NullWriter()
+            writer.write(black_box(n))
+            keep(writer)
 
     b.iter[call_fn]()
 
@@ -40,9 +43,12 @@ def bench_writer_simd[n: Int](mut b: Bencher) raises:
     @always_inline
     @parameter
     def call_fn():
-        var s1 = String()
-        s1.write(SIMD[DType.int32, simd_width_of[DType.int32]()](n))
-        _ = s1^
+        for _ in range(10**6):
+            var writer = NullWriter()
+            writer.write(
+                black_box(SIMD[DType.int32, simd_width_of[DType.int32]()](n))
+            )
+            keep(writer)
 
     b.iter[call_fn]()
 
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -72,8 +72,8 @@ from std.sys.intrinsics import _type_is_eq
 from std.bit import bit_width, byte_swap, pop_count
 from std.builtin._format_float import _write_float
 from std.builtin.device_passable import DevicePassable, DeviceTypeEncoder
-from std.builtin.format_int import _write_int
 from std.builtin.int import _FromInt
+from std.format._utils import _WriteBufferStack, _write_int_base_10
 from std.math import DivModable, Powable
 from std.memory import bitcast, memcpy, pack_bits
 from std.python import Python, PythonObject
@@ -2178,21 +2178,17 @@ struct SIMD[dtype: DType, size: Int](
         Args:
             writer: The object to write to.
         """
-
-        # Write an opening `[`.
-        comptime if Self.size > 1:
+        comptime if Self.size == 1:
+            _write_scalar(writer, self[0])
+        else:
             writer.write("[")
 
-        # Write each element.
-        for i in range(Self.size):
-            var element = self[i]
-            # Write separators between each element.
-            if i != 0:
-                writer.write(", ")
-            _write_scalar(writer, element)
+            for i in range(Self.size):
+                var element = self[i]
+                if i != 0:
+                    writer.write(", ")
+                _write_scalar(writer, element)
 
-        # Write a closing `]`.
-        comptime if Self.size > 1:
             writer.write("]")
 
     @no_inline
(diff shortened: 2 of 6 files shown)


mojo/stdlib/benchmarks/utils/bench_formatter.mojo
```diff
@@ -29,8 +29,10 @@ def bench_writer_int[n: Int](mut b: Bencher) raises:
     @parameter
     def call_fn():
         var s1 = String()
-        s1.write(n)
-        _ = s1^
+        for _ in range(10**6):
+            s1.write(black_box(n))
```
NathanSWard: Notably if the resulting value does not fit into the underlying `String`, the string is going to allocate meaning this benchmark is also accounting for the time it takes to allocate (which can have large variability).

If you take a look at `bench_string.mojo` you'll see I did something similar here but used a `NullWriter` which makes sure we don't introduce overhead of the allocation but also makes sure the `write` calls aren't compiled/optimized out via `keep(...)`.

martinvuyk: That should only affect the first execution out of 100 (and only if the resulting string is larger than 24 bytes in 64 bit systems). Could we maybe provide the `NullWriter` as part of the benchamark module?

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -2176,22 +2176,20 @@ struct SIMD[dtype: DType, size: Int](
         Args:
             writer: The object to write to.
         """
+        comptime if Self.size == 1:
```
NathanSWard, line 2181: Quick question on these optimizations take here. 
From looking over the implementation, it seems like was makes the "faster" is by manually having `SIMD` do it's own buffered writing. I assume this helps with `String` since when you write to a `String` w/out buffering it'll cause potentially a few different allocations.

I'm not sure this is however, the best solution here. I don't think we want the types that implement `Writable` to do their own buffered writing. It should be the `Writer` types that do the buffering.
For example, this is often why you have things like the `stdout` be buffered so when it's written to you don't do unnecessary flushing.

I feel like that is that case here too. If someone really wanted this, they should opt into it via `Buffered[String]` or something like that. I don't think doing to buffering inside of SIMD is the optimization we want to go for - also since if the `Writer` _is_ doing buffering, this will actually potentially add overhead as we are now wasting extra memory for every SIMD value we are writing.

NathanSWard (changes requested):

mojo/stdlib/std/builtin/simd.mojo
martinvuyk: When I benchmarked this by only changing the internal implementation (without touching the buffering) both benchmarks were faster than previously

The longer int and simd benchmarks were the only ones that were heavily improved by the buffering. The shorter benchmarks were still fast, but slowed down ~30% (I'm not 100% sure on this number since I'm pulling it from memory) due to the buffering

I'm going to remove the buffering and I'll leave the discussion for #6489

martinvuyk:
@NathanSWard I reworked the underlying algorithm to squeeze out some more performance, left some more detail in the PR description and updated the benchmark values. Now it's the pure numbers without any buffering

mojo/stdlib/test/io/test_write.mojo
```diff
@@ -141,23 +121,57 @@ def test_write_simd_padded() raises:
     SIMD[DType.int32, 2](12345).write_padded(s3, width=3)
     assert_equal(s3, "[12345,12345]")
 
-    s3 = String()
+    s3.resize(0)
```
NathanSWard, line 124: Small nit - not sure of the purpose of using `.resize(0)` here. Granted they end up having the same behavior and side this is test code preserving the backing allocation isn't really an issue. IMO `= String()` reads a bit better.

mojo/stdlib/test/io/test_write.mojo
```diff
@@ -141,23 +121,57 @@ def test_write_simd_padded() raises:
...
     SIMD[DType.int32, 2](-12345).write_padded(s3, width=7)
     assert_equal(s3, "[ -12345, -12345]")
 
-    s3 = String()
+    s3.resize(0)
     SIMD[DType.int8, 4](127, 1, 10, 0).write_padded(s3, width=6)
     assert_equal(s3, "[   127,     1,    10,     0]")
 
 
+def test_write_int() raises:
```
NathanSWard: Since this is specifically testing `write_int_base_10`, can you update the test name to indicate that? `def test_write_int_base_10() raises:`

mojo/stdlib/std/format/_utils.mojo
```diff
@@ -660,3 +667,72 @@ def _write_hex[
...
+    pad_str: StaticString = ""
+](mut writer: Some[Writer], decimal: Scalar, pad_width: Int = 0):
+    """A function that writes a base 10 integer with the lowest possible
+    latency. Anything that needs higher throughput should look elsewhere."""
+    # NOTE: can't use Byte(ord()) here due to recursion
+    comptime `0` = 0x30
+    comptime `-` = 0x2D
+
+    comptime dtype = _uint_type_of_width[bit_width_of[decimal.dtype]()]()
+    comptime buf_size = _max_int_base_10_digits_and_sign[dtype]()
```
NathanSWard: Is this a buffer overflow for signed types? 👀 

`_uint_type_of_width` returns the **unsigned** dtype of the same width. So `_max_int_base_10_digits_and_sign[dtype]()` sees `dtype.is_signed() == False` and never adds 1 for the sign character.

For `Int8.MIN = -128`, buf_size is computed as `1 + Int(log10(UInt8.MAX=255)) + 0 = 3`, but we need 4 bytes (`-`, `1`, `2`, `8`). 

Suggestion:
```mojo
comptime buf_size = _max_int_base_10_digits_and_sign[decimal.dtype]()
```
Keep `dtype` (the unsigned alias) for the `remainder` typing, but size the buffer from `decimal.dtype` so the sign character is accounted for.

mojo/stdlib/test/io/test_write.mojo
NathanSWard: There's also some gaps in the test coverage here:
- No `0` case (smallest value, exercises the `remainder < 10` branch with no prior loop iteration).
- No `Int128` (only `UInt128.MAX` is tested; signed 128-bit min/max isn't covered).

Adding `Int8(0)`, `Int8(127)`, `Int128.MIN`, `Int128.MAX`, seems like good extra coverage for these corner cases.

NathanSWard (changes requested):

mojo/stdlib/test/io/test_write.mojo
martinvuyk: I prefer fixing the performance of our tests when I find them. Some of them take way too long to execute. These do a realloc every time string.write is called which is pointless

mojo/stdlib/std/format/_utils.mojo
martinvuyk: Damn I didn't see this was using the uint dtype, it was meant to use the decimal dtype. Thank you for catching this!!!!!

Issue #6498, opened 2026-05-06
[BUG] Compiler can't prove `Variant` type built with `TypeList` can be `Copyable`


bgreni:
### Actual behavior
If you build a `Variant` with all `Copyable` types using a variadic type list, the compiler can't prove they are all `Copyable`, therefore, the resulting `Variant` type won't be `Copyable`.

### Expected behavior
The `Variant` type should be `Copyable` as is dictated by the `where` clause in it's definition.

### Steps to reproduce

```mojo
from std.utils import Variant

trait Producer(ImplicitlyDestructible, Movable):
    comptime Value: Copyable


@fieldwise_init
struct IntProducer(Producer):
    comptime Value = Int
    var x: Int


@fieldwise_init
struct StringProducer(Producer):
    comptime Value = String
    var s: String


comptime _ProducerValue[T: Producer]: Copyable = T.Value


struct OneOfRepro[*T: Producer](Producer):
    comptime Value = Variant[
        *TypeList[Trait=Producer, Self.T.values]().map[_ProducerValue]() # comptime member 'Value' type 'AnyStruct[Variant[*#kgen.param_list.tabulate(TypeList[T.values], [idx: __mlir_type.index] T.values[idx].Value)]]' does not conform to trait's required type 'Copyable'
    ]

def main():
    pass

```

### System information

```
Pixi version: 0.63.2
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=26.3.1=0
mojo           1.0.0b2.dev2026050406 
```

bgreni:
Looks like this was fixed by ae469bc39cfe965e0fae15bd8d75f2e18e922de1

PR #6504, opened 2026-05-08
[stdlib] Fix `Variant` Copyable conformance for `TypeList.map` packs (#6498)

mahendrarathore1742:
**Problem:** When `Variant[*Ts]` is built from a `TypeList.map[...]()` result, the compiler may fail to prove `AllCopyable[*Ts]`, so `Variant` loses `Copyable` even though every element is `Copyable`.

**Fix:** Add a specialized predicate in `mojo/stdlib/std/reflection/traits.mojo`:
- `comptime AllCopyable[*Ts: Copyable]: Bool = True`

**Test:** Add a regression in `mojo/stdlib/test/utils/test_variant.mojo` that maps `Producer.Value` types and asserts:
- `conforms_to(Variant[*ProducerValues], Copyable)`

**Scope/Risk:** Minimal and safe; only affects packs already constrained to `Copyable`.

**Closes:** #6498


mojo/stdlib/std/reflection/traits.mojo
```diff
@@ -36,6 +36,8 @@ Parameters:
 
 comptime _IsCopyablePredicate[T: AnyType]: Bool = conforms_to(T, Copyable)
 
+comptime AllCopyable[*Ts: Copyable]: Bool = True
+
 comptime AllCopyable[*Ts: AnyType]: Bool = Ts.all_satisfies[
     _IsCopyablePredicate,
 ]()
```
NathanSWard, line 43: Oh wow, this is a super clever trick!

mahendrarathore1742: Thanks! Happy to address any follow-up feedback

NathanSWard (approved):

mojo/stdlib/std/reflection/traits.mojo
NathanSWard: This lgtm! I'll sync it in :)

mojo/stdlib/std/reflection/traits.mojo
```diff
@@ -36,6 +36,8 @@ Parameters:
 
 comptime _IsCopyablePredicate[T: AnyType]: Bool = conforms_to(T, Copyable)
 
+comptime AllCopyable[*Ts: Copyable]: Bool = True
+
```
NathanSWard, line 40: Hey @mahendrarathore1742 - so it seems like this is actually causing some compilation errors. adding this new `AllCopyable` is now causing the following `redefinition` errors to occur 

```
Included from "oss/modular/mojo/stdlib/std/reflection/__init__.mojo":1:
oss/modular/mojo/stdlib/std/reflection/traits.mojo:41:1: error: invalid redefinition of 'AllCopyable'
comptime AllCopyable[*Ts: AnyType]: Bool = Ts.all_satisfies[
^
Included from "oss/modular/mojo/stdlib/std/reflection/__init__.mojo":1:
oss/modular/mojo/stdlib/std/reflection/traits.mojo:39:1: note: previous definition here
comptime AllCopyable[*Ts: Copyable]: Bool = True
^
```

I'm not sure if this is a viable solution, however, would you be willing to file a bug report (GitHub issue) that talks about the original problem you were encountering? It seems like this may need to be a compiler fix and not something we can simply do in the standard library.

NathanSWard (changes requested):

mojo/stdlib/std/reflection/traits.mojo
mahendrarathore1742: > Hey @NathanSWard, thanks for the feedback! I've filed a compiler bug report here: #6538 linking both #6498 and this PR for context.

NathanSWard: Thank @mahendrarathore1742 ! In the meantime I'm going to close this PR since it's not a possible solution and we can re-visit once we have a better solution from the compiler fix.

PR #6508, opened 2026-05-08
[Kernels] fix `ops.conv2d` on CPU with padded input

### LINKED ISSUE #6461: [BUG]: ops.conv2d on CPU silently corrupts boundary output rows of padded input
### Bug description

**Context.** I'm not a MAX or deep learning expert. I was using Claude Opus 4.7 to port the [VoxCPM](https://github.com/OpenBMB/VoxCPM) audio VAE to MAX, our reconstructed audio came out as saturated garbage, and the agent bisected the decoder layer-by-layer to a single op. It then minimized that into the three scripts attached here. I don't fully understand all the internals of the analysis, but Opus is confident this is a real kernel bug rather than a usage error and that the attached scripts are a clean MRE. I'm filing it on that basis. Apologies in advance if I've miscategorized anything — happy to rerun or clarify.

**TL;DR**

- `ops.conv2d` on CPU silently corrupts boundary output rows whenever its input was zero-padded (via `ops.pad` or via `conv2d`'s own `padding=`).
- An `ops.concat([zeros, x, zeros])` produces a bit-identical input tensor and yields correct output through the same `conv2d`. Three equivalent paths to the same conv, only one is right.
- Reproduces in one command: `uv run 01_mre.py` (no setup; PEP 723 metadata pins the deps). Confirmed on `26.2.0`, `26.3.0.dev2026042005`, and today's `26.3.0.dev2026042605` — pattern shifts between versions, never clean.

[01_mre.py](https://github.com/user-attachments/files/27111435/01_mre.py)
[02_determinism.py](https://github.com/user-attachments/files/27111434/02_determinism.py)
[03_shape_sweep.py](https://github.com/user-attachments/files/27111433/03_shape_sweep.py)

---

`max.graph.ops.conv2d` on CPU returns wrong values at boundary output rows whenever its input was zero-padded along a spatial axis. The wrongness takes one of three forms depending on shape and version:

- `NaN` at the affected positions (most common),
- finite but absurd-magnitude values (`~1e+36`, doesn't trip `isnan`), or
- finite values that are non-deterministic across fresh compilations of the same graph with the same input.

Whether the padding is supplied via `ops.pad` upstream or via `conv2d`'s own `padding=` argument makes no difference — both are broken. A mathematically identical input built with `ops.concat([zeros, x, zeros])` (same shape, same numeric values, same zeros in the same positions) produces correct output through the same `conv2d` op. So the wrong output is not a function of the values fed to `conv2d`; it's a function of how the runtime built that tensor.

The affected output rows are always contiguous runs of length ≈`K-1` adjacent to the left and/or right edge of the valid output region — the rows where the kernel window straddles the pad/real-input boundary. That's the geometry of the "border row" codepath in a tiled or im2col-style conv kernel, which is the most plausible location for the bug.

This is filed because a 1D convolution over a padded linear array — the most basic, textbook conv operation — silently corrupts output on perfectly ordinary shapes (`Cin=2048, Cout=1024, K=14`, etc.) on every released and nightly version we've tested. The wrongness is invisible without a reference (the `~1e+36` flavor doesn't trip `isnan`), and the workaround requires knowing this particular trick.

### Steps to reproduce

The three attached scripts are PEP 723 self-contained — the inline metadata pins the MAX nightly and points `uv` at the Modular nightly index, so a single command from a clean machine reproduces each, no venv setup, no installed deps:

```bash
uv run 01_mre.py        # the smoking gun
uv run 02_determinism.py   # uninitialized-memory evidence
uv run 03_shape_sweep.py   # tile-bug fingerprint
```

`uv` builds an ephemeral environment with `max==26.3.0.dev2026042605` (plus `numpy`, plus `torch` for `01` only — used as an independent reference), runs the script, and tears down.

### Expected output of `01_mre.py`

Without the bug, A and B would match C and the torch reference to ~3.7e-5 (float32 accumulation noise). On the affected build:

```
A) ops.pad + conv padding=0                nan= 7168  nan_positions=[6, 7, 8, 9, 10, 11, 12]  finite_max_abs=3.12e+00
B) conv2d(padding=(0,0,13,13))             nan= 7168  nan_positions=[6, 7, 8, 9, 10, 11, 12]  finite_max_abs=3.12e+00
C) ops.concat([z, x, z])  (workaround)     nan=    0  nan_positions=[]  finite_max_abs=3.74e-05
```

7 contiguous output positions (`t = 6..12` for `K=14`, `Lu=50`) carry NaN in all 1024 output channels, and the *non-NaN* values of A and B drift from the reference by up to 3.12 — so the kernel is producing wrong values *outside* the NaN region too.

### Sample output of `03_shape_sweep.py`

```
  Cin  Cout       nan  nan_positions
   16    16       112  [50, 51, 52, 53, 54, 55, 56]
   32    32       224  [50, 51, 52, 53, 54, 55, 56]
   64    64         0  []
  128   128         0  []
  256   256         0  []
  512   512      7168  [6, 7, 8, 9, 10, 11, 12, 50, 51, 52, 53, 54]…
 2048  1024     11264  [6, 7, 8, 9, 10, 11, 12, 53, 54, 55, 56]
 2048  2048     20480  [6, 7, 8, 9, 10, 11, 12, 54, 55, 56]
 1024  2048         0  []
```

Notice that `(2048, 1024)` is broken but the swapped `(1024, 2048)` is clean — strong fingerprint that the bug lives in tile-size / SIMD-tail dispatch logic rather than in the conv arithmetic.

### Sample output of `02_determinism.py`

On `26.2.0`, run 0 gives a "normal" pattern with `fin_sum=4.836e+00` and 7168 NaN, while runs 1–4 give `nan=0` but `fin_sum=7.907e+37`. On Apr 20 nightly the first compile yields `fin_max=2.199e+36` and subsequent compiles settle at `4.836e+00`. On Apr 26 nightly all 5 runs deterministically yield `nan=0` with `fin_max=2.434e+36` — same wrongness, harder to detect.

## Suspected internals

The geometry of the broken positions — contiguous runs of length `K-1` adjacent to the left/right edges of the valid output range — matches exactly the "border row" codepath of a tiled or im2col-style conv kernel: the rows where some kernel taps land in the padded region while others land in real input. A plausible mental model is that the codepath that fills the lowering buffer for those rows is failing to zero (or skip-and-add-zero) the columns corresponding to padded positions, so the conv reads whatever was in that memory.

This explains all three observations:

- C (concat-zeros) is clean because the concat materializes real zeros in the W axis, so every output row is an "interior" row from the conv's perspective; the buggy border codepath never runs.
- The non-determinism in `02_determinism.py` is the contents of that uninitialized buffer differing between fresh compiles.
- The non-monotonic shape dependence in `03_shape_sweep.py` is which kernel variant / tile width gets dispatched, with only some variants having the bug.

I'd start at the CPU `conv2d` border-row / lowering-buffer initialization for the RSCF layout, `groups=1`, `dilation=1`, no-stride configuration.

### System information

```shell
This project is managed with `uv` rather than `pixi`, so `pixi list max` is not applicable; the PEP 723 metadata at the top of each MRE pins the exact MAX version. `pixi info` for the host:


System
------------
       Pixi version: 0.63.2
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=26.3.1=0
                   : __archspec=1=m1


- Host: macOS 26.3.1, Darwin 25.3.0, arm64 (Apple Silicon, M-series).
- Python 3.12.11, `uv` 0.11.7.
- Primary version under test: `max==26.3.0.dev2026042605` (today's nightly).
- Also confirmed on `max==26.2.0` and `max==26.3.0.dev2026042005`. The bug exists on all three; the exact NaN/garbage pattern shifts between versions but it's never gone.
- Device: CPU (`DeviceRef.CPU()`) only. We have not been able to test on Metal because `ops.conv2d_transpose` fails to lower for our shapes (separate bug, not filed yet).
```

larme:
## Linked issue

Fixes #6461

## Type of change

- [ ] Documentation update

## Motivation

Fix 2d convolution on CPU with heavily padded input.

## What changed

- in `max/kernels/src/nn/conv/conv.mojo`, `_h_loop_static` used to assume padding size is always smaller than `micro_kernel_height`, hence sending middle of the row still contains padding-affected output columns to `_inner_loops_static`. This PR fix this problem by 
- This PR also fix `max/kernels/test/nn/test_static_conv.mojo` only return when failed without raising an error

## Testing

 This PR added a static 2D conv regression case in `max/kernels/test/nn/test_static_conv.mojo`:

- `H=1, W=50, R=1, S=14`
- `dilation=(1, 3)`
- `pad_w=(13, 13)`

This shape has output affecting padding wider than the old static boundary micro-kernel split, so it will catch the bug directly. 

## Checklist

      approval
      [pull request sizes](../mojo/CONTRIBUTING.md#about-pull-request-sizes))

Assisted-by: Claude

### DIFF
--- a/max/kernels/test/nn/test_static_conv.mojo
+++ b/max/kernels/test/nn/test_static_conv.mojo
@@ -12,7 +12,6 @@
 # ===----------------------------------------------------------------------=== #
 
 from std.math import ceildiv, isclose
-from std.random import rand
 from std.sys.info import simd_width_of
 
 from std.itertools import product
@@ -48,8 +47,8 @@ def test[
 ]() raises:
     # Output Shape.
     # fmt: off
-    comptime HO = (H + pad_h[0] + pad_h[1] - dilation[1] * (R - 1) - 1) // stride[0] + 1
-    comptime WO = (W + pad_w[0] + pad_w[1] - dilation[0] * (S - 1) - 1) // stride[1] + 1
+    comptime HO = (H + pad_h[0] + pad_h[1] - dilation[0] * (R - 1) - 1) // stride[0] + 1
+    comptime WO = (W + pad_w[0] + pad_w[1] - dilation[1] * (S - 1) - 1) // stride[1] + 1
     # fmt: on
     comptime type = DType.float32
     comptime simd_size = simd_width_of[type]()
@@ -70,15 +69,39 @@ def test[
         num_groups=num_groups,
     )
 
-    var input_ptr = List(length=N * H * W * C, fill=Scalar[type](0))
+    var input_size = N * H * W * C
+    var input_guard_size = C * (
+        W * max(pad_h[0], pad_h[1]) + max(pad_w[0], pad_w[1])
+    )
+    var input_storage_ptr = List(
+        length=input_size + 2 * input_guard_size, fill=Scalar[type](0)
+    )
+    var input_ptr = input_storage_ptr.unsafe_ptr() + input_guard_size
     var filter_ptr = List(length=R * S * C * F, fill=Scalar[type](0))
 
     # output from conv w/ dynamic and static shapes.
     var output_ptr_static = List(length=N * HO * WO * F, fill=Scalar[type](0))
     var output_ptr_dynamic = List(length=N * HO * WO * F, fill=Scalar[type](0))
 
-    rand(input_ptr)
-    rand(filter_ptr)
+    for i in range(input_guard_size):
+        input_storage_ptr[i] = Scalar[type](100.0)
+        input_storage_ptr[input_guard_size + input_size + i] = Scalar[type](
+            -100.0
+        )
+
+    var input_value = -8
+    for i in range(input_size):
+        input_ptr[i] = Scalar[type](Float64(input_value) * 0.01)
+        input_value += 1
+        if input_value > 8:
+            input_value = -8
+
+    var filter_value = -6
+    for i in range(R * S * C * F):
+        filter_ptr[i] = Scalar[type](Float64(filter_value) * 0.01)
+        filter_value += 1
+        if filter_value > 6:
+            filter_value = -6
 
     comptime layout_4d = Layout.row_major[4]()
     comptime layout_5d = Layout.row_major[5]()
@@ -204,7 +227,7 @@ def test[
                 "rerr",
                 abs(actual - expected) / abs(expected + 1e-10),
             )
-            return
+            raise Error("static conv output mismatch")
 
     # CHECK: Succeed
     print("Succeed")
@@ -213,7 +236,7 @@ def test[
     _ = packed_filter_ptr_static^
     _ = packed_filter_ptr_dynamic^
     _ = filter_ptr^
-    _ = input_ptr^
+    _ = input_storage_ptr^
 
 
 def main() raises:
@@ -242,6 +265,19 @@ def main() raises:
         Index(1, 1),  # dilation
         Index(1, 1),  # pad_h
         Index(1, 1),  # pad_w
+    ]()
+    test[
+        1,  # N
+        1,  # H
+        50,  # W
+        8,  # C
+        1,  # R
+        14,  # S
+        64,  # F
+        Index(1, 1),  # stride
+        Index(1, 3),  # dilation
+        Index(0, 0),  # pad_h
+        Index(13, 13),  # pad_w
     ]()
 
     # Each test will build a specialization of the conv kernel.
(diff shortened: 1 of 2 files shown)


max/kernels/test/nn/test_static_conv.mojo
```diff
@@ -204,7 +227,7 @@ def test[
                 "rerr",
                 abs(actual - expected) / abs(expected + 1e-10),
             )
-            return
+            raise Error("static conv output mismatch")
```
larme, line 230: I'm not sure why originally here only have a return instead of raising an error

yongmzhang (approved):

yongmzhang:
!sync

PR #6510, opened 2026-05-09
[Stdlib] DRY repeated unsafe_value() calls in WeakPointer

msaelices:
DRY redundant `unsafe_value()` calls in `WeakPointer`. `__del__`, `try_upgrade`, and `weak_count` each called `self._inner.unsafe_value()` two or three times in a row. Bind the inner control-block pointer once via `var` (or `ref` for `weak_count`).

## Codegen

Source-level cleanup, **not** a release-build perf change. Verified with the bazel-built mojo toolchain at `-O3`, before/after:

```
mojo build -O3 --emit=asm /tmp/driver.mojo
md5sum: identical (d655e8ae300e5c05e429640ff7857faa)
diff:   exit 0 (no differences)
```

LLVM IPO already collapses the redundant `unsafe_value()` calls in optimized builds.

### DIFF
--- a/mojo/stdlib/std/memory/arc_pointer.mojo
+++ b/mojo/stdlib/std/memory/arc_pointer.mojo
@@ -509,8 +509,11 @@ struct WeakPointer[T: Movable & ImplicitlyDestructible](
     @no_inline
     def __del__(deinit self):
         """Decrement the weak count and free the allocation if last."""
-        if self._inner and self._inner.unsafe_value()[].drop_weak():
-            free(self._inner.unsafe_value(), {count = 1})
+        if not self._inner:
+            return
+        var inner_ptr = self._inner.unsafe_value()
+        if inner_ptr[].drop_weak():
+            free(inner_ptr, {count = 1})
 
     def try_upgrade(self) -> Optional[ArcPointer[Self.T]]:
         """Attempts to obtain a strong reference.
@@ -519,9 +522,12 @@ struct WeakPointer[T: Movable & ImplicitlyDestructible](
             An `ArcPointer` sharing the allocation, or `None` if the
             payload has already been destroyed (strong count reached 0).
         """
-        if self._inner and self._inner.unsafe_value()[].try_add_strong():
-            return {ArcPointer[Self.T](_inner=self._inner.unsafe_value())}
-        return Optional[ArcPointer[Self.T]]()
+        if not self._inner:
+            return None
+        var inner_ptr = self._inner.unsafe_value()
+        if inner_ptr[].try_add_strong():
+            return {ArcPointer[Self.T](_inner=inner_ptr)}
+        return None
 
     def strong_count(self) -> UInt64:
         """Returns the current strong count, or 0 if the payload is gone.
@@ -543,12 +549,12 @@ struct WeakPointer[T: Movable & ImplicitlyDestructible](
             The approximate number of outstanding `WeakPointer`s.
         """
 
-        if self._inner:
-            var w = self._inner.unsafe_value()[].weak_count_with_implicit()
-            if self._inner.unsafe_value()[].strong_count() == 0:
-                return w
-            # If there are any strong remaining, we don't want to
-            # include the implicit weak in the returned count.
-            return w - 1
-        else:
+        if not self._inner:
             return 0
+        ref inner = self._inner.unsafe_value()[]
+        var w = inner.weak_count_with_implicit()
+        if inner.strong_count() == 0:
+            return w
+        # If there are any strong remaining, we don't want to
+        # include the implicit weak in the returned count.
+        return w - 1


mojo/stdlib/std/memory/arc_pointer.mojo
```diff
@@ -509,8 +509,11 @@ struct WeakPointer[T: Movable & ImplicitlyDestructible](
     @no_inline
     def __del__(deinit self):
         """Decrement the weak count and free the allocation if last."""
-        if self._inner and self._inner.unsafe_value()[].drop_weak():
-            free(self._inner.unsafe_value(), {count = 1})
+        if not self._inner:
+            return
+        var inner_ptr = self._inner.unsafe_value()
+        if inner_ptr[].drop_weak():
+            free(inner_ptr, {count = 1})
```
NathanSWard, line 516: One note about this, the multiple calls to `.unsafe_value()` doesn't introduce a branch in release builds (only debug builds) - hence the name "unsafe" value (since it doesn't do any runtime checking of the optional's "engaged" status at runtime.

mojo/stdlib/test/memory/test_arc.mojo
```diff
@@ -360,5 +360,82 @@ def test_lifetime_probe_arc_drop_with_weak_in_scope() raises:
...
+
+
+def test_tree_with_weak_parent_links() raises:
+    var root_dropped = False
+    var a_dropped = False
+    var b_dropped = False
+
+    var root = ArcPointer(
+        TreeNode(
+            1, ObservableDel(UnsafePointer(to=root_dropped).as_any_origin())
```
NathanSWard:
```suggestion
            1, ObservableDel(UnsafePointer(to=root_dropped))
```
The `as_any_origin` shouldn't beed needed here.

mojo/stdlib/test/memory/test_arc.mojo
```diff
@@ -360,5 +360,82 @@ def test_lifetime_probe_arc_drop_with_weak_in_scope() raises:
...
+    assert_false(b_dropped)
+    assert_false(root_dropped)
+
+    # Dropping root cascades: root payload destroyed -> children list dropped
+    # -> each child Arc dropped to zero -> child payloads destroyed -> child
+    # parent WeakPointers drop root's weak count -> root's allocation freed.
+    _ = root^
+    assert_true(root_dropped)
+    assert_true(a_dropped)
+    assert_true(b_dropped)
```
NathanSWard: I'm not sure these tests are actually testing deletion order correctly . Notably with the introducton of `AnyOrigins` via the `as_any_origin()` calls above, **ALL* values in the test get their lifetimes extended until the `_ = root^` not due to `Arc/Weak` correctly handling lifetimes, but instead due to the origin system.

mojo/stdlib/test/memory/test_arc.mojo
szbergeron: And further, this may cause Weak to behave strangely and not do ASAP deletion ^

mojo/stdlib/std/memory/arc_pointer.mojo
msaelices: Confirmed with an asm diff, `mojo build -O3 --emit=asm` produces byte-identical output before and after the DRY change (md5 matches, `diff` exits 0). So this commit is purely source-level readability. Happy to drop the commit altogether if you like

mojo/stdlib/test/memory/test_arc.mojo
msaelices: Good catch. Gone in the latest push (a05eee1648). Restructured `TreeNode` to be parametric on origin (`TreeNode[origin: MutOrigin]`).

mojo/stdlib/test/memory/test_arc.mojo
msaelices: You were right, the test wasn't proving what it claimed. Reworked it (latest push): swapped the per-node `ObservableDel` markers for a single shared `Int` counter held behind a `Pointer` with a concrete origin.

szbergeron:
Hi @msaelices, looking this over a bit more. I think the refactor in Weak itself is a positive but we're not sure we want to do the tree test as a standalone test. Mechanically, the existing tests for Arc and Weak should cover the behavior needed from them without introducing as much dependence on other language features.

If we want to come back otherwise and consider actually adding various tree data structures to std that's another story (and could act as a test in its own right), but that would require further discussion.

msaelices:
> Hi @msaelices, looking this over a bit more. I think the refactor in Weak itself is a positive but we're not sure we want to do the tree test as a standalone test. Mechanically, the existing tests for Arc and Weak should cover the behavior needed from them without introducing as much dependence on other language features.
> 
> If we want to come back otherwise and consider actually adding various tree data structures to std that's another story (and could act as a test in its own right), but that would require further discussion.

Great. Do you want me to remove the tests and keep the readability changes?

Also, I could create an issue in GH to discuss the tree structures. I think we should maintain at least one recursive data structure somewhere in the stdlib code base, just to notice when we break something (via compiler or stdlib). That was the actual intent when I wrote the tests for trees.

szbergeron:
> > Hi @msaelices, looking this over a bit more. I think the refactor in Weak itself is a positive but we're not sure we want to do the tree test as a standalone test. Mechanically, the existing tests for Arc and Weak should cover the behavior needed from them without introducing as much dependence on other language features.
> > If we want to come back otherwise and consider actually adding various tree data structures to std that's another story (and could act as a test in its own right), but that would require further discussion.
> 
> Great. Do you want me to remove the tests and keep the readability changes?
> 
> Also, I could create an issue in GH to discuss the tree structures. I think we should maintain at least one recursive data structure somewhere in the stdlib code base, just to notice when we break something (via compiler or stdlib). That was the actual intent when I wrote the tests for trees.

Both sound great to me 👍. I'm personally in favor of having trees of some form in the stdlib, though we want to motivate it enough and have concrete uses we can use to shake out any architectural/API issues. It's likely an area that would get used (and misused) with some frequency, so having something that stands up to whatever people may throw at it is tricky.

For this PR, if you want to pull out the tree test and keep the code changes that would work great.

msaelices:
> > > Hi @msaelices, looking this over a bit more. I think the refactor in Weak itself is a positive but we're not sure we want to do the tree test as a standalone test. Mechanically, the existing tests for Arc and Weak should cover the behavior needed from them without introducing as much dependence on other language features.
> > > If we want to come back otherwise and consider actually adding various tree data structures to std that's another story (and could act as a test in its own right), but that would require further discussion.
> > 
> > 
> > Great. Do you want me to remove the tests and keep the readability changes?
> > Also, I could create an issue in GH to discuss the tree structures. I think we should maintain at least one recursive data structure somewhere in the stdlib code base, just to notice when we break something (via compiler or stdlib). That was the actual intent when I wrote the tests for trees.
> 
> Both sound great to me 👍. I'm personally in favor of having trees of some form in the stdlib, though we want to motivate it enough and have concrete uses we can use to shake out any architectural/API issues. It's likely an area that would get used (and misused) with some frequency, so having something that stands up to whatever people may throw at it is tricky.
> 
> For this PR, if you want to pull out the tree test and keep the code changes that would work great.

Done! I got rid of the tests and adapted the PR description.

Created GH Issue to discuss about recursive structures: https://github.com/modular/modular/issues/6522

szbergeron:
!sync

barcharcraz:
We looked at this and we don't think it's an improvement.

Issue #6521, opened 2026-05-11
Mojo from Python per-call FFI overhead ~10x higher than PyO3


msaelices:
## Summary

Calling a trivial Mojo function from Python through `PythonModuleBuilder` measures **~150-175 ns** per call for a no-op and **~305-330 ns** for a two-`Int`-conversion `add(a, b)`. The same shape implemented in PyO3 measures **~17-24 ns** and **~30-32 ns** under identical conditions on the same machine.

Mojo's Python <-> Mojo FFI is roughly an **order of magnitude slower than PyO3's** on the simplest possible call shapes. Filing for visibility and to ask whether a `METH_FASTCALL`-equivalent fast path is on the roadmap for the `PythonModuleBuilder` dispatch.

## Environment

- **CPU**: Intel i7-12700H (12th Gen Alder Lake, 14C/20T, 4.7 GHz max)
- **OS**: Slimbook OS 24 (Ubuntu 24.04 base), Linux 6.17.0-19-generic
- **Mojo**: 1.0.0b2.dev2026051106 (aff922aa) — bazel toolchain (see `MODULE.bazel`)
- **Python**: 3.12.3 (system venv)
- **PyO3** (for comparison): 0.22 with `lto = true, opt-level = 3, codegen-units = 1`, rustc 1.80.1

## Methodology

- `timeit.repeat(number=2_000_000, repeat=7)`, report **minimum** across repeats (standard microbench practice; min is the closest estimate of true per-call cost, mean is contaminated by transient scheduler / cache / other-process noise)
- Process pinned to a single CPU core with `taskset -c 2` to remove migration jitter
- 3 independent benchmark runs per configuration; reported numbers are the min/range across those runs
- Baselines included: pure-Python `noop` and `add` functions, and the timeit-loop floor (`1 + 2` with no function call)

## Results (min ns per call)

| Call shape | Mojo 1.0.0b2.dev | PyO3 0.22 |
|------------|--------------|-----------|
| `noop(x)` (return arg untouched) | **~150-175** | **~17-24** |
| `add(a, b)` (two `Int`/`i64` conversions + add + wrap) | **~305-330** | **~30-32** |
| `py_noop(x)` baseline (pure Python) | ~22-28 | ~22-28 |
| `py_add(a, b)` baseline (pure Python) | ~29-31 | ~29-31 |
| `1 + 2` (timeit floor, no call) | ~6 | ~6 |

[...]

PR #6523, opened 2026-05-11
[Stdlib] Skip redundant GIL acquires on Python->Mojo FFI hot path

msaelices:
## Linked issue

Related to #6521. This PR addresses the lowest-hanging-fruit portion of that issue: redundant GIL bookkeeping on the Python -> Mojo trampoline. The bigger structural wins (METH_FASTCALL, borrowed-PythonObject) are noted as follow-ups below.

## Motivation

Issue #6521 reports that calling a trivial Mojo function from Python costs ~150-170 ns/call vs PyO3's ~17-24 ns/call. Reading the dispatch path and instrumenting it with `std.benchmark` shows the per-noop overhead breaks down roughly like this on Mojo 1.0.0b2.dev2026051106 (bazel toolchain) + CPython 3.12, single core via `taskset -c 2`:

| Component | ns/call |
|---|---|
| stdlib `_dispatch[is_method]` chain | ~76 |
| **Outer `GILAcquired` in `_py_c_function_wrapper`** | **~17** |
| **2× `PythonObject.__del__` with `GILAcquired`** | ~42 (this PR shaves ~12) |
| `PythonObject.from_borrowed` incref/wrap | ~20 |
| Tuple build by CPython (METH_VARARGS) | ~30-50 |

The two GIL-related rows together account for **~30 ns** of redundant work that the wrapper does on every call, so the GIL is already held by definition when CPython dispatches our `METH_VARARGS` trampoline, and the same applies to the destructor that runs as part of that trampoline.

### Commit 1: Skip redundant GIL acquire in `PythonObject.__del__`

`PythonObject.__del__` unconditionally wrapped its `Py_DecRef` in a `PyGILState_Ensure` / `PyGILState_Release` pair so it could run safely even on a thread that doesn't hold the GIL. In the common case the GIL *is* already held (we're inside a Python -> Mojo trampoline, or inside a `with Python()` block); the pair is then just two extra C calls into CPython.

| Primitive (cached fn ptr) | ns / call |
|---|---|
| `PyGILState_Ensure` + `PyGILState_Release` (pair) | 14.5 |
| `PyGILState_Check` | 8.3 |

So ~6 ns saved per `__del__`. The FFI trampoline destroys 2 `PythonObject`s per call (`self` and the args tuple), so this is **~12 ns/call** of saved overhead on the hot path.

### Commit 2: Drop outer `GILAcquired` from `_py_c_function_wrapper`

When CPython dispatches via `METH_VARARGS` / `METH_VARARGS | METH_KEYWORDS`, the calling thread already holds the GIL. PyO3, pybind11, and nanobind all rely on this invariant and do **not** re-acquire on entry. The Mojo wrapper, by contrast, opened an explicit `GILAcquired` block on every call.

Decomposed A/B bench, all running against the *current* stdlib (so this is apples-to-apples for the wrapper variants):

| Variant (return arg unchanged, like `def noop(x): return x`) | ns / call |
|---|---|
| A: stdlib `def_function` (current path) | **198.9** |
| B: hand-rolled wrapper, PythonObject wraps + outer GIL | 123.4 |
| C: hand-rolled wrapper, PythonObject wraps, **no outer GIL** | 106.4 |
| D: raw, no PythonObject wraps, no GIL (lower bound) | 63.8 |
| (Py lambda baseline) | 94.6 |

`B - C = 17 ns` is the direct measurement of this commit's saving.

## Combined effect

With both commits applied, the FFI trampoline shaves ~29 ns/call:

- Outer GIL removal (commit 2): **17 ns** (measured directly)
- Conditional GIL in `__del__` (commit 1): **~12 ns** (calculated from cached-pointer primitive measurements; the second-order effect — fewer C calls per destructor — applies twice per FFI call)

So for the `noop(x)` shape in issue #6521, expected improvement is ~170 → ~140 ns/call (**~17%**). For `add(a, b)` the relative improvement is smaller because the per-arg `Int(py=a)` conversion dominates the remaining overhead.


msaelices:
@JoeLoser's [stack](https://github.com/modular/modular/issues/6521#issuecomment-4426212508) covers items 2 + 3 from his decomposition (the changes in this PR), plus items 4 + 5 which I had identified but deferred. His after-numbers (≈22 ns noop / ≈42 ns add on M4) are dramatically better. Yielding, will close this PR once his stack lands publicly. Pivoting to item 6 (typed-arg fast paths) which he explicitly hasn't looked at.

msaelices:
Superseded: upstream now skips the `PyGILState_Ensure`/`Release` round-trip via `PyGILState_Check()` on the Python->Mojo FFI hot path. Closing as already implemented upstream.

PR #6526, opened 2026-05-12
[Stdlib] Typed-arg fast-path bindings: Int <-> PyLong

msaelices:
## Linked issue

Related to #6521. This PR addresses **item 6** of @JoeLoser's decomposition (typed-argument fast paths for `Int(py=...)`).

## Type of change

- [x] Performance improvement (measurements below)
- [x] New user-facing API: `PythonModuleBuilder.def_typed_function`

## Motivation

When users expose a Mojo function to Python via `PythonModuleBuilder.def_function`, they currently have to declare `PythonObject` arguments and convert internally:

```mojo
def add(a: PythonObject, b: PythonObject) raises -> PythonObject:
    return PythonObject(Int(py=a) + Int(py=b))
m.def_function[add]("add")
```

`Int(py=a)` goes through `__int__()` -> `PyNumber_Long`, which **allocates a fresh `PyLong` per argument per call** before extracting the `Py_ssize_t`. For `add(a, b)` that's two allocations on the args plus one for the return.

## What this PR adds

A typed-argument fast path that lets users declare concrete types:

```mojo
def add(a: Int, b: Int) -> Int:
    return a + b
m.def_typed_function[add]("add")
```

The trampoline calls `PyLong_AsSsize_t` directly on each borrowed tuple slot and `PyLong_FromSsize_t` on the return, so no `PythonObject` wraps, no `PyNumber_Long` allocation.

## Shapes supported in this PR

| Signature | Trampoline |
|---|---|
| `def f(a: Int) -> Int` (with or without `raises`) | `_trampoline_int_to_int` |
| `def f(a: Int, b: Int) -> Int` (with or without `raises`) | `_trampoline_int_int_to_int` |

Non-raising user functions are implicitly compatible with the raising signature; one overload per arity is enough.

Extension to `Float64` (via `PyFloat_AsDouble`/`PyFloat_FromDouble`), `Bool` (via `PyObject_IsTrue`/`PyBool_FromLong`), `UInt`, and `StringSlice` (via `PyUnicode_AsUTF8AndSize`) follows the same pattern. Left as follow-ups.

## Bench results

`./bazelw test //mojo/stdlib/benchmarks:python/bench_python_ffi_typed.mojo.bench --test_output=all`, Mojo `1.0.0b2.dev2026051106 (aff922aa)` bazel toolchain, CPython 3.13, single core (`taskset -c 2`):

| Variant | min ns/iter | median ns/iter |
|---|---|---|
| `a_def_function_add` (PythonObject + `Int(py=...)`) | 566 | 880 |
| `b_typed_int_int_add` (typed fast path) | **164** | 250 |

**Min delta: ~400 ns/iter saved (~71%)** on the `add(Int, Int) -> Int` shape.

Assisted-by: AI

### DIFF
--- a/mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
+++ b/mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
@@ -1,3 +1,15 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
 """Benchmark typed-argument fast-path wrappers vs the standard
 `def_function[user_func]` path.
 
@@ -92,7 +104,5 @@ def main() raises:
     _ = Python()
     var m = Bench(BenchConfig(num_repetitions=5, max_runtime_secs=2.0))
     m.bench_function[bench_a_def_function_add](BenchId("a_def_function_add"))
-    m.bench_function[bench_b_typed_int_int_add](
-        BenchId("b_typed_int_int_add")
-    )
+    m.bench_function[bench_b_typed_int_int_add](BenchId("b_typed_int_int_add"))
     print(m)
--- a/mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
+++ b/mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
@@ -32,8 +32,8 @@ def add_pyobj(a: PythonObject, b: PythonObject) raises -> PythonObject:
     return PythonObject(Int(py=a) + Int(py=b))
 
 
-# --- Variant B: typed fast-path — `def_function` overload picks the
-# right wrapper based on the user's function signature.
+# --- Variant B: typed fast-path via `def_typed_function`, which picks
+# the right wrapper based on the user's function signature.
 
 
 def add_int(a: Int, b: Int) -> Int:
--- a/mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
+++ b/mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
@@ -1,0 +1,98 @@
+"""Benchmark typed-argument fast-path trampolines vs the standard
+`def_function[user_func]` path.
+
+A: stdlib `def_function[add]` where `add` is
+   `def(PythonObject, PythonObject) raises -> PythonObject` and the body
+   does `Int(py=a) + Int(py=b)` (allocates two PyLongs internally
+   via PyNumber_Long).
+
+B: typed trampoline registered via `def_py_c_function`, with `add`
+   declared as `def(Int, Int) -> Int`. The trampoline calls
+   `PyLong_AsSsize_t` directly on each tuple slot, skipping the
+   PythonObject + PyNumber_Long round-trip.
+
+This is item 6 from issue #6521.
+"""
+
+from std.benchmark import (
+    Bench,
+    BenchConfig,
+    Bencher,
+    BenchId,
+    keep,
+)
+from std.python import Python, PythonObject
+from std.python.bindings import PythonModuleBuilder
+
+
+# --- Variant A: current def_function path with PythonObject args ---
+
+
+def add_pyobj(a: PythonObject, b: PythonObject) raises -> PythonObject:
+    return PythonObject(Int(py=a) + Int(py=b))
+
+
+# --- Variant B: typed fast-path — `def_function` overload picks the
+# right trampoline based on the user's function signature.
+
+
+def add_int(a: Int, b: Int) -> Int:
+    return a + b
+
+
+@parameter
+def bench_a_def_function_add(mut b: Bencher) raises:
+    _ = Python()
+    var m = PythonModuleBuilder("_ffi_typed_bench_a")
+    m.def_function[add_pyobj]("add")
+    var mod = m.finalize()
+    var f = mod.add
+    var arg_a = PythonObject(1)
+    var arg_b = PythonObject(2)
+
+    @always_inline
+    @parameter
+    def call_fn() raises:
+        for _ in range(1000):
+            var r = f(arg_a, arg_b)
+            keep(r)
+
+    b.iter[call_fn]()
+    keep(mod)
+    keep(arg_a)
+    keep(arg_b)
+    keep(f)
+
+
+@parameter
+def bench_b_typed_int_int_add(mut b: Bencher) raises:
+    _ = Python()
+    var m = PythonModuleBuilder("_ffi_typed_bench_b")
+    m.def_typed_function[add_int]("add")
+    var mod = m.finalize()
+    var f = mod.add
+    var arg_a = PythonObject(1)
+    var arg_b = PythonObject(2)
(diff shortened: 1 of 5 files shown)


msaelices:
This matches to item 6 on @JoeLoser's list, see https://github.com/modular/modular/issues/6521#issuecomment-4426212508

This would compose with the remaining items. Once his METH_FASTCALL + dispatch consolidation lands, the two optimizations stack: typed args avoid the `Int(py=...)` allocation, his work avoids the tuple build + dispatch indirection. End-state estimate is within ~1.2-1.5× of PyO3 on the `add` shape.

When this PR rebases on top of his work, the `_trampoline_*` functions can switch from METH_VARARGS to METH_FASTCALL by adopting the fastcall signature (`(self, args[], nargs)`) — a one-line change of registration plus replacing `PyTuple_GetItem(args, i)` with `args[i]`.

mojo/stdlib/std/python/_python_func_typed.mojo
```diff
@@ -0,0 +1,105 @@
...
+
+def _typed_int_to_int_wrapper[
+    user_func: _FnIntToInt
+](_self: PyObjectPtr, args: PyObjectPtr) -> PyObjectPtr:
+    ref cpy = Python().cpython()
+    try:
+        var a = _get_int_arg(cpy, args, 0)
+        if not a[1]:
+            return PyObjectPtr()
+        return cpy.PyLong_FromSsize_t(user_func(a[0]))
```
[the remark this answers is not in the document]
msaelices: Good catch. Fixed in cffc80ad42 — both 1-arg and 2-arg wrappers now call `_check_arity` (via `CPython.PyObject_Length`) before fetching args and set `PyExc_TypeError` on mismatch.

mojo/stdlib/std/python/_python_func_typed.mojo
```diff
@@ -0,0 +1,105 @@
...
+](_self: PyObjectPtr, args: PyObjectPtr) -> PyObjectPtr:
+    ref cpy = Python().cpython()
+    try:
+        var a = _get_int_arg(cpy, args, 0)
+        if not a[1]:
+            return PyObjectPtr()
+        var b = _get_int_arg(cpy, args, 1)
+        if not b[1]:
+            return PyObjectPtr()
+        return cpy.PyLong_FromSsize_t(user_func(a[0], b[0]))
```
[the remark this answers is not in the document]
msaelices: Fixed alongside the 1-arg case in cffc80ad42.

mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
```diff
@@ -0,0 +1,98 @@
...
+# --- Variant A: current def_function path with PythonObject args ---
+
+
+def add_pyobj(a: PythonObject, b: PythonObject) raises -> PythonObject:
+    return PythonObject(Int(py=a) + Int(py=b))
+
+
+# --- Variant B: typed fast-path — `def_function` overload picks the
+# right wrapper based on the user's function signature.
+
```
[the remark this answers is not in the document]
msaelices: Fixed in cffc80ad42 — comment now references `def_typed_function` to match the code.

mojo/stdlib/std/python/_python_func_typed.mojo
```diff
@@ -0,0 +1,131 @@
...
+    Py_ssize_t,
+)
+
+
+# Compile-time aliases for the supported user-function shapes.
+# A single raising signature per arity; non-raising user functions are
+# implicitly compatible with the raising form (they simply never raise),
+# which avoids overload ambiguity between raises and non-raises shapes.
+comptime _FnIntToInt = def(Int) thin raises -> Int
+comptime _FnIntIntToInt = def(Int, Int) thin raises -> Int
```
martinvuyk, line 40: I can't help fut feel like this is the wrong fix. If we got a way to reflect over a function argument and return types this could be solved generically for all types in python that could be converted more trivially directly to Mojo raw types

martinvuyk (commented):
Is there any particular reason this specific kind of conversion is needed? We should be able to justify hardcoding something like this

mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
```diff
@@ -0,0 +1,108 @@
...
+    var arg_b = PythonObject(2)
+
+    @always_inline
+    @parameter
+    def call_fn() raises:
+        for _ in range(1000):
+            var r = f(arg_a, arg_b)
+            keep(r)
+
+    b.iter[call_fn]()
```
NathanSWard, line 72: There is an overload that works with unified closures, lets use that instead of the old `@parameter` ones please.

```suggestion
    @always_inline
    def call_fn() raises {mut f}:
        for _ in range(1000):
            var r = f(arg_a, arg_b)
            keep(r)

    b.iter(call_fn)
```

mojo/stdlib/benchmarks/python/bench_python_ffi_typed.mojo
```diff
@@ -0,0 +1,108 @@
...
+    keep(mod)
+    keep(arg_a)
+    keep(arg_b)
+    keep(f)
```
NathanSWard, line 76: For benchmarks, please don't use `keep` as a way to silence warnings - `keep` should be used to explicitly prevent the compiler from optimizing out values **within** the benchmarked function.
If we need to silence "unused" warnings, use the `_ = value^` pattern instead.

Lastly, if we update this to use unified closures, I think we can remove these entirely.

mojo/stdlib/std/python/bindings.mojo
```diff
@@ -465,6 +471,55 @@ struct PythonModuleBuilder:
...
+    # Typed-argument fast-path API
+    # ===-------------------------------------------------------------------===#
+    # Users who declare their function with concrete Mojo types like
+    # `def f(a: Int, b: Int) -> Int` can register via `def_typed_function`
+    # to get a wrapper that converts via `PyLong_AsSsize_t` /
+    # `PyLong_FromSsize_t` directly, bypassing the `PythonObject` wrap
+    # and the `__int__()` -> `PyNumber_Long` allocation chain.
+
+    def def_typed_function[
+        func: _FnIntToInt
```
NathanSWard, line 484: These fast path optimizations feel very odd in the sense they only work with methods that are `fn(Int) -> Int` or `fn(Int, Int) -> Int`. How often are people defining functions that match this API 1-for-1 where this fast path will actually benefit them?

This feels like a bit too much "special-casing" where ideally we should improve the perf of _all_ python method calls via fixing the root `PythonObject`.

mojo/stdlib/std/python/_python_func_typed.mojo
NathanSWard: +1

msaelices: I’m probably not experienced enough yet with the new reflection logic in Mojo, but I will try to find a more generic solution. If you could give me any hints or guidance on it, I would really appreciate it.

msaelices: Also I will try to use variadics.

mojo/stdlib/std/python/bindings.mojo
msaelices: Sure, this was a first step to get your feedback. But I did it on purpose to ensure it could be done, as I am not an expert in python to mojo interoperability. I will work on making it more generic

martinvuyk: I'm not sure we can even express them generically in the type system yet, so don't bang your head on it for too long

msaelices:
Superseded by #6530. @NathanSWard's review pushed the right direction, instead of adding a per-type opt-in API, optimize `Int.__init__(py=...)` directly so every existing `def_function` binding benefits transparently. The new PR adds a single `PyLong_CheckExact` fast path that takes `Int(py=python_int)` from ~88 ns to ~4 ns (~22x faster), without introducing any new public surface. Thanks for the careful review @NathanSWard @martinvuyk

PR #6539, opened 2026-05-13
[stdlib] Fix base64 decode to return List[Byte]

### LINKED ISSUE #6535: [BUG] `b64decode` and `b16decode` should return `List[Byte]` instead of `String`
### Actual behavior
The decoding functions `b64decode` and `b16decode` in https://github.com/modular/modular/blob/main/mojo/stdlib/std/base64/base64.mojo return a `String`.

### Expected behavior
They should return `List[Byte]`. Not all binary strings correspond to valid `String` representations, so `List[Byte]` is the best, most general output format.

I have an old PR that implements this (and it's not rocket science), but I want to get buy-in first before rebasing.

### Steps to reproduce

``` mojo
def b64decode[
    *, validate: Bool = False
](str: StringSlice[mut=False, _]) raises -> String:
```

``` mojo
def b16decode(str: StringSlice[mut=False, _]) -> String:
```

Should probably be something like

```mojo
def b64decode[
    *, validate: Bool = False
](str: StringSlice[mut=False, _]) raises -> List[Byte]:
```

``` mojo
def b16decode(str: StringSlice[mut=False, _]) -> List[Byte]:
```

### System information

All versions affected.

mahendrarathore1742:
### Description

- Update b64decode and b16decode to return List[Byte] instead of String.
- Adjust base64 tests to validate byte outputs.
- Preserve decoding behavior while supporting non‑UTF8 binary data.

close #6535

### DIFF
--- a/mojo/stdlib/std/base64/base64.mojo
+++ b/mojo/stdlib/std/base64/base64.mojo
@@ -223,6 +223,9 @@ def b16decode(str: StringSlice[mut=False, _]) raises -> List[Byte]:
 
     Returns:
         The decoded bytes.
+
+    Raises:
+        If the input length is not divisible by 2.
     """
 
     comptime `A` = Byte(ord("A"))
--- a/mojo/stdlib/std/base64/base64.mojo
+++ b/mojo/stdlib/std/base64/base64.mojo
@@ -147,19 +147,13 @@ def b64decode[
     var data = str.as_bytes()
     var n = str.byte_length()
 
-    if n % 4 != 0:
-        raise Error("ValueError: Input length '", n, "' must be divisible by 4")
+    comptime if validate:
+        if n % 4 != 0:
+            raise Error(
+                "ValueError: Input length '", n, "' must be divisible by 4"
+            )
 
-    if n == 0:
-        return List[Byte]()
-
-    var output_len = (n // 4) * 3
-    if data[n - 1] == `=`:
-        output_len -= 1
-    if data[n - 2] == `=`:
-        output_len -= 1
-
-    var result = List[Byte](capacity=output_len)
+    var result = List[Byte](capacity=n)
 
     # This algorithm is based on https://arxiv.org/abs/1704.00605
     for i in range(0, n, 4):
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/base64/base64.mojo
```diff
@@ -147,13 +147,19 @@ def b64decode[
     var data = str.as_bytes()
     var n = str.byte_length()
 
-    comptime if validate:
-        if n % 4 != 0:
```
NathanSWard, line 151: I believe should should likely leve in the `comptime if validate` branch here.

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -241,13 +247,14 @@ def b16decode(str: StringSlice[mut=False, _]) -> String:
 
     var data = str.as_bytes()
     var n = str.byte_length()
-    debug_assert(n % 2 == 0, "Input length '", n, "' must be divisible by 2")
+    if n % 2 != 0:
+        raise Error("ValueError: Input length '", n, "' must be divisible by 2")
```
NathanSWard: **question** Also here, any reason this was update to a `rase` instead of the current `debug_assert` approach?

mojo/stdlib/std/base64/base64.mojo
mahendrarathore1742: Thanks! I moved the check out of the comptime if validate block to prevent out‑of‑bounds reads when ``` n % 4 != 0 ``` in non‑validate mode. If you want validation to be gated, I can instead guard the loop for malformed input and keep the strict raise only when ``` validate=True ```

mojo/stdlib/std/base64/base64.mojo
mahendrarathore1742: I switched to a runtime raise because debug_assert can be compiled out, which could lead to OOB access on odd‑length inputs. If you prefer the original behavior, I can keep debug_assert and add a safe guard/early return to avoid undefined behavior in release builds.

NathanSWard: For purposes of this PR - let's keep the change-set minimal and only have the change be updating the return type to be `List[Byte]`. We can fix the `debug_assert`/`validation` logic in a separate issue :)

mahendrarathore1742: Thanks for the guidance! I’ll keep this PR minimal and limit it to the return type change (String -> List[Byte]) and required test updates. I’ll revert the validation/debug_assert adjustments and handle them in a separate issue if needed.

NathanSWard (approved):
thank you!

The one thing we forgot to do was add a `changelog` entry for this - but I'll go ahead and just add it internally :)

NathanSWard:
!sync

Issue #6546, opened 2026-05-15
[Feature Request] Re-enable `ConvertibleToPython` conformance for builtin types


christoph-schlumpf:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Re-enabling `ConvertibleToPython` conformance for builtin types allows them to be used in generic code depending on `ConvertibleToPython` conformance.

### What is your motivation for this change?

This was discussed on Discord with Chris and others.
https://discord.com/channels/1087530497313357884/1503501113217781810 
> It is appealing to me because it is cleaner and less arbitrary, but it does mix more python stuff into standard files. It is probably the cleanest solution though.
> All this stuff will be much cleaner with extensions! 🙂
> In any case, I support # 2 if you think this would help.  Would you be willing to put up a PR?

Changes:
- Remove implicit init methods on PythonObject for the builtin types (`Bool`, `Int`, `Skalar`, `StringLiteral`, `StringSlice` and `String`).
- Add trait `ConvertibleToPython` to those types. This still allows for implicit conversion to `PythonObject`.
- Add unit tests to check `ConvertibleToPython` conformance and implicit conversion
- Aligns builtin types with custom types as both use the same implicit `PythonObject` init method.

As discussed with Chris I will open a PR for this change.

### Any other details?

This is related to [[mojo-stdlib] 3/x: Continue simplifying PythonObject implementation · modular/modular@5a23c2a](https://github.com/modular/modular/commit/5a23c2a12288b055561e8e01f7c58cbc5630c9f9) where conformance to `ConvertibleToPython` was removed from builtin types to enable implicit conversion to `PythonObject` for custom types conforming to `ConvertibleToPython`.

lattner:
I'm very +1 for this change, thank you for working on this Christoph!

PR #6547, opened 2026-05-15
[mojo][stdlib] Re-enable `ConvertibleToPython` conformance for builtin types

christoph-schlumpf:
## Linked issue

Relates to https://github.com/modular/modular/issues/6546

## Type of change

- [ ] Documentation update
- [x] (Re-enable) feature or public API (requires prior proposal or issue approval)
- [x] Refactor / internal cleanup (no user-visible change)

## Motivation

See https://github.com/modular/modular/issues/6546

## What changed

- Removed implicit init methods on PythonObject for the builtin types (`Bool`, `Int`, `Skalar`, `StringLiteral`, `StringSlice` and `String`).
- Added trait ConvertibleToPython to those types. This still allows for implicit conversion to PythonObject. 

## Testing

- Added unit tests to check `ConvertibleToPython` conformance and implicit conversion.
- Added two helper functions to `test-utils`.

## Checklist

      approval


lattner (approved):
This patch looks really great, thank you for cleaning this up for me @christoph-schlumpf !

christoph-schlumpf:
Some tests are failing at the moment because some former implicit `PythonObject` init functions were non-raising. I think converting something to a PythonObject should always be raising because there is a py environment involved. I will update the tests accordingly.

ConnorGray:
I also noticed the failing tests, but I'm less certain whether making these always raise the best solution. It's true that operations involving Python can fail due a Python environment being involved, but generally that failure occurs when we try do dynamically load the CPython `libpython.dylib`, and consequently will abort if it fails.

I wonder if an alternative model would be to add an associated error type to `ConvertibleToPython`, that can be defined as `Never` in the case of "trivial" types that ought never to fail conversion, e.g.

```
trait ConvertibleToPython:
    comptime PythonConversionError: Movable

    def to_python_object(self) raises Self.PythonConversionError -> PythonObject:
        ...

struct Bool(ConvertibleToPython):
    comptime PythonConversionError = Never # Never inhabited
    
    ...
```

I believe the compiler treats `-> raises Never` as not needing to be wrapped in a `try` block, so we avoid "polutting" callers with `try` blocks that are never used in practice.

Does that seem like a direction worth experimenting with @christoph-schlumpf? 🙂

christoph-schlumpf:
> I wonder if an alternative model would be to add an associated error type to `ConvertibleToPython`, that can be defined as `Never` in the case of "trivial" types that ought never to fail conversion, e.g.
> ...
> Does that seem like a direction worth experimenting with @christoph-schlumpf? 🙂

That's a great idea and I will experiment with it. This would allow for more flexibility, making the easy conversions easy to implement and use while allowing generic and more complex cases to raise an error at runtime if needed.

christoph-schlumpf:
I updated this PR with a parametric raising version of `ConvertibleToPython`.

It worked out very well. `Bool`, `Int` and `Scalar` types can now be converted to a `PythonObject` in non-raising functions and without `try` blocks while still allowing other types like `StringSlice` to require rising conversions.

This PR will open the door for other types to be (conditional) conformant to `ConvertibleToPython`.

Have a look, @ConnorGray. 🙂

christoph-schlumpf:
How to proceed with this PR? It is not assigned nor synced.

christoph-schlumpf:
Hi @ConnorGray ,

I've updated this PR to the latest nightly. It is (again) ready to be merged without conflicts.

Best regards,
Christoph

PR #6552, opened 2026-05-15
[stdlib] Add codepoint slicing

martinvuyk:
Add codepoint slicing

### DIFF
--- a/mojo/stdlib/std/collections/string/string_slice.mojo
+++ b/mojo/stdlib/std/collections/string/string_slice.mojo
@@ -481,17 +481,7 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
         )
         return Self(unsafe_from_utf8=self._slice[byte])
 
-    def __getitem__(self, *, codepoint: Some[Indexer]) -> Self:
-        """Gets the character at the specified position.
-
-        Args:
-            codepoint: The codepoint index.
-
-        Returns:
-            A `StringSlice` view containing the unicode codepoint at the
-            specified position.
-        """
-
+    def _get_codepoint(self, codepoint: Some[Indexer]) -> Self:
         var c_idx = index(codepoint)
         debug_assert[assert_mode="safe"](
             c_idx >= 0, "negative indexing is not supported"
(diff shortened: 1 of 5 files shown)


mojo/stdlib/std/collections/string/string_slice.mojo
```diff
@@ -522,6 +522,38 @@ struct StringSlice[mut: Bool, //, origin: Origin[mut=mut]](
...
+        Returns:
+            A new StringSlice containing the codepoints in the specified range.
+        """
+
+        var res: Self
+        if codepoint.start:
+            var cp = self[codepoint=codepoint.start.value()]
+            var cp_ptr = cp.unsafe_ptr()
+            var offset = Int(cp_ptr) - Int(self.unsafe_ptr())
+            res = {ptr = cp_ptr, length = self.byte_length() - offset}
```
barcharcraz: for the other slice subscripts we snap to a zero-sized range at the end of the collection if the starting index is off the end. Here we abort. We should try and match the existing behavior of slice.indices.

(we might want to change the "off the end" behavior at some point, but we should try and have the same behavior, or at least document the difference)

mojo/stdlib/test/collections/string/test_string_slice.mojo
```diff
@@ -1105,5 +1105,38 @@ def test_codepoint_indexing() raises:
     assert_equal(StringSlice("🔄🔥🔄")[codepoint=2], "🔄")
 
 
+def test_codepoint_slicing() raises:
```
barcharcraz, line 1108: test the "off the end" and "start>end" behaviors.

I think the current behavior is to snap "off the end" and "off the start" values to the end/start and to create a negative length view for the "reversed". 

For this PR I don't care so much about "start>end" because I think the current behavior of .indices and `span.__getitem__` are somewhat accidental in that case.

barcharcraz (changes requested):
Either make the "off the end" slicing behavior consistent with everything else or document the different behavior here.

mojo/stdlib/std/collections/string/string_slice.mojo
martinvuyk: The only two options I see to match slice.indices behavior would be to either:
- add an argument like `self[codepoint=codepoint.start.value(), empty_on_oob=True]` and add a branch to the codepoint indexer that returns a `StringSlice` of length 0 pointing at the end
  - we could also make returning an empty string on oob be the default
- use `slice.indices(self.count_codepoints())` which requires an extra pass through the string data

mojo/stdlib/test/collections/string/test_string_slice.mojo
martinvuyk: > I think the current behavior is to snap "off the end" and "off the start" values to the end/start and to create a negative length view for the "reversed".

why would we allow start > end ? why would a negative length `StringSlice` be valid? I've never read anything or had any discussion that indicates that is accepted in the stdlib (it is also unsafe outside of ASCII)

barcharcraz: I think it's accepted in python and we carried that behavior over. Again I don't care that much about that edge case, you can address only the "off the ends" one

mojo/stdlib/std/collections/string/string_slice.mojo
barcharcraz: yeah, you'd pull out the codepoint indexing into some private method and call it from both subscript methods. For the scalar subscripting you'd assert that it wasn't oob.

That internal method can just return a bool to indicate if it was oob or not, but the returned slice should probably have the same one-off-the-end pointer as you get for other collections.

martinvuyk: sounds like a good plan, I'll implement it

barcharcraz:
This also needs a changelog entry

barcharcraz (approved):
lgtm now

barcharcraz:
!sync

PR #6553, opened 2026-05-15
[stdlib] Add grapheme indexing

martinvuyk:
Add grapheme indexing

Remove the `def nth_grapheme(self, n: Int) -> Optional[Self]:` API from `String` and `StringSlice` and replace it with indexing similar to what we have for codepoints and bytes. We need to keep these APIs consistent. Users should use `self.graphemes.nth()` if they want to use an `Optional`-based API

### DIFF
--- a/mojo/stdlib/scripts/run-tests.sh
+++ b/mojo/stdlib/scripts/run-tests.sh
@@ -22,8 +22,7 @@ FILTER=""
 if [[ $# -gt 0 ]]; then
   FILTER="${1#./}" # remove leading relative file path if it has one
   if [[ -f ${FILTER} ]]; then # specific file
-    FILTER="//mojo/$(dirname "$FILTER"):$(basename "$FILTER")"
-    FILTER=$("$REPO_ROOT"/bazelw query "$FILTER")
+    FILTER="//mojo/$(dirname "$FILTER"):$(basename "$FILTER").test"
   else
     # specific test directory
     FILTER="${FILTER%/}" # remove trailing / if it has one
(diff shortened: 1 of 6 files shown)


mojo/stdlib/scripts/run-tests.sh
```diff
@@ -22,8 +22,7 @@ FILTER=""
 if [[ $# -gt 0 ]]; then
   FILTER="${1#./}" # remove leading relative file path if it has one
   if [[ -f ${FILTER} ]]; then # specific file
-    FILTER="//mojo/$(dirname "$FILTER"):$(basename "$FILTER")"
-    FILTER=$("$REPO_ROOT"/bazelw query "$FILTER")
+    FILTER="//mojo/$(dirname "$FILTER"):$(basename "$FILTER").test"
```
barcharcraz, line 25: this appears to be a stray diff

barcharcraz (changes requested):
Looks good except it appears you edited `mojo/stdlib/scripts/run-tests.sh`

mojo/stdlib/scripts/run-tests.sh
martinvuyk: it is intentional, it fixes that code and allows one to run a single test file directly without testing the whole directory

barcharcraz: we'd prefer a separate PR for stuff like this in the future, but it's OK this time

barcharcraz (approved):

barcharcraz:
!sync

barcharcraz:
This does need a changelog entry, but I can add one before landing

martinvuyk:
@barcharcraz I added a changelog entry

barcharcraz:
!sync

barcharcraz:
Thanks Martin. merged internally and should show up in the public GitHub tomorrow.

PR #6576, opened 2026-05-20
[Stdlib] Fast-path `String(py=...)` for exact Python `str` inputs

msaelices:
## Context

`String.__init__(py=...)` previously called `py.__str__()` for every input, paying for a `PyObject_Str` dispatch and a temporary `PythonObject` lifecycle even when the input was already a Python `str`.

For exact `PyUnicode` inputs, read the cached UTF-8 buffer directly via `PyUnicode_AsUTF8AndSize` on the borrowed pointer and skip the `__str__()` round trip.

## Changes

- `_cpython.mojo`: add `_PyUnicode_Type` + `PyUnicode_CheckExact`, mirroring the `_PyLong_Type` / `PyLong_CheckExact` pattern.
- `string.mojo`: gate on `PyUnicode_CheckExact`; fall through for subclasses with overridden `__str__` and non-`str` objects so CPython `str(x)` semantics are preserved.
- `bench_pyobject_convert.mojo`: primitive microbench (`bench_string_from_pystr`) so reviewers can reproduce the table.
- `bench_bindings/`: new `str_byte_len_def(s) -> PythonObject(String(py=s).byte_length())` exercising the input-side conversion end-to-end via the existing `def_function` harness, plus `str_byte_len_raw_fastcall` lower bound and `py_str_byte_len` baseline.

### Microbench (`bench_pyobject_convert.mojo`)

Local harness, single core, 5 repetitions of 100 iterations × 1000 conversions per iteration. Min-of-min per side.

| Bench | Before (slow path via `__str__()`) | After (fast path) | Δ |
|---|---|---|---|
| `bench_string_from_pystr` | **~103.6 ns** | **~34.3 ns** | **−69 ns, ~3× faster** |

### End-to-end (`bench_bindings/`)

`bazelw test //mojo/stdlib/benchmarks/python/bench_bindings:bench_bindings -c opt`, 2M iter × 7 repeats via `timeit.repeat`, min of repeats. Input `"hello world"`.

| Variant | Before | After | Δ |
|---|---|---|---|
| `str_byte_len_def` (`def_function`) | ~115.5 ns | **~85.2 ns** | **−30 ns (~26% faster)** |
| `str_byte_len_raw_fastcall` (lower bound) | ~24.9 ns | ~27.3 ns | within noise (unchanged path) |
| `py_str_byte_len(s)` (pure Python) | ~57 ns | ~50 ns | reference |

The microbench's ~69 ns primitive win is partly absorbed by the surrounding FFI dispatch, but ~30 ns still passes through to the user-visible call.

Assisted-by: AI


msaelices:
@JoeLoser or @NathanSWard could you please see if this is a good addition for optimizing FFI?

PR #6578, opened 2026-05-20
[Stdlib] Fast-path `PythonObject(value: Bool)` via cached `Py_True` / `Py_False` singletons

msaelices:
## Context

`PythonObject(value: Bool)` previously went through `cpy.PyBool_FromLong(...)`, which CPython implements as `branch + Py_INCREF` internally; calling it from Mojo pays one full `ExternalFunction` dispatch on top.

The cached `Py_True` / `Py_False` singletons are already available in CPython at well-known symbols (`Py_GetConstantBorrowed` on 3.13+, `_Py_TrueStruct` / `_Py_FalseStruct` on older builds). Reading them directly and doing the `Py_IncRef` inline saves one C-call dispatch per `Bool -> PythonObject` conversion.

## Changes

- `_cpython.mojo`: load `_Py_TrueStruct` / `_Py_FalseStruct` at `CPython.__init__` (mirrors the existing `Py_None` pattern); expose `Py_True()` / `Py_False()` accessors; expose named `Py_CONSTANT_{NONE,FALSE,TRUE}` constants near the `Py_GetConstantBorrowed` declaration. Version gate uses `major > 3 or (major == 3 and minor >= 13)` so a future Python 4.0 keeps using `Py_GetConstantBorrowed`.
- `python_object.mojo`: `PythonObject(value: Bool)` and the `Scalar[DType.bool]` overload's `DType.bool` branch both delegate to a private `_bool_to_pyobject(value)` helper that reads the singleton + inlines `Py_IncRef` against the in-scope `cpy` ref.
- `test_python_cpython.mojo`: asserts `Py_True()` / `Py_False()` return the same pointers as `PyBool_FromLong(1)` / `(0)`.
- `test_python_to_mojo.mojo`: extends `test_bool` with `Scalar[DType.bool]` round-trip coverage.
- `bench_bool_to_pyobject.mojo`: primitive microbench, 20 repetitions, covers both `Bool` and `Scalar[DType.bool]` overloads.
- `bench_bindings/`: new `bool_pass_def(b) -> PythonObject(Bool(py=b))` exercising the round-trip end-to-end via the existing `def_function` harness, plus `bool_pass_raw_fastcall` lower bound and `py_bool_pass` baseline.

### Microbench (`bench_bool_to_pyobject.mojo`)

`bazelw test //mojo/stdlib/benchmarks:python/bench_bool_to_pyobject.mojo.bench -c opt`, Mojo `1.0.0b2`, single core, 20 reps × 100 iters × 1000 conversions per iter. Min-of-min.

| Variant | min ns/conv |
|---|---|
| `PyBool_FromLong` (before) | ~34.8 ns |
| `Py_True` / `Py_False` + inline `Py_IncRef` (after) | **~30.0 ns** |

~5 ns / ~14% saved per `Bool` returned to Python.

### End-to-end (`bench_bindings/`)

`bazelw test //mojo/stdlib/benchmarks/python/bench_bindings:bench_bindings -c opt`, 2M iter × 7 repeats via `timeit.repeat`, min of repeats. ASCII input `True`.

| Variant | Before | After | Δ |
|---|---|---|---|
| `bool_pass_def` (`def_function`) | ~72.4 ns | **~53.4 ns** | **−19 ns (~26%, ~12 ns above noise floor)** |
| `bool_pass_raw_fastcall` (lower bound) | ~14-21 ns (run-to-run drift) | ~14-21 ns | unchanged baseline |
| `py_bool_pass(b)` (pure Python) | ~28-30 ns | ~28-30 ns | reference |

The microbench's ~5 ns delta composes through the FFI to a clearly measurable E2E gain on `bool_pass_def`. Noise floor on this machine is ~7 ns (estimated from raw-fastcall drift across runs); the gain sits above it.

Assisted-by: AI


mojo/stdlib/std/python/python_object.mojo
```diff
@@ -225,8 +225,13 @@ struct PythonObject(
...
             value: The boolean value.
         """
+        # Pick the cached singleton directly and INCREF. Skips the
+        # `c_long(Int(value))` cast chain that `PyBool_FromLong` requires
+        # and lets the branch + field reads inline at the call site.
         ref cpy = Python().cpython()
-        self = Self(from_owned=cpy.PyBool_FromLong(c_long(Int(value))))
+        var ptr = cpy.Py_True() if value else cpy.Py_False()
+        cpy.Py_IncRef(ptr)
+        self = Self(from_owned=ptr)
```
JoeLoser: **Suggestion** This snippet is the same as what `Self(from_borrowed=ptr)` does on lines 131-151 — read borrowed pointer, `Py_IncRef`, store. Perhaps we should just write this as:

```mojo
ref cpy = Python().cpython()
self = Self(from_borrowed=cpy.Py_True() if value else cpy.Py_False())
```

Should be the same codegen (the `from_borrowed` ctor will inline), but the refcount protocol stays encapsulated in one constructor instead of being re-spelled at every singleton-fast-path call site. As more fast paths land (Py_None, small-int cache, etc.) this is the pattern we'll want to standardize on.

If you have benchmark numbers showing `from_borrowed` is measurably slower here, please document the delta in the comment above so future maintainers don't try to "clean this up" and regress perf.

mojo/stdlib/std/python/python_object.mojo
```diff
@@ -225,8 +225,13 @@ struct PythonObject(
         Args:
             value: The boolean value.
         """
+        # Pick the cached singleton directly and INCREF. Skips the
+        # `c_long(Int(value))` cast chain that `PyBool_FromLong` requires
+        # and lets the branch + field reads inline at the call site.
```
JoeLoser: **Quibble**: this comment credits the savings to skipping the `c_long(Int(value))` cast chain, but those casts should compile to ~nothing on the target architecture. The PR description correctly identifies the real win — skipping the `ExternalFunction` dispatch through `PyBool_FromLong`. Perhaps we should align the in-code comment with the PR description so future readers don't optimize against the wrong cost model.

mojo/stdlib/std/python/python_object.mojo
```diff
@@ -253,8 +258,9 @@ struct PythonObject(
         ref cpy = Python().cpython()
 
         comptime if dtype == DType.bool:
-            var val = c_long(Int(value))
-            self = Self(from_owned=cpy.PyBool_FromLong(val))
+            var ptr = cpy.Py_True() if Bool(value) else cpy.Py_False()
+            cpy.Py_IncRef(ptr)
+            self = Self(from_owned=ptr)
```
JoeLoser: Same `from_borrowed` suggestion as above — and this branch is now nearly byte-identical to the `Bool` overload. Consider factoring into a private helper:

```mojo
@always_inline
fn _bool_to_pyobject(cpy: CPython, value: Bool) -> PythonObject:
    return PythonObject(from_borrowed=cpy.Py_True() if value else cpy.Py_False())
```

Then use this with both the `Bool` and `Scalar[DType.bool]` initializers.

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -1634,6 +1636,26 @@ struct CPython(Defaultable, Movable):
         self._PyBool_Type = self.lib.get_symbol[PyTypeObject](
             "PyBool_Type"
         ).value()
+        if self.version.minor >= 13:
+            # Py_GetConstantBorrowed is part of the Stable ABI since version 3.13.
+            # Py_CONSTANT_FALSE = 1, Py_CONSTANT_TRUE = 2.
+            self._Py_False = self.lib.call[
+                "Py_GetConstantBorrowed", PyObjectPtr
+            ](1)
```
JoeLoser: The comment names these (`Py_CONSTANT_FALSE = 1`, `Py_CONSTANT_TRUE = 2`) but the call sites still pass raw `1` / `2`. Not blocking, but consider adding named aliases near the `Py_GetConstantBorrowed` declaration:

```mojo
comptime Py_CONSTANT_NONE: c_uint = 0
comptime Py_CONSTANT_FALSE: c_uint = 1
comptime Py_CONSTANT_TRUE: c_uint = 2
```

mojo/stdlib/benchmarks/python/bench_bool_to_pyobject.mojo
```diff
@@ -0,0 +1,55 @@
...
+            for _ in range(1000):
+                var x = PythonObject(black_box(v))
+                keep(x)
+        except e:
+            abort(String(e))
+
+    b.iter(call_fn)
+
+
+def main() raises:
```
JoeLoser, line 71: Two suggestions for the benchmark:

1. Add a second `bench_function` for the `Scalar[DType.bool]` overload — it's a separate code path (different overload selection + the extra `Bool(value)` cast).
2. At ~5 ns deltas, min-of-min from 5 reps × 100 iters is borderline. Consider bumping `num_repetitions` to 20+ for tighter confidence intervals on the comparison numbers.

Also curious to get your thoughts on this style of benchmarking vs the full FFI ones I added in https://github.com/modular/modular/tree/main/mojo/stdlib/benchmarks/python/bench_bindings.

mojo/stdlib/std/python/python_object.mojo
msaelices: Tried `Self(from_borrowed=...)` first — it ended up slower in the bench. `PythonObject.__init__(*, from_borrowed:)` runs its own `Python().cpython()` lookup before calling `Py_IncRef`, so the body becomes two `cpython()` lookups per conversion. Measured `-c opt`, min-of-min of 20×100×1000 conversions:

| | min ns/conv |
|---|---|
| `PyBool_FromLong` (old) | ~34.8 |
| `from_borrowed=...` | ~33.4 |
| Inline `Py_IncRef` + `from_owned` (this PR) | ~30.0 |

So the explicit `Py_IncRef` is the optimization as it lets us reuse the `cpy` ref already in scope. Created a new reusable `_bool_to_pyobject` in [70e3c813b5]

mojo/stdlib/std/python/_cpython.mojo
```diff
@@ -1634,6 +1636,26 @@ struct CPython(Defaultable, Movable):
...
+            self._Py_True = PyObjectPtr(
+                upcast_from=self.lib.get_symbol[PyObject](
+                    "_Py_TrueStruct"
+                ).value()
+            )
+            self._Py_False = PyObjectPtr(
+                upcast_from=self.lib.get_symbol[PyObject](
+                    "_Py_FalseStruct"
+                ).value()
+            )
```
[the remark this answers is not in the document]
msaelices: The `_Py_*Struct` symbol fallback here mirrors the existing `_Py_NoneStruct` loader a few lines above. Same export-availability risk profile, if `_Py_NoneStruct` is missing on a given build, `Python().cpython()` would have already failed to initialize before we even get to the Bool block.

mojo/stdlib/std/python/_cpython.mojo
msaelices: Thanks. Fixed: https://github.com/modular/modular/pull/6578/changes/028039d13c3208deb9df0c4d6d62360c17fc6376

mojo/stdlib/benchmarks/python/bench_bool_to_pyobject.mojo
msaelices: Fixed items 1 and 2 here: https://github.com/modular/modular/pull/6578/changes/184baa8b769c944ae12ef3299bc63e2d28082f8b

msaelices: My take on benchmarks approach: for the kind of work we have been doing (fast paths on individual conversion primitives), the microbenches are the right primary signal.  They let us see a ~5 ns win or regression on `Bool` to `PythonObject` that would be lost in the noise of an E2E `noop_def` measurement and they are quicker for iterations, checking optimizations hypothesis, etc.

But bench_bindings/ is the ground truth for "does this actually help users?" because that's what users feel. Here, the composition story matters: 25 ns saved in a microbench might be 5 ns saved E2E, or even 0, if `METH_FASTCALL` work already dominated the overhead.

One idea is to have both kinds:

1. Microbench drives implementation iteration (what we did).
2. Also drop a bench_bindings/ style to confirm the win composes through the FFI, even if it's noisier.

Will add a bench_bindings/ one in this PR to explore it.

mojo/stdlib/std/python/python_object.mojo
msaelices: done. thanks

msaelices:
@JoeLoser could you PTAL?

mojo/stdlib/std/python/python_object.mojo
msaelices: Addressed in 70e3c813b5, extracting the body into a private `_bool_to_pyobject(value)` helper

msaelices:
@JoeLoser is there something more I could do to improve this PR

PR #6585, opened 2026-05-22
[stdlib] Add homogeneous tuple iteration

martinvuyk:
Add homogeneous tuple iteration.

This needs `InlineArray` to accept `Movable` elements


NathanSWard:
I'm a bit -1 on adding this. This feel very niche and I'm not sure why you wouldn't just use an `InlineArray` to achieve the same thing.
@martinvuyk - can you please move this to a GH issue instead of a PR so we can have a discussion there about it (RE the new stdlib contribution policy)? I'm going to close this PR in the mean time.

Issue #6570, opened 2026-05-19
[Feature Request] [Epic] Extend support for NVIDIA sm_120 / sm_121 consumer Blackwell GPUs


BradLarson:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

This is an umbrella tracking issue for sub-issues around expanding platform support for NVIDIA's `sm_120` / `sm_121` consumer Blackwell GPUs, as found in the RTX 50XX series or DGX Spark.

### What is your motivation for this change?

NVIDIA's Blackwell GPU platform is divided into two very different architectures: the `sm_100` / `sm_110` series (B200 / B300, Jetson Thor), and the `sm_120` / `sm_121` series (RTX 50XX, DGX Spark). The former is very well-supported and optimized in Mojo and MAX today, but due to major architectural differences the latter has a few areas still to be filled in and tuned.

Help is appreciated from those with access to `sm_120` / `sm_121` hardware to find and fix issues with missing paths in our kernels for those GPUs, as well as to help build and tune optimized kernels for them.

superpan:
Filed #6961: on DGX Spark (GB10, sm_121a), `max serve` crashes during device graph capture warmup because the VMM defragmenting allocator's `preBackForCapture` path calls `cuMemCreate` with `allocFlags.gpuDirectRDMACapable = 1`, which fails with `CUDA_ERROR_INVALID_DEVICE` since `CU_DEVICE_ATTRIBUTE_GPU_DIRECT_RDMA_WITH_CUDA_VMM_SUPPORTED = 0` on this integrated GPU. Plain VMM allocations work, and graph capture works with `MODULAR_DEVICE_CONTEXT_MEMORY_MANAGER_VMM=0`, so gating the flag on that device attribute should fix it. Full isolation matrix and a dependency-free ctypes repro in the issue; happy to test candidate fixes on this hardware.

PR #6590, opened 2026-05-22
[kernels] Fix topk sampling LAUNCH_OUT_OF_RESOURCES on Blackwell (sm_120 & sm_121)

### LINKED ISSUE #6488: Gemma-4-31B: CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES on DGX Spark GB10 (SM 12.1) — topk sampling kernel
## Environment

* **System**: NVIDIA DGX Spark (AI TOP ATOM variant)
* **CPU**: ARM64 (Grace), 20 cores
* **GPU**: NVIDIA GB10, SM 12.1a
* **Memory**: 128GB unified (shared CPU/GPU)
* **MAX Version**: 26.4.0.dev2026050306 (nightly, May 3 2026)
* **Model**: google/gemma-4-31b-it (local, NVMe)
* **OS**: Ubuntu, kernel 6.x, CUDA 13.0

## Summary

MAX Engine 26.4.0 nightly successfully **compiles and loads** Gemma-4-31B-IT on the DGX Spark GB10, but crashes on the **first inference request** due to a CUDA kernel resource limit violation in the top-K/top-P sampling kernel.

**Greedy sampling (temperature=0) works perfectly.** Only the fused top-K/top-P sampling kernel fails.

## What Works

| Step | Status | Time |
| -- | -- | -- |
| Model detection (Gemma4ForConditionalGeneration) | ✅ | instant |
| Language model graph build | ✅ | 3s |
| Language model kernel compilation (SM 12.1) | ✅ | 285s |
| Vision model compilation | ✅ | 5s |
| Weight loading | ✅ | \~30s |
| Device graph capture (batch 1..4) | ✅ | 36s |
| Server startup (OpenAI API) | ✅ | instant |
| `/v1/models` endpoint | ✅ | works |
| **Greedy sampling (temperature=0)** | ✅ | **works** |
| **Top-K/Top-P sampling (temperature>0)** | ❌ | **crashes** |

## Error

```
ERROR: Worker exception, shutting down: An error occurred in kernel entry point named "region_15":
An error occurred in kernel named "stub_386":
max/kernels/src/nn/topk_fi.mojo:1419:49 failed calling 
'topk_topp_sampling_from_prob_float32_int64_False_ddb3285f' 
on device cuda:0 with error 'CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES 
(too many resources requested for launch)'
```

The fused token sampling kernel requests more resources (registers/shared memory/threads per block) than SM 12.1 supports.

## Workaround

**Greedy sampling works.** Setting `temperature: 0` bypasses the broken kernel entirely:

```bash
curl http://localhost:8091/v1/chat/completions \
  -d '{"model":"...","messages":[{"role":"user","content":"Hello"}],"temperature":0,"max_tokens":100}'
# Returns valid response
```

### 1\. Memory Estimation Bug (unified memory)

`MemoryEstimator.free_memory()` reads CUDA-reported GPU memory (\~34GB) instead of system available memory (\~116GB). Models >34GB are rejected even though the unified memory pool is 128GB.

**Workaround**: Monkey-patch `MemoryEstimator.free_memory` to read `/proc/meminfo` on unified memory systems (DGX Spark, Grace Blackwell).

### 2\. Pydantic Deferred Type Annotations

`PipelineConfig` and `MAXModelConfig` have deferred type annotations referencing `torch.Tensor`. When the CLI creates these configs, torch hasn't been imported, causing `PydanticUserError`.

**Workaround**: Import torch and call `model_rebuild()` on both models before the CLI runs.

### 3\. CUDA Memory Overallocation (OOM)

Without constraints, CUDA allocates the entire unified memory pool during model loading + graph capture, OOMing the system.

**Workaround**: Use a CUDA memory limiter shim (LD_PRELOAD interception of cuMemAlloc) with a 96GB cap to leave headroom for the OS.

### 4\. Missing conda Dependencies

The `max` conda package doesn't declare many required Python dependencies. Had to manually add: tqdm, pydantic, pillow, transformers, safetensors, tokenizers, sentencepiece, accelerate, msgspec, opentelemetry-\*, pyzmq, llguidance, av, prometheus_client, python-json-logger, aiofiles, asgiref, openai, sse-starlette, pytorch.

## Reproduction

```bash
# On DGX Spark with Gemma-4-31B local
pixi init --channel https://conda.modular.com/max-nightly
pixi add max pytorch tqdm pydantic pillow transformers safetensors \
  tokenizers sentencepiece accelerate msgspec opentelemetry-api \
  opentelemetry-sdk opentelemetry-exporter-otlp-proto-http pyzmq \
  llguidance av prometheus_client python-json-logger aiofiles \
  asgiref openai sse-starlette
pixi add --pypi 'opentelemetry-exporter-prometheus==0.62b1'

# Apply workarounds (unified memory + pydantic + CUDA limiter)
# Then:
max serve --model-path /path/to/gemma-4-31b-it --port 8091 --max-batch-size 4

# temperature=0 works:
curl localhost:8091/v1/chat/completions -d '{"model":"...","messages":[{"role":"user","content":"Hello"}],"temperature":0}'
# ✅ Returns "Hello!"

# temperature>0 crashes:
curl localhost:8091/v1/chat/completions -d '{"model":"...","messages":[{"role":"user","content":"Hello"}]}'
# ❌ CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES
```

## Proposed Fix

The `topk_topp_sampling` kernel at `topk_fi.mojo:1419` needs SM 12.1-aware resource tuning:

1. Reduce thread block size for SM 12.x
2. Add `--maxrregcount` equivalent for SM 12.1 builds
3. Or provide a fallback non-fused sampling path

Happy to test any patches on our DGX Spark hardware.

**Related**: modular/modular#5920, modular/modular#6396

lightofbaldr:
Fixes #6488.

## What

The top-k/top-p sampling kernels in `max/kernels/src/nn/topk_fi.mojo` fail to launch on Blackwell GPUs with `CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES`, breaking all non-greedy text generation on those GPUs. This PR declares the missing launch-bound metadata on the three sampling kernels, and adds the RTX PRO 4000 Blackwell to the bazel `gpu_mapping`.

## Symptom

```
max generate --model-path google/gemma-3-1b-it --devices=gpu --prompt "The capital of France is"
# ...
max/kernels/src/nn/topk_fi.mojo:1419 ... topk_topp_sampling_from_prob_...
  on device cuda:0 with error 'CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES (too many resources requested for launch)'
```

The forward pass, KV cache, and greedy decoding (`--top-k 1`) all work — only the stochastic top-k/top-p sampling path fails.

## Root cause

These kernels launch with `block_dim = block_size` (default **1024**) but do **not** declare `MAX_THREADS_PER_BLOCK_METADATA`. Without the launch bound the compiler allocates registers per thread freely; on Blackwell the sampling kernels exceed 64 registers/thread, so `1024 threads × regs/thread` over-subscribes the 65,536-register-per-block file and the launch is rejected. (`topk_mask_logits`, a lighter kernel, fits at 1024 and is unaffected.)

The sibling kernels in `mojo/stdlib/std/algorithm/backend/gpu/reduction.mojo` already declare this metadata; the topk sampling kernels were simply missing it.

## Fix

Declare `MAX_THREADS_PER_BLOCK_METADATA` on the three sampling kernels (`TopKSamplingFromProbKernel`, `TopKTopPSamplingFromProbKernel`, `topk_softmax_sample_kernel`), so the compiler caps registers-per-thread to fit `block_size` threads. `block_size` is unchanged, so there is **no occupancy or performance change on architectures where the 1024-thread launch already fit** — this is purely a correctness fix for register-constrained targets, and applies to any such GPU, not only Blackwell.

Also adds `NVIDIA RTX PRO 4000` (`sm_120a`) to `bazel/common.MODULE.bazel` `gpu_mapping`/`supported_gpus` so the build/test toolchain recognizes it.

## Validation

`max/kernels/test/gpu/nn/test_topk_gpu_fi.mojo`, red → green:

| GPU | Arch | Before | After |
|-----|------|--------|-------|
| NVIDIA RTX PRO 4000 Blackwell | sm_120 | `LAUNCH_OUT_OF_RESOURCES` | ✅ all pass |
| NVIDIA GB10 (DGX Spark) | sm_121 | `LAUNCH_OUT_OF_RESOURCES` | ✅ all pass |

Also confirmed end-to-end: `max generate` with default sampling no longer crashes on either GPU.

Tested with MAX 26.3.0 / Mojo 1.0.0b1.


lightofbaldr:

lightofbaldr:
Contributes to #6570 (the sm_120 / sm_121 consumer Blackwell support epic). Found and fixed on both `sm_120` (RTX PRO 4000 Blackwell) and `sm_121` (DGX Spark GB10) hardware: the top-k/top-p sampling kernels were missing the `MAX_THREADS_PER_BLOCK_METADATA` launch bound their `reduction.mojo` siblings carry, so the 1024-thread launch over-subscribed the register file and failed with `LAUNCH_OUT_OF_RESOURCES`. Validated red→green in `test_topk_gpu_fi.mojo` on both GPUs, plus end-to-end `max generate`.

max/kernels/src/nn/topk_fi.mojo
```diff
@@ -673,6 +674,9 @@ def _block_reduce_value_count[
 @__name(
     t"topk_sampling_from_prob_{dtype}_{out_idx_type}_{deterministic}",
 )
+@__llvm_metadata(
```
BradLarson: Because this would alter behavior for all NVIDIA systems, this would require careful review from our kernels team and would take a while to land. I'd like to find a narrower fix for this issue that's instead localized to consumer Blackwell systems. It seems like possibly restricting `block_size` to 512 on GPUs with fewer than 2048 threads / SM in the `topk_topp_sampling_from_prob` launch might be a more focused fix for this specific GPU architecture. That seems like it works for me, locally.

dylan-stark: @BradLarson, is [this](https://github.com/modular/modular/compare/main...dylan-stark:modular:dylan-stark-topk-sm-120?diff=unified&w) too narrow? (Note: I set block size to 896 after a quick benchmark sweep showed ~20% better throughput over 512.)

I had run into the same issue and am happy to see the epic and @lightofbaldr's patch.

lightofbaldr: Thanks @BradLarson — reworked to localize this to consumer Blackwell as you suggested. Dropped the kernel-wide MAX_THREADS_PER_BLOCK_METADATA entirely; instead each sampling launch (topk_sampling_from_prob, topk_topp_sampling_from_prob, topk_softmax_sample) caps block_size to 512 only when threads_per_multiprocessor < 2048 (sm_120/sm_121 = 1536, datacenter = 2048), computed at compile time so datacenter codegen is untouched. Used min(block_size, 512) so smaller caller-provided blocks aren't bumped up.

Verified on a GB10 (sm_121): the full test_topk_gpu_fi suite passes all trials — topk_topp_sampling_from_prob and _topk_topp_sampling_fi, fp32/bf16, batch 1/4/8, N up to 32000, K up to 50. Happy to adjust the threshold or split this out further if you'd prefer.

dylan-stark (commented):
I can verify the patch works on my RTX 5060 ti (after updating `bazel/common.MODULE.bazel`).

One note: I think we should continue to honor the caller's requested block size, when given, with something like `block_size if block_size else 1024 if not [...] else 512`. That preserves the existing behavior (1024 is just a default) and just adds 512 as an alternate default for low thread counts. A user would then be free to bump up (I know it works up through 896) for their own tuning/benchmarking.

CINOAdam:
Thanks @dylan-stark for the benchmark — pushed `5c21407a` which raises the consumer-Blackwell cap from `512` → `896` to match what you measured on `sm_120`. 896 still fits the 1536-thread/SM budget at one block per SM, so the `LAUNCH_OUT_OF_RESOURCES` fix is preserved, and your sweep showed ~20% better throughput than 512.

Branch is now rebased clean on top of `modular/main` (HEAD `5c21407a`); the changes are scoped to the three sampling launches in `max/kernels/src/nn/topk_fi.mojo` and gated at compile time on `hw_info.threads_per_multiprocessor < 2048`, so datacenter codegen is untouched.

@BradLarson — would appreciate another look when you have a moment. Happy to re-run the `test_topk_gpu_fi` suite on GB10 again if useful; both the original 512 cap and now 896 pass it locally.

PR #6591, opened 2026-05-22
[Kernels][GPU] Fix attention sink on consumer Blackwell (sm_120/sm_121)

### LINKED ISSUE #6198: [BUG] [GPU][SM120] sink-aware flash attention is incorrect on RTX 5090: test_flash_attention_sink_kernel returns 1.0 instead of ~0.3013

lightofbaldr:
Fixes #6198.

## Symptom

On consumer Blackwell GPUs (sm_120 / sm_121), the sink-aware flash-attention path is numerically wrong — `test_flash_attention_sink_kernel` returns `1.0` instead of `~0.3013` (the QK logits are forced to 0, V is all 1, and each head has a positive sink logit, so the output should equal the probability mass on real keys, `num_keys/(num_keys+exp(sink))`).

## Root cause (two issues)

1. **Sink dropped on the generic path.** The optimized generic flash-attention kernel does not apply `sink_weights`. The specialized `sm90` / `sm100` / AMD kernels do. Consumer Blackwell (sm_120/sm_121) matches none of those specialized paths, so it falls to the generic kernel and the attention-sink logit is silently dropped from the softmax denominator → output `1.0`.

2. **Wrong per-head sink index in the naive path.** `softmax_kernel` indexed the per-head sink by `row_idx % num_heads`, which is only correct when `seq_q == 1` (decode). With the naive softmax shape `[batch*num_heads, seq_q, num_keys]`, multi-query rows then pick up the wrong head's sink.

## Fix

1. Exclude the `sink` case from `flash_attention_applicable` on architectures whose optimized flash path lacks sink support, routing it to the (correct) naive kernel:
   ```
   ... and not (sink and not (gpu_info == H100 or _is_sm10x_gpu(gpu_info) or has_amd_gpu_accelerator()))
   ```
2. Index the per-head sink by the row's outer (`batch*head`) dimension:
   ```
   umod(row_idx // ufloordiv(num_rows, shape[0]), num_heads)
   ```

## Validation

`test_flash_attention.mojo` passes in full — including `test_flash_attention_sink_kernel` (both heads now correct) — on **sm_121 (DGX Spark GB10)**, with no regressions in the other flash-attention configurations. Tested with MAX 26.3 / Mojo 1.0.0b1.


lightofbaldr:
Closing — main has an equivalent fix landed as part of `564a8f813d [Kernels] Improve short-axis GPU softmax kernel perf + Coord migration`. It uses `row_coords[0]` (the outer-dim index) for the per-head sink lookup, which is semantically the same as the `row_idx // ufloordiv(num_rows, shape[0])` derivation I had here. Plus it adds explanatory comments about the per-head reasoning. Thanks to the contributors on that PR.

PR #6594, opened 2026-05-22
[Stdlib][GPU] Enable NVFP4 (block-scaled FP4) on consumer Blackwell (sm_121)

lightofbaldr:
**Stacked on #6593** (sm_121a target). Until that merges, the diff here also shows #6593's 3-line target change; it will reduce to just the two stdlib fixes + test after #6593 lands.

## What

With the arch-specific `sm_121a` target in place (#6593), two stdlib gaps still blocked the arch-agnostic `naive_block_scaled_matmul` (NVFP4) path on the DGX Spark (GB10, sm_121):

1. **sm_121 missing from arch-capability predicates.** Added `_SM_121X_ARCHS` + `_is_sm_121x` / `_has_sm_121x`, wired into `_is_sm_120x_or_newer` / `_has_sm_120x_or_newer`. This is what gates the FP4 decode helper `cast_f4e2m1x2_to_fp16x2` (constraint `_is_sm_100x_or_newer`), among other Blackwell paths.

2. **float8→float hardware cast unlowered on sm_121.** The `_convert_float8_to_f16/f32` fast-paths emit a `pop.cast` whose lowering isn't implemented for the sm_121 backend (a mojo→LLVM lowering gap, independent of the PTX target). On sm_121, route float8 conversions to the existing pure bit-manipulation fallback — always numerically correct; the only cost is forgoing the hardware-accelerated conversion.

Both changes are **scoped to sm_121** (guarded by `_is_sm_121x()` / the new arch set) and do not change behavior on any other target.

## Result

Together with #6593, a block-scaled FP4 (NVFP4) GEMM **compiles, launches, and is numerically correct** on the GB10.

## Test

Adds `test_block_scaled_fp4_naive.mojo` — a known-answer correctness test for `naive_block_scaled_matmul` that needs no reference matmul: every FP4 element decodes to 1.0 (packed byte `0x22`) and every block scale is 1.0, so with a unit per-tensor scale `C[i, j] == K`. Verified on a DGX Spark (GB10): all 65,536 output elements exact. The naive path is arch-agnostic, so the test also runs on existing supported GPUs.


mojo/stdlib/std/sys/info.mojo
```diff
@@ -564,6 +564,19 @@ def _has_sm_120x() -> Bool:
     return _has_nvidia_gpu_any[_SM_120X_ARCHS]()
 
 
+comptime _SM_121X_ARCHS: List[String] = ["sm_121", "sm_121a"]
```
BradLarson: Could you pull out these platform detection changes into a dedicated PR, or remove everything else from the PR but these? These additions in this file are correct, and may actually clean up some of the platform detection issues elsewhere. I'd like to land them in isolation.

lightofbaldr: Thanks @BradLarson! Pulled the platform-detection changes into a dedicated PR: https://github.com/modular/modular/pull/6599 — just the _SM_121X_ARCHS predicates and the _is/_has_sm_120x_or_newer wiring in sys/info.mojo, nothing else. I'll rebase this PR and the rest of the sm_121 stack on top once it lands.

CINOAdam:
Rebased on top of current `modular/main` (HEAD `65725d81`) — the platform-detection bits @BradLarson asked to extract are now inherited from merged #6599 (`_SM_121X_ARCHS`, `_is_sm_121x`, `_has_sm_121x`), so the duplicated code is out of this PR and the diff is now scoped to:

  - `mojo/stdlib/std/builtin/simd.mojo` (+4/-2): import `_is_sm_121x` and gate the inline-PTX `cvt.rn.f32.e4m3x2` / `cvt.rn.f16x2.e4m3x2` fast paths so consumer Blackwell falls through to the generic conversion. The PTX instructions are not available on `sm_121`.
  - `max/kernels/test/gpu/linalg/test_block_scaled_fp4_naive.mojo` (new, 143 lines): known-answer NVFP4 GEMM test that runs on any NVIDIA GPU including consumer Blackwell.

Ready for re-review when you have a moment.

lightofbaldr:
Superseded by #6596, which carries this enablement commit rebased onto current `main` (with the `Idx[X]()` → `Idx[X]` nightly-API fix) alongside the dependent `block_scaled_matmul` routing change, verified on consumer-Blackwell hardware. Closing to consolidate — the enable change lives on in #6596.

PR #6596, opened 2026-05-22
[Kernels][GPU] Route NVFP4 operations to cuBLAS on consumer Blackwell (sm_120 / sm_121)

lightofbaldr:

`block_scaled_matmul` hard-asserted SM100 (B200) only, so NVFP4 was unreachable from the inference dispatch on consumer Blackwell. This routes it to cuBLAS kernels on `sm_120` / `sm_121`.

`test_cublaslt_scaled_nvfp4_matmul.mojo` now passes on `sm_120` / `sm_121` system, verified on GB10 (DGX Spark).


CINOAdam:
Rebased on top of current `modular/main` (HEAD `3fc20d40`), now stacked cleanly on #6594. The conflict against main was the in-flight `_width: Int → _width: SIMDSize` change in `compute_lambda_wrapper`; carried that over into the rebased commit so the dispatch wrapper builds against current main.

Diff summary post-rebase:

  - `max/kernels/src/linalg/fp4_quantization.mojo` (+189/-165): remove the hard `assert ctx.default_device_info.compute == B200.compute`, wrap the SM100 warp-specialized body in `comptime if ctx.default_device_info.compute == B200.compute`, add an `elif compute >= 12.0` branch that routes to `naive_block_scaled_matmul`. Most of the diff is the indentation shift on the existing SM100 body.
  - `max/kernels/src/linalg/fp4_utils.mojo` (+1/-1): tiny supporting tweak.
  - `max/kernels/test/gpu/linalg/test_block_scaled_fp4_dispatch_sm121.mojo` (new, 127 lines): end-to-end dispatch test exercising the new sm_121 path.

Verified on a GB10 (sm_121a): the known-answer NVFP4 GEMM driven through `block_scaled_matmul` routes to the naive path and produces correct results (all 65,536 output elements exact). No changes to SM100 codegen.

Ready for review.

max/kernels/src/linalg/fp4_quantization.mojo
```diff
@@ -1740,217 +1741,250 @@ def block_scaled_matmul[
     var a_scales = a_scales_device.as_unsafe_any_origin()
     var b_scales = b_scales_device.as_unsafe_any_origin()
 
-    comptime assert (
```
BradLarson, line 1743: This is an extreme change here to critical kernel paths, and is very difficult to review. Can you isolate this to the minimal changes necessary to add `sm_12x` to be registered correctly here? That shouldn't require this level of change.

max/kernels/src/linalg/fp4_quantization.mojo
```diff
@@ -697,10 +697,10 @@ def naive_block_scaled_matmul_kernel[
     elementwise_lambda_fn: Optional[elementwise_epilogue_type] = None,
 ](
     c: LayoutTensor[c_type, c_layout, MutAnyOrigin],
-    a: LayoutTensor[a_type, a_layout, MutAnyOrigin],
-    b: LayoutTensor[b_type, b_layout, MutAnyOrigin],
-    a_scales: LayoutTensor[a_scales_type, a_scale_layout, MutAnyOrigin],
-    b_scales: LayoutTensor[b_scales_type, b_scale_layout, MutAnyOrigin],
+    a: LayoutTensor[a_type, a_layout, ImmutAnyOrigin],
```
BradLarson: These aren't necessarily wrong, but the code was building before. Why these changes to mutability as part of this?

max/kernels/src/linalg/fp4_quantization.mojo
```diff
@@ -1708,9 +1708,10 @@ def block_scaled_matmul[
     comptime assert a_scales_device.rank == 5 and a_scales_device.flat_rank == 5
     comptime assert b_scales_device.rank == 5 and b_scales_device.flat_rank == 5
 
-    comptime assert (
-        ctx.default_device_info.compute == B200.compute
-    ), "This kernel is only supported on SM100"
+    # NVFP4 arch dispatch happens after the input tensors are set up below:
```
BradLarson: This comment isn't really necessary, people can follow the code flow below.

max/kernels/test/gpu/linalg/BUILD.bazel
```diff
@@ -589,6 +589,44 @@ mojo_test(
     ],
 )
 
+# Consumer-Blackwell (sm_120 / sm_121, e.g. RTX PRO / GB10 / DGX Spark) NVFP4
```
BradLarson: I don't think we want to add these tests to be run for every NVIDIA GPU.

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -3675,7 +3676,7 @@ def _convert_float8_to_f32[
     dtype: DType,
     size: SIMDSize,
 ](val: SIMD[dtype, size]) -> SIMD[DType.float32, size]:
-    comptime if _is_sm_9x_or_newer() and dtype in (
+    comptime if _is_sm_9x_or_newer() and not _is_sm_121x() and dtype in (
```
BradLarson: This will be incorrect for `sm_120` Blackwell devices, which should also be supported. What is needed is a function like `_is_sm12x_gpu()` that acts like `_is_sm10x_gpu()` and that can cover `sm_120` and `sm_121`.

max/kernels/test/gpu/linalg/BUILD.bazel
lightofbaldr: My fault, I forgot to remove that, let me address your comments and update everything. Thank you!

BradLarson:
Sorry, my test DGX Spark got pulled into use as a demo system and I just got another one set up. As a larger comment, I don't think the right approach is to use a naive NVFP4 kernel. Instead, just add the bare minimum code to route `sm_12x` devices to cuBLAS fallback NVFP4 kernels. That'll get you up and going with NVFP4 models on DGX Spark much faster. Also, anything that massively reworks our B200 Blackwell paths as in the above will most likely be rejected by our kernel team. The changes should be much more localized in scope, only adding the right comptime checks as needed.

lightofbaldr:
> Sorry, my test DGX Spark got pulled into use as a demo system and I just got another one set up. As a larger comment, I don't think the right approach is to use a naive NVFP4 kernel. Instead, just add the bare minimum code to route `sm_12x` devices to cuBLAS fallback NVFP4 kernels. That'll get you up and going with NVFP4 models on DGX Spark much faster. Also, anything that massively reworks our B200 Blackwell paths as in the above will most likely be rejected by our kernel team. The changes should be much more localized in scope, only adding the right comptime checks as needed.

@BradLarson 
Not a problem at all. I have a Spark and run RTX pro 4000's locally. I would love to try and knock out as many open issues for this hardware over the next month. Any guidance is welcome so I can bake that into my SOP. Thank you!

lightofbaldr:
Thanks — reworked to exactly that. sm_12x NVFP4 now routes to the cuBLASLt vendor fallback via localized comptime gates; the naive kernel is gone and the B200 / SM100 warp-specialized path is untouched:

- `[Stdlib]` route sm_12x fp8→f32/f16 to the software scalar path (the hardware path doesn't lower on consumer Blackwell)
- `[Kernels]` relax the B200 gates in `block_scaled_matmul`, `block_scaled_matmul_with_epilogue`, and vendor `_cublasLt_matmul` block-scaling setup so sm_12x reaches cuBLASLt; comptime-exclude the SM100 Mojo dispatch on sm_12x
- `[Kernels]` extend the existing `test_cublaslt_scaled_nvfp4_matmul` to also run on GB10

Verified end-to-end on GB10 (sm_121a, CUDA 13.4): cuBLASLt NVFP4 block-scaled is accepted, matches the naive reference within tolerance across all 8 shapes (4 bit-identical, the rest bf16-tight, max_abs ≤ 1.0), no NaN/Inf. `git diff -w` on the kernel change is +30/−6; the B200 codegen path is unchanged.

max/kernels/src/linalg/matmul/vendor/blas.mojo
```diff
@@ -1077,7 +1077,13 @@ def _cublasLt_matmul[
         msg="failed to set cublasLtMatmulDescAttribute for transb",
     )
 
-    comptime if _is_sm10x_gpu(ctx.default_device_info):
+    # cuBLASLt block scaling: SM100 (B200) and consumer Blackwell (sm_120 /
```
BradLarson: This comment is unnecessary and can be removed.

max/kernels/src/linalg/matmul/vendor/blas.mojo
```diff
@@ -1077,7 +1077,13 @@ def _cublasLt_matmul[
...
+    # sm_121, e.g. DGX Spark) share this host-side descriptor setup; the driver
+    # decides hardware support. B200 behavior is unchanged.
+    comptime if (
+        _is_sm10x_gpu(ctx.default_device_info)
+        or ctx.default_device_info.compute >= 12.0
```
BradLarson: Would it be preferable to have a `_is_sm12x_gpu()` function to match `_is_sm10x_gpu()`?

max/kernels/src/linalg/fp4_quantization.mojo
```diff
@@ -1549,9 +1549,16 @@ def block_scaled_matmul_with_epilogue[
     by the lambda.
     """
 
-    comptime assert _is_sm10x_gpu(
-        ctx.default_device_info
-    ), "This kernel is only supported on SM100"
+    # The vendor (cuBLASLt) block-scaled matmul below is host-side and
```
BradLarson: Same comments as above.

max/kernels/src/linalg/fp4_quantization.mojo
```diff
@@ -1708,9 +1715,16 @@ def block_scaled_matmul[
     comptime assert a_scales_device.rank == 5 and a_scales_device.flat_rank == 5
     comptime assert b_scales_device.rank == 5 and b_scales_device.flat_rank == 5
 
+    # B200 (SM100) runs the warp-specialized Mojo kernels below; consumer
```
BradLarson: This comment is unnecessary and can be removed.

max/kernels/test/gpu/linalg/BUILD.bazel
```diff
@@ -135,7 +135,15 @@ _EXTRA_CONSTRAINTS = {
...
     }),
     "test_cublaslt_scaled_mxfp8_matmul.mojo": ["//:b200_gpu"],
-    "test_cublaslt_scaled_nvfp4_matmul.mojo": ["//:b200_gpu"],
+    # NVFP4 cuBLASLt block-scaled matmul: B200 (SM100) and consumer Blackwell
+    # (sm_120 / sm_121, e.g. GB10 / DGX Spark), which routes to the same vendor
+    # cuBLASLt path. Verified on GB10 (sm_121a). Note: on sm_12x runners invoke
+    # via `bazelw run` — `bazelw test` needs the gpu-memory resource config.
+    "test_cublaslt_scaled_nvfp4_matmul.mojo": select({
+        "//:b200_gpu": ["//:b200_gpu"],
+        "//:gb10_gpu": ["//:gb10_gpu"],
```
BradLarson, line 140: Do we have a `gb10_gpu` Bazel device target?

max/kernels/test/gpu/linalg/BUILD.bazel
```diff
@@ -135,7 +135,15 @@ _EXTRA_CONSTRAINTS = {
         "//conditions:default": ["@platforms//:incompatible"],
     }),
     "test_cublaslt_scaled_mxfp8_matmul.mojo": ["//:b200_gpu"],
-    "test_cublaslt_scaled_nvfp4_matmul.mojo": ["//:b200_gpu"],
+    # NVFP4 cuBLASLt block-scaled matmul: B200 (SM100) and consumer Blackwell
```
BradLarson: This comment is unnecessary and can be removed.

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -3675,7 +3676,7 @@ def _convert_float8_to_f32[
     dtype: DType,
     size: SIMDSize,
 ](val: SIMD[dtype, size]) -> SIMD[DType.float32, size]:
-    comptime if _is_sm_9x_or_newer() and dtype in (
+    comptime if _is_sm_9x_or_newer() and not _is_sm_120x_or_newer() and dtype in (
```
BradLarson, line 3679: This may exceed the maximum line length, you may want to run `./bazelw run //:format` here.

max/kernels/src/linalg/fp4_quantization.mojo
```diff
@@ -1549,9 +1549,16 @@ def block_scaled_matmul_with_epilogue[
...
-        ctx.default_device_info
-    ), "This kernel is only supported on SM100"
+    # The vendor (cuBLASLt) block-scaled matmul below is host-side and
+    # arch-portable: SM100 (B200) and consumer Blackwell (sm_120 / sm_121, e.g.
+    # DGX Spark) both use it.
+    comptime assert (
+        _is_sm10x_gpu(ctx.default_device_info)
+        or ctx.default_device_info.compute >= 12.0
+    ), (
+        "block_scaled_matmul_with_epilogue is only supported on SM100 (B200) or"
```
BradLarson: For a more minimal change, the comment could simply be "This kernel is only supported on SM100 or SM120".

max/kernels/src/linalg/fp4_quantization.mojo
```diff
@@ -1708,9 +1715,16 @@ def block_scaled_matmul[
...
+    # B200 (SM100) runs the warp-specialized Mojo kernels below; consumer
+    # Blackwell (sm_120 / sm_121, e.g. DGX Spark) has no SM100 path and routes
+    # to the vendor (cuBLASLt) block-scaled fallback further down.
     comptime assert (
-        ctx.default_device_info.compute == B200.compute
-    ), "This kernel is only supported on SM100"
+        _is_sm10x_gpu(ctx.default_device_info)
+        or ctx.default_device_info.compute >= 12.0
+    ), (
+        "block_scaled_matmul is only supported on SM100 (B200) or consumer"
```
BradLarson: Again: "This kernel is only supported on SM100 or SM120"

max/kernels/src/linalg/fp4_quantization.mojo
BradLarson: You may need to add an `_is_sm_120x_gpu` condition here, too. At least, I did locally.

BradLarson: This may also need to be enabled for `sm_120x`.

lightofbaldr:
Thanks — all addressed:

- **`_is_sm12x_gpu(info)`** added to `gpu/host/info.mojo` (mirrors `_is_sm10x_gpu`, covers sm_120 + sm_121); every inline `compute >= 12.0` check now uses it.
- **Comments removed**, and the relaxed asserts reduced to `"This kernel is only supported on SM100 or SM120"`.
- **The additional gate sites** you flagged — plus two kernel-*internal* ones a dispatch-level pass misses: opened `quantize_dynamic_block_scaled` / `quantize_dynamic_scaled_fp4fp8` and `block_scales_interleave` / `block_scales_interleave_fp4` for sm_12x. The quantize path also needed the **write-direction** fp8 fix (`_convert_f32_to_float8`) — the twin of the read-direction one — so both directions now fall through to the software scalar path on consumer Blackwell.
- **Mutability (`ImmutAnyOrigin`)**: that was part of the naive-kernel approach, which this rewrite removes entirely — no longer present.
- **`gb10_gpu`**: yes, it is a real device target (generated from `@mojo_gpu_toolchains//:gpus.bzl`); kept it and added `rtx5090_gpu` for sm_120 coverage. Ran `./bazelw run //:format`.

Verified end-to-end on GB10 (sm_121a, CUDA 13.4): the matmul, quantize, and interleave paths all accept and match their references within tolerance, no NaN/Inf. The B200 / SM100 codegen path is unchanged.

BradLarson:
!sync

BradLarson:
!sync

BradLarson:
!sync

BradLarson:
Appreciate your patience on this, and the flexibility to reshape this for using the cuBLAS kernels. I finally got this working internally in our CI, and just merged. I noticed that there are couple of small things we may need on top of this to fully drive models end-to-end in NVFP4, and I can see about adding those if you didn't beat me to it.

lightofbaldr:
Thanks Brad — really appreciate you getting it working and reshaping it around the cuBLAS fallback. I've updated our branch, so CI should settle shortly. We haven't started on the end-to-end pieces, so please go ahead — what did you have in mind for the couple of things needed on top? If it's useful, we can validate on real sm_120 / sm_121 hardware on our side. Otherwise we'll leave it with you for now.

BradLarson:
When I tested with

```
./bazelw run //max/python/max/_entrypoints:pipelines -- generate --model nvidia/Llama-3.1-8B-Instruct-NVFP4 --prompt "The sky is blue because"
```

I hit a couple of remaining codepaths specific to NVFP4. I believe you had a fix for one of those in the SIMD checks in a previous PR (or version of this PR), but I know there were others I saw on a local testing machine.

lightofbaldr:
Brad — you were right on both counts: there were remaining sm_12x codepaths, and "a fix we had then removed" was the crux. We reproduced `nvidia/Llama-3.1-8B-Instruct-NVFP4` on a DGX Spark (sm_121a) and traced the first break past the merged reroute: `quantize_dynamic_block_scaled` (`max/python/max/nn/kernels.py`) emits rank-5 activation scales only for `_is_sm10x_gpu()`, so sm_12x falls to the rank-2 else-branch and trips `dynamic_block_scaled_matmul`'s rank-5 check. reroute_v2 taught the Mojo kernel to emit rank-5 for sm_12x, but the matching Python graph op stayed on the else-branch — the half removed in the cleanup.

To confirm it's arch-specific and not a general graph bug, we ran the same model on a B200: clean, 367 tok/s, no rank error. With the one-line restore, the Spark GB10 now generates coherent text at ~41 tok/s (was the rank-5 `ValueError`).

Opened #6769 with the fix (a `_is_sm12x_gpu()` helper + including it in the rank-5 branch). Happy to adjust — and we've got real sm_120/sm_121 hardware if you want anything else validated on-device.

PR #6617, opened 2026-05-28
collections/deque: guard `rotate` against empty deque to prevent uninitialized memory reads

Dev-X25874:
## Linked issue

N/A — trivial fix

## Type of change

## Motivation

`Deque.rotate(n)` on an empty deque reads from uninitialized memory.
When `n > 0` the positive branch decrements `_tail` and reads from
`_data + _tail` (one slot before the empty ring's sole position) before
writing to `_data + _head - 1`; when `n < 0` the negative branch reads
from `_data + _head` and writes to `_data + _tail`. In both cases the
source slot has never been initialized.

Python's `collections.deque` treats `rotate` on an empty (or zero-step)
deque as a no-op. The Mojo implementation had no equivalent guard, so
any call to `rotate` on an empty `Deque` was undefined behavior.

## What changed

Added an early-return guard at the top of `rotate`:

```mojo
if n == 0 or len(self) == 0:
    return
```

`len(self)` evaluates to `(self._tail - self._head) & (self._capacity - 1)`,
which is 0 exactly when the deque is empty regardless of the circular-buffer
wrap state. The guard short-circuits before either loop branch runs, so
no memory is touched. The `n == 0` arm is included for free to match
Python's semantics and eliminate the loop overhead on a trivially
no-op call.

No existing behavior changes — all current `test_rotate` cases use a
non-empty deque with `n` well within bounds.

## Testing

Added `test_rotate_empty_deque` to
`mojo/stdlib/test/collections/test_deque.mojo`, placed immediately after
the existing `test_rotate`. The new test:

- Creates an empty `Deque[Int]()`.
- Calls `rotate(1)`, `rotate(-1)`, and `rotate(0)` in sequence.
- Asserts `len(q) == 0` after each call.

On the unpatched code all three calls read from uninitialized memory;
with the fix they are all no-ops.

## Checklist

      approval
      smaller PRs where possible

### DIFF
--- a/mojo/stdlib/std/collections/deque.mojo
+++ b/mojo/stdlib/std/collections/deque.mojo
@@ -897,6 +897,8 @@ struct Deque[ElementType: Movable & ImplicitlyDestructible](
             n: Number of steps to rotate the deque
                 (defaults to 1).
         """
+        if n == 0 or len(self) == 0:
+            return
         if n < 0:
             for _ in range(-n):
                 (self._data + self._tail).init_pointee_move_from(
--- a/mojo/stdlib/std/collections/deque.mojo
+++ b/mojo/stdlib/std/collections/deque.mojo
@@ -892,6 +892,7 @@ struct Deque[ElementType: Movable & ImplicitlyDestructible](
 
         If `n` is positive, rotates to the right.
         If `n` is negative, rotates to the left.
+        On an empty deque or when `n == 0`, this is a no-op.
 
         Args:
             n: Number of steps to rotate the deque
--- a/mojo/stdlib/test/collections/test_deque.mojo
+++ b/mojo/stdlib/test/collections/test_deque.mojo
@@ -1038,8 +1038,6 @@ def test_rotate() raises:
 
 
 def test_rotate_empty_deque() raises:
-    # Rotating an empty deque must be a no-op; without the guard it reads
-    # from uninitialized memory (undefined behaviour).
     q = Deque[Int]()
     q.rotate(1)
     assert_equal(len(q), 0)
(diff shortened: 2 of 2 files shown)


Dev-X25874:

mojo/stdlib/std/collections/deque.mojo
```diff
@@ -897,6 +897,8 @@ struct Deque[ElementType: Movable & ImplicitlyDestructible](
             n: Number of steps to rotate the deque
                 (defaults to 1).
         """
+        if n == 0 or len(self) == 0:
```
JoeLoser, line 901: **Thought** WDYT of documenting this new behavior in the docstring? E.g. "On an empty deque or when `n == 0`, this is a no-op."

mojo/stdlib/test/collections/test_deque.mojo
```diff
@@ -1037,6 +1037,18 @@ def test_rotate() raises:
...
+def test_rotate_empty_deque() raises:
+    # Rotating an empty deque must be a no-op; without the guard it reads
+    # from uninitialized memory (undefined behaviour).
+    q = Deque[Int]()
+    q.rotate(1)
+    assert_equal(len(q), 0)
+    q.rotate(-1)
+    assert_equal(len(q), 0)
+    q.rotate(0)
+    assert_equal(len(q), 0)
```
JoeLoser, line 1047: **Suggestion** The current test cases only exercise the `len(self) == 0` half of the new branch added.

Can we please add a non-empty deque where we rotate(0) so the new branch is verified in isolation:

```mojo
    # n == 0 on a non-empty deque must also be a no-op.
    q2 = Deque(1, 2, 3)
    q2.rotate(0)
    assert_equal(len(q2), 3)
    assert_equal(q2[0], 1)
    assert_equal(q2[1], 2)
    assert_equal(q2[2], 3)
```

JoeLoser (approved):

mojo/stdlib/std/collections/deque.mojo
Dev-X25874: Good call — added a note to the docstring clarifying that an empty deque 
or n == 0 results in a no-op. Thanks for the suggestion!

mojo/stdlib/test/collections/test_deque.mojo
Dev-X25874: Agreed, the original test only covered the len(self) == 0 path. Added a 
non-empty deque case with rotate(0) to verify the n == 0 branch in 
isolation. Thanks for catching that!

PR #6619, opened 2026-05-28
[Kernels] Fix Tuple-returning shape() in state_space ops

ulmentflam:
## Linked issue

N/A -- trivial

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update

## Motivation

State-space ops in `max/kernels/src/state_space/*_ops.mojo` bridge the Mojo kernels to the Python graph layer, but three of them had `shape()` methods returning a parameterized `std.Tuple` that the MLIR kernel-package loader couldn't resolve silently blocking `KernelLibrary.add_path(state_space.mojoc)` for every Python caller. The bug was undetected because no existing integration test exercised the package-load path, but it surfaces the moment you call any of these ops from Python (Mamba's selective_scan_fwd wrapper, the new Mamba2 ssd_chunk_scan_combined, etc.). This patch swaps the Tuple returns for single IndexList[N] returns, matching the pattern already used by `gemv_and_partial_norm` in `builtin_kernels/linalg.mojo` for another multi-output op, unblocking the entire Python pipeline layer for state-space models without changing any kernel behavior.

## What changed

Replaces the `shape()` return type on three state_space op registrations (CausalConv1DUpdate, SelectiveScanUpdate, and the varlen scan update) from `Tuple[IndexList[A], IndexList[B]]` to a single `IndexList[N]` returning the primary OutputTensor's shape. The kernels themselves are untouched. Only the graph-build-time shape-inference hook changes. Per-OutputTensor shape is now inferred from each `execute()` body's OutputTensor allocations, matching the existing convention in `builtin_kernels/linalg.mojo`'s `gemv_and_partial_norm`, which is another multi-output op. Non-obvious: the `shape()` return-type narrowing is technically a public-surface API change, but no callers introspect these registrations directly (verified by grep), and behavior at call time is unchanged because op semantics already drove per-output allocation. Performance: zero runtime impact, purely a graph-build-time metadata fix.

## Testing

I ran the full test suite with `./bazelw` and verified no regressions on Nvidia DGX spark. I also ran a test graph from a python call to verify. 

## Checklist

      approval
- [X] PR is small and focused — I've split larger changes into a sequence of
- [X] I ran `./bazelw run format` to format my changes
- [X] If AI tools assisted with this contribution, I have included an


ulmentflam:
Looks like this has been fixed alternatively upstream. Closing it out.

PR #6646, opened 2026-06-06
[Stdlib] Fix `time.sleep()` truncation for small fractional seconds

### LINKED ISSUE #6583: [BUG] [stdlib] Function `time.sleep()` doesn't work for units smaller than seconds anymore
### Bug description

```mojo
time.sleep(1e-9)
```

gets truncated by

```mojo
    comptime NANOSECONDS_IN_SECOND = 1_000_000_000
    var total_secs = floor(sec)
    var tv_spec = _CTimeSpec(
        Int(total_secs),
        Int((sec - total_secs) * NANOSECONDS_IN_SECOND),
    )
    var req = UnsafePointer(to=tv_spec)
    var rem = _CPointer[_CTimeSpec, MutExternalOrigin]()
    _ = external_call["nanosleep", Int32](req, rem)
```

This was my original motivation for requesting modular/modular#6446, since there is no need whatsoever to be truncating this and the API should be clear in that it should be able to receive any SI time delta

### Steps to reproduce

...

### System information

...

BlueDestination:
## Linked issue
Fixes #6583

## Type of change
- [X] Bug fix (non-breaking change that fixes an issue)

## Motivation
Floating-point truncation causes `Int(0.999...)` → `0`, so `sleep(1e-9)` returns immediately.

## What changed
Replace `Int()` with `round()` in the fractional-seconds-to-nanoseconds conversion.

## Testing
All tests pass, including new cases for sub-second sleep durations (e.g., 1ms).

## Checklist
- [X] The linked issue above has been reviewed by a maintainer and is
      approval
- [X] PR is small and focused — I've split larger changes into a sequence of
- [X] I ran `./bazelw run format` to format my changes
- [X] If AI tools assisted with this contribution, I have included an
Assisted-by: AI


Mojo/stdlib/test/time/test_time.mojo
```diff
@@ -49,6 +49,32 @@ def time_capturing_function(iters: Int) -> Int:
...
 
+def test_sleep_small_fractional() raises:
+    """Test that sleep handles very small fractional seconds without truncation.
+    """
+    # sleep(1e-9) should not crash or return immediately due to Int() truncation.
+    # Before the fix, Int((sec - floor(sec)) * 1e9) would truncate near-zero
+    # floating-point values to 0, causing the nanosleep call to be skipped.
+    sleep(1e-9)
+    sleep(0.000000001)
+    sleep(0.0000001)
```
barcharcraz, line 61: was this one OK before? It seems like the bug should affect only the case where the floating point number of seconds is slightly less than 1 nanosecond (as in 1e9).

mojo/stdlib/std/time/time.mojo
```diff
@@ -358,9 +358,15 @@ def sleep(sec: Float64):
 
     comptime NANOSECONDS_IN_SECOND = 1_000_000_000
     var total_secs = floor(sec)
+    var nsecs = Int(round((sec - total_secs) * NANOSECONDS_IN_SECOND))
```
barcharcraz: is `round` right here? Shouldn't we _always_ round up (instead of down as previously). That way we can say "wait _at least_ s seconds".

Like I think `sleep(0+ulp)` should sleep for 1 nanosecond, even though it's much closer to zero.

barcharcraz (changes requested):
These changes look "correct" but I think fixing this by rounding to nearest is just as arbitrary as rounding down.

We should probably round up.

barcharcraz:
I could be misunderstanding this subtle bug, though.

barcharcraz:
is this derived from user [SharanRP](https://github.com/SharanRP)'s similar work?

mojo/stdlib/std/time/time.mojo
BlueDestination: Got it, updating to `ceil()`.

Mojo/stdlib/test/time/test_time.mojo
BlueDestination: Yes, only affects sub-ns values like `1e-9`.

BlueDestination (commented):
Addressed all review comments.

BlueDestination:
> is this derived from user [SharanRP](https://github.com/SharanRP)'s similar work?

Yes, I saw SharanRP's comment on the issue and was inspired by this approach.

barcharcraz (approved):

barcharcraz:
!sync

barcharcraz (approved):

barcharcraz:
!sync

PR #6668, opened 2026-06-11
[Kernels][Pipelines] NVFP4 fallback for pre-Blackwell NVIDIA GPUs via fused dequant-GEMV

### LINKED ISSUE #6667: [Feature Request] NVFP4 dequant fallback for pre-Blackwell NVIDIA GPUs (serve Gemma 4 NVFP4 on Ampere)
### Feature request

Support serving NVFP4 checkpoints (`float4_e2m1fnx2`) on pre-Blackwell NVIDIA GPUs via a dequantization fallback, following the precedent MAX already ships for MXFP4 on H100.

### Motivation

Gemma 4 is effectively unservable on common 24 GB Ampere cards (A10G / `g5.*`, RTX 3090/4090-class) today:

* bf16 `google/gemma-4-26B-A4B-it` is 51.6 GB — does not fit a single 24 GB card.
* The NVFP4 checkpoints fit comfortably (`RedHatAI/gemma-4-26B-A4B-it-NVFP4` is 16.4 GB; `nvidia/Gemma-4-26B-A4B-NVFP4` — the example repo id referenced in `gemma4/arch.py` — is 18.8 GB), but the NVFP4 kernels are compile-time gated to Blackwell:
  * `max/kernels/src/linalg/fp4_quantization.mojo:140/395/1536` — `comptime assert _is_sm10x_gpu(...)`
  * `max/kernels/src/linalg/fp4_utils.mojo:184` — `is_nvidia_gpu() and _is_sm_100x_or_newer()`
* The other public quantizations (AWQ/GPTQ/FP8/GGUF) are not in `gemma4`'s `supported_encodings` (`bfloat16`, `float4_e2m1fnx2`), so there is no path to serve any Gemma 4 variant on these cards.

### Precedent inside MAX

`max/kernels/src/linalg/mxfp4_dequant.mojo` ("MXFP4 dequantization kernel for H100 (SM90)") already implements "fp4 weights on hardware without fp4 tensor cores" by dequantizing to fp8/bf16 — used by `gpt_oss`, `kimik2_5` and `minimax_m2`. H100 has no fp4 tensor cores either; the requested feature extends the same idea to sm_8x with a bf16 target (Ampere has no fp8 tensor cores).

### Suggested shape

* Full weight-resident dequantization at load is not viable (26B-A4B in bf16 is 51.6 GB again), so the fallback needs on-the-fly dequant — either fused into the matmul tile loop (AWQ-kernel style: fp4 resident in VRAM, expand to bf16 in registers) or a per-layer scratch-buffer dequant + standard bf16 matmul.
* NVFP4 specifics vs the existing MXFP4 kernel: E2M1 values with FP8-E4M3 block scales, `sf_vector_size=16` (vs 32), and the 128x4 tiled scales layout (`max/python/max/nn/quant_config.py:420`).
* The `is_nvfp4` branches in the gemma4 pipeline (fused swiglu, MoE expert scales, EP dispatch) assume the Blackwell path and would need to route to the fallback.
* Replace the `comptime assert _is_sm10x_gpu(...)` gates with capability-based dispatch, in the spirit of modular/modular#6659 (pre-Ampere bf16 fallback).

### Context

Found while benchmarking Gemma 4 variants on an A10G (`g5.xlarge`). Related: modular/modular#6665 (E*B null `num_global_key_value_heads` crash, fix in modular/modular#6666).

I am happy to work on this — opening for design feedback first given the kernel scope.

msaelices:
## Summary

Serve NVFP4 checkpoints on GPUs without native FP4 matmul (pre-Blackwell NVIDIA). Fixes #6667.

The fallback is a **Marlin-style fused dequant-GEMV**: packed E2M1 is decoded in registers inside the matmul kernel, so no BF16 weight copy is ever materialized and per-token DRAM traffic is the packed bytes (~0.5 B/element vs ~4.5 for dequantize-then-matmul).

- `nvfp4_gemv` / `mo.gemv.nvfp4`: fused kernel behind `_matmul_float4`, which covers every NVFP4 linear in-tree (dense/MLP and the StackedLinear QKV projections). The unreachable NVFP4 branch of `quantized_fused_qkv_matmul` now raises a clear error instead of carrying dead fallback code.
- `dequant_nvfp4` / `mo.dequant.nvfp4`: NVFP4 generalization of the MXFP4 dequant kernel, kept for the MoE `Nvfp4DequantStrategy`.
- `cast_uint_to_fp4e2m1` rewritten branchless: the old LUT's runtime SIMD indexing spilled to local memory on every element (35x slowdown).
- Active only off SM100+ (Blackwell unchanged); clear errors for pre-interleaved (TCGEN 5D) scales and expert-parallel NVFP4.

## Performance (single-token decode, batch 1)

The fused GEMV is bandwidth-bound at batch 1 (per-token cost is essentially reading the packed weights once), so it matches the best available NVFP4 kernel on bandwidth-capable hardware and only pulls ahead where the vendor kernel is bandwidth-inefficient.

**NVIDIA L40S (sm_89, 46 GB; gemma-4 31B NVFP4, pure decode):**

| Kernel | tok/s |
|---|---|
| fused GEMV (this PR) | ~23.8 |
| vLLM nightly Marlin, same checkpoint and GPU | ~23.8 |

On the L40S both kernels sit at the memory-bandwidth bound for reading the 31B packed weights, so they are effectively **tied** — the fused GEMV leaves little on the table at batch 1, but it does not beat a well-tuned Marlin here.

**NVIDIA A10G (sm_86, 24 GB; gemma-4 12B NVFP4, warm decode):**

| Stage | tok/s |
|---|---|
| Initial dequantize-then-matmul fallback | 0.38 |
| + branchless E2M1 decode (LUT spill fix) | 0.94 |
| + fused GEMV (all NVFP4 linears, incl. QKV via StackedLinear) | **31.5 (83x)** |
| vLLM nightly Marlin, same checkpoint and GPU, eager | ~16 |

On the older A10G the fused GEMV is ~2x vLLM's Marlin, but that gap reflects Marlin being bandwidth-inefficient on that GPU rather than a general advantage (see the L40S tie above). Measured with device graph capture on; first request pays ~17 s of capture warmup, and model output is token-identical to the dequantize-then-matmul path at temperature 0.

Batched / large-M throughput is a separate kernel (tensor-core GEMM tile), out of scope here.

## Verification

CPU-reference smoke tests for both kernels (`test_nvfp4_gemv_smoke`, `test_nvfp4_dequant_smoke`); `nvidia/Llama-3.1-8B-Instruct-NVFP4` and `berkerdooo/gemma-4-12B-it-NVFP4` (with #6670) serve E2E with coherent output.

Out of scope (tracked in #6667): compressed-tensors NVFP4 parsing, Gemma 4 activation-memory headroom, tensor-core GEMM tile for large-batch prefill.

Assisted-by: AI


msaelices:
Perf follow-up: profiling the fallback on an A10G found the software LUT was the bottleneck — `cast_uint_to_fp4e2m1` indexed a `SIMD[float32, 16]` with a runtime index, which spills the vector to GPU local memory on every element (kernel ran at ~12 GB/s effective, fully local-memory bound). Replaced with a branchless exp/mantissa decode in 4135fb1: the dequant kernel now hits 422 GB/s (~35x) on the same device, verified bit-exact by the smoke test. End-to-end on the runtime-dequant serving experiment this was 2.5x (0.38 -> 0.94 tok/s on a 12B; remaining gap vs a fused Marlin-style dequant-GEMM is per-op dispatch overhead and the 3x memory traffic of dequantize-then-matmul).

msaelices:
Implemented the Marlin-style fused dequant-GEMV (4831155): one warp per output column decodes packed E2M1 in registers and dot-products against BF16 activations — the dequantized weight is never materialized, so per-token DRAM traffic drops ~9x and the per-forward transient buffers disappear. `_matmul_float4`'s pre-Blackwell path now routes through it (QKV keeps dequantize-then-matmul for now since `fused_qkv_ragged_matmul` owns the KV-cache write).

A10G, gemma-4 12B NVFP4, bs=1: per-linear 2.7 ms -> 0.133 ms through the graph executor (20x); end-to-end decode 0.38 -> 4.98 tok/s cumulative with the E2M1 LUT fix (vLLM nightly's Marlin on the same checkpoint: ~16 tok/s eager — remaining gap is the QKV path, M_TILE prefill scaling, and per-op dispatch). Correctness: scalar CPU reference in `test_nvfp4_gemv_smoke`, plus token-identical model output vs the dequant path.

msaelices:
QKV now also routes through the fused GEMV (split + `store_k/v_cache_ragged`, with K-equals-V handling for Gemma 4 full-attention layers) — no BF16 copy of any quantized weight is materialized anywhere on the fallback anymore.

Final numbers on the A10G (gemma-4 12B NVFP4, bs=1, warm decode): **31.5 tok/s**, vs 0.38 at the start of this work (83x) and ~16 tok/s for vLLM nightly's Marlin on the same checkpoint/GPU. 58-token prefill + 150-token generation sustains 29.6 tok/s. Output stays token-identical to the dequant path. Earlier per-request numbers in this thread (~5 tok/s) were first-request measurements that included device graph capture warmup.

msaelices:
Updated `main` into this branch (263 commits). Conflict resolution, **static-only (no GPU build yet)**:

- **`moe_fp8.py` imports**: union of both sides (`ep_mxfp4_max_padded_m` from main + `_is_pre_sm100_nvidia_gpu` from this branch).
- **`moe_fp8.py` `_strategy()`**: `Nvfp4Strategy` → `NvMxf4f8Strategy`. `main` removed `Nvfp4Strategy`; the native (post-SM100) NVFP4 path now uses `NvMxf4f8Strategy` (same `(quant_config, dtype)` ctor). The pre-SM100 `Nvfp4DequantStrategy` fallback branch is preserved.
- **`quant_strategy.py` `Nvfp4DequantStrategy`**: added the new `QuantStrategy` protocol kwargs (`a_scales_preshuffled`, `a_scales_max_padded_m` on `grouped_matmul`; `max_padded_M` on `fused_silu_quantize`), accepted-but-unused (the BF16 dequant fallback never preshuffles activation scales).

`py_compile` clean; the only remaining pyright notes are pre-existing (`Layer` dynamic attrs, the `max.python.max.nn` vs `max.nn` double-path false positive).

### Needs GPU validation (could not build locally)

```bash
./bazelw build //max/python/max/entrypoints:pipelines
# Pre-Blackwell fallback still correct (A10G/L40S):
./bazelw run //max/python/max/entrypoints:pipelines -- generate \
  --model berkerdooo/gemma-4-12B-it-NVFP4 --devices=gpu \
  --prompt "Explain tensor parallelism in one sentence." --max-new-tokens 64
# Expect coherent output, token-identical to pre-merge at temp 0.
# If on Blackwell, also smoke-test the native NvMxf4f8Strategy path.
```

msaelices:
recheck

Issue #6677, opened 2026-06-12
[Serve] Disabling the overlap scheduler changes generation at temperature 0


msaelices:
## Description

The same prompts produce materially different generations at `temperature=0` depending on scheduler mode. On a 46-prompt classification benchmark (`Qwen/Qwen3.5-4B`, A10G, identical prompts and sampling params):

| Server flags | Accuracy | Avg latency |
|---|---|---|
| default (overlap scheduler ON) | 85% (39/46) | ~430 ms |
| `--no-enable-overlap-scheduler --force` | 70% (32/46) | ~232 ms |

Reproducible back and forth: restoring the default scheduler restores 85%. The output distribution shifts noticeably (in our case the model stopped abstaining on low-signal inputs: 'unsure' recall dropped from 10/12 to 6/12). The large latency difference for identical workloads also looks suspicious.

At temperature 0 both modes should be near-deterministic and agree; a per-mode numerics or sampling divergence this large makes results measured in one mode invalid in the other — and since logprobs currently *require* the non-overlap mode, any logprob-based evaluation inherits the divergence.

Repro sketch: serve `Qwen/Qwen3.5-4B --max-length 4096 --max-batch-size 1` with and without the flags above; send identical chat requests with `temperature: 0` and compare outputs.

Assisted-by: AI

PR #6678, opened 2026-06-12
[Serve] Fix logprobs lost for multimodal-tokenizer architectures

### LINKED ISSUE #6676: [Serve] Logprobs come back null/empty on both endpoints with the non-overlap scheduler
## Description

With `--no-enable-overlap-scheduler --force`, requesting logprobs returns the logprobs object but with empty/null payloads:

- `/v1/completions` (`"logprobs": 1`): response contains `logprobs` with keys `[text_offset, token_logprobs, tokens, top_logprobs]` but `tokens` and `token_logprobs` are `null`.
- `/v1/chat/completions` (`"logprobs": true`): `logprobs.content` is `[]`.

Reproduced on commit ~current main (June 12) serving `Qwen/Qwen3.5-4B` on an A10G, batch size 1, temperature 0, max_tokens 8.

## Code-path notes (from reading the source, not yet live-debugged)

The request plumbing looks correct end to end: `completion_request.logprobs` → `TextGenerationRequest.logprobs` → `context.log_probabilities` (`tokenizer.py`) → `TextGenerationInputs.enable_log_probs`. `Qwen3_5Model` inherits `LogProbabilitiesMixin` via `LlamaModelBase`, so `compute_log_probabilities` is available.

One suspicious spot: `pipeline_variants/text_generation.py` appends one `batch_log_probabilities` entry **per decode step**, but `update_context_and_prepare_responses` (`pipeline_variants/utils.py`) consumes only `batch_log_probabilities[0]` and its docstring says it expects a single step — with multi-step decode every step after the first would silently lose its log probabilities. That may not be the only issue (it would explain partial rather than fully-null arrays).

Happy to help validate a fix — I have a selective-classification use case (label-token logprob gating) blocked on this.

Assisted-by: AI

msaelices:
## Summary

Fixes #6676. Three chained breaks left logprob responses empty (null/empty arrays) for any architecture whose tokenizer is vision-capable — Qwen3VL, Qwen3.5, and anything on `TextAndVisionTokenizer` — even for text-only requests:

1. `Qwen3VLTokenizer.new_context` did not forward `request.logprobs`/`request.echo` into the context, so `TextGenerationInputs.enable_log_probs` stayed `False` and the pipeline never computed log probabilities (traced live: router emits `logprobs=1`, scheduler sees `top=[0]`).
2. `TextAndVisionTokenizer.new_context` has the same omission.
3. Once logprobs flow, `TextAndVisionTokenizer.decode` crashes with `AttributeError: 'int' object has no attribute 'tolist'` — logprob assembly decodes one bare-int token id at a time, which the text tokenizer already guards.

A follow-up commit (from adversarial self-review) extends the same two fixes to the remaining multimodal tokenizers that drop the fields — **gemma4, idefics3, kimik2_5 and qwen2_5vl** (gemma4/idefics3 also lacked the bare-int decode guard) — and adds a regression test next to the existing `chat_template_options` field-drop test for the same function (`max/tests/integration/architectures/qwen3_5/test_tokenizer.py`).

## Verification

`Qwen/Qwen3.5-4B` on an A10G with `--no-enable-overlap-scheduler --force`: `/v1/completions` with `"logprobs": 1` now returns populated `token_logprobs`/`top_logprobs` (e.g. `[-0.209, -0.019]` for a one-word answer); previously the arrays came back null. Downstream use case (label-logprob gating for a classifier) produces a clean coverage/precision curve (100% precision at p≥0.95 on a 46-call validation set).

Known remaining gaps, intentionally out of scope: the `tokens` array in the completions logprobs payload is still null (assembly fills only `token_logprobs`/`top_logprobs`), and #6677 (non-overlap scheduler changes generation at temperature 0) affects anything measured in this mode.

Assisted-by: AI

### DIFF
--- a/max/python/max/pipelines/architectures/gemma4/tokenizer.py
+++ b/max/python/max/pipelines/architectures/gemma4/tokenizer.py
@@ -290,7 +290,7 @@ class Gemma4Tokenizer(TextAndVisionTokenizer):
         return templated_message
 
     async def decode(
-        self, encoded: npt.NDArray[np.integer[Any]], **kwargs
+        self, encoded: npt.NDArray[np.integer[Any]] | int, **kwargs
     ) -> str:
         """Decode tokens, preserving tool-related special tokens.
 
@@ -298,6 +298,10 @@ class Gemma4Tokenizer(TextAndVisionTokenizer):
         to selectively preserve them when skip_special_tokens=True by filtering
         unwanted special tokens before decoding.
         """
+        # Log-probability responses decode one token id (a plain int) at a
+        # time; match the text tokenizer's handling.
+        if isinstance(encoded, int):
+            encoded = np.array(encoded)
         skip_special_tokens = kwargs.get("skip_special_tokens", True)
 
         if not skip_special_tokens:
@@ -519,6 +523,8 @@ class Gemma4Tokenizer(TextAndVisionTokenizer):
             json_schema=json_schema,
             grammar=grammar,
             grammar_state=grammar_state,
+            log_probabilities=request.logprobs,
+            log_probabilities_echo=request.echo,
             sampling_params=request.sampling_params,
             images=image_metadata,
             vision_token_ids=self.vision_token_ids,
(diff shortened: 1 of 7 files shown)


k-w-w (approved):
Thanks for the fixes!

k-w-w:
!sync

PR #6699, opened 2026-06-19
[Pipelines] Load compressed-tensors NVFP4 Gemma 4 checkpoints

msaelices:
## Summary
Adds support for loading **compressed-tensors** NVFP4 Gemma 4 checkpoints (e.g. `RedHatAI/gemma-4-31B-it-NVFP4`), in addition to the existing modelopt NVFP4 path.

compressed-tensors NVFP4 and modelopt NVFP4 are the **same block format** — packed E2M1 weights with FP8-E4M3 block scales over 16-wide groups. They differ only in:

1. **HF config schema** — compressed-tensors uses `quant_method: "compressed-tensors"` with `config_groups.group_0.weights`.
2. **Tensor names** — `weight_packed` / `weight_global_scale` / `input_global_scale` vs modelopt's `weight` / `weight_scale_2` / `input_scale` (the per-block `weight_scale` matches).
3. **Global-scale convention** — compressed-tensors stores the **reciprocal**: llm-compressor uses `(FP4_MAX * FP8_MAX) / amax`, while modelopt uses `amax / (FP4_MAX * FP8_MAX)`.
4. **Global-scale shape** — `[1]` vs scalar `()`.

## Changes
- **`quant.py`** — `_parse_compressed_tensors_float4_config` parses the compressed-tensors schema into the same `QuantFormat.NVFP4` `QuantConfig` as modelopt (group_size 16, FP8-E4M3 block scales). Verified against the real checkpoint: gemma4 passes `ignored_modules_prefix="model.language_model."`, only `lm_head`+vision are ignored, so all text layers are correctly marked quantized.
- **`gemma4/weight_adapters.py`** — reconcile the compressed-tensors tensors onto the modelopt names the quantized `Linear` registers:
  - `weight_packed → weight`, `weight_global_scale → weight_scale_2`, `input_global_scale → input_scale`;
  - **invert** the global scales (the NVFP4 kernels *multiply* by `weight_scale_2` / `input_scale`) and squeeze `[1] → ()`.
  - modelopt checkpoints keep their names/scalars and are left untouched.

## Why the inversion matters
Without it the model loads and runs, but every weight is off by ~`weight_scale_2²`, producing gibberish. Confirmed numerically on layer 0 of `RedHatAI/gemma-4-31B-it-NVFP4`: `weight_global_scale = 10304` vs the matching modelopt `weight_scale_2 = 9.74e-5`; `input_global_scale = 33.5` vs `input_scale = 0.03125`.

## Validation
- New no-network unit test `test_nvfp4_weight_adapter.py` asserts the remap, the reciprocal, the `[1]→()` squeeze, and that the modelopt path is unchanged.
- End to end on an L40S: `RedHatAI/gemma-4-31B-it-NVFP4` answers *"The capital of France is Paris."* at ~28 tok/s, 9.1 GiB KV cache. (On a pre-Blackwell GPU this also exercises #6668; the loader changes here are hardware-independent.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


msaelices:
recheck

PR #6708, opened 2026-06-22
[Kernels][GPU] NVFP4 GEMM on the multistage qGEMM skeleton (~1.8-2.9x serving vs bespoke)

msaelices:
## Summary

NVFP4 (E2M1) GEMM built on MAX's mature multistage quantized mainloop
(`multistage_mma_q`) instead of the bespoke `nvfp4_gemm`, with the weight
repacked **once at load** and routed into pre-Blackwell serving behind
`MAX_NVFP4_SKELETON_GEMM=1` (default off). **Draft.**

### Depends on / blocked by (must merge first)

Stacked on the Gemma-4 NVFP4 series; lands after all three, then rebases onto
`main` (overlap disappears). The new commits here are the skeleton port itself
(~21, from `NVFP4 (E2M1) path on the multistage qGEMM skeleton` onward).

- **#6668** — NVFP4 fallback for pre-Blackwell NVIDIA GPUs (fused dequant-GEMV) — the bespoke path this offers a faster alternative to. *Partially included (earlier snapshot; rebase needed).*
- **#6698** — Right-size Gemma 4 activation-memory reservation. *Fully included (ancestor of this branch).*
- **#6699** — Load compressed-tensors NVFP4 Gemma 4 checkpoints — the loader this extends with the load-time repack. *Partially included (earlier snapshot).*

## Results (L40S / sm_89, gemma-4-31B-it-NVFP4)

**Batched serving throughput** (tok/s, 256 output tokens):

| workload | before (bespoke `nvfp4_gemm`) | after (skeleton + load-time repack) | speedup |
|---|---|---|---|
| 16 concurrent | ~118 | ~347 | **~2.9x** |
| 64 concurrent | ~445 | ~648 (maxlen 1024, dmu 0.84) / ~815 (maxlen 2048) | **~1.5–1.8x** |

**vs vLLM** (~1340 tok/s batched, same model/GPU, prior external measurement):
pre-Blackwell NVFP4 serving moves from **~33%** of vLLM (bespoke) to
**~48–61%** (this PR).

**Kernel-level** (microbench, manual wall-clock):

| metric | before (bespoke) | after (skeleton) |
|---|---|---|
| NVFP4 GEMM, M=482, gemma4 shapes | ~66 TFLOP/s | **~167 TFLOP/s** (~2.5x) |
| cold-weight HBM bandwidth, M=64 (decode) | ~21–31% of peak | **~65–76% of peak** |

group=16 / group=32 match an E2M1 dequant reference within bf16 rounding (loose tol); GGUF Q4 unregressed.
Load: ~10min → ~2.8min (parallelized host repack).

## What changed

- `TensorCore.load_b_nvfp4` (Marlin bit-trick decode) + comptime `is_nvfp4` flag (off ⇒ int4 path byte-identical).
- Fixed `num_scales_stages` for group < BK (was counting groups, not packed stages → stale scale at group=16).
- Load-time repack in the weight adapter: combined buffer becomes the sole registered weight (the original isn't uploaded), avoiding the in-graph per-step repack (~52% of decode) and the 2x-weight OOM; byte-exact numpy host replica of the GPU repack.
- `repack_nvfp4_g16` / `qmatmul_nvfp4_g16` custom ops + `_matmul_float4` routing; per-M autotuned configs (split-K).

## Validation

Byte-exact host repack vs the GPU kernel (4 shapes); engine-level skeleton-vs-bespoke
equivalence on identical weights; real 31B serving (original freed, coherent generation
matching bespoke, throughput as above). Flag-off path unchanged.

## Caveats

Combined weight is ~2GB larger (inline bf16 scales) ⇒ batch-64/maxlen-2048 is
memory-marginal on 46GB (use `--device-memory-utilization` / `--max-length`). One-time
~2.8min host repack at load (disk-caching is a follow-up). Remaining gap to vLLM is the
decode step beyond the FP4 GEMM (attention/KV/scheduler), out of scope.

## Test plan

- `bazel test //max/kernels/test/gpu/quantization:{test_nvfp4_skeleton_gemm,test_nvfp4_repack_e2e}.mojo.test`
- `//max/tests/integration/nn:nn_gpu_tests -k nvfp4_skeleton` (engine equivalence) and `-k repack_host` (byte-exact)
- `MAX_NVFP4_SKELETON_GEMM=1 max serve --model RedHatAI/gemma-4-31B-it-NVFP4`

---

_Assisted by: AI._


msaelices:
## Adversarial review — NVFP4 skeleton port

Reviewed the **skeleton-port net diff only** (`726edea48f^..HEAD`, ~4.5k lines / 18 files), not the stacked ancestors (#6668/#6698/#6699). Two independent adversarial passes (kernel/op side and Python/host-repack/wiring side), each told to assume every claim is wrong until the code proves it.

**Verdict: no confirmed data-corruption bug in the net diff.** The five load-bearing claims hold up under scrutiny, with a few precondition/wording/guard items worth fixing before un-drafting. Severity tags: 🔴 fix-before-merge · 🟡 concern · ⚪ nit.

### Claims verified
- ✅ **group<BK scale staging** — the `num_scales_stages = ceildiv(num_pipeline_stages-1, max(group//BK,1))+1` fix is correct *and general* for group<BK (not just g16): ring depth == pipeline depth, prefetch writes the correct oldest slot, SMEM sizing uses the same formula (no overrun), and the stage→k-sub-tile pairing is off-by-one-clean across the tile boundary.
- ✅ **Marlin E2M1 decode + 2^14 fold through the staged path** — bit-exact for all 16 codes incl. {0, ±0.5} and signed zero; the e==0 case rides through as an fp16 denormal; the scale product is computed in f32 before the bf16 cast.
- ✅ **flag-off ⇒ int4/GGUF byte-identical** — every new behavior is behind `comptime if is_nvfp4` / the `_NVFP4_USE_SKELETON_GEMM` env gate (default off; only exact `"1"` enables); for GGUF group≥BK the new loops degenerate to the original single-tile copy and the new stages formula equals the old whenever `group % BK == 0`. (One exception below.)
- ✅ **byte-exact host repack** — traced `_nvfp4_repack_host.py` step-by-step against `repack_nvfp4_for_sm8x`/`pack_Q_tile`: nibble/bit layout, `distribute[row_major(8,4)]` thread→element mapping, nested destination layout, scale permutation (`Layout((4,8),(16,1))`), fp8-e4m3 decode, RNE bf16 round, region order/sizes — all match.
- ✅ **original weight freed / no double-register** — combined uint8 buffer replaces the weight in the state dict, scale tensors are skipped, `Linear` registers only the combined buffer with `weight_scale=weight_scale_2=None`; no dangling ref, no GPU upload of the original. Parallel host repack is race-free (inputs materialized in the main thread; per-task output buffers).
- ✅ **no new SMEM/cp.async race** — the scale staging reuses the proven int4 discipline (`async_copy_wait_group` + block-wide `barrier()` before the cross-warp read); the bespoke-path WAR race does not recur.

### Fix before merge
- 🔴 **Unguarded partial-tile precondition** (`qmatmul_gpu.mojo`): no `comptime assert` for `N % BN == 0` / `K % BK == 0` (GEMM) nor `N % BN == 0` / `K % WGROUP == 0` (repack). `repack_nvfp4_for_sm8x` has no N%BN masking and the weight `copy_dram_to_sram` is unguarded against K%1024 (only the write is guarded). Holds for gemma4 shapes (N mult of 128/256, K mult of 1024) but will OOB-read on a non-conforming shape. → add the asserts. *(Confirm by running the repack on N%128≠0 / K%1024≠0 under `--config=asan`.)*
- 🟡 **"bit-exact" is overstated.** Both tests use loose tolerance (atol 1e-4…1e-3) AND the references bake in the *same* bf16 fold the kernel uses (scale fold at repack → product cast in decode = two bf16 roundings), so they cannot detect a fold-precision divergence from a true f32-accumulated reference. → soften the docstring/PR wording to "matches within bf16 rounding" and add one run against an independent f32/torch reference to bound the real error.
- 🟡 **Ungated default-path change** (`graph/weight.py:130-135`): the new single-device early return in `head_aware_col_sharding_strategy` affects *all* single-device callers, not just NVFP4, and is not behind the flag. Equivalent to the prior slice for the standard column shapes (so almost certainly a latent-bug fix), but it rides into a flagged PR. → guard it to the combined-buffer shape or call it out explicitly.

### Worth a guard (latent fragility, harmless today)
- 🟡 **e4m3fn scale-dtype gate vs MXFP4 e8m0 coercion** (`weight_adapters.py:89-90` vs `:180`): the skeleton fuse only matches `float8_e4m3fn` scales; an earlier pass coerces uint8 scales to `e8m0`. If an NVFP4 checkpoint ever delivered raw-uint8 block scales, the layer would be skipped by the fuse and mismatch the already-registered combined-buffer shape → **loud** load-time shape error (not silent). → assert `fused-prefixes == skeleton-Linears`, and load-test both modelopt and compressed-tensors (RedHatAI) NVFP4 variants under the flag.
- ⚪ `q_smem_usage` sizes the scale SMEM with `size_of[a_type]` not `size_of[scales_type]` — correct only because both are bf16 for NVFP4; an fp8-activation variant would under-allocate 2x → SMEM corruption. Use `scales_type` explicitly.
- ⚪ K-recovery `K = shape[1]//group_bytes*group_size` is a group=16-only algebraic coincidence (`5/8·16/10==1`). Add `comptime assert group_size == 16` on the NVFP4 entry (the op is already g16-hardcoded).
- ⚪ `tensor_core.mojo` `load_b_nvfp4` docstring says "assumes BK == group_size" — contradicts the group<BK purpose; stale, fix to avoid future misuse.

### Note on what the tests actually cover
The **active flagged path never calls the in-graph `repack_nvfp4_g16` op** — production repacks on the host (`nvfp4_repack_host`); the GPU repack op is exercised only by tests. So production correctness rests entirely on the host replica matching the GEMM's expected layout. Make sure `test_nvfp4_repack_e2e` diffs **host-output vs GPU-repack-output** (not host-to-host) on a real gemma4 shape — that one parity test is the single load-bearing safeguard for the whole load-time path.

---
*Adversarial review assisted by AI (two independent passes). Findings are code-trace based; the three items tagged "confirm by running" need a GPU run to settle.*

msaelices:
## Review follow-ups addressed (pushed in `8f8db2f955`)

Thanks for the two-pass adversarial review. No data-corruption bug ⇒ I addressed
the guards/clarity items:

- 🔴 **Partial-tile preconditions** — added `comptime assert`s: `matmul_gpu_nvfp4`
  now guards `N % 128 == 0` and `K % 32 == 0` (BK); `repack_nvfp4_for_sm8x` guards
  `group_size == 16`, `N % 64 == 0` (repack tile) and `K % WGROUP(128) == 0`. So a
  non-conforming shape fails at compile time instead of OOB-reading. (Note: the
  repack tiles in **128**-wide WGROUP chunks, not 1024 — all gemma4 K and the test
  shapes satisfy `K % 128 == 0`.)
- ⚪ **`q_smem_usage` scale dtype** — now sizes the scale SMEM by `bfloat16` (the
  kernel's `scales_type`) explicitly, so an fp8-activation variant can't under-allocate.
- ⚪ **`load_b_nvfp4` docstring** — corrected the stale "assumes BK == group_size"
  to document `group_size <= BK` via the `scale_sub` subgroup selection.
- 🟡 **"bit-exact" wording** — softened in the PR description to "matches within bf16
  rounding (loose tol)" for the dequant-reference tests. (The host-vs-GPU repack
  parity test stays genuinely **byte-exact** — `np.array_equal` — which is the
  single load-bearing safeguard you flagged, and it's green.)

### Acknowledged / notes

- 🟡 **`graph/weight.py` single-device early return** — kept, with an explicit
  comment: at `num_devices == 1` the head-aware math already returns the whole
  weight to device 0, so the early return is **behavior-preserving** for standard
  column shapes and only additionally handles the combined-buffer layout. Called
  out in-code.
- 🟡 **e4m3fn fuse gate vs MXFP4 e8m0** — the failure mode is a **loud** load-time
  shape error, not silent corruption; the byte-exact host-vs-GPU repack parity test
  guards the layout. A `fused-prefixes == skeleton-Linears` assert is reasonable
  future hardening.

### Re-validated after syncing `main` (toolchain bump) + these fixes

- Skeleton GEMM correctness (g16/g32): ~167 TFLOP/s, bit-close to the E2M1 reference.
- Byte-exact host repack vs the GPU kernel: pass (all shapes).
- Engine equivalence (skeleton vs bespoke): 3 passed.
- 31B serving @conc64: ~647 tok/s (≈ pre-merge; ~1.5x bespoke).

_Assisted by: AI._

Issue #6728, opened 2026-06-28
[BUG] Current 'Get Started with GPU Programming' crashes on compile.


maxz411:
### Actual behavior

Saw that there was another issue about this but has since been closed. 

Compiling the following code:

```
from std.sys import has_accelerator

from std.gpu.host import DeviceContext
from std.gpu import block_idx, thread_idx


def print_threads():
    """Print thread IDs."""

    print(block_idx.x, thread_idx.x, sep="\t")


def main() raises:
    comptime if not has_accelerator():
        print("No compatible GPU found")
    else:
        ctx = DeviceContext()
        print("block\tthread")
        ctx.enqueue_function[print_threads](
            grid_dim=2, block_dim=64
        )
        ctx.synchronize()
        print("Program finished")
```

as pasted directly from https://mojolang.org/docs/manual/gpu/intro-tutorial/ raises an error.
```
(gpu-intro)  gpu-intro mojo vadd.mojo
~/code/gpu-intro/vadd.mojo:1:1: error: Metal Compiler failed to compile metallib. Please submit a bug report.
from std.gpu.host import DeviceContext
^
mojo: error: failed to run the pass manager
```

This is on Apple M4 Max with 64 GB and macOS Tahoe 26.5.1

### Expected behavior
As per the docs, I should get:

```
block	thread
1	32
1	33
1	34
...
0	30
0	31
Program finished
```

### Steps to reproduce

https://mojolang.org/docs/manual/gpu/intro-tutorial/

Following the exact instructions here on M4 Max consistently yields this error for me.

### System information

(gpu-intro) ➜  gpu-intro pixi info
System
------------
       Pixi version: 0.71.1
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=26.5.1=0
                   : __archspec=1=m4
          Cache dir: /Users/max/Library/Caches/rattler/cache
       Auth storage: /Users/max/.rattler/credentials.json
   Config locations: No config files found

Global
------------
            Bin dir: /Users/max/.pixi/bin
    Environment dir: /Users/max/.pixi/envs
       Manifest dir: /Users/max/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: gpu-intro
            Version: 0.1.0
      Manifest file: /Users/max/code/gpu-intro/pixi.toml
       Last updated: 28-06-2026 04:54:52

Environments
------------
        Environment: default
           Features: default
[...]

PR #6730, opened 2026-06-29
[Docs] Add macOS Metal toolchain tip to GPU intro tutorial

maxz411:
## Summary

Adds a `:::tip` admonition to the [Get started with GPU programming](https://docs.modular.com/mojo/manual/gpu/intro-tutorial) tutorial, placed right below the complete `vector_addition.mojo` example.

Apple silicon users can hit the following error when compiling the example:

```
error: Metal Compiler failed to compile metallib. Please submit a bug report.
from std.gpu.host import DeviceContext
^
```

The tip points them to the fix:

```bash
xcodebuild -downloadComponent MetalToolchain
```

## Motivation

Relates to #6728. A maintainer noted this is "a frequent issue with targeting Apple silicon GPUs" and suggested making the fix more prominent in the docs — this PR does that by surfacing it directly in the tutorial where users hit the error.

## Notes

- Docs-only change; no code affected.
- Passes `./bazelw run format` (rumdl/lint: "No issues found").

🤖 Generated with [Claude Code](https://claude.com/claude-code) - Claude helped me with MDX formatting and stuff.

### DIFF
--- a/mojo/docs/manual/gpu/intro-tutorial.mdx
+++ b/mojo/docs/manual/gpu/intro-tutorial.mdx
@@ -850,6 +850,25 @@ def main() raises:
 
 </details>
 
+:::tip macOS
+
+If you're on Apple silicon and getting:
+
+```output
+error: Metal Compiler failed to compile metallib. Please submit a bug report.
+from std.gpu.host import DeviceContext
+^
+mojo: error: failed to run the pass manager
+```
+
+Please try:
+
+```bash
+xcodebuild -downloadComponent MetalToolchain
+```
+
+:::
+
 The `enqueue_function()` method enqueues the compilation and invocation
 of the `vector_addition()` kernel function, passing the input and output tensors
 as arguments. The `grid_dim` and `block_dim` arguments use the `num_blocks` and


maxz411:

KenJones-Modular:
Thanks for identifying the problem and contributing a suggested fix for this, @maxz411. I'm going to dig into this a bit further this week to determine:

- if there's a better location in the tutorial to place the callout
- if there are other failure cases that we should highlight as well
- how we can improve the requirements documentation at the start of the tutorial

So even if we don't accept your PR as-is, rest assured that we'll try to address your root concern as soon as possible.

maxz411:
> Thanks for identifying the problem and contributing a suggested fix for this, @maxz411. I'm going to dig into this a bit further this week to determine:
> 
> 
> 
> - if there's a better location in the tutorial to place the callout
> 
> - if there are other failure cases that we should highlight as well
> 
> - how we can improve the requirements documentation at the start of the tutorial
> 
> 
> 
> So even if we don't accept your PR as-is, rest assured that we'll try to address your root concern as soon as possible.

Thanks! Thank you for building Mojo by the way, it is truly amazing work!

PR #6754, opened 2026-07-07
[Kernels][GPU] Add native NVFP4 tensor-core GEMM for consumer Blackwell (sm_120 / sm_121)

lightofbaldr:
> **Stacked on #6596.** This branch builds on #6596 (route `block_scaled_matmul` to the naive kernel on consumer Blackwell), so the diff below includes its three commits. The change introduced by *this* PR is the final commit only — `[Kernels][GPU] Add native NVFP4 tensor-core GEMM for consumer Blackwell` — which adds `max/kernels/src/linalg/nvfp4_tensorcore.mojo` + its test. Please review/merge #6596 first; GitHub will retarget this PR to `main` automatically once it lands. Reviewing that single commit shows this PR's contribution in isolation.

---

Adds a warp-level block-scaled FP4 tensor-core GEMM using the
`mma.sync.aligned.m16n8k64.row.col.kind::mxf4nvf4.block_scale.scale_vec::4X`
instruction available on consumer Blackwell (RTX 50xx / DGX Spark GB10). This is
the performant alternative to the arch-agnostic naive CUDA-core path that #6596
routes to when the SM100 warp-specialized kernel is unavailable: on a GB10 it
reaches ~37 TFLOP/s at 2048³ (single-warp-per-tile, no shared-memory staging
yet), roughly 50× the naive path, and is bit-exact versus a host block-scaled
reference.

### Scale-factor layout

Scales are read from MAX's rank-5 interleaved SF layout
`[ceil(rows/128), ceil(K/64), 32, 4, 4]` (e4m3) via `sf_rank5_off`, which
reproduces the `set_scale_factor` mapping in `fp4_utils.mojo`. A K-group's four
e4m3 scales are contiguous, so they load as one `UInt32` for the MMA's
`scale_vec::4X` per-lane distribution.

### Requirements

Requires target `sm_121a` (plain `sm_121` yields `INVALID_PTX`) and a CUDA 13.1+
driver: the MMA emits PTX 9.1, which assembles on CUDA 13.0 but fails to load
there with `CUDA_ERROR_UNSUPPORTED_PTX_VERSION`. The kernel is provided as a
callable module in this PR; wiring it into the `block_scaled_matmul` dispatch
(behind driver detection, with the naive path as fallback) is a deliberate
follow-up.

### Verification

Known-answer test (every FP4 nibble decodes to 1.0 and every e4m3 scale is 1.0,
so `C[i, j] == K`) passes bit-exact (`bad == 0`) at 64³, 512²·256, 256×128×192,
and 2048³ on **both** consumer-Blackwell arches:

| Arch | GPU | Result |
|------|-----|--------|
| sm_120a | RTX 5050 | bit-exact, all shapes |
| sm_121a | DGX Spark GB10 | bit-exact, all shapes |

The test is gated to consumer-Blackwell hardware (`nvidia_gpu` + `manual`; see
`BUILD.bazel`) — the FP4 decode/dispatch comptime asserts fail to build on
H100 / A100 / AMD, so it cannot run in standard CI. Run locally with
`--test_tag_filters=` and `--local_resources=gpu-memory=<GiB>`.


max/kernels/test/gpu/linalg/test_nvfp4_tensorcore.mojo
```diff
@@ -0,0 +1,104 @@
+# ===----------------------------------------------------------------------=== #
```
BradLarson, line 1: This test won't be compatible with our internal B200 CI systems. Instead of adding a new test, can't this be tested by paths that already exercise the NVFP4 B200 matmuls?

lightofbaldr:
Closing in favor of #6596. Per the review guidance, the merge path for NVFP4 on consumer Blackwell (sm_120 / sm_121) is routing to the existing cuBLASLt vendor fallback with minimal comptime gates rather than a native/custom kernel. #6596 now does exactly that and is verified end-to-end on GB10 (sm_121a, CUDA 13.4). The native tensor-core GEMM here stays on the branch as a possible follow-up optimization once the cuBLAS path lands.

PR #6761, opened 2026-07-10
[Mojo] Fix Mojo importer redirected cache isolation across Pixi environments

### LINKED ISSUE #6534: [BUG]  libKGENCompilerRTShared.so: cannot open shared object file
### Actual behavior

When using multiple environment in pixi + using the python module `mojo.importer`, we get the following error:

```
      from causal_conv1d_mojo.fwd import dispatch
  ImportError: libKGENCompilerRTShared.so: cannot open shared object file: No such file or
  directory
```

### Expected behavior
It should work

### Steps to reproduce

Since this require pixi and multiple environments i made a small repo to reproduce with nearly empty source files:

https://github.com/gabrieldemarmiesse/modular-runpath-bug

```bash
git clone https://github.com/gabrieldemarmiesse/modular-runpath-bug.git
cd modular-runpath-bug

# 1. Compile in 'extra' env → RUNPATH baked to .pixi/envs/extra/...
pixi run -e extra python main.py

# Notice the absolute path here
readelf -d hello/__mojocache__/*.so | grep RUNPATH

# 2. Routine clean reinstall — only the default env comes back.
rm -rf .pixi && pixi install

# 3. Use the cache from default env. RUNPATH points to nonexistent
#    .pixi/envs/extra/.../modular/lib → ImportError.
pixi run python main.py
```

none of those steps are unusual. Deleting the pixi environment is common to force a reinstall or clean up the repo.

### System information

mojo-compiler==1.0.0b1

mahendrarathore1742:
### PR description

1. Fixes Mojo importer redirected cache handling when __mojocache__ is read-only
2. Redirected cache paths now include an environment-specific identifier so artifacts built in one Pixi environment are not reused in another
3. Prevents stale .so files with invalid absolute RUNPATH from causing libKGENCompilerRTShared.so import failures
4. Adds regression coverage:-
    - unit test for [_resolve_cache_dir](vscode-file://vscode-app/home/msr/Downloads/code-stable-x64-1769004726/VSCode-linux-x64/resources/app/out/vs/code/electron-browser/workbench/workbench.html) env-specific redirected caches
    - integration test verifying two read-only Mojo imports use separate redirected cache directories

### Validation

1. python -m py_compile [importer.py](http://_vscodecontentref_/1) mojo/python/test/test_importer.py
2. python -m py_compile mojo/integration-test/python-extension-modules/readonly-cache/test_module.py

close #6534


dgurchenkov:
@Ahajha this looks like a proper bug fix in Mojo, do you agree? Can you please decide if to merge it or close (or maybe we have a different fix).

0xamlab (commented):
**Verdict up front:** the env-keying mechanism this PR adds to the *redirected* cache branch works as designed (verified end-to-end below), but the PR in its current form (1) does **not** fix the scenario reported in #6534, which uses a *writable* in-tree `__mojocache__`, (2) breaks one pre-existing integration test and its own new integration test — CI on this PR has been red since 2026-07-10, and (3) needs a rebase over the `mojo/` → `Mojo/` tree move (`015ae3d280`, 2026-09-02), which is why the PR currently shows as unmergeable. Details and reproduction evidence below.

> Disclosure: I work with thyn-ai. This review was prepared with AI assistance and verified/edited by a human before posting, per the [AI Tool Use Policy](https://github.com/modular/modular/blob/main/AI_TOOL_POLICY.md). No part of this was posted automatically.

## What I tested

I could not find any prior review here, so I ran the full matrix myself:

- **Reporter's exact repro** ([gabrieldemarmiesse/modular-runpath-bug](https://github.com/gabrieldemarmiesse/modular-runpath-bug), `mojo-compiler==1.0.0b1`, linux/amd64, pixi 0.81.0, containerized ubuntu:24.04 + gcc).
- **Current toolchain**: two pixi envs (`default`/`extra`, solve-group shared) on `mojo-compiler 1.2.0.dev2026092005` from `max-nightly` (matches the pin on current `main`, `579c250800`), linux/arm64, as a non-root user so the read-only detection in `_cache_dir_is_writable` works.
- PR head `391221fe26` applied by overlaying the patched `importer.py` into the installed `mojo` package; the PR's `test_importer.py` and `readonly-cache/test_module.py` run with pytest against that package.

## 1. The bug is real and still present on current `main`

Reporter's repro on 1.0.0b1, step by step:

```
$ pixi run -e extra python main.py          # compiles + caches
dispatch.hello() -> hello from mojo
$ readelf -d hello/__mojocache__/dispatch.hash-08bd64fdd471af19.so
(NEEDED)  Shared library: [libKGENCompilerRTShared.so]
(RUNPATH) Library runpath: [/work/repro/.pixi/envs/extra/lib/python3.13/site-packages/modular/lib]
$ rm -rf .pixi && pixi install              # only the default env comes back
$ pixi run python main.py
ImportError: libKGENCompilerRTShared.so: cannot open shared object file: No such file or directory
```

Same result on the nightly (`1.2.0.dev2026092005`), with RUNPATH `/work/envtest/.pixi/envs/extra/lib` (conda layout). Root cause, for the record: `MojoImporter.find_spec` keys the cache filename solely on a hash of the `.mojo` sources (`dispatch.hash-<source_hash>.so`), while the artifact's `DT_NEEDED`/`DT_RUNPATH` are absolute per-environment. Any second environment that hits the same cache reuses the foreign artifact, and the loader fails as soon as the original env's path disappears.

Two linker details worth noting:

- The baked entry is `DT_RUNPATH` (not `DT_RPATH`), so `LD_LIBRARY_PATH` is searched *before* it. The reporter's `LD_LIBRARY_PATH=$CONDA_PREFIX/lib` activation didn't help on the PyPI wheel layout because the library lives at `$CONDA_PREFIX/lib/python3.13/site-packages/modular/lib`, not `$CONDA_PREFIX/lib`. (On the conda layout, where it *is* `$CONDA_PREFIX/lib`, that activation would have masked the bug.)
- `$ORIGIN`-relative RUNPATH (the auditwheel/delocate pattern suggested in the repro README) can't fix the general case: the cached artifact lives in an arbitrary user source tree, so its relative path to the env's lib dir is not fixed. `$ORIGIN` works for vendored, fixed-layout wheels, not for a cache outside the env.

## 2. The PR does not fix the reporter's scenario (writable in-tree cache)

The env-id is only added on the **redirected** branch of `_resolve_cache_dir` (read-only source tree → Modular cache folder). The reporter's repro takes the **in-tree** branch (writable `__mojocache__`), which is unchanged. Verified on both toolchains — patched importer, stale in-tree cache still present:

```
1.0.0b1, pixi run python main.py            -> ImportError: libKGENCompilerRTShared.so (exit 1)
nightly,  pixi run python main.py           -> ImportError: libKGENCompilerRTShared.so (exit 1)
```

Deleting `__mojocache__` and re-running succeeds, confirming the failure is stale-artifact reuse, not the patch. So as-is, merging this would close #6534 without fixing its repro.

## 3. The PR's mechanism does work for its actual target (redirected cache)

Read-only source tree, shared `MODULAR_CACHE_DIR`, two envs, nightly:

- **Unpatched**: extra env compiles to `$MODULAR_CACHE_DIR/.mojo_cache/python_extensions/hello/dispatch/...` (RUNPATH → extra env). Remove the extra env, run from default → same `ImportError` (bug reproduces through the redirected path too).
- **Patched**: default env compiles its own artifact and imports fine (`dispatch.hello() -> hello from mojo`), under `python_extensions/76a577f8daed4c84/...` — and `76a577f8daed4c84` is exactly `sha256("/work/envtest/.pixi/envs/default")[:16]`, i.e. the env id is derived as documented.

The approach (keying the shared redirected cache by install root) is sound, deterministic, and cheap. PR unit tests: **5/5 pass** against the patched nightly package.

## 4. CI is red on this PR's own tests — root-caused

The 2026-07-10 `large-oss-linux` run failed `//mojo/integration-test/python-extension-modules/readonly-cache:main`, and I reproduce the same 2 failures locally (2 failed, 1 passed):

1. `test_read_only_tree_auto_redirects_without_config` (pre-existing): asserts the artifact under `python_extensions/<name>`; with the env-id segment it now lands under `python_extensions/<env_id>/<name>`. The PR doesn't update this assertion.
2. `test_read_only_tree_redirected_cache_is_env_specific` (new in this PR): the expected path is missing the `.mojo_cache` component — `_modular_cache_root()` returns `$MODULAR_CACHE_DIR/.mojo_cache`, so artifacts land under `<cache_dir>/.mojo_cache/python_extensions/<env_id>/<name>`, not `<cache_dir>/python_extensions/<env_id>/<name>`.

The PR description's validation was `python -m py_compile ...` only — actually running pytest would have caught both. Note the unit tests pass because they mock `_modular_cache_root()` directly.

## 5. Rebase needed

`main` moved `mojo/python/mojo/` → `Mojo/python/mojo/` (and the test dirs) on 2026-09-02. I diffed the PR's `importer.py` against current `main`'s: the only delta is this PR's own change, so the rebase is mechanical — but the file paths in the PR no longer exist upstream (`mergeable_state: dirty`).

## Suggestions

- To actually close #6534, the env/install identity (ideally including the compiler version, not just the path — see below) needs to key the **in-tree** cache too, e.g. folded into `cache_filename` next to the source hash. Alternative: resolve the runtime library from the *current* env at import time (preload `libKGENCompilerRTShared` via `ctypes.CDLL(..., RTLD_GLOBAL)` from `get_package_root()` before the `ExtensionFileLoader` runs — glibc then satisfies the `DT_NEEDED` by soname from the loaded object and the stale RUNPATH never matters; macOS dyld semantics differ and would need care). Or re-scope this PR to the redirected branch with an honest "partially addresses #6534" linkage and a follow-up issue for the in-tree branch.
- Residual gaps in the env-id itself, for awareness: it's keyed on the install *path*, so reinstalling a *different compiler version* at the same path reuses stale artifacts (the source-hash-only cache keying is a pre-existing instance of the same class); and old entries are orphaned with no GC (harmless, but the cache grows).
- macOS: this diff is pure Python path selection, so no dylib install-name/`LC_RPATH` regression surface; I did not test macOS behavior of the baked paths themselves.

Happy to re-run any of this on updated commits — the containerized repro/scripts are deterministic and quick (~4 min).

**Summary for maintainers:** direction is right for the redirected-cache case; needs (a) rebase over the `Mojo/` move, (b) the two integration-test fixes above, and (c) either extension to the in-tree cache branch or an explicit scope-down vs #6534.

mahendrarathore1742:
@0xamlab  Thank you for the detailed review and for root-causing the integration test failures!

    I have updated the branch:
    1. **Rebased onto latest `main`**, resolving the merge conflicts resulting from the `mojo/` -> `Mojo/` directory migration.
    2. **Fixed both integration test assertions** in `readonly-cache/test_module.py`:
       - Added the `env_id` segment in `test_read_only_tree_auto_redirects_without_config`.
       - Included the `.mojo_cache` component in `test_read_only_tree_redirected_cache_is_env_specific`.
    3. **PR Scope:** Re-scoped the PR description to *partially address #6534* for the redirected cache scenario in read-only source trees, leaving in-tree cache isolation for a follow-up.

PR #5637, opened 2025-12-07
[Stdlib] Add implicit Codepoint(StringLiteral) constructor

KrxGu:
Fixes #5490 

This adds an implicit constructor to `Codepoint` that accepts a compile-time `StringLiteral` parameter, making it much easier to create codepoint constants.

### What changed
- Added `__init__[value: __mlir_type.\`!kgen.string\`](out self, lit: StringLiteral[value])`
- Constructor validates at compile-time that the literal is exactly 1 byte
- Computes the codepoint value using `alias` for compile-time evaluation
- Added comprehensive tests covering ASCII, whitespace, and compile-time usage

### Why
Currently, the only way to create a `Codepoint` from a literal is `Codepoint.ord("x")`, which is verbose and can generate suboptimal IR. The stdlib has been working around this with patterns like:
```mojo
alias ` ` = UInt8(ord(" "))
alias `\t` = UInt8(ord("\t"))
```

Now you can just write:
```
var space = Codepoint(" ")
var tab = Codepoint("\t")
```

The constructor enforces the single-byte constraint at compile-time, so invalid literals like Codepoint("") or Codepoint("ab") will fail to build rather than causing runtime issues.

### Testing
./bazelw test //mojo/stdlib/test/collections:test_codepoint.mojo.test passes
Added test_char_from_stringliteral() with coverage for ASCII, whitespace, and compile-time contexts

PR #6762, opened 2026-07-10
[stdlib] Add Codepoint constructor from StringLiteral

### LINKED ISSUE #5490: [Feature Request] Implicit ctor Codepoint(StringLiteral)
### Review Mojo's priorities

### What is your request?

Adding an implicit Codepoint constructor with equivalent semantics to the following function:
```mojo
@always_inline("nodebug")
fn codepoint_from_string_literal[value: __mlir_type.`!kgen.string`](s: StringLiteral[value]) -> Codepoint:
    alias sl: StringLiteral[value] =  {}
    constrained[len(sl) == 1]()
    alias cp = Codepoint.ord(sl)
    return cp
```

Currently there is no safe, efficient and ergonomic way to construct a Codepoint.
Codepoint.ord exists but generates bloated llvm ir. So the only efficient way to construct a Codepoint is an alias statement. Leading to patterns like these: https://github.com/modular/modular/blob/53e05e781812f63b8011b03a9dcdc60bbffd7d77/mojo/stdlib/stdlib/collections/string/string_slice.mojo#L1905-L1913

### Any other details?

_No response_

odanilosalve:
Fixes #5490

Adds a `__init__` overload for `Codepoint` that takes a `StringLiteral`. The literal gets validated at compile time to contain exactly one Unicode codepoint, and the codepoint value is computed as a compile-time constant.

## Why

`Codepoint.ord("X")` works today but it's a runtime call, so you can't use it in a `comptime` context. With this overload you write `Codepoint("A")` directly, and can do `comptime c = Codepoint("X")`. Multi-byte UTF-8 like `"🔥"` works the same way.

## How

- Plain signature, no `@implicit`. ConnorGray's review on #5637 recommended dropping `@implicit` because implicit conversions on `Codepoint` would surprise more than help.
- Uses `comptime assert` (not `__comptime_assert` / `constrained[]`) — stdlib is moving to this.
- Validates UTF-8 by reusing `Codepoint.unsafe_decode_utf8_codepoint` on the literal's bytes.

## Context

PR #5637 (KrxGu) has been working on the same change since December but seems stalled. I incorporated the review feedback that's already accumulated (martinvuyk, NathanSWard, ConnorGray) and am opening fresh to keep things moving. Happy to coordinate if KrxGu wants to drive this to merge instead — I commented on both #5490 and #5637.

## Testing

Added `test_char_from_stringliteral` covering ASCII, whitespace, the null char, multi-byte UTF-8 (`🔥` → `0x1F525`), and `comptime` usage. Full `test_codepoint.mojo` passes (43/43) against a locally rebuilt stdlib.

### DIFF
--- a/mojo/stdlib/std/collections/string/codepoint.mojo
+++ b/mojo/stdlib/std/collections/string/codepoint.mojo
@@ -136,6 +136,62 @@ struct Codepoint(Comparable, ImplicitlyCopyable, Intable, Movable, Writable):
         """
         self._scalar_value = UInt32(Int(codepoint))
 
+    @always_inline("nodebug")
+    def __init__(out self, lit: StringLiteral):
+        """Construct a `Codepoint` from a single-codepoint `StringLiteral`.
+
+        This constructor validates at compile-time that the literal contains
+        exactly one Unicode codepoint, and computes the codepoint value as a
+        compile-time constant. This provides an ergonomic and efficient way to
+        create codepoints from string literals without runtime overhead.
+
+        Args:
+            lit: A string literal containing exactly one Unicode codepoint.
+
+        Constraints:
+            The string literal must contain exactly one Unicode codepoint.
+            Multi-byte UTF-8 sequences are supported (e.g., "🔥").
+
+        Examples:
+
+        ```mojo
+        from std.collections.string import Codepoint
+        from std.testing import assert_equal
+
+        # Create codepoints from ASCII literals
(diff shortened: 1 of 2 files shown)


odanilosalve:

barcharcraz:
!sync

barcharcraz:
This has been merged internally and should be reflected in the open source GitHub repo by tomorrow morning.

PR #6769, opened 2026-07-13
[Kernels][GPU] NVFP4: emit rank-5 activation scales on consumer Blackwell (sm_120/sm_121)

lightofbaldr:
### Summary
`quantize_dynamic_block_scaled` (`max/python/max/nn/kernels.py`) emits **rank-5** interleaved activation scales only for `_is_sm10x_gpu()` (B200); on consumer Blackwell (`sm_120`/`sm_121`) it falls to the **rank-2** else-branch, which then trips `dynamic_block_scaled_matmul`'s "both a_scales and b_scales must be rank 5" check — blocking NVFP4 end-to-end on consumer Blackwell.

`reroute_v2` taught the **Mojo kernel** to emit rank-5 activation scales for `sm_12x`, but the matching **Python graph op** stayed on the else-branch — this is the half that was removed during the #6596 cleanup. This restores it: one `_is_sm12x_gpu()` helper (mirroring `_is_sm10x_gpu()`) + including it in the rank-5 branch.

Contributes to #6570 (consumer Blackwell sm_120/sm_121 support); pairs with the reroute_v2 kernel side.

### Verification (real hardware, all three points)
- **`sm_121a` (DGX Spark GB10)** — without this fix, `nvidia/Llama-3.1-8B-Instruct-NVFP4` fails with the rank-5 `ValueError` at the first NVFP4 layer.
- **B200** — the same model runs clean (367 tok/s, no rank error) → confirms the break is **`sm_12x`-specific**, not device-independent.
- **`sm_121a` with this fix** — the model now generates coherent text ("The sky appears blue to us because of…") at ~41 tok/s.

Because `_matmul_float4` is the path every NVFP4 layer takes, this single branch unblocks the whole forward.

### DIFF
--- a/max/python/max/nn/kernels.py
+++ b/max/python/max/nn/kernels.py
@@ -6552,6 +6552,14 @@ def _is_sm10x_gpu() -> bool:
         return False
 
 
+def _is_sm12x_gpu() -> bool:
+    """Checks if the current accelerator is NVIDIA SM120/SM121 (consumer Blackwell)."""
+    try:
+        return accelerator_architecture_name().startswith("sm_12")
+    except Exception:
+        return False
+
+
 def _is_apple_gpu() -> bool:
     """Checks if the current accelerator is an Apple (Metal) GPU."""
     try:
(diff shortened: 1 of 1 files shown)


BradLarson (approved):
Cool, yup I think this was the last missing piece for getting some basic models working. Thanks!

BradLarson:
!sync

BradLarson:
!sync

PR #6780, opened 2026-07-15
[bazel] Fix modular_shared_library for path mapping

keith:
The env vars we were using previously are nicer but are not compatible
with CppLink path mapping.


Ahajha:
!sync

PR #6786, opened 2026-07-17
[Stdlib] Add mul_no_contraction to prevent FMA contraction

### LINKED ISSUE #6785: [Feature Request] Allow disabling FMA contraction
### Review Mojo's priorities

### What is your request?

I would like an option to be able to disable FMA contraction when evaluating mathematical expressions. It appears that Mojo always lowers to FMA contractions when algebraically possible.

However, in some applications, this is not desirable. Some expressions may be numerically well-conditioned only without FMA contractions.

Also, some HPC simulation codes must have near-perfect geometric symmetry, which is broken by FMA contractions in expressions like a*x + b*y, since an FMA contraction will break the symmetry between the x and y coordinates.

### Any other details?

I have constructed a temporary workaround using inline asm:
```
  def rounded_multiply(a: Float64, b: Float64) -> Float64:
      return inlined_assembly[
          "mul.rn.f64 $0, $1, $2;",
          Float64,
          constraints="=d,d,d",
          has_side_effect=False,
      ](a, b)
```

mahendrarathore1742:
Adds mul_no_contraction() to std.math for users who need separate multiply and add operations without FMA contraction. This is needed for HPC codes requiring geometric symmetry and numerical precision where FMA contraction would break symmetry (e.g., a*x + b*y).

Usage:
  from std.math import mul_no_contraction
  var result = mul_no_contraction(a, b) + c

Fixes #6785


zgarapetian:
@mahendrarathore1742 Apologies for the late followup here, but could you please update the PR if you are working on it still?

mahendrarathore1742:
> @mahendrarathore1742 Apologies for the late followup here, but could you please update the PR if you are working on it still?

okay i'll update the PR

zgarapetian:
@mahendrarathore1742 I just checked the referenced issue and see its been closed. Thanks for taking a look at this and if you feel your addition is still needed please file an issue for it and feel free to reopen this PR referencing the new issue

PR #6788, opened 2026-07-18
[Kernels] Fix static conv width-padding boundaries

### LINKED ISSUE #6712: [BUG]: `conv2d` on CPU returns silently wrong / nondeterministic output with large width-padding
### Bug description

Assisted by: Claude

On the CPU device, `ops.conv2d` silently returns incorrect and nondeterministic output when the convolution has large width-padding on a `permute`d (non-contiguous) NHWC input.

This shows up with the common "conv1d via conv2d" idiom: take an `(N, C, L)` signal, `permute` it to NHWC, convolve with a `(1, k)` kernel, and `permute` back. A dilated/transposed conv built this way needs `padding = k-1`, which is exactly the large width-padding that triggers the bug. It is **not** specific to `(1, k)` kernels — a `(3, 3)` kernel with large width-padding fails too. Small padding, H-padding, and native (contiguous) NHWC inputs are unaffected.

The same convolution oriented on the **H axis** (a `(k, 1)` kernel with H-padding) is stable and correct, this is a usable workaround.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = ["modular==26.4", "numpy"]
# ///
"""conv2d on CPU with large width-padding on a permuted NHWC input (the
conv1d-via-conv2d idiom). Compares against a numpy reference (correctness) and
runs the same compiled graph 20x (determinism). Both error and spread should be 0."""
import numpy as np
from max.driver import CPU
from max.dtype import DType
from max.engine import InferenceSession
from max.graph import DeviceRef, Graph, TensorType, ops

DEV = DeviceRef.CPU()
CIN, COUT, K, W = 3, 2, 8, 16
PAD = K - 1
W_KERNEL = np.random.default_rng(0).standard_normal((COUT, CIN, K)).astype(np.float32)
X = np.random.default_rng(1).standard_normal((1, CIN, W)).astype(np.float32)


def reference():
    xp = np.pad(X, ((0, 0), (0, 0), (PAD, PAD)))
    out = np.zeros((1, COUT, xp.shape[-1] - K + 1), np.float32)
    for o in range(out.shape[-1]):
        out[:, :, o] = np.einsum("ncs,fcs->nf", xp[:, :, o:o + K], W_KERNEL)
    return out


def build(kernel_on_h: bool):
    w = W_KERNEL
    with Graph("g", input_types=[TensorType(DType.float32, (1, CIN, W), device=DEV)]) as g:
        (t,) = g.inputs
        if kernel_on_h:   # (k,1) kernel, H-padding -> correct + deterministic (workaround)
            x = ops.permute(ops.unsqueeze(t.tensor, 3), [0, 2, 3, 1])   # (1,W,1,C)
            wr = np.ascontiguousarray(np.transpose(w[:, :, :, None], (2, 3, 1, 0)))
            y = ops.conv2d(x, ops.constant(wr, DType.float32, device=DEV), padding=(PAD, PAD, 0, 0))
            g.output(ops.squeeze(ops.permute(y, [0, 3, 1, 2]), 3))
        else:             # (1,k) kernel, W-padding -> wrong + nondeterministic (BUG)
            x = ops.permute(ops.unsqueeze(t.tensor, 2), [0, 2, 3, 1])   # (1,1,W,C)
            wr = np.ascontiguousarray(np.transpose(w[:, :, None, :], (2, 3, 1, 0)))
            y = ops.conv2d(x, ops.constant(wr, DType.float32, device=DEV), padding=(0, 0, PAD, PAD))
            g.output(ops.squeeze(ops.permute(y, [0, 3, 1, 2]), 2))
    return g


def check(name, kernel_on_h):
    m = InferenceSession(devices=[CPU()]).load(build(kernel_on_h))
    outs = [m.execute(X)[0].to_numpy() for _ in range(20)]
    err = float(np.abs(outs[0] - reference()).max())          # vs numpy: correctness
    spread = float(np.abs(np.max(outs, 0) - np.min(outs, 0)).max())  # run-to-run: determinism
    print(f"{name}: error_vs_reference={err:.4f}  run_to_run_spread={spread:.4f}")


check("W-padding (1,k)", kernel_on_h=False)
check("H-padding (k,1)", kernel_on_h=True)
```

Save as `repro.py` and run:

```
uv run --prerelease allow repro.py
```

(`--prerelease allow` is needed because `modular` depends on a prerelease `mojo`.)

### Expected vs actual

Both error and spread should be `0` (a convolution is correct and deterministic). Actual:

```
W-padding (1,k): error_vs_reference=3.4720  run_to_run_spread=5.0426
H-padding (k,1): error_vs_reference=0.0000  run_to_run_spread=0.0000
```

The `(1,k)` case is both wrong (large error vs the reference) and nondeterministic — the exact nonzero values vary from process to process (consistent with an uninitialized-memory read). The `(k,1)` case is correct and stable across runs.

### System information

```shell
Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=15.6.1=0
                   : __archspec=1=m1

max==26.4
mojo=1.0.0b2
```

blinding-pixels:
## Linked issue

Fixes #6712

(#6712 is currently unassigned and `Needs Triage` — no maintainer review yet. Opening this to surface the concrete fix; happy to hold for direction on the approach.)

## Type of change

## Motivation

The static conv padding path miscomputes outputs near the padding boundary. At `wo=16` (K=8) it returns 23 where the correct value is 21, because the static path doesn't derive per-tile padding adjustments the way the non-static path does. (I picked this up as an entry point into the conv kernels — small and self-contained.)

## What changed

Compute the left and right padding-impact boundaries from the convolution's padding, stride, dilation, input width, and kernel width — the same geometry the non-static path already uses. The static path keeps its fixed-width micro-kernel tiles, but now: marks every tile that has a true padding boundary, passes the tile's output-width offset into `_inner_loops_static`, and computes each tap's padding adjustment relative to that tile offset. Net −71 lines in `conv.mojo` (37 added, 108 removed).

## Testing

Added a deterministic 16-case regression sweep. All-ones inputs with packed filters give an analytical oracle: each output must equal the exact number of valid taps times the channel count, which checks every output element and inherently rejects NaNs. The fixed kernel passes all 16 cases exactly. As a negative control, the same test fails against the original implementation in 3/3 uncached runs at the K=8 boundary (`wo=16`, expected 21, actual 23).

## Checklist

- [ ] The linked issue above has been reviewed by a maintainer and is agreed-upon, or this is a trivial fix that does not need prior approval
- [x] PR is small and focused
- [x] If AI tools assisted with this contribution, I have included an `Assisted-by:` trailer

Assisted-by: AI

---

### DIFF
--- a/max/kernels/src/nn/conv/conv.mojo
+++ b/max/kernels/src/nn/conv/conv.mojo
@@ -1723,15 +1723,34 @@ struct ConvDirectNHWC[
             # Point to (n, 0, ho, f_tile_offset) mapped in input
             var output_base = output_curr_image + (f_tile_offset + F * WO * ho)
 
-            # The entire row fits in one micro kernel.
-            comptime if WO <= micro_kernel_height:
+            comptime left_pad_impact_end = min(
+                ceildiv(Self.conv_attr.pad_left(), Self.conv_attr.strides()[1]),
+                WO,
+            )
+            comptime right_pad_impact_start = max(
+                (
+                    W
+                    + Self.conv_attr.pad_left()
+                    - S * Self.conv_attr.dilations()[1]
+                )
+                // Self.conv_attr.strides()[1]
+                + 1,
+                left_pad_impact_end,
+            )
+            comptime for tile_wo in range(0, WO, micro_kernel_height):
+                comptime tile_height = min(micro_kernel_height, WO - tile_wo)
+                comptime padded_left = tile_wo < left_pad_impact_end
+                comptime padded_right = (
+                    tile_wo + tile_height > right_pad_impact_start
+                )
                 self._inner_loops_static[
-                    WO,
+                    tile_height,
                     micro_kernel_width,
-                    True,
-                    True,
+                    padded_left,
+                    padded_right,
                     has_residual,
                     last_c_tile,
+                    tile_wo,
                 ](
                     input_base,
                     filter_base,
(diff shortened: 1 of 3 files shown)


blinding-pixels:

blinding-pixels:
Hey, just checking whether this fix is still wanted. I realize the branch is now well behind `main` and has conflicts. I’d be happy to rebase it, resolve the conflicts, and rerun the tests if the approach is still useful.

PR #6800, opened 2026-07-23
[Stdlib][GPU] Fix warp shuffles for Int/UInt on Apple and NVIDIA GPUs

### LINKED ISSUE #6799: [BUG] Warp shuffle ops and warp.prefix_sum fail with Int/UInt on Apple GPU (Metal backend internal error)
### Actual behavior
On Apple silicon GPUs, calling `warp.prefix_sum` (or any warp shuffle:
`shuffle_up`, `shuffle_down`, `shuffle_idx`, `shuffle_xor`) with a value of
type `Int` or `UInt` fails at kernel launch with an internal error in the
Metal backend:

```
Unhandled exception caught during execution: At oss/modular/mojo/stdlib/std/gpu/host/device_context.mojo:2828:17:
Failed to create compute pipeline state (GPU machine code generation): Compiler encountered an internal error
To get more accurate error information, set MODULAR_DEBUG=device-sync-mode.
```

The same code works correctly with `Int64`, `UInt64`, `Int8`, `Float32` and
other dtypes — only the index dtypes `DType.int` / `DType.uint` (i.e. `Int` /
`UInt`) are affected.

The cause is in `_shuffle_apple_helper`
(`mojo/stdlib/std/gpu/primitives/warp.mojo`). 64-bit values must be split
into two 32-bit shuffles because the AIR `llvm.air.simd_shuffle*` intrinsics
have no 64-bit overloads, but the split path is selected by enumerating
dtypes:

```mojo
comptime if dtype in (DType.int64, DType.uint64):
    # split into two 32-bit shuffles
```

`DType.int` / `DType.uint` are distinct dtypes (MLIR `index` / `uindex`) that
compare unequal to `DType.int64` / `DType.uint64`, even though they are
64-bit wide on Apple silicon. They fall through to the generic branch, which
emits `llvm.air.simd_shuffle_up` with an `i64` operand — an overload that
does not exist in AIR — and Metal machine code generation fails with an
internal error.

### Expected behavior
`warp.prefix_sum(Int(...))` and the warp shuffle operations should compile and produce correct results on Apple GPUs with `Int` / `UInt` values, the same as they already do for `Int64` / `UInt64`.

## Note: NVIDIA is affected too; AMD already has the right approach

Based on code analysis, the NVIDIA path (`_shuffle` in the same file) has the
same dtype-enumeration problem: `Int` / `UInt` match none of the enumerated
branches and hit `comptime assert False, "unhandled shuffle dtype"`, so on
NVIDIA GPUs the same code fails at compile time instead of with a backend
internal error.

The AMD helpers (`_shuffle_amd_helper`, `_dpp_move`) already dispatch on the
value's bit width (`size_of` / `bit_width_of`) instead of enumerating dtypes,
so they handle the index dtypes correctly. The fix is to unify the Apple and
NVIDIA helpers on the same width-based approach, e.g.:

```mojo
comptime if dtype.is_integral() and bit_width_of[dtype]() == 64:
    # split into two 32-bit shuffles
```

This is a one-line condition change per backend and is done in the linked PR.

### Steps to reproduce

Minimal reproducer (`repro.mojo`):

```mojo
from std.gpu import global_idx
from std.gpu.host import DeviceContext
from std.gpu.primitives import warp


def kernel(output: UnsafePointer[Int64, MutAnyOrigin]):
    var i = global_idx.x
    # Works with Int64(i); fails with Int(i):
    var scan = warp.prefix_sum[exclusive=True](Int(i))
    output[i] = Int64(scan)


def main() raises:
    with DeviceContext() as ctx:
        var buf = ctx.enqueue_create_buffer[DType.int64](4)
        buf.enqueue_fill(0)
        ctx.enqueue_function[kernel](buf, grid_dim=1, block_dim=4)
        ctx.synchronize()
        with buf.map_to_host() as host:
            for i in range(4):
                print(host[i])
```

Run on an Apple silicon Mac:

```bash
mojo repro.mojo
```

- With `Int(i)` or `UInt(i)`: the error above (Metal pipeline creation fails
  with an internal compiler error).
- With `Int64(i)` / `UInt64(i)` / `Int8(i)` / `Float32(i)`: works and prints
  the expected exclusive scan `0 0 1 3`.

### System information

```
Mojo 1.0.0b3.dev2026072006 (7d0f0c04)
macOS 15.7.7 (Darwin 24.6.0), Apple M2 Pro
```

SultanovAR:
## Linked issue

`Fixes #6799 `

## Type of change

- [ ] Documentation update

## Motivation

On Apple silicon GPUs, `warp.prefix_sum` (and `shuffle_up` / `shuffle_down` /
`shuffle_idx` / `shuffle_xor`) with an `Int` or `UInt` value fails at kernel
launch with a Metal backend internal error:

```
Failed to create compute pipeline state (GPU machine code generation): Compiler encountered an internal error
```

The Apple and NVIDIA shuffle helpers select the 64-bit
split-into-two-32-bit-shuffles path by enumerating dtypes
(`dtype in (DType.int64, DType.uint64)`). The index dtypes `DType.int` /
`DType.uint` compare unequal to `int64` / `uint64` but are also 64-bit wide on
GPU targets, so they fell through: on Apple to the generic branch that emits
`llvm.air.simd_shuffle*` with an `i64` operand (an overload that doesn't exist
in AIR, crashing Metal codegen), and on NVIDIA to the
`comptime assert False, "unhandled shuffle dtype"` catch-all.

The AMD helpers (`_shuffle_amd_helper`, `_dpp_move`) already dispatch on bit
width and are unaffected.

## What changed

- `_shuffle` (NVIDIA) and `_shuffle_apple_helper`: select the 64-bit split
  path with `dtype.is_integral() and bit_width_of[dtype]() == 64` instead of
  enumerating dtypes — a one-line condition change per backend, unifying all
  three backends on the width-based approach already used by the AMD helpers.
- `test_shuffle.mojo`: added `Int` / `UInt` variants of the
  `shuffle_idx` / `shuffle_up` / `shuffle_down` / `shuffle_xor` tests,
  mirroring the existing `int64` ones. These exercise the fixed helpers
  directly; `warp.prefix_sum` is built on `shuffle_up` on the affected
  backends.

## Testing

On an Apple M2 Pro (macOS 15.7.7, Mojo nightly `1.0.0b3.dev2026072306`):

- `max/kernels/test/gpu/basics/test_shuffle.mojo` passes with the patched
  stdlib and reproduces the failure against the unpatched stdlib (Metal
  internal error on the `DType.int` cases), confirming the new tests cover
  the bug.
- Also verified `warp.prefix_sum[exclusive=True/False]` with `Int` / `UInt`
  values produces correct scans on-device with the patched stdlib.
- NVIDIA/AMD paths validated by code inspection and CI (no NVIDIA/AMD
  hardware locally); the NVIDIA change is the same split path already used
  for `int64`/`uint64`, now also selected for the index dtypes.
- Note: `test_shuffle.mojo` is excluded on Apple GPU in CI (pre-existing
  Metal backend limitation unrelated to this change), so the Apple path was
  verified manually on-device; CI exercises the new tests on the NVIDIA/AMD
  paths.

## Checklist

      approval

### DIFF
--- a/max/kernels/test/gpu/basics/test_shuffle.mojo
+++ b/max/kernels/test/gpu/basics/test_shuffle.mojo
@@ -112,6 +112,14 @@ def test_shuffle_idx_int64(ctx: DeviceContext) raises:
     _shuffle_idx_launch_helper[DType.int64, 1](ctx)
 
 
+def test_shuffle_idx_int(ctx: DeviceContext) raises:
+    _shuffle_idx_launch_helper[DType.int, 1](ctx)
+
+
+def test_shuffle_idx_uint(ctx: DeviceContext) raises:
+    _shuffle_idx_launch_helper[DType.uint, 1](ctx)
+
+
 def _shuffle_up_launch_helper[
     dtype: DType, simd_width: Int
 ](ctx: DeviceContext) raises:
@@ -172,6 +180,14 @@ def test_shuffle_up_fp16_packed(ctx: DeviceContext) raises:
 
 def test_shuffle_up_int64(ctx: DeviceContext) raises:
     _shuffle_up_launch_helper[DType.int64, 1](ctx)
+
+
+def test_shuffle_up_int(ctx: DeviceContext) raises:
+    _shuffle_up_launch_helper[DType.int, 1](ctx)
+
+
+def test_shuffle_up_uint(ctx: DeviceContext) raises:
+    _shuffle_up_launch_helper[DType.uint, 1](ctx)
 
 
 def _shuffle_down_launch_helper[
@@ -236,6 +252,14 @@ def test_shuffle_down_int64(ctx: DeviceContext) raises:
     _shuffle_down_launch_helper[DType.int64, 1](ctx)
 
 
+def test_shuffle_down_int(ctx: DeviceContext) raises:
+    _shuffle_down_launch_helper[DType.int, 1](ctx)
+
+
+def test_shuffle_down_uint(ctx: DeviceContext) raises:
+    _shuffle_down_launch_helper[DType.uint, 1](ctx)
+
+
 def _shuffle_xor_launch_helper[
     dtype: DType, simd_width: Int
 ](ctx: DeviceContext) raises:
@@ -290,6 +314,14 @@ def test_shuffle_xor_fp16_packed(ctx: DeviceContext) raises:
 
 def test_shuffle_xor_int64(ctx: DeviceContext) raises:
     _shuffle_xor_launch_helper[DType.int64, 1](ctx)
+
+
+def test_shuffle_xor_int(ctx: DeviceContext) raises:
+    _shuffle_xor_launch_helper[DType.int, 1](ctx)
+
+
+def test_shuffle_xor_uint(ctx: DeviceContext) raises:
+    _shuffle_xor_launch_helper[DType.uint, 1](ctx)
 
 
 def _warp_reduce_launch_helper[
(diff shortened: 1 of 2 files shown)


SultanovAR:

SultanovAR:
recheck

pprovins:
Nice find and thank you for the PR!

pprovins:
!sync

Issue #6803, opened 2026-07-23
[Stdlib] `Counter` should be constructible from an arbitrary iterable


msaelices:
### Summary

`Counter[V]` can currently only be built from a variadic pack of values or a
`List[V]`. It cannot be built directly from an arbitrary iterable, so the
common tallying idiom `Counter(some_iterable)` does not work. In particular,
counting the characters of a string (the canonical `Counter` example in Python)
requires a manual loop today.

### Motivation

Python's `collections.Counter` accepts any iterable:

```python
>>> Counter("aaabbaacdd")
Counter({'a': 5, 'b': 2, 'd': 2, 'c': 1})
```

Mojo already has the iteration protocol (`Iterable` / `IterableOwned` /
`Iterator` in `std.iter`), and `String` exposes iterators over codepoints,
grapheme slices, and bytes. `List` already grew a generic
`__init__[IterableType: Iterable](ref iterable, out self: List[...Element])`
constructor (see `mojo/stdlib/std/collections/list.mojo:473`). `Counter` should
follow the same idiom so that these all work:

```mojo
var c = Counter("aaabbaacdd".codepoints())   # infers Counter[Codepoint]
var c2 = Counter(my_list)                     # any List[V]
var c3 = Counter(my_set)                      # any Iterable of a KeyElement
```

### Proposed API

A borrowed-iterable constructor that infers the value type from the iterator's
element type, following the same shape as `Dict.fromkeys`'s borrowed overload
(`mojo/stdlib/std/collections/dict.mojo:785`):

```mojo
def __init__[
    IterableType: Iterable,
](
    out self,
    ref iterable: IterableType,
) where (
    IterableType.IteratorType[origin_of(iterable)].Element == Self.V
):
    self = Self()
    for var value in iterable:
        var key = rebind_var[Self.V](value^)
        self[key] = self[key] + 1
```

`Self.V` already carries the `KeyElement & Copyable & ImplicitlyDeletable`
bound at the struct level, so the `where` equality transitively constrains the
element type without a separate `conforms_to`.

This replaces the existing `__init__(out self, items: List[Self.V])` overload,
which becomes redundant (a `List` is an `Iterable`) and would otherwise be
ambiguous with the generic form. `Counter(my_list)` keeps working (and now
infers `V`); the variadic and empty constructors are unchanged.

### Design note: why not `out self: Counter[...Element]`?

The obvious mirror of `List`'s iterable constructor
[...]

PR #6804, opened 2026-07-23
[Stdlib] Add Counter constructor from an arbitrary iterable

msaelices:
## Summary

Adds a `Counter.__init__` overload that builds a `Counter` from any `Iterable`,
tallying each produced value. This enables the common counting idiom that Python
users expect:

```mojo
var c = Counter(["a", "a", "a", "b", "b", "c"])   # infers Counter[String]
var by_byte = Counter(String("aaabbaacdd").bytes())  # infers Counter[Byte]
```

The value type is inferred from the iterator's element type via
`where Element == Self.V`, the same shape as `Dict.fromkeys`'s borrowed
overload. This **replaces** the previous `Counter(items: List[V])` constructor,
which is now redundant (a `List` is an `Iterable`) and would otherwise be
ambiguous with the generic form. `Counter(my_list)` keeps working and now infers
`V`; the empty and variadic constructors are unchanged.

Design tracked in #6803, including why the `List`-style
`out self: Counter[...Element]` form does not work for a hash container (baking
the origin into `Self.V` trips the exclusivity checker on key insertion).

## Test plan

New `test_construction_from_iterable` / `test_construction_from_iterable_unicode`
in `test_counter.mojo` cover list-literal, `List`-variable, and empty-iterable
construction, plus unicode: multi-byte grapheme `String` keys and byte-level
counting of a non-ASCII string.

```
./bazelw test //mojo/stdlib/test/collections:test_counter.mojo.test   # passes
# no regressions in dict/set/list:
./bazelw test //mojo/stdlib/test/collections:test_{dict,set,list}.mojo.test
```

Note: `test_counter.mojo` is otherwise commented out (compiler bug MOCO-3763 +
the Int/SIMD migration); these new functions are added live and run.

Draft pending consensus on #6803.

Assisted-by: AI


willwray:
This is a useful addition, thanks.

The last commit [7748b55](https://github.com/modular/modular/pull/6804/commits/7748b550fc6cd3869f730461723fcd34c38759d9) merged badly with the changelog, creating a large unrelated diff.
Please redo.
(Maybe drop the commit, add the changelog entry as separate final commit, rebase on main and force push)

zgarapetian:
@msaelices Thanks for this, could you please follow @willwray's suggestion and update the PR?

msaelices:
> @msaelices Thanks for this, could you please follow @willwray's suggestion and update the PR?

Sure! done! Thanks

zgarapetian:
Thank you!

!sync

zgarapetian:
!sync

willwray (approved):
👍 Thanks Manuel and Zareh

PR #6814, opened 2026-07-28
[Kernels] Add Gemma-4-31B-IT-NVFP4 MLP decode rows to SM100 NVFP4 tuning table

rachfop:
Gemma-4-31B-IT-NVFP4's MLP decode shapes (down N=5376/K=21504, gate_up N=43008/K=5376, M=1) miss the SM100 NVFP4 tuning table, which carried DeepSeek shapes only, and fall to the hardcoded small-M fallback in `block_scaled_dispatch.mojo`. The fallback picks `k_group_size` by K-iteration parity, so for gate_up's 21 K-iterations it never tries kg=3, which is faster. Adds the two tuned rows; `tuning_configs.mojo` regenerated with tuning_codegen.

B200, vs the 26.5.0.dev20260715 pin fallback: down -19.6% (16.8 us), gate_up -8.4% (27.0 us); about 8% single-request decode throughput combined.

## Test plan

`sm100_nvfp4_tuning_codegen_test` passes, locally (`--check`) and on a B200 host (`./bazelw test --config=prebuilt-mojo`).

### DIFF
--- a/max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
+++ b/max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
@@ -2091,6 +2091,48 @@ def _get_tuning_list_sm100_nvfp4() -> List[TuningConfigSM100]:
         # index: [0]
         TuningConfigSM100(
             M=1,
+            M_end=2,
+            N=5376,
+            K=21504,
+            mma_shape=Index(256, 16, 32),
+            cta_group=2,
+            cluster_shape=Index(2, 1, 1),
+            block_swizzle_size=8,
+            rasterize_order=RasterOrder(1),
+            swapAB=True,
+            k_group_size=2,
+            num_accum_pipeline_stages=1,
+            num_clc_pipeline_stages=0,
+            num_split_k=1,
+            num_pipeline_stages=0,
+            is_small_bn=True,
+            batch_size=1,
+        ),
+        # Automatically generated from [tuning_table_sm100_nvfp4.yaml]
+        # index: [1]
+        TuningConfigSM100(
+            M=1,
+            M_end=2,
+            N=43008,
+            K=5376,
+            mma_shape=Index(128, 8, 32),
+            cta_group=1,
+            cluster_shape=Index(1, 1, 1),
+            block_swizzle_size=8,
+            rasterize_order=RasterOrder(1),
+            swapAB=True,
+            k_group_size=3,
+            num_accum_pipeline_stages=1,
+            num_clc_pipeline_stages=0,
+            num_split_k=1,
+            num_pipeline_stages=0,
+            is_small_bn=True,
+            batch_size=1,
+        ),
+        # Automatically generated from [tuning_table_sm100_nvfp4.yaml]
+        # index: [2]
+        TuningConfigSM100(
+            M=1,
             M_end=17,
             N=18432,
             K=7168,
@@ -2109,7 +2151,7 @@ def _get_tuning_list_sm100_nvfp4() -> List[TuningConfigSM100]:
             batch_size=1,
         ),
         # Automatically generated from [tuning_table_sm100_nvfp4.yaml]
-        # index: [1]
+        # index: [3]
         TuningConfigSM100(
             M=17,
             M_end=33,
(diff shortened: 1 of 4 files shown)


max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
```diff
@@ -2089,6 +2089,48 @@ def _get_tuning_list_sm100_nvfp4() -> List[TuningConfigSM100]:
...
+            cta_group=2,
+            cluster_shape=Index(2, 1, 1),
+            block_swizzle_size=8,
+            rasterize_order=RasterOrder(1),
+            swapAB=True,
+            k_group_size=2,
+            num_accum_pipeline_stages=1,
+            num_clc_pipeline_stages=0,
+            num_split_k=1,
+            num_pipeline_stages=0,
```
danteisalive, line 2107: I don't think `num_pipeline_stages=0` is a correct value

max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
```diff
@@ -2089,6 +2089,48 @@ def _get_tuning_list_sm100_nvfp4() -> List[TuningConfigSM100]:
...
+            M_end=2,
+            N=43008,
+            K=5376,
+            mma_shape=Index(128, 8, 32),
+            cta_group=1,
+            cluster_shape=Index(1, 1, 1),
+            block_swizzle_size=8,
+            rasterize_order=RasterOrder(1),
+            swapAB=True,
+            k_group_size=3,
```
danteisalive, line 2124: I've never tested `k_group_size=3`. could you please check the accuracy with this config?

max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
```diff
@@ -2089,6 +2089,48 @@ def _get_tuning_list_sm100_nvfp4() -> List[TuningConfigSM100]:
...
+            num_accum_pipeline_stages=1,
+            num_clc_pipeline_stages=0,
+            num_split_k=1,
+            num_pipeline_stages=0,
```
danteisalive, line 2128: same here. there is a chance that we don't use `num_pipeline_stages` value so it's ignored but it worth to check. `num_pipeline_stages` cannot be 0

max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
danteisalive: you can add a test case for this particular shape and config into one of our unit tests

max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
rachfop: 0 means auto-compute — it's the field default and what every entry in this table uses. Dispatch maps it to `None`, and the config computes the stage count from the smem budget, aligned to `k_group_size`. The kernel never sees 0.

max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
rachfop: 0 means auto-compute — it's the field default and what every entry in this table uses. Dispatch maps it to `None`, and the config computes the stage count from the smem budget, aligned to `k_group_size`. The kernel never sees 0.

max/kernels/src/linalg/matmul/gpu/sm100_structured/default/tuning_configs.mojo
rachfop: Added a test for this exact shape and config (gate_up, kg=3) in `test_matmul_sm100_block_scaled_fp4_small_bn.mojo` — it passes accuracy against the vendor BLAS reference on B200. The config is legal: 21 k-iters, 21 % 3 == 0.

rachfop: Added a test for this exact shape and config (gate_up, kg=3) in `test_matmul_sm100_block_scaled_fp4_small_bn.mojo` — it passes accuracy against the vendor BLAS reference on B200. The config is legal: 21 k-iters, 21 % 3 == 0.

Issue #6458, opened 2026-04-26
[Feature Request] Add equality comparison for DeviceContext (not based on device ID alone)


ratulb:
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://docs.modular.com/mojo/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

**Problem**
Currently, there is no built‑in way to check if two DeviceContext instances represent the same logical context (not just the same device ID).

Example:
```mojo
var dc1 = DeviceContext(1)
var dc2 = dc1
var dc3 = DeviceContext(1)
```
dc1 and dc2 → same context (same instance)

dc1 (or dc2) and dc3 → different contexts, even though device ID (1) is identical

Comparing by device ID is insufficient because DeviceContext(1) can be called multiple times, potentially returning distinct objects that are not equivalent.

**Workaround (hacky)**
One hacky option is:

```mojo
var same_context = dc1._handle.value() == dc2._handle.value()
```
This compares the internal handles, but it relies on implementation details and is not recommended for production code.

**Request**

Please add a proper equality API for DeviceContext in the public interface, such as:

dc1 == dc2 operator overload

dc1.__equal__(dc2) method

Or explicit def is_same_context(other: DeviceContext) -> Bool

The equality should be defined as same underlying runtime context (i.e., same internal handle), not just same device ID.

This would make the API more robust and avoid reliance on private members.

**Additional context**
This was identified as a missing feature by a Mojo team member.

Thanks

### What is your motivation for this change?

Thanks

PR #6863, opened 2026-08-07
[MAX] Add DeviceContext identity equality

ratulb:
## Linked issue

Related to #6458

## Type of change

- [ ] Documentation update

## Motivation

`DeviceContext` has no equality operators, so two contexts can only be compared by device ID (`api()`). There's no way to tell whether two contexts wrap the same underlying runtime context — e.g. after copying a context or passing it by value, or when deciding whether two contexts on the same device are actually the same resource. Issue #6458 requests a public equality API.

## What changed

- Add `__eq__` / `__ne__` to `DeviceContext` in `max/mojo/max/gpu/host/device_context.mojo`. Equality is defined on the internal native `_handle`: copies of the same context compare equal; separately constructed contexts on the same device compare unequal. `_owning` is deliberately excluded from identity, so a non-owning wrapper around the same native context compares equal to the owning one.
- Add `test_context_equality` to `test_device_context_cpu.mojo` covering copies, freshly constructed contexts, and the `!=` negation. It runs from `main()` alongside the existing tests.

## Testing

Not run locally — this environment has no GPU/toolchain, so the change could not be compiled or executed here. Coverage is provided by the CI bazel `mojo_test` job, which compiles and runs `test_device_context_cpu.mojo.test`, including the new `test_context_equality`.

## Scope note

This covers the old driver only. The HAL/neo-driver `DeviceContext` also lacks equality; a matching patch (comparing the underlying `Stream` via `__is__`) belongs in `mojo/stdlib/std/gpu/host/_device_context_hal.mojo`, which is not present in this checkout. Issue #6458 is therefore left open for that half.

Assisted-by: AI

### DIFF
--- a/max/kernels/test/gpu/device_context/test_device_context_cpu.mojo
+++ b/max/kernels/test/gpu/device_context/test_device_context_cpu.mojo
@@ -13,7 +13,7 @@
 
 from max.gpu.host import DeviceContext, DeviceContextArray
 from std.memory import alloc
-from std.testing import assert_equal, assert_true
+from std.testing import assert_equal, assert_false, assert_true
 
 
 def test_empty_func(ctx: DeviceContext) raises:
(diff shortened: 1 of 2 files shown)


ratulb:

ratulb:
Hi — the only failing check on this PR is //max/tests/tests/kv_cache:test_ucx_env, and as far as I can tell it isn't related to this change (which only touches device_context.mojo and test_device_context_cpu.mojo). I could be wrong, so I wanted to double-check before assuming anything.

Looking at the test, it appears to read the real /dev/infiniband on the runner (only pci_devices_root/ib_class_root are overridden with fake paths). On a host where that directory is visible, the real host NIC set filters out all the test's fake devices, so local_ib_device() returns None. I reproduced this locally: it passes on a host without /dev/infiniband and fails on one with it — so it looks environment-dependent rather than caused by this PR.

Can we confirm whether this is a known pre-existing failure?

manoj-rajagopalan (approved):

JoeLoser:
!sync

PR #6872, opened 2026-08-09
[stdlib] Fix unsafe_memcmp sign for chunked comparisons

### LINKED ISSUE #6859: [BUG] unsafe_memcmp returns the wrong sign when size_of[T]() is a multiple of 4
### Bug description

**Actual behavior**

`unsafe_memcmp` returns the wrong sign whenever `size_of[type]()` is a multiple of 4. It reinterprets the buffers as signed `Int32` chunks, so a chunk whose high bit is set compares as negative:

```mojo
from std.memory import alloc, unsafe_memcmp

def main():
    var a = alloc[UInt32](1)
    var b = alloc[UInt32](1)
    a.unsafe_write(0x8000_0001)   # bytes: 01 00 00 80
    b.unsafe_write(0x0000_0001)   # bytes: 01 00 00 00
    print("as UInt32:", unsafe_memcmp(a.as_imm(), b.as_imm(), 1), "(want 1)")
    print("as bytes :", unsafe_memcmp(a.as_imm().unsafe_bitcast[Byte](),
                                      b.as_imm().unsafe_bitcast[Byte](), 4))
    a.unsafe_free(); b.unsafe_free()
```

```text
as UInt32: -1 (want 1)
as bytes : 1
```

`a > b` on both readings — byte-wise, the first differing byte is `0x80` against `0x00`, and numerically `0x80000001 > 0x00000001`. The two calls compare the identical bytes and disagree, which pins the bug on the chunking rather than on either interpretation of the data.

**Expected behavior**

`1`, matching the documented first-differing-byte semantics and matching what the byte-wise call already returns.

### Steps to reproduce

Save the snippet above as `repro.mojo` and run `mojo run repro.mojo`.

Any element type whose size is a multiple of 4 is affected, and the trigger is simply that the first differing 4-byte chunk has its high bit set — so roughly half of all differing `UInt32` pairs compare backwards. Comparing the same memory as `Byte` gives the correct answer, so the chunked path needs to do its comparison unsigned.

### System information

Reproduces at the current head of `main` built from source — `mojo --version` reports `Mojo 1.1.0.dev0` — on macOS (Apple Silicon). Little-endian; the mechanism is endianness-sensitive in which byte lands in the high position but the signed-versus-unsigned defect is not.

mahendrarathore1742:
## Problem
    `unsafe_memcmp` in `std.memory` reinterpreted memory buffers as signed `Int32`
  chunks when comparing element types whose size is a multiple of 4 (`size_of[type]()
  % 4 == 0`). When a differing 4-byte chunk had its high bit set (e.g., `0x8000_0001`
  vs `0x0000_0001`), signed 32-bit comparison evaluated the high-bit chunk as a
  negative number (`-2147483647 < 1`), causing `unsafe_memcmp` to return `-1` instead
  of `1`.
    
    ## Solution
    1. Updated `unsafe_memcmp` in `mojo/stdlib/std/memory/memory.mojo` to reinterpret
  buffers as unsigned `UInt32` chunks instead of signed `Int32`.
    2. Verified that the result sign reflects the correct first-differing-byte order
  and matches the byte-wise comparison path (`Byte`).
    
    ## Testing
    Added `test_memcmp_high_bit_and_multiples_of_4` to
  `mojo/stdlib/test/memory/test_memory.mojo` covering:
    - Repro case with high-bit-set `UInt32` chunks (`0x8000_0001` vs `0x0000_0001`)
    - Multiples of 4 element sizes (4, 8, 12, 16 bytes)
    - Asserting equality of result sign between chunked path and byte-wise path
  (`Byte`)
  
  Fixes #6859

### DIFF
--- a/mojo/stdlib/std/memory/memory.mojo
+++ b/mojo/stdlib/std/memory/memory.mojo
@@ -148,11 +148,11 @@ def unsafe_memcmp[
     """
     var byte_count = count * size_of[type]()
 
-    comptime if size_of[type]() % size_of[DType.int32]() == 0:
+    comptime if size_of[type]() % size_of[UInt32]() == 0:
         return _memcmp_impl(
-            s1.unsafe_bitcast[Int32](),
-            s2.unsafe_bitcast[Int32](),
-            byte_count // size_of[DType.int32](),
+            s1.unsafe_bitcast[UInt32](),
+            s2.unsafe_bitcast[UInt32](),
+            byte_count // size_of[UInt32](),
         )
 
     return _memcmp_impl(
--- a/mojo/stdlib/std/memory/memory.mojo
+++ b/mojo/stdlib/std/memory/memory.mojo
@@ -28,11 +28,13 @@ from std.traits import (
 )
 from std.sys import _libc as libc
 from std.ffi import external_call
+from std.bit import byte_swap
 from std.sys import (
     align_of,
     codegen_unreachable,
     get_defined_string,
     is_gpu,
+    is_little_endian,
     llvm_intrinsic,
     simd_bit_width,
     simd_width_of,
@@ -58,6 +60,9 @@ def _memcmp_impl_unconstrained[
         var s1i = s1[unsafe_offset=i]
         var s2i = s2[unsafe_offset=i]
         if s1i != s2i:
+            comptime if is_little_endian() and size_of[dtype]() > 1:
+                s1i = byte_swap(s1i)
+                s2i = byte_swap(s2i)
             return 1 if s1i > s2i else -1
     return 0
 
@@ -76,6 +81,9 @@ def _memcmp_opt_impl_unconstrained[
             var s1i = s1[unsafe_offset=i]
             var s2i = s2[unsafe_offset=i]
             if s1i != s2i:
+                comptime if is_little_endian() and size_of[dtype]() > 1:
+                    s1i = byte_swap(s1i)
+                    s2i = byte_swap(s2i)
                 return 1 if s1i > s2i else -1
         return 0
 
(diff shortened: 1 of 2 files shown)


mahendrarathore1742:
Hi @NathanSWard, this PR is ready for review. Thank you!

mojo/stdlib/test/memory/test_memory.mojo
```diff
@@ -568,6 +568,91 @@ def test_memcmp_simd_zero_bytes() raises:
...
+
+@fieldwise_init
+struct SixteenByteStruct(TrivialRegisterPassable):
+    var a: UInt64
+    var b: UInt64
+
+
+def test_memcmp_high_bit_and_multiples_of_4() raises:
+    var a4 = alloc[UInt32]({count = 1}).unsafe_leak()
+    var b4 = alloc[UInt32]({count = 1}).unsafe_leak()
```
NathanSWard: **suggestion**: Since this is test code, instead of doing manual memory management, I would prefer if we just use `List[UInt32]` here. It would cleanup the code a bit.
So cases like this would just be:
```suggestion
    var a4 = List[UInt32](length=1, fill=0x8000_0001)
    var b4 = List[UInt32](length=1, fill=0x0000_0001)
```

mojo/stdlib/std/memory/memory.mojo
```diff
@@ -58,6 +60,9 @@ def _memcmp_impl_unconstrained[
         var s1i = s1[unsafe_offset=i]
         var s2i = s2[unsafe_offset=i]
         if s1i != s2i:
+            comptime if is_little_endian() and size_of[dtype]() > 1:
+                s1i = byte_swap(s1i)
+                s2i = byte_swap(s2i)
```
NathanSWard: **question**: I saw this byte-swapping endian stuff was added after the initial PR. The intent of this PR was just to fix the issues of the signed-ness of the `Int32`/`UInt32` - are we able to keep the in scope here?
The endianness is a different issue that would better addressed as a separate PR/followup with its own set of tests.

NathanSWard (changes requested):

mojo/stdlib/std/memory/memory.mojo
mahendrarathore1742: Yes, you're right. I've reverted the byte_swap/endianness changes and kept this PR scoped to just the Int32 → UInt32 signedness fix. The endianness issue can be addressed in a separate follow-up PR.

NathanSWard (approved):
Awesome, thank you for the changes, and this looks great!

NathanSWard:
!sync

PR #6893, opened 2026-08-14
[doc] Fixed reflection example.

gsmyridis:
## Linked issue

N.A.

Fixes #

## Type of change

- [x] Documentation update

## Motivation

The example in the documentation of reflection is wrong. If I try to run the code I get the following compilation error:

```shell
➜ mojo run ../refl.mojo
/Users/gsmyridis/Projects/compilers/modular/../refl.mojo:8:20: error: cannot materialize comptime value of type 'Array[StringSpan[ImmStaticOrigin], Int(len(#kgen.struct_field_types<:trait<@std::@traits::@anytype::@AnyType> T>))]' to runtime because it is not 'ImplicitlyCopyable'
        print(names[i])
                   ^
/Users/gsmyridis/Projects/compilers/modular/../refl.mojo:8:20: note: use 'materialize' to explicitly materialize the value
        print(names[i])
                   ^
                   materialize[ ]()
/Users/gsmyridis/.pixi/envs/mojo/bin/mojo: error: failed to parse the provided Mojo source module
```

## What changed

I added `materialize`. 

## Testing

N.A. 

## Checklist

      approval


mojo/stdlib/std/reflection/__init__.mojo
```diff
@@ -32,7 +32,7 @@ struct Point:
 def print_fields[T: AnyType]():
     comptime names = reflect[T].field_names()
     comptime for i in range(reflect[T].field_count()):
-        print(names[i])
+        print(materialize[names[i]]())
```
JoeLoser, line 35: Two notes:

1. The identical snippet appears twice more in `reflect.mojo` (module docstring at line 44, `Reflected[T]` docstring at line 172) and both still fail to build. Worth fixing here — `reflect.mojo` is the file readers reach for `field_names()`.

2. The public manual materializes the whole array once instead of per element (`mojo/docs/manual/metaprogramming/reflection.mdx:267`):

```mojo
var names = materialize[reflect[T].field_names()]()
comptime for i in range(reflect[T].field_count()):
    print(names[i])
```

Both compile; matching the manual keeps one idiom in the docs. Non-blocking.

JoeLoser (approved):
LGTM, thanks for the fix! Can you also fix the two other docstrings that have this same error? They are are at `reflect.mojo:44` and `reflect.mojo:172`. After that, happy to sync this in and land it.

mojo/stdlib/std/reflection/__init__.mojo
gsmyridis: Regarding the second point, since `names` is a runtime variable and not `comptime`, the `materialize` is not needed at all. 

Would you prefer to:

- keep the example as is
- remove the `materialize`
- make it consistent with docstrings, that is, make it `comptime` and then `materialize` each `name[i]`

gsmyridis: @JoeLoser Pinging just in case.

JoeLoser: Thanks for poking at this and clarifying! We can keep that as-is then. I'll sync this in and merge it, thanks!

JoeLoser:
!sync

PR #6902, opened 2026-08-16
[stdlib] Remove opaque-pointer hack from WeakPointer

msaelices:
## Summary

`WeakPointer` stored its control block as `Optional[Pointer[NoneType, _]]`
and round-tripped through `unsafe_bitcast` on every access to work around
**MOCO-3525**, which prevented `Optional[Pointer[Self._inner_type, _]]`
from being used as a struct field.

The compiler now supports recursive dependent pointer fields. This is the
same improvement that let `LinkedList` and `IntervalTree` drop their
opaque-pointer nodes (see 8d80886). `ArcPointer` already stores its inner
pointer with the real type (`Pointer[Self._inner_type, _]`); this brings
`WeakPointer` in line by:

- Typing `_inner_ptr_type` as `Pointer[Self._inner_type, MutUntrackedOrigin]`.
- Deleting the `# FIXME MOCO-3525` workaround and every `unsafe_bitcast`.

Net: **10 insertions, 44 deletions**, no `unsafe_bitcast` left in the struct.

## Testing

`./bazelw test //mojo/stdlib/test/memory:test_arc.mojo.test` passes and the
full `//mojo/stdlib/std` builds cleanly. This is a non-functional change
(same layout and semantics, just a properly-typed pointer instead of an
opaque one).

Assisted-by: AI

### DIFF
--- a/mojo/stdlib/std/memory/arc_pointer.mojo
+++ b/mojo/stdlib/std/memory/arc_pointer.mojo
@@ -469,8 +469,7 @@ struct WeakPointer[T: Movable & Deinitable](
     """
 
     comptime _inner_type = _ArcPointerInner[Self.T]
-    # FIXME MOCO-3525: use Pointer[Self._inner_type, MutUntrackedOrigin]
-    comptime _inner_ptr_type = Pointer[NoneType, MutUntrackedOrigin]
+    comptime _inner_ptr_type = Pointer[Self._inner_type, MutUntrackedOrigin]
     var _inner: Optional[Self._inner_ptr_type]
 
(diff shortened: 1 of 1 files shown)


NathanSWard (approved):
Thanks for this! And nice job catch the other case of the `OpaquePointer` hack for recursively-dependent types :)

NathanSWard:
!sync

NathanSWard:
Closing this PR as it was already merged internally. Thanks!

See here: https://github.com/modular/modular/pull/6902#issuecomment-5331610053

PR #6914, opened 2026-08-19
[Nix] Add flake building the release from source

sequencer:
Add a Nix flake (flake.nix + nix/) that rebuilds the open-source Mojo release artefacts from the sources in this repository, with a toolchain built entirely from source by nixpkgs — no prebuilt binary is downloaded or executed anywhere, and the whole build runs natively in the Nix sandbox (no FHS environment, no patchelf, no nix-ld):

- Bazel: nixpkgs bazel_9 (bootstrapped from source).
- Compiler: nixpkgs llvmPackages_22 clang/lld/LLVM tools laid out like Modular's prebuilt tarball and described by the tree's own bazel/public-patches/clang.BUILD (nix/toolchain-repos.nix).
- Sysroot: assembled from nixpkgs glibc/gcc (libstdc++, libgcc, crt) and the system libraries LLDB links (ncurses, libedit, libbsd, libmd) — the nixpkgs analogue of the Ubuntu jammy sysroot tarball.
- Python: nixpkgs CPython (single version, the default python3) in rules_python's hermetic-runtime layout, replacing every python-build-standalone download; llvm-ifs from nixpkgs LLVM.
- nix/bazelrc.nix wires it all into Modular's unmodified Bazel build via --override_repository/--override_module, and gives every linked binary the Nix dynamic linker + store rpaths.
- nix/deps.nix: fixed-output repository cache (source archives and registry metadata only — audited to contain no binaries) + lockfile; nix/sdk.nix: the offline 11k-action build (LLVM/MLIR/Clang/LLDB + the Mojo compiler, which then compiles the standard library), laid out like the wheels' modular/ root; nix/mojo.nix wraps it for direct use; nix/wheels.nix + nix/mkwheel.py produce deterministic wheels (linux_<arch> tag: the binaries are Nix-linked, not manylinux); nix/release.nix assembles wheels + a PEP 503 index for offline pip install + a reproducible source tarball.
- MAX (max, max-core, ...) is excluded on purpose: its core is not open source and this flake refuses prebuilt binaries.  release-compare diffs our wheels' file lists against the published ones (identical except the closed-source gpu-query).

Verified on x86_64-linux: the dependency fetch is deterministic (two independent fetches, identical FOD hash); the offline sandbox build succeeds; mojo run / build / repl, mojo-lldb breakpoints, lldb-server, lldb-dap and mojo-lsp-server all work; `nix flake check` passes.

See nix/README.md for the design and verification notes.


sequencer:
Sorry wrong repo...

PR #6934, opened 2026-08-20
[stdlib] Fixed `_CTimeSpec.as_nanoseconds` for macOS

### LINKED ISSUE #6839: [BUG] macOS os.stat timestamps are too large by up to ~1000 seconds
### Bug description

**Actual behavior**

On macOS, the timestamps returned by `os.stat` are too large by up to ~1000 seconds, because the sub-second field is scaled as if it were microseconds when `struct stat` actually supplies nanoseconds.

For one file, compared against CPython on the same path with no modification in between:

```text
CPython os.stat().st_mtime_ns : 1786052440625291247   (sec 1786052440, nsec 625291247)
Mojo    os.stat().st_mtimespec: 1786053065291247000ns
```

The Mojo value is exactly `1786052440 * 1_000_000_000 + 625291247 * 1000`. The nanoseconds field is multiplied by 1000, which carries `625` whole seconds into the timestamp and leaves a wrong sub-second remainder. The size of the error scales with the nanoseconds value, so it ranges from 0 up to ~999.99 seconds (~16.7 minutes) depending on the file.

This affects `st_atimespec`, `st_mtimespec`, `st_ctimespec`, and `st_birthtimespec`. Any code comparing file times, or comparing a file time against `time.now()`, gets wrong answers — and the error is inconsistent between files, so ordering comparisons can invert.

**Expected behavior**

`st_mtimespec.as_nanoseconds()` should equal CPython's `st_mtime_ns` for the same file (1786052440625291247 in the example above).

### Steps to reproduce

```mojo
from std.os import stat

def main() raises:
    print(String(stat("/etc/hosts").st_mtimespec))
```

Compare against `python3 -c "import os; print(os.stat('/etc/hosts').st_mtime_ns)"`.

Root cause: `_CTimeSpec` in `std/time/time.mojo` documents its second field as "nanoseconds on linux and usec on mac" and `as_nanoseconds()` multiplies by `_NSEC_PER_USEC` on non-Linux targets. That assumption is right for `struct timeval` (`gettimeofday`), but `struct stat` on macOS carries `struct timespec` fields whose `tv_nsec` is already nanoseconds. `os.stat` populates `_CTimeSpec` from those timespecs, so the microsecond scaling is applied to a value that needs none.

Note this makes `_CTimeSpec` ambiguous by construction — its unit depends on which syscall filled it. Distinguishing the timespec and timeval cases (or normalizing at the fill site) would prevent the same bug from recurring.

### System information

macOS (Apple Silicon), current head of `main` built from source. macOS-specific — the Linux branch of `as_nanoseconds()` adds `tv_subsec` unscaled and is correct.

CC: @bgreni

gsmyridis:
## Linked issue

Fixes #6839

## Type of change

- [ ] Documentation update

## Motivation

The referenced issue explain the issue in detail. 

## What changed

`_CTimeSpec` for macOS assumed that the second field expressed μsec, while in reality it expresses nanos. See: https://developer.apple.com/documentation/kernel/timespec. 

## Testing

- I added a test related to the correct expressions of `_CTimeSpec` to nanoseconds. This is the root of the bug.
- I also added a test comparing the common time fields in `os.stat` agains Python. This is test could be removed since the root cause is tested from the previous bug. I added it because the bug emerged in the context of file stat.

## Checklist

      approval


MuhammadNiazAli (commented):
Good simplification, unifying tv_subsec into tv_nsec removes the platform branch entirely in as_nanoseconds and makes the struct's contract way clearer, previously that field's unit silently depended on the OS which is a footgun waiting to happen. The math checks out too, dropping the _NSEC_PER_USEC multiplication makes sense since the field is now nanoseconds on both platforms instead of usec on mac.

One thing worth confirming before merging, this only works correctly if every place that populates this struct (like wherever clock_gettime or gettimeofday results get assigned into tv_sec/tv_nsec) has also been updated to normalize to nanoseconds on the mac path. If any producer still writes usec values into this field expecting the old conversion logic to handle it, this rename will silently produce wrong timestamps instead of a compile error. Worth grepping for other usages of tv_subsec in the codebase to make sure nothing was missed.

gsmyridis:
> One thing worth confirming before merging, this only works correctly if every place that populates this struct (like wherever clock_gettime or gettimeofday results get assigned into tv_sec/tv_nsec) has also been updated to normalize to nanoseconds on the mac path. If any producer still writes usec values into this field expecting the old conversion logic to handle it, this rename will silently produce wrong timestamps instead of a compile error. Worth grepping for other usages of tv_subsec in the codebase to make sure nothing was missed.

`clock_gettime` (`gettimeofday` is not called anywhere) is an FFI call, so you do not populate it yourself.

gsmyridis:
@JoeLoser pinging just in case.

zgarapetian:
Thanks for this @gsmyridis! Could you add a changelog entry please?

gsmyridis:
> Thanks for this @gsmyridis! Could you add a changelog entry please?

@zgarapetian Done :).

mojo/docs/nightly-changelog.md
```diff
@@ -416,7 +416,7 @@ This release completes the removal of APIs deprecated during the v1.0 cycle.
 - A `where` clause naming a type that an enclosing `where` clause constrained
   to a tighter trait can now be proven. Calling a method declared
   `where Ts.contains[T]()` with such a `T` failed with `lacking evidence to
-  prove correctness`, even though `T` was plainly in `Ts`.
```
zgarapetian, line 444: Can you undo this tab change?

mojo/docs/nightly-changelog.md
```diff
@@ -426,7 +426,7 @@ This release completes the removal of APIs deprecated during the v1.0 cycle.
 
 - `mojo build` can cross-compile to RISC-V again. Emitting LLVM IR, assembly,
   or an object for a `riscv32` or `riscv64` triple failed with `target '...'
-  is not supported by this build`.
```
zgarapetian, line 454: Same tab nit here

mojo/docs/nightly-changelog.md
```diff
@@ -463,7 +463,7 @@ This release completes the removal of APIs deprecated during the v1.0 cycle.
 
 - Fixed `ceildiv()` returning `0` for unsigned operands near the type's
   maximum value. The unsigned code path computed `numerator + denominator -
-  1`, which overflows and wraps for large operands; it now derives the
```
zgarapetian, line 491: Same tab nit here

gsmyridis:
@zgarapetian Sorry about that, it was formatted on-save. Should be good now.

zgarapetian:

zgarapetian (approved):
Thanks for the fixes!

zgarapetian:
!sync

zgarapetian:
Thanks again @gsmyridis

PR #6945, opened 2026-08-21
[stdlib] Ignore whitespace in `b64decode`

### LINKED ISSUE #3446: [BUG] b64decode does not handle whitespaces

Nithesh8678:
## Linked issue

Fixes #3446

## Type of change

- [ ] Documentation update

## Motivation

`b64decode` raised on any whitespace in its input. That makes it fail on
perfectly ordinary base64 text: MIME encoders and the `base64` command line
tool both wrap their output at a fixed column, so anything read from a file
or a mail header arrives with embedded newlines. A user hitting this has to
know to strip the whitespace themselves before calling the API.

The issue reports the simple form of this:

```mojo
b64decode("Qm9 uam91cg==")  # expected "Bonjour", got a ValueError
```

Python's `base64.b64decode` ignores whitespace, so this is also a Mojo/Python
parity gap.

## What changed

`b64decode` now filters ASCII whitespace (space, tab, newline, carriage
return, form feed, vertical tab) out of the input before decoding, via a new
`_is_ascii_whitespace` helper.

Two user-visible consequences:

- Whitespace-separated base64 now decodes correctly.
- The "length must be divisible by 4" check now counts only the significant
  characters. An existing test asserted the old message for
  `"invalid base64 string"` (21 characters); with the two spaces ignored the
  length is 19, and the test was updated accordingly.

Non-whitespace characters outside the alphabet still raise, so genuinely
malformed input is not silently accepted.

## Testing

- Added `test_b64decode_whitespace` covering the case from the issue, base64
  wrapped across multiple lines, a mix of all six whitespace bytes, and an
  input whose significant length is still not divisible by 4 (must still
  raise).
- Updated the existing length-error assertion described above.
- `./bazelw test //mojo/stdlib/test/base64/...` passes.
- Ran the wider `collections`, `os`, and `pathlib` suites as well: 91 tests
  pass, 0 failures.
- Ran `./bazelw run format`; it reported no changes.

## Checklist

      approval
- [x] PR is small and focused

> Note: this contribution was AI-assisted (`Assisted-by: AI` trailer is in the
> commit). Flagging it explicitly so reviewers can calibrate their review.


mojo/stdlib/std/base64/base64.mojo
```diff
@@ -129,19 +150,29 @@ def b64encode(input_bytes: ImmSpan[Byte, _]) -> String:
...
     """
     comptime `=` = Byte(ord("="))
-    var data = str.as_bytes()
-    var n = str.byte_length()
+
+    var data = List[Byte](capacity=str.byte_length())
+    for char in str.as_bytes():
+        if not _is_ascii_whitespace(char):
+            data.append(char)
+    var n = len(data)
```
JoeLoser, line 171: This is the main thing I'd like changed. Before this PR `b64decode` read directly out of the caller's buffer via `str.as_bytes()` and allocated nothing but `result`. Now every call — whitespace or not — allocates a `List[Byte]` the size of the input and fills it with a per-byte `append`, each of which carries a capacity check the optimizer may or may not hoist. For a multi-megabyte payload with no whitespace in it, which is the normal case, that's a full extra allocation and copy bought for nothing.

Two ways out, either is fine by me:

**Keep it zero-copy in the common case.** Count significant bytes in one cheap vectorizable pass, and only build the filtered list if that count differs from `str.byte_length()`:

```mojo
var raw = str.as_bytes()
var n = 0
for char in raw:
    n += Int(not _is_ascii_whitespace(char))
if n == len(raw):
    # existing fast path, decode straight out of `raw`
```

**Or drop the copy entirely.** You need one counting pass anyway to do the `n % 4` check, so after that you can drive the decode loop off a cursor that advances past whitespace when fetching each of the four bytes in a quad. No allocation at all, and it also removes the need to keep `data` and `str` conceptually in sync.

The first is the smaller diff and probably the right call for this PR; the second is nicer if you feel like it.

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -129,19 +150,29 @@ def b64encode(input_bytes: ImmSpan[Byte, _]) -> String:
 def b64decode(str: StringSlice[mut=False, _]) raises -> List[Byte]:
     """Performs base64 decoding on the input string.
 
+    Whitespace (spaces, tabs, newlines, carriage returns, form feeds, and
+    vertical tabs) is ignored, matching the behavior of Python's
+    `base64.b64decode`. This allows decoding base64 text that has been
+    wrapped across multiple lines.
```
JoeLoser: "matching the behavior of Python's `base64.b64decode`" isn't right, and it's the kind of inaccuracy that will bite a user who reads it and trusts it. Python's default `validate=False` throws away *all* non-alphabet bytes, not just whitespace:

```python
>>> base64.b64decode('Qm9!uam91cg==')
b'Bonjour'
>>> base64.b64decode('Qm9$$$uam91cg==')
b'Bonjour'
```

This PR keeps those raising — which I agree is the better behavior — so Mojo is deliberately stricter than Python here. Please state Mojo's actual rule and cite the reason that motivates it rather than claiming parity, something like:

> Whitespace (spaces, tabs, newlines, carriage returns, form feeds, and vertical tabs) is ignored, so base64 that has been wrapped across lines — as MIME encoders and the `base64` command line tool produce — decodes without preprocessing. Unlike Python's `base64.b64decode`, non-whitespace characters outside the alphabet still raise.

The last sentence is the useful part to index on IMO: it tells a reader coming from Python exactly where the two differ.

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -129,19 +150,29 @@ def b64encode(input_bytes: ImmSpan[Byte, _]) -> String:
...
-    var n = str.byte_length()
+
+    var data = List[Byte](capacity=str.byte_length())
+    for char in str.as_bytes():
+        if not _is_ascii_whitespace(char):
+            data.append(char)
+    var n = len(data)
 
     if n % 4 != 0:
         raise Error("ValueError: Input length '", n, "' must be divisible by 4")
```
JoeLoser: Now that whitespace is filtered first, this reports a length the caller cannot see anywhere in the value they passed: `b64decode("invalid base64 string")` is 21 bytes and this says `19`. Your test at line 77 needed an explanatory comment for exactly this reason, which is a good sign the message itself should carry the explanation:

```mojo
raise Error(
    "ValueError: Input length '",
    n,
    "' (ignoring whitespace) must be divisible by 4",
)
```

Then the test comment can go away too. What do you think?

mojo/stdlib/std/base64/base64.mojo
```diff
@@ -73,6 +73,27 @@ def _ascii_to_value(char: Byte) raises -> Byte:
         )
 
 
+@always_inline
+def _is_ascii_whitespace(char: Byte) -> Bool:
```
JoeLoser: The stdlib already has a predicate for this: `Codepoint.is_posix_space()` in `std/collections/string/codepoint.mojo`, which accepts the same six bytes *plus* `\x1c`, `\x1d` and `\x1e`. So `_is_ascii_whitespace` becomes a third, silently different notion of "whitespace" in the library, under a name general enough that a future reader will assume it agrees with the others.

I'm not asking you to widen the set — narrower is defensible here, `\x1c`-`\x1e` have nothing to do with how base64 gets wrapped in the wild. But either reuse the existing predicate, or keep this local helper and have the name or a note say what it actually accepts and why base64 uses a narrower set than the rest of the stdlib. A one-line comment pointing at `is_posix_space` and saying "deliberately narrower" would be enough.

mojo/stdlib/test/base64/test_base64.mojo
```diff
@@ -86,9 +87,29 @@ def test_b64decode() raises:
...
     with assert_raises(
-        contains="ValueError: Unexpected character ' ' encountered"
+        contains="ValueError: Unexpected character '!' encountered"
     ):
-        _ = b64decode("invalid base64 string!!!")
+        _ = b64decode("abc!")
+
+
+def test_b64decode_whitespace() raises:
+    # Whitespace is ignored, matching Python's `base64.b64decode`.
```
JoeLoser: Same parity inaccuracy as the docstring — Python drops all non-alphabet bytes, not just whitespace, so this comment describes behavior the code intentionally doesn't have. I'd suggest rewording alongside the docstring.

Do you mind adding the test case from #3446:

```mojo
assert_equal(b64decode("YQ==\n"), bytes_of("a"))
```

mojo/stdlib/test/base64/test_base64.mojo
```diff
@@ -86,9 +87,29 @@ def test_b64decode() raises:
...
+
+    # A mix of tabs, newlines, carriage returns, and spaces.
+    assert_equal(b64decode("\t Y Q\r\n=\v=\f"), bytes_of("a"))
+
+    # Whitespace that makes the effective length not divisible by 4 must
+    # still raise.
+    with assert_raises(
+        contains="ValueError: Input length '3' must be divisible by 4"
+    ):
+        _ = b64decode("a b c")
```
JoeLoser, line 127: There's at least two new behaviors this PR introduces that would warrant some test cases I'd say.

**Whitespace-only input now succeeds.** `b64decode("   ")` and `b64decode("\n")` previously raised on the divisibility check; they now filter to nothing, pass `0 % 4 == 0`, and return an empty list. This matches Python, which is fine, but let's add a test case.

```mojo
assert_equal(b64decode("   "), List[Byte]())
```

**Filtering plus an invalid character.** One interesting interaction is that whitespace is removed and the alphabet check still fires on what's left. We should add a test case with both of these in play:

```mojo
with assert_raises(
    contains="ValueError: Unexpected character '!' encountered"
):
    _ = b64decode("ab c!")
```

Separately, note what the swap at line 92 cost. `"invalid base64 string!!!"` exercised the alphabet check on a realistic malformed string; `"abc!"` is the minimum that reaches it. The case above recovers most of that value, so I'd add it rather than trying to restore the old input.

JoeLoser (commented):
Thanks for the fix! I left some comments inline that I'd like to see addressed.

JoeLoser (changes requested):

JoeLoser (approved):

PR #6946, opened 2026-08-21
[stdlib] Add `List.drain()`

### LINKED ISSUE #5312: [Feature Request] Add `drain`/`take_items` API to `List`

Nithesh8678:
## Linked issue

Fixes #5312

## Type of change

- [ ] Documentation update

## Motivation

`Dict` has `take_items()`, which moves entries out of the dictionary as you
iterate. `List` has no equivalent, so moving elements out of an owned list
into another collection forces a copy of each element even though the source
is about to be discarded. The linked issue points at concrete places this
bites, including a `TODO` in `Dict`'s own initialization noting that it
should transfer key/value pairs rather than copy them.

Consuming iteration (`for x in list^`) already exists, but it takes the whole
list by value. That does not help when the list is borrowed and must stay
alive, which is the case in the issue's motivating example:

```mojo
for ref key_value in zip(keys.drain(), values.drain()):
    self._data.insert(key_value^)
```

## What changed

Adds `List.drain()`, returning a `_ListDrainIter` that moves each element out
of the list as it is yielded. It mirrors `Dict.take_items()` in naming and
docstring style.

The one design point worth review: `List` has contiguous storage and no
tombstones, so unlike `Dict` it cannot cheaply update its bookkeeping per
element. Instead `_ListDrainIter.__deinit__` destroys whatever was not
drained and then sets `_len = 0`, so the list ends up empty whether or not
iteration ran to completion, and each element is destroyed exactly once. This
is the same approach `_ListIterOwned` already uses. `_data`/`_capacity` are
left alone so the allocation is still freed normally when the list deinits.

`drain()` requires only `Movable & Deinitable`, not `Copyable`.

## Testing

Added four tests to `test_list.mojo`:

- `test_list_drain` — full drain, list is empty afterwards and still usable.
- `test_list_drain_move_only` — works with a non-`Copyable` element type.
- `test_list_drain_destroys_elements_if_partially_consumed` — uses the
  existing `ObservableElement` helper to assert that after draining one of
  two elements and dropping the iterator, `dels == 2` and `copies == 0`, i.e.
  no leak and no double-free.
- `test_list_drain_empty`.

`./bazelw test //mojo/stdlib/test/collections/...` passes. Also ran the
`base64`, `os` and `pathlib` suites: 91 tests pass, 0 failures. Ran
`./bazelw run format`; no changes.

## Checklist

      approval
- [x] PR is small and focused

> Note: this contribution was AI-assisted (`Assisted-by: AI` trailer is in the
> commit). Flagging it explicitly so reviewers can calibrate their review.

### DIFF
--- a/mojo/stdlib/std/collections/list.mojo
+++ b/mojo/stdlib/std/collections/list.mojo
@@ -183,7 +183,9 @@ struct _ListDrainIter[T: Movable & Deinitable, origin: MutOrigin](
             self._src[].unsafe_ptr().unsafe_offset(self._index),
             count=len(self._src[]) - self._index,
         )
+        var old_size: Int = self._src[]._len
         self._src[]._len = 0
+        self._src[]._annotate_shrink(old_size)
 
     @always_inline
     def __iter__(var self) -> Self.IteratorOwnedType:
--- a/mojo/stdlib/std/collections/list.mojo
+++ b/mojo/stdlib/std/collections/list.mojo
@@ -152,6 +152,60 @@ struct _ListIterOwned[T: Movable & Deinitable](
         return (iter_len, {iter_len})
 
 
+@fieldwise_init
+struct _ListDrainIter[T: Movable & Deinitable, origin: MutOrigin](
+    IterableOwned, Iterator, Movable
+):
+    """An iterator over a List's elements that moves them out of the list,
+    effectively draining the list.
+
+    Parameters:
+        T: The type of the elements in the list.
+        origin: The mutable origin of the List.
+    """
+
+    comptime Element = Self.T
+    comptime IteratorOwnedType = Self
+
+    var _src: Pointer[List[Self.T], Self.origin]
+    var _index: Int
+
+    def __init__(out self, ref[Self.origin] list: List[Self.T]):
+        self._src = Pointer(to=list)
+        self._index = 0
+
+    @always_inline
+    def __deinit__(deinit self):
+        # Destroy the elements that have not yet been drained, then mark the
+        # source list as empty. `_data`/`_capacity` are left untouched so the
+        # list's underlying allocation is freed normally when it deinits.
+        unsafe_destroy_n(
+            self._src[].unsafe_ptr().unsafe_offset(self._index),
+            count=len(self._src[]) - self._index,
+        )
+        self._src[]._len = 0
+
+    @always_inline
+    def __iter__(var self) -> Self.IteratorOwnedType:
+        return self^
+
+    def __next__(mut self) raises StopIteration -> Self.Element:
+        if self._index >= len(self._src[]):
+            raise StopIteration()
+        self._index += 1
+        return (
+            self._src[]
+            .unsafe_ptr()
+            .unsafe_offset(self._index - 1)
+            .unsafe_take_pointee()
+        )
+
+    @always_inline
+    def bounds(self) -> Tuple[Int, Optional[Int]]:
+        var iter_len = len(self._src[]) - self._index
+        return (iter_len, {iter_len})
+
+
 @explicit_destroy(
     "Use `deinit_with()` to explicitly destroy a `List` of"
     " non-`Deinitable` elements"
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/collections/list.mojo
```diff
@@ -755,6 +811,31 @@ struct List[T: AnyType, /](
...
+        """Iterate over the list's elements and move them out of the list,
+        effectively draining the list.
+
+        Returns:
+            An iterator of owned elements that moves them out of the list.
+
+        Examples:
+
+        ```mojo
+        var my_list = [1, 2, 3]
```
JoeLoser: `warning` · non-blocking · docs

A bare bracket literal infers `Array[Int, 3]`, which has no `drain()` — so the only example for this new API doesn't compile. Your tests get this right (`var list: List = [1, 2, 3]`); the docstring should match. Same defect #6973 is cleaning up elsewhere in this file.

```suggestion
        var my_list: List[Int] = [1, 2, 3]
```

mojo/stdlib/test/collections/test_list.mojo
```diff
@@ -782,6 +782,70 @@ def test_list_iter_owned_bounds() raises:
...
+    for value in list.drain():
+        total += value
+
+    assert_equal(total, 6)
+    assert_equal(len(list), 0)
+
+    # The list is still usable (and empty) after being drained.
+    list.append(4)
+    var expected: List = [4]
+    assert_equal(list, expected)
```
JoeLoser: `warning` · non-blocking · testing

#5312 asks specifically to "ensure that `zip` is compatible with this (supports mutable iterators)", and its motivating example is `zip(keys.drain(), values.drain())`. Nothing here exercises that, so the one use case the API was requested for is unverified — and if it doesn't compose with `zip`, that's better found now than after the API is public.

mojo/stdlib/std/collections/list.mojo
```diff
@@ -152,6 +152,62 @@ struct _ListIterOwned[T: Movable & Deinitable](
         return (iter_len, {iter_len})
 
 
+@fieldwise_init
```
JoeLoser: `minor` · non-blocking · design

Redundant next to the explicit `__init__` below, and `_ListIterOwned` directly above has neither. It also exposes a second constructor that takes an arbitrary `_index`, which is exactly the field that must start at 0 for `__deinit__`'s destroy range to be sound.

Issue #6920, opened 2026-08-19
[Feature Request] [stdlib] Migrate the remaining abort tests to _assert_aborts


JoeLoser:
### Review Mojo's priorities

- [X] I have read the [roadmap and priorities](<https://mojolang.org/docs/roadmap.html#overall-priorities>) and I believe this request falls within the priorities.

### What is your request?

`std.testing._assert_aborts()` replaced the old `-D test=N` + lit RUN-line pattern for testing code that aborts. I converted `mojo/stdlib/test/collections/test_span_bounds_abort.mojo` along with it, but about 40 other stdlib tests still use one of the two older patterns. This tracks moving them over.

The first group is the lit tests that dispatch on `-D test=N` through a `comptime if get_defined_int["test"]()` chain, one RUN line per case:

| File | Cases |
| -- | -- |
| `mojo/stdlib/test/collections/test_list_bounds_abort.mojo` | 5 |
| `mojo/stdlib/test/collections/string/test_string_span_bounds_abort.mojo` | 4 |
| `mojo/stdlib/test/collections/string/test_string_span_codepoint_grapheme_bounds_abort.mojo` | 9 |
| `mojo/stdlib/test/memory/test_fail/layout_invalid_alignment_abort.mojo` | 2 |
| `mojo/stdlib/test/memory/test_fail/alloc_zst_fail_tests.mojo` | 1 |
| `mojo/stdlib/test/collections/string/test_string_unicode_panic.mojo` (argv dispatch, `not not %t`) | 6 |

Each RUN line becomes a `test_*` function, and the files become plain `mojo_test` targets that drop out of `_LIT_TESTS` in their BUILD file.

The second group is tests that get a whole `mojo_filecheck_test(expect_crash = True)` target for a single abort. There are 34:

* `collections/`: 15, being `test_negative_index_*` (9), `test_check_bounds_{assert_error,direct_call}`, `test_binary_heap_assert_empty_{pop,peek}`, `test_optional_none_abort`, `test_list_getitem_invalid_index_int`
* `memory/uninit_check/`: 7 poison tests, all sharing the same `-D` flag, so they merge into one file
* `builtin/`: 5, being `test_debug_assert_{default_error,mode_all_error,location,no_message}` and `test_range_len_overflow_abort`
* `itertools/`: 2, `test_{take,drop}_negative_count`
* `os/`: 2, `test_trap` and `test_trap_gpu`
* one each in `logger/`, `memory/`, `memory/pointer/`

[...]

PR #6952, opened 2026-08-21
refactor(stdlib): migrate test_list_bounds_abort to _assert_aborts

gabo3454675:
## Summary

Convert the lit-based abort test for List slice bounds checking to use .

## Changes

- Replaced 5 RUN lines and  dispatch chain with 5 focused  functions
- Each test uses a trigger closure with  following the pattern in 
- Removed test from  in BUILD.bazel (now auto-discovered by  glob)

## Motivation

This is part of issue #6920 - migrating abort tests to use  reduces boilerplate and makes tests more maintainable.

## Testing

The test file follows the established pattern from  which is already merged.


mojo/stdlib/test/memory/test_fail/alloc_zst_fail_tests.mojo
```diff
@@ -32,19 +26,17 @@ def test_alloc_zst_count_negative_fails() raises:
...
-    # CHECK_1: alloc: `Layout.count()` must be > 0
-    var layout = Layout[ZST](count=-1)
-    var ptr = alloc(layout).unsafe_leak()
-
-    # CHECK_1-NOT: is never reached
-    assert_equal(0, len(ptr[]))
+    def trigger() raises -> None:
+        var layout = Layout[ZST](count=-1)
+        var ptr = alloc(layout).unsafe_leak()
+        assert_equal(0, len(ptr[]))
```
JoeLoser: `error` · blocking · correctness

`assert_equal` was removed from the `std.testing` import above but is still called here, so the file doesn't compile:

```
error: use of unknown declaration 'assert_equal'; did you mean to import it from 'std.testing'?
```

It's also unreachable — the `alloc` on the line above is what aborts. Dropping this line and the `dealloc` below reads better than re-importing something that can never run.

mojo/stdlib/test/memory/test_fail/BUILD.bazel
```diff
@@ -1,11 +1,31 @@
-load("//bazel:api.bzl", "lit_tests")
+load("//bazel:api.bzl", "lit_tests", "mojo_test")
 
 package(default_visibility = ["//visibility:private"])
 
+_MOJO_TESTS = [
+    "layout_invalid_alignment_abort.mojo",
```
JoeLoser: `error` · blocking · build

This PR deletes the only `# RUN:` line from `alloc_zst_fail_tests.mojo`, but that file isn't in `_MOJO_TESTS`, so the `glob` below still hands it to `lit_tests` — and lit fails a file with no run line.

```suggestion
    "alloc_zst_fail_tests.mojo",
    "layout_invalid_alignment_abort.mojo",
```

mojo/stdlib/test/collections/string/test_string_unicode_panic.mojo
```diff
@@ -10,54 +10,92 @@
...
+    var s = String("😀longlonglonglong")
 
-def main() raises:
-    if len(argv()) <= 1:
-        return
-    var test = argv()[1]
-    if test == "1":
-        var s = String("😀longlonglonglong")
-        # CHECK_1: does not lie on a codepoint boundary.
+    def trigger() raises {s} -> None:
```
JoeLoser: `error` · blocking · correctness

`{s}` captures immutably, so `resize` has no matching overload — `no matching method in call to 'resize'` at lines 21, 33, 56 and 68. Every trigger in this file mutates its `String`.

`{mut s}` fixes it; I built and ran the first one that way and it passes.

```suggestion
    def trigger() raises {mut s} -> None:
```

mojo/stdlib/test/memory/test_fail/layout_invalid_alignment_abort.mojo
```diff
@@ -11,22 +11,34 @@
...
+
+    def trigger() raises {too_small} -> None:
         var _layout = Layout[Int32](count=1, alignment=too_small)
-    elif std.sys.get_defined_int["ALLOC_TEST"]() == 2:
-        var not_pow_of_2 = 127
-        # CHECK-TEST2: ABORT: {{.*}}: Alignment is invalid. Must be a power of two and >= to the types natural alignment.
+
+    _assert_aborts(
+        trigger,
+        contains="Alignment is invalid. Must be a power of two and >= to the types natural alignment.",
```
JoeLoser: `warning` · blocking · style

`mojo format` rewraps this string (and the one on line 39) — it runs past the column limit. It also reflows `s[byte=-1:]` back to `s[byte= -1:]` in `test_string_span_bounds_abort.mojo` line 51 and `test_string_span_codepoint_grapheme_bounds_abort.mojo` lines 53 and 104. `./bazelw run //:format` clears all five.

mojo/stdlib/test/collections/string/test_string_unicode_panic.mojo
```diff
@@ -10,54 +10,92 @@
...
         s.resize(1)
```
alisterburt, line 21: Looks like this is hitting a compiler error
https://github.com/modular/modular/actions/runs/32741202382/job/97475835392?pr=6952#step:3:219

The problem is that you are capturing `s` and by default this capture give you an immutable reference. 

```suggestion
    def trigger() raises {mut s} -> None:
s.resize(1)
```

You would fix this by explicitly capturing a mutable reference

alisterburt (changes requested):
Hey @gabo3454675

Thanks for opening this PR! There are some linting and compilation errors I've provided some hints about in the comments. In future if you make sure the code is linted and compiles before opening a PR that would be great.

NathanSWard:
Closing this PR as we've not hear back yet from the original author @gabo3454675 - please re-open if you'd like to readdress these issues, thanks!

PR #6954, opened 2026-08-21
[stdlib] Add a timeout to assert_aborts's re-exec'd child

### LINKED ISSUE #6911: [BUG] [stdlib] Add a timeout to assert_aborts's re-exec'd child
### Review Mojo's priorities

- [x] I have read the [roadmap and priorities](https://mojolang.org/docs/roadmap.html#overall-priorities) and I believe this request falls within the priorities.

### What is your request?

Within `_assert_aborts`, `_spawn_and_check` does a plain blocking `child.wait()` with no timeout. If the closure under test has a bug and hangs instead of aborting (deadlock, infinite loop, blocked I/O), the parent blocks forever and the whole test binary/CI job hangs with no diagnostic.

Need a timeout on the wait, and a clear error (kill the child, report "did not abort within Ns") when it fires.

N/A

### Any other details?

_No response_

jayhemnani9910:
## Linked issue

Fixes #6911

## Type of change

## Motivation

`_assert_aborts` runs the closure under test in a re-exec'd child, and
`_spawn_and_check` waits on that child with a plain blocking `Process.wait()`.
If the closure hangs instead of aborting (deadlock, infinite loop, blocked
I/O), the parent blocks forever. The test binary stops making progress and the
CI job is eventually killed by the harness, with nothing in the output saying
which assertion was responsible.

## What changed

- `_assert_aborts` takes a keyword-only `timeout` in seconds, defaulting to
  60.0, threaded through `_assert_aborts_impl` to `_spawn_and_check`. All three
  are underscore-prefixed internals, so there is no public API change.
- `_spawn_and_check` now polls `Process.poll()` on a 1 ms interval instead of
  blocking, and gives up once the deadline passes.
- On expiry it kills the child, reaps it, and raises an error naming the limit
  and including the child's captured output.
- The timeout is reported before the existing `term_signal` check. We send the
  kill ourselves, so the child comes back terminated by SIGKILL, and letting
  that reach the `term_signal` branch would report a hang as a passing abort
  assertion.

## Testing

- Added `test_raises_when_closure_does_not_abort_in_time`, which runs a closure
  that sleeps and asserts the raise, following the `assert_raises` pattern
  already used by the negative tests in that file.
- `./bazelw test //mojo/stdlib/test/testing:test_assert_aborts.mojo.test`
  passes 6 of 6. The new test completes in 501 ms against `timeout=0.5`, so the
  timeout path is actually exercised rather than passing vacuously.
- `./bazelw test //mojo/stdlib/test/collections:test_span_bounds_abort.mojo.test //mojo/stdlib/test/testing/...`
  passes 9 tests. That covers the other five `_assert_aborts` call sites in the
  tree, which all pass `contains=` by keyword and so are unaffected by the new
  keyword-only parameter.
- `./bazelw run format` is clean and made no changes.

## One thing noticed while tracing callers

`Process.__deinit__` at `mojo/stdlib/std/os/process.mojo:248` also does a bare
blocking `self.wait()`, so dropping a `Process` whose child is hung blocks there
too. That predates this change and is out of scope here. Happy to file a
separate issue if that is useful.

## Checklist

      agreed-upon (#6911 carries the `accepted` label)
- [x] PR is small and focused (53 insertions, 7 deletions across 2 files)
- [x] AI tools assisted with this contribution; the commit carries an
      `Assisted-by: AI` trailer

Implementation assisted by Claude Code.

### DIFF
--- a/mojo/stdlib/std/testing/assert_aborts.mojo
+++ b/mojo/stdlib/std/testing/assert_aborts.mojo
@@ -33,9 +33,7 @@ from std.time import perf_counter, sleep
 comptime _LOCATION_ENV = "__MOJO_TEST_EXPECT_ABORT_LOCATION_TARGET"
 # Path the child's stdout/stderr get redirected to before running.
 comptime _OUTPUT_ENV = "__MOJO_TEST_EXPECT_ABORT_OUTPUT"
-# How long to wait for the child to abort before killing it. Kept well under a
-# typical CI test timeout so a hang is reported here, with a diagnostic, rather
-# than by the harness killing the whole binary.
+# How long to wait for the child to abort before killing it.
 comptime _DEFAULT_TIMEOUT = 60.0
 # How often to check on the child while waiting for it. The child re-execs the
 # whole test binary, so this is small next to the work it does.
--- a/mojo/stdlib/std/testing/assert_aborts.mojo
+++ b/mojo/stdlib/std/testing/assert_aborts.mojo
@@ -34,10 +34,9 @@ comptime _LOCATION_ENV = "__MOJO_TEST_EXPECT_ABORT_LOCATION_TARGET"
 # Path the child's stdout/stderr get redirected to before running.
 comptime _OUTPUT_ENV = "__MOJO_TEST_EXPECT_ABORT_OUTPUT"
 # How long to wait for the child to abort before killing it.
-comptime _DEFAULT_TIMEOUT = 60.0
-# How often to check on the child while waiting for it. The child re-execs the
-# whole test binary, so this is small next to the work it does.
-comptime _POLL_INTERVAL = 0.001
+comptime _DEFAULT_TIMEOUT = 30.0
+# How often to check on the child while waiting for it.
+comptime _POLL_INTERVAL = 0.01
 
 
 @always_inline
@@ -165,8 +164,6 @@ def _spawn_and_check(
     var timed_out = False
     try:
         var child = Process.run(String(self_argv[0]), rest)
-        # Poll rather than block, so a closure that hangs instead of aborting
-        # fails here instead of wedging the whole test binary.
         var deadline = perf_counter() + timeout
         status = child.poll()
         while not status.has_exited():
@@ -195,8 +192,6 @@ def _spawn_and_check(
             t" used for assert_aborts.\nError: {e}"
         )
 
-    # Checked before `term_signal`: we killed the child, so it *is* signaled,
-    # and reading that as a pass would turn a hang into a false success.
     if timed_out:
         raise Error(
             t"assert_aborts: expected the process to abort, but it was still"
--- a/mojo/stdlib/std/testing/assert_aborts.mojo
+++ b/mojo/stdlib/std/testing/assert_aborts.mojo
@@ -27,16 +27,27 @@ from std.sys._io import stderr, stdout
 from std.sys._libc import dup2
 from std.sys.compile import SanitizeAddress
 from std.tempfile import NamedTemporaryFile
+from std.time import perf_counter, sleep
 
 # Which call site (file:line:col) the re-exec'd child should run.
 comptime _LOCATION_ENV = "__MOJO_TEST_EXPECT_ABORT_LOCATION_TARGET"
 # Path the child's stdout/stderr get redirected to before running.
 comptime _OUTPUT_ENV = "__MOJO_TEST_EXPECT_ABORT_OUTPUT"
+# How long to wait for the child to abort before killing it. Kept well under a
+# typical CI test timeout so a hang is reported here, with a diagnostic, rather
+# than by the harness killing the whole binary.
+comptime _DEFAULT_TIMEOUT = 60.0
+# How often to check on the child while waiting for it. The child re-execs the
+# whole test binary, so this is small next to the work it does.
+comptime _POLL_INTERVAL = 0.001
 
 
 @always_inline
 def _assert_aborts(
-    f: Some[def() raises], *, contains: Optional[String] = None
+    f: Some[def() raises],
+    *,
+    contains: Optional[String] = None,
+    timeout: Float64 = _DEFAULT_TIMEOUT,
 ) raises:
     """Asserts that calling `f` aborts the process.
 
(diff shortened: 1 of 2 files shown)


mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -27,16 +27,27 @@ from std.sys._io import stderr, stdout
...
 
 # Which call site (file:line:col) the re-exec'd child should run.
 comptime _LOCATION_ENV = "__MOJO_TEST_EXPECT_ABORT_LOCATION_TARGET"
 # Path the child's stdout/stderr get redirected to before running.
 comptime _OUTPUT_ENV = "__MOJO_TEST_EXPECT_ABORT_OUTPUT"
+# How long to wait for the child to abort before killing it. Kept well under a
+# typical CI test timeout so a hang is reported here, with a diagnostic, rather
+# than by the harness killing the whole binary.
+comptime _DEFAULT_TIMEOUT = 60.0
+# How often to check on the child while waiting for it. The child re-execs the
```
JoeLoser: `suggestion` · non-blocking · design

The budget is per call site, so a file whose code under test genuinely hangs pays 60s at *every* `_assert_aborts`. Bazel's default test timeout is 300s, so from the fifth assertion on the harness kills the binary first and the diagnostic this exists to print never appears.

With #6920 moving ~40 files onto this helper, a smaller default — or one deadline shared across the binary — would keep it reachable.

mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -148,9 +164,21 @@ def _spawn_and_check(
...
     try:
         var child = Process.run(String(self_argv[0]), rest)
-        status = child.wait()
+        # Poll rather than block, so a closure that hangs instead of aborting
+        # fails here instead of wedging the whole test binary.
+        var deadline = perf_counter() + timeout
+        status = child.poll()
+        while not status.has_exited():
+            if perf_counter() >= deadline:
+                _ = child.kill()
```
JoeLoser: `minor` · non-blocking · correctness

If the child exits on its own between the deadline check and this `kill()`, `wait()` returns its real status but `timed_out` is already committed, so a correct abort is reported as a hang. One more poll before killing closes the window:

```suggestion
                status = child.poll()
                if status.has_exited():
                    break
                _ = child.kill()
```

mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -27,16 +27,27 @@ from std.sys._io import stderr, stdout
...
 from std.tempfile import NamedTemporaryFile
+from std.time import perf_counter, sleep
 
 # Which call site (file:line:col) the re-exec'd child should run.
 comptime _LOCATION_ENV = "__MOJO_TEST_EXPECT_ABORT_LOCATION_TARGET"
 # Path the child's stdout/stderr get redirected to before running.
 comptime _OUTPUT_ENV = "__MOJO_TEST_EXPECT_ABORT_OUTPUT"
+# How long to wait for the child to abort before killing it. Kept well under a
+# typical CI test timeout so a hang is reported here, with a diagnostic, rather
+# than by the harness killing the whole binary.
```
alisterburt:
```suggestion
# How long to wait for the child to abort before killing it.
```

mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -27,16 +27,27 @@ from std.sys._io import stderr, stdout
...
+comptime _DEFAULT_TIMEOUT = 60.0
```
alisterburt:
```suggestion
comptime _DEFAULT_TIMEOUT = 30.0
```

Let's drop this a bit to keep CI times under control

mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -27,16 +27,27 @@ from std.sys._io import stderr, stdout
...
+# How often to check on the child while waiting for it. The child re-execs the
+# whole test binary, so this is small next to the work it does.
```
alisterburt:
```suggestion
# How often to check on the child while waiting for it.
```

mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -148,9 +164,21 @@ def _spawn_and_check(
         _ = unsetenv(_OUTPUT_ENV)
 
     var status: ProcessStatus
+    var timed_out = False
     try:
         var child = Process.run(String(self_argv[0]), rest)
-        status = child.wait()
+        # Poll rather than block, so a closure that hangs instead of aborting
+        # fails here instead of wedging the whole test binary.
```
alisterburt:
```suggestion
```

mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -169,6 +197,15 @@ def _spawn_and_check(
             t" used for assert_aborts.\nError: {e}"
         )
 
+    # Checked before `term_signal`: we killed the child, so it *is* signaled,
+    # and reading that as a pass would turn a hang into a false success.
```
alisterburt:
```suggestion
```

mojo/stdlib/std/testing/assert_aborts.mojo
```diff
@@ -27,16 +27,27 @@ from std.sys._io import stderr, stdout
...
 comptime _LOCATION_ENV = "__MOJO_TEST_EXPECT_ABORT_LOCATION_TARGET"
 # Path the child's stdout/stderr get redirected to before running.
 comptime _OUTPUT_ENV = "__MOJO_TEST_EXPECT_ABORT_OUTPUT"
+# How long to wait for the child to abort before killing it. Kept well under a
+# typical CI test timeout so a hang is reported here, with a diagnostic, rather
+# than by the harness killing the whole binary.
+comptime _DEFAULT_TIMEOUT = 60.0
+# How often to check on the child while waiting for it. The child re-execs the
+# whole test binary, so this is small next to the work it does.
+comptime _POLL_INTERVAL = 0.001
```
alisterburt:
```suggestion
comptime _POLL_INTERVAL = 0.01
```

alisterburt (changes requested):
Hey @jayhemnani9910, thanks for taking the time to open this PR! Implementation looks good. I've added a few suggestions for changes to defaults and removing/simplifying some of the LLM generated comments.

In future if you could simplify your PR description that would be great, your description looks AI generated isn't particularly clear or concise.

PR #6969, opened 2026-08-23
[stdlib] Fix `SIMD.__pow__` with float literal exponents

### LINKED ISSUE #6940: [BUG] Scalar[dt] ** 0.5 fails to compile for float16/bfloat16/float32 (pass-manager error)
#### Summary

Raising a `Scalar[dt]` to the power `0.5` (`x**0.5`) fails to instantiate for
`DType.float16`, `DType.bfloat16`, and `DType.float32` with a pass-manager error
("constraint failed: unsupported type combination" from `std/builtin/simd.mojo`).
`DType.float64` works. `std.math.sqrt(x)` handles every float dtype and is a valid
workaround.

This **passes on `1.0.0b2` (2cf4d08a)** and **fails on `1.0.0` stable (ed45d567)**.

#### Actual behavior

```text
$ mojo repro_scalar_pow.mojo        # Mojo 1.0.0 stable
error: failed to run the pass manager
note: constraint failed: unsupported type combination   (std/builtin/simd.mojo)
```

#### Expected behavior

```text
$ mojo repro_scalar_pow.mojo        # Mojo 1.0.0b2
float16 => 2.0
bfloat16 => 2.0
float32 => 2.0
float64 => 2.000000000000565
```

#### Steps to reproduce

Save as `repro_scalar_pow.mojo` (also at `docs/dev/reproducers/repro_scalar_pow.mojo`
in Odyssey):

```mojo
def f[dt: DType](x: Scalar[dt]) -> Scalar[dt]:
    return x**0.5

def main():
    var v1 = Scalar[DType.float16](4.0)
    print("float16 =>", f(v1))
    var v2 = Scalar[DType.float32](4.0)
    print("float32 =>", f(v2))
    var v3 = Scalar[DType.float64](4.0)
    print("float64 =>", f(v3))
```

Run:

```bash
mojo repro_scalar_pow.mojo
```

#### Impact

Code that computes `x**0.5` on `Scalar` values of float16/bfloat16/float32 (e.g. a
generic `sqrt` helper) fails to compile on stable 1.0.0. The Odyssey framework hit this
in `_sqrt_typed` (`normalization.mojo`) and switched to `std.math.sqrt`, which works.

#### Environment

- Mojo version: 1.0.0 (ed45d567) — also reproduced with 1.0.0b2 (2cf4d08a) for comparison
- OS: Linux x86_64 (Ubuntu 24.04 container), glibc 2.39
- Installed via `pip install mojo==1.0.0` / `mojo==1.0.0b2` (PyPI)

iMostfa:
## Linked issue

Fixes #6940

## Type of change

- [ ] Documentation update

## Motivation

Fixes what appears to be a regression caused by b6c836e725
## Testing

1. I tested by running the example in the issue.
2. I added a test case in `[mojo/stdlib/test/builtin/test_simd.mojo](https://github.com/modular/modular/compare/main...iMostfa:modular:fix-simd-pow-float-literal?expand=1#diff-aadb21e1549244f42cf08a5f6b5c64db3e08bf6475db2531f078fc0071db3ad3)`

## Checklist

      approval

### DIFF
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -1260,6 +1260,9 @@ struct SIMD[dtype: DType, length: SIMDLength](
         The literal is converted to `Self`, so the exponent always has the
         same dtype as the base.
 
+        Constraints:
+            The SIMD dtype must be floating point.
+
         Args:
             exp: The exponent value.
 
@@ -1267,6 +1270,9 @@ struct SIMD[dtype: DType, length: SIMDLength](
             A SIMD vector where each element is raised to the power of the
             specified exponent value.
         """
+        comptime assert (
+            Self.dtype.is_floating_point()
+        ), "a float literal exponent requires a floating point SIMD type"
         return _pow(self, Self(exp))
 
     # TODO(#22771): remove this overload.
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -1247,6 +1247,13 @@ struct SIMD[dtype: DType, length: SIMDLength](
         )
         comptime if Self.length == exp.length and self.dtype == exp.dtype:
             return _pow(self, rebind[Self](exp))
+        elif exp.dtype.is_floating_point() and Self.dtype.is_floating_point():
+            # `_pow` asks the base and exponent dtypes to match, so cast
+            # a mismatched float exponent (e.g. a float literal, which
+            # defaults to float64) to the base dtype.
+            return _pow(
+                self, Self(rebind[Scalar[exp.dtype]](exp).cast[Self.dtype]())
+            )
         else:
             return _pow(
                 self,
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -1247,18 +1247,27 @@ struct SIMD[dtype: DType, length: SIMDLength](
         )
         comptime if Self.length == exp.length and self.dtype == exp.dtype:
             return _pow(self, rebind[Self](exp))
-        elif exp.dtype.is_floating_point() and Self.dtype.is_floating_point():
-            # `_pow` asks the base and exponent dtypes to match, so cast
-            # a mismatched float exponent (e.g. a float literal, which
-            # defaults to float64) to the base dtype.
-            return _pow(
-                self, Self(rebind[Scalar[exp.dtype]](exp).cast[Self.dtype]())
-            )
         else:
             return _pow(
                 self,
                 SIMD[exp.dtype, self.length](rebind[Scalar[exp.dtype]](exp)),
             )
+
+    @always_inline("nodebug")
+    def __pow__(self, exp: FloatLiteral) -> Self:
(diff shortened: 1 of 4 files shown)


mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -1234,6 +1234,13 @@ struct SIMD[dtype: DType, length: SIMDLength](
         )
         comptime if Self.length == exp.length and self.dtype == exp.dtype:
             return _pow(self, rebind[Self](exp))
+        elif exp.dtype.is_floating_point() and Self.dtype.is_floating_point():
+            # `_pow` asks the base and exponent dtypes to match, so cast
```
JoeLoser: `minor` · non-blocking · comment

`_pow` requires matching dtypes only on its *float* path; the integral path deliberately accepts a mismatched `exp` and routes it to `_powi` / `base * base`. Stated as a blanket constraint, this invites a later "simplification" that folds the branch into the `else` and quietly drops integer exponents off that fast path.

```suggestion
            # `_pow`'s float path needs the base and exponent dtypes to match
            # (its integral path does not), so cast a mismatched float
            # exponent — e.g. a float literal, which defaults to float64.
```

mojo/stdlib/test/builtin/test_simd.mojo
```diff
@@ -1772,6 +1772,13 @@ def test_pow() raises:
         F32x4(0.0, 1.0, 1.414213562, 1.732050808),
     )
 
+    assert_almost_equal(Float16(4.0) ** 0.5, Float16(2.0))
+    assert_almost_equal(BFloat16(4.0) ** 0.5, BFloat16(2.0))
+    assert_almost_equal(Float32(4.0) ** 0.5, Float32(2.0))
+    assert_almost_equal(
+        f32x4_val**0.5, F32x4(0.0, 1.0, 1.414213562, 1.732050808)
+    )
```
JoeLoser: `suggestion` · non-blocking · testing

All four of these use a `float64` literal exponent. The new branch also takes over a case that previously went to the `else` — same dtype, different length, e.g. `f32x4_val ** Float32(0.5)`. One assertion on that pins the reroute, so a future refactor can't silently change which path it takes.

mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -1253,6 +1253,28 @@ struct SIMD[dtype: DType, length: SIMDLength](
...
+            exp: The exponent value.
+
+        Returns:
+            A SIMD vector where each element is raised to the power of the
+            specified exponent value.
+        """
+        comptime assert (
+            Self.dtype.is_floating_point()
+        ), "a float literal exponent requires a floating point SIMD type"
+        return _pow(self, Self(exp))
```
iMostfa, line 1276: i updated the approach, and now we prefer just `FloatLiteral` and we also have the `is_floating_point` as a compile-time check.

@barcharcraz 


i think this tightens up the situations we do the casting, wdyt ?

barcharcraz (approved):
This is an improvement, I'd like to see scalar bases in the "compile fail" tests, but I think this is mergable.

My only remaining concern is that overload resolution happens before we enforce this constraint, so we can't really "move on" in the case where "self" isn't floating point. I don't think that's a big deal here.

PR #6973, opened 2026-08-24
[stdlib][docs] Fix List docstring examples to annotate list literals

### LINKED ISSUE #6875: [Docs] List initialization needs a type spezifier
### Where is the problem?

https://mojolang.org/docs/std/collections/list/List/

### What can we do better?

According given documentation lists can be initialized with

`var str_list = ["a", "b", "c"]` but I tested this using version 1.0.0 which resulted in str_list being off type Array[String, 3], which cannot implicitly be converted to list. 

The correct way to initialize List[String] is, by stating the type:

`let str_list: List[String] = ["a", "b", c""]`

### Anything else?

_No response_

yassernamez03:
This fixes the `List` documentation examples in `mojo/stdlib/std/collections/list.mojo`.

In Mojo 1.0.0, bare bracket literals such as `[1, 2, 3]` create an `Array`, not a `List`. I updated the affected documentation examples to explicitly specify `List[T]` and corrected the owned-iteration examples so they match the current Mojo behavior.

## Linked issue

Fixes #6875

## Type of change

* [ ] Bug fix (non-breaking change that fixes an issue)
* [ ] Performance improvement (includes benchmark results below)
* [x] Documentation update
* [ ] New feature or public API (requires prior proposal or issue approval)
* [ ] Refactor / internal cleanup (no user-visible change)
* [ ] Build, CI, or tooling change

## Motivation

The `List` documentation currently contains examples that use bare bracket literals without specifying the `List` type. In Mojo 1.0.0, these literals are inferred as fixed-size `Array` values rather than `List` values. As a result, some of the documented examples do not work as described.

## What changed

* Added explicit `List[Int]` and `List[String]` type annotations to the affected examples.
* Corrected the owned-iteration examples to print the already-owned value directly.
* Updated the list multiplication example so that the `List` type is established before multiplication.

The changes are limited to the `List` class-level documentation examples.

## Testing

I tested the updated examples with Mojo 1.0.0 and verified that the corrected examples compile and run as expected.

## Checklist

* [x] The linked issue above has been reviewed by a maintainer and is
  approval
* [x] PR is small and focused — I've split larger changes into a sequence of
  [[pull request sizes](https://github.com/modular/modular/blob/main/mojo/CONTRIBUTING.md#about-pull-request-sizes)](https://github.com/modular/modular/blob/main/mojo/CONTRIBUTING.md#about-pull-request-sizes))
* [ ] I ran `./bazelw run format` to format my changes
* [ ] I added or updated tests to cover my changes
* [x] If AI tools assisted with this contribution, I have included an

### DIFF
--- a/mojo/stdlib/std/collections/list.mojo
+++ b/mojo/stdlib/std/collections/list.mojo
@@ -188,8 +188,8 @@ struct List[T: AnyType, /](
     # With initial size and fill value
     var filled = List[Float64](length=10, fill=0.0)
 
-    # With initial values and inferred type (Int)
-    var numbers = [1, 2, 3, 4, 5]
+    # With initial values and an explicit element type
+    var numbers: List[Int] = [1, 2, 3, 4, 5
     ```
 
     Be aware of the following characteristics:
@@ -199,8 +199,8 @@ struct List[T: AnyType, /](
       improves performance:
 
       ```mojo
-      var int_list = [1, 2, 3]        # List[Int]
-      var str_list = ["a", "b", "c"]  # List[String]
+      var int_list: List[Int] = [1, 2, 3]
+      var str_list: List[String] = ["a", "b", "c"]
       # var mixed = [1, "hello"]      # Error! All elements must be same type
       ```
 
@@ -213,7 +213,7 @@ struct List[T: AnyType, /](
       assignment creates a deep copy of all elements:
 
       ```mojo
-      var list1 = [1, 2, 3]
+      var list1: List[Int] = [1, 2, 3]
       var list2 = list1.copy()        # Deep copy
       list2.append(4)
       print(list1)   # => [1, 2, 3]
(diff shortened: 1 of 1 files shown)


mojo/stdlib/std/collections/list.mojo
```diff
@@ -323,9 +324,9 @@ struct List[T: AnyType, /](
         print(i, fruits[i])
 
     # Iterate by ownership (consumes the list)
-    var temps = ["a", "b", "c"]
+    var temps: List[String] = ["a", "b", "c"]
```
JoeLoser: `warning` · non-blocking · docs

The class docstring is fixed, but the per-method examples below it have the identical defect — `capacity`, `append`, `resize`, `shrink`, `reverse`, `count` and roughly twenty others all open with a bare `var list = [...]`. `List.append`'s own example doesn't compile today:

```
var list = [1, 2, 3, 4, 5]
list.append(6)  # 'Array[Int, Int(5)]' value has no attribute 'append'
```

The same one-line annotation fixes each, and doing them here beats leaving the file half-corrected.

PR #6974, opened 2026-08-24
[stdlib] Fix Int division by zero

### LINKED ISSUE #6898: [BUG] `Int` division by zero is silently undefined: returns garbage, returns 0, or crashes with an unsymbolicated stack trace
### Actual behavior
Dividing an `Int` by zero has no defined behaviour. Depending on how the
denominator reaches the division, the same operation either returns a
nondeterministic value, returns `0`, or crashes the process with a raw stack
dump. Nothing in the [operators documentation](https://docs.modular.com/mojo/manual/operators/)
mentions division by zero, so there is no way to know this without testing for it.

### Expected behavior
Python — whose syntax and semantics Mojo otherwise follows closely — raises
`ZeroDivisionError` here. Code ported from Python silently produces wrong
results instead of failing.

So either:

1. **Python parity** — raise `ZeroDivisionError`, or
2. if the branch is considered too expensive for hot paths: **document it as
   undefined behaviour** in the operators manual, and trap it in debug builds
   (as Rust does), so it fails loudly during development rather than corrupting
   data silently.

The current state is the worst of both: undocumented, inconsistent between
operators, and inconsistent between runs of the same program.

### A. Hard crash

```mojo
def divide(numerator: Int, denominator: Int) -> Int:
    return numerator / denominator

def main():
    var values = [7, 0, 3]
    for v in values:
        print(v, "->", divide(100, v))
```

Output (identical across 3 runs):

```
7 -> 14 #0 0x000076c2505bef8e (.../libKGENCompilerRTShared.so+0xfbf8e)
#1 0x000076c2505bc0a6 (.../libKGENCompilerRTShared.so+0xf90a6)
#2 0x000076c2505bfdd0 (.../libKGENCompilerRTShared.so+0xfcdd0)
#3 0x000076c253645330 (/lib/x86_64-linux-gnu/libc.so.6+0x45330)
#4 0x000076c21c0011a7
mojo: error: execution crashed
```

The message never mentions division by zero.

### B. Silent nondeterministic value

```mojo
def divide(numerator: Int, denominator: Int) -> Int:
    return numerator / denominator

def main():
    var segments = 7
    var duration = 5
    var steps = duration / segments   # == 0
    print("computed denominator:", divide(3, steps))

    var zero = 0
    print("constant denominator:", divide(3, zero))
```

Three consecutive runs:

```
computed denominator: 0   constant denominator: 140736790674968
computed denominator: 0   constant denominator: 140726301657880
computed denominator: 0   constant denominator: 140736819750648
```

The second value changes on every run and looks like a stack address. Which of
the two operands produces garbage versus `0` is not stable either — in a
slightly different arrangement of the same code the roles were reversed.

### C. `//` and `%`

With the same list-based shape that crashes for `/`, both `//` and `%` instead
return `0` silently:

```mojo
def main():
    var values = [0]
    for v in values:
        print("floordiv:", 100 // v)   # 0
        print("mod:", 100 % v)         # 0
```

### System information

System
------------
       Pixi version: 0.76.0
        TLS backend: rustls
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.8.0=0
                   : __glibc=2.39=0
                   : __archspec=1=skylake
          Cache dir: /home/stefan/.cache/rattler/cache
       Auth storage: /home/stefan/.rattler/credentials.json

Global
------------
            Bin dir: /home/stefan/.pixi/bin
    Environment dir: /home/stefan/.pixi/envs
       Manifest dir: /home/stefan/.pixi/manifests/pixi-global.toml

As I am using uv:
uv pip list
Package                 Version
----------------------- -------
click                   8.4.2
mblack                  26.5.0
mojo                    1.0.0
mojo-compiler           1.0.0
mojo-compiler-mojo-libs 1.0.0
mojo-lldb-libs          1.0.0
mypy-extensions         1.1.0
pathspec                1.1.1
platformdirs            4.11.2

mahendrarathore1742:
### Summary
    This PR addresses undefined behavior when dividing `Int` by zero. Previously, integer division by zero    
  resulted in silent 0 returns, garbage values, or unsymbolicated crashes depending on the operator and       
  codegen path. 
  
    To ensure safe, deterministic failure on invalid input, division by zero for scalar integers now          
  immediately and safely traps execution, avoiding silent data corruption.
  
    ### Changes
    - Updated `__truediv__`, `__floordiv__`, `__mod__`, and `__divmod__` in `std.builtin.simd` to trap via    
  `_abort_base()` when evaluating scalar integers with a `0` denominator.
    - Changed `@always_inline("builtin")` to `@always_inline("nodebug")` in `__truediv__` to support control  
  flow for the zero check.
    - Added a `FileCheck` test (`test_int_div_by_zero.mojo`) to ensure the trap triggers deterministically on 
  scalar division by zero.
    - Removed outdated `assert_raises` zero-division tests from `test_uint.mojo` and `test_int.mojo`.         
    - Updated standard library changelog and operator proposals to document the new edge case behavior.       

   Fixes #6898

### DIFF
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -51,6 +51,7 @@ from std.collections.string.string import (
 from std.hashlib.hasher import Hasher
 from std.math import Ceilable, CeilDivable, Floorable, Truncable
 from std.math.math import _call_ptx_intrinsic, trunc
+from std.os.os import _abort_base
 from std.sys import (
     CompilationTarget,
     _RegisterPackType,
@@ -1099,7 +1100,7 @@ struct SIMD[dtype: DType, length: SIMDLength](
             mlir_value=__mlir_op.`pop.mul`(self._mlir_value, rhs._mlir_value)
         )
 
-    @always_inline("builtin")
+    @always_inline("nodebug")
     def __truediv__(self, rhs: Self) -> Self:
         """Computes `self / rhs`.
 
@@ -1111,6 +1112,9 @@ struct SIMD[dtype: DType, length: SIMDLength](
             `self[i] / rhs[i]`.
         """
         comptime assert Self.dtype.is_numeric(), "the SIMD type must be numeric"
+        comptime if Int(Self.length) == 1 and Self.dtype.is_integral():
+            if rhs == Self(0):
+                _abort_base()
         return Self(
             mlir_value=__mlir_op.`pop.div`(self._mlir_value, rhs._mlir_value)
         )
@@ -1130,6 +1134,10 @@ struct SIMD[dtype: DType, length: SIMDLength](
             `floor(self / rhs)` value.
         """
         comptime assert Self.dtype.is_numeric(), "the type must be numeric"
+
+        comptime if Int(Self.length) == 1 and Self.dtype.is_integral():
+            if rhs == Self(0):
+                _abort_base()
 
         var is_zero_mask = Self._Mask(fill=Self.dtype.is_integral()) and rhs.eq(
             0
@@ -1154,6 +1162,10 @@ struct SIMD[dtype: DType, length: SIMDLength](
             The remainder of dividing self by rhs.
         """
         comptime assert Self.dtype.is_numeric(), "the type must be numeric"
+
+        comptime if Int(Self.length) == 1 and Self.dtype.is_integral():
+            if rhs == Self(0):
+                _abort_base()
 
         var is_zero_mask = Self._Mask(fill=Self.dtype.is_integral()) and rhs.eq(
             0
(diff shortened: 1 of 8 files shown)


mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -1099,7 +1100,7 @@ struct SIMD[dtype: DType, length: SIMDLength](
             mlir_value=__mlir_op.`pop.mul`(self._mlir_value, rhs._mlir_value)
         )
 
-    @always_inline("builtin")
+    @always_inline("nodebug")
```
barcharcraz, line 1103: I'm not sure we can make this no-longer built-in.

bgreni: Yeah I agree that is probably a non-starter

NathanSWard: If it a trade off for safety vs compile-time capabilities IMO it's okay to make this non-builtin.

NathanSWard: We can provide "safe" and "builtin" ways to do this at compile time. Via like `a.checked_div(b)` or `a.unsafe_div(b)` which could be builtins.

PR #6978, opened 2026-08-25
[stdlib] Abort on negative length in `List.shrink`

### LINKED ISSUE #6857: [BUG] List.shrink with a negative length destroys an element twice and leaves len() negative
### Bug description

**Actual behavior**

`List.shrink` with a negative argument runs one destructor too many, destroys an element twice, and leaves `len()` negative.

```mojo
struct Obs(Deinitable, Movable):
    var id: Int

    def __init__(out self, id: Int):
        self.id = id

    def __deinit__(deinit self):
        print("deinit", self.id)

def main():
    var l = List[Obs]()
    l.append(Obs(0))
    l.append(Obs(1))
    print("shrinking to -1")
    l.shrink(-1)
    print("len =", len(l))
```

```text
shrinking to -1
deinit 0
deinit 0
deinit 1
len = -1
```

Two live objects produce three destructor calls, and element 0 is destroyed twice. The list is left reporting a negative length, so every subsequent length-derived computation on it is wrong.

This is reachable from public API without touching `shrink` directly:

```mojo
from std.collections import Counter

def main() raises:
    _ = Counter[Int](1, 1, 2).most_common(-1)
```

**Expected behavior**

A negative length aborts, or clamps to 0. Exactly two destructors run, both within `[0, len)`, and `len()` is never negative. Python's `Counter.most_common(n)` returns `[]` for `n <= 0`.

### Steps to reproduce

Save either snippet and run `mojo run <file>`.

`shrink` has a guard, but it only rejects a new length *greater* than the current one, so a negative value passes through and is then used as a loop bound and written to the length field.

This is adjacent to modular/modular#6832, which covered `most_common(n)` aborting when `n` exceeds the number of unique keys. That is a different path — this one is the negative side, and it corrupts the list rather than aborting. Clamping `n` to `[0, len]` in `most_common` fixes the public entry point, but `shrink` should reject a negative length on its own regardless, since it is the one performing the destruction.

### System information

Reproduces at the current head of `main` built from source — `mojo --version` reports `Mojo 1.1.0.dev0` — on macOS (Apple Silicon). Platform-independent.

chiliec:
## Problem

`List.shrink` guards only against a `new_length` **greater** than the current length. A negative `new_length` slips past that check and is then used directly as the destroy offset/count and written to `_len`:

```mojo
unsafe_destroy_n(self._data.unsafe_offset(new_length), count=len(self) - new_length)
...
self._len = new_length
```

With two live elements, `shrink(-1)` runs **three** destructor calls (double-destroying element 0) and leaves `len()` at `-1`, corrupting every subsequent length-derived computation. It's reachable without calling `shrink` directly — any path that forwards an unclamped value into it.

Repro from #6857:

```mojo
struct Obs(Deinitable, Movable):
    var id: Int
    def __init__(out self, id: Int): self.id = id
    def __deinit__(deinit self): print("deinit", self.id)

def main():
    var l = List[Obs]()
    l.append(Obs(0)); l.append(Obs(1))
    l.shrink(-1)                 # double-destroys, len() == -1
    print("len =", len(l))
```

## Fix

Add a guard that aborts when `new_length < 0`, mirroring the existing over-length abort, and document the behavior in the docstring. A negative length is a programming error (like an over-length shrink), so aborting is consistent with the sibling case rather than silently clamping.

Note: `Counter.most_common` — the public path called out in the issue — already clamps with `max(0, min(n, len(items)))` at HEAD, so this change hardens `shrink` itself, which the issue explicitly asks for ("`shrink` should reject a negative length on its own regardless, since it is the one performing the destruction").

## Test

`test_list_shrink_negative_abort.mojo` (FileCheck, registered in `_FILECHECK_TESTS` + `_EXPECT_CRASH`) asserts the abort message fires and that **no** destructor runs before it (`CHECK-NOT: deinit`).

Verified locally against a patched `std` built with the pinned nightly (`Mojo 1.1.0.dev2026082405`):
- Before the fix: `shrink(-1)` corrupts the list / crashes; the test's `CHECK` for the abort message does not match (test bites).
- After the fix: `shrink(-1)` aborts cleanly with the new message and no element destructor runs first; normal `shrink` paths (e.g. `shrink(2)`, `shrink(0)`) are unchanged.

I could not run the full Bazel suite in my environment, so CI here is the authoritative signal for the wider stdlib.

Fixes #6857

### DIFF
--- a/Mojo/stdlib/std/collections/list.mojo
+++ b/Mojo/stdlib/std/collections/list.mojo
@@ -1262,7 +1262,8 @@ struct List[T: AnyType, /](
             With no new value provided, the new length must be smaller than or
             equal to the current one. Elements at the end are discarded.
 
-            Calls abort() if the new length is larger than the current length.
+            Calls abort() if the new length is larger than the current length,
+            or if the new length is negative.
 
         Examples:
 
@@ -1278,6 +1279,13 @@ struct List[T: AnyType, /](
                 " current size. If you want to make the List bigger, provide a"
                 " value to fill the new slots with. If not, make sure the new"
                 " size is smaller than the current size."
+            )
+
+        if new_length < 0:
+            abort(
+                t"You are calling List.shrink with a negative value:"
+                t" {new_length}. The new size must be in the range [0,"
+                t" {len(self)}]."
             )
 
         unsafe_destroy_n(
--- a/Mojo/stdlib/test/collections/test_list_shrink_negative_abort.mojo
+++ b/Mojo/stdlib/test/collections/test_list_shrink_negative_abort.mojo
@@ -1,0 +1,36 @@
+# ===----------------------------------------------------------------------=== #
+# Copyright (c) 2026, Modular Inc. All rights reserved.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions:
+# https://llvm.org/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===----------------------------------------------------------------------=== #
+#
+# Regression test: `List.shrink` with a negative `new_length` used to pass the
+# over-length guard, then run one destructor too many (destroying an element
+# twice) and store a negative value in `len()`. It must now abort instead of
+# corrupting the list.
+#
+# ===----------------------------------------------------------------------=== #
+
+from std.testing import _assert_aborts, TestSuite
+
+
+def test_list_aborts_on_negative_shrink_size() raises:
+    var l: List = [1, 2, 3]
+
+    def shrink() raises {mut l} -> None:
+        l.shrink(-1)
+
+    _assert_aborts(
+        shrink, contains="You are calling List.shrink with a negative"
+    )
+
+
+def main() raises:
+    TestSuite.discover_tests[__functions_in_module()]().run()


mojo/stdlib/std/collections/list.mojo
```diff
@@ -1232,6 +1233,12 @@ struct List[T: AnyType, /](
                 " size is smaller than the current size."
             )
 
+        if new_length < 0:
+            abort(
+                "You are calling List.shrink with a negative new_size. The new"
```
JoeLoser: `minor` · non-blocking · naming

The parameter is `new_length`, so `new_size` sends the reader looking for an argument that doesn't exist. The neighbouring message has the same drift, but a new message doesn't have to inherit it.

```suggestion
                "You are calling List.shrink with a negative new_length. The"
                " new length must be in the range [0, len(self)]."
```

mojo/stdlib/test/collections/test_list_shrink_negative_abort.mojo
```diff
@@ -0,0 +1,47 @@
...
+
+    var l = List[Obs]()
+    l.append(Obs(0))
+    l.append(Obs(1))
+
+    # No destructor runs before the abort fires (the old bug ran three
+    # destructors here, double-destroying element 0).
+    # CHECK-NOT: deinit
+    # CHECK: You are calling List.shrink with a negative new_size
+    l.shrink(-1)
```
JoeLoser: `suggestion` · non-blocking · testing

`resize(unsafe_uninit_length=-1)` reaches this same guard (`-1 <= self._len` forwards straight into `shrink`), and that is the path a caller is far likelier to hit than calling `shrink(-1)` directly. Worth a second file covering it, so the hardening of that entry point can't be dropped later without a test noticing.

Mojo/stdlib/std/collections/list.mojo
```diff
@@ -1232,6 +1233,12 @@ struct List[T: AnyType, /](
                 " size is smaller than the current size."
             )
 
+        if new_length < 0:
+            abort(
+                "You are calling List.shrink with a negative new_size. The new"
+                " size must be in the range [0, len(self)]."
+            )
```
NathanSWard, line 1289: **suggestion**: Can we include the value of `new_length` and the current length in the error message?

```suggestion
            abort(
                t"You are calling List.shrink with a negative value: {new_length}. The new size must be in the range [0, {len(self)}]."
            )
```

Mojo/stdlib/test/collections/test_list_shrink_negative_abort.mojo
```diff
@@ -0,0 +1,47 @@
+# ===----------------------------------------------------------------------=== #
```
NathanSWard, line 1: **suggestion (blocking)**:
For the abort test, we don't use the file-check stuff anymore, but instead this should just be a normal test file but you can use the `std.testing._assert_aborts` testing infra.

```
from std.testing import _assert_aborts

def test_list_aborts_on_negative_shrink_size() raises:
    var list = [1, 2, 3]
    def shrink() {imm}:
        list.shrink(-1)
    _assert_aborts(shrink, contains="You are calling List.shrink with a negative")
```

NathanSWard (changes requested):

NathanSWard (approved):

NathanSWard:
!sync

PR #6980, opened 2026-08-25
[stdlib] Read unsigned SIMD dtypes from Python via `PyLong_AsSize_t`

### LINKED ISSUE #6850: [BUG] SIMD(py=) reads all integral dtypes through signed PyLong_AsSsize_t: UInt64 >= 2**63 fails to round-trip, negative Python ints silently wrap
### Actual behavior

`SIMD.__init__(*, py=)` converts every integral dtype through the **signed** `PyLong_AsSsize_t`,
with no unsigned branch. Two symptoms follow:

1. A negative Python int silently wraps into an unsigned Mojo scalar instead of raising.
2. `Mojo -> Python -> Mojo` does not round-trip for unsigned values `>= 2**63`, even though the
   Mojo -> Python direction deliberately uses the *unsigned* entry point `PyLong_FromSize_t`.

```
UInt64(py=PythonObject(-1)) = 18446744073709551615
Mojo -> Python: 18446744073709551615
Python -> Mojo raised: Python int too large to convert to C ssize_t
```

Left uncaught, the second one terminates the program:

```
Unhandled exception caught during execution: Python int too large to convert to C ssize_t
mojo: error: execution exited with a non-zero result: 1
```

### Expected behavior

- `UInt64(py=PythonObject(UInt64.MAX))` should return `18446744073709551615`. The value survived the
  Mojo -> Python leg intact; only the read-back fails.
- `UInt64(py=PythonObject(-1))` should raise, matching CPython's `PyLong_AsSize_t`, which raises
  `OverflowError` for negative values — not silently produce the unsigned wraparound.

## Steps to reproduce

Repro file:

```mojo
from std.python import PythonObject


def main() raises:
    # 1. A negative Python int silently wraps instead of raising.
    print("UInt64(py=PythonObject(-1)) =", UInt64(py=PythonObject(-1)))

    # 2. Mojo -> Python -> Mojo does not round-trip for values >= 2**63.
    var py = PythonObject(UInt64.MAX)
    print("Mojo -> Python:", py)
    try:
        print("Python -> Mojo:", UInt64(py=py))
    except e:
        print("Python -> Mojo raised:", e)
```

Command:

```bash
mojo run uint64_py_roundtrip.mojo
```

### Where it comes from

As of 1.0.0rc0:

- Mojo -> Python (correct, unsigned) — `mojo/stdlib/std/python/python_object.mojo:257-262`:
  `elif dtype.is_unsigned(): var val = c_size_t(value.cast[DType.uint]()); self = Self(from_owned=cpy.PyLong_FromSize_t(val))`
- Python -> Mojo (missing the unsigned branch) — `mojo/stdlib/std/builtin/simd.mojo:984-985`:
  `elif Self.dtype.is_integral() and bit_width_of[Self.dtype]() <= 64: self = Scalar[Self.dtype](Python.py_long_as_ssize_t(py.__int__()))`
- `Python.py_long_as_ssize_t` (`mojo/stdlib/std/python/python.mojo:572-591`) is a thin wrapper over
  `PyLong_AsSsize_t`.

The asymmetry is untested: `mojo/stdlib/test/python/test_python_object.mojo:171-174`
(`test_num_conversion`) only stringifies `PythonObject(UInt64(0xFEDCBA0987654321))` and never
converts it back.

Separately, `SIMD.__init__(*, py:)`'s docstring documents only `Raises: If the conversion to double
fails`, which does not describe the integral path at all.

### Suggested fix

Give the integral path an unsigned branch that goes through `PyLong_AsSize_t` (mirroring the
`PyLong_FromSize_t` used in the other direction), keep the signed path on `PyLong_AsSsize_t`, and
update the `Raises:` clause to cover integral overflow.

## System information

- Mojo 1.0.0rc0 (73af5f88), installed via `pixi` from the `max-nightly` conda channel
- Platform: macOS (osx-arm64), Darwin 24.6.0

chiliec:
## Problem

`SIMD.__init__(*, py=)` converts every integral dtype through the **signed** `PyLong_AsSsize_t`, with no unsigned branch. Two bugs follow:

1. A negative Python int silently wraps into an unsigned Mojo scalar instead of raising.
2. `Mojo -> Python -> Mojo` does not round-trip for unsigned values `>= 2**63`. The Mojo→Python leg deliberately uses the unsigned `PyLong_FromSize_t`, but the read-back overflows `Py_ssize_t`:

```mojo
from std.python import PythonObject

def main() raises:
    print(UInt64(py=PythonObject(-1)))          # was: 18446744073709551615 (silent wrap)
    var py = PythonObject(UInt64.MAX)           # Mojo -> Python: fine
    print(UInt64(py=py))                         # was: raises "too large to convert to C ssize_t"
```

Left uncaught, the second case terminates the program.

## Fix

- Bind CPython's `PyLong_AsSize_t` in `_cpython.mojo` (declaration + field + load + wrapper), mirroring the existing `PyLong_AsSsize_t` binding exactly.
- Add `Python.py_long_as_size_t`, which surfaces the `(size_t)-1` error sentinel via `PyErr_Occurred()` (the unsigned sentinel is `size_t.MAX`, not `-1`).
- Route **unsigned** integral dtypes in `SIMD(py=)` through the unsigned helper via a `comptime` branch; signed dtypes keep the signed path, so negative values still convert correctly for them.

This matches CPython semantics: `PyLong_AsSize_t` raises `OverflowError` for negative inputs, and unsigned values up to `2**64 - 1` read back losslessly — symmetric with the `PyLong_FromSize_t` used on the way out.

## Test

Adds `test_scalar_from_python_unsigned` to `test_python_to_mojo.mojo`: `UInt64.MAX` and `2**63` round-trips, `UInt32.MAX`, the negative-raises case, and that signed `Int64` still accepts `-1`.

Verified locally against a patched `std` built with the pinned nightly (`Mojo 1.1.0.dev2026082405`), with a real CPython:
- Full `test_python_to_mojo.mojo`: **8 passed, 0 failed** with the fix.
- Sabotage check: reverting the SIMD branch to signed-only makes `test_scalar_from_python_unsigned` fail (7 passed, 1 failed), confirming the test bites.

I could not run the full Bazel suite in my environment, so CI here is the authoritative signal for the broader stdlib.

Fixes #6850

### DIFF
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -1003,7 +1003,19 @@ struct SIMD[dtype: DType, length: SIMDLength](
             # NOTE: if dtype is not float64, we truncate.
             self = Scalar[Self.dtype](float_value)
         elif Self.dtype.is_integral() and bit_width_of[Self.dtype]() <= 64:
-            self = Scalar[Self.dtype](Python.py_long_as_ssize_t(py.__int__()))
+            comptime if Self.dtype.is_unsigned():
+                # Read unsigned dtypes through the unsigned entry point so that
+                # values in `[2**63, 2**64)` round-trip (the signed path
+                # overflows on them) and negative Python ints raise instead of
+                # silently wrapping. Mirrors the Mojo -> Python direction, which
+                # already uses the unsigned `PyLong_FromSize_t`.
+                self = Scalar[Self.dtype](
+                    Python.py_long_as_size_t(py.__int__())
+                )
+            else:
+                self = Scalar[Self.dtype](
+                    Python.py_long_as_ssize_t(py.__int__())
+                )
         else:
             self = Scalar[Self.dtype]()
             comptime assert False, "unsupported dtype"
--- a/mojo/stdlib/std/builtin/simd.mojo
+++ b/mojo/stdlib/std/builtin/simd.mojo
@@ -990,7 +990,8 @@ struct SIMD[dtype: DType, length: SIMDLength](
             py: The PythonObject to convert.
 
         Raises:
-            If the conversion to double fails.
+            If the conversion to double fails, or if the value is out of range
+            for an unsigned dtype (a negative or too-large Python int).
         """
 
         comptime if Self.dtype.is_floating_point():
--- a/mojo/stdlib/test/python/test_python_to_mojo.mojo
+++ b/mojo/stdlib/test/python/test_python_to_mojo.mojo
@@ -88,11 +88,6 @@ def test_numpy_float() raises:
 
 
 def test_scalar_from_python_unsigned() raises:
-    # Unsigned scalars must read through the unsigned CPython entry point so
-    # that Mojo -> Python -> Mojo round-trips for values >= 2**63 and negative
-    # Python ints raise instead of silently wrapping (regression test for
-    # #6850).
-
     # Values in [2**63, 2**64) survived the Mojo -> Python leg (via the
     # unsigned `PyLong_FromSize_t`) and must survive the read-back too.
     assert_equal(UInt64(py=PythonObject(UInt64.MAX)), UInt64.MAX)
--- a/mojo/stdlib/test/python/test_python_to_mojo.mojo
+++ b/mojo/stdlib/test/python/test_python_to_mojo.mojo
@@ -87,5 +87,28 @@ def test_numpy_float() raises:
     assert_equal(Float64(py=py_numpy_float), mojo_float)
 
 
+def test_scalar_from_python_unsigned() raises:
+    # Unsigned scalars must read through the unsigned CPython entry point so
+    # that Mojo -> Python -> Mojo round-trips for values >= 2**63 and negative
+    # Python ints raise instead of silently wrapping (regression test for
+    # #6850).
+
+    # Values in [2**63, 2**64) survived the Mojo -> Python leg (via the
+    # unsigned `PyLong_FromSize_t`) and must survive the read-back too.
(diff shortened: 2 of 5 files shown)


mojo/stdlib/std/builtin/simd.mojo
```diff
@@ -1003,7 +1003,19 @@ struct SIMD[dtype: DType, length: SIMDLength](
             # NOTE: if dtype is not float64, we truncate.
             self = Scalar[Self.dtype](float_value)
         elif Self.dtype.is_integral() and bit_width_of[Self.dtype]() <= 64:
-            self = Scalar[Self.dtype](Python.py_long_as_ssize_t(py.__int__()))
+            comptime if Self.dtype.is_unsigned():
```
JoeLoser, line 1007: **Suggestion** (Non-blocking): The constructor's `Raises:` section above still says only "If the conversion to double fails." After this change, an unsigned dtype also raises on a negative or out-of-range Python int. Mind updating the docstring here?

mojo/stdlib/test/python/test_python_to_mojo.mojo
```diff
@@ -87,5 +87,28 @@ def test_numpy_float() raises:
     assert_equal(Float64(py=py_numpy_float), mojo_float)
 
 
+def test_scalar_from_python_unsigned() raises:
+    # Unsigned scalars must read through the unsigned CPython entry point so
```
JoeLoser: **Suggestion** Drop this test comment as it's clear from the test case name and body.

JoeLoser (approved):
Thanks for working on this! The fix looks good to me, please just address my suggestions before I land this internally.

It may not be bad to add a small changelog entry under the Fixed section as well.

Issue #6781, opened 2026-07-15
[BUG]: `ops.resize` fails to compile on GPU for NEAREST and BILINEAR interpolation types


bgreni:
### Bug description

The `resize` op fails to compile on GPU when it works fine on CPU.

FYI I did have Claude sketch up a patch but its a fairly sizeable change so I have't opened a PR yet.

### Steps to reproduce

```python

import numpy as np
from max.driver import CPU, Accelerator, Buffer, accelerator_count
from max.dtype import DType
from max.engine import InferenceSession
from max.graph import DeviceRef, Graph, KernelLibrary, TensorType, ops

SHAPE = [1, 3, 32, 32]
OUT = [1, 3, 64, 64]


def try_compile(device, mode):
    ref = DeviceRef.from_device(device)
    graph = Graph(
        f"resize_{mode}",
        input_types=[TensorType(DType.float32, shape=SHAPE, device=ref)],
        kernel_library=KernelLibrary(),
    )
    with graph:
        x, = graph.inputs
        graph.output(
            ops.resize(
                x.tensor,
                shape=OUT,
                interpolation=getattr(ops.InterpolationMode, mode),
            )
        )
    model = InferenceSession(devices=[device]).load(graph)
    array = np.random.rand(*SHAPE).astype(np.float32)
    return model.execute(Buffer.from_numpy(array).to(device))[0]


devices = [("CPU", CPU()), ("Accelerator", Accelerator())]

for name, device in devices:
    for mode in ("NEAREST", "BILINEAR", "BICUBIC"):
        try:
            try_compile(device, mode)
            print(f"  {name:6s} {mode:9s} OK")
        except Exception as error:
            print(f"  {name:6s} {mode:9s} FAIL -- {error}")
```

```
 CPU    NEAREST   OK
  CPU    BILINEAR  OK
  CPU    BICUBIC   OK
  Accelerator NEAREST   FAIL -- Failed to compile the model. Please file an issue, all models should be correct by construction and this error should have been caught during construction.
For more detailed failure information enable the `max-debug.source-tracebacks` config key (for example, `Graph.debug.source_tracebacks = True` or `MODULAR_DEBUG=source-tracebacks`).
  Accelerator BILINEAR  FAIL -- Failed to compile the model. Please file an issue, all models should be correct by construction and this error should have been caught during construction.
[...]

PR #6981, opened 2026-08-25
[kernels] Added GPU target for resize with nearest neighbours interpolation method

gsmyridis:
## Linked issue

Partially addresses #6781.

This PR implements the public, kernel-side portion of GPU support for nearest-neighbor resize. The issue also covers linear interpolation, which is outside the scope of this change.

An accompanying internal graph-compiler change is required to remove the `MO_HostOnly` constraint from `mo.resize.nearest`( @joshpeterson , see also comment from @JoeLoser ).

## Type of change

- [ ] Documentation update

## Motivation

Nearest-neighbor resize currently cannot be used with accelerator-backed MAX graph tensors. This affects every accelerator backend, not only Metal (see issue #6781).

The graph compiler rejects the operation during device placement because `mo.resize.nearest` is declared host-only. On the kernel side, `resize_nearest_neighbor` previously instantiated `elementwise` with its default `target="cpu"` because the selected graph-compiler target was not forwarded through `ResizeNearest.execute`.

The nearest-neighbor implementation is already based on `elementwise`, so it can support accelerators by threading the selected target through the graph compiler and kernel.

## What changed

- Added a `target` parameter to `resize_nearest_neighbor`.
- Forwarded the target to the enclosing `elementwise` invocation.
- Updated `ResizeNearest.execute` to accept and forward the graph compiler’s
  selected target.
- Clamped computed source coordinates to `[0, input_dim - 1]`. This prevents a
  negative index when half-pixel transformation is combined with `Floor`
  rounding during upsampling.
- Added a GPU kernel test covering:
  - Default half-pixel and `HalfDown` nearest-neighbor upsampling.
  - Half-pixel and `Floor` rounding, including the lower-bound clamp.

The existing CPU target remains the default.

## Testing

The following targets pass:

- `./bazelw test //max/kernels/test/gpu/nn:test_resize_nearest.mojo.test`
- `./bazelw test //max/kernels/test/nn:test_resize.mojo.test`

The graph-compiler kernel registration also builds successfully:

- `./bazelw build //max:builtin_kernels`

Linear interpolation requires a separate GPU kernel and is not changed by this PR. I will work on this next.

## Checklist

      approval


max/kernels/src/nn/resize.mojo
```diff
@@ -230,7 +234,9 @@ def resize_nearest_neighbor[
         output.raw_store(out_idx, input.ptr[in_idx])
 
     # TODO (#21439): can use unsafe_memcpy when scale on inner dimension is 1
```
gsmyridis: I can add this as well. Is it blocked by something?

max/kernels/src/nn/resize.mojo
gsmyridis: I was wondering, is there a reason we have this here instead of a method to `RoundMode`?

gsmyridis: What does `@__parameter` do here?

PR #7012, opened 2026-08-27
[stdlib] Fix Python finalization by running global destructors (#6860)     

### LINKED ISSUE #6860: [BUG] The CPython interpreter is never finalized, so buffered Python output is lost when stdout is a pipe
### Bug description

**Actual behavior**

The CPython interpreter is never finalized when a Mojo process exits — `Py_FinalizeEx` does not run. So anything still sitting in Python's stdout buffer is discarded, and Python `atexit` handlers never fire.

CPython line-buffers stdout when it is a terminal and block-buffers it when it is a pipe, so the symptom appears exactly when output is redirected:

```mojo
from std.python import Python

def main() raises:
    _ = Python.evaluate("print('hello from Python')")
    print("hello from Mojo")
```

Run with stdout piped:

```text
hello from Mojo
```

The Python line is gone. Run directly to a terminal and both lines appear, which is what makes this easy to miss — it works interactively and fails under `|`, `>`, `$(...)`, and every CI log capture.

**Expected behavior**

Both lines, in order, regardless of whether stdout is a terminal or a pipe. Python `atexit` handlers should also run.

### Steps to reproduce

```bash
mojo run repro.mojo          # both lines
mojo run repro.mojo | cat    # only the Mojo line
```

The interpreter is initialized lazily and stored in a global, but nothing calls `Py_FinalizeEx` on the way out. Beyond lost output, that also means any Python-side cleanup registered with `atexit` is skipped — including buffered file writes from Python code, not just stdout.

A caller can work around it today by flushing explicitly from Python before returning, but that requires knowing the problem exists, and it does not help with `atexit`.

### System information

Reproduces at the current head of `main` built from source — `mojo --version` reports `Mojo 1.1.0.dev0` — on macOS (Apple Silicon). The buffering behavior that exposes it is CPython's and is platform-independent.

mahendrarathore1742:
### Description
    This fixes a bug where the embedded CPython interpreter was never finalized when a Mojo process exited. Previously, this     
  caused CPython's block-buffered `stdout` (used in piped or CI environments) to silently drop output, and prevented Python  `atexit` handlers from running.
  
    **Root cause:** 
    The global cleanup mechanism (`_deinit_wrapper` in `std/ffi/__init__.mojo`) was properly freeing the memory backing Mojo globals, but it was not invoking their destructors. This meant `_PythonGlobal.__deinit__` was skipped on exit.                 
  
    **Fix details:**
    * Updated `_deinit_wrapper` to safely call `unsafe_deinit_pointee()` right before `dealloc`. This ensures `_PythonGlobal.    
  __deinit__` runs, which safely calls `CPython.destroy()` and `Py_FinalizeEx()`.
    * Added a guard inside `CPython.destroy()` to skip finalization if the interpreter failed to initialize (`init_error`).      
    * Added return code checking for `Py_FinalizeEx` to print a warning to `stderr` if it fails.
  
    ### Testing
    * Added `test_python_finalize.mojo`, which verifies:
      1. Python `print()` correctly flushes its output without dropping characters.
      2. Python `atexit` handlers successfully trigger on process exit (verified end-to-end using a subprocess).                 
    * Ran the full `//mojo/stdlib/test/python/...` test suite across multiple Python versions locally (all 56/56 tests passed).  
Resolves #6860


Mojo/stdlib/std/ffi/__init__.mojo
```diff
@@ -1048,14 +1048,17 @@ struct _Global[
...
         opaque_ptr: OptionalPointer[NoneType, UntrackedOrigin[mut=True]]
     ):
-        # Deinitialize and deallocate the storage.
+        # Destroy the value, then deallocate the storage.
         if opaque_ptr:
+            var typed_ptr = opaque_ptr.unsafe_value().unsafe_bitcast[
+                Self.StorageType
+            ]()
+            comptime if conforms_to(Self.StorageType, Deinitable):
+                typed_ptr.unsafe_deinit_pointee()
```
alisterburt, line 1057: unfortunately this introduces some undefined behavior, `OwnedDLHandle` instances are also stored in the global table can be null so your `unsafe_deinit_pointee()` can dereference a null pointer

e.g. from `_find_dylib()` in this file
```mojo
def _find_dylib[
    name: StaticString = "", abort_on_failure: Bool = True
](paths: List[Path]) -> OwnedDLHandle:
    try:
        return _try_find_dylib[name](paths)
    except e:
        comptime if abort_on_failure:
            abort(String(e))
        else:
            return OwnedDLHandle(unsafe_uninitialized=True)
```

Added #7063 to track and need that to be solved before this can go in

alisterburt (changes requested):
Hey @mahendrarathore1742 ! Thanks for the PR 🙂 Generally looks good but unfortunately has to stay blocked for now on #7063. Feel free to take that on if you'd like to get this in!

I added a few comments. Would you mind explaining the test failures you saw with your subprocess test that was removed? I'm not sure the test that is leftover correctly tests the functionality and the PR body still references the removed test.

We will need to check this pretty carefully internally as it means we are now unmapping dynamically linked libraries correctly at process end where we weren't before, apparently for CUDA libs (which we load for MAX) this dlclose is a source of segfaults and hangs

Mojo/stdlib/std/ffi/__init__.mojo
mahendrarathore1742: Hi @alisterburt,

Thanks for the review — I've addressed the feedback:

1. Added a null guard in `_DLHandle.close()` so it safely skips calling `dlclose` on uninitialized handles, fixing #7063.
2. Fully restored `test_python_atexit_via_subprocess()` to properly verify that Python's `atexit` handlers fire during `Py_FinalizeEx`.
3. Rebased onto `main` and resolved the `mojo/` → `Mojo/` directory rename conflicts.

Ready for another look — thanks again!

PR #7023, opened 2026-08-29
[stdlib] Migrate string span abort tests to _assert_aborts

zhoujingfighting:
Convert the lit-based abort tests in collections/string to use std.testing._assert_aborts(). Each test case is now a separate test_ function discovered by TestSuite:
- test_string_span_bounds_abort.mojo (4 cases)
- test_string_span_codepoint_grapheme_bounds_abort.mojo (9 cases)

Removes both files from _LIT_TESTS in BUILD.bazel so they run as standard mojo_test targets.

Part of #6920


NathanSWard:
@copilot resolve the merge conflicts in this pull request

PR #7026, opened 2026-08-30
[stdlib] Fix `String.find` phantom match past 2**simd_width bytes

### LINKED ISSUE #7024: [BUG] `String.find` returns a phantom match for single-byte needles past 64 KiB
### Bug description

# `String.find` returns a phantom match for single-byte needles past 64 KiB

## Summary

`_memchr_impl` truncates the match index to the width of the SIMD comparison
mask. On a target where `simd_width_of[DType.bool]() == 16` the mask is a
`UInt16`, so any match at or beyond byte 65536 is reported at `index % 65536`
— an offset where the needle is not present.

This is silent data corruption, not a crash: `find` returns a plausible
in-range index and callers have no way to notice. `String.split` is built on
`find`, so it inherits the bug and quietly emits extra pieces, breaking any
line-oriented parsing of text over 64 KiB.

- **Severity:** wrong results, no diagnostic.
- **Blast radius:** `StringSpan.find`, `String.find`, and everything layered on
  them — `split`, `splitlines`, `replace`, `count`, `__contains__`.
- **Trigger:** a **single-byte** needle and a haystack longer than 65536 bytes.
  Multi-byte needles take `_memmem_impl`, which does the same arithmetic
  correctly and is unaffected.

## Prior art

Searched `modular/modular` issues before filing — this does not appear to be
reported:

- `memchr`, `pack_bits`, `_memchr_impl` — only optimization proposals
  (#5997, #6044, #6107) touching the same neighbourhood, no correctness reports.
- `split truncat`, `memchr 64KB` — no results.
- #1355 "`String.find()` returns wrong value when starting index is non-zero"
  is a different bug (relative vs absolute index) and was fixed in 2023.

#5912 "Use pre-scanned capacity in `_split()` for explicit separators" touches
`_split`, but proposes a capacity optimization and does not mention wrong
results.

## Environment

| | |
|---|---|
| Mojo | `1.0.0 (ed45d567)` — also present in `1.1.0.dev2026082905` sources |
| Platform | macOS 15 (Darwin 25.6.0), Apple silicon, arm64 |
| `simd_width_of[DType.bool]()` | 16 |

Any target with a bool SIMD width of 16 or 32 is affected (the cap is
`2**width`). A 64-lane target would truncate at 2^64 and so is effectively
safe.

## Reproduction

(check "steps to reproduce" section)

## Cause

(check "steps to reproduce" section)

## Fix

Do the addition in `Int`. Only the lane offset needs to come out of the mask.

```diff
--- a/mojo/stdlib/std/collections/string/string_span.mojo
+++ b/mojo/stdlib/std/collections/string/string_span.mojo
@@ -2632,9 +2632,7 @@ def _memchr_impl[
         var mask = pack_bits(bool_mask)
         if mask:
-            return haystack.unsafe_offset(
-                Int(type_of(mask)(i) + count_trailing_zeros(mask))
-            )
+            return haystack.unsafe_offset(i + Int(count_trailing_zeros(mask)))
```

This is exactly what the neighbouring `_memmem_impl` already does at line 2723,
which is why multi-byte needles are unaffected:

```mojo
            var offset = i + Int(count_trailing_zeros(mask))
```

The narrowing cast buys nothing — `count_trailing_zeros(mask)` is at most
`bool_mask_width`, and `i` is already an `Int`.

### Verification

Both versions transcribed into user code and run side by side on this target:

| needle at | current | with the fix |
|---:|---:|---:|
| 1,000 | 1,000 | 1,000 |
| 65,535 | 65,535 | 65,535 |
| 65,536 | **0** | 65,536 |
| 70,000 | **4,464** | 70,000 |
| 131,072 | **0** | 131,072 |
| 2,000,000 | **33,920** | 2,000,000 |

## Also worth fixing

`mojo/stdlib/benchmarks/utils/bench_memmem.mojo:175` carries the same pattern:

```mojo
var offset = Int(type_of(mask)(i) + count_trailing_zeros(mask))
```

It is a benchmark rather than shipped code, but if it is ever a reference for
the real implementation it will propagate the bug back.

## Why the existing tests miss it

`mojo/stdlib/test/collections/string/test_mem_functions.mojo` does cover the
vectorized path, but only just:

```mojo
def test_memchr_vectorized() raises:
    # Long string to exercise the SIMD path in _memchr_impl.
    var s = String("a") * 200 + "z"
```

201 bytes enters the SIMD loop but never reaches `i == 65536`, so the
truncation cannot show. The gap is the *magnitude* of the index, not the code
path.

## Suggested regression test

```mojo
def test_memchr_past_the_mask_width() raises:
    # A single-byte needle beyond 2**simd_width_of[DType.bool]() bytes must
    # still report its true index; it used to come back as index % 65536.
    for at in materialize[[65535, 65536, 70000, 200000]]():
        var buf = List[UInt8](length=at + 40, fill=0x61)  # "a"
        buf[at] = 0x0A  # "\n"
        var s = String(StringSlice(unsafe_from_utf8=buf))
        assert_equal(s.find("\n"), at)
        assert_equal(len(s.split("\n")), 2)
```

### Steps to reproduce

```mojo
def main() raises:
    for at in materialize[[1000, 65535, 65536, 70000, 2000000]]():
        var buf = List[UInt8](length=at + 40, fill=0x61)  # all "a"
        buf[at] = 0x0A                                    # one newline
        var s = String(StringSlice(unsafe_from_utf8=buf))
        print("needle at", at, "-> find returned", s.find("\n", 0))
```

```
needle at 1000    -> find returned 1000       ok
needle at 65535   -> find returned 65535      ok
needle at 65536   -> find returned 0          WRONG
needle at 70000   -> find returned 4464       WRONG   (70000  - 65536)
needle at 2000000 -> find returned 33920      WRONG   (2000000 - 30*65536)
```

Note `find` reports index `0` for a needle at 65536 — position 0 holds `'a'`.

### How it surfaced: `split`

```mojo
var body = List[UInt8](length=70000, fill=0x61)
var s = String(StringSlice(unsafe_from_utf8=body))
for i in range(4):
    s += "\nshort" + String(i)

# 5 pieces expected, the first 70000 bytes long
# actual: 6 pieces, longest 65535
```

The byte total is preserved, so nothing looks lost — the long piece is simply
cut into 65535-byte chunks, as though extra separators were present. Reading a
text file of records, every column after the cut shifts by one row.

There is a threshold that makes this confusing to hit: whether the vectorized
loop reaches the needle at all depends on
`vectorized_end = align_down(length, 16)`. A needle in the final unaligned
16 bytes is found by the scalar tail loop and comes back correct. So appending
a few characters to a file can flip the result between right and wrong, which
reads like nondeterminism.

## Cause

`mojo/stdlib/std/collections/string/string_span.mojo`, `_memchr_impl`, line 2635:

```mojo
    for i in range(0, vectorized_end, bool_mask_width):
        var bool_mask = haystack.unsafe_load[width=bool_mask_width](i).eq(
            first_needle
        )
        var mask = pack_bits(bool_mask)
        if mask:
            return haystack.unsafe_offset(
                Int(type_of(mask)(i) + count_trailing_zeros(mask))
            )
```

`pack_bits` on a 16-lane bool vector returns a `UInt16`. `type_of(mask)(i)`
casts the *byte offset* `i` — which grows to the length of the haystack — into
that 16-bit type before the addition, so the sum wraps mod 65536. The `Int(...)`
on the outside then widens a value that has already lost its high bits.

### System information

pixi info
System
------------
       Pixi version: 0.77.1
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=26.6.2=0
                   : __archspec=1=m4
          Cache dir: /Users/alexbrou/Library/Caches/rattler/cache
       Auth storage: /Users/alexbrou/.rattler/credentials.json

Global
------------
            Bin dir: /Users/alexbrou/.pixi/bin
    Environment dir: /Users/alexbrou/.pixi/envs
       Manifest dir: /Users/alexbrou/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: mojo-bitcoin
            Version: 0.1.0
      Manifest file: /Users/alexbrou/Code/mojo/mojo-bitcoin/pixi.toml
       Last updated: 29-08-2026 22:37:22

Environments
------------
        Environment: default
           Features: default
           Channels: https://conda.modular.com/max, conda-forge
   Dependency count: 2
       Dependencies: mojo, max
   Target platforms: linux-aarch64, osx-arm64, linux-64
  Resolved platform: osx-arm64 (archspec=m1, macos=13.0, __unix)
   Minimum platform: osx-arm64 (__osx >=11.0, __unix)
    Prefix location: /Users/alexbrou/Code/mojo/mojo-bitcoin/.pixi/envs/default
              Tasks: test, vanity, bench, deps

pixi list max
Installed for: osx-arm64
Name      Version  Build             Size  Kind   Source
max       26.5.0   3.14release  15.61 MiB  conda  https://conda.modular.com/max
max-core  26.5.0   release      63.63 MiB  conda  https://conda.modular.com/max

this was tested with 1.0.0 (ed45d567) and also with 1.1.0.dev2026082905 (nightly) and the problem exists in both

chiliec:
## What

Closes #7024.

`_memchr_impl` (the SIMD path behind `StringSpan.find` / `String.find`, and
everything layered on it — `split`, `splitlines`, `replace`, `count`,
`__contains__`) computed the vectorized match offset as:

```mojo
Int(type_of(mask)(i) + count_trailing_zeros(mask))
```

`pack_bits` on a `bool_mask_width`-lane vector returns a `bool_mask_width`-bit
unsigned integer. `type_of(mask)(i)` casts the *byte offset* `i` — which grows
to the length of the haystack — into that narrow type **before** the addition,
so the sum wraps `mod 2**bool_mask_width`. On a target where
`simd_width_of[DType.bool]() == 16` any single-byte match at or beyond byte
65536 is reported at `index % 65536`; where it is `32`, the wrap is at 2**32.
The result is silent wrong data (a plausible in-range index), not a crash.

## Fix

Do the addition in `Int`; only the lane offset needs to come out of the mask:

```mojo
i + Int(count_trailing_zeros(mask))
```

This is exactly what the neighbouring `_memmem_impl` already does
(`string_span.mojo`, the `var offset = i + Int(count_trailing_zeros(mask))`
line), which is why multi-byte needles were unaffected. The identical pattern
in `stdlib/benchmarks/utils/bench_memmem.mojo` (flagged in the issue) is fixed
too so it can't propagate the bug back if used as a reference.

## Tests

Adds `test_memchr_past_the_mask_width` to
`stdlib/test/collections/string/test_mem_functions.mojo`, asserting `_memchr`
returns the true index for a needle at/after the mask boundary (65535, 65536,
70000, 200000). The existing `test_memchr_vectorized` only reached index 201,
so the truncation could never show — the gap was the *magnitude* of the index,
not the code path.

## Validation (exact commands + real results)

I verified on a Linux x86_64 host where `simd_width_of[DType.bool]() == 32`, so
the real wrap point is 2**32 (4 GiB). Two constraints on this box shaped how I
proved it, stated plainly:

1. A 4 GiB `List` + the `String` copy exceeds the machine's RAM, so an
   end-to-end `find` at the true boundary can't be run here (it OOM-kills). The
   maintainer-suggested `find`/`split` test at indices ≤200000 therefore passes
   both before and after the fix **on this width-32 target** — it only goes red
   on a 16-lane-bool target (e.g. arm64), which CI covers. It is included as the
   protective regression guard the issue asked for.

2. To get a genuine RED→GREEN **here**, I transcribed the two offset
   expressions and drove them with the *real* stdlib `pack_bits` and
   `count_trailing_zeros` at this host's actual `simd_width_of[DType.bool]()`,
   reconstructing the exact `mask` the loop produces (lane 0 set) at a byte
   offset just past `2**bool_mask_width`:

   ```
   simd_width_of[DType.bool]() = 32
   mask bit width = 32
   i (true offset) = 4294967424
   buggy offset (Int(type_of(mask)(i) + ctz)) = 128        <- WRONG (wrapped)
   fixed offset (i + Int(ctz))                = 4294967424  <- correct
   RESULT: RED->GREEN confirmed
   ```

   The buggy expression wraps `4294967424 → 128`; the fixed expression returns
   the true offset. This exercises the precise arithmetic changed by the diff.

I could **not** run the full `./bazelw test //mojo/stdlib/test/...` suite on
this host (it needs bazel and a matching nightly `mojo`; the locally installed
compiler is release `1.0.0`, which can't build the nightly stdlib sources). The
change is a one-token narrowing removal that makes `_memchr_impl` match the
already-correct sibling `_memmem_impl`, plus a benchmark line and an additive
test — happy to adjust if CI surfaces anything.

First-time contributor to this area; grateful for review.


Mojo/stdlib/test/collections/string/test_mem_functions.mojo
```diff
@@ -76,6 +76,23 @@ def test_memchr_vectorized() raises:
...
 
+def test_memchr_past_the_mask_width() raises:
+    # A single-byte needle beyond 2**simd_width_of[DType.bool]() bytes must
+    # still report its true index. The vectorized offset used to be computed as
+    # `Int(type_of(mask)(i) + count_trailing_zeros(mask))`, which narrowed the
+    # byte offset `i` to the width of the pack_bits mask before the addition, so
+    # any match at or beyond 2**mask_width came back as `index % 2**mask_width`
+    # (e.g. index 65536 reported as 0 on a 16-lane-bool target). The fix mirrors
+    # `_memmem_impl`: keep `i` an `Int` and only widen the lane offset.
+    for at in materialize[[65535, 65536, 70000, 200000]]():
```
barcharcraz, line 87: do we really need this materialize?

Mojo/stdlib/test/collections/string/test_mem_functions.mojo
```diff
@@ -76,6 +76,23 @@ def test_memchr_vectorized() raises:
...
 
 
+def test_memchr_past_the_mask_width() raises:
+    # A single-byte needle beyond 2**simd_width_of[DType.bool]() bytes must
+    # still report its true index. The vectorized offset used to be computed as
+    # `Int(type_of(mask)(i) + count_trailing_zeros(mask))`, which narrowed the
+    # byte offset `i` to the width of the pack_bits mask before the addition, so
+    # any match at or beyond 2**mask_width came back as `index % 2**mask_width`
+    # (e.g. index 65536 reported as 0 on a 16-lane-bool target). The fix mirrors
+    # `_memmem_impl`: keep `i` an `Int` and only widen the lane offset.
```
barcharcraz, line 86:
```suggestion
    # Regression test for GH-7024: 
    # String.find returns a phantom match for single-byte needles past 64 KiB
```

barcharcraz (approved):
Some minor feedback.

Thank you for the PR,

barcharcraz (approved):

PR #7050, opened 2026-08-31
[stdlib][mojo-lang][proposal] Add paramfor_lower_bound stub to preallocate vector during unrolling

martinvuyk:
## Type of change

- [x] Performance improvement (includes benchmark results below)

## Motivation

The idea is to add a hint to preallocate the compiler vector accumulator for copmtime for loop unrolling. It should be a change that doesn't meaningfully degrade short loops, but has an impact for larger ones.

## What changed

- Created a new paramfor_lower_bound stub that is required
  - It could be made optional as well if it is deemed best.
- Had to change `SIMD.cast()` to not use `Tuple.__contains__` since it would make the paramfor stub recursive when used in combination with the range iterators
- A limitation that this approach introduces is that no iterator in the stdlib will be able to have a `__len__` function that directly or indirectly uses any `comptime for` and be used at compile time as a `comptime for` iterator.

## Other notes

I didn't know what the best way to benchmark the effect of this would be, not sure how to best benchmark compile-times per se, and not sure what loop size to benchmark. Without knowing the actual impact, I don't know if this is really worth the churn.

I relied heavily on an LLM for the compiler parts of this, but went back and forth several times until the code looked as similar to other code in the same modules and ensured the change was a minimal as possible. I have little experience with C++ so I may have missed some details.

Assisted-by: AI


dgurchenkov:
@martinvuyk we are not ready yet to accept feature PRs to the compiler. Yet at some moment this would change.

To prepare to that, do you mind update the PR body to explain the problem that is being addressed (is this slow compile time, runtime perf, usability etc) and add reasonable examples explaining the solution? From a quick glance I'm not quite sure what is the goal, so if you add details it'd be great! And we can discuss the implementation after that.

PR #7064, opened 2026-09-02
[stdlib] Guard _DLHandle.close against a null handle

### LINKED ISSUE #7063: [stdlib] _DLHandle.close calls dlclose on a possibly-null handle
In `mojo/stdlib/std/ffi/__init__.mojo` `_DLHandle.close` calls `dlclose` unconditionally:

```mojo
def close(mut self):
    _ = dlclose(self.handle)
    self.handle = {}
```

`self.handle` is an `OptionalPointer` and can be null. POSIX leaves `dlclose` on a handle not returned by `dlopen` undefined. The return value is discarded, so any error is silent.

e.g. `_find_dylib` returns this null pointer

```mojo
# _find_dylib, abort_on_failure=False
except e:
    return OwnedDLHandle(unsafe_uninitialized=True)   # _DLHandle({}) — null
```

modular/modular#7012 adds a missing destructor call to `_deinit_wrapper`, which makes every dylib-backed global run `__deinit__` at process exit — including the ten above with a null handle.

## Proposed fix

Guard at the dereference site `_DLHandle.close()`

```mojo
def close(mut self):
    if self.handle:
        _ = dlclose(self.handle)
    self.handle = {}
```

chiliec:
## Linked issue

Fixes #7063

## Type of change

## Motivation

`_DLHandle.close` calls `dlclose(self.handle)` unconditionally, but `self.handle`
is an `OptionalPointer` that can legitimately be null. `_find_dylib` (with
`abort_on_failure=False`) returns an `OwnedDLHandle(unsafe_uninitialized=True)`
whose handle is null when a library can't be found. When such a handle is
destroyed, `OwnedDLHandle.__deinit__` calls `close()`, which reaches
`dlclose(NULL)`.

POSIX leaves `dlclose` on a handle not returned by `dlopen` undefined, and on
glibc `dlclose(NULL)` segfaults, so destroying a null-handle library crashes the
process. The `dlclose` return value is also discarded, so any error is silent.

## What changed

Guard the `dlclose` call on a non-null handle in `_DLHandle.close`. The
`self.handle = {}` reset is unchanged, so a closed handle still reads as null.

Added a regression test in `stdlib/test/sys/test_dlhandle.mojo` that constructs
an uninitialized (null) `OwnedDLHandle` and lets it be destroyed.

## Testing

Verified locally with a nightly Mojo toolchain (`mojo 1.1.0.dev2026090105`) by
packaging the edited `std` and running the test file against it:

```
mojo package stdlib/std -o /tmp/std.mojoc
MODULAR_MOJO_MAX_IMPORT_PATH=/tmp mojo run stdlib/test/sys/test_dlhandle.mojo
```

- Fixed: `23 tests run: 23 passed, 0 failed` (including the new
  `test_owned_dlhandle_destroy_null_handle`).
- RED→GREEN: reverting the guard (back to the unconditional `dlclose`) and
  re-running crashes with `execution crashed` inside `dlclose`; restoring the
  guard passes. A standalone C reproducer (`dlclose(NULL)`) segfaults on this
  host's glibc, confirming the underlying behavior.

`mojo format` reports both changed files unchanged.

I could not run the full Bazel test suite locally; validated the affected file
as described above. Happy to adjust.

Assisted-by: Claude (Anthropic)


mojo/stdlib/test/sys/test_dlhandle.mojo
```diff
@@ -358,5 +358,11 @@ def test_owned_dlhandle_automatic_cleanup() raises:
     create_and_destroy_handle()
 
 
+def test_owned_dlhandle_destroy_null_handle() raises:
+    var lib = OwnedDLHandle(unsafe_uninitialized=True)
+    assert_true(not lib.__bool__(), "uninitialized handle should be null")
```
JoeLoser:
```suggestion
    assert_true(not lib, "uninitialized handle should be null")
```

There's no need to explicitly call `__bool__()` here. We could make this more explicit, e.g. `assert_false(lib)` too.

JoeLoser (approved):
Thanks for working on this! One small test cleanup and then I'm happy to sync this in and land it.

alisterburt (commented):
@chiliec you also have some merge conflicts that will need resolving before merge

Mojo/stdlib/std/ffi/__init__.mojo
```diff
@@ -679,7 +679,11 @@ struct _DLHandle(Boolable, ImplicitlyCopyable, RegisterPassable):
             using `OwnedDLHandle` which automatically manages the library
             lifetime.
         """
-        _ = dlclose(self.handle)
+        # `dlclose` on a handle not returned by `dlopen` is undefined; on glibc
+        # a null handle segfaults, and a null handle is reachable (e.g.
+        # `_find_dylib` hands back an `unsafe_uninitialized` handle on failure).
```
alisterburt:
```suggestion
```

Mojo/docs/site/nightly-changelog.md
```diff
@@ -497,6 +502,13 @@ This release completes the removal of APIs deprecated during the v1.0 cycle.
 
 ## Fixed
 
+- Destroying an [`OwnedDLHandle`](/docs/std/ffi/OwnedDLHandle/) that holds a
+  null handle no longer crashes the process. `close()` called `dlclose()`
+  unconditionally, and `dlclose(NULL)` segfaults on glibc. A null handle is
+  reachable through the normal API: an internal library lookup that cannot find
+  its library hands back an uninitialized handle, so the crash landed at the
+  end of the handle's scope rather than at the failed load.
```
alisterburt:
```suggestion
- Destroying an [`OwnedDLHandle`](/docs/std/ffi/OwnedDLHandle/) that holds a
  null handle no longer crashes the process.
```

Mojo/docs/site/nightly-changelog.md
```diff
@@ -136,6 +136,11 @@ This version is still a work in progress.
 
 ## Library changes
 
+- [`OwnedDLHandle`](/docs/std/ffi/OwnedDLHandle/) now conforms to `Boolable`.
+  It already had a `__bool__()` reporting whether the handle is valid, but the
+  missing conformance kept it out of anything generic over `Boolable`,
+  including `assert_false()`.
```
alisterburt:
```suggestion
- [`OwnedDLHandle`](/docs/std/ffi/OwnedDLHandle/) now conforms to `Boolable`.
```

alisterburt (approved):

PR #7065, opened 2026-09-02
[docs] Add a parallelism page for MAX inference

### LINKED ISSUE #6926: [Docs] Add a dedicated parallelism page for MAX
### Where is the problem?

https://max.modular.com/cli/serve/

### What can we do better?

MAX supports tensor, data, and expert parallelism for inference, but there is no dedicated page in the docs that explains any of them. The material that does exist lives almost in release notes and CLI reference doc.

This proposes a new page at `docs/max/serve/parallelism.mdx` under the Deployment chapter (similar to the vLLM doc here: https://docs.vllm.ai/en/stable/serving/parallelism_scaling/).

I am happy to draft a doc PR for this if the proposal makes sense.

## What we have today

Parallelism is currently scattered around the doc. Some examples:

| Location | What it covers |
|---|---|
| `max/python/docs/cli/serve.rst:38-57` | "Running on multiple GPUs" — `--devices` syntax, `gpu:all`, `CUDA_VISIBLE_DEVICES` warning. ~20 lines, and it is mainly about device selection. |
| `docs/max/container.mdx:205-209` | `--devices` for multi-GPU container deployment |
| `docs/max/models.mdx` | Generated support table with a Multi-GPU Yes/No column |
| `docs/max/environment-variables.mdx:97` | `MAX_SERVE_EPLB_PROFILE` (expert-parallel load balancing) |
| `docs/max/packages.mdx:441` | Points at the models table to check multi-GPU support |

In https://max.modular.com/develop/, the doc puts Parallelism as one of the key features there:

> Data parallelism, tensor parallelism, weight sharding, collective communication, and device management.

It promises the topic, but links to nothing.

## Why we need this page

I don't expect this page to explain the concepts in detail (we can link to the [LLM Inference Handbook](https://handbook.modular.com/inference-optimization/data-tensor-pipeline-expert-hybrid-parallelism/)), but we need a page where people can go if they have questions about parallelism in MAX.

Some reasons:

### 1. Tensor parallelism has no flag, and the rule is undocumented

Users arriving from vLLM or SGLang search for `--tensor-parallel-size` or similar flags do not find it, and have no page telling them that `--devices gpu:0,1,2,3` is the TP knob.

### 2. The parallelism flags are undiscoverable as a group

One has to go to the CLI doc to understand the usage. It would be helpful to users if there were a dedicated page explicitly shown in the left navigation bar, with some exlanations and code examples for different use cases.

| Mode | CLI flag | Documented |
|---|---|---|
| Tensor | none, implicit from `--devices` | nowhere |
| Data | `--data-parallel-degree` | https://max.modular.com/cli/serve/ |
| Expert | `--ep-size`, `--ep-use-allreduce` | https://max.modular.com/cli/serve/ |
| Devices | `--devices` | https://max.modular.com/cli/serve/ |

### 3. Data parallelism silently redefines `--max-batch-size`

Per https://max.modular.com/releases/v26.1/#26-1-max-serve, enabling DP changed what `--max-batch-size` means per replica. This is an important change and impacts how users understand batches in MAX. A user who sets both flags gets different behavior than they expect, and the only explanation is in a release note for a version they may have skipped. You can't expect users to go to the release notes to find the answer.

## Proposed page

**Location:** `docs/max/serve/parallelism.mdx`

Same genre as `prefix-caching.mdx` and `speculative-decoding.mdx`, a serving feature you turn on and tune. It sits naturally in the existing serve sidebar.

Outline:

- Intro. 
   - Basic explanations about parallelism and why you need it.
- Supported strategy (using a table here)
  - Use a table here to show supported parallelism strategy, namely DP, EP, TP, their use case, with their flags. Link concept details to https://handbook.modular.com/inference-optimization/data-tensor-pipeline-expert-hybrid-parallelism/
  - BTW, pipeline parallelism seems not supported in MAX. Feel free to correct me if I am wrong, but if it does not support it, the doc should mention it.
- Running on a single node
  - Show CLI commands here,  DP / TP / EP, hybrid configs
- Running on multiple nodes
  - I am not sure if MAX supports this as I did not see a flag like  vllm’s `--nnodes` and `--node-rank`. Skip this section if this feature is not supported
  - Show CLI commands here if MAX supports this
- Interaction with batch size
  - the v26.1 DP semantics, promoted out of the release note.
- Tuning parallelism
  - Talk about other useful flags
- Next steps

### Anything else?

_No response_

Sherlock113:
## Linked issue

Fixes https://github.com/modular/modular/issues/6926

## Type of change

- [x] Documentation update

## Motivation

MAX supports tensor, data, and expert parallelism for inference, but no page in the docs explained any of them. The material that existed was spread across the `max serve` CLI reference and release notes. Learn more in https://github.com/modular/modular/issues/6926

## What changed

- Added a dedicated parallelism page at `docs/max/serve/parallelism.mdx`, under the Deployment chapter in the sidebar. It covers single node only, per reviewer guidance. There's no multi-node section; instead a short call-out points multi-node workloads at Modular Cloud.
- Fixed a typo on the layouts page

## Testing

Didn't test the commands on GPUs, but every command (except the TP+DP example) on the page was checked against architecture code and the recipe configs in `max/python/max/pipelines/architectures/*/recipes/`:

- TP+EP (`--data-parallel-degree 1 --ep-size 8`) matches `deepseekV3/recipes/nvfp4_tpep_8x_b200.yaml`.
- DP+EP (`--data-parallel-degree 8 --ep-size 8`) matches `deepseekV3/recipes/nvfp4_8x_b200.yaml`.
- The DP example matches the `--data-parallel-degree 2 --devices gpu:0,1` example from the 25.6 release notes.
- The `--ep-size` constraint (1 or the total GPU count on a single node) is enforced in `deepseekV3/model.py:124-129` and `qwen3/model.py:117`.
- The DP+EP description (dense components and shared experts replicated, routed experts divided) follows `deepseekV3/model.py:91-93`, `nn/moe/moe.py:359` (routed experts divided by `ep_size`), and `nn/moe/moe.py:387-397` (shared experts built with no `ep_size` division).

One question about TP+DP: is this actually supported? The TP+DP tab uses `--data-parallel-degree 2` on eight GPUs, expecting TP=4 per replica, but I did not find any recipe using a partial degree.

BTW, I am not able to preview the doc page locally, so please help build it to see if the build succeeds. Thanks. 

## Checklist

      approval
- [ ] PR is small and focused — I've split larger changes into a sequence of

### DIFF
--- a/docs/max/serve/parallelism.mdx
+++ b/docs/max/serve/parallelism.mdx
@@ -106,7 +106,8 @@ combinations depend on support for the selected model.
 
 Because the tensor parallelism degree is derived rather than set directly, you
 choose a hybrid configuration by combining a device count, a data parallelism
-degree, and an expert parallelism size. On eight GPUs:
+degree, and an expert parallelism size. For example, if you have
+8 GPUs, you have the following hybrid options:
 
 | `--data-parallel-degree` | `--ep-size` | Resulting configuration                                                                                          |
 |--------------------------|-------------|------------------------------------------------------------------------------------------------------------------|
--- a/docs/max/serve/parallelism.mdx
+++ b/docs/max/serve/parallelism.mdx
@@ -1,0 +1,208 @@
+---
+title: Parallelism
+description: Distribute a model across multiple GPUs with tensor, data, or expert parallelism
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Parallelism distributes model computation or incoming requests across multiple
+GPUs. You might consider parallelism when a model doesn't fit comfortably on a
+single GPU or when distributing work could help meet your throughput or
+concurrency goals.
+
+MAX supports tensor parallelism, data parallelism, and expert parallelism on a
+single node. For certain models, you can combine strategies to match the model
+architecture and available hardware.
+
+## Supported strategies
+
+The following table summarizes how each strategy works, when to use it, and the
+relevant MAX CLI options.
+
+| Strategy                    | How it works                                                              | When to use it                                                              | CLI options                          |
+|-----------------------------|---------------------------------------------------------------------------|-----------------------------------------------------------------------------|--------------------------------------|
+| Tensor parallelism (TP)     | Splits model weights and layer computation across GPUs                    | A model doesn't fit on one GPU                                              | None, TP is implicit in `--devices`  |
+| Data parallelism (DP)       | Runs multiple model replicas and distributes incoming requests among them | A model fits in each replica, and you need higher throughput or concurrency | `--data-parallel-degree`             |
+| Expert parallelism (EP)     | Distributes Mixture of Experts (MoE) expert weights across GPUs           | You need to distribute the expert weights of a supported MoE model          | `--ep-size`                          |
+
+:::note
+
+MAX currently doesn't support pipeline parallelism.
+
+:::
+
+For more conceptual information, see
+[data, tensor, pipeline, expert, and hybrid parallelism](https://handbook.modular.com/inference-optimization/data-tensor-pipeline-expert-hybrid-parallelism/)
+in the LLM Inference Handbook.
+
+## Serve a model on multiple GPUs
+
+The following examples all run on a single node.
+
+### Tensor parallelism
+
+MAX doesn't provide a separate tensor parallelism flag. For a model that
+supports tensor parallelism, use `--devices` to select the GPUs you want to
+shard across. This example serves a model with TP=4:
+
+```bash
+max serve \
+  --model google/gemma-3-27b-it \
+  --devices gpu:0,1,2,3
+```
+
+MAX applies tensor parallelism automatically across the devices you give it. It
+derives the tensor parallelism degree from the device count and the data
+parallelism degree:
+
+```text
+tensor parallelism degree = number of devices / data parallelism degree
+```
+
+The device count must be divisible by the data parallelism degree. With the
+default `--data-parallel-degree 1`, serving this model on four GPUs gives you
+TP=4.
+
+### Data parallelism
+
+Set `--data-parallel-degree` to the number of replicas you want. This example
+serves two independent replicas, one per GPU:
+
+```bash
+max serve \
+  --model meta-llama/Llama-3.1-8B-Instruct \
+  --devices gpu:0,1 \
+  --data-parallel-degree 2
+```
+
+When data parallelism is active, `--max-batch-size` sets the maximum batch size
+**per replica**, not across the whole server. For example,
+`--data-parallel-degree 8` with `--max-batch-size 32` allows up to 32 requests
+per replica and 256 requests across the server.
+
+This changed in [MAX 26.1](/releases/v26.1/#26-1-max-serve). Before that
+release, the same flags meant an aggregate maximum of 32 requests, with each
+replica capped at 4. When carrying a configuration forward from an earlier
+version, divide the previous `--max-batch-size` by the data parallelism degree
+to keep the same aggregate maximum.
+
+### Expert parallelism
+
+Expert parallelism applies only to supported MoE models. Set `--ep-size` to the
+total number of GPUs:
+
+```bash
+max serve \
+  --model nvidia/DeepSeek-V3.1-NVFP4 \
+  --devices gpu:0,1,2,3,4,5,6,7 \
+  --ep-size 8
+```
+
+For a single-node deployment, `--ep-size` must be either `1` (EP disabled, the
+default) or the total GPU count. Values in between are not supported.
+
+### Hybrid parallelism
+
+Some models and inference workloads benefit from combining parallelism
+strategies. This approach can distribute model computation, requests, and MoE
+expert weights in different ways across the same set of GPUs. The available
+combinations depend on support for the selected model.
+
+Because the tensor parallelism degree is derived rather than set directly, you
+choose a hybrid configuration by combining a device count, a data parallelism
+degree, and an expert parallelism size. On eight GPUs:
+
+| `--data-parallel-degree` | `--ep-size` | Resulting configuration                                                                                          |
+|--------------------------|-------------|------------------------------------------------------------------------------------------------------------------|
+| `1` (default)            | unset       | TP=8, DP=1: one replica, weights sharded across all GPUs                                                         |
+| `2`                      | unset       | TP=4, DP=2: two replicas, each sharded across four GPUs                                                          |
+| `1`                      | `8`         | TP=8, EP=8: attention layers tensor-sharded across all eight GPUs; MoE experts distributed across all eight GPUs |
+| `8`                      | `8`         | DP=8, EP=8: each GPU runs one attention replica; requests and MoE experts distributed across all eight GPUs      |
+
+<Tabs>
+  <TabItem value="tp-dp" label="TP+DP">
+
+Set `--data-parallel-degree` to the replica count. This example serves eight
+GPUs as two data-parallel replicas of a TP=4 model:
+
+```bash
+max serve \
+  --model google/gemma-4-31B-it \
+  --devices gpu:0,1,2,3,4,5,6,7 \
+  --data-parallel-degree 2
+```
+
+  </TabItem>
+  <TabItem value="tp-ep" label="TP+EP">
+
+Keep `--data-parallel-degree` at the default of `1` and set `--ep-size` to the
+device count:
+
+```bash
+max serve \
+  --model nvidia/DeepSeek-V3.1-NVFP4 \
+  --devices gpu:0,1,2,3,4,5,6,7 \
+  --data-parallel-degree 1 \
+  --ep-size 8
+```
+
+  </TabItem>
+  <TabItem value="dp-ep" label="DP+EP">
+
+To give each GPU a separate batch shard for attention, set both
+`--data-parallel-degree` and `--ep-size` to the device count. Dense components
+such as attention and MLP layers, along with shared experts, are replicated on
+each GPU, while routed experts are divided across all eight GPUs:
+
+```bash
+max serve \
+  --model nvidia/DeepSeek-V3.1-NVFP4 \
+  --devices gpu:0,1,2,3,4,5,6,7 \
+  --data-parallel-degree 8 \
+  --ep-size 8
+```
+
+  </TabItem>
+</Tabs>
+
+Not every model supports parallelism, and support for the hybrid combinations
+above is added per architecture. To check whether a specific model supports
+multi-GPU execution, see the **Multi-GPU** column in [MAX models](/models).
+
+:::note
+
+The examples on this page cover a single node. For multi-node and large-scale
+workloads, use [Modular Cloud](https://docs.modular.com/), which manages
+distributed deployment for you.
+
+:::
+
+## Tune parallelism
+
+After choosing a strategy, use the following flags to tune the behavior. The
+**Config** column shows the config class each flag maps to when you configure a
+pipeline programmatically.
+
+| Flag                                    | Description                                                                                                                                                                                                          | Config                                                                               |
+|-----------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------|
+| `--ep-use-allreduce`                    | Uses allreduce for cross-device communication in expert parallelism. Performance depends on the model and interconnect, so benchmark your deployment first.                                                          | [`PipelineRuntimeConfig`](/api/python/generated/max.pipelines.PipelineRuntimeConfig) |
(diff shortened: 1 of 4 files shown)


docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+architecture and available hardware.
+
+## Supported strategies
+
+The following table summarizes how each strategy works, when to use it, and the
+relevant MAX CLI options.
+
+| Strategy                    | How it works                                                              | When to use it                                                              | CLI options                          |
+|-----------------------------|---------------------------------------------------------------------------|-----------------------------------------------------------------------------|--------------------------------------|
+| Tensor parallelism (TP)     | Splits model weights and layer computation across GPUs                    | A model doesn't fit on one GPU                                              | None, TP is implicit in `--devices`  |
```
scottamain: The CLI option description here is confusing. It sounds like you’re saying "None, you don’t need a CLI option," but then say it needs the `--devices` option, so that’s contradictory (because the `--devices` flag isn’t required). I think the more clear description might be:


```suggestion
| Tensor parallelism (TP)     | Splits model weights and layer computation across GPUs                    | A model doesn't fit on one GPU                                              | [`--devices`](/cli/serve/#--devices-devices) with multiple GPU IDs  |
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+| Data parallelism (DP)       | Runs multiple model replicas and distributes incoming requests among them | A model fits in each replica, and you need higher throughput or concurrency | `--data-parallel-degree`             |
+| Expert parallelism (EP)     | Distributes Mixture of Experts (MoE) expert weights across GPUs           | You need to distribute the expert weights of a supported MoE model          | `--ep-size`                          |
```
scottamain: It’s not necessary to link every instance of an API or CLI command to the reference, but we should link at least the first mention, and especially in a reference table like this:

```suggestion
| Data parallelism (DP)       | Runs multiple model replicas and distributes incoming requests among them | A model fits in each replica, and you need higher throughput or concurrency | [`--data-parallel-degree`] (/cli/serve/#--data-parallel-degree-data_parallel_degree)            |
| Expert parallelism (EP)     | Distributes Mixture of Experts (MoE) expert weights across GPUs           | You need to distribute the expert weights of a supported MoE model          | [`--ep-size`](/cli/serve/#--ep-size-ep_size)                          |
```

scottamain: Typo? If this is what you intended, then I think you should instead rephrase so it doesn’t read "experts expert." I can imagine how that’s the case, but I don’t think it’s necessary to be so grammatically specific to specify "expert weights form a Mixture of Experts model;" I think just deleting the extra experts is still clear about the meaning.

```suggestion
| Expert parallelism (EP)     | Distributes Mixture of Experts (MoE) weights across GPUs           | You need to distribute the expert weights of a supported MoE model          | `--ep-size`                          |
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+:::
+
+For more conceptual information, see
+[data, tensor, pipeline, expert, and hybrid parallelism](https://handbook.modular.com/inference-optimization/data-tensor-pipeline-expert-hybrid-parallelism/)
+in the LLM Inference Handbook.
+
+## Serve a model on multiple GPUs
+
+The following examples all run on a single node.
```
scottamain: I think this section needs a better segue to establish more context. Maybe something like this:

```suggestion
As summarized in the table above, you can serve a model across multiple GPUs using one or more parallelism strategies. The following sections provide more information about how to parallelize a model with `max serve` on a single node.
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+### Tensor parallelism
+
+MAX doesn't provide a separate tensor parallelism flag. For a model that
```
scottamain: We shouldn’t ever need to say what the user _doesn’t_ do. Focus on what they _do_ do.

```suggestion
For a model that
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+```
+
+The device count must be divisible by the data parallelism degree. With the
+default `--data-parallel-degree 1`, serving this model on four GPUs gives you
+TP=4.
+
+### Data parallelism
+
+Set `--data-parallel-degree` to the number of replicas you want. This example
+serves two independent replicas, one per GPU:
```
scottamain: I think this is getting ahead of the reader by talking about TP + DP before even introducing DP on its own. I think this explanation should move into the next section.

````suggestion
MAX applies tensor parallelism automatically across the devices you give it. 

### Data parallelism

Set `--data-parallel-degree` to the number of replicas you want, combined
with `--devices` to select which GPUs to use.

Of course, the parallelism degree cannot exceed the number of
devices available. If the parallelism degree is less than the number of
devices, MAX will also enable tensor parallelism to a degree
based on this simple formula:

```text
tensor parallelism degree = number of devices / data parallelism degree
```

This example serves two independent replicas, one per GPU:
````

File: docs/max/serve/parallelism.mdx [hunk 6]
```diff
@@ -0,0 +1,208 @@
...
+max serve \
+  --model meta-llama/Llama-3.1-8B-Instruct \
+  --devices gpu:0,1 \
+  --data-parallel-degree 2
+```
+
+When data parallelism is active, `--max-batch-size` sets the maximum batch size
+**per replica**, not across the whole server. For example,
+`--data-parallel-degree 8` with `--max-batch-size 32` allows up to 32 requests
+per replica and 256 requests across the server.
```
scottamain, line 83: Please keep terminology consistent. I think you mean "node" here:

```suggestion
**per replica**, not across the whole node. For example,
`--data-parallel-degree 8` with `--max-batch-size 32` allows up to 32 requests
per replica and 256 requests across the node.
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+This changed in [MAX 26.1](/releases/v26.1/#26-1-max-serve). Before that
+release, the same flags meant an aggregate maximum of 32 requests, with each
+replica capped at 4. When carrying a configuration forward from an earlier
+version, divide the previous `--max-batch-size` by the data parallelism degree
+to keep the same aggregate maximum.
```
scottamain: This is a really old release now and we currently don’t support older releases, so we should remove this paragraph.

```suggestion
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+Some models and inference workloads benefit from combining parallelism
+strategies. This approach can distribute model computation, requests, and MoE
+expert weights in different ways across the same set of GPUs. The available
+combinations depend on support for the selected model.
+
+Because the tensor parallelism degree is derived rather than set directly, you
+choose a hybrid configuration by combining a device count, a data parallelism
+degree, and an expert parallelism size. On eight GPUs:
+
+| `--data-parallel-degree` | `--ep-size` | Resulting configuration                                                                                          |
```
scottamain: The line breaks in here bug me:

```suggestion
| <code style={{whiteSpace:'nowrap'}}>--data-parallel-degree</code> | <code style={{whiteSpace:'nowrap'}}>--ep-size</code> | Resulting configuration                                                                                          |
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+### Hybrid parallelism
+
+Some models and inference workloads benefit from combining parallelism
+strategies. This approach can distribute model computation, requests, and MoE
+expert weights in different ways across the same set of GPUs. The available
+combinations depend on support for the selected model.
+
+Because the tensor parallelism degree is derived rather than set directly, you
+choose a hybrid configuration by combining a device count, a data parallelism
+degree, and an expert parallelism size. On eight GPUs:
```
scottamain: This table into could be more clear:

```suggestion
degree, and an expert parallelism size. For example, if you have
8 GPUs, you have the following hybrid options:
```

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+| `--data-parallel-degree` | `--ep-size` | Resulting configuration                                                                                          |
+|--------------------------|-------------|------------------------------------------------------------------------------------------------------------------|
+| `1` (default)            | unset       | TP=8, DP=1: one replica, weights sharded across all GPUs                                                         |
+| `2`                      | unset       | TP=4, DP=2: two replicas, each sharded across four GPUs                                                          |
+| `1`                      | `8`         | TP=8, EP=8: attention layers tensor-sharded across all eight GPUs; MoE experts distributed across all eight GPUs |
+| `8`                      | `8`         | DP=8, EP=8: each GPU runs one attention replica; requests and MoE experts distributed across all eight GPUs      |
+
+<Tabs>
```
scottamain, line 122: This section needs a sentence to introduce it after the table.

docs/max/serve/parallelism.mdx
```diff
@@ -0,0 +1,208 @@
...
+workloads, use [Modular Cloud](https://docs.modular.com/), which manages
+distributed deployment for you.
+
+:::
+
+## Tune parallelism
+
+After choosing a strategy, use the following flags to tune the behavior. The
+**Config** column shows the config class each flag maps to when you configure a
+pipeline programmatically.
```
scottamain, line 184: I don’t think it’s realistic that external developers are configuring pipelines with the Python API like this, so my instinct is we should delete the Config column. Although this might be accurate, these APIs are not mature, they’re almost certainly going to change soon, and the audience for them is tiny. So I think this is going to confuse most people and become a maintenance problem for this page as the APIs change... But I could be wrong, and @rachfop should weigh in on whether these pipeline configs are important to our users at this point.

scottamain (approved):
This is great, thank you Sherlock! I left several suggestions to improve the clarity in some places, but nothing blocking. Overall it reads well and is super useful.

Let’s wait for @rachfop to check on correctness of the different hybrid parallelism, but this otherwise LGTM.

docs/max/serve/parallelism.mdx
Sherlock113: Sorry, this is just a typo 😂

docs/max/serve/parallelism.mdx
Sherlock113: Noted!

docs/max/serve/parallelism.mdx
Sherlock113: I don't understand why `<code>` should be used here, but I applied your suggestion anyway. I guess the markdown code format is not rendered correctly in table for some reason?

docs/max/serve/parallelism.mdx
Sherlock113: oh, I added it simply because the spec decoding doc has it (https://github.com/modular/modular/blob/main/docs/max/serve/speculative-decoding.mdx#tune-speculative-decoding), so just to be consistent here.

docs/max/serve/parallelism.mdx
Sherlock113: Actually I still prefer `server` here. If you are running two max serve processes on one 8-GPU node, such as `--devices gpu:0,1,2,3` for one, `gpu:4,5,6,7` for the other, you are serving two different models, or two configurations of one model, on a single machine. Each process gets its own `--max-batch-size` and its own `data_parallel_degree`, so each computes its own `total_max_batch` independently.

In this case, "256 requests across the node" is not accurate. The node is carrying 512, or whatever the two processes sum to. "Across the server" stays true for each, because each max serve instance is a server.

docs/max/serve/parallelism.mdx
Sherlock113: Added

docs/max/serve/parallelism.mdx
rachfop: partial DP isn't supported. dp must be 1 or the full GPU count. The TP+DP tab should be dropped.

docs/max/serve/parallelism.mdx
scottamain: I see. In that case, you’re using "server" to mean a software service, not a server in the datacenter sense, but that’s not clear. (Although I agree that "server" more commonly/accurately means what you intend here, I think the term is very ambiguous and it’s sometimes used to refer to a larger concept of "datacenter servers" as a hardware concept.) Please rephrase to remove "server" and replace it with "MAX serving instance" or something that emphasizes "MAX" as the the entity.

docs/max/serve/parallelism.mdx
scottamain: @rachfop thank you! But that’s not what this comment is about. Please share your thoughts about the table here.

PR #7102, opened 2026-09-09
[stdlib] Fix `exp` flushing subnormal results to zero and overflowing early

### LINKED ISSUE #7098: [stdlib] Accuracy issues in `math.exp`
`exp` returns `0.0` for every Float64 argument whose true result is subnormal, and `inf` about one unit before the true overflow point:

```
exp(-708.745)    0.0    libm  1.570208859696427e-308
exp(-745.0)      0.0    libm  5e-324
exp(709.4365)    inf    libm  1.2716195926832784e+308
cosh(709.4365)   inf    libm  6.358097963416392e+307
```

`cosh` is `0.5 * exp(|x|) + 0.5 / exp(|x|)`, so it inherits the second one.

parted out from #7068

alisterburt:
## Linked issue

Fixes #7098

## Type of change

- [ ] Documentation update

## Motivation

`exp` returns `0.0` whenever the true result is subnormal, and `inf` just
short of real overflow. Softmax hits the first one routinely: a small
probability becomes exactly zero, then `-inf` through `log`.

## What changed

`_ldexp_impl` scaled by writing `exp + bias` into the exponent field, which
only encodes `[-1022, 1023]`. `exp` reaches `-1023` and `1024`, the reserved
zero and inf patterns. It now scales in two halves so neither factor leaves
the range, which fixes `ldexp` too, and the clamps in `exp` widen to the
real limits.

Float32 had the same bug. Its existing test wanted `inf` from `exp(89)` and
was getting it from the broken scale factor, so it failed once the scaling
was right.

## Checklist

      approval
      [pull request sizes](https://github.com/modular/modular/blob/main/Mojo/docs/contributing/issue-pr-etiquette.md#keep-pull-requests-small))

### DIFF
--- a/Mojo/stdlib/std/math/math.mojo
+++ b/Mojo/stdlib/std/math/math.mojo
@@ -519,14 +519,22 @@ def _ldexp_impl[
         return res
 
     comptime integral_type = FPUtils[dtype].integral_type
-    var m = exp.cast[integral_type]() + SIMD[integral_type, width](
-        FPUtils[dtype].exponent_bias()
-    )
 
-    return x * type_of(x)(
-        from_bits=m
-        << SIMD[integral_type, width](FPUtils[dtype].mantissa_width())
-    )
+    @always_inline
+    def pow2(e: SIMD[dtype, width]) -> SIMD[dtype, width]:
+        var m = e.cast[integral_type]() + SIMD[integral_type, width](
+            FPUtils[dtype].exponent_bias()
+        )
+        return SIMD[dtype, width](
+            from_bits=m
+            << SIMD[integral_type, width](FPUtils[dtype].mantissa_width())
+        )
+
+    # Halving keeps both factors within the representable exponent range, so
+    # exponents that would encode as 0 or as the inf/nan pattern still scale
+    # correctly and underflow stays gradual.
+    var half = floor(exp * 0.5)
+    return x * pow2(half) * pow2(exp - half)
 
 
 @always_inline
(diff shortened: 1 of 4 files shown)


Mojo/stdlib/std/math/math.mojo
```diff
@@ -519,14 +519,22 @@ def _ldexp_impl[
...
+        return SIMD[dtype, width](
+            from_bits=m
+            << SIMD[integral_type, width](FPUtils[dtype].mantissa_width())
+        )
+
+    # Halving keeps both factors within the representable exponent range, so
+    # exponents that would encode as 0 or as the inf/nan pattern still scale
+    # correctly and underflow stays gradual.
+    var half = floor(exp * 0.5)
+    return x * pow2(half) * pow2(exp - half)
```
barcharcraz, line 537: doesn't this reduce precision?

Mojo/stdlib/std/math/math.mojo
```diff
@@ -519,14 +519,22 @@ def _ldexp_impl[
         return res
 
     comptime integral_type = FPUtils[dtype].integral_type
-    var m = exp.cast[integral_type]() + SIMD[integral_type, width](
-        FPUtils[dtype].exponent_bias()
-    )
```
barcharcraz, line 524: I think this actually does handle subnormals correctly. with -1023 you get m=0 and then x*0, which should work fine with subnormals (the zero exponent field indicates a subnormal).

I think the real issue is coming from elsewhere?

Let's fully understand the AI's proposed mechanism before we start blindly implementing things

barcharcraz (changes requested):

Mojo/stdlib/std/math/math.mojo
barcharcraz: oh, wait, you get the right exponent but nuke it because you lack the leading 1

PR #7118, opened 2026-09-11
[stdlib] Add a defaulted Iterator.count() method

### LINKED ISSUE #7116: Add a defaulted `Iterator.count()` method.
Provide an O(N) way to count iterator elements without manually iterating with a counter or converting to a list.<br><br>**NOTE**: There is a bit of bikeshedding allowed for this. Ideally something like `.length()` maybe also be nice to be consistent with our collection APIs, but this **should not** be confused with `__length__` or the `len()` function as these should always operation in `O(1)` time. `Iterator.count()` on the other hand is an `O(N)` operation.

`Iterator` implementations can then override this behavior if they have a "faster" way of computing the count of the `Iterator`. E.g. Span's `Iterator` can can do this in `O(1)` time.

It should look something like:

```
trait Iterator: 
    def count(var self) -> Int:
        var count = 0
        try:
            while True:
                _ = self.__next__()
                if count == Int.MAX:
                    abort("Iterator.count: overflow encountered")
                count += 1
        except StopIteration:
            pass
        return count
```

neuregex:
Fixes #7116

Adds a defaulted `count(var self) -> Int` to the `Iterator` trait, following the sketch in the issue: it consumes the iterator, returns how many elements it yielded, and aborts if the count would exceed `Int.MAX`. It uses the same `Deinitable` assert as `nth` (MOCO-3947). Iterators that know their length in O(1), like Span's, can override it; I left those overrides for a follow-up.

**Tests:** new `test/iter/test_count.mojo` (basic, empty, partially consumed, and a custom override). Also added a changelog entry.

**Checked** on main `da4f8b76` with nightly `1.1.0.dev2026091105` (`mojo precompile` + `mojo run`):

- the new test doesn't compile without the change and passes 4/4 with it
- no new failures in the existing `iter/` and `itertools/` tests
- `mojo format` leaves both files unchanged

### DIFF
--- a/Mojo/stdlib/std/iter/__init__.mojo
+++ b/Mojo/stdlib/std/iter/__init__.mojo
@@ -242,10 +242,8 @@ trait Iterator(Deinitable, Movable):
         Examples:
 
         ```mojo
-        var l = [10, 20, 30]
-        print(iter(l).count())          # 3
-        print(iter(l).count())          # 3: the list itself is not consumed
-        print(range(0, 10, 3).count())  # 4
+        var nums = [10, 20, 30]
+        print(iter(nums).count())  # 3
         ```
         """
         # Same workaround as `nth`: drop it once MOCO-3947 lets us put the
--- a/Mojo/stdlib/std/iter/__init__.mojo
+++ b/Mojo/stdlib/std/iter/__init__.mojo
@@ -56,6 +56,7 @@ from std.builtin.rebind import downcast
 
 
 from std.builtin.variadics import TypeList
+from std.os import abort
 
 
 # ===-----------------------------------------------------------------------===#
@@ -216,6 +217,51 @@ trait Iterator(Deinitable, Movable):
             return self.__next__()
         except StopIteration:
             return None
+
+    def count(var self) -> Int:
+        """Consumes the iterator and returns how many elements it yielded.
+
+        Each element is advanced past and destroyed, so the iterator is
+        exhausted afterwards. Unlike `len()`, which is O(1), this walks the
+        whole iterator. An iterator that can compute its remaining length
+        cheaply can override this method.
+
+        Returns:
+            The number of elements the iterator yielded before it stopped.
+
+        Preconditions:
+            The iterator must yield at most `Int.MAX` elements.
+
+        Constraints:
+            `Self.Element` must conform to `Deinitable` so the elements can
+            be discarded.
+
+        Performance:
+            O(n), where n is the number of remaining elements.
+
+        Examples:
+
+        ```mojo
+        var l = [10, 20, 30]
+        print(iter(l).count())          # 3
+        print(iter(l).count())          # 3: the list itself is not consumed
+        print(range(0, 10, 3).count())  # 4
+        ```
+        """
+        # Same workaround as `nth`: drop it once MOCO-3947 lets us put the
+        # bound in a `where` clause on the method.
+        comptime assert conforms_to(Self.Element, Deinitable)
+
+        var n = 0
+        try:
+            while True:
+                _ = self.__next__()
+                if n == Int.MAX:
+                    abort("Iterator.count: overflow encountered")
+                n += 1
+        except StopIteration:
+            pass
+        return n
 
 
 @always_inline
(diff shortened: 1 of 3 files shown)


Mojo/stdlib/std/iter/__init__.mojo
```diff
@@ -217,6 +218,51 @@ trait Iterator(Deinitable, Movable):
...
+            O(n), where n is the number of remaining elements.
+
+        Examples:
+
+        ```mojo
+        var l = [10, 20, 30]
+        print(iter(l).count())          # 3
+        print(iter(l).count())          # 3: the list itself is not consumed
+        print(range(0, 10, 3).count())  # 4
+        ```
```
NathanSWard, line 247: **suggestion**: This example can just show how `.count()` works - no need to show that the list is not consumed. That's not a specific property of `count()` that just a general `Iterator` property.
Also a nicer variable name than `l` would be preferred.

```suggestion
        ```mojo
        var nums = [10, 20, 30]
        print(iter(sums).count())          # 3
        ```
```

Mojo/stdlib/test/iter/test_count.mojo
```diff
@@ -0,0 +1,79 @@
...
+    var l = [10, 20, 30, 40]
+    assert_equal(iter(l).count(), 4)
+
+
+def test_count_borrowed_does_not_consume_source() raises:
+    var l = [1, 2, 3]
+    assert_equal(iter(l).count(), 3)
+    # Counting a borrowed iterator leaves the list untouched.
+    assert_equal(iter(l).count(), 3)
+    assert_equal(len(l), 3)
```
NathanSWard: **suggestion** - please remove this test as it's not testing anything specific to "count". This is instead testing how iterators work.

Mojo/stdlib/test/iter/test_count.mojo
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_count_owned() raises:
+    var l: List[Int] = [100, 200, 300]
+    assert_equal(iter(l^).count(), 3)
+
+
+def test_count_string_elements() raises:
+    var l = [String("a"), String("b"), String("c")]
+    assert_equal(iter(l).count(), 3)
```
NathanSWard: **suggestion**: Ditto with these two tests too - please also remove them as they are not testing anything specific to "count".

Mojo/stdlib/test/iter/test_count.mojo
```diff
@@ -0,0 +1,79 @@
...
+    var it = iter(l)
+    _ = next(it)
+    _ = next(it)
+    assert_equal(it^.count(), 3)
+
+
+def test_count_range() raises:
+    assert_equal(range(10).count(), 10)
+    assert_equal(range(0, 10, 3).count(), 4)
+    assert_equal(range(5, 5).count(), 0)
```
NathanSWard: **suggestion**: Please also remove this test - this is not testing `count()`'s implementation but rather testing the `range`'s `Iterator` implementation.

NathanSWard (changes requested):
Thanks for implementing this! I left some comments in the tests to clean them up a bit as a handful of tests aren't really testing anything specific to `count()`. We generally try to make sure any newly added unit tests actually binds some behavior of the newly added functionality and not a pre-existing one.

NathanSWard (approved):
thanks!

enimo999-sketch:
My suggestion would be to introduce an IteratorHelper trait ( which inherits from Iterator ) to keep the iterator trait clean. The IteratorHelper trait would not be in the prelude, and thus the freestanding function ( and not method ) count would be preferred. Since mojo is a typed superset of python, mojo needs to think about how to handle iterator helpers, consumers and adapters. In my view, the method should be called __count on the IteratorHelper trait.

EDIT:
I am highly sceptical of this method. Firstly, itertools.count already exists. Second, the 'count' method could easily be achieved using the Python built-in 'sum' function and generator comprehension with the idiom sum(1 for _ in it).
Also, the issue of overflow is not clear. What if you want an Int64 instead of an Int? For example, Rust uses usize. With the sum function, this could easily be achieved using either a [Sum trait](https://doc.rust-lang.org/stable/std/iter/trait.Sum.html#tymethod.sum) or an Add trait.

PR #7126, opened 2026-09-13
[mblack] Fix grammar cache to account for content changes

### LINKED ISSUE #7111: [BUG] `mblack` cache doesn't account for grammar changes
### Actual behavior

running `mojo format` produces error messages when running in an env that specifies a version of Mojo that has outdated grammar rules vs a previously used env.  All invocations for this command share a global grammar cache, which can be problematic in the right workstation environment.

Example:
`error: cannot format emberregex/set_approx.mojo: '_python_symbols' object has no attribute 'old_comptime_assert_stmt'`

### Expected behavior

No mojo commands should be affected by commands run in a different mojo project.

### Steps to reproduce

1. Open a project that uses the latest nightly and run `mojo format`
2. Open another project pinned to Mojo 1.0 and run `mojo format` again, observe errors

### System information

```
      Pixi version: 0.78.0
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=26.6.2=0
                   : __archspec=1=m3
```

mahendrarathore1742:
Fixes #7111

The `mblack` grammar pickle cache was keyed only by the grammar filename and Python version, not by the grammar file's content. Two different Mojo environments with different `Grammar.txt` files produced the same cache
filename (e.g. `Grammar3.12.0.0final.pickle`), so switching environments caused stale grammar errors like:

    error: cannot format emberregex/set_approx.mojo: '_python_symbols' object has
    no attribute 'old_comptime_assert_stmt'

## Root Cause

`_generate_pickle_name()` in `driver.py` builds the cache filename as `Grammar` + Python version + `.pickle`. When using a shared `cache_dir`, only the basename is kept — so all Mojo environments collide on the same file. The `_newer()` timestamp check then serves the stale pickle.

## Fix

Include a SHA-256 hash of the grammar file's content (truncated to 12 hex chars) in the pickle filename. Different grammar versions now produce different cache files and never collide.

This also removes the earlier MOTO-1264 workaround (force-regeneration when `cache_dir` is `None`), since content-based keying is a proper superset fix.

## Changes

- `Mojo/tools/mblack/src/mblib2to3/pgen2/driver.py` — added `content_hash` parameter to `_generate_pickle_name()` and compute it in  `load_packaged_grammar()`
- `Mojo/tools/mblack/tests/test_grammar_cache.py` — new test file with 7 tests covering cache key generation and cross-environment isolation

### DIFF
--- a/Mojo/tools/mblack/src/mblib2to3/pgen2/driver.py
+++ b/Mojo/tools/mblack/src/mblib2to3/pgen2/driver.py
@@ -38,6 +38,7 @@ __author__ = "Guido van Rossum <guido@python.org>"
 __all__ = ["Driver", "load_grammar"]
 
 # Python imports
+import hashlib
 import io
 import logging
 import os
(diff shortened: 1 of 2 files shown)


npanchen (approved):
thanks for the fix. I'd prefer yours as mine is on a top on another stack

